PR #1185, opened 2020-07-01
Store whether we had a timeout on a file in the parsing cache

### LINKED ISSUE #1156: semgrep 0.12.0 is extremely slow compared to 0.10.1 when matching multiple rules in multiple files
**Performance issue**
For some reasons, probably due to individual matching of files in 0.12.0, it appears that semgrep is much slower than 0.10.1

```
$ semgrep --version
0.10.1
```

`$ time semgrep -f ~/Code/njsscan/njsscan/rules/semantic_grep/ ~/Downloads/dvna-master/`

```
real	0m21.668s
user	0m9.579s
sys	0m2.495s
```

```
$ semgrep --version
0.12.0
```
```
$ time semgrep -f ~/Code/njsscan/njsscan/rules/semantic_grep/ ~/Downloads/dvna-master/
running 73 rules...
```
```
real	6m33.271s
user	6m10.971s
sys	0m9.231s
```

Source Repo: https://github.com/appsecco/dvna
Rules: https://github.com/ajinabraham/njsscan/tree/master/njsscan/rules/semantic_grep

aryx:
This should help fix #1156
See the comment in the diff for more information.

Test plan:

develop/home/pad/semgrep/tests/PERF $ rm -rf /tmp/xxx/
develop/home/pad/semgrep/tests/PERF $ time yy -use_parsing_cache /tmp/xxx -lang js -e '$X == $X' timeout.js
+ /home/pad/github/semgrep/semgrep-core/_build/default/bin/Main.exe -use_parsing_cache /tmp/xxx -lang js -e $X == $X timeout.js
timeout.js:1:0: Timeout

Fatal error: exception Common.Timeout
...
5.528 secs
develop/home/pad/semgrep/tests/PERF $ time yy -use_parsing_cache /tmp/xxx -lang js -e '$X == $X' timeout.js
+ /home/pad/github/semgrep/semgrep-core/_build/default/bin/Main.exe -use_parsing_cache /tmp/xxx -lang js -e $X == $X timeout.js
timeout.js:1:0: Fatal Error: Common.Timeout
Fatal error: exception Common.Timeout
...
0.498 secs

### DIFF
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -384,13 +384,13 @@ let parse_generic lang file =
   (*e: [[Main_semgrep_core.parse_generic()]] resolve names in the AST *)
   Left ast
  (* This is a bit subtle, but we now store in the cache whether we had
-  * a timeout on this file. Indeed, semgrep now calls semgrep-core
-  * per rule, and if one file timeout during parsing, it would timeout
-  * for each rule, but we don't want to wait each time 5sec for each rule.
-  * So here we store the exn in the cache, and below we reraise it
+  * an exception on this file, especially Timeout. Indeed, semgrep now calls
+  * semgrep-core per rule, and if one file timeout during parsing, it would
+  * timeout for each rule, but we don't want to wait each time 5sec for each
+  * rule. So here we store the exn in the cache, and below we reraise it
   * after we got it back from the cache.
   *)
- with Timeout -> Right Timeout
+ with exn -> Right exn
   ))
   in
   match v with
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -359,6 +359,7 @@ let parse_generic lang file =
   then Parse_cpp.init_defs !Flag_parsing_cpp.macros_h;
   (*e: [[Main_semgrep_core.parse_generic()]] use standard macros if parsing C *)
 
+  let v =
   cache_computation file (fun file ->
     (* we may use different parsers for the same file (e.g., in Python3 or
      * Python2 mode), so put the lang as part of the cache "dependency".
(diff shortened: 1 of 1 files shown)


aryx:
Still quite slow, 0.5s it's still pretty long, not sure why, but it's still 10x better than what was before.

semgrep-core/bin/Main.ml
```diff
@@ -380,8 +382,20 @@ let parse_generic lang file =
...
-  ast
+  Left ast
+ (* This is a bit subtle, but we now store in the cache whether we had
+  * a timeout on this file. Indeed, semgrep now calls semgrep-core
+  * per rule, and if one file timeout during parsing, it would timeout
+  * for each rule, but we don't want to wait each time 5sec for each rule.
+  * So here we store the exn in the cache, and below we reraise it
+  * after we got it back from the cache.
+  *)
+ with Timeout -> Right Timeout
```
brendongo, line 397: Should we store just timeouts or any parse error as well?

semgrep-core/bin/Main.ml
```diff
@@ -380,8 +382,20 @@ let parse_generic lang file =
   Naming_AST.resolve lang ast;
   Constant_propagation.propagate lang ast;
   (*e: [[Main_semgrep_core.parse_generic()]] resolve names in the AST *)
-  ast
+  Left ast
```
brendongo, line 385: What does `Left` and `Right` do?

semgrep-core/bin/Main.ml
aryx: yep, maybe better cache any exn.
I think only when we got a timeout things are really annoying performance wise because we
will do n times 5s, but it's true that maybe sometimes parsing takes only 4s but end up in a parse error.
I'll change the diff.

brendongo (approved):

semgrep-core/bin/Main.ml
aryx: Left and Right are constructors defined in pfff/.../Common.ml
They are a pretty standard idiom in OCaml and haskell to represent an alternative between 2 types.
It's defined as  type ('a, 'b) either = Left of 'a | Right of 'b

PR #1186, opened 2020-07-01
Add a new -maxout to limit memory consumption

aryx:
In the past semgrep-core provoked the reboot of the machine of a few users
because it was using lots of memory (probably because of >||> and the
explosion of paths with '...').
This diff adds the possibility to limit the amount of memory
semgrep-core should use. It then returns an exception instead of crashing
before exhausing all the swap of the user's machine.

Test plan:
uncomment the line in the diff about Gc.full_major and then:

time yy -verbose -lang js -maxout 10_000_000 -rules_file ../../data/basic.yml   l10000.js
+ /home/pad/github/semgrep/semgrep-core/_build/default/bin/Main.exe -verbose -lang js -maxout 10_000_000 -rules_file ../../data/basic.yml l10000.js
maxout allocated memory: 72597504
{
  "matches": [],
  "errors": [
    {
      "check_id": "FatalError",
      "path": "l10000.js",
      "start": { "line": 1, "col": 1 },
      "end": { "line": 1, "col": 1 },
      "extra": { "message": "Fatal Error: Out of memory", "line": "/**" }
    }
  ],
  "stats": { "okfiles": 0, "errorfiles": 1 }
}
0.669 secs

### DIFF
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -139,6 +139,7 @@ let ncores = ref 1
 let use_parsing_cache = ref ""
 let target_file = ref ""
 let timeout = ref 0
+let maxout = ref 4_000_000_000 (* 4 GB *)
 
 (*s: constant [[Main_semgrep_core.action]] *)
 (* action mode *)
@@ -347,6 +348,24 @@ let timeout_function lang = fun f ->
   then f ()
   else Common.timeout_function ~verbose:!verbose timeout f
 
+(* from https://discuss.ocaml.org/t/todays-trick-memory-limits-with-gc-alarms/4431 *)
+let run_with_memory_limit limit f =
+  let limit_memory () =
+    let mem = (Gc.quick_stat ()).Gc.heap_words in
+    if mem > limit / (Sys.word_size / 8)
+    then begin
+        if !verbose
+        then pr2 (spf "maxout allocated memory: %d"
+                      (mem * (Sys.word_size / 8)));
+        raise Out_of_memory
+      end
+  in
+  let alarm = Gc.create_alarm limit_memory in
+  Fun.protect f ~finally:(fun () ->
+      Gc.delete_alarm alarm;
+      Gc.compact ()
+  )
+
 (*****************************************************************************)
 (* Parsing *)
 (*****************************************************************************)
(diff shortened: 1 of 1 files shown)


brendongo (approved):

semgrep-core/bin/Main.ml
```diff
@@ -347,6 +348,24 @@ let timeout_function lang = fun f ->
   then f ()
   else Common.timeout_function ~verbose:!verbose timeout f
 
+(* from https://discuss.ocaml.org/t/todays-trick-memory-limits-with-gc-alarms/4431 *)
+let run_with_memory_limit limit f =
```
brendongo, line 352: Does this mean semgrep-core will use 4GB split among all the jobs or each one can use 4GB?

ievans:
can we rename this flag to be more descriptive and also include the implied unit?

semgrep-core/bin/Main.ml
```diff
@@ -530,7 +549,16 @@ let iter_generic_ast_of_files_and_get_matches_and_exn_to_errors f files =
...
+         run_with_memory_limit !maxout (fun () ->
+           (f file lang ast, [])
+           (* to test -maxout, you can uncomment the code below, to give
+            * a chance to the Gc.create_alarm to run even if the program does
+            * noi even need to run the Gc. However, this has a slow perf
+            * penality on small programs, which is why it's better to keep
+            * it in comment when you're not testing -maxout.
+            *)
+           (*|> (fun v -> Gc.full_major(); v) *)
+         )
```
mjambon, line 561: It would be nice to have a permanent test for this. You could add a `--test` or `--debug` flag that would enable the `Gc.full_major ()` (or an environment variable).

semgrep-core/bin/Main.ml
```diff
@@ -347,6 +348,24 @@ let timeout_function lang = fun f ->
...
+        then pr2 (spf "maxout allocated memory: %d"
+                      (mem * (Sys.word_size / 8)));
+        raise Out_of_memory
+      end
+  in
+  let alarm = Gc.create_alarm limit_memory in
+  Fun.protect f ~finally:(fun () ->
+      Gc.delete_alarm alarm;
+      Gc.compact ()
+  )
```
mjambon, line 367: I didn't know about this way.

mjambon (approved):

brendongo:
Probably worth looking into: it looks like this diff makes benchmark much slower

Issue #1170, opened 2020-06-29
Log exception/crash information to a file to aid in debugging


brendongo:
Saving a log of the last run somewhere would be useful when debugging user crashes. We can also use this as an opportunity to bring some sanity into the existing logging (or in our case print statements and global variables).

Useful links: http://nathanielobrown.com/blog/posts/quick_and_dirty_python_logging_lesson.html, https://docs.python.org/3/howto/logging-cookbook.html

PR #1188, opened 2020-07-02
Use logging instead of print to output to stderr

brendongo:
logging framework allows us to add debug logs to a file
and have more sanity around verbosity.

Does not yet create a file handler for the logs. Just reaches feature parity for now.

### DIFF
--- a/semgrep/semgrep/__main__.py
+++ b/semgrep/semgrep/__main__.py
@@ -1,9 +1,17 @@
 #!/usr/bin/env python3
+import logging.handlers
 import sys
 
 from semgrep.cli import cli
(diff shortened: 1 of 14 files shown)


brendongo:
Related to https://github.com/returntocorp/semgrep/issues/1170

brendongo:
TODO: cli flags should set logging levels

aryx (approved):
looks good. Not a Python expert but this looks ok to me.

aryx:
Now we need the log file ... I used to be able to see the semgrep-core command semgrep was using, but now -v does not display it anymore.

Issue #438, opened 2020-04-10
Move default Docker analysis directory


mschwager:
This directory is currently:

```python
REPO_HOME_DOCKER = "/home/repo/"
```

To shorten Docker volume command line arguments let's move this to something like `/code`.

`docker run -v "${PWD}:/home/repo"` is now coming up a lot - `docker run --rm -v "${PWD}:/code"` is a bit shorter and easier to remember. Also, `sgrep` doesn't require the directory to be a code repository, as far as I can tell. So `repo` is a bit of a misnomer.

We'll probably have to support both locations for a period of time as we deprecate the old one.

DrewDennison:
I like this. I think it used to support both `/home/repo` and then I added `/src` which is what pre-commit requires for it's docker image tools. My vote is for `/src`

mschwager:
Fixed in #1192.

PR #1192, opened 2020-07-02
Fix #1191, #438, use '/src' as Docker code volume

### LINKED ISSUE #1191: Fix version check caching in Docker image
We cache version checks for a day to speed up subsequent `semgrep` runs. This functionality is currently broken in our Docker image due to volume weirdness. We set the version cache file as such:

```
ENV SEMGREP_VERSION_CACHE_PATH=/tmp/.cache/semgrep_version
```

The `.cache` directory must exist in a non-root, writable directory. The ensures the Docker container continues to work if someone runs the container as a non-root user with `docker run -u ...` (`pre-commit` does this). The problem is that the `/tmp` directory does not persist across container runs, which nullifies the cache. This means we call out with a version check every run, which is a potentially slow operation.

One option is to create the `.cache` in the `/home/repo` directory, which is used to inject the user's code via a volume into the container. This will persist the cache file across container runs. One issue with this approach is that we cannot always rely on `/home/repo` being available. E.g. when using `pre-commit` we must use `/src`, or a user could, in theory, use whatever they want. We can remedy this by detecting if `--precommit` is set and updating the cache path accordingly - this is a reasonable solution.

Although I think it'd be easiest to simplify things and always use the `/src` directory as the code mount point. That way we're not maintaining multiple directory locations based on how `semgrep` is run. This has been brought up before: https://github.com/returntocorp/semgrep/issues/438. We'll also be able to remove the `--precommit` flag which is one less thing for user's to worry about. The downside here is this is a backwards incompatible change, so user's will have to update how they run the Docker image. We could alleviate this somewhat by detecting if we're running against `/home/repo` and telling them to use `/src` instead.

I purpose tackling the above issue and #438 and simplifying things. Thoughts?

mschwager:
Trying the old directory:

```
$ docker run --rm -v "${PWD}:/home/repo" returntocorp/semgrep:develop --config semgrep-local.yaml semgrep/setup.py
Detected Docker environment using old code volume, please use '/src' instead of '/home/repo'
```

Speedup:

```
$ time docker run --rm -v "${PWD}:/src" returntocorp/semgrep:develop --config semgrep-local.yaml semgrep/setup.py
running 1 rules...

real	0m2.239s
user	0m0.041s
sys	0m0.049s
```

```
$ time docker run --rm -v "${PWD}:/src" returntocorp/semgrep:develop --config semgrep-local.yaml semgrep/setup.py
running 1 rules...

real	0m1.456s
user	0m0.050s
sys	0m0.040s
```

### DIFF
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -26,8 +26,9 @@ from semgrep.util import print_stderr
 
 IN_DOCKER = "SEMGREP_IN_DOCKER" in os.environ
 IN_GH_ACTION = "GITHUB_WORKSPACE" in os.environ
-REPO_HOME_DOCKER = "/home/repo/"
-REPO_HOME_DOCKER_PRECOMMIT = "/src/"
+
+SRC_DIRECTORY = Path(os.environ.get("SEMGREP_SRC_DIRECTORY", Path("/") / "src"))
+OLD_SRC_DIRECTORY = Path("/") / "home" / "repo"
 
 TEMPLATE_YAML_URL = (
     "https://raw.githubusercontent.com/returntocorp/semgrep-rules/develop/template.yaml"
@@ -72,17 +73,19 @@ def resolve_targets(targets: List[str]) -> List[Path]:
     ]
 
 
-def adjust_for_docker(in_precommit: bool = False) -> None:
-    repo_home = REPO_HOME_DOCKER_PRECOMMIT if in_precommit else REPO_HOME_DOCKER
-
+def adjust_for_docker() -> None:
     # change into this folder so that all paths are relative to it
     if IN_DOCKER and not IN_GH_ACTION:
-        if not Path(repo_home).exists():
+        if OLD_SRC_DIRECTORY.exists():
             raise SemgrepError(
-                f'you are running semgrep in docker, but you forgot to mount the current directory in Docker: missing: -v "${{PWD}}:{repo_home}"'
+                f"Detected Docker environment using old code volume, please use '{SRC_DIRECTORY}' instead of '{OLD_SRC_DIRECTORY}'"
             )
-    if Path(repo_home).exists():
-        os.chdir(repo_home)
+        if not SRC_DIRECTORY.exists():
+            raise SemgrepError(
+                f"Detected Docker environment without a code volume, please include '-v \"${{PWD}}:{SRC_DIRECTORY}\"'"
+            )
+    if SRC_DIRECTORY.exists():
+        os.chdir(SRC_DIRECTORY)
 
 
 def get_base_path() -> Path:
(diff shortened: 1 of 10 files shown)


mschwager:
The `pre-commit` GHA will depend on the latest changes to the `:develop` Docker image, so the failure is expected.

underyx:
@mschwager could you explain what caused the speedup you mention?

mschwager:
> @mschwager could you explain what caused the speedup you mention?

The version check caching now works! This didn't work previously because we didn't have a consistent directory we could reliably write files to and persist them across container runs. This is now always the `/src` directory (or configurable by the `SEMGREP_VERSION_CACHE_PATH` env variable) and we achieve persistence via same Docker volume used to analyze the code. This means a `.cache/semgrep_version` file will now show up in your code repository (instead of the home directory) when running the Docker container.

brendongo:
Can we have commit titles not refer to issue numbers? Probably better to have them in the body if needed

mschwager:
> Can we have commit titles not refer to issue numbers? Probably better to have them in the body if needed

I'd like to - the problem is that the PR doesn't auto-close the related issues if it's in the body, only if it's in the title. A small annoyance, but I blame Github :)

EDIT: Huh looks like you can include it in the body: https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword - maybe I'm operating on old information :man_shrugging:. I can remove it from the title in the future :+1:

brendongo (approved):

semgrep/semgrep/config_resolver.py
```diff
@@ -72,17 +73,19 @@ def resolve_targets(targets: List[str]) -> List[Path]:
...
 
-def adjust_for_docker(in_precommit: bool = False) -> None:
-    repo_home = REPO_HOME_DOCKER_PRECOMMIT if in_precommit else REPO_HOME_DOCKER
-
+def adjust_for_docker() -> None:
     # change into this folder so that all paths are relative to it
     if IN_DOCKER and not IN_GH_ACTION:
-        if not Path(repo_home).exists():
+        if OLD_SRC_DIRECTORY.exists():
+            raise SemgrepError(
```
DrewDennison, line 80: we should just symlink for now and warn?

semgrep/semgrep/version.py
```diff
@@ -21,8 +21,7 @@
 )
 VERSION_CACHE_PATH = Path(
     os.environ.get(
-        "SEMGREP_VERSION_CACHE_PATH",
-        Path(os.path.expanduser("~")) / ".cache" / "semgrep_version",
+        "SEMGREP_VERSION_CACHE_PATH", Path.home() / ".cache" / "semgrep_version",
```
DrewDennison, line 24: hmm I would be surprised if a tool dropped files in my home folder. I would expect `/tmp`

PR #1199, opened 2020-07-06
autofix: Handle case when no fix is requested

underyx:
I'm not sure in what case this can happen, but I saw this error in semgrep.live logs:

    UnboundLocalError: local variable 'fixobj' referenced before assignment

I think the only way that error can happen is if a rule_match without a fix is sent in to this function.

But maybe instead of this commit, we should track down why the function was called in the first place if no fix was requested.

### DIFF
--- a/semgrep/semgrep/autofix.py
+++ b/semgrep/semgrep/autofix.py
@@ -121,6 +121,8 @@ def apply_fixes(
                     raise SemgrepError(
                         f"unable to use regex to modify file {filepath} with fix '{fix}': {e}"
                     )
+            else:
+                continue
             # endif
             if not dryrun:
                 _write_contents(rule_match.path, fixobj.fixed_contents)


minusworld:
> But maybe instead of this commit, we should track down why the function was called in the first place if no fix was requested.

I think I might've added --autofix and --dryrun to every invocation of Semgrep on the backend assuming it would be fine. 😬  Good catch, and thanks for cleaning up my mess 🙇

underyx:
> I think I might've added --autofix and --dryrun to every invocation of Semgrep on the backend assuming it would be fine. 😬 Good catch, and thanks for cleaning up my mess 🙇

No worries! This code should be more correct anyway for regular CLI users, too. Either `--autofix` should be ignored for rules that have no fixes, or just like this PR does, we should make `autofix.py` expect being called without a fix.

minusworld (approved):

PR #1202, opened 2020-07-07
Added note to install pkg-config in development instructions

chmccreery:
`opam install --deps-only -y` fails without it


nbrahms (approved):
👍

brendongo (approved):

PR #1205, opened 2020-07-07
pfff: bring in python2 parse fixes

brendongo:
- Parse python files with trailing whitespace
- Parse python2 tuple as parameter in function/lambda definition
- Parse python3.8 positional only parameters (PEP 570)
- Parse python2 implicit array in comprehension


nbrahms (approved):
lgtm

PR #1206, opened 2020-07-07
Adding cases and functionality

emjin:
Tested with make test


aryx (approved):
because of the merge/rebase there are lots of irrelevant files but the code in
the relevant commit looks good!
Let's get this in so we can iterate and fix the remaining unit tests

PR #1207, opened 2020-07-07
Handle file-not-found errors more gracefully

### LINKED ISSUE #706: Gracefully handle missing files
**Describe the bug**
If you run semgrep on a file that doesn't exist, it panics:
```
➜  Downloads semgrep script.py -l python -e '$X == $X'
an internal error occured while invoking the semgrep engine: unknown exception: Unix.Unix_error(Unix.ENOENT, "stat", "script.py").

An error occurred while invoking the semgrep engine; please help us fix this by filing an an issue at https://semgrep.dev
```

**To Reproduce**
Run semgrep on a file that doesn't exist.

**Expected behavior**
An error about a file not existing.

**Screenshots**

**Environment**
pip package v0.6.0

chmccreery:
Re-factored target manager init method to use attr.s for clarity and passed output_handler to target_manager. Fixes #706

### DIFF
--- a/semgrep/semgrep/target_manager.py
+++ b/semgrep/semgrep/target_manager.py
@@ -5,11 +5,16 @@ from typing import List
 from typing import NewType
 from typing import Set
 
+import attr
+
 from semgrep.error import _UnknownLanguageError
+from semgrep.error import FilesNotFoundError
+from semgrep.error import SemgrepError
+from semgrep.output import OutputHandler
 from semgrep.semgrep_types import Language
 from semgrep.util import partition_set
+from semgrep.util import print_stderr
 from semgrep.util import sub_check_output
-
 
 FileExtension = NewType("FileExtension", str)
 
@@ -62,26 +67,22 @@ def lang_to_exts(language: Language) -> List[FileExtension]:
         raise _UnknownLanguageError(f"Unsupported Language: {language}")
 
 
+@attr.s(auto_attribs=True)
 class TargetManager:
-    def __init__(
-        self,
-        includes: List[str],
-        excludes: List[str],
-        targets: List[str],
-        respect_git_ignore: bool,
-    ) -> None:
-        """
-            Handles all file include/exclude logic for semgrep
+    """
+        Handles all file include/exclude logic for semgrep
 
-            If respect_git_ignore is true then will only consider files that are
-            tracked or (untracked but not ignored) by git
-        """
-        self._targets = targets
-        self._includes = includes
-        self._excludes = excludes
-        self._respect_git_ignore = respect_git_ignore
+        If respect_git_ignore is true then will only consider files that are
+        tracked or (untracked but not ignored) by git
+    """
 
-        self._filtered_targets: Dict[str, Set[Path]] = {}
+    targets: List[str]
+    includes: List[str]
+    excludes: List[str]
+    respect_git_ignore: bool
+    output_handler: OutputHandler
+
+    _filtered_targets: Dict[str, Set[Path]] = {}
 
     @staticmethod
     def resolve_targets(targets: List[str]) -> Set[Path]:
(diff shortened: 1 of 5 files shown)


semgrep/semgrep/target_manager.py
```diff
@@ -62,26 +67,22 @@ def lang_to_exts(language: Language) -> List[FileExtension]:
...
-        self._respect_git_ignore = respect_git_ignore
+        If respect_git_ignore is true then will only consider files that are
+        tracked or (untracked but not ignored) by git
+    """
+
+    targets: List[str]
+    includes: List[str]
+    excludes: List[str]
+    respect_git_ignore: bool
+    output_handler: OutputHandler
```
brendongo, line 83: TargetManager should not need to know how to handle output. Probably better to have functions return relevant info and the caller can handle how it wants to deal with outputing said info.

semgrep/semgrep/target_manager.py
```diff
@@ -237,23 +238,31 @@ def filtered_files(self, lang: Language) -> Set[Path]:
...
             return self._filtered_targets[lang]
 
-        targets = self.resolve_targets(self._targets)
-        explicit_files, directories = partition_set(lambda p: not p.is_dir(), targets)
+        targets = self.resolve_targets(self.targets)
+
+        files, directories = partition_set(lambda p: not p.is_dir(), targets)
+
+        # Error on non-existent files
+        explicit_files, nonexistent_files = partition_set(lambda p: p.is_file(), files)
```
brendongo, line 246: You might need to handle symlinked files as well?

semgrep/semgrep/target_manager.py
chmccreery: Can you elaborate on this a bit more? I understand the preference for not making output_handler a required argument, but don't really understand the alternative you are suggesting (how would you change the FilesNotFoundError?)

semgrep/semgrep/target_manager.py
chmccreery: It looks like is_file() should cover regular files and symlinks (and I just tested with symlinks and got the behavior I expected- but maybe that is not the desired behavior): https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file 
Or do you mean that symlinked files should be handled differently instead of lumped in with regular files?

semgrep/semgrep/target_manager.py
nbrahms: @brendongo IMO we should really have a context object do this (a la Bento) or a reader monad or the like. But passing output_handler here is exactly how we separate knowing how to handle output from the TargetManager.

brendongo: I mean: TargetManager should concern itself with file includes excludes and not need to know how to use an output_handler object. Resolved targets can return a tuple with the second item being problematic files.

semgrep/semgrep/target_manager.py
brendongo: Oh hmm so it does. Not sure why tests are failing then

semgrep/semgrep/target_manager.py
brendongo: TargetManager should not need to know that any errors are "handled" or that they are even errors to begin with. The TargetManager just knows that of the files you handed it, these happened to not exist and it lets you know that. It's up to the caller to deal with this information.

semgrep/semgrep/target_manager.py
chmccreery: At least one of them is because we added the parameter to TargetManager but the test file still thinks it only takes 4 arguments... I'm working on it

semgrep/semgrep/target_manager.py
brendongo: Thinking about this more and the list of files that don't exist is unrelated to the language so might be better placed in the constructor instead of in this code path. Seems like a significant endeavor would be supportive of getting this PR in as is with injection implementation and revisiting that and constructor in a follow up.

brendongo: you might want to think about the edge cases of a rule with multiple languages calling get_files multiple times though

chmccreery: Good thought! We just tested this case and it seems to print out the right thing (only once)

brendongo (approved):

PR #1233, opened 2020-07-10
Ruby and semgrep, v0

aryx:


mjambon (approved):
looks good to me :+1:

semgrep-core/parsing/Parse_ruby_tree_sitter.ml
```diff
@@ -27,14 +27,66 @@ module G = AST_generic
...
 
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
-let token2 (_tok : Tree_sitter_run.Token.t) =
-  PI.fake_info "XXX"
-let str (_tok : Tree_sitter_run.Token.t) =
-  "STRXXX", PI.fake_info "XXX"
+
+(* mostly a copy of Parse_info.full_charpos_to_pos_large *)
```
nbrahms, line 41: Might be worth noting that most of this can go in a language-agnostic library in the future.

semgrep-core/tests/python/concrete_syntax.py
```diff
@@ -1,11 +1,13 @@
 def foo():
```
nbrahms, line 1: Why's this file changing?

nbrahms (approved):

semgrep-core/parsing/Parse_ruby_tree_sitter.ml
aryx: yep, totally, I'll move it to a tree_sitter_utils.ml when I do Parse_java_tree_sitter.ml (or the typescript one).

semgrep-core/tests/python/concrete_syntax.py
aryx: It's because to start the examples from ruby I copy-pasted a few one from python (which themselves were copy-pasted
from examples for javascripts) and realized some python examples were not as powerful as they could be (the javascript examples were better for instance).

PR #1234, opened 2020-07-10
Rev pfff

nbrahms:
Updates pfff to include a python2 parse fix.


nbrahms:
Abandoning this as it's unnecessary for its stated purpose

PR #1236, opened 2020-07-10
Update dev docs, create semgrep-core/bin symlink

mjambon:
I also updated the `bindings.c` files to suppress a compilation warning.


DrewDennison (approved):

semgrep-core/Makefile
```diff
@@ -4,8 +4,10 @@ all:
 
 	dune build
 	dune build ./_build/default/tests/test.bc
+	test -e bin || ln -s _build/install/default/bin .
```
aryx, line 7: What is that? semgrep-core/bin/ actually contains the source files of the main of the binaries (ok bin might not be
the best name).
We need at least to rename bin/Main.ml (the main entry point of semgrep-core) to something else, maybe main/Main.ml?

PR #1239, opened 2020-07-10
Mypy: configure for tests

### LINKED ISSUE #699: Enable mypy on semgrep tests
It took lots of effort to get type hinting working on tests so we just disabled it. See http://calpaterson.com/mypy-hints.html which suggests this too. Here's a discussion anyway to try re-adding type hints down the line.

nbrahms:
We want mypy to help us make sure we're calling methods correctly in
test code, to help us avoid unnecessary CI cycles. However, we don't
need to be strict with making sure we're defining types.

Here I add a second mypy config just for test code, with this less
restrictive configuration.

Closes #699 .


.pre-commit-config.yaml
```diff
@@ -30,9 +30,18 @@ repos:
     rev: "v0.761"
     hooks:
       - id: mypy
-        exclude: ^semgrep/tests/.+$
+        exclude: ^semgrep\/tests\/.+$
```
underyx, line 33: whaaat how did this work before? 🤯

nbrahms: my guess is that the pre-commit regex dialect is ok with forward slashes being unescaped?

underyx (approved):

PR #1242, opened 2020-07-10
Clara/taint analysis

### LINKED ISSUE #1212: Incorporate taint tracking in python wrapper
Semgrep-core already has support for rudimentary taint analysis. We should wire it up to the python CLI to make it more accessible to users!

The current core code takes a yaml file specifying sources, sinks, and sanitizers. This seems like a lot of arguments to include directly on the commandline (like what is done with the -e flag), so it may make most sense to pass in the yaml file using some variant of the --config flag.

A few options:
1. We could use the same --config flag to specify the file but then use a second flag like --taint to specify how to interpret that file 
2. We could use a completely new flag like --taint-config to specify the yaml
3. We could use the same --config flag to specify the file and include a line in the yaml file itself indicating how to interpret the file
4. Any other ideas??? 

Welcoming feedback/suggestions/opinions!

chmccreery:
Could use more re-factoring but I need some advice/consultation on how to test the current code (ex. all the regex behaviors and equivalences) to make sure I am not breaking anything


semgrep-core/data/basic_tainting.yml
```diff
@@ -1,5 +1,6 @@
 rules:
   - id: classic
+    mode: track
```
nbrahms: FWIW, I'd probably `s/track/taint` here for specificity

semgrep-core/data/basic_tainting.yml
```diff
@@ -11,5 +12,18 @@ rules:
       - sanitize(...)
       - sanitize1(...)
     message: A user input source() went into a dangerous sink()
+    languages: [python, python]
```
nbrahms:
```suggestion
    languages: [python]
```

semgrep/semgrep/core_runner.py
```diff
@@ -217,6 +217,72 @@ def _write_equivalences_file(self, fp: IO, equivalences: List[Equivalence]) -> N
...
         fp.flush()
 
+    def _run_core_command(
+        self,
+        patterns_json: List[Any],
+        patterns: List[Pattern],
+        targets: List[Path],
+        language: Language,
+        rule: Rule,
+        rules_flag: str,
```
nbrahms: I'd probably make this an enum instead of a `str` to avoid future potential bugs.

semgrep/semgrep/core_runner.py
```diff
@@ -225,111 +291,93 @@ def _run_rule(
...
-                ]
-
-                if equivalences:
-                    self._write_equivalences_file(equiv_file, equivalences)
-                    cmd += ["-equivalences", equiv_file.name]
+                )
 
-                core_run = sub_run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        else:
+            for language, all_patterns_for_language in self._group_patterns_by_language(
```
nbrahms: This function is probably complex enough that we want to subfunction it.

I can show you a fun PyCharm-fu to automate this.

semgrep/semgrep/evaluation.py
```diff
@@ -207,16 +207,21 @@ def evaluate(
         }
     ]
     debug_print(str(pattern_ids_to_pattern_matches))
-    valid_ranges_to_output = evaluate_expression(
-        rule.expression,
-        pattern_ids_to_pattern_matches,
-        flags={RCE_RULE_FLAG: allow_exec},
-        steps_for_debugging=steps_for_debugging,
-    )
+    if rule._mode == "search":
```
nbrahms: We probably want to be consistent with the `_mode` comparisons, so we have a consistent fallback mode.

semgrep/semgrep/semgrep_main.py
```diff
@@ -61,8 +61,18 @@ def validate_single_rule(
             + [e.span for e in extra_key_spans],
             help=help_msg,
         )
-
-    # Raises InvalidRuleSchemaError if fails to parse
+    else:
```
nbrahms: I'm guessing we can probably simplify this logic with the above. Happy to hack through it with you.

semgrep/semgrep/evaluation.py
chmccreery: good point- just fixed this

semgrep-core/data/basic_tainting.yml
chmccreery: what does s/track/taint mean? something like track-mode: True where track-mode: False is the default?

chmccreery: Oohh just swap the words 🤦‍♀️  yes will do

semgrep/semgrep/core_runner.py
chmccreery: Like this? https://docs.python.org/3.4/library/enum.html  
So it would be of type Mode and would evaluate to search = 1 and taint = 2 or something like that?

semgrep/semgrep/core_runner.py
```diff
@@ -150,11 +150,16 @@ def _flatten_rule_patterns(self, rules: List[Rule]) -> Iterator[Pattern]:
     def _group_patterns_by_language(
         self, rules: List[Rule]
     ) -> Dict[Language, List[Pattern]]:
-        # a rule can have multiple patterns inside it. Flatten these so we can send semgrep a single yml file list of patterns
-        patterns: List[Pattern] = list(self._flatten_rule_patterns(rules))
+
         by_lang: Dict[Language, List[Pattern]] = collections.defaultdict(list)
-        for pattern in patterns:
-            by_lang[pattern.language].append(pattern)
+        if rules[0]._mode == "taint":
```
DrewDennison: `[0]` always makes me nervous because what if `rules` is empty or there are mixed rules and you only consider the first case. I would generally tend to use map, filter, partition logic here.

semgrep/semgrep/core_runner.py
```diff
@@ -225,111 +296,58 @@ def _run_rule(
...
-            # semgrep-core's results for other patterns.
-            patterns_regex, patterns = partition(
-                lambda p: p.expression.operator == OPERATORS.REGEX,
-                all_patterns_for_language,
-            )
-            if patterns_regex:
-                patterns_json = [pattern.to_json() for pattern in patterns_regex]
+            if rule._mode == "taint":
+                pattern_json = rule._raw.copy()
+                del pattern_json["mode"]
```
DrewDennison, line 309: do we also support `metadata`?

semgrep/semgrep/evaluation.py
```diff
@@ -207,16 +207,21 @@ def evaluate(
         }
     ]
     debug_print(str(pattern_ids_to_pattern_matches))
-    valid_ranges_to_output = evaluate_expression(
-        rule.expression,
-        pattern_ids_to_pattern_matches,
-        flags={RCE_RULE_FLAG: allow_exec},
-        steps_for_debugging=steps_for_debugging,
-    )
+    if rule._mode == "track":
```
DrewDennison: make a constant?

semgrep/semgrep/rule.py
```diff
@@ -65,7 +67,33 @@ def __init__(self, raw: YamlTree[YamlMap]) -> None:
...
+        # check taint/search mode
+        self._expression, mode = self._taint_or_search_patterns_validation(self._yaml)
+        self._mode = mode
+
+    def _taint_or_search_patterns_validation(
+        self, rule: YamlTree[YamlMap]
+    ) -> Tuple[BooleanRuleExpression, str]:
+
+        rule_raw = rule.value
+        _mode = rule_raw.get("mode")
```
DrewDennison: make `mode` a constant?

.pre-commit-config.yaml
```diff
@@ -40,32 +40,32 @@ repos:
...
 
-  - repo: https://github.com/returntocorp/semgrep
-    rev: 'v0.12.0'
-    hooks:
-      - id: semgrep-develop
-        name: Semgrep Python
-        types: [python]
-        exclude: "^semgrep/tests/.+$|^install-scripts/.+$|^release-scripts/.+$|^semgrep/setup.py$"
-        args: ['--config', 'https://semgrep.live/p/python', '--error']
+  # - repo: https://github.com/returntocorp/semgrep
```
nbrahms: I assume you don't want to comment changes to this file?

semgrep/semgrep/core_runner.py
nbrahms: I think `enum` is the wrong word for me to use.

I mean this:
```python
Mode = NewType("Mode", str)

SearchMode = Mode("search")
TaintMode = Mode("taint")
```

.pre-commit-config.yaml
chmccreery: Oh yeah, we just changed this so that it would run on my machine, right? Or should I delete the commented-out pieces and commit the change? I guess I didn't understand what the impact of this was/if it was supposed to be pushed up

semgrep/semgrep/core_runner.py
chmccreery: Within search-mode there is still the pattern-regex option, if that's what you're talking about? We don't have metadata as a separate mode.

semgrep/semgrep/evaluation.py
chmccreery: done

semgrep/semgrep/rule.py
chmccreery: done

semgrep/semgrep/core_runner.py
chmccreery: done

semgrep/semgrep/core_runner.py
chmccreery: re-factored it so this isn't even a list (it was always [rule] before, and doesn't need to become a list until later)

semgrep/semgrep/core_runner.py
nbrahms: There's an auxiliary `metadata` field outside of `patterns`, which I assume is what Drew is referring to?

semgrep/semgrep/semgrep_main.py
```diff
@@ -46,23 +46,37 @@ def validate_single_rule(
...
-        extra_keys: Set[str] = rule_keys - YAML_ALL_VALID_RULE_KEYS
-        extra_key_spans = sorted([rule.key_tree(k) for k in extra_keys])
-        help_msg = None
-        if extra_keys:
-            help_msg = f"Unexpected keys {extra_keys} found. Is one of these a typo of {missing_keys}?"
+    extra_keys = rule_keys - YAML_ALL_VALID_RULE_KEYS
+    extra_key_spans = sorted([rule.key_tree(k) for k in extra_keys])
+    missing_keys = YAML_MUST_HAVE_KEYS - rule_keys
+
+    if missing_keys and extra_keys:
```
nbrahms, line 53: 👍

semgrep/semgrep/semgrep_types.py
```diff
@@ -57,14 +64,18 @@ class OPERATORS:
     OPERATORS.EQUIVALENCES,
     OPERATORS.REGEX,
 }
+YAML_TOP_LEVEL_PATTERN_KEYS = {
+    pattern_name
+    for op in YAML_VALID_TOP_LEVEL_OPERATORS
+    for pattern_name in OPERATOR_PATTERN_NAMES_MAP[op]
+}
+YAML_TAINT_KEYS = {"pattern-sinks", "pattern-sources", "pattern-sanitizers"}
```
nbrahms: nit:
```python
YAML_TAINT_KEYS = YAML_TAINT_MUST_HAVE_KEYS | { "pattern-sanitizers" }
```

semgrep/semgrep/rule.py
```diff
@@ -65,7 +72,35 @@ def __init__(self, raw: YamlTree[YamlMap]) -> None:
...
+        self._expression, self._mode = self._taint_or_search_patterns_validation(
+            self._yaml
+        )
+
+    def _taint_or_search_patterns_validation(
+        self, rule: YamlTree[YamlMap]
+    ) -> Tuple[BooleanRuleExpression, Mode]:
+
+        rule_raw = rule.value
+        _mode = rule_raw.get("mode")
```
nbrahms: You can just use `rule_raw.get("mode", DEFAULT_MODE)` here, and remove a case below.

nbrahms (approved):
I left some stylistic suggestions, but the functionality looks 👌

semgrep/semgrep/semgrep_main.py
chmccreery: 😄

semgrep/semgrep/rule.py
chmccreery: It won't let me put more than 1 arg into get() because rule_raw is a YamlMap object, not a dictionary. Do you have another idea? This approach would be cleaner...

semgrep/semgrep/core_runner.py
chmccreery: Ah, great! I just added a check for this to say it isn't supported

semgrep/semgrep/rule.py
chmccreery: It didn't like the "or" approach either, so I went with a ternary, which has a similar effect wrt code structure and complexity

PR #1248, opened 2020-07-13
pre-commit-config housekeeping

underyx:


.pre-commit-config.yaml
```diff
@@ -22,12 +22,13 @@ repos:
       - id: black
 
   - repo: https://github.com/asottile/reorder_python_imports
-    rev: v1.9.0
+    rev: v2.3.0
     hooks:
       - id: reorder-python-imports
+        args: [--py36-plus]
```
mschwager, line 28: What does this get us?

mschwager (approved):

.pre-commit-config.yaml
underyx: From commit message:

```
This way it'll update legacy imports automatically,
such as mock -> unittest.mock
and mypy_extensions.TypedDict -> typing.TypedDict.
```

I can imagine people leaving in accidentally copy-pasted `from __future__` imports for instance. And people at a previous employer of mine were repeatedly confused and installing `mock` when it was already in the standard library.

mschwager: Nice, that kept burning me with `dataclasses`

PR #1249, opened 2020-07-13
Ruby support v1

aryx:


aryx:
This will definitely help https://github.com/returntocorp/enterprise/issues/251

PR #1254, opened 2020-07-13
add ruby to supported languages

DrewDennison:
merge after #1249


DrewDennison:
actually #1249 has all these changes

PR #1272, opened 2020-07-14
Allow pattern to be inferred from expressions and statements (and later, statement lists?)

emjin:
Changed range matcher so that it will first look for a statement, then, if not found, an expression to match. This will allow pattern_from_code to handle a much wider variety of code.

Tested with arrays_and_funcs.py, set_cookie.py, typed_func.java.


aryx:
Should we merge this?

emjin:
> Should we merge this?

I think it's ready now.

aryx (approved):
Looks great!

PR #1276, opened 2020-07-14
Allow untyped function definitions to match typed definitions

### LINKED ISSUE #1270: [Python] Untyped function pattern does not match typed function code
**Describe the bug**
A pattern like:
```python
def $FUNC(...):
    ...
```
does not match a typed function definition in python.

**To Reproduce**
See https://semgrep.live/d8e5/.

**Expected behavior**
Patterns without a declared type should match both typed and untyped function definitions.

**Environment**
semgrep.live

nbrahms:
This enables writing untyped function patterns, and matching typed code
in Python and JavaScript:

```
def $FUNC(...):
  ...
```

will match

```
def foo(x) -> str:
  return str(x)
```

for example.

Fixes #1270


nbrahms:
@aryx In general, does it make sense to replace `m_option m_type_` with `m_option_none_can_match_some m_type_` in all instances?

Places where we do not do this now:
- variable definitions
- unnamed parameters
- names

Places where we already do this:
- named parameters

nbrahms:
> @aryx In general, does it make sense to replace `m_option m_type_` with `m_option_none_can_match_some m_type_` in all instances?
> 
> Places where we do not do this now:
> 
> * variable definitions
> * unnamed parameters
> * names
> 
> Places where we already do this:
> 
> * named parameters

FWIW I'd expect

```
$X = $Y
```

to match

```
x: int = 1
```

aryx (approved):
Yes we can probably use m_option_non_can_match_some in the other instances you mentioned.
That's a good idea.

PR #1282, opened 2020-07-15
Test semgrep-core linking

brendongo:


brendongo:
Included necessary fixes in https://github.com/returntocorp/semgrep/pull/1281

PR #1284, opened 2020-07-15
Skeleton to use the Go tree-sitter parser

aryx:
This is part1 of adding support for the Go tree-sitter-based parser.
This diff just integrate the go tree-sitter library in the build system
and create a boilerplate Parse_go_tree_sitter.ml to be filled later.

Test plan:
make

### DIFF
--- a/semgrep-core/parsing/Parse_go_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_go_tree_sitter.ml
@@ -28,44 +28,44 @@ let blank (env : env) () =
 let todo (env : env) _ =
    failwith "not implemented"
 
-let map_int_literal (env : env) (tok : CST.int_literal) =
+let int_literal (env : env) (tok : CST.int_literal) =
   token env tok (* int_literal *)
 
-let map_identifier (env : env) (tok : CST.identifier) =
+let identifier (env : env) (tok : CST.identifier) =
   token env tok (* identifier *)
 
-let map_raw_string_literal (env : env) (tok : CST.raw_string_literal) =
+let raw_string_literal (env : env) (tok : CST.raw_string_literal) =
   token env tok (* raw_string_literal *)
 
-let map_rune_literal (env : env) (tok : CST.rune_literal) =
+let rune_literal (env : env) (tok : CST.rune_literal) =
   token env tok (* rune_literal *)
 
-let map_float_literal (env : env) (tok : CST.float_literal) =
+let float_literal (env : env) (tok : CST.float_literal) =
   token env tok (* float_literal *)
 
-let map_imaginary_literal (env : env) (tok : CST.imaginary_literal) =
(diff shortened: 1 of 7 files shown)


aryx:
The goal of this PR is also to document the different steps to add support for a new tree-sitter parser.

mjambon (approved):
woohoo

semgrep-core/parsing/Parse_java_tree_sitter.ml
```diff
@@ -50,13 +51,11 @@ module G = AST_generic
 
 [@@@warning "-32"]
 
-type env = unit
+type env = H.env
 
-let token (env : env) (_tok : Tree_sitter_run.Token.t) =
-  failwith "not implemented"
+let token (env : env) (_tok : Tree_sitter_run.Token.t) = H.token
```
nbrahms, line 56: should this be in a Java parsing PR instead?

aryx: yes it should ...

semgrep-core/parsing/Parse_go_tree_sitter.ml
```diff
@@ -0,0 +1,1547 @@
...
+  let v2 = token env v2 (* identifier *) in
+  todo env (v1, v2)
+
+let empty_labeled_statement (env : env) ((v1, v2) : CST.empty_labeled_statement) =
+  let v1 = token env v1 (* identifier *) in
+  let v2 = token env v2 (* ":" *) in
+  todo env (v1, v2)
+
+let interpreted_string_literal (env : env) ((v1, v2, v3) : CST.interpreted_string_literal) =
+  let v1 = token env v1 (* "\"" *) in
```
nbrahms, line 101: FWIW I'd love to see us, in the future, clearly separating boilerplate vs. contributed code.

E.g.
```ocaml
match Boilerplate.interpreted_string_literal env v with
  (v1, v2, v3) -> todo env (v1, v2, v3)
```
here

nbrahms (approved):

semgrep-core/parsing/Parse_go_tree_sitter.ml
nbrahms: The motivation is to reduce, as much as possible, any manual work when rev'ing the grammar later.

aryx: I don't think we can call Boilerplate from Parse_go_tree_sitter.ml. We copy-paste Boilerplate.ml initially but
in my next diff I will update every final "actions" in Parse_go_tree_sitter.ml. 
When we rev'ing the grammar later, we will regenerate CST.ml and the OCaml compiler will help us pinpoint
all the place in Parse_go_tree_sitter.ml we need to update. I don't think it will be too much work.

PR #1286, opened 2020-07-15
Remove benchmark, don't run tests on MD changes

rcoh:
We can do a lot more smart filtering of what tests we run when, but this a conservative change just to test out the machinery.


brendongo (dismissed):

brendongo:
This probably means required tests don't ever run on PRs with just md files changed instead of autopassing those tests. We can see if it's fine to rely on an admin override to merge for now instead of having to modify the action to pass given certain conditions.

brendongo (approved):

PR #1294, opened 2020-07-16
Revert "Remove sdist from pypi release"

underyx:
Reverts returntocorp/semgrep#1263

The original issue was that `pip install --no-build-isolation --no-binary :all: semgrep` would fail with a 'semgrep-core not found' error.

After this commit, running the same command will still fail with the same error as before, because pip will download the last release with an sdist available. So I'd argue this commit was not an improvement.

If you run `pip install --no-build-isolation --no-binary :all: semgrep==0.15.0` with the version pinned, you get `ERROR: Could not find a version that satisfies the requirement semgrep==0.15.0` which I'd say is a worse, more confusing error message than `semgrep-core not found`.

Created https://github.com/returntocorp/semgrep/issues/1295 for a more long term solution.

### DIFF
--- a/release-scripts/build-wheels.sh
+++ b/release-scripts/build-wheels.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 set -e
 pip install setuptools wheel
-cd semgrep && python setup.py bdist_wheel
+cd semgrep && python setup.py sdist bdist_wheel
 # Zipping for a stable name to upload as an artifact
 zip -r dist.zip dist


brendongo (approved):

mschwager (changes requested):
I think we need to come up with a better solution here.

mschwager (approved):
I think #1295 is probably sufficient :+1:

Although I still wonder if we should provide a sdist at all given that you can just as easily download the corresponding repository tarball from a Github release and build things that way. Having sdists available seems like it will just set users up for failure since it's unlikely that they'll have `semgrep-core` available. This could be a documentation issue and/or a `setup.py` improvement.

Perhaps it's best to revamp `setup.py`, assume sdist installation is for internal use-cases and power-users, and go from there. We could even have a required env variable pointing to `semgrep-core` be a per-requisite for installation. Although, to that end, I think we'll need to provide more obvious ways to obtain `semgrep-core` (better docs? easily available as a GHA artifact?).

underyx:
> Having sdists available seems like it will just set users up for failure since it's unlikely that they'll have semgrep-core available.

That's right, but I still think they'll fail worse and have less of an idea what to do if we don't have sdists available.

PR #1302, opened 2020-07-17
output: Fix duplicate rule IDs in SARIF output

underyx:
I'm not sure why rules started having duplicates, but I gave up after 5 minutes of digging. Let's just assume that's expected?

### DIFF
--- a/semgrep/semgrep/output.py
+++ b/semgrep/semgrep/output.py
@@ -156,6 +156,8 @@ def build_sarif_output(rule_matches: List[RuleMatch], rules: FrozenSet[Rule]) ->
     - which links to this schema https://github.com/oasis-tcs/sarif-spec/blob/master/Schemata/sarif-schema-2.1.0.json
     - full spec is at https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html
     """
+    # SARIF doesn't allow duplicate IDs, we dedupe here
+    rules_by_id = {rule.id: rule for rule in rules}
     output_dict = {
         "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
         "version": "2.1.0",
@@ -164,7 +166,7 @@ def build_sarif_output(rule_matches: List[RuleMatch], rules: FrozenSet[Rule]) ->
                 "tool": {
                     "driver": {
                         **_sarif_tool_info(),
-                        "rules": [rule.to_sarif() for rule in rules],
+                        "rules": [rule.to_sarif() for rule in rules_by_id.values()],
                     }
                 },
                 "results": [match.to_sarif() for match in rule_matches],


semgrep/semgrep/output.py
```diff
@@ -156,6 +156,8 @@ def build_sarif_output(rule_matches: List[RuleMatch], rules: FrozenSet[Rule]) ->
     - which links to this schema https://github.com/oasis-tcs/sarif-spec/blob/master/Schemata/sarif-schema-2.1.0.json
     - full spec is at https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html
     """
+    # SARIF doesn't allow duplicate IDs, we dedupe here
+    rules_by_id = {rule.id: rule for rule in rules}
```
mschwager, line 160: I think it'd be worth it to keep digging and see why we're hitting this issue. If there's a bug earlier in the program causing this then we'd like to know about it.

What I think is happening: we currently have nothing preventing multiple rules from using the same rule ID. There's been some discussion around this recently, but I think this is undesirable behavior. IMO we should fail to run if `semgrep` detects multiple rules with the same ID. In this case I suspect we have a couple rules using the same ID, which is causing this issue.

Not only is this undesirable from a rule cleanliness point of view, but I believe these changes will introduce other subtle bugs. E.g. with these changes, if two rules use the same ID we will silently ignore some rules/findings:

```
In [1]: {r['id']: r['value'] for r in [{'id': 1, 'value': 'foo'}, {'id': 1, 'value': 'bar'}]}           
Out[1]: {1: 'bar'}
```

Even worse, we will only do this with SARIF output, so other outputs will appear normal. I think the fix here is to revamp our rule IDs. Early in the program we should detect duplicate IDs and warn the user and error out, IMO.

underyx: That's quite possible. Are we certain that multiple rules with the same rule ID is a bug and not a feature, though?

mschwager: > That's quite possible. Are we certain that multiple rules with the same rule ID is a bug and not a feature, though?

I definitely think it's a bug, but I'm open to hearing why it may be a feature. Here are some problems I see with duplicate rule IDs:

* This breaks rule discovery/usage in a number of different contexts: Google search, Github issues,`nosem` functionality, any future features around whitelisting/blacklisting individual rules, etc. I.e. using a rule ID in any of those contexts no longer lands me on a single rule. Google searching a rule ID may bring up two different results for two different people. Using `nosem` exceptions may disable more rules than intended. Rules should be one-to-one with their rule ID, at least insofar as we can control.
* As seen here, most tools assume a rule or linter is one-to-one with its ID. As far as I know every other linter/static analysis tool maintains this one-to-one relationship. Breaking this assumption will make it difficult to play well with other tools or systems.
* Also as seen here, a deficiency with duplicate rule IDs is the tendency to use `dict`s from ID -> rule, which will silently remove rules with the same ID from consideration. This is mostly something we'll have to consider internally, but a design prone to subtle bugs is not ideal.

I'd be curious to hear the advantages to duplicate rule IDs and how they weigh against the costs.

underyx: > I'm open to hearing why it may be a feature

I have no idea, I couldn't really understand what's going on in the Rule class. I was imagining stuff like, maybe a user would want to use multiple rule definitions with the same ID but various path include/excludes. Or different metadata for different patterns? One such case that comes to mind now is:

```yaml
rules:
  - id: wrong-user-agent
     patterns:
       - pattern: user_agent="..."
       - pattern-not: user_agent="foo"
     paths:
        - src/foo/**
  - id: wrong-user-agent
     patterns:
       - pattern: user_agent="..."
       - pattern-not: user_agent="bar"
     paths:
        - src/bar/**
```

mschwager (changes requested):
With the current implementation we're going to lose rules/results. If anything, we need to do something like:

```python
rules_by_id = collections.defaultdict(list)
for rule in rules:
    rules_by_id.append(rule)
```

mschwager:
I investigated and thought about this a bit more. My takeaway is that we _really_ don't want to support multiple rules using the same rule ID. I did a little testing of our SARIF output. Consider the following rules and code:

```
$ cat same.yaml 
rules:
- id: test-id
  message: Message 1
  pattern: foo(...)
  languages: [python]
  severity: ERROR
- id: test-id
  message: Message 2
  pattern: bar(...)
  languages: [python]
  severity: ERROR
```

```
$ cat test.py 
foo()
bar()
```

Not only does the SARIF output produce multiple `results` with the same `ruleId`, which is causing trouble with downstream applications, but it is also problematic in other areas of the schema. For example, under `rules`:

```
                    "rules": [
                        {
                            "id": "tmp.test-sarif.test-id",
                            "name": "tmp.test-sarif.test-id",
                            "shortDescription": {
                                "text": "Message 2"
                            },
                            "fullDescription": {
                                "text": "Message 2"
                            },
                            "defaultConfiguration": {
                                "level": "error"
                            },
                            "properties": {
                                "precision": "very-high",
                                "tags": []
                            }
                        },
                        {
                            "id": "tmp.test-sarif.test-id",
                            "name": "tmp.test-sarif.test-id",
                            "shortDescription": {
                                "text": "Message 1"
                            },
                            "fullDescription": {
                                "text": "Message 1"
                            },
                            "defaultConfiguration": {
                                "level": "error"
                            },
                            "properties": {
                                "precision": "very-high",
                                "tags": []
                            }
                        }
                    ]
```

Here we have two rules with the same ID and different messages. How would we determine which rule message to use if we combine rules with the same ID? If SARIF is _the_ canonical output format for static analysis results, and we're producing results that are fundamentally incompatible with the format, then I think that's a good indicator that we're taking the wrong approach. This means other static analysis tools subscribe to the one-rule <-> one-rule-id philosophy, and unless we have a really good reason not to, we should adopt that as well.

My recommendation is to have Semgrep error out if it detects multiple rules using the same ID and require users to correct the issue before proceeding. That way we can correct the user's error as soon as possible before it has the opportunity to cause subtle bugs in downstream applications.

mschwager:
I still don't think this is a great idea because it opens the door for silently hiding results, but if you'd like to move forward then I won't get in the way. You're correct that the `rule.id` is fully qualified, so that somewhat mitigates the concern:

```
ipdb> rules
frozenset({<Rule id=tmp.test-id>, <Rule id=tmp.test-id>})
ipdb> [r.id for r in rules]
['tmp.test-id', 'tmp.test-id']
```

We do have quite a few rules with duplicate rule-id's - here's what I could find:

```
{'eqeq-is-bad': 3,
 'telnet-request': 3,
 'disallow-old-tls-versions2': 2,
 'bypass-tls-verification': 4,
 'disallow-old-tls-versions1': 2,
 'http-request': 2,
 'ftp-request': 3,
 'render-template-string': 2,
 'avoid-raw': 2,
 'avoid-content-tag': 2,
 'avoid-html-safe': 2,
 'var-in-script-tag': 5,
 'var-in-href': 4,
 'incorrect-default-permission': 5,
 'use-of-md5': 2,
 'use-of-sha1': 2,
 'cookie-missing-httponly': 2,
 'avoid-bind-to-all-interfaces': 2,
 'use-of-weak-rsa-key': 2,
 'useless-if-conditional': 2,
 'useless-if-body': 2,
 'autoescape-disabled': 2,
 'jwt-exposed-credentials': 2,
 'jwt-none-alg': 2,
 'hardcoded-jwt-secret': 2,
 'eval-detected': 2,
 'template-explicit-unescape': 3,
 'insecure-hash-algorithm-md5': 3,
 'insecure-hash-algorithm-sha1': 3,
 'insufficient-rsa-key-size': 2,
 'insufficient-dsa-key-size': 2,
 'insecure-cipher-algorithm-rc4': 2,
 'insecure-cipher-algorithm-blowfish': 2,
 'path-traversal-open': 2,
 'template-href-var': 2,
 'template-autoescape-off': 2,
 'open-redirect': 2,
 'disabled-cert-validation': 2}
```

Note that many of these are in the same file, so fully qualified rule-id's won't save us.

semgrep/semgrep/output.py
ievans: Big 👍  to @mschwager comments about the downsides of duplicate IDs. I'd lean towards not letting a user define the same `id` in a single YAML file -- more trouble than it's worth to diagnose later, and a very uncommon usecase if we did make it a feature IMO.

DrewDennison:
@mschwager thanks for pulling up stats and doing the digging here. I strongly agree we should have uniq rule ids that are global and universal. I vote we close this and update our docs and lints saying that in a single file, there can never be duplicate rule ids. Then I believe fully-qualified / future namespaced rules will be globally unique

DrewDennison:
@underyx are you good with closing this?

underyx:
Very happy to close! When I brought this up, I just wanted to resolve the discussion finally :D

PR #1305, opened 2020-07-17
Go Metavariable Parsing Bug

colleend:
To fix parsing issue 638, added another pattern matching case to m_module_name_prefix to interpret import string as metavariable.


nbrahms:
I think you need to get your pfff changes merged first

nbrahms (approved):

PR #1307, opened 2020-07-17
Set TREESITTER_LIBDIR

mjambon:
These two other branches should be merged at the same time:

* https://github.com/returntocorp/ocaml-tree-sitter/pull/74
* https://github.com/returntocorp/ocaml-tree-sitter-lang/pull/1

The main highlight is that the semgrep-core build now uses a base image that comes with the right version of ocaml and most external opam packages pre-installed.

I had to fix the mess that GHA creates during the git checkout (or container launch). This is taken care of by the post-checkout scripts. Its output shows what's being done.


aryx (approved):
sounds good.
Not sure why the CI failure, but would be good to fix that first :)

PR #1310, opened 2020-07-20
Match filenames in metavars as literal strings

### LINKED ISSUE #638: can't use metavars in golang imports
This idiom works in python and javascript, but not golang: https://semgrep.live/8pB

```
    pattern: import ( $X )
```
fails to parse

nbrahms:
Previous work (8cc2954a) added the ability to match metavars to file
names within directives. This work was done specifically to enable
matching and reporting of Go imports.

However, reporting was incomplete here, as pretty-printing has not been
implemented for directives or their tributary AST nodes.

Here I instead bind a literal string expression to the metavariable,
rather than a FileName node. Since we have pretty-printing for such
expressions, reporting now works.

A side effect of this change is that the following pattern:

```go
  import "$X"

  x = $X
```

will match

```go
  import "fmt"

  x = "fmt"
```

with `$X` bound to `"fmt"`, which is a tad odd. This doesn't seem
particularly awful, and is particularly unlikely, so it's left as
a dangling edge case for now.

I've added a test pattern that can be used to verify functionality:

```bash
  semgrep --config semgrep-core/tests/go/imports2.sgrep \
    semgrep-core/tests/go/imports.go
```

Fixes #638

### DIFF
--- a/semgrep-core/matching/Generic_vs_generic.ml
+++ b/semgrep-core/matching/Generic_vs_generic.ml
@@ -161,10 +161,11 @@ let m_qualified_name a b =
 let m_module_name_prefix a b =
   match a, b with
   (* metavariable case *)
-  | A.FileName((a_str, _) as a1), B.FileName(_) as b_file
-    when MV.is_metavar_name a_str ->
-    let (b_file1, _) = b_file in
-    envf a1 (B.Modn b_file1)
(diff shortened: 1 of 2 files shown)


colleend (approved):
Should we get rid of the Modn addition to the AST in pfff because that's no longer necessary?

nbrahms:
> Should we get rid of the Modn addition to the AST in pfff because that's no longer necessary?

It's a good question, but it seems reasonable to me that `any` AST node should include module names 🤷

PR #1311, opened 2020-07-20
Install semgrep-core dependencies on setup

brendongo:
Now `make config; make setup; make` builds semgrep on machine with necessary build dependencies

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ rebuild:
 setup:
 	git submodule update --init --recursive
 	opam update -y
+	opam install -y --deps-only ./semgrep-core
 	opam install -y --deps-only ./pfff
 	cd ocaml-tree-sitter && ./configure && ./scripts/install-tree-sitter-lib
 	opam install -y --deps-only ./ocaml-tree-sitter


Makefile
```diff
@@ -42,6 +42,7 @@ rebuild:
 setup:
 	git submodule update --init --recursive
 	opam update -y
+	opam install -y --deps-only ./semgrep-core
 	opam install -y --deps-only ./pfff
 	cd ocaml-tree-sitter && ./configure && ./scripts/install-tree-sitter-lib
 	opam install -y --deps-only ./ocaml-tree-sitter
```
mjambon, line 48: I don't expect this to work because pfff and ocaml-tree-sitter are dependencies of semgrep-core. Maybe moving the line `opam install -y --deps-only ./semgrep-core` to last position works?

mjambon: I see it passes the tests. I don't know how this works. Is it due to the opam files found in `semgrep-core/`? In this case, it should be fine to remove `opam install -y --deps-only ./pfff` and `opam install -y --deps-only ./ocaml-tree-sitter`.

mjambon (approved):

mjambon:
This is now in `develop`. Closing.

PR #1312, opened 2020-07-21
Make error classes frozen

### LINKED ISSUE #1221: Semgrep 0.14 only showing first parse error
In Semgrep 0.14.0, it seems to only report the first parsing error. There are other files with parsing errors, and the parsing error changes across runs. Also, it seem that the parsing error now only shows the filename instead of the full path.
For my monorepo I can't tell which __init__.py has the parse error.

Running with `--verbose` does not show all parse errors.

brendongo:
Prior to this commit, it looked like only one error was being reported and other errors were
being silently hidden: https://github.com/returntocorp/semgrep/issues/1221. Root cause was that
all errors of the same type had the same hash and were considered duplicate errors https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/output.py#L242

There are issues with non frozen attr objects generating hashes. For more information: https://www.attrs.org/en/stable/hashing.html

This PR converts our error objects into frozen attr objects and removes explicit attr hash flags.
Adds a test that verifies an Error object hashes as expected and added semgrep-rule to check
for explicit hash flag setting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Fixed
 - Parse nested object properties in parameter destructuring in JavaScript
 - Parse binding patterns in ECMAScript 2021 catch expressions
+- Was mistakenly reporting only one of each type of issue even if multiple issues exist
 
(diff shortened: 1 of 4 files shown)


nbrahms (approved):
lgtm

Issue #1095, opened 2020-06-22
Pattern searching '#include' doesn't work in C


mschwager:
Similar to other languages, it would be great if we supported C's "import" functionality using '#include'. Currently, it looks like we can't even parse a pattern like '#include $X':

```
$ cat test.c 
#include <stdio.h>
#include <string.h>
#include "local.h"

int main() {
   printf("Hello, World!");
   return 0;
}
```

```
$ semgrep -l c -e '#include $X' test.c
...
  File "semgrep/semgrep/rule_lang.py", line 278, in parse_yaml_preserve_spans
    f"Something went wrong parsing Yaml (expected a YamlTree as output): {PLEASE_FILE_ISSUE_TEXT}"
Exception: Something went wrong parsing Yaml (expected a YamlTree as output): An error occurred while invoking the semgrep engine; please help us fix this by creating an issue at https://semgrep.dev
```

It looks like we're failing before even reaching `semgrep-core`. Maybe @rcoh knows what's going on.

rcoh:
Can you see what happens on my branch that fixes the none issue? Will debug when I get online this morning

rcoh:
As I suspected, `#include` is being interpreted as a YAML comment :roll_eyes: will fix

Issue #285, opened 2020-03-12
Module aliasing for Javascript


aryx:
@aryx 
Took some time to rethink your comment and try different code examples.

I agree with you, it would be way more practical to specify the module.
I broke down possible module import examples to this list:

```javascript
/**/
let foo = require('./foo');
foo.something('yo');

/**/
let fooSome = require('./foo').something;
fooSome('yo1');

/**/
import * as foo1 from './foo';
foo1.something('yo2')

/**/
import {something} from './foo';
something('yo3')

/**/
import foofoo from "./foo";
foofoo.something('yo4')

/**/
import { something as smth } from "./foo";
smth('yo5')

/**/
import("./foo").then(foo2 => {
  foo2.something('yo6');
})

/**/
(async function() {
  const foo3 = await import('./foo');
  foo3.something('yo7')
})()
```
and here is how the rule to catch em all would look like: 
```yaml
rules:
  - id: imports
    patterns:
        - pattern-either:
            - pattern: |
                import $X from "./foo";
                $X.something(...);
            - pattern: |
                $E = require("./foo");
                $E.something(...);
    message: "imports test"
    languages: [javascript]
    severity: ERROR
```

so the basic idea is to specify the exact module path
```
import $X from "./foo";
```
but have all the aliasing handled
```
$X.something(...);
```

_Originally posted by @inkz in https://github.com/returntocorp/sgrep/issues/197#issuecomment-596828741_

aryx:
related to issue #162

inkz:
just discovered an interesting case which I forgot to mention
```javascript
const foo = require('./foo')
const { smth, oneMore } = foo

smth("yo");
```

inkz:
most of this stuff is working now, so I believe that this issue can be closed

PR #1316, opened 2020-07-21
Bump cheatsheet functionality and fill a few gaps

mschwager:
We're pretty close to parity with our officially support languages except for a few gaps:

- JS import aliasing, this is enough of a painful mess with all the different `import`/`require` features that I'll tackle this separately. Tracking here: #285
- "Import" functionality in C - tracking here: #1095 
- Ruby still has quite a few gaps, I'll tackle those in a separate change


mschwager:
See also #1317, #1318, and #1319 tracking separately.

semgrep-core/tests/generate_test_matrix.py
```diff
@@ -40,6 +40,15 @@
     "exprstmt": "Expression and Statement",
 }
 
+LANGUAGE_EXCEPTIONS = {
+    "java": [
+        "naming_import",
```
ievans, line 45: Is there an issue tracking this?

semgrep-core/tests/generate_test_matrix.py
```diff
@@ -89,7 +98,14 @@ def generate_cheatsheet(root_dir: str):
...
                 entry = (read_if_exists(sgrep_path), sgrep_path,  read_if_exists(code_path), code_path)
                 print((lang, entry))
-                output[lang][VERBOSE_FEATURE_NAME.get(category, category)][VERBOSE_SUBCATEGORY_NAME.get(subcategory, subcategory)].append(entry)
+                feature_name = VERBOSE_FEATURE_NAME.get(category, category)
+                subcategory_name = VERBOSE_SUBCATEGORY_NAME.get(subcategory, subcategory)
+                language_exception = (
+                    feature_name in LANGUAGE_EXCEPTIONS.get(lang, [])
+                    or subcategory in LANGUAGE_EXCEPTIONS.get(lang, [])
+                )
+                if not language_exception:
```
ievans, line 107: is it really worth having language exception? I'd prefer to be explicit: "we don't have test cases for this, it is probably not supported"

ievans (approved):

semgrep-core/tests/generate_test_matrix.py
mschwager: From what I can tell, Java doesn't support import aliasing. To address the comment below as well, this list _is_ the explicit declaration that a language doesn't support something. A missing test tells me "someone forgot to test this, or maybe it's not supported" whereas this list tells me "this language doesn't support this feature." Or, at least, that's what I was hoping. Is there a better way we can state that?

PR #1320, opened 2020-07-21
Only print out errors to stderr when not using TEXT output

### LINKED ISSUE #1243: semgrep 0.14.0 python API issues
**Describe the bug**
I am trying to upgrade to semgrep 0.14.0.

Setting`quiet=True` is not respected by the progress bar and initial message, `running `x` rules.`
Ref: https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/semgrep_main.py#L250
Also, this will trigger an exception.

Example:

```
    io_capture = StringIO()
    output_handler = OutputHandler(
        OutputSettings(
            output_format=OutputFormat.JSON,
            output_destination=None,
            quiet=True,
            error_on_findings=False,
            strict=True,
        ),
        stdout=io_capture,
    )
    semgrep_main.main(
        output_handler=output_handler,
        target=[str(t) for t in targets],
        pattern="",
        lang="",
        config=str(config),
        **kwargs,
    )
    output_handler.close()
    return json.loads(io_capture.getvalue())
```
When `quiet=True`semgrep shows `running x rules` message and progress bar and triggers and exception,
```
running 73 rules...
  8%|█████▊                                                                |6/73
 in _invoke_semgrep
    return json.loads(io_capture.getvalue())
  File "/xxx/.pyenv/versions/3.7.7/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/xxx/.pyenv/versions/3.7.7/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/xxx/.pyenv/versions/3.7.7/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
It looks like when `quiet=True`, `io_capture.getvalue()` returns and empty string. Everything works as expected when `quite` is set to `True`

**Environment**
0.14.0 Pypi

brendongo:
JSON and SARIF format  includes errors in the output so printing to
stderr is redundant.


brendongo:
Fixes https://github.com/returntocorp/semgrep/issues/1243

semgrep/semgrep/output.py
```diff
@@ -242,7 +242,10 @@ def handle_semgrep_error(self, error: SemgrepError) -> None:
         if error not in self.error_set:
             self.semgrep_structured_errors.append(error)
             self.error_set.add(error)
-            logger.error(str(error))
+            if self.settings.output_format == OutputFormat.TEXT:
```
nbrahms, line 245: If we were fancy this would be a method on `OutputFormat`, so we don't have to remember to update this as we monkey with `OutputFormat` objects.

But this works too :)

nbrahms (approved):

PR #1324, opened 2020-07-22
Add 'metavariable-regex' operator

### LINKED ISSUE #1217: pattern-regex on a single metavariable
I've seen several cases where it would be helpful to make pattern-regex operator not on a full range, but on a single metavariable.

E.g.: 
```
- pattern: 'foo($X)
- pattern-regex:
   - regex: '[0-9]'
   - metavariable: '$X'
 ```

### PR DESCRIPTION
**Pull request by mschwager**
Fix #1217. The 'metavariable-regex' operator filters finding's by metavariable value against a Python re.match compatible expression.

### DIFF
--- a/docs/configuration-files.md
+++ b/docs/configuration-files.md
@@ -18,6 +18,7 @@ Contents:
   * [`patterns`](configuration-files.md#patterns)
   * [`pattern-either`](configuration-files.md#pattern-either)
   * [`pattern-regex`](configuration-files.md#pattern-regex)
+  * [`metavariable-regex`](configuration-files.md#metavariable-regex)
   * [`pattern-not`](configuration-files.md#pattern-not)
   * [`pattern-inside`](configuration-files.md#pattern-inside)
   * [`pattern-not-inside`](configuration-files.md#pattern-not-inside)
@@ -109,6 +110,7 @@ The below optional fields must reside underneath a `patterns` or `pattern-either
 
 | Field | Type | Description |
 | :--- | :--- | :--- |
+| [`metavariable-regex`](configuration-files.md#metavariable-regex) | `map` | Search metavariables for [Python `re.match`](https://docs.python.org/3/library/re.html#re.match) compatible expressions. |
 | [`pattern-not`](configuration-files.md#pattern-not) | `string` | Logical NOT - remove findings matching this expression. |
 | [`pattern-inside`](configuration-files.md#pattern-inside) | `string` | Keep findings that lie inside this pattern. |
 | [`pattern-not-inside`](configuration-files.md#pattern-not-inside) | `string` | Keep findings that do not lie inside this pattern. |
@@ -196,6 +198,30 @@ rules:
 
 *Note that single (`'`) and double (`"`) quotes [behave differently](https://docs.octoprint.org/en/master/configuration/yaml.html#scalars)
 in YAML syntax. Single quotes are typically preferred when using backslashes (`\`) with `pattern-regex`.*
+
+### `metavariable-regex`
+
+The `metavariable-regex` operator searches metavariables for a [Python `re.match`](https://docs.python.org/3/library/re.html#re.match)
+compatible expression. This is useful for filtering results based on a [metavariable's](/docs/pattern-features.md#metavariables)
+value.
+
+**Example**
+
+The `metavariable-regex` operator is a mapping which requires the
+`metavariable` and `regex` keys. It can be combined with other pattern operators:
+
+```yaml
+rules:
+  - id: insecure-methods
+    patterns:
+      - pattern: module.$METHOD(...)
+      - metavariable-regex:
+          metavariable: '$METHOD'
+          regex: '(insecure1|insecure2|insecure3)'
+    message: "module using insecure method call"
+    languages: [python]
+    severity: ERROR
+```
 
 ### `pattern-not`
 
--- a/semgrep/semgrep/evaluation.py
+++ b/semgrep/semgrep/evaluation.py
@@ -13,7 +13,6 @@ logger = logging.getLogger(__name__)
 
 from semgrep.constants import RCE_RULE_FLAG
 from semgrep.error import NEED_ARBITRARY_CODE_EXEC_EXIT_CODE
-from semgrep.error import InvalidRuleSchemaError
 from semgrep.error import SemgrepError
 from semgrep.error import UnknownOperatorError
 from semgrep.pattern_match import PatternMatch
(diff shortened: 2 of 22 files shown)

### CODE REVIEW

File: semgrep/semgrep/evaluation.py [hunk 1]
```diff
@@ -134,6 +161,30 @@ def _evaluate_single_expression(
             }
         )
         return output_ranges
+    elif expression.operator == OPERATORS.METAVARIABLE_REGEX:
+        if (
+            not isinstance(expression.operand, dict)
+            or "metavariable" not in expression.operand
+            or "regex" not in expression.operand
+        ):
+            raise SemgrepError(
```
ievans, line 170: is this the right error type? Don't we have an error type that represents invalid (semantically) YAML pattern?

ievans (approved):
Looks great @mschwager 

@aryx with this landing, should we soft-deprecate / stop advertising O'Caml regex inside semgrep-core?

semgrep/semgrep/evaluation.py
mschwager: Good catch! Are you thinking of `InvalidPatternError`? That seems like the best fit here

mschwager: Actually, `InvalidRuleSchemaError` may be better since this isn't a pattern

docs/configuration-files.md
```diff
@@ -197,6 +199,30 @@ rules:
 *Note that single (`'`) and double (`"`) quotes [behave differently](https://docs.octoprint.org/en/master/configuration/yaml.html#scalars)
 in YAML syntax. Single quotes are typically preferred when using backslashes (`\`) with `pattern-regex`.*
 
+### `metavariable-regex`
```
nbrahms, line 202: ❤️ for documentation

semgrep/semgrep/evaluation.py
```diff
@@ -99,7 +126,8 @@ def _evaluate_single_expression(
                 f"at least one rule needs to execute arbitrary code; this is dangerous! if you want to continue, enable the flag: {RCE_RULE_FLAG}",
                 code=NEED_ARBITRARY_CODE_EXEC_EXIT_CODE,
             )
-        assert expression.operand, "must have operand for this operator type"
+        if not isinstance(expression.operand, str):
+            raise InvalidRuleSchemaError("pattern-where-python must have a string value")
```
nbrahms: Is this the correct error message? What does this case have to do with `pattern-where-python`?

semgrep/semgrep/rule.py
```diff
@@ -172,11 +179,11 @@ def _parse_boolean_expression(
...
 
     @staticmethod
-    def _validate_operand(operand: YamlTree) -> str:  # type: ignore
-        if not isinstance(operand.value, str):
+    def _validate_operand(operand: YamlTree) -> Union[str, dict]:  # type: ignore
+        if not (isinstance(operand.value, str) or isinstance(operand.value, dict)):
             raise InvalidRuleSchemaError(
                 short_msg="invalid operand",
-                long_msg=f"type of `pattern` must be a string, but it was a {type(operand.unroll()).__name__}",
+                long_msg=f"type of `pattern` must be a string or dict, but it was a {type(operand.unroll()).__name__}",
```
nbrahms: This may be confusing, as the `dict` presumably won't be in a `pattern` key...

semgrep/tests/e2e/rules/metavariable-regex-multi-regex.yaml
```diff
@@ -0,0 +1,13 @@
+rules:
+  - id: metavar-test-multi-regex
+    patterns:
+      - pattern: 'metavariable_regex_test.$X($Y)'
```
nbrahms, line 4: I'd be curious to see a negative test case for:
```yaml
- pattern:
    metavariable: '$X'
    regex: 'method1'
```

semgrep/semgrep/evaluation.py
mschwager: This was more of a code hygiene fix since `operand` is no longer guaranteed to be a string, so this appeases mypy. Also, when Python is run in optimize mode `assert` statements are removed, so we shouldn't rely on them:

```
$ python --help
...
-O     : remove assert and __debug__-dependent statements; add .opt-1 before
         .pyc extension; also PYTHONOPTIMIZE=x
-OO    : do -O changes and also discard docstrings; add .opt-2 before
         .pyc extension
...
```

semgrep/semgrep/rule.py
mschwager: I think this should be "type of `operand` ...". The function is `_validate_operand` with `operand` as an argument. I can bump the message

semgrep/tests/e2e/rules/metavariable-regex-multi-regex.yaml
mschwager: That gives me:

```
running 1 rules...
	invalid yaml: wrong rule fields
An error occurred while invoking the semgrep engine; please help us fix this by creating an issue at https://semgrep.dev
```

Is there something specific you're trying to test for here? This PR isn't as precise as it could be. Ideally we'd want to avoid all operators allowing `str|dict` and instead only allow specific operators to be their specific type, but that's a much larger effort and refactor.

mschwager:
Looks like we have to use `SemgrepError` in `semgrep/semgrep/evaluation.py`. Using any error that inherits from `ErrorWithSpan` requires access to the YAML span corresponding to that operator, which we don't have during evaluation.

There are a few suboptimal changes in this PR, but it'd probably be easier to do a refactor in a separate change than try to do everything at once here. The biggest one I see is: it'd be great to do _all_ rule validation in `semgrep/semgrep/rule.py` rather than dispersed across the codebase. We should define the type we expect each operator to be and compare the rules against that early in the program. This allows us to include all our validation in one place and print useful span information if we see any validation errors.

aryx:
@ievans the OCaml regexps were used mostly for strings, with patterns like foo("=~/xxx/"), and maybe
some fieldname in Javascript, so they were never used for metavariables restrictions.

docs/configuration-files.md
```diff
@@ -197,6 +199,30 @@ rules:
...
+The `metavariable-regex` operator is a mapping which requires the
+`metavariable` and `regex` keys. It can be combined with other pattern operators:
+
+```yaml
+rules:
+  - id: insecure-methods
+    patterns:
+      - pattern: module.$METHOD(...)
+      - metavariable-regex:
+          metavariable: '$METHOD'
```
aryx, line 219: Why the quotes around $METHOD?
Why the quotes around the regexp below? The main advantage of YAML is not having
to use "" as in JSON around strings.
We do not use quotes around semgrep patterns for examples.

aryx (commented):
Looks great. I remember @ievans wanted to express different kinds of constraints on metavariables.
Do we need something more than regexps? (we have "typed metavariables" in semgrep-core,
but do you have something else in mind?)

Issue #1173, opened 2020-06-30
Order of named fields in struct matters


dlukeomalley:
In Go, I _think_ the order of named fields in structs doesn’t matter (much like Python keyword args), but it seems Semgrep doesn’t understand that:

not working: https://semgrep.live/oqjX
not working: https://semgrep.live/zdjn
working: https://semgrep.live/pKde

dgryski:
Value ordering in struct literals only applies to un-keyed literals.  (Note that golint complains about un-keyed literals because they break compatibility with newer versions of packages with extra fields.). Matching on field assignments for unkeyed literals might be nice, but would require full type information from third-party packages.

PR #1332, opened 2020-07-22
Go Struct Order Fix

colleend:
fix to the go struct ordering bug, issue #1173 . Added a case to match ellipses on both sides -- this case will match anything, regardless of order. There is an issue where this will match a correct line twice.


aryx:
I don't think it's the right way to fix the issue. One problem is that a pattern like f(..., 1, 2, ...) should match f(0,1,2,3) but should not match f(0,2,1,3) but with this PR it will.

The main issue is that the generic AST generated for composite literal is not great. Here is what I get with pfff -dump_ast on your Go example for the last variable 'c':

aryx:
```
                 (DefStmt
                    ({ name = ("c", ()); attrs = [(KeywordAttr (Var, ()))];
                       info =
                       { id_resolved = ref (None); id_type = ref (None);
                         id_const_literal = ref (None) };
                       tparams = [] },
                     (VarDef
                        { vinit =
                          (Some (Call ((IdSpecial (New, ())),
                                   ((),
                                    [(ArgType
                                        (TyName
                                           (("Address", ()),
                                            { name_qualifier = None;
                                              name_typeargs = None })));
                                      (Arg
                                         (Tuple
                                            [(Id (("street", ()),
                                                { id_resolved = ref (None);
                                                  id_type = ref (None);
                                                  id_const_literal =
                                                  ref (None) }
                                                ));
                                              (L (String ("hello", ())))]));
                                      (Arg
                                         (Tuple
                                            [(Id (("name", ()),
                                                { id_resolved = ref (None);
                                                  id_type = ref (None);
                                                  id_const_literal =
                                                  ref (None) }
                                                ));
                                              (L (String ("what", ())))]));
                                      (Arg
                                         (Tuple
                                            [(Id (("city", ()),
                                                { id_resolved = ref (None);
                                                  id_type = ref (None);
                                                  id_const_literal =
                                                  ref (None) }
                                                ));
                                              (L (String ("SF", ())))]))
                                      ],
                                    ())
                                   )));
                          vtype = None })))
                 ],
               ()))
           })))
  ]
```

aryx:
We don't want the composite literal Foo {a: 1, b:2 } to be converted in a flat list of arguments of tuples. Instead we want to generate a Record (see AST_generic.ml) or a Container (Dict) which then have their elements matched in any order (see Generic_vs_generic.ml and the code to handle Dict with m_container_set_or_dict_unordered_elements).

aryx:
For example here is the generic AST generated for 'x = {a:2, b:3}' in Python.
```
+ /home/pad/pfff/pfff -dump_ast dict.py
[(ExprStmt (
    (Assign (
       (Id (("x", ()),
          { id_resolved = ref (None); id_type = ref (None);
            id_const_literal = ref (None) }
          )),
       (),
       (Container (Dict,
          ((),
           [(Tuple
               [(Id (("a", ()),
                   { id_resolved = ref (None); id_type = ref (None);
                     id_const_literal = ref (None) }
                   ));
                 (L (Int ("2", ())))]);
             (Tuple
                [(Id (("b", ()),
                    { id_resolved = ref (None); id_type = ref (None);
                      id_const_literal = ref (None) }
                    ));
                  (L (Int ("3", ())))])
             ],
           ())
          ))
       )),
    ()))
  ]
master/home/pad/pfff/tests/python/parsing $ 
```

aryx:
So the right fix is probably to modify go_to_generic.ml in the pfff repository to better handle CompositeLit, and for example
if all its arguments are of the form InitKeyValue, then we can convert that in a New ("Address", Dict [ ...])

aryx (changes requested):
The test files in this PR are good, but the right fix is to first modify pfff/.../go_to_generic.ml to better handle CompositeLit.

semgrep-core/matching/Generic_vs_generic.ml
```diff
@@ -16,6 +16,7 @@
  *)
 (*e: pad/r2c copyright *)
 open Common
+open List
```
aryx: We don't have any ocaml coding style with @mjambon yet, but I usually try to avoid to use 'open' and prefer to qualify functions used in other modules because it's harder then when you see just  'foo arg' to see to which module it belongs to. You can use IDEs to help, but it still IMHO makes the code slightly harder to read.

There are a few exceptions when it's just annoying to qualify the same module again and again. For example I usually do open AST_generic when I know a module will exclusively use entities defined in this AST.

aryx (dismissed):
see conversation

semgrep-core/matching/Generic_vs_generic.ml
```diff
@@ -995,7 +996,17 @@ and m_list__m_argument (xsa: A.argument list) (xsb: A.argument list) =
   (* dots: ..., can also match no argument *)
   | [A.Arg (A.Ellipsis _i)], [] ->
       return ()
-
+  (* Fix to issue 1173. Case where there are ellipsis to the left and right of a statement.
+    * Order should not matter in this case, as we can see from the go structs example.
+    * However, there is an issue where this will match a correct line twice occasionally.
+    * Look into m_list_in_any_order for correction in this case. *)
+   | (A.Arg (A.Ellipsis _i))::xsa, xsb
+     when let rev_xsa = rev xsa in
```
nbrahms: Is this trying to get the last element in the list?

If so, I think you can do `List.nth xsa (List.length xsa - 1)`.

I'm also wondering what happens with ellipses in the middle of the list; e.g. `..., a, ..., b, ...`. I'm wondering if we need some sort of recursive solution here? (cc. @aryx )

semgrep-core/matching/Generic_vs_generic.ml
```diff
@@ -995,7 +996,17 @@ and m_list__m_argument (xsa: A.argument list) (xsb: A.argument list) =
   (* dots: ..., can also match no argument *)
   | [A.Arg (A.Ellipsis _i)], [] ->
       return ()
-
+  (* Fix to issue 1173. Case where there are ellipsis to the left and right of a statement.
+    * Order should not matter in this case, as we can see from the go structs example.
+    * However, there is an issue where this will match a correct line twice occasionally.
+    * Look into m_list_in_any_order for correction in this case. *)
+   | (A.Arg (A.Ellipsis _i))::xsa, xsb
```
nbrahms: To what extent is our desire to match in any order due to the pattern syntax; vs. the parent AST node?

I might imagine instead defining two different arg matching paths, using one from, e.g., function argument list, and using the other from object, import, and struct definition lists. cc. @aryx for thoughts here.

semgrep-core/matching/Generic_vs_generic.ml
aryx: There's also Common2.list_last.
As I said, improving the arguments matching for Composite literals is not the right approach.
We already have code to handle matching elements in any order for Dict, fields, etc. We just need
to generate something better in go_to_generic.ml from CompositeLit.

semgrep-core/matching/Generic_vs_generic.ml
aryx: I'm not sure I understand what you mean, but when we are in a object/records/hash definition context, then
the semantic of the language says that the order of the field does not really matter, which is why we want to match those in any order.
For function arguments, we match certain arguments in any order, but only for the keyword arguments (ArgKwd in AST_generic.ml).

nbrahms: Yup, what you say is what I was suggesting: whether a list of arguments is order-independent should probably depend on the semantic context of the language (which ideally is represented in the parent node type of the generic AST).

The change here, in contrast, I believe uses the pattern API to determine whether matches are order-independent, rather than the semantic context.

semgrep-core/matching/Generic_vs_generic.ml
mjambon: Whatever you all do, please avoid exceptions when you can (unless you do want a fatal error): if you want the last element of a list, such function should return an option because the list could be empty. Similarly, don't use `List.hd`, use pattern matching or a function that returns an option, like:
```ocaml
let first = function [] -> None | x :: _ -> Some x
```

semgrep-core/matching/Generic_vs_generic.ml
colleend: done!

nbrahms (approved):

semgrep-core/matching/Generic_vs_generic.ml
colleend: done! thanks for the comments!

semgrep-core/matching/Generic_vs_generic.ml
colleend: fixes issues in pfff together!

PR #1334, opened 2020-07-22
Add setup-ocaml as a submodule

mjambon:
It's for building and pushing custom Docker images with most dependencies pre-installed. In particular, this includes the version of ocaml we want and pre-installed opam packages.

> https://github.com/returntocorp/setup-ocaml


aryx:
Why does this have to be a submodule of semgrep? We will call stuff from inside setup-ocaml/ from the Makefile in semgrep?

PR #1335, opened 2020-07-23
Try more effective CI triggers

mjambon:
* Long workflows (> 10 min or so) were disabled on feature branches so as to not block PRs.
* The "env tests" workflow now runs on every push so we know if there's a problem as soon as the commit is pushed.
* The "env tests" cancels the previous run if there's one (on the same branch).

It looks like the docker.yml flow does some sort of deployment that would be useful for manual testing? I commented it out because it takes a long time.


brendongo:
Pulled in most changes to https://github.com/returntocorp/semgrep/pull/1347

PR #1347, opened 2020-07-23
ci: split build tests from unit/e2e tests

brendongo:
All unit/e2e tests are in tests.yaml while all built tests are in built-test.yaml

Everything in tests.yaml runs in ~5 minutes


mjambon (approved):
awesome

Issue #1200, opened 2020-07-06
run pre-commit hook without Docker


ievans:
cc @underyx 

Now that semgrep is pip installable, we should be able to run the pre-commit hook via a pip install rather than requiring docker.

Also the instructions semgrep.live are missing a `v` in front of the revision.

mschwager:
One downside here is we won't be able to run `pre-commit` with the bleeding edge changes on `develop`. Perhaps we use Docker for `develop` and dogfood that internally, and offer the `pip` package externally as the default/paved road.

brendongo:
Good point @mschwager do we care about the experience of running cutting edge develop as a pre-commit hook locally or just checking that it doesnt break anything? We can run another pre-commit-config in CI that includes running semgrep-develop docker image if that is the case.

ievans:
@mschwager @brendongo tbh I'm not sure we care about testing that, can probably do the cheap thing for now

ievans:
This was closed in https://github.com/returntocorp/semgrep/pull/1401

PR #1349, opened 2020-07-23
pre-commit: use pypi install of semgrep instead of docker

brendongo:
Running with docker leads to slowness due to filesystem issues with
OSX. Since we have a semgrep published in pypi might as well use it.

Note that we have to have a to use a separate repo because of how
semgrep is built.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,28 +50,28 @@ repos:
         additional_dependencies: ["flake8-bugbear==20.1.4"]
         args: ["--select=B,E9,F63,F7,F82"]
 
-  - repo: https://github.com/returntocorp/semgrep
-    rev: 'v0.14.0'
+  - repo: https://github.com/returntocorp/semgrep-pre-commit
+    rev: 'v0.16.0'
     hooks:
-      - id: semgrep-develop
+      - id: semgrep
         name: Semgrep Python
         types: [python]
         exclude: "^semgrep/tests/.+$|^install-scripts/.+$|^release-scripts/.+$|^semgrep/setup.py$"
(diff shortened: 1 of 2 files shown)


brendongo:
Relevant https://github.com/returntocorp/semgrep/issues/1200

mschwager:
> Note that we have to have a to use a separate repo because of how semgrep is built.

Could you elaborate on this? We can't just update `.pre-commit-hooks.yaml`?

brendongo:
If you want to point to a github repo python project it looks like pre-commit goes to the project root and runs `pip3 install .` so assumes there is a top level setup.py and does not work for us since we don't want users to have to build semgrep-core and such

mschwager:
> If you want to point to a github repo python project it looks like pre-commit goes to the project root and runs `pip3 install .` so assumes there is a top level setup.py and does not work for us since we don't want users to have to build semgrep-core and such

Ahh, interesting. I wish they would allow you to specify a repository's subdirectory :/

My concern here is that it'll be easy for these files and repositories to get out of sync. And this is Yet Another Repository that we have to worry about, and another level of indirection that people will need to know about in the `semgrep` ecosystem. I wonder if we could instead include a dummy `setup.py` file in the root of this repository and list `semgrep` in [`additional_dependencies`](https://pre-commit.com/#config-additional_dependencies). Or even include the necessary files/information in https://github.com/returntocorp/semgrep-pre-commit in this repository instead.

brendongo:
Interesting. Yeah we could have a setup.py at this repo with semgrep as a dependency :D and just add a comment that this is for pre-commit

brendongo:
Moving to https://github.com/returntocorp/semgrep/pull/1401

Issue #1330, opened 2020-07-22
Add Native (non-WSL) Windows support to semgrep


mschwager:
Creating this issue to begin tracking requests for Semgrep support on Windows. Note that we expect Semgrep to work via `pip` installation on WSL, so this issue is focused on native Windows support. If you need _native_ Windows support please chime in here :+1: 

Depends on:
- [x] #2956
- [ ] #2991
- [x] #3061
- [ ] [issue needed] Eliminate calls to fork(), which requires parser to run in the same semgrep process, but has a memory leak as a workaround for a segfault that would need to be fixed ([segfault issue](https://github.com/returntocorp/ocaml-tree-sitter-semgrep/issues/65)). The memory leak may be acceptable for small scans.
- [x] [issue needed] Set up Windows testing in CI

parsiya:
Attaching the error after running `pip3 install semgrep` on Windows 10 (1909) and Python 3.8.1. Seems to be a similar issue to #1336 but not the path separator problem.

I am running it via WSL2 so it's not a deal breaker but it would be nice to have Windows support for the VS Code extension (unless I can figure out how to run it via WSL2).

The log is attached but the important part appears to be:

```
error: can't copy 'C:\Users\phakimian\AppData\Local\Temp\1\pip-install-r20usipa\semgrep-files\semgrep-core': doesn't exist or not a regular file
```
Looking at the log, I cannot see the file being copied. 

Unrelated: There's also this warning so something to look out for in the future.

```
DEPRECATION: Could not build wheels for semgrep which do not use PEP 517. pip will fall back to legacy
'setup.py install' for these. pip 21.0 will remove support for this functionality. A possible replacement is to
fix the wheel build issue reported above. You can find discussion regarding this at
https://github.com/pypa/pip/issues/8368.
```

[pip-install-semgrep-windows.txt](https://github.com/returntocorp/semgrep/files/5070821/pip-install-semgrep-windows.txt)

bobby-lin:
Bumping this up. Is this work in progress?

zain-CSsol:
Yes i think it's still in progress i am also backtracking the issue for solution, anyhow if any one of you came to know the solution then do let me know

DrewDennison:
related to #2252 and #2565 we would need to do cross-platform builds as part of our release process

DrewDennison:
marking as low priority given WSL and docker workarounds

fasiha:
[...]

PR #1353, opened 2020-07-24
support windows path

### LINKED ISSUE #1336: Invalid Path Separator on Windows
"pip3 install semgrep" fails on windows 10 with Python 3.7.8 and pip 20.1.1 the error seems to be an invalid path separator char.
error: can't copy 'XXXXXXXXXXXXXX\Local\Temp\pip-install-cq40rzma\semgrep-files/semgrep-core': doesn't exist or not a regular file

Anyone here know how to fix that?

DrewDennison:
Fix #1336

### DIFF
--- a/semgrep/setup.py
+++ b/semgrep/setup.py
@@ -86,7 +86,9 @@ class PostInstallCommand(install):
             os.putenv("SKIP_NUITKA", "TRUE")
             if "osx" in distutils.util.get_platform():
                 with chdir(repo_root):
-                    os.system(os.path.join(repo_root, "release-scripts", "osx-release.sh"))
+                    os.system(
+                        os.path.join(repo_root, "release-scripts", "osx-release.sh")
+                    )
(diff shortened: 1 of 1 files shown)


dlukeomalley:
@DrewDennison To call get this to done-done, can we also deploy a Semgrep rule to avoid this issue in the future? Suspect it's an easy mistake for us to make and we don't formally test on or support Windows.

mschwager:
I'm not sure, but I suspect this won't be enough to support Windows. `semgrep-core` likely has to be re-compiled for that environment. This is a good change regardless - it's always good to be using filesystem independent paths. However, it may actually be best to detect a Windows environment in `setup.py`, print something like "Windows is not supported yet, please see #1330", and exit vs. having some weird runtime error with `semgrep-core`.

dlukeomalley:
+1, I don't want to yet give the impression we support Windows. Perhaps hard exiting is too strong, but a nice loud warning instead and a note to switch to WSL.

PR #1359, opened 2020-07-24
highlights in cheatsheet output

ievans:


aryx (commented):
Noice! Can get rid of the //ERROR: comment then? (which changes the position though).

aryx (commented):
maybe you can just replace the //ERROR: with the same number of spaces ...

aryx (approved):
lgtm

PR #1363, opened 2020-07-24
Use version constant

brendongo:


mschwager:
Typically the convention for Python packages is to include this kind of information in the top-level `__init__.py` file for the package. E.g. here's how I do it in Dlint: https://github.com/dlint-py/dlint/blob/master/dlint/__init__.py#L16, then in `setup.py`: https://github.com/dlint-py/dlint/blob/master/setup.py#L29. This will also fix your CI problem because then `setup.py` can run without requiring that the package's dependencies are already installed.

semgrep/semgrep/__init__.py
```diff
@@ -0,0 +1 @@
+__VERSION__ = "0.17.0"
```
mschwager, line 1: I believe the convention here is to use lowercase, e.g. `__version__`

mschwager (approved):
LGTM! Always nice to minimize the number of places you have to bump when releasing a new version :+1:

PR #1369, opened 2020-07-27
Use -timeout for the matching process and improve debugging experience

aryx:


semgrep-core/bin/Main.ml
```diff
@@ -55,16 +55,16 @@ module J = Json_type
 let env_debug = "SEMGREP_CORE_DEBUG"
 let env_profile = "SEMGREP_CORE_PROFILE"
 
+(* see also verbose/... flags in Flag_semgrep.ml *)
+(* to test things *)
+let test = ref false
```
ievans, line 60: is it a flag? a mode to run the test suite? what is its semantic effect? perhaps a more descriptive name would be helpful

PR #1373, opened 2020-07-27
tests: add basic perf regression test

### LINKED ISSUE #1362: Benchmark CI job that loudly fails performance regressions
**Describe the bug**
Current Benchmark CI job gathers performance data but does not block PRs on regressions

**Expected behavior**
Given a set of rules and a project we should have some target run time in mind and the CI job should exit as a failure
and block a PR when it goes above such target time.

brendongo:
Removes old benchmarking code and add a simple test that asserts semgrep finishes running on a specific ruleset/target in the expected amount of time. These tests will run as part of qa tests which for now are only run after merges to develop


semgrep/tests/performance/test_ci_perf.py
```diff
@@ -1,203 +1,49 @@
...
-        RepoCase("https://github.com/getsentry/sentry", "f25ea5dc", "python"),
-        RepoCase(
-            "https://github.com/bkimminich/juice-shop.git", "e03629b", "javascript"
-        ),
-    ],
-    ids=lambda case: case.test_id,
-)
-def test_default_packs(run_semgrep_in_tmp, clone_github_repo, benchmark, repo_case):
+@pytest.mark.qa
+def test_perf(clone_github_repo):
```
mschwager, line 8: Instead of testing 3 repositories in one function manually can we instead [parametrize](https://docs.pytest.org/en/latest/parametrize.html) the function to take a repo, SHA hash, and expected runtime? This will improve extensibility and allow us to easily test one set of parameters from the command line (via `pytest`'s `-k` flag). I think `xdist` will also allow us to parallelize the 3 repositories.

mschwager (commented):
I really like that we're simplifying things here. The previous benchmarking functionality felt a bit over engineered, which made it difficult to understand and error prone. My one high-level question is: do we trust this approach to not be flaky in the face of wildly different hardware? My machine may be different from your machine which may be different from the CI machine. Ideally the tests should run and pass on each.

install-scripts/latest-artifact-for-branch.py
```diff
@@ -100,12 +100,6 @@ def install_semgrep_core(url: str) -> None:
         sys.exit(1)
 
 
-def download_benchmarks(url: str, output: str) -> None:
```
mschwager, line 103: I think we can remove this whole file, it's only referenced in `benchmark.yml`.

semgrep/tests/performance/test_ci_perf.py
brendongo: Will make this change

brendongo: Actually this only tests two repositories with a shared rule repo and the repositories are executed differently (cause of --exclude in one of them) and have different expected timeout. So not a lot of shared code. Thoughts on parametrizing if/when we add more cases in the future?

mschwager: Sounds good :+1:

semgrep/tox.ini
```diff
@@ -11,4 +11,3 @@ setenv =
     PIPENV_VERBOSITY = -1
 
 [pytest]
```
mschwager: Looks like we can remove this whole `[pytest]` key now.

mschwager (approved):

brendongo:
talked on slack re different machine performance: not sure how to really handle it but having this as a canary on merges to develop is probably a good first step and we can improve later as needed.

semgrep/tox.ini
brendongo:
```suggestion
```

PR #1374, opened 2020-07-27
Upgrade ocaml-tree-sitter runtime

mjambon:
This comes with a renaming of polymorphic variants. It wasn't tricky, just slightly tedious.

This upgrades the ocaml-tree-sitter runtime, and in particular the matching/backtracking engine. I expect this to be less efficient than before but hopefully it's negligible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@
 #
 .PHONY: build
 build:
-	$(MAKE) -C build-core
+	$(MAKE) build-core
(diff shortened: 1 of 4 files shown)


aryx:
hmm is this a duplicate of the one I already accepted?

mjambon:
> hmm is this a duplicate of the one I already accepted?

yes, this set of changes is included in the PR #1376 you already accepted.

PR #1376, opened 2020-07-28
Update to reflect more name changes

mjambon:
Extends https://github.com/returntocorp/semgrep/pull/1374 to support changes from https://github.com/returntocorp/ocaml-tree-sitter/pull/82

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@
 #
 .PHONY: build
 build:
-	$(MAKE) -C build-core
+	$(MAKE) build-core
 	cd semgrep && python3 -m pipenv install --dev
 
 .PHONY: build-core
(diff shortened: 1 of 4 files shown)


aryx (approved):
Looks better indeed! I like it.
How did you automate the change to the Parse_xxx_tree_sitter.ml? sed?

mjambon:
> How did you automate the change to the Parse_xxx_tree_sitter.ml? sed?

I only partially automated this, using emacs macros (F3 to start recording, F4 to stop recording, F4 again to replay).

What I didn't automate is turning `Bin_op_foo` to just `foo` because it requires identifying the prefix to remove. In theory, the prefix `Bin_op_` could be derived from the name of the containing function `binary_operator` but the rules are complicated and this should have been done in ocaml (where the rules are already implemented).

Issue #1361, opened 2020-07-24
Semgrep on JuiceShop stalling on three.js


brendongo:
**Describe the bug**
Running `time semgrep -f ~/Code/njsscan/njsscan/rules/semantic_grep/ ~/Downloads/juice-shop-master/` takes forever but running `time semgrep -f ~/Code/njsscan/njsscan/rules/semantic_grep/ ~/Downloads/juice-shop-master/ --exclude three.js` takes 90 seconds.

It looks like parsing step is only taking 0.5 seconds to parse three.js so this is stalling in the match step.

Rules: https://github.com/ajinabraham/njsscan/tree/master/njsscan/rules/semantic_grep
Source: https://github.com/bkimminich/juice-shop

**Expected behavior**
- If there are problematic files we should respect the semgrep-core timeout flag and report a timeout instead of hanging indefinitely.
- We should also be able to match three.js in a reasonable amount of time so maybe we have a bug in the matching code

**Environment**
Exists on 0.16.0

brendongo:
From slack: @aryx is going to have semgrep-core respect the timeout flag in the matching step and return a timeout error as a stop gap while we look into the root cause

aryx:
We can probably close this one. At least it does not stall anymore.

DrewDennison:
@aryx see https://returntocorp.slack.com/archives/C6A1GN4P3/p1603390422139900 (this is a regression)

mschwager:
Related #1981?

DrewDennison:
Fixed by https://github.com/returntocorp/semgrep/pull/2045. @DrewDennison to test

PR #1379, opened 2020-07-28
Better Timeout error indicating the ruleid of the offending rule

aryx:
This will help diagnose more quickly issues like
https://github.com/returntocorp/semgrep/issues/1361

Test plan:
you need some modifs to core_runner.py to not capture stderr (should
be fixed hopefully soon), and then you can do:

(semgrep) pad@yrax:~/semgrep/tests/PERF$ export SEMGREP_CORE_DEBUG=1
(semgrep) pad@yrax:~/semgrep/tests/PERF$ semgrep -v -f ajin2.yaml three.js
loaded 1 configs in 0.020273208618164062
running 2 rules from 1 config ajin2.yaml
running 2 rules...
Running rule node_nosqli_js_injection
Debug mode On
Executed as: semgrep-core -lang javascript -rules_file /tmp/tmp1aheu78p -j 8 -target_file /tmp/tmp8qu5slnl -use_parsing_cache /tmp/tmpy3k0oivh
Profile mode On
disabling -j when in profiling mode
Parsing /tmp/tmp1aheu78p
Analyzing three.js
PARSING: three.js
timeout (we abort)
returned JSON: {
  "matches": [],
  "errors": [
    {
      "check_id": "Timeout",
      "path": "three.js",
      "start": { "line": 1, "col": 1 },
      "end": { "line": 1, "col": 1 },
      "extra": {
        "message": "Timeout: with ruleid: 0..0.0.10",
        "line": "/**"
      }
    }
  ],

### DIFF
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -580,10 +580,25 @@ let iter_generic_ast_of_files_and_get_matches_and_exn_to_errors f files =
             *)
             |> (fun v -> (if !test then Gc.full_major()); v)
          ))
-       (* note that Error_code.exn_to_error now recognized Timeout
-        * and will generate a TimeoutError code for it
+       with
+       (* note that Error_code.exn_to_error already handles Timeout
+        * and would generate a TimeoutError code for it, but we intercept
+        * Timeout here to give a better diagnostic.
         *)
-       with exn -> [], [Error_code.exn_to_error file exn]
+        | (Timeout | Out_of_memory) as exn ->
+            let str_opt =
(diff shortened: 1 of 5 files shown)


aryx:
Note that right now all those Timeout, OutOfMemory errors are converted to just "parse error" by semgrep-python.
@brendongo we may want to give better error message for those? Especially the timeout now has the ruleid in its message
(but it's the ruleid of the semgrep_core yaml file, e.g. 1.0.0.12, so you need to revert back to the original rule name).

brendongo (approved):

PR #1382, opened 2020-07-28
Go package bug

colleend:
Tests to ensure that change to go parser in pfff works.


semgrep-core/tests/go/package2.sgrep
```diff
@@ -0,0 +1 @@
+package $X
```
aryx, line 1: wrong PR, but yes you can call those tests misc_package2.sgrep. We try to follow some conventions in the test filenames (which are then used to generate the cheatsheets), so for those features, just put a misc_prefix.

PR #1384, opened 2020-07-28
Match arbitrary content with f"..."

### LINKED ISSUE #979: Python f"..." fails to match f-strings with interpolated expressions
(Follow up to #654.)

Currently (https://semgrep.live/Rey8),

* `f"..."` matches `f"zero"` but not `f"one {a}"`, `f"two {a} {b}"`,
* `f"...{...}..."` matches `f"one"` but not `f"zero"`, `f"two {a} {b}"`,
* `f"...{...}...{...}..."` matches `f"two {a} {b}"` but not `f"zero"`, `f"one {a}"`,

and so on. There still seems to be no way to write an pattern that matches all f-strings; you have to specify the exact number of interpolated expressions.

I would expect

* `f"..."` to match `f"zero"`, `f"one {a}"`, `f"two {a} {b}"`,
* `f"...{...}..."` to match `f"one {a}"`, `f"two {a} {b}"` but not `f"zero"`,
* `f"...{...}...{...}..."` to match `f"two {a} {b}"` but not `f"zero"`, `f"one {a}"`,

and so on.

nbrahms:
This will now match any arbitrary inclusion of interpolated string
content.

It also sort-of implements some additional use cases, as trailing
ellipses in patterns like f"...{$X}..." can now match arbitrary content
as well.

Definitely not perfect, and some thoughts about what the appropriate API
is here are in order.

Fixes returntocorp/semgrep#979.

### DIFF
--- a/semgrep-core/matching/Generic_vs_generic.ml
+++ b/semgrep-core/matching/Generic_vs_generic.ml
@@ -1065,11 +1065,13 @@ and m_arguments_concat a b =
   match a,b with
   | [], [] ->
       return ()
+
   (*s: [[Generic_vs_generic.m_arguments_concat()]] ellipsis cases *)
-  (* dots '...' for string literal, can also match no argument *)
-  | [A.Arg (A.L (A.String("...", _a)))], [] ->
+  (* dots '...' for string literal, can match any number of arguments *)
+  | [A.Arg (A.L (A.String("...", _)))], _xsb ->
       return ()
 
+  (* specific case: f"...{$X}..." will properly extract $X from f"foo {bar} baz" *)
   | A.Arg (A.L (A.String("...", a)))::xsa, B.Arg(bexpr)::xsb ->
     (match Normalize_generic.constant_propagation_and_evaluate_literal bexpr
      with
@@ -1081,6 +1083,7 @@ and m_arguments_concat a b =
       | None ->
         (m_arguments_concat xsa (B.Arg(bexpr)::xsb))
       )
+
   (*e: [[Generic_vs_generic.m_arguments_concat()]] ellipsis cases *)
   (* the general case *)
   | xa::aas, xb::bbs ->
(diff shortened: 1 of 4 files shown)


nbrahms:
https://github.com/returntocorp/semgrep-rules/pull/746 necessary to make tests pass

mjambon:
Whatever syntax is chosen, I would recommend keeping it in "alpha" and subject to change for a while, since string matching in itself is a whole can of worms, and we're doing this inside a semgrep pattern which makes matters even more delicate. There's always a compromise between having an easy syntax and leaving enough room for future extensions.

So... I don't have a great sense of how specific a useful string pattern would have to be, which isn't very helpful.

There's one thing that bothers me in the current notation (if I understand it correctly). It's that there's no distinction between the first occurrence of a captured object and a backreference. For example:
```
function $X() {  // "capture" X
  ...
  $X();  // "backreference" to X
}
```
The above is fine when dealing with identifiers which are not substrings of anything, but what would the following mean:
```
var msg = "...{$X}...";
var msg2 = "...{$X}...";
```
What does the first `$X` capture? Is it a syntax error?

aryx:
Not sure what you mean Martin. Can you write a semgrep.live example that matches and that bothers you?

aryx:
Not sure what's the right syntax either, but I agree that the tests should pass (after all it was marked as a TODO) so
let's accept this.

aryx (approved):

nbrahms:
FWIW, I understand @mjambon 's comments to refer to the (still) broken test case in semgrep-core/tests/python/equivalence_f_string_3.py. Since it's ambiguous what content a metavariable will capture, it can be ambiguous what enforcing metavariable equality will do in such a situation.

@mjambon I agree that it makes sense for us to think of this as still an incomplete feature. For the purposes of this PR I'm trying to follow the Pareto principle and get us closer to being useful for a wide swath of use cases. We'll still have broken edge cases like this one, though.

mjambon:
@nbrahms : @aryx explained to me that `"...{$X}..."` is to match an interpolated string, i.e. the curly braces are a feature of the target language, not of the semgrep pattern language. I thought it meant "X matches any substring of a simple string constant".

Issue #1391, opened 2020-07-29
[Go] Wrong generic AST for short var decl


aryx:
a,b := 1,2  is actually parsed as   b,a := 1,2

See https://semgrep.live/Qkvx/?version=develop

This was reported here https://github.com/returntocorp/semgrep/issues/1300#issuecomment-663670111

PR #1393, opened 2020-07-29
Tests showing fix to wrong Go matching

aryx:
This closes issue #1391

Test plan:
test file included


aryx:
The CI failures are because of a todoruleid in semgrep-rules. I'm forcing push this and I'll fix semgrep-rules after.

Issue #1297, opened 2020-07-16
Semgrep 0.15.0 performance issues


ajinabraham:
**Describe the bug**
```
$  time semgrep -f ~/Code/njsscan/njsscan/rules/semantic_grep/ ~/Downloads/juice-shop-master/
running 105 rules...
 10%|███████████████▍                                                                                                                                   |11/105
```

More than 15 minutes and still waiting.

Rules: https://github.com/ajinabraham/njsscan/tree/master/njsscan/rules/semantic_grep
Source: https://github.com/bkimminich/juice-shop

## Enhancement 

It would be nice to add an end to end test suite with semgrep that runs before release updates on real apps to catch performance bugs.

**Environment**
```
$ semgrep --version
0.15.0
```

ajinabraham:
Also is there a way to identify rules that are slow?

ajinabraham:
It took ~1 minute to scan single file with 105 rules
```
time semgrep -f ~/Code/njsscan/njsscan/rules/semantic_grep/ ~/Downloads/dvna-master/server.js
running 105 rules...
 10%|███████████████▍
...SNIPPED...
real	0m53.124s
user	0m12.414s
sys	0m8.551s
```

minusworld:
> Also is there a way to identify rules that are slow?

Hey @ajinabraham, we are actually working on gathering performance stats for each rule right now. We also need to know which rules are fast and slow. This'll probably be done tomorrow or Monday and we'll be able to check out runtimes on all the rules together.

minusworld:
@ajinabraham From the JS rules, it looks like `audit.puppeteer-injection` eats up a LOT of time. You can check out the runtimes of all rules in this chart:

![semgrep-rules-public-repos-benchmarks](https://user-images.githubusercontent.com/2374948/87871974-de818280-c969-11ea-910a-86f1a4fa2e8d.png)

ajinabraham:
@minusworld Nice!
Can I run this benchmark locally? Also, can we know which file contributes to the most time?

minusworld:
@ajinabraham 
The benchmarks are now set to run on our [semgrep-rules](https://github.com/returntocorp/semgrep-rules) repo. It uses `pytest` and `pytest-benchmark`, so you should be able to run the benchmarks locally.

You can pass in rule-directory and a git-repo.

Something like this should work:
```
[...]

PR #1395, opened 2020-07-29
Filter rules not containing any specific strings contained in the source

aryx:
It's useless to run the matching engine on a file not containing
the 'eval' string if your rule mentions 'eval'.

This seems to improve the speed on semgrep on Zulip and Ajin
"benchmarks".

Test plan:
make test still works!

(semgrep) pad@yrax:~/work/semgrep/zulip$
(semgrep) pad@yrax:~/work/semgrep/zulip$ time semgrep --dangerously-allow-arbitrary-code-execution-from-rules --config tools/semgrep.yml .
running 9 rules...
100%|                                                                                                                                                    |9/9

real	0m5.479s
user	0m24.776s
sys	0m3.019s

(semgrep) pad@yrax:~/work/semgrep/zulip$ export SEMGREP_CORE_EXTRA=''
(semgrep) pad@yrax:~/work/semgrep/zulip$ time semgrep --dangerously-allow-arbitrary-code-execution-from-rules --config tools/semgrep.yml .
running 9 rules...
100%|                                                                                                                                                    |9/9

real	0m4.415s
user	0m16.017s
sys	0m3.082s

On juice-shop with Ajin rules I got from 8m52 to 7m52, and more importantly
got from 50 Timeout errors to just 18 or so.

### DIFF
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -975,6 +975,8 @@ let options () =
     (*e: [[Main_semgrep_core.options]] other cases *)
     "-use_parsing_cache", Arg.Set_string use_parsing_cache,
     " <dir> save and use parsed ASTs in a cache at given directory. Caller responsiblity to clear cache";
+    "-no_filter_rules_regexp", Arg.Clear Flag.filter_rules_regexp,
+    " do not filter rules using regexp optimization";
     "-debug", Arg.Set Flag.debug,
     " add debugging information in the output (tracing)";
     "-debug_matching", Arg.Set Flag.debug_matching,
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -975,8 +975,10 @@ let options () =
     (*e: [[Main_semgrep_core.options]] other cases *)
     "-use_parsing_cache", Arg.Set_string use_parsing_cache,
     " <dir> save and use parsed ASTs in a cache at given directory. Caller responsiblity to clear cache";
-    "-no_filter_rules_regexp", Arg.Clear Flag.filter_rules_regexp,
-    " do not filter rules using regexp optimization";
+    "-filter_irrelevant_rules", Arg.Set Flag.filter_irrelevant_rules,
+    " filter rules not containing any strings in target file";
+    "-no_filter_irrelevant_rules", Arg.Clear Flag.filter_irrelevant_rules,
+    " do not filter rules";
     "-debug", Arg.Set Flag.debug,
     " add debugging information in the output (tracing)";
     "-debug_matching", Arg.Set Flag.debug_matching,
--- a/semgrep-core/matching/Rules_filter.ml
+++ b/semgrep-core/matching/Rules_filter.ml
@@ -1,0 +1,155 @@
+(* Yoann Padioleau
+ *
+ * Copyright (C) 2020 r2c
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * version 2.1 as published by the Free Software Foundation, with the
+ * special exception on linking described in file license.txt.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the file
+ * license.txt for more details.
+ *)
+open Common
+module Flag = Flag_semgrep
+module R = Rule
+module V = Visitor_AST
+open AST_generic
+
+(*****************************************************************************)
+(* Prelude *)
+(*****************************************************************************)
+(* Rules filtering using regexps.
+ *
+ * It is useless to run the engine on rules containing specific IDs
+ * (e.g., eval) that are never mentioned in the file.
+ *
+ * We did something similar in Coccinelle I think. This also has been
+ * mentioned many times (by Clint, HN, etc.).
+ *
+ * notes: I tried to use the ocaml-re (Re) regexp libray instead of Str
+ * because I thought it would be faster, and because it offers regexp
+ * combinators (alt, rep, etc.) which might be useful at some point to
+ * handle patterns containing explicit DisjExpr. However when running
+ * on Zulip codebase with zulip semgrep rules, Str is actually faster
+ * than Re.
+ *
+ * todo:
+ *  - we could even avoid parsing the file by using a lazy AST
+ *    in Semgrep_generic.check.
+ *)
+
+(*****************************************************************************)
+(* Helpers  *)
+(*****************************************************************************)
+
+(* Those functions using Re instead of Str are actually slower on
+ * Zulip.
+let regexp_matching_str s =
+  Re.str s
+let compile_regexp t =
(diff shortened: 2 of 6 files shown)


aryx:
This can help a bit https://github.com/returntocorp/semgrep/issues/1297

aryx:
Goes down form 34s with lots of timeout errors on dvna with njsscan rules to 30s with 13 timeout errors.

semgrep-core/matching/Rules_filter.ml
```diff
@@ -0,0 +1,155 @@
...
+ * (e.g., eval) that are never mentioned in the file.
+ *
+ * We did something similar in Coccinelle I think. This also has been
+ * mentioned many times (by Clint, HN, etc.).
+ *
+ * notes: I tried to use the ocaml-re (Re) regexp libray instead of Str
+ * because I thought it would be faster, and because it offers regexp
+ * combinators (alt, rep, etc.) which might be useful at some point to
+ * handle patterns containing explicit DisjExpr. However when running
+ * on Zulip codebase with zulip semgrep rules, Str is actually faster
```
nbrahms, line 36: 😛

aryx:
@mjambon did you have similar experience with Re vs Str? I might have used it wrong.

semgrep-core/matching/Rules_filter.ml
```diff
@@ -0,0 +1,155 @@
...
+  } in
+  visitor any;
+  !res
+
+(*****************************************************************************)
+(* Entry point *)
+(*****************************************************************************)
+
+let filter_rules_relevant_to_file_using_regexp2 rules lang file =
+  let str = Common.read_file file in
```
nbrahms, line 123: Have you looked at the implications of running this on, e.g., a large minified js file? It looks like we need to load the entire file into a single block of memory here (not sure how the internals of ocaml Str work, but I assume it wraps a contiguous character array).

It might be worth wrapping this in a max memory guard.

Presumably a streaming solution is even better, but probably an annoying implementation.

aryx: if we're slow with read_file, it will be far slower with parsing it and matching it.

semgrep-core/matching/Rules_filter.ml
```diff
@@ -0,0 +1,155 @@
...
+let reserved_str str =
+  str = "..." ||
+  Matching_generic.is_regexp_string str
+
+(*****************************************************************************)
+(* ID extractor  *)
+(*****************************************************************************)
+let extract_specific_strings lang any =
+  let res = ref [] in
+  let visitor = V.mk_visitor {V.default_visitor with
```
nbrahms, line 95: Is this correctly exhaustive?

I'm wondering what will happen with file and package names, e.g.:

```go
import "fmt"
```

```java
package foo;
```

nbrahms: Reflecting on ^^, I think we need a mechanism to log which rules are rejected for each file. Can we print this to stderr in verbose / debug mode?

semgrep-core/matching/Rules_filter.ml
```diff
@@ -0,0 +1,155 @@
...
+       *)
+      xs |> List.for_all (fun x ->
+         let t = regexp_matching_str x in
+         let re = compile_regexp t in
+         run_regexp re str
+    )
+
+    in
+    if !Flag.debug && not match_
+    then pr2 (spf "filtering rule %s" rule.R.id);
```
nbrahms, line 149: ... and it's here 😛 . Can we add the filename to the debug message?

semgrep-core/matching/dune
```diff
@@ -7,6 +7,7 @@
    pfff-config
    pfff-h_program-lang
    pfff-lang_GENERIC pfff-lang_GENERIC-analyze
+   pfff-lang_js pfff-lang_js-analyze
```
nbrahms, line 10: I assume this shouldn't be added to this PR?

semgrep-core/bin/Main.ml
```diff
@@ -975,6 +975,8 @@ let options () =
     (*e: [[Main_semgrep_core.options]] other cases *)
     "-use_parsing_cache", Arg.Set_string use_parsing_cache,
     " <dir> save and use parsed ASTs in a cache at given directory. Caller responsiblity to clear cache";
+    "-no_filter_rules_regexp", Arg.Clear Flag.filter_rules_regexp,
```
nbrahms: Can we invert this flag?

IMO this performance optimization could result in a surprising behavior, and I can anticipate it being quite difficult for users to debug.

Since most users are likely to not encounter a serious performance issue, I'd rather default filtering to `off`, then recommend turning this on in our performance troubleshooting documentation.

cc. @ievans @dlukeomalley for additional thoughts.

aryx:
Actually there is not even need for a regexp. But String might be slower than Str no @mjambon ?

semgrep-core/matching/Rules_filter.ml
aryx: Yes use -debug

aryx: It was not exhaustive first, but after a few iterations i got make test and semgreprule test working

semgrep-core/matching/Rules_filter.ml
aryx: It is already display before as PARSING foo.js

semgrep-core/matching/dune
aryx: Actually i need this because of the ugly reference to Astjs.special

semgrep-core/matching/Rules_filter.ml
mjambon: I haven't used Re. Based on the [project description](https://github.com/ocaml/ocaml-re), it's faster than Pcre and Str only on large regexps. This is presumably due to using an algorithm with polynomial rather than exponential asymptotic complexity, but with a greater startup cost due to the construction of an automaton.

mjambon: I was reading this article on the subject the other day: https://swtch.com/~rsc/regexp/regexp1.html
It's interesting to grasp how the path-merging algorithm works and why it has no problem handling regular expressions made of many nodes.

semgrep-core/matching/Rules_filter.ml
nbrahms: Isn't parsing guarded by max_memory though? Or does that also apply to this load?

semgrep-core/matching/Rules_filter.ml
nbrahms: 👍

semgrep-core/matching/dune
nbrahms: "neat"

semgrep-core/bin/Main.ml
aryx: ok, I'll invert the flag for now.

aryx:
Yep, the String module does not really have the right function and from stack overflow it looks like poeple suggest to use Str.

semgrep-core/matching/Rules_filter.ml
aryx: The whole parsing and matching is guarded by max_memory. But again, if we exceed memory with read_file, there's no way we can generate the AST for this file, which uses far more memory.

PR #1398, opened 2020-07-29
Clara/python file

chmccreery:


semgrep-core/tests/generate_test_matrix.py
```diff
@@ -12,14 +12,17 @@
...
 
 FEATURES = ["dots", "equivalence", "metavar", "misc"]
+
+VERBOSE_REGEXP_SYNTAX = "OCaml Syntax: \"=~/<regexp>/\""
+
 VERBOSE_FEATURE_NAME = {
-    "dots": "Ellipsis",
-    "equivalence": "Equivalences",
-    "metavar": "Metavariables",
+    "dots": "Wildcard Matches (...)",
```
mschwager, line 19: I'm glad we're improving these! The naming was a bit confusing/inconsistent. Although, we should keep these names consistent with how we describe them in the docs: https://github.com/returntocorp/semgrep/blob/develop/docs/pattern-features.md. Looks like that covers "Ellipsis Operator", "Equivalences", "Metavariables", and "Deep Expression Operator."

chmccreery: @mschwager yes, I agree that consistency is important- thanks for bringing this up! I am happy to update the docs to match these, too, if you think that makes sense. What I was trying to do here was not so much re-name the operators (ellipsis are ellipsis after all) but instead re-frame the cheatsheet's headings as "the problem the user is trying to solve", rather than "the operator they need", which they may not know the name of yet

chmccreery:
I guess I forgot to request a review on this and actually merge it in last week

semgrep-core/tests/generate_test_matrix.py
```diff
@@ -157,26 +159,28 @@ def generate_cheatsheet(root_dir: str):
...
-                     'highlights': higlights,
+                    'highlights': highlights,
                 }
                
 
                 #print((lang, entry))
                 feature_name = VERBOSE_FEATURE_NAME.get(category, category)
                 subcategory_name = VERBOSE_SUBCATEGORY_NAME.get(subcategory, subcategory)
+                if (category == "regexp" and subcategory == "string"):
+                    subcategory_name = VERBOSE_REGEXP_SYNTAX
```
ievans, line 183: is there a better way to do this?

ievans (approved):
thanks for improving this so much!

semgrep-core/tests/generate_test_matrix.py
chmccreery: Only other way I could think of was to re-name the actual test files, because currently this subcategory mapping is based on those names, but that felt kind of hacky too, and I wasn't sure if other things depended on those names

PR #1400, opened 2020-07-29
Update CONTRIBUTING.md

sabrinabrogren:


brendongo:
Does `pipenv install --dev` not already handle this?

sabrinabrogren:
> Does `pipenv install --dev` not already handle this?

Unless I did it incorrectly, it didn't for me. Side note: I also don't seem to have pytest, even after running `pip install -e .`. Not sure if I messed up somewhere. Though I did run through the steps several times to make sure I did them correctly.

brendongo (approved):

sabrinabrogren:
Making this a draft for now because this may be something that's just necessary on my machine...

sabrinabrogren:
Ah, so the reason things weren't working for me is because I was running in `semgrep` not `semgrep/semgrep`. So this change is not necessary, but I'm going to make another PR with a short clarification, since I'm not the only one who's made this mistake.

PR #1401, opened 2020-07-29
Add python backed pre-commit hook

brendongo:
Running with docker leads to slowness due to filesystem issues with
OSX. Since we have a semgrep published in pypi might as well use it.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,8 +4,16 @@
   language: docker_image
   entry: returntocorp/semgrep:develop
 
-- id: semgrep
+- id: semgrep-docker
   name: semgrep
   description: This hook runs semgrep
   language: docker_image
   entry: returntocorp/semgrep:latest
+
+- id: semgrep
+  name: semgrep
+  entry: semgrep
+  language: python
+  args: []
+  require_serial: false
+  additional_dependencies: []
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,4 +1,4 @@
-- id: semgrep-develop
+- id: semgrep-docker-develop
   name: semgrep
   description: This hook runs semgrep:develop
   language: docker_image
@@ -14,6 +14,3 @@
   name: semgrep
   entry: semgrep
   language: python
-  args: []
-  require_serial: false
-  additional_dependencies: []
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@
 from setuptools import setup
 
 setup(
-    name="pre_commit_dummy_package",
+    name="semgrep_pre_commit_package",
     version="0.0.0",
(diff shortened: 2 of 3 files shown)


.pre-commit-hooks.yaml
```diff
@@ -4,8 +4,16 @@
...
   entry: returntocorp/semgrep:develop
 
-- id: semgrep
+- id: semgrep-docker
   name: semgrep
   description: This hook runs semgrep
   language: docker_image
   entry: returntocorp/semgrep:latest
+
+- id: semgrep
```
mschwager, line 13: Will we be breaking backwards compatibility at all for existing pre-commit configurations? I don't _think_ so. Since the Docker image uses the Python package and they present the same interface I think we're fine here. Thoughts?

If there's no backwards compatibility issues, and we go with `semgrep-docker`, we should update the first one to `semgrep-docker-develop`. Although we'll also have to update a few of our internal pre-commit configurations. Maybe it's not worth it, what do you think?

setup.py
```diff
@@ -0,0 +1,10 @@
+# type: ignore
+# Used for pre-commit since it expects a setup.py in repo root
+# for actual setup.py see semgrep/setup.py
+from setuptools import setup
+
+setup(
+    name="pre_commit_dummy_package",
```
mschwager: Should we instead use something like `semgrep_pre_commit_package`? That way if we ever see this string anywhere it'll give us a hint as to where it's coming from. We may also want to grab this PyPI package name just so there's no funny business ever going on with that.

brendongo: Went with your suggestion and parked the name

.pre-commit-hooks.yaml
brendongo: There should not be any issues since when you specify a pre-commit hook you specify the git tag the repo is checked out at so any existing users that use the id semgrep will still actually use the docker container one.

brendongo:
Closes: https://github.com/returntocorp/semgrep/issues/1200

.pre-commit-hooks.yaml
```diff
@@ -4,8 +4,16 @@
...
+  name: semgrep
+  entry: semgrep
+  language: python
+  args: []
```
mschwager: It looks like these are the default values for `args`, `require_serial`, and `additional_dependencies`, so we shouldn't need to re-specify them here.

.pre-commit-hooks.yaml
```diff
@@ -4,8 +4,16 @@
   language: docker_image
   entry: returntocorp/semgrep:develop
 
-- id: semgrep
+- id: semgrep-docker
```
mschwager, line 7: Not sure why Github won't allow me to comment on the unchanged line above... Should we move `semgrep-develop` -> `semgrep-docker-develop` above to stay consistent here?

mschwager (approved):

PR #1413, opened 2020-07-30
Implemented for loops

emjin:
Added pretty-printing for for loops (tested on Python, Java), which necessitated adding basic pattern cases, and added patterns to abstract away dots in the condition or expression


brendongo:
🔥

aryx (approved):

PR #1367, opened 2020-07-27
Use -max_memory, not -timeout, and fix comments

aryx:
This addresses the comments in https://github.com/returntocorp/semgrep/pull/1186

test plan:
pad@yrax:~/semgrep/tests/PERF$ yy -test -lang js -max_memory 100 -rules_file ../../data/basic.yml l10000.js
+ /home/pad/github/semgrep/semgrep-core/_build/default/bin/Main.exe -test -lang js -max_memory 100 -rules_file ../../data/basic.yml l10000.js
{ "matches": [], "errors": [], "stats": { "okfiles": 1, "errorfiles": 0 } }
pad@yrax:~/semgrep/tests/PERF$ yy -test -lang js -max_memory 10 -rules_file ../../data/basic.yml l10000.js
+ /home/pad/github/semgrep/semgrep-core/_build/default/bin/Main.exe -test -lang js -max_memory 10 -rules_file ../../data/basic.yml l10000.js
{
  "matches": [],
  "errors": [
    {
      "check_id": "FatalError",
      "path": "l10000.js",
      "start": { "line": 1, "col": 1 },
      "end": { "line": 1, "col": 1 },
      "extra": { "message": "Fatal Error: Out of memory", "line": "/**" }
    }
  ],
  "stats": { "okfiles": 0, "errorfiles": 1 }
}

PR #1416, opened 2020-07-30
added --max-memory flag

sabrinabrogren:
Initial changes, haven't tested yet


sabrinabrogren:
Tried to test by running the same tests here but in the CLI: https://github.com/returntocorp/semgrep/pull/1367 but I didn't get any errors. The command I was running is `python -m semgrep --config ../semgrep-core/data/basic.yml --max-memory 1 --lang js ../semgrep-core/tests/PERF/l10000.js`

sabrinabrogren:
@aryx The -max_memory flag in semgrep-core appears to be hidden behind -test currently (see https://github.com/returntocorp/semgrep/pull/1367/files#diff-775b9086b9f1a4b7c4c824ee62e8ae41R582). Is it going to stay that way? The changes in this PR were made assuming that wouldn't be the case.

brendongo:
@sabihb you can include the change in this PR moving the -test out of the ocaml code.

aryx:
You don't need to call semgrep-core with -test.
The -test is is just a flag to trigger certain conditions that makes it easier to test features.
Here, if you use -max_memory with a low number (e.g., 1MB), you might not get a Out_of_memory error
that we want to generate because we rely on an alarm in the GC (Garbage collector), but if you use semgrep on small problems, you will not even trigger the GC. The -test flag allows to explicitely calls the Gargave collector, which then gives a chance to run the alarm and raise the Out_of_memory error.

sabrinabrogren:
This started to get messy, so I moved everything over here: https://github.com/returntocorp/semgrep/pull/1426

PR #1421, opened 2020-07-31
Match "OtherAttribute" attributes in any order

### LINKED ISSUE #1406: Matching Python annotations containing "." characters
**Describe the bug**
When trying to match an annotation on a python method that contains a "." character the match fails

**To Reproduce**

In the following code snippet:

```
@ns.route('/store/<int:checklist_type>/<int:maturity>')
@api.response(404, 'Validation error', message)
class QuestionSprintStoreCollection(Resource):

    @api.expect(authorization, store_list_items)
    @api.marshal_with(message, 'Success')
    @api.response(400, 'No results found', message)
    def put(self, checklist_type, maturity):
        """
        Store list of question sprint items.
```

This pattern will match the class annotator:

```
- pattern: |
      @ns.route($ROUTE)
      class $C(...):
```

but a similar pattern to match the class annotator "@api.response" fails:

```
- pattern: |
        @api.response(...)
        def $METH(...):
          ...
```

If I remove the . in the api.response string I can write a match pattern that works.

**Expected behavior**
Pattern should match the annotation

**Environment**
docker and semgrep.live

nbrahms:
We already have code to match keyword and named attributes in any order
in an attribute list, and to match with "less is ok" functionality.

This commit extends that to work with "OtherAttribute" AST nodes as
well. Note that it is not fully implemented here, as I took a shortcut
that would leave this pattern:

```python
@x("a")
@x("a")
def foo:
  ...
```

still matching this code:

```python
@x("a")
def foo:
  pass
```

I'm leaving this as a TODO for now.

Fixes #1406.

### DIFF
--- a/semgrep-core/matching/Generic_vs_generic.ml
+++ b/semgrep-core/matching/Generic_vs_generic.ml
@@ -1298,13 +1298,25 @@ and m_list__m_attribute (xsa: A.attribute list) (xsb: A.attribute list) =
       with Not_found -> fail ()
       )
   (*e: [[Generic_vs_generic.m_list__m_attribute]] [[NamedAttr]] pattern case *)
-  (* the general case *)
-  | xa::aas, xb::bbs ->
-      m_attribute xa xb >>= (fun () ->
-      m_list__m_attribute aas bbs
+  (*s: [[Generic_vs_generic.m_list__m_attribute]] [[OtherAttribute]] pattern case *)
+  (* Matches language-specific attributes
+   * In general, we assume we want to match these with less-is-ok and in any order.
+   * As a motivating example, see tests/python/misc_attributes.sgrep.
+   * The algorithm here is a super-cheap approach to accomplish this, but leaves behind the
+   * matched node from the right-hand-side, which means that node will be available for matching
+   * of other attributes. IMO this is a pretty rare edge case (essentially it only appears if the
+   * left-hand side declares the same attribute twice), so I'm leaving it as a TODO here.
+   * In the future, we should create a matching helper that delegates continued matching, using the
+   * unmatched remnants of the right-hand list. Creating such a helper would also help us DRY the
+   * KeywordAttr and NamedAttr matching code above.
+   * -- nbrahms July 2020
+   *)
+  | ((A.OtherAttribute _) as a)::xsa, xsb ->
+      m_list_in_any_order ~less_is_ok:true m_attribute [a] xsb >>= (fun () ->
+        (* TODO: remove b match from rhs *)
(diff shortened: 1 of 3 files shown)


semgrep-core/matching/Generic_vs_generic.ml
```diff
@@ -1298,13 +1298,25 @@ and m_list__m_attribute (xsa: A.attribute list) (xsb: A.attribute list) =
       with Not_found -> fail ()
       )
   (*e: [[Generic_vs_generic.m_list__m_attribute]] [[NamedAttr]] pattern case *)
-  (* the general case *)
```
nbrahms: Note that "the general case" is actually just the specific "OtherAttribute" case

mschwager (approved):
Noice

mjambon:
I don't know how this works but I appreciate the brevity of the code change.

nbrahms:
Needs https://github.com/returntocorp/semgrep-rules/pull/758 merged first

aryx:
This might work but might be slower than the code in in m_list__m_attribute which goes more directly
to the matching keyword or named attribute. It tries to avoid the more costly all_elem_and_rest_of_list.

aryx:
Maybe in practice it does not make much difference, and using the naive m_list_in_any_order is not that much slower.

colleend (approved):

nbrahms:
Timed on https://github.com/returntocorp/semgrep/runs/941906725; no noticeable difference in performance

aryx:
we might want to delete the m_list__m_attribute() function then. It should be dead now (not sure why ocaml
didn't warn about it).

aryx:
ah I see, because there are more places where we compare attributes.

semgrep-core/matching/Generic_vs_generic.ml
```diff
@@ -1846,7 +1846,7 @@ and m_entity a b =
   { A. name = a1; attrs = a2; tparams = a4; info = a5 },
   { B. name = b1; attrs = b2; tparams = b4; info = b5 } ->
     m_ident_and_id_info_add_in_env_Expr (a1, a5) (b1, b5) >>= (fun () ->
-    (m_list__m_attribute) a2 b2 >>= (fun () ->
+    (m_list_in_any_order ~less_is_ok:true m_attribute a2 b2) >>= (fun () ->
```
aryx, line 1849: better to remove the body of m_list__m_attribute and call 
m_list_in_any_order ... in it that way all the places where we look for attributes will have this new behavior.
(need to check the possible perf regression when applied to all the instances).

PR #1422, opened 2020-07-31
Debug logs were being hidden even with --verbose flag

### LINKED ISSUE #1365: SEMGREP_CORE_DEBUG=1 does not work anymore with semgrep
as explained in CONTRIBUTING.md, one way to debug semgrep-core from semgrep is to run
export SEMGREP_CORE_DEBUG=1 and if you do:
pipenv run semgrep -f ../semgrep-core/tests/PERF/ajin.yaml ../semgrep-core/tests/PERF/three.js

will get some debugging output from semgrep-core. There is also SEMGREP_CORE_PROFILE.
However this does not work anymore. My guess is that semgrep now redirect the stdout and stderr of semgrep
but this is really annoying to debug things for me.

Can we get a special option or have -v let stderr message from semgrep-core to reach the terminal?

brendongo:
To prevent logs being handled twice semgrep does not propagate logs
from semgrep module to root logging handler so calls to logging.info/debug
were not being handled properly. Really we should not be sending logs
to the root logger and should go through the logger of the submodule.

Fixed calls to go through the correct logger and added a semgrep
rule to alert when the root logger function is used.

Closes https://github.com/returntocorp/semgrep/issues/1365

### DIFF
--- a/semgrep-local.yaml
+++ b/semgrep-local.yaml
@@ -28,3 +28,16 @@ rules:
       https://www.attrs.org/en/stable/hashing.html for more info.
     languages: [python]
     severity: ERROR
+
+  - id: using-root-logger
+    patterns:
+      - pattern: logging.$METHOD(...)
+      - metavariable-regex:
+          metavariable: '$METHOD'
+          regex: '(debug|info|warn|warning|error)'
+    message: >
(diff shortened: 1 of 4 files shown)


aryx (approved):
Thxsemgrep! Nice rule

PR #1436, opened 2020-08-04
Tree-sitter CST to AST for Go

aryx:


aryx:
It achieves 100% parsing on my ~/work/lang-go/  benchmark (was at 99.76% before with the
pfff-based Go parser).

mjambon:
great

PR #1438, opened 2020-08-04
* semgrep-core/tests/go/misc_exprstmt_vs_expr.go: Test files

### LINKED ISSUE #1370: Fix Golang newline inconsistencies
Originally noticed in this comment: https://github.com/returntocorp/semgrep/pull/1357#issuecomment-663603212

In Go a newline introduces hidden semicolons. This ends up creating inconsistencies when patterns do or do not include a newline at the end of the pattern. This is especially noticeable when using a file vs. a CLI flag as the pattern input since files tend to include newlines whereas CLI arguments do not. Let's make this behavior more consistent by fixing up these inconsistencies :+1:

aryx:
This closes #1370

Test plan:
test file included
make test


aryx:
The test failures are in semgrep-rules. I will fix semgrep-rules in another diff.

Issue #1441, opened 2020-08-04
[JS] JSX semgrep patterns parse error


aryx:
https://semgrep.dev/8xxq

PR #1442, opened 2020-08-04
Test file showing support for JSX semgrep pattern using attr={expr}

aryx:
This closes issue #1441

Test plan:
test file included


aryx:
I'll fix the error in semgrep-rules in another diff.

PR #1443, opened 2020-08-04
Create Dockerfile-nonroot

### LINKED ISSUE #1437: Non root docker image
Hi

The docker image for semgrep runs as root. I think a new image should be released with a less privileged user

javixeneize:
Created non root dockerfile. Closes #1437

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,8 @@ RUN semgrep-core -version
 COPY semgrep /semgrep
 RUN HOMEBREW_SYSTEM='NOCORE' python -m pip install /semgrep
 RUN semgrep --version
-
+RUN adduser -D -u 1000 semgrep
+USER 1000
 ENV SEMGREP_IN_DOCKER=1
 ENV SEMGREP_VERSION_CACHE_PATH=/src/.cache/semgrep_version
 ENV PYTHONIOENCODING=utf8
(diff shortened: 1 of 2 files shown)


javixeneize:
Apparently the image can’t be pushed to docker hub. Can somebody have a look please?

brendongo:
Thanks for the PR @javixeneize. For security we don't push to dockerhub from PRs coming from a forked repo (changes will be pushed after the PR is merged) so don't worry about the failing docker test. 

@DrewDennison is there a reason we need the docker to run as root or can these changes be made to the main dockerfile instead of having a second dockerfile?

DrewDennison:
@brendongo I don't believe semgrep needs to be root. We can probably have a `semgrep` user inside the main docker file.

javixeneize:
I can amend it if needed. On top of that, I have done the change using as a base the most recent tag/hash in docker hub (It’s confusing, maybe you should use a more standardised naming for tags). The image with the latest tag is very different to this one (the binary is in /root/.local/bin) and it needs a different dockerfile, which was my initial commit. Is this the right image to modify?

Thanks

brendongo:
You probably want to make those changes here: https://github.com/returntocorp/semgrep/blob/develop/Dockerfile#L56
Also the hashes on dockerhub correspond to commits on Github. The official named tags are latest, develop, and semver versions (like 0.17.0)

javixeneize:
I have added the changes to the main Dockerfile and removed the extra file

Thanks

brendongo (approved):

brendongo:
Thanks @javixeneize!

PR #1450, opened 2020-08-04
Added "..." for string literals, modified block

### LINKED ISSUE #1372: Code-to-pattern should suggest `"..."` for string literals

emjin:
- String literals are generalized as "..." instead of $X
- If a block only has one statement, all possibilities for that statement are shown
- Fixed issue causing an infinite loop on top level of block

Note that "metavariable" is still being used as the tag when literals are replaced with "..."


aryx (approved):
What was the infinite loop? Can be useful to add comments like
(* bugfix: if you do X then you'll get an infinite loop *)

Issue #1397, opened 2020-07-29
Add Ruby repositories to QA testing


mschwager:
In an effort to gain more confidence in our Ruby support we should include some Ruby repositories in our QA testing process. This will confirm our parsing support, Ruby rules in semgrep-rules, and prevent regressions.

Related Ruby efforts:

- https://github.com/returntocorp/enterprise/issues/251
- https://github.com/returntocorp/enterprise/issues/270
- https://github.com/returntocorp/enterprise/issues/272
- https://github.com/returntocorp/semgrep-rules/issues/726
- https://github.com/returntocorp/semgrep/issues/1269

brendongo:
Repos used in QA testing are in https://github.com/returntocorp/semgrep/blob/develop/semgrep/tests/public_repos.py

https://github.com/returntocorp/semgrep/blob/develop/semgrep/tests/qa/test_public_repos.py will need to be modified to run ruby rules. Right now it only does python go and javascript

minusworld:
We also have benchmarking for Ruby rules on `semgrep-rules`.

https://github.com/returntocorp/semgrep-rules/blob/develop/tests/performance/.test_public_repos.yaml

mschwager:
Re-opening since we reverted this change due to failing tests: https://github.com/returntocorp/semgrep/runs/946554625

PR #1455, opened 2020-08-05
Refactor QA tests to allow passing specific language(s) per-repository

mschwager:


semgrep/tests/conftest.py
```diff
@@ -234,4 +234,4 @@ def pytest_generate_tests(metafunc):
         if metafunc.config.getoption("--qa"):
             metafunc.parametrize("public_repo_url", public_repos.ALL_REPOS)
         else:
-            metafunc.parametrize("public_repo_url", public_repos.PASSING_REPOS[:5])
+            metafunc.parametrize("public_repo_url", public_repos.PASSING_REPOS)
```
mschwager: This is here so I can test these changes on all the repos. I'll remove before pushing.

mschwager:
The goal here is to refactor this code to accomplish https://github.com/returntocorp/semgrep/issues/1397. The tests failed the original PR due to non-Ruby failures. The goal here is to re-add those tests, but only run the Ruby language against the repositories.

brendongo (approved):

PR #1458, opened 2020-08-05
timeout threshold flag

### LINKED ISSUE #1431: Stop running rules on file that keeps timing out
With `--timeout` flag it is still possible for a file to cause a timeout on every single rule.

Make default behavior to be to stop trying to scan a file after X number of rule timeouts. Add a flag that modifies the value of X.

This will significantly speed up run times when running on a codebase with a problematic file.

sabrinabrogren:
Set `max-timeouts` to 3 for now, but open to suggestions on that (and everything else, including the flag name).


semgrep/tests/e2e/test_check.py
```diff
@@ -209,3 +209,29 @@ def test_max_memory(run_semgrep_in_tmp, snapshot):
...
     )
+
+
+def test_max_timeouts(run_semgrep_in_tmp, snapshot):
+    # Check that semgrep-core timeouts are properly handled
+
+    snapshot.assert_match(
+        run_semgrep_in_tmp(
+            "rules/long.yaml",
+            options=["--timeout", "1", "--max-timeouts", "1"],
```
brendongo: Probably worth adding a test that has max-timeouts > 1 and checking that it doesn't stop running rules until it hits max-timeouts. Can probably just create another rule file that duplicates the rule in long.yaml a bunch of times.

sabrinabrogren: Will do!

mschwager:
Had a more general question on Slack taking it back a step:

> Hmm, I wonder if it's even better to take a step back here. Would it make more sense to avoid adding another flag here and leverage something like `--exclude`? I.e. if we timeout more than once on the same file we print a message saying something like "Timing out on <filename>, consider excluding file from analysis with `--exclude`". The downside is they may have to restart their analysis, the upside seems to be avoiding a potentially confusing flag. As a `semgrep` user I'm not really sure when a file could timeout, and I don't really know a reasonable value to set here unless I need it to try more rules against a timing out file
>
> Or maybe the language is just a bit confusing to me. Something like `--timeout-retries` makes more sense to me, although that may imply you're retrying the same rule. Thoughts?

What do you think @brendongo?

semgrep/semgrep/core_runner.py
```diff
@@ -441,11 +457,19 @@ def _run_rules(
...
                 )
                 findings_by_rule[rule] = rule_matches
                 debugging_steps_by_rule[rule] = debugging_steps
                 all_errors.extend(errors)
+                for err in errors:
+                    if isinstance(err, MatchTimeoutError):
+                        file_timeouts[err.path] += 1
+                        if (
+                            self._timeout_threshold != 0
+                            and file_timeouts[err.path] == self._timeout_threshold
```
mschwager: We can be slightly more defensive here by using `>=`, which could prevent an infinite loop in the future.

semgrep/semgrep/output.py
```diff
@@ -243,9 +245,24 @@ def handle_semgrep_errors(self, errors: List[SemgrepError]) -> None:
...
         for path in errors.keys():
+            num_errs = len(errors[path])
+            errors[path].sort()
+            error_msg = f"Warning: {num_errs} timeout error(s) in {path} when running the following rules: [{separator.join(errors[path])}]"
+            if num_errs == self.settings.timeout_threshold:
+                error_msg += f"\nSemgrep stopped running rules on {path} after {num_errs} timeout error(s). See `--timeout-threshold` for more info."
+            print_threshold_hint = print_threshold_hint or (
+                num_errs > 5 and not self.settings.timeout_threshold
+            )
+            logger.error(with_color(colorama.Fore.RED, error_msg))
```
mschwager, line 258: Do we typically use `with_color` with our loggers? I see it's only currently used in `semgrep/error.py` and not in any `logger` calls. I'll defer to @brendongo here.

sabrinabrogren: I used it here because the messages were red before, so when aggregating them and creating a new error message I added `with-color` to make them look like the other not aggregated error messages.

brendongo: I think there are other errors that have color added as part of their __str__ function that get written to log so this is fine.

brendongo (approved):
Can you make the PR title more descriptive before merging

PR #1465, opened 2020-08-07
ci: shellcheck only run on .sh files

brendongo:
shellcheck was running on files with sh anywhere in the filename

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -83,7 +83,7 @@ repos:
         name: shellcheck
         language: docker_image
         entry: koalaman/shellcheck:v0.7.1
-        files: ".sh$"
+        files: "[.]sh$"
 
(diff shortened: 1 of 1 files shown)


underyx (approved):

.pre-commit-config.yaml
```diff
@@ -83,7 +83,7 @@ repos:
         name: shellcheck
         language: docker_image
         entry: koalaman/shellcheck:v0.7.1
-        files: ".sh"
+        files: ".sh$"
```
underyx:
```suggestion
        files: "\.sh$"
```

This would probably be even more accurate?

underyx:
```suggestion
        files: "[.]sh$"
```

or rather this, cause I'm not sure how to escape when there's YAML involved

PR #1468, opened 2020-08-07
Update pre-commit to use python package

brendongo:
Use 0.18.0 which uses pypi package in pre-commit instead of a docker container.
Add pre-commit config that uses docker container develop image in CI only


mschwager (approved):
Nice, I like the manual Docker image dogfooding

Issue #1471, opened 2020-08-07
Timeout still happens even if default is 0


TMFRook:
**Describe the bug**
When timeout is not provided, the documentation says it defaults to 0 (which means there is no timeout). However, it still timeouts.

**To Reproduce**
1. Run `semgrep --config=https://semgrep.dev/p/r2c-CI .`
2. Returns `Warning: Semgrep exceeded timeout when running contrib.nodejsscan.eval_yaml_deserialize.yaml_deserialize2 on api_manager/static/api_manager/libs/jquery_2.1.3/js/jquery.min.js. See `--timeout` for more info.`.
3. `semgrep -h` says: `--timeout TIMEOUT     Maximum time to spend running a rule on a single file in seconds. If set to 0 will not have time limit. Defaults to 0.`

**Expected behavior**
Should never timeout since that's what the doc says.

**Screenshots**
![image](https://user-images.githubusercontent.com/68116003/89695148-f8ffa980-d8e0-11ea-97d8-5d7475cd32ba.png)

![image](https://user-images.githubusercontent.com/68116003/89695205-1af92c00-d8e1-11ea-9189-6aaca9be15d1.png)

**Environment**
Running off official binary: 
![image](https://user-images.githubusercontent.com/68116003/89695103-d40b3680-d8e0-11ea-8608-2f919939c656.png)

brendongo:
Thanks for filling this issue @TMFRook! Looks like we didn't remove https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/bin/Main.ml#L349 which is setting a default timeout of 5 seconds for javascript files.

brendongo:
A side effect of removing the default timeout of 5 seconds in js will mean if a project has large minified javascript files semgrep will potentially take forever to run.

Thoughts on setting default timeout to 5 seconds @mschwager @TMFRook?

TMFRook:
I don't have an opinion on the default timeout setting. However, it is odd to me that it would choke on seemingly small-ish files. They're only ~82K big. Is this expected?

mschwager:
> A side effect of removing the default timeout of 5 seconds in js will mean if a project has large minified javascript files semgrep will potentially take forever to run.
> 
> Thoughts on setting default timeout to 5 seconds @mschwager @TMFRook?

[...]

PR #1472, opened 2020-08-07
Remove javascript default timeout of 5 sec in core

brendongo:
We now expose timeout in semgrep cli so let's respect that.
This does mean that if there are large minified js files in a project
we will no longer timeout on them and semgrep may potentially take
forever by default.

See discussion on https://github.com/returntocorp/semgrep/issues/1471

### DIFF
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -343,16 +343,9 @@ let cache_file_of_file filename =
  * with Timeout -> raise Timeout | _ -> ...
  *)
 let timeout_function lang = fun f ->
-  let timeout =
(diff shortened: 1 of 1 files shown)


aryx:
What we could do once we have the javascript tree-sitter parser is start with it, which should hopefully handle minified files.

brendongo:
Closing for now. Can reopen if moving to treesitter doesn't remove the need to special case JS

aryx:
Yes, still can't totally switch to tree-sitter. Right now we use tree-sitter only if the pfff parser. I tried to put tree-sitter first but we currently have regressions so we still need to wait a bit before this PR.

PR #1481, opened 2020-08-10
Match against destructured parameters with metavariables in js

### LINKED ISSUE #1005: JavaScript: Tracking destructured object's property inside function
What would be the way to track destructured object's property inside function?

example:
```javascript
function test({foo}, bar) {
    console.log(foo)
}
```

trying to track `foo` from the example I came up with several patterns but none of them works:

link: https://semgrep.live/3qBG

```yaml
  pattern: |
    function $F ($FOO,...) {
      ...
      console.log($FOO);
      ...
    }
```

```yaml
  pattern: |
    function $F ({$FOO},...) {
      ...
      console.log($FOO);
      ...
    }
```

```yaml
  pattern: |
    function $F ($INPUT,...) {
      ...
      console.log($INPUT.foo);
      ...
    }
```

```yaml
  patterns:
    - pattern-inside: function $F ($FOO,...) {...}
    - pattern: console.log($FOO);
```

```yaml
  patterns:
    - pattern-inside: function $F ({$FOO},...) {...}
    - pattern: console.log($FOO);
```

nbrahms:
Previously, we could not match these, as we could not reuse the symbol
in

  let $FOO = arg.$FOO

when matching

  let foo = arg.foo

Here we add the ability to match pure identifier nodes against
identifier expressions, allowing this match.

Fixes #1005.

### DIFF
--- a/semgrep-core/matching/Matching_generic.ml
+++ b/semgrep-core/matching/Matching_generic.ml
@@ -221,6 +221,14 @@ let rec equal_ast_binded_code (a: AST.any) (b: AST.any) : bool = (
       a =*= b
   | A.I _, A.E (A.Id (b_id, _)) ->
     (* Allow identifier nodes to match pure identifier expressions *)
+
+    (* You should prefer to add metavar as expression (A.E), not id (A.I),
+     * (see Generic_vs_generic.m_ident_and_id_info_add_in_env_Expr)
+     * but in some cases you have no choice and you need to match an expression
+     * metavar with an id metavar.
+     * For example, we want the pattern 'const $X = foo.$X' to match 'const bar = foo.bar'
+     * (this is useful in the Javascript transpilation context of complex pattern parameter).
+     *)
       equal_ast_binded_code a (A.I b_id)
   | _, _ ->
       false
--- a/semgrep-core/matching/Matching_generic.ml
+++ b/semgrep-core/matching/Matching_generic.ml
@@ -198,8 +198,8 @@ let (fail : tin -> tout) = fun _tin ->
 (* pre: both 'a' and 'b' contains only regular code; there are no
  * metavariables inside them.
  *)
-let equal_ast_binded_code (a: AST.any) (b: AST.any) : bool =
-  match a, b with
+let rec equal_ast_binded_code (a: AST.any) (b: AST.any) : bool = (
+  let res = (match a, b with
   | A.I _, A.I _
   | A.N _, A.N _
   | A.E _, A.E _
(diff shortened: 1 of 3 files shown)


semgrep-core/matching/Matching_generic.ml
```diff
@@ -218,16 +218,22 @@ let equal_ast_binded_code (a: AST.any) (b: AST.any) : bool =
...
+      a =*= b
+  | A.I _, A.E (A.Id (b_id, _)) ->
+    (* Allow identifier nodes to match pure identifier expressions *)
+      equal_ast_binded_code a (A.I b_id)
   | _, _ ->
       false
+  ) in
+
+  if !Flag.debug_matching && not res
+  then begin
```
nbrahms, line 238: Note that I also print debugging information even when the two node types differ.

semgrep-core/matching/Matching_generic.ml
```diff
@@ -198,8 +198,8 @@ let (fail : tin -> tout) = fun _tin ->
 (* pre: both 'a' and 'b' contains only regular code; there are no
  * metavariables inside them.
  *)
-let equal_ast_binded_code (a: AST.any) (b: AST.any) : bool =
-  match a, b with
+let rec equal_ast_binded_code (a: AST.any) (b: AST.any) : bool = (
```
nbrahms, line 201: This isn't tail recursive, but it only ever gets to a stack depth of 2, so 🤷

semgrep-core/matching/Matching_generic.ml
```diff
@@ -218,16 +218,22 @@ let equal_ast_binded_code (a: AST.any) (b: AST.any) : bool =
...
-      let res = a =*= b in
-      if !Flag.debug_matching && not res
-      then begin
-        pr2 (spf "A = %s" (str_of_any a));
-        pr2 (spf "B = %s" (str_of_any b));
-      end;
-      res
-
+      a =*= b
+  | A.I _, A.E (A.Id (b_id, _)) ->
```
nbrahms, line 222: @aryx @mjambon I'm open to suggestions on how to do this better :)

aryx: We might need also to handle the symetric pattern
A.E (A.ID (a_id, _)), A.I _

semgrep-core/tests/js/misc_destructuring_2.sgrep
```diff
@@ -0,0 +1,3 @@
+function $F ({$FOO}, ...) {
```
aryx, line 1: I would also add the simpler  const $X = foo.$X; more obvious testcase.

semgrep-core/matching/Matching_generic.ml
aryx: Maybe add a comment such as
"you should prefer to add metavar as expression (A.E), not id (A.I), see Generic_vs_generic.m_ident_and_id_info_add_in_env_Expr  but in some cases you have no choice
and you need to match an expression metavar with an id metavar.
For example, we want the pattern 'const $X = foo.$X'  to match 'const bar = foo.bar' (this is useful in the Javascript transpilation context of complex pattern parameter)."

aryx (approved):

semgrep-core/tests/js/misc_destructuring_2.sgrep
nbrahms: Oddly enough you can not write this pattern 🤷

nbrahms: Apparently you can only write this pattern via transpilation from destructuring; it can not be parsed directly.

PR #1488, opened 2020-08-11
minor readme fixes

ievans:


README.md
```diff
@@ -83,7 +83,7 @@ Once installed, Semgrep can be run with single rule patterns or entire rule pack
 
 ```sh
 # Check for Python == where the left and right hand sides are the same (often a bug)
-$ semgrep -e `$X==$X` --lang=py path/to/src
+$ semgrep -e `$X == $X` --lang=py path/to/src
```
mschwager: I don't think we want to use backticks here. In bash this is equivalent to invoking a subshell (same as `$(...)`). E.g. this gives me:

```
$ python -m semgrep -e `$X == $X` --lang=python tests/
==: command not found
```

This should be:

```
$ semgrep -e '$X == $X' --lang=py path/to/src
```

mschwager (dismissed):

README.md
ievans: Good catch (it's present in `develop` right now but I'll fix it with this PR)

dlukeomalley (approved):
Thanks for the improvements, these are great.

PR #1495, opened 2020-08-11
Adding support for JUnit XML format (#1388)

### LINKED ISSUE #1388: Support for JUnit XML format as output
Would love to see an option to get output as JUnit XML, since many CI/CD tools support this out of the box. This can be done via this module:
https://github.com/kyrus/python-junit-xml

nightwatchcyber:
Closes #1388


brendongo:
Thanks for the PR @nightwatchcyber! We are currently thinking through how we want to handle different output formats for semgrep going forward and will get back to early next week.

DrewDennison:
@nightwatchcyber thanks again for the PR. We've going to try to get this in this week. A couple small requests that we may get to before you do 1) can we pin `junit_xml` and `xml_diff` just to prevent against malicious updates in the future. 2) we're going to change the flag  to `--fmt json`, `--fmt sarif`, and thus `--fmt junit` just to make the API more extensible. Thanks again for the contribution!

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## [Unreleased]
 - `--timeout-threshold` flag to set the maximum number of times a file can timeout before it is skipped
+- Added support for the JUnit XML report format (`--junit-xml`)
```
brendongo, line 6: Should rebase and move these to a new unreleased

semgrep/Pipfile
```diff
@@ -10,9 +10,11 @@ tox = "*"
 pytest-snapshot = "*"
 appdirs = "*"
 pytest-xdist = "*"
+xmldiff = "*"
 
 [packages]
 semgrep = {editable = true,path = "."}
+junit-xml = "*"
```
brendongo: No need to put this here since the dependencies in setup.py are installed when semgrep is installed
```suggestion
```

semgrep/setup.py
```diff
@@ -126,13 +126,14 @@ def run(self):
     long_description_content_type="text/markdown",
     url="https://github.com/returntocorp/semgrep",
     install_requires=[
+        "attrs>=19.3.0",
         "colorama>=0.4.3",
+        "junit_xml>=1.9",
```
brendongo:
```suggestion
        "junit_xml==1.9",
```

semgrep/tests/e2e/test_check.py
```diff
@@ -42,6 +43,16 @@ def test_terminal_output(run_semgrep_in_tmp, snapshot):
     )
 
 
+def test_junit_xml_output(run_semgrep_in_tmp, snapshot):
```
brendongo, line 46: Thanks for adding tests!

nightwatchcyber:
I updated the PR to pin the file but didn't do anything about changing the CLI command yet

brendongo:
Merged in #1720. Thanks @nightwatchcyber!

PR #1507, opened 2020-08-13
Add flag to ignore explicit files with unknown extensions

brendongo:
By default semgrep treats explicitly passed files with unknown extension as possibly any language and so runs all rules on said files. This adds a flag so that semgrep will treat these files as if they matched no language and will so run no rules on them.

Useful for https://github.com/returntocorp/semgrep-action/issues/35

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -109,9 +109,9 @@ def cli() -> None:
         help="Scan all files even those ignored by a projects gitignore(s)",
     )
     parser.add_argument(
-        "--ignore_unknown_extentions",
+        "--skip-unknown-extensions",
         action="store_true",
-        help="By default, explicity passed targets with an extension semgrep does not understand is treated as if it matches all languages. Setting this flag will have semgrep ignore such files.",
+        help="Scan only known file extensions, even if unrecognized ones are explicitly targeted.",
     )
 
     config.add_argument(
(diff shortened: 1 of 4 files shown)


semgrep/semgrep/cli.py
```diff
@@ -108,6 +108,11 @@ def cli() -> None:
         action="store_true",
         help="Scan all files even those ignored by a projects gitignore(s)",
     )
+    parser.add_argument(
+        "--ignore_unknown_extentions",
```
underyx:
```suggestion
        "--ignore-unknown-extensions",
```

(will require changes later as well)

underyx (changes requested):

semgrep/semgrep/cli.py
```diff
@@ -108,6 +108,11 @@ def cli() -> None:
         action="store_true",
         help="Scan all files even those ignored by a projects gitignore(s)",
     )
+    parser.add_argument(
+        "--ignore-unknown-extensions",
+        action="store_true",
+        help="By default, explicity passed targets with an extension semgrep does not understand is treated as if it matches all languages. Setting this flag will have semgrep ignore such files.",
```
underyx:
```suggestion
        help="Scan only known file extensions, even if unrecognized ones are explicitly targeted.",
```

underyx:
It's too late for me to properly think about this right now, so I'm not approving yet. This option feels a bit out of place and I wonder if a better solution might exist. Will try to figure it out proper tomorrow.

semgrep/semgrep/cli.py
```diff
@@ -108,6 +108,11 @@ def cli() -> None:
         action="store_true",
         help="Scan all files even those ignored by a projects gitignore(s)",
     )
+    parser.add_argument(
+        "--ignore-unknown-extensions",
```
underyx:
```suggestion
        "--skip-unknown-extensions",
```

Ignore can be misinterpreted as "ignore that these extensions are unknown and scan anyway"

underyx (approved):
Nice, looks good!

PR #1512, opened 2020-08-14
Dockerfile: Ensure /src exists

nbrahms:
Because we now run without root as the user, we need to create this
directory first.

We should release 0.19.2 after this in order to unblock CI users.


DrewDennison (approved):

DrewDennison:
@nbrahms thanks. I can cut 0.19.2 for you late tonight

PR #1529, opened 2020-08-18
Update CONTRIBUTING.md

ekse:
The documentation refers to semgrep_core but the folder is named semgrep-core.


mschwager (approved):
Good catch! Thank you!

PR #1531, opened 2020-08-18
Only respect nosem after a comment

### LINKED ISSUE #1053: Consider requiring comment symbol in 'nosem' functionality
Currently 'nosem' functionality looks for the first line of a finding including " nosem". This yields false positives in situations like:

```python
def foobar():
    nosemblance_of_reality()
```

When we really only care about situations like:

```python
def foobar():
    insecure_func()  # nosem
```

One improvement here would be to require a comment character at the beginning of 'nosem' detection. The reason I initially decided against this was to avoid having to keep track of all possible comment styles in all languages, and a line containing " nosem" seems unlikely (this is why I chose to include a " " in the beginning), but maybe it's worth it. A good first pass could be something like `(#|//|/*) nosem`.

nbrahms:
This avoids treating a code symbol that happens to be named "nosem" as
an indicator to disable semgrep.

Bonus: Remove spurious unused capturing group from semgrep regex.

Fixes #1053.

### DIFF
--- a/semgrep/semgrep/constants.py
+++ b/semgrep/semgrep/constants.py
@@ -47,7 +47,7 @@ NOSEM_INLINE_RE = re.compile(
     # * We want multi-language support, so we cannot strictly look for
     #   Python comments that begin with '# '
     #
-    r" nosem(?::[\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?",
+    r"(?:#|\/\/|\/\*)\s+nosem(?::[\s]?(?P<ids>(?:[^,\s](?:[,\s]+)?)+))?",
     re.IGNORECASE,
 )
 COMMA_SEPARATED_LIST_RE = re.compile(r"[,\s]")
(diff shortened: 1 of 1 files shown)


nbrahms:
Note that `nosem` won't be available in languages with other comment syntaxes, but these will now get raised as bugs we can address.

semgrep/semgrep/constants.py
```diff
@@ -47,7 +47,7 @@ def is_json(self) -> bool:
     # * We want multi-language support, so we cannot strictly look for
     #   Python comments that begin with '# '
     #
-    r" nosem(?::[\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?",
+    r"(?:#|\/\/|\/\*)\s+nosem(?::[\s]?(?P<ids>(?:[^,\s](?:[,\s]+)?)+))?",
```
brendongo:
```suggestion
    r"(?:#|[/]{2,3}|[/][*]|[(][*])\s+nosem(?::[\s]?(?P<ids>(?:[^,\s](?:[,\s]+)?)+))?",
```
Handle Ocaml as well

mschwager:
I would lean toward "doing nothing" here since we haven't had any complaints or bugs reported, and doing nothing is always the cheapest possible solution. But I don't have strong feelings if you want to move forward with this :+1:

brendongo:
I lean on doing nothing too. Closing for now til existing behavior is an issue for someone

PR #1532, opened 2020-08-18
Add Semgrep integration playbook docs

mschwager:
There's still some TODO. I'm putting this up to get early feedback. One high-level thought I'm having: take what we currently have and include a more concrete "r2c implementation" at the bottom. I.e. after synthesizing all this information, here's the approach we've chosen for ourselves. I think this will be helpful internally as well since most of this information is a bit abstract.

Linking https://github.com/returntocorp/enterprise/issues/298.

Rendered view: https://github.com/returntocorp/semgrep/blob/c107cedca4c08c86d022dbf84bc86e0c7dca0f39/docs/integration-playbook.md.

### DIFF
--- a/docs/integration-playbook.md
+++ b/docs/integration-playbook.md
@@ -1,0 +1,188 @@
+# Semgrep Integration Playbook
+
+This document describes common integration methods for including Semgrep in
+your workflows. This includes strategies for when and where to include Semgrep,
+advice on managing Semgrep in large codebases with many pre-existing issues,
+using Semgrep to help ensure the security of your organization, etc.
+
+Contents:
+
+* [Integration Points](#integration-points)
+  * [Development](#development)
+  * [Security](#security)
+* [Lifecycle Management](#lifecycle-management)
+  * [Installing Semgrep](#installing-semgrep)
+  * [Managing Rules](#managing-rules)
+
+## Integration Points
+
+This section describes common integration points where Semgrep is effective.
+
+### Development
+
+Integrating Semgrep into development workflows allows you to catch bugs and
+security issues before they reach production environments. There are many
+points where Semgrep can be integrated into development workflows. Generally
+speaking, integrating Semgrep earlier in the development pipeline allows you
+to catch mistakes sooner and easily fix any issues it surfaces. Correcting an
+issue in your code editor is much easier than waiting days or weeks for a
+security or QA team to surface the issue. However, CI systems are often more
+reliable and repeatable than results from developer's environments. This
+section will move from earliest to latest in the development pipeline.
+
+**One-off Analyses**
+
+Of course, Semgrep can always be [run locally](https://github.com/returntocorp/semgrep#getting-started)
+to investigate findings, search for new patterns, or simply to tinker with the
+tool. Running one-off analyses provides maximal flexbility while losing
+reliability, repeatablility, and consistency. We recommend using this approach
+during initial investigations and then moving to one of the below approaches
+once the tool has been tuned and [configured](https://github.com/returntocorp/semgrep/blob/develop/docs/configuration-files.md)
+to your environemnt.
+
+**In-editor**
+
+In-editor integration allows you to see Semgrep findings inside your editor.
+This is the fastest way to get Semgrep results and offers the convenience of
+seeing findings alongside the code you're editing.
+
+In-editor integration is available for the following code editors:
+
+* [Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=semgrep.semgrep)
+
+**Git Hooks**
+
+Git hooks integration allows you to check for findings before code is pushed
+to a remote repository. This allows you to quickly see issues without waiting
+for a CI process to run or a code reviewer to notice them.
+
+r2c recommends using the [`pre-commit`](https://pre-commit.com/) framework for
+Git hook integration. However, the following is a complete listing of Git hook
+integrations:
+
+* [`pre-commit`](https://github.com/returntocorp/semgrep/blob/develop/docs/integrations.md#pre-commit-hook)
+* [Git Hooks](https://www.atlassian.com/git/tutorials/git-hooks)
+
+**Continuous Integration**
+
+Continuous Integration (CI) allows you to check each commit for findings and
+halt the run if necessary. CI is typically invoked on a remote server when
+changes are pushed to a [VCS](https://en.wikipedia.org/wiki/Version_control).
+Including Semgrep in CI gives the most assurance that no findings will fall
+through the cracks. CI environments are often more reliable than expecting
+uniformity and consistency across many development environments.
+
+The following is a complete listing of CI integration avenues alongside our
+existing CI provider documentation:
+
+* [Semgrep CI Providers](https://github.com/returntocorp/semgrep/blob/develop/docs/integrations.md#continuous-integration)
+* [`semgrep-action` GitHub Action](https://github.com/marketplace/actions/semgrep-action)
+* [Coinbase Salus](https://github.com/coinbase/salus)
+
+### Security
+
+Security teams will naturally have an adversarial relationship with development.
+Security is blocking, not enabling. Security is a cost center, not a value
+producer. Security is fundamentally destructive, not constructive. It is for
+these reasons that security must help our friends in development by choosing
+solutions, technologies, tools, and products that reduce friction as much as
+possible. We must be available for questions and provide guidance, while not
+letting perfect be the enemy of good. Although we must sometimes be secretive,
+we must strive to be understandable and clear. We must enable confidence and
+take on the bulk of integration work since we are fundamentally the ones making
+a request.
+
+This section will take a different approach than the development section above.
+Security will often be making requests of development teams and thus will likely
+need to reference the above integration points. However, this section will
+instead discuss pre-integration research, integration point trade-offs,
+smoothly rolling out Semgrep, etc. In other words, strategic considerations
+instead of purely technical capabilities.
+
+**Pre-integration Research**
+
+Before integrating Semgrep into your workflows you will likely need to do a
+bit of preliminary research. This includes asking yourself questions like:
+What languages, frameworks, and technologies is this project using? What
+vulnerability classes am I concerned about? What integration point(s) can I
+leverage to maximize developer productivity while minimizing risk of security
+issues? Research will likely start with a **one-off analysis** as mentioned
+above. This section will discuss trade-offs to consider while performing
+pre-integration research.
+
+After performing a language and framework asset inventory you may start to
+build a corpus of rules you'd like to run against the project. r2c has developed
+many openly available rule sets. E.g.
+
+* You discover mostly Python code: [`/p/python`](https://semgrep.dev/p/python)
+* You're concerned about Golang security: [`/p/gosec`](https://semgrep.dev/p/gosec)
+* You'd like to replace your slow FindSecBugs tooling: [`/p/findsecbugs`](https://semgrep.dev/p/findsecbugs)
+* You discover Flask web server usage: [`/p/flask`](https://semgrep.dev/p/flask)
+
+Although security is the primary focus, there are also many packs that look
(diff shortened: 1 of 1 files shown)


docs/integration-playbook.md
```diff
@@ -0,0 +1,188 @@
...
+The following is a complete listing of CI integration avenues alongside our
+existing CI provider documentation:
+
+* [Semgrep CI Providers](https://github.com/returntocorp/semgrep/blob/develop/docs/integrations.md#continuous-integration)
+* [`semgrep-action` GitHub Action](https://github.com/marketplace/actions/semgrep-action)
+* [Coinbase Salus](https://github.com/coinbase/salus)
+
+### Security
+
+Security teams will naturally have an adversarial relationship with development.
```
minusworld: This reads pessimistically to me. While I agree that pointing out the friction between devs and appsec can evoke the kind of response we want, I think as written it has the danger of engendering that friction further. How could we rephrase this in such a way that Semgrep presents itself as a solution to the problem of developer/security friction?

minusworld (commented):
I really like the way this is shaping up. I left one comment.

docs/integration-playbook.md
```diff
@@ -0,0 +1,188 @@
+# Semgrep Integration Playbook
+
+This document describes common integration methods for including Semgrep in
+your workflows. This includes strategies for when and where to include Semgrep,
```
dlukeomalley: Who does "your" refer to in this case? Is it any user/integrator, or tailored for a specific audience, say security engineer?

docs/integration-playbook.md
```diff
@@ -0,0 +1,188 @@
...
+advice on managing Semgrep in large codebases with many pre-existing issues,
+using Semgrep to help ensure the security of your organization, etc.
```
dlukeomalley:
```suggestion
advice on managing Semgrep in large codebases with many pre-existing issues, 
and using Semgrep to help ensure the security of your organization.
```

docs/integration-playbook.md
```diff
@@ -0,0 +1,188 @@
...
+
+Integrating Semgrep into development workflows allows you to catch bugs and
+security issues before they reach production environments. There are many
+points where Semgrep can be integrated into development workflows. Generally
+speaking, integrating Semgrep earlier in the development pipeline allows you
+to catch mistakes sooner and easily fix any issues it surfaces. Correcting an
+issue in your code editor is much easier than waiting days or weeks for a
+security or QA team to surface the issue. However, CI systems are often more
+reliable and repeatable than results from developer's environments. This
+section will move from earliest to latest in the development pipeline.
```
dlukeomalley, line 33: What do you think about taking a stronger stance here and recommending that Semgrep be integrated as early in the development workflow as possible?

```suggestion
Semgrep is most effective when results can be addressed early in the development workflow.
In general we recommend running Semgrep as close to code-conception time as possible, ideally in the editor, on commit, or at CI-time. Each integration point has pros/cons, with CI being the most reliable due to its homogenous and repeatable environment. This section will discuss each integration point.
```

docs/integration-playbook.md
```diff
@@ -0,0 +1,188 @@
...
+
+**One-off Analyses**
```
dlukeomalley, line 35: What do you think about striking this section in the spirit of it not being an integration point?

docs/integration-playbook.md
```diff
@@ -0,0 +1,188 @@
...
+
+Of course, Semgrep can always be [run locally](https://github.com/returntocorp/semgrep#getting-started)
+to investigate findings, search for new patterns, or simply to tinker with the
+tool. Running one-off analyses provides maximal flexbility while losing
+reliability, repeatablility, and consistency. We recommend using this approach
+during initial investigations and then moving to one of the below approaches
+once the tool has been tuned and [configured](https://github.com/returntocorp/semgrep/blob/develop/docs/configuration-files.md)
+to your environemnt.
+
+**In-editor**
```
dlukeomalley, line 45: As part of the playbook, are there recommendations you have about what rules to run that are particularly good / fast for IDE? How should I think about using IDE with my devs?

docs/integration-playbook.md
```diff
@@ -0,0 +1,188 @@
...
+Contents:
+
+* [Integration Points](#integration-points)
+  * [Development](#development)
+  * [Security](#security)
+* [Lifecycle Management](#lifecycle-management)
+  * [Installing Semgrep](#installing-semgrep)
+  * [Managing Rules](#managing-rules)
+
+## Integration Points
```
dlukeomalley, line 19: What do you think about a visual to highlight these plugin points visually?

docs/integration-playbook.md
dlukeomalley: +1

dlukeomalley (changes requested):
@mschwager I started a review and left a few comments (I have not yet finished the entire document). Biggest questions for me are:

1. What is the goal of this work and what outcome for Semgreps users will it enable?
2. Who is the intended audience?
3. How will we measure success of this document as it pushed forward the answer to #1?

mschwager:
>     1. What is the goal of this work and what outcome for Semgreps users will it enable?
>     2. Who is the intended audience?

I'd recommend taking a look at the original issue: https://github.com/returntocorp/enterprise/issues/298. Specifically:

> What outcome is made possible for our users?
> 
>     Internal stakeholders can be confident in the security posture of code r2c produces
>     Internal stakeholders can be confident that static analysis policies are consistently applied
>     Internal engineers trust that security continuously has their back and have a point of reference when things go wrong
>     Internal security members can ensure security at scale and have a consistent playbook with which to reference
>     External users have a "best in class" playbook for guiding their Semgrep integration experience

>     3. How will we measure success of this document as it pushed forward the answer to #1?

I'm not sure I have a good answer for measuring the "success" of a document. This is first and foremost a source of truth for holding our internal security processes accountable with regards to Semgrep. On top of that, we're making it externally available so other organizations have a resource for high-level, strategic integration planning.

docs/integration-playbook.md
mschwager: I meant to show that this document contains these things, but not only these things. Better phrasing may be "This includes, but is not limited to, strategies for ..."

docs/integration-playbook.md
mschwager: Generally I think visuals can be useful for showing complex interactions, but I tend to avoid them for a few reasons: 1.) they're not searchable, 2.) they're time consuming to create, update, and maintain (what service was I using to create that image? how can others access it to update it? clicking around in a UI is slow vs. updating text). I agree that visuals can help enhance existing information, which is what I think you're getting at here, but I'd leave that for a v2, or for someone else to include and maintain if they feel strongly about it.

docs/integration-playbook.md
mschwager: I don't think it's strictly better to integrate Semgrep as early as possible in the dev workflow - there are trade-offs to each approach. Relying on local dev machines gives faster feedback, but is unreliable. Depending on where you are on the speed vs. reliability spectrum you may choose a different approach. This is why it's often best to go with both, you get speed and reliability, but there may be other constraints like you can't install software on devs machines, or you can't easily integrate into CI (some other team owns that process). Further, if you're doing information gathering by running Semgrep in CI it may not be appropriate yet to include findings in a devs editor.

docs/integration-playbook.md
mschwager: I agree that it's not a technical integration point, but it can be a useful workflow. If I, as an AppSec or QA engineer, can't easily integrate into any of the other purposed integration points, then I can always setup a calendar reminder to run Semgrep once a week against my company's monorepo and manually create issues. Perhaps this is a stop-gap solution until I can integrate into CI, but the QA team typically takes a while to process new tool requests. Note this is also referenced below in "Pre-integration Research."

I agree that this is a little hamfisted in this list, but I do think it can be valuable. Is there a better way we can include this information?

docs/integration-playbook.md
mschwager: What about something like: "This document describes common integration methods for you, the intrepid integrator, to include Semgrep in your organization's workflows."

Since the integrator could be an AppSec engineer, QA engineer, or regular 'ol Software Engineer I'm hesitant to mention any specific role.

docs/integration-playbook.md
mschwager: Hmm, this is an interesting consideration that isn't addressed in this doc. Similarly as above, this is another good reason you may choose CI over in-editor or vice versa: CI can include more, and more comprehensive, rules, and in-editor you'll likely want a limited, fast set of rules.

It's hard to come up with a generalized heuristic for good/fast rules in the IDE. One loose heuristic I could see would be only checking for banned APIs (e.g. specific function calls, classes, or methods) in the IDE, and looking for multiline violations (e.g. basically, anything that would use `...` for multiline statements) in CI. Since we don't have a great understanding of the heuristic ourselves, I think it'd be better to kick this suggestion down the line until we understand the problem space better. I'll add a TODO for a "Performance Considerations" section :+1:

docs/integration-playbook.md
clintgibler: I propose this wording instead:

> Historically, in many companies, there was periodic friction between engineering and security teams when developers wanted to push forward and release new features quickly, while the security teams wanted to slow down or stop new releases until it could be ensured that the new code met a certain security bar.

> But times have changed. Now, it's essential for modern security teams to enable the engineering teams they support to move quickly *and* securely, by choosing the right solutions, technologies, tools, and products that reduce friction as much as possible. This is what Semgreps aims to help security teams achieve: being business *enablers* with a seat at the table.

> Security teams must be available for questions and provide guidance, while not letting perfect be the enemy of good. We must enable confidence and take on the bulk of integration work since we are fundamentally the ones making a request.

> This section will take a different approach ...

docs/integration-playbook.md
clintgibler: @mschwager - that wording makes sense to me.

docs/integration-playbook.md
clintgibler: +1 @mschwager makes sense to me.

docs/integration-playbook.md
clintgibler: Also, you probably want to have a very lean, high signal set of rules for IDE (analogous to the CI rule set), because you don't want to annoy with FPs, whereas in CI/audit workflow you may be OK with running things that are a bit noisier. 

+1 getting broad coverage of IDEs is a non starter in many medium+ sized companies.

docs/integration-playbook.md
clintgibler: I like that we have discussion of the pros and cons of various approaches. I'm not sure if this is the right doc for a discussion of those tradeoffs to live permanently, but in the vein of not letting perfect be the enemy of good enough, I'm fine with keeping this content here and moving it later.

docs/integration-playbook.md
clintgibler: Perhaps a separate doc or at least section on how we think about what to include in CI vs IDE would be useful at some point. But I'd punt on this for now.

docs/integration-playbook.md
clintgibler: I like the idea of having gifs or at least links to videos for some of these, but I'd probably punt this to v2.

docs/integration-playbook.md
colleend: I think that adding in perf considerations would be great 👍

clintgibler:
@mschwager Thanks for spec-ing this out! I think after updating the beginning of the `### Security` section to change the overall sentiment/angle, I'm down to ship this and then we can move on to dogfooding internally 👍

clintgibler (approved):
Pls adjust the tone/angle of the security/development team interactions section intro then I'd say ship it 🚀

minusworld (approved):
changes look good. 👍

aryx:
What's the status of this PR? Should we merge it @mschwager ?

mschwager:
Closing this due to staleness - if we need it in the future it's here.

PR #1535, opened 2020-08-19
Update perf tests to use correct rules directory

mschwager:
We were searching the entire `njsscan` repository for rules files (anything ending in `.yaml`). This found https://github.com/ajinabraham/njsscan/blob/master/njsscan/rules/missing_controls.yaml, which is not a rules file. We should instead be using this directory: https://github.com/ajinabraham/njsscan/tree/master/njsscan/rules/semantic_grep.


underyx (approved):

mschwager:
This should fix failing CI: https://github.com/returntocorp/semgrep/runs/1002841740

PR #1538, opened 2020-08-19
Multiple assign Fix

### LINKED ISSUE #1490: Semgrep cannot detect assignment for multiple variables in one line
**Describe the bug**
When you have multiple variable declarations in one line and want to find one of those declarations with a pattern, semgrep cannot find it.

**To Reproduce**
This works [in C](https://semgrep.dev/Nx4N/). However in languages like [Ruby](https://semgrep.dev/wpxd/) and [Python](https://semgrep.dev/kWLA/), the pattern fails to find a variable declaration.

**Expected behavior**
The behavior that occurs in the C example is what's expected.

colleend:
This should fix #1490, multiple assignment matching, if the number of variables and the number of values are equal.


aryx (approved):
perfect

Issue #1460, opened 2020-08-06
Semgrep doesn't warn if yaml rules contain unexpected values e.g. due to missing whitespace


disconnect3d:
When we include a `-pattern-not:` instead of `- pattern-not:` as `rules[patterns]` key in the multi-pattern yaml semgrep rules, it will not be treated as a `pattern-not`.

The screenshots below show an example of that from the semgrep tutorial.

**Screenshots**
![image](https://user-images.githubusercontent.com/10009354/89531228-2f113080-d7f0-11ea-98cb-7f5c4a05dafe.png)

vs

![image](https://user-images.githubusercontent.com/10009354/89531285-43edc400-d7f0-11ea-8b8c-bb7ef1bd1332.png)

**What can we do**
This could be ignored, as that (most likely) comes from the yaml parser itself. However, semgrep could validate all the resulting yaml and warn if there is something unexpected?

underyx:
Thinking we should replace our homegrown simple schema validation code with jsonschema.

nbrahms:
Fixed in 0.21

PR #1542, opened 2020-08-19
Use jsonschema for rule config validation

nbrahms:
Previously we used an ersatz schema validation system scattered
throughout different parts of our code base, and requiring significant
manual work to update.

Here we substitute in jsonschema as a validation system. This should
allow us to catch more problems (e.g. the motivating example in #1460).

Advantages:
- Centralized schema definition using a publicly documented definition
  language
- Properly handles recursion in patterns, etc.
- Uniform errors
- Adds support for previously missing items (e.g. 'metavariable-regex')

Disadvantages:
- Adds an external dependency to our application


semgrep/semgrep/rule_lang.py
```diff
@@ -328,11 +334,15 @@ def construct_object(self, node: Node, deep: bool = False) -> YamlTree:
     yaml = YAML()
     yaml.Constructor = SpanPreservingRuamelConstructor
     data = yaml.load(StringIO(contents))
+
+    schema_path = Path(__file__).parent / "rule_schema.yaml"
+    with schema_path.open() as fd:
+        validation_schema = yaml.load(fd).unroll()
```
mschwager: Can we store this as a const Python `dict` somewhere? That way we won't have to read something from disk every time we perform validation or include a file in the package MANIFEST.

semgrep/Pipfile
```diff
@@ -12,6 +12,7 @@ appdirs = "*"
 pytest-xdist = "*"
 
 [packages]
+jsonschema = "~=3.2.0"
```
mschwager: I believe we include dependencies in `setup.py` then rely on `path = "."` below for local installation.

semgrep/semgrep/rule_lang.py
nbrahms: I much prefer using the external file, as we can publish it. I'm happy to load it once, though.

DrewDennison (approved):

aryx:
Impressive! Far cleaner indeed.
Although this jsonschema specification is quite verbose ... specifying the same thing with yacc would be more compact (if it was possible).

PR #1556, opened 2020-08-20
Match generic unbraced blocks to single-statement blocks

### LINKED ISSUE #1548: Semgrep 0.20.0 is matching things that were not matched before or missing matches
**Describe the bug**

**EXAMPLE** missing match
Semgrep matches changes across versions for the same rule.

```
      - pattern: |
          if (password == $X) {
              ...
          }
      - pattern: |
          if ($X == password) {
              ...
          }
```
The above pattern is not working anymore.

**To Reproduce**
Working on 0.18.0: https://semgrep.dev/editor/?version=0.18.0
Failing on 0.20.0 https://semgrep.dev/Wep2/?version=0.20.0

**Expected behavior**
https://semgrep.dev/editor/?version=0.18.0

**Screenshots**
N/A

**Environment**
semgrep.dev

nbrahms:
This fixes a regression from returntocorp/pfff@53808c5690, which
explicitly identified blocks on conditionals.

This previous change was necessary to distinguish between conditionals
with no else block, and conditionals with an empty else block.

However, as a side effect it caused us to no longer match
single-statement blocks with unbraced if blocks.

This commit adds that equivalence in the Semgrep layer.

Fixes #1548.

### DIFF
--- a/semgrep-core/matching/Generic_vs_generic.ml
+++ b/semgrep-core/matching/Generic_vs_generic.ml
@@ -1466,9 +1466,9 @@ and m_stmt a b =
     m_tok a0 b0 >>= (fun () ->
     (* too many regressions doing m_expr_deep by default; Use DeepEllipsis *)
     m_expr a1 b1 >>= (fun () ->
-    m_stmt a2 b2 >>= (fun () ->
+    m_block a2 b2 >>= (fun () ->
     (* less-is-more: *)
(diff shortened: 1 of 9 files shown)


aryx:
Ohoho, very nice! I was thinking to fix ast_js_build.ml but this is a better fix I think. Nice code equivalence,
and it will  work for all languages!

aryx (approved):

PR #1559, opened 2020-08-21
Tolerate errors in secondary 'make clean' commands

mjambon:
(due to missing Makefile.config and whatnot)


aryx (approved):
I dont see the difference but fine.

aryx (approved):
Ah ok there is the minus in front

PR #1563, opened 2020-08-21
Show repo URL in QA tests

mschwager:
The QA tests currently show something like:

```
semgrep/tests/qa/test_public_repos.py::test_semgrep_on_repo[public_repo_url7]
```

This updates them to use the `ids` kwarg to show something like:

```
semgrep/tests/qa/test_public_repos.py::test_semgrep_on_repo[https:/github.com/keybase/pykeybasebot]
```

I also updated the non-QA smoketest to choose a random sample of repos rather than the same 5 every time.

### DIFF
--- a/semgrep/tests/conftest.py
+++ b/semgrep/tests/conftest.py
@@ -1,7 +1,6 @@
 import contextlib
 import json
 import os
-import random
 import shutil
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


mschwager:
Okay, nevermind, looks like we can't do a `random.sample` during non-QA.

underyx (approved):

PR #1566, opened 2020-08-21
Fix jsonschema deprecation warning

nbrahms:
... by pinning validator.


semgrep/semgrep/rule.py
```diff
@@ -88,21 +88,17 @@ def _parse_boolean_expression(
...
-                        )
-                        yield BooleanRuleExpression(
-                            operator=operator,
-                            pattern_id=None,
-                            children=list(sub_expression),
-                            operand=None,
-                        )
-                    else:
-                        raise Exception("Internal error: bad schema")
+                    sub_expression = self._parse_boolean_expression(
```
nbrahms, line 93: Also just found some extra lines to delete here as well

brendongo (approved):

Issue #1518, opened 2020-08-17
Parse ?? operator in js


brendongo:
**Describe the bug**
Semgrep unable to parse ?? operator https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

**To Reproduce**
 https://semgrep.dev/1L8G/

**Expected behavior**
See doc link above

mjambon:
This will be supported when we switch to the tree-sitter implementation.

nbrahms:
Duplicate of #1351

aryx:
This works now! no more parse error.

Issue #1519, opened 2020-08-17
Parse ?. in javascript


brendongo:
**Describe the bug**
Semgrep unable to parse operator chaining in javascript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

**To Reproduce**
https://semgrep.dev/9Rrr/

**Expected behavior**
See doc above

mjambon:
linking to https://github.com/tree-sitter/tree-sitter-javascript/issues/134

nbrahms:
@brendongo Is this JavaScript, Flow, or TypeScript? IIRC, `?.` is not supported in JavaScript.

brendongo:
Out on Node14 apparently

aryx:
I think it's a proposed extension not yet approved. @mjambon knows more.

aryx:
This still does not work.

aryx:
And it's not supported yet by the tree-sitter-javascript

aryx:
So @mjambon we now need to update ocaml-tree-sitter-lang with the latest javascript grammar and hope we don't have too many modifications to do?

mjambon:
The first task is to check that it works with typescript. Yesterday I saw failed tests in tree-sitter-typescript. Not sure if Max fixed them  on the javascript side or if it the typescript grammar extension has to be updated to reflect new rules in javascript. Also I'm working on semgrep for tsx support at the moment, let's try to minimize conflicts.

aryx:
This works with ?develop

Issue #1496, opened 2020-08-12
Takes more than 5 seconds to parse large js files


brendongo:
**Describe the bug**
Many repos contain either minified vendored code (jquery.js), webpacked bundles, or just large javascript files (See https://github.com/signalapp/Signal-Desktop). Semgrep timeout when trying to parse these large files (matching a complex pattern will surely take even more time).

**Expected behavior**
Either semgrep skips these large files by default and mentions them to the user or we are able to quickly parse them.

Many of the tests in https://github.com/returntocorp/semgrep/blob/develop/semgrep/tests/public_repos.py#L264 fail because of this

minusworld:
Similar to https://github.com/returntocorp/semgrep/issues/1125

aryx:
I've removed the 5s timeout and Signal-desktop seems to parse at 100% correctly, so let's close that.
It's still slow, but it parse the 189 files in 40s

PR #1570, opened 2020-08-21
Mark newly passing tests

brendongo:


brendongo:
Does not quite close https://github.com/returntocorp/semgrep/issues/1518, https://github.com/returntocorp/semgrep/issues/1519 and https://github.com/returntocorp/semgrep/issues/1496 since while it no longer crashes when failing to parse these, it does not yet parse them in a way that allows semgrep to match patterns with them (AST still missing some things) So will keep these issues open.

mschwager (approved):

PR #1573, opened 2020-08-21
Run all repo tests on pull request

brendongo:


brendongo:
Feedback cycle is better if the unit tests and QA tests were separate jobs so running QA tests don't delay failed unit tests. Will do a follow up for that.

semgrep/tests/conftest.py
```diff
@@ -232,9 +232,5 @@ def pytest_runtest_setup(item):
 def pytest_generate_tests(metafunc):
     if "public_repo_url" in metafunc.fixturenames:
         non_qa_smoketest_count = 5  # Arbitrarily chosen
-        repos = (
-            public_repos.ALL_REPOS
-            if metafunc.config.getoption("--qa")
-            else public_repos.PASSING_REPOS[:non_qa_smoketest_count]
-        )
+        repos = public_repos.ALL_REPOS
```
underyx, line 235: How much extra time does this add to total test time?

underyx (approved):

semgrep/tests/conftest.py
brendongo: 10 minutes

PR #1577, opened 2020-08-24
Move repository-local Semgrep config to cloud app

mschwager:
I put these rules under my Github user's namespace. In the future it'd be helpful to have organization managed rules so that anyone in the org can update the configuration. For now we can use this stop-gap solution and anyone with requests can contact me.


minusworld (approved):
lgtm

PR #1584, opened 2020-08-25
Add a 'none' language to config files to support arbitrary regex scanning.

minusworld:
Add a 'none' language to config files to support arbitrary regex scanning. This enables the ability to use regexes without having to explicitly specify target files on the command line.

Previously, to scan arbitrary files with pattern-regex, you had to specify the target files on the command line. This is not how Semgrep usually works, nor how ripgrep works.

The 'none' language only supports 'pattern-regex', 'patterns', and 'pattern-either' clauses.
The 'none' language makes it easy to support scanning using pattern-regex without having
to specify target files on the CLI.

This also helps address the divergent behavior between includes on the command line and
the 'paths' and 'include' key. The 'paths' and 'include' key did not explicitly scan
included files when in the config file. This behavior persists UNLESS using the 'none'
language, in which case it will scan all files (still respecting the behavior of
all includes and excludes, both on the CLI and the config file.)

This commit also added support for HTML comments with '--test'.

Tests are included for this feature.


semgrep/semgrep/__init__.py
```diff
@@ -1 +1 @@
-__VERSION__ = "0.20.0"
+__VERSION__ = "0.20.0-dev"
```
brendongo: Did you mean to update this and pfff?

minusworld: No I did not, in fact I meticulously worked to ensure that didn't happen and it seemed to have happened anyway :(

minusworld:
image

brendongo (approved):

PR #1588, opened 2020-08-25
Support to search not only expr/stmts but also types (and patterns)

aryx:
When writing rules for OCaml, and when searching OCaml code, it's quite
frequent what you look for is not an expression or a statement but
a type. Moreover, we may need that soon for Typescript too where
people may want to specify rules just for the typing parts of
Typescript.

test plan:
semgrep --exclude parsing_errors -l ml -e : $X ref .

lang_ruby/parsing/parser_ruby_helpers.mli
2:val state_override : bool ref

tests/ml/visual/ref.mli
4:val y: int ref

### DIFF
--- a/semgrep-core/matching/Generic_vs_generic.mli
+++ b/semgrep-core/matching/Generic_vs_generic.mli
@@ -12,6 +12,9 @@ val m_stmts_deep : less_is_ok:bool->
   (AST.stmt list, AST.stmt list) Matching_generic.matcher
 (*e: signature [[Generic_vs_generic.m_stmts_deep]] *)
 
+val m_type_ :   (AST_generic.type_, AST_generic.type_) Matching_generic.matcher
+val m_pattern : (AST_generic.pattern, AST_generic.pattern) Matching_generic.matcher
+
 (*s: signature [[Generic_vs_generic.m_any]] *)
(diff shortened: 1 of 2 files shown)


ievans:
@aryx can you clarify the expression here? What is the colon between the `-e` and `$X ref` for?

aryx:
```
it's really semgrep -e ': $X ref', github removed some quote
```

aryx:
The reason for the ':' in the pattern is to disambiguate "type expressions" from "real expressions".
In OCaml 'int' can be a type, but it can also be a value, it depends on the context. The namespace for types and values are clearly separated
because they occur in different syntactic contexts. It's clear when you write the declaration 'type foo = ...' that after = you expect a type, or in   let foo = ...  that here you expect an expression.
However in semgrep if you just type 'foo' as a pattern, there is no way we can know whether you wanted to find the use of the value 'foo', or the use of the type 'foo'. To disambiguate I force to add a ':' before the type, so semgrep can know you look for a type.
I use ':' because it's a token used for typing in OCaml, as in  let (convert: int -> float) = fun i -> float_of_int i

aryx:
We should document it at some point, but right now nobody is looking directly for type, except me in OCaml code.

PR #1590, opened 2020-08-25
Use latest ocaml-tree-sitter-lang

aryx:
Martin fixed the issue with unnamed pattern-tokens
so we have no more Blank and have real tokens!
He also updated the Ruby grammar which was modified quite heavily
which led to quite a few changes in the CST and boilerplate.

test plan:
+ /home/pad/github/semgrep/semgrep-core/_build/default/bin/Main.exe -dump_tree_sitter_cst string.js
|   | Exp_stmt
|   |   | Exp
|   |   |   Assign_exp
|   |   |   |   Choice_member_exp
|   |   |   |   | Id
|   |   |   |   | a
|   |   |   | "="
|   |   |   |   Choice_this
|   |   |   |   | Str
|   |   |   |   |   DQUOT_rep_choice_imm_tok_pat_de5d470_DQUOT
|   |   |   |   |   | "\""
|   |   |   |   |   |   | Imm_tok_pat_de5d470
|   |   |   |   |   |   | foo
|   |   |   |   |   | "\""
|   |   | Auto_semi
|   |   | ""

We now have the content of the string 'foo'! no more Blank


mjambon (approved):

mjambon:
Note that due to https://github.com/returntocorp/ocaml-tree-sitter/pull/105, there are now newer files for javascript and typescript. They remove references to `semgrep_dots` (which were only partially removed and causing some build issues under some circumstances) and upgrade to the latest version of source tree-sitter-javascript grammar.

aryx:
Not sure what's the build macOS test failure. I can't even click on the error.

PR #1598, opened 2020-08-27
Update dockerfile base image to include ocaml 4.10

### LINKED ISSUE #1594: Provide a working Dockerfile for building on alpine
The current dockerfile to build semgrep, seems buggy. I could not build an docker image with it. 
there is a strang git pull in the 16th line and the opam switch also failed 

Tests on a Windows 10 host

brendongo:


Dockerfile
```diff
@@ -6,14 +6,15 @@
...
+# Pinned version of ocaml/opam2:alpine-3.12-ocaml-4.10 as of 8/27/2020
+FROM ocaml/opam2@sha256:58d4f8d57b671b4f273c9e97466d1d594dae7ca2b3829a1573d5e97eba8597cb as build-semgrep-core
 
 USER root
 RUN apk add --no-cache perl m4
 USER opam
 
 WORKDIR /home/opam/opam-repository
-RUN git pull && opam update && opam switch create 4.10.0+flambda
+RUN opam update && opam switch create 4.10.0+flambda
```
mschwager, line 17: Should we remove `opam update` too? That way we know exactly what we're getting. Is that call necessary to do the `switch`?

brendongo: Good call

mjambon: `opam update` is like `apt update`. It is needed so we can install any package, not just those known at the time the base docker image was created. So let's keep it. :-)

mjambon: I think, without the `git pull` in the local opam repository, you'll get only access to the packages available at the time the base image was created, which will get old and break eventually. I would leave the `git pull` - or switch to a different base image like we do for the rest of CI, without pinning it to a specific version.

mjambon (approved):
approving this to get it out of the way.

Dockerfile
mjambon: note that fetching most opam packages when running `opam install` relies on git anyway.

brendongo:
Oh I see. Looks like the way this image interacts with opam is through git so would not fix https://github.com/returntocorp/semgrep/issues/1594 anyway. Real fix will be to switch over to an image that interacts with opam not through git. Closing this PR and will go with real fix in a seperate PR.

PR #1600, opened 2020-08-27
Allow access to the javascript stuff to be reused by typescript.

mjambon:


aryx:
I thought we agree just removing the mli was simpler

PR #1608, opened 2020-08-28
jsonschema: allow metavariable-regex in pattern-either

brendongo:
It is useful to specify multiple metavar regex in a pattern-either. It is different
from adding an or to the regex since you might want to match a different metavariable

See https://github.com/dgryski/semgrep-go/blob/master/hostport.yml

### DIFF
--- a/semgrep/semgrep/rule_schema.yaml
+++ b/semgrep/semgrep/rule_schema.yaml
@@ -21,6 +21,7 @@ definitions:
       - $ref: '#/definitions/pattern-inside'
       - $ref: '#/definitions/pattern'
       - $ref: '#/definitions/pattern-regex'
+      - $ref: '#/definitions/metavariable-regex'
   taint-content:
     type: array
     items:


nbrahms:
What does adding `metavariable-regex` in `patterns-either` actually do? Discussing with @mschwager we couldn't figure out what it even should do, hence we specified it should only live in `patterns`.

nbrahms:
FWIW, I don't have access to the linked repo.

Are you trying to set a constraint on the result of all matches in the `patterns-either`? In that case, I would expect:
```yaml
- patterns:
  - patterns-either:
    - ...
    - ...
  - metavariable-regex:
   ...
```

or do you want multiple metavariable constraints?
```yaml
- patterns:
  - pattern: ...
  - metavariable-regex:
    ...
  - metavariable-regex:
    ...
```

If you want at least one metavariable to match one out of a set of metavariable-regex definitions, I don't think this change will accomplish that. Is it possible to add a test case for what you're trying to accomplish?

nbrahms (changes requested):
See comments

nbrahms:
Also @brendongo ... is there an issue that this PR relates to?

brendongo:
They renamed it [https://github.com/dgryski/semgrep-go/blob/master/hostport.yml_](https://github.com/dgryski/semgrep-go/blob/master/hostport.yml_)

See https://r2c-community.slack.com/archives/CK86BJ5DW/p1598644415085600

mschwager:
We had a internal conversation to discuss next steps here. The TL;DR is `metavariable-regex` does not logically make sense under a `pattern-either` operation. `metavariable-regex` is fundamentally a logical AND operation with a `pattern` or `pattern`s that provides a metavariable. We should make this more clear in our documentation. But, for now we're going to stick with disallowing it underneath `pattern-either`. This allows us to keep a simpler API and nudge users towards familiar workflows. With that being said, we realize that we unintentionally broke backwards compatibility and thus broke existing rules, and for that we'd like to apologize @dgryski.

@nbrahms did come up with an equivalent rule that doesn't require `metavariable-regex` underneath `pattern-either`: https://semgrep.dev/s/OkbB. Would that cover your use-case @dgryski?

dgryski:
LGTM.  Not apologies necessary -- we all understand these rule schemas are under development.  I'll replace my rule with the one from @nbrahms.

Issue #1514, opened 2020-08-14
Unable to do case-insensitive matching in Java


nightwatchcyber:
**Describe the bug**
Unable to do case-insensitive matching in Java

**To Reproduce**
My test case here:
https://semgrep.dev/nightwatchcyber:regex_case_insensitive_match

Based on:
https://semgrep.dev/editor?registry=java.lang.security.audit.crypto.desede-is-deprecated

**Expected behavior**
Ability to match with case-sensitivity flags normally used in Java ("{?i}" or "/i")

**Screenshots**

**Environment**
Live Editor

aryx:
Right now we use the OCaml regexp syntax, which is not great and does not support /i.
We could switch to the PCRE (the one used in Perl) which is better and support /i and /m.
@ievans what do you think?

aryx:
I wonder if switching to another regexp engine would break many semgrep-rules @minusworld ?

nbrahms:
Hi @nightwatchcyber ... 

Can you confirm that https://semgrep.dev/2L8L/ works for you? Over time we are working to deprecate the OCAML-based Regex syntax and replace it with `metavariable-regex`.

I'm going to close this issue out; please reopen it if you find yourself still blocked.

nbrahms:
> I wonder if switching to another regexp engine would break many semgrep-rules @minusworld ?

@aryx cf. https://github.com/returntocorp/semgrep/issues/688

nightwatchcyber:
@nbrahms your solution works for me, thank you!

aryx:
Should work now! You can use /i for case-insensitive search.

Issue #688, opened 2020-05-05
Support fully featured regexp in strings (e.g. PCRE)


nbrahms:
**Is your feature request related to a problem? Please describe.**
A motivating example is to match a 16-digit hex field (e.g. when searching for a hard-coded hexidecimal API key):
```python
$KEY = "=~/[A-Fa-f0-9]{16}/"
```

With vanilla ocaml regex, I'd need to instead write:

```python
$KEY = "=~/[A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9][A-Fa-f0-9]/"
```

**Describe the solution you'd like**
Support more fully featured regex dialects. For example, https://mmottl.github.io/pcre-ocaml/ would allow detection of pcre.

DrewDennison:
Would also be nice since `pattern-regex` will have PCRE support. Alternatively, we could make it so all regexing is done in python and you have to do something like 

```yaml
patterns:
    - pattern: $KEY = "..."
    - pattern-regex: "[A-Fa-f0-9]{16}"
```

and let the and logic make the magic happen. Related to #629

mschwager:
Can we close this? You need to use `pattern-inside`, but what @nbrahms is describing is now available.

aryx:
Should work now

PR #1610, opened 2020-08-31
Use PCRE regexp syntax in semgrep "=~/xxx/" pattern

aryx:
This fixes
https://github.com/returntocorp/semgrep/issues/1514
and
https://github.com/returntocorp/semgrep/issues/688

test plan:
test files included


aryx:
@ajinabraham I've seen a few rules in nodejjscan that will not work with the new syntax.
Here is the diff to update crypto_node.yaml (that we include in semgrep-rules/contrib/nodejjscan/
```
   - pattern-either:
     - pattern: |
-        $X.createCipheriv("=~/aes-\([0-9]+\)-ecb/", ...)
+        $X.createCipheriv("=~/^aes-([0-9]+)-ecb$/", ...)
     - pattern: |
-        $X.createDecipheriv("=~/aes-\([0-9]+\)-ecb/", ...)
+        $X.createDecipheriv("=~/^aes-([0-9]+)-ecb$/", ...)
```

aryx:
The failures are in semgrep-rules because of the change of syntax.
I'll update semgrep-rules in a separate PR. 
There is a mutual dependency between the CI of those 2 repos so I'll force merge this one.

underyx:
Copy-pasted from Slack for you @aryx:

> • I think this is a textbook insecure default, if /foo/ matches thai food as well, people won’t notice while testing. If the problem is the old, inverse one, that /foo/ doesn’t match thai food even if you want it to, that will be clear during the rule writing and testing process.
> • This causes false positives in our very own examples too: https://semgrep.dev/s/8xQy/?version=0.22.0 — I know we can update it, but the fact that this is how we originally wrote the rule should probably taken as evidence that the old regex behavior was more natural.
> • I think the phrase “regex matching” we use in the changelog, as well as in common parlance, implies that a full match is required. The behavior where any substring matching is okay I’ve seen usually called “regex search”. See: python re module.
> • The syntax is "=~/foo/" where the equals sign seems to suggest we are test whether the string is equal to the given pattern, not whether it’s a partial match

Couldn't we just always wrap the given pattern in `^` and `$` to get the old behavior back and undo the breaking change?

ajinabraham:
I have around 59 individual lines across 7 rule files using the old syntax. I can change them, but as @underyx mentioned if It is something that can be wrapped without causing a big problem to the regex conventions or writing patterns, wouldn't that be a better backward compatible approach?

aryx:
The original convention was a bad convention. It was using the ugly syntax of OCaml regexp because that's what was available at that time and I was too lazy to use a proper syntax. We should never have forced users to escape pipe or parenthesis because only in OCaml you have to do those things. I was also using the wrong matching function in OCaml, Str.string_match instead of Str.search_forward which is what everybody else is doing. So this was by accident that we had this bad syntax and bad semantic.

aryx:
I understand that it's a bit annoying to have to update your rules and that being backward compatible is nice, but we're still in the early stage of semgrep and it's nice to be able to update the language and remove ugly corners.

ievans:
@aryx @DrewDennison can you make sure we update the docs for this? https://github.com/returntocorp/semgrep/blob/develop/docs/pattern-features.md#strings

ajinabraham:
@aryx Are these changes present in semgrep 0.22.0?
I tried both develop and 0.22.0 with the recommended changes of adding ^ and $.
https://semgrep.dev/s/D892?version=develop
https://semgrep.dev/s/D892?version=0.22.0

Both returns no match.

aryx:
You should not escape any more parenthesis.

Here is the relevant part in Changelog.md:
- Breaking: regular expressions in semgrep string patterns (e.g., `"=~/foo/"`)
  are now using the PCRE (Perl Compatible Regular Expressions) syntax instead of
  the OCaml syntax. This means you should not escape parenthesis for grouping
  or escape pipes for dijunctions (e.g., use simply `"=~/foo|bar/"` instead of
  `"=~/foo\|bar/"`). You can also use more advanced regexp features available
  in PCRE such as case-insensitive regexps with '/i' (e.g., `"=~/foo/i"`).
  The semantic of matching changes also to look for the regexp anywhere
  in the string, not just at the beginning, which means if you want to
  enforce a format for the whole string, you will now need to use the '^' anchor
  character (e.g., `"=~/^o+$/"` to check if a string contains only a sequence
  of 'o').

aryx:
See https://semgrep.dev/s/WeLy/

ajinabraham:
Thanks for the clarification. Noted and updated in njsscan 0.1.3

PR #1611, opened 2020-08-31
Parallelize --test functionality

mschwager:
I manually compared the before/after output to confirm it was the same. Before/after:

```
real	0m55.896s
user	0m47.893s
sys	0m7.067s
```

```
real	0m17.683s
user	1m26.406s
sys	0m11.285s
```

### DIFF
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -426,9 +426,14 @@ class CoreRunner:
             ]
         except re.error as err:
             raise SemgrepError(f"invalid regular expression specified: {err}")
-        re_fn = functools.partial(get_re_matches, patterns_re)
-        with multiprocessing.Pool(self._jobs) as pool:
-            matches = pool.map(re_fn, targets)
+
+        if len(targets) > 1:
+            re_fn = functools.partial(get_re_matches, patterns_re)
+            with multiprocessing.Pool(self._jobs) as pool:
+                matches = pool.map(re_fn, targets)
+        else:
+            matches = [get_re_matches(patterns_re, targets[0])]
+
         outputs.extend(
             single_match for file_matches in matches for single_match in file_matches
         )
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -132,12 +132,14 @@ class CoreRunner:
         timeout: int,
         max_memory: int,
         timeout_threshold: int,
+        testing: bool = False,
     ):
         self._allow_exec = allow_exec
         self._jobs = jobs
         self._timeout = timeout
         self._max_memory = max_memory
         self._timeout_threshold = timeout_threshold
+        self._testing = testing
 
     def _flatten_rule_patterns(self, rules: List[Rule]) -> Iterator[Pattern]:
         """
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/core_runner.py
```diff
@@ -426,9 +426,14 @@ def handle_regex_patterns(
             ]
         except re.error as err:
             raise SemgrepError(f"invalid regular expression specified: {err}")
-        re_fn = functools.partial(get_re_matches, patterns_re)
-        with multiprocessing.Pool(self._jobs) as pool:
-            matches = pool.map(re_fn, targets)
+
+        if len(targets) > 1:
```
mschwager: This is required because we cannot do a `multiprocessing.Pool` inside another `Pool`. We need to avoid the double `Pool` with `--test` using `Pool` and our rules in `semgrep-rules` using `pattern-regex` (uses `Pool` here). Since our tests only have one corresponding target file we can limit our `multiprocessing` use to > 1 target.

mschwager: https://stackoverflow.com/questions/6974695/python-process-pool-non-daemonic

nbrahms: Have you looked into using a thread pool instead of a process pool? We got around most of these nasty issues in Bento by using multithreading inside a single process.

This won't work, of course, if your tests are mutating global state.

nbrahms: It looks like you're assuming that we never test with more than one file? IMO that's a bit fragile. Would it make more sense to set a TESTING global and check against that?

semgrep/semgrep/test.py
```diff
@@ -195,62 +198,75 @@ def confusion_matrix_to_string(confusion: List[int]) -> str:
     return f"TP: {tp}\tTN: {tn}\tFP: {fp}\tFN: {fn}"
 
 
+def invoke_semgrep_multi(filename, *args, **kwargs):
+    try:
+        output = invoke_semgrep(filename, *args, **kwargs)
+    except Exception as error:
+        return (filename, error, {})
+    else:
+        return (filename, None, output)
```
nbrahms, line 207:
```suggestion
    try:
        output = invoke_semgrep(filename, *args, **kwargs)
        return (filename, None, output)
    except Exception as error:
        return (filename, error, {})
```

nbrahms (approved):

semgrep/semgrep/core_runner.py
nbrahms: cf. http://lucasb.eyer.be/snips/python-thread-pool.html

semgrep/semgrep/test.py
mschwager: I try to only include the line I'm expecting an exception from in the `try` statement. This minimizes: "Where did this exception get thrown from?" Obviously this is an innocent enough change since we don't expect the `return` to raise, but I believe these are semantically equivalent so I'll stick to my rule of thumb here.

semgrep/semgrep/core_runner.py
mschwager: Good catch! Although I did slightly better than a global.

mschwager: Hmm, interesting. I'm going to save that for a future improvement. I've mostly achieved what I set out to with this, so I'm going to push it over the line :+1:

PR #1615, opened 2020-08-31
Remove install-script release functionality

mschwager:
We're removing the install-script installation procedure in favor of `pip install`. The install script is a large maintenance burden and provides no benefits over `pip`, `brew`, or `docker`. The most direct translation will be `pip`. Using `pip` allows you to easily uninstall Semgrep whereas the install-script leaves files around your filesystem. This also allows us to remove Nuitka functionality, which greatly complicated our build and release procedures.


semgrep/Makefile
```diff
@@ -1,7 +1,3 @@
-all:
```
mschwager, line 1: These commands were only used in `make all` in `ubuntu-release.sh` and `osx-release.sh`.

brendongo (approved):

PR #1624, opened 2020-09-02

aryx:


aryx:
@mschwager not sure what's with the circle CI error. I don't see any .circleci/ directory in semgrep now.

PR #1627, opened 2020-09-02
Remove Semgrep binary checks from ubuntu release

mschwager:
The `0.22.0` release ubuntu verification failed: https://github.com/returntocorp/semgrep/runs/1058629525

This is because we're no longer including the `semgrep` Nuitka binary. This PR removes the checks for that binary. We're still confirming that the tarball and SHA256 hash are correct, but we're no longer confirming the `semgrep` binary works because it does not exist.

We may remove the tarball and hash altogether in the future. If that happens we can remove all the ubuntu release functionality. Removing the ubuntu release functionality simplifies our release process and lowers our maintenance burden. However, some users may find it useful to be able to download a pre-compiled `semgrep-core` artifact. This is a discussion for another time, for now let's fix the release scripts :+1:

### DIFF
--- a/scripts/Dockerfile.16.04
+++ b/scripts/Dockerfile.16.04
@@ -4,11 +4,8 @@ RUN test -n "$version"
 RUN apt-get update && apt-get install -y curl sudo ca-certificates --no-install-recommends \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
-COPY ubuntu-generic.sh ./install.sh
-RUN chmod +x install.sh
-COPY validate-install.sh ./validate-install.sh
-RUN chmod +x validate-install.sh
+COPY ubuntu-generic.sh ./ubuntu-generic.sh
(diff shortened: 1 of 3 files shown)


scripts/Dockerfile.16.04
```diff
@@ -4,11 +4,8 @@ RUN test -n "$version"
 RUN apt-get update && apt-get install -y curl sudo ca-certificates --no-install-recommends \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/*
-COPY ubuntu-generic.sh ./install.sh
-RUN chmod +x install.sh
-COPY validate-install.sh ./validate-install.sh
-RUN chmod +x validate-install.sh
+COPY ubuntu-generic.sh ./ubuntu-generic.sh
```
mschwager, line 7: Using `ubuntu-generic` everywhere makes this much easier to `grep` for.

brendongo (approved):

PR #1629, opened 2020-09-02
Create semgrep.yml

DrewDennison:


ievans:
@DrewDennison seems like this is superseded now

PR #1634, opened 2020-09-03
minimal typescript support

mjambon:
It doesn't have tests at this point but shouldn't break anything either.


aryx (approved):

aryx:
What is going on with the tests failures inCI?

mjambon:
> What is going on with the tests failures inCI?

This was due to changes in `combine_infos` which I hadn't merged into my branch.

mjambon:
I added all the pure javascript (non-jsx) tests as typescript tests using symlinks. I added just one typescript-specific test. Everything passed without particular difficulties, which is nice.

aryx:
great! let's merge

Issue #877, opened 2020-06-01
Ellipsis operator does not parse in Java Interfaces


minusworld:
**Describe the bug**
It appears that the ellipsis operator does not parse when writing patterns that involve Java interfaces.  Here is an example: 
https://semgrep.live/2ZZz

Interestingly, leaving out the ellipsis operator works like I want:
https://semgrep.live/pKK3

**Expected behavior**
If the working example is expected behavior, great! We should document this somewhere. Otherwise, we should make ellipsis work for Java interfaces.

aryx:
Another good one for @IagoAbal

IagoAbal:
Hi,

I can take this one, if nobody did already @aryx?

Without having really looked carefully into the code, but having played around with the example, I think semgrep doesn't have an use for `...` here. As far as I can tell, semgrep considers a Java interface as a set of methods (hence unordered), and an interface-pattern matches an interface if each method-pattern matches a different method in it. As @minusworld has already noticed, if you omit the `...` it just works.

Possible actions:
1. Make sure that documentation is clear about this.
2. Parse `...` in Java interfaces. (I think it's more user-friendly than a parse error and this issue is kind of proof of that.)
    a. Simply ignore them.
    b. Issue a warning.

aryx:
Yes I vote for 2. Actually we support them in class_member_declaration (see the use of DeclEllipsis in parser_java.mly), so we should support them also for interface member declaration.

IagoAbal:
I think this was supposed to be implemented by https://github.com/returntocorp/pfff/commit/81ea364572db3cdd55f2b3b85523ab936c5732f0 ("Java Class Ellipsis in body") but the `interface_member_declaration` rule of the parser wasn't updated accordingly.

I have a patch that fixes that, now I have to check if the semgrep engine already handles `...` in interfaces' bodies.

PR #1637, opened 2020-09-03
tests: java: Add regression test for #877

IagoAbal:


IagoAbal:
Hmm, sorry I didn't run the tests locally but I wasn't expecting it to fail. I will take a look tomorrow, but it's weird because the original example https://semgrep.live/2ZZz from #877 now works. Please let me know if you spot any obvious mistake in the test case.

aryx:
The CI failure seems unrelated. Not sure what's going on here. Let's merge this, maybe some CI flakyness.

aryx (approved):

PR #1638, opened 2020-09-03
pypi: publish manylinux wheel as any so works on alpine

### LINKED ISSUE #1636: Support pip install in alpine
Semgrep manylinux wheel is compatible with alpine but the manylinux spec is actually not alpine compatible
so pip will skip over this wheel and try to build from source.

Let `pip install semgrep` work on alpine

Will probably need to publish an "any" compatible wheel and just error out when windows is detected.

brendongo:
Closes: https://github.com/returntocorp/semgrep/issues/1636

### DIFF
--- a/semgrep/setup.py
+++ b/semgrep/setup.py
@@ -46,10 +46,11 @@ class bdist_wheel(_bdist_wheel):
         if "macosx" in plat:
             plat = f"macosx_{MIN_OSX_VERSION}_x86_64"
 
-        # The binary we build is statically linked & manylinux compatible, so change platform
-        # accordingly
+        # The binary we build is statically linked & manylinux compatible and alpine compatible
+        # there is no way to specify works with alpine on pypi so set platform as any.
+        # Note that semgrep-core is still incompatible with Windows
         if plat == "linux_x86_64":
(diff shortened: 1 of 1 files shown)


semgrep/setup.py
```diff
@@ -46,10 +46,11 @@ def get_tag(self):
...
-        # The binary we build is statically linked & manylinux compatible, so change platform
-        # accordingly
+        # The binary we build is statically linked & manylinux compatible and alpine compatible
+        # there is no way to specify works with alpine on pypi so set platform as any.
+        # Note that semgrep-core is still incompatible with Windows
         if plat == "linux_x86_64":
-            plat = "manylinux1_x86_64"
+            plat = "any"
         elif plat == "linux_i686":
             plat = "manylinux1_i686"
```
mschwager, line 55: Should we bump this one too?

brendongo: We don't actually build on i686 or publish for it so I'm fine leaving as it

mschwager: I would remove that `elif` then. Otherwise it's confusing to see these two conditions together.

mschwager (approved):

PR #1646, opened 2020-09-04
Update CLI help messages with path to appropriate docs

clintgibler:
Just updating the path we spit out in the Semgrep CLI on where to learn more, now that we've slimmed the README.


brendongo (approved):

ievans:
@clintgibler would it be better to link them to github.com? Many users will not know where the source code lives, eg if they installed via pip or brew

PR #1655, opened 2020-09-08
Delete Typescript  symbolic links test files

aryx:
I think it is simpler to add a test to run the typescript
parser on the js/ dir

test plan:
make test


mjambon:
great, thank you.

PR #1657, opened 2020-09-09
Fix timeout handling on languages using tree-sitter and pfff

aryx:
This also switches to float for specifying timeout (easier to test things).

We were incorrectly capturing the Timeout exn in Parse_code
when the first parser (usually the tree-sitter one) was timing out.
This then caused to use the second parser (usually the pfff one) without
any more timeout.

test plan:
  semgrep -l js --timeout 0.5 -e 'FOO' tests/PERF/timeout.js
now finishes quickly and does not invoke the pfff parser.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,18 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## NEXT VERSION
 
 ### Added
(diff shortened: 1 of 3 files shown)


aryx:
@brendongo I tried to modify also cli.py to use float instead of int, but then pre-commit yelled at me with
19 errors in 4 files which seems unrelated. 
How do I make work mypi and please pre-commit?

PR #1659, opened 2020-09-09
* semgrep-core/semgrep.yml: Starting dogfooding

aryx:
Dogfooding!

test plan:
wait for CI to kick in and expect no failure

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,16 @@
+version: 2.1
+jobs:
+
+  semgrep:
+    docker:
+      - image: returntocorp/semgrep:develop
+    working_directory: /src
+    steps:
+      - checkout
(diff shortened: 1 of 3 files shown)


DrewDennison (approved):

.circleci/config.yml
```diff
@@ -0,0 +1,16 @@
+version: 2.1
+jobs:
+
+  semgrep:
+    docker:
+      - image: returntocorp/semgrep:develop
+    working_directory: /src
+    steps:
+      - checkout
+      - run: semgrep --config semgrep-core/semgrep.yml --error --strict --exclude ocaml --exclude TODO --exclude _build semgrep-core
```
mschwager, line 10: Can we store this configuration in the cloud app instead of locally? Even better to dogfood two things at once.

aryx: We already dogfood the cloud config in semgrep-app. Here I want to remove extra layers,
and it's a good way to dogfood our circleCI support :)

PR #1662, opened 2020-09-09
ci: upload to pypi on release

brendongo:


.github/workflows/release.yml
```diff
@@ -250,6 +250,31 @@ jobs:
...
+          path: manylinux-wheel
+      - name: Download Osx Artifact
+        uses: actions/download-artifact@v1
+        with:
+          name: osx-wheel
+          path: osx-wheel
+      - name: Unzip
+        run: unzip ./manylinux-wheel/dist.zip
+      - name: Unzip OSX Wheel
+        # Don't unzip tar.gz
```
mschwager: Is this because it already exists from `unzip ./manylinux-wheel/dist.zip`? If so, we could add the additional context: `# Don't unzip tar.gz because it already exists from ./manylinux-wheel/dist.zip`.

mschwager (approved):

.github/workflows/release.yml
brendongo:
```suggestion
        # Don't unzip tar.gz because it already exists from ./manylinux-wheel/dist.zip.
```

PR #1664, opened 2020-09-10
Handling more types in typescript

aryx:
test plan:
make test


mschwager:
It looks like we're adding [typed metavariable](https://github.com/returntocorp/semgrep/blob/develop/docs/pattern-features.md#typed-metavariables) support for Typescript. Should we update the docs?

aryx:
We don't have the typed metavariable for Typescript yet. We just allow to match on parameter type declaration and field type declaration and variable declaration.
Regarding type what we do for Typescript is not different than what we we for Java and Go where there's also type declarations.

PR #1665, opened 2020-09-10
Respect --json flag with --test

### LINKED ISSUE #1592: Output JSON for test arg
```bash
semgrep --test --json --config <rules> <files>
```

```json
{
   "results": {
       "nodejsscan.no-xss": {
            "triggers": 8,
            "found": [{"path": "rel/path/to/file", "start": {"line": 30, "col": 3}}],
            "expected": [{"path": "rel/path/to/file", "start": {"line": 30, "col": 3}}],
            "pass": true
        }
    },
   "all_pass": false
}
```

mschwager:
Fix #1592.

Requested by @ajinabraham.

I'd also recommend running with `--quiet` to avoid collecting any subprocess output. E.g.

```
$ python -m semgrep --quiet --json --test ~/r2c/semgrep-rules/python/django/best-practice/
{
    "config_missing_tests": [],
    "config_with_errors": [],
    "results": {
        "~/r2c/semgrep-rules/python/django/best-practice/use-onetoonefield.yaml": {
            "todo": 0,
            "checks": {
                "use-onetoonefield": {
                    "tp": 1,
                    "tn": 0,
                    "fp": 0,
                    "fn": 0,
                    "passed": true,
                    "expected_reported": {
                        "~/r2c/semgrep-rules/python/django/best-practice/use-onetoonefield.py": {
                            "expected": [
                                28
                            ],
                            "reported": [
                                28
                            ]
                        }
                    }
                }
            }
        },
        ...
    }
}
```


mschwager:
@ajinabraham will this work for you?

ajinabraham:
@mschwager This looks great! Much better than what I asked for.

Couple of queries

So for true positives 

`expected` are the lines with  `# ruleid: xxx` annotation and a match
`reported` are all matched lines irrespective of the annotation.

and for true negatives does annotating with `# ok` has any significance?
How is false positive and false negative determined? 

Few useful asks

1. On a test failure, can the code do a non zero exit? The use case for this will a repo of rules and a CI running
 `python -m semgrep --quiet --json --test <repo_root>`
2. On failure, does this provide exact and easy to debug diff per rule/file showing what rule id was not matched or extra matches are found, similar to `py.test -vvv` ?
Ref: https://github.com/ajinabraham/njsscan/runs/1089108827?check_suite_focus=true#step:8:40
3. Can I configure rules and tests in separate repository? 
    Reason being, It's cleaner to work with and package rules alone without tests https://github.com/ajinabraham/njsscan/tree/master/njsscan/rules/semantic_grep
4. Can the test runner produce a warning/error when a rule id does not have an associated test case? This enforces/promote people to write tests for all rules.
5. What is meant by `passed` is `true`? Is is just expected = reported ?
6. Would it make more sense to 
    a. rename `expected_reported` to `matches` 
    b. append `expected` and `reported` with `_lines`
 This makes it's easy for people to understand at first look.

mschwager:
> So for true positives
> 
> `expected` are the lines with `# ruleid: xxx` annotation and a match
> `reported` are all matched lines irrespective of the annotation.

Yes, that's correct.

> and for true negatives does annotating with `# ok` has any significance?
> How is false positive and false negative determined?

You can use `# ok:<rule-id>` to annotate true negatives. This helps prevent false positives. If a true negative annotation ever produces a reported line, then the `--test` functionality will error out. This would be considered a false positive.

A false negative is the opposite, and is covered by the `--test` functionality as well. That is, a false negative would mean a `# ruleid:<rule-id>` was specified, but the line wasn't reported. The `--test` functionality checks that `expected == reported`.

> On a test failure, can the code do a non zero exit? The use case for this will a repo of rules and a CI running
>        `python -m semgrep --quiet --json --test <repo_root>`

Yes, I'll make the JSON and regular output modes return the same exit code in all situations.

> On failure, does this provide exact and easy to debug diff per rule/file showing what rule id was not matched or extra matches are found, similar to `py.test -vvv` ?

We likely do not have the full power or flexibility that `pytest` has. However, now that we have JSON output it's possible to build nicer tooling on top of this functionality. We do show the reported lines vs. the expected lines which should be enough information to investigate a failure. E.g.

```
$ python -m semgrep --quiet --test ~/r2c/semgrep-rules/
...
(TODO: 1) ~/r2c/semgrep-rules/python/django/security/injection/sql/sql-injection-using-db-cursor-execute.yaml
	✖ sql-injection-db-cursor-execute                              TP: 6 TN: 1 FP: 0 FN: 1
	test: ~/r2c/semgrep-rules/python/django/security/injection/sql/sql-injection-using-db-cursor-execute.py, expected: [7, 9, 11, 17, 26, 34, 50], reported: [7, 9, 11, 17, 26, 34]
```

> Can I configure rules and tests in separate repository?

Not at the moment. Please open a new feature request for this change.

> Can the test runner produce a warning/error when a rule id does not have an associated test case? This enforces/promote people to write tests for all rules.

In `--json` mode we return the `config_missing_tests` key. This is a list of rule files that are missing tests. In normal output this is shown as:

```
$ python -m semgrep --quiet --test ~/r2c/semgrep-rules/
The following config files are missing tests:
	~/r2c/semgrep-rules/python/flask/security/xss/audit/direct-use-of-jinja2.yaml
	~/r2c/semgrep-rules/python/django/performance/access-foreign-keys.yaml
	~/r2c/semgrep-rules/javascript/angular/security/detect-angular-trust-as-html-method.yaml
...
```

> What is meant by `passed` is `true`? Is is just expected = reported ?

`passed` is `(fp == 0) and (fn == 0)` - [see here](https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/test.py#L294)

> Would it make more sense to
>        a. rename `expected_reported` to `matches`
>        b. append `expected` and `reported` with `_lines`
>        This makes it's easy for people to understand at first look.

Sure, that language makes more sense to me :+1:

brendongo (approved):

Issue #1666, opened 2020-09-10
Update base docker image used for ocaml projects


mjambon:
* change [ocaml-layer setup](https://github.com/returntocorp/ocaml-layer) to as to push to r2c's DockerHub account instead of [mjambon](https://hub.docker.com/repository/docker/mjambon/r2c-ocaml)
* set up automatic weekly updates (ocaml-layer already provides a config for this)

PR #1668, opened 2020-09-11
Use base image for docker build

mjambon:
Two changes:

* Use `returntocorp/ocaml` images instead of `mjambon/r2c-ocaml`.
* Also use this image in the dockerfile, which will speed up the build and simplify maintenance.

Background: our [ocaml-layer setup](https://github.com/returntocorp/ocaml-layer) now runs a weekly job that builds base images (alpine and ubuntu) with the desired version of ocaml and most or all external dependencies. This is meant to speed up CI jobs.

Relies on #1666

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,19 +6,15 @@
 # of the 'semgrep' wrapping.
 #
 
-FROM ocaml/opam2:alpine@sha256:4c2ce9a181b4b12442a68fc221d0b753959ec80e24eae3bf788eeca4dcb9a293 as build-semgrep-core
+FROM returntocorp/ocaml:alpine as build-semgrep-core
 
-USER root
-RUN apk add --no-cache perl m4
-USER opam
+USER user
+WORKDIR /home/user
 
-WORKDIR /home/opam/opam-repository
-RUN git pull && opam update && opam switch create 4.10.0+flambda
-
-COPY --chown=opam .gitmodules /semgrep/.gitmodules
-COPY --chown=opam .git/ /semgrep/.git/
-COPY --chown=opam semgrep-core/ /semgrep/semgrep-core/
-COPY --chown=opam scripts /semgrep/scripts
+COPY --chown=user .gitmodules /semgrep/.gitmodules
+COPY --chown=user .git/ /semgrep/.git/
+COPY --chown=user semgrep-core/ /semgrep/semgrep-core/
+COPY --chown=user scripts /semgrep/scripts
 
 WORKDIR /semgrep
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -6,19 +6,15 @@
 # of the 'semgrep' wrapping.
 #
 
-FROM ocaml/opam2:alpine@sha256:4c2ce9a181b4b12442a68fc221d0b753959ec80e24eae3bf788eeca4dcb9a293 as build-semgrep-core
+FROM returntocorp/ocaml:alpine as build-semgrep-core
```
brendongo: Can we pin this to a specific hash. We want this public docker image to build as deterministically as possible

mjambon: I was wondering about that. In addition to reproducibility, we want to stay current with the ocaml packages and reduce the amount of maintenance. How about we use the latest image but we print the image ID at runtime (`docker build`) for troubleshooting purposes?

On DockerHub, we could set a tag based on the date e.g. `returntocorp/ocaml:alpine-2020-09-11` in addition to `returntocorp/ocaml:alpine`. I would still reference `returntocorp/ocaml:alpine` in CI builds but a message would indicate that `returntocorp/ocaml:alpine-2020-09-11` is available for troubleshooting purposes.

I'll take care of tagging with dates. This should be helpful to see what changed from one week to another.

brendongo: My worry with having a dynamically updating image behind a single tag is that if someone has returntocorp/ocaml:alpine downloaded on the computer from 5 weeks ago when they docker build again it wont actually pull the latest but use the locally cached one. Not sure what level of caching GithubActions does which might also affect this. Will errors even occur when this happens or is it just a cache miss for dependencies and we end up downloading the needed ones anyway.

aryx (commented):
Lgtm

Dockerfile
mjambon: > if someone has returntocorp/ocaml:alpine downloaded on the computer from 5 weeks ago when they docker build again it wont actually pull the latest

I hadn't realized that. That's a good point.

mjambon: So what I did is the following:

* The generated dockerfile used to build the ocaml base image contains a dummy command with the current date (e.g. `RUN echo 2020-09-11`), which ensure the ocaml image is always fresh.
* Such ocaml images are built weekly.
* Each image is doubly tagged: as just `alpine` and as `alpine-2020-09-11`. See https://hub.docker.com/repository/docker/returntocorp/ocaml

For semgrep's Dockerfile, I suggest using the version like `alpine-2020-09-11`, which won't change and is friendlier than a hash. This makes it easy to communicate with users, and easy for us to see when it gets old and should be updated.

For our CI, I suggest sticking with the latest image such as `alpine`, and it's sufficiently easy to find out which one was used for a given job, by looking at the list of tags on dockerhub.

brendongo (approved):

mjambon:
ok, I'll merge this after resolving the conflict. I'm assuming the circleci error `submodule not found` can be ignored.

PR #1672, opened 2020-09-14
* Dockerfile: add git in semgrep docker image

aryx:
This seems necessary for circleCI to be able to apply semgrep in CI
on git repositories using submodules.
See https://app.circleci.com/pipelines/github/returntocorp/semgrep/785/workflows/8bc50c5e-7891-45e6-8111-1ad8224ab79b/jobs/1418 for an example of error

test plan:
wait for it to be accepted, pushed to docker, and check if next
PR get a working circle CI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,10 @@ RUN ./semgrep-core/_build/install/default/bin/semgrep-core -version
 FROM python:3.7.7-alpine3.11
 LABEL maintainer="support@r2c.dev"
 
+# ugly: circle CI requires a valid git command when running semgrep
+# on a repository containing submodules
+RUN apk add --no-cache git
+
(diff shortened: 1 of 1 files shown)


aryx:
On my machine the docker images goes from 172MB to 188MB when adding git.

brendongo:
Maybe shouldn't use the semgrep docker image as the base image in circleCI but instead pick a base image with git installed and run `pip3 install semgrep==0.23.0` and run semgrep.

aryx:
I don't want 0.23, I want the latest! so it's easier to dogfood new feeatures.

aryx:
But yeah, we should maybe build another docker image for circleCI for regular people.
I just want a way to make a docker image I can use in the CI of pfff and semgrep with the very latest
semgrep, one that matches develop.

PR #1676, opened 2020-09-15
Use latest pfff

aryx:
This should fix the error regarding ounit2 and JSON in CI


aryx:
The failure in CI is due to the new constant propagation fix of @IagoAbal genering an error in semgrep-rules testsuite. I'll fix those in a separate diff. The circle CI error is because of the lack of git in the docker develop image which this PR should fix.

PR #1677, opened 2020-09-15

aryx:


aryx:
e2e test error should be fixed by https://github.com/returntocorp/semgrep-rules/pull/861

PR #1680, opened 2020-09-15
enable typescript extensions

DrewDennison:


semgrep/semgrep/target_manager.py
```diff
@@ -54,8 +55,10 @@ def lang_to_exts(language: Language) -> List[FileExtension]:
     """
     if language in {"python", "python2", "python3", "py"}:
         return PYTHON_EXTENSIONS
-    elif language in {"js", "javascript"}:
+    elif language in {"js", "jsx", "javascript"}:
         return JAVASCRIPT_EXTENSIONS
+    elif language in {"ts", "tsx" "typescript"}:
```
nbrahms, line 60: Do we need to filter out `.d.ts`?

nbrahms (approved):

semgrep/semgrep/target_manager.py
mjambon: Yes, we need to filter out `.d.ts` (not 100% sure if it should ever work, but it doesn't work at the moment). A hack I used to produce parsing stats is to treat `.d.ts` as the extension of `foo.d.ts`. I think it violates the usual meaning of "extension". Applying a more generic filter on file names seems like a proper solution here.

aryx:
I usually try in my commit/PR to have a test plan section that shows how to test the new feature.
Here I guess maybe  semgrep -l ts -e '$X' /path/to/tsfile?

PR #1683, opened 2020-09-16
Add TSX support

mjambon:
This adds support for the the tree-sitter tsx parser:

* typescript and tsx each have their own, quasi-identical parser.
* parsing results in slightly different CSTs.
* those CSTs are merged into a single CST type which we'll have to maintain by hand. This is the file `semgrep-core/parsing/CST_tree_sitter_typescript.ml` which a manual merge of the two generated `CST.ml` files.

Depends on pfff PR https://github.com/returntocorp/pfff/pull/294

Note: I haven't tested any of this. It builds. I'll run basic tests tomorrow.


aryx (approved):
Again i would have just merged Lang.TSX with typescript but maybe it could help in the futur to separate them.

PR #1685, opened 2020-09-16
Add env variable for appending to user agent

mschwager:
This enables us to specify a `SEMGREP_USER_AGENT_APPEND` environment variable. We can use this to better understand which environments requests are coming from or even better debug support requests. Naturally this will be opt-in for users by specifying the env variable.


Dockerfile
```diff
@@ -61,6 +61,7 @@ RUN adduser -D -u 1000 semgrep
 USER 1000
 ENV SEMGREP_IN_DOCKER=1
 ENV SEMGREP_VERSION_CACHE_PATH=/tmp/.cache/semgrep_version
+ENV SEMGREP_USER_AGENT_APPEND="(Docker)"
```
minusworld, line 64: Should this be `None` so that we don't get faulty data in non-Docker environments?

minusworld (dismissed):

Dockerfile
mschwager: I'm not following. Can you elaborate? This only affects the Semgrep Docker image/container's environment, so there shouldn't be any faulty data elsewhere. In fact, it can even be overwritten at runtime with `docker run -e SEMGREP_USER_AGENT_APPEND=...`.

minusworld: Oh, I see -- this is in the Dockerfile 🤦‍♂️  I thought it was in the code itself. 👍

minusworld (approved):

PR #1690, opened 2020-09-17
Add regression test for pfff#289

IagoAbal:
Follows: pfff@0ae0779e6 ("generic: Fix bug in constant propagation for Python (#289)")

### DIFF
--- a/semgrep-core/tests/python/constprop_used_twice.py
+++ b/semgrep-core/tests/python/constprop_used_twice.py
@@ -1,0 +1,13 @@
+import flask
+from flask import response as r
+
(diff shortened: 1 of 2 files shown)


aryx (approved):
Looks good.
I try to follow an informal naming strategy in those tests where the prefix of the test is a big feature category (metavar, dots, equivalence) and misc_ for all the other stuff.
The naming strategy is partially used to generate the cheatsheet for each language (see the scripts/ dir)).  So those test maybe should be renamed equivalence_const_prop_used_twice.py but it's ok for now.

Issue #1237, opened 2020-07-10
Make --config take an array of args or be allowed to be passed in multiple times


DrewDennison:
so you can pass in multiple flags or an array to --config and get perf speed ups from only having to parse once.

DrewDennison:
@colleend @brendongo is this really finished? PRs seem closed not merged

colleend:
ah yes, just linked the merged PR

PR #1693, opened 2020-09-17
adding support for running multiple YAML urls with flag 'configs'

colleend:
Feature request addition for issue #1237. 

Added a flag, --configs that allows you to input a list of configuration files separated by a comma (NO SPACE AFTER). 

**Test Plan**
Can test with python -m semgrep --configs \<rule1\> \<rule2\> \<rule3\> ...:
Ex:  python -m semgrep --configs https://semgrep.dev/s/0gZr https://semgrep.dev/s/NxvL

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -59,7 +59,6 @@ def cli() -> None:
         "-f",
         "--config",
         nargs='+',
-        type=str,
         help=(
             "YAML configuration file, directory of YAML files ending in "
             ".yml|.yaml, URL of a configuration file, or semgrep registry entry "
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -47,12 +47,6 @@ def cli() -> None:
     # config options
     config = parser.add_argument_group("config")
     config_ex = config.add_mutually_exclusive_group()
-    config_ex.add_argument(
-        "--configs",
-        help=("List of configuration files seperated by comma. The type of permitted files is "
-              "specified in the help section of --config."
-              "Separate like \"<url1>,<url2>\" with no space in between."),
-    )
 
     config_ex.add_argument(
         "-g",
@@ -64,6 +58,8 @@ def cli() -> None:
     config_ex.add_argument(
         "-f",
         "--config",
+        nargs='+',
+        type=str,
         help=(
             "YAML configuration file, directory of YAML files ending in "
             ".yml|.yaml, URL of a configuration file, or semgrep registry entry "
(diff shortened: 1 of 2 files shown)


aryx:
I usually like to have in the commit message or PR a section called
Test plan:
where I describe how to test the change (to check that it actually works).
Here you linked in the PR the related issue, but I still think it's better to repeat the test plan in the commit message (later people can use git blame on the file and seeing a test plan help in my experience).

colleend:
Good point, I might write some unit tests for the flag too 🧪

aryx:
unit tests are better, but I'm ok with just a Test plan.

colleend:
Awesome, write a brief test plan in the comment, let me know if I should do anything else 👍

colleend:
adding nargs to the --config results in regression in performance, so I'm adding the --configs flag back.

mschwager:
> adding nargs to the --config results in regression in performance, so I'm adding the --configs flag back.

Could you elaborate on the regression in performance? I wouldn't expect it to behave any differently, we're essentially just adding a `for` loop. Obviously if we specify multiple configs it will take more time because we're running multiple analyses, but I don't _think_ that's what's going on here.

After briefly looking at the CI failures it looks like something was off with the argument processing, which may mean there was a bug in the multi-config implementation.

The advantage I see to `--config` instead of `--config` and `--configs` is we provide a simpler interface to users. They don't have to change or remember anything to run multiple configurations - it's just multiple values passed to the same argument. It's also simpler to maintain on our end. Instead of multiple arguments to consider it's just a `for` loop over the configs, whether it's 1 config or 5 configs. Thoughts?

aryx:
I am tempted to vote too for using multiple --config url1 --config url2 instead of the separate --configs, but at the same time I don't know any other command-line tools that allows to accumulate things by using the same flag. Actually inn OCaml the default CLI library will just pick the last one. If you do semgrep-core -lang php -lang c ... then the selected language will be 'c'.

brendongo:
FWIW it looks like it failed functionality tests and not performance tests in the config nargs commit

brendongo:
Looks like what we want is not nargs but the "append" action https://docs.python.org/3/library/argparse.html#action

But looking into this more, we would want to parse all the config files into a single datastructure of rules to run instead of looping over the config files cause if we do the later we don't get any performance benefit and end up parsing all the files NUM CONFIG number of times anyway so this is effectively invoking semgrep multiple times. Might be worth refactoring config logic cause I'm not entirely sure what config_ids are and what assumptions our codebase makes about them.

brendongo (changes requested):
-

brendongo:
Used append action on https://github.com/returntocorp/semgrep/pull/1730

Issue #1705, opened 2020-09-21
Typescript not recognized


msorens:
**Describe the bug**
Only `ts` is working for typescript. Neither `tsx` nor `typescript` are recognized, as they are supposed to be.

```
$ semgrep -f mysemgrep
error: invalid language
14 |     message: "onDestroy not implemented"
15 |     languages: [typescript]
   |                ^^^^^^^^^^^^
16 |     severity: ERROR

unsupported language typescript
```

Reason--there is a missing comma:
https://github.com/returntocorp/semgrep/blob/d3fe469c7edc44f179cf7754836ed5f34d073614/semgrep/semgrep/target_manager.py#L61

Change this:
```
elif language in {"ts", "tsx" "typescript"}:
```
To this:
```
elif language in {"ts", "tsx", "typescript"}:
```

aryx:
Ahah, semgrep can find this bug now!
+ /home/pad/semgrep/_build/default/bin/Main.exe -lang python -e {..., "..." "...", ...} semgrep/
semgrep/semgrep/target_manager.py:61
     elif language in {"ts", "tsx" "typescript"}:
semgrep/semgrep/target_manager.py:61
     elif language in {"ts", "tsx" "typescript"}:

aryx:
@mschwager @minusworld a new python rule to put in semgrep-rules/ ?
{..., "..." "....", ...} is probably a bug with a missing comma typo.

aryx:
I'll add it in the local semgrep.yml

aryx:
Should work now.

PR #1707, opened 2020-09-21
Use new python rule to detect wrong string concatenation in sets

aryx:
This should help detect bugs like the one reported in
 https://github.com/returntocorp/semgrep/issues/1705

test plan:
make check
wait for CI to see if circleCI spot the reported bug above.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,13 +8,7 @@ jobs:
     working_directory: /src
     steps:
       - checkout
-      - run:
-          name: "Pull Submodules"
-          command: |
-            apk add git
-            git submodule init
-            git submodule update --remote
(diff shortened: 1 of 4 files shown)


aryx:
good job circleCI, reported the bug in 16s!

mschwager:
We have an official rule for this: https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/correctness/common-mistakes/string-concat-in-list.yaml. Should we instead update it to include `set`s instead of using this local version?

aryx:
Sure, we should extend it to sets! I'll use the new --configs of @colleend after so I can run the R2C rules and the local semgrep.yml rules in CI.

PR #1708, opened 2020-09-21
docs/pattern-features.md: fix golang snippet

srenatus:
I'm not 100% certain that this was the original intention behind the code
snippet, but it seems probable to me. The code that was there previously wasn't
valid.


mschwager:
Good catch, thank you!

mschwager (approved):

Issue #1691, opened 2020-09-17
Matching arrow functions regardless of storage scope 


mgaudet:
**Describe the bug**

I would expect that `(...) => { ... }` would match the following lines

```
var k = (a,...rest) => { return 10; }
var j = (...rest) => { return 12; }
var l = (a,b) => rest[0]; 
var i = (a) => { return 10; }
```

However, this doesn't appear to work right now. Yet, if you drop the `var`, it -does- seem to work. 

*edit*: Oh wild: I added a `let` binding at the bottom, and the sense of what works flipped: Now the `var`s work, but the `var`less don't. https://semgrep.dev/s/q6WR/

**To Reproduce**
https://semgrep.dev/s/Kd1j/

**Expected behavior**

Scope doesn't affect arrow matches. 

**Screenshots**

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?

aryx:
It's because var foo = (a) => { stuff }
was considered like an alternative way to define a function, so this is really transformed internally
in function foo(a) { stuff }, which is then not considered an anonymous function (a lambda).
I probably did that because people use the short arrow syntax to define toplevel functions, but I should probably restrict the internal transformation only when people use  const foo = (a) => { stuff }.
I'll fix it.

PR #1710, opened 2020-09-21
Test file for arrow pattern matching arrows stored in vars

aryx:
Arrows in vars used to be consider toplevel functions, when really
they should still be considered lambdas.

test plan:
test files included
make test


aryx:
This closes https://github.com/returntocorp/semgrep/issues/1691

Issue #1697, opened 2020-09-19
[Javascript] Match on decorators


aryx:
We support attributes in Java, OCaml, Python, etc.. Javascript now has "decorators" which are similar.

PR #1722, opened 2020-09-23
More TODOs in Parse_typescript_tree_sitter.ml

aryx:


aryx:
This should help https://github.com/returntocorp/semgrep/issues/1697

Issue #1724, opened 2020-09-23
RFC: Add primitive for mathematical comparisons of metavariable


mschwager:
In an attempt to further reduce our reliance on `pattern-where-python`, I'd like to propose adding a new primitive for performing basic mathematical comparisons on metavariables. First let's look at how `pattern-where-python` is used in [`semgrep-rules`](https://github.com/returntocorp/semgrep-rules):

<details> 
  <summary>Semgrep rules examples</summary>

```
$ rg 'pattern\-where\-python' -A 5
contrib/dlint/redos.yaml
6:  - pattern-where-python: |
7-      # COPIED FROM https://github.com/dlint-py/dlint/blob/master/dlint/redos/detect.py
8-
9-      import sre_constants
10-      import sre_parse
11-

go/gorilla/security/audit/handler-attribute-read-from-multiple-sources.yaml
41:  - pattern-where-python: vars['$KEY'].strip('"') == vars['$ATT']
42-  message: "Attribute $ATT is read from two different sources: '$X[$KEY]' and '$Y.$ATT'. Make sure this is intended, as this\
43-    \ could cause logic bugs if they are treated as if they are the same object."
44-  languages: [go]
45-  severity: WARNING

java/lang/security/audit/crypto/weak-rsa.yaml
17:  - pattern-where-python: |-
18-      int(vars['$BITS']) < 2048

go/lang/correctness/overflow/overflow.yaml
11:  - pattern-where-python: int(vars['$NUM'].replace('"', '')) > 32767 or int(vars['$NUM'].replace('"', '')) < -32768
12-- id: interter-overflow-int32
13-  message: Potential Integer overflow made by strconv.Atoi result conversion to int32
14-  languages: [go]
15-  severity: WARNING
16-  patterns:
--
21:  - pattern-where-python: |-
22-      int(vars['$NUM'].replace('"', '')) > 2147483647 or int(vars['$NUM'].replace('"', '')) < -2147483648

go/lang/correctness/permissions/file_permission.yaml
12:  - pattern-where-python: |
13-      int(vars['$PERM'], 8) > 0o600
14-- id: incorrect-default-permission
15-  message: Expect permissions to be `0600` or less for ioutil.WriteFile()
16-  metadata:
17-    cwe: 'CWE-276: Incorrect Default Permissions'
--
24:  - pattern-where-python: |
25-      int(vars['$PERM'], 8) > 0o600
26-- id: incorrect-default-permission
[...]

PR #1728, opened 2020-09-24
Proof of concept of semgrep_core -eval <jsonfile> for pattern-where-python

aryx:
This is a possible solution to https://github.com/returntocorp/semgrep/issues/1724

test plan:
```
$ cat simple_expr.json
{
  "metavars": {
    "$X": "42",
    "$Y": "\"foo\"",
    "$Z": "x+1"
  },
  "language": "python",
  "code": "$X > 0 and $Y == \"foo\""
}
$ /home/pad/semgrep/_build/default/bin/Main.exe -eval simple_expr.json
true

### DIFF
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -924,6 +924,8 @@ let all_actions () = [
   Common.mk_action_2_arg Datalog_experiment.gen_facts;
   "-dump_il", " <file>",
   Common.mk_action_1_arg Datalog_experiment.dump_il;
+  "-eval", " <JSON file>",
+  Common.mk_action_1_arg Eval_generic.eval_json_file;
  ]
 
 (*e: function [[Main_semgrep_core.all_actions]] *)
--- a/semgrep-core/matching/Eval_generic.ml
+++ b/semgrep-core/matching/Eval_generic.ml
@@ -159,7 +159,8 @@ and eval_op op values code =
   | G.Div, [Int i1; Int i2] -> Int (i1 / i2)
   | G.Mod, [Int i1; Int i2] -> Int (i1 mod i2)
 
-  | G.Eq, [String s1; String s2] -> Bool (s1 = s2)
+  | G.Eq, [v1; v2] -> Bool (v1 = v2)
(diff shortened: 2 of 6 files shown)


aryx:
@mschwager try to "hack" this eval. Any non-trivial input will be rejected.

ievans (approved):
This is amazing @aryx. 🤯 

Can we support <, <=, >=, and 'x in [1,2,3]' operators? All but the last should be trivial. I think this would let us replace pretty much all the one-line `pattern-where-python` operators.

aryx:
Done, you can now use $X in [1,2,3] and a few more operators.

mschwager:
Nice! This is excellent - I have a few questions:

From our discussion on the ticket, this is manually evaluating basic Python expressions? From what I can tell, since there's no Python runtime it's impossible to get any kind of complex code execution.

Two of our most common use-cases are: 1.) [strings that look like integers](https://github.com/returntocorp/semgrep-rules/blob/develop/go/lang/correctness/overflow/overflow.yaml#L11), e.g. `"2147483647"` and 2.) [comparisons on integers of different bases](https://github.com/returntocorp/semgrep-rules/blob/develop/go/lang/correctness/permissions/file_permission.yaml#L13), e.g. `0700 > 0o600`. Is there a good way to handle these cases with this functionality? This is where the `strip` and `base` arguments came in in the original ticket.

Also, I think we can avoid handling string operations. Between the OCaml PCRE regexs, `metavariable-regex`, and basic metavariable string comparison (e.g. `$METAVAR="password"`) I think we're covered here. It's not necessarily hurting anything, but it would be less maintenance, less potential for abuse, less duplicated functionality.

aryx:
Yes it's impossible to get any kind of complex code execution. It just interprets simple arithmetic expressions (and also
the in python operator with lists).

Right now if the metavar contains an integer (e.g., $X in the simple_expr.json in the PR), it is automatically converted to
an int if the int_of_string OCaml function does not return an exception. Note that the lexical syntax for numbers in Python
and OCaml is pretty much the same, so I think int_of_string will handle numbers written in decimal or octal or hexa,
so comparison such as 0o700 > 0o600 should work).
If it's needed, we can augment the intepreter to handle a few python builtins.

PR #1729, opened 2020-09-24
Update javascript and typescript parsers for latest tree-sitter grammars

mjambon:
Relies on https://github.com/returntocorp/pfff/pull/308


semgrep-core/parsing/Parse_javascript_tree_sitter.ml
```diff
@@ -715,19 +712,18 @@ and class_body (env : env) ((v1, v2, v3) : CST.class_body) =
...
-    )
+  let v1 = anon_choice_exp env v1 in
+  let v2 =
+    match v2 with
+    | `DOT tok (* "." *)
+    | `QMARKDOT (* "?." *) tok -> token env tok
   in
-  let v2 = token env v2 (* "." *) in
   let v3 = identifier env v3 (* identifier *) in
   ObjAccess (v1, v2, PN v3)
```
aryx, line 729: I guess at some point we will need to extend ast_js.ml to describe more precisely this ?. construct instead
of agglolmerating both in ObjAccess.

semgrep-core/parsing/Parse_javascript_tree_sitter.ml
```diff
@@ -874,20 +871,30 @@ and constructable_expression (env : env) (x : CST.constructable_expression) : ex
...
+                  Apply (IdSpecial (Encaps true, t1),
+                         (t1, v1::xs, t2))
+             )
+           in
+           v2
+       | `Choice_this_QMARKDOT_args (v1, v2, v3) ->
+           let v1 = primary_expression env v1 in
+           let _v2 = token env v2 (* "?." *) in
+           let v3 = arguments env v3 in
+           Apply (v1, v3)
```
aryx, line 898: again, probably need a new construct instead of putting in Apply.

semgrep-core/parsing/Parse_javascript_tree_sitter.ml
```diff
@@ -1024,6 +1031,9 @@ and expression (env : env) (x : CST.expression) : expr =
         | `GTGTGTEQ tok -> G.ASR, token env tok (* ">>>=" *)
         | `LTLTEQ tok -> G.LSL, token env tok (* "<<=" *)
         | `STARSTAREQ tok -> G.Pow, token env tok (* "**=" *)
+        | `AMPAMPEQ tok -> G.LogAndAss, token env tok (* "&&=" *)
```
aryx: AST_generic.And is the logical and. (BitAnd is the bitwise and '&').

aryx (changes requested):
No need for LogAndAss.
Otherwise looks ok (it's a big diff. Hopefully you just moved around things
and the compiler did the rest).

semgrep-core/parsing/Parse_javascript_tree_sitter.ml
mjambon: I fixed this for js and ts. It should transpile into the right thing now.

aryx (approved):

semgrep-core/parsing/Parse_javascript_tree_sitter.ml
mjambon: I added some to-dos. I suggest we take care of this later. At this point `?.` is simply treated like normal access in all 3 cases (property access, function call, array access).

mjambon:
> (it's a big diff. Hopefully you just moved around things and the compiler did the rest).

Yes, the real changes were only about the points you raised already (optional chaining of 3 kinds for both js and ts, and the logical assignment operators).

The new ocaml-tree-sitter should produce more stable output from now on, making clearer diffs and easier updates.

PR #1730, opened 2020-09-26
Allow multiple configs

### LINKED ISSUE #1237: Make --config take an array of args or be allowed to be passed in multiple times

brendongo:
Multiple config files can now be passed to semgrep cli `semgrep --config foo.yaml --config bar.yaml`

Also refactors config handling logic to it's own class

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -58,6 +58,7 @@ def cli() -> None:
         "-f",
         "--config",
         action="append",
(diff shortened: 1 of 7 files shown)


aryx:
Looks good. How does it relate to the one @colleend initially wrote?

colleend (approved):
Looks good to me 👍

PR #1731, opened 2020-09-26
Rename 'anon_' types

mjambon:
This is just a mechanical replacement of generated names.


aryx (approved):
Great. Emacs macros again to automate?

mjambon:
> Emacs macros again to automate?

No. I had one window with the diffs on `CST.ml`, I grabbed each new type name (left-hand side of type definitions) and replaced the old one with the new (search and replace), manually. I did consider automating the repetitive search-and-replace but it would have required solid regexps and I wasn't in the mood. Maybe I should have made the effort.

Issue #1562, opened 2020-08-21
Report basic statistics when run with no flags


disconnect3d:
It would be nice if Semgrep reported a number of INFO/LOW/WARN/ERROR findings when its simply run with `semgrep [-f <config>]`, probably as the last line it prints out.

It seems some statistics are displayed when `--json` is invoked thx to https://github.com/returntocorp/semgrep/pull/302, but that's not very convenient.

The proposed solution would make it more convenient to develop rules based on big codebases.

ievans:
Other suggestions I've heard for the default output:

* the # of files, # of bytes, or # of lines of code processed. This can be useful if you run and think that you're scanning a lot of code, but it's actually mostly under a `.gitignore`.

ievans:
@disconnect3d can you share more about your reasoning for this? I'm trying to decide whether to put it in the default output or behind a flag like `--verbose`

disconnect3d:
@ievans Sure! I sometimes develop/adjust rules on big codebases and I want to be certain that a given change to reduce false positives did not exclude any true positives. Same for the opposite: if I change a rule to be more general, I want to know right away if I hit more results.

ievans:
@disconnect3d for that, is it sufficient to just report # of findings, or is the INFO/WARN/ERROR count meaningful to you?

disconnect3d:
The `# of findings` would be sufficient for my needs, though I bet the `INFO/WARN/ERROR` would be (more) interesting for running Semgrep on CI/CD in order to e.g. fail on errors but also display info/warn findings?

ievans:
Closed in #1732, cc @underyx for the suggestion about putting this into semgrep-action (the CI integration!)

Issue #1551, opened 2020-08-20
Parser error output does not report full path to file with parse error


ievans:
Parser error output does not report full path to file with parse error:

```
ine@imbp4 ~/D/r/t/highcharts (master) [2]> semgrep -e '$X==$X' --lang python --lang javascript
warn: parse error
  --> callback.js:1
...
```

But there is no callback.js at root. 

```
find . -name "callback.js"
./exporting-server/java/highcharts-export/highcharts-export-web/target/highcharts-export-web/WEB-INF/jspf/callback.js
```

aryx:
+1 on that.

Issue #1550, opened 2020-08-20
Clean up parser warning UX


ievans:
I'm looking for feedback on the UX of the final semgrep output. I propose the following. When run in an interactive shell, semgrep will print the following final line.

If (internal tool) warnings exist:

```
current> Warnings exist. Run with `--strict` to turn warnings into errors.
proposed> Ran 37 rules over 3817 files: 0 findings
3 files could be not analyzed: run with `--verbose` for details; run with `--strict` to exit non-zero if one or more files cannot e analyzed
```

If (internal tool) warnings do not exist (related to #1562):

```
current> no output
proposed> Ran 37 rules over 3820 files: 0 findings 
proposed> Ran 37 rules over 3820 files: 42 findings (2 info, 3 warn, 37 error)
```

The tricky thing about this text is differentiating _internal tool errors_ from _rules labeled error_.

ievans:
cc @brendongo @DrewDennison @aryx @sabihb @dlukeomalley for thoughts

clintgibler:
I like what you've proposed.

> Warnings exist. Run with `--strict` to turn warnings into errors.

This is unclear whether semgrep found bugs or had some internal problems.

```
proposed> Ran 37 rules over 3817 files: 0 findings
3 files could be not analyzed: ...
```

I like this: printing how many rules were ran over how many finds and a quick summary of findings tells you it's working, or at least points out when you're doing something wrong ("huh, I expected to be running 100 rules but only ran 3").

Can we print more context around **why** files couldn't be analyzed? "Could not be analyzed" makes me wonder, "Well, why not?" At the very least we could mention something about it being due to an error parsing source code files.

```
proposed> Ran 37 rules over 3820 files: 42 findings (2 info, 3 warn, 37 error)
```
I love the extra context here.

As I was thinking about this I had this idea about making it easier to report parser bugs. I think we've already worked on this, but I think we can do more: https://github.com/returntocorp/semgrep/issues/1652.

aryx:
sgtm too. I like the proposed output.

aryx:
Who can do the work? @brendongo @mschwager who is familiar with this part of the codebase?

Issue #384, opened 2020-04-06
Provide debugging context in sgrep-lint error messages


nbrahms:
Currently, sgrep-lint error messages can be difficult for a user to develop with.

We should improve these sgrep-lint error messages to:
- [ ] Detect and indicate possibly bad YAML formatting (e.g. dedenting a "pattern" key outside a "patterns" block)
- [ ] Indicate expected vs. actual YAML schema
- [ ] Clearly indicate which layer of code the error occurs within (sgrep-lint vs. sgrep)

dlukeomalley:
Blocked on https://github.com/returntocorp/sgrep/issues/383

nbrahms:
> Blocked on #383

I don't actually think this should be blocked on #383, unless we want to enforce a unified error-reporting format.

dlukeomalley:
Rereading, I agree with you. There is a desire to have more meaningful error messages in sgrep-lint, that will come from sgrep itself. That's originally why I added `blocked`.

rcoh:
Mostly fixed by #773

nbrahms:
cc. @mschwager

ievans:
@nbrahms looks like this is pretty much done now?

ievans:
I think these are fixed in https://github.com/returntocorp/semgrep/pull/1732. @nbrahms feel free to re-open

PR #1732, opened 2020-09-26
Clean up parser warning UX

ievans:
- [x] Added a `--debug` flag and moved most of the output under `--verbose` to it. 
- [x] Parse and other errors are mentioned at final output, but not individually displayed unless `--verbose` is passed

- [x] Explicitly mark semgrep errors as being "semgrep internal" (Closes https://github.com/returntocorp/semgrep/issues/384)
```
semgrep warn: parse error
  --> invalid_python.py:1
1 | def foo(a)
  |           ^^^^^
Could not parse invalid_python.py as python
```

- [x] Closes https://github.com/returntocorp/semgrep/issues/1550
```
ran 79 rules on 77 files: 14 findings
```

- [x] Closes https://github.com/returntocorp/semgrep/issues/1562 (currently only under `--verbose`)
```
findings summary: 7 warning, 7 error, 0 info
```

- [x] Closes https://github.com/returntocorp/semgrep/issues/1551
```
semgrep warn: parse error
  --> tests/e2e/targets/bad/invalid_python.py:1
1 | def foo(a)
  |           ^^^^^
Could not parse invalid_python.py as python
```

- [x] We display Level.ERROR semgrepErrors by default still; it's just parse, OOM, timeout errors, etc. (anything with .WARN) that now are not individually listed without `--verbose`

Outstanding issues: assumes that # files could not be analyzed == # of errors returned from semgrep core. Probably not a show-stopper.


semgrep/semgrep/core_runner.py
```diff
@@ -498,6 +501,7 @@ def _run_rules(
                             max_timeout_files.append(err.path)
 
         all_errors = dedup_errors(all_errors)
+        # logger.debug(all_targets)
```
brendongo:
```suggestion
```

semgrep/semgrep/output.py
```diff
@@ -399,7 +409,10 @@ def build_output(self, color_output: bool) -> str:
         elif output_format == OutputFormat.SARIF:
             return build_sarif_output(self.rule_matches, self.rules)
         elif output_format == OutputFormat.TEXT:
-            return "\n".join(build_normal_output(self.rule_matches, color_output))
+            return "\n".join(
+                list(build_normal_output(self.rule_matches, color_output))
+                + ([self.stats_line] if self.stats_line else [])
```
brendongo: Some users expect no output to be printed when no matches are found. IMO a stats line should be under --verbose and output should only ever be findings.

brendongo (changes requested):

semgrep/semgrep/output.py
ievans: Even in an interactive session @brendongo? I agree if run under a headless shell but for interactive it makes more sense that it tells you how many files it ran on and that there were 0 findings. I get that that differs from grep, but I'm trying to think about the non-poweruser UX.

aryx (commented):
Great! Thx for fixing all of that.

semgrep/semgrep/output.py
```diff
@@ -399,7 +409,13 @@ def build_output(self, color_output: bool) -> str:
         elif output_format == OutputFormat.SARIF:
             return build_sarif_output(self.rule_matches, self.rules)
         elif output_format == OutputFormat.TEXT:
-            return "\n".join(build_normal_output(self.rule_matches, color_output))
+            stats_line = []
+            if sys.stderr.isatty():
```
brendongo: Still not sure why we are passing stats line to output and gating tty instead of outputting to stderr like all other non finding output.

brendongo: Relevant slack convo: https://returntocorp.slack.com/archives/C6A1GN4P3/p1601598463387000?thread_ts=1601596568.384500&cid=C6A1GN4P3

brendongo: It's also less code: call logger.info here: https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/output.py#L346 ?

brendongo (approved):

PR #1733, opened 2020-09-26
update cheatsheet generator

ievans:


semgrep-core/tests/generate_test_matrix.py
```diff
@@ -65,7 +65,7 @@
     ],
 }
 
-EXCLUDE = ["TODO", "GENERIC", "fuzzy", "lint", 'EQUIV', 'e2e', 'SYNTHESIZING', 'NAMING', 'PERF', 'TAINTING']
+EXCLUDE = ["TODO", "GENERIC", "fuzzy", "lint", 'EQUIV', 'e2e', 'SYNTHESIZING', 'NAMING', 'PERF', 'TAINTING', 'EVAL']
```
aryx, line 68: Maybe i should put all those dirs under a NOT_A_LANG durectory

aryx (approved):

semgrep-core/tests/generate_test_matrix.py
ievans: yeah that's probably a good idea, or put the languages under lang/ or something

PR #1738, opened 2020-09-29
Use dune-build-info instead of script for bin/version.ml

aryx:
Feel free to reject this diff. It's just that i'm slightly annoyed
when I type make that it each time generates this bin/version.ml.

Fortunately, this does not cause the recompilation of the binary (not sure
what dune is doing; maybe it's computing md5sum of the file to see if it
actually changed, because 'touch bin/version.ml' does not cause any
recompilation).

test plan:
before:
$ /home/pad/semgrep/_build/default/bin/Main.exe -version
semgrep-core version: v0.24.0-31-g0bcf6652, pfff: 0.42

after:
$ /home/pad/semgrep/_build/default/bin/Main.exe -version
semgrep-core version: n/a, pfff: 0.42

after make install
$ semgrep-core -version
semgrep-core version: 0.2.0-1126-g0bcf6652-dirty, pfff: 0.42

note that you get the version only when you install. Also
dune-build-info does not use git describe ... --tags so you
get this 0.2.0 (not sure where it comes from).
Note also I don't understand why before we we getting
v0.24.0 when the last tag was actually 0.25.0.

In theory dune-build-info can leverage version information in the dune-project
but I didn't use that yet.
Like I said, feel free to reject this diff.


brendongo:
It looks like the semgrep-core build artifact is listed as version n/a https://github.com/returntocorp/semgrep/runs/1182909215?check_suite_focus=true#step:6:6 
even though the install script calls make install https://github.com/returntocorp/semgrep/blob/develop/scripts/install-alpine-semgrep-core#L54 are we copying the wrong executable maybe?

aryx:
yep, we copy the built semgrep-core binary in _build/ instead of the one installed in ~/.opam/.../bin/semgrep-core/
They used to be the same but now I guess with dune-build-info the 'make install' does more stuff.
Good catch, I should update the install-alpine-semgrep-core last line

aryx:
Let's abandon this. Does not provide much value.

PR #1739, opened 2020-09-29
Move non-languages tests under semgrep-core/tests/OTHER

aryx:
All those uppercase directories mixed with languages were ugly anway.
This simplifies also the generate_matrix cheatsheet script

test plan:
make test


ievans (approved):
thanks @aryx!

PR #1758, opened 2020-09-30
Nathan/fix lints

nbrahms:


brendongo (approved):

mschwager:
Do we know why these issues weren't being caught in CI?

semgrep/semgrep/test.py
```diff
@@ -215,9 +219,11 @@ def _generate_line(test_file, matches):
     )
 
 
-def invoke_semgrep_multi(filename, *args, **kwargs):
+def invoke_semgrep_multi(
+    filename: Path, *args: Path, **kwargs: Any
+) -> Tuple[Path, Optional[Exception], Any]:
     try:
-        output = invoke_semgrep(filename, *args, **kwargs)
+        output = invoke_semgrep(filename, list(args), **kwargs)
```
mschwager: These two pieces of code are not strictly equivalent. Are you sure this is what you want?

nbrahms: I can't figure out how this code was working before TBH, but `invoke_semgrep` definitely takes a list in that position.

mschwager: If we look at `config_with_tests` we can see it contains data like:

```python
ipdb> pp config_with_tests                                                                           
[(PosixPath('~/r2c/semgrep-rules/python/requests/security/no-auth-over-http.yaml'),
  [PosixPath('~/r2c/semgrep-rules/python/requests/security/no-auth-over-http.py')]),
 (PosixPath('~/r2c/semgrep-rules/python/requests/security/disabled-cert-validation.yaml'),
  [PosixPath('~/r2c/semgrep-rules/python/requests/security/disabled-cert-validation.py')]),
 (PosixPath('~/r2c/semgrep-rules/python/requests/best-practice/use-timeout.yaml'),
```

With [`starmap`](https://docs.python.org/3/library/multiprocessing.html?highlight=multiprocessing#multiprocessing.pool.Pool.starmap) we pass these in as arguments to `invoke_semgrep_multi`. E.g.

* `filename` -> `PosixPath('~/r2c/semgrep-rules/python/requests/security/no-auth-over-http.yaml')`
* `*args` -> `[PosixPath('~/r2c/semgrep-rules/python/requests/security/no-auth-over-http.py')]` (but it could be more with additional positional arguments passed in)

`invoke_semgrep_multi` was meant to be a loose wrapper around `invoke_semgrep` with `multiprocessing`, so I went with `*args` and `**kwargs`. So it is true that `list(args)` will work, but that wasn't really the intent here. If someone passes in an additional positional argument it won't work as expected. I think you want `*args: Any` and no `list` in the `invoke_semgrep` call.

mschwager: We could've even included `filename` in the `*args` passthru here, but we needed to `return` it below.

nbrahms: Let's not fight with the typechecker. We control the signature and call of the method, so let's just use it explicitly.

nbrahms: FWIW I tested this on `semgrep-rules` and it does the right thing TM

PR #1761, opened 2020-10-01
Use own junit-xml fork

brendongo:
junit-xml 1.9 does not have a tarball published on pypi. This means
in homebrew formula we have to pin to a git resource which makes it
difficult to autobump the formula. This PR relies on a fork of
junit-xml that has the tarball published.

Relevant ticket for original project to publish tarball has no updates
after 6 months: https://github.com/kyrus/python-junit-xml/issues/69

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -48,7 +48,7 @@ ignore_missing_imports = True
 ignore_missing_imports = True
 
 # junit_xml
-[mypy-junit_xml]
+[mypy-junit_xml_2]
 ignore_missing_imports = True
 
 # tqdm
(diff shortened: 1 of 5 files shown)


brendongo:
Fork Source: https://github.com/returntocorp/python-junit-xml
Fork Pypi: https://pypi.org/project/junit-xml-2/

brendongo:
Based on @mschwager's suggestion might be easier to just pull relevant function from the package into semgrep instead of having to deal with dependencies. The relevant code is pretty straightforward dictionary parsing and is unlikely to change anyway.

brendongo:
Looks like homebrew-auto-pr works despite this now. https://github.com/Homebrew/homebrew-core/pull/62215 . Closing.

PR #1763, opened 2020-10-01
Attempt to handle parsing errors from tree-sitter.

mjambon:
It looks like I didn't add the exception catcher at the right spot because the original exceptions aborts the program:
```
$ ./_build/install/default/bin/semgrep-core -lang ts -e x toto.ts
toto.ts:1:0: Fatal Error: Tree_sitter_run.Tree_sitter_error.Error(_)

Fatal error: exception Tree_sitter_run.Tree_sitter_error.Error(_)
```
... but at least it compiles.

Hopefully my conversion from the `Tree_sitter_run.Tree_sitter_error.t` is correct. I had to add a field for holding the entire "token" (region of the error), which populates the `str` field on the semgrep side.

I'll let you take it from here or give me guidance on where/how to catch the exception.

This depends on https://github.com/returntocorp/ocaml-tree-sitter/pull/114

### DIFF
--- a/semgrep-core/dune
+++ b/semgrep-core/dune
@@ -1,5 +1,5 @@
 (env
 (dev
-(flags (:standard -w +42-52-6)))
+(flags (:standard -w -52-6)))
 (release
 (flags (:standard -O3))))
--- a/semgrep-core/bin/Main.ml
+++ b/semgrep-core/bin/Main.ml
@@ -582,6 +582,17 @@ let get_final_files xs =
   Common2.uniq_eff (files @ explicit_files)
 (*e: function [[Main_semgrep_core.get_final_files]] *)
 
+let mk_tree_sitter_error (err : Tree_sitter_run.Tree_sitter_error.t) =
+  let start = err.start_pos in
+  let loc = {
+    PI.str = err.substring;
+    charpos = 0; (* fake *)
+    line = start.row + 1;
+    column = start.column;
+    file = err.file.name;
+  } in
+  Error_code.({ loc; typ = ParseError; sev = Error })
+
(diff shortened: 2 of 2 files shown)


semgrep-core/dune
```diff
@@ -1,5 +1,5 @@
 (env
 (dev
-(flags (:standard -w +42-52-6)))
+(flags (:standard -w -52-6)))
```
aryx, line 3: Why removing 42?

aryx (approved):
Ok i ll take it from there

semgrep-core/dune
mjambon: This is what warning 42 tells us:
```
Warning 42: this use of x relies on type-directed disambiguation, it will not compile with OCaml 4.00 or earlier.
```

If 42 is enabled, the code gets really ugly when record fields from multiple modules need to be accessed. This was a problem for this code:

```ocaml
let mk_tree_sitter_error (err : Tree_sitter_run.Tree_sitter_error.t) =
  let start = err.start_pos in
  let loc = {
    PI.str = err.substring;
    charpos = 0; (* fake *)
    line = start.row + 1;
    column = start.column;
    file = err.file.name;
  } in
  Error_code.({ loc; typ = ParseError; sev = Error })
```
With warning 42, we'd have to `open` 2 modules (`Tree_sitter_run.Tree_sitter_error` and `Tree_sitter_run.Src_file`), or access fields as `err.Tree_sitter_error.substring` in combination with `let open Tree_sitter_run in`, or create short module aliases... even though the compiler and the reader know exactly which type we're dealing with from the type annotation on the function argument.

Issue #1713, opened 2020-09-21
[typescript] Probable bug in parsing


msorens:
**Describe the bug**

Running semgrep on my large-ish code base yielded these remarkably few probable parsing problems--nice job, folks!

These files compile cleanly and are currently running in production so, as far as I know, they should be parsable.
```
warn: parse error
  --> data-bags.requests.ts:1
1 | import { Injectable } from '@angular/core';
  | 
= help: If the code appears to be valid, this may be a semgrep bug.
Could not parse data-bags.requests.ts as ts

warn: parse error
  --> cookbook-details.component.ts:1
1 | import { Component, OnInit, OnDestroy } from '@angular/core';
  | 
= help: If the code appears to be valid, this may be a semgrep bug.
Could not parse cookbook-details.component.ts as ts

warn: parse error
  --> entities.ts:1
1 | import {
  | 
= help: If the code appears to be valid, this may be a semgrep bug.
Could not parse entities.ts as ts
```

**To Reproduce**

My codebase is open source so you can go right to, well, the source: https://github.com/chef/automate/tree/master/components/automate-ui
 

**Expected behavior**
No parsing errors.

**Environment**
Version 0.24.0

mjambon:
Thanks for the report, @msorens. I'll look into this.

aryx:
With the better error reporting I now get those parse error locations:
```
warn: parse error
  --> entities.ts:20
20 |   readonly [id: string]: T;
   |             ^^^
= help: If the code appears to be valid, this may be a semgrep bug.
Could not parse entities.ts as ts

warn: parse error
  --> data-bags.requests.ts:20
20 |   : Observable<DataBagDetailsSuccessPayload> {
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
= help: If the code appears to be valid, this may be a semgrep bug.
Could not parse data-bags.requests.ts as ts

Warnings exist. Run with `--strict` to turn warnings into errors.

```

aryx:
This can help in the futur to provide more information on the kind of constructs we do not parse well, which also makes it
easier to create simpler test file showing the issue (minimal test case).

msorens:
[...]

PR #1767, opened 2020-10-01
Convert tree-sitter parse error exn to Parse_info.Parsing_error

aryx:
There is already some code to automatically intercept Parse_info exns
to display their location, so better to use a consistent scheme
across pfff and tree-sitter parsers.

test plan:
```
(semgrep) pad@yrax:~/semgrep/tests/OTHER/parsing_errors$ semgrep -l ts -e 'FOO' .
warn: parse error
  --> err.ts:2
2 |          return 1+
  |          ^^^^^^^^^
Could not parse err.ts as ts

Warnings exist. Run with `--strict` to turn warnings into errors.
```

better than the previous error that was always reported on the first line.
This also improves error location for the errors reported in
https://github.com/returntocorp/semgrep/issues/1713

### DIFF
--- a/semgrep-core/parsing/Parse_tree_sitter_helpers.ml
+++ b/semgrep-core/parsing/Parse_tree_sitter_helpers.ml
@@ -92,6 +92,45 @@ let combine_tokens env xs =
       let t = token env x in
       t
 
+let mk_tree_sitter_error (err : Tree_sitter_run.Tree_sitter_error.t) =
+  let start = err.start_pos in
+  let loc = {
+    PI.str = err.substring;
+    charpos = 0; (* fake *)
+    line = start.row + 1;
+    column = start.column;
+    file = err.file.name;
+  } in
+  loc
+
+let convert_tree_sitter_exn_to_pfff_exn f =
+  try f ()
+  with
+  (* The case below is what we would like to do! However if
+   * you use Parallel.invoke to invoke the tree-sitter parser, this
+   * code below will never trigger. Indeed, unmarshalled exn
+   * can't be used in match or try or used for structural equality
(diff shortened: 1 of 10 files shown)


semgrep-core/parsing/Parse_tree_sitter_helpers.ml
```diff
@@ -92,6 +92,45 @@ let combine_tokens env xs =
...
+  | exn ->
+      let s = Common.exn_to_s exn in
+      if s = "Tree_sitter_run.Tree_sitter_error.Error(_)" then begin
+        let t = Obj.repr exn in
+        let info = Obj.field t 1 in
+        let (ts_error : Tree_sitter_run.Tree_sitter_error.t) = Obj.obj info in
+        let loc = mk_tree_sitter_error ts_error in
+        let info = { PI.token = PI.OriginTok loc; transfo = PI.NoTransfo } in
+        raise (PI.Parsing_error info)
+      end else raise exn
```
mjambon, line 131: To get the constructor, you can do `Printexc.exn_slot_name exn` (since ocaml 4.02). But I don't see how to do the rest without `Obj`, other than catching the exception in the child (or make the parser return an Ok/Error).

mjambon (approved):

semgrep-core/parsing/Parse_tree_sitter_helpers.ml
aryx: yep, the parser could return an algebraic data type. Only exn (open variants) have this marshalling issue.
Still, I like this hacky solution :)

aryx (commented):
But removing the use of Parallel.invoke will also solve the issue and let us use exceptions properly.

Issue #1190, opened 2020-07-02
Write CI documentation for Jenkins


underyx:
**Is your feature request related to a problem? Please describe.**

Our CI integration docs have many providers but are missing Jenkins.

**Describe the solution you'd like**

I think Jenkins might have a more complicated configuration than just committing a YAML file, and these docs would be more complex than others? Or maybe that changed in the past couple years.

dlukeomalley:
@underyx Is this an external request and/or near term needed?

underyx:
I assumed it was only cause it appears on r2c.dev

dlukeomalley:
Good eye. Note that there is a distinction in what we support as part of the OSS project and what we spend time supporting for our commercial users. I'll follow-up with you in person to talk through the distinction and how to communicate that clearly on our websites and project.

mschwager:
https://github.com/returntocorp/semgrep/blob/develop/docs/integrations.md#jenkins

PR #1770, opened 2020-10-01
Add Jenkins information to integrations

mschwager:


ievans (approved):
https://semgrep.dev/packs --> wrong url?

Also can we dedupe this with semgrep.dev?
Screen Shot 2020-10-01 at 1 18 46 PM

mschwager:
Fixes https://github.com/returntocorp/semgrep/issues/1190

> wrong url?

:+1: 

> Also can we dedupe this with semgrep.dev?

I agree that we should have a single source of truth. I'll defer to @underyx on this for semgrep.dev.

Since we're planning on a larger docs change effort, maybe we delay this until then?

PR #1772, opened 2020-10-02
Improved tags in Sarif output

hunt3rkillerz:
The SARIF output for the scans currently only outputs CWE and/or OWASP when those values are present in the rule meta-data. This change will use a shortened version of the CWE (cwe-xxx) and OWASP (owasp-AX) meta values as part of the SARIF output.

### DIFF
--- a/semgrep/semgrep/rule.py
+++ b/semgrep/semgrep/rule.py
@@ -261,25 +261,9 @@ class Rule:
         Tags to display on SARIF-compliant UIs, such as GitHub security scans.
         """
         if "cwe" in self.metadata:
-            # Handling unexpected formatting of CWE meta-data
-            try:
-                # Assumed CWE format is CWE-[0-9]+[:]?.*
-                cwe_id = self.metadata["cwe"].split("CWE-")[1]
-                cwe_tag= "cwe-"
-                for char in cwe_id:
-                    if char.isdigit():
-                        cwe_tag += char
-                    else:
-                        break
-                yield cwe_tag
-            except Exception:
-                yield "cwe"
+            yield self.metadata["cwe"]
         if "owasp" in self.metadata:
-            try:
-                # Assumed OWASP format is AX: Description
-                yield "owasp-{}".format(self.metadata["owasp"].split(":")[0])
-            except Exception:
-                yield "owasp"
+            yield f"OWASP-{self.metadata['owasp']}"
 
     @property
     def languages(self) -> List[Language]:
--- a/semgrep/semgrep/rule.py
+++ b/semgrep/semgrep/rule.py
@@ -267,7 +267,7 @@ class Rule:
                 cwe_id = self.metadata["cwe"].split("CWE-")[1]
                 cwe_tag= "cwe-"
                 for char in cwe_id:
-                    if char.is_integr():
+                    if char.isdigit():
                         cwe_tag += char
                     else:
                         break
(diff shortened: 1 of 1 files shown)


hunt3rkillerz:
Can any clarification regarding the CI error be provided. It doesn't appear to me if this issue is related to my change.

brendongo:
@hunt3rkillerz it isn't. That's our job to push a testable image to https://semgrep.dev/editor/

For security pushing from a PR from an outside contributor isn't enabled. We'll make that more clear/not run that task.

hunt3rkillerz:
Thanks @brendongo for the clarification. Also, it would be great to get some 👀 on this change (it's not letting me set reviewers).

brendongo:
@hunt3rkillerz sorry for the delay. I'll find someone who is more familiar with our sarif output to take a look

semgrep/semgrep/rule.py
```diff
@@ -261,9 +261,25 @@ def sarif_tags(self) -> Iterator[str]:
         Tags to display on SARIF-compliant UIs, such as GitHub security scans.
         """
         if "cwe" in self.metadata:
-            yield "cwe"
+            # Handling unexpected formatting of CWE meta-data
```
mschwager: Hi @hunt3rkillerz, thanks for the PR. Is it reasonable to simply return the string from the metadata if it exists? This avoids all the heavy string parsing. E.g.

```python
if "cwe" in self.metadata:
    yield self.metadata['cwe']
if "owasp" in self.metadata:
    yield f"OWASP-{self.metadata['owasp']}"
```

Note that I prepended OWASP because our convention hasn't included it, but has for CWE:

```
javascript/lang/security/audit/path-traversal/path-join-resolve-traversal.yaml
47:    owasp: 'A5: Broken Access Control'
48:    cwe: 'CWE-22: Improper Limitation of a Pathname to a Restricted Directory (Path Traversal)'

javascript/lang/security/audit/non-constant-sql-query.yaml
10:    owasp: 'A1: Injection'
11:    cwe: "CWE-89: Improper Neutralization of Special Elements used in an SQL Command ('SQL Injection')"

javascript/lang/security/audit/unknown-value-with-script-tag.yaml
14:    owasp: 'A7: Cross-site Scripting (XSS)'
15:    cwe: "CWE-79: Improper Neutralization of Input During Web Page Generation ('Cross-site Scripting')"
```

hunt3rkillerz: Hi @mschwager, the reason I tried to cut down the tags is I noticed that there was some inconsistent formatting in the rules repo (some rules had the full description and some did not). 

Additionally, I'm not sure if there is a string length limit on tags when you upload a SARIF to Github's API (happy to test this if the preference is to avoid the string parsing).

mschwager: I think simply returning what's in the metadata key (e.g. `self.metadata['cwe']`) would be easiest. If there are inconsistencies, or we want to improve the metadata, let's do so in the rules themselves instead of accounting for all cases here. If you have a list of rules whose metadata could be improved I'd be happy to help out :+1:

hunt3rkillerz: The changes above have been implemented.

hunt3rkillerz: Regarding rules with Metadata issues. I can probably go through the rules repo and see about cleaning up some formatting.

mschwager: Awesome, thank you!

brendongo (approved):

mschwager (approved):

mschwager:
Thanks again @hunt3rkillerz!

PR #1774, opened 2020-10-02
Make `--include` additive rather than subtractive

### LINKED ISSUE #1099: Add functionality to include additional file extensions to targets
Currently, the `--include` flag filters targets based on the specified argument instead of adding them. E.g. if we specify `languages: [java]` we filter targets based on the `.java` extension, we then filter targets again based on the `--include` flag. This doesn't really make any sense since we've already done the initial filtering.

It makes more sense for `--include` to be additive. E.g. if for some reason you have Java code with the `.txt` extension you can specify `--include "*.txt"` to also consider those files. This also allows us to use `pattern-regex` over arbitrary file extensions instead of having to use a hack like in #1093: `$(find /path/to/directory -name "*.ext")`.

This is also odd for using `--include` on a directory like `--include additional-directory/` when you could just include that directory in the list of targets like `semgrep --config sem.yaml additional-directory/`. 

Let's make this workflow easy!

mschwager:
Fix #1099.

I opened this to get the discussion going. This has been a constant pain for me: there's no way to add additional file extensions for analysis. And since we're already filtering files based on their corresponding language file extensions, a `--include` becomes a double filter, which isn't very useful.

This flag is used in `semgrep-action`: https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/semgrep.py#L184. Will this change break that functionality? Why are we using `--include` there and not specifying the files as targets?

I also briefly searched for `--include` use in the cloud app codebase and didn't find anything. Does the "Ignored Paths" functionality use `--include` or `--exclude`? Am I missing anything?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+- The `--include` flag is now additive rather than subtractive. The
+  `language` specification already filters based on file extension, a
+  `--include` would further filter this list of files - a double filter. Now,
+  e.g., a `--include "*.txt"` will add text files for analysis.
+
 ## [0.26.0](https://github.com/returntocorp/semgrep/releases/tag/v0.26.0) - 2020-09-30
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


brendongo:
We were using --include instead of explicit target because explicit target will run all rules on that target. For semgrep-action we want to pass the list of files to scan but we want semgrep to then figure out which rules to run on which file depending on extension. See https://github.com/returntocorp/semgrep/issues/1099#issuecomment-702852013

mschwager:
Closing this out as it breaks a core workflow - we'll continue to think on this so we can provide an ergonomic solution to the problem of including additional files.

PR #1798, opened 2020-10-08
* semgrep-core/pfff: Use latest

aryx:


aryx:
The failure in semgrep-rules will be addressed in a separate PR!

PR #1800, opened 2020-10-08
Do not consider $_GET and other PHP superglobals as metavariables

### LINKED ISSUE #1790: $_GET should not be considered a metavariable in PHP
https://semgrep.dev/s/4BoZ

aryx:
This is a quick fix. The right solution is to pass the language
to is_metavar_name function (or use an escape mechanism).
This closes https://github.com/returntocorp/semgrep/issues/1790

test plan:
test file included
make test


aryx:
Ok great, back to working CI!

PR #1806, opened 2020-10-08
adding support for running all js rules on ts files automatically.

colleend:
Should fix issue https://github.com/returntocorp/enterprise/issues/427. 

Test plan: 
1. create a file helloworld.ts with code:
`let helloWorld = "test"`
2. create a file helloworld.yaml with pattern:
``` 
rules:
  - id: find-unverified-transactions
  pattern: |
    helloWorld
  message: |
    In $METHOD, there's a call to make_transaction() without first calling verify_transaction() on the Transaction object.
  languages:
  - js
  severity: WARNING
```

3. run: 
`python -m semgrep -f <path to helloworld.yaml> <directory containing helloworld.ts>`

NOTE:
* Performance is taking a hit because we are running semgrep on TS files. We should edit baseline performance in order to reflect this.


semgrep/semgrep/target_manager.py
```diff
@@ -40,14 +40,15 @@
 ALL_EXTENSIONS = (
     PYTHON_EXTENSIONS
     + JAVASCRIPT_EXTENSIONS
+    + TYPESCRIPT_EXTENSIONS
     + JAVA_EXTENSIONS
     + C_EXTENSIONS
     + GO_EXTENSIONS
+    + RUBY_EXTENSIONS
```
brendongo, line 48: Why was this removed?

brendongo:
Is there a reason we're hacking this into the rule yaml instead of just changing the rule yamls to include typescript as a target language?

dlukeomalley:
@DrewDennison @brendongo @colleend Any updates on this work? Looking to clean up outstanding issues on our internal tracking boards 🤗

colleend:
There's been discussion to hold this off until TS repos are QA'd. The reason some tests are failing is because TS parsing is failing on some of the repositories.

DrewDennison (approved):

PR #1811, opened 2020-10-08
Support ... in JSX attribute list

### LINKED ISSUE #1803: [JSX] Allow ... in list of attributes

aryx:
This closes https://github.com/returntocorp/semgrep/issues/1803

Right now the '...' are actually filtered at matching time,
because we implicitely treat all jsx attribute list as an unordered
list of partial attributes (implicit '...').
This may not be completely consistent with what do we for Dict and Set,
but most rules are written that way for now.

Test plan:
test file included
make test


aryx:
The perf regressions seems unrelated.

PR #1812, opened 2020-10-08
adding yml to output and upload SARIF file during each pull/push

colleend:
All other yml files in workflows were for release/testing purposes, so I created a new file for organization. 

@dlukeomalley does this seem ok?

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,26 @@
+name: Semgrep
+on:
+  pull_request:
+  push:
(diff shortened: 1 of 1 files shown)


colleend:
Would need to specify running on public repos, which the github env variables don't actually have anything you can access yet: 
https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables

Or we could specify the repos we want this action to run on. Seems like this may be more difficult than expected, unless someone has good understanding of github actions.

PR #1824, opened 2020-10-13
Add 'metavariable-comparison' frontend to expression eval

### LINKED ISSUE #1724: RFC: Add primitive for mathematical comparisons of metavariable

mschwager:
Fixes #1724.

Things notably out-of-scope for the first iteration of this functionality:

1.) Multi-language expressions
2.) Multi-metavariable expressions

Per (1), let's start small with a language we know well and expand from there as needed. Per (2), we can leverage the logical AND functionality of 'patterns' here instead of multiple metavariables in the same expression. Again, let's start small and iterate as needed.


semgrep/semgrep/evaluation.py
```diff
@@ -11,13 +11,12 @@
 from typing import Tuple
 
 logger = logging.getLogger(__name__)
-# disable logging from this module most of the time
```
mschwager, line 14: I think this may have been unintentionally left in. It makes it difficult to debug this module because it's impossible to set the logger level to `DEBUG` with this here. Now that we have the `--debug` flag I think this is no longer necessary, or unintentional.

semgrep/semgrep/evaluation.py
```diff
@@ -193,14 +254,6 @@ def _evaluate_single_expression(
         )
         return output_ranges
     elif expression.operator == OPERATORS.METAVARIABLE_REGEX:
-        if (
```
mschwager, line 196: This is no longer needed with the `jsonschema` changes.

semgrep/semgrep/evaluation.py
ievans: I did it intentionally -- it was really difficult to debug other parts of the program with the large amount of output this module spews. But very open to other approaches

mschwager: Hmm, I usually try to err on the side of over-outputting with debug output vs. under-outputting. It's much easier to filter existing output than include additional output information in an already released version. If you're looking for a specific message in the debug output you can always:

```
$ semgrep --debug ... 2>&1 | grep 'compiled result'
```

Or, if you want to squelch a particularly noisy log line you can do something like:

```
$ semgrep --debug  ... 2>&1 | grep -v 'after filter'
```

Another good option here is including `%(name)s` in our [logger format](https://docs.python.org/3/library/logging.html#logrecord-attributes) in `semgrep/semgrep/util.py`. Then we could do something like the following to squelch a single module's logging:

```
$ semgrep --debug  ... 2>&1 | grep -v 'semgrep.evaluation'
```

docs/configuration-files.md
```diff
@@ -19,6 +19,7 @@ Contents:
   * [`pattern-either`](configuration-files.md#pattern-either)
   * [`pattern-regex`](configuration-files.md#pattern-regex)
   * [`metavariable-regex`](configuration-files.md#metavariable-regex)
+  * [`metavariable-comparison`](configuration-files.md#metavariable-comparison)
```
aryx, line 22: could also be called pattern-where-python-safe

aryx: Maybe the only safe thing people want to express are comparisons, so maybe it's not a bad name but it does not really give room for being more than that. For example some people wanted to check if $X in [1, 42, 57]
and it's not really a comparison (or maybe it is).

docs/configuration-files.md
```diff
@@ -224,6 +226,64 @@ rules:
     severity: ERROR
 ```
 
+### `metavariable-comparison`
+
+The `metavariable-comparison` operator compares metavariables against a basic [Python comparison](https://docs.python.org/3/reference/expressions.html#comparisons)
+expression. This is useful for filtering results based on a [metavariable's](/docs/pattern-features.md#metavariables) value.
```
aryx: maybe say also "and as opposed to pattern-where-python is guaranteed to be safe and so can be computed without
--unsafe-eval-do-at-your-own-risk CLI parameter".

docs/configuration-files.md
```diff
@@ -224,6 +226,64 @@ rules:
...
+The `metavariable-comparison` operator is a mapping which requires the
+`metavariable` and `comparison` keys. It can be combined with other pattern operators:
+
+```yaml
+rules:
+  - id: superuser-port
+    patterns:
+      - pattern: set_port($ARG)
+      - metavariable-comparison:
+          metavariable: '$ARG'
```
aryx, line 245: why the quote around the metavariable? we don't put quote to express patterns.

docs/configuration-files.md
```diff
@@ -224,6 +226,64 @@ rules:
...
+          comparison: '$ARG < 1024'
```
aryx, line 246: same, why the quote? The main interest of yaml is to remove the need for quotes no?

docs/configuration-files.md
```diff
@@ -224,6 +226,64 @@ rules:
...
+    message: "module setting superuser port"
+    languages: [python]
+    severity: ERROR
+```
+
+This will catch code like `set_port(80)` or `set_port(443)`, but not `set_port(8080)`.
+
+The `metavariable-comparison` operator also takes optional `base: int` and
```
aryx, line 254: the number syntax itself contains the base, why specify the base here?
OCaml, and I guess Python will be able to read 0xa or 10 or 0o82 or whatever syntax and convert to an internal number.

docs/configuration-files.md
```diff
@@ -224,6 +226,64 @@ rules:
...
+`strip: bool` keys. These keys set the integer base the metavariable value
+should be interpreted as and remove quotes from the metavariable value,
```
aryx, line 256: again, I don't think we need strip, if the metavariable matches a string, then -eval will do the right thing and allow only operations on strings.

docs/configuration-files.md
aryx: Also why specify the metavariable ... just export all the metavariables in the current matched state and pass it to semgrep-core -eval.
Also this prevents to use multiple metavariables comparisons as $X < $Y.

semgrep/semgrep/evaluation.py
```diff
@@ -56,6 +55,68 @@ def get_re_range_matches(
...
+
+        any_matching_ranges = any(
+            pm.range == _range
+            and metavariable in pm.metavars
+            and compare_range_match(
+                metavariable,
+                comparison,
+                strip,
+                base,
+                pm.metavars[metavariable]["abstract_content"],
```
aryx, line 110: @brendongo might get rid of the use of abstract_content at some point, which is quite fragile.

semgrep/semgrep/metavariable_comparison.py
```diff
@@ -0,0 +1,34 @@
...
+from semgrep.constants import PLEASE_FILE_ISSUE_TEXT
+from semgrep.constants import SEMGREP_PATH
+from semgrep.error import SemgrepError
+from semgrep.util import sub_check_output
+
+
+def metavariable_comparison(
+    metavariable: str, comparison: str, content: Union[int, float]
+) -> bool:
+    # semgrep-core requires the 'language' key be before the 'code' key,
```
aryx: I can fix that and sort on the semgrep-core side. This is ugly indeed otherwise to impose an order.
Sorry.

semgrep/semgrep/metavariable_comparison.py
```diff
@@ -0,0 +1,34 @@
...
+        temp_file.flush()
+        cmd = [SEMGREP_PATH, "-eval", temp_file.name]
+        try:
+            output = sub_check_output(cmd)
+        except subprocess.CalledProcessError as ex:
+            raise SemgrepError(
+                f"error invoking semgrep with:\n\t{' '.join(cmd)}\n\t{ex}\n{PLEASE_FILE_ISSUE_TEXT}"
+            )
+
+    return output.strip() == b"true"
```
aryx, line 32: Note that if it returns NONE, this means the expression passed was not considered safe or part of the handled subset. Maybe you want to return that information to the user.

aryx (commented):
Looks good to me. I disagree with some of the language design (the need for the metavar field, or base, or strip), but maybe users like it that way. We can A/B test it :)

docs/configuration-files.md
mschwager: IMO, naming it `pattern-where-python-safe` will make people question why they'd ever need `pattern-where-python` and what the differences between the two are. Then we'd have to document all the differences, which may be hard to grok for someone without a program analysis background. E.g. "what's the difference between an expression and a statement", "why can't I call functions X, Y, and Z?", etc. This would also tie our hands if we ever decided to open up the list of available languages. I.e. would we also have a `pattern-where-javascript-safe`?

Keeping this distinct from `pattern-where-python` allows for a simpler mental model of our operators IMO.

docs/configuration-files.md
mschwager: Hah, I think YAML provides a bit more than lack of quotes over JSON :D

I've been burned one too many times by YAML interpreting string data as a different scalar type, so I always quote strings by default now /shrug. An interesting example here was including git short hashes in YAML data, so on the off chance your commit hash was like `1234567` YAML would interpret it as an integer instead of the usual string like `5dbfa6d`. That was a fun bug to track down.

I don't have strong feelings - feel free to change the examples in the docs if you'd like.

docs/configuration-files.md
mschwager: Specifying the metavariable allows for the `base` and `strip` behavior. We could do something like make `metavariables`, `bases`, and `strips` all lists or something so we can specify multiple, but I went with the simplest approach here.

docs/configuration-files.md
mschwager: I don't think that's true at the moment. The regex for integer detection is currently `^[0-9]+$`, which misses `0x`, `0o`, etc. Since the metavariable content is always initially a string, I wanted to provide a mechanism for parsing the string value in different bases.

Generally, if you look at our `semgrep-rules`, the majority of rules using `pattern-where-python` perform a simple comparison, and often parse a number in a different base, or perform a string strip. I know this implementation isn't the maximally powerful/generalized solution, it instead aims to cover the majority of use cases, not all of them, and hit the sweet spot between power and usability (and minimize `--dangerously...` usage). E.g.

```
24:  - pattern-where-python: |
25-      int(vars['$PERM'], 8) > 0o600
```

```
18:  - pattern-where-python: |-
19-      int(vars['$BITS']) < 2048
```

```
18:  - pattern-where-python: |-
19-      int(vars['$SIZE']) < 128
```

```
11:  - pattern-where-python: int(vars['$NUM'].replace('"', '')) > 32767 or int(vars['$NUM'].replace('"', '')) < -32768
```

```
21:  - pattern-where-python: |-
22-      int(vars['$NUM'].replace('"', '')) > 2147483647 or int(vars['$NUM'].replace('"', '')) < -2147483648
```

docs/configuration-files.md
mschwager: `strip` is for strings that look like integers. E.g.

```
21:  - pattern-where-python: |-
22-      int(vars['$NUM'].replace('"', '')) > 2147483647 or int(vars['$NUM'].replace('"', '')) < -2147483648
```

```
11:  - pattern-where-python: int(vars['$NUM'].replace('"', '')) > 32767 or int(vars['$NUM'].replace('"', '')) < -32768
```

Performing string operations on `"2147483647"` won't be helpful in this situation.

semgrep/semgrep/evaluation.py
mschwager: It's used in quite a few places, so we'll have to tread carefully if/when we do.

semgrep/semgrep/metavariable_comparison.py
mschwager: No worries! That was going to be one of my v2 feature requests. That and `float` support :)

mschwager (commented):
There's lot of thoughts here! I'll include some additional conversation @ievans and I had with the hopes that that will clear up some of my intentions with these changes:

MS:

> My longer-term thoughts were that pattern-where-python would remain as a "proving ground" for rules, and if patterns emerge in those rules we could lift them up into an official operator. If we can't completely get rid of pattern-where-python then we should at least minimize --dangerously... usage, however if we make some sacrifices on the power of Semgrep we could always remove it.

IE:

> what’s the advantage of being explicit with this new metavariable-comparison key? ie why not just call pattern-where-python with the generic-AST and if it can’t do it, either fallback to our current evaluator or compliain that --dangerous needs to be set
> and we could implement pattern-where-<lang> for any of our supported languages

MS:

> We originally had 25 pattern-where-python rules - with metavariable-regex we were able to cut that to 18. With metavariable-comparison we can cut that to 4. So I suspect those two operators cover the vast majority of things people would need surrounding code execution-type things.

IE:

> ie why not just call pattern-where-python with the generic-AST and if it can’t do it, either fallback to our current evaluator or compliain that --dangerous needs to be set

MS:

> That seems like a reasonable solution. This is more subjective, but I tend to prefer explicit solutions vs. implicit/automagic ones. The current solution is easy to validate, easy to reason about, easy to extend, but obviously lacks some of the raw power/generalization of pure code. It's aiming to cover the majority of use cases, not all of them, and hit the sweet spot between power and usability (and minimize --dangerously... usage)

And with that, I'll let everyone sleep on their thoughts.

semgrep/semgrep/evaluation.py
aryx: I now use in semgrep-core a logging library inspired by the python one, and you can use a config file to specify the different levels for the different modules outside the code. See https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/log_config.json.ex1

docs/configuration-files.md
aryx: sounds fair.

docs/configuration-files.md
aryx: Got it, no the examples are fine, I was just worried I misunderstood something with YAML.

aryx:
I agree with Isaac that I would rather piggyback on pattern-where-python: than using
a new construct, but I'm fine trying matt's idea and see how people react to it.

ievans:
On reflection I really like where this is heading. I wonder if the best approach is to wait for the generic-AST evaluator to grow just a bit more powerful, and then switch to a world where we remove --dangerously. If we can really get down to just a few rules, is it worth maintaining? The registry does a tremendous amount of work to try to make sure we only run rules from trusted contexts. 

In this world, we could just have:
```
pattern-where-code:
    language: python
    code: $X.strip() > int($Y)
```

Alternatively, we could switch to a safe-by-design evaluator that is more full featured like the ones mentioned in https://github.com/returntocorp/semgrep/issues/1724. 

But I think there's not a huge rush to jump to either of these worlds just yet.

aryx (approved):
Ok, let's try this then!

docs/configuration-files.md
mschwager: @aryx I had a thought here: would it make more sense for `semgrep-core` to simply take in the data as the corresponding JSON type instead of performing string parsing? I.e. a JSON integer is an integer, a JSON float is a float, a JSON string is a string, etc, instead of trying to interpret the data with regexs. That way Python can interpret the metavariable data it receives, then `semgrep-core` can simply operate on the JSON types.

IMO Python is better equipped to deal with things like string parsing, while OCaml is obviously better equipped to deal with program analysis tasks.

aryx: Yes, good idea (but no Python is not better equipped for string parsing :) )

aryx: I'll modify -eval to accept directly the right Json type.

aryx: @mschwager https://github.com/returntocorp/semgrep/pull/1839

mschwager:
> I wonder if the best approach is to wait for the generic-AST evaluator to grow just a bit more powerful, and then switch to a world where we remove --dangerously.

Unless we make some compromises on Semgrep's power we can never fully remove `--dangerously`. E.g. it's highly unlikely the ReDoS rule can ever be ported to an evaluator.

> If we can really get down to just a few rules, is it worth maintaining? The registry does a tremendous amount of work to try to make sure we only run rules from trusted contexts.

It seems you've caught on to one of my ulterior motives here! A side-goal here is to make the risk:reward of `pattern-where-python` so skewed that we can remove the ability to run those rules from the registry in our service, even with a whitelist. That is, the risk of this functionality is very high, while the reward is very low (`4 / 758 = 0.52%` of our rules will use `pattern-where-python`). We can instead say: "If you want to use `pattern-where-python` you'll have to run that rule locally." I won't get into all the details here, but it's on my mind.

> Alternatively, we could switch to a safe-by-design evaluator that is more full featured like the ones mentioned in #1724.

This seems like a noble goal, but I'm concerned about the implementation. I'm not aware of any secure mechanisms for this in the Python ecosystem, and I would also like to see if any existing evaluators have had some kind of professional security assessment done. Remember, this is unfettered RCE on the line, so we should tread carefully.

Anyways, I just wanted to make sure I've addressed everything here. Any other thoughts?

docs/configuration-files.md
underyx: I like the resolution here. I also would add that this suggestion would make the unsafe variant our de facto default, which is something we should always strive to avoid.

docs/configuration-files.md
```diff
@@ -224,6 +226,64 @@ rules:
...
+      - pattern: set_port($ARG)
+      - metavariable-comparison:
+          metavariable: '$ARG'
+          comparison: '$ARG < 1024'
+    message: "module setting superuser port"
+    languages: [python]
+    severity: ERROR
+```
+
+This will catch code like `set_port(80)` or `set_port(443)`, but not `set_port(8080)`.
```
underyx, line 252: Could we also document whether it can compare strings to, for instance, ensure alphabetical ordering?

mschwager: Although `semgrep-core` is wired up for performing basic string comparisons I didn't include this in the Python interface. The `pattern-regex` operator can be used for string operations. I tend to split these use-cases in my mind like this: `pattern-regex` is useful for performing string operations and `pattern-comparison` is useful for performing numeric operations.

underyx: I was thinking of a pattern like

```
$FIRST = ...
$SECOND = ...
```
and then ensuring `$FIRST < $SECOND` to check for alphabetical order in a whole file.

1. Is there a way to do this with regex? 🤔 
2. Could we specify in the docs that only numeric comparisons are supported, then?

mschwager: > ... and then ensuring $FIRST < $SECOND to check for alphabetical order in a whole file.

I don't think I understand what you're trying to achieve here. You want to determine if a non-code file is in alphabetical order? Like a CSV file? This can't be done with a regex, especially because these operators only handle a single metavariable.

> Could we specify in the docs that only numeric comparisons are supported, then?

Sure thing :+1:

underyx: I meant, in a `constants.py`, which looks like:

```py
FOO_LOGIN = "me"
FOO_PASS = "hunter2"

BAR_PASS = "hunter3"
```

I'd write

```yaml
  pattern: |
    $FIRST = ...
    $SECOND = ...
  metavariable-comparison:
    comparison: $FIRST < $SECOND
```

to ensure that `BAR_PASS` is defined first and thus the whole file is in alphabetical order.

underyx: Reference: https://eslint.org/docs/rules/sort-vars

mschwager: Oh I see what you mean. That behavior is not currently supported. Lexicographical ordering of strings seems like a fairly niche use-case that I'd like to avoid for the v1 of this functionality in the interest of moving forward.

mschwager:
I'm planning on merging this soon so we can have it for the `0.28.0` release.

One last random thought I had regarding metavariable comparison: don't forget we always have implicit metavariable "equals" or "not equals" based on the metavariable name used:

- Equals: https://semgrep.dev/s/P9oq/
- Not equals: https://semgrep.dev/s/Jq0Q/

This will work for basic string data equality comparisons.

PR #1826, opened 2020-10-14
WIP integrate spacegrep/spacecat into semgrep

mjambon:
Note: Most commits appearing here are old since we're importing an external repo while preserving history. You can ignore commits older than this week.

Main changes so far:

* The spacegrep repo which used to be separate and private is now a folder at the root of the `semgrep` repo.
* The main makefile and CI were updated to build the executables `spacegrep` and `spacecat` (both are statically linked on linux just like `semgrep-core`, and "mostly statically linked" on mac; all of this in theory, we'll see if it works when we have semgrep tests invoking `spacegrep`).

The plan is now to call `spacegrep` from the python application when the language specified in the yaml input is `generic`. I will make changes to the `spacegrep` command so as to emit json output just like `semgrep-core` does.


aryx (commented):
Why not keep spacegrep as a separate repo and use it as a submodule?
That way you can also release spacegrep independently of semgrep.

aryx (approved):
Let's go for it then.

PR #1832, opened 2020-10-14
Java annotation patterns regardless of context

### LINKED ISSUE #1829: catching Java annotations regardless of the context
in Java it is possible for annotations to be applied to targets (contexts) such as a class, a method, a field etc). I'd like to write semgrep patterns that match an annotation in any context (the pattern is just `@SomeAnnotation(...)` and does not have class `$CLASS { ... }` or `$RETURN $METHOD(...) {...}` in it.
I'm writing rules that really only care about the annotation, the idea being to flag code for an analyst to review further. At the moment I have used pattern-either: and write effectively the same pattern for class and for method. In the message I then might refer to $CLASS or $METHOD which may or may not be bound. I could use a more generic name for the metavariable but it makes the pattern harder to read. I could write separate rules but its too micro and repetitive for my liking.
Ideally, I'd write a more generic singular pattern that can match in many contexts. 

catching Java annotations regardless of the context
```
@SomeAnnotation(...)
```

aryx:
Fixes https://github.com/returntocorp/semgrep/issues/1829

test plan:
test file included
make test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Added
 - Added matching with partial patterns for function signatures or class headers
-  for Javascript/Typescript.
+  for Javascript/Typescript and Java.
+- Added matching of annotation patterns for Java (@SomeAnnot(...)) in any
+  context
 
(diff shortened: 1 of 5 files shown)


aryx:
Note that not all annotations are currently stored in the AST, only annotations for class and methods, but not yet
for packages, types, etc, so you may miss some use sites. Still, better than before at least.

PR #1836, opened 2020-10-14
Registryv2 edits

colleend:
for registry v2 caching; outputs & gzips results file to be stored in semgrep_run_output. 

Test plan: 

run semgrep with the --save-output flag in semgrep-rules. Ex: `python -m semgrep --test --strict --test-ignore-todo --dangerously-allow-arbitrary-code-execution-from-rules --save-output $PWD`


semgrep/semgrep/cli.py
```diff
@@ -200,6 +200,13 @@ def cli() -> None:
     output.add_argument(
         "--json", action="store_true", help="Output results in JSON format."
     )
+    output.add_argument(
+        "--save-output",
```
ievans: Perhaps it should be `--save-test-output-dir` to be more specific?

semgrep/semgrep/test.py
```diff
@@ -328,6 +331,18 @@ def generate_file_pairs(
         print(json.dumps(output, indent=4, separators=(",", ": ")))
         sys.exit(exit_code)
 
+    # save the results to json file. this is really ratchet.
+    if save_output:
+        list_to_output = []
+        with open('semgrep_runs_output.json', 'w') as f:
```
ievans: `semgrep_runs_output.json` could be a constant?

mschwager:
We already have the `--output` flag. Can we honor that here as well instead of adding a new flag? Also, are we that concerned about size here? This dataset doesn't seem very large to me, i.e. might not warrant `tar`ing. Seems like we could achieve most of this without such special case code. Thoughts?

colleend:
Hm I think functionality in this case is different because output saves the results of the current run, but we want to run on every semgrep-rules file and save the corresponding output every time we run save-output. The result is pretty large, 100,000+ lines, and we need to upload it as a github artifact (which prevents us from uploading 100,000+ line files).

semgrep/semgrep/cli.py
colleend: how about --save-test-output-tar?

ievans (approved):
Looks great @colleend

PR #1838, opened 2020-10-15
Use the new ocamllint semgrep rules and fix warnings

aryx:
Dogfood!

test plan:
make test
make check

### DIFF
--- a/semgrep_ocaml.yml
+++ b/semgrep_ocaml.yml
@@ -1,0 +1,210 @@
+rules:
+- id: ocaml.lang.best-practice.ref.ocamllint-ref-incr
+  pattern: $X := $X + 1
+  message: You should use `incr`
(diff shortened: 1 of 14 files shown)


semgrep_ocaml.yml
```diff
@@ -0,0 +1,210 @@
+rules:
```
aryx, line 1: At some point I would rather like use --config https://semgrep.dev/r/ocaml but I could not find a way to create an ocaml ruleset, so I've created a policy and downloaded the corresponding yaml file.

aryx:
No real bugs (those were all warnings), but still feels good to cleanup a bit.

aryx:
fixed 12 findings.
I have 288 in pfff/, argh, I really want bento archive here.

PR #1839, opened 2020-10-15
New JSON interface for -eval metavars, handle float, better json parsing

aryx:
This will help #1724
I am now using JSON value types instead of parsing strings for the metavars
as suggested by Matt.
I also support the different JSON field in any order, and
added basic support for operations on float.

test plan:
test file included


mschwager (approved):
:tada: :tada: :tada:

PR #1848, opened 2020-10-16
Spacegrep integration in Python layer

minusworld:
This PR integrates spacegrep into the Python layer. It can be invoked using the language "generic".

```
$ semgrep -l generic -e '## $X' README.md
README.md
63:## Getting
98:## Examples
112:## Try
146:## Resources
160:## Usage
162:## Command
166:## Exit
174:## Upgrading
189:## Contributing
204:## Commercial
ran 1 rules on 1 files: 10 findings
```

A few things of note:
- The spacegrep offsets are bit off; the function `_patch_lines` fixes that for now.
- Semgrep needs the `extra['lines']` key which is a list of matched lines. `_patch_lines` handles this for now.
- This integration assumes spacegrep was installed somewhere that is on the system PATH.
- An observation: Semgrep metavariables include the dollar sign '$' in front of them in the JSON output; spacegrep excludes the dollar sign at the moment.


semgrep/semgrep/spacegrep.py
```diff
@@ -0,0 +1,93 @@
...
+        with open(match["path"], "r") as fin:
+            data = fin.readlines()
+        start_l = match.get("start", {}).get("line", 0)
+        start_c = match.get("start", {}).get("col", 0)
+        end_l = match.get("end", {}).get("line", 0)
+        end_c = match.get("end", {}).get("col", 0)
+
+        lines = data[start_l - 2 : end_l - 1]
+        lines[0] = lines[0][start_c - 1 :]
+        lines[-1] = lines[-1][: end_c - 1]
```
mjambon: ok, I'll look into this. I specifically added 1 here and there before exporting to json because a comment I found in the ocaml code said so.

mjambon (approved):

brendongo (changes requested):
Is spacegrep executable correctly packaged in wheel and ~docker images~?

mjambon:
@brendongo 
> Is spacegrep executable correctly packaged in wheel?

I don't know if there's any subtlety that I may have missed. I'm not familiar with python wheels.

@minusworld added tests and they pass, after I made some fixes. The status of `spacegrep` and `spacecat` is the same as `semgrep-core`. All 3 are executables built from ocaml/C/C++ source. They're statically linked for linux and "mostly statically linked" for osx.

brendongo:
Maybe worth taking a look at https://github.com/returntocorp/semgrep/blob/develop/semgrep/setup.py#L88 for how the python scripts install semgrep-core in the correct place.

mjambon:
@brendongo I would have missed it if you hadn't looked into it. Is there a way to test this? automatically?

brendongo:
The python unit and e2e tests should be failing since they don't have spacegrep installed in their environments... not sure why they pass tbh :D

brendongo:
To test locally: `python setup.py sdist bdist_wheel` in repo_root/semgrep will generate the wheel file. To test if that wheel file contains all the necessary files I would make sure `which spacegrep` doesnt find anything, install the wheel using `pip3 install PATH_TO_DOT_WHL_FILE` and then checking that `which spacegrep` now points to where it should

brendongo:
IMO we could get this in as is, not advertise the functionality and work on getting it building on pip in a follow up PR. For now BYO spacegrep might be fine.

But it looks like we correctly fail tests now so yay :D

mjambon:
@brendongo I'll try to get this fixed today-ish. I'll learn something.

mjambon:
Here's what I did:

* re-enabled the installation of executables into `/usr/bin` for the tests to work. An alternative would be to first create the wheel package and then install it and use it for the tests. This is done somewhere else and it looks like too much too change for now.
* added `spacegrep` and `spacecat` to the wheel package just like `semgrep-core`. This is not tested, see point above.

That's not much but it works in theory. In practice, we're getting a missing blank in a test result, which seems to come out of nowhere (maybe a library that wasn't pinned and changed its behavior). You can see it [here](https://github.com/returntocorp/semgrep/pull/1848/checks?check_run_id=1284435787#step:8:112). We expect the string
```
This \noperation
```
but instead we get
```
This\noperation
```
This is from the file `semgrep/tests/e2e/snapshots/test_check/test_registry_rule/results.json`. Let me know if I should just remove the space in that file and move on.

brendongo (approved):

Issue #1721, opened 2020-09-23
Autofix inadvertently changing code outside the match


msorens:
**Describe the bug**
Autofix properly updated the matched pattern with the replacement text--but it also changed the code outside the match and it should not have. I first found this by running semgrep on the command line with a YAML file; then reproduced precisely the same behavior at semgrep.dev.

**To Reproduce**
In the playground (https://semgrep.dev/s/7g4k/), run the sample and apply the autofix. It correctly changes the match but also deletes the immediately adjacent semicolon.

**Expected behavior**
Surrounding code should be 100% preserved.

**Screenshots**
![image](https://user-images.githubusercontent.com/6817500/93963640-92a8dc00-fd12-11ea-9b3a-f2cf68cfd591.png)

**Environment**
v 0.24.0

chmccreery:
Found the same bug in python: https://semgrep.dev/s/LrkX/

brendongo:
Looks like we might have some disparity on how UI and CLI treat end column (inclusive vs exclusive). I think the webUI is treating it as exclusive, but the autofix code treats it as inclusive. Will have to double check with semgrep-core on what actual intention is.

Evidence:
https://semgrep.dev/s/RrNv/?stepByStep=1 shows finding range as 0-5
WebUI highlights 0-4 but autofix replaces characters 0-5.

nbrahms:
cc. @aryx

aryx:
Here is the code in semgrep-core/reporting/JSON_report.ml
```
let json_range min max =
  (* pfff (and Emacs) have the first column at index 0, but not r2c *)
  let adjust_column x = x + 1 in

  let min = PI.token_location_of_info min in
  let max = PI.token_location_of_info max in
  let len_max = String.length max.PI.str in
  J.Object [
    "line", J.Int min.PI.line;
    "col", J.Int (adjust_column min.PI.column);
    "offset", J.Int min.PI.charpos;
  ],
  J.Object [
    "line", J.Int max.PI.line;
    "col", J.Int (adjust_column (max.PI.column + len_max));
    "offset", J.Int (max.PI.charpos + len_max);
  ]

```

aryx:
So it's adding the length of the matched thing to the starting offset, so I guess this is exclusive in brendon's terminology.
On brendon's example, here is what semgrep-core report:
```
+ /home/pad/semgrep/_build/default/bin/Main.exe -lang python -rules_file test.yaml test.py
[...]

Issue #1768, opened 2020-10-01
Go: displayed types are not complete


disconnect3d:
Not sure if that is a bug, but the displayed types by `message` key in the rule format are a bit incomplete.

PoC:
```
$ cat a.yml
rules:
- id: some-rule
  pattern: func ($O *$TYPE) $METHOD($A *$TA, $B *$TB) { ... }
  message: arg1=`$TA` arg2=`$TB`
  severity: WARNING
  languages: [go]

$ cat a.go
package heh
func (o *T) Foo(a *some.NamespacedType, b *some.OtherType) { }
func (o *T) Bar(a *interface{}, b *struct{}) { return }

$ semgrep -f a.yml | grep severity
running 1 rules...
severity:warning rule:some-rule: arg1=`some NamespacedType` arg2=`some OtherType`
severity:warning rule:some-rule: arg1=`interface` arg2=`{}`
```

There are three weird cases:
1. The `some.NamespacedType` is displayed without dots, as `some NamespacedType`
2. The `interface{}` is displayed as just `interface`
3. The `struct{}` is displayed as `{}`

brendongo:
Convenience repro link: https://semgrep.dev/s/QwNJ

Checking metavars locally with --json has output: 
```
"metavars": {
          "$TB": {
            "start": {
              "line": 2,
              "col": 44,
              "offset": 55
            },
            "end": {
              "line": 2,
              "col": 58,
              "offset": 69
            },
            "abstract_content": "some OtherType",
            "unique_id": {
              "type": "AST",
              "md5sum": "93c4313b9bdeb983dad54c719c7c7dff"
            }
          },
          "$B": {
            "start": {
              "line": 2,
              "col": 41,
              "offset": 52
            },
            "end": {
              "line": 2,
              "col": 42,
              "offset": 53
            },
            "abstract_content": "b",
            "unique_id": {
              "type": "id",
              "value": "b",
              "kind": "Param",
              "sid": 2
            }
          },
          "$TA": {
            "start": {
              "line": 2,
              "col": 20,
              "offset": 31
            },
            "end": {
              "line": 2,
              "col": 39,
              "offset": 50
            },
            "abstract_content": "some NamespacedType",
            "unique_id": {
              "type": "AST",
              "md5sum": "2af73acda8772054a355b37e8fa59db3"
            }
          },
          "$A": {
            "start": {
              "line": 2,
              "col": 17,
              "offset": 28
            },
            "end": {
              "line": 2,
[...]

PR #1859, opened 2020-10-19
Use metavariable ranges

brendongo:
Before this commit semgrep was using the abstract_context associated
with each metavariable assuming it was what the metavariable matched with
this is not always the case since the abstract context removes some concrete
syntax.

This PR relies on the more stable line ranges reported with the metavariable
as the value of the metavariable.

### DIFF
--- a/semgrep/semgrep/pattern_match.py
+++ b/semgrep/semgrep/pattern_match.py
@@ -8,7 +8,7 @@ from semgrep.semgrep_types import Range
 
 class PatternMatch:
     """
-        Encapsulates a section of code that matches a single pattern
+    Encapsulates a section of code that matches a single pattern
     """
 
     def __init__(self, raw_json: Dict[str, Any]) -> None:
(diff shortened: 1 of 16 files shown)


brendongo:
Fixes any lingering issues in https://github.com/returntocorp/semgrep/issues/1768 and fixes https://github.com/returntocorp/semgrep/issues/1721

semgrep/semgrep/pattern_match.py
```diff
@@ -63,5 +63,23 @@ def end(self) -> Dict[str, Any]:
...
+    def get_metavariable_value(self, metavariable: str) -> str:
+        """
+            Use metavars start and end to read into the file to find what the
+            metavariable in this pattern match maps to in the file
+
+            Assumes METAVARIABLE is a key in self.metavars
+        """
+        # Offsets are start inclusive and end exclusive
+        start_offset = self.metavars[metavariable].get("start").get("offset")
+        end_offset = self.metavars[metavariable].get("end").get("offset")
```
underyx: If `start` or `end` are missing as keys, we'll fail anyway, so I'll update this to `["end"]["offset"]` to make the error message more clear.

underyx (approved):

aryx:
Great!

PR #1869, opened 2020-10-22
Update README.md

dlukeomalley:


pabloest:
Re: the text, IMO “…for running a modern AppSec program” is a bit too marketing-y for the README, and something like “hosted infrastructure to apply and enforce Semgrep policies” would be more straightforward.

PR #1875, opened 2020-10-22
Error out in --test if ruleid or ok not in reported IDs

mschwager:
This should help identify when we use `ruleid: <rule-id>` or `ok: <rule-id>` with a rule-id that doesn't exist.

Also includes random 'black' fixes that went unnoticed.

cc @brendongo do you know how `pre-commit` fails locally, but passed in CI?


brendongo:
Should we also error out if the reported rule-id has no corresponding finding? i.e. do a symmetric difference instead of difference?

brendongo (approved):

PR #1876, opened 2020-10-22
Ignore tree-sitter error nodes

mjambon:
This switches to the latest version of ocaml-tree-sitter (runtime + generated code), which removes error nodes from the CST and lists them as errors alongside the CST.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ build:
 
 .PHONY: install
 install:
+	$(MAKE) -C spacegrep install
 	$(MAKE) -C semgrep-core install
 	python3.7 -m pip install semgrep
 
--- a/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
@@ -2219,22 +2219,23 @@ and declaration (env : env) (x : CST.declaration) =
 (* Entry point *)
 (*****************************************************************************)
 let parse file =
- H.convert_tree_sitter_exn_to_pfff_exn (fun () ->
-  let ast =
-    Parallel.backtrace_when_exn := false;
-    Parallel.invoke Tree_sitter_csharp.Parse.file file ()
-  in
-  let env = { H.file; conv = H.line_col_to_pos file } in
+  H.wrap_parser
+    (fun () ->
+       Parallel.backtrace_when_exn := false;
+       Parallel.invoke Tree_sitter_csharp.Parse.file file ()
+    )
+    (fun cst ->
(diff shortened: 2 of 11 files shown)


aryx:
Do you have an example of a toy js file containing an error that can be now analyzed with semgrep because
we skip the error token?

mjambon:
How to investigate a failed test.
--

The logs shown by GHA look like this:
```
____________ test_semgrep_on_repo[https://github.com/jekyll/jekyll] ____________
[XPASS(strict)] ruby parse error on '!old_scope.key? "type"'
```

Look into the `public_repos.py` file and locate the test case for which an error was reported. For example:
```python
    xfail_repo(
        {"repo": "https://github.com/jekyll/jekyll", "languages": ["ruby"],},
        reason="ruby parse error on '!old_scope.key? \"type\"'",
    ),
```

Find the repository's git URL and clone it:
```
$ git clone https://github.com/jekyll/jekyll.git
```

To investigate tree-sitter parsing errors, use `parse-ruby` installed by `make install` like `semgrep-core`:
```
$ parse-ruby lib/jekyll/frontmatter_defaults.rb
...
|   |   |   |   |   |   |   |   |   |   |   |   |   | end
|   |   |   |   |   |   |   |   |   |   |   Line_brk
|   |   |   |   |   |   |   |   |   |   |   ""
|   |   |   |   |   |   |   | end
|   |   |   |   Line_brk
|   |   |   |   ""
Error: File lib/jekyll/frontmatter_defaults.rb, line 190, characters 9-23:
        true
      else
        !old_scope.key? "type"
         ^^^^^^^^^^^^^^
      end
    end
node type: ERROR
children: [
  call
]
Unrecognized construct
total lines: 240
error lines: 1
error count: 1
success: 99.58%
```
This file is now parsed without error by semgrep-core, since errors are ignored at the moment. We obtain an almost complete parse tree, only lacking the node corresponding to the line shown in the error message.

mjambon:
@aryx I checked the first 3 tests (jekyll, highcharts, NodeGoat), which were all expecting parsing errors in the `FAILING_REPOS` list in `public_repos.py`. Parsing no longer fails, which is reported as a problem.

I suggest adding a "strict" mode to `semgrep-core` so as to fail on a file that can't be parsed without an error, even if a valid parse tree is returned. The default mode would be a "safe" mode and would only print the errors as warnings.

Once this is in place and the tests run in strict mode, we'll see if all the expectations are satisfied again like before.

aryx (approved):
Excellent.

aryx:
hmm, semgrep-rules was also failing, not just the external github repo parsing stuff.

PR #1878, opened 2020-10-23
Fix CI by being backward compatible and throw an exn when tree-sitter returns an error

aryx:
test plan:
make test
also make test in semgrep-rules now works


aryx:
The perf regression test is failing because we now calls the Pfff parser when we get an error in the Tree-sitter one, instead of just returning always an AST and skipping the errors.

Issue #1879, opened 2020-10-23
Match code even if the file contains parse errors, leverage tree-sitter error-recovery mechanism


aryx:

aryx:
Most of the work has been done by martin before.
This just finishes the task by integreating martin's work at the right place in semgrep-core.

PR #1880, opened 2020-10-23
Leveraging tree-sitter error recovery in semgrep-core

aryx:
This is just the first step to help
https://github.com/returntocorp/semgrep/issues/1879

test plan:
make
make test


aryx:
Not sure how the baseline is computed, but the perf regression is probably due again to the fact that we 
now try also the pfff parser where before we would not call it.

PR #1884, opened 2020-10-23
Test files showing we now visit nested JSX elements

### LINKED ISSUE #1841: Allow JSX patterns to be matched against nested JSX elements
https://semgrep.dev/s/erlE should match

aryx:
Fixes https://github.com/returntocorp/semgrep/issues/1841

test plan:
test file included
make
make test
also tried in semgrep and against semgrep-rules and no regressions


aryx:
There's perf regression CI failure, but I don't see how this diff entails a perf regression.

PR #1887, opened 2020-10-23
Fix off by one in autofix

### LINKED ISSUE #1721: Autofix inadvertently changing code outside the match

brendongo:
Also modified autofix tests to actually verify inline autofix

Closes https://github.com/returntocorp/semgrep/issues/1721

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Added
 
 ### Fixed
+- More off by one fixes in autofix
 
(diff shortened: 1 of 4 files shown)


DrewDennison (approved):

brendongo:
Perf test failure expected after ts change

PR #1891, opened 2020-10-23
adding the flag disable-language-overrides

colleend:
The --disable-language-overrides flag disables js rules from running on ts files automatically.

**Test Plan:**

To test this, create a repository that has a ts rule and ts code. In the ts rule, make sure the `languages` key only includes language javascript.

Use

`python -m semgrep -f ../semgrep-rules/ts/<rule file> ../semgrep-rules/ts`

to check that the rule will run.

Then use

`python -m semgrep --disable-language-overrides -f ../semgrep-rules/ts/helloworld.yaml ../semgrep-rules/ts`

to check that the rule doesn't run when the disable-language-overrides flag is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## Next
 
 ### Added
+- disable-language-overrides flag to stop semgrep from automatically running javascript
+  rules on typescript files.
 
 ### Fixed
 - More off by one fixes in autofix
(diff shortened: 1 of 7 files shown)


DrewDennison:
thanks @colleend. After a quick chat with @brendongo, I think we should save this for either a patch release or go back to the drawing table for js / ts since it's very unlikely we will need this flag and it clutters the help / dev UX.

colleend:
Is there any update on whether we want to add js/ts on every js rule or keep automatically running js rules on ts files?

DrewDennison:
yeah let's keep automatically running js rules on ts files but Brendon didn't think we needed to merge this until / unless there is a problem which is unlikely

colleend:
ok, we can save this until there's a user request then

PR #1894, opened 2020-10-26
Upgrade javascript and typescript grammars from tree-sitter

mjambon:
I left some todo notes for the constructs that need changes (I think?) in the AST.


aryx (approved):

aryx:
ok Down from 6 to 2 errors on semgrep-app!

semgrep warn: parse error
  --> frontend/src/components/AddToPolicyButton.tsx:19
19 |   const [policies, setPolicies] = useState<Policy[] | undefined>();
   |                                                                ^^^
Could not parse AddToPolicyButton.tsx as ts

semgrep warn: parse error
  --> frontend/src/react-bootstrap-table2-editor.d.ts:2
2 |   export default function cellEditFactory(data: any): void;
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Could not parse react-bootstrap-table2-editor.d.ts as ts

ran 1 rules on 163 files: 0 findings

PR #1898, opened 2020-10-27
Add examples for the 'generic' cheatsheet

mjambon:
Add examples for `generic` language which is the user-facing name for spacegrep. These examples are used to populate the cheatsheet on the semgrep.dev website. The `GENERIC` folder was renamed to `POLYGLOT` to avoid confusion and problems on case-insensitive file systems.

This is part of adding support for `generic` on semgrep.dev, see https://github.com/returntocorp/semgrep-app/pull/950 (private repo)


aryx (approved):
Great! POLYGLOT is a better name indeed.
Nice tests also, powerful!

PR #1913, opened 2020-10-29
Spacegrep - fix a parsing bug and a matching bug

### LINKED ISSUE #1896: spacegrep bug - indented ellipsis should match empty block
Pattern:
```
{
  ...
}
```

Document 1, doesn't match:
```
{
}
```

Document 2, matches:
```
{}
```

Both documents should match. Use `spacecat --debug` to see the difference between document 1 and document 2.

mjambon:
Fixes https://github.com/returntocorp/semgrep/issues/1896

This is all very technical and not necessarily interesting. It fixes a bug I found by accident while testing the spacegrep integration. I'm working on adding tests and discovering/fixing more bugs.

### DIFF
--- a/spacegrep/src/lib/Match.ml
+++ b/spacegrep/src/lib/Match.ml
@@ -110,6 +110,16 @@ let doc_matches_dots ~dots last_loc doc =
       else
         None
 
+let rec pat_matches_empty_doc pat =
+  match pat with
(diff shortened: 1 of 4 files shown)


aryx (approved):

colleend (approved):
nice!

PR #1921, opened 2020-10-29
Remove semgrep-core/finding/, and enable back LSP

aryx:
The file filtering is now handled in the Python wrapper so
we don't need this code anymore. Morever the code uses
the Dune.glob library which conflicts with the LSP library, so
time to remove it anyway.

test plan:
make


aryx:
The semgrep failure is normal because we use the docker returntocorp/semgrep:develop image, which needs to be updated to not fail ... mutual recursivity problem here (we have the same problem with semgrep-rules/).

aryx:
@mjambon I get a few CI failures in the "Push Semgrep docker image". I think it's because we don't do an opam update there, and so we don't see the recent "lsp" package that I now need.

mjambon:
The reason for not doing `opam update` from the dockerfile was due to someone's inability to access some remote git repository requested by `opam update`, due to some firewalling. The solution was to not run `opam update` at all, which is incorrect but makes the build more reproducible. A bunch of packages are also potentially downloaded from all over the place during `opam install` but this is mostly solved by having the packages pre-installed in the base image.

Anyway, we should be able to switch to a new base image to solve this particular problem. The image used by `Dockerfile` is explicitly tagged with a date, again to help with reproducibility when a user runs into a problem. That's why we have a problem with the dockerfile and not with the rest of CI jobs, which use the latest weekly base image.

mjambon:
I edited the `Dockerfile` to use the latest base image from https://hub.docker.com/repository/registry-1.docker.io/returntocorp/ocaml/tags?page=1

aryx:
Thx martin

PR #1926, opened 2020-10-29
Move 'todook' test annotation to look like other annotations

mschwager:
I'm working on https://github.com/returntocorp/semgrep-docs/issues/22, and it's much easier if all our testing annotations look the same :+1:


brendongo (approved):

mschwager:
I tested this against our lone `todook` test and it worked as expected: https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/correctness/useless-comparison.py. I'll update the test once this is in.

ievans (approved):

PR #1929, opened 2020-10-30
Fix impossible matching on files not containing a newline

### LINKED ISSUE #1928: Can't match file with no newline with spacegrep (generic mode)
Repro:
```
$ echo -n x > foo.generic
$ semgrep -l generic -e x foo.generic
```

but it should find one match, as in the following:
```
$ echo > bar.generic
$ echo -n x >> bar.generic
$ semgrep -l generic -e x bar.generic
bar.generic
2:x
```

The `spacegrep` command line shows a match when `semgrep` doesn't:
```
$ echo -n x | spacegrep x --output-format semgrep
{
  "matches": [
    {
      "check_id": "0",
      "path": "<stdin>",
      "end": { "line": 1, "col": 2, "offset": 1 },
      "extra": { "metavars": {}, "lines": [ "x" ] }
    }
  ],
  "errors": [],
}
```

Impact: prevents "hello world" examples from working in the live editor. Seems like any program containing a newline character is ok, though.

mjambon:
This was an off-by-one error in counting lines, leading to the conclusion that the file was gibberish due to having too much material per line.

Spacegrep will now print something on stderr if a file is skipped.

Also, I added support in spacecat for reading from files instead of stdin, and fixed the spacing between match results just so the output looks better.

Fixes https://github.com/returntocorp/semgrep/issues/1928

### DIFF
--- a/spacegrep/src/bin/Spacecat_main.ml
+++ b/spacegrep/src/bin/Spacecat_main.ml
@@ -9,6 +9,7 @@ open Spacegrep
 type config = {
   is_pattern: bool;
   debug: bool;
+  input_files: string list;
 }
 
 let is_pattern_term =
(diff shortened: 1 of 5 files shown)


brendongo:
Does the python side read and do anything to spacegrep stderr?

aryx (approved):

PR #1939, opened 2020-10-31
Fix matching of trailing dots

mjambon:
Writing a test for this required the ability, in unit tests, to compare expected matches with the actual matches. So I added this functionality. Previously we were just counting how matches were returned without looking what they captured.

So, the failing test was the following:
```
$ seq 0 20
0
1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
$ seq 0 20 | spacegrep '5...'
5
```

The spacegrep result `5` above is incorrect. The dots should match up to 10 lines after the last item captured before the dots. The result should be:
```
$ seq 0 20 | spacegrep '5...'
5
6
7
8
9
10
11
12
13
14
15
```

It now works correctly.

The value of this fix is primarily for pedagogical purposes, because it shows that the dots match at most 10 lines. Trailing dots are otherwise not expected to be a useful feature.

### DIFF
--- a/spacegrep/src/bin/Spacegrep_main.ml
+++ b/spacegrep/src/bin/Spacegrep_main.ml
@@ -55,7 +55,7 @@ let run_all ~debug ~force ~output_format ~highlight patterns docs =
                 printf "match document from %s against pattern from %s\n%!"
                   (Src_file.source_string doc_src)
                   (Src_file.source_string pat_src);
-              (pat_id, Match.search pat doc)
+              (pat_id, Match.search doc_src pat doc)
             ) patterns
       in
       (doc_src, matches)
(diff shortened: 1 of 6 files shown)


jonatanmurga (approved):

jonatanmurga (approved):

aryx:
Who is @jonatanmurga ? and how can he approves those pull requests without being a reviewer?

aryx (approved):

mjambon:
@aryx it looks like unsolicited github users can leave a review if they want but their approval doesn't make the branch mergeable. What happened on the other PR is that @jonatanmurga left a comment as part of an unsolicited review, but I still had to force-merge the branch, using my admin rights.

PR #1943, opened 2020-11-01
spacegrep - detect binary files more aggressively + tests

mjambon:
This allows a `spacegrep` scan to exclude all binary files by looking at the frequency of ascii control characters, which is an improvement over what we had before (which is just based on the frequency of `\n` and doesn't always work to exclude binary files). This is useful when using the spacegrep command recursively e.g.
```
spacegrep 'eval(...)' -d .
```


aryx (approved):
Cool!

PR #1951, opened 2020-11-03
Improve the examples documenting the generic mode in the live editor.

mjambon:
I added more examples of things that work and things that don't work for the generic mode. These examples show up as the cheatsheet on the live editor page. Since "generic" is not a programming language and it doesn't really know any syntax, this is a completely new beast for a user.

Please let me know if you have a suggestion to make the discovery of the generic mode easier.


aryx (approved):
Excellent. Helps to see the limitations compared to semgrep.

PR #1959, opened 2020-11-04
Fix not running multiple configs.

### LINKED ISSUE #1912: Semgrep with multiple configs only uses the last one.
**Describe the bug**
Semgrep with multiple configs only uses the last one.

**Screenshots**
image

P2

**Environment**
pip install, v0.29.0

minusworld:
Fixes https://github.com/returntocorp/semgrep/issues/1912.

Patch 'config_id' with an incrementally increasing integer. Added unit test to make sure two configs are resolved.

### DIFF
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -85,8 +85,13 @@ class Config:
         for i, config in enumerate(configs):
             try:
                 # Patch config_id to fix https://github.com/returntocorp/semgrep/issues/1912
+                resolved_config = resolve_config(config)
+                if not resolved_config:
+                    logger.debug(f"Could not resolve config for {config}. Skipping.")
+                    continue
+                # Extract key and value
                 resolved_config_key, resolved_config_yaml_tree = next(
-                    iter(resolve_config(config).items())
+                    iter(resolved_config.items())
                 )
                 patched_resolved_config: Dict[str, YamlTree] = {}
                 patched_resolved_config[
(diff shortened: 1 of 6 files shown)


semgrep/semgrep/config_resolver.py
```diff
@@ -82,9 +82,23 @@ def from_config_list(
             except SemgrepError as e:
                 errors.append(e)
 
-        for config in configs:
+        for i, config in enumerate(configs):
             try:
-                config_dict.update(resolve_config(config))
+                # Patch config_id to fix https://github.com/returntocorp/semgrep/issues/1912
+                resolved_config = resolve_config(config)
```
brendongo, line 88: Can we push the fix down to resolve_config level? I think it just makes the key a constant when downloading but we can add some part of the url to the key?

brendongo (approved):
I'm fine with merging as is. Up to you if you want to push fix down a layer

semgrep/semgrep/config_resolver.py
minusworld: Yeah, but we also have to do it for files, dirs, etc.
I do agree that moving it down is cleaner. I'll spend 15 minutes looking at it to see if an easy implementation jumps out; otherwise I'll merge as-is.

minusworld: I gave it another shot pushing the fix down; something about doing so caused the rule_ids not to be output properly...

image

Merging as-is.

Issue #1949, opened 2020-11-02
Error when trying to run semgrep that installed with python


omerlh:
**Describe the bug**
When installing semgrep via pip, I am getting the following error:

semgrep --debug views/
running 1 rules from 1 config .semgrep.yml 
running 1 rules...
Running rule find-missing-sri...
spacegrep output could not be parsed as JSON.
Traceback (most recent call last):
  File "/usr/local/bin/semgrep", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/semgrep/__main__.py", line 17, in main
    cli()
  File "/usr/local/lib/python3.9/site-packages/semgrep/cli.py", line 378, in cli
    semgrep.semgrep_main.main(
  File "/usr/local/lib/python3.9/site-packages/semgrep/semgrep_main.py", line 223, in main
    rule_matches_by_rule, debug_steps_by_rule, semgrep_errors, num_targets = CoreRunner(
  File "/usr/local/lib/python3.9/site-packages/semgrep/core_runner.py", line 531, in invoke_semgrep
    findings_by_rule, debug_steps_by_rule, errors, num_targets = self._run_rules(
  File "/usr/local/lib/python3.9/site-packages/semgrep/core_runner.py", line 499, in _run_rules
    rule_matches, debugging_steps, errors, rule_targets = self._run_rule(
  File "/usr/local/lib/python3.9/site-packages/semgrep/core_runner.py", line 390, in _run_rule
    output_json = run_spacegrep(patterns, targets)
  File "/usr/local/lib/python3.9/site-packages/semgrep/spacegrep.py", line 49, in run_spacegrep
    errors.extend(output_json["errors"])
KeyError: 'errors'

The same config file with the same source files and same semgrep version is working is expected when semgrep is installed with brew or running inside a container

aryx:
@mjambon ? "spacegrep output could not be parsed as JSON."

mjambon:
This was a packaging bug, sorry about that. This will be fixed in the next release of semgrep (0.31.0) which should happen next week.

Issue #1937, opened 2020-10-30
Wrong spacegrep executable installed by pip (was: Adding support for CFML)


reillybr:
**Describe the solution you'd like**

It would be great to have support for Coldfusion/CFML in semgrep.  (Edit:  I'm looking at #1392 as well for some guidance on trying to take a shot at adding it.)

I can successfully run basic rules on CFML code in the web Live Editor, but it doesn't seem to work on the command line.  

Error output:

% **semgrep -f test.yaml vuln.cfm**
running 1 rules...
spacegrep output could not be parsed as JSON.
Traceback (most recent call last):
  File "/usr/local/bin/semgrep", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/semgrep/__main__.py", line 17, in main
    cli()
  File "/usr/local/lib/python3.8/dist-packages/semgrep/cli.py", line 378, in cli
    semgrep.semgrep_main.main(
  File "/usr/local/lib/python3.8/dist-packages/semgrep/semgrep_main.py", line 223, in main
    rule_matches_by_rule, debug_steps_by_rule, semgrep_errors, num_targets = CoreRunner(
  File "/usr/local/lib/python3.8/dist-packages/semgrep/core_runner.py", line 531, in invoke_semgrep
    findings_by_rule, debug_steps_by_rule, errors, num_targets = self._run_rules(
  File "/usr/local/lib/python3.8/dist-packages/semgrep/core_runner.py", line 499, in _run_rules
    rule_matches, debugging_steps, errors, rule_targets = self._run_rule(
  File "/usr/local/lib/python3.8/dist-packages/semgrep/core_runner.py", line 390, in _run_rule
    output_json = run_spacegrep(patterns, targets)
  File "/usr/local/lib/python3.8/dist-packages/semgrep/spacegrep.py", line 49, in run_spacegrep
    errors.extend(output_json["errors"])
KeyError: 'errors'

--
SAMPLE CODE (vuln.cfm):

```
<h3>cfexecute</h3>
<!--- This executes a command and outputs it to a file --->

<cfexecute name = "#form.command#"
arguments = "#form.args#"
outputFile = "C:\Temp\output.txt"
timeout = "1">

<!--- This reads and outputs the entire file contents. --->
<cfscript>
myfile = FileRead(#url.file#);
WriteOutput("#myfile#");
</cfscript>
```

SAMPLE RULES FILE (test.yaml):

```
rules:
[...]

PR #1966, opened 2020-11-05
Add error handling to spacegrep

minusworld:
See Slack convo: https://returntocorp.slack.com/archives/C6A1GN4P3/p1604543123342900, in which confusing errors were produced.

### DIFF
--- a/semgrep/semgrep/spacegrep.py
+++ b/semgrep/semgrep/spacegrep.py
@@ -43,7 +43,6 @@ def run_spacegrep(patterns: List[Pattern], targets: List[Path]) -> dict:
                 p = sub_run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
                 p.check_returncode()
                 raw_output = p.stdout
-                raw_error = p.stderr
 
                 output_json = _parse_spacegrep_output(raw_output)
                 output_json["matches"] = _patch_id(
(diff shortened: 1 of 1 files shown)


minusworld:
Help with https://github.com/returntocorp/semgrep/issues/1937.
Fixes https://github.com/returntocorp/semgrep/issues/1949.

semgrep/semgrep/spacegrep.py
```diff
@@ -38,15 +39,32 @@ def run_spacegrep(patterns: List[Pattern], targets: List[Path]) -> dict:
                 str(target),
                 pattern_str,
             ]
-            p = sub_run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-            raw_output = p.stdout
-            raw_error = p.stderr
+            try:
+                p = sub_run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                p.check_returncode()
```
brendongo, line 44: Is there a reason to use check_returncode then catch the exception instead of checking non-zero exit code in a conditional?

brendongo (approved):

semgrep/semgrep/spacegrep.py
minusworld: Primarily because I don't know what the return code will be (except for 0), so I'm not sure how to handle any non-zero return codes. My thought was to use this `check_returncode` shorthand, which raises a `CalledProcessError`, and then propagate the error upward with an `except` clause for now.

mjambon (approved):
thank you guys

PR #1984, opened 2020-11-06
Clarify CLI documentation for '--include' and '--no-git-ignore'.

mjambon:


semgrep/semgrep/cli.py
```diff
@@ -102,14 +102,30 @@ def cli() -> None:
         "--include",
         action="append",
         default=[],
-        help="Scan only files or directories that match this pattern; --include='*.jsx' will scan"
-        " the following: foo.jsx, src/foo.jsx, foo.jsx/bar.sh. --include='src' will scan src/foo.py"
-        " as well as a/b/src/c/foo.py. Can add multiple times.",
+        help="Filter files or directories by path. The argument is a pattern"
```
brendongo: According to python docs they are called "glob-style patterns" https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.match idk if we care to add that detail.

semgrep/semgrep/cli.py
```diff
@@ -102,14 +102,30 @@ def cli() -> None:
...
-        help="Scan all files even those ignored by a projects gitignore(s)",
+        help="Don't skip files ignored by git."
+        " Scanning starts from the root folder specified on the semgrep"
+        " command line."
+        " Normally, if the scanning root is within a git repository, "
+        " only the tracked files and the new files"
+        " would be scanned. Git submodules and git-ignored files would"
+        " normally be skipped."
+        " This option will disable git-aware filtering."
+        " It does nothing if the scanning root is not"
```
brendongo:
```suggestion
        " Setting this flag does nothing if the scanning root is not"
```

brendongo (approved):

PR #1989, opened 2020-11-09
adding conversion from Kotlin CST to generic for literals.

colleend:
Try making!

* not sure about uni_character_literal, but I think it should be something along these lines. Converting uni_character_literal to returning a str * tok resulted in having to put `todos` in other places because they were expected `uni_character_literal` to return a `tok`. 
* Got rid of extra spaces between comments and start of line.
* Also the issue we were seeing earlier was due to me renaming `Unsi_lit` to `Unsigned_lit` in line 407 😓 .

### DIFF
--- a/semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
@@ -1,36 +1,38 @@
- (* PUT YOUR NAME HERE
-  *
-  * Copyright (c) PUT YOUR COPYRIGHT HERE
-  *
-  * This program is free software; you can redistribute it and/or
-  * modify it under the terms of the GNU General Public License (GPL)
-  * version 2 as published by the Free Software Foundation.
-  *
-  * This program is distributed in the hope that it will be useful,
-  * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
```diff
@@ -270,10 +272,10 @@ let multi_line_string_content (env : env) (x : CST.multi_line_string_content) =
...
 
 let uni_character_literal (env : env) ((v1, v2, v3) : CST.uni_character_literal) =
-  let v1 = token env v1 (* "\\" *) in
-  let v2 = token env v2 (* "u" *) in
-  let v3 = token env v3 (* pattern [0-9a-fA-F]{4} *) in
-  todo env (v1, v2, v3)
+  let v1 = str env v1 (* "\\" *) in
+  let v2 = str env v2 (* "u" *) in
+  let v3 = str env v3 (* pattern [0-9a-fA-F]{4} *) in
+  fst v3, PI.combine_infos (snd v1) [snd v3]
```
aryx, line 278: You can concat the 3 fst and combine the 3 snd

aryx (approved):

aryx:
The goal is to be able to -dump_ast the smallest kotlin program

Issue #1947, opened 2020-11-01
Other occurrence of expression is not recognized


msorens:
**Describe the bug**
If I change $ITEM2 in the rule to just $ITEM it should work, but it does not match.

<img width="792" alt="image" src="https://user-images.githubusercontent.com/6817500/97812325-74010380-1c35-11eb-97de-4811a12a04ab.png">

**To Reproduce**
https://semgrep.dev/s/6WXK/

Note: This sample actually contains 3 bugs; filing separately.

**Expected behavior**
Should get a match if `$ITEM2` is replaced with `$ITEM` in the rule.

**What is the priority of the bug to you?**
Is this a P0 (blocking your adoption of Semgrep or workflow), P1 (important to fix or quite annoying), P2 (regular bug that should get fixed)?
P2

**Environment**
semgrep 0.29.0

mschwager:
This looks like a TS vs. JS issue. E.g. if I run the same rule with JS selected it works as expected: https://semgrep.dev/s/olX0/?version=develop.

cc @aryx @mjambon it looks like our metavariable matching is off for TS vs. the equivalent JS.

aryx:
Nice find @mschwager 
Apparently the tree-sitter-based parser (which is used when running in TS mode) does not return the right thing for a foreach with a const iterator. The one in pfff (which is used when running in JS mode because tree-sitter-javascript fails on the code because it contains type annotations), returns the right thing, hence the different behavior in JS vs TS mode.

aryx:
works now on develop: https://semgrep.dev/s/ZWOw/?version=develop

PR #2015, opened 2020-11-11
[JS/TS] Generate correct AST for foreach (const x of y)

aryx:
We were using the multivardef_pattern special variable to
transform the above definition in
foreach (!MULTIVARDEF = x of y)

but there's no need to use multivardef_pattern when the
pattern is a simple variable

test plan:
test file included
make test


aryx:
Fixes https://github.com/returntocorp/semgrep/issues/1947

PR #2019, opened 2020-11-11
fix: semgrep was only finding one config file when passed dir

brendongo:
Code assumed that only a single config was being returned by resolve_config but when passed a directory it may contain many configs.


DrewDennison (approved):

ajinabraham:
Thanks for the quick fix and added tests.

PR #2022, opened 2020-11-11
Fix failing homebrew-core release action

brendongo:
Looks like linuxbrew and homebrew can sometimes drift and result in
the homebrew bump formula action to fail. Suggested workaround on
https://github.com/dawidd6/action-homebrew-bump-formula/issues/18
is to use macos-latest instead of ubuntu-latest so the action will
not try to use linuxbrew


colleend (approved):
🙏

Issue #1719, opened 2020-09-22
Go Object Matching


colleend:
**Describe the bug**
Go object matching with ellipsis on either side is not matching as expected. 

**To Reproduce**
https://semgrep.dev/s/ere0/

**Expected behavior**
Would expect the pattern to match on the http.Request object.

aryx:
I can't load your snippet. I get "Error loading snippet 'ere0': o.replaceAll is not a function"
@chmccreery might know why.

chmccreery:
Strange- I think I know what line of code may have caused the error you just mentioned, @aryx, but I am not able to reproduce it (the snippet loads just fine for me?)

colleend:
Oh interesting, whenever I try to load the snippet again, I'm also getting the same error @aryx was getting...I'll try to make another semgrep.live link then.

chmccreery:
@colleend do you want to screenshare with me and debug together? I think I know the problem roughly, but since I can't reproduce it locally, it would help to be on the call with you.

colleend:
Talked to @chmccreery , it was actually the version of JS our web browsers are running that resulted in that error.

chmccreery:
We are changing the code so that any web browser can load snippets. Sorry about that, and thanks for filing the issue, @aryx! Fixed in: https://github.com/returntocorp/semgrep-app/pull/685

aryx:
@colleend want to take a look at it? Sounds similar to other bugs related to ellipsis you fixed with other languages.
As before you can dump the pattern AST and code AST and figure out in Generic_vs_generic why it does not match.
Note that I've added some logging capability so feel free to ask me if you need to setup logging.

colleend:
yes I'll take a look! Yay for logging 🥳

aryx:
This now works on develop: https://semgrep.dev/s/ere0/?version=develop

PR #2027, opened 2020-11-12
[Go] Generate correct AST for init key values in new expression

aryx:
Parse_go_tree_sitter.ml was incomplete and didn't handle a case that led
to init key values in new expressions (e.g., &foo {fld1: 1, fld2: 2}) to
not be present in the AST.
It was correctly done in the pfff parser, which led to a mismatch
between the AST of the pattern and the AST of the code.

test plan:
test file included
make test


aryx:
Fixes https://github.com/returntocorp/semgrep/issues/1719

PR #2030, opened 2020-11-12
Allow unqualified attribute pattern to match imported attribute

### LINKED ISSUE #2029: Name resolution should not prevent simple pattern to match the same code
This address this previously reported issue:
https://github.com/returntocorp/semgrep/issues/1877#issuecomment-722449548

aryx:
Fixes https://github.com/returntocorp/semgrep/issues/2029

We used to force people to use fully qualified attribute as in
@org.com.Attr
class X { ...}

to match code like

import org.com.Attr;
@Attr
class X { }

but some users find this confusing.

test plan:
test file included


aryx:
We should probably do that not just for imported attributes but any imported entities
so code like foo(...) would match code where foo was imported, not requiring the user to fully qualify foo in the pattern.
Sometime semgrep should be like grep :)

mjambon (approved):

Issue #1981, opened 2020-11-06
indefinite hang when running r/javascript rule


ievans:
**Describe the bug**

Semgrep hangs for 20+ minutes when running r/javascript on itself. The first rule is `javascript.jquery.security.audit.jquery-insecure-selector.jquery-insecure-selector`.

**To Reproduce**

```
ine@imbp4 ~/D/r/semgrep (ie/fix-help) [130]> time semgrep --config="r/javascript"
using config from https://semgrep.dev/r/javascript. Visit https://semgrep.dev/registry to see all public rules.
downloading config...
running 124 rules...
  0%|                                                                                                                                                                        |0/124  0%|                                                                                                                                                                        |0/124
Traceback (most recent call last):
  File "/usr/local/bin/semgrep", line 33, in <module>
    sys.exit(load_entry_point('semgrep==0.30.0', 'console_scripts', 'semgrep')())
  File "/usr/local/Cellar/semgrep/0.30.0/libexec/lib/python3.9/site-packages/semgrep/__main__.py", line 17, in main
    cli()
  File "/usr/local/Cellar/semgrep/0.30.0/libexec/lib/python3.9/site-packages/semgrep/cli.py", line 378, in cli
    semgrep.semgrep_main.main(
  File "/usr/local/Cellar/semgrep/0.30.0/libexec/lib/python3.9/site-packages/semgrep/semgrep_main.py", line 223, in main
    rule_matches_by_rule, debug_steps_by_rule, semgrep_errors, num_targets = CoreRunner(
  File "/usr/local/Cellar/semgrep/0.30.0/libexec/lib/python3.9/site-packages/semgrep/core_runner.py", line 531, in invoke_semgrep
    findings_by_rule, debug_steps_by_rule, errors, num_targets = self._run_rules(
  File "/usr/local/Cellar/semgrep/0.30.0/libexec/lib/python3.9/site-packages/semgrep/core_runner.py", line 499, in _run_rules
    rule_matches, debugging_steps, errors, rule_targets = self._run_rule(
  File "/usr/local/Cellar/semgrep/0.30.0/libexec/lib/python3.9/site-packages/semgrep/core_runner.py", line 424, in _run_rule
    findings_for_rule, debugging_steps = evaluate(
  File "/usr/local/Cellar/semgrep/0.30.0/libexec/lib/python3.9/site-packages/semgrep/evaluation.py", line 362, in evaluate
    valid_ranges_to_output = evaluate_expression(
[...]

PR #2045, opened 2020-11-14
Set default timeout of 30 seconds on semgrep-core calls from semgrep

mjambon:
This is instead of no timeout.

I don't know why `pytest` hangs for me during `tests/e2e/test_semgrep_rules_repo.py`. Let's see if the same thing happens in CI.

Works around https://github.com/returntocorp/semgrep/issues/1981


mjambon:
It would be good to have a test that ensures that the timeout works as it should, especially since both `semgrep` and `semgrep-core` have their own timeout option on the command line.

Another issue is that `semgrep-core` honors the timeout value I specify with `-timeout 3` but `semgrep` will not time out before about 6 seconds. So, if I specify `--timeout 10`, it works as expected:
```
$ time semgrep -l js -e 'function(...,$Z,...) {... $Y = <... $Z ...>; ...}' ../semgrep-core/tests/OTHER/PERF/swagger-ui-bundle.js --timeout 10
Warning: 1 timeout error(s) in ../semgrep-core/tests/OTHER/PERF/swagger-ui-bundle.js when running the following rules: [-]

real	0m10.454s
user	0m10.003s
sys	0m0.483s
```
... but if I pass `--timeout 2` (or another value up to 6), it still takes 6 seconds to return:
```
(semgrep) 18:54:36 martin@mj2 [timeout] ~/semgrep/semgrep $ time semgrep -l js -e 'function(...,$Z,...) {... $Y = <... $Z ...>; ...}' ../semgrep-core/tests/OTHER/PERF/swagger-ui-bundle.js --timeout 2
Warning: 1 timeout error(s) in ../semgrep-core/tests/OTHER/PERF/swagger-ui-bundle.js when running the following rules: [-]

real	0m6.058s
user	0m0.332s
sys	0m0.080s
```

aryx:
Would be good to investigate the 6s vs 2s.
I use export SEMGREP_CORE_DEBUG=1 and semgrep --verbose to see what's going on.

semgrep/semgrep/cli.py
```diff
@@ -159,9 +160,9 @@ def cli() -> None:
     config.add_argument(
         "--timeout",
         type=int,
-        default=0,
+        default=DEFAULT_TIMEOUT,
```
aryx, line 163: Weird that you have to specify the default value here and in the main() function in semgrep_main.py
@nbrahms some Python tricks to avoid the duplication?

aryx (approved):

mschwager:
This is a breaking change. Consider that files taking 5+ seconds to analyze will now be skipped. This means files that users were previously analyzing will no longer be analyzed, which could come as a surprise.

Instead of hardcoding a timeout is it possible to optimize the analysis?

aryx:
Yes it would be good to optimize the analysis, but that would require far more time. My guess is that if the analysis does not finish in 5s, that means you are already in a patological case where it's just annoying to have to wait 10 or 20 minutes or more when everything else is fast.

DrewDennison:
5s seems too low to me. I think we want at least 1 stdev so maybe something like 30 seconds for timeout?

DrewDennison (changes requested):
blocking so we don't skip a bunch of files if 5 seconds is too low.

semgrep/semgrep/constants.py
```diff
@@ -17,6 +17,8 @@
 DEFAULT_CONFIG_FILE = f".{DEFAULT_SEMGREP_CONFIG_NAME}.yml"
 DEFAULT_CONFIG_FOLDER = f".{DEFAULT_SEMGREP_CONFIG_NAME}"
 
+DEFAULT_TIMEOUT = 5  # seconds
```
DrewDennison:
```suggestion
DEFAULT_TIMEOUT = 30  # seconds
```

mjambon:
I can make it 30 seconds. I'll add it to the release notes.

It's still a breaking change, but the benefits outweigh the risks imo.

semgrep/semgrep/cli.py
```diff
@@ -159,9 +160,9 @@ def cli() -> None:
...
         help=(
-            "Maximum time to spend running a rule on a single file in seconds. If set to 0 will not have time limit. Defaults to 0."
+            "Maximum time to spend running a rule on a single file in seconds. If set to 0 will not have time limit. Defaults to {} s.".format(DEFAULT_TIMEOUT)
```
brendongo, line 165:
```suggestion
            f"Maximum time to spend running a rule on a single file in seconds. If set to 0 will not have time limit. Defaults to {DEFAULT_TIMEOUT} s."
```

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## [0.31.1](https://github.com/returntocorp/semgrep/releases/tag/v0.31.1) - 2020-11-11
 
+### Changed
```
brendongo, line 15: Should go under # Next not 0.31.1

mjambon:
I just added timeout support on the spacegrep calls.

mjambon:
Now the tests are failing but I don't know why, or why they weren't failing earlier.

DrewDennison (approved):

mjambon:
I don't know why the changelog reminder is displeased.

I see this warning. I don't see if that's what causes the failure or if it's something else:
```
warning: refname 'pull/2045/merge' is ambiguous.
```
It may have to do with the squashing and force-pushing I did on my commits (I had to set a correct email address for the CLA bot to identify me).

ievans:
@DrewDennison @mjambon  -- will this appropriately set output so that the user will see that some files failed to be processed? E.g. as we do with semgrep-core when it times out, they show up alogn with the parse errors ("N files could not be analyzed, run with --verbose for more info"). If not I am quite worried about silently failing the user

aryx:
I think I return the offending file and rule in the timeout exception returned by semgrep-core.

ievans:
@aryx yes but this is the python wrapper killing the subprocess, so that won't apply

aryx:
I thiught the
Python wrapper was passing the timeout parameter to semgrepcore

aryx:
Both semgrepcore and spacegrep are in control regarding timeout. They are setting the alarm. The python wrapper just
Pass down the parameter to the subcommands

ievans:
@aryx thanks for clarification, after reading the code I see you are right. I thought from the PR desription that this would cause the python wrapper to just kill the subprocess after 30sec but that was a wrong interpretation

mjambon:
Here's what happens in case of a timeout: warning message + exit 0
```
$ semgrep -l js -e 'function(...,$Z,...) {... $Y = <... $Z ...>; ...}' ../semgrep-core/tests/OTHER/PERF/swagger-ui-bundle.js
Warning: 1 timeout error(s) in ../semgrep-core/tests/OTHER/PERF/swagger-ui-bundle.js when running the following rules: [-]
```

PR #2048, opened 2020-11-14
Add '--timeout' option to spacegrep.

### LINKED ISSUE #2046: Honor '--timeout' option in generic mode (spacegrep)
Currently, the time limit for processing a file is set with e.g. `semgrep --timeout 2` but is only honored by the `semgrep-core` back-end. It is not honored by `spacegrep` which invoked for language=generic.

mjambon:
I think it won't work on Windows but it's simple and gets the job done.

Fixes https://github.com/returntocorp/semgrep/issues/2046

### DIFF
--- a/spacegrep/src/bin/Spacegrep_main.ml
+++ b/spacegrep/src/bin/Spacegrep_main.ml
@@ -6,6 +6,8 @@ open Printf
 open Cmdliner
 open Spacegrep
 
+let timeout_exit_code = 3
+
 type when_use_color =
   | Auto
   | Always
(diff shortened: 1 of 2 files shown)


aryx (approved):

spacegrep/src/bin/Spacegrep_main.ml
```diff
@@ -23,6 +25,7 @@ type config = {
   pattern: string option;
   pattern_files: string list;
   doc_files: string list;
+  timeout: int option;
```
aryx, line 28: Note that the semgrep-core accepts actually a float, so you can timeout at less than 1s, but then it's more complicated, you can't use alarm, you need to use a more specialized API (see semgrep-core/cli/Main.ml timeout function).
Note also that in semgrep-core the timeout is per processed file, not the full lenght of the program run.

Maybe you can add a comment that the int unit is in seconds.

Issue #2033, opened 2020-11-12
Inline // nosem not working


chmccreery:
**Describe the bug**
When chef runs Semgrep in CI, they have several lines that get flagged every time, even though they are tagged with comments like // nosem: ruleid

**To Reproduce**
Create a file in go that violates the rule go.grpc.security.grpc-client-insecure-connection.grpc-client-insecure-connection and add a // nosem: go.grpc.security.grpc-client-insecure-connection.grpc-client-insecure-connection to it. Then run semgrep 

**Expected behavior**
This line of code should not show up in Semgrep findings on the commandline

**What is the priority of the bug to you?**
P2 - this was a P0 for msorens, but since // nosem without rule id specified seems to work, he is unblocked now. Others may still become blocked, too, though.

**Environment**
Docker using custom Buildkite setup
Language: Go

mschwager:
It works for me here: https://semgrep.dev/s/B2BA/

I also tested this out locally and it worked there too. :thinking:

msorens:
@mschwager Could you post details of how you tested from the command line locally. It is not working for me and I am seeing some new unusual behavior--it is putting my entire file path in the rule id on the first line here!
```
severity:error rule:Users.msorens.code.go.src.github.com.chef.automate.semgrep.grpc-client-insecure-connection: Found an insecure gRPC connection. This creates a connection without encryption to a gRPC server. A malicious attacker could tamper with the gRPC message, which could compromise the machine.
10:    conn, err := grpc.Dial(address, grpc.WithInsecure()) // nosem: grpc-client-insecure-connection
```

mschwager:
By default the `nosem` functionality requires the full rule-id, e.g.

```
// nosem: Users.msorens.code.go.src.github.com.chef.automate.semgrep.grpc-client-insecure-connection
```

You can disable this functionality with the `--no-rewrite-rule-ids` flag, which will allow:

```
// nosem: grpc-client-insecure-connection
```

msorens:
OK, let's back up a bit. Here are the details of my particular issue.

I have in my file this bit of code:
```
func conn() grpc.ClientConnInterface {
	cc, err := grpc.Dial("localhost:1234", grpc.WithInsecure()) // nosem: go.grpc.security.grpc-client-insecure-connection.grpc-client-insecure-connection
	if err != nil {
		panic(err)
	}
	return cc
}
```

In my CI pipeline I am invoking semgrep-agent...
```
[...]

PR #2055, opened 2020-11-16
Add an "is_ignored" field to JSON output

underyx:
This makes it possible for semgrep-agent to tell when a finding has been
ignored with a nosem comment.

Changes were tested with the included test named
`test_nosem_rule__with_disable_nosem`,
as well as reading through the updated snapshots.

### DIFF
--- a/semgrep/semgrep/semgrep_main.py
+++ b/semgrep/semgrep/semgrep_main.py
@@ -9,6 +9,8 @@ from typing import Optional
 from typing import Set
 from typing import Tuple
 
+import attr
+
 import semgrep.config_resolver
 from semgrep.autofix import apply_fixes
 from semgrep.constants import COMMA_SEPARATED_LIST_RE
@@ -123,9 +125,9 @@ def rule_match_nosem(rule_match: RuleMatch, strict: bool) -> bool:
 
 def invoke_semgrep(config: Path, targets: List[Path], **kwargs: Any) -> Any:
     """
-        Call semgrep with config on targets and return result as a json object
+    Call semgrep with config on targets and return result as a json object
 
-        Uses default arguments of MAIN unless overwritten with a kwarg
+    Uses default arguments of MAIN unless overwritten with a kwarg
     """
     io_capture = StringIO()
     output_handler = OutputHandler(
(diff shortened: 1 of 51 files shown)


underyx:
Force-pushed a changelog entry.

mschwager:
Looks like this is related to https://github.com/returntocorp/semgrep/issues/2033. Although, I don't understand: why does semgrep-agent need to know if a finding has been ignored? Seems like an interesting data point, but shouldn't Semgrep just not return the finding if it's been ignored?

Side note: this would be much easier to review if the refactor and functionality changes were done separately :)

mschwager:
Actually, taking a step back... I don't understand what the original bug here was. Could you briefly describe the issue, and how this solves it?

underyx:
@mschwager we want to be able to show and crunch data about ignored issues in Semgrep Community. This will be able to flip issues from `NEW` to `IGNORED` status in semgrep.dev's db instead of `FIXED` being the only resolution option.

underyx:
I re-ran the benchmark job, not sure what could make semgrep +5% slower here :/

Is the baseline from the current develop version?

underyx:
Seems like the benchmark re-run fixed it!

semgrep/semgrep/semgrep_main.py
```diff
@@ -231,12 +233,18 @@ def main(
 
     output_handler.handle_semgrep_errors(semgrep_errors)
 
+    rule_matches_by_rule = {
```
brendongo, line 236: Should this only run as an else to the conditional below?

underyx (commented):
@brendongo re: your deleted comment:

> Could go either way. I ended up including `is_ignored: false` on each run just to make the JSON output schema more consistent.

brendongo:
> @brendongo re: your deleted comment:
> 
> > Could go either way. I ended up including `is_ignored: false` on each run just to make the JSON output schema more consistent.

Yeah it was more clear on second reading. :D

brendongo (approved):
LGTM. If anything I would change is_ignored to something like has_no_sem or something to make it more obvious what it does but that seems like a nit.

mschwager:
Discussed out-of-band - not related to https://github.com/returntocorp/semgrep/issues/2033. LGTM :+1: although yeah I agree that naming tying this closer to `nosem` would be helpful

underyx:
`ignored_with_nosem`?

underyx:
Though what if we add more methods of ignoring, for instance for ignoring a rule ID for a whole file with a marker on line 1? It's conceivable that new ignore methods wouldn't be named nosem anymore 🤔

underyx:
My issue with just `has_nosem` is that without the word "ignore" someone looking at the JSON output who isn't aware of `// nosem` functionality would have zero idea what this is supposed to mean; they might as well think it's about people's noses or something.

brendongo:
> Though what if we add more methods of ignoring, for instance for ignoring a rule ID for a whole file with a marker on line 1? It's conceivable that new ignore methods wouldn't be named nosem anymore 🤔

hmm makes sense. I can imagine we might want to know how a finding was ignored if there are many ways to ignore. I'm fine with whatever you choose here @underyx can always change this later.

underyx:
So I'll merge as is. I expect we'll be the only users of this field so changing it later won't be likely to hurt anyone.

PR #2058, opened 2020-11-17
WIP: Upgrade ruby grammar

mjambon:
This upgrade of the Ruby grammar went pretty well, I would say faster than I was expecting. It's because most changes didn't extend the AST. They were mostly fixes in the precedences and in which kind of expression is allowed where.

There's just one todo left, which is about how we represent ranges with a missing bound like `..42` or `73...`.


aryx:
There's are regression in semgrep-rules on ruby/lang/security/nested-attributes-bypass.yaml

aryx:
Hmm I think it's a mistake in semgrep-rules, they were missing space in keyword arguments which led now
'foo arg:false'   to be parsed as a foo (arg (:false)   instead of foo (arg: false).
I'll fix semgrep-rules

aryx (approved):

PR #2065, opened 2020-11-17
Include rest of Semgrep subcategories in cheatsheet generation

mschwager:
This will give us all the Semgrep features by language, which will help for language graduation.

### DIFF
--- a/scripts/generate_test_matrix.py
+++ b/scripts/generate_test_matrix.py
@@ -21,8 +21,6 @@ yaml = YAML(typ="rt")
 
 FEATURES = ["dots", "equivalence", "metavar", "misc"]
 
-VERBOSE_REGEXP_SYNTAX = 'OCaml Syntax: "=~/<regexp>/"'
-
 VERBOSE_FEATURE_NAME = {
     "dots": "Wildcard Matches (...)",
     "equivalence": "Helpful Features",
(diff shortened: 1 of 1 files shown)


scripts/generate_test_matrix.py
```diff
@@ -82,14 +81,16 @@
         "import",
         "typed",
     ],
-    "regexp": ["string"],
+    "regexp": ["string", "fieldname"],
```
colleend, line 84: what does 'fieldname' represent?

mschwager: Looks like a regexp in an object key: https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/tests/js/regexp_fieldname.sgrep. It's currently only implemented for JS, but I could see it being used in other languages too (e.g. `dict` key in Python).

ievans:
@mschwager I like. Are we updating the build process so that generate cheatsheet is called as part of the build?

ievans (approved):

mschwager:
> @mschwager I like. Are we updating the build process so that generate cheatsheet is called as part of the build?

Already done! E.g. you can find the "test-matrix" JSON file here: https://github.com/returntocorp/semgrep/actions/runs/368794473

Do we need the HTML file autogenerated as well?

scripts/generate_test_matrix.py
colleend: oh interesting, didn't know that we supported that 😮 , looks good to me!

colleend (approved):

ievans:
Awesome @mschwager  Nah I don't think so!

Issue #2052, opened 2020-11-16
ellipsis operator and metavariables inside imports for TypeScript


inkz:
**Is your feature request related to a problem? Please describe.**
Currently there is no opportunity to catch and track arbitrary `import ... from ...`s in TypeScript
e.g.
trying to find `jwt_decode` library import on code like this

```typescript
import jwt_decode from "jwt-decode";

function ra() {
  return 1
}
```

with pattern like this

```
import jwt_decode from "...";
```
https://semgrep.dev/s/KdR5

wont return results

trying to use pattern like this

```
import jwt_decode from $X;
```
https://semgrep.dev/s/4B6z

will return an error

**Describe the solution you'd like**
It would be very useful to have an opportunity to create patterns for arbitrary imports for TS
e.g.
```
import $X from $Y;
```
or
```
import $Z from "...";
```

PR #2072, opened 2020-11-18
[JS/TS] support import x from "..."

aryx:
Fix part 1 of https://github.com/returntocorp/semgrep/issues/2052

test plan:
test file included


aryx:
the CI failure is due to a TODO in semgrep-rules/ that now works :)
I'll fix it in a separate PR

PR #2074, opened 2020-11-18
Allow ignoring findings with // nosemgrep

underyx:
The original keyword, nosem, isn't very easy to Google or understand if
you aren't yet aware of Semgrep.

For backwards compatibility, this commit doesn't remove the 'nosem'
keyword.

This change was tested by adding more lines to the basic/nosem.py test
target and manually reading through the updated snapshots.

### DIFF
--- a/semgrep/semgrep/constants.py
+++ b/semgrep/semgrep/constants.py
@@ -46,17 +46,18 @@ class OutputFormat(Enum):
 NOSEM_INLINE_RE = re.compile(
     # We're looking for items that look like this:
     # ' nosem'
-    # ' nosem: example-pattern-id'
+    # ' nosemgrep: example-pattern-id'
     # ' nosem: pattern-id1,pattern-id2'
-    # ' NOSEM:pattern-id1,pattern-id2'
+    # ' NOSEMGREP:pattern-id1,pattern-id2'
     #
     # * We do not want to capture the ': ' that follows 'nosem'
     # * We do not care about the casing of 'nosem'
     # * We want a comma-separated list of ids
     # * We want multi-language support, so we cannot strictly look for
     #   Python comments that begin with '# '
+    # * nosem and nosemgrep should be interchangeable
     #
-    r" nosem(?::[\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?",
+    r" nosem(?:grep)?(?::[\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?",
     re.IGNORECASE,
 )
 COMMA_SEPARATED_LIST_RE = re.compile(r"[,\s]")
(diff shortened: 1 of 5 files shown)


underyx:
Thanks @msorens for bringing this to our attention!

semgrep/tests/e2e/targets/basic/nosem.py
```diff
@@ -1,2 +1,6 @@
 test_nosem_func()  # nosem: rules.test-nosem
+test_nosem_func()  # NOSEM: rules.test-nosem
+test_nosem_func()  # nosemgrep: rules.test-nosem
+test_nosem_func()  # NOSEMGREP
+test_nosem_func()  # nOseMgREP: rules.test-nosem
```
underyx, line 5: Screenshot 2020-11-18 at 12 12 40

underyx:
CI failure fixed in https://github.com/returntocorp/semgrep/pull/2075

semgrep/semgrep/constants.py
```diff
@@ -46,17 +46,18 @@ def is_json(self) -> bool:
...
     #
     # * We do not want to capture the ': ' that follows 'nosem'
     # * We do not care about the casing of 'nosem'
     # * We want a comma-separated list of ids
     # * We want multi-language support, so we cannot strictly look for
     #   Python comments that begin with '# '
+    # * nosem and nosemgrep should be interchangeable
     #
-    r" nosem(?::[\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?",
+    r" nosem(?:grep)?(?::[\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?",
```
msorens, line 60: Could make the user experience much less (potentially) frustrating by being more forgiving of whitespace here.

Without this change then `//nosem:  some-rule-id` (two spaces there) will be treated as just `// nosem`.
```suggestion
    r" nosem(?:grep)?(?::[\s]*(?P<ids>([^,\s](?:[,\s]+)?)+))?",
```

mschwager:
From a technical perspective this looks fine to me :+1:

I'll copy a comment I made from an internal conversation:

> I'll preface by saying I don't have too strong of feelings here. The downsides I see: 1.) fragmentation further reduces googleability, 2.) more complicated `nosem` behavior is a bigger maintenance burden and more likely to attract bugs or confusing behavior, 3.) we're lengthening a syntax that's already prone to very long lines, and 4.) linters do already use shorter syntax's - flake8 uses `noqa` as opposed to `noqualityassurance` :man_shrugging:

chmccreery:
I really like this. I find "nosemgrep" way more intuitive than "nosem" which my brain always parses first as "nose m" which makes no sense (#nosense?)

chmccreery (approved):

underyx:
@chmccreery hah! I was just wondering about noses the other day :D https://github.com/returntocorp/semgrep/pull/2055#issuecomment-728294659

ievans:
@underyx love that you jumped on this; should we just remove mention of // nosem from the docs? I'd be in favor of that.

Can we update or at least make a ticket to track updating https://github.com/returntocorp/semgrep-docs/blob/main/docs/ignoring-findings.md too?

mschwager:
Personally I wouldn't pull this in, but I don't have too strong of feelings and won't block. Looks like it's been approved, so I'll let you move forward at your discretion :+1: 

Unassigning myself

DrewDennison (approved):

PR #2076, opened 2020-11-18
Name resolution should not prevent simple expr pattern to match the same code

### LINKED ISSUE #2031: golang package import not binding metavariable
The snippet below works if you use a generic metavariable (eg $X) for the line inside $F, but not if you try to re-reference the $P metavar from the import.

```
import $P "github.com/sirupsen/logrus"

func $F() {
  ...
  $P.WithError(...).Info(...)
  ...
}
```

https://semgrep.dev/s/3e7K

aryx:
We used to force to fully qualify entities but this is confusing,
especially when you use metavar for imported module/entity which prevent
you to reuse the same metavar later.

Related to https://github.com/returntocorp/semgrep/issues/2029
but this time for expression identifiers (Id), not attributes.

Fixes https://github.com/returntocorp/semgrep/issues/2031

Note that semgrep-rules will fail because of some code
that now correctly match but was not tagged as. I'll fix that
later (mutual recursivity issue between CI of semgrep and semgrep-rules).

test plan:
many test files included
make test
x


aryx:
@emjin this is probably also solving the issue you were working on.

aryx:
The CI failure is due to some code that is now correctly matching in semgrep-rules. I'll do a PR for that.

ievans:
@DrewDennison @aryx this seems like an important and breaking change. Are we sure we want to do this? It may suddenly result in a lot more matches for users. At a minimum it should be a headline in the CHANGELOG?

aryx:
We definitely want to do this. There has been many issues closed by this fix, because it's just too confusing for users.
One can still fully qualify an entity and match code that will alias entity! So the nice aliasing equivalence still works! It's just
that we don't want to force it on the user.
Yes we definitely add something in the Changelog.

PR #2084, opened 2020-11-18
php can't render because of /<, />, this lets these examples render

colleend:


mschwager:
Should we use [`<code>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code) instead?

colleend:
hmm so I think even with \<code\> tags, it recognizes <?php> as another tag and doesn't render it as it should 😓 

just checked this, using \<code\> still results in sadness :')

mschwager (approved):
Cool, no worries, just wanted to see if that'd work

PR #2087, opened 2020-11-18
php tests hooray

colleend:


semgrep-core/tests/php/dots_args.php
```diff
@@ -1,8 +1,8 @@
 <?php
-
 function foo() {
   //ERROR:
     foo(1,2,3,4,5);
   //ERROR:
     foo(5);
 }
+?>
```
aryx, line 8: you actually don't need that in practice.
Most php tools will accept code without the end ?>

aryx (approved):
thanks!

semgrep-core/tests/php/dots_args.php
colleend: good to know 👍

PR #2088, opened 2020-11-19
adding binary expr transformation

colleend:
Test plan:
Run `semgrep-core/_build/default/cli/Main.exe --dump_ast tests/kotlin/binary-func.kt`

* Transforms all the binary expressions, along with Super and call references for primary expressions.

_Not sure about:_
* Should Call_ref return DotAccess?
* What should Infix_exp and Check_exp return?


semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
```diff
@@ -516,28 +517,29 @@ and binary_expression (env : env) (x : CST.binary_expression) =
         | `Is_op x -> is_operator env x
         )
       in
+      let v2_id = Id (("", v2), empty_id_info()) in
```
aryx, line 520: We could add an operator for that maybe. It's actually used also in Python (but right now I use
OE_In, and OE_NotIn, but really we could make In and NotIn operators like Plus/... in AST_generic.operator.

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
```diff
@@ -1369,25 +1371,30 @@ and primary_constructor (env : env) ((v1, v2) : CST.primary_constructor) =
...
   | `Call_ref (v1, v2, v3) ->
       let v1 =
         (match v1 with
-        | Some x -> simple_identifier env x
-        | None -> todo env ())
+        | Some x ->
+            let id = simple_identifier env x in
+            Id(id, empty_id_info())
+        | None ->
+            let fake_id = ("None", fake "None") in
```
aryx, line 1386: better to keep the todo than using fake thing when you're not sure.

aryx (approved):

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
colleend: adding in AST_generic sounds great, we could probably also add in the Is/NotIs operators in there too 🤔

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
aryx: yep, probably can add Is and Isnot; I have no idea what it is though. Is it a type check?
maybe we could convert it in an InstanceOf which is part of the 'special' type.

colleend: yep, they're type checks -- if we convert is and isnot both to instanceof, would we need to differentiate between if it's isnot or is when doing matching?

aryx: In theory you can create an Ast with a regular Not before the InstanceOf. But problably simpler
you add Is and IsNot in AST_generic.operator.

colleend: got it! I'll add it to the AST_generic then 😄

PR #2090, opened 2020-11-19
* semgrep-core/parsing/Parse_csharp_tree_sitter.ml: List.concat_map does not exist in 4.09.1

aryx:
test plan:
make now works on my machine


aryx:
cc @Sjord just a few small fixes

PR #2101, opened 2020-11-19
ts tests

colleend:
Mostly the same as js ones.


aryx:
Note that we run all the tests/js/ in typescript mode so it would be good to really make those new test files specific to typescript, like adding types here and there so it looks more like a ts file than a js file.

aryx (approved):

PR #2108, opened 2020-11-21
Enforce "shortest match" rule in spacegrep

mjambon:
Until now, the pattern `function ... foo` would match the whole `function bar function foo` instead of just `function foo`. This is now fixed by starting the search from all possible locations, allowing overlapping matches, but eliminating the longer of two matches that end at the same location.

I also fixed the behavior of patterns with leading dots such as `... foo`. Now, they may only match at the beginning of a block. Consequently, `...` now correctly matches the whole document.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Changed
 
+- In generic mode, shorter matches are now always preferred over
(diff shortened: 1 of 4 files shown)


mjambon:
I found a few issues which I'll fix:

- [x] add test case for nested matches (hadn't found one earlier)
- [x] fix incorrect elimination of a longer match that end at the same location as the current match (need to use a lookup map or table rather than only comparing to the latest match)
- [x] update the semgrep changelog

aryx (approved):

PR #2113, opened 2020-11-23
Do not follow symlinks

### LINKED ISSUE #2005: semgrep-core errors on symlinks
Running against: https://github.com/mapbender/mapbender-starter

```
$ python -m semgrep --pattern '$X == $X' --lang php /tmp/mapbender-starter/ --verbose
	unknown exception: (Sys_error
  "/tmp/mapbender-starter/application/bin/doctrine.php: No such file or directory")
An error occurred while invoking the semgrep engine; please help us fix this by creating an issue at https://github.com/returntocorp/semgrep
```

That file is a symlink:

```
$ ls -alh mapbender-starter/application/bin/doctrine.php 
lrwxrwxrwx 1 user user 39 Nov 10 10:20 mapbender-starter/application/bin/doctrine.php -> ../vendor/doctrine/orm/bin/doctrine.php
```

We should either: A.) follow symlinks in `semgrep-core`, or B.) resolve symlinks in Python before passing them to `semgrep-core`.

nbrahms:
Symlinks break semgrep, as they often point outside of the path actually
mounted into the core library, or are otherwise unresolveable by the
core.

In the tradition of comparable UNIX tools (such as ripgrep), we disable
link following.

In the future, we may want to add a switch to allow link following, but
let's wait until this is actually needed to avoid flag bloat in the
immediate.

Fixes #2005.

### DIFF
--- a/semgrep/semgrep/target_manager.py
+++ b/semgrep/semgrep/target_manager.py
@@ -143,6 +143,10 @@ class TargetManager:
         )
 
     @staticmethod
+    def _is_valid(path: Path) -> bool:
+        return path.exists() and not path.is_symlink()
+
+    @staticmethod
     def _expand_dir(
         curr_dir: Path, language: Language, respect_git_ignore: bool
     ) -> Set[Path]:
@@ -171,7 +175,11 @@ class TargetManager:
             """
             Return set of all files in curr_dir with given extension
             """
-            return set(p for p in curr_dir.rglob(f"*{extension}") if p.is_file())
+            return {
+                p
+                for p in curr_dir.rglob(f"*{extension}")
+                if p.is_file() and TargetManager._is_valid(p)
+            }
 
         extensions = lang_to_exts(language)
         expanded: Set[Path] = set()
@@ -234,7 +242,7 @@ class TargetManager:
         """
         expanded = set()
         for target in targets:
-            if not target.exists():
+            if not TargetManager._is_valid(target):
                 continue
 
             if target.is_dir():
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/target_manager.py
```diff
@@ -171,7 +175,11 @@ def _find_files_with_extention(
             """
             Return set of all files in curr_dir with given extension
             """
-            return set(p for p in curr_dir.rglob(f"*{extension}") if p.is_file())
+            return {
```
mschwager, line 182: Should we apply the same logic to the `_parse_output` loop above? That way we have the same behavior between git tracked files and not.

Also, should we lift `is_file` into `_is_valid`?

mjambon (approved):

semgrep/semgrep/target_manager.py
nbrahms: There's another branch that follows directories, so we don't want to include `is_file`

dlukeomalley:
@nbrahms Drive by question: is there any external indication that a symlink has been skipped to the user? Could see behavior like this causing some confusion (as I have been confused in the path by something I didn't realize was a symlink).

nbrahms:
> @nbrahms Drive by question: is there any external indication that a symlink has been skipped to the user? Could see behavior like this causing some confusion (as I have been confused in the path by something I didn't realize was a symlink).

Not currently I think. `ripgrep` has a `--files` flag, which lists the files that are checked, which we could implement in the future.

PR #2127, opened 2020-11-26
Enable json output with '-json'

### LINKED ISSUE #1469: Core when run with -json does not output json
**Describe the bug**
Running the OCaml core with the `-json` flag does not emit JSON output

**To Reproduce**
Build the core,  then:
```
$ ./_build/default/bin/Main.exe -f tests/js/metavar_try_catch.sgrep -json -lang js tests/js/metavar_try_catch.js
tests/js/metavar_try_catch.js:1
 try {
   const x = 1 / 0;
 } catch {}
```

**Expected behavior**
`-json` should emit json

**Environment**
local build on commit `3f0f93e0ec`

mjambon:
The `-json` option is now honored when the pattern is specified with `-e` or `-f`, that is not with a rules file. However, I didn't get rid of much code like I wanted to. Some legacy code is still used for:

* printing results as text (which is not supported for rules files)
* supporting lang=fuzzy which is not one of the semgrep languages (`Lang.t` etc.)

Anyway, I'm happy that `-json` works so we can see all the details that the text output doesn't show.

I also added pretty-printing for the json output, for user comfort. Normal semgrep usage involves large inputs and a relatively small number of matches, so the cost associated with pretty-printing should not be an issue.

Sample output:
```
~/semgrep/semgrep-core $ semgrep-core -lang java -e 'class $X' tests/java/deep_cond.java -json
{
  "matches": [
    {
      "check_id": "-e/-f",
      "path": "tests/java/deep_cond.java",
      "end": { "line": 1, "col": 10, "offset": 9 },
      "extra": {
        "message": "",
        "metavars": {
          "$X": {
            "start": { "line": 1, "col": 7, "offset": 6 },
            "end": { "line": 1, "col": 10, "offset": 9 },
            "abstract_content": "Foo",
            "unique_id": {
              "type": "AST",
              "md5sum": "1f0c9b0d0f3f351a3dd7a4e7ca77c675"
            }
          }
        }
      }
    }
  ],
  "errors": [],
}
```

Fixes https://github.com/returntocorp/semgrep/issues/1469

### DIFF
--- a/semgrep-core/cli/Main.ml
+++ b/semgrep-core/cli/Main.ml
@@ -122,9 +122,11 @@ let lang = ref "unset"
 (*s: constant [[Main_semgrep_core.include_dirs]] *)
 (*e: constant [[Main_semgrep_core.include_dirs]] *)
 
-(*s: constant [[Main_semgrep_core.output_format_json]] *)
-let output_format_json = ref false
-(*e: constant [[Main_semgrep_core.output_format_json]] *)
+type output_format = Text | Json
+
+(*s: constant [[Main_semgrep_core.output_format]] *)
+let output_format = ref Text
+(*e: constant [[Main_semgrep_core.output_format]] *)
 (*s: constant [[Main_semgrep_core.match_format]] *)
 let match_format = ref Matching_report.Normal
 (*e: constant [[Main_semgrep_core.match_format]] *)
@@ -489,7 +491,7 @@ let create_ast file =
 
 (*s: type [[Main_semgrep_core.pattern]] *)
 type pattern =
-  | PatGen of Pattern.t
+  | PatGen of (Lang.t * Pattern.t)
   (*s: [[Main_semgrep_core.pattern]] other cases *)
   | PatFuzzy of Ast_fuzzy.tree list
   (*e: [[Main_semgrep_core.pattern]] other cases *)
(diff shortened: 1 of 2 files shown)


mjambon:
Some tests fail because they're called with rules_files and without the `-json` option. Maybe `-json` should be implied by the presence of a rules file. I'll look into this after the weekend.

aryx:
Yes I think -rules_file implies -json.
Maybe worth adding explicitely those -json in the python wrapper

semgrep-core/cli/Main.ml
```diff
@@ -626,52 +627,64 @@ let print_matches_and_errors files matches errs =
...
 (*e: function [[Main_semgrep_core.print_matches_and_errors]] *)
 
 (*****************************************************************************)
 (* Semgrep -e/-f *)
 (*****************************************************************************)
-(*s: function [[Main_semgrep_core.semgrep_with_one_pattern]] *)
-(* simpler code path compared to semgrep_with_rules *)
-let semgrep_with_one_pattern xs =
+
+(*s: function [[Main_semgrep_core.legacy_semgrep_with_one_pattern]] *)
```
aryx, line 655: please do not modify those comments starting with s:, e:, x:, They are used by my
literate programming tool syncweb, and the tool gonna get confused by those renames.

aryx (approved):
Looks good. 
Please fix the test failures before merging though.

mjambon:
ok, so pretty-printing the json results with `Yojson.Safe.prettify` worsened performance by an extra 50%. It's surprising. It uses ocaml's `Format` module which can be slow on some large or unusual output. I don't know if we're producing really large code snippets or something. Anyway, I turned off pretty-printing unconditionally to avoid surprises. The user should use `jq` to pretty-print json on the command line (or `ydump` which comes with the yojson library).

PR #2129, opened 2020-11-27
Parse keywords

Sjord:
Parse keywords as NamedAttr if not KeywordAttr exists in pfff.

Should I create an issue or pull request to add these keywords to pfff?


aryx (approved):
Yes you can make a PRs on pfff. I usually put keywords shared by multiple languages
in the keyword_attribute type, and keywords used until now in only one language
in the other_attribute_operator type.
 Here I think override, and virtual could be added to keyword_attribute.
Maybe "ref" is the same than Mutable? Ther other I would put in other_attribute_operator like I said, unless you know their semantic and you can find an equivalent keyword in keyword_attribute.

PR #2130, opened 2020-11-27
Parse anonymous class in C#.

Sjord:
E.g.

```csharp
var a = new { someprop = "Somevalue" };
```

### DIFF
--- a/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
@@ -813,11 +813,15 @@ and expression (env : env) (x : CST.expression) : AST.expr =
       in
       let v4 =
         (match v4 with
-        | Some tok -> token env tok (* "," *)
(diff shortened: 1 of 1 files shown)


semgrep-core/parsing/Parse_csharp_tree_sitter.ml
```diff
@@ -813,11 +813,15 @@ and expression (env : env) (x : CST.expression) : AST.expr =
...
       let v4 =
         (match v4 with
-        | Some tok -> token env tok (* "," *)
-        | None -> todo env ())
+        | Some tok -> Some (token env tok) (* "," *)
+        | None -> None)
       in
       let v5 = token env v5 (* "}" *) in
-      todo env (v1, v2, v3, v4, v5)
+      AnonClass {
```
aryx, line 820: Hmm should we use anonymous objects, a.k.a Record instead of AnonClass?

aryx (approved):

PR #2131, opened 2020-11-27
Parse construction initializer

Sjord:
E.g.

```csharp
   public SampleCollection() : this(3)
   {
   }
```

We transform this into `Call(This, 3)`, and prepend that to the function body.
Perhaps we should actually use the method name in here somewhere, or call
`.ctor`.


semgrep-core/parsing/Parse_csharp_tree_sitter.ml
```diff
@@ -2091,11 +2091,14 @@ and declaration (env : env) (x : CST.declaration) : stmt =
         | None -> None)
       in
       let v6 = function_body env v6 in
+      let fbody = (match v5 with
```
aryx, line 2094: v6 is probably itself a Block, so maybe better to check if it is and prepend the this call inside the Block itself, 
to avoid a Block (Block  ...) (this extra Block might later prevent some matching I think).

aryx (approved):

PR #2134, opened 2020-11-28
Parse using block in C#

Sjord:
E.g.
```csharp
using (Font font3 = new Font("Arial", 10.0f), font4 = new Font("Arial", 10.0f))
{
   // Use font3 and font4.
}
```

We convert the using block to a `try` block, with the declaration outside the
try. This is also what C# does internally. Should we call `Dispose` in the
finally? Can we know on what we should call `Dispose`? Is this the correct way
forward, or should we use `OSWS_With`?


aryx:
Yes, we should define a new Stmt construct and merge the Python OSWS_With construct with that.
Maybe we can call it WithUsingResource? I think Java has something similar too.

aryx (approved):

Sjord:
Thanks, I have filed an issue: [Add using/with to AST_generic.ml · Issue #342 · returntocorp/pfff](https://github.com/returntocorp/pfff/issues/342)

PR #2136, opened 2020-11-30
Prevent crash matching metavar Id to IdSpecial

### LINKED ISSUE #2081: NoTokenLocation error during python fstring parse
Semgrep at develop, running with `$ semgrep -e '$X' -l py semgrep/semgrep` gives me:

```
	unknown exception: Parse_info.NoTokenLocation("Match returned an empty list with no token location information; this may be fixed by adding enclosing token information (e.g. bracket or parend tokens) to the list's enclosing node type.")
```

nbrahms:
A match to a bare IdSpecial(ConcatString, _) will cause
a NoTokenLocation error; instead we must match to Call(IdSpecial, [...])
(from which token information can be taken).

I prevent this match manually.

Fixes #2081.


aryx (approved):
Excellent! Good catch.

PR #2139, opened 2020-11-30
ci: check access to secrets before pushing

brendongo:
PRs from forks will not try and fail to push to dockerhub


brendongo:
Should fix issue where PRs from external contributors keep seeing the failed docker CI job

PR #2145, opened 2020-12-02
added code to generate better table, also added in exceptions to new …

colleend:
* added exceptions to new tests
* added code for rendering new tests
* added code to generate table ordered by Alpha, Beta, GA features
* added code to generate headers for Alpha, Beta, GA groups

### DIFF
--- a/scripts/generate_test_matrix.py
+++ b/scripts/generate_test_matrix.py
@@ -51,14 +51,18 @@ VERBOSE_SUBCATEGORY_NAME = {
     "syntax": "Single Statements",
     "exprstmt": "Expression and Statement",
     "typed": "Typed Metavariables",
+    "class": "Classes",
+    "anno": "Annotations",
 }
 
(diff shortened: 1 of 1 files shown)


minusworld (approved):
\o/

scripts/generate_test_matrix.py
```diff
@@ -93,6 +98,33 @@
     "deep": ["exprstmt"],
 }
 
+ALPHA_FEATURES = {
```
mschwager, line 101: I don't think this is set in stone yet. I would hold off here until we've finalized the language maturity doc :+1:

colleend: ah yes, good catch

colleend:
per Matt's suggestion, I'm going to close this PR so I can separate it into two:

1. For adding in tests and exceptions
2. For adding in the table-changing code

PR #2151, opened 2020-12-02
Basic support for metavariable matching multiple arguments in JS

### LINKED ISSUE #2150: Allow special metavariables like $...X to match multiple arguments
This is really useful when working with the autofix feature, so the fix line can
refer to list of arguments in the original code.

aryx:
Fixes https://github.com/returntocorp/semgrep/issues/2150
There is probably more work after.

test plan:
see include files
make test


semgrep-core/matching/Generic_vs_generic.ml
```diff
@@ -1085,6 +1085,19 @@ and m_list__m_argument (xsa: A.argument list) (xsb: A.argument list) =
   | [A.Arg (A.Ellipsis _i)], [] ->
       return ()
 
+  | A.Arg (A.Id ((s, tok), _idinfo))::xsa, xb::xsb
+    when MV.is_metavar_ellipsis s ->
+      (* can match 1 or more arguments (or 0 is ok too?) *)
```
nbrahms, line 1090: My intuition is that we could match 0 items, but this might lead to weird matching in edge cases...

I'd expect to be able to match:
```python
unsafe_api_call($ARGS...)
```

to
```python
unsafe_api_call()
```

so I could have an autofix:
```python
safe_api_call($ARGS...)
```

(I'm not sure how we're imagining performing autofix here, do we need to include the delimiter token inside the match?)

semgrep-core/matching/Matching_generic.mli
```diff
@@ -62,6 +62,7 @@ val check_and_add_metavar_binding :
 (*s: signature [[Matching_generic.has_ellipsis_stmts]] *)
 val has_ellipsis_stmts : AST_generic.stmt list -> bool
 (*e: signature [[Matching_generic.has_ellipsis_stmts]] *)
+val inits_and_rest_of_list: 'a list -> ('a list * 'a list) list
```
nbrahms, line 65: The more I think about this, the more I think we need to include the delimiter in the match

nbrahms (approved):
Snazzy 👍

semgrep-core/matching/Generic_vs_generic.ml
aryx: I was just scared that if it matches nothing, then we will get some NoTokenLocation errors.
But yes my intuition is that I think also it should match 0 or more.

PR #2158, opened 2020-12-02
tests for function definitions

colleend:


mschwager (approved):
:tada:

Can we add this to the features in `generate_test_matrix.py`?

colleend:
Yep, currently working on a pull request that separates the entries and the table-generation!

PR #2161, opened 2020-12-03
Add pre-commit hook for re-indenting OCaml code

mjambon:
It is set up such that if `ocp-indent` is missing, no error is reported. This allow pure python users to not install `ocp-indent`, which would be needed when `pre-commit run --all`.

CI will check the indentation of OCaml files.


brendongo (approved):

aryx (approved):

aryx:
Those CI tests take forever now ...

PR #2173, opened 2020-12-03
adding in support for the new tests

colleend:


scripts/generate_test_matrix.py
```diff
@@ -52,14 +52,20 @@
     "exprstmt": "Expression and Statement",
     "typed": "Typed Metavariables",
     "expr_operator": "Deep Expression Operator",
+    "class": "Class Definitions",
```
mschwager: I would keep the function and class feature names similar, i.e. either use `class_def` and `func_def` or `class` and `func`.

mschwager (approved):

scripts/generate_test_matrix.py
colleend: To do this, I'll have to change the names of all the test files -- will push the change in a sec

PR #2174, opened 2020-12-03
Add 'dev-setup' target for install ocp-indent and merlin.

mjambon:
`make dev-setup` should do the right thing. It relies on an opam file that contains only dev dependencies. The `{dev}` constraints are for something else, so I didn't use that, based on info found at https://khady.info/opam-sandbox.html.


aryx (approved):
Thx!
Looks perfect.

PR #2175, opened 2020-12-03
[JS] use $...ARGS instead of $ARGS...

aryx:
This create less ambiguity when parsing old patterns like <...$VAR...>

alt: look for $XXX...> and do an yyback 4 and reutrn a metavar

different take on https://github.com/returntocorp/semgrep/issues/2150

test plan:
make
make test


nbrahms (approved):

brendongo (approved):

aryx:
cc @ajinabraham

aryx:
@ajinabraham this may not be our final design, but at least it will not break nodejsscan for the next release.

PR #2178, opened 2020-12-03
code for generating headers

colleend:
As detailed in https://github.com/returntocorp/semgrep/pull/2145/files

### DIFF
--- a/scripts/generate_test_matrix.py
+++ b/scripts/generate_test_matrix.py
@@ -47,6 +47,7 @@ VERBOSE_SUBCATEGORY_NAME = {
     "var": "Variables",
     "naming_import": "Import Renaming/Aliasing",
     "constant_propagation": "Constant Propagation",
+    "fieldname": "Field Names",
     "syntax": "Single Statements",
     "exprstmt": "Expression and Statement",
     "typed": "Typed Metavariables",
@@ -55,6 +56,7 @@ VERBOSE_SUBCATEGORY_NAME = {
     "anno": "Annotations",
     "func_def": "Function Definitions",
     "key_value": "Object or Dictionary Key Value Pairs",
+    "typed_fieldaccess": "Typed Metavariable Field Access",
 }
 
 LANGUAGE_EXCEPTIONS = {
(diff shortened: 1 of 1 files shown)


scripts/generate_test_matrix.py
```diff
@@ -47,7 +47,6 @@
     "var": "Variables",
     "naming_import": "Import Renaming/Aliasing",
     "constant_propagation": "Constant Propagation",
-    "fieldname": "Field Names",
```
mschwager: Should we keep this and not include it in any of the alpha/beta/GA feature lists?

colleend: I thought about leaving it in, but I think because it's javascript-specific it doesn't really belong in a cheatsheet with more generalized tests. It also leads to key errors in the new code 😓

mschwager: I don't think there's anything inherently Javascript-specific to this feature, only that we've only implemented it for Javascript at the moment. I could see similar implementations for any language that supports `key_value`-like behavior, like Python and Ruby.

If we're getting errors I would lean toward fixing the errors instead of removing things until they go away :+1:

colleend: i guess my question is -- are we planning on supporting it in other languages? If we're unsure, I feel like it's still better to remove it and place it back in later because it might give other people expectations that we'll be supporting this in other languages 🤔 

I'm also not sure why we're keeping in regex_fieldname (which is currently only in js) as compared to something like metavar_typed_fieldaccess, which has tests in java and go 🤔

mschwager: > If we're unsure, I feel like it's still better to remove it and place it back in later because it might give other people expectations that we'll be supporting this in other languages

That's a good point. I had three reasons for including it originally: 1.) we'll continue to run the test and ensure the functionality continues to work, 2.) we'll understand our feature coverage by language for all features, not just common ones, and 3.) we'll include that feature's example code in our JSON output, which helps people trying to understanding the functionality in, e.g. the cloud app editor.

If people are concerned about this feature in other languages we can point them at our shiny new list that doesn't include it in any support tiers :)

> I'm also not sure why we're keeping in regex_fieldname (which is currently only in js) as compared to something like metavar_typed_fieldaccess, which has tests in java and go

Huh, TIL about `metavar_typed_fieldaccess`! Similar to above, I'd vote we include this in `VERBOSE_SUBCATEGORY_NAME` and `CHEATSHEET_ENTRIES` for completeness sake.

colleend: i see, makes sense to me 👍 good point that we'll have the alpha, beta, ga docs out soon. 

Ok, i'll add in metavar typed fieldaccess and fieldname back in 👍

mschwager (approved):
Noice

PR #2184, opened 2020-12-04
Parse explicit interface specifier in C#

Sjord:
E.g.
```csharp
ContentItem IItemContainer.CurrentItem
{
    get { return CurrentItem; }
}
```

The value is not really used anywhere yet.


aryx (approved):

aryx:
it might be useful to collect all those parsing snippet tests somewhere, as a basic regression testsuite.
Maybe in semgrep-core/tests/csharp/parsing/?

Issue #1392, opened 2020-07-29
C# support


mackowski:
I'm opening a ticket to indicate interest in support for the C# Language.
I can help with developing it if you can point me how to start and where to look.

dlukeomalley:
Thank you for filing this @mackowski. Adding @aryx and @nbrahms to discuss what adding support for a new language looks like. We'd love to get more folks involved there.

aryx:
Are you familiar with OCaml? the only remaining part to handle C# needs to be written in OCaml (in semgrep/semgrep-core/parsing/ like the Parse_ruby_tree_sitter.ml in this directory for example, but here it would be a Parse_csharp_tree_sitter.ml. We can generate the boilerplate for this file, but we still need to convert that in the generic AST (defined in pfff/h_program-lang/AST_generic.ml)

mackowski:
I do not know OCaml, but it seems to be fun to learn it. Anyway, as I did not do it before I do not know how much time is needed to do it. I am ready to learn something new and help you with this project, but I also do not want to be a roadblock.
If you can generate the boilerplate in a new branch, I can try to start writing mapping, I will use other languages as examples. If you have capacity and willingness to do this faster do not wait for me, I will review the code to understand it better.

aryx:
Sounds great! @mjambon can you add csharp in ocaml-tree-sitter-lang? I'll create the Parse_csharp_tree_sitter.ml boilerplate with a few cases filled in as a starting point.

mjambon:
@aryx I just added csharp to ocaml-tree-sitter-lang. I parsed one sample program successfully on the first try but didn't run large-scale stats.

aryx:
Ok,  I'll create the Parse_csharp_tree_sitter.ml boilerplate with a few cases filled in as a starting point.

ievans:
@aryx looks like you added the basic parser; did you want to follow up with @mackowski ?

aryx:
No I didn't. I guess I need to add a few simple cases to be a better starting point and maybe some documentation on where to find the relevant AST definitions.

aryx:
The main work @mackowski  is to replace all the todo in https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/parsing/Parse_csharp_tree_sitter.ml with the relevant construct defined in https://github.com/returntocorp/pfff/blob/master/h_program-lang/AST_generic.ml
Not trivial, but not impossible :)

mackowski:
[...]

PR #2185, opened 2020-12-04
Parse anonymous methods in C#

Sjord:
E.g.
```csharp
TestDelegate testDelB = delegate(string s) { Console.WriteLine(s); };
```

Even though we convert the `async` keyword to a keyword list, this currently
doesn't have anywhere to go. A `function_definition` has no attrs; these
normally go in the entity.

Relates to #1392


aryx (approved):
yes, we might need to add the ability to store attributes inside function_definition.

Issue #2196, opened 2020-12-06
spacegrep: Add -i option to spacegrep for case-insensitive searches


mjambon:
This is useful for text search in general. Spacegrep is focused on ascii support, so this option would only handle the downcasing of ascii letters.

aryx:
We can close this no @mjambon ?
You need to use "Fixes #<issue  number>" or "Closes #<issue>" in the commit message to get github to automatically close the issue
once the related PR is merged.  "Implements #<issue>" does not close automatically the issue.

PR #2197, opened 2020-12-06
Add command-line option '-i' for case-sensitive matching in spacegrep.

mjambon:
Implements #2196 

@minusworld don't hesitate to comment if you have suggestions about command-line design, like things you'd expect or find standard or convenient.


aryx (approved):
lgtm

PR #2198, opened 2020-12-06
spacegrep optimizations

mjambon:
* Only peek into the first 4096 (first FS block or so) to determine whether to skip an input file.
* Don't keep all file contents in memory until the end. This saves a lot of memory when matches are only in some files. Still a problem when all the files match.

I tested this just with `spacegrep hello .` on the whole semgrep repo after a build. ~~Getting 1 min 10 now for what used to take 1 min 35.~~

None of this is in the critical path of semgrep usage, since it only applies when scanning whole directories without filtering based on file extensions. It's just for making spacegrep more pleasant when used directly from the command line and in demos.

### DIFF
--- a/spacegrep/src/bin/Spacegrep_main.ml
+++ b/spacegrep/src/bin/Spacegrep_main.ml
@@ -363,5 +363,12 @@ let parse_command_line name =
    Entry point for calling the command 'spacegrep' directly.
 *)
 let main () =
+  (*
+     Make the GC work less aggressively than the default.
+     This wastes memory but makes everything up to twice as fast.
+     TODO: clean benchmarks
+  *)
+  Gc.set { (Gc.get()) with Gc.space_overhead = 500 };
+
   let config = parse_command_line "spacegrep" in
   run config
--- a/spacegrep/src/bin/Spacegrep_main.ml
+++ b/spacegrep/src/bin/Spacegrep_main.ml
@@ -43,9 +43,22 @@ let run_all
     ~case_sensitive ~debug ~force ~output_format ~highlight ~warn
     patterns docs =
   let matches =
-    List.map (fun get_doc_src ->
-      let doc_src = get_doc_src () in
-      let doc_type = File_type.classify doc_src in
+    List.filter_map (fun (get_doc_src : ?max_len:int -> unit -> Src_file.t) ->
+      (*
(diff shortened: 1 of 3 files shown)


aryx (approved):

spacegrep/src/bin/Spacegrep_main.ml
```diff
@@ -43,9 +43,22 @@ let run_all
...
+    List.filter_map (fun (get_doc_src : ?max_len:int -> unit -> Src_file.t) ->
+      (*
+         We inspect the first 4096 bytes to guess whether the file type.
+         This saves time on large files, by reading typically just one
+         block from the file system.
+      *)
+      let peek_length = 4096 in
+      let partial_doc_src = get_doc_src ~max_len:peek_length () in
+      let doc_type = File_type.classify partial_doc_src in
+      let doc_src =
```
aryx: Shoudn't this be put later when you know you're not in the Minified or Binary case? Otherwise
it defeats the purpose of the optimization no? Looks like you gonna read the whole doc in every paths,
whatever the result of the classify on the peek_length read doc.

mjambon: oh, I really rushed this. Thanks for taking a look.

aryx: weird that you got better performance ... it should have been the same amount of time than before.

mjambon: I fixed this, which seems to improves the performance overall and on a single big binary file, but otherwise it's slower on a repo with mostly small source files. I don't understand why. I'll try to get to the bottom of this before merging. I might also postpone this task since it's fun but not urgent.

underyx:
hey @mjambon, heads up, and sorry if this messes with your workflow: I just rebased onto the latest develop and force-pushed your branch to try to debug a semgrep-action bug

mjambon:
ok, so I checked was was slow. It was the parsing step. I wondered if freeing the memory previously occupied by file contents could be what was making things harder for the GC, so I tried to make the GC less aggressive. And it worked, with `space_overhead` = 500.

I haven't tried it, but it might be possible to get even better performance overall by processing each input file in its own process.

Still no automatic benchmark. Will worry about it another day.

PR #2205, opened 2020-12-07
Parse conversion operator overload in C#

Sjord:
E.g.
```csharp
public static implicit operator string(SpaceString s) { ... }
```

Name the methods `op_Explicit` and `op_Implicit`, just like C# does.

Also, add an example C# file that demonstrates this behaviour. This file is not
yet parsed in a test setup.


Sjord:
Could we set up unit-testing easily, that runs `semgrep-core -dump_ast -lang cs tests/csharp/parsing/*.cs`?

aryx:
good idea. Just to check that all the tests files under csharp/parsing/ do not trigger any todo ()?

aryx (approved):

PR #2214, opened 2020-12-09
Use new tree-sitter parsing result type, which includes error stats.

mjambon:
I did the minimum for compatibility with the new result type. We can now do something with the `stat` field; see https://github.com/returntocorp/ocaml-tree-sitter/blob/master/src/run/lib/Parsing_result.mli.

### DIFF
--- a/semgrep-core/parsing/Parse_c_tree_sitter.mli
+++ b/semgrep-core/parsing/Parse_c_tree_sitter.mli
@@ -1,2 +1,2 @@
 
-val parse: Common.filename -> Ast_c.program Parse_tree_sitter_helpers.result
+val parse: Common.filename -> Ast_c.program Tree_sitter_run.Parsing_result.t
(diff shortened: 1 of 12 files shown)


mjambon:
Semgrep doesn't seem to like ocaml type expressions with underscores instead of type variables like `_ list`.
image

mjambon:
I added support for `_` in type expressions in pfff (https://github.com/returntocorp/pfff/pull/351).

For this PR, I'll just use `'a` instead of `_`.

aryx:
That's lots of CI failures.

aryx (changes requested):
fix the CI failure.

mjambon:
Fascinating. `'a` and `_` are not equivalent. A local function definition of type `'a foo` requires `'a` to be bound to the same type everywhere it's used. There is no such restriction if we use `_` instead of `'a`.

I worked around this by making the function not local (since `_` isn't supported by the semgrep check just yet).

mjambon:
ok, CI passed. Merging.

PR #2216, opened 2020-12-09
Track match performance

mjambon:
This is focused on match performance. The duration of each test is tracked separately. Right now it's just two tests:
* control named `empty`
* a test for a pattern that gives takes about 35 seconds to complete on my old laptop. Should be less than this in CI. If/when we fall under a second, we might consider creating a slower test.


semgrep-core/perf/run-perf-suite
```diff
@@ -0,0 +1,86 @@
+#! /usr/bin/env python3
```
aryx, line 1: look at you, writing python scripts now!

aryx (approved):

PR #2218, opened 2020-12-09
Support for partial try, catch and finally

aryx:


underyx:
wow, awesome, you actually did it! :heart:

PR #2222, opened 2020-12-09
Run parsing tests on C#

Sjord:
Just parse the .cs files and check that no errors occur. This tests whether
constructs are implemented in `parsing/Parse_csharp_tree_sitter.ml`.

It's currently pretty generic, inspired by `lang_regression_tests`. However, we currently only have C# to test this way, so maybe it's too generic?


aryx (approved):
Excellent!

PR #2225, opened 2020-12-09
[JS] support for method chaining, e.g. $O.foo() ... .bar()

### LINKED ISSUE #2208: [Java] Add the ability to use ellipsis (or equivalent) for method chaining.

The ellipsis `...` operator does not seem to support usage in method chaining. https://semgrep.dev/docs/writing-rules/pattern-syntax/#ellipsis-operator

For example, we want to ensure a fluent API / builder-like (from an external library) is invoked in the right order. We do not care about any method calls in between. More specifically, we use jOOQ which can leak DB connections if `.stream()`  is invoked without a prior `.fetch`.

In the case below `...` denotes that any number of other method calls may be done before building the query.

```java
// Bad, may leak connections
db.selectFrom()...
  .stream()
  ...

// Better
db.selectFrom()...
  .fetch()
  .stream()
  ...
```

I could imagine writing a rule to check for JS/TS to ensure `Future` invoke/define `.catch(...)` or something like that.

The `...` operator or an equivalent wild-card that supports ignoring interstitial method calls or subsequent method calls. I am thinking we may want something else than `...` or a way to escape it as `db....stream()` ends up being a lot of periods.

Maybe a regex, but that may defeat the purpose of using `semgrep`.

**Additional context**

https://r2c-community.slack.com/archives/C018NJRRCJ0/p1607366986093500

aryx:
Fixes https://github.com/returntocorp/semgrep/issues/2208

test plan:
test files included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - Add entries here
 
+### Added
+
+- Support for '...' in chains of method calls in JS, e.g. `$O.foo() ... .bar()`
+
 ## [0.34.0](https://github.com/returntocorp/semgrep/releases/tag/v0.34.0) - 2020-12-09
 
 ### Added
--- a/semgrep-core/matching/Generic_vs_generic.ml
+++ b/semgrep-core/matching/Generic_vs_generic.ml
@@ -128,6 +128,28 @@ let make_dotted xs =
         B.DotAccess (acc, tok, B.EId e)) base xs
 (*e: function [[Generic_vs_generic.make_dotted]] *)
 
+let rec obj_and_method_calls_of_expr = function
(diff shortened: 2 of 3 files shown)


mjambon (approved):

ievans:
@aryx I love how fast you did this. Quick thoughts:

* are the spaces around `...` required?
* we should add this to the docs for pattern syntax and I think to the language matrix as well?
* since it only works in Javascript to start -- should it be marked as experimental? I'd hate for people to start thinking that it's a bug that we don't have it in other languages when it is actually an enhancement.

aryx:
Yes we could add both $...ARGS and this ... for method chaining in the experimental doc.
And no the space are not required.

aryx:
hmm actually we probably need to revert this for JS because of some weird interaction with ASI in Javascript (Automatic Semicolon Insertion). See https://github.com/returntocorp/semgrep-rules/issues/1055

PR #2226, opened 2020-12-09
Prim expr

colleend:
Basic primary expression support for Kotlin.

Needs Work:
* Lambda functions: assign doesn't work yet. Also need to work on function args and typing.
* When expressions: when_entry needs work. Have yet to do range_test and type_test.
* Return@<id> expressions: should maybe edit AST_generic to have return be of type `tok * expr option * label_ident * sc` instead of `tok * expr option *sc` because currently the simple_identifier <id> is not being stored.

Unsure about:
* conversion of if_expr to if_stmt and then otherexpr. 

Test Plan:
run `_build/default/cli/Main.exe -dump_ast tests/kotlin/prim_expr.kt` to get: 

```
Pr(
  [DefStmt(
     ({name=EId(("foo", ())); attrs=[]; tparams=[];
       info={id_resolved=Ref(None); id_type=Ref(None);
             id_const_literal=Ref(None); };
       },
      FuncDef(
        {fkind=(Function, ()); fparams=[]; frettype=None;
         fbody=Block(
                 [ExprStmt(
                    Call(
                      Id(("foo", ()),
                         id_const_literal=Ref(None); }), []), ());
                  ExprStmt(
                    Lambda(
                      {fkind=(Function, ()); fparams=[]; frettype=None;
                       fbody=ExprStmt(
                               Call(IdSpecial((Op(Plus), ())),
                                 [Arg(
                                    Id(("x", ()),
                                      {id_resolved=Ref(None);
                                       id_type=Ref(None);
                                       id_const_literal=Ref(None); }));
                                  Arg(
                                    Id(("y", ()),
                                      {id_resolved=Ref(None);
                                       id_type=Ref(None);
                                       id_const_literal=Ref(None); }))]), ());
                       }), ());
                  ExprStmt(
                    Container(List,
                      [L(Int(("1", ()))); L(Int(("2", ())));
                       L(Int(("3", ())))]), ());
                  ExprStmt(
                    OtherExpr(OE_StmtExpr,
                      [S(
                         If((),
                           Call(IdSpecial((Op(Lt), ())),
                             [Arg(
                                Id(("a", ()),
                                  {id_resolved=Ref(None); id_type=Ref(
                                   None); id_const_literal=Ref(None); }));
                              Arg(
                                Id(("b", ()),
                                  {id_resolved=Ref(None); id_type=Ref(
                                   None); id_const_literal=Ref(None); }))]),
                           ExprStmt(
                             OtherExpr(OE_StmtExpr, [S(Return((), None, ()))]),
                             ()), None))]), ())]);
         })))])
```


semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
```diff
@@ -355,7 +355,7 @@ let line_string_content (env : env) (x : CST.line_string_content) =
 let return_at (env : env) ((v1, v2) : CST.return_at) =
   let v1 = token env v1 (* "return@" *) in
   let v2 = simple_identifier env v2 in
-  todo env (v1, v2)
+  (v1, Some(v2))
```
mjambon, line 358: Note that you don't need parentheses around `v2`. It's more idiomatic to write:
```ocaml
(v1, Some v2)
```

mjambon (approved):
looks good to me!

aryx:
I think we can use a TodoExpr for those weird return@.
Regarding if_expr, you can probably use the Conditional constructor (it's used for ternary
expressions in C/JS/... as in foo() ? 1 : 0.

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
```diff
@@ -1432,41 +1465,47 @@ and primary_expression (env : env) (x : CST.primary_expression) : expr =
...
                 )
               in
-              todo env (v1, v2, v3, v4)
+              (v1, Some(v4))
          )
        in
-       todo env (v1, v2, v3, v4, v5)
+       let (v6, v7) = v5 in
+       let if_stmt = If (v1, v3, v6, v7) in
+       OtherExpr (OE_StmtExpr, [S if_stmt])
```
aryx, line 1496: maybe you can use Conditional instead, but it accepts 3 expressions.

aryx (approved):
Great progress!

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
colleend: ah, got it -- good catch! I'll fix this in my next push 👍

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
colleend: ah yes -- the 3 expressions part is difficult because v5 returns a statement 😢

colleend:
good idea for todoexpr!! i'll change that in the next push. 

I think Conditional is hard because v5 returns a statement 😱

PR #2236, opened 2020-12-11
Parse finalizer in C#

Sjord:
E.g.
```csharp
~MyClass() { }
```

Also called destructor in other languages.

Relates to #1392

### DIFF
--- a/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
@@ -2191,14 +2191,23 @@ and declaration (env : env) (x : CST.declaration) : stmt =
        let v1 = List.concat_map (attribute_list env) v1 in
        let v2 =
          (match v2 with
-          | Some tok -> token env tok (* "extern" *)
-          | None -> todo env ())
+          | Some tok -> [KeywordAttr (Extern, token env tok)] (* "extern" *)
+          | None -> [])
        in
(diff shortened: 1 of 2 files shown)


semgrep-core/parsing/Parse_csharp_tree_sitter.ml
```diff
@@ -2191,14 +2191,23 @@ and declaration (env : env) (x : CST.declaration) : stmt =
...
-          | None -> todo env ())
+          | Some tok -> [KeywordAttr (Extern, token env tok)] (* "extern" *)
+          | None -> [])
        in
        let v3 = token env v3 (* "~" *) in
        let v4 = identifier env v4 (* identifier *) in
        let v5 = parameter_list env v5 in
        let v6 = function_body env v6 in
-       todo env (v1, v2, v3, v4, v5, v6)
+       let name = ("Finalize", v3) in
```
aryx, line 2201: We could generate a name like ~HelloWorldFinalizer instead of just finalize.
You can also merge multiple tokens into one when it makes sense, her for example v3 and v4 would
be merged. See the use of PI.combine_infos in Parse_go_tree_sitter.ml for example where we have 3 tokens for
a string "foo" (the token for the first '"', then a token for 'foo', then another token for the last '"', but we merge all 3 tokens in a single one and we generate a String("foo",  <combined_token_info>)

aryx (approved):

PR #2245, opened 2020-12-12
Parse event fields in C#

Sjord:
Add a function to create variable definition statements, that also adds
attributes to the entity. Add an `event` NamedAttr to event fields. Maybe this
should actually be a keyword attr?

Related to #1392

### DIFF
--- a/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
@@ -62,6 +62,18 @@ let arg_to_expr (a : argument) =
   | ArgKwd (_, e) -> e (* TODO maybe ArgKwd is also impossible here *)
   | _ -> raise Impossible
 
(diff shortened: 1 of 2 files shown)


semgrep-core/parsing/Parse_csharp_tree_sitter.ml
```diff
@@ -62,6 +62,18 @@ let arg_to_expr (a : argument) =
   | ArgKwd (_, e) -> e (* TODO maybe ArgKwd is also impossible here *)
   | _ -> raise Impossible
 
+let var_def_stmt (decls : (entity * variable_definition) list) (attrs : attribute list) =
+  let stmts = List.map (fun (ent, def) ->
+    let ent = {
```
aryx, line 67: in OCaml you can write let ent = { ent with attrs = ent.attrs @ attrs }
and the other ent field will be copied.

aryx (approved):

PR #2246, opened 2020-12-12
Parse lock in C#

Sjord:
E.g.
`lock (obj) { ... }`

Similar to Java's `synchronized` block.

### DIFF
--- a/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
@@ -1368,7 +1368,7 @@ and statement (env : env) (x : CST.statement) =
        let v3 = expression env v3 in
        let v4 = token env v4 (* ")" *) in
        let v5 = statement env v5 in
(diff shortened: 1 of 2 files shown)


semgrep-core/parsing/Parse_csharp_tree_sitter.ml
```diff
@@ -1368,7 +1368,7 @@ and statement (env : env) (x : CST.statement) =
        let v3 = expression env v3 in
        let v4 = token env v4 (* ")" *) in
        let v5 = statement env v5 in
-       todo env (v1, v2, v3, v4, v5)
+       OtherStmt (OS_Sync, [E v3; S v5])
```
aryx, line 1371: Actually for those kind of stmts that have inside other statements, it's better to use the
OtherStmtWithStmt construct. It's better than OtherStmt because when we build
a control-flow graph for a generic AST (see pfff/lang-GENERIC/analyze/IL_to_CFG.ml),
we need to know those nested stmts,

aryx (approved):

PR #2180, opened 2020-12-04
Fix metavar anno

colleend:
fix the js file that couldn't be parsed

PR #2251, opened 2020-12-14
Add basic error checking to test matrix

mschwager:
Adds the following error checking:

- Add `--strict` and ensure we exit == 0. This will error out if we fail to run the test.
- Ensure we have >= 1 finding in a test, otherwise it's useless. Really it should be findings == `ERROR` count in the test file, but that's an improvement for another day.


scripts/generate_test_matrix.py
```diff
@@ -65,7 +65,7 @@
     "c": ["naming_import", "class_def", "anno"],
     "ruby": ["naming_import", "typed", "anno"],
     "python": ["typed"],
-    "js": ["typed", "anno"],
+    "js": ["typed"],
```
mschwager, line 68: Unrelated improvement per https://github.com/returntocorp/semgrep/pull/2180.

minusworld (approved):

Issue #2054, opened 2020-11-16
semgrep thows exception on python setup.py install


ajinabraham:
**Describe the bug**
```
Processing semgrep-0.31.1-cp36.cp37.cp38.py36.py37.py38-none-macosx_10_14_x86_64.whl
Installing semgrep-0.31.1-cp36.cp37.cp38.py36.py37.py38-none-macosx_10_14_x86_64.whl to /Users/ajinabraham/Code/Mobile-Security-Framework-MobSF/v/lib/python3.7/site-packages
Adding semgrep 0.31.1 to easy-install.pth file
Traceback (most recent call last):
  File "setup.py", line 56, in <module>
    install_requires=requirements,
  File "/Users/ajinabraham/Code/Mobile-Security-Framework-MobSF/v/lib/python3.7/site-packages/setuptools/__init__.py", line 144, in setup
    return distutils.core.setup(**attrs)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/Users/ajinabraham/Code/Mobile-Security-Framework-MobSF/v/lib/python3.7/site-packages/setuptools/command/install.py", line 67, in run
    self.do_egg_install()
  File "/Users/ajinabraham/Code/Mobile-Security-Framework-MobSF/v/lib/python3.7/site-packages/setuptools/command/install.py", line 117, in do_egg_install
    cmd.run(show_deprecation=False)
  File "/Users/ajinabraham/Code/Mobile-Security-Framework-MobSF/v/lib/python3.7/site-packages/setuptools/command/easy_install.py", line 424, in run
    self.easy_install(spec, not self.no_deps)
  File "/Users/ajinabraham/Code/Mobile-Security-Framework-MobSF/v/lib/python3.7/site-packages/setuptools/command/easy_install.py", line 673, in easy_install
    return self.install_item(None, spec, tmpdir, deps, True)
  File "/Users/ajinabraham/Code/Mobile-Security-Framework-MobSF/v/lib/python3.7/site-packages/setuptools/command/easy_install.py", line 720, in install_item
[...]

PR #2253, opened 2020-12-14
Move package information to dunder attributes

mschwager:
Also, general packaging cleanup in anticipation of making some improvements for https://github.com/returntocorp/semgrep/issues/2054.

Using dunder (double underscore) module attributes is a common way for specifying this information: https://github.com/psf/requests/blob/master/requests/__version__.py. The primary advantage here is that other tools/packages and programmatically query this information from the `semgrep` module.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,5 +109,5 @@ gitclean:
 
 .PHONY: bump
 bump:
-	$(SED) 's/__version__ = ".*"/__VERSION__ = "$(SEMGREP_VERSION)"/g' semgrep/semgrep/__init__.py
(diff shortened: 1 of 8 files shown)


mschwager:
Ugh, I give... this is some kind of pip/pipenv hell I'm not trying to get into right now.

mschwager:
- https://stackoverflow.com/questions/49396115/pipenv-warning-invalid-requirement-parse-error-at-extra
- https://github.com/pypa/pipenv/issues/3230
- https://github.com/pypa/pipenv/issues/4142
- https://stackoverflow.com/questions/59552380/pip3-warning-could-not-generate-requirement-for-distribution-ertifi-2019-11#--stacks-s-tooltip-umezhkbn

Issue #2255, opened 2020-12-14
spacegrep: Memoize + get rid of the 10-line limitation in ellipsis


mjambon:
In spacegrep aka generic matching, an ellipsis pattern (`...`) can span at most 10 lines. This restriction has served as a convenient way to avoid polynomial matching complexity O(n^(k+1)) where k is the number of `...` in the pattern. It was also useful to work around the problem of not returning the shortest match, which was fixed by #2108 .

Memoizing intermediate match results should change the complexity from O(n^(k+1)) to O(n*(k+1)) without limiting the span of ellipses. Similar work is in progress on semgrep-core.

aryx:
Can you port semgrep Caching to spacegrep and fix that now?

mjambon:
I'm not sure if we'll end up implementing this, given the complexity of caching. We did implement that for regular semgrep, and the results on real programs were disappointing. Spacegrep is a little different because it has a much simpler AST, and it may benefit from caching more than semgrep's AST. I'm really not sure what to expect.

A starting point could be the [toy matcher implementation](https://github.com/returntocorp/semgrep/tree/4fe3a2feaa054975e2eedef0896a72067fa28500/toy-matcher) residing in the semgrep repo, which was used to make a proof-of-concept before moving on to the implementation in semgrep.

aryx:
@mjambon can we close this once your PR on #5377 is in?

mjambon:
@aryx yes. I'm setting it to autoclose.

PR #2256, opened 2020-12-15
Preliminaries for matching with memoization

mjambon:
This branch adds two main things:
* updated and renamed test suite for matching performance (`semgrep-core/perf`)
* toy sequence matcher

The toy matcher operates on sequences of chars, and matches a pattern over the entirety of the input sequence. See the readme for the motivations and usage.

I added optional memoization of the recursive match function, which gives satisfying results (30x speedup on the `quadratic` test, 600x speedup on the `cubic` test). The difficulty in the implementation is to ensure that the full environment is captured in the cache key and then to use suitable `hash` and `equal` functions. Creating cache keys and looking up cache entries is expensive, and in this implementation it turned out that only caching a random 1/3 of all computations was 1.5x faster than caching them all.

These tests run as part of semgrep-core's `make test`.

Plan for semgrep-core:

* port the kind of memoization we did here to semgrep-core statements.
* leave memoization optional to make performance comparisons easier - at least initially.

The plan for spacegrep is similar. It may be needed even more than for semgrep-core, but spacegrep currently avoids problems by limiting the span of ellipses to 10 lines (task #2255).


semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+(*
+   Accumulator use to store the symbols matched by the current ellipsis.
+
+     None = we're not in an ellipsis
+     Some None = we're in an ellipsis but it has no name and it's not capturing
+
+     Some (Some ("foo", acc)) = acc is the stack of symbols, i.e. the list
+     of symbols captured so far by the current ellipsis in reverse order
+*)
+type ellipsis = (string * char list) option option
```
aryx: maybe better to define a new type for those 3 variants instead of using option option.
type capture_ellipsis = NotInEllipsis | EllipsisButNoCapture | EllipsisCapture of string * char list

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+   - Clarify semgrep-core's matching algorithm by not having to deal with any
+     of the real-world complexity.
+   - Keep the semgrep-core implementation separate and easy to follow,
+     without functors.
+*)
+
+open Printf
+
+type pattern_atom =
+  | Any_symbol         (* match any single symbol *)
```
aryx: Why not Any_char? Why symbol?

aryx: Why not closer to what semgrep does and use:
type pattern =
  | Literal of char
  | Metavar of string
  | Ellipsis

instead of split in 2 with pattern_atom and pattern with those option string.
Also why do you need to capture the content of an Ellipsis?
Is it to imitate the ugly thing I do in semgrep-core to capture the range of the sequence of matched statements?

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+     Some (Some ("foo", acc)) = acc is the stack of symbols, i.e. the list
+     of symbols captured so far by the current ellipsis in reverse order
+*)
+type ellipsis = (string * char list) option option
+
+type stat = {
+  mutable match_calls: int;
+}
+
+let sample_pattern : pattern = [
```
aryx: Then this could be just:
let sampe_pattern = [
  Literal 'A'; Metavar "thing"; Ellipsis; Metavar "thing"]

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+  Symbol 'A', None;
+  Any_symbol, Some "thing";
+  Ellipsis, None;
+  Backref "thing", None;
+]
+
+(*
+   Unlike in semgrep syntax, we distinguish metavariable assignments
+   from metavariable dereferencing.
```
aryx, line 194: why?

aryx: To make things more explicit? What do you win by having Backref and the string option in pattern to
represent those 2 different things?

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+      check_pattern_ new_env pat
+
+let check_pattern pat =
+  check_pattern_ [] pat
+
+(*
+   Initialize the capture of symbols by an ellipsis.
+   'init_acc' is a stack, holding symbols in reverse order.
+*)
+let init_ellipsis opt_name init_acc =
```
aryx: this would be clearer if we had 3 constructors instead of Some Some.

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+(*
+   Extend the environment by adding captured subsequences:
+   - add the named ellipsis that ended just before the current match,
+     if applicable
+   - add the named atom that was just matched, if applicable
+*)
+let extend env opt_ellipsis opt_name captured_sequence =
+  let env =
+    match opt_ellipsis with
+    | None -> env       (* we were not in an ellipsis *)
```
aryx: You would not need those extra comments if None was NotInEllipsis.

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+      printf "match%a%a%a%a\n"
+        print_ellipsis ellipsis
+        print_env env
+        print_pat_head pat
+        print_input_head input
+    else if match_calls = max_trace_lines + 1 then
+      printf "[exceeded max trace lines = %i]\n%!" max_trace_lines
+
+module Cache_key = struct
+  type t = ellipsis * env * pattern * input
```
aryx, line 364: Can you add a comment explaining the intuition behind this caching idea. What do you try to cache
and why it avoids work.

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+*)
+let match_input ?(trace = true) ?(cache = false) root_pat root_input =
+  let stat = { match_calls = 0 } in
+  let get_from_cache = ref (fun _ellipsis _env _pat _input -> assert false) in
+
+  let rec match_ (ellipsis : ellipsis) (env : env) pat input =
+    trace_match_call ~trace stat ellipsis env pat input;
+    !get_from_cache ellipsis env pat input
+
+  and uncached_match ellipsis env pat input =
```
aryx, line 479: would be nice  to have the return type here. What does a matcher return?

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+    fun ellipsis env pat input ->
+      get tbl compute ellipsis env pat input
+end
+
+(*
+   Main matching function.
+
+   Checks if a pattern matches the entire input sequence.
+   Returns the captured symbols or sequences of symbols for which a
+   name was specified in the pattern.
```
aryx, line 463: Maybe a small comment saying that essentially the code below
recurse on the pattern list, and inside this recurse on the input list.

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+                 match_ (init_ellipsis opt_name []) env pat input
+             | Backref name ->
+                 (match Env.find_opt name env with
+                  | Some [] ->
+                      let env = extend env ellipsis opt_name [] in
+                      match_ None env pat input
+                  | _ ->
+                      None
+                 )
+             | _ -> None
```
aryx: Maybe better to be explicit here, given there's not so many cases in pattern_atom:
| Any_symbol  | Symbol _ -> None

semgrep-core/toy-matcher/Matcher.ml
aryx: Apparently either None when no match or a Some with a triplet with X, the environment, and the rest of the input.

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+            | None ->
+                if in_ellipsis then
+                  (* skip input symbol and retry *)
+                  let ellipsis = extend_ellipsis ellipsis symbol in
+                  match_ ellipsis env orig_pat input
+                else
+                  None
+  in
+  if cache then
+    get_from_cache := Memoize.create uncached_match
```
aryx: Would be great to have an example where I can understand how memoizing would improve the performance.
On your sample_pattern, with a simple input like A B C C C C C C C C B C C
why do we avoid redoing the same work thx to memoization?

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+let pseudo_random_string len pick_from =
+  let n = String.length pick_from in
+  assert (n > 0);
+  Random.init 0;
+  String.init len (fun _i ->
+    pick_from.[Random.int n]
+  )
+
+(*
+   This is equivalent in semgrep to searching for a pattern like
```
aryx, line 714: This is great!

semgrep-core/toy-matcher/Matcher.ml
```diff
@@ -0,0 +1,562 @@
...
+   '$A; ... foo;', in a file where the statement 'foo;' doesn't exist.
+
+   With the naive match algorithm, the complexity is O(n^2) where n is the
```
aryx, line 717: can you explain why it's n^2, by unfolding the work done each time.

aryx (changes requested):
This looks good. I would just like a few more comments.

semgrep-core/toy-matcher/Matcher.ml
mjambon: > Why not Any_char? Why symbol?

It's because we don't really care that they are of type `char`. It could be anything.

> Why not closer to what semgrep does and use:
> type pattern =
> | Literal of char
> | Metavar of string
> | Ellipsis
>
> instead of split in 2 with pattern_atom and pattern with those option string.

I picked the list because the list type is well-known. It makes it clear that the pattern is structured as a list rather than any tree.

> Also why do you need to capture the content of an Ellipsis?
> Is it to imitate the ugly thing I do in semgrep-core to capture the range of the sequence of matched statements?

Yes. Like anything that determines the result of the function, it must be made part of the cache key.

semgrep-core/toy-matcher/Matcher.ml
mjambon: I added a bunch of comments to the file. For this specifically, I tried explaining things in comments but concluded it's better to run `make bench` and observe the traces.

semgrep-core/toy-matcher/Matcher.ml
mjambon: This is a lot harder to explain than I thought. See other my comment. Best illustration is a trace.

semgrep-core/toy-matcher/Matcher.ml
mjambon: done

semgrep-core/toy-matcher/Matcher.ml
mjambon: done

semgrep-core/toy-matcher/Matcher.ml
mjambon: I clarified this

semgrep-core/toy-matcher/Matcher.ml
mjambon: I added  something.

semgrep-core/toy-matcher/Matcher.ml
mjambon: I did that.

semgrep-core/toy-matcher/Matcher.ml
mjambon: I explained this in the file.

semgrep-core/toy-matcher/Matcher.ml
mjambon: I think it's clearer this way. I added comments about the differences between semgrep and this.

semgrep-core/toy-matcher/Matcher.ml
mjambon: I did that.

mjambon:
I made various changes, mostly to clarify things.

There's just one thing that's not working, I left todo notes. Essentially, a semgrep pattern like `$A; ... foo($B);` can benefit from caching but not `$A; ... foo($A);` because in the second case, the value captured by `$A` is used in the tail of the pattern. Right now the toy matcher won't benefit even in the first case. This is because `$A` isn't discarded from the environment when constructing the cache keys for input matched against `... foo($B)`. (Caching does work in some of the the tests since we support matching any symbol without capturing its value, which doesn't trigger the name pollution problem)

aryx (approved):
Better. Not sure what's going on with the CI though.

aryx:
Fix the ci so we can merge it.

aryx:
Not sure why you have all those pfff, ocaml-tree-sitter, and ocaml-tree-sitter-lang modifications. Maybe merge develop first?

semgrep-core/parsing/Caching.ml
```diff
@@ -0,0 +1,163 @@
...
+     traversal of statements.
+  *)
+  let stack = ref [] in
+  let add_to_stack f = stack := f :: ! stack in
+
+  let visitor = Visitor_AST.mk_visitor {
+    Visitor_AST.default_visitor with
+    kident = (fun (_k, _) (id, _tok) ->
+      if debug then
+        printf "kident %s\n" id;
```
aryx, line 100: You can use the Logging module in pfff/commons/. It's not too bad. Look for the use of logger in other files.

aryx (approved):
Ok let's merge this because this is getting really big.

mjambon:
I forgot we had a pending pull request for this. Sorry about that. I kept adding changes to it and it wasn't in a mergeable state yesterday. It looks like you found a way around the errors, which is good.

PR #2288, opened 2020-12-18
type outline for kotlin

colleend:
Basic type support for kotlin. 

**Test Plan:**
Run `semgrep-core -dump_ast tests/kotlin/types.kt`. This gives me the AST:
```
Pr(
  [DefStmt(
     ({
       name=EId(("foo", ()),
               id_const_literal=Ref(None); });
       attrs=[]; tparams=[]; },
      FuncDef(
        {fkind=(Function, ()); fparams=[]; frettype=None;
         fbody=Block(
                 [ExprStmt(
                    Call(
                      Id(("x", ()),
                         id_const_literal=Ref(None); }),
                      [Arg(OtherExpr(OE_StmtExpr, [S(Return((), None, ()))]));
                       Arg(IdSpecial((This, ())))]), ())]);
         })))])

``` 

__Todo:__
* add type modifier information to type_ function (after annotations and everything else are done)

_Unsure about:_
* TyTuple return for user_type -- is this ok?
* function_type is not using v1 currently. Will need to think about how to add this in.

### DIFF
--- a/semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
@@ -467,7 +467,7 @@ and anon_choice_param_b77c1d8 (env : env) (x : CST.anon_choice_param_b77c1d8) =
    | `Param x -> parameter env x
    | `Type x ->
        let v1 =  type_ env x in
-       ParamPattern (PatType v1)
+       ParamClassic (param_of_type v1)
   )
 
 and assignment (env : env) (x : CST.assignment) =
@@ -1347,22 +1347,14 @@ and nullable_type (env : env) ((v1, v2) : CST.nullable_type) =
   let v2 = List.map (token env) (* "?" *) v2 in
   (match v2 with
    | hd::tl -> TyQuestion (v1, hd)
-   | _ ->
-       let q = Parse_info.fake_info "?" in
-       TyQuestion (v1, q)
+   | [] -> raise Impossible (* see repeat1($._quest) in grammar.js *)
   )
 
 and parameter (env : env) ((v1, v2, v3) : CST.parameter) : parameter =
   let v1 = simple_identifier env v1 in
   let v2 = token env v2 (* ":" *) in
   let v3 = type_ env v3 in
-  let param = {
-    pname = Some v1;
-    ptype = Some v3;
-    pdefault = None;
-    pattrs = [];
-    pinfo = empty_id_info();
-  } in
+  let param = { (param_of_id v1) with ptype = Some v3 } in
   ParamClassic param
 
 and parameter_modifiers (env : env) (x : CST.parameter_modifiers) =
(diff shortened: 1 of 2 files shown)


aryx:
The test plan is not great because we actually don't see any type in the AST. fparams is currently empty.

aryx:
Also I don't understand why the body of the function in the AST is this complicated Call (x, return this)
when in the code you have just return x;
Are you sure you put the dump of the right file?

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
```diff
@@ -465,8 +465,9 @@ and annotation (env : env) (x : CST.annotation) =
 and anon_choice_param_b77c1d8 (env : env) (x : CST.anon_choice_param_b77c1d8) =
   (match x with
    | `Param x -> parameter env x
-   | `Type x -> let _ =  type_ env x in
-       raise Todo
+   | `Type x ->
+       let v1 =  type_ env x in
+       ParamPattern (PatType v1)
```
aryx: I tyink you can build a ParamClassic with p_type = v1, and p_name = None,
like we do in C for function signatures.

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
```diff
@@ -1344,13 +1345,25 @@ and nullable_type (env : env) ((v1, v2) : CST.nullable_type) =
     )
   in
   let v2 = List.map (token env) (* "?" *) v2 in
-  todo env (v1, v2)
+  (match v2 with
+   | hd::tl -> TyQuestion (v1, hd)
+   | _ ->
```
aryx: in grammar.js they use repeat1($._quest)
and _quest: $ => "?"
which means there is at least one "?", so you can write the pattern:
| [] -> raise Impossible (* see repeat1($._quest) in grammar.js *)

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
aryx: There is even AST_generic.param_of_type() to help build this parameter_classic type.

semgrep-core/parsing/Parse_kotlin_tree_sitter.ml
```diff
@@ -1344,13 +1345,25 @@ and nullable_type (env : env) ((v1, v2) : CST.nullable_type) =
...
+       let q = Parse_info.fake_info "?" in
+       TyQuestion (v1, q)
+  )
 
 and parameter (env : env) ((v1, v2, v3) : CST.parameter) : parameter =
   let v1 = simple_identifier env v1 in
   let v2 = token env v2 (* ":" *) in
   let v3 = type_ env v3 in
-  todo env (v1, v2, v3)
+  let param = {
```
aryx: You can factorize a bit and do let param = { (AST_generic.param_of_id v1) with ptype = v3 }

aryx (changes requested):

colleend:
🤦‍♀️ !! thanks for noticing!! fparams is empty because I initially set the fparams to empty list 🤦 🤦 

I'll make the changes to the current code and think of a better test plan. Ty for the comments 🎉 

I'll probably need to make some more changes to Function or assign type to get some good tests going (I'll do this after break). In the meantime, I'll just push the changes to your comments 👍 👍

aryx (approved):

PR #2308, opened 2020-12-23
Consistently use `#! /usr/bin/env bash' as shebang

iamleot:
This increase portability of semgrep on platforms not having bash as
/bin/bash.

This is not just a possible theoretical problem but actually helps helping build
semgrep on NetBSD!

(And, FTR, I was finally able to build semgrep and can use it on NetBSD with mostly just this patch and a couple of notes that I should tidy up and share (but probably only relevant for NetBSD and pkgsrc users! :))

### DIFF
--- a/scripts/build-wheels.sh
+++ b/scripts/build-wheels.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
 set -e
 pip install setuptools wheel
 cd semgrep && python setup.py sdist bdist_wheel
--- a/scripts/osx-release.sh
+++ b/scripts/osx-release.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
 set -e
 brew install opam pkg-config coreutils
 opam init --no-setup --bare;
--- a/scripts/ubuntu-generic.sh
+++ b/scripts/ubuntu-generic.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
 set -e
 version="${VERSION:?Set a version to install}"
 tarball=semgrep-v$version-ubuntu-16.04.tgz
--- a/scripts/ubuntu-release.sh
+++ b/scripts/ubuntu-release.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
 #
 # Make a semgrep release for Linux x86_64.
 #
--- a/scripts/validate-docker-release.sh
+++ b/scripts/validate-docker-release.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#! /usr/bin/env bash -e
 version="${GITHUB_REF/refs\/tags\//}"
 version="${version:-$VERSION}"
 version="${version:?Version must be set}"
--- a/scripts/validate-osx-release.sh
+++ b/scripts/validate-osx-release.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
 set -e
 
 echo "Loading version from file: $(cat version)"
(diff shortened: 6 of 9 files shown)


ievans (approved):
this looks great @iamleot -- thanks for the contribution! @cla-bot check

tagging @DrewDennison or @brendongo for packaging/devops review

@mjambon or @minusworld  -- could we write a semgrep generic-mode rule for #!/bin/bash -> #! /usr/bin/env bash? 

My attempt: https://semgrep.dev/s/vzxn

iamleot:
Thanks @ievans!

ievans:
@cla-bot check pls (seems flaky)

iamleot:
> @mjambon or @minusworld -- could we write a semgrep generic-mode rule for #!/bin/bash -> #! /usr/bin/env bash?
> 
> My attempt: https://semgrep.dev/s/vzxn

Really a minor cosmetic nit: in my experience it is more common to have `#!/usr/bin/env bash` instead of `#! /usr/bin/env bash` (with a white space between the `#!` and `/usr/bin/env`). In this patch I have added a whitespace in order to be consistent with existent shebangs in semgrep codebase.

In [pkgsrc](https://www.pkgsrc.org/) package system we have the following [sed(1)](https://man.NetBSD.org/sed.1) invocation to spot shebangs (via [pkgsrc/mk/check/check-interpreter.mk:67](https://github.com/NetBSD/pkgsrc/blob/f54a860ccbb6dad4a216275b4819768f09caf373/mk/check/check-interpreter.mk#L67)):

```sh
sed -n -e '1s/^#![[:space:]]*\([^[:space:]]*\).*/\1/p' -e '1q'
```

(...and then the interpreter in `\1` is double-checked to be existent (so possible packages installing scripts but not having the corresponding interpreter available as dependency are catched), but that is not useful in the semgrep context! So we can ignore it! :))

While checking possible `[[:space:]]*` between the `#!` and the `/bin/bash` would be nice I think that checking only the first line is even more important in this case to avoid possible false positives in the middle of text files.
I have tried to adjust the @ievans' rule trying to address that by trying:

 `=~/#!\/bin\/bash/`

(both unquoted and quoted via a single quote (`'...'`) but it seems to not match anything.)
Regarding the other possible problem: only catching the first line of the file, it seems that I wasn't able to find any possible way to express that by skimming the docs.

Anyway, these are mostly my side-notes on thinking aloud about the rule proposed by @ievans! :) (i.e. not strictly related to this PR :))

iamleot:
> While checking possible `[[:space:]]*` between the `#!` and the `/bin/bash` would be nice I think that checking only the first line is even more important in this case to avoid possible false positives in the middle of text files.
> I have tried to adjust the @ievans' rule trying to address that by trying:
> 
> `=~/#!\/bin\/bash/`
> 
> (both unquoted and quoted via a single quote (`'...'`) but it seems to not match anything.)
> Regarding the other possible problem: only catching the first line of the file, it seems that I wasn't able to find any possible way to express that by skimming the docs.

Of course I have not skimmed enough! (`pattern-regex` can solve the first problem and it seems by accident using `^` in the regex also address the second problem (are multiple lines concatenated together in the generic mode?)):

https://semgrep.dev/s/gLle

underyx (approved):
Agreed, this is good practice. We should probably write a semgrep rule for it, too! Thanks for the patch.

Edit: oh, I clicked directly to the review screen and failed to notice the discussion above. I see I'm not the only one with these thoughts :D

underyx:
This one works only with the standard no-space rule, but adds a rule ID, message, and matching any binary, not just `bash`. https://semgrep.dev/s/underyx:hardcoded-shebang-path

aryx (approved):

iamleot:
Thanks for review/discussion/merge @ievans, @underyx and @aryx!

Happy holidays!

PR #2319, opened 2020-12-31
Use Arrow as function type in lambda with => in C#

Sjord:


aryx (approved):
This is actually currently not used in Semgrep, but maybe one day.

Issue #361, opened 2020-04-01
"Match everything" inside javascript arrays


inkz:
Not able to create a pattern that would match any array or an array with one specific item

https://sgrep.live/0PB
https://sgrep.live/K9L
examples:
```yaml
var $X = [...];
```
```javascript
var x = [3,2,1]
```
and 
```yaml
var $X = [...,1];
```
```javascript
var x = [3,2,1]
```

PR #2322, opened 2020-12-31
Try use latest pfff with the new constant propagation dataflow-based analysis

aryx:
Test plan:
lots of test regression. We need to figure this out.

### DIFF
--- a/semgrep-core/analyzing/Datalog_experiment.ml
+++ b/semgrep-core/analyzing/Datalog_experiment.ml
@@ -112,7 +112,7 @@ let instr env x =
   match x.i with
   | Assign (lval, e) ->
       (match lval, e.e with
-       | {base = Var n; offset = NoOffset},
+       | {base = Var n; offset = NoOffset; constness = _},
          Literal (AST.Int (s)) ->
            let v = var_of_name env n in
            let h = heap_of_int env s in
--- a/semgrep-core/matching/Generic_vs_generic.ml
+++ b/semgrep-core/matching/Generic_vs_generic.ml
@@ -369,8 +369,8 @@ and m_ident_and_empty_id_info a1 b1 =
 *)
 and m_id_info a b =
   match a, b with
-    { A. id_resolved = _a1; id_type = _a2; id_const_literal = _a3 },
-    { B. id_resolved = _b1; id_type = _b2; id_const_literal = _b3 }
+    { A. id_resolved = _a1; id_type = _a2; id_constness = _a3 },
+    { B. id_resolved = _b1; id_type = _b2; id_constness = _b3 }
     ->
       (* old: (m_ref m_resolved_name) a3 b3  >>= (fun () ->
        * but doing import flask in a source file means every reference
--- a/semgrep-core/matching/Normalize_generic.ml
+++ b/semgrep-core/matching/Normalize_generic.ml
@@ -81,9 +81,13 @@ let normalize_import_opt is_pattern i =
 let rec eval x =
   match x with
   | L x -> Some x
-  | Id (_, { id_const_literal = {contents = Some x}; _})
-  | DotAccess (IdSpecial (This, _), _, EId (_, {id_const_literal = {contents = Some x}; _})) ->
-      Some x
+  | Id (_, { id_constness = {contents = Some x}; _})
+  | DotAccess (IdSpecial (This, _), _, EId (_, {id_constness = {contents = Some x}; _})) ->
+      (match x with
+       | Lit x -> Some x
+       (* todo: do something better? *)
+       | Cst _ | NotCst -> None
+      )
 
   | Call(IdSpecial((Op(Plus | Concat) | ConcatString _), _), args)->
       let literals = args |> unbracket |> Common.map_filter (fun (arg) ->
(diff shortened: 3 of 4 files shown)


aryx:
I've tried to integrate the latest pfff with your dataflow const analysis and fixed a few compilation errors in semgrep,
but I get lots of test regressions.

IagoAbal:
I will take a look ASAP, should have some time tomorrow. I was also preparing another PR to Pfff with a small fix and some clean ups.

IagoAbal:
This may be partly due to the AST-to-IL translation, some issues are independent of PR #361. I was able to reproduce some of the problems on 76580e4c2c1b91f177b612a94bdf3097602f5be8 (current HEAD of develop) by changing `semgrep-core/pfff/analyze/Constant_propagation.ml` in the following way:

```
--- a/lang_GENERIC/analyze/Constant_propagation.ml
+++ b/lang_GENERIC/analyze/Constant_propagation.ml
@@ -336,4 +336,13 @@ let propagate2 lang prog =
   ()
 
 let propagate a b = Common.profile_code "Constant_propagation.xxx" (fun () ->
-  propagate2 a b)
+  propagate2 a b;
+  let module V = Visitor_AST in
+  let v = V.mk_visitor { V.default_visitor with
+                         V.kfunction_definition = (fun (_k, _) def ->
+                           let xs = AST_to_IL.stmt def.fbody in
+                           ignore xs
+                         );
+                       } in
+  v (Pr b)
+)
```
I'm trying to figure out why is this...

IagoAbal:
One of the failures, tests/js/deep_expr_and_ellipsis.js, can be reproduced with pfff -il_generic. That example, and in particular the `super(props);` statement, triggers the second `raise Impossible` in `AST_to_IL.call_special`:

```ocaml
and call_special _env (x, tok) =
  (match x with
   | G.Op _ | G.IncrDecr _ ->
       raise Impossible (* should be intercepted before *)
   | G.This | G.Super | G.Self | G.Parent ->
       raise Impossible (* should be intercepted before *)
```

IagoAbal:
The other error is due to tests/lua/parsing/varargs.lua:
```lua
local function test(a, b, ...)
   print(a, b)
   print(...)
   return {...}
end
```
which triggers:
```ocaml
  | G.Ellipsis _ | G.TypedMetavar (_, _, _)| G.DisjExpr (_, _)
  | G.DeepEllipsis _ | G.DotAccessEllipsis _
    -> sgrep_construct (G.E eorig)
```
Is e.g. `print(...)` valid Lua code? If not, maybe we could fix that test example?

IagoAbal:
BTW, we may prefer not to show warnings from AST-to-IL translation in normal mode (just in verbose mode). The test files are incomplete pieces of code so I guess it's normal that we get so many unresolved-id warnings.

All this makes me think that, if we add the new flow-senstive constant propagation based on IL, we will be running AST-to-IL all the time, so to be safe perhaps we could remove all the fatal errors from AST-to-IL and replace them with warnings. At least until AST_to_IL is fairly complete and tested.

IagoAbal:
This `-lang go -f tests/go/dots_stmts3.sgrep tests/go/dots_stmts3.go` no longer works after enabling flow-sensitive constant propagation.

tests/go/dots_stmts3.sgrep:
```go
var $SOME = "..."
...
var $OTHER = $X
...
var $LAST = $SOME + $OTHER
```
tests/go/dots_stmts3.go:
```go
package Foo

func foo() {
        //ERROR:
        var x = "foobar"
        //var y = "bla";
        if true {
                x = "bar"
        }
        var z = 3
        if (false) { 
                // ignore me
        }
        var d1 = x + z;
}
```

As far as I can tell, the issue here is that Semgrep's matching environment takes id_info into consideration, and once the analysis is flow-sensitive, different occurrences of a variable may have different _constness_ info attached. In particular, I think `$SOME` is first matched against the  first occurrence of `x` with constness `Some (Lit (String ("foobar", ())))`, then when we try to match `$SOME + $OTHER` against `x + z`, matching `$SOME` against the last occurrence of `x` fails because this last occurrence of `x` has constness `Some (Cst Cstr)`.

See in `Matching_generic`:
```ocaml
let rec equal_ast_binded_code (a: MV.mvalue) (b: MV.mvalue) : bool = (
  let res = (match a, b with
    | MV.Id _, MV.Id _
    | MV.E _, MV.E _
    | MV.S _, MV.S _
    | MV.P _, MV.P _
    | MV.T _, MV.T _
    | MV.Args _, MV.Args _
      ->
        (* Note that because we want to retain the position information
         * of the matched code in the environment (e.g. for the -pvar
         * sgrep command line argument), we can not just use the
         * generic '=' OCaml operator as 'a' and 'b' may represent
         * the same code but they will contain leaves in their AST
         * with different position information. So before doing
         * the comparison we just need to remove/abstract-away
         * the line number information in each ASTs.
        *)
        let a = MV.abstract_position_info_mval a in
        let b = MV.abstract_position_info_mval b in
        a =*= b
```

See also in `Metavars_generic`:
```ocaml
let mvalue_to_any = function
  | E e -> G.E e
  | S s -> G.S s
  (* bugfix: do not return G.I id. We need the id_info because
   * it can be used to check if two metavars are equal and have the same
   * sid (single unique id).
  *)
  | Id (id, Some idinfo) -> G.E (G.Id (id, idinfo))
```

I guess the fix is to simply abstract constness info too before comparing two mvalues.

aryx:
Ok for now I'll remove the call to Propagate_dataflow so we can still use the latest modifs to pfff
(lua) and semgrep.
We can fix the bugs you mention later.

PR #2325, opened 2021-01-02
Include LICENSE in source distributions

bollwyvl:
Thanks for this remarkable tool!

This just ensures the LICENSE is included in source distributions.

Motivation: looking into packaging this for [conda-forge](https://github.com/conda-forge/staged-recipes/issues/12965), which likes to ensure all license files are included in distributions (even if the license doesn't _demand_ it).

Thanks again!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE


bollwyvl:
hah, probably won't take the time to CLA-NDA-BBQ or whatever for a one-liner config change :blush:

bollwyvl:
And I even put it in the wrong spot... I'll just go ahead and close this now! Will report back if i make progress on packaging it!

PR #2326, opened 2021-01-04
Use package_data for package binaries instead of scripts

### LINKED ISSUE #2054: semgrep thows exception on python setup.py install

mschwager:
Fixes https://github.com/returntocorp/semgrep/issues/2054.

There will undoubtedly be build failures I'll have to fix, I'm just putting this up to get early eyes on if I'm missing anything obvious, or obvious reasons why this won't work. I won't get into _all_ the details as to why I chose `package_data` for the binaries instead of `scripts` - comments in #2054 cover that.

**TL;DR is [`package_data`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#package-data) is a much cleaner solution for including arbitrary data files (including binaries), and allows us to easily reference the resultant installed file via `pkg_resources`.** We will no longer have to muck around with system binaries/scripts, as [`scripts`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#scripts)/`install_scripts` does, and we can instead directly reference files inside the installed package.

Potential buggy areas:

* There is now more opportunity for a `semgrep` : `semgrep-core` version mismatch. E.g. `semgrep` is somehow missing the `package_data`, falls back on an old `semgrep-core` version, and the version mismatch causes runtime issues. I believe this issue is highly unlikely to occur in practice, and existed before but wasn't explicitly defined. I think a better solution in the long-term is to avoid looking for system-level binaries and only reference `package_data` unless an env variable pointing to `semgrep-core` is explicitly defined. I believe we can live with this deficiency for now, and move toward a cleaner, easier to extend/improve/understand solution right now.

I tested this out with a local `.whl` file install, and a mocked PyPI install via https://github.com/steiza/simplepypi - both worked as expected. I also expect this to solve @blshkv's problem in https://github.com/returntocorp/semgrep/issues/2054.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -95,6 +95,8 @@ jobs:
       - name: Install Tox and any other packages
         run: pip install pipenv==2018.11.26 wheel==0.34.2 tox==3.15.0
       - name: Run Tox
+        env:
+          SEMGREP_SKIP_BIN: true
         run: |
           cd semgrep
           tox -e py  # Run tox using the version of Python in `PATH`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   the type of a field
 - Constant propagation for Java final fields (using this.field syntax)
 
+### Changed
+
+- Packaging and `setup.py` functionality (`.whl` and `pip` install unchanged):
+  `SEMGREP_SKIP_BIN`, `SEMGREP_CORE_BIN`, and `SPACEGREP_BIN` now available
+
 ### Fixed
 - correctly match the same metavariable for a field when used at a definition
   site and use site for Java
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,7 +58,7 @@ COPY --from=build-semgrep-core \
 RUN ln -sf spacegrep /usr/local/bin/spacecat
 
 COPY semgrep /semgrep
-RUN HOMEBREW_SYSTEM='NOCORE' python -m pip install /semgrep
+RUN SEMGREP_SKIP_BIN=true python -m pip install /semgrep
 RUN semgrep --version
 
 RUN mkdir -p /src
--- a/semgrep/semgrep/bin/__init__.py
+++ b/semgrep/semgrep/bin/__init__.py
@@ -1,0 +1,1 @@
+# Binaries will be installed in this module directory and referenced via 'pkg_resources'
--- a/semgrep/semgrep/util.py
+++ b/semgrep/semgrep/util.py
@@ -14,6 +14,7 @@ from typing import Tuple
 from typing import TypeVar
 from urllib.parse import urlparse
 
+import pkg_resources
 from colorama import Fore
 from tqdm import tqdm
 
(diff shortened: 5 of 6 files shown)


mschwager:
Note this will also require brew formula changes.

.github/workflows/tests.yml
```diff
@@ -95,6 +95,8 @@ jobs:
       - name: Install Tox and any other packages
         run: pip install pipenv==2018.11.26 wheel==0.34.2 tox==3.15.0
       - name: Run Tox
+        env:
+          SEMGREP_SKIP_BIN: true
```
mschwager, line 99: This replaces the `if "TOX_ENV_NAME" in os.environ` check before.

brendongo:
Doesn't look like you're missing anything. FYI HOMEBREW_SYSTEM is set automatically when doing a `brew install` but no reason we can't just set the new environment var in the brew formula.

brendongo (approved):

mschwager:
> FYI HOMEBREW_SYSTEM is set automatically when doing a `brew install` but no reason we can't just set the new environment var in the brew formula.

That's what I figured. I thought it'd be better to have a single, explicit env variable we provide rather than try to enumerate all the environments we _don't_ want to build the binaries in our setup script.

Issue #1931, opened 2020-10-30
Support for Lua language


mwild1:
**Is your feature request related to a problem? Please describe.**

Add support for the [Lua language](https://lua.org/).

**Additional context**

Lua is used in a wide variety of contexts including many video games, business logic software, and within notable open-source software such as Apache, MediaWiki, Redis, VLC and Wireshark.

There is currently one popular static analysis tool for Lua, [luacheck](https://github.com/luarocks/luacheck). However it is not as powerful or extensible as semgrep.

As a relatively small language with a relatively [simple syntax](https://www.lua.org/manual/5.4/manual.html#9), hopefully it wouldn't be tremendously difficult to support!

aryx:
There is actually already a tree-sitter grammar for LUA: https://github.com/Azganoth/tree-sitter-lua
so integrating LUA should not be too difficult.
We already have a few people adding support for C# and kotlin by also usint a tree-sitter-based parser.

aryx:
If you're willing to learn a bit OCaml, you could try to add yourself support for LUA!
See the work of @Sjord for C# support here: https://github.com/returntocorp/semgrep/issues/1392

Ruin0x11:
I did a lot of work on this:

https://github.com/Ruin0x11/semgrep/tree/lua
https://github.com/Ruin0x11/pfff/tree/lua
https://github.com/Ruin0x11/semgrep-grammars/tree/lua
https://github.com/Ruin0x11/ocaml-tree-sitter/tree/lua
https://github.com/Ruin0x11/ocaml-tree-sitter-lang/tree/lua
https://github.com/Ruin0x11/tree-sitter-lua/tree/fix-call-expression

At this stage it can parse Lua source code to the generic AST format.

Blocked by #2280 and https://github.com/Azganoth/tree-sitter-lua/pull/7. There are also still some other parse errors caused by `tree-sitter-lua`.

aryx:
@DrewDennison we should add Lua and C# on https://dashboard.semgrep.dev/languages

DrewDennison:
will do. Note that there are some metrics already https://dashboard.semgrep.dev/metric/semgrep.core.c-sharp.parse.pct https://dashboard.semgrep.dev/metric/semgrep.core.tests.lua.unit-test.num

aryx:
Ok we're getting really close to get Lua support in semgrep!

aryx:
Hopefully we can release in the next version of semgrep next week (0.37)

[...]

PR #2327, opened 2021-01-04
Enable back warnings for Parse_lua_tree_sitter.ml

aryx:
test plan:
make


aryx:
This will help https://github.com/returntocorp/semgrep/issues/1931
cc @Ruin0x11

aryx:
The parsing stats (via semgrep-core -lang lua -parsing_stats lua/testes/) goes from 12% to 98%.
Hopefully we get good stats on all the corpus @Ruin0x11 included in the projects.txt file.

PR #2328, opened 2021-01-05
Use deriving eq to automatically generate equal_xxx functions for generic AST

aryx:
The use of Lib_AST.abstract_info_xxx is a bit ugly and potentially also
slower than using deriving eq with some well placed adhoc equal functions
(e.g., equal_tok to abstract away information position).
This also abstract away the constness that we need when we will plug
Iago's dataflow-based constness analysis.

test plan:
make test


aryx:
cc @IagoAbal

PR #2329, opened 2021-01-05
Parse async/await in C#

Sjord:
Even though `async` is mapped to a keyword attribute, sometimes this attribute
is ignored. For example, lambda functions currently don't have a place to put
attributes, since they have no entity.

Related to #1392


aryx (approved):
Yes I think you're right, we need to add a fattr: attribute; field in function_def, so lambdas
can have attributes. I had a similar problem when handling javascript.

PR #2335, opened 2021-01-06
Enable flow-sensitive constant propagation

IagoAbal:


IagoAbal:
Requires returntocorp/pfff#369.

aryx:
Note that if this requires a fix to pfff not yet merged, you can still play with the submodule and reference it from this diff. That way the CI will run with the pfff you require.
Try cd semgrep-core/pfff; git pull; git checkout origin/<branch>
and add a commit with this specific pfff.
That way we can know if the CI would work with this pfff change.

aryx (changes requested):
reference the right version of pfff needed for this PR, so we can see whether the CI would then work.

IagoAbal:
> reference the right version of pfff needed for this PR, so we can see whether the CI would then work.

I can't push a branch to Semgrep's repository it seems, so I added my own repo as a remote and checked out my branch. Not sure what effect that had, but checks are not running at all now it seems.

IagoAbal:
OK let's see now. I would expect all the timeout errors to go away. But I think there was another weird Ruby parsing error.

IagoAbal:
PR returntocorp/pfff#373 removes the parsing error on some Ruby example containing divisions by zero. One failing test left!

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "pfff"]
 	path = semgrep-core/pfff
-	url = https://github.com/returntocorp/pfff.git
+	url = https://github.com/IagoAbal/pfff.git
```
aryx: you need to undo that. I've merged the div_by_zero so you should be able to reference the original pfff repo.
Note that Ruin0x11 is an extenral contributor and he managed to modify semgrep and pfff without 
referencing his own fork of pfff. Once you make a PR to the pfff repo, you can pull it and reference it from your semgrep repo (git pull origin; git checkout <branchname> where branchname is exactly the same name than the branch pushed to the main pfff repo).
If's some git gymnastic. Let me know next time you need that and I can help.

semgrep-core/matching/Generic_vs_generic.ml
```diff
@@ -1196,16 +1206,10 @@ and m_arguments_concat a b =
...
-       with
-       | Some _ ->
-           (* can match nothing *)
-           (m_arguments_concat xsa xsb) >||>
-           (* can match more *)
-           (m_arguments_concat ((A.Arg (A.L (A.String("...", a))))::xsa) xsb)
-       | None ->
-           (m_arguments_concat xsa (B.Arg(bexpr)::xsb))
-      )
+      (* can match nothing *)
```
aryx, line 1209: could you put in comment the call to Normalize_generic.consta... 
and explain why we don't need it anymore now that we have the right constant_propagation?

aryx (changes requested):
just the submodule thing and we should be able to merge this if semgrep-rules have been modified correctly.

aryx:
Ok just one semgrep-rules failure remaining! Woohoo, we're getting really close.

.gitmodules
IagoAbal: Aha, it seems possible, although it's a bit more complicated than that: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally

IagoAbal: Thanks for the pointer, I'll do like that next time!

semgrep-core/matching/Generic_vs_generic.ml
IagoAbal: I don't fully understand why it was needed before. However I found the original commit: e775c1df4f036783f0cbbb1097ad0641578c4358.  Maybe it's a good idea that @ulziibay reviews the changes to `Generic_vs_generic.m_arguments_concat` and shed some light on this issue?

aryx: ulzii is not anymore in the company.
I'll try to merge it today and fix the last issues.

aryx (commented):
merged develop and trying here: https://github.com/returntocorp/semgrep/pull/2386

aryx:
merged via #2386

PR #2336, opened 2021-01-06
Parse C# semgrep patterns

aryx:

PR #2338, opened 2021-01-06
Add stubs to support semgrep pattern constructs in C#

mjambon:
This and more was done in parallel by @aryx while I was sleeping in https://github.com/returntocorp/semgrep/pull/2336.
We should just close this PR unless we see something special.


aryx:
You can drop it i think

Issue #2237, opened 2020-12-11
Custom policy locally with Docker results in unknown exception: End_of_file


zythosec:
**Describe the bug**
Running a custom policy consisting of PHP/Javascript rules locally with the latest Docker image on https://github.com/rapid7/hackazon (vulnerable web application) results in an internal error:

docker run --rm -v "${PWD}:/src" returntocorp/semgrep:0.34.0 --config hackazon.yaml
running 16 rules...
an internal error occured while invoking semgrep-core:
	unknown exception: End_of_file
An error occurred while invoking the semgrep engine; please help us fix this by creating an issue at https://github.com/returntocorp/semgrep

**To Reproduce**
```
git clone https://github.com/rapid7/hackazon
docker run --rm -v "${PWD}:/src" returntocorp/semgrep:0.34.0 --config hackazon.yaml
```
hackazon.yml
```yml
rules:
- id: php.lang.security.weak-crypto.weak-crypto
  patterns:
  - pattern: $FUNC(...);
  - metavariable-regex:
      metavariable: $FUNC
      regex: crypt|md5|md5_file|sha1|sha1_file|str_rot13
  message: |
    Detected usage of weak crypto function. Consider using stronger alternatives.
  metadata:
    references:
    - https://www.php.net/manual/en/book.sodium.php
    - https://github.com/FloeDesignTechnologies/phpcs-security-audit/blob/master/Security/Sniffs/BadFunctions/CryptoFunctionsSniff.php
    dev.semgrep.actions:
    - block
    - notify
    semgrep.url: https://semgrep.dev/r/php.lang.security.weak-crypto.weak-crypto
  languages: [php]
  severity: ERROR
- id: php.lang.security.preg-replace-eval.preg-replace-eval
  patterns:
  - pattern: preg_replace(...);
  - pattern-not: preg_replace("...", ...);
  message: |
    Calling preg_replace with user input in the pattern can lead to arbitrary
    code execution. The eval modifier (`/e`) evaluates the replacement argument
    as code.
  metadata:
    references:
    - https://www.php.net/manual/en/function.preg-replace.php
    - https://www.php.net/manual/en/reference.pcre.pattern.modifiers.php
    - https://github.com/FloeDesignTechnologies/phpcs-security-audit/blob/master/Security/Sniffs/BadFunctions/PregReplaceSniff.php
    dev.semgrep.actions:
    - block
    - notify
[...]

PR #2350, opened 2021-01-07
Augment unknown semgrep-core unknown error message with the rule ID

minusworld:
Augment unknown semgrep-core error message with the rule ID it's trying to run. Added a unit test to make sure the rule ID is reported in the error message. Error message comes with a "skip this rule" recommendation.

Based on these issues:
- https://github.com/returntocorp/semgrep/issues/2237
- https://github.com/returntocorp/semgrep-rules/issues/1058

This should enable slightly easier debugging.


aryx (approved):
This is really useful. Thx!

PR #2354, opened 2021-01-08
[Java] change the syntax of method chaining to require extra "."

aryx:
This also fix a bug where db. ... .stream() was not allowed
because name is not part of primary in the grammar.

test plan:
test files included


aryx:
Closes https://github.com/returntocorp/semgrep/issues/2208

PR #2361, opened 2021-01-09
Metavariable interpolation for pattern-inside

### LINKED ISSUE #1699: pattern-inside metavariables are not bound in the message
**Describe the bug**
Metavariables used in a `pattern-inside` are not available in the `message` and are left unexpanded.

**To Reproduce**
https://semgrep.dev/s/gpoK/

**Expected behavior**
`$CLASS` in the above example should be expanded to `Foo`.

Is this a known limitation?

minusworld:
Fixes https://github.com/returntocorp/semgrep/issues/1699.

Only works with pattern + pattern-inside and with pattern-either + pattern.
Uses ranges to determine whether to apply metavariable interpolation. E.g., if a pattern-inside clause contains the pattern, it will interpolate the metavariables from the pattern-inside clause.

Does not work with pattern-not-inside.

This should handle cases with pattern + pattern-inside. This doesn't handle
pattern-not-inside because it is difficult to determine metavariables for
exclusion ranges. For example: imaagine a pattern-not-inside for 'def $CLASS(): ...'
and a file has multiple classes inside. How do we infer which metavariable was
intended for interpolation? As such, this will fix the immediate issue and should
handle the most common case.
Another corner case is: what should we do with nested metavars? Imagine 'def $FUNC(): ...'
and code with nested functions. Did we want the top-level function? The lowest-level? What
about other nesting cases? ¯\_(ツ)_/¯ Right now it will prefer the largest PatternMatch range

### DIFF
--- a/semgrep/semgrep/evaluation.py
+++ b/semgrep/semgrep/evaluation.py
@@ -375,19 +375,19 @@ def evaluate(
 
     # Addresses https://github.com/returntocorp/semgrep/issues/1699,
     # where metavariables from pattern-inside are not bound to messages.
-    # This should handle most cases. There are some edge cases we don't account for, such
-    # as a '$VAR' used in both pattern-inside and pattern-not-inside clauses. How do
-    # we infer which metavariable was intended for interpolation anyway? As such, this will
-    # fix the immediate issue and # should handle most cases. I'm leaving this comment here
-    # so that anyone who works on this in the future knows there will be some weirdness later. =X
-    all_pattern_match_metavariables: Dict[str, str] = {}
+    # This should handle cases with pattern + pattern-inside. This doesn't handle
+    # pattern-not-inside because it is difficult to determine metavariables for
+    # exclusion ranges. For example: imaagine a pattern-not-inside for 'def $CLASS(): ...'
+    # and a file has multiple classes inside. How do we infer which metavariable was
+    # intended for interpolation? As such, this will fix the immediate issue and should
+    # handle the most common case.
+    # Another corner case is: what should we do with nested metavars? Imagine 'def $FUNC(): ...'
+    # and code with nested functions. Did we want the top-level function? The lowest-level? What
+    # about other nesting cases? ¯\_(ツ)_/¯ Right now it will prefer the largest PatternMatch range.
+    all_pattern_match_metavariables: Dict[str, List[PatternMatch]] = defaultdict(list)
     for pattern_match in pattern_matches:
-        all_pattern_match_metavariables.update(
-            {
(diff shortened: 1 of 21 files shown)


aryx:
Do we still want this? What's up with the reviews?

semgrep/semgrep/evaluation.py
```diff
@@ -373,9 +373,27 @@ def evaluate(
         logger.debug(f"compiled result {valid_ranges_to_output}")
         logger.debug("-" * 80)
 
+    # Addresses https://github.com/returntocorp/semgrep/issues/1699,
+    # where metavariables from pattern-inside are not bound to messages.
+    # This should handle cases with pattern + pattern-inside. This doesn't handle
+    # pattern-not-inside because it is difficult to determine metavariables for
+    # exclusion ranges. For example: imaagine a pattern-not-inside for 'def $CLASS(): ...'
```
brendongo:
```suggestion
    # exclusion ranges. For example: imagine a pattern-not-inside for 'def $CLASS(): ...'
```

brendongo: IMO these should go in the ticket for metavariables in pattern-not-inside instead of as a comment in the code

semgrep/semgrep/evaluation.py
```diff
@@ -391,12 +409,30 @@ def evaluate(
...
-    for metavar in pattern_match.metavars:
-        msg_text = msg_text.replace(
-            metavar, pattern_match.get_metavariable_value(metavar)
-        )
+    for metavar_text in all_pattern_match_metavariables:
+        replace_text = metavar_text
+        try:  # Always prefer the pattern match metavariable first.
+            replace_text = pattern_match.get_metavariable_value(metavar_text)
+        except KeyError:  # If one isn't present, retrieve the value from all metavariables.
+            pattern_matches_with_metavars_that_enclose_match: List[PatternMatch] = list(
```
brendongo, line 423: This seems ok but feels very fragile. Is there a way we can have a pattern_match object contain a list of all metavars contained within it then in interpolation we can just do a lookup?

brendongo (approved):
LGTM. Seems fragile but I think without a major change to the pattern matching boolean logic  would be hard to have a cleaner solution.

semgrep/semgrep/evaluation.py
minusworld: Potentially: we might be able to populate sibling pattern nodes with all matched metavariables.

But then, I thought that wouldn't be so much more efficient than gathering up all the metavariables for all the patterns. 🤷‍♂️

PR #2363, opened 2021-01-09
Add Rust parsing support

Ruin0x11:
Adds Rust parsing support. This is up to the CST -> AST level. There is no pattern support yet.

Related PRs:

https://github.com/returntocorp/ocaml-tree-sitter/pull/134
https://github.com/returntocorp/ocaml-tree-sitter-lang/pull/5
https://github.com/returntocorp/pfff/pull/370
https://github.com/returntocorp/semgrep-grammars/pull/14

At the moment only one test on the Semgrep side is failing. It's weird since it parses correctly from tree-sitter.

Also, because there were so many language constructs that needed to be added, there are still a lot of TODOs for parts like flow control. I'm not entirely sure what those parts of the code are supposed to do yet. Some parts of the tree-sitter parser still need fixing also. For example, the `use` statement is an expression when it comes out of tree-sitter, but the Rust language docs say you can simplify it to a dotted identifier.


Ruin0x11:
As suggested, I'll break up this PR into multiple parts instead. The first PR is here: https://github.com/returntocorp/semgrep/pull/2365

PR #2370, opened 2021-01-11
Add most of the Rust parser code except new features

Ruin0x11:
Related: https://github.com/returntocorp/pfff/pull/372

I tried not to add many new language features this time, leaving everything else as a todo for other PRs.


semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -31,6 +31,11 @@ module G = AST_generic
 (*****************************************************************************)
 type env = unit H.env
 let token = H.token
+let str = H.str
+let sc = PI.fake_info ";"
```
aryx: You can reuse G.sc

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -48,289 +53,334 @@ let token = H.token
...
+       | `Esc_seq tok -> let (s, _) = str env tok in s (* escape_sequence *)
+       | `Str_content tok -> let (s, _) = str env tok in s (* string_content *)
       )
-    ) v2
-  in
-  let v3 = token env v3 (* "\"" *) in
-  todo env (v1, v2, v3)
+    ) v2 in
+  let rdquote = token env v3 (* "\"" *) in
+  G.String ((String.concat "" strs), ldquote)
```
aryx: We actually need the token to represent the full range. See the use of PI.combine_infos in Parse_go_tree_sitter.ml for example.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -48,289 +53,334 @@ let token = H.token
...
   (match x with
-  | `Str_lit x -> map_string_literal env x
-  | `Raw_str_lit tok -> token env tok (* raw_string_literal *)
-  | `Char_lit tok -> token env tok (* char_literal *)
-  | `Bool_lit x -> map_boolean_literal env x
-  | `Int_lit tok -> token env tok (* integer_literal *)
-  | `Float_lit tok -> token env tok (* float_literal *)
+   | `Str_lit x -> map_string_literal env x
+   | `Raw_str_lit tok -> G.String (str env tok) (* raw_string_literal *)
+   | `Char_lit tok -> G.String (str env tok) (* char_literal *)
```
aryx: There is a G.Char I think to represent characters.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -48,289 +53,334 @@ let token = H.token
...
+let map_literal_pattern (env : env) (x : CST.literal_pattern): G.pattern =
+  (match x with
+   | `Str_lit x -> G.PatLiteral (map_string_literal env x)
+   | `Raw_str_lit tok -> G.PatLiteral (G.String (str env tok)) (* raw_string_literal *)
+   | `Char_lit tok -> G.PatLiteral (G.Char (str env tok)) (* char_literal *)
+   | `Bool_lit x -> G.PatLiteral (map_boolean_literal env x)
+   | `Int_lit tok -> G.PatLiteral (G.Int (str env tok)) (* integer_literal *)
+   | `Float_lit tok -> G.PatLiteral (G.Float (str env tok)) (* float_literal *)
+   | `Nega_lit (v1, v2) ->
+       let (neg, _) = str env v1 (* "-" *) in
```
aryx: again we may need to use PI.combine_info. The reason is that in semgrep we rely on the tokens to compute the range of an expression when we return a match, to correctly highlight a matched region, but if the tok in the generic AST does not contain the right information, the range might be wrong, for example here the "-" part may not be highlighted.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -339,2122 +389,2262 @@ let map_macro_rule (env : env) ((v1, v2, v3) : CST.macro_rule) =
...
+   | `Func_type x -> map_function_type env x
   )
 
-and map_anon_choice_field_id_2c46bcf (env : env) (x : CST.anon_choice_field_id_2c46bcf) =
+and map_struct_name (env : env) (x : CST.anon_choice_field_id_2c46bcf): G.name =
   (match x with
-  | `Id tok ->
-      token env tok (* pattern (r#)?[a-zA-Zα-ωΑ-Ωµ_][a-zA-Zα-ωΑ-Ωµ\d_]* *)
-  | `Scoped_type_id x -> map_scoped_type_identifier env x
+   | `Id tok -> let ident = ident env tok in (* pattern (r#)?[a-zA-Zα-ωΑ-Ωµ_][a-zA-Zα-ωΑ-Ωµ\d_]* *)
```
aryx: You can maybe use some of the helpers functions in AST_generic_helpers module, for example
name_of_id to convert a single ident into a name.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -339,2122 +389,2262 @@ let map_macro_rule (env : env) ((v1, v2, v3) : CST.macro_rule) =
...
+         G.pname = Some self;
+         pdefault = None;
+         ptype = Some type_;
+         pattrs = attrs;
+         pinfo = G.empty_id_info ();
+       } in
+       G.ParamClassic param
+   | `Vari_param tok -> let ellipsis = token env tok in (* "..." *)
+       G.ParamEllipsis ellipsis
+   | `X__ tok -> let underscore = token env tok in (* "_" *)
```
aryx: Maybe you can use ParamPattern (PatUnderscore tok)

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -339,2122 +389,2262 @@ let map_macro_rule (env : env) ((v1, v2, v3) : CST.macro_rule) =
...
+       let mutability = Option.map (fun tok ->
+         let tok = token env tok (* "mut" *) in
+         G.KeywordAttr (G.Mutable, tok)
+       ) v2 in
+       let expr = map_expression env v3 in
+       G.Ref (ref_, expr)
+   | `Try_exp (v1, v2) ->
+       let v1 = map_expression env v1 in
+       let v2 = token env v2 (* "?" *) in
+       todo env (v1, v2)
```
aryx: could be a Call IdSpecial with Nullish. I think Typescript or Javascript have a similar operator.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -339,2122 +389,2262 @@ let map_macro_rule (env : env) ((v1, v2, v3) : CST.macro_rule) =
...
-        | None -> todo env ())
-      in
-      let v2 = token env v2 (* "for" *) in
-      let v3 = map_pattern env v3 in
-      let v4 = token env v4 (* "in" *) in
-      let v5 = map_expression env v5 in
-      let v6 = map_block env v6 in
-      todo env (v1, v2, v3, v4, v5, v6)
-  | `Const_blk x -> map_const_block env x
+   | `Unsafe_blk (v1, v2) ->
```
aryx, line 1121: You can use OtherStmtWithStmt and add a new UnsafeBlock constructor in
other_stmt_with_stmt_operator type.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -339,2122 +389,2262 @@ let map_macro_rule (env : env) ((v1, v2, v3) : CST.macro_rule) =
...
+       let unsafe = token env v1 (* "unsafe" *) in
+       let block = map_block_expr env v2 in
+       block (* TODO UnsafeBlock? *)
+   | `Async_blk (v1, v2, v3) ->
```
aryx, line 1126: Same, AsyncBlock next to UnsafeBlock in the other_stmt_with_stmt_operator.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -339,2122 +389,2262 @@ let map_macro_rule (env : env) ((v1, v2, v3) : CST.macro_rule) =
...
+  let if_ = token env v1 (* "if" *) in
+  let cond = map_expression env v2 in
+  let body = map_block env v3 in
+  let else_ = Option.map (fun x -> map_else_clause env x) v4 in
+  let if_stmt = G.If (if_, cond, body, else_) |> G.s in
+  G.OtherExpr (G.OE_StmtExpr, [G.S if_stmt])
+
+and map_if_let_expression (env : env) ((v1, v2, v3, v4, v5, v6, v7) : CST.if_let_expression): G.expr =
+  let if_ = token env v1 (* "if" *) in
+  let let_ = token env v2 (* "let" *) in
```
aryx, line 1508: Maybe you can use LetPattern

aryx (approved):
I didn't have the time to fully review everything but looks like a good start.
If you can fix the merge  conflicts (I think it's because you didn't indent correctly the first Parse_rust_tree_sitter.ml you committed with pre-commit/ocp-indent and I added such a commit later).

semgrep-core/parsing/Parse_rust_tree_sitter.ml
Ruin0x11: The semantics are different, it's more like a try-catch operator using Rust's `Result<T, E>` type. It replaces this

```rust
    let f = File::open("username.txt");
    let mut f = match f {
        Ok(file) => file,
        Err(e) => return Err(e),
    };
```

with this

``` rust
   let file = File::open("username.txt")?;
```

aryx: I see. There's a related thing in Kotlin with ?.  https://kotlinlang.org/docs/reference/null-safety.html
We might introduce indeed a new operator in AST_generic.ml for that. Not sure about a good name.

Ruin0x11:
`pfff` PR for OSWS: https://github.com/returntocorp/pfff/pull/375

semgrep-core/parsing/Parse_rust_tree_sitter.ml
Ruin0x11: It's called the [safe navigation operator](https://en.wikipedia.org/wiki/Safe_navigation_operator).

Perhaps `SafeNavigation`?

Ruin0x11: Actually it's already in the code. It's called `Elvis`, but Kotlin's `?` doesn't actually perform the same function as the elvis operator. I'll just use that for now.

aryx: Yes, it does not have to be exactly the same semantic. Right now we don't do any deep analysis about those operators, so as long as you don't use the same operator for 2 different constructs in the same language, you're fine. For example it's ok to use TyPointer in Rust for &, because there is no * in Rust, but for C++ we need to introduce a new TyRef because we don't want *x and &x to match to the same generic AST construct.

Ruin0x11: > there is no * in Rust

Actually there is, and it's called the "deref" operator. 

> Used for immutable dereferencing operations, like *v.

https://doc.rust-lang.org/std/ops/trait.Deref.html

Ruin0x11: Or if you were meaning pointer types, Rust has those also: https://doc.rust-lang.org/std/primitive.pointer.html

aryx: Hmm, in that case we may want to add TyRef that is currently in comment, and use TyRef for & and TyPointer for * for Rust (anyway I will need TyRef when I port tree-sitter-cpp to the generic AST).

Ruin0x11: Ok, I added it for now.

aryx:
Can you fix the merging conflicts?

Ruin0x11:
I can't tell what the conflicting files are because I don't have write access, so I merged develop.

aryx:
yep, it's a big mess; too many people modifying pfff and semgrep-core at the same time; plus some people forgot to setup pre-commit so we have also indentation conflicts.
Argh; I'll try to merge everything today.

aryx:
Still some conflicts. I've fetched your branch and merged develop here:
https://github.com/returntocorp/semgrep/pull/2385

aryx:
merged in #2385

PR #2378, opened 2021-01-12
Parse LINQ queries in C#

Sjord:
E.g.

```csharp
        var greeting = 
            from word in words
            let title = textInfo.ToTitleCase(word)
            select word + title;
```
I started converting LINQ queries to method calls, which the C# compiler also does internally. So this would get mapped like

```csharp
words.Select(word => (word, textInfo.ToTitleCase(word)).Select((word, title) => word + title)
```

I would like some feedback on this whether this is the correct approach. As you can see it gets pretty complex. Does this make sense? Or do you we need a AST representation for LINQ queries?

I currently map it to actual [Enumerable methods](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.select?view=net-5.0), but I guess semgrep doesn't care about that. How accurate does this needs to be?

### DIFF
--- a/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_csharp_tree_sitter.ml
@@ -70,6 +70,28 @@ let var_def_stmt (decls : (entity * variable_definition) list) (attrs : attribut
   ) decls in
   stmt1 stmts
 
+type direction =
+  | Ascending
+  | Descending
+
+and linq_query_part = 
+  | From of (tok * (type_ option * ident) * expr)
+  | Group of (tok * expr * expr)
+  | Select of (tok * expr)
+  | Into of (tok * ident * linq_query_part list)
+  | Join of (tok * (type_ option * ident) * expr * expr * expr * ident option)
+  | Let of (tok * ident * expr)
+  | OrderBy of (tok * (expr * direction) list)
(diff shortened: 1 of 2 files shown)


aryx:
I think it's ok to transpile, unless at some point people really want to match and find only Linq queries, in which case we might
need to extend the generic AST. As long as we transpile both the Code and the pattern, things should be fine.

semgrep-core/parsing/Parse_csharp_tree_sitter.ml
```diff
@@ -70,6 +70,87 @@ let var_def_stmt (decls : (entity * variable_definition) list) (attrs : attribut
   ) decls in
   stmt1 stmts
 
+type direction =
+  | Ascending
+  | Descending
+
+and linq_query_part =
```
aryx, line 77: Nice. We might put that at some point in AST_generic.ml instead, especially if another programming
language has a similar feature (e.g., F#, but Python has also some List comprehensions which are a bit
similar, even though far more limited).

aryx (approved):
Very nice!

aryx:
It has the "Draft" mark, so I can't merge it. Remove the tag and I'll merge the PR.

aryx:
I've found a way to remove the Draft tag. It's merged now.

PR #2381, opened 2021-01-13
Make pattern-not-regex valid with regex-only rules.

minusworld:
Silly me, I forgot to add `pattern-not-regex` to valid operators when working with regex-only rules. 😆


brendongo (approved):

brendongo:
Thanks for adding tests!

PR #2389, opened 2021-01-13

aryx:


aryx:
@mjambon if you look at the modifs in AST_generic.ml they are needed for metavariable comparison to work.
With this I was able to get more tests running with the calls to Caching.xxx, but I still had lots of assertion failures.

aryx:
hmm some CI failures. I hope it's not related to the const propagation I merged force before.

aryx:
Issue is tracker here: https://github.com/returntocorp/semgrep/issues/2396

Issue #2287, opened 2020-12-17
Ruby parser doesn't handle `rescue` blocks correctly


chgg-kboberg:
**Describe the bug**
Ruby parser doesn't correctly handle `rescue` blocks in function definitions, e.g.

```
def $METHOD(...)
  ...
end
```
should match
```
def someFunction(foo)
  someOtherFunction()
rescue SomeClass::Error
  errorHandler()
end
```

**To Reproduce**
https://semgrep.dev/s/gpde/

**Expected behavior**
Semgrep should find a match but does not.

**What is the priority of the bug to you?**
This is a P1.  You can work around this with explicit handling of `rescue`: https://semgrep.dev/s/Qwvo/
**Environment**
semgrep.dev

mschwager:
Hi @chgg-kboberg,

Nice find, this is quite an odd bug. It looks like we successfully match when the optional `begin`/`end` are included: https://semgrep.dev/s/3eoo/.

Following along [here](https://www.rubyguides.com/2019/06/ruby-rescue-exceptions/) it looks like they are optional like you're example specifies. cc @aryx or @mjambon for a parsing assist.

mjambon:
From @mschwager 's example https://semgrep.dev/s/3eoo/, it looks like a parsing error because the correct `end` isn't included in the match.
![image](https://user-images.githubusercontent.com/343265/102542064-177f6980-4066-11eb-99ee-a833ff16c897.png)

(but maybe it's something else)

aryx:
Hopefully I can work on it next week.  A  it swamped this week with backlogs.

PR #2390, opened 2021-01-13
Added tests for ruby rescue bug

emjin:


emjin:
Closes #2287

aryx:
Again, better to have a nice test plan: section in the commit message (for semgrep it can be as simple as
'test plan: make test, see new test files included.'
Also better to include the issue it closes in the commit message if possible.

aryx (approved):

aryx (commented):
We try to follow some informal conventions on the name of the test files. Some of them are used to generate the cheatsheets in semgrep.live (see semgrep/scripts/generate_test_matrix.py).
Here maybe you could call it dots_rescue_function.rb.

PR #2391, opened 2021-01-14
Parse semgrep expression in Rust parser

Ruin0x11:
Adds semgrep patterns and an alternate toplevel entry point to the Rust parser.


semgrep-core/tests/Test.ml
```diff
@@ -258,6 +258,12 @@ let lang_regression_tests =
     let lang = Lang.Lua in
     regression_tests_for_lang files lang
   );
+  "semgrep Rust" >::: (
+    let dir = Filename.concat tests_path "rust" in
+    let files = Common2.glob (spf "%s/*.rs" dir) in
```
aryx, line 263: Did you forget to add some test files in the PR?
The glob might raise an exn if there is no file in the directory I think.

aryx (approved):
Could you remove your modif on pfff as it generates a conflicts (and the modification shown have already been merged).

aryx (changes requested):
Remove modif to pfff so I can merge.

aryx:
Not sure why, but above on the "This branch has conflicts that must be resolved" I can't even click on the "Use the command line" link; it does not show any help anymore (it used to show a list of git commands to run to fix the conflict).
Github bug?

brendongo:
It probably has something to do with pfff being a submodule?

aryx:
Maybe. I thought I had similar conflicts before and was able to click on "use the command line" but maybe it was never a conflict about a submodule.

semgrep-core/tests/Test.ml
Ruin0x11: Yes, I did. I added them.

aryx (approved):

PR #2393, opened 2021-01-14
Add Rust trait definitions to parser

Ruin0x11:
There are some constructs that don't fit cleanly in the generic AST, like trait bounds. It's mostly because trait bounds in Rust can include lifetimes, which aren't types, and "higher ranked trait bounds" that have other syntax: https://doc.rust-lang.org/nomicon/hrtb.html

I think the current implementation is pretty close, though.

### DIFF
--- a/semgrep-core/parsing/Parse_rust_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_rust_tree_sitter.ml
@@ -562,18 +562,6 @@ and map_type_argument (env : env) (x : CST.anon_choice_type_39799c3): G.type_arg
    | `Blk x -> let block_expr = map_block_expr env x in todo env x
   )
 
-and map_trait_bound (env : env) (x : CST.anon_choice_type_d689819) =
-  (match x with
-   | `Type x -> let ty = map_type_ env x in todo env x
-   | `Life x -> let lt = map_lifetime env x in todo env x
-   | `Higher_ranked_trait_bound x -> let tb = map_higher_ranked_trait_bound env x in
-       todo env x
-   | `Remo_trait_bound (v1, v2) ->
-       let v1 = token env v1 (* "?" *) in
-       let v2 = map_type_ env v2 in
-       todo env (v1, v2)
-  )
-
 and map_tuple_pattern_list (env : env) ((v1, v2) : CST.anon_pat_rep_COMMA_pat_2a80f16): G.pattern list =
   let pattern_first = map_pattern env v1 in
   let pattern_rest =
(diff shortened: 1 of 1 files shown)


semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -2143,22 +2125,41 @@ and map_statement (env : env) (x : CST.statement): G.stmt list =
...
-      G.DefStmt def |> G.s
-  | `Asso_type x -> map_associated_type env x
+      G.FieldStmt (G.DefStmt def |> G.s)
+  | `Asso_type x -> G.FieldStmt (map_associated_type env x)
   | `Macro_invo x -> let invo = map_macro_invocation env x in
-      G.ExprStmt (invo, sc) |> G.s
+      G.FieldStmt (G.ExprStmt (invo, sc) |> G.s)
+
+and map_higher_ranked_trait_bound (env : env) ((v1, v2, v3) : CST.higher_ranked_trait_bound): (G.type_parameter list * G.type_) =
+  let for_ = token env v1 (* "for" *) in
```
aryx, line 2146: BTW at some point you may want to remove the [@@@ warning "-26-27"] directives at the top
to get the unused_variable check from the OCaml compiler back on, so you'll have to prefix
those unused tokens with an '_', e.g. let _for = token ... here.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -2143,22 +2125,41 @@ and map_statement (env : env) (x : CST.statement): G.stmt list =
...
+  let type_parameters = map_type_parameters env v2 in
+  let type_ = map_type_ env v3 in
+  (type_parameters, type_)
+
+and map_trait_bound (env : env) (x : CST.anon_choice_type_d689819): G.type_ option =
```
aryx, line 2151: yep, let's skip the most complex constructs for now.

semgrep-core/parsing/Parse_rust_tree_sitter.ml
```diff
@@ -2616,8 +2617,20 @@ and map_item_kind (env : env) outer_attrs visibility (x : CST.item_kind): G.stmt
...
+       let fields = map_trait_block env v7 in
+       let class_def = {
+         G.ckind = (G.Trait, trait);
+         G.cextends = [];
+         G.cimplements = trait_bounds;
+         G.cmixins = [];
+         G.cbody = fields;
+       } in
+       let ent = {
+         G.name = G.EId (ident, G.empty_id_info ());
```
aryx, line 2629: You can use the G.basic_entity helper instead of building the record yourself.

aryx (approved):

PR #2405, opened 2021-01-17
use latest ocaml-tree-sitter-lang

aryx:


aryx:
cc @Ruin0x11 had to fix Parse_lua_tree_sitter.ml and Parse_rust_tree_sitter.ml to compile with the latest tree-sitter grammars
of lua and rust.

PR #2406, opened 2021-01-17
Better handle Lua multi assignment/declaration

### LINKED ISSUE #2404: Lua method calls not fully handled
Hi folks, congratulations on the release, and thanks again @Ruin0x11 and everyone involved with adding Lua support!

I just ran the sample `$X == $X` rule to get started, and found an issue.

**Describe the bug**
Method calls (of the form `object:method()` seem to be only partly handled.

In Lua `foo:bar("abc")` is syntactically equivalent to `foo.bar(foo, "abc")`, but it is not parsed this way by semgrep.

**To Reproduce**
I'd link to semgrep.dev, but Lua is not listed there (yet? :) ).

Lua code:

```
return foo:one() == foo:two()
```

incorrectly matches the pattern `$X == $X`.

**Expected behavior**

I'm not certain, but I feel that probably the solution is translating to the canonical `object.method(object)` form in the AST (and then translating patterns the same way).

Method calls are used extensively, and right now it appears that semgrep cannot match them correctly at all. I think that makes this a pretty high priority issue.

**Environment**
I tested with semgrep 0.37.0 from pip.

Ruin0x11:
Closes #2404.


semgrep-core/parsing/Parse_lua_tree_sitter.ml
```diff
@@ -703,34 +700,26 @@ and map_variable_declarator_expr (env : env) (x : CST.variable_declarator): G.ex
...
-  )
-  in
-  { G.name = s; attrs = []; tparams = [] }
-
-
+and map_variable_declarator (env : env) (x : CST.variable_declarator): G.expr =
+  match x with
+  | `Id tok ->
+      G.Id (identifier env tok, G.empty_id_info ()) (* pattern [a-zA-Z_][a-zA-Z0-9_]* *)
+  | `Self tok ->
```
aryx: Self is not part of variable_declarator anymore in the latest tree-sitter-lua. I've updated
ocaml-tree-sitter-lang and semgrep in https://github.com/returntocorp/semgrep/pull/2405
you probably need to merge develop to get the latest and resubmit a PR.

aryx (changes requested):

semgrep-core/tests/lua/assignment.lua
```diff
@@ -0,0 +1,19 @@
...
+foo = foo
+
+qux = foo:bar() == foo:baz()
+
+qux = foo:bar() == foo:bar()
+
+local a, b, c = 1, 2, 3
+
+-- ERROR: match
+a, b, c = a
```
aryx, line 16: This should not match, unless I don't understand the semantic of equal in Lua.
Is it equivalent to  OCaml let (a,b,c) = a? 
In that case it should be also transformed into an Assign (Tuple (a,b,c), a)

semgrep-core/tests/lua/assignment.lua
```diff
@@ -0,0 +1,19 @@
...
+
+-- ERROR: match
```
aryx, line 18: Again, I don't think it should match.
This should be transformed into
Assign (a, Tuple(a,b,c))

aryx (approved):

semgrep-core/tests/lua/assignment.lua
Ruin0x11: No, `a, b = 1, 2` is the equivalent of `a = 1; b = 2`. It's similar to Go.

semgrep-core/tests/lua/assignment.lua
Ruin0x11: Same as above.

aryx (approved):

Issue #2396, opened 2021-01-14
Restore we45/Vulnerable-Flask-app  in CI (was Matchtimeout error only in CI)


mschwager:
https://github.com/returntocorp/semgrep/runs/1704014556

```
 E               subprocess.CalledProcessError: Command '['/home/runner/work/semgrep/semgrep/semgrep/.tox/py/bin/python', '-m', 'semgrep', '--config=rules/regex-sentinel.yaml', '--strict', '--json', PosixPath('/home/runner/.cache/semgrep-tests/we45-Vulnerable-Flask-App-latest')]' returned non-zero exit status 9.
```

https://github.com/returntocorp/semgrep/runs/1704014685

```
E   Failed: Running on https://github.com/we45/Vulnerable-Flask-App with lang javascript had errors (['/home/runner/.local/share/virtualenvs/semgrep-GqZ0EjEI/bin/python', '-m', 'semgrep', '--pattern', '$SENTINEL = 87518275812375164', '--lang', 'javascript', '--json', PosixPath('/home/runner/.cache/semgrep-cache/we45-Vulnerable-Flask-App-latest')]): [
E       {
E           "type": "MatchTimeoutError",
E           "code": 9,
E           "path": "/home/runner/.cache/semgrep-cache/we45-Vulnerable-Flask-App-latest/app/static/loader.js",
E           "rule_id": "-"
E       }
E   ]
```

cc @aryx @brendongo

mschwager:
Looks like [`loader.js`](https://github.com/we45/Vulnerable-Flask-App/blob/master/app/static/loader.js) is a large minified file, but I'm unable to reproduce locally:

```
$ python -m semgrep -l js -e '$SENTINEL = 87518275812375164' /tmp/Vulnerable-Flask-App/ --verbose --debug
Running rule -...

...ran on 1 files
semgrep ran in 0:00:12.168811 on 1 files
findings summary: 0 error
ran 1 rules on 1 files: 0 findings
```

aryx:
cc @mjambon

aryx:
There was also some flakyness in CI on btcexport, sometimes working, sometimes not, but there
was actually 2 entries in the test_public_repos.py file, so I removed one; maybe it fixes the problem.

aryx:
Fix also coinbase-pro-node in public_repos.py that I've commented the xfail_repo when doing "the big refactoring".

mjambon:
[...]

PR #2410, opened 2021-01-18
* semgrep/tests/public_repos.py: fix CI, move repo in xfail_repo

aryx:
test plan:
Unable to reproduce the failure locally. make tests works on my
machine and matt's machine. Something happens in CI, maybe slower
CI machine leads to some timeout.


aryx:
This is a temporary fix for https://github.com/returntocorp/semgrep/issues/2396

PR #2413, opened 2021-01-19
Add Rust trait bounds / where clauses

Ruin0x11:
Fixes a `todo` for Rust `where` clauses: https://doc.rust-lang.org/rust-by-example/generics/where.html

I intentionally avoided adding anything in `pfff` and left all the new types internal to the parser. However, I want to know if there's a good way of putting the clauses in the generic AST somehow.

### DIFF
--- a/semgrep-core/parsing/Parse_rust_tree_sitter.ml
+++ b/semgrep-core/parsing/Parse_rust_tree_sitter.ml
@@ -61,6 +61,24 @@ type function_declaration_rs = {
   retval: G.type_ option;
 }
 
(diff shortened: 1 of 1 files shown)


aryx:
I think you should use the type_parameter_constraint to represent those Where constraints,
and then there is already support for intersection and union types in the AST, see TyOr and TyAnd
so a: Trait1 + TraitB could maybe represented as a Extends (TyAnd (TyId "Trait1", "TyId "traitB")

aryx (approved):

Issue #2349, opened 2021-01-07
[Perf] A basic linter for semgrep rules (detect really-slow patterns)


aryx:
Some patterns can be too general and can leads to too many matches that can cause the semgrep engine to
timeout.
Maybe we could at least warn about such patterns.

aryx:
We could use spacegrep to write rules to run on semgrep-rules/, but for now I think a few hardcoded AST-based checks in semgrep-core might be enough.
For example I've found in semgrep-rules some patterns like: $FOO
that will match everything (the rules contains a few pattern-inside that combines with this pattern, but semgrep-core does not know about that, and this leads to some timeout).

aryx:
example of bad patterns:

```
$FOO
```
```
...
$FOO
...
```

```
...
any pattern
...
```

this last one should be rewritten

aryx:
Note that in many of those cases we want to explicicely match a metavar or a ..., so spacegrep might not be able to help.

aryx:
maybe I can plug in 'semgrep --validate' used in semgrep-rules @minusworld to report possible performance problems?

aryx:
related issue that was causing performance problems: https://github.com/returntocorp/semgrep/issues/2102

aryx:
because of typescript rules comtaining just pattern: '$PROP' if I remember correctly

aryx:
Bonus: fix semgrep-rules
Bonus: plug in web UI to give warnings

aryx:
examples of slow rules we removed to detect: https://github.com/returntocorp/semgrep-rules/pull/1059#event-4182269988

aryx:
Have a look at Check_semgrep.ml
Should be called by semgrep -validate

aryx:
Another bug to detect: identical patterns:
see hardcoded_passport_secret in semgrep-rules/ where 
  - pattern: |
      $F = require("passport-auth0").Strategy;
      ...
      $P.clientSecret = "...";
      ...
      new $F($P, ...);

is duplicated once.

aryx:
We may need to parse the whole yaml file and have a global view of all the patterns, not just analyze one by one separately
how we do in semgrep-core -validate_pattern.

aryx:
For the duplicated pattern check, would be nice to write it via semgrep :) meta! like
pattern: $X
...
pattern: $X

PR #2420, opened 2021-01-20
Check_rule semgrep rule metachecker

aryx:


aryx:
closes https://github.com/returntocorp/semgrep/issues/2349

PR #2422, opened 2021-01-20
Typed metavariables in go match literal strings

### LINKED ISSUE #2401: Typed metavariables in Go doesn't match constants
**Describe the bug**
Constants are not matched when using typed metavariables in Go

**To Reproduce**
https://semgrep.dev/s/emjin:go-string

**Expected behavior**
Expected first two to both match

Unknown, user bug

**Environment**

emjin:
Closes #2401

Fixed typo in Generic_vs_generic that expects strings to be encoded as "str" rather than "string"

Added a test specifically for literals and modified metavar_typed_args appropriately.


aryx (approved):
Excellent!

PR #2423, opened 2021-01-20
Explicit --timeout 0 should disable timeouts

### LINKED ISSUE #2411: Rules with "too many matches" generate confusing "Timeout" error
**Describe the bug**
I am using semgrep in Azure Pipeline for PR analysis. Semgrep is getting timedout for a small JS file (~6K lines) in 5 seconds. I have seen multiple threads related to timeouts in JS but did not find any conclusion or plan of action. I have tried the command locally with only this file but no luck. Is there other workaround or any other option that I could use to get a scan no matter the time it takes? 

I have seen the use of -max_match_per_file  flag being used for semgrep-core but not part of the CLI. Is it possible to use this with CLI? Will this help with time outs?

**To Reproduce**
Unfortunately, I can not share the JS file. Here is the command. The failing rule is javascript.browser.security.raw-html-concat.raw-html-concat

semgrep   --config "p/clientside-js"  --timeout 0 --error --verbose --disable-version-check --include "********/application.js" --debug --strict

**Expected behaviour**
I expect the scan to complete as I've given timeout is 0.

P0 - we are using semgrep to provide a report of secure coding practices to a client but since this single is not getting scanned we are unable to use semgrep.

**Environment**
Python package - 0.37.0

nbrahms:
Previously we defaulted to 5 seconds when timeout was 0, but our
documentation states that explicitly setting timeout to 0 should disable
timeouts.

Note that, practically, the OCaml core will _never_ receive an unset
`-timeout` option, as the Python wrapper defaults to 30 seconds.

Fixes #2411.

### DIFF
--- a/semgrep-core/parsing/Parse_code.ml
+++ b/semgrep-core/parsing/Parse_code.ml
@@ -210,24 +210,7 @@ let just_parse_with_lang lang file =
       *)
       run file [
         TreeSitter Parse_javascript_tree_sitter.parse;
-        Pfff (throw_tokens (fun file ->
-          let f () =
-            Parse_js.parse file
-          in
-          (* timeout already set in caller, then good to go *)
-          if !Flag.timeout <> 0.
-          then f ()
-          else begin
-            try
-              Common.timeout_function 5 (fun () ->
-                logger#info "running the pfff JS parser with 5s timeout";
-                f ()
-              )
-            with Timeout ->
-              logger#debug "Timeout, transforming in parse error";
-              raise Parsing.Parse_error
-          end
-        ));
+        Pfff (throw_tokens Parse_js.parse)
       ]
         Js_to_generic.program
 
--- a/semgrep-core/cli/Main.ml
+++ b/semgrep-core/cli/Main.ml
@@ -1189,7 +1189,7 @@ let options () =
     " only use tree-sitter-based parsers";
 
     "-timeout", Arg.Set_float Flag.timeout,
-    " <float> time limit to process one input program (in seconds)";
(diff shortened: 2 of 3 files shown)


brendongo:
Maybe we should change the default timeout to not be zero. The reason we had a non zero timeout for JS was so large minified JS files don't cause semgrep to stall.

brendongo:
Oh I see default is 30 nvm

brendongo (approved):

semgrep-core/parsing/Parse_code.ml
```diff
@@ -210,24 +210,7 @@ let just_parse_with_lang lang file =
...
-          then f ()
-          else begin
-            try
-              Common.timeout_function 5 (fun () ->
-                logger#info "running the pfff JS parser with 5s timeout";
-                f ()
-              )
-            with Timeout ->
-              logger#debug "Timeout, transforming in parse error";
-              raise Parsing.Parse_error
```
aryx, line 228: This line is why I don't see how he possibly got a timeout. He should have had a parse error,
and I'm afraid with this patch we will have regressions for some users because a file will now take forever
because it does not parse with tree-sitter and we default to the one in pfff which is really bad at error recovery.
This timeout is a trick to go around bad error-recovery algorithm.
This timeout is an internal thing, that should not leak to the user.

jithurjacob: @aryx - I've added the screenshot for timeout error - https://github.com/returntocorp/semgrep/issues/2411#issuecomment-764681925

Issue #1256, opened 2020-07-13
Metavariable information discarded when using typed metavariables


emjin:
Metavariable not being properly extracted from pattern-where-python

Pattern at: https://semgrep.live/6JDY

Trace:

```
$ semgrep --dangerously-allow-arbitrary-code-execution-from-rules -l go -f ~/go/src/github.com/dgryski/semgrep-go/wrongerr.yml
running 1 rules...
error evaluating a where-python expression: `vars[‘$ERR1’] != vars[‘$ERR’].split(' ‘)[-1]`: ‘$ERR1’
error evaluating a where-python expression: `vars[‘$ERR1’] != vars[‘$ERR’].split(' ‘)[-1]`: ‘$ERR1’
error evaluating a where-python expression: `vars[‘$ERR1’] != vars[‘$ERR’].split(' ‘)[-1]`: ‘$ERR1’
error evaluating a where-python expression: `vars[‘$ERR1’] != vars[‘$ERR’].split(' ‘)[-1]`: ‘$ERR1’
```
```
$ cat !$
cat ~/go/src/github.com/dgryski/semgrep-go/wrongerr.yml
rules:
- id: wrong-else-nil-check
  message: Checking the wrong err variable in an if block
  languages: [go]
  severity: WARNING
  patterns:
  - pattern-either:
          - pattern: “if $ERR := ...; ($ERR1 : error) != nil { ... }”
          - pattern: “if $ERR = ...; ($ERR1 : error) != nil { ... }”
  - pattern-where-python: vars[‘$ERR1’] != vars[‘$ERR’].split(' ’)[-1]
```

nbrahms:
is `$ERR1` a valid metavariable? for some reason I thought one couldn't use numerals

dgryski:
Yes, because this is a valid file:
```
rules:
- id: wrong-else-nil-check
  message: Checking the wrong err variable in an if block
  languages: [go]
  severity: WARNING
  patterns:
  - pattern-either:
          - pattern: "if $ERR := ...; $ERR1 != nil { ... }"
          - pattern: "if $ERR = ...; $ERR1 != nil { ... }"
  - pattern-where-python: vars['$ERR1'] != vars['$ERR'].split(' ')[-1]
```

mschwager:
Hi all,

[...]

PR #2441, opened 2021-01-21
Fix bug that caused m_compatible_type to only bind the type

emjin:
Previously, when matching a type with an id, we would only call m_type_option_with_hook, which would add the metavariable for the type into the environment if applicable, but not the actual typed metavar. Similarly, the C string metavariable pointer type change made the same error.

Now, we bind the matcher to the matcher for the typed metavar.

Test plan: make test with cases in metavar_typed_reuse


emjin:
I believe this closes #1256, or at least my part of it. (Apologies for the delay)

aryx (approved):

PR #2446, opened 2021-01-22
Add rust macros parsing

Ruin0x11:
I removed the TODOs related to Rust macros.

I want to know if we should have a new `expr` type for these macros, or if it would better to just wrap everything in lists of `any`. They are very different from the macro definitions available in `pfff`, which are more C-like,


aryx:
For now maybe list of anys ... like there is in AST_generic.macro_definition.macrobody type. You can keep your clean macro AST type and convert into anys.
Later we might integrate your macro type in AST_generic.ml and merge it with the current AST_generic.macro_definition.

aryx:
Can you merge develop and rerun CI?
There are lots of failures here.

Ruin0x11:
I'll create a new PR incorporating the project structure changes.

PR #2450, opened 2021-01-22
Only warn on unrunnable pattern-where-python, unless --strict

### LINKED ISSUE #2427: Rule with code execution is not skipped and fails whole scan (and fails pipeline as a result)
**Describe the bug**
Sorry to create is as bug, while it also could be considered as feature. But for me it currently fails whole pipeline.
I'm running semgrep with setting --config=semgrep-rules/, where "semgrep-rules/" is pulled from repo.
In case of there is a rule with code execution inside, whole scan stopped and says:

`at least one rule needs to execute arbitrary code; this is dangerous! if you want to continue, enable the flag: --dangerously-allow-arbitrary-code-execution-from-rules`

Adding this flags is not acceptable in my and most probably many other cases

**To Reproduce**
Run scan on any project with "--config=semgrep-rules/" 

**Expected behavior**
Skip rules with code execution if no flag "--dangerously-allow-arbitrary-code-execution-from-rules", just print warning but do not stop whole scan

P0 - fails pipeline.

**Environment**
official binary

mschwager:
Fixes #2427.

```
$ cat code.py 
1 == 1
```

```
$ cat safe.yml 
rules:
- id: eqeq-is-bad1
  pattern: 1 == 1
  message: 1 == 1 is a useless equality check
  languages: [python]
  severity: ERROR
```

```
$ cat dangerous.yml 
rules:
- id: eqeq-is-bad2
  patterns:
  - pattern: $X == $X
  - pattern-where-python: bool(vars['$X'])
  message: X == X is a useless equality check
  languages: [python]
  severity: ERROR
- id: eqeq-is-bad3
  pattern: 3 == 3
  message: 3 == 3 is a useless equality check
  languages: [python]
  severity: ERROR
```

```
$ python -m semgrep --config /tmp/dangerous-test/ /tmp/dangerous-test/code.py 
skipping rule 'tmp.dangerous-test.eqeq-is-bad2' due to pattern-where-python usage, pass '--dangerously-allow-arbitrary-code-execution-from-rules' to enable
running 2 rules...
/tmp/dangerous-test/code.py
severity:error rule:tmp.dangerous-test.eqeq-is-bad1: 1 == 1 is a useless equality check
1:1 == 1
ran 2 rules on 1 files: 1 findings
```

```
$ python -m semgrep --config /tmp/dangerous-test/ /tmp/dangerous-test/code.py --strict
skipping rule 'tmp.dangerous-test.eqeq-is-bad2' due to pattern-where-python usage, pass '--dangerously-allow-arbitrary-code-execution-from-rules' to enable
failing due to unrunnable pattern-where-python and --strict
```

```
$ python -m semgrep --config /tmp/dangerous-test/ /tmp/dangerous-test/code.py --dangerously-allow-arbitrary-code-execution-from-rules
running 3 rules...
/tmp/dangerous-test/code.py
severity:error rule:tmp.dangerous-test.eqeq-is-bad1: 1 == 1 is a useless equality check
1:1 == 1
severity:error rule:tmp.dangerous-test.eqeq-is-bad2: X == X is a useless equality check
1:1 == 1
ran 3 rules on 1 files: 2 findings
```

### DIFF
--- a/semgrep/semgrep/evaluation.py
+++ b/semgrep/semgrep/evaluation.py
@@ -228,9 +228,6 @@ def _evaluate_single_expression(
                 f"at least one rule needs to execute arbitrary code; this is dangerous! if you want to continue, enable the flag: {RCE_RULE_FLAG}",
                 code=NEED_ARBITRARY_CODE_EXEC_EXIT_CODE,
             )
-        if not isinstance(expression.operand, str):
-            raise SemgrepError("pattern-where-python must have a string value")
-
         # Look through every range that hasn't been filtered yet
         for pattern_match in list(flatten(pattern_ids_to_pattern_matches.values())):
             # Only need to check where-python clause if the range hasn't already been filtered
@@ -240,7 +237,7 @@ def _evaluate_single_expression(
                     f"WHERE is {expression.operand}, metavars: {pattern_match.metavars}"
                 )
                 if _where_python_statement_matches(
-                    expression.operand, pattern_match.metavars
+                    expression.operand, pattern_match.metavars  # type: ignore
                 ):
                     output_ranges.add(pattern_match.range)
         add_debugging_info(
--- a/semgrep/semgrep/util.py
+++ b/semgrep/semgrep/util.py
@@ -6,6 +6,7 @@ import subprocess
 import sys
 from typing import Any
 from typing import Callable
+from typing import Dict
 from typing import IO
 from typing import Iterable
 from typing import List
@@ -172,3 +173,17 @@ def compute_semgrep_path() -> str:
 
 def compute_spacegrep_path() -> str:
     return compute_executable_path("spacegrep")
+
+
+def recursive_has_key(pred: Callable, d: Dict) -> bool:
+    if pred(d):
+        return True
+    result = False
+    for v in d.values():
+        if isinstance(v, dict):
+            result = result or recursive_has_key(pred, v)
+        elif isinstance(v, list):
+            result = result or any(
+                recursive_has_key(pred, i) for i in v if isinstance(i, dict)
+            )
+    return result
--- a/semgrep/semgrep/util.py
+++ b/semgrep/semgrep/util.py
@@ -175,16 +175,17 @@ def compute_spacegrep_path() -> str:
     return compute_executable_path("spacegrep")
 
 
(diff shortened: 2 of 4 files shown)


semgrep/semgrep/evaluation.py
```diff
@@ -228,9 +228,6 @@ def _evaluate_single_expression(
                 f"at least one rule needs to execute arbitrary code; this is dangerous! if you want to continue, enable the flag: {RCE_RULE_FLAG}",
                 code=NEED_ARBITRARY_CODE_EXEC_EXIT_CODE,
             )
-        if not isinstance(expression.operand, str):
```
mschwager, line 231: This is guaranteed by the schema now :+1:

semgrep/semgrep/evaluation.py
```diff
@@ -240,7 +237,7 @@ def _evaluate_single_expression(
                     f"WHERE is {expression.operand}, metavars: {pattern_match.metavars}"
                 )
                 if _where_python_statement_matches(
-                    expression.operand, pattern_match.metavars
```
mschwager, line 243: Same as above, but `mypy` doesn't know this.

semgrep/semgrep/util.py
```diff
@@ -172,3 +173,18 @@ def compute_semgrep_path() -> str:
 
 def compute_spacegrep_path() -> str:
     return compute_executable_path("spacegrep")
+
+
+def recursive_has_key(pred: Callable, d: Dict) -> bool:
```
brendongo: More a recursive generic boolean than just has_key right?

brendongo (approved):
Just need a better name for util function and function level comment on usage would be helpful

semgrep/semgrep/util.py
mschwager: Yeah, something like `recursive_has_item` makes more sense

dlukeomalley:
@mschwager Will Semgrep return a failing exit code, but run as many rules as it can, with this change? From a trust perspective our philosophy has been to loudly error if a rule cannot run (to avoid a false sense of "Semgrep passed, everything must be good). In the context of a CI run I think this is particularly important - skipped rules result in errors unless explicitly overriden or ignored.

PR #2458, opened 2021-01-26
Memoize matches for lists of statements

mjambon:
This goes with https://github.com/returntocorp/pfff/pull/381


semgrep-core/cli/Main.ml
```diff
@@ -726,15 +733,17 @@ let json_of_exn e =
 (*****************************************************************************)
 
 (*s: function [[Main_semgrep_core.semgrep_with_rules]] *)
-let semgrep_with_rules rules files =
+let semgrep_with_rules ~with_opt_cache rules files =
```
aryx, line 736: An alternative is to put this in Flag_semgrep.ml as a global, so you don't have to thread it all around.

semgrep-core/core/Caching.ml
```diff
@@ -0,0 +1,376 @@
...
+   Decorate a pattern and target ASTs to make the suitable for memoization
+   during matching.
+*)
+
+module MV = Metavariable
+
+open Printf
+open AST_generic
+
+let debug = false
```
aryx, line 11: You can use the pfff/commons/logging.ml module.
git grep logger for example of use inside semgrep.

semgrep-core/core/Caching.ml
```diff
@@ -0,0 +1,376 @@
...
+
+(* A set of metavariables. Access cost is O(log n). *)
+module Names = AST_generic.String_set
```
aryx, line 14: You know, in practice it might be better to use a list ... there are very few metavariables usually
in patterns, and in the past I sometimes found that plain list can be faster than binary-trees set.

semgrep-core/core/Metavariable.ml
```diff
@@ -15,9 +15,15 @@
  * license.txt for more details.
 *)
 (*e: pad/r2c copyright *)
+open Printf
 open Common
 module G = AST_generic
 
+(* Provide hash_* and hash_fold_* for the core ocaml types *)
+open Ppx_hash_lib.Std.Hash.Builtin
```
aryx, line 23: This is annoying ... deriving eq and show do not need this.

semgrep-core/parsing/Parse_target.ml
```diff
@@ -280,7 +280,6 @@ let parse_and_resolve_name_use_pfff_or_treesitter lang file =
   Naming_AST.resolve lang ast;
   Constant_propagation.propagate_basic lang ast;
   Constant_propagation.propagate_dataflow ast;
-  (* TODO:  Caching.prepare_target ast; commented because of test regressions *)
```
aryx, line 283: No need for prepare_target here anymore?

aryx (approved):
Big diff.
I don't follow everything, but does not look like it breaks what was before so let's merge this.

semgrep-core/core/Metavariable.ml
mjambon: I'm not sure why I had to resort to this. It seems wrong indeed.

semgrep-core/parsing/Parse_target.ml
mjambon: That's correct. It was in charge of numbering the node IDs, but now it's done at the time of creating the stmt node, rather than mutating the field later. The motivation was the ppx_hash extension which refuses to run on mutable fields.

PR #2464, opened 2021-01-26
Use bloom filters to reduce tree traversal

### LINKED ISSUE #2352: [Perf] use nathan's bloom filter idea to improve matching performance
see https://returntocorp.slack.com/archives/C6A1GN4P3/p1609872936216400 for more information

emjin:
Closes #2352

When deciding whether or not to match a subtree, we can definitely choose not to if essential parts of the pattern are not present. It is cheap and space efficient to first hash these essential parts using a bloom filter, which will guarantee a "no" answer if it is not present. Then, when matching the pattern against a node, we only traverse further if all essential parts of the pattern are present in that node.

For the current implementation:
- the essential parts are non-metavariable idents and string literals
- a bloom filter is stored at every statement node

The initial bloom annotation is worst case O(sn), where s is the number of strings and n is the number of statement nodes. This is because each string for a given statement is added to that statement's bloom filter and each of its parents'. Note that this is a *very* loose bound

Later computations will be O(s_p), where s_p is the number of strings in the pattern.

Test plan: make test (correctness should still hold), benchmarking to see if it is faster. Will also need to tweak the parameters passed into Bloomf.create

This implementation makes use of the bloomf library in opam


mjambon:
@pad can we eliminate this semgrep rule:
```
severity:warning rule:ocaml.lang.best-practice.string.ocamllint-str-first-chars: Use instead `Str.first_chars`
82:  AST_generic_.is_metavar_name str || (String.length str > 4 && (String.sub str 0 4) = "$...") || is_regexp_string str
```
`Str.first_chars "abc" 4` is the same as `String.sub "abc" 0 4`, both raise the same exception. And `Str` requires the str library.

mjambon (approved):

aryx:
I like the semgrep check. It's easier to read, but I let @emjin decide if she thinks it's a useless rule.
You can update the semgrep-rules repo to remove the rule if you want.

semgrep-core/Analyzing/Bloom_annotation.ml
```diff
@@ -41,9 +44,44 @@ module B = Bloom_filter
...
 (*****************************************************************************)
 
+(* Note: If the bottom most node has n strings and k parents, it will take
+ * O(kn) to add it to all the lists. Since it will also take O(kn) to add it
+ * to all the bloom filters, this is acceptable, but it would not be
+ * necessary if we used a linked list
+*)
+
+let push v (l : 'a list ref) = Common.push v l
+;;
```
aryx: you don't need those ;;
They are need only in ocamltop (the OCaml REPL).

semgrep-core/Analyzing/Bloom_annotation.ml
```diff
@@ -53,23 +91,94 @@ module B = Bloom_filter
...
+      )
+    );
+    V.kconstness = (fun (k, _) x ->
+      (match x with
+       | Lit (String (str, _tok)) ->
+           if not (special_literal str) then push str res
+       | _ -> k x
+      )
+    );
+    V.kid_info = (fun (k, _) x -> k x);
```
aryx, line 118: maybe a comment here reminding that the default behavior of kid_info is to not call the continuation? which
is why we redefine here to call the continuation so we can also index propagated constants inside id_constness
(that would not be indexed by the default visitor).

semgrep-core/Analyzing/Bloom_annotation.ml
```diff
@@ -53,23 +91,94 @@ module B = Bloom_filter
...
+    V.kstmt = (fun (k, _) x ->
+      (* First statement visited is the current statement *)
+      if !top_level then (top_level := false; k x)
+      else
+        (* For any other statement, recurse to add the filter *)
+        (let strs = statement_strings x in
```
aryx: I usually prefer begin/end rather than () to mark a block of instructions.

semgrep-core/Analyzing/Bloom_annotation.ml
```diff
@@ -53,23 +91,94 @@ module B = Bloom_filter
...
+      (match x with
+       | Lit (String (str, _tok)) ->
+           if not (special_literal str) then push str res
+       | _ -> k x
+      )
+    );
+    V.kid_info = (fun (k, _) x -> k x);
+    V.kstmt = (fun (k, _) x ->
+      (* First statement visited is the current statement *)
+      if !top_level then (top_level := false; k x)
```
aryx: why do we do something special for the top statement?
why we don't set its bloom filter too?

semgrep-core/Analyzing/Bloom_annotation.ml
```diff
@@ -53,23 +91,94 @@ module B = Bloom_filter
...
 
-let bloom_of_expr _e =
-  B.create ()
-
-let bloom_of_stmt _st =
-  B.create ()
+let list_of_pattern_strings any =
+  let res = ref [] in
+  let visitor = V.mk_visitor {
+    V.default_visitor with
```
aryx, line 143: looks like duplicated code with above. Maybe you can factorize.

semgrep-core/Analyzing/Bloom_annotation.ml
```diff
@@ -53,23 +91,94 @@ module B = Bloom_filter
...
 
 (* TODO: visit AST and set the s_bf field in statements *)
-let annotate_program _ast =
-  ()
+let annotate_program ast =
+  let visitor = V.mk_visitor {
+    V.default_visitor with
+    V.kstmt = (fun (_k, _) x ->
+      let bf = B.create () in
+      let ids = statement_strings x in
```
aryx, line 174: This seems like duplicated with the statement_strings code, why not factorize? (which would also
make useless the is_toplevel variable).

semgrep-core/Core/AST/Bloom_filter.ml
```diff
@@ -15,29 +16,45 @@
...
 (* API entry points *)
 (*****************************************************************************)
 
+let pp _formatter _bf = ()
+;;
+
 let create () =
-  { added = false }
+  { added = false;
+    filter = B.create ~error_rate:0.01 10 }
```
aryx, line 38: maybe some quick explanation of what means the parameter and what is the intuition behind  those numbers?

aryx (changes requested):

semgrep-core/Analyzing/Bloom_annotation.ml
emjin: This is because of how I visit so that I can get all the strings in a statement (including its substatements). I use the visitor until I come to a statement S. Then, I stop that visitor, and recursively call the statement_strings function, which starts a ref for a list of strings and creates a new visitor for S. However, S was already visited by the previous visitor, so it should not be revisited. I couldn't quite figure out how to do this with only one visitor call, though I'm afraid this will use more memory.

semgrep-core/Analyzing/Bloom_annotation.ml
emjin: I wasn't sure what the best way to do this was. The code is very similar in some ways, but statement_strings is a recursive function, called recursively at each statement node, whereas list_of_pattern_strings visits the entire AST. I could make functions to produce kident and kexpr which take in a ref, I just wasn't sure if it would end up making the code less readable. If I start making kexpr more complicated, I'll factorize it.

semgrep-core/Matching/Generic_vs_generic.ml
```diff
@@ -1643,143 +1654,152 @@ and m_list__m_stmt_uncached ~flattened (xsa: A.stmt list) (xsb: A.stmt list) =
...
-        ))
-
-  | A.ExprStmt _, _  | A.DefStmt _, _  | A.DirectiveStmt _, _
-  | A.Block _, _  | A.If _, _  | A.While _, _  | A.DoWhile _, _  | A.For _, _
-  | A.Switch _, _  | A.Return _, _  | A.Continue _, _  | A.Break _, _
-  | A.Label _, _  | A.Goto _, _  | A.Throw _, _  | A.Try _, _
-  | A.Assert _, _
-  | A.OtherStmt _, _ | A.OtherStmtWithStmt _, _ | A.WithUsingResource _, _
-    -> fail ()
+  let pattern_strs = Bloom_annotation.list_of_pattern_strings (S a) in
```
aryx: No! I don't think it's a good idea to do this each time in m_stmt.
Either we also pre-compute also on the pattern the bloom filter, and set it in the s_bf field of the pattern,
or we leave the logic outside Generic_vs_generic. Generic_vs_generic is already quite complicated
so I would really prefer to have the optimizations done outside, e.g., in Semgrep_generic.ml as it was before.

aryx (changes requested):

semgrep-core/Analyzing/Bloom_annotation.ml
aryx: ok, sounds fair.

semgrep-core/Matching/Generic_vs_generic.ml
emjin: This was an experiment---I was planning on asking you for advice on Monday, but I'm not getting any speedup on stmt lists because it doesn't check the pattern against a bloom filter first. See my reply to Martin in eng.

semgrep-core/Analyzing/Bloom_annotation.ml
emjin: See earlier comment (the need to union all the child statements' strings and add it to the statement's bloom filter). There may be a better way of doing this (perhaps passing in a ref list of bloom filters?), but I'm not sure

emjin:
> I like the semgrep check. It's easier to read, but I let @emjin decide if she thinks it's a useless rule.
> You can update the semgrep-rules repo to remove the rule if you want.

I left the rule and changed my code, but there's probably a way to do what I want with regexs (precede with (...)*?)

aryx:
Why the CI failures? Can you fix that so we can merge it after?

semgrep-core/CLI/Main.ml
```diff
@@ -1288,6 +1288,8 @@ let options () =
     " <dir> save and use parsed ASTs in a cache at given directory. Caller responsiblity to clear cache";
     "-filter_irrelevant_rules", Arg.Set Flag.filter_irrelevant_rules,
     " filter rules not containing any strings in target file";
+    "-bloom_filter", Arg.Set Flag.use_bloom_filter,
+    "-no_bloom_filter";
```
aryx, line 1292: This seems wrong, this string is supposed to comment what -bloom_filter means.

aryx (approved):

PR #2468, opened 2021-01-27
Run again unit tests that were in pfff but now in semgrep

aryx:
test plan:
make test


aryx:
cc @IagoAbal and @emjin some of your test files in pfff moved now in semgrep-core/tests/OTHER/

Issue #2429, opened 2021-01-21
internal: Experiment with using real rules in semgrep-core


aryx:

aryx:
```
#pros to rewrite (or start move main stuff) in ocaml:
# - DONE holistic rules checker
# - holistic optimisations, see pattern: $PROP issues, order 
#   of calling the engine matters! "query rewrite optimizations"?
# - DONE JSON/jsonnet experiment
# - DONE clean formula rewrite (still nobody did it after months)
#   https://github.com/returntocorp/semgrep/issues/1489
# - lots of ugly corner cases with pattern-not and metavar, cleaner in
#   typed world maybe to fix all of that (look recent diff by grayson)
# - python setup/install mess, just one binary after! (can even link
#   spacegrep in)
# - even eval we now do it ocaml side, and for spacegrep we could link
#   directly with the spacegrep ocaml library
# - each time we need to duplicate feature on python and ocaml side
#   (e.g., profiling)
# - finalize equivalences, without duplication of work with 2 sides
# - finalizing tainting, without duplication of work with 2 sides
# - easier switch to list of ranges for core matches, no need modify
#   2 sides, refactoring in same (typed!) program
# - just 6000 LOC to port, and anyway like I said lots of duplicated/redundant
#   structures also defined in OCaml anyway
# - perf win? no need those intermediate AST caching?
# - road for semgrep config file! policy 2.0! with jsonnet easier!
# - most R2C non-ocaml developers migrated to work on semgrep-app or
#   semgrep-agent/action anyway, even brendon is busy with kubernetes and matt
#   and grayson with security work

#cons if decide for full rewrite, need to port many things:
# - network access for config and flexible way to refer to rules/config
# - file set spec (but I used to have that in ocaml side too
# - .gitignore.
# - autofix
# - pattern-regexp
# - jsonschema check
# - sarif, other external-contributor Python contributions?
#   but we have arguably more ocaml contribs than python contrib anyway
# - ???
```

PR #2471, opened 2021-01-27
Improve full rule handling in semgrep-core

aryx:


aryx:
This continues work on https://github.com/returntocorp/semgrep/issues/2429

PR #2482, opened 2021-02-01
Switch to Pcre for the ocaml regexp parsing library

aryx:
The Re OCaml library does not support negative lookup assertion
regexps like (?!foo) which are used in semgrep-rules.

test plan:
test file included
make test

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@
 
 FROM returntocorp/ocaml:alpine-2020-10-28 as build-semgrep-core
 
+USER root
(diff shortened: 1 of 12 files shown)


aryx:
See also https://stackoverflow.com/questions/49305881/excluding-a-specific-word-in-ocaml-regex

aryx:
At some point we should switch the handling of "=~/.../" in semgrep to also rely on this PCRE regexp library,
and also to use the `ANCHORED flag to behave like Python re.match(), to be more consistent with the other
use of regexps in semgrep rule file.

PR #2490, opened 2021-02-02
Basic support for pattern-regex in semgrep-full-rule-in-ocaml

aryx:
test plan:
$ /home/pad/semgrep/_build/default/CLI/Main.exe -lang python -config regexp.jsonnet regexp.py
regexp.py:3
     foo("128.0.0.1")

### DIFF
--- a/semgrep-core/Core/Rule.ml
+++ b/semgrep-core/Core/Rule.ml
@@ -44,7 +44,7 @@ type xlang =
 
 type regexp = string * Pcre.regexp
 let pp_regexp fmt (s, _) =
-  Format.fprintf fmt "%s" s
+  Format.fprintf fmt "\"%s\"" s
 let equal_regexp (s1, _) (s2, _) = s1 = s2
 
 type xpattern = {
--- a/semgrep-core/Engine/Semgrep.ml
+++ b/semgrep-core/Engine/Semgrep.ml
@@ -21,6 +21,7 @@ module R = Rule
 module MR = Mini_rule
 module PM = Pattern_match
 module G = AST_generic
+module PI = Parse_info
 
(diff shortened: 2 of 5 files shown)


semgrep-core/Core/Rule.ml
```diff
@@ -44,7 +44,7 @@ type xlang =
 
 type regexp = string * Pcre.regexp
 let pp_regexp fmt (s, _) =
-  Format.fprintf fmt "%s" s
+  Format.fprintf fmt "\"%s\"" s
```
mjambon, line 47: I usually use `%S`, which prints a double-quoted ocaml string literal. It escapes a bunch of characters too in addition to escaping `"`.
```
Format.fprintf fmt "%S" s
```

semgrep-core/Engine/Semgrep.ml
```diff
@@ -130,11 +135,34 @@ let (mini_rule_of_pattern: R.t -> (R.pattern_id * Pattern.t) -> MR.t) =
     languages =
       (match r.R.languages with
        | R.L (x, xs) -> x::xs
-       | R.LNone | R.LGeneric -> failwith "TODO: LNone | LGeneric"
+       | R.LNone | R.LGeneric -> raise Impossible
```
mjambon, line 138: For this, I use `assert false`. One advantage is that the exception contains the source location, even if no stack trace is collected or printed. Another advantage is that it's standard, see https://caml.inria.fr/pub/docs/manual-ocaml/expr.html#sss:expr-assertion

semgrep-core/Core/Rule.ml
aryx: Good to know, will fix.

PR #2501, opened 2021-02-04
Caching improvements, making it usable

mjambon:
The previous implementation of caching was incorrectly caching the results of matching flattened lists of statements (for deep ellipsis matching). It's no longer the case, each flattened list is now identified for caching purposes by the ID of the its first and last elements.

The code was also simplified a little by computing the flattened list eagerly. This allows us to know the last element of the list, which we use in the cache key. I didn't see a performance difference in my tests but there might be one. Maybe some day we can revise this part of the code so as to avoid doing this flattening over and over again.

There are now 3 levels of caching:
- no caching
- moderate caching (default): only consult the cache if the current set of captures referenced later is empty
- maximum caching: doesn't care if the rest of the pattern contains backreferences, which improves performance in some cases and makes it worse in others

Finally, I added and renamed performance tests for completeness and consistency. Because test cases have been renamed, I disable caching by default. We'll first merge this branch, wait for benchmarks to be published, and only then I'll flip the switch to enable caching by default so as to see the performance change on the graphs.

`make perf` gives me the following results (with default = moderate caching):
```
$ make perf
--- default settings ---
cd ./perf && ./run-perf-suite
empty: 0.127337 s
stmts-simple: 1.328697 s
stmts-ellipsis-ab: 2.304741 s
stmts-ellipsis-aa: 9.155018 s
stmts-double-ellipsis-abc: 0.290724 s
stmts-double-ellipsis-abb: 0.442996 s
stmts-double-ellipsis-aab: 0.370489 s
stmts-double-ellipsis-aba: 4.933527 s
stmts-double-ellipsis-aaa: 0.404022 s
--- no caching ---
cd ./perf && ./run-perf-suite --no-cache
empty: 0.103258 s
stmts-simple: 1.316770 s
stmts-ellipsis-ab: 9.150210 s
stmts-ellipsis-aa: 9.652244 s
stmts-double-ellipsis-abc: 4.733505 s
stmts-double-ellipsis-abb: 4.766145 s
stmts-double-ellipsis-aab: 0.395519 s
stmts-double-ellipsis-aba: 4.891320 s
stmts-double-ellipsis-aaa: 0.387988 s
--- maximum caching ---
cd ./perf && ./run-perf-suite --max-cache
empty: 0.104390 s
stmts-simple: 1.319819 s
stmts-ellipsis-ab: 2.106228 s
stmts-ellipsis-aa: 11.948184 s
stmts-double-ellipsis-abc: 0.291730 s
stmts-double-ellipsis-abb: 0.630560 s
stmts-double-ellipsis-aab: 0.505430 s
stmts-double-ellipsis-aba: 0.850217 s
stmts-double-ellipsis-aaa: 0.486438 s
```


semgrep-core/Core/Caching.ml
```diff
@@ -1,8 +1,97 @@
...
+   Due to ellipsis patterns ('...'), some sequences of statements end
+   up being matched against the same pattern repeatedly. Memoization
+   is any technique consisting in caching the results of a function call.
+   We use memoization on the function(s) that match a sub-pattern against
+   a sub-sequence of statements in the target AST.
+
+   The goal is to make matching faster. Memory usage is assumed to be
+   less problematic.
+
+   Below is a list of sample patterns or targets, in semgrep-js syntax,
```
aryx, line 13: This is great documentation.

aryx (approved):

aryx:
Can you fix the conflict? (I think it's due to Emma recent changes)?

semgrep-core/Core/Caching.ml
emjin: +1

semgrep-core/Core/Caching.ml
```diff
@@ -71,6 +168,22 @@ let diff_backrefs bound_metavars ~new_backref_counts ~old_backref_counts =
    This function decorates the pattern AST with the set of metavariables
    that are used in the rest of the pattern. This set is consulted at runtime
    to determine whether a captured metavariable should go into the cache key.
+
+   We only consult the cache at strategic locations:
+   - on an ellipsis pattern, for every position in the target stmt list,
+   - and only if the rest of the pattern doesn't contain free metavariables.
```
emjin: Is this supposed to say "bound metavariables"? (Or am I confused about something)

emjin (approved):

semgrep-core/Core/Caching.ml
mjambon: I'll clarify.

mjambon: I think my use of the terms "free" and "bound" is just wrong.

mjambon: Anyway, the notion is that `let x = 1 in x + 1` is not parametrized whereas `x + 1` alone is parametrized by `x`.

mjambon: Changing it to: "and only if the rest of the pattern is not parametrized by earlier captures"

PR #2508, opened 2021-02-05
Add a TyInterfaceAnon type to the generic AST

### LINKED ISSUE #2376: Go Pattern Matching With ([]interface{}) Type Assertion
**Describe the bug**

[Interface types in Go](https://golang.org/ref/spec#Interface_types) can be type asserted to any valid type at compile time (though they may panic at runtime). An example use case for this is an abstract data mapping function that returns an interface type whose result is later parsed into primitive Go types. It is unclear in this case whether there is a possible pattern to match this code scenario or if there is a potential gap in Go syntax handling.

**To Reproduce**

Steps to reproduce the behavior, ideally a link to https://semgrep.dev: https://semgrep.dev/s/qx68/

Code:

```go
package main

func main() {
    var v interface{}
    _ = v.([]interface{})
}
```

Configuration:

```yaml
rules:
- id: go-interface-type-assertion
  pattern: |
    $VALUE.([]interface{})
  message: |
    Match Go interface type assertion
  severity: WARNING
```

Result:

```
Pattern could not be parsed as a Golang semgrep pattern
```

Also errors if changing the pattern to `v.([]interface{})`.

**Expected behavior**

Match on `_ = v.([]interface{})` line.

P2

**Environment**

Discovered locally on 0.36.0, verified on semgrep.dev.

emjin:
Closes #2376
    
Explicitly handle interface types in go patterns by directly translating interface types into a generic AST interface type value.
    
### DIFF
--- a/semgrep-core/Core/AST/AST_generic.ml
+++ b/semgrep-core/Core/AST/AST_generic.ml
@@ -1163,6 +1163,7 @@ and type_ =
    * via a TyName. Here we have flexible record types (a.k.a. rows in OCaml).
   *)
   | TyRecordAnon of tok (* 'struct/shape', fake in other *)* field list bracket
+  | TyInterfaceAnon of tok (* 'struct/shape', fake in other *)* field list bracket
 
   (* sgrep-ext: *)
   | TyEllipsis of tok
--- a/semgrep-core/Parsing/pfff/go_to_generic.ml
+++ b/semgrep-core/Parsing/pfff/go_to_generic.ml
@@ -144,14 +144,7 @@ let top_func () =
 
     | TStruct (t, v1) -> let v1 = bracket (list struct_field) v1 in
         G.TyRecordAnon (t, v1)
-    | TInterface (t, v1) -> let v1 = bracket (list interface_field) v1 in
-        let s = gensym () in
-        let ent = G.basic_entity (s, t) [] in
-        let def = G.ClassDef { G.ckind = (G.Interface, t);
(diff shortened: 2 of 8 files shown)


semgrep-core/Core/AST/AST_generic.ml
```diff
@@ -1163,6 +1163,7 @@ and type_ =
    * via a TyName. Here we have flexible record types (a.k.a. rows in OCaml).
   *)
   | TyRecordAnon of tok (* 'struct/shape', fake in other *)* field list bracket
+  | TyInterfaceAnon of tok (* 'struct/shape', fake in other *)* field list bracket
```
aryx, line 1166: the (cryptic) comment applies only to TyRecordAnon.

semgrep-core/Parsing/pfff/go_to_generic.ml
```diff
@@ -599,11 +586,8 @@ let top_func () =
         G.ImportAll (itok, module_name, v1)
 
   and program xs =
-    anon_types := [];
-    let xs = list top_decl xs in
-    let arg_types = !anon_types |> List.map (fun x -> G.DefStmt x |> G.s) in
-    (* TODO? put after program and imports? *)
-    arg_types @ xs
+    list top_decl xs
+  (* TODO? put after program and imports? *)
```
aryx, line 590: can remove this comment now

aryx (approved):
Great!

Issue #2507, opened 2021-02-05
Kotlin support -lang kt and -lang kotlin


colleend:
**Describe the bug**
-lang kt and -lang kotlin from the semgrep repository currently result in 'language not found' errors.

aryx:
what does "from the semgrep repository" means in this context?
What is the command-line that fails?

SPoint42:
This 2 lines : 

`docker run -v ${PWD}:/src -v c:\semgrep-rules\Level1\Kotlin\:/rules  returntocorp/semgrep /src --config /rules/ --lang kt`
`docker run -v ${PWD}:/src -v c:\semgrep-rules\Level1\Kotlin\:/rules  returntocorp/semgrep /src --config /rules/ --lang kotlin`
returns : 

`unsupported language: kt. supported languages are: C#, JSON, Json, c, cs, csharp, generic, go, golang, java, javascript, js, json, jsx, lua, ml, none, ocaml, php, py, python, python2, python3, rb, regex, ruby, ts, tsx, typescript`

or `unsupported language: kotlin. supported languages are: C#, JSON, Json, c, cs, csharp, generic, go, golang, java, javascript, js, json, jsx, lua, ml, none, ocaml, php, py, python, python2, python3, rb, regex, ruby, ts, tsx, typescript`

with 


```rules:
  - id: kotlin-path-traversal
    patterns:
      - pattern-either:
          - pattern: |
              createTempDir($VAR1,$VAR2)
          - pattern: |
              createTempFile($VAR1,$VAR2)
          - pattern: |
              createTempDir($VAR1,$VAR2, File("..."))
          - pattern: |
              createTempFile($VAR1,$VAR2, File("..."))
    languages:
      - kt
    message:
      Untrusted user input in createTempDir()/createTempFile() can endup in Directory Traversal Attacks.
    severity: ERROR
    metadata:
      owasp: 'A5: Broken Access Control'
      cwe: 'CWE-23: Relative Path Traversal
```

or with
 
```rules:
  - id: kotlin-path-traversal
    patterns:
      - pattern-either:
          - pattern: |
              createTempDir($VAR1,$VAR2)
          - pattern: |
              createTempFile($VAR1,$VAR2)
          - pattern: |
              createTempDir($VAR1,$VAR2, File("..."))
          - pattern: |
              createTempFile($VAR1,$VAR2, File("..."))
    languages:
      - kotlin
    message:
[...]

PR #2511, opened 2021-02-06
add kotlin to extensions

colleend:


DrewDennison (approved):

aryx:
@colleend the CI is failing because of a end-to-end test that assert the shape of the error message about languages.
See my comment in the file:
# coupling: if you add a constant here, modify also ALL_EXTENSIONS below
# and you probably also need to update _LANGS_TO_EXTS_INTERNAL
# You may also have to regenerate some test snapshots with
# pipenv run pytest tests/e2e/test_rule_parser.py --snapshot-update

aryx:
This closes https://github.com/returntocorp/semgrep/issues/2507

aryx (approved):

PR #2517, opened 2021-02-08
Remove Python 3.7 dev requirement

mschwager:
Let's see if this passes CI :crossed_fingers:

### DIFF
--- a/semgrep/Pipfile
+++ b/semgrep/Pipfile
@@ -14,6 +14,3 @@ xmldiff = "*"
 
 [packages]
 semgrep = {editable = true,path = "."}
-
-[requires]
-python_version = "3.7"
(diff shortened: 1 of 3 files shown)


semgrep/Pipfile
```diff
@@ -14,6 +14,3 @@ xmldiff = "*"
 
 [packages]
 semgrep = {editable = true,path = "."}
-
-[requires]
-python_version = "3.7"
```
aryx, line 19: or can we require > 3.6 or something?

aryx (approved):

semgrep/Pipfile
mschwager: I was just looking into this. Apparently the answer is "no": https://github.com/pypa/pipenv/issues/1050 :-1:

PR #2525, opened 2021-02-09
Fix 'make setup', which was not updating the tree-sitter runtime library

mjambon:
if it had been previously checked out.

Follow-up to https://github.com/returntocorp/ocaml-tree-sitter/pull/143


mjambon:
~~oops, I updated another submodule by accident.~~

update: fixed

aryx (approved):

PR #2529, opened 2021-02-10
Upgrade typescript grammar to version 'for-js-patched-3f8b62f'

mjambon:
https://github.com/returntocorp/tree-sitter-typescript/tree/for-js-patched-3f8b62f


mjambon:
This is where we have a parsing error:
https://github.com/zulip/zulip/blob/90636d5e81e7aa0ce734d1105806117db060d7a6/tools/debug-require-webpack-plugin.ts#L61
image

aryx:
Can you  fix the CI or should we drop this PR?

mjambon:
I added support for the latest commit by Max, which fixes the regression on tuple types that we had observed. Our stats indicates that there are other regressions, but let's do one thing at a time.

These latest changes come with support for tuple type members of the form `foo?` or `...foo` which are [new in typescript 4.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html). This required changing the type of tuple members in the generic AST and in the JS AST in pfff.

semgrep-core/Core/AST/AST_generic.ml
```diff
@@ -1226,6 +1226,11 @@ and other_type_argument_operator =
   | OTA_Todo
   (*e: type [[AST_generic.other_type_argument_operator]] *)
 
+and tuple_type_member =
```
aryx: hmm nah. You can use TyQuestion for TyTupOpt
and we can define a new constructor for TyRest.
No need to change all those xxx_to_generic.ml just because Typescript does thing differently.

aryx (changes requested):

semgrep-core/Core/AST/AST_generic.ml
mjambon: I didn't realize there was already a TyQuestion in the generic AST when I wrote my comment on the pfff pull request earlier. https://github.com/returntocorp/pfff/pull/394#discussion_r576309641

Let me summarize what needs to change:
* Use plain `type_` for tuple type members, remove the following from AST_generic, :
```
and tuple_type_member =
  | TyTupMember of type_  (* simple tuple type element *)
  | TyTupOpt of type_  (* optional tuple type element in typescript: string? *)
  | TyTupRest of type_  (* rest tuple type element in typescript: ...string *)
```
* Add `TyRest` to `type_` in the generic AST:
```
and type_ =
...
  | TyPointer of tok * type_
  | TyRef of tok * type_ (* C++/Rust *)

  | TyQuestion of type_ * tok (* a.k.a option type *)
  | TyRest of tok * type_ (* '...foo' e.g. in a typescript tuple type *)
  ^^^^^^^^^^^^^^^^^^^^^

  (* intersection types, used for Java Cast, and in Typescript *)
  | TyAnd of type_ * tok (* & *) * type_
...
```
* The order of tuple elements in the generic AST follows the concrete syntax in mainstream languages, that's why we have `type_ * tok` for `foo?` but `tok * type_` for `...foo`.

mjambon: The mapping from JS to generic AST is:
```ocaml
and tuple_type_member x =
  match x with
  | TyTupMember x -> type_ x
  | TyTupOpt (x, tok) -> TyQuestion (type_ x, tok)
  | TyTupRest (tok, x) -> TyRest (tok, type_ x)
```

aryx:
you nee to resolve the conflict. I think you need to merge develop and pin to the latest pfff.

semgrep-core/Parsing/pfff/js_to_generic.ml
```diff
@@ -405,6 +405,11 @@ and type_ x =
   | TypeTodo (categ, xs) ->
       G.OtherType (G.OT_Todo, (G.TodoK categ)::(List.map any xs))
 
+and tuple_type_member x =
+  match x with
+  | TyTupMember x -> type_ x
+  | TyTupOpt (x, tok) -> TyQuestion (type_ x, tok)
```
aryx, line 410: Supposed to be G.TyQuestion

aryx (approved):

mjambon:
> you nee to resolve the conflict. I think you need to merge develop and pin to the latest pfff.

Semgrep doesn't build with the latest pfff. This branch is now pointing to an earlier version of pfff + my changes --> https://github.com/returntocorp/pfff/tree/compat

aryx (approved):

PR #2531, opened 2021-02-10
Update Ruby parser from latest tree-sitter-ruby

mjambon:


aryx (approved):
Wow that's lots of changes. Good job martin.
Also your ocaml-tree-sitter-lang split was really the good idea; so good to update
parser independently.

PR #2538, opened 2021-02-10
analyzing: constness: Set function inputs as non-constant

IagoAbal:
Let's consider the following example:

    def update_system(password):
        if cond():
            password = "abc"
        # OK: it's not guaranteed to be "abc"!
        set_password(password)
    
Here `password` is *not* a constant, because there is a path where its
value comes from an input. But for this to work as expected we must tell
the analysis that function inputs are non-constant!

test plan:
make test


aryx (approved):
Looks good.

IagoAbal:
Thanks! I updated the commit message, the example I used there actually works because `eval` handles it well. So I just use the same example as the test.

Issue #975, opened 2020-06-11
Import inconsistencies between languages


mschwager:
```
$ semgrep --version
0.10.1
```

Consider the following Python code:

```
$ cat test.py 
import os.path.join
```

When using a pattern looking for imports we get the following results:

```
$ semgrep --json --lang python --pattern 'import $X' test.py | python -mjson.tool
{
    "errors": [],
    "results": [
        {
            "check_id": "-",
            "end": {
                "col": 20,
                "line": 1
            },
            "extra": {
                "lines": "import os.path.join",
                "message": "import os",
                "metadata": {},
                "metavars": {
                    "$X": {
                        "abstract_content": "os",
                        "end": {
                            "col": 10,
                            "line": 1,
                            "offset": 9
                        },
                        "start": {
                            "col": 8,
                            "line": 1,
                            "offset": 7
                        },
                        "unique_id": {
                            "md5sum": "aa9320c62d85df5585dec78521910201",
                            "type": "AST"
                        }
                    }
                },
                "severity": "ERROR"
            },
            "path": "test.py",
            "start": {
                "col": 1,
                "line": 1
            }
        }
    ]
}
```

Here the `$X` metavariable content is `os`. This doesn't feel correct looking at the code, I'd expect it to be `os.path.join`. An argument could be made for this behavior, but it doesn't even seem consistent across languages. For example, consider the following Java code:

```
$ cat test.java 
import deep.imp.statement;
```

Running a similar import check gives the following results:

```
$ semgrep --json --lang java --pattern 'import $X;' test.java | python -mjson.tool
{
    "errors": [],
    "results": [
        {
            "check_id": "-",
            "end": {
                "col": 26,
                "line": 1
            },
            "extra": {
                "lines": "import deep.imp.statement;",
                "message": "import statement;",
                "metadata": {},
                "metavars": {
                    "$X": {
                        "abstract_content": "statement",
                        "end": {
                            "col": 26,
                            "line": 1,
                            "offset": 25
                        },
                        "start": {
                            "col": 17,
                            "line": 1,
                            "offset": 16
                        },
                        "unique_id": {
                            "md5sum": "876b2cf6cf7324525b2283402c5df855",
                            "type": "AST"
                        }
                    }
                },
                "severity": "ERROR"
            },
            "path": "test.java",
            "start": {
                "col": 1,
                "line": 1
            }
        }
    ]
}
```

[...]

PR #2541, opened 2021-02-10
[Java] import $LIB should binds $LIB to a full import path

### LINKED ISSUE #2502: import $VAR metavariable does not match entire import path
**Describe the bug**
when typing the pattern ```import $VAR```, the metavariable $VAR matches only the last portion of the import path, not the entire import path. 

**To Reproduce**
Here's an example:
https://semgrep.dev/s/oxN6/

**Expected behavior**
Would expect to match the entire path. In the example given, I would expect the pattern to match both of the example imports.

P1 -- would like to have this fixed so that we can write some rules for snowflake 👍 

P.S: this is @colleend with snowflake account 😎

aryx:
This closes https://github.com/returntocorp/semgrep/issues/2502

test plan:
See how below abstract_content now contains the full path, not just
the last BouncyCastleProvider element.

$ /home/pad/semgrep/_build/default/CLI/Main.exe -lang java -f metavar_name.sgrep metavar_name.java -json
{"matches":[{"check_id":"-e/-f","path":"metavar_name.java","start":{"line":2,"col":1,"offset":15},"end":{"line":2,"col":61,"offset":75},"extra":{"message":"","metavars":{"$LIB":{"start":{"line":2,"col":8,"offset":22},"end":{"line":2,"col":61,"offset":75},"abstract_content":"org bouncycastle crypto provider BouncyCastleProvider","unique_id":{"type":"AST","md5sum":"8e0da5e5f63c8f7baf9ce1b42b240f90"}}}}}],"errors":[],"stats":{"okfiles":1,"errorfiles":0}}

### DIFF
--- a/semgrep-core/Core/Metavariable.ml
+++ b/semgrep-core/Core/Metavariable.ml
@@ -53,7 +53,9 @@ type mvar = string
  * use of AST_generic.any for patterns.
 *)
 type mvalue =
+  (* TODO: get rid of Id, N generalize it *)
   | Id of AST_generic.ident * AST_generic.id_info option
+  | N of AST_generic.name
   | E of AST_generic.expr
   | S of AST_generic.stmt
   | Ss of AST_generic.stmt list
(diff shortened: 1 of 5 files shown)


aryx:
This should also help https://github.com/returntocorp/semgrep/issues/975

PR #2542, opened 2021-02-10
Just add some .mli in spacegrep

aryx:
As I was trying to integrate spacegrep in my new
full-rules-in-ocaml engine (see semgrep-core/Engine/Semgrep.ml),
I was missing a few .mli files.

test plan:
make


mjambon:
Please, as a habit, include a headline that summarizes the purpose of the file (why it has to exist), for each source file. Copying it from the ml file is usually fine.

aryx:
yep, you're right

PR #2557, opened 2021-02-12
Add section about extending language support

mjambon:
This adds a section about extending language support to `CONTRIBUTING.md`.

Reviewers: take a look at the list of languages and let me know if you see some that are misclassified.

(I also need to update ocaml-tree-sitter)


aryx (approved):

colleend:
this is awesome! 🙏

PR #2562, opened 2021-02-12
Improve --generate-config capabilities

mschwager:
This came out of a conversation with the folks over at Gitlab.

Let's say you're hacking on a particular pattern:

```
$ semgrep --lang python --pattern '$X == $X' ...
```

Wouldn't it be nice if you could quickly turn this into a configuration file and continue hacking from there?

```
$ semgrep --lang python --pattern 'hashlib.md5(...)' -g
Template config successfully written to .semgrep.yml
$ cat .semgrep.yml 
rules:
- id: eqeq-is-bad
  pattern: hashlib.md5(...)
  message: $X == $X is a useless equality check
  languages:
  - python
  severity: ERROR
```

Okay, the `message` is off, but you can take it from there. Wouldn't it be nice if you could write to an arbitrary file?

```
$ semgrep --lang python --pattern 'hashlib.md5(...)' -g arbitrary.yml
Template config successfully written to arbitrary.yml
$ cat arbitrary.yml 
rules:
- id: eqeq-is-bad
  pattern: hashlib.md5(...)
  message: $X == $X is a useless equality check
  languages:
  - python
  severity: ERROR
```

This takes the existing `--generate-config` flag and makes it respect `--lang`, `--pattern`, and take an optional filename :+1:

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -51,13 +51,6 @@ def cli() -> None:
     config = parser.add_argument_group("config")
     config_ex = config.add_mutually_exclusive_group()
     config_ex.add_argument(
-        "-g",
-        "--generate-config",
-        action="store_true",
-        help=f"Generate starter configuration file, {DEFAULT_CONFIG_FILE}",
-    )
-
-    config_ex.add_argument(
         "-f",  # for backwards compatibility
         "-c",
         "--config",
@@ -69,11 +62,19 @@ def cli() -> None:
             "name. See https://semgrep.dev/docs/writing-rules/rule-syntax for information on configuration file format."
         ),
     )
-
     config_ex.add_argument(
         "-e",
         "--pattern",
         help="Code search pattern. See https://semgrep.dev/docs/writing-rules/pattern-syntax for information on pattern features.",
+    )
+    config.add_argument(
+        "-g",
+        "--generate-config",
+        action="store",
+        nargs="?",
+        const=DEFAULT_CONFIG_FILE,
+        type=argparse.FileType("x"),
+        help=f"Generate starter configuration file. Defaults to {DEFAULT_CONFIG_FILE}.",
     )
     config.add_argument(
         "-l",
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/cli.py
```diff
@@ -69,12 +62,20 @@ def cli() -> None:
...
         "-e",
         "--pattern",
         help="Code search pattern. See https://semgrep.dev/docs/writing-rules/pattern-syntax for information on pattern features.",
     )
+    config.add_argument(
+        "-g",
+        "--generate-config",
+        action="store",
+        nargs="?",
+        const=DEFAULT_CONFIG_FILE,
```
mschwager, line 75: Apparently `const` is used instead of `default` when `nargs="?"` and the optional argument isn't specified: https://docs.python.org/3/library/argparse.html#nargs

semgrep/semgrep/cli.py
```diff
@@ -69,12 +62,20 @@ def cli() -> None:
...
+        type=argparse.FileType("x"),
```
mschwager, line 76: The `"x"` file mode only allows writing if the file doesn't already exist. This covers the existing `if Path(DEFAULT_CONFIG_FILE).exists()` check.

semgrep/semgrep/config_resolver.py
```diff
@@ -457,34 +471,18 @@ def resolve_config(config_str: str) -> Dict[str, YamlTree]:
     return config
 
 
-def generate_config() -> None:
-    import requests  # here for faster startup times
+def generate_config(fd: IO, lang: Optional[str], pattern: Optional[str]) -> None:
+    config = DEFAULT_CONFIG
```
mschwager, line 475: I took out the existing get from `TEMPLATE_YAML_URL`, let me know if this is a bad idea. It looks like we rarely (never?) update that URL, and I figured having one less call here wasn't a bad thing: https://semgrep.dev/docs/faq/#what-network-requests-are-made.

brendongo (approved):
LGTM. Do you want to add a test?

ievans (approved):
+1 to @brendongo comment, a test would be nice

I love this change!

PR #2571, opened 2021-02-16
Recognize Go `:=' encoded as AssignOp(Eq)

### LINKED ISSUE #2440: go - constant propagation does not work with short variable assignment
It appears that Golang support for constant propagation does not cover `:=` short assignment statement. It works with `var` declaration though.

https://semgrep.dev/s/fproulx-boostsecurity:go-const-propag-broken

IagoAbal:
Closes #2440


IagoAbal:
For now I have avoided passing around the language in AST_to_IL, I would like to wait until there are more language specific rules in AST_to_IL. But if you think it is needed then we can do that already.

aryx (approved):

aryx:
yep I agree, better to wait we have more needs to pass the language.

Issue #1489, opened 2020-08-11
RFC: Simplify pattern matching API


nbrahms:
## Why

- Our current API has many different possible keys, and their interactions are hard to fully define
- Our current API doesn't meet well-known mental models
- Some functionality of our current API is broken (e.g. `pattern-inside` with `pattern-either`)

## Needs

- Minimize the number of API components of the matching API
- Ideally fit the API within a well-known mental model
- Create an easily maintainable and well-documented API

## Proposal

- [Boolean set logic](https://en.wikipedia.org/wiki/Boolean_algebra) is well understood and well researched; let's use it
- Make five API components available:
  - `pattern`: Specifies a single Semgrep pattern; value is always code text; result is the product of a range (a.k.a. a fully connected set of code byte offsets) and a set of metavariable matches
  - `regex`: Specifies a single regular expression; value is always text; result is the product of a range and an empty set of metavariable matches
  - `metavariable`: Specifies constraints on metavariable bindings; value is an object of constraint parameters; result is the product of the universal set and a set of metavariable constraints; must be used in a conjunction
  - `and`: Performs a set intersection of two or more matches; the intersection is non-empty iff all identically named metavariables match
  - `or`: Performs a set union of two or more matches; the returned set is a fully disjoint set of range-metavariable products; the union of two products with identical ranges but different metavariables is a set of two range-metavariable products
  - `not`: Takes the set complement of its single argument's match; value is always `and`, `or`, or `pattern`
- For ease of transition, all components must be nested under a top-level `match` key in the rule YAML
- When compiling the rule, DeMorgan's and Distributive Laws are applied to simplify the match; the resulting match will be a simple union of intersections, each of which must include one non-complement pattern (see below)

### Examples

- A single match:
```yaml
match:
  pattern: foo(...)
```

- Match all except an edge case:
```yaml
match:
  and:
    - pattern: foo(...)
    - not:
        pattern: foo(..., safe=True)
```

- Multiple possible taint avenues:
```yaml
match:
  and:
    - or:
      - pattern: |
           def $FUNC(...):
             ...
             $R = requests.$METHOD(...)
             ...
      - pattern: |
           def $FUNC(...):
             ...
             $R = requests.$METHOD.get(...)
             ...
    - pattern: |
           $DATA = $R.get("...", ...)
           ...
    - or:
      - pattern: some_sink($DATA)
      - pattern: some_other_sink($DATA)
```

- Complement ranges are disallowed:
```yaml
# INVALID
[...]

PR #2572, opened 2021-02-17
Change pattern-not-regex to filter when regex overlaps with a match

### LINKED ISSUE #2465: pattern-not-regex does not properly filter out matches
**Describe the bug**
`pattern-not-regex` does not correctly filter out matches that I expect it would.

**To Reproduce**
https://semgrep.dev/s/clintgibler:pattern-not-regex-bug.

**Expected behavior**
I would expect this rule to match the first line, but **not** the latter two. However, all three lines are matched.

```yaml
    patterns:
      - pattern-regex: "Foo"
      - pattern-not-regex: "Foo-"
      - pattern-not-regex: "-Foo"
```

```
Foo==1.1.1

Foo-Bar==3.0.8
Bar-Foo==3.0.8
```

P1

minusworld:
This changes the behavior from filtering when a match contains the regex to filtering when a regex overlaps with a match in any way.

Closes https://github.com/returntocorp/semgrep/issues/2465.


mschwager:
It looks like `pattern-regex` behaves like `pattern` while `pattern-not-regex` behaves like `pattern-not-inside`. Is this intended? I'm having a hard time understanding what `pattern-not-regex` is supposed to find. Looking at something like `pattern-not-regex: '[Xx]'`, does this mean anything that's not `[Xx]`? If so that'd be every non-`Xx` character in the code. I kinda confused on what the bounds of `pattern-not-regex` are supposed to be :confused:

aryx:
Note that in the new RFC for boolean composition of @nbrahms at https://github.com/returntocorp/semgrep/issues/1489#issuecomment-778368664 there is more patern-xxx-inside,
just And, Or, Not, and I think negation is indeed removing anything that overlaps with a "positive" match.

aryx (approved):

PR #2574, opened 2021-02-17
Allow 'nosem' in HTML.

### LINKED ISSUE #2566: Per line ignore for HTML files
Currently, it is not possible to ignore lines for HTML files. I've used exclude to ignore the file entirely.

Ability to ignore lines with nosemgrep comments like other languages.

minusworld:
Closes https://github.com/returntocorp/semgrep/issues/2566.

Changes spacegrep slightly to return the whole line for a match.

Filters out IDs from nosem strings that are not alphumeric, underscores, or dashes. This keeps the current nosem regex intact while allowing trailing characters such as required by HTML comments: `-->`.

### DIFF
--- a/semgrep/semgrep/semgrep_main.py
+++ b/semgrep/semgrep/semgrep_main.py
@@ -105,15 +105,13 @@ def rule_match_nosem(rule_match: RuleMatch, strict: bool) -> bool:
         )
         return True
 
+    # Strip quotes to allow for use of nosem as an HTML attribute inside tags.
+    # HTML comments inside tags are not allowed by the spec.
     pattern_ids = {
-        pattern_id.strip()
+        pattern_id.strip().strip("\"\'")
         for pattern_id in COMMA_SEPARATED_LIST_RE.split(ids_str)
         if pattern_id.strip()
     }
-
-    # Strip comments to allow for use of nosem as an HTML attribute inside tags.
-    # HTML comments inside tags are not allowed by the spec.
-    pattern_ids = set([pid.strip("\"\'") for pid in pattern_ids])
 
     # Filter out ids that are not alphanum+dashes+underscores+periods.
     # This removes trailing symbols from comments, such as HTML comments `-->`
--- a/semgrep/semgrep/semgrep_main.py
+++ b/semgrep/semgrep/semgrep_main.py
@@ -2,6 +2,7 @@ import json
 import logging
 from io import StringIO
 from pathlib import Path
+from re import sub
 from typing import Any
 from typing import Dict
 from typing import List
@@ -109,6 +110,9 @@ def rule_match_nosem(rule_match: RuleMatch, strict: bool) -> bool:
         for pattern_id in COMMA_SEPARATED_LIST_RE.split(ids_str)
         if pattern_id.strip()
     }
+    # Filter out ids that are not alphanum+dashes+underscores.
+    # This removes trailing symbols from comments, such as HTML comments' `-->`.
+    pattern_ids = set(filter(lambda x: not sub(r"[\w-]+", "", x), pattern_ids))
 
     result = False
     for pattern_id in pattern_ids:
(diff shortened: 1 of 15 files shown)


semgrep/semgrep/semgrep_main.py
```diff
@@ -109,6 +110,9 @@ def rule_match_nosem(rule_match: RuleMatch, strict: bool) -> bool:
         for pattern_id in COMMA_SEPARATED_LIST_RE.split(ids_str)
         if pattern_id.strip()
     }
+    # Filter out ids that are not alphanum+dashes+underscores.
+    # This removes trailing symbols from comments, such as HTML comments' `-->`.
+    pattern_ids = set(filter(lambda x: not sub(r"[\w-]+", "", x), pattern_ids))
```
mschwager: It'd probably be better to modify `NOSEM_INLINE_RE`, if possible, to allow arbitrary trailing symbols after the pattern ID list. This would also be helpful for C-like languages using something like:

```c
statement; /* nosem: pattern-id1, pattern-id2 */
```

Or even Python's

```python
statement  '''nosem: pattern-id1, pattern-id2'''
```

Although I'm not sure how you distinguish `*/` from being included as a pattern ID, or `pattern-id2'''` being used instead of `pattern-id2`. :thinking:

Another thing to consider: I think this might be the first time we're taking a hard stance that ids must be alphanum+dashes+underscores. Until now I think it's been any string. We certainly can make this requirement, but there could be downstream effects (user's existing `nosem` functionality breaking), and this may be inconsistent with other parts of the Semgrep codebase.

minusworld: I was intimidated by that RE and didn't wanna touch it 😅 

This is the "right" approach, though, so I'll take another crack at it tomorrow.

minusworld: I think it's going to be difficult to handle this case without explicitly looking for comment-ending syntax. If I'm thinking about this correctly, comment syntaxes are, by definition, unique character combinations, which means that are difficult to find with regexes in a generic fashion.

Some options here include:
1) specifying a rigid format for nosem comments, such as not allowing any spaces or using the current implementation of removing things that don't look like rule IDs.
2) explicitly removing known ending comment syntaxes.

I'm going forward with (2) because I'm afraid of unintended consequences like you pointed out, @mschwager . I also worry about "guessing" at rule IDs because I would implement a simple "does this have a readable character" filter, but can imagine edge cases where someone may not want readable characters as rule IDs. And what about Unicode? 😅 

(2) is crude but because comment endings are so unique, I think it will avoid problems for a while. It just has the downside of having to be kept up-to-date.

I am curious about @aryx's or @mjambon's or @emjin's thoughts on the matter, since they are the PA experts and all. :)

aryx: I would vote for (1), simple regexp like \bnosem:\s*[\w-_]+

minusworld: I decided to stick with the original filtering implementation, since it's a bit cleaner than (2) and (1) still nabs things like `--` from `-->`.

lmk if you think we should revisit this.

aryx (approved):
LGTM

spacegrep/src/lib/Semgrep.ml
```diff
@@ -51,7 +51,7 @@ let make_semgrep_json doc_matches : Semgrep_t.match_results =
           let ((pos1, _), (_, pos2)) = match_.region in
           let metavars = List.map convert_capture match_.named_captures in
           let lines =
-            Src_file.region_of_pos_range src pos1 pos2
+            Src_file.lines_of_pos_range src pos1 pos2
             |> String.split_on_char '\n'
           in
           let extra = {
```
mjambon, line 57: This change results in returning the full line(s) containing the match rather than just the match:
```
function() {
  hello("world");
}
```
If `"hello"` matches, json `lines` field is now populated with `["  hello(\"world\");", ""]`.

This comment is meant to document https://github.com/returntocorp/semgrep/pull/2604

aryx: So why the CI didn't fail on this PR? Why the error came later?

PR #2567, opened 2021-02-16
* semgrep-core/tree-sitter-lang/semgrep-c-sharp: update to latest

aryx:

PR #2575, opened 2021-02-17
Update csharp

Sjord:
Replaces and closes #2567.

Update the C# parser to the latest upstream tree-sitter parser.


aryx (approved):
Awesome job @Sjord 
Thx a lot!
I hope we can announce C# alpha support very soon.
How much work still remain?

PR #2576, opened 2021-02-17
Start integrating spacegrep directly in semgrep-full-rule-in-ocaml

aryx:
We want semgrep-core to be able to use full rules (not just "mini rules")
so we need to handle the [generic] language and calling spacegrep
directly.

This is the first part towards spacegrep integration in semgrep-core.
We just parse the pattern; we currently do not run the engine.

test plan:
make test


aryx:
@mjambon I've modified a few spacegrep files so I'll wait for your approval before merging this.

aryx:
(Also I may need your help to get the toplevel makefile to work again).

PR #2577, opened 2021-02-17
Exclude constructed IdSpecial from matches

### LINKED ISSUE #2540: Go type assertion missing token location info
**Describe the bug**
Type assertion label for go is missing token location info

**To Reproduce**
https://semgrep.dev/s/jRDl

**Expected behavior**
Should return results

emjin:
Closes #2540

Automatically fail when matching a metavariable with IdSpecial (Instanceof, _) or (New, _)

Test plan: make test with added test


aryx (approved):
I usually use tests/.../misc_xx.sgrep for those kinds of edge cases tests.
Remember that many of our tests are used to populate the cheatsheet on semgrep.dev, so better to keep those tests clean (see the generate_matrix script)

Issue #2548, opened 2021-02-11
metavariable-regex not working with pattern-inside?


xmo-odoo:
When using `pattern-inside` and `metavariable-regex` which tries to filter one of the `pattern-inside`'s metavariables, the match seems to never happen https://semgrep.dev/s/8ykq

```yaml
  patterns:
    - pattern-either: 
      - pattern: |
          return None
      - pattern: |
          return
    - pattern-inside: |
        class $C:
            def $METHOD(...):
              ...
    - metavariable-regex:
        metavariable: "$METHOD"
        regex: "^[^_].*"
```
The intent is to match all methods which are "public" (not starting with `_`) and use a bare `return` or a `return None`. Commenting the `metavariable-regex`, all methods with the return are matched (including those we want to exclude).

Duplicating the `pattern-inside` so the metavariables are extracted from a `pattern`, things work as expected: https://semgrep.dev/s/gLGx
```yaml
  patterns:
    - pattern-inside: |
        class $C:
            ...
    - pattern-either:
      - pattern: |
              def $METHOD(...):
                ...
                return
      - pattern: |
              def $METHOD(...):
                ...
                return None
    - metavariable-regex:
        metavariable: "$METHOD"
        regex: "^[^_].*"
```

The problem with the workaround is that I also want to match functions which don't return (and thus have an implicit `return None`), but I don't see how I could match a `pattern-not: return $X` inside a method with the proper name.

mschwager:
Hey @xmo-odoo,

Thanks for bringing this to our attention, I'm going to investigate what's going on. This looks similar to https://github.com/returntocorp/semgrep/issues/1699, hopefully we can solve this "metavariables with `pattern-inside`" in a generalized way :+1:

DrewDennison:
@xmo-odoo thanks for the bug report. Just checking what the priority of this fix is for you and if there are any timelines you're looking for a fix by?

xmo-odoo:
@DrewDennison this is a low priority thing, I'm looking into semgrep and trying to use it personally (in part in order to better understand the quirks of pattern combinations, I don't think my mental model is correct yet) but that's about it.

This I decided to report because I spent a while trying to find out what I was doing wrong and a reduced test case showed it didn't seem to be supported (sometimes I can't get a rule to work properly *in situ* but I can't get the mismatch to trigger in a reduced test case, in those cases I rather assume I am doing something wrong somehow, even if I don't know what).

[...]

PR #2580, opened 2021-02-17
Clean up evaluation.py

mschwager:
Working on this for https://github.com/returntocorp/semgrep/issues/2548. `evaluation.py` is starting to make my head explode trying to follow everything. This cleans it up a bit :+1:

### DIFF
--- a/semgrep/semgrep/evaluation.py
+++ b/semgrep/semgrep/evaluation.py
@@ -171,11 +171,12 @@ def _evaluate_single_expression(
     flags: Optional[Dict[str, Any]] = None,
 ) -> Set[Range]:
 
-    ranges_for_pattern = (
-        [x.range for x in pattern_ids_to_pattern_matches.get(expression.pattern_id, [])]
-        if expression.pattern_id
-        else []
-    )
+    if not expression.pattern_id:
+        raise SemgrepError(f"expected expression '{expression}' to have pattern_id")
+
+    ranges_for_pattern = [
+        x.range for x in pattern_ids_to_pattern_matches.get(expression.pattern_id, [])
+    ]
 
     if expression.operator == OPERATORS.AND:
         # remove all ranges that don't equal the ranges for this pattern
--- a/semgrep/semgrep/evaluation.py
+++ b/semgrep/semgrep/evaluation.py
@@ -3,7 +3,6 @@ import re
 from collections import defaultdict
 from pathlib import Path
 from typing import Any
-from typing import cast
 from typing import Dict
 from typing import Iterable
 from typing import List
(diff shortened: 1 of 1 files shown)


semgrep/semgrep/evaluation.py
```diff
@@ -172,129 +172,104 @@ def _evaluate_single_expression(
     flags: Optional[Dict[str, Any]] = None,
 ) -> Set[Range]:
 
-    assert expression.pattern_id, f"<internal error: expected pattern id: {expression}>"
```
mschwager, line 175: `assert`s are a no-op most of the time (e.g. when using `-OO` for optimization, which Python does when building wheels). The below `if` check should cover this :+1:

semgrep/semgrep/evaluation.py
```diff
@@ -486,13 +459,11 @@ def _evaluate_expression(
     steps_for_debugging: List[Dict[str, Any]],
     flags: Optional[Dict[str, Any]] = None,
 ) -> Set[Range]:
-    if (
-        expression.operator == OPERATORS.AND_EITHER
-        or expression.operator == OPERATORS.AND_ALL
-    ):
-        assert (
```
mschwager, line 493: Same as above, if we care about this `assert` we should perform a real check and `raise` if necessary.

semgrep/semgrep/evaluation.py
```diff
@@ -340,14 +314,6 @@ def group_by_pattern_id(
     return by_id
 
 
-def safe_relative_to(a: Path, b: Path) -> Path:
```
mschwager, line 343: This function is unused :broom:

semgrep/semgrep/evaluation.py
brendongo: I think the below if will silently continue if the pattern_id is falsy? (which i guess is the same behavior as an optimized out assert) Do we want to make this noisier or is it fine to continue?

semgrep/semgrep/evaluation.py
```diff
@@ -172,129 +171,105 @@ def _evaluate_single_expression(
...
-    metavars_for_patterns = get_metavar_debugging_info(
-        expression, pattern_ids_to_pattern_matches
+    ranges_for_pattern = (
+        [x.range for x in pattern_ids_to_pattern_matches.get(expression.pattern_id, [])]
+        if expression.pattern_id
+        else []
     )
 
     if expression.operator == OPERATORS.AND:
-        add_debugging_info(
```
brendongo, line 185: Nice

brendongo (approved):

semgrep/semgrep/evaluation.py
mschwager: I'll convert it to a `raise SemgrepError` to mimic the previous behavior :+1:

PR #2582, opened 2021-02-18
Start integrating spacegrep directly in semgrep-full-rule-in-ocaml (take 2)

aryx:
We want semgrep-core to be able to use full rules (not just "mini rules")
so we need to handle the [generic] language and calling spacegrep
directly.

This is the first part towards spacegrep integration in semgrep-core.
We just parse the pattern; we currently do not run the engine.

test plan:
make test


aryx:
This supersedes https://github.com/returntocorp/semgrep/pull/2576 which I did not manage to make work.
dune seems confused by having englobing dune-project, so I kept the semgrep core dune project in semgrep-core.

aryx:
It's not as nice, because we require spacegrep to be installed before being able to use it from semgrep-core, but that will do for now (we do the same with ocaml-tree-sitter).

Issue #988, opened 2020-06-12
Add test coverage for `--debugging-json`


rcoh:
Currently this feature is completely untested.

mschwager:
Closed by #2596.

PR #2596, opened 2021-02-19
More evaluation.py refactors + --debugging-json tests

mschwager:
This PR is mostly more `evaluation.py` refactors. While working on that I noticed it was hard to ensure these changes weren't breaking anything, so I added testing for `--debugging-json` as well. That should fix this longstanding issue: https://github.com/returntocorp/semgrep/issues/988. Since Semgrep App relies on `--debugging-json` it's good to have tests for this functionality.

As far as refactors go, here's what was done:

1. Simplified and consolidated the `add_debugging_info` / `get_metavar_debugging_info` functionality.
2. Exchanged the arbitrary `flags` kwarg for `allow_exec`. We're only passing around one flag, a simple bool is much easier.
3. Cleaned up the main `try` / `except` in `_where_python_statement_matches` to more easily tell where exceptions come from.
4. Added debugging typing to more easily track the requisite information.
5. Removed some unnecessary functions that were only called in one place.

I ran the snapshot test before and after the changes. The only difference was all steps now contain `metavar_ranges`, whereas before it was conditionally defined for some steps. The `ranges` union type is hard enough to keep track of, lets minimize conditional types and data :+1:


semgrep/semgrep/evaluation.py
```diff
@@ -263,34 +265,25 @@ def _evaluate_single_expression(
     else:
         raise UnknownOperatorError(f"unknown operator {expression.operator}")
 
-    metavars_for_patterns = get_metavar_debugging_info(
```
mschwager, line 266: These two calls were only called in two places: `_evaluate_single_expression` and `_evaluate_expression`. We can turn this into a single location as was done in this change by pulling the functionality up a level in `_evaluate_expression`. This makes for a much cleaner solution.

semgrep/semgrep/evaluation.py
```diff
@@ -504,9 +478,15 @@ def _evaluate_expression(
             expression,
             pattern_ids_to_pattern_matches,
             ranges_left,
-            steps_for_debugging,
-            flags=flags,
+            allow_exec=allow_exec,
         )
+
+    add_debugging_info(
```
mschwager, line 485: Here's where they were lifted out.

semgrep/tests/e2e/snapshots/test_debugging_json/test_debugging_json/results.json
```diff
@@ -0,0 +1,310 @@
...
+                {
+                  "$X": "8136864be65be7e901b2dfd746807f2a"
+                }
+              ]
+            ]
+          }
+        },
+        {
+          "filter": "pattern-not-inside",
+          "metavar_ranges": {},
```
mschwager, line 52: These were the only changes in the snapshot, before, this key was conditionally included per-step.

semgrep/semgrep/semgrep_types.py
```diff
@@ -157,3 +159,9 @@ def __eq__(self, other: Any) -> bool:
             and self.end == other.end
             and self.vars_match(other)
         )
+
+    def __lt__(self, other: Any) -> bool:
```
mschwager, line 163: This was required for sorting. I'm not sure which makes more sense here: 1.) "less than" means the start and end are smaller, or 2.) "less than" means end - start is smaller. Either way, it doesn't really matter, we just need to sort to get predictable results for comparison.

mjambon (approved):

semgrep/semgrep/semgrep_types.py
brendongo: Do we need to break ties for stability or is it fine?

brendongo (approved):

semgrep/semgrep/semgrep_types.py
mschwager: Ahhh, good call! I was banging my head against a wall all Friday wondering why I kept getting different results on Python 3.7 vs. 3.8. Looks like it's because there's no stable tie breaker, like you mentioned :+1: 

I'll add one - I think (end - start) with tie breaks going to smaller line numbers should do the trick.

mschwager:
Test failures are unrelated, merging...

PR #2597, opened 2021-02-19
This change the generic ASTs to compute the integer or float values for literals

### LINKED ISSUE #2595: Semantic equivalence of numbers
Looking for foo(8) should also report code written as foo(0x8); they are semantically equivalent.

aryx:
This closes https://github.com/returntocorp/semgrep/issues/2595

test plan:
test file included
make test


semgrep-core/Matching/Generic_vs_generic.ml
```diff
@@ -910,6 +910,25 @@ and m_literal a b =
     -> fail ()
 (*e: function [[Generic_vs_generic.m_literal]] *)
 
+and m_wrap_m_int_opt (a1, a2) (b1, b2) =
+  match a1, b1 with
+  (* iso: semantic equivalence of value! 0x8 can match 8 *)
+  | Some i1, Some i2 when i1 =|= i2 -> return ()
+  | _ ->
```
emjin, line 921: Maybe comment what happens in the other case?

semgrep-core/Analyzing/Dataflow_constness.ml
```diff
@@ -140,10 +140,10 @@ let literal_of_int i =
   let i_str = string_of_int i in
   (* TODO: use proper token when possible? *)
```
emjin, line 141: Change this comment? It looks like comparing large integers now relies on having the token be the string representation of the integer. It might also be better to call it "i_str" instead of tok. I was a little confused about what we were going to do if the integers/floats were too large

semgrep-core/Parsing/pfff/ruby_to_generic.ml
```diff
@@ -335,8 +335,9 @@ and atom x =
 and literal x =
   match x with
   | Bool x -> G.L (G.Bool (wrap bool x))
-  | Num x -> G.L (G.Int (wrap string x))
-  | Float x -> G.L (G.Float (wrap string x))
+  (* TODO: put real numbers here *)
```
emjin, line 338: Is this meant to be left as a todo?

emjin (approved):

semgrep-core/Analyzing/Dataflow_constness.ml
aryx: no the comment is about something else, that is when building AST elements, we should try to reuse tokens from
the existing AST to get better error reporting.
If the integer/floats are too large, we default to string comparisons.

semgrep-core/Parsing/pfff/ruby_to_generic.ml
aryx: yep, I was lazy and didn't finish using real numbers for Ruby, because I know the Ruby lexer is a mess right now.

semgrep-core/Matching/Generic_vs_generic.ml
```diff
@@ -910,6 +910,28 @@ and m_literal a b =
...
+   * a syntax OCaml int_of_string could not parse,
+   * we default to a string comparison *)
+  | _ ->
+      let a1 = Parse_info.str_of_info a2 in
+      let b1 = Parse_info.str_of_info b2 in
+      m_wrap m_string (a1, a2) (b1, b2)
+
+and m_wrap_m_float_opt (a1, a2) (b1, b2) =
+  match a1, b1 with
+  (* iso: semantic equivalence of value! 0x8 can match 8 *)
```
IagoAbal, line 927: Minor thing, the example for floats could be `0.123e1 can match 1.23` for example.

IagoAbal (approved):
Looks good! But there may be some small issue since it seems we got a regression:

In https://github.com/home-assistant/home-assistant:

- homeassistant/components/hassio/http.py
- homeassistant/components/zha/core/device.py
- ...

now fail to parse.

IagoAbal:
Minimal example:
```python
MAX_UPLOAD_SIZE = 1024 * 1024
```

semgrep-core/Matching/Generic_vs_generic.ml
```diff
@@ -910,6 +910,28 @@ and m_literal a b =
...
 
+and m_wrap_m_int_opt (a1, a2) (b1, b2) =
+  match a1, b1 with
+  (* iso: semantic equivalence of value! 0x8 can match 8 *)
+  | Some i1, Some i2 when i1 =|= i2 -> return ()
+  (* if the integers (or floats) were too large or were using
+   * a syntax OCaml int_of_string could not parse,
+   * we default to a string comparison *)
+  | _ ->
+      let a1 = Parse_info.str_of_info a2 in
```
IagoAbal, line 922: The integer literal may come from constant propagation and in that case the token was created with `Parse_info.fake_info`. Passing one of these fake tokens to `str_of_info` will make Semgrep crash with a `NoTokenLocation` exception.

There are different things we could do here... for example, we could change

```ocaml
| Some i1, Some i2 when i1 =|= i2 -> return ()
```

to

```ocaml
| Some i1, Some i2 -> if i1 =|= i2 then return () else fail ()
```

We could perhaps also make `str_of_info` get the first string argument of `FakeTokStr`? And perhaps `token_location_of_info` could return throw a warning and return a dummy `token_location`?

semgrep-core/Analyzing/Dataflow_constness.ml
IagoAbal: Actually we would have to be more careful now, if we evaluate `2 * 42` we cannot just reuse the token of either `2` or  `42`, we would need to change it's `str` field, since this may be used for matching.

aryx:
Thx! Yep good idea i ll
Change str_of_info

PR #2603, opened 2021-02-22
Use 'get_metavariable_value' instead of 'abstract_content' and cache lookups

mschwager:


semgrep/semgrep/evaluation.py
```diff
@@ -206,6 +206,20 @@ def _evaluate_single_expression(
...
+        output_ranges = {
+            _range
+            for _range in ranges_left
+            if not any(
+                _range.is_range_enclosing_or_eq(pattern_range)
+                or pattern_range.is_range_enclosing_or_eq(_range)
+                for pattern_range in ranges_for_pattern
+            )
+        }
     elif expression.operator == OPERATORS.WHERE_PYTHON:
```
mschwager, line 223: I refactored `WHERE_PYTHON` down since it's more similar to `METAVARIABLE_REGEX` and `METAVARIABLE_COMPARISON` :broom:

aryx (approved):

aryx:
Note that all this code may soon disappear ... Right now semgrep-core can process a whole rule, with and/or/not, pattern-regexp, spacegrep, metavariable-regexp, metavariable-comparison, YAML/JSON/JSONNET, etc. Just pattern-where-python is not handled. The goal is that
soon semgrep will pass a whole rule to semgrep-core, so evaluation.py should be used just for pattern-where-python.

mschwager:
> Note that all this code may soon disappear ... Right now semgrep-core can process a whole rule, with and/or/not, pattern-regexp, spacegrep, metavariable-regexp, metavariable-comparison, YAML/JSON/JSONNET, etc. Just pattern-where-python is not handled. The goal is that
> soon semgrep will pass a whole rule to semgrep-core, so evaluation.py should be used just for pattern-where-python.

Ahh, cool. Do all the tests pass in `semgrep-rules`? To start, it seems we'll have to replicate the existing behavior so we don't break anything for existing users.

aryx:
Yes, right now I still have 28 mismatch (difference of behavior between semgrep -config and semrep-core -config) in semgrep-rules. It's not that bad because there are 2800 ruleid: in semgrep-rules, so we are at 1% right now. Hopefully I can get down to 0 this week, thanks to your refactoring it will be easier for me to understand and reproduce exactly what our evaluation.py does.

mschwager:
> Yes, right now I still have 28 mismatch (difference of behavior between semgrep -config and semrep-core -config) in semgrep-rules. It's not that bad because there are 2800 ruleid: in semgrep-rules, so we are at 1% right now. Hopefully I can get down to 0 this week, thanks to your refactoring it will be easier for me to understand and reproduce exactly what our evaluation.py does.

Nice! One other consideration is that `--debugging-json`, which Semgrep App relies on, is pretty tightly tied to the Python evaluation implementation. Not sure how much of a pain it is to port this functionality to OCaml

PR #2604, opened 2021-02-22
* Restore test_spacegrep test on markdown

aryx:
I have no idea what I'm doing. This test used to fail in CI on
unrelated PR, so I comment it; it was actually working locally.
And now I put it back and it also fails locally, so I could update
the snapshot.
No idea why it failed in CI and worked locally, and now why it also
does not work locally.

Test plan:
pytest tests/e2e/test_spacegrep.py

### DIFF
--- a/semgrep/tests/e2e/snapshots/test_spacegrep/test_spacegrep/rulesspacegrepmarkdown.yaml-spacegrepmarkdown.md/results.json
+++ b/semgrep/tests/e2e/snapshots/test_spacegrep/test_spacegrep/rulesspacegrepmarkdown.yaml-spacegrepmarkdown.md/results.json
@@ -208,12 +208,12 @@
             "end": {
               "col": 14,
               "line": 203,
-              "offset": 10948
+              "offset": 10946
             },
             "start": {
               "col": 4,
               "line": 203,
-              "offset": 10938
+              "offset": 10936
             },
             "unique_id": {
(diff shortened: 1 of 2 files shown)


mschwager (approved):
Sounds like we forgot to run `--snapshot-update` somewhere along the line? Either way, this LGTM, but I defer to @mjambon or @minusworld on if the offset changes affect anything

mjambon:
@minusworld changed what spacegrep puts into the `lines` field, which is now the full lines containing the match rather than just the region of the match (which didn't have the beginning of the first line or the end of the last line). Since we're calling the `spacegrep` executable and @mschwager mentions a `--snapshot-update` option, it looks like the test failure perhaps wasn't triggered in @minusworld's original PR https://github.com/returntocorp/semgrep/pull/2574 due some older version of `spacegrep` being used.

* I don't know about this `--snapshot-update`, but let's make sure the version of `spacegrep` built from the current PR is used.
* The new offset by 2 characters may be legitimate, not sure. We don't have unit tests for this in spacegrep, which is bad.

Plan:
* @mschwager or @minusworld is the spacegrep executable cached? and we used the that instead of a fresh build? If nobody answers, I'll look into it myself.
* I'm adding unit tests to spacegrep for `Src_file.lines_of_pos_range` and friends so that it's one less thing to worry about.

aryx:
Sounds like a good plan martin. Thx for the investigation!

PR #2607, opened 2021-02-23
Add unit tests for spacegrep's output, change 'lines' field

mjambon:
This changes the value of the `lines` field of the json object representing a match, which is returned to semgrep. Previously, the lines `"a\nb\n"` would be split into `["a"; "b"; ""]`. Now it's just `["a"; "b"]`.

Let's see how many tests this breaks! 😬


mjambon:
There were some errors in tox tests apparently while running with `--lang javascript` which shouldn't have anything to do with spacegrep. I'm re-running GitHub Actions.

aryx (approved):

Issue #2298, opened 2020-12-18
Duplicates across rulesets should be eliminated


msorens:
**Describe the bug**
I recently changed a policy from one ruleset (`r2c-ci`) to multiple rulesets. Upon adding a dummy test file to trigger a finding in CI, the finding was reported twice. I have not verified, but I suspect the rule is present in two rulesets so is therefore reporting twice:

<img width="425" alt="image" src="https://user-images.githubusercontent.com/6817500/102645668-b021f200-4117-11eb-9f99-97b380f20780.png">

**To Reproduce**
Here are the rulesets I used:
<img width="173" alt="image" src="https://user-images.githubusercontent.com/6817500/102645936-1c9cf100-4118-11eb-9336-64c7c50fdaa9.png">

I then just added a test.py file to my repository with the contents `5 == 5`.

**Expected behavior**
Finding reported just once.

**Screenshots**
NA

**What is the priority of the bug to you?**
Is this a P0 (blocking your adoption of Semgrep or workflow), P1 (important to fix or quite annoying), P2 (regular bug that should get fixed)?

P2 -- regular

**Environment**
v 0.35.0

msorens:
Another data point. I was wondering why my open findings jumped from 21 to 25 just recently. I took the real output of my nightly semgrep run on my codebase, trimmed the output, and sorted to show the issue succinctly:
```
dgryski.semgrep-go.errtodo.err-todo > components/automate-deployment/pkg/a1upgrade/filemover.go:217
dgryski.semgrep-go.errtodo.err-todo > components/automate-deployment/pkg/client/deployer.go:1107
dgryski.semgrep-go.errtodo.err-todo > components/automate-gateway/gateway/services.go:1018
dgryski.semgrep-go.errtodo.err-todo > components/compliance-service/inspec-agent/runner/runner.go:339
dgryski.semgrep-go.errtodo.err-todo > components/infra-proxy-service/storage/postgres/orgs.go:181
dgryski.semgrep-go.errtodo.err-todo > components/infra-proxy-service/storage/postgres/servers.go:107
dgryski.semgrep-go.errtodo.err-todo > components/ingest-service/pipeline/publisher/bulk_publisher.go:55
dgryski.semgrep-go.errtodo.err-todo > components/session-service/server/server.go:479
dgryski.semgrep-go.errtodo.err-todo > components/teams-service/storage/postgres/postgres.go:243
[...]

PR #2608, opened 2021-02-23
Deduplicate rules prior to run (Closes: #2298)

hlieberman-rd:
Because the rule objects contain context from the ruleset they came
from, we can't directly compare them. Instead, we have to do a little
dance and indirectly deduplicate them by their id.

Signed-off-by: Harlan Lieberman-Berg <harlan@rebelliondefense.com>

### DIFF
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import time
+from collections import OrderedDict
 from pathlib import Path
 from typing import Any
 from typing import Dict
@@ -133,16 +134,9 @@ class Config:
             # re-write the configs to have the hierarchical rule ids
             configs = self._rename_rule_ids(configs)
 
-        rules = []
-        seen = set()
-        for rule in [rule for rules in configs.values() for rule in rules]:
-            if rule.id in seen:
-                next
-            else:
-                rules.append(rule)
-                seen.add(rule.id)
-
-        return rules
+        return list(
+            OrderedDict.fromkeys([rule for rules in configs.values() for rule in rules])
+        )
 
     @staticmethod
     def _safe_relative_to(a: Path, b: Path) -> Path:
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -133,7 +133,16 @@ class Config:
             # re-write the configs to have the hierarchical rule ids
             configs = self._rename_rule_ids(configs)
 
-        return [rule for rules in configs.values() for rule in rules]
+        rules = []
+        seen = set()
+        for rule in [rule for rules in configs.values() for rule in rules]:
+            if rule.id in seen:
+                next
+            else:
+                rules.append(rule)
+                seen.add(rule.id)
+
+        return rules
 
     @staticmethod
     def _safe_relative_to(a: Path, b: Path) -> Path:
--- a/semgrep/semgrep/rule.py
+++ b/semgrep/semgrep/rule.py
@@ -68,6 +68,15 @@ class Rule:
         if any([lang in REGEX_ONLY_LANGUAGE_KEYS for lang in self._languages]):
             self._validate_none_language_rule()
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, type(self)):
+            return False
+
+        return self._raw == other._raw
+
+    def __hash__(self) -> int:
+        return hash(self.id)
+
     def _validate_none_language_rule(self) -> None:
         """
         For regex-only rules, only patterns, pattern-either, and pattern-regex is valid.
--- a/semgrep/tests/e2e/snapshots/test_check/test_multiple_configs_file/results.json
+++ b/semgrep/tests/e2e/snapshots/test_check/test_multiple_configs_file/results.json
@@ -78,44 +78,6 @@
         "col": 13,
         "line": 3
       }
-    },
-    {
-      "check_id": "rules.eqeq-is-bad",
-      "end": {
-        "col": 26,
-        "line": 3
-      },
-      "extra": {
-        "is_ignored": false,
-        "lines": "    return a + b == a + b",
-        "message": "useless comparison operation `a + b == a + b` or `a + b != a + b`; possible bug?",
-        "metadata": {},
-        "metavars": {
-          "$X": {
-            "abstract_content": "a+b",
-            "end": {
-              "col": 17,
-              "line": 3,
-              "offset": 60
-            },
-            "start": {
-              "col": 12,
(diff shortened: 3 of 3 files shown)


hlieberman-rd:
This saves about 10% on the intersection of r2c-ci and r2c-best-practices (234 deduped vs 259).

aryx:
CI is failing. Needs investigation.
You can run the tests locally with: cd /path/to/semgrep_repo/semgrep; make test

aryx (changes requested):
Failure in CI.
There's also probably something shorter than this in Python. @mschwager ?

hlieberman-rd:
The CI failures appear unrelated; origin/develop is failing on my machine
as well.--
Harlan Lieberman-Berg
Head of Security
Rebellion Defense

aryx:
I don't think this one is failing in develop: FAILED tests/e2e/test_check.py::test_multiple_configs_file - AssertionError: ...

But yes, thx for reporting I'll try to fix develop.

aryx:
@cla-bot check

aryx:
@cla-bot check

aryx:
Yeah, the verification/cla-signed approved it!

semgrep/semgrep/config_resolver.py
```diff
@@ -133,7 +133,16 @@ def get_rules(self, no_rewrite_rule_ids: bool) -> List[Rule]:
             # re-write the configs to have the hierarchical rule ids
             configs = self._rename_rule_ids(configs)
 
-        return [rule for rules in configs.values() for rule in rules]
+        rules = []
+        seen = set()
+        for rule in [rule for rules in configs.values() for rule in rules]:
+            if rule.id in seen:
```
mschwager: Rule IDs are not a one-to-one mapping with a rule ([see e.g. here](https://github.com/returntocorp/semgrep-rules/blob/develop/go/lang/correctness/permissions/file_permission.yaml)), so we can't use `rule.id` to filter seen rules. If we define a reasonable `__eq__` comparator on the `rule.Rule` class, then we could simply do something like the following here:

```
return set(rule for rules in configs.values() for rule in rules)
```

And the `set` will handle everything for us :+1:

hlieberman-rd: Hi @mschwager!

Good to know; I will have to extend Rule with `__eq__` and `__hash__`.

Unfortunately, we can't use sets here for the rules themselves; the order being non-deterministic causes serious problems with the snapshot tests.  I talked to @brendongo about this on Slack, and the best option that we came to was to continue maintaining them as a list for the moment pending a more significant overhaul of this section of the code.

I experimented with passing it through `dict.fromkeys`, however, with Python 3.6, we'd be relying on a Cython implementation detail; dict was not guaranteed to preserve insertion order until Python 3.7.

The other option is to pass it through set as an intermediary and then sort the output, however, that would require adding `__lt__` and regenerating all the tests.

Happy to do either; let me know which is preferable.

mschwager: Hey there,

Ahh, interesting, good catch. We could use `OrderedDict.fromkeys` - something like:

```python
# Remove duplicates, preserve order
return list(OrderedDict.fromkeys(rule for rules in configs.values() for rule in rules))
```

Would that minimize snapshot thrashing?

brendongo:
Fixed issue on develop and rerunning tests

brendongo (dismissed):

brendongo (approved):

nbrahms:
@aryx is your review still pertinent, or should I dismiss it? 🙏

aryx:
There's still this CI failure: FAILED tests/e2e/test_check.py::test_multiple_configs_file - AssertionError: ...
 that does not happen on develop.
We can't merge until all the tests pass.

semgrep/semgrep/rule.py
```diff
@@ -68,6 +68,15 @@ def __init__(self, raw: YamlTree[YamlMap]) -> None:
...
             self._validate_none_language_rule()
 
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, type(self)):
+            return False
+
+        return self._raw == other._raw
+
+    def __hash__(self) -> int:
+        return hash(self.id)
```
mschwager, line 78: I think this may be problematic given the above definition of `__eq__`. Since `id` is not guaranteed to be unique this means we could have rules that produce the same hashable value, but are not `==`. I.e. they overwrite each other in a `dict` or `set`, but are not equivalent. If we want something really quick and easy, although not particularly fast, we could do something like `hash(json.dumps(self._raw))` or something like that.

hlieberman-rd: @mschwager That's OK -- hash equivalence isn't treated as the same as equality.

For example:

```python3
class Foo:
    def __init__(self, a, b):
        self.a = a
        self.b = b

    def __eq__(self, other):
        if not isinstance(other, type(self)):
            return False
        return (self.a == other.a and self.b == other.b)

    def __hash__(self):
        return hash(self.a)

x = Foo("abc", "def")
y = Foo("abc", "xyz")
z = Foo(888, "def")

print(hash(x))
print(hash(y))
print(hash(z))
print(x == y)
print(x == z)
print(y == z)
```
returns
```
-4144991442677912395
-4144991442677912395
888
False
False
False
```

Equivalence never checks `__hash__`.  Membership does (a la `x in y`), but only as a performance improvement.  Hash collisions are, afaik, only a performance hit, never a logical mistake.  (The performance hit of doing the deep comparison between the raw dicts on id collision is less than the hashing and serializing the json on every comparison).

I believe the test failure you're seeing @aryx is actually the corrected behavior: the snapshot version appears to be producing two hits from identical rules in two files.

```
>>> with open("tests/e2e/snapshots/test_check/test_multiple_configs_file/results.json") as x:
...     y = x.read()
...
>>> import json
>>> foo = json.loads(y)
>>> foo["results"][2] == foo["results"][0]
True
```

I will push a new patch correcting the snapshot.

hlieberman-rd: Sorry @mschwager, I misread your concern.  dicts and sets also check equality, not just hashing.  Extending the above Foo class example:

```
bar = set()
bar.add(x)
bar.add(y)
bar.add(z)
print(len(bar))
bar.add(x)
bar.add(y)
print(len(bar))
```
returns
```
3
3
```

mschwager: Ahh, TIL that hashable types rely on `__eq__` too! [`__hash__`](https://docs.python.org/3/reference/datamodel.html#object.__hash__) was a lengthy, but helpful read. Thanks for the clarifications! This LGTM :+1:

brendongo:
Thanks @hlieberman-rd this should go in with next week's release!

Issue #2078, opened 2020-11-18
javascript timeout errors on Signal-Desktop WASM autogenerated code


ievans:
**Describe the bug**
Semgrep times out running rules against three large files in Signal desktop:

```
ine@imbp4 ~/D/r/t/Signal-Desktop (development)> cat libtextsecure/libsignal-protocol.js | wc -l
   25286
ine@imbp4 ~/D/r/t/Signal-Desktop (development)> cat components/mp3lameencoder/lib/Mp3LameEncoder.js | wc -l
   46699
ine@imbp4 ~/D/r/t/Signal-Desktop (development)> cat js/Mp3LameEncoder.min.js | wc -l
   46699
```

**To Reproduce**

$ git clone https://github.com/signalapp/Signal-Desktop
$ time semgrep --config=r/javascript.jquery.security.audit.jquery-insecure-selector.jquery-insecure-selector Signal-Desktop

**Expected behavior**
I think Semgrep should be able to parse a 25K LoC file and run a rule in <30 sec.

**What is the priority of the bug to you?**
Timeouts are correctly reported as of v.32, so marking as low-pri.

**Environment**
Semgrep v.32

aryx:
v32.0 probably solves that

DrewDennison:
@ievans there is now a 30s timeout. Can you retry with latest?

ievans:
Thanks @DrewDennison. As of v.32 we now timeout successfully and have only three files that we fail on! I'll edit the issue to reflect

```
time semgrep --config=r/javascript.jquery.security.audit.jquery-insecure-selector.jquery-insecure-selector
using config from https://semgrep.dev/r/javascript.jquery.security.audit.jquery-insecure-selector.jquery-insecure-selector. Visit https://semgrep.dev/registry to see all public rules.
downloading config...
running 1 rules...
Warning: 1 timeout error(s) in libtextsecure/libsignal-protocol.js when running the following rules: [javascript.jquery.security.audit.jquery-insecure-selector.jquery-insecure-selector]
Warning: 1 timeout error(s) in components/mp3lameencoder/lib/Mp3LameEncoder.js when running the following rules: [javascript.jquery.security.audit.jquery-insecure-selector.jquery-insecure-selector]
Warning: 1 timeout error(s) in js/Mp3LameEncoder.min.js when running the following rules: [javascript.jquery.security.audit.jquery-insecure-selector.jquery-insecure-selector]
js/views/recorder_view.js
[...]

PR #2618, opened 2021-02-24
Optimise function(...) { ... } matching

aryx:
This helps https://github.com/returntocorp/semgrep/issues/2078

test plan:
semgrep-core -lang js -f perf/input/function-ellipsis.sgrep
perf/input/l50000.js
now terminates!

function-ellipsis2.sgrep still does not terminate, but we should
probably skip WASM code to handle it.


aryx:
This diffs also add a debug_timeout flag that leverages the new full-rule-in-ocaml engine code to debug rules.
Pretty useful.

PR #2621, opened 2021-02-24
Cache PatternMatch metavariable access

mschwager:
`get_metavariable_value` is often run in a tight loop. Let's read the file from disk once, get all the metavariable data, then cache it going forward instead of reading from disk every time.

### DIFF
--- a/semgrep/semgrep/pattern_match.py
+++ b/semgrep/semgrep/pattern_match.py
@@ -66,6 +66,9 @@ class PatternMatch:
         return end
 
     def _read_metavariable_values(self) -> Dict[str, str]:
+        """
+        Read self.path and lookup all values of metavariables in self.metavars
+        """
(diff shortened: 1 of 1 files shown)


brendongo (approved):

semgrep/semgrep/pattern_match.py
```diff
@@ -64,25 +65,33 @@ def end(self) -> Dict[str, Any]:
             del end["offset"]
         return end
 
-    @functools.lru_cache(maxsize=None)
+    def _read_metavariable_values(self) -> Dict[str, str]:
```
brendongo, line 68:
```suggestion
    def _read_metavariable_values(self) -> Dict[str, str]:
        """
        Read self.path and lookup all values of metavariables in self.metavars
        """
```

Issue #2601, opened 2021-02-22
Language support for YAML


marcelocarlos:
**Is your feature request related to a problem? Please describe.**

Not exactly, but my main interest/plan to use a semgrep would be to analyse IaC, such as kubernetes definitions. I can see HCL is being already tracked (https://github.com/returntocorp/semgrep/issues/1147) and hopefully available soon, however, I couldn't find references to YAML. I could obviously get away with converting it to JSON, but I believe having native support to YAML would simplify things as well as encourage contributions and growth of rules for kubernetes and other yaml-based solutions.

**Describe the solution you'd like**

Language support for YAML. I found a tree-sitter https://github.com/ikatyang/tree-sitter-yaml for YAML which might help to speed things up a bit.

**Describe alternatives you've considered**

I tried using the generic parser (spacegrep) but it has proven to be fairly difficult to work with YAML for more complex rules. Converting YAML to JSON was better, but still not ideal since it requires conversion of files before checking them.

**Additional context**

I'd love to see semgrep making IaC resources first class citizens and having YAML + HCL would be a fantastic start to allow people to contribute and add new rules to it. Being able to use the same solution for programming languages checks as well as IaC would be absolutely fantastic!

Thank you for the great work so far! 👍

aryx:
Yep YAML is on our roadmap for Q1 2021. Cc @DrewDennison

emjin:
Hey! @marcelocarlos I was wondering if you had specific examples you had in mind for yaml patterns that you could share. Our first priority is to implement yaml for semgrep in a way that is helpful to the people who want it. Thanks!

marcelocarlos:
Hi @emjin, thanks for getting in touch.

What I have in mind is running checks for kubernetes definitions, such as checking if certain labels are defined in the template, if cpu/memory limits are set and within certain ranges, if a `securityContext` is defined and contains `runAsNonRoot: true` in it, etc. Basically, being able to cover rules/patterns to ensure kubernetes definitions follow best practices for pods, services, deployments, etc.

aryx:
@marcelocarlos could you try to write those checks using an imaginary semgrep syntax for YAML?
That way we have concrete examples to follow.

fproulx-boostsecurity:
Any chance this might ship in next release of `semgrep` ?
So many great IaC (Kubernetes, etc..) opportunities.

aryx:
[...]

PR #2630, opened 2021-02-24
Use the yaml package to parse yaml code into the generic AST

emjin:
The yaml package we use to parse ocaml has a low level stream API that gives the tokens lexed. We use this stream API to turn yaml into a generic AST, with manually generated tokens. These tokens are all FakeTokStr, but the position information is set. Note that if you have

- id: some_id

this will consider the column of "id" to be 2. Values are 0 indexed (this may need to be changed?)

Does not currently handle alias/anchors or semgrep patterns. Doing this may require manually separating out ellipses

Experiment in support of #2601 

Test plan: parsing the below yaml snippet should result in
```
rules:
- id: my_pattern_id
  pattern: |
    if ( $X ) $Y;
  message: |
    Semgrep found a match
  severity: WARNING
```
```
E(
  Container(Dict,
    [Tuple(
       [L(String(("rules", ())));
        Container(Array,
          [Container(Dict,
             [Tuple([L(String(("id", ()))); L(String(("my_pattern_id", ())))]);
              Tuple(
                [L(String(("pattern", ())));
                 L(String(("if ( $X ) $Y;
", ())))]);
              Tuple(
                [L(String(("message", ())));
                 L(String(("Semgrep found a match
", ())))]);
              Tuple([L(String(("severity", ()))); L(String(("WARNING", ())))])])])])]))
```

### DIFF
--- a/semgrep-core/Parsing/other/yaml_to_generic.ml
+++ b/semgrep-core/Parsing/other/yaml_to_generic.ml
@@ -1,0 +1,121 @@
+module Y = Yaml
+module S = Yaml.Stream
+module E = Yaml.Stream.Event
+module M = Yaml.Stream.Mark
+module A = AST_generic
+
+exception Unimplemented
+exception ParseError of string
+
+(* Parser for YAML. The tree-sitter grammar for YAML is somewhat complicated,
+ * since YAML is whitespace sensitive. Instead, we use the OCaml Yaml module,
+ * which we also use to parse YAML for semgrep. To get positions, we use the
+ * low-level Stream API, which we parse into a generic AST
+*)
+
+(* Helper functions *)
+
+let fst (a, _) = a
+
+let p_token = function
+  | E.Stream_start _ -> "stream start"
+  | E.Stream_end -> "stream end"
+  | E.Document_start _ -> "document start"
+  | E.Document_end _ -> "document end"
+  | E.Sequence_start _ -> "sequence start"
+  | E.Sequence_end -> "sequence end"
+  | E.Mapping_start _ -> "mapping start"
+  | E.Mapping_end -> "mapping end"
+  | E.Alias _ -> "alias"
+  | E.Scalar { value; _ } -> "scalar " ^ value
+  | E.Nothing -> "nothing"
+
+let mk_err err v {E.start_mark = {M.line; M.column; _}; _} =
+  Printf.sprintf "%s %s at line %d column %d" err (p_token v) line column
+
+let get_res = function
+  | Result.Error (`Msg s) -> raise (ParseError s)
+  | Result.Ok v -> v
+
+let any_of_list = function
+  | [x] -> A.E x
+  | xs -> A.E (A.Seq xs) (* Change to array? *)
+
+let tok (index, line, column) str file =
+  {Parse_info.token = Parse_info.FakeTokStr (str, Some ({str; charpos = index; line; column; file}, 0));
+   Parse_info.transfo = NoTransfo}
+
+let mk_tok {E.start_mark = {M.index; M.line; M.column}; _} str file =
+  tok (index, line, column) str file
+
+let mk_bracket ({E.start_mark = {M.index = s_index; M.line = s_line; M.column = s_column}; _},
+                {E.end_mark = {M.index = e_index; M.line = e_line; M.column = e_column}; _}) v file =
+  tok (s_index, s_line, s_column) "(" file, v, tok (e_index, e_line, e_column) ")" file
+
+(* Parser functions *)
+
+let make_alias anchor pos file : A.name =
+  A.Id ((anchor, (mk_tok pos "" file)), {id_resolved = ref None; id_type = ref None; id_constness = ref None})
+
+let make_scalar _anchor _tag pos value file : A.expr =
+  A.L (A.String (value, mk_tok pos value file))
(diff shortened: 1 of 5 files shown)


semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -0,0 +1,121 @@
...
+
+(* Parser for YAML. The tree-sitter grammar for YAML is somewhat complicated,
+ * since YAML is whitespace sensitive. Instead, we use the OCaml Yaml module,
+ * which we also use to parse YAML for semgrep. To get positions, we use the
+ * low-level Stream API, which we parse into a generic AST
+*)
+
+(* Helper functions *)
+
+let fst (a, _) = a
```
aryx, line 18: This is in Stdlib, no need to define it, it's a builtin.

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -0,0 +1,121 @@
...
+let mk_err err v {E.start_mark = {M.line; M.column; _}; _} =
+  Printf.sprintf "%s %s at line %d column %d" err (p_token v) line column
+
+let get_res = function
+  | Result.Error (`Msg s) -> raise (ParseError s)
+  | Result.Ok v -> v
+
+let any_of_list = function
+  | [x] -> A.E x
+  | xs -> A.E (A.Seq xs) (* Change to array? *)
```
aryx, line 42: yep, Array.

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -0,0 +1,121 @@
...
+
+let tok (index, line, column) str file =
```
aryx, line 44: You can look at the code of Parse_tree_sitter_helpers.token which also build token
from tree-sitter-info.
  (* Parse_info is 1-line based and 0-column based, like Emacs *)

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -0,0 +1,121 @@
...
+  {Parse_info.token = Parse_info.FakeTokStr (str, Some ({str; charpos = index; line; column; file}, 0));
```
aryx, line 45: We don't want a FakeTokStr here. Use OriginTok (see Parse_tree_sitter_helpers.token() again).

OriginTok means that the token is in the original file. Here you build token that correspond to things
present in the original file. FakeTokStr is to build fake token, like a fake ";" that actually is not in the original file,
but that we use to help parse in a grammar.

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -0,0 +1,121 @@
...
+   Parse_info.transfo = NoTransfo}
+
+let mk_tok {E.start_mark = {M.index; M.line; M.column}; _} str file =
+  tok (index, line, column) str file
+
+let mk_bracket ({E.start_mark = {M.index = s_index; M.line = s_line; M.column = s_column}; _},
```
aryx, line 51: in those cases we might need to build a FakeTok if the parenthesis are actually not in the yaml file.

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -0,0 +1,121 @@
...
+                {E.end_mark = {M.index = e_index; M.line = e_line; M.column = e_column}; _}) v file =
+  tok (s_index, s_line, s_column) "(" file, v, tok (e_index, e_line, e_column) ")" file
+
+(* Parser functions *)
+
+let make_alias anchor pos file : A.name =
+  A.Id ((anchor, (mk_tok pos "" file)), {id_resolved = ref None; id_type = ref None; id_constness = ref None})
```
aryx, line 58: You can use A.empty_id_info()

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -0,0 +1,121 @@
...
+
+let make_scalar _anchor _tag pos value file : A.expr =
+  A.L (A.String (value, mk_tok pos value file))
```
aryx, line 61: you can put a TODO here saying we may want to generate a A.Float or A.Bool depending on the string content.

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -0,0 +1,121 @@
...
+    A.Tuple (mk_bracket (pos1, pos2) [key; value] file)
+  in
+  read_stream ()
+
+let parse_yaml str =
+  get_res (S.parser str) |> parse "dummy_file_name"
+
+(* Entry point *)
+
+let program _file = (* parse_yaml file *) raise Unimplemented
```
aryx, line 119: Why not implemented?
We actually have a file here, so we can have the right position information.
The position information for a pattern is not that important; it's really important for the yaml target file though,
because we report ranges on those target file.

BTW you may need to pass down more information than a file later, so it's good practice to define
type env = {
  file: Common.filename;
}
and pass that around instead of just the file.
That way you can add more info later.

aryx (approved):

aryx:
Ok let's merge this, we can improve later.

PR #2632, opened 2021-02-24
matching: Match variables with literals modulo constness

### LINKED ISSUE #374: Support dataflow for constants in conditionals
@aryx I think we should be able to support this type of dataflow for constants now?

```javascript
const x = "asdf";

if (x === "asdf") {
    console.log("foobar")
}

```

https://sgrep.live/Ken

IagoAbal:
As a result of adding this feature, we have to be more rigorous when
assigning constness to variables. Now setting constness to a literal
means that the specific occurrence of the variable can be replaced by
the literal. And in `x = E` the variable `x` stands for a label rather
than a value, so we cannot set its constness to any other value than
_Top_ (`NotCst`).

Closes #374

test plan:
make test


semgrep-core/Analyzing/Dataflow_constness.mli
```diff
@@ -1,7 +1,11 @@
 
-type mapping = AST_generic.constness Dataflow.mapping
+module G = AST_generic
```
aryx, line 2: please no module aliasing in .mli (also semgrep for ocaml does not like it because it's not in
the current grammar for ocaml in pfff)

aryx (approved):

semgrep-core/Analyzing/Constant_propagation.ml
```diff
@@ -304,22 +313,21 @@ let propagate_basic lang prog =
          | N (Id (id, id_info)) ->
              (match find_id env id id_info with
               | Some literal ->
-                  id_info.id_constness := Some (Lit literal)
+                  join_constness_ref id_info.id_constness (Lit literal)
```
aryx, line 316: But here id_constness should always be None no? this join is really a set_constness_ref?
Or is there a situation where it can be already set to a Some?

semgrep-core/Analyzing/Dataflow_constness.ml
```diff
@@ -362,14 +362,5 @@ let update_constness (flow :F.cfg) mapping =
            | Some c -> refine_constness_ref constness c
           )
       | ___else___ -> ()
-    );
-
-    (* Update LHS constness according to the output env. *)
```
aryx, line 367: maybe you can keep a comment saying "we used to update the LHS constness according to the output env
but in x = E, x is a label, and should not be substituted for its value".

aryx (changes requested):

semgrep-core/Analyzing/Dataflow_constness.mli
aryx: Note that you can do 'make check' inside semgrep-core/ and it will run semgrep on the ocaml code.
(you need to have semgrep in your path, so you can do pipenv shell from semgrep/semgrep/ to get semgrep in your PATH).

semgrep-core/Analyzing/Constant_propagation.ml
IagoAbal: Note that in the first step, `var_stats`, I go through all assignments `x = E` and assign `x` the constness `NotCst`. Here we are visiting a name, but we don't know whether this name is acting as a label (l-value) or as an r-value. If it were an r-value then it's correct that constness would be `None`, but it is an l-value then it would be `NotCst`. Perhaps I should document this better!

semgrep-core/Analyzing/Dataflow_constness.mli
IagoAbal: Fixed! Thanks for the tip, I still get problems to run Python's semgrep wrapper (it worked but then the setup changed and it no longer works). But I'll try to get it running.

semgrep-core/Analyzing/Constant_propagation.ml
IagoAbal: Hmmm, maybe this is a bit hacky, it may be more clear if I change the way the AST is visited.

IagoAbal: The way I do it with `join_constness_ref` is a bit tricky but right now I think it may be the simplest way. The visitor for the AST doesn't allow to visit r-value expressions only I think. It would be easier if we update constness using IL, since in IL it's very clear where a variable is being read (that's where we want to update the constness) and where it acts as a label (here we don't want to set the constness). But right now the AST-to-IL transformation only works for individual functions and not for entire programs, so we would need to extend it first.

What do you think?

aryx: yep, would be better to do that in IL where lvalue/rvalue is better separated.
Also right now we don't really detect well lvalue in the AST because you can do tuple assignement like (a, b, c) = (1, 2, 3).
In the IL it would be simpler because we could decompose those as a = 1; b = 2; c = 3, or if the rhs is not a tuple as in
(a, b, c) = foo(), we should generate in the IL   _tmp = foo(); a = _tuple_access(_tmp, 1); b = _tuple_accss(_tmp, 2); ...

IagoAbal: OK then I will pause this PR, work on a new PR to translate a full generic `program` into `IL` (maybe with some TODOs), then get back to this PR. May take some days though.

(Note that AST-to-IL already handles tuples in the LHS of assigments.)

IagoAbal:
The best would still be to rely on IL for this, but to make sure there is no regression we need a pretty complete translation from AST to IL. So I will see if there is any other thing we can do to get this merged sooner. I need to look into the `Lrvalue` module more carefully, but it could be a possibility.

aryx:
Lrvalue.ml is ugly. The IL is better for proper lrvalue computation

IagoAbal:
There is now a bug report (#2805) that relates to setting the constness of lvalue variables. We could use the same trick as in Naming_AST, where basically you set an `lvalue` flag when visiting the LHS of a definition. It's not perfect because e.g. in C you have `*x = E;` where `x` is actually an rvalue. But it shouldn't have a negative effect on our typical use case I think, and it would be temporary until we can switch to IL. Would that be fine?

IagoAbal:
There is now a separate issue #2805 for the problem of labels getting constness info. I will think again on a solution that doesn't require completing the AST-to-IL translation yet, even if it may not be ideal. We can probably do the same way as in Naming_AST.

IagoAbal:
This PR got a lot more complicated than it should have been due to the lvalue-vs-rvalue issue. But that was reported separately in #2805 and fixed by #3042, so I am closing this PR and I will open a fresh one.

PR #2637, opened 2021-02-25
Benchmarks

mjambon:
I set up dummy benchmarks triggered by each commit to the `benchmarks` branch in addition to the regular daily benchmarks on the main branch. Let's first see if the dummy benchmarks get triggered properly.

Update: I got the dummy benchmarks to pass after a bit of trial-and-error.

... and I started a manual run of the real benchmarks on the CircleCI host via ssh. It may take a while.


perf/Makefile
```diff
@@ -0,0 +1,15 @@
...
+run:
+	./run-benchmarks --docker returntocorp/semgrep:develop
+
+.PHONY: dummy
+dummy:
+	./run-benchmarks --dummy --docker returntocorp/semgrep:develop --upload
+
+.PHONY: clean
+clean:
+	git clean -dfX
```
aryx, line 15: I don't like this ... because it perfoms it on the whole repo. I think I got file that I keep around deleted because
I did make clean from the wrong place one time.

perf/README.md
```diff
@@ -0,0 +1,55 @@
+Semgrep benchmarks
```
aryx, line 1: great readme

aryx (approved):
This is great!

aryx:
Seems to work, https://dashboard.semgrep.dev/metrics is populated.

perf/Makefile
mjambon: It's local, I just double-checked. Capital `X` ensures only the git-ignored files are removed, as opposed to lowercase `x` which will remove all files not tracked by git.

mjambon:
@aryx I need to ignore the exit status of semgrep. Right now it's nonzero if it finds a match, which kills the benchmarks.

perf/Makefile
aryx: yes but it removed my skip_list.txt ... which is my local gitignore

PR #2642, opened 2021-02-25
Script to automate the oncall email

aryx:
You first need to create a token to access github API.
Then you can create a "base" board with
./curl.shell $GITHUB_TOKEN > yesterday.json

test plan:
```
$ ./curl.sh $GITHUB_TOKEN > today.json
$ ./_build/default/oncall.exe -base yesterday.json today.json

To Discuss (UNASSIGNED)

	- [story] Figma can run `nginx` rules on their codebase and report that the performance was good
	<https://github.com/returntocorp/semgrep-app/issues/1736> ()
	- read timeout on downloading the semgrep config !PRIORITY:HIGH!
	<https://github.com/returntocorp/semgrep/issues/2620> ()

Unassigned To do

Assigned To do

	- Analyzing the entire Figma codebase with Figma's policies takes less than 5 minutes !PRIORITY:HIGH!
	<https://github.com/returntocorp/semgrep-app/issues/1727> (DrewDennison colleend)
	- Parser error when parsing obfuscated JavaScript (CLI and Online Rules Editor) !PRIORITY:HIGH!
	<https://github.com/returntocorp/enterprise/issues/572> (underyx)
	- Negation not working with subexpressions
	<https://github.com/returntocorp/semgrep/issues/2457> (aryx)
	- Clearly document that metavariables match all expressions, and how to exclude array assignment
	<https://github.com/returntocorp/semgrep/issues/2305> (brendongo)
	- metavariable-regex not working with pattern-inside?
	<https://github.com/returntocorp/semgrep/issues/2548> (mschwager)
	- Prevent users from naming two notifications the same thing _without_ sending a Sentry error
	<https://github.com/returntocorp/semgrep-app/issues/1724> (chmccreery)
	- Make deployment id and org name more visible in UI
	<https://github.com/returntocorp/semgrep-app/issues/1739> (chmccreery)

Closed since last email (good job guys!)

	- [Php] Add ellipsis nested statement feature support
	<https://github.com/returntocorp/semgrep/issues/2622> (aryx)
	- Go parsed with pfff does not do nested statement matching
	<https://github.com/returntocorp/semgrep/issues/2636> (aryx)
	- [MacOS] semgrep-core now uses libpcre but is not linked statically in macOS !PRIORITY:HIGH!
	<https://github.com/returntocorp/semgrep/issues/2624> (mjambon brendongo)
	- [Python] Constant tracking do not follow assignment operators
	<https://github.com/returntocorp/semgrep/issues/2616> (IagoAbal)
```

### DIFF
--- a/scripts/oncall-email/curl.shell
+++ b/scripts/oncall-email/curl.shell
@@ -1,4 +1,7 @@
-#!/bin/bash
+#!/bin/sh
+
+#usage: $ curl.shell $GITUB_TOKEN > yesterday.json
+#see README.md for more information.
 
 script='query {
   repository(owner:\"returntocorp\", name:\"semgrep\") {
--- a/scripts/oncall-email/curl.shell
+++ b/scripts/oncall-email/curl.shell
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+script='query {
+  organization(login:\"returntocorp\") {
+    project(number:50) {
+      columns(first:10) {
+       	nodes {
+          name
+       	  cards {
+       	    nodes {
+              content{
+                __typename
+                ...on Issue{
+                  title
+                  url
+                  assignees(first:3) {
+                     nodes {
+                      login
+                    }
+                  }
+                  closed
+                  createdAt
+                  updatedAt
+                  labels(first:5) {
+                    nodes{
+                      name
+                    }
+                  }
+                }
+              }
+              isArchived
+              note
+              creator {
+                login
+              }
+       	    }
+       	  }
+       	}
+      }
+	  }
+  }
+}'
+script="$(echo $script)"
+
+token=$1
+curl -H "Authorization: bearer $1" -X POST -d "{ \"query\": \"$script\" }" https://api.github.com/graphql | jq
--- a/scripts/oncall-email/curl.shell
+++ b/scripts/oncall-email/curl.shell
@@ -1,6 +1,26 @@
 #!/bin/bash
 
 script='query {
+  repository(owner:\"returntocorp\", name:\"semgrep\") {
+    name
+    issues(last:50) {
+      nodes{
+        title
+        url
+        assignees(first:3) {
+          nodes { login }
+        }
+        closed
+        createdAt
+        updatedAt
+        labels(first:5) {
+          nodes {
+            name
+          }
+        }
+      }
+    }
+  }
   organization(login:\"returntocorp\") {
     project(number:50) {
       columns(first:10) {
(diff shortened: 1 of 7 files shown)


mschwager:
IMO keeping the on-call email manual is advantageous because: 1.) it encourages the on-call engineers to stay up-to-date on all the on-call tasks, 2.) it provides a human element to the task, I think people will be less responsive to a bot barking orders at them, and 3.) it will almost always necessarily include details that a bot can never fully realize, e.g. customer details from Salesforce, a private Slack message update, cross-repo information, intra-team knowledge about who's best suited for a task, etc.

The initial purpose of the email was to remove a synchronous meeting to free up time for deep work for the rest of the team. IMO, one person taking 15min to gather the current day's on-call info and send to the team isn't unreasonable, and could be considered part of the on-call duties, but others may feel differently. We should do whatever the team feels is best. What do you think?

aryx:
I think it can still be manual, but with some help from the script to automate some of the tasks.
There's lots of untapped information in our github repo/issues/projects/teams/... that we could use with those kinds of scripts, for example flagging Assigned task in the customer dashboard that didn't get any update in the last few days. Scripts are great for that.

scripts/oncall-email/curl.shell
```diff
@@ -0,0 +1,46 @@
...
+                login
+              }
+       	    }
+       	  }
+       	}
+      }
+	  }
+  }
+}'
+script="$(echo $script)"
```
mjambon, line 66: Please explain what this script is for, for the sake of everyone who will stumble upon it.

You should also explain the intention of the line `script="$(echo $script)"`, which appears to be replacing whitespace by single spaces.

See https://github.com/mjambon/reasonable-shell-script for more guidelines on writing shell scripts.

scripts/oncall-email/oncall.ml
```diff
@@ -0,0 +1,193 @@
+open Common
```
mjambon, line 1: Please include a summary of what this file is about, like we all should do for all source files.

scripts/oncall-email/curl.shell
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
mjambon: `/bin/bash` won't pick up the version of bash installed by brew (I believe it's not installed in `/bin` where you'd find the antiquated bash 3.x shipped by Apple).

mjambon (changes requested):
see comments

tvfmmfvt:
Love

PR #2644, opened 2021-02-25
Get the latest semgrep docker image when running benchmarks from 'make'

mjambon:
This leaves control over which docker image to use when calling `run-benchmarks` directly (I'm assuming that's better).

### DIFF
--- a/perf/Makefile
+++ b/perf/Makefile
@@ -4,10 +4,12 @@
 #
 .PHONY: run
 run:
+	docker pull returntocorp/semgrep:develop
 	./run-benchmarks --docker returntocorp/semgrep:develop
 
 .PHONY: dummy
 dummy:
+	docker pull returntocorp/semgrep:develop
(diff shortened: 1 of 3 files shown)


emjin (approved):

perf/Makefile
```diff
@@ -4,10 +4,12 @@
 #
 .PHONY: run
 run:
+	docker pull returntocorp/semgrep:develop
```
aryx: but we also want to run bencmark with the local semgrep in your PATH no?
using the container is also slower.

aryx (approved):

perf/Makefile
mjambon: For using the local `semgrep`, you just run the script `./run-benchmarks` without options.

mjambon: I added clarifications to readme and makefile.

PR #2651, opened 2021-02-26
Add semgrep features to yaml

emjin:
Supports #2601

Recognize specific strings as "ellipses" and string literals with dollar signs as metavariables. Does not yet support deep expressions. Later, ... will need to be preprocessed and converted into fake ellipses.

Test plan:
Ensure that

```
__sgrep_ellipses__: __sgrep_ellipses__
rules:
    $X: __sgrep_ellipses__
```

parses to something corresponding to
```
...
rules:
    $X: ...
```

### DIFF
--- a/semgrep-core/Parsing/Parse_target.ml
+++ b/semgrep-core/Parsing/Parse_target.ml
@@ -298,7 +298,12 @@ let just_parse_with_lang lang file =
   | Lang.R ->
       failwith "No R parser yet; improve the one in tree-sitter"
   | Lang.Yaml ->
-      failwith "No Yaml parser yet, parsed only for semgrep in Parse_rule.ml"
+      let ch = open_in file in
+      let s = really_input_string ch (in_channel_length ch) in
+      close_in ch;
+      { ast = Yaml_to_generic.program {str = s; file};
+        errors = [];
+        stat = Parse_info.default_stat file}
 
 (*****************************************************************************)
 (* Entry point *)
--- a/semgrep-core/Parsing/other/yaml_to_generic.mli
+++ b/semgrep-core/Parsing/other/yaml_to_generic.mli
@@ -1,2 +1,7 @@
-val program: string -> AST_generic.program
+type env = {
+  str: string;
+  file: Common.filename
+}
+
+val program: env -> AST_generic.program
 val any: string -> AST_generic.any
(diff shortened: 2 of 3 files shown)


aryx:
In the test plan you can actually put the concrete command and the concrete output so we can see
that it actually parses to the right thing.

semgrep-core/Parsing/Parse_target.ml
```diff
@@ -298,7 +298,12 @@ let just_parse_with_lang lang file =
   | Lang.R ->
       failwith "No R parser yet; improve the one in tree-sitter"
   | Lang.Yaml ->
-      failwith "No Yaml parser yet, parsed only for semgrep in Parse_rule.ml"
+      let ch = open_in file in
```
aryx, line 301: You can use Common.read_file

semgrep-core/Parsing/Parse_target.ml
```diff
@@ -298,7 +298,12 @@ let just_parse_with_lang lang file =
...
+      let s = really_input_string ch (in_channel_length ch) in
+      close_in ch;
+      { ast = Yaml_to_generic.program {str = s; file};
```
aryx, line 304: would be better to read the file in Yaml_to_generic.program, so that the interface of Yaml_to_generic.program
remains simple (Common.filename -> AST_generic.program)

semgrep-core/Parsing/other/yaml_to_generic.mli
```diff
@@ -1,2 +1,7 @@
-val program: string -> AST_generic.program
+type env = {
+  str: string;
+  file: Common.filename
+}
+
+val program: env -> AST_generic.program
```
aryx, line 6: I'd rather have val program: Common.filename -> AST_generic.program.

semgrep-core/Parsing/other/yaml_to_generic.mli
```diff
@@ -1,2 +1,7 @@
-val program: string -> AST_generic.program
+type env = {
```
aryx, line 1: this should not be exposed in the mli. This is an internal data structure that you use to pass
around information in recursive function, but the interface of the module should remain simple.

aryx (approved):
I left some comments, but again you can fix them in another PR with other stuff.
Let's merge this.

PR #2652, opened 2021-02-26
Explicitly include or disinclude bloom filters

emjin:
Bloom filters are currently turned off by default, so this will actually test them as intended


mjambon (approved):
great, thanks

Issue #2646, opened 2021-02-25
Issue with ellipsis matching introduced in version v0.41.0


robinjmurphy:
**Describe the bug**

The tests for one of our semgrep rules have started failing since version v0.41.0 was released.

**To Reproduce**

Rule and tests code here: https://semgrep.dev/s/5QR0/

Running this against v0.40.0 (as is the current version in the playground at time of writing) works fine.

However, running with the v0.41.0 release (and v0.41.1):

```
❯ docker run --rm -v "${PWD}:/src" returntocorp/semgrep:6c8281bafe839bff9266397176b1658099b785d2 --quiet --test .
Traceback (most recent call last):
  File "/usr/local/bin/semgrep", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/semgrep/__main__.py", line 17, in main
    cli()
  File "/usr/local/lib/python3.9/site-packages/semgrep/cli.py", line 411, in cli
    semgrep.test.test_main(args)
  File "/usr/local/lib/python3.9/site-packages/semgrep/test.py", line 432, in test_main
    generate_file_pairs(
  File "/usr/local/lib/python3.9/site-packages/semgrep/test.py", line 318, in generate_file_pairs
    tested = {
  File "/usr/local/lib/python3.9/site-packages/semgrep/test.py", line 319, in <dictcomp>
    filename: score_output_json(
  File "/usr/local/lib/python3.9/site-packages/semgrep/test.py", line 203, in score_output_json
    raise Exception(
Exception: found results on ok'ed lines - lines={27} path=/src/wrongerrcheck/wrong-err-check.go
```

**Expected behavior**

The statement on line 27 should be fine. However, it is failing. It looks like this `pattern-not` is broken:

```yaml
    - pattern-not: |
        if $ERR_B, ... = <...$F(...)...>; $ERR_B != nil {
          ...
        }
```

Interestingly, if the second variable is changed from an underscore `_` to a 3 character variable name the rule works 🤔

i.e.

```go
	// ok:wrong-err-check
	if err, foo := something3(); err != nil {
		return
	}
```

**What is the priority of the bug to you?**
Is this a P0 (blocking your adoption of Semgrep or workflow), P1 (important to fix or quite annoying), P2 (regular bug that should get fixed)?

P1

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?

Docker. Also Python in CI.

I wonder if this bug was introduced in #2618?

[...]

PR #2657, opened 2021-02-26
Remove the ... optimisation which introduces some regressions

aryx:
This helps https://github.com/returntocorp/semgrep/issues/2646
but this is not the right fix.
The only difference is that we return multiple times the same thing
in semgrep-core, which should not matter, but it seems to matter
for the semgrep wrapper.
Note that semgrep-core -config is not affected by the PR.

test plan:
$ semgrep --debug --verbose --config
tests/OTHER/rules/regression_uniq_or_ellipsis.yaml tests/OTHER/rules/

returns only 1 match (like in 0.40)

### DIFF
--- a/semgrep-core/Core/Mini_rule.ml
+++ b/semgrep-core/Core/Mini_rule.ml
@@ -70,5 +70,6 @@ and severity = Error | Warning | Info
 (*s: type [[Rule.t]] *)
 (* alias *)
 type t = rule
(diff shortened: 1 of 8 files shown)


mschwager:
This should be superseded by https://github.com/returntocorp/semgrep/pull/2659

brendongo:
Closing since issue fixed by PR above

PR #2660, opened 2021-02-26
Ej 2601 yaml ellipses

emjin:
Complete basic support for semgrep ellipses and metavariables
    
Adjust the line number of tokens, convert ellipses, add tests

Supports #2601
    
Test plan: makes test

### DIFF
--- a/semgrep-core/Parsing/other/yaml_to_generic.ml
+++ b/semgrep-core/Parsing/other/yaml_to_generic.ml
@@ -37,12 +37,12 @@ let get_res = function
   | Result.Ok v -> v
 
 let tok (index, line, column) str file =
-  {Parse_info.token = Parse_info.OriginTok {str; charpos = index; line; column; file};
+  {Parse_info.token = Parse_info.OriginTok {str; charpos = index; line = line + 1; column; file};
    Parse_info.transfo = NoTransfo}
 
 let mk_tok {E.start_mark = {M.index; M.line; M.column}; _} str file =
   (* their tokens are 0 indexed for line and column, AST_generic's are 1 indexed for line, 0 for column *)
-  tok (index, line + 1, column) str file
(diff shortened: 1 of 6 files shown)


semgrep/semgrep/target_manager_extensions.py
```diff
@@ -33,6 +33,7 @@
 CSHARP_EXTENSIONS = [FileExtension(".cs")]
 RUST_EXTENSIONS = [FileExtension(".rs")]
 KOTLIN_EXTENSIONS = [FileExtension(".kt")]
+YAML_EXTENSIONS = [FileExtension(".yaml")]
```
aryx, line 36: The the comment above about test_rule_parser.py --snapshot-update
This is why the CI is failing.

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -12,11 +12,6 @@ exception ParseError of string
  * low-level Stream API, which we parse into a generic AST
 *)
 
-type env = {
```
aryx, line 15: I think you should keep this type, because it would be easier to extend.
I just didn't want to expose this internal type in the .mli. This is internal details that users
of the module should not care about.

aryx (approved):

aryx:
Just fix the CI and merge

semgrep-core/Parsing/other/yaml_to_generic.ml
emjin: If I don't put it in the mli, how do I create the env to pass in when I call the function in parse_target?

aryx: This is what I had in mind: https://github.com/returntocorp/semgrep/pull/2670/files

PR #2679, opened 2021-03-03
[WIP] Filtering irrelevant rules using regexps by analyzing the whole rule

aryx:
This is similar to the filtering of irrelevant patterns I was doing
in Mini_rules.filter.ml, but analyzing the whole rule, not individual
patterns to skip entire rules. By also making the generic AST parsing
lazy, we can even sometimes skip parsing a file.

Right now the regexp is hardcoded for the rule used to match
big-js, but subsequents commits will fix that.

test plan:
$ time yy -profile -debug -lang js -config bench/big-js/semgrep_slow.yml
bench/big-js/
=> 5m27
$ time yy -profile -filter_irrelevant_rules -debug -lang js -config bench/big-js/semgrep_slow.yml bench/big-js/
=> 0.6s

### DIFF
--- a/semgrep-core/CLI/Main.ml
+++ b/semgrep-core/CLI/Main.ml
@@ -499,7 +499,7 @@ let filter_files_with_too_many_matches_and_transform_as_timeout matches =
     )
   in
   new_matches, new_errors
-[@@profiling]
+[@@profiling "Main.filter_too_many_matches"]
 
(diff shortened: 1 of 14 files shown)


aryx:
This PR is just the skeleton code around Analyze_rule.regexp_of_formula function.
I'll implement this function in another PR.

PR #2682, opened 2021-03-03
Settle on one of 'vars', 'metavars', 'metavariables'

mschwager:
I can't keep all these different naming schemes straight in my head.

### DIFF
--- a/semgrep/semgrep/rule_match.py
+++ b/semgrep/semgrep/rule_match.py
@@ -79,10 +79,6 @@ class RuleMatch:
         return self._extra
 
     @property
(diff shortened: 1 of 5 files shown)


semgrep/semgrep/rule_match.py
```diff
@@ -78,10 +78,6 @@ def path(self) -> Path:
     def extra(self) -> Dict[str, Any]:
         return self._extra
 
-    @property
-    def metavars(self) -> Dict[str, Any]:
```
mschwager, line 82: Unused

aryx (approved):

PR #2684, opened 2021-03-03
Add "rust" to list of supported languages

amietn:
In addition to "rs", it should also be possible to use "rust".


ievans (approved):
@amietn thanks for making your first contribution to semgrep! lgtm

PR #2689, opened 2021-03-03
Add metavariable propagation within 'patterns'

### LINKED ISSUE #2548: metavariable-regex not working with pattern-inside?

mschwager:
Fixes #2548.

I tried to make this as minimal as possible. Our evaluator implementation is a bit difficult to wrestle with this functionality. My first thought was to build propagation on top of the `PatternMatch`'s, but that's a difficult approach because they're treated as the "source of truth." I.e. they're used at the end to gather relevant pattern match information, so recursively mutating them would be difficult. That and they're not returned at the end, like the output `Range`'s. So I decided to build it on top of the `Range`'s recursively passed around and returned at the end.


semgrep/tests/e2e/snapshots/test_debugging_json/test_debugging_json/results.json
```diff
@@ -12,13 +12,15 @@
               [
                 72,
                 174,
+                {},
```
mschwager, line 15: These changes will require updating Semgrep App [here](https://github.com/returntocorp/semgrep-app/blob/b8722ed/frontend/src/types.ts#L135)

semgrep/tests/e2e/snapshots/test_message_interpolation/test_message_interpolation/rulesmessage_interpolationmulti-pattern-inside.yaml-message_interpolationmulti_pattern_inside_nested.py/results.json
```diff
@@ -10,7 +10,7 @@
       "extra": {
         "is_ignored": false,
         "lines": "            print(\"blah\")",
-        "message": "'print(\"blah\")' detected in method 'B' in class 'A'",
+        "message": "'print(\"blah\")' detected in method 'C' in class 'A'",
```
mschwager, line 13: Looking at this rule, this message feels "more correct" to me, but that's open to interpretation :man_shrugging:

mschwager: Some additional context: the `stabilize_evaluation_ordering` function above was created to address ordering, and thus evaluation, differences in different Python versions (Python 3.7+ orders dicts/sets internally by default now, which affects our evaluation). We can flip this back by setting `reverse=False`, however, I think tighter bounds on which code block we're inside (function, class, etc) will provide a better UX. It's easier to grok a finding referring to the innermost block vs. the outermost block which could be many lines away.

semgrep/semgrep/evaluation.py
```diff
@@ -347,28 +429,27 @@ def evaluate(
...
-            all_pattern_match_metavariables[metavar_text].append(pattern_match)
+    propagated_metavariable_lookup = {
+        _range: {
+            metavariable: pm.get_metavariable_value(metavariable)
+            for pm in pattern_matches
+            for metavariable in _range.propagated_metavariables
+            if compare_propagated_metavariable(_range, pm, metavariable)
+        }
+        for _range in valid_ranges_to_output
+    }
```
minusworld, line 440: w00t \o/

minusworld (commented):
Cool, this looks great! 👍  I looked over the code and nothing stood out to me as something that needs to be addressed.

I do, however, have one ask: The matching is getting more and more complicated. Would you be amenable to adding docstring comments to the functions you added? That's
- `stabilize_evaluation_order`
- `compare_propagated_metavariable`
- `filter_ranges_with_propagation`

brendongo:
Nice! +1 on function level comments. Should we make sure the server is updated and released before merging this?

mschwager:
I can add some function-level comments :+1: , but generally my take on comments is:

> [Don’t comment bad code—rewrite it.](https://www.goodreads.com/quotes/857949-don-t-comment-bad-code-rewrite-it)

There's definitely a time and place for comments when decisions are non-obvious, but they often cause more harm then good IMO. Comments often:

- Are flat out incorrect. The code has changed and someone forgot to update the comment, the code has moved around and is now distant from its corresponding comment, the comment is stale and no longer applies to the code, etc. Comments can lie, code can't.
- Waste lots of space providing redundant information. Most comments I see re-describe "how" something is done rather than "why." I can see how it's done by looking at the code, whereas it's often more helpful to explain why a non-obvious decision was made.
- Are less useful than the git blame, corresponding PR, corresponding issue, anyway. Rarely does a comment answer a question I have about a particular piece of code. I typically find the git blame/PR/issue to be much more enlightening (and here we are, hi :wave:).

The Comments chapter in [Clean Code](https://www.oreilly.com/library/view/clean-code-a/9780136083238/) (Chapter 4) does a much better job explaining the intricacies here than I can. Actually, that whole book is gold, highly recommend.

> Should we make sure the server is updated and released before merging this?

Yeah, I'm not sure about the best way to maintain backwards compatibility here. The only backwards incompatible change is in the debugger functionality, which I don't think many external users are using. We could either update the server when we bump the default Semgrep version to the latest version, or we could bump it now and try to deal with both the old and new format for `--debugging-json`.

mschwager:
Okay, I think this is actually the perfect example of comment redundancy. What kind of function-level comments are you thinking?

```python
def filter_ranges_with_propagation(
    ranges_left: Set[Range],
    ranges_for_pattern: Set[Range],
    predicate: Callable[[Range, Range], bool],
    metavariable_propagation: bool,
) -> Set[Range]:
    ...
```

Something like:

```python
"""
Filter ranges based on current pattern's results and predicate. Propagate metavariables, if enabled.
"""
```

My question is, what does this provide on top of the function name, argument names, and type information? I don't feel too strongly here, so I'm happy to add these comments, but I don't feel that it contributes much.

On the other hand, I feel like `stabilize_evaluation_ordering` is non-obvious, which is why I pulled it out into its own function instead of calling `sorted` in multiple places. My thought is that the function name is attempting to communicate something non-obvious vs. adding comments. Does that make sense? I can move that comment to a docstring if it makes a difference.

minusworld:
I think that is a good principle in general. However, this is something that's been extended by many people with a lot of types and naming conventions (some of which change in the middle of the algorithm). I think I'm mainly just hoping for a written description of the "intention" of the function. Or said another way, of the "step" in the process. Anyway, my 2c.

As for the redundant comment above, it does help me understand better the purpose of the function. 🤷‍♂️  Maybe I'm weird like that.

brendongo:
I find comments useful to know what the intention of the function is so someone in the future trying to debug or extend are aware of what the function was assumed to do in the past. So for `compare_propagated_metavariable` for example, should this only be called if the caller is sure metavariable is in both relevant dictionaries. Is raising an exception there documented behavior or is that something that I can mess with since it isn't?

brendongo:
That being said, we're probably absorbing a lot of evaluation.py functionality into semgrep-core in the next 3 months so don't feel too strongly re: comments here so just pre-approving too.

brendongo (approved):

minusworld:
> That being said, we're probably absorbing a lot of evaluation.py functionality into semgrep-core in the next 3 months so don't feel too strongly re: comments here so just pre-approving too.

Good point - I'll add my stamp for posterity 😆

minusworld (approved):

mschwager:
The finer points of code commenting is probably the ultimate [bikeshed](https://en.wiktionary.org/wiki/bikeshedding), but if we do want to enforce something like that it'd probably be best to use something like [this](http://pylint-messages.wikidot.com/messages:c0111) instead of relying on human involvement :man_shrugging:

aryx:
I think we should have forced the user to rewrite its rule to associate the metavariable-regexp to the right patterns, with the appropriate nested patterns:

PR #2700, opened 2021-03-04
Generate pattern which will match targets

emjin:
Given two or more targets, generate a pattern by creating potential patterns in layers. 

For example, if our targets are 
` foo(m, n)` and `foo(bar(2), 3)`, we will try the patterns: 

`...`, `$X(...)`, `foo(...)`, `foo($X, ...)`, `foo(m, ...)` vs `foo($X(...), ....)`, `foo($X, $Y)`, and `foo($X, n)` vs `foo($X, 3)`. Since the last one is terminal, we will return `foo($X, $Y)` as our generated pattern.

### DIFF
--- a/semgrep-core/CLI/Main.ml
+++ b/semgrep-core/CLI/Main.ml
@@ -1213,7 +1213,7 @@ let all_actions () = [
   Common.mk_action_2_arg Test_synthesizing.expr_at_range;
   "-synthesize_patterns", " <l:c-l:c> <file>",
   Common.mk_action_2_arg Test_synthesizing.synthesize_patterns;
-  "-generate_patterns", " <file>",
+  "-generate_patterns", " <l:c-l:c>+ <file>",
   Common.mk_action_n_arg Test_synthesizing.generate_pattern_choices;
 
   "-stat_matches", " <marshalled file>",
--- a/semgrep-core/CLI/Main.ml
+++ b/semgrep-core/CLI/Main.ml
@@ -1213,6 +1213,8 @@ let all_actions () = [
   Common.mk_action_2_arg Test_synthesizing.expr_at_range;
   "-synthesize_patterns", " <l:c-l:c> <file>",
   Common.mk_action_2_arg Test_synthesizing.synthesize_patterns;
+  "-generate_patterns", " <file>",
+  Common.mk_action_n_arg Test_synthesizing.generate_pattern_choices;
 
   "-stat_matches", " <marshalled file>",
   Common.mk_action_1_arg stat_matches;
--- a/semgrep-core/Synthesizing/Pattern_from_Targets.ml
+++ b/semgrep-core/Synthesizing/Pattern_from_Targets.ml
@@ -136,7 +136,7 @@ let lookup_pattern pattern s =
 let set_prev { prev = _; count; mapping } prev' = { prev = prev'; count; mapping }
 
 (*****************************************************************************)
-(* Algorithm *)
+(* Pattern generation *)
 (*****************************************************************************)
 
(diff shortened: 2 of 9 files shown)


semgrep-core/CLI/Main.ml
```diff
@@ -1213,6 +1213,8 @@ let all_actions () = [
   Common.mk_action_2_arg Test_synthesizing.expr_at_range;
   "-synthesize_patterns", " <l:c-l:c> <file>",
   Common.mk_action_2_arg Test_synthesizing.synthesize_patterns;
+  "-generate_patterns", " <file>",
```
aryx: it also takes a list of ranges no? so the comment string should be <l:c-l:c>+ <file>

semgrep-core/Synthesizing/Pattern_from_Targets.ml
```diff
@@ -0,0 +1,381 @@
...
+open Common
+module Set = Set_
+
+exception InvalidSubstitution
+exception UnsupportedTargetType
+
+(*****************************************************************************)
+(* Prelude *)
+(*****************************************************************************)
+(* The main target intersection algorithm.
```
aryx: Maybe you can add some high-level explanations of the algorithm, and how it differs from Pattern_from_code

semgrep-core/Synthesizing/Pattern_from_Targets.ml
```diff
@@ -0,0 +1,381 @@
...
+ *
+ * related work:
+ *  - coccinelle spinfer?
+*)
+
+(*****************************************************************************)
+(* Types *)
+(*****************************************************************************)
+
+type stage = DONE | ANY of any | LN of any
```
aryx: Some comments, what are they useful for? what LN stands for?

semgrep-core/Synthesizing/Pattern_from_Targets.ml
```diff
@@ -0,0 +1,381 @@
...
+type env = { prev : any; count : int; mapping : (expr * expr) list; }
+type pattern_instrs = (env * any * ((stage * ((any -> any) -> any -> any)) list)) list
```
aryx: this type is quite complicated. At least give an alias to the 3rd element, with a good
name so we can get an idea of what it is useful for.

semgrep-core/Synthesizing/Pattern_from_Targets.mli
```diff
@@ -0,0 +1,40 @@
+
+(*
+ * The pattern_from_code feature allows users to highlight code and
```
aryx: Looks like a comment for Pattern_from_code.mli, not Pattern_from_targets.

semgrep-core/Synthesizing/Synthesizer.mli
```diff
@@ -1,2 +1,4 @@
 (* range as "start row:start col-end row:end col" -> filename -> (label * pattern) list *)
 val synthesize_patterns: string -> string -> (string * string) list
+
```
aryx, line 4: a one-line description comment, and how it differs from syntesize_patterns.

aryx (changes requested):

semgrep-core/Synthesizing/Pattern_from_Targets.mli
```diff
@@ -0,0 +1,63 @@
+
+(*
+ * The pattern_from_targets feature allows users to highlight separate code
```
aryx, line 3: Great comment! I think it's better to keep the long detailed explanation in the .ml and just comments on the interface and general purpose in the .mli.

aryx (approved):

PR #2703, opened 2021-03-05
Make rule schema a bit nicer

underyx:
Mainly adds titles, but also simplifies some `anyOf` stuff to use `oneOf` instead.


semgrep/semgrep/rule_schema.yaml
```diff
@@ -127,6 +135,7 @@ definitions:
     type: object
     properties:
       pattern-not:
+        title: (deprecated, use pattern-not-inside) Do not return finding where Semgrep pattern matches exactly
```
mschwager: Is this deprecated? I don't think we have that language anywhere else

underyx: Discussed many times before, thought I'd just document it as such and let a discussion start here :D

underyx: cc @dlukeomalley

mschwager: I suspect that discussion will greatly slow down these unrelated changes. We'll probably need to figure out a deprecation plan, docs updates, etc, if we do in fact want to deprecate `pattern-not`. If you want to move quickly with these changes it'd probably be easiest to keep `pattern-not` deprecation separate :man_shrugging:

semgrep/semgrep/rule_schema.yaml
```diff
@@ -3,6 +3,7 @@ $schema: http://json-schema.org/draft-07/schema#
 definitions:
   patterns-content:
     type: array
+    title: "Return finding where all of the nested conditions are true"
```
mschwager, line 6: Nit: this one has quotes for `title` and the others don't, not sure if this is intentional

semgrep/semgrep/rule_schema.yaml
```diff
@@ -71,6 +72,7 @@ definitions:
     type: object
     properties:
       pattern:
+        title: Return finding where Semgrep pattern matches exactly
```
mschwager, line 75: Another nit: we could ape the descriptions from the [docs schema table](https://semgrep.dev/docs/writing-rules/rule-syntax/#schema) if that's easier/more consistent

mschwager (approved):

aryx:
ok merging then, it's been a while.

PR #2705, opened 2021-03-05
Add --vim flag for Syntastic

mschwager:
Before moving forward with this, how about we combine this with `--emacs` and use a `--oneline` argument for both? That way we can avoid the CLI flag explosion for all output formats. Looks like `--emacs` came from https://github.com/returntocorp/semgrep/pull/2495. @Ruin0x11 does that seem reasonable to you?

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -263,6 +263,11 @@ def cli() -> None:
         action="store_true",
         help="Output results in Emacs single-line format.",
     )
+    output.add_argument(
+        "--vim",
+        action="store_true",
+        help="Output results in vim single-line format.",
+    )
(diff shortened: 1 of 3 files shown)


Ruin0x11:
The pattern output for `--emacs` specifically matches a certain regex that is shared by the GCC compiler, which lets Emacs change the color of a line depending on if the match was a warning or error.

If it's changed it would probably break out of the box. But if the output of `--emacs` needs to be changed it's not an insurmountable problem. For the Emacs plugin I could add a new regex that matches a different syntax, but it would capture less info.

Personally, I would prefer if they stayed separate.

aryx (approved):

aryx:
I agree with ruin, better not mix vim and emacs :)

PR #2708, opened 2021-03-06
Parse local function declarations

Sjord:
I.e. a function defined within a method.

### DIFF
--- a/semgrep-core/Parsing/tree_sitter/Parse_csharp_tree_sitter.ml
+++ b/semgrep-core/Parsing/tree_sitter/Parse_csharp_tree_sitter.ml
@@ -1599,6 +1599,7 @@ and statement (env : env) (x : CST.statement) =
        let v1 = List.map (modifier env) v1 in
        let v2 = return_type env v2 in
        let v3 = identifier env v3 (* identifier *) in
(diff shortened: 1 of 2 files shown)


semgrep-core/Parsing/tree_sitter/Parse_csharp_tree_sitter.ml
```diff
@@ -1609,7 +1610,20 @@ and statement (env : env) (x : CST.statement) =
...
-       todo env (v1, v2, v3, v4, v5, v6, v7)
+       let tparams = type_parameters_with_constraints v4 v6 in
+       let idinfo = empty_id_info () in
+       let ent = {
+         name = EN (Id (v3, idinfo));
+         attrs = v1;
+         tparams;
+       } in
+       let def = AST.FuncDef {
+         fkind = (AST.Method, tok);
```
aryx, line 1621: should be AST.Function maybe?

aryx: Not that it really matters; it's not used by the engine right now.

aryx (approved):

Issue #1218, opened 2020-07-08
Merge pattern-inside and pattern, pattern-not-inside and pattern-not


ievans:
As discussed by @nbrahms : we haven't seen any instances where anyone would want to use pattern rather than pattern-inside. We could just alias the keys and stop advertising pattern-inside and pattern-not-inside.

mschwager:
Unassigning myself in case someone can get to this before I can.

DrewDennison:
cc @AzureUmbra

ievans:
Go away stale bot

dlukeomalley:
@ievans Adding a `priority:medium` label, thanks for resurrecting this ticket. It has continued to cause confusion for new (and existing) users and is worth fixing near-term (rather than "future").

aryx:
Actually just found an example where we can't merge pattern-not and pattern-not-inside: https://semgrep.dev/s/PJ7Y/?stepByStep=1

We could merge pattern: and patter-inside, but for negation the difference matters :(

aryx:
Here are example of semgrep-rules that now works if you treat pattern-not and patter-not-inside differently:
Great: a test file now works: ./javascript/browser/security/new-function-detected.yaml
Great: a test file now works: ./python/django/best-practice/upsell_django_environ.yaml
Great: a test file now works: ./python/django/security/audit/xss/template-translate-as-no-escape.yaml
Great: a test file now works: ./python/django/security/audit/xss/template-translate-no-escape.yaml
Great: a test file now works: ./python/django/security/injection/path-traversal/path-traversal-file-name.yaml
Great: a test file now works: ./python/django/security/injection/path-traversal/path-traversal-join.yaml
Great: a test file now works: ./ruby/lang/security/model-attributes-attr-accessible.yaml

For some examples we could rewrite the test file so that the difference does not matter (many examples use multiple times the same function or class which leads to more matching that it should), but not sure we can always do that.

aryx:
Here is another example where merging pattern and pattern-inside would lead to a different behavior:
https://semgrep.dev/s/W8w4/?stepByStep=1

aryx:
the problem is that And is not commutative anymore with pattern-inside and pattern. The evaluation order of
pattern-inside:x AND pattern-inside:y AND pattern:z  matters. We currently rely on
evaluatin first pattern:z and then comparing it with x, and y.

ievans:
@aryx should I close this as #wontfix and make a new issue to improve our documentation or UX?

[...]

PR #2718, opened 2021-03-09
[wip] Alias pattern-not to pattern-not-inside

minusworld:


ievans:
fix for https://github.com/returntocorp/semgrep/issues/1218 ?

ievans:
@aryx is this obviated by the work you are doing in the core? can you fix #1218 there?

aryx:
yes, semgrep-core already treats pattern and pattern-inside as the same.

aryx:
actually I've found recently examples in our rules preventing us from merging pattern and pattern-inside:.

PR #2722, opened 2021-03-09
Use globals instead of labels for optimization flags

aryx:
They are annoying to thread around. Globals are for non-functional
properties (e.g., optimizations).

test plan:
make test

### DIFF
--- a/semgrep-core/CLI/Main.ml
+++ b/semgrep-core/CLI/Main.ml
@@ -83,6 +83,7 @@ let debug = ref false
 (*e: constant [[Main_semgrep_core.debug]] *)
 let profile = ref false
 
+(* report matching times per file *)
 let report_time = ref false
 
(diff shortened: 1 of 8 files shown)


mjambon:
I understand the desire to not pass around a gazillion parameters and also consistency.

The problem with mutable globals is the same as usual:
* they're mutable so it's not clear what is initialized and when,
* it's not clear which settings are available, compared to having them all in a record

What about a single `conf` param that holds the result of parsing the command line? 

The `conf` type would be defined in a separate module so that most other modules in the application can depend on it. We could also make the `conf` object global if we don't want to pass a `conf` parameter everywhere. It would be initialized once by the command line (we can ensure the initialization happens only once, if necessary). Compared to the current solution, this is just one mutable object to worry about, and it's not really mutable anyway, it's only a late initialization.

PR #2723, opened 2021-03-09
Adding a new config parameter for the matching engine

aryx:
This is the first step towards enabling/disabling individual
features on a per-rule or per-pattern basis, e.g., disabling
constant propagation for some rules.

test plan:
make test

### DIFF
--- a/semgrep-core/Matching/Generic_vs_generic.ml
+++ b/semgrep-core/Matching/Generic_vs_generic.ml
@@ -30,6 +30,7 @@ module B = AST_generic
 module MV = Metavariable
 module AST = AST_generic
 module Flag = Flag_semgrep
+module Config = Config_semgrep
 module H = AST_generic_helpers
 
 (* optimisations *)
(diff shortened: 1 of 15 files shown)


semgrep-core/Matching/Generic_vs_generic.ml
```diff
@@ -598,11 +599,11 @@ and m_expr a b =
    * const a = "foo"; ... a == "foo" would be catched by $X == $X.
   *)
   | A.L(a1), b1 ->
-      (match Normalize_generic.constant_propagation_and_evaluate_literal b1 with
-       | Some b1 ->
-           m_literal_constness a1 b1
-       | None -> fail ()
-      )
+      if_config (fun x -> x.Config.constant_propagation)
```
aryx, line 602: @IagoAbal we can now enable/disable if we want.

aryx: Remains to hook this up in the rule format to let user enable/disable equivalences/features.

IagoAbal: Great!

IagoAbal (approved):
Looks great to me.

minusworld:
@aryx It looks like this introduces a Config object. How would I enable/disable constant propagation using `semgrep-core`?

aryx:
we need to add a field in the rule to allow enable/disable individual features. This will be easier once we do the "big refactoring" in evaluation.py where we just pass the whole rule file to semgrep-core.

PR #2727, opened 2021-03-10
Convert sgrep ellipses to yaml parseable ellipses

emjin:
The user will input a file like

```
- id
   ...
   - language: ocaml
   ...
```

The yaml parser won't understand this, so it needs to be converted to

```
- id
 - __sgrep_ellipses__: __sgrep_ellipses__
 - language: ocaml
 - __sgrep_ellipses__: __sgrep_ellipses__
```

To do this, we do a preprocessing pass where we replace ... with __sgrep_ellipses__, adding a - if possible

Test plan:
make test with new tests

Also, run `semgrep-core -dump_pattern tests/yaml/semgrep/ellipses_stress.sgrep -lang yaml` to test ellipses conversion on different cases


semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -172,15 +176,135 @@ let parse env parser : A.expr =
...
+  String.concat "\n" (process_lines lines [] [])
+
+
 (*****************************************************************************)
 (* Entry points *)
 (*****************************************************************************)
 
 let program file =
-  let str = Common.read_file file in
+  let str = preprocess_yaml (Common.read_file file) in
```
aryx, line 302: you should not need that for the target file. Only the pattern can contain ellipsis.

semgrep-core/tests/yaml/dots_list.sgrep
```diff
@@ -0,0 +1,5 @@
+...
```
aryx, line 1: nice!

semgrep-core/Parsing/other/yaml_to_generic.ml
```diff
@@ -172,15 +176,135 @@ let parse env parser : A.expr =
   in
   read_stream ()
 
+let make_pattern_expr e =
+  match e with
+  | A.Container(A.Dict, (_lp, [x], _rp)) -> A.E x
```
aryx, line 181: comment here: "if a pattern contains just a field and a value, the user probably want to only
match this field, not the whole enclosing container".
I guess we don't need the PartialSingleField trick. Cool.

semgrep-core/tests/yaml/semgrep/ellipses_stress.sgrep
```diff
@@ -0,0 +1,22 @@
+rules:
```
aryx, line 1: This files does not have a corresponding .yaml? Maybe put it in tests/yaml/semgrep/pattern/

aryx (approved):
great!

semgrep-core/tests/yaml/semgrep/ellipses_stress.sgrep
aryx: oh sorry, you actually did that :)

PR #2729, opened 2021-03-10
Move semgrep-core sources to semgrep-core/src

mjambon:
image

If this disrupts other people's work, I can redo this during the weekend. It takes 5 minutes. The only thing to be careful about is to move the submodules one by one with `git mv tree-sitter-lang/semgrep-foo ...`, rather than `git mv tree-sitter-lang ...` (the latter fails to change the paths in `.gitmodules`).


aryx:
ok, let's do it, but this weekend, and after every PRs affecting semgrep-core has been merged.

mjambon:
sounds good to me!

Issue #2704, opened 2021-03-05
Parse Java symbols containing accented characters and the new record extension


Floby:
**Describe the bug**
Semgrep fails to parse java code because of diacritics.

**To Reproduce**
Import or declare a symbol containing an accented character.
```java
class Tireur {
  private string cibleVisée;
}
```

**Expected behavior**
semgrep reads the symbol into its AST as javac does.

**What is the priority of the bug to you?**
P0 as our code is standardised to use french words, containing accents where needed.

**Environment**
semgrep@0.41.1 installed via pip on ubuntu 18.04

ievans:
Hello @Floby, welcome and thanks for creating an issue with so much detail! 

I believe that there is a bug here but I'm not able to reproduce. I tried on the live editor https://semgrep.dev/s/6n4B and on a local test case with your reproduction having the accent aigu. 

```
ine@imbp4 ~/D/r/t/synthetic> semgrep -e 'class $X { ...  }' --lang=java .
fr.java
1:class Tireur {
2:  private string cibleVisée;
3:}
ran 1 rules on 1 files: 1 findings
```

Can you provide a playground editor (https://semgrep.live) snippet that fails to parse? or more detail about your local setup?

cc @nbrahms or @mjambon who might have more details

Floby:
OK, my bad for not testing this code in the sandbox.

this one indicates an error on the accented character (https://semgrep.dev/s/2xeD)

```java

public class Usecase {
    public Usecase() {
    }

    public void exécuter(Commande commande) {
    }

    public static record Commande(String param) {
        public static Commande aPartirDe(Chose chose) {
            return new Commande(chose.enParam());
        }
    }
}
```
However, removing the `record` statement makes semgrep able to parse the file. `record` is still experimental so it makes sense that semgrep does not parse it, but it is still unexpected for the parse error to be reported elsewhere :)

aryx:
I think the issue is that we use 2 parsers for Java, the pfff one and tree-sitter one, and both choke on it, but then we report the parse error of pfff.

aryx:
We probably now should use first the tree-sitter Java parser instead of pfff one, so the error reported would be the one of tree-sitter.
Also, tree-sitter-java can probably parse partially this file by skipping the record keyword, so @ievans we probably need a way to tell semgrep-app of partial parse errors for the UI to not stop but run the pattern (and possibly highlight a problematic skipped token).

Floby:
[...]

PR #2733, opened 2021-03-10
[Java] use latest pfff where the AST for names and Dot change

aryx:
This gets pfff and tree-sitter java parser closer to each other.

test plan:
$ semgrep-core -diff_pfff_tree_sitter field.java
does not show any difference


aryx:
This will help https://github.com/returntocorp/semgrep/issues/2704 by allowing us to switch to tree-sitter first for parsing java.

PR #2741, opened 2021-03-11
[JSON] transform the Record in a Dict

aryx:
This will help apply JSON patterns on YAML files by being
consistent in how we treat objects.

test plan:
make test

### DIFF
--- a/semgrep-core/Parsing/pfff/json_to_generic.ml
+++ b/semgrep-core/Parsing/pfff/json_to_generic.ml
@@ -32,11 +32,15 @@ let expr x =
             | G.FieldStmt
                 ({s = G.DefStmt ({G.name = G.EN (G.Id (id, _)); _},
                                  FieldDefColon ({vinit = Some e; _})); _}) ->
-                id, e
-            | _ -> failwith (spf "not a JSON field: %s" (G.show_expr e))
+                Left (id, e)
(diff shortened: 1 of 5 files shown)


semgrep-core/Parsing/pfff/json_to_generic.ml
```diff
@@ -13,19 +12,45 @@
...
+      | Record (lp, xs, rp) ->
+          let ys = xs |> List.map (function
+            | G.FieldStmt
+                ({s = G.DefStmt ({G.name = G.EN (G.Id (id, _)); _},
+                                 FieldDefColon ({vinit = Some e; _})); _}) ->
+                id, e
+            | _ -> failwith (spf "not a JSON field: %s" (G.show_expr e))
+          ) in
+          let zs = ys |> List.map (fun (id, e) ->
+            G.Tuple (G.fake_bracket [G.L (G.String id); e])) in
```
emjin: Don't JSON records have real brackets we can use?

emjin (approved):

semgrep-core/Parsing/pfff/json_to_generic.ml
aryx: I use them, see the use of lp and rp above; here it's for a single field, like foo: "bar", which we convert internally in a Tuple (we could use the token for ":" for the bracket ...).

PR #2760, opened 2021-03-15
* semgrep-core/semgrep.opam: switch to easy_logging.git

aryx:
Path towards using comby in semgrep, which currently has conflicts
with easy_logging because of ppx_deriving < 5 constraint

test plan:
make test, and wait for CI to work


aryx:
actually this introduces huge performance regressions.

aryx:
with 4.09.1, after this semgrep-core takes 1.1s at startup, just doing semgrpe-core -version takes 1.1s.
I used strace to look at what it does and it's weird, it's opening ocaml/ld.conf, and does lots of read.
weird.

aryx:
seems to be related with OCAMLRUNPARAM=b

aryx:
I'm reverting, spent the whole morning trying to figure out why. Weird.

PR #2769, opened 2021-03-16
Fix symlink to bin/semgrep-core + explain -max_memory

mjambon:


semgrep-core/src/cli/Main.ml
```diff
@@ -1327,7 +1337,10 @@ let options () =
     "-timeout", Arg.Set_float timeout,
     " <float> time limit to process one input program (in seconds); 0 disables timeouts (default is 0)";
     "-max_memory", Arg.Set_int max_memory,
-    " <int> maximum memory to use (in MB)";
+    " <int> maximum memory available (in MB); allows for clean termination \
+     when running out of memory. This value should be less than the actual \
```
aryx, line 1341: You cant have long explanation here with Arg i think. The layout will be wrong

PR #2772, opened 2021-03-17
Add benchmarks to cover semgrep advertised packs

emjin:
Use repositories from https://github.com/returntocorp/semgrep/blob/develop/semgrep/tests/public_repos.py to benchmark packs advertised on semgrep.dev, most notably r2c-ci and r2c-security-audit.

Also, add a --internal flag, similar to dummy, for internal-only tests


perf/run-benchmarks
```diff
@@ -75,7 +85,6 @@ SEMGREP_VARIANTS = [
     SemgrepVariant("no-cache", "-bloom_filter -no_opt_cache"),
     SemgrepVariant("max-cache", "-bloom_filter -opt_max_cache"),
     SemgrepVariant("no-bloom", "-no_bloom_filter"),
-    SemgrepVariant("no-gc-tuning", "-bloom_filter -no_gc_tuning"),
```
mjambon: Did you mean to remove "no-gc-tuning"?

mjambon (approved):
Looks good to me.

I think we need to include a description of what each test is about, i.e. things like language, rule set, expectations, importance, and anything special about them. If you already know these things, please put them to the python script as comments. In the future we should print their description as part of the benchmark logs or add a `--describe` option/subcommand.

perf/run-benchmarks
```diff
@@ -169,13 +179,9 @@ def run_benchmarks(docker: str, dummy: bool, upload: bool) -> None:
                 metric_name = ".".join([name, "duration"])
                 print(f"------ {name} ------")
                 duration = run_semgrep(docker, corpus, variant)
-                msg = f"{metric_name} = {duration:.3f} s"
-                print(msg)
-                results.append(msg)
+                print(f"{metric_name} = {duration:.3f} s")
                 if upload:
                     upload_result(metric_name, duration)
-    for msg in results:
```
aryx: Can you also put that back? I think you got a merge problem, you deleted martin's last modifications to this file.

perf/run-benchmarks
```diff
@@ -50,10 +47,23 @@ CORPUSES = [
         "njsbox", "input/njsscan/njsscan/rules/semantic_grep", "input/dropbox-sdk-js"
     ),
     Corpus("zulip", "input/semgrep.yml", "input/zulip"),
+    Corpus("apache", "input/django.yml", "input/libcloud"),
+    Corpus("dropbox", "input/flask.yml", "input/pytest-flakefinder"),
+    Corpus("coinbase", "input/rules", "input/bifrost"),
+    Corpus("netflix", "input/rules", "input/lemur"),
+    Corpus("draios", "input/rules", "input/sysdig-inspect"),
+    Corpus("0c34", "input/golang.yml", "input/govwa"),
```
aryx, line 65: what does oc34 mean?

perf/bench/0c34/prep
```diff
@@ -0,0 +1,38 @@
...
+# rule_dir: input/golang.yaml
+# target_dir: input/govwa
+#
+# Uses sh because bash is not installed within the semgrep docker container.
+#
+set -eu
+
+mkdir -p input
+
+cp golang.yml input/
```
aryx: please, explain how you generated this golang.yml file. Did you curl
a certain r/golang? then please put the command in comment, so if we update the golang rules we know how to update golang.yml

perf/bench/0c34/prep
```diff
@@ -0,0 +1,38 @@
...
+cd input
+
+# Obtain a shallow clone of a git repo for a specific commit
+shallow_clone() {
```
aryx, line 19: @mjambon maybe we should move those helpers in a common.sh somewhere and source it from all the prep script?

perf/bench/coinbase/prep
```diff
@@ -0,0 +1,40 @@
...
+# target_dir: input/bifrost
+#
+# Uses sh because bash is not installed within the semgrep docker container.
+#
+set -eu
+
+mkdir -p input
+mkdir -p input/rules
+
+cp r2c-ci.yml input/rules
```
aryx: again, can you include the command you used to generate those files,
and I would just cat them together and generate a single file instead of putting in input/rules/
to be more consistent with the other corpus which use a single .yml file (which is also useful for semgrep-core).

perf/bench/netflix/prep
```diff
@@ -0,0 +1,40 @@
...
+# target_dir: input/lemur
+#
+# Uses sh because bash is not installed within the semgrep docker container.
+#
+set -eu
+
+mkdir -p input
+mkdir -p input/rules
+
+cp r2c-ci.yml input/rules
```
aryx: if we plan to use that a lot, maybe we could share them instead of duplicating them, so if we update them, we have just to update them at one place.
maybe in perf/common_rules/ @mjambon ?

perf/run-benchmarks
```diff
@@ -50,10 +47,23 @@ CORPUSES = [
         "njsbox", "input/njsscan/njsscan/rules/semantic_grep", "input/dropbox-sdk-js"
     ),
     Corpus("zulip", "input/semgrep.yml", "input/zulip"),
+    Corpus("apache", "input/django.yml", "input/libcloud"),
```
aryx, line 55: great list!

aryx (changes requested):
How long does it take to run those new benchs? Do we remain in a reasonable time like
less than 1min per bench?

aryx (approved):
very nice! good work.

perf/bench/0c34/prep
mjambon: sure, I'll do something like that. It's really just this "shallow_clone" function.

PR #2777, opened 2021-03-18
[semgrep-core] Treat differently pattern: and pattern-inside:

aryx:
This fixes one mismatch in semgrep-rules.
This also changes the regexp to list expected errors in files

test plan:
test file included
semgrep-core -test_rules ~/github/semgrep-rules/
now returns 12 mismatches


aryx:
related to https://github.com/returntocorp/semgrep/issues/1218 but unfortunately reintroduce the difference.

PR #2793, opened 2021-03-21
analyzing: naming: Take care of Go's short assignment

### LINKED ISSUE #2452: Go: metavariables do not work for captured variables with captured  + captured in closures
I want to write a semgrep rule for the following (problematic) Go code:

```go
package foo

import "fmt"

func f(x func ()) {
	go x()
}

func foo(ch chan int) {

	for i := range ch {
		f(func() {
			fmt.Println(i)
		})
	}
}
```

The rule should match if the loop variable `i` in the `for` clause is captured by the closure's body.

My attempt:

```
for $I := range ... {
  ...
  f(func(...) {
    ...
    fmt.Println($I)
    ...
  })
  ...
}
```

(Link: https://semgrep.dev/s/W8OA )

This seems to work.
But if we correct the Go code by shadowing the loop variable with a local variable (a common pattern), the rule still matches.

```go
package foo

import "fmt"

func f(x func ()) {
	go x()
}

func foo(ch chan int) {

	for i := range ch {
                i := i // copy i into a loop-body-local variable so that the closure captures a reference to it instead of the loop variable
		f(func() {
			fmt.Println(i)
		})
	}
}
```

**Expected behavior**

Semgrep probably needs to understand the semantics of Go closure captures and Go shadowing.

Also, I have not found a way to express "variable used anywhere in function body".
I would have expected that the following pattern works, but it doesn't:  `func(...) { <... $I ...> }`  

First time I use semgrep, so no pressure.

**Environment**
Official binary from pip.

IagoAbal:
Closes #2452

test plan:
make test


aryx:
Looks good! The CI failures seem unrelated.

aryx (approved):

PR #946, opened 2020-06-09
target_manager: handle all file filtering in python instead of ocaml

brendongo:
This PR adds logic in python to manage which files to run semgrep rules on.

- Respects per rule file ignores and only runs semgrep-core on necessary files instead of post-filtering
- Regex and semgrep-core rules share logic on what files to run on
- Code in semgrep-core that handles file include/exclude was not removed
- Adds support for respecting gitignore (for now turned off. we can turn on and add a flag in a follow up PR)
- Allows cli include/exclude behavior to mimic rule path config style so we don't have two separate include/exclude syntaxes but will expose the cli changes in a follow up PR
- passing list of files to semgrep-core as a file will also do in a follow up PR

PR #2796, opened 2021-03-22
Remove unused functions

mschwager:
It looks like most of these are unused after https://github.com/returntocorp/semgrep/pull/946, but not all.


aryx (approved):
Sure! Less python code is better code :)

PR #2797, opened 2021-03-22
Include formatted errors in sarif output

### LINKED ISSUE #2748: The --sarif switch will swallow rule errors.
**Describe the bug**
Given that you have a broken rule, an error will be output when running semgrep. That is unless you add the `--sarif` switch.

**To Reproduce**
I created a [SSCCE](http://www.sscce.org/) at https://gitlab.com/dsearles/semgrep-sarif-bug

**Expected behavior**
An error should be output when rules are broken no matter which output options are passed to `semgrep`.

Is this a P0 (blocking your adoption of Semgrep or workflow), P1 (important to fix or quite annoying), P2 (regular bug that should get fixed)?

P1

**Environment**

The link to the example above uses docker in the example, but I was able to also reproduce it locally with the `semgrep` binary installed with homebrew.

mschwager:
Fixes #2748.

Example from above issue:

```
$ python -m semgrep --config /tmp/test.yaml /tmp/index.js --sarif | pjq
{
    "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
    "version": "2.1.0",
    "runs": [
        {
            "tool": {
                "driver": {
                    "name": "semgrep",
                    "semanticVersion": "0.43.0",
                    "rules": []
                }
            },
            "results": []
        }
    ],
    "invocations": [
        {
            "toolExecutionNotifications": [
                {
                    "descriptor": {
                        "id": "InvalidRuleSchemaError"
                    },
                    "message": {
                        "text": "One of these properties is missing: 'pattern', 'pattern-either', 'pattern-regex', 'patterns'"
                    },
                    "level": "error"
                },
                {
                    "descriptor": {
                        "id": "SemgrepError"
                    },
                    "message": {
                        "text": "no valid configuration file found (1 configs were invalid)"
                    },
                    "level": "error"
                }
            ]
        }
    ]
}
```


semgrep/semgrep/error.py
```diff
@@ -57,6 +57,7 @@ def to_dict(self) -> Dict[str, Any]:
         return {
             "type": self.__class__.__name__,
             "code": self.code,
+            "level": self.level.name.lower(),
```
mschwager, line 60: The `.lower()` shenanigans are to stay consistent with `ErrorWithSpan` and `--json` output :+1:

aryx (approved):

PR #2800, opened 2021-03-23
build a 'semgrep-dev' extended container for benchmarks etc.

mjambon:
This builds a docker image that's slightly more convenient than the semgrep one:
* sets the entry point to `bash` instead of `semgrep`
* installs some basic packages

It is intended for CI jobs that will run semgrep inside that container, such as benchmarks.

testing status: the actual `docker build` ran fine on my machine, i.e. the new dockerfile should be fine. We need to see if the image is pushed correctly to DockerHub from the GitHub Actions config. This will should be triggered when the branch is merged (due to `push` event).


mjambon:
The build went fine. The image will be pushed to https://hub.docker.com/repository/docker/returntocorp/semgrep-dev when we merge the PR.

.github/workflows/docker.yml
```diff
@@ -45,3 +45,20 @@ jobs:
...
+            -f dockerfiles/semgrep-dev.Dockerfile \
+            -t returntocorp/semgrep-dev .
+      - name: Push develop image if on develop branch
+        if: ${{ github.event_name == 'push'}}
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+          repository: returntocorp/semgrep-dev
+          tags: develop
```
mjambon, line 64: The most dangerous piece is the `repository` line here. We just have to make sure it's a different name than the other image.

aryx (approved):

PR #2803, opened 2021-03-23
Fix the 'unique_id' field of metavariable captures in json exported by spacegrep

### LINKED ISSUE #2778: [spacegrep] missing matches when same metavariables used in different patterns
See:
https://semgrep.dev/s/GeLB/?stepByStep=1

It's because internally spacegrep returns for each metavariable a unique_id that depends on the location (it should not),
and then 2 metavariables that actually bind to the same content get different unique_ids and can't be matched together.

mjambon:
Fixes https://github.com/returntocorp/semgrep/issues/2778 as indicated there by @aryx

### DIFF
--- a/semgrep-core/src/reporting/JSON_report.ml
+++ b/semgrep-core/src/reporting/JSON_report.ml
@@ -48,6 +48,11 @@ let string_of_resolved = function
  * the callers of sgrep (sgrep-lint) can check if multiple metavariables
  * reference the same entity, or reference exactly the same code.
  * See pfff/.../Naming_AST.ml for more information.
+ *
+ * TODO: provide a typed interface for the json object because it's really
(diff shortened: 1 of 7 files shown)


mjambon:
```
  patterns:
  - pattern: RUN $CMD
  - metavariable-regex:
      metavariable: $CMD
      regex: (shutdown|service|ps|free|top|kill|mount|ifconfig|nano|vim)
```
now matches `RUN foo` instead of just `RUN shutdown` etc, which is incorrect.

Edit: I had made a mistake, putting the metavariable name instead of its capture into the field identifying the capture. The test above should now pass.

mjambon:
Waiting on this PR for the tests to pass https://github.com/returntocorp/semgrep-rules/pull/1186

aryx (approved):

PR #2814, opened 2021-03-24
Add dockerfile into container as /Dockerfile

mjambon:
GitHub Action for building semgrep and semgrep-dev containers seems to be pushing the same image twice when building two different images in the same job.

Having the dockerfile included in the container will help troubleshoot this kind of problems.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,3 @@
-Dockerfile
-dockerfiles
 semgrep/build
 semgrep/dist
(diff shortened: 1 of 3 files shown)


mjambon:
Merging this because it's minor and I need to troubleshoot the GitHub Action snafu that happens when merging into `develop`.

aryx:
sure, looks good.

PR #2821, opened 2021-03-24
Try to use plain docker commands instead of actions/checkout@v2

mjambon:
The GitHub Action `actions/checkout@v2` seems to do nothing that can't be done by calling docker commands directly. It has its own API with undocumented features and oddities, wasting my time.

Let's see if using docker commands works or if I'm missing something.


mjambon:
So this works at least partially. We need to merge into develop to check the rest.

PR #2823, opened 2021-03-25
Move 'docker pull' to before the first 'docker build' to see if it's faster.

mjambon:


mjambon:
Merging to `develop` to see if the conditionals work.

PR #2825, opened 2021-03-25
Allow ellipsis in ArrayGet

### LINKED ISSUE #2819: ... operator in PHP ($foo[...]) cannot be parsed.
**Describe the bug**
Having a Semgrep pattern with `...` in a `$foo[...]` yields a "Pattern could not be parsed as a PHP semgrep pattern."

**To Reproduce**
https://semgrep.dev/s/clintgibler:php-ellipsis-brackets-bug

**Expected behavior**
I expect the pattern `$foo[...]` to match `$foo[some_arg]`.

P1 (important to fix or quite annoying)

emjin:
Closes #2819

Incorporates changes for https://github.com/returntocorp/pfff/pull/408

Test plan:
sc -dump_pattern tests/php/sgrep/dots_in_brackets.php -lang php should not give parse error


semgrep-core/tests/php/sgrep/dots_in_brackets.php
```diff
@@ -0,0 +1 @@
+$foo[...]
```
aryx, line 1: you forgot the commit the .sgrep I think

aryx (approved):

semgrep-core/tests/php/sgrep/dots_in_brackets.php
aryx: ah, got it, you put it in the sgrep/ directory just to check if it parses.
I think it's better when possible to make a proper .php .sgrep in tests/php/ instead.

PR #2832, opened 2021-03-25
Use semgrep-core benchmarks as a submodule of run-benchmarks

emjin:
Removed -bloom_filter, changed -fast to -filter_irrelevant_rules, removed main in semgrep_core_benchmarks and instead called it in run-benchmarks (this necessitated changing the naming convention to underscores)

### DIFF
--- a/perf/run-benchmarks
+++ b/perf/run-benchmarks
@@ -4,6 +4,13 @@
 # and report the time it takes. Optionally upload the results to the semgrep
 # dashboard.
 #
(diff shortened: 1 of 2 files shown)


mjambon (approved):
great, thx!

aryx:
Note that grayson is close to add the --experimental flag to semgrep that directly pass the rule to semgrep-core, see https://github.com/returntocorp/semgrep/pull/2836 so we will be able to test --fast directly in run-benchmarks.

PR #2834, opened 2021-03-25
Ellipses in template strings don't match string literals

### LINKED ISSUE #2780: Regex matching fails with literal backticks
**Describe the bug**
Semgrep regex will not find matches within backticks, making rule creation for Javascript template strings impossible

**To Reproduce**
https://semgrep.dev/s/Kl7P

**Expected behavior**
I would like to match the JS template string with variable expansion in line 1, but not line 2. 

P2. String expansion with tainted variables are an issue we want to use semgrep to identify

emjin:
Closes #2780

When writing a pattern like

```
`...${...}...`
```

the user generally expects the ..., which is confined within the ${}, to imply that the other string also has ${} in it. Since in concat strings the arguments not within ${} are string literals, this means ellipsis should not match string literals

Test plan: make test


aryx:
git merge develop; the 2 CI failures has been fixed in develop I think.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1316,10 +1316,17 @@ and m_arguments_concat a b =
...
-  | xa::aas, xb::bbs ->
-      m_argument xa xb >>= (fun () ->
-        m_arguments_concat aas bbs
-      )
+  | xa::aas, xb::bbs -> (
+      (* exception: for concat strings, don't have ellipsis match   *)
+      (* string literals since string literals are implicitly not   *)
+      (* interpolated, and ellipsis implicitly is                   *)
+      match xa, xb with
+      | A.Arg (A.Ellipsis _), A.Arg (A.L _) ->
```
aryx: hmm, maybe be even more explicit and rquire A.L (A.String here?) instead of just any literal?

aryx (approved):

Issue #2773, opened 2021-03-17
[Internal] Pass the whole rule to semgrep-core in evaluation.py


aryx:
To get semgrep-core -fast, we need to pass the whole rule to semgrep-core, instead of individual patterns like it's the case right now. At some point we also want to pass a big set of rules to semgrep-core at once too, to dedup similar patterns in different rules, but for now let's focus on just passing a whole rule.

This is the most important thing to do to improve the performance of semgrep I think.

We can start by having a --experimental flag to semgrep that use a different code path to pass a whole rule to semgrep-core.
Semgrep-core itself accept a whole rule with the -config <file> flag (instead of the misnamed current -rules_file).

aryx:
cc @mschwager @nbrahms @DrewDennison

aryx:
Note that there are still 28 mismatches between semgrep and semgrep-core -config when used in semgrep-rules, and a few more regressions when using -fast. I need to fix them, but in the mean time we can in parallel work on updating the python wrapper to pass the whole rule to semgrep-core.

aryx:
The only feature not handled by semgrep-core -config is the pattern-where-python, so we may need to keep some of the evaluation.py logic for that, but this should not be too hard because the interface between semgrep and semgrep-core should be similar: semgrep-core will still return a set of matches with metavariables (it's just that it handled the boolean logic composition itself so there's less work in evaluation.py to do).

mschwager:
This seems like a fairly large task, and since it's an internal ask I'm going to remove from the customers board and defer on prioritization (@dlukeomalley, @minusworld, @DrewDennison, @nbrahms?).

This seems like a great start, but I think there's a lot of functionality depending on the current implementation. I'm curious what the plan is for making this transition. We can start dogfooding internally once we have a switch to prefer this implementation, as you mentioned. What's next from there? I would suspect a switch in `semgrep-agent` and/or `semgrep-app`, which we can then use to test it out in real workflows.

Some other non-exhaustive thoughts on what this could affect (we don't necessarily need answers now):

- The `--debugging-json` functionality (`semgrep-app` debugger) depends on the current implemenation.
- It looks like taint mode depends on the current implementation.
  - Maybe we just remove taint mode at this point, but it's a consideration.
- Broader testing ecosystem. `semgrep-rules` tests is a start, but I'm curious about the integration tests in this repository.
  - We can probably just run integration tests side-by-side with the new flag.
[...]

PR #2836, opened 2021-03-26
Integrate fast v2

minusworld:


minusworld:
First step in https://github.com/returntocorp/semgrep/issues/2773

semgrep/semgrep/core_runner.py
```diff
@@ -535,10 +535,97 @@ def _run_rules(
...
+                yaml.dump({"rules": [rule._raw]}, rule_file)
+                rule_file.flush()
+
+                cmd = [SEMGREP_PATH] + [
+                    "-lang",
+                    language,
+                    "-fast",
+                    "-json",
+                    "-config",
+                    rule_file.name,
```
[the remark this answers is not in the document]
minusworld: This is a false positive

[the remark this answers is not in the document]
aryx: nice PR format though :) First time I see one. Love the small font with "how do i suppress this ..."

aryx:
error in CI: pkg_resources.DistributionNotFound: The 'ruamel.yaml.clib>=0.1.2; platform_python_implementation == "CPython" and python_version < "3.9"' distribution was not found and is required by ruamel.yaml

aryx (approved):
This is great!

semgrep/semgrep/core_runner.py
```diff
@@ -535,10 +535,97 @@ def _run_rules(
...
+        List[SemgrepError],
+        Set[Path],
+        Dict[Any, Any],
+    ]:
+        from itertools import chain
+        from collections import defaultdict
+
+        outputs: Dict[Rule, List[RuleMatch]] = defaultdict(list)
+        errors: List[SemgrepError] = []
+        for rule, language in tuple(
```
aryx, line 555: Later on we can try to pass multiple rules at the same time if they have the same language and same set of target files
so we can start to dedup identical patterns and even even identical subformulas used in different rules (e.g., in nodejsscan, ajin reuse many times the same subformulas).

aryx:
I actually also get the ruamel CI error in an unrelated PR: https://github.com/returntocorp/semgrep/pull/2839

semgrep/semgrep/core_runner.py
```diff
@@ -535,10 +535,97 @@ def _run_rules(
...
+                target_file.write("\n".join(map(lambda p: str(p), targets)))
+                target_file.flush()
+                yaml = YAML()
+                yaml.dump({"rules": [rule._raw]}, rule_file)
+                rule_file.flush()
+
+                cmd = [SEMGREP_PATH] + [
+                    "-lang",
+                    language,
+                    "-fast",
```
aryx, line 573: In theory -fast is another thing, but we can bundle it in --experimental.

aryx:
I've merged develop @minusworld to fix some CI errors, so don't forget to pull the latest in your branch.

semgrep/semgrep/cli.py
```diff
@@ -324,6 +324,11 @@ def cli() -> None:
             "containing a 'nosem' comment at the end."
         ),
     )
+    output.add_argument(
+        "--experimental",
```
mschwager, line 328: Can we name this something less generic like `--core-direct-rules`?

minusworld: I think it's probably fine; this is really only meant for internal use. I think Pad was thinking that this would be an entrypoint for future experimental work, too. In any case, we can always change it later - it's just a string after all.

aryx: yep, --core-direct-rules sounds good.

PR #2844, opened 2021-03-26
Adding the new Comby OPAM package as a dependency

aryx:
test plan:
$ opam install --deps-only .


aryx:
I'll add the dependency @mjambon in the ocaml docker base you have; I need to wait though, because this version of comby does not work for what we want.

PR #2845, opened 2021-03-26
Remove xmldiff in favor of simple conversion func

mschwager:
Less dependencies, less problems

### DIFF
--- a/semgrep/tests/e2e/test_check.py
+++ b/semgrep/tests/e2e/test_check.py
@@ -14,6 +14,7 @@ GITHUB_TEST_GIST_URL = (
 )
 
 
+# https://stackoverflow.com/a/10077069
 def etree_to_dict(t):
     d: Dict[str, Dict] = {t.tag: {}}
     children = list(t)
--- a/semgrep/tests/e2e/test_check.py
+++ b/semgrep/tests/e2e/test_check.py
@@ -16,6 +16,10 @@ GITHUB_TEST_GIST_URL = (
 
 # https://stackoverflow.com/a/10077069
 def etree_to_dict(t):
+    """
+    A simple and sufficient XML -> dict conversion function. This function is
+    used to perform basic XML test data comparisons.
+    """
     d: Dict[str, Dict] = {t.tag: {}}
     children = list(t)
     if children:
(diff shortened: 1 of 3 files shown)


semgrep/tests/e2e/test_check.py
```diff
@@ -12,6 +14,27 @@
 )
 
 
+def etree_to_dict(t):
```
mschwager, line 18: This is a nice little snippet I've used a few times. Taken from: https://stackoverflow.com/a/10077069

It's not a perfect XML <-> Dict/JSON, but it's sufficient for this simple usage :+1:

aryx: maybe add this note in a comment then? (the link to stackoverflow) and that the alternative was to use xmldiff but it adds an annoying dependency?

DrewDennison:
do we need to add in any changes in setup.py?

semgrep/tests/e2e/test_check.py
mschwager: Eh, generally I subscribe more to the ["Don’t comment bad code -- rewrite it"](https://www.goodreads.com/quotes/857949-don-t-comment-bad-code-rewrite-it) philosophy, as opposed to lots of comments, but I can add a reference to the stackoverflow post. If anyone needs more context in the future they can always come see this conversation, after all.

mschwager:
> do we need to add in any changes in setup.py?

No, since this is a dev (test) only dependency we should be good in `setup.py`.

semgrep/tests/e2e/test_check.py
aryx: I prefer to add comments about design decisions in the code; git blame + git log + jump to github is 3 hops away to get the information.

aryx: in 2 months someone can come here and replace your code by xmldiff because it does not know what was the problem with xmldiff in the first place.

aryx: I also like "dont comment bad code -- rewrite it", but that does not prevent to add comments in the new good code about  alternative approach and what was wrong with those alternatives approach.

mschwager:
@aryx this should be good to go

aryx (approved):

PR #2848, opened 2021-03-26
Update ruamel, unify calls, add security tests

mschwager:


semgrep/semgrep/rule_lang.py
```diff
@@ -40,9 +39,10 @@ def get(cls) -> Dict[str, Any]:  # type: ignore
         Not thread safe.
         """
         if not cls._schema:
+            yaml = YAML()
```
mschwager, line 42: This defaults to the safe mode, see test below for more info

mschwager:
CI errors (https://github.com/returntocorp/semgrep/pull/2848/checks?check_run_id=2205094835) appear to be from some special validation we're performing. I'll investigate...

semgrep/semgrep/rule_lang.py
```diff
@@ -391,24 +391,24 @@ def _validation_error_message(error: jsonschema.exceptions.ValidationError) -> s
     any_of_invalid_keys = set()
     required = set()
     banned = set()
-    for m in (c.message for c in contexts):
-        if RuleValidation.BAD_TYPE_SENTINEL in m:
-            bad_type.add(m)
-        if RuleValidation.INVALID_SENTINEL in m:
-            ix = m.find(RuleValidation.INVALID_SENTINEL)
+    for context in contexts:
```
mschwager, line 394: We can remove YAML string parsing here altogether, and instead rely on the member variables `validator_value` and `instance` from here: https://python-jsonschema.readthedocs.io/en/stable/errors/#jsonschema.exceptions.ValidationError. The tests were failing because this code was relying on implementation details of `ruamel`. This change fixes up the tests too :+1:

semgrep/setup.py
```diff
@@ -103,7 +103,9 @@ def find_executable(env_name, exec_name):
         "junit_xml==1.9",
         "requests>=2.22.0",
         # Pin exact version of 'ruamel.yaml' because of unstable API.
-        "ruamel.yaml==0.16.10",
+        # Ensure default type is still 'rt' (round-trip) and that RoundTrip
+        # inherits from Safe when upgrading the version.
```
aryx, line 107: We need a semgrep rule running on our dependency! cc @ievans !

semgrep/tests/unit/test_yaml_parsing.py
```diff
@@ -83,3 +86,19 @@ def test_multiple_configs():
         rules = config.get_rules(True)
         assert len(rules) == 3
         assert {"rule1", "rule2", "rule3"} == set([rule.id for rule in rules])
+
+
+def test_default_yaml_type_safe():
+    s = '!!python/object/apply:os.system ["echo Hello world"]'
+
+    # Safe, returns the object
+    default_yaml = YAML()
```
aryx, line 95: ok, I guess we can also use a unit test to ensure upgrade are safe :)

aryx (approved):

PR #2850, opened 2021-03-27
Add note for faster CI

mjambon:
For faster CI, we pre-install opam packages in our docker images. If you add or change an opam dependency, please let me know and I'll take care of it. Or you can do it yourself by updating https://github.com/returntocorp/ocaml-layer/blob/master/common-config.sh
The resulting docker images `returntocorp/ocaml:alpine` and `returntocorp/ocaml:ubuntu` are updated automatically once a week.

### DIFF
--- a/semgrep-core/semgrep.opam
+++ b/semgrep-core/semgrep.opam
@@ -19,6 +19,14 @@ bug-reports: "https://github.com/returntocorp/semgrep/issues"
 # A big dependency is pfff, but it's now a submodule so it's not listed here
 #TODO: restore  "bisect_ppx" {>= "2.5.0"} once can use ppxlib 0.22.0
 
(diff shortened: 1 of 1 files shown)


semgrep-core/semgrep.opam
```diff
@@ -19,6 +19,14 @@ bug-reports: "https://github.com/returntocorp/semgrep/issues"
 # A big dependency is pfff, but it's now a submodule so it's not listed here
 #TODO: restore  "bisect_ppx" {>= "2.5.0"} once can use ppxlib 0.22.0
 
+# For semgrep CI to be fast, we try to pre-install these packages as part of
+# of the base docker image. When you add a new package or change a version
```
aryx, line 23: Double "of"

aryx (approved):
Good to have here indeed

Issue #1834, opened 2020-10-14
Support more expressive paths syntax


Isaac0616:
**Is your feature request related to a problem? Please describe.**
Semgrep supports path include/exclude with the Python glob syntax.
https://github.com/returntocorp/semgrep/blob/develop/docs/configuration-files.md#paths
The examples given in the doc are 
```
- "*.jinja2"
- "*_test.go"
- "project/tests"
- project/static/*.js
```
There is no way to specify something like js files in all subdirectories of `project/static`.

**Describe the solution you'd like**
One potential solution is using regex as the pattern (`project/static/.*[.]js$`).

**Describe alternatives you've considered**
Not sure how useful in general, other ideas are
* support anding the paths (`project/static` and `*.js`)
* syntax to match arbitrary level of directories (`project/static/**/*.js`)

aryx:
I like **/*.js, it's in many shells now.

underyx:
Hey @Isaac0616, we're trying to decide where to place this on our roadmap. Could you give us some idea of how important this feature would be for you?

Isaac0616:
There are not a lot (`dir/` and `*.ext` cover most cases), but I encountered several times during migrating some rules we used to use regex to include/exclude files.
Not super urgent, but it helps reduce false positives.

karoun:
This seems pretty important, since otherwise there's no easy way (that I can tell) to specify "file of type within subdirectory" without using `*`s for every possible file depth.

```yaml
# What I want
- "foo/**/*.txt"

# What I can do
- "foo/*.txt"
- "foo/*/*.txt"
- "foo/*/*/*.txt"
```

clintgibler:
I also think that being able to do `**` in path inclusions/exclusions would be quite useful 👍

fproulx-boostsecurity:
👍 👍 👍 on this, this is super useful to exclude any findings in says `**/migrations/*` for `typescript`

PR #2851, opened 2021-03-29
[WIP] Use pathspec to match include/exclude files.

Isaac0616:
This is an attempt to address #1834.
If this approach looks promising, I can proceed with updating unit tests, etc.

Something to discuss:
- The pathspec's semantics is similar to the current `match_glob` but not entirely compatible.
   I think the major one is that `b/c/` won't match `a/b/c/d/e.py`.
   It has to be `**/b/c/` to retain the same meaning.
   Do you think this is acceptable?
   (btw, is there a way to match only the prefix of a path in the current `match_glob`)
- Should we reject some special patterns which have special meaning in pathspec (e.g. path starts with `!` or `#`)?
   They are unlikely to be provided, but it is nice to be cautious.

### DIFF
--- a/semgrep/setup.py
+++ b/semgrep/setup.py
@@ -106,6 +106,7 @@ setuptools.setup(
         "ruamel.yaml==0.16.10",
         "tqdm>=4.46.1",
         "packaging>=20.4",
+        "pathspec>=0.8.0",
         "jsonschema~=3.2.0",
         # Include 'setuptools' for 'pkg_resources' usage. We shouldn't be
         # overly prescriptive and pin the version for two reasons: 1) because
--- a/semgrep/semgrep/target_manager.py
+++ b/semgrep/semgrep/target_manager.py
@@ -11,6 +11,7 @@ from typing import List
 from typing import Set
 
 import attr
+import pathspec
 
 from semgrep.config_resolver import resolve_targets
 from semgrep.error import FilesNotFoundError
@@ -187,14 +188,6 @@ class TargetManager:
         return expanded
 
     @staticmethod
-    def match_glob(path: Path, globs: List[str]) -> bool:
-        """
-        Return true if path or any parent of path matches any glob in globs
-        """
-        subpaths = [path, *path.parents]
-        return any(p.match(glob) for p in subpaths for glob in globs)
-
-    @staticmethod
     def filter_includes(arr: Set[Path], includes: List[str]) -> Set[Path]:
         """
         Returns all elements in arr that match any includes pattern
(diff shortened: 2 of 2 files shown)


semgrep/setup.py
```diff
@@ -106,6 +106,7 @@ def find_executable(env_name, exec_name):
         "ruamel.yaml==0.16.10",
         "tqdm>=4.46.1",
         "packaging>=20.4",
+        "pathspec>=0.8.0",
```
Isaac0616, line 109: What is the right steps to add a new third-party library?
The README said: "Runtime dependencies may not be installed via pipenv. They must be specified in setup.py."
So, I added this line.
However, I still failed the mypy import test.

DrewDennison (commented):
thanks for finding `pathspec` looks nice. Adding @mschwager who is the expert on how python deps work

semgrep/setup.py
nbrahms: cc. @brendongo @underyx

fproulx-boostsecurity:
👍 great contribution

semgrep/setup.py
mschwager: This looks correct to me. Dependencies for mainline functionality go in `setup.py`, while dev dependencies go in the `Pipfile`. For mypy you probably have to add the following to `mypy.ini`:

```
# pathspec
[mypy-pathspec.*]
ignore_missing_imports = True
```

brendongo:
Sorry for the delay just got back from vacation so taking a look at this now. Looks great does pathspec have full feature parity with how .gitignore works?

> The pathspec's semantics is similar to the current match_glob but not entirely compatible.
> I think the major one is that b/c/ won't match a/b/c/d/e.py.
> It has to be **/b/c/ to retain the same meaning.
> Do you think this is acceptable?
> (btw, is there a way to match only the prefix of a path in the current match_glob)

seems fine. Afaicr we had b/c/ match a/b/c/d/e.py because we didn't have the ** functionality. So this is an improvement. We will just have to update the help text of includes/excluldes and make a note in the changelog

> Should we reject some special patterns which have special meaning in pathspec (e.g. path starts with ! or #)?
> They are unlikely to be provided, but it is nice to be cautious.

Yeah sounds like a good idea, we wouldnt need to support those anyway.

@Isaac0616 this looks good feel free to proceed 🔥 

I'd feel better if we pinned to a specific version of pathspec given project inactivity

@DrewDennison looks like pathspec uses Mozilla Public License v2 are we good with this?

Isaac0616:
While I am updating the unit and e2e tests, I found that there seems to be a bug in `pathspec` library.
I reported it to the author (https://github.com/cpburnz/python-path-specification/issues/46) and hopefully it can be resolved there.

Isaac0616:
Keep this open. Didn't receive response from the pathspec author. Will think about alternative approaches this week.

Issue #2787, opened 2021-03-19
mode taint should be improved minimally to make it somewhat useful


fproulx-boostsecurity:
I have a test case here (which BTW is broken in `0.42.0`+ -- filed separate bug here https://github.com/returntocorp/semgrep/issues/2786) 

Here is a playground that shows basically the only scenario that works.
If the other scenarios listed would be made to work this would make mode taint actually useful...

https://semgrep.dev/s/fproulx-boostsecurity:dom-xss-taint-tracking

aryx:
@IagoAbal want to take a look? Does some work on the IL could fix the issue?

fproulx-boostsecurity:
Basically if the following scenarios (`mode: taint`) were made to work, that would at least allow to implement some injections rules that have a good number of `TPs` and also many `FNs`. But that's a conscious choice, for my use case that's a valid trade off (High Signal / Low Noise). Better to have many good, easy to understand and fix `TPs`, than wasting time on many hard to validate `FPs`, .

```
    // NOT WORKING - Direct use
    // todoruleid: typescript.nestjs.security.test-taint-mode
    return React.createElement("a", { href: location.hash });

    // NOT WORKING - Mutate before use
    const c = location.hash + "bbb";
    // todoruleid: typescript.nestjs.security.test-taint-mode
    return React.createElement("a", { href: c });

    // NOT WORKING - Mutate before use - in-line
    // todoruleid: typescript.nestjs.security.test-taint-mode
    return React.createElement("a", { href: "aa" + location.search });
```

IagoAbal:
Hi! I think it shouldn't take much work to handle these examples. It just probably requires a bit more work on our IL as @aryx is suggesting. I will take a look at it this week!

IagoAbal:
Most of these examples can be make to work by introducing extra temporary variables during AST-to-IL. The dataflow analyses only track facts on simple variables. This is clean and convenient. Perhaps we should consider changing the IL to force some subexpressions to be variables instead.

IagoAbal:
@fproulx-boostsecurity The first examples do not work because the dataflow analysis kind of assumes that the _source_ will be a function, it is expecting something like:

```yaml
pattern-sources:
    - source(...)
```
I will fix that so it works also for sources that are object attributes.

[...]

PR #2856, opened 2021-03-30
matching: Make deep-expr patterns match inside records

IagoAbal:
Related-to: #2787

test plan:
make test


aryx (approved):
ok, let's hope performance is not impacted.
The best would be to run the new benchmarks of martin (in semgrep/perf/run-benchmarks) before and after and check for any perf regressions.

PR #2870, opened 2021-04-01
Add YAML language support for --test

### LINKED ISSUE #2799: `semgrep --test` a yaml file
Currently, `semgrep --test` doesn't allow a yaml file to be the test file even the rules and test files are separated.
For example, `semgrep --test --config rules/ tests/` will ignore `x.yaml` in the following dir structure.
```
project/
  rules/
    x.yml
  tests/
    x.py
    x.yaml
```
The current workaround is using other file extensions for the generic patterns, but I think once the yaml language rule is supported (#2601), we need some way to support it.

I can see that semgrep will confuse if both rules and tests with yaml/yml extension are in the same directory.
Not sure what's the best solution.
Maybe use special extension that semgrep would treat it as yaml during testing.

mschwager:
Fixes #2799.

Manually tested by running `--test` against `semgrep-rules` and the following directory:

```
$ cat foo.yaml 
rules:
- id: my_pattern_id
  pattern: |
    rules:
        ...
  message: bad
  severity: WARNING
  languages: [yaml]
```

```
$ cat foo.test.yaml 
# ruleid: my_pattern_id
rules:
    foo: bar
```

```
$ python -m semgrep --quiet --test /tmp/yaml-test/
1 yaml files tested
check id scoring:
(TODO: 0) /tmp/yaml-test/foo.yaml
	✔ my_pattern_id                                                TP: 1 TN: 0 FP: 0 FN: 0
final confusion matrix: TP: 1 TN: 0 FP: 0 FN: 0
```

We'll need to add a note here about this special case: https://semgrep.dev/docs/writing-rules/testing-rules/


nbrahms:
Do you want a test case?

mschwager:
> Do you want a test case?

Unfortunately we don't have any tests for `--test` yet, which is becoming tougher and tougher as it gets more complex. I can create a few, but it will take me a bit longer to wire it up.

nbrahms:
> > Do you want a test case?
> 
> Unfortunately we don't have any tests for `--test` yet, which is becoming tougher and tougher as it gets more complex. I can create a few, but it will take me a bit longer to wire it up.

KK. I'll approve this in the meantime.

nbrahms (approved):

mjambon:
:+1: for updating the changelog!

PR #2881, opened 2021-04-05
analyzing: tainting: Allow arbitrary expressions to be sources

IagoAbal:
This fixes 3 out of the 5 non-working examples in #2787.

### DIFF
--- a/semgrep-core/src/analyzing/Dataflow_tainting.ml
+++ b/semgrep-core/src/analyzing/Dataflow_tainting.ml
@@ -45,6 +45,7 @@ type mapping = unit Dataflow.mapping
 (* this can use semgrep patterns under the hood *)
 type config = {
   is_source: IL.instr -> bool;
+  is_source_exp: IL.exp -> bool;
   is_sink: IL.instr -> bool;
   is_sanitizer: IL.instr -> bool;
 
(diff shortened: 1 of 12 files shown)


semgrep-core/src/analyzing/Dataflow_tainting.ml
```diff
@@ -77,6 +78,49 @@ let option_to_varmap = function
   | Some lvar -> VarMap.singleton (str_of_name lvar) ()
 (*e: function [[Dataflow_tainting.option_to_varmap]] *)
 
+(*****************************************************************************)
+(* Tainted *)
+(*****************************************************************************)
+
+let sanitized config instr =
+  match instr.i with
+  | Call (_, {e=Lvalue({base=Var(("sanitize",_),_);_}); _}, [])
```
aryx, line 87: I would add -> true for the Call case and
then just do | _ -> config.is_sanitizer instr

semgrep-core/src/analyzing/Dataflow_tainting.ml
```diff
@@ -77,6 +78,49 @@ let option_to_varmap = function
...
+    | Operator (_, es)
+      -> List.exists (tainted env config) es
+    | Record fields
+      -> List.exists (fun (_, e) -> tainted env config e) fields
+    | Cast (_, e)
+      -> tainted env config e
+  in
+  config.is_source_exp exp || go_into exp.e
+
+let tainted_instr env config instr =
```
aryx, line 111: what was wrong with just using rvars_of_instr and looking if one of them is tainted?
Why the explicit matching of instr cases here? rvars_of_instr was wrong?

semgrep-core/src/analyzing/Dataflow_tainting.ml
```diff
@@ -126,68 +165,30 @@ let (transfer: config -> flow:F.cfg -> unit Dataflow.transfn) =
...
-              | None -> None
-              | Some lvar ->
-                  (* one taint argument propagate the taint to the lvar *)
-                  if rvars |> List.exists (fun rvar ->
-                    VarMap.mem (str_of_name rvar) in')
-                  then Some lvar
-                  else None
-             )
-        )
+        if tainted_instr in' config x then
```
aryx, line 168: wow, nicer! shorter is good!

semgrep-core/src/analyzing/Dataflow_tainting.ml
```diff
@@ -126,68 +165,30 @@ let (transfer: config -> flow:F.cfg -> unit Dataflow.transfn) =
...
-              | None -> None
-              | Some lvar ->
-                  (* all clean arguments should reset the taint *)
-                  if rvars |> List.for_all (fun rvar ->
-                    not (VarMap.mem (str_of_name rvar) in'))
-                  then Some lvar
-                  else None
-             )
-        )
+        if tainted_instr in' config x then
```
aryx, line 186: shorter again! Good job.

semgrep-core/tests/Test.ml
```diff
@@ -136,6 +138,53 @@ let regression_tests_for_lang ~with_caching files lang =
...
+  in
+  let expected = Error_code.expected_error_lines_of_files [file] in
+  Error_code.compare_actual_to_expected actual expected
+
+let tainting_tests_for_lang files lang =
+  files |> List.map (fun file ->
+   (Filename.basename file) >:: (fun () ->
+    let rules_file =
+      let (d,b,_e) = Common2.dbe_of_filename file in
+      let candidate1 = Common2.filename_of_dbe (d,b,"rules") in
```
aryx, line 180: I think you can use a .yaml instead of .rules

aryx (approved):
Great work!

semgrep-core/src/analyzing/Dataflow_tainting.ml
IagoAbal: Good point, fixed.

semgrep-core/src/analyzing/Dataflow_tainting.ml
IagoAbal: Note that `go_into` is not recursive, but `tainted` is. Every time we call `tainted` on a subexpression we are checking if the subexpression matches any pattern in `pattern-sources`. I will add a comment to the code.

IagoAbal: In principle the only interesting subexpressions are the l-values, but we need the `eorig` of those lvalues in order to match. We could have some new `xyz_of_instr` function to return them and it would work as well. But note that the current code allows for _any arbitrary expression_ to be a source of taint... you could e.g. say that `1 + 1` is a source. A more interesting example could be `NULL`  as a source... This could perhaps be interesting for advanced users to play around?

semgrep-core/tests/Test.ml
IagoAbal: I agree .yaml it's better, fixed!

PR #2886, opened 2021-04-05
Make deep expressions traverse OtherStmt, Lambda nodes

### LINKED ISSUE #2878: [golang] `go func() { ... }` doesn't parse as a pattern
**Describe the bug**
The `go` keyword seems to break pattern parsing, but not target parsing.

**To Reproduce**
https://semgrep.dev/s/1n85/

**Expected behavior**
Pattern should parse (and match)

**Screenshots**
P1 (prevents rule-writing)

### LINKED ISSUE #2879: [golang] Deep expressions don't work with `go` keyword
**Describe the bug**
Deep expression matching doesn't work with `go` statements

**To Reproduce**
https://semgrep.dev/s/kLre/

**Expected behavior**
Pattern should match

**Screenshots**

Unknown priority 😢

nbrahms:
A few easily recurseable cases were omitted from deep-statement
matching: OtherStmt nodes, and Lambda nodes. The former can refer to an
arbitrary number of `any` nodes, and we update to traverse any wrapped
`E` nodes. The latter we treat like any other `FunctionDef` node.

The `OtherStmt` change should be relatively innocuous, as these
statements are relatively rare in code. I'm a bit worried about the
performance impacts of the Lambda change, however, as a minified js
file could easily contain a very large number of anonymous functions.

To that end, let's not merge this until we can confirm that we are not
creating a performance regression with this change.

Fixes #2879.

Fixes #2878.

Perhaps not the most complete argument re. performance, but looking at `p/nodejsscan` (which makes heavy use of deep expressions) against `angular.min.js` (which makes heavy use of anonymous functions) yields no significant regression:

Before this change:
```
semgrep --config p/nodejsscan node_modules/angular/angular.min.js  233.27s user 12.69s system 98% cpu 4:08.47 total
```

After this change:
```
semgrep --config p/nodejsscan node_modules/angular/angular.min.js  229.75s user 12.45s system 98% cpu 4:05.77 total
```

PR checklist:
- [x] changelog is up to date


semgrep-core/src/matching/SubAST_generic.ml
```diff
@@ -32,6 +32,11 @@ let go_really_deeper_stmt = ref true
 (* Sub-expressions and sub-statements *)
 (*****************************************************************************)
 
+let subexprs_of_any_list xs = xs |> List.fold_left (fun x -> (function
+  | E e -> e :: x
+  | _ -> x
+)) []
```
mjambon, line 38: Your use of fold_left is weird to me. I would write it as:
```ocaml
let subexprs_of_any_list anys =
  List.fold_left (fun acc any ->
    match any with
      | E e -> e :: acc
      | _ -> acc
  ) [] anys
```
The part that's weird to me in your version is to not have the two arguments side by side (`acc any` and `[] anys` in my version).

mjambon (approved):
Looks good to me.

semgrep-core/src/matching/SubAST_generic.ml
nbrahms: ^^ yeah looks better

PR #2887, opened 2021-04-06
Check benchmark correctness and display results in chart

### LINKED ISSUE #2852: [Benchmark] make sure the different SemgrepVariant return the same findings
Probably need to pass --json to semgrep and then make sure the result finding sets are equal to the base (the std SemgrepVariant).

emjin:
Closes #2852

Add flag --plot_benchmarks to automatically graph the benchmarks

Check errors and results against the std benchmark. For each benchmark, it will output an error if the results do not match, and a warning if it counts more errors in the benchmark than the std. Currently, errors do not compare the message, because the experimental variants aren't propagating metavariables properly. We achieve this by using the class SemgrepResult, which compares results using an object that elides the message first. This should be restored after the experimental variant is fixed.

Some of the experimental results are still different---it seems that the column numbers are sometimes off by one.

Test plan:
./run_benchmarks
./run_benchmarks --plot_benchmarks

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -55,8 +55,12 @@ ignore_missing_imports = True
 [mypy-tqdm.*]
 ignore_missing_imports = True
 
-# packaging
-[mypy-packaging.*]
+# matplotlib
+[mypy-matplotlib.*]
+ignore_missing_imports = True
+
+# pandas
+[mypy-pandas]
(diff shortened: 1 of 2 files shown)


aryx (approved):
Great!

Issue #2890, opened 2021-04-06
Access generic AST from other languages (Semgrep "API")


aryx:
Just an issue to track this.

ievans:
@aryx I thought there was some progress on this? At least a JSON dump?

aryx:
Yes, we exported the AST using martin's atd tool, which make it easier to export/import in other languages.
Now we need customers for this feature.

PR #2891, opened 2021-04-06
AST generic ATD definition v1

aryx:
This will help https://github.com/returntocorp/semgrep/issues/2890

test plan:
make

PR checklist:
- [ ] changelog is up to date


mjambon:
Looks great. On the use of atdgen:

Maybe it's better to use classic variants than polymorphic ones. Things were different when atdgen was created. Things like the following weren't possible for a constructor `Foo.Bar`:
```ocaml
let f (x : Foo.t) =
  match x with
  | Bar -> ...
  ...
``` 
Instead we had to write:
```ocaml
let f x =
  match x with
  | Foo.Bar -> ...
  ...
```
or
```ocaml
open Foo (* and hope we're not shadowing another 'Bar' *)
let f x =
  match x with
  | Bar -> ...
  ...
```

With polymorphic variants, we could already do this:
```ocaml
let f (x : Foo.t) =
  match x with
  | `Bar -> ...
  ...
```
which was the main reason why I thought polymorphic variants should be the default at the time.

Regarding versioning and compatibility: assuming the AST will be provided to pure consumers, we should reserve the right to add record fields without considering them a breaking change. The user's code should tolerate the addition of new record fields. This doesn't work with tuples because their length must always be known, that's why we should have a preference for records. Extensions also don't work with sum types: we can't add new cases without breaking consumer (reader side) code. Also, we shouldn't remove cases either because we expect the user to rely on exhaustiveness checks in pattern matching. Examples:

Original interface:
```ocaml
type record = { a: int }
type tuple = int * string
type sum = A | B
```
OK change:
```ocaml
type record = { a: int; b: string }
```
Breaking changes:
```ocaml
type tuple = int * string * float
type sum = A | B | C
(* also breaking: *)
type sum = A
```

mjambon (approved):

aryx:
Yep, not a big fan of poly variant, but I kept the default. It's easy to change.
Thanks for the notes on atdgen.

PR #2895, opened 2021-04-07
Use copied junit_xml instead of external dependency

mschwager:
I'm a fan of the following from Go:

> [A little copying is better than a little dependency.](https://go-proverbs.github.io/)

The whole dependency is basically a single file, let's just use it :+1:. Especially considering we're planning on using our own `junit_xml_2` fork. This is basically the same thing without the dependency hassle.

### DIFF
--- a/semgrep/semgrep/external/junit_xml.py
+++ b/semgrep/semgrep/external/junit_xml.py
@@ -1,0 +1,433 @@
+# type: ignore
+import collections
+import re
+import sys
+import xml.dom.minidom  # nosem: python.lang.security.use-defused-xml.use-defused-xml
+import xml.etree.ElementTree as ET  # nosem: python.lang.security.use-defused-xml.use-defused-xml
+
+"""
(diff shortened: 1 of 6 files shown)


semgrep/semgrep/external/junit_xml.py
```diff
@@ -0,0 +1,433 @@
...
+    xml_string = ET.tostring(xml_element, encoding=encoding)
+    # is encoded now
+    xml_string = _clean_illegal_xml_chars(xml_string.decode(encoding or "utf-8"))
+    # is unicode now
+
+    if prettyprint:
+        # parseString() works just on correctly encoded binary strings
+        xml_string = xml_string.encode(encoding or "utf-8")
+        # fmt: off
+        xml_string = xml.dom.minidom.parseString(xml_string)  # nosem: contrib.dlint.dlint-equivalent.insecure-xml-use
```
mschwager, line 269: This string is generated internally, so we should be okay here.

brendongo (approved):

PR #2899, opened 2021-04-07
Add tests for known slow rule-file combinations

emjin:
Add two tests targeting different methods known to be unnecessarily slow. See https://github.com/returntocorp/semgrep/pull/2868 and https://github.com/returntocorp/semgrep/pull/2869

PR checklist:
- [ ] changelog is up to date


mjambon (approved):

aryx (commented):
We should try to restrict ourseleves to patterns in those micro benchmark, not full rule.

PR #2905, opened 2021-04-08
Add benchmarking info to README

emjin:
PR checklist:
- [x] changelog is up to date

### DIFF
--- a/perf/run-benchmarks
+++ b/perf/run-benchmarks
@@ -152,6 +152,9 @@ SEMGREP_VARIANTS = [
     SemgrepVariant("experimental", "-fast", "--experimental"),
 ]
 
+# For when you just want to test a single change
+STD_VARIANTS = [SemgrepVariant("std", "")]
+
 
 # This class allows us to put semgrep results in a set and compute set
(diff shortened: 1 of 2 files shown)


aryx (approved):
Good

perf/run-benchmarks
```diff
@@ -390,6 +398,9 @@ def main() -> None:
     parser.add_argument(
         "--upload", help="upload results to semgrep dashboard", action="store_true"
     )
+    parser.add_argument(
+        "--std_only", help="only run the default semgrep", action="store_true"
```
mjambon, line 402: I'd like to change option flags to using only dashes (`--std-only`, `--plot-benchmarks`), as it is most common and expected by users.

emjin: Will do! Our options flags are somewhat disorganized; I'm thinking about separating the corpuses into small (< 10 s), medium (< 2 min), and large (rest), and having corresponding flags

aryx: I recently read that the fact that we use -- for flags is an accident; originally short was supposed to use -x and long argument +xxx :)

PR #2913, opened 2021-04-09
analyzing: tainting: Add basic cross-function capabilities

### LINKED ISSUE #2787: mode taint should be improved minimally to make it somewhat useful

IagoAbal:
See: tests/tainting_rules/ts/2787-fun.ts
    
Closes #2787
    
test plan:
make test

PR checklist:
- [ x] changelog is up to date


IagoAbal:
I think my changes to `Naming_AST` are causing the test failures, but I believe it's not because there is any real error, we just need to update the expected output.

aryx:
The changes to the snapshot look good to me, but now 2 tests in semgrep-rules are failing.
You need to investigate those regressions.

CHANGELOG.md
```diff
@@ -15,14 +15,16 @@ will now match
 var { exec } = require("child_process");
 exec("dangerous");
 ```
```
aryx, line 34: you need to merge develop, this is an old version of CHANGELOG.md

semgrep-core/tests/tainting_rules/ts/2787-fun.ts
```diff
@@ -0,0 +1,16 @@
+// https://github.com/returntocorp/semgrep/issues/2787
```
aryx: I think it would be better to name the file with the feature it covers, like
basic_interprocedural.{ts,yaml}
and then put like you did in it a comment so people can refer to the relevant issue if they need more context.

semgrep-core/src/analyzing/Dataflow_tainting.ml
```diff
@@ -85,37 +88,44 @@ let option_to_varmap = function
...
+   * a source, this would infer that `"aa" + location.href + "bb"` is tainted.
+   * Also note that any arbitrary expression can be source! *)
+  let is_tainted = tainted config fun_env env in
   let go_into = function
-    | Lvalue {base=Var var;_}
+    | Lvalue {base=Var var; _}
       -> VarMap.mem (str_of_name var) env
+         || Hashtbl.mem fun_env (str_of_name var)
+    | Lvalue {base=VarSpecial(This, _); offset=Dot fld; _}
+      -> Hashtbl.mem fun_env (str_of_name fld)
```
aryx, line 106: str_of_name does not use sid, so if a local has the same name than a tainted function, it will report the local var
as tainted.

semgrep-core/src/analyzing/Dataflow_tainting.ml
```diff
@@ -41,6 +41,9 @@ module VarMap = Dataflow.VarMap
 type mapping = unit Dataflow.mapping
 (*e: type [[Dataflow_tainting.mapping]] *)
 
+(* Tracks tainted functions. *)
+type fun_env = (Dataflow.var, unit) Hashtbl.t
```
aryx, line 45: maybe we need to add the sid here? Dataflow.var is just a string.

semgrep-core/src/analyzing/Dataflow_tainting.mli
```diff
@@ -19,6 +22,6 @@ type config = {
 
 (*s: signature [[Dataflow_tainting.fixpoint]] *)
 (* main entry point *)
-val fixpoint : config -> IL.cfg -> mapping
+val fixpoint : config -> fun_env -> IL.name option -> IL.cfg -> mapping
```
aryx, line 38: maybe a comment here about what fun_env and IL.name option are.

semgrep-core/src/analyzing/Naming_AST.ml
```diff
@@ -352,7 +352,10 @@ let is_local_or_global_ctx env lang =
   | InClass -> (
       match lang with
       (* true for Java so that we can type class fields *)
-      | Lang.Java -> true
+      | Lang.Java
+      (* true for JS/TS so that we can resolve class methods *)
```
aryx, line 355: maybe rename is_resolvable_name_ctx env lang

semgrep-core/src/core/ast/AST_generic_helpers.ml
```diff
@@ -38,6 +38,14 @@ module M = Map_AST
 let str_of_ident = fst
 (*e: constant [[AST_generic.str_of_ident]] *)
 
+let name_of_entity ent =
+  match ent.name with
+  | EN (Id (i, pinfo))
+  | EN (IdQualified ((i, _), pinfo))
+    -> Some (i, pinfo)
+  | ___else___
```
aryx: better be explicit with remaining cases here, especially if there's not so many such cases.

semgrep-core/src/core/il/IL.ml
```diff
@@ -169,8 +169,11 @@ and offset =
    * Note that Dot is used to access many different kinds of entities:
    *  objects/records (fields), classes (static fields), but also
    *  packages, modules, namespaces depending on the type of 'var' above.
+   *
+   * old: Previously the field was an `ident` but in some cases we want a
+   *      proper (resolved) name here.
```
aryx: more comment, "and point directly to the method, e.g., in this.foo() we can resolve foo; This is useful for our
poor's man interprocedural analysis".

aryx (changes requested):
Looks good.
So this works only if the method is defined before it's used right?

semgrep-core/tests/tainting_rules/ts/2787-fun.ts
```diff
@@ -0,0 +1,16 @@
...
+import React, { Component } from "react";
+class BadPractice extends Component {
+  dangerous() {
```
aryx: So if this method is defined after render, it will not detect the tainting?
The order matters right?

CHANGELOG.md
IagoAbal: Thanks for the heads up, I will do.

semgrep-core/tests/tainting_rules/ts/2787-fun.ts
IagoAbal: Yes, as of now, that is correct. We could iterate the methods in topological order wrt their dependencies.

IagoAbal:
Thanks for the feedback! I'm investigating the failing tests and then I'll address all your comments.

IagoAbal:
Just checking if the tests pass now (I did run them for a while on my laptop but it's too slow). I still have to address your comments, but I'll do that tomorrow (hopefully, if the tests pass).

semgrep-core/src/analyzing/Dataflow_tainting.ml
IagoAbal: In this case we are converting the `IL.name` to a string "<source name>:<sid>", I clarified that in a comment. I did it this way because it's how the `Dataflow.mapping`s work, but we could also have a proper key type `string * AST_generic.sid`.

semgrep-core/src/analyzing/Dataflow_tainting.ml
IagoAbal: In this case `str_of_name` does use the `sid`, but I see there are many `str_of_name` defined in different modules so maybe the confusion comes from that.

IagoAbal:
Thanks a lot for the review! I think I addressed all your comments.

CHANGELOG.md
aryx: you merged but you didn't put those new additions in the right section. They should not be with the 0.46 since it actually didn't make the cut. It should be in the "Coming up next" section.

aryx (approved):
Great comments!

CHANGELOG.md
IagoAbal: Ooops, sorry, I didn't notice that it had been bumped already to 0.46. Fixing it in PR #2955.

PR #2921, opened 2021-04-12
Support many suffix delimiters in filename, only check end

mschwager:
This provides centralized functionality to determine if a path is a config file, YAML test file, or something else.


semgrep/semgrep/util.py
```diff
@@ -174,3 +179,37 @@ def compute_semgrep_path() -> str:
 
 def compute_spacegrep_path() -> str:
     return compute_executable_path("spacegrep")
+
+
+def liststartswith(l: List[Any], head: List[Any]) -> bool:
```
brendongo: Does List[str] not work?

semgrep/semgrep/util.py
```diff
@@ -20,6 +21,10 @@
 
 T = TypeVar("T")
 
+YML_EXTENSIONS = {".yml", ".yaml"}
+YML_SUFFIXES = [[ext] for ext in YML_EXTENSIONS]
+YML_TEST_SUFFIXES = [[".test", ext] for ext in YML_EXTENSIONS]
```
brendongo, line 26: Wondering if we should be using soemthing like .r2ctest or something specific in case a user actually has .test.yaml?

brendongo (approved):

semgrep/semgrep/util.py
mschwager: `str` would work, but this function also works with list's containing anything that defines `==`, not just strings.

semgrep/semgrep/util.py
mschwager: Hmm, we could do that, but we've already released a version using `.test`, and all our `semgrep-rules` use `.test` (not that that's unchangeable). I'm less worried about YAML test files (`.test.yaml`) because that tends to be a much more tightly controlled directory structure, i.e. it's typically just Semgrep rules and associated tests. The config file suffix comparison should be much stricter because it runs across diverse directories, i.e. arbitrary user repositories.

semgrep/semgrep/util.py
mschwager: Maybe `T` makes more sense?

brendongo:
Probably want to update changelog

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 - fixed single field pattern in JSON, allow `$FLD: { ... }` pattern
+- Config detection in files with many suffix delimiters, like `this.that.check.yaml`.
+  More concretely: configs end with `.yaml`, YAML language tests end with `.test.yaml`,
+  and everything else is handled by its respective language extenion (e.g. `.py`).
```
mschwager:
```suggestion
  and everything else is handled by its respective language extension (e.g. `.py`).
```

Issue #2806, opened 2021-03-24
Allow multiple rules to be triggered by the same line in test files.


Isaac0616:
**Is your feature request related to a problem? Please describe.**
Sometimes, different rules will be triggered by the same line.
However, we are not able to comment two `ruleid: ` on the same line.
We can try to avoid it by changing the test code, but it would be more convenient to comment multiple rule ids.

This happens when we have two rules to catch similar things.
One is for a more generic case.
The other is for a more specific case with more actionable message.

**Describe the solution you'd like**
Maybe we can allow a comma separated list of rule ids like `# ruleid: id-a, id-b` or be able to comment in multiple lines like
```
# ruleid: id-a
# ruleid: id-b
target
```

Isaac0616:
Found that there used to be another issue (#753)

minusworld:
Closed by https://github.com/returntocorp/semgrep/pull/2923

minusworld:
@Isaac0616 - I ended up doing comma-separated since that was a bit easier to implement. Please let us know if it works as expected. Thanks! 👍 🙏

Isaac0616:
Thanks! I will check it out.

PR #2923, opened 2021-04-13
Multiple test annotations

minusworld:
Adds the ability to include multiple rule IDs for test annotations such as `ruleid` and `ok` on the same line. Rule IDs must be comma-separated.

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/test.py
+++ b/semgrep/semgrep/test.py
@@ -111,22 +111,22 @@ def _annotations(annotation: str) -> Set[str]:
 
 def line_has_todo_rule(line: str) -> bool:
     todo_rule_annotations = _annotations(TODORULEID)
-    return any([annotation in line for annotation in todo_rule_annotations])
+    return any(annotation in line for annotation in todo_rule_annotations)
 
 
 def line_has_rule(line: str) -> bool:
     rule_annotations = _annotations(RULEID)
-    return any([annotation in line for annotation in rule_annotations])
+    return any(annotation in line for annotation in rule_annotations)
 
 
 def line_has_ok(line: str) -> bool:
     rule_annotations = _annotations(OK)
-    return any([annotation in line for annotation in rule_annotations])
+    return any(annotation in line for annotation in rule_annotations)
 
 
 def line_has_todo_ok(line: str) -> bool:
     rule_annotations = _annotations(TODOOK)
-    return any([annotation in line for annotation in rule_annotations])
+    return any(annotation in line for annotation in rule_annotations)
 
 
 def score_output_json(
--- a/semgrep/semgrep/test.py
+++ b/semgrep/semgrep/test.py
@@ -416,7 +416,7 @@ def generate_file_pairs(
 
     for filename, rr in passed_results_first.items():
         print(f"(TODO: {rr['todo']}) {filename}")
-        for check_id, check_results in rr["checks"].items():
+        for check_id, check_results in sorted(rr["checks"].items()):
             print(generate_check_output_line(check_id, check_results))
             if not check_results["passed"]:
                 print(generate_matches_line(check_results))
--- a/semgrep/semgrep/test.py
+++ b/semgrep/semgrep/test.py
@@ -13,6 +13,7 @@ import argparse
 import collections
 import functools
 import json
+import logging
 import multiprocessing
 import sys
 import tarfile
@@ -32,6 +33,8 @@ from semgrep.util import is_config_suffix
 from semgrep.util import is_config_test_suffix
 from semgrep.util import partition
 
+logger = logging.getLogger(__name__)
+
 SAVE_TEST_OUTPUT_JSON = "semgrep_runs_output.json"
 SAVE_TEST_OUTPUT_TAR = "semgrep_runs_output.tar.gz"
 
@@ -43,6 +46,12 @@ TODOOK = "todook"
 OK = "ok"
 
 
+def _remove_ending_comments(rule: str) -> str:
+    for _, end in COMMENT_SYNTAXES:
+        rule = rule.strip() if end == "\n" else rule.strip().replace(end, "")
+    return rule
+
+
 def normalize_rule_ids(line: str) -> Set[str]:
     """
     given a line like `     # ruleid:foobar`
@@ -52,11 +61,9 @@ def normalize_rule_ids(line: str) -> Set[str]:
     _, rules_text = line.strip().split(":")
     rules_text = rules_text.strip()
     rules = rules_text.split(",")
-    # strictly remove ending HTML comment
-    rules = [rule.strip().replace("-->", "") for rule in rules]
-    # strictly remove ending OCaml comment
-    rules = [rule.strip().replace("*)", "") for rule in rules]
-    return set(filter(None, [rule.strip() for rule in rules]))
+    # remove comment ends for non-newline comment syntaxes
+    rules_clean = map(lambda rule: _remove_ending_comments(rule), rules)
+    return set(filter(None, [rule.strip() for rule in rules_clean]))
 
 
 def compute_confusion_matrix(
@@ -174,6 +181,9 @@ def score_output_json(
                             effective_line_num
                         )
             except ValueError:  # comment looked like a test annotation but couldn't parse
+                logger.warning(
+                    f"Could not parse {line} as a test annotation in file {test_file_resolved}. Skipping this line"
+                )
                 continue
 
     for result in json_out["results"]:
--- a/semgrep/semgrep/test.py
+++ b/semgrep/semgrep/test.py
@@ -156,20 +156,25 @@ def score_output_json(
             todo_ok_in_line = line_has_todo_ok(line)
             num_todo += int(todo_rule_in_line) + int(todo_ok_in_line)
 
-            if (not ignore_todo and todo_rule_in_line) or rule_in_line:
-                rule_ids = normalize_rule_ids(line)
-                for rule_id in rule_ids:
-                    ruleid_lines[test_file_resolved][rule_id].append(effective_line_num)
-            if (not ignore_todo and todo_rule_in_line) or ok_in_line:
-                rule_ids = normalize_rule_ids(line)
-                for rule_id in rule_ids:
-                    ok_lines[test_file_resolved][rule_id].append(effective_line_num)
-            if ignore_todo and todo_ok_in_line:
-                rule_ids = normalize_rule_ids(line)
-                for rule_id in rule_ids:
-                    todo_ok_lines[test_file_resolved][rule_id].append(
-                        effective_line_num
-                    )
+            try:
+                if (not ignore_todo and todo_rule_in_line) or rule_in_line:
+                    rule_ids = normalize_rule_ids(line)
+                    for rule_id in rule_ids:
+                        ruleid_lines[test_file_resolved][rule_id].append(
+                            effective_line_num
+                        )
+                if (not ignore_todo and todo_rule_in_line) or ok_in_line:
+                    rule_ids = normalize_rule_ids(line)
+                    for rule_id in rule_ids:
+                        ok_lines[test_file_resolved][rule_id].append(effective_line_num)
+                if ignore_todo and todo_ok_in_line:
+                    rule_ids = normalize_rule_ids(line)
+                    for rule_id in rule_ids:
+                        todo_ok_lines[test_file_resolved][rule_id].append(
+                            effective_line_num
+                        )
+            except ValueError:  # comment looked like a test annotation but couldn't parse
+                continue
 
     for result in json_out["results"]:
         reported_lines[str(Path(result["path"]).resolve())][result["check_id"]].append(
--- a/semgrep/semgrep/test.py
+++ b/semgrep/semgrep/test.py
@@ -35,13 +35,20 @@ SAVE_TEST_OUTPUT_JSON = "semgrep_runs_output.json"
 SAVE_TEST_OUTPUT_TAR = "semgrep_runs_output.tar.gz"
 
 
-def normalize_rule_id(line: str) -> str:
+def normalize_rule_ids(line: str) -> Set[str]:
     """
     given a line like `     # ruleid:foobar`
     or `      // ruleid:foobar`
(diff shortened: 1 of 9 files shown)


minusworld:
Closes https://github.com/returntocorp/semgrep/issues/2806

semgrep/semgrep/test.py
```diff
@@ -109,6 +116,8 @@ def line_has_ok(line: str) -> bool:
         or "# ok:" in line
         or "//ok:" in line
         or "// ok:" in line
+        or "<!--ok:" in line
+        or "<!-- ok:" in line
```
minusworld: Also allows `ok` in HTML comments in this PR. Wasn't in there before

aryx:
Only test failure seems actually related to the PR ...:
FAILED tests/e2e/test_cli_test.py::test_cli_test_multiline_annotations - Asse...

semgrep/semgrep/test.py
mschwager: Should we do the same for `line_has_todo_rule` and `line_has_todo_ok`?

semgrep/semgrep/test.py
```diff
@@ -35,13 +35,20 @@
...
     """
-    return line.strip().split(":")[1].strip().split(" ")[0].strip()
+    _, rules_text = line.strip().split(":")
+    rules_text = rules_text.strip()
+    rules = rules_text.split(",")
+    # strictly remove ending HTML comment
+    rules = [rule.strip().replace("-->", "") for rule in rules]
+    # strictly remove ending OCaml comment
+    rules = [rule.strip().replace("*)", "") for rule in rules]
+    return set(filter(None, [rule.strip() for rule in rules]))
```
mschwager: You could simplify to:

```
return {rule.strip() for rule in rules if rule.strip()}
```

semgrep/semgrep/test.py
```diff
@@ -160,17 +169,19 @@ def score_output_json(
             num_todo += int(todo_rule_in_line) + int(todo_ok_in_line)
 
             if (not ignore_todo and todo_rule_in_line) or rule_in_line:
-                ruleid_lines[test_file_resolved][normalize_rule_id(line)].append(
-                    effective_line_num
-                )
+                rule_ids = normalize_rule_ids(line)
+                for rule_id in rule_ids:
```
mschwager: This could be:

```
ruleid_lines[test_file_resolved][rule_id].extend(rule_ids)
```

semgrep/semgrep/test.py
```diff
@@ -160,17 +169,19 @@ def score_output_json(
...
+                    ruleid_lines[test_file_resolved][rule_id].append(effective_line_num)
             if (not ignore_todo and todo_rule_in_line) or ok_in_line:
-                ok_lines[test_file_resolved][normalize_rule_id(line)].append(
-                    effective_line_num
-                )
+                rule_ids = normalize_rule_ids(line)
+                for rule_id in rule_ids:
```
mschwager: Same as above --^

semgrep/tests/unit/test_semgrep_test.py
```diff
@@ -0,0 +1,69 @@
...
+
+# cf. https://docs.python.org/3/library/itertools.html#itertools-recipes
+def powerset(iterable):
+    "powerset([1,2,3]) --> () (1,) (2,) (3,) (1,2) (1,3) (2,3) (1,2,3)"
+    s = list(iterable)
+    return chain.from_iterable(combinations(s, r) for r in range(len(s) + 1))
+
+
+COMMENT_SYNTAXES = (("#", "\n"), ("//", "\n"), ("<!--", "-->"), ("(*", "*)"))
+ANNOTATIONS = ("ruleid", "ok")
```
mschwager: Looks like we're missing the TODO equivalents.

mschwager (commented):
We should probably remove this comment:

```
# TODO: -- re-enable this
# assert len(set(reported_lines[file_path][check_id])) == len(
#    reported_lines[file_path][check_id]
# ), f"for testing, please don't make rules that fire multiple times on the same line ({check_id} in {file_path} on lines {reported_lines[file_path][check_id]})"
```

cc @ievans who originally included it, and on whether it's important or not.

semgrep/tests/unit/test_semgrep_test.py
```diff
@@ -0,0 +1,69 @@
...
+from semgrep.test import normalize_rule_ids
+
+# cf. https://docs.python.org/3/library/itertools.html#itertools-recipes
+def powerset(iterable):
+    "powerset([1,2,3]) --> () (1,) (2,) (3,) (1,2) (1,3) (2,3) (1,2,3)"
+    s = list(iterable)
+    return chain.from_iterable(combinations(s, r) for r in range(len(s) + 1))
+
+
+COMMENT_SYNTAXES = (("#", "\n"), ("//", "\n"), ("<!--", "-->"), ("(*", "*)"))
```
mschwager: It seems overly fragile to be redefining these here. I.e. now we'll have to update two places when updating this information. Can we define these in one place, like `test.py`, and use them from there?

IMO we should define the language comment syntax's and autogenerate all the annotations from there. E.g. `['#', '//', '(*', ...]` -> `['# ruleid:', '// ruleid:', ..., '# ok:', '// ok:', ...]`. Then we'll also be less likely to forget annotation definitions for certain comment syntax's.

semgrep/semgrep/test.py
minusworld: I'm not really sure we use the feature anymore, but I guess there's no harm in adding it 👍

semgrep/semgrep/test.py
minusworld: I wanted to remove empty strings though, which `filter(None, ...)` does

semgrep/tests/unit/test_semgrep_test.py
minusworld: Yeah, that's a good idea 👍

semgrep/semgrep/test.py
mschwager: The `if rule.strip()` portion of the comprehension will remove empty strings:

```
In [49]: {s.strip() for s in ['   ', '', ' test'] if s.strip()}                                         
Out[49]: {'test'}
```

semgrep/semgrep/test.py
mschwager: I see a few places:

```
$ rg 'todook' semgrep-rules | wc -l
3
$ rg 'todoruleid' semgrep-rules | wc -l
19
```

semgrep/tests/unit/test_semgrep_test.py
minusworld: Done!

semgrep/tests/unit/test_semgrep_test.py
minusworld: Done

semgrep/semgrep/test.py
minusworld: This is actually appending `effective_line_num` so I left it as-is for now

semgrep/semgrep/test.py
```diff
@@ -34,14 +35,28 @@
 SAVE_TEST_OUTPUT_JSON = "semgrep_runs_output.json"
 SAVE_TEST_OUTPUT_TAR = "semgrep_runs_output.tar.gz"
 
+COMMENT_SYNTAXES = (("#", "\n"), ("//", "\n"), ("<!--", "-->"), ("(*", "*)"))
```
mschwager, line 41: Noice!

semgrep/semgrep/test.py
```diff
@@ -34,14 +35,28 @@
...
     """
     given a line like `     # ruleid:foobar`
     or `      // ruleid:foobar`
     return `foobar`
     """
-    return line.strip().split(":")[1].strip().split(" ")[0].strip()
+    _, rules_text = line.strip().split(":")
+    rules_text = rules_text.strip()
+    rules = rules_text.split(",")
+    # strictly remove ending HTML comment
```
mschwager: If you want to generalize this so we don't need to remember to update these as well we could do something like:

```python
rules = [
    rule.strip() if end == "\n" else rule.strip().replace(end, "")
    for start, end in COMMENT_SYNTAXES
]
```

Then we could also remove the `rule.strip()` call in the `return` statement.

semgrep/semgrep/test.py
```diff
@@ -79,50 +94,32 @@ def _test_compute_confusion_matrix() -> None:
...
-    return (
-        "#todoruleid:" in line
-        or "# todoruleid:" in line
-        or "// todoruleid:" in line
-        or "//todoruleid:" in line
-        or "(*todoruleid:" in line
-        or "(* todoruleid:" in line
-    )
+    todo_rule_annotations = _annotations(TODORULEID)
+    return any([annotation in line for annotation in todo_rule_annotations])
```
mschwager: Noice, I love all the code removal.

Fun Pythonism/nitpick: using a generator instead of a list in `any`/`all` calls will be more memory efficient and produce the same result. E.g. `any(i for i in iterable)` vs. `any([i for i in iterable])`. Since these lists are small it won't have much of an effect, but for larger lists/datasets it will. This is because the list comprehension builds the whole list in-memory, then calls `any`. A generator only holds a single element in memory at a time, and since `any`/`all` will exit early if they can it becomes quite a bit more efficient :+1: 

Not a huge deal here, but using generator statements with these calls will always be more efficient with no downside.

semgrep/semgrep/test.py
```diff
@@ -159,18 +156,25 @@ def score_output_json(
...
-                ok_lines[test_file_resolved][normalize_rule_id(line)].append(
-                    effective_line_num
-                )
-            if ignore_todo and todo_ok_in_line:
-                todo_ok_lines[test_file_resolved][normalize_rule_id(line)].append(
-                    effective_line_num
-                )
+            try:
+                if (not ignore_todo and todo_rule_in_line) or rule_in_line:
+                    rule_ids = normalize_rule_ids(line)
```
mschwager, line 168: We can pull the `rule_ids = normalize_rule_ids(line)` line out of the `if` statements since it's called the same way in each.

semgrep/semgrep/util.py
```diff
@@ -5,6 +5,8 @@
 import subprocess
 import sys
 from pathlib import Path
+from itertools import chain
```
mschwager: We already have the `itertools` module imported, should just use `itertools.chain`, etc? :man_shrugging:

semgrep/semgrep/test.py
```diff
@@ -159,18 +156,25 @@ def score_output_json(
...
+                    for rule_id in rule_ids:
+                        ok_lines[test_file_resolved][rule_id].append(effective_line_num)
+                if ignore_todo and todo_ok_in_line:
+                    rule_ids = normalize_rule_ids(line)
+                    for rule_id in rule_ids:
+                        todo_ok_lines[test_file_resolved][rule_id].append(
+                            effective_line_num
+                        )
+            except ValueError:  # comment looked like a test annotation but couldn't parse
```
mschwager, line 183: Where is this `ValueError` expected to occur from? I suspect it'll be confusing if we silently fail for malformed annotations when people are expecting them to be testing their rules. Should we fail louder?

semgrep/semgrep/test.py
minusworld: Yeah, we could, but then it runs on every line rather than only the lines with comment syntaxes. It looks factorable, but I decided to leave it that way.

semgrep/semgrep/test.py
minusworld: I'll add a logging comment 👍

semgrep/semgrep/test.py
minusworld: Sounds like a good CR-style Semgrep rule ;)

semgrep/semgrep/test.py
```diff
@@ -31,17 +33,37 @@
...
     or `      // ruleid:foobar`
     return `foobar`
     """
-    return line.strip().split(":")[1].strip().split(" ")[0].strip()
+    _, rules_text = line.strip().split(":")
+    rules_text = rules_text.strip()
+    rules = rules_text.split(",")
+    # remove comment ends for non-newline comment syntaxes
+    rules_clean = map(lambda rule: _remove_ending_comments(rule), rules)
+    return set(filter(None, [rule.strip() for rule in rules_clean]))
```
mschwager, line 66: We don't need the `.strip()` call after `_remove_ending_comments`.

semgrep/tests/e2e/test_cli_test.py
```diff
@@ -35,3 +35,16 @@ def test_cli_test_suffixes(run_semgrep_in_tmp, snapshot):
...
+def test_cli_test_multiline_annotations(run_semgrep_in_tmp, snapshot):
+    results = run_semgrep_in_tmp(
+        "rules/cli_test/multiple_annotations/",
+        options=["--test"],
+        target_name="cli_test/multiple_annotations/",
+        output_format="text",
+    )
+    snapshot.assert_match(
+        results,
+        "results.json",
```
mschwager, line 49: This isn't actually outputting JSON - I would either change `output_format="json"` or use `results.txt` :+1:

mschwager (approved):
LGTM, lots of improvements!

semgrep/tests/e2e/targets/cli_test/multiple_annotations/multiple-annotations-bad.py
```diff
@@ -0,0 +1,8 @@
+# ruleid: basic-test
```
aryx, line 1: why this file was called multiple-annotations-bad.py? Looks like it's actually the correct file, and the
file below (multiple-annotations.py) contains bad syntax

aryx: it's also not even used in the test code so I wonder why it was added in the PR.

aryx: fixed here I think: https://github.com/semgrep/semgrep/pull/10189

PR #2927, opened 2021-04-13
Single array elements also are parsed as individual field

### LINKED ISSUE #2926: [Yaml] the '- "String"' YAML pattern does not match the yaml content
https://semgrep.dev/s/GWpn

emjin:
Closes #2926
    
It is convenient in a semgrep pattern to consider a single array element, like `- "Job/Read"`, to be a pattern matching a field of an array rather than an array with a single element
    
PR checklist:
- [x] changelog is up to date


semgrep-core/src/parsing/other/yaml_to_generic.ml
```diff
@@ -224,6 +224,7 @@ let make_pattern_expr e =
   (* If the user creates a pattern with a single field, assume they just want *
    * to match the field, not the whole enclosing container                    *)
   | A.Container(A.Dict, (_lp, [x], _rp)) -> A.E x
+  | A.Container(A.Array, (_lp, [x], _rp)) -> A.E x
```
aryx, line 227: However this will match anything, not just in an Array context. For Dict it was ok, because the
element was a Tuple(key,val) which can appear only in a Dict, but for Array it's not the case.

aryx (approved):

semgrep-core/src/parsing/other/yaml_to_generic.ml
emjin: hm should we make some sort of distinction then?

aryx: let's try like this, see if people complain :)

PR #2935, opened 2021-04-14
Run performance benchmarks on commit

emjin:
Restructure run-benchmarks so that it can be used in the test suite that runs on commit. In particular, we only want to run small tests, since these run every commit, so a flag (`small-only`) was necessary to specify that. Also, to reuse compare_perf.py (in scripts), added a flag (`save-to`) to specify a timings file.

Test plan: commit

PR checklist:
- [x] changelog is up to date


mjambon (approved):
very nice, thank you!

perf/run-benchmarks
```diff
@@ -379,9 +390,18 @@ def run_benchmarks(
...
+    print("\n".join(results_msgs))
+
+    if summary_file_path:
+        with chdir(called_dir):
+            summary_file = open(summary_file_path, "w+")
+            summary_file.write(durations)
+            summary_file.close()
+
     if plot_benchmarks:
+        import matplotlib.pyplot as plt
```
aryx, line 402: sometimes Python is convenient :)

mjambon: btw this package needs to be installed to unbreak the daily benchmarks: https://app.circleci.com/pipelines/github/returntocorp/semgrep/3265/workflows/b09dbf63-092e-4626-aaec-a9af8eec4e4d/jobs/3878

emjin: hm weird on github actions hiding it behind the if was fine

emjin: Also @mjambon are we already running the benchmarks nightly? I'm not seeing on the dashboard runs after 4/4 (other than the ones I did yesterday and today), but that might be when I broke it with matplotlib

emjin: I've been working on running the benchmarks nightly using github actions, so I might have duplicated work

emjin: Looking at the CircleCI benchmarks, are they running the latest run-benchmarks? The line number highlighted indicates that the import is still at the top of the file. (If not, we can update the run-benchmarks file being used and won't have to install the moedule!)

mjambon: If you navigate to the [list of jobs in CircleCI](https://app.circleci.com/pipelines/github/returntocorp/semgrep?branch=develop), you can see the matplotlib error is gone, but there's [another problem](https://app.circleci.com/pipelines/github/returntocorp/semgrep/3352/workflows/6941e8d2-5516-478d-b9ea-3a3952df1aed/jobs/3964).

emjin: hm, not sure why this one is happening---is there an easy command to reproduce this locally?

PR #2943, opened 2021-04-14
Upload benchmarks on pull request merge

emjin:
To track semgrep performance, run the small and medium benchmarks, with all variants, on each pull request merge (push to develop), where small is under 30 seconds and medium is under 3 minutes. We will then be able to plot semgrep perf changes over time

PR checklist:
- [ ] changelog is up to date


.github/workflows/upload-benchmarks.yml
```diff
@@ -0,0 +1,75 @@
+name: Upload-benchmarks
+
+on:
+  workflow_dispatch:
+  pull_request:
```
aryx: if we just want to do it on merge to develop, we should remove this line no? We should not do that on each PR?

.github/workflows/upload-benchmarks.yml
```diff
@@ -0,0 +1,75 @@
...
+
+jobs:
+  build-core:
+    name: Build semgrep-core
+    runs-on: ubuntu-latest
+    container: returntocorp/ocaml:alpine
+    steps:
+      - name: Pre-checkout fixes
+        run: |
+          sudo chmod -R 777 /github
```
aryx, line 17: what is this? Do we do that in the other workflow?

aryx (approved):

.github/workflows/upload-benchmarks.yml
emjin: That's just for testing---I'm actually going to change it to nightly

.github/workflows/upload-benchmarks.yml
emjin: Yes, this is copied from tests.yml, the build and test semgrep-core job.

aryx: put a comment then, like coupling: this is copy-pasted from tests.yml#build and test semgrep-core job

perf/bench/big-js/prep
```diff
@@ -36,6 +36,6 @@ shallow_clone() {
 name="big-js"
 #TODO: the other repo are using https: but this repo is currently private
 # so I use ssh
```
mjambon, line 38: ^ remove the old comment

mjambon (commented):
This is a lot of duplicated code that's going to be hard to maintain and slow to run. Compare with
https://github.com/returntocorp/semgrep/blob/develop/.circleci/config.yml
which uses the latest docker build from the develop branch. We should not have to rebuild semgrep each time we want to use it.

PR #2945, opened 2021-04-15
Add run-benchmarks --filter-corpus and --filter-variant

aryx:
This allows to run specific corpus with specific variants from the CLI

test plan:
$ ./run-benchmarks --filter-corpus netflix --filter-variant '^(std|experimental)$' --plot-benchmarks

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/perf/run-benchmarks
+++ b/perf/run-benchmarks
@@ -15,6 +15,7 @@ import argparse
 import copy
 import json
 import os
+import re
 import subprocess
 import time
 import urllib.request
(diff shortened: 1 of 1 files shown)


perf/run-benchmarks
```diff
@@ -439,6 +440,18 @@ def main() -> None:
     parser.add_argument(
         "--std-only", help="only run the default semgrep", action="store_true"
     )
+    parser.add_argument(
+        "--filter-corpus",
+        metavar="REGEXP",
```
emjin, line 448: I would include an example of a regex string you could use (like you gave in the commit) in the help message

perf/run-benchmarks
```diff
@@ -439,6 +440,18 @@ def main() -> None:
...
+        type=str,
+        help="run the corpus only if it satisfies the regexp",
+    )
+    parser.add_argument(
+        "--filter-variant",
+        metavar="REGEXP",
```
emjin, line 454: Similarly here

emjin (approved):

perf/run-benchmarks
mjambon: Perhaps a `list` subcommand (or `--list` option) would be useful. It would list the name of the corpuses and variants in alphabetical order, so we know what to filter for without inspecting the source code.

mjambon (approved):

PR #2953, opened 2021-04-16
Improve stability of CI perf

emjin:
Retire perf-test job (the juice-shop test will never alarm because we are significantly faster, the dvna test can be subsumed into run-benchmarks) for simplicity and to avoid potential conflicting.

Run tests multiple times, take the min of ours and the max of theirs.

Ignore minor differences in small tests.

PR checklist:
- [ ] changelog is up to date


.github/workflows/tests.yml
```diff
@@ -154,58 +154,59 @@ jobs:
...
-        with:
-          python-version: ${{ matrix.python }}
-      - name: Run 0.29.0 Timing Benchmark
-        run: |
-          pip3 install semgrep==0.29.0
-          semgrep --version
-          semgrep-core -version
-          python3 -m semgrep --version
-          export PATH=/github/home/.local/bin:$PATH
+  # perf-test:
```
aryx: if you comment all of this, you need to put a comment why you did that.

aryx (approved):

.github/workflows/tests.yml
emjin: Oops I meant to delete this since you've added the other two tests to bench

PR #2958, opened 2021-04-17
feat: limit max characters in output line

### LINKED ISSUE #2159: Max line length for any finding
In a minified JS file, showing the "matched line" will basically mean the whole file is printed. This is pretty painful to navigate around. And since every match in the file is a new finding, the whole file is printed again and again.

Ideally, I would like to only show the matched section (the part that semgrep highlights/bolds) if its small and the whole line is >200chars. And if matched section is also massive, then semgrep just says "redacted due to length". A CLI flag can then allow enabling this display.

Another option is just default to never showing any line output > 200 chars and saying "Redacted".
 
**Additional context**

ankush:
- limit max characters in line
- add flag to control maximum characters, defaults to 160

Screenshot (color changed to yellow temporarily):

semgrep_strip_lines

closes #2159


semgrep/semgrep/output.py
```diff
@@ -97,7 +101,26 @@ def finding_to_line(
                 else:
                     line_number = f"{start_line + i}"
 
+                if per_line_max_chars_limit and len(line) > per_line_max_chars_limit:
+                    stripped = True
+                    is_first_line = i == 0
+                    if is_first_line:
+                        line = (
+                            line[
+                                start_col - 1 : start_col - 1 + per_line_max_chars_limit  # type: ignore
```
brendongo, line 110: Why is type checking ignored here?

brendongo: Oh I see. We do a dictionary get on line 82-83. No change necessary then will fix at some point in the future

semgrep/semgrep/output.py
```diff
@@ -97,7 +101,26 @@ def finding_to_line(
...
+                            ]
+                            + ELLIPSIS_STRING
+                        )
+                        if start_col > 1:  # type: ignore
+                            line = ELLIPSIS_STRING + line
+                    else:
+                        line = line[:per_line_max_chars_limit] + ELLIPSIS_STRING
+                    line = line + colorama.Style.RESET_ALL
```
brendongo, line 119:
```suggestion
```
Is this a left over from having the line be yellow in the screen shot?

brendongo (changes requested):

semgrep/semgrep/output.py
ankush: No, this was added to handle a case where a section is highlighted but trailing RESET_ALL is cut off. In which case following text will also get coloured untill another RESET_ALL is encountered. I am not sure if this is most graceful way to handle it.


I tried to handle over-all stripping of lines in two ways, before coloring and after coloring. Doing it before coloring is more complicated because of the way "coordinates" are passed to function that colours it.

ankush: Now that I think about it, better way would be to add it in last matched line and not universally.

Sounds good?

brendongo: Hmm i see. Can you add a comment above this line mentioning that it is to handle that case

ankush: Done @brendongo

ievans (approved):
Looks good to me, thanks @ankush! Maybe add a test case if you have time?

ankush:
@ievans I'll try to push it in future, right now I don't have much free time. 😐

ievans:
@brendongo want to merge?

brendongo (approved):

PR #2968, opened 2021-04-19
Use the semgrep-dev container

mjambon:
* This adds instructions on how to use the `semgrep-dev` container.
* This also sets the current directory to the home directory instead of `/` in the `semgrep-dev` container.
* I realized that we were not using the `semgrep-dev` extended image for the daily benchmarks. Now we do.
* The dummy benchmarks should now run according to the default schedule, which includes pull requests. Update: they ran and passed on this PR, yay!

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/perf/README.md
+++ b/perf/README.md
@@ -64,6 +64,10 @@ for CI jobs which run more or less in a consistent environment.
 Troubleshooting CI with the semgrep-dev Docker image
 --
 
+_CI uses CircleCI or GitHub Actions, configured in the standard places
+(`.circleci`, `.github/workflows`). See those files to determine which
+jobs run and when._
+
 We maintain a Docker build that comes with `semgrep`
 pre-installed. It is meant for daily benchmarks and other
 jobs that use the development version of semgrep. The image URL is
--- a/perf/README.md
+++ b/perf/README.md
@@ -60,3 +60,37 @@ $ make
 
 This will *not upload* the results to the dashboard, as it is reserved
 for CI jobs which run more or less in a consistent environment.
+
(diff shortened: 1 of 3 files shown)


perf/README.md
```diff
@@ -60,3 +60,37 @@ $ make
 
 This will *not upload* the results to the dashboard, as it is reserved
 for CI jobs which run more or less in a consistent environment.
+
+Troubleshooting CI with the semgrep-dev Docker image
```
emjin, line 64: It might be useful to previously state that we run these benchmarks nightly, and give the path to the CircleCI config file, since we have the docker troubleshooting info

perf/README.md
```diff
@@ -60,3 +60,37 @@ $ make
...
+$ docker run -it returntocorp/semgrep-dev:develop  # starts bash in container
+```
+
+If you want to test some of your local code inside the container, you
+would mount your folder using the `-v` option. The usage is
+`-v SRC:DST` where SRC is an absolute path to your original folder and DST is
+the absolute path you want it to have in the container. The home
+folder for this image is set to `/home/semgrep`. For example you'd do
+this:
+
```
emjin, line 89: Thank you for this!

emjin (approved):

perf/README.md
mjambon: I didn't mention specifics about CI because (1) this is independent from the CI setup and (2) we will forget to update this document when the schedule or the nature of the jobs changes.

mjambon: I added a generic note about where to find the CI config.

PR #2969, opened 2021-04-19
Skip tree-sitter installation if in homebrew

brendongo:
tree-sitter exists as a homebrew formula. Can rely on that instead
of installing tree-sitter from source when building in homebrew

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/scripts/install-ocaml-tree-sitter
+++ b/scripts/install-ocaml-tree-sitter
@@ -5,5 +5,13 @@
 set -eu
 
 cd ocaml-tree-sitter
-./scripts/install-runtime
+
+# In homebrew, tree-sitter formula is used as dependency
+# instead of installing manually so skip installing tree-sitter-lib
+if [[ -z "${HOMEBREW_SYSTEM+x}" ]]; then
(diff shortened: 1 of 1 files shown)


brendongo:
Relevant removal of special case in ocaml-tree-sitter repo: https://github.com/returntocorp/ocaml-tree-sitter/pull/194

mjambon:
This is fine iff this installs the correct version of libtree-sitter, the one that matches what we used to generate the tree-sitter parsers. This is currently `0.19.4`. It might work with past and future bugfix releases (`0.19.x`). It wouldn't work with `0.18`.

The tree-sitter version we need is currently set in [this script](https://github.com/returntocorp/ocaml-tree-sitter/blob/e5bbdda167c9eba5fb60fb71d58120a1ec7bf276/scripts/install-tree-sitter-lib#L7). We could move this version ID to its own file if that helps.

mjambon:
^ Follow up at https://github.com/returntocorp/semgrep/issues/2989

brendongo:
Going to revert this

PR #2970, opened 2021-04-19
Add symlink to LICENSE so setup.py includes it in wheel

### LINKED ISSUE #2957: Provide a LICENSE file along with pypi release
Currently it can only be assumed from meta data under what license the pypi release is shipped.
INAL but I think LGPL somehow requires to ships the LICENSE text along with a release, which isn't the case if just pulled from pypi

brendongo:
Closes: https://github.com/returntocorp/semgrep/issues/2957


mschwager:
Do we need to include it in `MANIFEST.in` as well?

brendongo:
Nope setuptools automatically detects LICEN[SC]E[.txt|.md] files

mschwager:
> Nope setuptools automatically detects LICEN[SC]E[.txt|.md] files

Oooh, noice

mschwager (approved):

PR #2973, opened 2021-04-20
Let RuleMatch handle retrieving line information

### LINKED ISSUE #2972: nosemgrep comment doesn't work with pattern-regex
**Describe the bug**

Rules using pattern-regex (such as [generic.secrets.security.detected-aws-account-id](https://semgrep.dev/r/generic.secrets.security.detected-aws-account-id)) cannot be muted with a `# nosemgrep` comment.

**To Reproduce**

https://semgrep.dev/s/okZ6

**Expected behavior**

The nosem comments should be picked up so these lines don't return a finding.

P0 from user on Slack: https://r2c-community.slack.com/archives/CK86BJ5DW/p1618918412288900?thread_ts=1618917327.288400&cid=CK86BJ5DW

mschwager:
Fixes #2972.

We currently have many ways of retrieving and modifying `lines` information. This caused a subtle bug in #2972 where `nosemgrep` didn't work because it didn't have the additional `lines` context including the comment. We should let `RuleMatch` gather this information for all findings (e.g. `pattern-regex`, `semgrep-core`, `spacegrep`, etc).

### DIFF
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -93,7 +93,7 @@ def get_re_matches(
                     "line": _offset_to_line_no(match.end(), contents),
                     "col": _offset_to_col_no(match.end(), contents),
                 },
-                "extra": {"lines": [contents[match.start() : match.end()]]},
+                "extra": {},
             }
         )
         for pattern_id, pattern in patterns_re
--- a/semgrep/semgrep/rule_match.py
+++ b/semgrep/semgrep/rule_match.py
@@ -109,9 +109,6 @@ class RuleMatch:
 
         Assumes file exists.  Note that start/end line is one-indexed
         """
-        if "lines" in self.extra:
-            return self.extra["lines"]
-
         with self.path.open(
             buffering=1, errors="replace"
         ) as fin:  # buffering=1 turns on line-level reads
--- a/semgrep/semgrep/rule_match.py
+++ b/semgrep/semgrep/rule_match.py
@@ -31,6 +31,7 @@ class RuleMatch:
     _start: Dict[str, Any] = attr.ib(repr=str)
     _end: Dict[str, Any] = attr.ib(repr=str)
     _extra: Dict[str, Any] = attr.ib(repr=False)
+    _lines_cache: Dict[str, Dict[str, List[str]]] = attr.ib(repr=False)
 
     # optional attributes
     _is_ignored: Optional[bool] = attr.ib(default=None)
@@ -64,6 +65,7 @@ class RuleMatch:
             start,
             end,
             extra,
+            {},
         )
 
     @property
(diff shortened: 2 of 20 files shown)


semgrep/semgrep/core_runner.py
```diff
@@ -93,7 +93,7 @@ def get_re_matches(
                     "line": _offset_to_line_no(match.end(), contents),
                     "col": _offset_to_col_no(match.end(), contents),
                 },
-                "extra": {"lines": [contents[match.start() : match.end()]]},
+                "extra": {},
```
mschwager, line 96: This was the actual bug. The `lines` information only included the matched portion of the regex, not the whole line. Letting `RuleMatch` grab this information for us fixes the issue.

semgrep/semgrep/rule_match.py
```diff
@@ -109,9 +109,6 @@ def lines(self) -> List[str]:
 
         Assumes file exists.  Note that start/end line is one-indexed
         """
-        if "lines" in self.extra:
```
mschwager, line 112: Let's use the file contents as the source of truth, like we do for metavariables.

semgrep/semgrep/spacegrep.py
```diff
@@ -88,13 +87,6 @@ def run_spacegrep(
                     output_json["matches"] = _patch_id(
                         pattern, output_json.get("matches", [])
                     )
-                    # dedent lines
```
mschwager, line 91: To be consistent with other parts of the codebase, let's not dedent here. We don't do this for regular patterns, e.g.

```
$ cat /tmp/test.py
if True:
    print(1 == 1)  # This is a comment
```

```
$ python -m semgrep --json -e '$X == $X' -l python /tmp/test.py
{
    "results": [
        {
            "check_id": "-",
            "path": "/tmp/test.py",
            ...
            "extra": {
                ...
                "lines": "    print(1 == 1)  # This is a comment"
            }
        }
    ],
    "errors": []
}
```

Note the whitespace in `lines` for a regular pattern.

minusworld (approved):
lgtm! Thanks

semgrep/semgrep/rule_match.py
```diff
@@ -109,13 +111,20 @@ def lines(self) -> List[str]:
...
-            return self.extra["lines"]
+        start_line = self.start["line"] - 1
+        end_line = self.end["line"]
```
nbrahms, line 118: Suggestion:

Rename this to `line_after_end` or also subtract 1

semgrep/semgrep/rule_match.py
```diff
@@ -109,13 +111,20 @@ def lines(self) -> List[str]:
...
-        ) as fin:  # buffering=1 turns on line-level reads
-            return list(itertools.islice(fin, self.start["line"] - 1, self.end["line"]))
+        try:
+            return self._lines_cache[start_line][end_line]
+        except KeyError:
+            pass
+
+        # buffering=1 turns on line-level reads
+        with self.path.open(buffering=1, errors="replace") as fd:
+            result = list(itertools.islice(fd, start_line, end_line))
```
nbrahms, line 127: With above suggestion, you'd either also rename `end_line` here, or add 1 back on.

nbrahms (approved):

semgrep/semgrep/rule_match.py
mschwager: Added a comment addressing - it's kinda confusing with the `islice` call below :+1:

PR #2978, opened 2021-04-21
analyzing: Infer global constant declarations in JS/TS

### LINKED ISSUE #554: Match variables inside JS Template literals
Couldn't figure out a way to match this.

https://semgrep.live/xvz

How to match line 2 precisely?

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

IagoAbal:
Even if the `const' qualifier is missing; it's enough if the variable is
resolved, it's set to a literal, and it's only assigned once.

Closes: #554

PR checklist:
- [ x] changelog is up to date

### DIFF
--- a/semgrep-core/src/analyzing/Constant_propagation.ml
+++ b/semgrep-core/src/analyzing/Constant_propagation.ml
@@ -116,15 +116,20 @@ let var_stats prog : var_stats =
 
   let hooks =
     { V.default_visitor with
-      V.kdef = (fun (k, _v) x ->
+      V.kdef = (fun (k, v) x ->
         match x with
         | { name = EN (Id (id,
                            { id_resolved = {contents = Some(_kind, sid)}; _})); _},
-          VarDef ({ vinit = Some _; _ }) ->
+          VarDef ({ vinit = Some e; _ }) ->
             let var = (H.str_of_ident id, sid) in
             let stat = get_stat_or_create var h in
             incr stat.lvalue;
-            k x
+            (* We can't do `k x` otherwise the variable-declaration itself is
+             * re-visited but now interpreted as an assignment (because of
+             * v_vardef_as_assign_expr in Visitor_AST), and we would be wrongly
+             * incrementing `stat.lvalue` a second time. Instead, we just need
+             * to visit the defining expression. *)
+            v (E e)
         | _ -> k x
       );
       V.kexpr = (fun (k, vout) x ->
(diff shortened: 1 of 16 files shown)


semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -116,15 +116,19 @@ let var_stats prog : var_stats =
...
         | { name = EN (Id (id,
                            { id_resolved = {contents = Some(_kind, sid)}; _})); _},
-          VarDef ({ vinit = Some _; _ }) ->
+          VarDef ({ vinit = Some e; _ }) ->
             let var = (H.str_of_ident id, sid) in
             let stat = get_stat_or_create var h in
             incr stat.lvalue;
-            k x
+            (* We can't do `k x` otherwise the variable-declaration itself is
+             * re-visited but now interpreted as an assignment, and we would be
```
aryx: Please, add "because of v_vardef_as_assign_expr in Visitor_AST.ml".
This v_vardef_as_assign_expr is maybe not a good thing ... We should maybe do that in Generic_vs_generic
instead of in the visitor.

aryx (approved):
Looks good. Need to look for the CI errors though.

IagoAbal:
There seem to be other semgrep-rules failing now... looking into it.

IagoAbal:
For the tests to pass here we need to update semgrep-rules, but updating semgrep-rules will make the tests fail with develop.

So I checked that all e2e tests passed except for `test_semgrep_rules_repo`: 
```
pytest tests/e2e/ -k 'not test_semgrep_rules_repo'
```
And I opened a PR to semgrep-rules with the required changes: https://github.com/returntocorp/semgrep-rules/pull/1269.
And I check that this modified Semgrep passes the tests for the modified semgrep-rules.

IagoAbal:
Wait, better run all QA and Unit tests too...

IagoAbal:
OK, unit tests pass too. QA tests, 127 pass and 17 "xfail", but those 17 "xfailures" also happen with develop so I guess it's fine?

aryx:
yes, xfail are ok.

semgrep/tests/e2e/snapshots/test_check/test_basic_rule__absolute/results.json
```diff
@@ -56,12 +56,12 @@
             "end": {
               "col": 14,
               "line": 3,
-              "offset": 25
+              "offset": 62
```
aryx, line 59: Why the offset changed after this PR?

aryx: Ah ok got it, because of the window.prompt() modification to the test file.

aryx (approved):

PR #3004, opened 2021-04-27
MVP for --watch

ievans:
As requested by many rule authors, I've implemented an MVP of a --watch mode for semgrep. It will re-run Semgrep if the local or remote config file changes. cc those who had mentioned this, including @smcpeak @devd @julianthome @DrewDennison 

PR checklist:
- [ ] changelog is up to date
- [ ] paginating and clearing the output


ievans:
after discussion with @mschwager we can't see this being an improvement over prefixing semgrep invocation with the unix `watch` utility. closing

mschwager:
Just to follow up, here's the command I've used:

```
$ watch -c -d python -m semgrep --config /tmp/test.yaml /tmp/test.py
```

Issue #2744, opened 2021-03-11
Produce a human-readable performance report from the output of --json-time


mjambon:
Semgrep now has a `--json-time` option that dumps one duration per pair (rule, target). On real code bases, this is a lot and it's not human-readable.

Let's create a script that takes this json output and produces a human-readable report. This report might include things like:
* total matching time
* total matching time per rule
* total matching time per target
* or only top 10 slowest targets by matching time
* or only top 10 slowest rules by matching time
* same as above but divided by the number of lines in the target
* other?

Parsing time is also probably useful, but it's complicated to obtain since we cache the AST, which should be cheaper than full parsing but also has its cost. We could report the total time spent by semgrep (not semgrep-core) on each target, and subtract the matching time. This would give us the time spent on "other things than matching", which could be substantial.

aryx:
sounds like a good start

Issue #2880, opened 2021-04-03
Tools integrating semgrep  programatically breaks as the API is being updated on releases.


ajinabraham:
**Is your feature request related to a problem? Please describe.**
njsscan was using semgrep 0.38 and a recent upgrade to 0.45 breaks tests

```
        output_handler = OutputHandler(
>           OutputSettings(
                output_format=OutputFormat.JSON,
                output_destination=None,
                error_on_findings=False,
                verbose_errors=False,
                strict=False,
                timeout_threshold=3,
                json_stats=False,
                output_per_finding_max_lines_limit=None,
            ),
            stdout=io_capture,
        )
E       TypeError: __new__() missing 1 required positional argument: 'json_time'

```

Current njsscan integration with semgrep: https://github.com/ajinabraham/libsast/blob/master/libsast/core_sgrep/helpers.py#L9

**Describe the solution you'd like**

Can we have two sets of APIs?

For seamless integration and maintenance, the semgrep API has to follow a contract with third-party tools. I suggest an internal and external API. The internal API changes and evolves with semgrep and the external one is fixed, documented, and changes only on major releases / a major security update.

**Describe alternatives you've considered**
NA

**Additional context**
NA

mschwager:
cc @mjambon who added the `json_time` functionality (https://github.com/returntocorp/semgrep/pull/2702). Is there a reason this isn't optional and defaulting to `False`?

@ajinabraham you can fix this by adding a `json_time=False` line to your `OutputSettings` call :+1:

ajinabraham:
I fixed it by adding `json_time`. @mschwager, it would be nice to have a stable API for programmatic integration with semgrep.

aryx:
@mschwager what about we have a semgrep_api_v1.py somewhere, with big comments at the top we can't change the signatures of the functions?

mschwager:
> @mschwager what about we have a semgrep_api_v1.py somewhere, with big comments at the top we can't change the signatures of the functions?

I like it, that sounds good to me :+1:

aryx:
For reference, I'm doing something similar on the generic AST side: https://github.com/returntocorp/semgrep/pull/2891
with a generic_ast_v1.atd

brendongo:
I think if we have https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/semgrep_main.py#L132 actually set fields based on arguments it would work as an API. This function is run in tests intending it to be the entry point for calling from other python files

[...]

PR #3016, opened 2021-04-28
Add useful information when reporting times

emjin:
Output a human readable time report when --time (formerly --json-time) is used without --json

Addresses https://github.com/returntocorp/semgrep/issues/2744

For every rule/file combination, report the amount of time spent parsing, matching, and in total on that file for that rule by semgrep-core. This will not capture all the time spent, because there is processing done on a per-rule basis, but because files are run in parallel it is hard to get a good comparison of times between the total time spent on each rule and the time spent on each file.

When --time is used without --json, create a short summary giving the total time, parse time, match time, and listing the 5 files that semgrep-core spent the longest on, as well as the 5 rules that used the most time

TODO (will make a separate PR): rule parse time can be substantial, needs to be captured

Test plan: run 
```
semgrep --config /Users/emma/workspace/semgrep/perf/bench/zulip/input/semgrep.yml /Users/emma/workspace/semgrep/perf/bench/zulip/input/zulip --strict --time --timeout 0 --verbose --no-git-ignore 
```

PR checklist:
- [x] changelog is up to date


emjin:
The failure is because I wanted to replace --json-time with --time (and dummy benchmarks runs on develop; run-benchmarks has been updated to pass --time). Should we keep --json-time around in case users have scripts relying on it? Or does that seem unlikely?

emjin:
https://github.com/returntocorp/semgrep/issues/2880 is relevant: I want to rename `json_time` to `output_time` in OutputSettings since that would be more accurate, but I don't want to break anyone's integration. @mschwager Should I just keep it as `json_time`?

semgrep/semgrep/core_runner.py
```diff
@@ -298,14 +298,18 @@ def _parse_core_output(
     def _add_match_times(
         self,
         rule: Rule,
-        match_time_matrix: Dict[Tuple[str, str], float],
+        match_time_matrix: Dict[Tuple[str, str], Tuple[float, float, float]],
```
mschwager, line 301: Can we turn this type into an object/class `Dict[Tuple[str, str], Tuple[float, float, float]]`? It's difficult to understand what each of these pieces of data represent with just the types, and makes for very long type hints. It'll also be easier to extend in the future if we have to :+1:

semgrep/semgrep/core_runner.py
```diff
@@ -635,7 +653,7 @@ def _run_rules_direct_to_semgrep_core(
                 )
         # end for rule, language ...
 
-        return outputs, {}, errors, set(Path(p) for p in target_manager.targets), {}
+        return outputs, {}, errors, all_targets, match_time_matrix
```
mschwager, line 656: It looks like the result here has changed. Are `set(Path(p) for p in target_manager.targets)` and `all_targets.union(targets)` the same thing?

semgrep/semgrep/output.py
```diff
@@ -409,7 +494,7 @@ class OutputSettings(NamedTuple):
     output_per_finding_max_lines_limit: Optional[int]
     output_per_line_max_chars_limit: Optional[int]
     json_stats: bool
-    json_time: bool
+    output_time: bool
```
mschwager, line 497: Yeah, this will potentially break downstream applications.

Technically `OutputSettings` isn't part of our officially exposed API, only `semgrep_main.invoke_semgrep`. However, we may want to come up with a better story here since external consumers seem to be running into trouble with this.

We could limit our API exposure here by making everything but `output_format` optional and setting a sensible default. That way consumers specifying `OutputSettings` don't have to explicitly set all these kwargs, and we can change them more easily. That doesn't solve this particular issue of moving to `output_time`, but it could limit the damage in the future.

cc @brendongo on thoughts here.

semgrep/semgrep/spacegrep.py
```diff
@@ -45,7 +54,7 @@ def run_spacegrep(
 ) -> dict:
     matches: List[dict] = []
     errors: List[dict] = []
-    targets_time: Dict[str, float] = {}
+    targets_time: Dict[str, Tuple[float, float, float]] = {}
```
mschwager, line 57: I just noticed all these kwargs fall prey to mutable default argument gotchas: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

semgrep/semgrep/core_runner.py
emjin: set(Path(p) for p in target_manager.targets) was incorrect because the Set here only contained the directory we were running. I believe all_targets is the list of targets actually run against

semgrep-core/src/cli/Main.ml
```diff
@@ -618,53 +618,58 @@ let iter_files_and_get_matches_and_exn_to_errors f files =
...
-                | Out_of_memory ->
-                    logger#info "OutOfMemory on %s" file;
-                    Error_code.OutOfMemory str_opt
-                | _ -> raise Impossible
-               )],
-            (file, 0.0)
-          )
-      | exn when not !fail_fast ->
-          [], [Error_code.exn_to_error file exn], (file, 0.0)
+      let (matches, errors, (file, parse_time, match_time)), run_time =
```
aryx, line 621: We need to define a record at this point. so you can write 
let matches_and_more = ...

type match_result = {
  matches: ...;
  errors: ...
  file: Common.filename;
  profiling: profiling;
}
and profiling = {
  parsing_time: Float;
  match_time: float;
  run_time: float
}

semgrep-core/src/cli/Main.ml
```diff
@@ -618,53 +618,58 @@ let iter_files_and_get_matches_and_exn_to_errors f files =
...
+                   (match exn with
+                    | Timeout ->
+                        logger#info "Timeout on %s" file;
+                        Error_code.Timeout str_opt
+                    | Out_of_memory ->
+                        logger#info "OutOfMemory on %s" file;
+                        Error_code.OutOfMemory str_opt
+                    | _ -> raise Impossible
+                   )],
+                (file, 0.0, 0.0)
```
aryx, line 666: maybe use a function like
  empty_profile file

semgrep/semgrep/core_runner.py
aryx: yep agree, we need to do the same also on the OCaml side. Those tuples are getting big.

aryx (approved):
Ok we need to cleanup the code with all those tuples on the Python and OCaml side, but this diff is already big, so let's do that in a separate PR.
I'll merge this but we need to cleanup after.

semgrep/semgrep/spacegrep.py
ievans: @mschwager didn't @brendongo write a semgrep rule for this?

PR #3018, opened 2021-04-28
Add: ignore .git directories when resolving targets to scan

brendongo:
.git directory contains a lot of objects that semgrep does not need
to scan. Because generic language can potentially scan any file with
any extension it does end up scanning the large object files in the .git
directory. This is normally not a problem since we only scan files that
are tracked by git but if `--no-git-ignore` flag is set it will end up scaning
the .git directory. With this PR, generic langauage should see a speed \
up when running on a directory with .git folders with `--no-git-ignore`

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/target_manager.py
+++ b/semgrep/semgrep/target_manager.py
@@ -230,9 +230,6 @@ class TargetManager:
 
         files, directories = partition_set(lambda p: not p.is_dir(), targets)
 
-        # Filter out .git directories
-        directories = set(d for d in directories if d.resolve().name != ".git")
-
         # Error on non-existent files
         explicit_files, nonexistent_files = partition_set(lambda p: p.is_file(), files)
         if nonexistent_files:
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/target_manager.py
```diff
@@ -237,7 +239,7 @@ def filtered_files(self, lang: Language) -> Set[Path]:
 
         targets = self.expand_targets(directories, lang, self.respect_git_ignore)
         targets = self.filter_includes(targets, self.includes)
-        targets = self.filter_excludes(targets, self.excludes)
+        targets = self.filter_excludes(targets, self.excludes + [".git"])
```
mschwager, line 242: Should we constantize this list of default excludes in a global variable?

aryx:
Can we merge this?

semgrep/semgrep/target_manager.py
brendongo: Can probably wait til we add more things

brendongo:
Note that with usual behavior .git is already ignored but if we are scanning a git repo with --no-git-ignore it also scans the entire .git repo which never seems like something we want to do anyway. There is some techdebt in our benchmarking code that makes it necessary to run with --no-git-ignore for now so this will make the data we get from benchmarking more like real world scanning

aryx (approved):

PR #3021, opened 2021-04-29
Release scripts

nbrahms:
PR to begin automating our release process

Author: Nathan Brahms <nathan@returntocorp.com>
Date:   Thu Apr 29 10:50:13 2021 -0700

    Add script to bump and push version change

    Automates steps 3-4 in release process.

Author: Nathan Brahms <nathan@returntocorp.com>
Date:   Wed Apr 28 09:22:23 2021 -0700

    Add cut release script

    Automates step 1 of our release process

### DIFF
--- a/scripts/release/util.py
+++ b/scripts/release/util.py
@@ -1,0 +1,54 @@
+import os
+import sys
+from typing import Any
+from typing import Callable
+from typing import Optional
+from typing import Sequence
+
+import click
+import sh
+
+
+_RELEASE_VERSION: Optional[str] = None
+
+
+def abort(code: int, message: str) -> None:
+    click.secho(message, err=True, fg="red")
+    sys.exit(code)
+
+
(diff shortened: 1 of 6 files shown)


scripts/release/util.py
```diff
@@ -0,0 +1,77 @@
+import os
+import sys
+from typing import Any
+from typing import Callable
+from typing import Optional
+from typing import Sequence
+
+import click
```
brendongo, line 8: Why are we using click?

brendongo (approved):

scripts/release/util.py
nbrahms: For `echo` mostly, but also if we want to add flags or options later it's easier than optparse.

I'm not really worried about dependency chain here since it's just for internal use.

brendongo: Yeah sounds good

PR #3022, opened 2021-04-29
Make ocaml-tree-sitter no longer a submodule

mjambon:
With this, `ocaml-tree-sitter` is no longer a git submodule of `semgrep`. This saves time and space when running `git submodule update --init --recursive`, by not fetching 650+ MB of tree-sitter-* repos (submodules of ocaml-tree-sitter) that we don't need for the semgrep-core build.

PR checklist:
- [x] changelog is up to date


brendongo (approved):
Looks good. And hombrew formula should continue working with this change

mjambon:
I made further changes so as to build the ocaml code of ocaml-tree-sitter directly as part of semgrep-core (like pfff and others) rather than making an opam install.

I also added bash code to avoid errors and clean things up during the transition.

scripts/install-ocaml-tree-sitter
```diff
@@ -4,6 +4,59 @@
 #
 set -eu
 
-cd ocaml-tree-sitter
-./scripts/install-runtime
-./configure
+# Temporary block to help with transition (2021-04-29):
+# ocaml-tree-sitter used to be a submodule, and if it's still there,
```
aryx, line 8: Very nice

.gitignore
```diff
@@ -166,3 +166,6 @@ TAGS
 # Do not check in compiled binaries
 semgrep/semgrep/bin/semgrep-core
 semgrep/semgrep/bin/spacegrep
+
+# Git repo that's not a submodule
+/ocaml-tree-sitter
```
aryx, line 171: Not super convinced of having this special case. I d rather have a submodule like ocaml-treesitter-lib-and-rntime

mjambon: hmm, maybe we should have a bare-bones git repo for the ocaml code of ocaml-tree-sitter, and use it as submodule in what's currently ocaml-tree-sitter.

aryx: Exactly!

mjambon: It requires a bit of restructuring otherwise I'd do it right away.

PR #3035, opened 2021-05-04
Parse empty yaml files

### LINKED ISSUE #2981: semgrep fails to parse empty yaml file
**Describe the bug**
semgrep will through the parse error if the target yaml file is empty.

**To Reproduce**
Can't not reproduce in the playground.
Can reproduce by `semgrep --verbose --pattern 'test:' --lang yaml empty.yaml`

**Expected behavior**
No error.
It seems that other languages (at least Python) won't throw error for the empty file.

P2

**Environment**
semgrep 0.48.0 on MacOS

emjin:
Closes https://github.com/returntocorp/semgrep/issues/2981

Test plan: 
```
touch empty.yaml
sc -f tests/generic/metavar_stmt.sgrep empty.yaml -lang yaml
```

Should return nothing and not crash

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - JS/TS: Infer global constants even if the `const` qualifier is missing (#2978)
 
 ### Fixed
+- Empty yaml files do not crash
+
 ### Changed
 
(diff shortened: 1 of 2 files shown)


brendongo:
Is it worth adding an empty yaml file to tests so future PRs dont break this functionality?

aryx (approved):
yep would be better with a test file, we could have semgrep-core/tests/yaml/empty.yaml and a empty.sgrep with $WHATEVER: $WHATEVER

PR #3042, opened 2021-05-05
analyzing: Do not assign constness to labels

### LINKED ISSUE #2805: Constant propagation causes =~"//" string matching to match assigned variable
**Describe the bug**
To be honest, I'm not sure if this is a bug to be fixed, or just unintuitive behavior. It was surprising to me as a user.

If you have a pattern like `"=~/http:///"` and are matching code like:

~~~python
a = "http://example.com"
~~~

The pattern will match that line **twice**, once for the right hand side string AND (the surprising part) the variable assigned, `a`. 

This caused me to believe Semgrep was duplicating results, though this constant propagation is perhaps intended behavior.

**To Reproduce**
https://semgrep.dev/s/clintgibler:http-leak-source-code

**Expected behavior**
I expected only the right hand side, `"http://example.com"` to match, but based on how constant propagation works, the current behavior might be the "correct behavior."

I created this issue to track this behavior because I think it's unintuitive for users, but I don't know if we want to "fix" it 🤷 

**Screenshots**

P2

**Environment**

IagoAbal:
Closes #2805

PR checklist:
- [x] changelog is up to date


aryx (approved):
Great!

PR #3045, opened 2021-05-05
Fix autofix for multiline patterns

chmccreery:
The yaml parser we use in semgrep keeps a newline character at the end of any multi-line output. 

Meanwhile, semgrep.dev/editor configs mostly return yaml with multi-line output (the | character). 

This PR strips up to one \n at the end of a yaml node, in order to accurately parse the message, autofix, etc. that was returned by semgrep.dev/editor.

This newline character was causing [a bug ](https://github.com/returntocorp/semgrep-app/issues/1999) with autofix where newline characters were inserted at the end of a match. The newline character was not causing a bug with general pattern matching, because semgrep-core strips the whitespace automatically.

PR checklist:
- [x] changelog is up to date

(co-written and debugged by Clara and Emma)


mschwager:
YAML provides an indicator for exactly this purpose: [Block Chomping Indicator](https://yaml.org/spec/1.2/spec.html#id2794534)

E.g. `message: |-` instead of `message: |`.

Rather than hardcode this functionality I think it'd be better to stick to the YAML spec, if we can. There could be some instances where users want this functionality, and want to distinguish between `|-` and `|`.

Here's what I'm seeing in the rule:

```
$ curl -L -H "User-Agent: Semgrep/0.0.0" https://semgrep.dev/s/clintgibler:php-cookie-autofix
rules:
- fix: |
    $wgRequest->getCookie($VALUE)
  id: my_pattern_id
  languages:
  - php
  message: |
    Directly accessing of $_COOKIE is not considered best practice, please use `$wgRequest->getCookie(...)` instead.
  pattern: |
    $_COOKIE[$VALUE]
  severity: WARNING
```

Could we instead turn this into `fix: $wgRequest->getCookie($VALUE)` or `fix: |- ...`? It seems to me that the code is doing the correct thing here and including a newline at the end because that's what the YAML is telling it to do.

What do you think?

mschwager:
I think turning `fix: |` -> `fix: |-` would be the best solution :+1:

emjin:
Oh interesting; I wasn't aware of that YAML syntax. `fix: |-` sounds good to me. Maybe we should just make this the default in all cases, since we never expect the extra new line, and I think it would be confusing to have a mix (most users probably don't know the distinction)

chmccreery:
Yeah, I didn't know about `|-` either, thanks for bringing it up @mschwager!

Although I agree with Emma that, at least in the case of autofix, since people may not know about `|-`, we might want to strip it in order to handle "fix" the same way that semgrep-core handles "pattern"s. It seems strange that these are not treated the same way. Until yesterday, I didn't know that autofix was handled in the python wrapper, so I was super confused why it was substituting the '\n' into the middle of the code but not demanding a '\n' to match in the code. 

(semgrep.live was also recently enhanced to use single-line when possible, so it would now be `fix: $wgRequest->getCookie($VALUE)` anyway I think)

mschwager:
> (semgrep.live was also recently enhanced to use single-line when possible, so it would now be fix: $wgRequest->getCookie($VALUE) anyway I think)

Yup, if I go to https://semgrep.dev/s/clintgibler:php-cookie-autofix and click "Advanced" it shows `fix` on a single line. If I run it and apply the fixes it works as expected now. Can we close this out? Or at least wait until someone complains about it here instead of the playground?

> since people may not know about |-, we might want to strip it in order to handle "fix" the same way that semgrep-core handles "pattern"s.

I don't think we want to unconditionally remove trailing newlines from all string values in our YAML file. It feels a bit heavy handed for something that's already been fixed at the source of issue, and, again, is according to the YAML spec. At the very least, if we're concerned about `fix` we should limit it to that :man_shrugging:

chmccreery:
I would be fine with limiting it to `fix` and still think that's the right thing to do for consistency with `pattern`.

This bug took a long time to trace back to its source, so my only concern with leaving it open until someone complains again, is that it may cost another couple of engineering days to figure out what the problem is. It is true that the immediate problem has been fixed (for now).

mschwager (approved):

PR #3046, opened 2021-05-05
Fix location of same-file findings-divider (--------) 

chmccreery:
When multiple findings exist for a file, we print a ------ between them. However, for autofix, we were then printing the autofix after that line, visually grouping it with the wrong item. 

The easiest fix is to move the autofix print above finding_to_line, which is where the concluding ----- gets printed. Alternatively, we could keep the logic for printing the divider one level up, in build_normal_output.

Before:
Screen Shot 2021-05-05 at 3 55 22 PM
After:
Screen Shot 2021-05-05 at 4 06 09 PM

PR checklist:
- [ ] changelog is up to date


mschwager (approved):
Nice! :bug: :hammer:

PR #3048, opened 2021-05-06
Add types for profiling structure

### LINKED ISSUE #3036: [tech debt] cleanup https://github.com/returntocorp/semgrep/pull/3016
We need to use proper types (record for OCaml/data class for Python?) for the profiling information on the Python side and OCaml side instead of having those 3-uples everywhere.
We should not use unzip4, we should have a proper sum_profiling_result or something.

emjin:
On the OCaml and the Python side, refactor to use a type for the result reporting and the profile data, respectively. Also, add --json-time to ensure backwards compatibility

Closes https://github.com/returntocorp/semgrep/issues/3036

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -17,6 +17,7 @@ module MR = Mini_rule
 module R = Rule
 module J = JSON
 module FT = File_type
+module RP = Reporting
 
 (*****************************************************************************)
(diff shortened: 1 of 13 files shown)


aryx (approved):
Perfect.
I would rename Reporting.ml Report.ml maybe, to be more consistent with the other files
in this directory, but no big deal.

PR #3051, opened 2021-05-06
Use OutputFormat in tests instead of strings

mschwager:


semgrep/tests/conftest.py
```diff
@@ -101,7 +102,7 @@ def _run_semgrep(
         stderr=subprocess.STDOUT if stderr else subprocess.PIPE,
     )
 
-    if output_format in {"json", "sarif"} and not stderr:
+    if output_format.is_json() and not stderr:
```
emjin, line 105: Is also including JSON_DEBUG here intended?

emjin (approved):

semgrep/tests/conftest.py
mschwager: `JSON_DEBUG` is gone after https://github.com/returntocorp/semgrep/pull/3050!

emjin: <3

brendongo (approved):

PR #3057, opened 2021-05-07
Always show semgrep-core error output + pass debug/verbose flag down to semgrep-core

mjambon:
This is a start to get better visibility on errors happening in semgrep-core.

Main changes:
* anything that semgrep-core prints to stderr is now printed on the standard output of semgrep
* `--verbose` or `--debug` result in the `-debug` flag being passed to semgrep-core

This combination makes `semgrep --debug` show the verbose log of semgrep-core, and does so without resorting to an environment variable.

@aryx I'm not confident about the changes I made in `_run_rules_direct_to_semgrep_core` because I haven't tested them by hand and I don't know if we have tests for it in CI.

PR checklist:
- [x] changelog is up to date

^ I still need to update the changelog.


semgrep/semgrep/util.py
```diff
@@ -154,8 +154,8 @@ def sub_run(cmd: List[str], **kwargs: Any) -> Any:
 def sub_check_output(cmd: List[str], **kwargs: Any) -> Any:
     """A simple proxy function to minimize and centralize subprocess usage."""
     # fmt: off
-    if QUIET:
-        kwargs = {**kwargs, "stderr": subprocess.DEVNULL}
+    #if QUIET:
```
brendongo: Why is this commented out?

semgrep/semgrep/core_runner.py
```diff
@@ -253,49 +256,98 @@ def _run_core_command(
...
+        logger.debug(core_run.stderr.decode("utf-8", errors="replace"))
+
+        semgrep_output = core_run.stdout.decode("utf-8", errors="replace")
+        semgrep_error_output = core_run.stderr.decode("utf-8", errors="replace")
+
+        # Always print semgrep-core's error output, which includes
+        # semgrep-core's logging if it was requested.
+        if semgrep_error_output != "":
+            name = f"[process {os.getpid()}, rule '{rule.id}']"
+            print(
```
brendongo: let's. use logger class so this goes to stderr and respects verbosity
```suggestion
            logger.debug(
```

semgrep/semgrep/core_runner.py
```diff
@@ -253,49 +256,98 @@ def _run_core_command(
...
+    ) -> Dict[str, Any]:
+        logger.debug(core_run.stderr.decode("utf-8", errors="replace"))
+
+        semgrep_output = core_run.stdout.decode("utf-8", errors="replace")
+        semgrep_error_output = core_run.stderr.decode("utf-8", errors="replace")
+
+        # Always print semgrep-core's error output, which includes
+        # semgrep-core's logging if it was requested.
+        if semgrep_error_output != "":
+            name = f"[process {os.getpid()}, rule '{rule.id}']"
```
brendongo: Why do we need the pid of the python process?

semgrep/semgrep/core_runner.py
```diff
@@ -253,49 +256,98 @@ def _run_core_command(
...
-                        f"unexpected json output while invoking semgrep-core with rule '{rule.id}':\n{PLEASE_FILE_ISSUE_TEXT}"
-                    )
+            output_json = self._extract_core_output(rule, patterns, core_run)
+            return output_json
 
-            output_json = self._parse_core_output(out_bytes, err_bytes, returncode)
+    def _extract_core_output(
+        self, rule: Rule, patterns: List[Pattern], core_run: subprocess.CompletedProcess
+    ) -> Dict[str, Any]:
+        logger.debug(core_run.stderr.decode("utf-8", errors="replace"))
```
brendongo: We decode stderr twice probably worth moving line 275 below to be above 272

semgrep/semgrep/core_runner.py
```diff
@@ -253,49 +256,98 @@ def _run_core_command(
...
+
+        semgrep_output = core_run.stdout.decode("utf-8", errors="replace")
+        semgrep_error_output = core_run.stderr.decode("utf-8", errors="replace")
+
+        # Always print semgrep-core's error output, which includes
+        # semgrep-core's logging if it was requested.
+        if semgrep_error_output != "":
```
brendongo, line 280: We probably don't want to print something just because error is not empty. There have been situations of a stray stderr output in semgrep-core but we want just json to come out of calling semgrep with --json --quiet for example. logger.debug suggestion below will respect verbosity in this case

brendongo (changes requested):
Should we also be making these changes on `_run_rules_direct_to_semgrep_core` so behavior is the same with `--optimizations`

semgrep/semgrep/util.py
mjambon: oops, sorry about that!

semgrep/semgrep/core_runner.py
mjambon: um yes, what I did is dumb. I wanted to show the pid of semgrep-core so we can match it with strace output that occurs earlier.

semgrep/semgrep/core_runner.py
mjambon: There shouldn't be stray logging in semgrep-core any more than in semgrep. If there is, we can fix it easily. It's just that I don't want to hide any non-fatal problem we may have in semgrep-core.

It should definitely obey `--quiet` though, and it looks like `logger.info` does just that.

mjambon: More specifically, I don't want to go back and forth with users and ask them to re-run their jobs with special flags when they run into a problem.

mjambon:
> Should we also be making these changes on `_run_rules_direct_to_semgrep_core` so behavior is the same with `--optimizations`

They both call `output_json = self._extract_core_output(rule, patterns, core_run)` which takes care of printing the stderr output captured by core_run.

semgrep/semgrep/core_runner.py
mjambon: I dropped the goal of printing the child's pid because I'd have to use Popen() and it seems more complicated than it's worth.

brendongo (approved):

mjambon:
I didn't go as far as I wanted but I'm going to move on to other tasks.

I still think it would be ideal to bundle up the input files (rules and targets; probably not the parsing cache) into a tarball with a standard name e.g. `repro.tar.gz` which could be uploaded as an artifact. I see that we do something similar in `test.py`. I'll add my thoughts to a github issue and we can implement this later.

PR #3068, opened 2021-05-10
Report the relative path of files instead of absolute

emjin:
Test plan:
run `semgrep --config /Users/emma/workspace/semgrep/perf/bench/dummy/input/dummy/rules /Users/emma/workspace/semgrep/perf/bench/zulip/input/zulip/zerver/ --time` in `semgrep/semgrep`

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/output.py
+++ b/semgrep/semgrep/output.py
@@ -1,6 +1,7 @@
 import contextlib
 import json
 import logging
+import pathlib
 import sys
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


brendongo:
Why are we changing the col line limit?

emjin:
Since it's not printing out the absolute path, I don't need as many characters, and it was more convenient for me to have it a bit shorter

mjambon (approved):

PR #3071, opened 2021-05-11
Add v0 opt-in aggregate metrics

brendongo:
This PR adds the ability to capture opt-in, anonymized metrics behind the environment variable `SEMGREP_SEND_METRICS` or the CLI flag `--enable-metrics`.

The motivation for this change is to prioritize engineering work improving Semgrep’s performance and language support. We’ve added a dizzying number of languages to over the past year (~13), and also became committers to the tree-sitter project, which powers parsing. This has significantly increased support load and development cost across parsers, generic AST, the matcher, and community rules. If someone contributes a slow rule, it’s difficult to trace the stack to understand which part is making it slow. We have aggressive performance goals for Semgrep (think: in-editor speed) so capturing these metrics helps us understand real-world performance and be data-driven in prioritization.

Our decisions around metrics are informed by these principles:

1. Transparency: Collect and use data in a way that is clearly explained to the user and benefits them
2. User control: Put users in control of their data at all times
3. Limited data: Collect what is needed, de-identify where possible, and delete when no longer necessary

The specific metrics collected, descriptions of fields, example values, and additional context are included in the PRIVACY.md file.

Note for GitLab SAST users: by default the security scanner semgrep-sast will turn on these metrics via an environment variable.

We’re opening a comment period to collect feedback about the metrics coming in v0.51.0. For future releases, we’re also considering adding more advanced profiling and metrics, and making the metrics opt-out instead of opt-in. We’d appreciate community input on that as well.

We’d deeply appreciate any input on this topic.

In sum, for now this PR only the following gathers anonymized aggregate metrics:

* semgrep version
* number of files scanned
* number of rules run
* number of findings ignored with nosem
* types of errors returned by semgrep-core
* return code
* total number of bytes scanned
* number of findings
* total run time

PR checklist:

*  changelog is up to date

### DIFF
--- a/semgrep/semgrep/metric_manager.py
+++ b/semgrep/semgrep/metric_manager.py
@@ -5,8 +5,6 @@ from typing import Dict
 from typing import List
 from typing import Optional
 
-import requests
-
 from semgrep.constants import SEMGREP_USER_AGENT
 
 METRICS_ENDPOINT = "https://stats.semgrep.dev"
@@ -78,6 +76,8 @@ class _MetricManager:
         Send metrics to the metrics server. Is a noop if SEMGREP_SEND_METRICS
         env var is not set
         """
+        import requests
+
         if os.environ.get("SEMGREP_SEND_METRICS"):
             metrics = self.as_dict()
             headers = {"User-Agent": SEMGREP_USER_AGENT}
--- a/semgrep/semgrep/metric_manager.py
+++ b/semgrep/semgrep/metric_manager.py
@@ -1,3 +1,4 @@
+import hashlib
 import logging
 from typing import Any
 from typing import Dict
@@ -5,6 +6,8 @@ from typing import List
 from typing import Optional
 
 from semgrep.constants import SEMGREP_USER_AGENT
+from semgrep.rule import Rule
+
 
 METRICS_ENDPOINT = "https://metrics.semgrep.dev"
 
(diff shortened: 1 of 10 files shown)


semgrep/semgrep/metric_manager.py
```diff
@@ -0,0 +1,95 @@
...
+logger = logging.getLogger(__name__)
+
+
+class _MetricManager:
+    """
+
+    Made explicit decision to be verbose in setting metrics instead
+    of something more dynamic (and thus less boiler plate code) to
+    be very explicit in what metrics are being collected
+    """
```
adamberman, line 26: Might be nice to put a comment describing how this class is meant to be used/why it is a singleton for ease of use. But also totally fine if you just want to follow up with a commit later for expediency :)

semgrep/semgrep/metric_manager.py
```diff
@@ -0,0 +1,95 @@
...
+            headers = {"User-Agent": SEMGREP_USER_AGENT}
+
+            try:
+                r = requests.post(
+                    METRICS_ENDPOINT, json=metrics, timeout=10, headers=headers
+                )
+                r.raise_for_status()
+                logger.info("Sent anonymized metrics.")
+            except Exception as e:
+                logger.info("Failed to send anonymized metrics.")
```
adamberman: should we log the exception?

adamberman (approved):
Approved, but two overall comments (totally fine to add later IMO): 1) we should add tests for this, and 2) have you manually tested with semgrep-agent to make sure things are still working there?

brendongo:
Waiting on privacy doc

DrewDennison (changes requested):
Let's review spec

DrewDennison (approved):
LGTM

adamberman:
LGTM too

PR #3081, opened 2021-05-12
Add --max-target-bytes option to semgrep and set default to 1 MB

mjambon:
This filters out any target larger than max_target_size, which is set to 1,000,000 bytes by default. This is a change from previous versions of semgrep, and I edited the changelog accordingly.

I added a parser for sizes like `1.5 MB`, `42KiB` and such because it's relatively simple compared to using a dedicated package for it. I wrote tests (which pass) but I don't think they will run until I put them in some file with the other unit tests.

PR checklist:
- [x] changelog is up to date


mjambon:
2 tests are failing. I think they're minified files so it would make sense that they're now excluded from the scan with the default options. Maybe I'll add `--max-target-size 0` to all the semgrep commands.

mjambon:
I just moved the two tests out of the "xfail" section. They should now pass since we're excluding the files that were minified incorrectly wrt recent javascript.

semgrep/semgrep/cli.py
```diff
@@ -190,6 +192,18 @@ def cli() -> None:
         ),
     )
 
+    config.add_argument(
+        "--max-target-size",
+        type=parse_size,
```
emjin, line 197: I was confused by this---is parse_size a class? My understanding was that the argument is casted to this type

semgrep/tests/unit/test_target_manager.py
```diff
@@ -93,6 +94,23 @@ def test_filter_exclude():
     assert len(TargetManager.filter_excludes(all_files, ["foo/*.go"])) == 17
 
 
+def test_filter_by_size():
```
emjin, line 97: Thanks for the test cases!

emjin (approved):

semgrep/semgrep/cli.py
nbrahms: yeah I'd expect `type=int` here.

nbrahms: oh nm looks like it's in semgrep.size

semgrep/semgrep/size.py
```diff
@@ -0,0 +1,63 @@
...
+        number, unit = tokens
+    else:
+        raise ValueError(f"Invalid representation for a number of bytes: '{input}'")
+    if unit in UNITS:
+        return int(float(number) * UNITS[unit])
+    else:
+        raise ValueError(f"Invalid representation for a number of bytes: '{input}'")
+
+
+TESTS = [
```
nbrahms: Please change:

This should go in the `tests/` directory.

semgrep/semgrep/size.py
```diff
@@ -0,0 +1,63 @@
...
+    ("7 GB", 7 * 1000 * 1000 * 1000),
+    ("8 TB", 8 * 1000 * 1000 * 1000 * 1000),
+    # test binary units
+    ("1512 KiB", 1512 * 1024),
+    ("7 GiB", 7 * 1024 * 1024 * 1024),
+    ("8 TiB", 8 * 1024 * 1024 * 1024 * 1024),
+]
+
+
+def test_parse_size() -> None:
```
nbrahms: Please change:

Move to a tests/ file

semgrep/semgrep/semgrep_main.py
```diff
@@ -160,6 +160,7 @@ def main(
     no_git_ignore: bool = False,
     timeout: int = DEFAULT_TIMEOUT,
     max_memory: int = 0,
+    max_target_size: int = 0,
```
nbrahms: Suggestion:

```suggestion
    max_target_bytes: int = 0,
```

Adding units to the variable name removes need for documentation

semgrep/semgrep/size.py
```diff
@@ -0,0 +1,63 @@
+#
```
nbrahms: Could also consider importing https://pypi.org/project/datasize/

semgrep/semgrep/target_manager.py
```diff
@@ -209,10 +210,29 @@ def filter_includes(arr: Set[Path], includes: List[str]) -> Set[Path]:
...
+        """
+        if max_target_size <= 0:
+            return arr
+        else:
+            return set(
+                path
+                for path in arr
+                if TargetManager._is_valid(path)
+                and os.path.getsize(path) <= max_target_size
+            )
```
nbrahms: Suggestion:

```suggestion
            return {
                path
                for path in arr
                if TargetManager._is_valid(path)
                and os.path.getsize(path) <= max_target_size
            }
```

nbrahms (changes requested):
Let's move the test code

semgrep/semgrep/target_manager.py
brendongo: instead of os.path.getsize(path) we can also call path.stat().st_size but I think functionally both do the same thing

semgrep/semgrep/semgrep_main.py
mjambon: nice, let's do that

semgrep/semgrep/cli.py
mjambon: yeah it has to be a function that takes a string. I might use the `datasize` package Nathan suggests in another thread; I saw an older one mentioned in stackoverflow answers and it seemed overkill.

mjambon:
It should be good to merge now. The option name is now `--max-target-bytes`.

Issue #3077, opened 2021-05-11
Enspeeden Gitlab eslint.direct-object-detection


nbrahms:
This rule is 10x slower than all the other rules from Gitlab. Let's make it fast so that Gitlab users have a better experience:

https://gitlab.com/gitlab-org/security-products/analyzers/semgrep/-/blob/main/rules/eslint.yml#L751

"Faster" in this case might mean that it runs in under 10 seconds on a modern Mac on returntocorp/semgrep-app.

brendongo:
Other numbers that might be relevant:
- Running above rule on our private semgrep-app repo takes 8 minutes without optimizations
- running above rule on same repo with optimizations takes 35 seconds
- For context running the other 92 rules in that rules repo takes 70 seconds without opttimizations

mschwager:
I wonder if we could speed this up with something like:

```yaml
  patterns:
    - pattern: $O[$ARG]
    - pattern-not: $O["..."]
    - pattern-not-inside: |
        $ARG = [$V];
        ...
        $O[$ARG]
    - pattern-not-inside: |
        $ARG = $V;
        ...
        $O[$ARG]
```

I think tightening the bounds on the `pattern-not-inside`s should speed things up. In my quick testing on `semgrep-app` the before/after is:

```
real	5m45.649s
user	6m58.732s
sys	0m5.394s
```

```
real	0m14.273s
user	0m27.280s
sys	0m3.010s
```

Looks like a 15x speedup?

mschwager:
Since `pattern-inside` and `pattern-not-inside` are inclusive the results should be the same.

brendongo:
Nathan didn't update the issue: he made a PR https://gitlab.com/gitlab-org/security-products/analyzers/semgrep/-/merge_requests/33/diffs

aryx:
What does enspeeden mean?

mschwager:
With the deep expression improvements from https://github.com/returntocorp/semgrep/pull/3083 we still see a ~15x speedup:

```yaml
patterns:
  - pattern: $O[$ARG]
  - pattern-not: $O["..."]
  - pattern-not-inside: |
      $ARG = [$V];
      ...
      <... $O[$ARG] ...>
  - pattern-not-inside: |
      $ARG = $V;
      ...
      <... $O[$ARG] ...>
  - metavariable-regex:
      metavariable: "$V"
      regex: "[0-9]+"
      metavariable: "$ARG"
      regex: "[^0-9]+"
```

```
[...]

PR #3086, opened 2021-05-12
JS/TS: Propagate constant definitions without declaration

IagoAbal:
Relates-to: #3077
    
test plan:
make test

PR checklist:
- [x] changelog is up to date


semgrep-core/tests/js/cp_undeclared.js
```diff
@@ -0,0 +1,3 @@
+foo = "undeclared variable"
+//ERROR:
+console.log(foo)
```
mschwager: Will this also work in a function like so: https://semgrep.dev/s/squaresurf:eslint_detect-object-injection?

emjin: +1, I tried 

```
value = obj["..."]
```

and then 

```
function x() {
    foo = "undeclared variable"
    //ERROR:
    value = obj[foo]
}
```

and didn't get results, but did with just 

```
foo = "undeclared variable"
//ERROR:
value = obj[foo]
```

Oddly, this doesn't show up when you use

```
function x() {
   foo = "undeclared variable"
   //ERROR:
   console.log(foo)
}
```

mschwager: > Oddly, this doesn't show up when you use ...

Hmm, I'm not sure I follow. I wouldn't expect that to match `value = obj["..."]` because there's no `obj`, or `value`?

IagoAbal: There is something weird indeed, I'll take a look. Thanks both!

emjin: oops, @mschwager to clarify, with the test case Iago has

```
   foo = "undeclared variable"
   //ERROR:
   console.log(foo)
```

matches as expected, and so does

```
function x() {
   foo = "undeclared variable"
   //ERROR:
   console.log(foo)
}
```

IagoAbal: There were several problems at play, first `foo` was incorrectly resolved as a local variable instead of a global, and then the dataflow constant propagation didn't propagate constants inside nested lvalues. Thanks for catching this!

IagoAbal:
Sorry for the noise this change in Naming_AST was trickier than I anticipated, there is still a false negative in the semgrep-rule `react-props-in-state` that I need to investigate. It's weird because `semgrep-core -config` does find the match, but `semgrep` doesn't.

IagoAbal:
It should work now, but the changes to Naming_AST are a bit tricky so I would really appreciate feedback from @aryx.

semgrep-core/src/analyzing/Naming_AST.ml
```diff
@@ -430,16 +433,23 @@ let resolve2 lang prog =
...
            * construct
            * todo? should add those somewhere instead of in_lvalue detection? *)
           VarDef ({ vinit; vtype }) when is_resolvable_name_ctx env lang ->
-            (* Need to visit expressions first so that type is populated *)
-            (* If we do var a = 3, then var b = a, we want to propagate the type of a *)
-            k x;
+            (* Need to visit expressions first so that type is populated, e.g.
+             * if we do var a = 3, then var b = a, we want to propagate the type of a. *)
+            (* Note that we are careful to visit each part of the declaration
+             * separately in order to avoid the v_vardef_as_assign_expr
```
aryx: yep, we need to remove the v_vardef_as_assign_expr in the visitor; it's too subtle. We need to move the logic in generic_vs_generic.ml and perform the equivalence there.

semgrep-core/src/analyzing/Naming_AST.ml
```diff
@@ -587,7 +597,13 @@ let resolve2 lang prog =
             add_ident_current_scope id resolved env.names;
             set_resolved env id_info resolved;
             k x
-        | OtherPat _ ->
+        | OtherPat _
+          (* This interacts badly with implicit JS/TS declarations. It causes
```
aryx, line 610: add a big TODO here, that says we should fix the AST of JS/TS so those f({ foo}) patterns do not show
as regular variables.
I think @mjambon wanted to rewrite the pattern part of ast_js.ml at some point.

aryx (approved):

semgrep-core/src/analyzing/Naming_AST.ml
IagoAbal: Done! FYI yesterday I opened a related issue: #3103.

semgrep-core/src/analyzing/Naming_AST.ml
IagoAbal: Let me look into that next week.

aryx:
Can we push this Iago? I think it was good.

IagoAbal:
I am trying to rebase/merge against the ocamlformat changes, but I'm having some problems.

aryx:
yes next time we need to do those big refactoring after everybody merged their stuff.

PR #2883, opened 2021-04-05
Flexible ruamel.yaml versioning

ajinabraham:
PR checklist:
- [ ] changelog is up to date

ruamel.yaml 0.17.0 is yanked.

https://pypi.org/project/ruamel.yaml/0.17.0/#history

PR #3088, opened 2021-05-12
Update comment on ruamel.yaml version

mjambon:
PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/setup.py
+++ b/semgrep/setup.py
@@ -101,7 +101,7 @@ setuptools.setup(
         "attrs>=19.3.0",
         "colorama>=0.4.3",
         "requests>=2.22.0",
-        # Pin exact version of 'ruamel.yaml' because of unstable API.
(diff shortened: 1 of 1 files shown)


mschwager:
I was actually hoping to remove this comment altogether after https://github.com/returntocorp/semgrep/pull/2883.

This part is obviously no longer relevant (unstable API dependency removed in https://github.com/returntocorp/semgrep/pull/2848):

> Pin exact version of 'ruamel.yaml' because of unstable API.

And this part is now guaranteed by a [unit test](https://github.com/returntocorp/semgrep/blob/develop/semgrep/tests/unit/test_yaml_parsing.py#L91), so we don't need to manually ensure anything:

> Ensure default type is still 'rt' (round-trip) and that RoundTrip inherits from Safe when upgrading the version.

mjambon:
@mschwager I'm dropping this PR (which is broken anyway), I'll let you edit the comment when you get a chance.

PR #3091, opened 2021-05-13
Add configNameHash and rulesHash to metrics

brendongo:
To be able to determine performance regressions it would be useful to have a hash of all the rules run so we can track how changes in versions with the same rules hash perform.

For now the aggregate rules_hash is the sha256 hash of the sha256 hashes of raw rules. So changes to order of rules will give us a different hash and changes to non-pattern fields will also give us a different hash. This limitation seems fine for now

PR checklist:
- [ ] changelog is up to date

Tested manually for now since getting a test for this will be more involved. The e2e tests should capture that this doesnt cause breakages

### DIFF
--- a/semgrep/semgrep/metric_manager.py
+++ b/semgrep/semgrep/metric_manager.py
@@ -28,6 +28,8 @@ class _MetricManager:
 
     def __init__(self) -> None:
         self._project_hash: Optional[str] = None
+        self._configs_hash = ""
+        self._rules_hash = ""
         self._return_code: Optional[int] = None
         self._version: Optional[str] = None
         self._num_rules: Optional[int] = None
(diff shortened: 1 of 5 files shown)


adamberman (approved):

semgrep/semgrep/metric_manager.py
```diff
@@ -112,7 +130,7 @@ def send(self) -> None:
 
             try:
                 r = requests.post(
-                    METRICS_ENDPOINT, json=metrics, timeout=10, headers=headers
+                    METRICS_ENDPOINT, json=metrics, timeout=2, headers=headers
```
mschwager, line 133: Could also use `version.VERSION_CHECK_TIMEOUT` here, although that will conflate two things.

semgrep/tests/unit/test_metric_manager.py
```diff
@@ -0,0 +1,91 @@
...
+    import requests
+
+    class block_network(socket.socket):
+        def __init__(self, *args, **kwargs):
+            raise Exception("Network call blocked")
+
+    socket.socket = block_network  # type: ignore
+
+    # test that network is blocked
+    with pytest.rasies(Exception):
```
mschwager, line 84: `rasies` -> `raises`

PR #3093, opened 2021-05-13
Add per rule timing to metrics

brendongo:
Add per rule parse, match, and run times to metrics. With these metrics it will be possible to determine where bottlenecks in performance are.

With this PR we always ask semgrep-core/spacegrep for timing information and the --time flag determines whether or not it is added to the output instead of whether or not it is even collected.

Also with this PR metrics object is only generated when metrics would even be sent anyway.

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -540,6 +540,5 @@ def cli() -> None:
                 timeout_threshold=args.timeout_threshold,
                 skip_unknown_extensions=args.skip_unknown_extensions,
                 severity=args.severity,
-                report_time=output_time,
                 optimizations=args.optimizations,
             )
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -363,16 +363,20 @@ class CoreRunner:
             for target in output_time_json["targets"]:
                 if "match_time" in target and "path" in target:
                     profiling_data.set_run_times(
-                        rule.id,
-                        target["path"],
+                        rule,
+                        Path(target["path"]),
                         Times(
-                            parse_time=target["parse_time"],
                             match_time=target["match_time"],
                             run_time=target["run_time"],
                         ),
                     )
+                    profiling_data.set_file_parse_time(
+                        Path(target["path"]), target["parse_time"]
+                    )
         if "rule_parse_time" in output_time_json:
-            profiling_data.set_parse_time(rule.id, output_time_json["rule_parse_time"])
+            profiling_data.set_rule_parse_time(
+                rule, output_time_json["rule_parse_time"]
+            )
 
     def _run_rule(
         self,
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -366,12 +366,10 @@ class CoreRunner:
                         rule,
                         Path(target["path"]),
                         Times(
+                            parse_time=target["parse_time"],
                             match_time=target["match_time"],
                             run_time=target["run_time"],
                         ),
-                    )
-                    profiling_data.set_file_parse_time(
-                        Path(target["path"]), target["parse_time"]
                     )
         if "rule_parse_time" in output_time_json:
             profiling_data.set_rule_parse_time(
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -118,7 +118,6 @@ class CoreRunner:
         timeout: int,
         max_memory: int,
         timeout_threshold: int,
-        report_time: bool,
     ):
         self._output_settings = output_settings
         self._allow_exec = allow_exec
@@ -126,7 +125,6 @@ class CoreRunner:
         self._timeout = timeout
         self._max_memory = max_memory
         self._timeout_threshold = timeout_threshold
-        self._report_time = report_time
 
     def _flatten_rule_patterns(self, rules: List[Rule]) -> Iterator[Pattern]:
         """
@@ -247,15 +245,13 @@ class CoreRunner:
                 str(self._timeout),
                 "-max_memory",
                 str(self._max_memory),
+                "-json_time",
             ]
 
             equivalences = rule.equivalences
             if equivalences:
                 self._write_equivalences_file(equiv_file, equivalences)
                 cmd += ["-equivalences", equiv_file.name]
-
-            if self._report_time:
-                cmd += ["-json_time"]
 
             if self._output_settings.debug:
                 cmd += ["-debug"]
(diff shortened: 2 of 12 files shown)


brendongo:
~~This is still buggy but doesn't need to get in for 0.51.0~~

Fixed

semgrep/semgrep/cli.py
```diff
@@ -540,6 +540,5 @@ def cli() -> None:
                 timeout_threshold=args.timeout_threshold,
                 skip_unknown_extensions=args.skip_unknown_extensions,
                 severity=args.severity,
-                report_time=output_time,
```
emjin, line 543: There were a lot of changes and I lost track a bit---how is `self.settings.output_time` in output.py being set? It controls whether or not the timing summary is outputted when we're using the text output.

semgrep/semgrep/metric_manager.py
```diff
@@ -38,6 +40,9 @@ def __init__(self) -> None:
         self._run_time: Optional[float] = None
         self._total_bytes_scanned: Optional[int] = None
         self._errors: List[str] = []
+        self._rule_hashes: List[str] = []
```
emjin: What's the difference between _rules_hash and rule_hashes? (Specifically, what's the former for?)

emjin (dismissed):

brendongo:
This metric is a little large. Will work on reducing by sending the relevant parse/match rates we want instead of all this raw data

aryx:
Can we merge this?

brendongo:
> Can we merge this?

Not yet. This makes metrics too large. Work still needs to be done to do summarization on cli side before reporting back

brendongo:
From @emjin:

What we care about is:
Overall bps number for semgrep
Per rule bps number (run_time - parse_time) and per rule bps match_time
Per file bps parse_time
Per rule bps parse_time
Outliers (@nathan suggested Top 10 slowest files?)

brendongo:
file_stats object example:
`{'size': 742, 'numTimesScanned': 60, 'parseTime': 0.0018460750579833984, 'matchTime': 0.03446817398071289, 'runTime': 0.05040717124938965}`

rule_stats object example:
`{'ruleHash': 'bb929471a63480629dc352b80e089ea36336c00ea46cba41de348c283ddcde79', 'parseTime': 0.0016939640045166016, 'matchTime': 0.06403851509094238, 'runTime': 0.14647603034973145, 'bytesScanned': 412192}`

semgrep/semgrep/cli.py
brendongo: It is set in cli.py depending on whether --time or --json-time is passed

semgrep/semgrep/core_runner.py
```diff
@@ -246,16 +244,14 @@ def _run_core_command(
...
                 str(self._max_memory),
+                "-json_time",
             ]
 
             equivalences = rule.equivalences
             if equivalences:
                 self._write_equivalences_file(equiv_file, equivalences)
                 cmd += ["-equivalences", equiv_file.name]
 
-            if self._report_time:
```
brendongo, line 257: Always asks semgrep-core to report timing information even if we don't print them out (user did not specify `--time`)

brendongo: This makes the _report_time option unnecessary in core_runner.py

semgrep/semgrep/output.py
```diff
@@ -62,8 +62,8 @@ def _build_time_target_json(
 
     target_json["path"] = path_str
     target_json["num_bytes"] = num_bytes
-    timings = [profiling_data.get_run_times(rule.id, path_str) for rule in rules]
-    target_json["parse_times"] = [timing.parse_time for timing in timings]
+    target_json["parse_time"] = profiling_data.get_file_parse_time(target)
```
brendongo: Changed `parse_times` in the json time output to just report a single time instead of a list of times. Does this affect our benchmarking stuff @emjin @colleend?

semgrep/semgrep/formatter/text.py
```diff
@@ -130,16 +130,14 @@ def _build_text_timing_output(
             rule["id"]: functools.reduce(
                 lambda x, y: (x[0] + y[0], x[1] + y[1]),
                 (
-                    (t["run_times"][i] - t["parse_times"][i], t["match_times"][i])
```
brendongo: @emjin not sure what the intention is here. Is it important that the output is run_time - parse_time?

semgrep/semgrep/formatter/text.py
```diff
@@ -130,16 +130,14 @@ def _build_text_timing_output(
...
                     for t in time_data["targets"]
                 ),
                 (time_data["rule_parse_info"][i], 0.0),
             )
             for i, rule in enumerate(time_data["rules"])
         }
-        file_parsing_time = sum(
-            sum(target["parse_times"]) for target in time_data["targets"]
-        )
+        file_parsing_time = sum(target["parse_time"] for target in time_data["targets"])
```
brendongo: I'm assuming file_parsing_time was the total time spent parsing all files?

semgrep/semgrep/formatter/text.py
emjin: The run_time includes the time it took the file to parse, which, because it varies based on whether the file is cached or not, we don't want to count

semgrep/semgrep/formatter/text.py
emjin: Yes (note: all files, for all rules---not just the first time)

semgrep/semgrep/output.py
emjin: This is fine for the benchmarking I did (I don't use -json-time), but @colleend might have changed that

colleend: no worries, benchmarking still doesn't use -json-time 👍

semgrep/semgrep/formatter/text.py
brendongo: Oh I see. I'll put some of this back into profiler then. Didn't consider this :D

semgrep/semgrep/profiling.py
```diff
@@ -11,19 +14,108 @@
...
+
+    def get_file_num_times_scanned(self, target: Path) -> int:
+        """
+        Returns number of times a file was scanned with rules.
+        Assumes that each entry to set_run_times means a target
+        was scanned once
+        """
+        return self._file_num_times_scanned.get(target, 0)
+
+    def set_run_times(self, rule: Rule, target: Path, times: Times) -> None:
```
emjin, line 86: Not sure if `set` is the right description for this, since it accumulates information

emjin (approved):

PR #3109, opened 2021-05-14
Gracefully handle C# constructs that are not yet supported

IagoAbal:
We just map them to `Other*` nodes in the Generic AST.

Also completed the translation of a bunch of constructs.

PR checklist:
- [x] changelog is up to date


semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
```diff
@@ -268,8 +268,15 @@ end
 (* Disable warning against unused 'rec' *)
 [@@@warning "-39"]
 
-let todo (env : env) _ =
-  failwith "not implemented"
+let todo_expr _env _ = AST.OtherExpr (AST.OE_Todo, [])
```
aryx: could be good to at least pass a Parse_info.t to it (You can use the Info any), so in the AST we know the original location of the expression/stmt/... not handled yet.

semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
```diff
@@ -922,7 +940,7 @@ and argument (env : env) ((v1, v2, v3) : CST.argument) : AST.argument =
   (* TODO return Ast.ArgKwd if Some v1 *)
   AST.Arg v3
 
-and initializer_expression (env : env) ((v1, v2, v3, v4) : CST.initializer_expression) =
+and initializer_expression (env : env) ((v1, v2, v3, v4) : CST.initializer_expression) : tok * expr list * tok =
```
aryx: you can use expr list bracket, for those times surround by () or {} or [].
See AST_generic.bracket type.

semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
```diff
@@ -1354,14 +1382,35 @@ and expression (env : env) (x : CST.expression) : AST.expr =
...
+       let v2 = token env v2 (* "with" *) in
+       let v3 = token env v3 (* "{" *) in
+       let v4 =
+         (match v4 with
+          | Some x -> with_initializer_expression env x
+          | None -> [])
+       in
+       let v5 = token env v5 (* "}" *) in
+       let with_fields = AST.Record((v3, v4, v5)) in
+       (* THINK:
```
aryx: yes maybe. They are also used in OCaml, so when the same construct is used in different languages I tend to
promote the construct from OE_xxx to a real construct.

semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
```diff
@@ -1587,8 +1636,8 @@ and statement (env : env) (x : CST.statement) =
         | `Case_exp (v1, v2) ->
             let v1 = token env v1 (* "case" *) in
             let v2 = expression env v2 in
-            todo env (v1, v2)
-        | `Defa tok -> todo env tok (* "default" *)
+            todo_stmt env (v1, v2)
```
aryx: There is a Default and Case in AST_generic.ml

aryx (approved):
cool!

semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
IagoAbal: Good point, done!

semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
IagoAbal: Good point, done!

IagoAbal: Maybe let's wait for someone to have the need for it, when C# support is a bit more mature.

semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
IagoAbal: Good point, done!

semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
IagoAbal: These are special labels, C# allows stuff like:
```
switch (n)
{
    case 1:
        cost += 25;
        break;
    case 2:
        cost += 25;
        goto case 1;
    case 3:
        goto default;
    default:
        Console.WriteLine("Invalid selection.");
}    
```
We could just generate `"__switch_case_1"` and `"__switch_default"` labels and map `goto case` and `goto default` to regular `goto`s.

toy-matcher/Matcher.ml
```diff
@@ -176,29 +201,34 @@ let sample_pattern : pattern =
...
-      | Backref name ->
-          if not (List.mem name env) then
-            failwith ("backreference to invalid name: " ^ name) );
+      (match atom with
+       | Any_symbol -> ()
+       | Symbol _ -> ()
+       | Ellipsis -> ()
+       | Backref name ->
+           if not (List.mem name env) then
+             failwith ("backreference to invalid name: " ^ name)
```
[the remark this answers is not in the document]
IagoAbal: This one and the rest seem like false positives (?).

aryx:
yes, I need to fix semgrep/ocaml

PR #3112, opened 2021-05-14
Add to timing output to test.py

colleend:
This PR makes it so that you can run the command:
` semgrep --test --test-ignore-todo --dangerously-allow-arbitrary-code-execution-from-rules --time --json $$PWD`
and receive timing information in the outputted json.

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -488,7 +488,7 @@ def cli() -> None:
     if args.test:
         # the test code (which isn't a "test" per se but is actually machinery to evaluate semgrep performance)
         # uses managed_output internally
-        semgrep.test.test_main(args)
+        semgrep.test.test_main(args, output_settings)
 
     # The 'optional_stdin_target' context manager must remain before
     # 'managed_output'. Output depends on file contents so we cannot have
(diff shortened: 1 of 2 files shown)


mschwager:
Do we really need this? I tend to see benchmarking and correctness as two independent forms of testing. My understanding is that the `--test` testing functionality is primarily for rule correctness testing. The test cases are so small that I would expect them to run almost instantly, so timing information isn't particularly useful. On the other hand, if we're trying to profile performance I would expect to pass `--time` against a particular config and code file. Is there a use-case I'm missing here?

mschwager:
Ah, I see https://github.com/returntocorp/enterprise/issues/632 now.

But still, is there a reason we need this functionality under `--test`? Looking at:

> The next thing we'll probably want to do is benchmark these on full repositories. This way we get data closer to the real world since --test will only run on unit test files.

I think this agrees with my previous comment. Aren't timing and `--test` mostly independent? If we're going to benchmark against full repositories I wouldn't expect to be using the `--test` functionality :thinking:

colleend:
hmm, that's a good point! I think the times using --test and --time simultaneously can generate a good basepoint measure for rules, but it's definitely far more accurate to time rules on real-world repositories. 

I was thinking when creating this PR that there might be a case where we keep benchmarking performance using --test and --time, but that doesn't seem like the case if we're able to test the rules on real-world repositories.

I'll close out this PR draft and we can perhaps add to the benchmarking code in order to generate some good estimates of how fast our rules are currently 👍

PR #3114, opened 2021-05-14
Preview ocamlformat's "sparse" style

mjambon:
This is a preview only. Don't merge.


aryx:
I guess we can close that then.

Issue #3105, opened 2021-05-13
auto-inserted 'else' branch in ocaml AST


mjambon:
**Describe the bug**

Either this shouldn't match or the [rules that detect useless `else`s](https://semgrep.dev/r/ocaml.lang.best-practice.ifs.ocamllint-useless-else) should be removed.

```
$ cat toto.ml
if true then huge
$ semgrep-core -lang ml -e 'if $COND then $BODY else ()' toto.ml
toto.ml:1
 if true then huge
```
^ should not match because there's no `else` in the source code.

Same code in the live editor (fails due to [something else at the moment](https://github.com/returntocorp/semgrep-app/issues/2518)):
https://semgrep.dev/s/J3nk

aryx:
But why do we have this bug now? What is the PR that broke this?

mjambon:
This started showing up with the ocamlformat reformat. First I thought it was due to semgrep only checking files that changed so I didn't look further.

aryx:
I think it's fixed now.

PR #3123, opened 2021-05-17
Fix semgrep for OCaml useless-else false positives

aryx:
The generic AST for 'if foo then bar else ()' was using Conditional
which didn't have an option, so whether or not you were using an else,
the generic AST would always contain an 'else ()' in the AST.
This PR fixes that by using If instead of Conditional in
ml_to_generic.ml

test plan:
test file included.
also
$ make install
< under pipenv shell>
$ cd ~/github/semgrep
$ make check

now works witout returning any findings

PR checklist:
- [x] changelog is up to date


aryx:
This closes https://github.com/returntocorp/semgrep/issues/3105

semgrep-core/old/lang_php/spatch_php.ml
```diff
@@ -143,7 +143,9 @@ let parse file =
   toks
   |> List.iter (fun tok ->
          let line = Parse_info.line_of_info tok in
-         let annot = Hashtbl.find hline_env line in
+         let annot =
+           try Hashtbl.find hline_env line with Not_found -> raise Not_found
```
IagoAbal, line 147: This line looks strange. I'm guessing it's a left-over from debugging?

IagoAbal (approved):

IagoAbal:
There are a couple of regressions perhaps related to the mapping of `try-with`:

> 2021-05-17T15:26:23.5452092Z E           (TODO: 0) semgrep-rules/ocaml/lang/best-practice/hashtbl.yaml
> 2021-05-17T15:26:23.5453131Z E           	✖ hashtbl-find-outside-try                                     TP: 1 TN: 0 FP: 1 FN: 0
> 2021-05-17T15:26:23.5454568Z E           	test: /tmp/pytest-of-runner/pytest-0/popen-gw1/test_semgrep_rules_repo0/semgrep-rules/ocaml/lang/best-practice/hashtbl.ml, expected lines: [3], reported lines: [3, 10]
> 2021-05-17T15:26:23.5455896Z E           (TODO: 0) semgrep-rules/ocaml/lang/best-practice/list.yaml
> 2021-05-17T15:26:23.5456809Z E           	✖ list-find-outside-try                                        TP: 1 TN: 0 FP: 1 FN: 0
> 2021-05-17T15:26:23.5458177Z E           	test: /tmp/pytest-of-runner/pytest-0/popen-gw1/test_semgrep_rules_repo0/semgrep-rules/ocaml/lang/best-practice/list.ml, expected lines: [3], reported lines: [3, 10]

semgrep-core/src/parsing/pfff/ml_to_generic.ml
```diff
@@ -247,50 +298,16 @@ and expr = function
...
-      in
-      let st = G.For (t, header, G.exprstmt v5) |> G.s in
-      G.OtherExpr (G.OE_StmtExpr, [ G.S st ])
   | ExprTodo (t, xs) ->
       let t = todo_category t in
       let xs = list expr xs in
       G.OtherExpr (G.OE_Todo, G.TodoK t :: List.map (fun x -> G.E x) xs)
+  | If _ | Try _ | For _ | While _ | Sequence _ ->
+      let s = stmt e in
+      G.OtherExpr (G.OE_StmtExpr, [ G.S s ])
```
IagoAbal, line 310: Should be `G.OE_StmtExpr2` used here instead of `G.OE_StmtExpr`?

semgrep-core/old/lang_php/spatch_php.ml
aryx: no it's to remove the semgrep warning about the use of Hashtbl.find :)

IagoAbal: Aha! Assuming that there is no `try-with` that catches it, we could use a `failwith` (if we expect it to happen) or log an error (if we don't).

Issue #2123, opened 2020-11-25
Semgrep changing current directory causes unexpected file resolving behaviour


Sjord:
Semgrep [changes to the current directory to SEMGREP_SRC_DIRECTORY in some cases](https://github.com/returntocorp/semgrep/blob/b28753dcc3d3aa24bdae1802321cd8e0661d219c/semgrep/semgrep/config_resolver.py#L241-L253). This can lead to unexpected behaviour when resolving relative files.

**To Reproduce**

```sh
% docker run -t -i --entrypoint sh --workdir /myproject -v "$PWD:/myproject" returntocorp/semgrep
/myproject $ ls -l .myconfig/semgrep.yml 
-rw-r--r--    1 semgrep  semgrep       1239 Nov 25 10:56 .myconfig/semgrep.yml
/myproject $ semgrep -f .myconfig/semgrep.yml 
unable to find a config; path `.myconfig/semgrep.yml` does not exist (since you are running in docker, you cannot specify arbitary paths on the host; they must be mounted into the container)
no valid configuration file found (1 configs were invalid)
/myproject $ 
```

As you can see, .myconfig/semgrep.yml exists, and semgrep claims it does not exist. But semgrep changed the working directory, and it is silently looking in /src.

**Expected behavior**

Perhaps semgrep should look in SEMGREP_SRC_DIRECTORY without changing the current directory. That way, at least the error message would say `path '/src/.myconfig/semgrep.yml' does not exist`. Or should it first look in the current directory and then in SEMGREP_SRC_DIRECTORY?

**What is the priority of the bug to you?**

P2

**Environment**

returntocorp/semgrep docker image. I am trying to run this in GitLab CI.

ievans:
Thanks for this bug report @Sjord. As a workaround, have you seen https://github.com/returntocorp/semgrep-action? Many other semgrep users are using that successfully to run in GitLab CI, and it has a lot of other features that may be appropriate. Someone just contributed support for Gitlab SAST output to it. Anyways may save you some time as you are trying to get it set up in GitLab CI.

aryx:
@spencerdrak is this still an issue?
Closing in doubt.

Sjord:
This is still a problem. Could you reopen this?

```
/myproject # semgrep --version
0.94.0
/myproject # ls -l semgrep.yml
-rw-r--r--    1 root     root          3014 Jan 27 09:03 semgrep.yml
/myproject # semgrep -f semgrep.yml
[...]

PR #3125, opened 2021-05-17
Avoid using "chdir"

### LINKED ISSUE #1215: Make semgrep's "chdir" usage more resilient
Semgrep's usage within Docker combined with our `chdir` call caused for a difficult to track down bug. The related fixes were:

- https://github.com/returntocorp/semgrep-rules/pull/696
- https://github.com/returntocorp/semgrep-app/pull/165

Along with https://github.com/returntocorp/semgrep/pull/1192 for the original change.

The issue was our Github Action calls `make test`: https://github.com/returntocorp/semgrep-rules-test-action-develop/blob/master/entrypoint.sh, which runs this `test` target: https://github.com/returntocorp/semgrep-rules/blob/develop/Makefile. This was failing on the 2nd validate call on the `c` directory.

The failure was due to the `chdir` call in `semgrep/config_resolver.py` when run in a Docker environment. Initially the `/src` directory doesn't exist because the volume hasn't been mounted, then on the first validate call it avoid the `chdir` because it doesn't exist. Next the version caching code creates the `/src` directory. Then the next validate call sees that that directory does exist and calls `chdir`. This moves the script from the correct location such that it can no longer find the rules.

We could confirm this behavior with the following:

```
/github/workspace # ls /src
ls: /src: No such file or directory
/github/workspace # semgrep --validate --config=./python .
Configuration is valid - found 119 valid configuration(s) and 0 configuration error(s).
/github/workspace # ls /src
/github/workspace # semgrep --validate --config=./c .
Configuration is invalid - found 0 valid configuration(s) and 1 configuration error(s).
unable to find a config; path `c` does not exist (since you are running in docker, you cannot specify arbitrary paths on the host; they must be mounted into the container)
Please fix the above errors and try again.
```

If possible, we should make this `chdir` call more resilient. I see a few options:

- Remove it all together, if possible (this may play poorly with rule ID rewriting)
- Un`chdir` somewhere later in the script. This feels potentially fragile, what if we error out or have an exception? What if we fail to `chdir` correctly? The plus side is this is fairly straightforward, but there may be gotchas.
- Others?

mschwager:
Fixes #1215.

This thing bit GitLab, and is so often a sharp edge for users. We're well past fixing it at this point :+1: 

I'm not 100% clear on everything `adjust_for_docker` was supposed to accomplish. My understanding is that it provided two things:

1. A sane default target inside Docker environments where we expect the code volume to be.
2. A mechanism for cleaning our filesystem-dependent rule IDs. In other words, the default volume is `/src` and we don't want all rule IDs to be prepended with `src`. Performing a `chdir` to the code directory fixes this.

These code changes account for both of the above issues without performing a `chdir`. The problems with `chdir` have been highlighted extensively in #1215. Let me know if I'm missing any more cases `adjust_for_docker` was covering.

These changes also have the benefit of explicitly stating what state to alter when we're running in Docker vs. implicitly performing a `chdir` and you're supposed to know why we're doing that :+1: 

I manually tested this by hardcoding `ADJUST_FOR_DOCKER` to `True` and `False` and noting that it did the correct thing.

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -9,15 +9,18 @@ import semgrep.semgrep_main
 import semgrep.test
 from semgrep import __VERSION__
 from semgrep.bytesize import parse_size
+from semgrep.constants import ADJUST_FOR_DOCKER
 from semgrep.constants import DEFAULT_CONFIG_FILE
 from semgrep.constants import DEFAULT_MAX_CHARS_PER_LINE
 from semgrep.constants import DEFAULT_MAX_LINES_PER_FINDING
 from semgrep.constants import DEFAULT_MAX_TARGET_SIZE
+from semgrep.constants import DEFAULT_TARGET
 from semgrep.constants import DEFAULT_TIMEOUT
 from semgrep.constants import MAX_CHARS_FLAG_NAME
 from semgrep.constants import MAX_LINES_FLAG_NAME
 from semgrep.constants import OutputFormat
 from semgrep.constants import RCE_RULE_FLAG
+from semgrep.constants import SEMGREP_SRC_DIRECTORY
 from semgrep.constants import SEMGREP_URL
 from semgrep.dump_ast import dump_parsed_ast
 from semgrep.error import SemgrepError
@@ -46,7 +49,7 @@ def cli() -> None:
     parser.add_argument(
         "target",
         nargs="*",
-        default=[os.curdir],
+        default=DEFAULT_TARGET,
         help=(
             "Search these files or directories. Defaults to entire current "
             "working directory. Implied argument if piping to semgrep."
@@ -446,13 +449,6 @@ def cli() -> None:
     # set the flags
     semgrep.util.set_flags(args.debug, args.quiet, args.force_color)
 
-    # change cwd if using docker
-    try:
-        semgrep.config_resolver.adjust_for_docker()
-    except SemgrepError as e:
-        logger.exception(str(e))
-        raise e
-
     output_format = OutputFormat.TEXT
     if args.json or args.json_time or args.debugging_json:
         output_format = OutputFormat.JSON
@@ -490,10 +486,15 @@ def cli() -> None:
         # uses managed_output internally
         semgrep.test.test_main(args)
 
+    if args.target == DEFAULT_TARGET and ADJUST_FOR_DOCKER:
+        target = [str(SEMGREP_SRC_DIRECTORY)]
+    else:
+        target = args.target
+
     # The 'optional_stdin_target' context manager must remain before
     # 'managed_output'. Output depends on file contents so we cannot have
     # already deleted the temporary stdin file.
-    with optional_stdin_target(args.target) as target, managed_output(
+    with optional_stdin_target(target) as target, managed_output(
         output_settings
     ) as output_handler:
         if args.dump_ast:
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -487,14 +487,14 @@ def cli() -> None:
         semgrep.test.test_main(args)
 
     if args.target == DEFAULT_TARGET and ADJUST_FOR_DOCKER:
-        target = [str(SEMGREP_SRC_DIRECTORY)]
+        target_input = [str(SEMGREP_SRC_DIRECTORY)]
     else:
-        target = args.target
+        target_input = args.target
 
     # The 'optional_stdin_target' context manager must remain before
     # 'managed_output'. Output depends on file contents so we cannot have
     # already deleted the temporary stdin file.
-    with optional_stdin_target(target) as target, managed_output(
+    with optional_stdin_target(target_input) as target, managed_output(
         output_settings
     ) as output_handler:
         if args.dump_ast:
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/config_resolver.py
```diff
@@ -35,16 +38,8 @@
 
 logger = logging.getLogger(__name__)
 
-IN_DOCKER = "SEMGREP_IN_DOCKER" in os.environ
-IN_GH_ACTION = "GITHUB_WORKSPACE" in os.environ
-
-SRC_DIRECTORY = Path(os.environ.get("SEMGREP_SRC_DIRECTORY", Path("/") / "src"))
-OLD_SRC_DIRECTORY = Path("/") / "home" / "repo"
```
mschwager, line 42: This has been deprecated for 10.5 months at this point (https://github.com/returntocorp/semgrep/pull/1192), I think we're safe to remove it.

semgrep/semgrep/config_resolver.py
```diff
@@ -35,16 +38,8 @@
...
-
 RULES_REGISTRY = {"r2c": "https://semgrep.dev/c/p/r2c"}
 
-MISSING_RULE_ID = "no-rule-id"
```
mschwager, line 46: I think this was originally here because we didn't have an explicit schema. Since `semgrep.rule.Rule` specifies `id` as `str` instead of `Optional[str]`, and an `id` is required by the JSON schema I think we're safe to remove this now.

semgrep/semgrep/config_resolver.py
```diff
@@ -144,25 +139,29 @@ def _safe_relative_to(a: Path, b: Path) -> Path:
...
+            relative_path = SEMGREP_SRC_DIRECTORY
+        else:
+            relative_path = Path.cwd()
+
         at_path = Path(config_id)
-        at_path = Config._safe_relative_to(at_path, Path.cwd())
+        at_path = Config._safe_relative_to(at_path, relative_path)
 
-        prefix = ".".join(at_path.parts[:-1]).lstrip("./").lstrip(".")
+        prefix = ".".join(at_path.parts[:-1]).lstrip("./")
```
mschwager, line 154: This 2nd `lstrip` isn't necessary. The first will strip all the characters included in its argument:

```
In [1]: "./.foo".lstrip("./").lstrip(".")                                                               
Out[1]: 'foo'

In [2]: "./.foo".lstrip("./")                                                                           
Out[2]: 'foo'
```

semgrep/semgrep/config_resolver.py
```diff
@@ -144,25 +139,29 @@ def _safe_relative_to(a: Path, b: Path) -> Path:
...
+
         return prefix
 
     @staticmethod
     def _rename_rule_ids(valid_configs: Dict[str, List[Rule]]) -> Dict[str, List[Rule]]:
-        transformed = {}
-        for config_id, rules in valid_configs.items():
-            transformed[config_id] = [
-                rule.with_id(
-                    f"{Config._convert_config_id_to_prefix(config_id)}{rule.id or MISSING_RULE_ID}"
```
mschwager, line 161: This was the only usage of `MISSING_RULE_ID`.

semgrep/semgrep/config_resolver.py
```diff
@@ -251,21 +250,6 @@ def resolve_targets(targets: List[str]) -> List[Path]:
...
-        if OLD_SRC_DIRECTORY.exists():
-            raise SemgrepError(
-                f"Detected Docker environment using old code volume, please use '{SRC_DIRECTORY}' instead of '{OLD_SRC_DIRECTORY}'"
-            )
-        if not SRC_DIRECTORY.exists():
-            raise SemgrepError(
-                f"Detected Docker environment without a code volume, please include '-v \"${{PWD}}:{SRC_DIRECTORY}\"'"
-            )
-    if SRC_DIRECTORY.exists():
-        os.chdir(SRC_DIRECTORY)
```
mschwager, line 266: :fire:

mschwager:
I think this will fix https://github.com/returntocorp/semgrep/issues/2123 too.

semgrep/semgrep/cli.py
```diff
@@ -490,10 +486,15 @@ def cli() -> None:
...
+    if args.target == DEFAULT_TARGET and ADJUST_FOR_DOCKER:
+        target = [str(SEMGREP_SRC_DIRECTORY)]
+    else:
+        target = args.target
+
     # The 'optional_stdin_target' context manager must remain before
     # 'managed_output'. Output depends on file contents so we cannot have
     # already deleted the temporary stdin file.
-    with optional_stdin_target(args.target) as target, managed_output(
+    with optional_stdin_target(target) as target, managed_output(
```
brendongo: Do you think it's worth changing one of the `target` s to prevent confusion?

semgrep/semgrep/config_resolver.py
```diff
@@ -144,25 +139,29 @@ def _safe_relative_to(a: Path, b: Path) -> Path:
 
     @staticmethod
     def _convert_config_id_to_prefix(config_id: str) -> str:
+        if ADJUST_FOR_DOCKER:
```
brendongo, line 142: Can you add a comment that this is so that the rule-ids are not prepended with src (and also  that this assumes the rules will be somewhere in SEMGREP_SRC_DIRECTORY)

brendongo (approved):

brendongo (approved):

PR #3130, opened 2021-05-18
matching: Prefer Id over E when binding typed metavariables to names

IagoAbal:
When a metavar binds to a name this must be represented using the Id
constructor.

This fixes a -config regression wrt Python's Semgrep wrapper affecting:
semgrep-rules/java/lang/security/audit/xss/no-direct-response-writer.java

The problem can be reproduced with a simple pattern-and:

    patterns:
    - pattern: |
        (PrintWriter $WRITER).println(...)
    - pattern-not: $WRITER.println("...")

on this Java code:

    PrintWriter writer = response.getWriter();
    writer.println("foo");

The typed metavar $WRITER used in the positive pattern is bound to
`MV.E (N (Id "writer"))` whereas the non-typed metavar $WRITER used in
the negative pattern is bound to `MV.Id "writer"`. Then, when we check
if there is a negative match that filters out our positive match, we
can't find one because `MV.E` and `MV.Id` are different kinds of
values.

test plan:
make test # test included

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -1000,9 +1000,12 @@ and m_compatible_type typed_mvar t e =
       m_type_ t (A.TyPointer (t1, TyBuiltin ("char", tok))) >>= fun () ->
       envf typed_mvar (MV.E e)
   (* for matching ids *)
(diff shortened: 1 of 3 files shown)


aryx (approved):
Good catch!

PR #3131, opened 2021-05-18
Rename files following Emma's critics

aryx:
At the last code review, Emma pointed out a few files had bad names.
Maybe Match_rules.ml is better than just Semgrep.ml, and
Match_patterns.ml
better than Semgrep_generic.ml

test plan:
make test

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/matching/Matching_generic.mli
+++ b/semgrep-core/src/matching/Matching_generic.mli
@@ -25,10 +25,10 @@ and tout = tin list
  * information tin, and it will return something (tout) that will
  * represent a match between element A and B.
  *)
-(* currently 'a and 'b are usually the same type as we use the
+(* currently A and B are usually the same type as we use the
  * same language for the host language and pattern language
  *)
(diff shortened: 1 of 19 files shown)


semgrep-core/src/matching/Matching_generic.mli
```diff
@@ -181,11 +180,7 @@ val m_string_prefix : string -> string -> tin -> tout
 
 (*e: signature [[Matching_generic.m_string_prefix]] *)
 val m_string_ellipsis_or_regexp_or_default :
-  ?m_string_for_default:(string, string) matcher ->
-  string ->
-  string ->
-  tin ->
-  tout
+  ?m_string_for_default:string matcher -> string -> string -> tin -> tout
```
IagoAbal: Could this be `string matcher -> string matcher` ?

IagoAbal (approved):
Nice!

semgrep-core/src/matching/Matching_generic.mli
aryx: yep

PR #3134, opened 2021-05-18
Revert "Avoid using "chdir" (#3125)"

mschwager:
This reverts commit 43d3b4a5721235dcc5538edd9f36ff3a8351c58c.


brendongo:
Reverting since this change makes semgrep in docker report absolute paths instead of relative paths when reporting findings

brendongo (approved):

Issue #2453, opened 2021-01-25
PHP - Inconsistent behaviour of wildcard matches in nested statement


erwanlr:
**Describe the bug**
Wildcard matches (...) fail to properly work in a nested statement (such as foreach and if) in PHP.

For example, the pattern:
```
$VARS = $_GET[$KEY];
...
foreach ($VARS as $VAR) {
  ...
  do_something($VAR);
  ...
}
```

Will not match the code
```php
$stacks = $_GET['p'];
foreach ($stacks as $value) {
  do_something($value);
}
```

But will match
```php
$stacks = $_GET['p'];
foreach ($stacks as $value) {
  $a = 'a';
  do_something($value);
}
```

**To Reproduce**
https://semgrep.dev/s/Kl8d

**Expected behavior**
Line 3 should match

**What is the priority of the bug to you?**
P0

erwanlr:
Not sure if also related to this issue:

https://semgrep.dev/s/rdk0

```
$VAR = $_GET[$KEY];
...
do_something($VAR);
```

Will not match
```php
if (isset($_GET['q'])) {
  $page2 = $_GET['q'];
}
do_something($page2);
```

To match it, the pattern would need to be like below, but then produces too many matches
```
...
$VAR = $_GET[$KEY];
...
do_something($VAR);
```

erwanlr:
Another one: https://semgrep.dev/s/N487 which is even weirder, as no match will be found at all with a pattern like `pattern: do_something(...);`

Let me know if I should open different issue reports for those

emjin:
For the last one (https://semgrep.dev/s/N487), the discrepancy is because one of the else bodies is empty. Not sure if we'd want to update the behavior to be more intuitive, but https://semgrep.dev/s/ydXO should get you want you want in the meantime.

erwanlr:
Thanks @emjin, however for the last one, the pattern in the linked semgrev dev was not the one I talked about for some reason (I guess I was testing things).

https://semgrep.dev/s/nd9Z

What I meant with this one is that `pattern: do_something(...);` will not match:

```php
if (isset($_GET['q'])) {
  // ruleid:issue
  do_something($_GET['q']);
} else {
  $a = 'a';
}
```

But will match

```php
if (isset($_GET['q'])) {
  $a = 2;
  // ruleid:issue
  do_something($_GET['q']);
} else {
  $a = 'a';
}
```

[...]

PR #3144, opened 2021-05-19
[PHP] removing more stmt list in ast_php.ml to avoid to use stmt1

aryx:
This will help https://github.com/returntocorp/semgrep/issues/2453

test plan:
see test files included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 - PHP: parsing correctly ... and metavariables in parameters
+- PHP: parsing correctly functions with a single statement in their body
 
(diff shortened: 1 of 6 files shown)


IagoAbal (approved):
There are some changes in Pfff like:

```
- let implicit_assigns = [] in (* TODO? *)
...
- A.f_body = implicit_assigns @ method_body env m.f_body;
+ A.f_body = (* implicit_assigns @ *) method_body env m.f_body;
```
Here I think it would be better to either remove the `(* implicit_assigns @ *)`  comment completely or replace it with a more elaborated TODO comment.

aryx:
yeah, but I don't remember what those things were for ... I think it was when there was an abstract interpreter for PHP in pfff (for Facebook), but I removed it to simplify things for semgrep development.

PR #3145, opened 2021-05-19
Evaluate interpolated strings during constant propagation

### LINKED ISSUE #3127: ruby: string constant propagation doesn't seem to work with double quotes and embedded newlines
**Describe the bug**
Given the pattern:
```
  patterns:
    - pattern-either: 
      - pattern: |
          foo(...)
    - pattern-not: |
        foo("...")
```
 and the code:
```
a = foo(data)
a = foo("foo: bar")

const_str1 = "foo: bar"
a = foo(const_str1)

const_str2 = 'foo: bar'
a = foo(const_str2)

const_str3 = "foo: bar\n"
a = foo(const_str3)

const_str4 = 'foo: bar\n'
a = foo(const_str4)
```

The call to `foo(data)`  **and** `foo(const_str3)` both match, but none of the other calls.

**To Reproduce**
https://semgrep.dev/s/aegarbutt-stripe:ruby-string-with-newline

**Expected behavior**
I expect only `foo(data)` to match.

IagoAbal:
Closes #3127

PR checklist:
- [x] changelog is up to date


semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -252,6 +256,20 @@ and eval_op env op args =
       | __else__ -> None )
   | __else__ -> None
 
+and eval_concat_string env args : literal option =
+  let go_concat : arguments -> string option =
+    List.fold_left
+      (fun res arg ->
+        match arg with
+        | Arg e ->
+            res >>= fun r ->
```
aryx: can use let* r = res in
...

aryx (approved):

PR #3148, opened 2021-05-19
Move normal outputting to text formatter

mschwager:
Move final output format option to its own formatter.

### DIFF
--- a/semgrep/semgrep/formatter/text.py
+++ b/semgrep/semgrep/formatter/text.py
@@ -1,0 +1,258 @@
+import functools
+import itertools
+from pathlib import Path
+from typing import Any
+from typing import Dict
+from typing import Iterator
+from typing import List
+from typing import Optional
+
+import colorama
+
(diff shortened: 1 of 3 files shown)


brendongo (approved):
LGTM. Non blocking comment: I was initially concerned we were adding formatting options to the rule.extra object just cause we also called it extra. Maybe formatting_opts or something might be a better name but w/e

mschwager:
> LGTM. Non blocking comment: I was initially concerned we were adding formatting options to the rule.extra object just cause we also called it extra. Maybe formatting_opts or something might be a better name but w/e

Yeah, `extra` here is not ideal. It basically exists because it wasn't clear to me initially what formatters needed what data. E.g. time data, stats data, debug data. Different formatters optionally make use of different data. I think it's more clear now - we could improve the `BaseFormatter` interface to ingest these different types of data in a consistent way. Since that's unrelated to this particular change, I'll save it for an additional refactor for another day.

PR #3154, opened 2021-05-19
Simplify verbosity flags

### LINKED ISSUE #2016: verbose or -v flag has confusing UX
_[edited by @ievans after discussion with @tkisason on Slack]_

**Describe the bug**

When you run semgrep with --verbose or -v option, the output may not be changed.

**To Reproduce**

```
~/Downloads> git clone https://github.com/tkisason/vulnapi.git
Cloning into 'vulnapi'...
remote: Enumerating objects: 85, done.
remote: Counting objects: 100% (85/85), done.
remote: Compressing objects: 100% (59/59), done.
remote: Total 85 (delta 39), reused 59 (delta 22), pack-reused 0
Unpacking objects: 100% (85/85), done.
~/Downloads> semgrep -v --config "https://semgrep.dev/c/r/python" vulnapi
using config from https://semgrep.dev/r/python. Visit https://semgrep.dev/registry to see all public rules.
downloading config...
running 259 rules...
100%|█████████████████████████████████████████|259/259
vulnapi/main.py
severity:warning rule:python.django.security.audit.django-ratelimit.missing-ratelimit.missing-ratelimit: Function 'set_notifications_addr' is missing a rate-limiting decorator.
High volume traffic to this function could starve application
resources. Consider adding rate limiting from a library such
as 'django-ratelimit'.

153:@app.post("/me/notifications/")
154:async def set_notifications_addr(
155:    notification: Notification,
156:    request: Request,
157:    current_user: User = Depends(get_current_active_user),
158:):
159:    email = await request.json()
160:    for i, user in enumerate(users_db):
161:        if user["username"] == current_user.username:
162:            users_db[i] = {**user, **email}
163:            print(users_db)
164:    return email
severity:error rule:python.jwt.security.jwt-hardcode.jwt-python-hardcoded-secret: Hardcoded JWT secret or private key is used.
This is a Insufficiently Protected Credentials weakness: https://cwe.mitre.org/data/definitions/522.html
Consider using an appropriate security mechanism to protect the credentials (e.g. keeping secrets in environment variables)

14:SECRET_KEY = "123456"
15:ALGORITHM = "HS256"
16:ACCESS_TOKEN_EXPIRE_MINUTES = 600
17:today = time.strftime("%Y-%m-%d")
18:
19:accounts_db = json.loads(open("data/accounts.json").read())
20:users_db = json.loads(open("data/users.json").read())
21:sqldb = sqlite3.connect("data/data.sqlite")
22:

```

Same with --verbose or -v, before and after config. 

**Expected behavior**

We should see an indication that the verbose mode is enabled and instructions on how to get even more verbosity. Perhaps (`Running in verbose mode, parse errors will be displayed in output. Run with `--debug` for even more detail.`)

P2 probably

**Environment**

Latest semgrep from brew on macos. Tried it with docker, same situation. 

```
~/Downloads> semgrep --version
0.30.0
~/Downloads>
~/Downloads> brew info semgrep
semgrep: stable 0.30.0 (bottled), HEAD
```

brendongo:
This PR standardizes our logging so that there is an easy mapping
between desired verbosity (--quiet, normal, --verbose, --debug) and
logging levels.

Moving forward, --quiet, normal, --verbose, and --debug will correspond
to CRITICAL, INFO, VERBOSE, and DEBUG logging levels respectively.
A given verbosity level will see logs that are at its level and lower.

This PR also makes --quiet, --verbose, and --debug mutually exclusive
command line flags.

Closes: https://github.com/returntocorp/semgrep/issues/2016

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -1,4 +1,3 @@
-import logging
 import os
 import time
 from collections import OrderedDict
@@ -32,8 +31,9 @@ from semgrep.rule_lang import YamlMap
 from semgrep.rule_lang import YamlTree
 from semgrep.util import is_config_suffix
 from semgrep.util import is_url
+from semgrep.verbose_logging import getLogger
 
-logger = logging.getLogger(__name__)
+logger = getLogger(__name__)
 
 IN_DOCKER = "SEMGREP_IN_DOCKER" in os.environ
 IN_GH_ACTION = "GITHUB_WORKSPACE" in os.environ
@@ -98,7 +98,7 @@ class Config:
                 # Patch config_id to fix https://github.com/returntocorp/semgrep/issues/1912
                 resolved_config = resolve_config(config)
                 if not resolved_config:
-                    logger.debug(f"Could not resolve config for {config}. Skipping.")
+                    logger.verbose(f"Could not resolve config for {config}. Skipping.")
                     continue
 
                 for (
@@ -463,7 +463,7 @@ def resolve_config(config_str: str) -> Dict[str, YamlTree]:
     else:
         config = load_config_from_local_path(config_str)
     if config:
-        logger.debug(f"loaded {len(config)} configs in {time.time() - start_t}")
+        logger.verbose(f"loaded {len(config)} configs in {time.time() - start_t}")
     return config
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Fixed
 
 ### Changed
+- Moved some debug logging to verbose logging
 
 ## [0.53.0](https://github.com/returntocorp/semgrep/releases/tag/v0.53.0) - 2021-05-26
 
--- a/semgrep/semgrep/__main__.py
+++ b/semgrep/semgrep/__main__.py
@@ -1,9 +1,4 @@
 #!/usr/bin/env python3
-from semgrep.verbose_logging import install_verbose_logging
-
-install_verbose_logging()
-
-import logging.handlers
 import sys
 
 from semgrep import __VERSION__
@@ -11,13 +6,14 @@ from semgrep.cli import cli
 from semgrep.error import OK_EXIT_CODE
 from semgrep.error import SemgrepError
 from semgrep.metric_manager import metric_manager
+from semgrep.verbose_logging import getLogger
 
 
 def main() -> int:
     # When running semgrep as a command line tool
     # silence root level logger otherwise logs higher
     # than warning are handled twice
-    logger = logging.getLogger("semgrep")
+    logger = getLogger("semgrep")
     logger.propagate = False
     metric_manager.set_version(__VERSION__)
     try:
--- a/semgrep/semgrep/__main__.py
+++ b/semgrep/semgrep/__main__.py
@@ -1,4 +1,8 @@
 #!/usr/bin/env python3
+from semgrep.verbose_logging import install_verbose_logging
+
+install_verbose_logging()
+
 import logging.handlers
 import sys
 
--- a/semgrep/semgrep/autofix.py
+++ b/semgrep/semgrep/autofix.py
@@ -1,4 +1,3 @@
-import logging
 import re
 from pathlib import Path
 from typing import Dict
@@ -9,8 +8,9 @@ from typing import Tuple
 from semgrep.error import SemgrepError
 from semgrep.rule import Rule
 from semgrep.rule_match import RuleMatch
+from semgrep.verbose_logging import getLogger
 
-logger = logging.getLogger(__name__)
+logger = getLogger(__name__)
 
 SPLIT_CHAR = "\n"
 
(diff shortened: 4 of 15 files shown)


mschwager:
Instead of doubling down and further complicating things to solve the complications, how do you feel about removing/simplifying this functionality? As far as I can tell, `--verbose` doesn't do much of anything:

```python
$ rg verbose semgrep/semgrep
semgrep/semgrep/output.py
360:    verbose_errors: bool = False  # to do: rename to just 'verbose'
467:                error.level != Level.WARN or self.settings.verbose_errors
515:                    f"{error_stats}; run with --verbose for details or run with --strict to exit non-zero if any file cannot be analyzed"

semgrep/semgrep/cli.py
362:        "--verbose",
474:        verbose_errors=args.verbose,

semgrep/semgrep/metric_manager.py
23:    Made explicit decision to be verbose in setting metrics instead

semgrep/semgrep/semgrep_main.py
42:    verbose: bool = False,
59:    if verbose:
```

The `semgrep_main` verbose functionality can easily be turned into a `logging.debug`. As for `output`, I think we can base `verbose_errors` on the `--debug` flag. Then we could even make `--verbose` an alias of `--debug`. Thoughts?

brendongo:
Debug also passes down the debug flag to semgrep-core and results in a lot of output being printed out that is really only useful if you are actually digging into a bug. If you're just trying to understand what semgrep is doing it's too much.

mschwager:
> Debug also passes down the debug flag to semgrep-core ...

Unless I'm missing something, I don't think this is correct. I see the `semgrep-core` debug flag based on:

```python
if self._output_settings.debug:
    cmd += ["-debug"]
```

And `output_settings.debug` is based on `debug=args.debugging_json`. I think this further highlights that this is too complicated and confusing. If we can't keep any of this straight, how can we expect our users to know what to do?

> ... if you are actually digging into a bug. If you're just trying to understand what semgrep is doing it's too much.

I don't fully understand the difference here. Both of these sound like "you're trying to understand what semgrep is doing" to me :confused:. I think if we come up with a clear, concise split of which flag is supposed to do what that would help both us and our users. Something like: `--verbose` turns on debugging output for Semgrep Python, `--debug` turns on additional debugging output from `semgrep-core`. It doesn't have to be that split, but I think something that simple and clear would help all parties involved :)

The changes in this PR lead me to wonder, when do I use `logging.verbose` vs. `logging.debug` vs. `logging.info`? And I suspect that will further exacerbate the issues we're having here. Basically, I still think adding _more_ complexity here is going in the wrong direction.

Given the recent `--debugging-json` difficulties, in addition to this confusion, would it make more sense to take a step back, think about how we would like this to look from first principles, and implement that instead of bolting on more things?

brendongo:
The way I see it, you want `--verbose` to give users more detail on things we sweep under an aggregation in normal runs.
"Running 5 rules..." ok in verbose mode what are those 5 rules? What files are scanned? What errors are you skipping?

`--debug` is when you think something went wrong and you want semgrep to step through everything it does. This includes semgrep-core logs, yes but also includes lots more breadcrumbing from the python code as well (think logging every network call, file read, etc.)

Maybe a high level summary is verbose tells you what semgrep is doing given the arguments you passed while debug in addition to that tells you the outcome of everything semgrep did that you as a user don't necessarily have control over? Is this a good enough split for now?

The intention of this PR is to make it more obvious when each log is made visible and even make it possible to have a log level between INFO and DEBUG. By pushing the decision to output a log to the logger, we don't need to get into a situation of having to check a log level before logging a message.

mschwager:
That `debug`|`verbose` split makes sense to me :+1:. I wonder if there's some way we could encourage/enforce those distinctions when people are creating new log messages? My concern is that this is mostly going to be tribal knowledge, and there will be a lot of nitpicking on log message level and/or it won't be clear what level to use when.

Overall, I'm not too concerned. I'm happy to accept this as-is, although I'm not sure that this is necessarily simplifying things. Removing our global flags here could also help clarify this functionality (similar to passing verbose around), but that may require some changes to our `tqdm` usage :man_shrugging:

mschwager:
At this point, if you think this is a step in the right direction, we should just roll with it. We can always make further improvements going forward :+1:

mschwager (approved):

brendongo:
Going to follow up with PR removing global verbosity flags

semgrep/semgrep/config_resolver.py
```diff
@@ -463,7 +463,7 @@ def resolve_config(config_str: str) -> Dict[str, YamlTree]:
     else:
         config = load_config_from_local_path(config_str)
     if config:
-        logger.debug(f"loaded {len(config)} configs in {time.time() - start_t}")
+        logger.verbose(f"loaded {len(config)} configs in {time.time() - start_t}")
```
aryx, line 466: if there's time information, that means we can't use --verbose for our pytests and snapshots.

QuinceyJames:
Hi! I just noticed that this modification to python's logger module has caused a downstream bug in my code.

Python recommends against defining custom levels: https://docs.python.org/3/howto/logging.html#custom-levels
> Defining your own levels is possible, but should not be necessary, as the existing levels have been chosen on the basis of practical experience. However, if you are convinced that you need custom levels, great care should be exercised when doing this, and it is possibly a very bad idea to define custom levels if you are developing a library. That’s because if multiple library authors all define their own custom levels, there is a chance that the logging output from such multiple libraries used together will be difficult for the using developer to control and/or interpret, because a given numeric value might mean different things for different libraries.

I found a workaround to get my code to work but I wanted to give a heads up!

Issue #2664, opened 2021-03-01
Statically detect unused metavariable-regex (or metavariable-regexp with wrong scope)


JoelPagliuca:
**Describe the bug**
in simple cases this isn't obviously obstructive, but in cases with lots of `pattern-either`s we found this forced us to split our rules across multiple files with lots of duplicated logic

**To Reproduce**
https://semgrep.dev/s/4xK5 produces no results
https://semgrep.dev/s/PJKz produces 1 result

**Expected behavior**
1 result for both examples

**What is the priority of the bug to you?**
P1 - quite annoying

**Environment**
`brew` distributed `v0.41.1`

ievans:
thanks for the clear reproductions, I agree this looks like a bug! Probably need an optimization to not filter if metavariable didn't occur in patterns, @aryx a case for a new feature easy to enable in the core now?

mschwager:
The `metavariable-regex` functionality will notify of this if you run with `--debug` locally:

```
$ echo 'set("hello maintainers")' | python -m semgrep --debug --config /tmp/metavariable.yaml -
loaded 1 configs in 0.04578685760498047
running 1 rules from 1 config /tmp/metavariable.yaml_0 
running 1 rules...
Running rule tmp.my_pattern_id...


--> rule (tmp.my_pattern_id) has findings on filepath: /tmp/tmpeacuze2n
after filter 'and': {<Range start=0 end=24 vars={'$SET': '4e1fedb0242220863f38d97b0e28db56'}>}
after filter 'metavariable_regex': {<Range start=0 end=24 vars={'$SET': '4e1fedb0242220863f38d97b0e28db56'}>}
metavariable '$UNUSED' missing in range '<Range start=0 end=24 vars={'$SET': '4e1fedb0242220863f38d97b0e28db56'}>'
after filter 'metavariable_regex': set()
after filter 'and_all': set()
compiled result set()
--------------------------------------------------------------------------------
...ran on 1 files
semgrep ran in 0:00:00.256416 on 1 files
findings summary: 0 warning
ran 1 rules on 1 files: 0 findings
```

We could always promote this to a warning so we wouldn't need `--debug`:

```
$ echo 'set("hello maintainers")' | python -m semgrep --config /tmp/metavariable.yaml -
running 1 rules...
metavariable '$UNUSED' missing in range '<Range start=0 end=24 vars={'$SET': '4e1fedb0242220863f38d97b0e28db56'}>'
[...]

PR #3157, opened 2021-05-20
engine: match_rules: Merge metavar bindings when intersecting matches

IagoAbal:
This fixes a -config regression wrt Python's Semgrep wrapper affecting:
javascript/sequelize/security/audit/sequelize-weak-tls-version.yaml

Minimal example:

    patterns:
    - pattern-inside: |
        {
          dialect: $DIALECT,
          ...
         }
    - pattern: |
        {
          minVersion: 'TLSv1'
        }
    - metavariable-regex:
        metavariable: $DIALECT
        regex: "['\"](mariadb|mysql|postgres)['\"]"

When intersecting the two patterns we must propagate the binding for
$DIALECT, so we can evaluate the `metavariable-regex`. If the binding is
lost then `metavariable-regex` evaluates to `false` and the match gets
dropped.

test plan:
make test # test included

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -282,24 +282,27 @@ let inside_compatible x y =
   | Some R.Inside, None -> false
 
 let intersect_ranges config xs ys =
+  let left_merge r1 r2 =
+    (* [r1] extended with [r2.mvars], assumes [included_in config r1 r2] *)
+    let r2_only_mvars =
+      r2.mvars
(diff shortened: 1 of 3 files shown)


aryx (approved):
excellent

aryx:
Note that an alternative would be to force the metavariable-regexp to have the right "scope", meaning it should be And'ed with pattern-inside, and then the 2 together would be And'ed with the other pattern.
That way we may not need to propagate those metavars.

aryx:
I'll add a comment in the code

aryx:
related issue: https://github.com/returntocorp/semgrep/issues/2664

PR #3121, opened 2021-05-17
C# alpha support

IagoAbal:
Closes #2345

test plan:
make test

PR checklist:
- [x] changelog is up to date

PR #3162, opened 2021-05-20
[csharp] Fix syntax error in tests/csharp/metavar_call.cs

### LINKED ISSUE #3161: C# tests are broken
Looks like we're getting parse errors on some of our C# tests: https://github.com/returntocorp/semgrep/runs/2631102847?check_suite_focus=true

Also confirmed here:

```
$ cat semgrep-core/tests/csharp/metavar_call.cs | docker run --rm -i returntocorp/semgrep:0.52.0 -e '$X == $X' -l csharp -
```

IagoAbal:
Fixes: 4ca2980cf55 ("C# alpha support (#3121)")

Closes #3161

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/tests/csharp/metavar_call.cs
+++ b/semgrep-core/tests/csharp/metavar_call.cs
@@ -1,7 +1,9 @@
-public static void Test()
+public class Test
 {
-    // ERROR:
-    Foo(1, 2);
-
-    return 1;
+    public static int Test()
+    {
+        // ERROR:
+        Foo(1, 2);
+        return 1;
+    }
 }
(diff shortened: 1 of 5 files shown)


semgrep-core/tests/csharp/metavar_call.cs
```diff
@@ -1,4 +1,4 @@
-public static void Test
+public static void Test()
```
mschwager: Does `public` work outside of a class context?

IagoAbal: My C# knowledge is very rusty, but the tree-sitter grammar accepts this but perhaps it's over-approximating. Do you know C# better? Do I just drop `public`? Otherwise  I can change all the tests to make use of classes.

IagoAbal: OK I just put everything inside a class so there is no doubt that it is valid C#.

mschwager (approved):

PR #3164, opened 2021-05-20
naming: Infer types also in implicit variable declarations

### LINKED ISSUE #3122: Use constant propagation with typed metavariables
It would be awesome if we could combine two of our most powerful operations: constant propagation and typed metavariables.

This snippet highlights the desired behavior: https://semgrep.dev/s/XLKN/?version=develop

Note the `$O[($ARG : float)]` not matching `value = obj[two]`.

This would be particularly helpful for Gitlab's `eslint.detect-object-injection` rule: https://gitlab.com/gitlab-org/security-products/analyzers/semgrep/-/blob/main/rules/eslint.yml.

IagoAbal:
This makes typed metavariables work in languages where variables are
implicitly declared "by assignment" such as in Python, Ruby, PHP, and in
"bad-style" JS.

Note: We were repeating the same piece of code too many times so I
decided to factor out some code into a new `declare_var` function.

Closes #3122

test plan:
make test # test included

PR checklist:
- [x] changelog is up to date


aryx (approved):

aryx:
Great. I wonder if we should have a step before naming that transform those Assign in VarDef, so then we don't have to think about those languages that implicitely declares variables. Are there other places except Naming_AST.ml where we have to think about whether an Assign is a Vardef?

IagoAbal:
> Great. I wonder if we should have a step before naming that transform those Assign in VarDef, so then we don't have to think about those languages that implicitely declares variables. Are there other places except Naming_AST.ml where we have to think about whether an Assign is a Vardef?

We do it also in Constant_propagation.ml, but this may change the day that we have a complete AST-to-IL translation and we can do the constant propagation on the IL.

PR #3169, opened 2021-05-21
Allow nested 'pattern-either'

### LINKED ISSUE #3168: Nested 'pattern-either' produces invalid YAML schema
The playground is changing my YAML rules when I try to share then, so I'll have to show these inline.

These examples are a bit artificial, but the original rule exhibiting this behavior was a more complex, valid extension of these examples.

`patterns` -> `patterns` works:

```
$ cat /tmp/test.yaml 
rules:
- id: test-patterns
  languages: [python]
  severity: ERROR
  message: Test patterns
  patterns:
  - patterns:
    - pattern: bar(...)

$ cat /tmp/test.py | python -m semgrep --config /tmp/test.yaml -
running 1 rules...
/tmp/tmp42mxeyg7
severity:error rule:tmp.test-patterns: Test patterns
1:bar('test')
```

`patterns` -> `pattern-either` works:

```
$ cat /tmp/test.yaml 
rules:
- id: test-patterns
  languages: [python]
  severity: ERROR
  message: Test patterns
  patterns:
  - pattern-either:
    - pattern: bar(...)

$ cat /tmp/test.py | python -m semgrep --config /tmp/test.yaml -
running 1 rules...
/tmp/tmpcl29ffhb
severity:error rule:tmp.test-patterns: Test patterns
1:bar('test')
```

`pattern-either` -> `patterns` works:

```
$ cat /tmp/test.yaml 
rules:
- id: test-patterns
  languages: [python]
  severity: ERROR
  message: Test patterns
  pattern-either:
  - patterns:
    - pattern: bar(...)

$ cat /tmp/test.py | python -m semgrep --config /tmp/test.yaml -
running 1 rules...
/tmp/tmpgx8kv6zx
severity:error rule:tmp.test-patterns: Test patterns
1:bar('test')
```

`pattern-either` -> `pattern-either` fails:

```
$ cat /tmp/test.yaml 
rules:
- id: test-patterns
  languages: [python]
  severity: ERROR
  message: Test patterns
  pattern-either:
  - pattern-either:
    - pattern: bar(...)

$ cat /tmp/test.py | python -m semgrep --config /tmp/test.yaml -
semgrep error: Invalid rule schema
  --> /tmp/test.yaml:7
7 |   - pattern-either:
8 |     - pattern: bar(...)

Additional properties are not allowed ('pattern-either' was unexpected)

no valid configuration file found (1 configs were invalid)
```

mschwager:
Fixes #3168.

I'm still on the fence as to whether we actually want this or not. A nested `pattern-either` can always be combined with it's parent `pattern-either` to produce an equivalent rule. Currently I'm leaning toward allowing this for a few reasons:

1. Users often aren't concerned about the technical precision of their rule, and want it to Just Work.
  i. This is especially apparent in very long rules. It can be difficult to tell at first glance you're in a very long `pattern-either`.
2. We allow other, basic noop operations as seen in #3168.
3. We can easily write a meta rule to catch and warn on this behavior rather than hard fail.

Overall, I think being liberal in what we accept ([Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle)) and getting out of the user's way is best. Thoughts?

### DIFF
--- a/semgrep/semgrep/rule_schema.yaml
+++ b/semgrep/semgrep/rule_schema.yaml
@@ -19,9 +19,11 @@ definitions:
       - $ref: '#/definitions/metavariable-comparison'
   pattern-either-content:
     type: array
+    title: "Return finding where any of the nested conditions are true"
     items:
       anyOf:
       - $ref: '#/definitions/patterns'
+      - $ref: '#/definitions/pattern-either'
       - $ref: '#/definitions/pattern-inside'
       - $ref: '#/definitions/pattern'
       - $ref: '#/definitions/pattern-regex'
--- a/semgrep/tests/e2e/rules/nested-pattern-either.yaml
+++ b/semgrep/tests/e2e/rules/nested-pattern-either.yaml
@@ -1,0 +1,12 @@
+rules:
+- id: test
+  message: test
+  severity: WARNING
+  languages: [javascript]
+  pattern-either:
+    - pattern-either:
+      - pattern: nested_patterns_func('foo', ...)
+      - pattern: nested_patterns_func('bar', ...)
+    - pattern-either:
+      - pattern: nested_patterns_func($X, 1, ...)
+      - pattern: nested_patterns_func($X, 2, ...)
--- a/semgrep/tests/e2e/snapshots/test_check/test_nested_pattern_either_rule/results.json
+++ b/semgrep/tests/e2e/snapshots/test_check/test_nested_pattern_either_rule/results.json
@@ -1,0 +1,103 @@
+{
+  "errors": [],
+  "results": [
+    {
+      "check_id": "rules.test",
+      "end": {
+        "col": 35,
+        "line": 2
+      },
+      "extra": {
+        "is_ignored": false,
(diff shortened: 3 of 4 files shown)


semgrep/semgrep/rule_schema.yaml
```diff
@@ -19,9 +19,11 @@ definitions:
       - $ref: '#/definitions/metavariable-comparison'
   pattern-either-content:
     type: array
+    title: "Return finding where any of the nested conditions are true"
```
mschwager, line 22: Unrelated: to be consistent with `patterns-content` above.

brendongo:
I'm supportive of allowing it in the spirit of robustness. But we should check that it does work with `--optimizations all` and we're not adding more work to get ocaml rule running to feature parity.

aryx (approved):
looks good to me.

aryx:
There was no change to evaluation.py; the code was probably ready to accept such nested pattern-either. I think the same is true for semgrep-core, so we should be good to go; it should work with or without --optimizations.

PR #3171, opened 2021-05-21
Add config file option for benchmarks

minusworld:
This adds the ability to set up benchmark jobs with a config YAML file.

An example file looks like this:
```
runs:
- name: test_python
  repos:
  - url: https://github.com/django/django
    commit_hash: 7cca22964c09e8dafc313a400c428242404d527a
  rule_configs:
  - 'r/python'
  opts: []
- name: test_ruby
  repos:
  - url: https://github.com/rails/rails
    commit_hash: f7b65ee0c41c534d1360f50cc136b01b8b63271b
  rule_configs:
  - 'r/ruby'
  opts: []
- name: test_javascript
  repos:
  - url: https://github.com/bkimminich/juice-shop
    commit_hash: e07059edd3b4ed0527add13d0dd5c8570daa3bcc
  rule_configs:
  - 'r/javascript'
  opts: [] 
```

Run it with `./run-benchmarks --config <file>`

### DIFF
--- a/perf/run-benchmarks
+++ b/perf/run-benchmarks
@@ -21,9 +21,19 @@ import time
 import urllib.request
 from collections import defaultdict
 from contextlib import contextmanager
-from typing import Iterator
+from typing import Iterator, List, Optional
 from typing import Tuple
+from pathlib import Path
 from RepositoryTimePerRule import RepositoryTimePerRule
+from config import SemgrepBenchmarkConfig, BenchmarkRunSetupData
+import logging
+import sys
+
+logger = logging.getLogger(__file__)
+logger.setLevel(logging.DEBUG)
+handler = logging.StreamHandler(stream=sys.stderr)
+handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
+logger.addHandler(handler)
 
 import requests
 import semgrep_core_benchmark  # type: ignore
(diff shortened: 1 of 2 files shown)


perf/run-benchmarks
```diff
@@ -453,18 +579,21 @@ def run_benchmarks(
...
+    elif gitlab:
         corpuses = GITLAB_CORPUSES
         variants = GITLAB_VARIANTS
-    if small_only:
+    elif small_only:
         corpuses = SMALL_CORPUSES
-    if all:
+    elif all:
         corpuses = SMALL_CORPUSES + MEDIUM_CORPUSES + LARGE_CORPUSES
+    elif config_file:
```
colleend, line 591: should this be `elif not config_file is None`?

perf/run-benchmarks
```diff
@@ -453,18 +579,21 @@ def run_benchmarks(
     corpuses = SMALL_CORPUSES + MEDIUM_CORPUSES
     if dummy:
         corpuses = DUMMY_CORPUSES
-    if internal:
+    elif internal:
         corpuses = INTERNAL_CORPUSES
-    if gitlab:
```
colleend, line 458: I feel like there might be a case where you want to run internal and github corpuses? not super important right now tho 🤷

colleend (approved):
looks good :D exciting progress!

perf/run-benchmarks
minusworld: Yeah, I didn't want to change any of the existing functionality for right now. If we decide to make things easier in the future, we can do that

perf/run-benchmarks
minusworld: On this line (https://github.com/returntocorp/semgrep/pull/3171/files#diff-903f08c7f7a15ccdd118e5e7e62172cac2dd2e20ce2b1d8bbd31bf75f6808994R763), the `config_file` parameter is either set to a `Path` or is `None`. Python evaluates `None` as `False`, so if `config_file` is `None` this will be skipped.

perf/run-benchmarks
colleend: yep, sounds good 👍

PR #3173, opened 2021-05-24
Support globstar (**) syntax in paths include/exclude with wcmatch.

Isaac0616:
In #2851, I tried to use `pathspec` to support globstar (**) syntax (#1834).
However, there is a bug which prevents matching absolute paths and I can't get the library to fix it.
I tried another library `wcmatch` to support the globstar syntax in this pull request.
[`wcmatch`](https://facelessuser.github.io/wcmatch/) is more actively maintained and supports similar APIs/semantic to standard library's `glob` and `pathlib`.
It makes this updates backward compatible (supposed to be, mostly).
(In contrast to `pathspec`, which is slightly different and follows gitignore's semantics.)

In this pull request, I only changed the `match_glob` function.
This is probably the easiest way / minimum changes to support the case that pattern `foo/bar` should match paths `x/foo/bar`, `foo/bar/x`, and `x/foo/bar/x`.
(`wcmatch` follows bash's semantics, so the pattern should be `**/foo/bar/**` to natively match `x/foo/bar`, `foo/bar/x`, and `x/foo/bar/x`).
It is not very clean and might have performance impact (hopefully not too bad compare to the original).
I can see some complexity comes from that we have to shell out to git to respect gitignore.
I am thinking whether there is a better way to implement `TargetManager`, but if this pull request is acceptable, that can be a future work.
I will also have to figure out a good benchmark for such changes.

### Test plan
I updated the unit test to add the test cases for globstar and question mark and make sure the previous tests still pass.
I also made the unit tests to compare the actual content of the `Set` instead of only comparing the length of the `Set`.
(Some e2e tests failed, but they seems to fail before this pull request.)

### DIFF
--- a/semgrep/semgrep/target_manager.py
+++ b/semgrep/semgrep/target_manager.py
@@ -12,6 +12,7 @@ from typing import List
 from typing import Set
 
 import attr
+from wcmatch import glob as wcglob
 
 from semgrep.config_resolver import resolve_targets
 from semgrep.error import FilesNotFoundError
@@ -189,12 +190,21 @@ class TargetManager:
         return frozenset(expanded)
 
     @staticmethod
-    def match_glob(path: Path, globs: List[str]) -> bool:
+    def preprocess_path_patterns(patterns: List[str]) -> List[str]:
+        """Convert semgrep's path include/exclude patterns to wcmatch's glob patterns.
+
+        In semgrep, pattern "foo/bar" should match paths "x/foo/bar", "foo/bar/x", and
+        "x/foo/bar/x". It implicitly matches zero or more directories at the beginning and the end
+        of the pattern. In contrast, we have to explicitly specify the globstar (**) patterns in
+        wcmatch. This function will converts a pattern "foo/bar" into "**/foo/bar" and
+        "**/foo/bar/**". We need the pattern without the trailing "/**" because "foo/bar.py/**"
+        won't match "foo/bar.py".
         """
-        Return true if path or any parent of path matches any glob in globs
-        """
-        subpaths = [path, *path.parents]
-        return any(p.match(glob) for p in subpaths for glob in globs)
+        result = []
+        for pattern in patterns:
+            result.append("**/" + pattern)
+            result.append("**/" + pattern + "/**")
+        return result
 
     @staticmethod
     def filter_includes(arr: FrozenSet[Path], includes: List[str]) -> FrozenSet[Path]:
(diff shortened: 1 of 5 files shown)


Isaac0616:
I tried to be backward compatible and made pattern `foo/bar` matches paths `x/foo/bar`, `foo/bar/x`, and `x/foo/bar/x`.
In other words, there are implicit `**` before and after the pattern (more precisely, `foo/bar` is `**/foo/bar` + `**/foo/bar/**`).
We can discuss whether we should keep the behavior.
Requiring explicit `**` before the pattern allow `foo/bar` to match only from the root folder, which seems useful.
For the `**` after the pattern, I can see that it is handy if the pattern is a directory.

semgrep/semgrep/target_manager.py
```diff
@@ -211,7 +217,12 @@ def filter_includes(arr: Set[Path], includes: List[str]) -> Set[Path]:
     def filter_excludes(arr: Set[Path], excludes: List[str]) -> Set[Path]:
         """
         Returns all elements in arr that do not match any excludes pattern
+
+        If includes is empty, returns arr unchanged
```
brendongo:
```suggestion
        If excludes is empty, returns arr unchanged
```

semgrep/setup.py
```diff
@@ -108,6 +108,7 @@ def find_executable(env_name, exec_name):
         "tqdm>=4.46.1",
         "packaging>=20.4",
         "jsonschema~=3.2.0",
+        "wcmatch>=8.2",
```
brendongo:
```suggestion
        "wcmatch==8.2",
```
Can we pin to a specific version of wcmatch for now. A change in file targeting behavior would have an outsized impact on semgrep behavior.

brendongo (commented):
Thanks for the PR @Isaac0616 looks good. Really appreciate the tests. Something idk if you've considered is what happens to glob matching when a user runs `semgrep --config xyz --exclude */bar/* foo` and it turns out the full path of foo is `/a/b/c/bar/d/e/foo`?

Isaac0616:
Hi @brendongo,
Sorry for the delay, I was working on something with a tight deadline.
I updated the commit according to your comments.
Regarding the question of `semgrep --config xyz --exclude '**/bar/**' foo`, `foo` won't be excluded because we are providing the relative path.
If we on the other hand use `semgrep --config xyz --exclude '**/bar/**' /a/b/c/bar/d/e/foo`, `foo` will be excluded.
I think this behavior is consistent without the changes in this commit.
That is, `semgrep --config xyz --exclude 'bar' /a/b/c/bar/d/e/foo` will exclude `foo` but `semgrep --config xyz --exclude 'bar' foo` won't (`bar` implies `**/bar/**` in our current setting).
Does it answer your question?

semgrep/semgrep/target_manager.py
```diff
@@ -193,7 +194,12 @@ def match_glob(path: Path, globs: List[str]) -> bool:
         Return true if path or any parent of path matches any glob in globs
         """
         subpaths = [path, *path.parents]
-        return any(p.match(glob) for p in subpaths for glob in globs)
+        return any(
+            wcpathlib.PurePath(p).match(
+                globs, flags=wcpathlib.GLOBSTAR | wcpathlib.DOTGLOB
+            )
+            for p in subpaths
+        )
```
Isaac0616: wcmatch's `PurePath.match` is slower and looping like this is quite a performance hit.
Let me try another approach.

semgrep/semgrep/target_manager.py
```diff
@@ -188,12 +189,21 @@ def expand_targets(
         return expanded
 
     @staticmethod
-    def match_glob(path: Path, globs: List[str]) -> bool:
+    def preprocess_path_patterns(patterns: List[str]) -> List[str]:
```
Isaac0616, line 193: @brendongo 
Because calling `wcmatch`'s `PurePath.match` for each subpaths is too slow, I switched to this approach.
Let me know what do you think.
It looks a bit hacky, but I believe we can use unit tests to ensure it works as expected.

brendongo: Yeah looks good to me!

brendongo:
@Isaac0616 thanks for the PR it looks great! Do you mind fixing the conflict and we can get this merged

brendongo (approved):

mjambon:
Hi, we're planning to revert some of this behavior to get closer to what git does. See https://github.com/returntocorp/semgrep/issues/6835

tl;dr according to the [gitignore spec](https://git-scm.com/docs/gitignore) `a/b` is anchored and is equivalent to `/a/b` while `a` is unanchored and equivalent to `**/a`. we will implement this.

PR #3181, opened 2021-05-24
semgrep_main: Get project URL from CI env var

underyx:
PR checklist:
- [ ] changelog is up to date


chmccreery:
@underyx I decided against this and went with the manual-file-search of .git/config in order to keep the .git/config file as the source of truth (based on what you mentioned with https vs ssh, I realized we might double-count repositories if some runs happened it GitLab CI and others happened locally). 

Ok to close this?

underyx:
Yup!

PR #3182, opened 2021-05-24
fix: _raw in Rule object accurately reflects raw rule file

### LINKED ISSUE #3179: Rule hash metric should be independent of file path
**Describe the bug**
A config yaml should hash to the same value no matter where the file is stored

**To Reproduce**
Download a config yaml to two different locations, and print the metric rule hash.

**Expected behavior**
Hash should be same value.

**Environment**
Local 0.52

brendongo:
Rule._raw will now accurate reflect the unchanged rule as it was read instead of modifying the rule_id in that yaml
when rewriting the rule_id to reflect path

Closes https://github.com/returntocorp/semgrep/issues/3179

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -128,7 +128,7 @@ class Config:
         configs = self.valid
         if not no_rewrite_rule_ids:
             # re-write the configs to have the hierarchical rule ids
-            configs = self._rename_rule_ids(configs)
+            self._rename_rule_ids(configs)
 
         return list(
             OrderedDict.fromkeys([rule for rules in configs.values() for rule in rules])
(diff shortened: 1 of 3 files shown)


brendongo:
Verified that running with and without `--no-rewrite-rule-ids` generates the same rule hashes

mschwager (approved):
> Verified that running with and without --no-rewrite-rule-ids generates the same rule hashes

If we're relying on this, can you add a test to guarantee this moving forward?

brendongo:
Because we shell out to run semgrep with arguments I think it's a little involved. I found some other issues with this method of hashing so will be refactoring to something more stable. Will get this in as is but I think the new solution will be more testable.

PR #3183, opened 2021-05-24
Add a backup way of obtaining git url if docker image does not have git

chmccreery:
This should work for any git configuration. 
First, we try to obtain the project url through git ls-remote --get-url
If that fails, we manually look at the .git/config file to pull out the url ourselves

I also considered using the environment variable CI_PROJECT_URL from GitLab.
I decided to go this route because it was easier to test, and works for all git projects, not just for GitLab.
The tradeoff is this is slower and a little more complicated - requires file IO rather than an env variable that might "just work"
I am definitely open to reversing that decision, though!

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/semgrep_main.py
+++ b/semgrep/semgrep/semgrep_main.py
@@ -301,8 +301,11 @@ The two most popular are:
         )
     except Exception as e:
         logger.debug(f"Failed to get project url from 'git ls-remote': {e}")
-        # add \n to match urls from git ls-remote (backwards compatability)
-        project_url = manually_search_file(".git/config", ".com", "\n")
+        try:
+            # add \n to match urls from git ls-remote (backwards compatability)
+            project_url = manually_search_file(".git/config", ".com", "\n")
+        except Exception as e:
+            logger.debug(f"Failed to get project url from .git/config: {e}")
 
     project_hash = (
         hashlib.sha256(project_url.encode()).hexdigest() if project_url else None
--- a/semgrep/semgrep/semgrep_main.py
+++ b/semgrep/semgrep/semgrep_main.py
@@ -6,10 +6,11 @@ import time
 from io import StringIO
 from pathlib import Path
 from re import sub
-from typing import Any, Tuple
+from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Tuple
 from typing import Union
 
 import attr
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/semgrep_main.py
```diff
@@ -288,16 +293,20 @@ def main(
...
             ["git", "ls-remote", "--get-url"],
             encoding="utf-8",
             stderr=subprocess.DEVNULL,
         )
-        project_hash = hashlib.sha256(project_url.encode()).hexdigest()
     except Exception as e:
-        logger.debug(f"Failed to generate project hash: {e}")
+        logger.debug(f"Failed to get project url from 'git ls-remote': {e}")
+        # add \n to match urls from git ls-remote (backwards compatability)
+        project_url = manually_search_file(".git/config", ".com", "\n")
```
brendongo: Probably also want to put this in a try-except since file reading can potentially fail (doesn't exits, wrong permissions, wrong format, etc.)

semgrep/semgrep/util.py
```diff
@@ -161,6 +162,19 @@ def sub_check_output(cmd: List[str], **kwargs: Any) -> Any:
     return result
 
 
+def manually_search_file(path: str, search_term: str, suffix: str) -> Optional[str]:
+    """
+    Searches a file for the given search term and, if found,
+    returns the first word that contains that search term
+    """
+    fd = open(path, mode="r")
```
brendongo: Do we want to check file exists before reading? If we are in a subdirectory of the git repo .git/config will not exist

brendongo: We might also want to use the with open(...) as f: syntax https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files so python handles closing the file for us

chmccreery: fixed

semgrep/semgrep/semgrep_main.py
chmccreery: fixed

semgrep/semgrep/util.py
chmccreery: woop didn't see the second comment - one min

chmccreery: ok fixed 😄

brendongo (approved):

PR #3193, opened 2021-05-25
test_rules: Handle the special case of .test.yaml files

IagoAbal:
test plan:
semgrep-core -test_rules path/to/semgrep-rules/yaml # now works

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Test_engine.ml
+++ b/semgrep-core/src/engine/Test_engine.ml
@@ -57,6 +57,9 @@ let test_rules ?(ounit_context = false) xs =
           * FT.Config ((* | FT.Json*) FT.Jsonnet) when not ounit_context -> true
           *)
          | _ -> false)
+    |> Common.exclude (fun filepath ->
+           (* .test.yaml files are YAML target files rather than config files! *)
+           Filename.check_suffix filepath ".test.yaml")
     |> Skip_code.filter_files_if_skip_list ~root:xs
   in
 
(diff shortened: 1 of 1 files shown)


semgrep-core/src/engine/Test_engine.ml
```diff
@@ -57,6 +57,9 @@ let test_rules ?(ounit_context = false) xs =
           * FT.Config ((* | FT.Json*) FT.Jsonnet) when not ounit_context -> true
           *)
          | _ -> false)
+    |> List.filter (fun filepath ->
```
aryx: You can use Common.exclude and remove the not.

semgrep-core/src/engine/Test_engine.ml
```diff
@@ -89,12 +92,24 @@ let test_rules ?(ounit_context = false) xs =
...
-                    if b = b2 && ext <> ext2 && ext2 <> "jsonnet" then
-                      Some (Filename.concat d file2)
-                    else None)
+                    let path2 = Filename.concat d file2 in
+                    (* In order to distinguish YAML test files from rule files,
+                     * YAML test files use the .test.yaml extension.
+                     * THINK: If we used .yml or .test_yaml we would not need
+                     *        to handle this case separately. *)
+                    match
+                      Filename.chop_suffix_opt ~suffix:".test.yaml" file2
```
aryx, line 101: I'm surprised ocamlformat use this indentation. Are you sure you set pre-commit correctly?

aryx (approved):

aryx:
I try to have a 'test plan:' section in all my PRs/commits describing how another person can test the changed code.
Here maybe you could say "semgrep-core -test_rules ~/github/semgrep-rules/meta/ now works.

semgrep-core/src/engine/Test_engine.ml
IagoAbal: I think it's correctly setup, it probably does that because otherwise the line would be too long. I can introduce a `let` to split it up.

IagoAbal:
> I try to have a 'test plan:' section in all my PRs/commits describing how another person can test the changed code.
> Here maybe you could say "semgrep-core -test_rules ~/github/semgrep-rules/meta/ now works.

My bad, I added it now. BTW I think `meta` is now `yaml/semgrep`.

PR #3194, opened 2021-05-25
adding more repositories for rules we don't have timing info for

colleend:
* This adds in 3 repositories that add in small java, c, and dockerfile repositories.
* I used a script to calculate which rules we didn't have timing info for, and these three languages did pop up in higher numbers.
* Also made it so `output-time-per-rule-json` always uses the standard variants.

_Test Plan:_
Run 
`./run-benchmarks --output-time-per-rule-json test.json`

These are the results I get for running on the new repositories: 

[test.txt](https://github.com/returntocorp/semgrep/files/6542922/test.txt)


perf/run-benchmarks
```diff
@@ -453,6 +453,9 @@ def run_benchmarks(
         variants = GITLAB_VARIANTS
     elif small_only:
         corpuses = SMALL_CORPUSES
+    elif output_time_per_rule_json:
```
minusworld: Do we need this? If we use this flag, we can still use it with the other flags like `--std-only` and `--small-only` independent of whether it reports back average rule runtimes

colleend: right, the reason why I added this was because I wanted to make sure that the `--output-times-per-json` flag is run with `--std-only`. The times per-repository aren't accurate without running `--std-only` (because it will only output the latest run for a specific repository, so if all the variants run then it will output the results for the last optimized run). The times per-rule also come out not completely accurate because they are added together then divided for each run 🤔 

Perhaps a better way to do this is to throw an error if `output-time-per-rule` isn't rule with `std-only`.

perf/bench/r2c-rules/dockerfile.yml
```diff
@@ -0,0 +1,689 @@
...
+  paths:
+    include:
+    - '*dockerfile*'
+    - '*Dockerfile*'
+- id: generic.dockerfile.best-practice.nonsensical-command.nonsensical-command
+  patterns:
+  - pattern: RUN $CMD
+  - metavariable-regex:
+      metavariable: $CMD
+      regex: (shutdown|service|ps|free|top|kill|mount|ifconfig|nano|vim)
```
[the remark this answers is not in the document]
colleend: these are not applicable bc this is not an actual dockerfile.

perf/run-benchmarks
minusworld: OK, let's add a TODO comment explaining why this is the case and that, in the future, we'll support aggregate times for all variants

minusworld: Add a logging INFO line that reports that this flag will only run with std benchmarks

minusworld (approved):

PR #3196, opened 2021-05-26
parse_rule: Parse pattern-not-regex patterns

IagoAbal:
test plan:
semgrep-core -test_rules path/to/semgrep-rules/generic/secrets # now works

PR checklist:
- [ ] changelog is up to date


IagoAbal:
I was a bit surprised that it was that easy, so I took a look at the Python wrapper to check the handling of `pattern-not-regex` and I think this is correct.

aryx (approved):

aryx:
excellent

PR #3201, opened 2021-05-26
Add hard timeout for benchmarks

minusworld:
Adds a hard timeout option for benchmarks which will kill the job after the timeout has passed. This is for development purposes so we can test large benchmark jobs faster

PR checklist:
- [ ] changelog is up to date


perf/run-benchmarks
```diff
@@ -276,15 +185,13 @@ def output_differences(
     return fd_len, sd_len
 
 def run_semgrep(
-    docker: str, corpus: Corpus, variant: SemgrepVariant, include_time: bool
-) -> Tuple[float, bytes]:
+    docker: str, corpus: Corpus, variant: SemgrepVariant, include_time: bool, hard_timeout: int = 0
+) -> Tuple[float, dict]:
     args = []
     common_args = [
-        "--strict",
```
colleend, line 283: do we not want to run with strict?

perf/RepositoryTimePerRule.py
```diff
@@ -53,20 +53,23 @@ def times_per_file_to_times_per_rule(self, repo_name: str, times_per_file_bytes:
...
+
         # this is for rule_id -> average file time, want to remove all 0 run-times
         repo_time_per_rule_no_zeroes: Dict[str, TotalTimeAndTargetNum] = defaultdict(TotalTimeAndTargetNum)
+
         for target in times_per_file['time']['targets']:
-            total_time_per_rule = list(map(add, target['parse_times'], total_time_per_rule))
-            for idx, parse_time in enumerate(target['parse_times']):
-                if parse_time > 0:
-                    repo_time_per_rule_no_zeroes[rule_ids[idx]] = repo_time_per_rule_no_zeroes[rule_ids[idx]] + TotalTimeAndTargetNum(parse_time, 1)
+            total_time_per_rule = list(map(add, target['run_times'], total_time_per_rule))
```
colleend, line 61: good catch with the 'parse_times' -> 'run_times' change!

colleend (approved):
lgtm!

perf/run-benchmarks
minusworld: Oh, good catch - I was debugging, so I removed it and forgot to put this back in

PR #3206, opened 2021-05-27
Use metavariables in messages for experimental

emjin:
Separate metavariable propagation into a different function and use it in run_rules_direct_to_semgrep_core. Re-expose the message check in run-benchmarks

Test plan: ./run-benchmarks

PR checklist:
- [x] changelog is up to date


aryx (approved):
Lots of unrelated formatting changes. Not sure what's going on here, but looks good.

PR #3209, opened 2021-05-27
[Internal] Using Semgrep.atd also in semgrep-core for JSON comm

aryx:
This is cleaner and will help evolve the interface, for example
to allow empty metavariables.

test plan:
make test

PR checklist:
- [x] changelog is up to date


aryx:
have to investigate the CI failures, not sure what's going on here ...

mjambon (approved):
looks good to me!

semgrep-core/src/cli/Main.ml
```diff
@@ -737,23 +738,25 @@ let semgrep_with_patterns lang (rules, rule_parse_time) files_or_dirs =
...
-      (* so we don't get also the profile output of Common.main_boilerplate*)
-      Common.profile := Common.ProfNone;
-      flds @ [ ("profiling", json) ] )
-    else flds
-  in
+  let res = JSON_report.match_results_of_matches_and_errors files res in
+  (* TODO need change type match_results.time to a choice
+     let res =
+       if !profile then (
+         let json = JSON_report.json_of_profile_info !profile_start in
```
aryx, line 745: @emjin I've commented this code for now during my migration to atd to specify the JSON output of semgrep-core, but anyway is this code still used? The profiling information it returns use a different format from the other profiling information with a key and then a time: and count: fields. Is this used python-side?

aryx: I don't see any use of the "profiling" string python side (I see use of the "time", but not of the "profiling" key which is used below), so I guess it's safe to remove?

aryx:
Not sure why but the rails repo fails locally on my machine. Seems to work in CI though. Weird.

semgrep-core/src/cli/Main.ml
emjin: I thought this was used if -profile was passed into semgrep-core? But if not then it's not being used

PR #3215, opened 2021-05-27
fix: don't do file extension filtering in semgrep-core

brendongo:
semgrep-cli has logic that allows files without the correct extension to
be explicitly passed and scanned for a given language. With `--optimizations all`
this behavior is broken since semgrep-core is filtering out files without expected
filenames when doing whole rule matching. This PR removes that filtering.

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -349,13 +349,6 @@ let xlang_of_string s =
       let lang = lang_of_string s in
       R.L (lang, [])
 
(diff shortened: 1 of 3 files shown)


emjin:
Looks good to me, tagging @aryx because I'm not sure if this logic will affect something else

aryx (approved):

aryx:
I'll add the PR comment in the code, so if someone wants to put back files_or_dirs he will know he should not.

PR #3218, opened 2021-05-27
Add ability to use paths as config strings with --config files in ./run-benchmarks

minusworld:
PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/perf/RepositoryTimePerRule.py
+++ b/perf/RepositoryTimePerRule.py
@@ -21,7 +21,7 @@ class TotalTimeAndTargetNum:
         self.total_rule_time = total_rule_time
         self.num_targets = num_targets
 
-    def __add__(self, other: TotalTimeAndTargetNum) -> TotalTimeAndTargetNum:
+    def __add__(self, other: "TotalTimeAndTargetNum") -> "TotalTimeAndTargetNum":
         added_total_rule_time = self.total_rule_time + other.total_rule_time
         added_num_targets = self.num_targets + other.num_targets
         return TotalTimeAndTargetNum(added_total_rule_time, added_num_targets)
--- a/perf/RepositoryTimePerRule.py
+++ b/perf/RepositoryTimePerRule.py
@@ -1,81 +1,115 @@
 import json
 import logging
 import sys
-
 from collections import defaultdict
-from typing import Dict, List
 from operator import add
+from typing import Dict
+from typing import List
+from typing import Optional
 
 logger = logging.getLogger(__file__)
 logger.setLevel(logging.INFO)
 handler = logging.StreamHandler(stream=sys.stderr)
-handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
+handler.setFormatter(
+    logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+)
 logger.addHandler(handler)
+
 
 class TotalTimeAndTargetNum:
     def __init__(self, total_rule_time: float = 0.0, num_targets: int = 0):
         self.total_rule_time = total_rule_time
         self.num_targets = num_targets
 
-    def __add__(self, other):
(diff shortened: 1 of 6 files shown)


perf/config.py
```diff
@@ -1,33 +1,125 @@
...
+import logging
+import sys
+import time
 from pathlib import Path
+from typing import List
+from typing import Optional
+from urllib.parse import urlparse
+
 import attr
+import requests
```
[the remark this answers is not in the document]
aryx: yeah!

aryx:
What is the test plan:?
What is an example of a command using that new feature? (I just don't understand the title of the PR; I thought
--config was already accepting paths to a config file).

minusworld:
> What is the test plan:?
> What is an example of a command using that new feature? (I just don't understand the title of the PR; I thought
> --config was already accepting paths to a config file).

This is for `perf/` benchmarking; this enables the ability to use filesystem paths as part of a benchmarking config which we added earlier.

An example benchmark conifg would look like this:

```
runs:
- name: test_ruby
  repos:
  - url: https://github.com/presidentbeef/brakeman
    commit_hash: d6421d4878ba7b13b5c52dca051360d8b4d1e3eb
  rule_configs:
  - '~/semgrep-rules/ruby'
  opts: []
- name: test_go
  repos:
  - url: https://github.com/astaxie/beego
    commit_hash: 8f3d1c5f42fce57e83e1c3f7d180477595db7cca 
  rule_configs:
  - r/go
  - p/dgryski
- name: test_python
  repos:
  - url: https://github.com/pallets/flask
    commit_hash: 65a0a4b585781b5dd3db4009de3195e2ac7f049f
  rule_configs:
  - r/python
  opts: []
```

An example of running this would look like this: `./run-benchmarks --config <config_file> --std-only`

perf/RepositoryTimePerRule.py
```diff
@@ -1,81 +1,115 @@
...
 logger.addHandler(handler)
 
+
 class TotalTimeAndTargetNum:
     def __init__(self, total_rule_time: float = 0.0, num_targets: int = 0):
         self.total_rule_time = total_rule_time
         self.num_targets = num_targets
 
-    def __add__(self, other):
+    def __add__(self, other: "TotalTimeAndTargetNum") -> "TotalTimeAndTargetNum":
```
colleend, line 24: should this be in quotes? not sure if this matters 😅  <-- ignore this, I'm sure it doesn't

perf/config.py
```diff
@@ -1,33 +1,145 @@
...
+            return None
+        except requests.HTTPError as e:
+            logger.warning(
+                f"There was an issue connecting to '{rule_config_url}'; error: {e}"
+            )
+            return None
+        return r.text
+
+    def _fetch_rule_config_from_path(self, path: Path) -> Optional[str]:
+        pass
```
colleend, line 61: what's the main purpose of this function? 😲

perf/run-benchmarks
```diff
@@ -278,19 +277,16 @@ def run_semgrep(
     try:
         stdout_decoded = res.stdout.decode("utf-8")
         semgrep_results = json.loads(stdout_decoded)
-    except (UnicodeDecodeError, ValueError, json.JSONDecodeError):
```
colleend, line 281: are we not catching the UnicodeDecodeError anymore?

colleend (commented):
Thanks for refactoring & adding --clean!!

colleend (approved):
tested, works great 👍

perf/run-benchmarks
minusworld: Flake8 told me that UnicodeDecodeError is a subclass of ValueError. So if we catch ValueError we will also catch UnicodeDecodeError.

perf/config.py
minusworld: Oops, should've deleted this! I'll open another PR to remove.

perf/run-benchmarks
colleend: ah gotcha, that makes sense 👍

PR #3225, opened 2021-05-28
Consolidate suppressed CLI arguments

mschwager:


semgrep/semgrep/cli.py
```diff
@@ -422,6 +414,17 @@ def cli() -> None:
         help=argparse.SUPPRESS,
         # help="Deprecated alias for --json + --debug",
     )
+    output.add_argument(
+        "--save-test-output-tar",
```
mschwager, line 418: This argument is only used internally. Let's suppress it so it doesn't clutter up the help text, and so external users don't inadvertently start relying on it.

semgrep/semgrep/cli.py
```diff
@@ -422,6 +414,17 @@ def cli() -> None:
...
+        action="store_true",
+        help=argparse.SUPPRESS,
+        # help="Save --test output for use in semgrep-app registry",
+    )
+    output.add_argument(
+        "--synthesize-patterns",
```
mschwager, line 424: This is the old pattern recommendation functionality. It's also internal-only, and unused (?). Can we remove this from the Python interface at some point? cc @emjin

emjin: Yup, I don't think there are any near-term plans to revive pattern-from-code

brendongo (approved):

PR #3230, opened 2021-05-28
fix: set timings to null instead of -1 when nonexistent in metrics

brendongo:
This makes it easier for our backend to run aggregations when it doenst have to consider -1 values. It should already have desired behavior when the metric is set to null on non-existence.


emjin (approved):

emjin:
These look like the same errors as in https://github.com/returntocorp/semgrep/pull/3229/files, so probably unrelated to both

Issue #3159, opened 2021-05-20
Python try/except ellipsis missing cases


mschwager:
In Python, `try`/`except` blocks can contain multiple exceptions to catch. Sometimes we'll want to check if a particular exception type is included at all, not just if it's _the only_ exception type caught. E.g. https://semgrep.dev/s/B43v/

mschwager:
Fixed by https://github.com/returntocorp/semgrep/pull/3233

PR #3233, opened 2021-05-28
fixing mismatching error for ellipses in excepthandler

colleend:
This is a fix for https://github.com/returntocorp/semgrep/issues/3159, where ellipsis in the `except` clause of a try-except python statement does not match variations where there is only one error type.

_Test Plan:_
Run 
`semgrep-core/_build/install/default/bin/semgrep-core -lang python -f tests/python/misc_except_matching.sgrep tests/python/misc_except_matching.py`

There should be three matches :D


colleend:
not sure why lint isn't passing -- it passes locally 🤔

aryx:
@brendongo introduced some regressions in a previous commit. I'm reverting it here: https://github.com/returntocorp/semgrep/pull/3241
if you merge develop you should get back the tests in green.

aryx (approved):

PR #3234, opened 2021-05-29
fix: stablize project_hash even if username passwords are set

### LINKED ISSUE #3232: Project hash metric depends on passwords
**Describe the bug**
The project hash should be a stable function of the project location, and not dependent on username or password.

**To Reproduce**
Run `echo $(git ls-remote --get-url) | sha256sum` in a gitlab CI multiple times

**Expected behavior**
Our metrics return the same value from run to run

**Environment**
GitLab CI

brendongo:
Gitlab SAST adds username and password to their git ls-remote that changes per-run of CI. This PR stabilizes by removing username and password when the url is of the format: `scheme://username:password@hostname/path`

For now if the format is not of that form we don't do any sanitation of the url.

### DIFF
--- a/semgrep/semgrep/metric_manager.py
+++ b/semgrep/semgrep/metric_manager.py
@@ -6,6 +6,7 @@ from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Set
+from urllib.parse import urlparse
 
 from semgrep.constants import SEMGREP_USER_AGENT
 from semgrep.profiling import ProfilingData
@@ -45,8 +46,23 @@ class _MetricManager:
 
         self._send_metrics = False
 
-    def set_project_hash(self, project_hash: Optional[str]) -> None:
-        self._project_hash = project_hash
+    def set_project_hash(self, project_url: Optional[str]) -> None:
+        """
+        Standardizes url then hashes
+        """
+        if project_url is None:
+            self._project_hash = None
+        else:
+            parsed_url = urlparse(project_url)
+            if parsed_url.scheme == "https":
+                # Remove optional username/password from project_url
+                sanitized_url = f"{parsed_url.hostname}{parsed_url.path}"
+            else:
+                # For now don't do anything special with other git-url formats
+                sanitized_url = project_url
+
+            project_hash = hashlib.sha256(sanitized_url.encode()).hexdigest()
+            self._project_hash = project_hash
 
     def set_configs_hash(self, configs: List[str]) -> None:
         """
--- a/semgrep/semgrep/metric_manager.py
+++ b/semgrep/semgrep/metric_manager.py
@@ -1,5 +1,4 @@
 import hashlib
-import logging
 from pathlib import Path
 from typing import Any
 from typing import Dict
--- a/semgrep/semgrep/metric_manager.py
+++ b/semgrep/semgrep/metric_manager.py
@@ -1,4 +1,5 @@
 import hashlib
+import logging
 from pathlib import Path
 from typing import Any
 from typing import Dict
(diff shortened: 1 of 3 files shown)


brendongo:
Closes https://github.com/returntocorp/semgrep/issues/3232

semgrep/semgrep/metric_manager.py
```diff
@@ -46,8 +46,23 @@ def __init__(self) -> None:
...
+        """
+        Standardizes url then hashes
+        """
+        if project_url is None:
+            self._project_hash = None
+        else:
+            parsed_url = urlparse(project_url)
+            if parsed_url.scheme == "https":
+                # Remove optional username/password from project_url
+                sanitized_url = f"{parsed_url.hostname}{parsed_url.path}"
```
mschwager: Should we include other parsed URL information? E.g. `params`, `query`, `fragment`, or `port`?

Since we're generalizing the solution to all possible `git` repositories, and not just Gitlab/Github, there could be all kinds of URLs returned, not just what we've come to expect from those SCMs.

I suspect the failure mode of _not_ including this other information is that some weird SCM somewhere will have duplicate project hashes because the hostname + paths are all the same, and it uses the query string for lookup, or something like that. Basically, we over count a particular project hash because we avoided additional context.

The failure mode of including additional information seems to be the issue we faced originally: we accidentally include information that makes the metric useless. Then we have under counting instead of the over counting mentioned above.

It's hard to say what all the weird types of URLs we might receive are. Or the weird `git` authentication schemes leveraging this URL. We want to get this right insofar as we can, though, because every time we change this we either invalidate the old data, or make it difficult to connect the old and new.

What do you think?

semgrep/semgrep/metric_manager.py
```diff
@@ -46,8 +46,23 @@ def __init__(self) -> None:
...
-    def set_project_hash(self, project_hash: Optional[str]) -> None:
-        self._project_hash = project_hash
+    def set_project_hash(self, project_url: Optional[str]) -> None:
+        """
+        Standardizes url then hashes
+        """
+        if project_url is None:
+            self._project_hash = None
+        else:
+            parsed_url = urlparse(project_url)
```
mschwager: Should we catch a `ValueError` here, and perform some error handling (return `None`?), or would we rather fail loudly? The [`urlparse`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) docs state when this exception might occur.

semgrep/semgrep/metric_manager.py
adamberman: I think you're right to be broad, but not so broad as to cause us to create differences between different runs as the same repo. @brendongo do we have a sense of where that line lines for gitlab? wondering if you're excluding query for example b/c that might change run-to-run?

semgrep/semgrep/metric_manager.py
adamberman: maybe catch and re-raise a more helpful error? but i'm also fine with skipping this for now for the sake of expediency

adamberman (approved):
Approved pending resolution of Matt's line of question

semgrep/semgrep/metric_manager.py
brendongo: I think we prefer to undercount instead of possiibly overcounting. For now none of the urls we care about seem to use these extra url information so the risk of invalidating old data/not being able to connect old and new is low

semgrep/semgrep/metric_manager.py
brendongo: Caught and falls back to hashing the entire URL

PR #3244, opened 2021-05-31
Allow $...ARGS to also match an empty list

### LINKED ISSUE #3177: In Ruby, `$...ARGS` doesn't match all of the same places as `...` (Needed for autofix)
**Describe the bug**
In ruby, when using the ellipsis operator (i.e. `...`, it can match on 0 or more additional function parameters. I expect that the spread metavariable operator (i.e. `$...ARGS`) can match on the exact same set of items.

In practice, the ellipsis operator is matching 0 or more additional function parameters while the spread metavariable is only matching on 1 or more.

**To Reproduce**
For both of these test cases, I expect lines 1 through 3 to match, but the second only matches 1 and 2.

Use of `...` matches when 0 or more additional parameters are passed: https://semgrep.dev/s/1nAl
Use of `$...ARGS` only matches when 1 or more additional parameters are passed: https://semgrep.dev/s/9qY2

**Expected behavior**
`$...ARGS` matches the same content as `...`

Fixing this would make autofixes much more capable.

**Environment**
Both semgrep.dev and a local build

aryx:
This closes https://github.com/returntocorp/semgrep/issues/3177

I probably need to also change a few things Python side and in
JSON_report.ml to allow metavars without any content (an empty list)
otherwise we might get some NoTokenLocation errors

test plan:
test file included

PR checklist:
- [x] changelog is up to date


aryx:
I ended up not changing Semgrep.atd and use a start position = end position for metavariables with an empty list of arguments.
I think @brendongo mentioned that.

PR #3247, opened 2021-06-01
[Python] transform assigns in class defs as VarDef

### LINKED ISSUE #3212: [Python] semgrep cannot match class code, with or without aliasing
I need to match a whole class with a pattern and not use pattern-inside to find code inside a pattern
and semgrep does not match the class code...

reproduce:
https://semgrep.dev/s/E595

expected:
  match the class
  
 result:
  matches nothing

aryx:
This closes https://github.com/returntocorp/semgrep/issues/3212

test plan:
test file included
also
```
+ /home/pad/yy/_build/default/src/cli/Main.exe -dump_ast misc_metavar_field_anyorder.py
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("LOL", ()),
                 id_constness=Ref(None); }));
       attrs=[]; tparams=[]; },
      ClassDef(
        {ckind=(Class, ());
	 ...
         cbody=[FieldStmt(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("hah", ()),
                                id_constness=Ref(None); }));
                      attrs=[]; tparams=[]; },
                     VarDef({vinit=Some(L(Int((Some(1), ()))));
		     vtype=None; }))
		     ...
```

which now shows a VarDef for "hah"

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   properties](https://github.com/tree-sitter/tree-sitter-javascript/pull/166)
 - JavaScript parsing: [Allow default export for any declaration](https://github.com/tree-sitter/tree-sitter-javascript/pull/168)
 - Metavariables in messages are filled in when using `--optimizations all`
+- Python: class variables are matched in any order (#3212)
 
 ### Changed
 - $...ARGS can now match an empty list of arguments, just like ... (#3177)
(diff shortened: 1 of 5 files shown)


aryx:
Ultimately we should do that for the other Assign in Python and JS which are really disguised VarDef.

IagoAbal:
> Ultimately we should do that for the other Assign in Python and JS which are really disguised VarDef.

I guess we would have to do that in Naming_AST since that's where we can recognize which assigns are actually VarDefs.

PR #3248, opened 2021-06-01
Experimental support for Comby

aryx:
This is just the first PR for comby.
Next steps are to handle metavariables, and also
to figure out why the line/col and the number of matches
returned by Comby seems wrong

test plan:
test file included

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -63,6 +63,7 @@ and xpattern_kind =
   | Sem of Pattern.t * Lang.t (* language used for parsing the pattern *)
   | Spacegrep of Spacegrep.Pattern_AST.t
   | Regexp of regexp
+  | Comby of string
 
 (* used in the engine for rule->mini_rule and match_result gymnastic *)
 and pattern_id = int [@@deriving show, eq]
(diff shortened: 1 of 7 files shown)


semgrep-core/tests/OTHER/rules/comby_basic.js
```diff
@@ -0,0 +1,14 @@
+function foo() {
+ //ERROR:
+    foo(1,2);
+
+ //TODO should detect that:
```
aryx: seems to work on the comby playground: bit.ly/3i8SgLW weird.

semgrep-core/src/core/Rule.ml
```diff
@@ -73,8 +74,7 @@ let mk_xpat pat pstr =
   incr count;
   { pat; pstr; pid = !count }
 
-let is_regexp xpat =
-  match xpat.pat with Regexp _ -> true | Sem _ | Spacegrep _ -> false
+let is_regexp xpat = match xpat.pat with Regexp _ -> true | _ -> false
```
IagoAbal, line 77: It's better to enumerate all cases when it's just 3 left.

PR #3253, opened 2021-06-02
Propagate metavariables into match results in experimental

emjin:
After a pattern composition is performed, there may be more metavariables pertain to the actual result range. When reporting results, report those metavariables in addition

Test plan:

In semgrep/semgrep/tests/e2e/ run

semgrep --config rules/metavariable_propagation/metavariable-regex-propagation.yaml targets/metavariable_propagation/metavariable-regex-propagation.py --json --optimizations

to confirm that the message returned uses foo, not $FUNC.

To ensure this does not affect anything else, make optimizations default in core_runner and run all tests, checking against known errors

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -196,7 +196,9 @@ let (range_to_pattern_match_adjusted :
       message = r.R.message (* keep pattern_str which can be useful to debug *);
     }
   in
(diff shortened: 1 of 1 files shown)


aryx (approved):
Typo in evaluate

IagoAbal (approved):

PR #3256, opened 2021-06-02
ci: add pr comment on non-blocking perf regressions

### LINKED ISSUE #3100: Add PR comment on non blocking performance regressions
We want performance regressions to block the build if they are beyond a certain threshold but just
alert when they are slower but not past the threshold.

Github actions just has a binary pass fail so a suggested improvement is to add a github comment to
warn of slight regressions.

Relevant: https://github.com/unsplash/comment-on-pr

brendongo:
Closes: https://github.com/returntocorp/semgrep/issues/3100

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/scripts/compare_perf.py
+++ b/scripts/compare_perf.py
@@ -67,9 +67,11 @@ def main() -> None:
         total_latest += latest_time
 
         # Send PR comment if performance is slower but not past blocking threshold
-        if latest_time < baseline_time * 1.1 or latest_time - baseline_time < 3.5:
+        if latest_time > baseline_time * 1.1:
             send_comment(
-                "Potential non-blocking slowdown", github_token, pull_request_number
+                f"Potential non-blocking slowdown latest time {latest_time} is over 12 percent slower than baseline {baseline_time}. See run output for more details",
+                github_token,
+                pull_request_number,
             )
 
     # Assert the rules in aggregate are not more than 6% slower than baseline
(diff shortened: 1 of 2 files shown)


scripts/compare_perf.py
```diff
@@ -37,6 +66,12 @@ def main() -> None:
         total_baseline += baseline_time
         total_latest += latest_time
 
+        # Send PR comment if performance is slower but not past blocking threshold
+        if latest_time < baseline_time * 1.1 or latest_time - baseline_time < 3.5:
+            send_comment(
+                "Potential non-blocking slowdown", github_token, pull_request_number
+            )
```
mjambon, line 75: 1. I would include the numbers in the comment: baseline and latest times in seconds, percentage difference, reason why the warning is triggered (percentage or absolute difference?), and a summary of the protocol (what did we run and compare against?).
2. 10% slowdown seems reasonable for a warning. Maybe 5% is fine too, since it's only a comment.

If you're not sure, @emjin or other PA dev can do it later.

mjambon (approved):

scripts/compare_perf.py
brendongo: Yeah makes sense to be more clear in the error than just "there was a slowdown" i'll update

brendongo: Updated it to 7%.

For (1) this script doesn't actually have the information on what protocol is being compared. Probably fine for now as that change will need to update how the timing files are generated

PR #3258, opened 2021-06-02
fix: fall back to no optimizations when there are taint rules

brendongo:
PR checklist:
- [ ] changelog is up to date


aryx (approved):

aryx:
We should do that also for rules that use pattern-where-python (or --allow-arbitrary-code-to-run flag).

PR #3259, opened 2021-06-03
add: support equivalences in whole rule core running

brendongo:
PR checklist:
- [ ] changelog is up to date


brendongo:
@aryx not sure how much work it is to do this the right way in semgrep-core since it does have access to the entire rule. If it's a simple fix might be worth just doing that instead of this hack?

aryx (approved):
Yes, we could handle it directly in semgrep-core, because the equivalence is stored
in the rule, but at the same time we may also want at some point to apply a list
of general user-defined equivalences, kind of a list of builtin equivalences, so
does not hurt to also pass those equivalences around and have the -equivalences flag
also in whole-rule-in-ocaml.

Note that Matt suggested we might deprecate this feature, because it's not really used.

PR #3262, opened 2021-06-03
Clean up name handling

Sjord:
Properly return a `name` in `name` and `simple_name`.

PR checklist:
- [x] changelog is up to date

There are no functional changes, should I put something in the changelog?


aryx (approved):

aryx:
thx @Sjord ! No need to update changelog in those cases where it's just internal refactoring/cleaning.

PR #3266, opened 2021-06-04
Handle regexp parse errors gracefully

emjin:
Create an exception for regexp parse errors with good error messages and handle it python side

Test plan:

In semgrep/semgrep/tests/e2e, run semgrep --config rules/regex-invalid.yaml --optimizations

You should see

running 1 rules...
Invalid regexp in rule: '(abc': missing ) at position 4

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/parsing/Parse_mini_rule.ml
+++ b/semgrep-core/src/parsing/Parse_mini_rule.ml
@@ -28,6 +28,9 @@ exception InvalidLanguageException of string * string
 exception InvalidPatternException of string * string * string * string
 
 (*e: exception [[Parse_rules.InvalidPatternException]] *)
+
+exception InvalidRegexpException of string * string
+
 (*s: exception [[Parse_rules.UnparsableYamlException]] *)
 exception UnparsableYamlException of string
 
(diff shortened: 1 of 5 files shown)


emjin:
Failures seem to be due to https://github.com/returntocorp/semgrep-rules/commit/1e444476e5b9a203069b3f7861e8cee0a49db9e4; merging https://github.com/returntocorp/semgrep/pull/3268 will fix

semgrep-core/src/parsing/Parse_mini_rule.ml
```diff
@@ -28,6 +28,9 @@ exception InvalidLanguageException of string * string
 exception InvalidPatternException of string * string * string * string
 
 (*e: exception [[Parse_rules.InvalidPatternException]] *)
+
+exception InvalidRegexpException of string * string
```
aryx, line 32: By the way, I don't think I introduced those exceptions, but there's no need to have them suffixed by Exception.
We know it's an exception :)
(It's fine in this PR because it's good to be consistent, but we should fix those later).

aryx (approved):
Great!

PR #3267, opened 2021-06-04
Take innermost matching container when using pattern-inside

emjin:
If you have

```
rules:
- id: test_rule_id
   patterns:
   - pattern-inside: |
       def $FUNC(...):
          ...
   - pattern: print($MSG)
```

and

```
class A():
    def B():
        def C():
            print("blah")
```

Expect `$FUNC` to bind `C` rather than `B`

Test plan:

semgrep --config rules/message_interpolation/multi-pattern-inside.yaml targets/message_interpolation/multi_pattern_inside_nested.py

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/evaluation.py
+++ b/semgrep/semgrep/evaluation.py
@@ -440,7 +440,16 @@ def create_output(
             )
             output.append(rule_match)
 
-    return sorted(output, key=lambda rule_match: rule_match._pattern_match.range.start)
+    # Reversing the output is a terrible hack. It is useful because running whole rule in semgrep-core
+    # will return the results backwards compared to running only the patterns in semgrep-core
+    #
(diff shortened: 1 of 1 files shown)


emjin:
Merging https://github.com/returntocorp/semgrep/pull/3268 will fix the error

emjin:
This actually fixes the behavior of tests/e2e/test_check.py::test_nested_patterns_rule and tests/e2e/test_check.py::test_nested_patterns_either_rule, so I would prefer to just update the snapshots. That being said, this is still a terrible hack

aryx (approved):
I don't really understand it ... but let's merge this if this fixes mismatches.

IagoAbal:
This broke the tests, I just regenerated the oracles, please check if the new oracles make sense.

PR #3274, opened 2021-06-04
Use map_opt for running functions on options

Sjord:
Run f(x) for Some(x), and return None for None. We can't use Common.map_opt,
since we have to pass an extra parameter `env` everywhere.

Technical change, so no changelog needed.


aryx:
I'll accept this because the code is shorter and better, but there is no need to define your own
map_opt; you can reuse Common.map_opt because you can use "partial evaluation" in OCaml meaning
that you can do calls like
```
 let res = Common.map_opt (expression env) some_expr_opt in
```

(expression env) is expression "partially evaluated" and passed env.

PR #3283, opened 2021-06-06
Support unsafe block in C#

Sjord:
E.g.
```
unsafe { *i = *i + 1 }
```

PR checklist:
- [ ] changelog is up to date


aryx (approved):

IagoAbal:
I think we should prefer tests that match things inside the constructs. Parsing is kind of a "solved" issue here, in the sense that we don't crash anymore when a construct is not yet properly mapped to Generic (we instead insert a "TODO" node).

PR #3288, opened 2021-06-07
[php] parsing: Convert stmt-expr patterns into pure expr patterns

### LINKED ISSUE #3191: echo not supported as pattern-sinks in PHP taint mode
**Describe the bug**
echo is not supported as a `pattern-sinks` in taint mode in PHP

**To Reproduce**
https://semgrep.dev/s/DJWq

**Expected behavior**
Line 6 and 9 in the playground linked above should match and not result in `Semgrep exited with an unspecified error`

P0: blocking your adoption of Semgrep or workflow

IagoAbal:
Closes #3191

test plan:
make test # test included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/tests/tainting_rules/php/echo.yaml
+++ b/semgrep-core/tests/tainting_rules/php/echo.yaml
@@ -1,0 +1,12 @@
+rules:
+  - id: test-taint-echo
+    languages:
+      - php
+    message: Working!
+    pattern-sources:
+      - $_GET[...]
+    pattern-sanitizers:
+      - htmlspecialchars(...)
+    pattern-sinks:
+      - echo ...;
+    severity: ERROR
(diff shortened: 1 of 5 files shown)


IagoAbal:
@aryx I did as you suggested. IMO if this is something we want to do in general, then perhaps it should be a Generic-to-Generic transformation in `Parse_pattern.any`.

semgrep-core/tests/tainting_rules/php/echo.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - id: test-taint-echo
+    languages:
+      - php
+    message: Working!
+    pattern-sources:
+      - $_GET[...]
+    pattern-sanitizers:
+      - htmlspecialchars(...)
+    pattern-sinks:
+      - echo(...);
```
aryx: I think we could write echo ..., without the parenthesis.

aryx (approved):
Yes maybe we could generalize to other languages; let's keep it just for PHP for now.

PR #3291, opened 2021-06-07
Improve implementation of taking innermost container

emjin:
Better solution than https://github.com/returntocorp/semgrep/pull/3267

PR checklist:
- [ ] changelog is up to date


brendongo:
Nice! Much cleaner than the last solution. We can revisit whether we actually want both results surfaced but for now having feature parity between optimization flags is nice 👍

brendongo (approved):

PR #3292, opened 2021-06-07
Add support for metavariable-pattern in Semgrep

### LINKED ISSUE #3257: Metavariable patterns

Let's say I'm looking for calls to the following functions:

- `sqlx.Get($_X, $_Y, $QUERY, ...)`
- `$OBJ.Get($_X, $QUERY, ...)`

I'd like to be able to match cases where `$QUERY` is an expression like:
- `$X + $Y`
- `fmt.Sprintf("...", $X)`
- etc

It'd be nice to be able to define patterns that `$QUERY` must match in a concise way.

It's possible to work around this by just enumerating all the cases - for the example above, we'd have:

- `sqlx.Get($_X, $_Y, $X + $Y, ...)`
- `sqlx.Get($_X, $_Y, fmt.Sprintf("...", $X), ...)`
- `$OBJ.Get($_X, $X + $Y, ...)`
- `$OBJ.Get($_X, fmt.Sprintf("...", $X), ...), ...)`

... but this gets unwieldy as more clauses are introduced.

IagoAbal:
NOTE: This is only available via semgrep-core (requires --optimizations)!

Closes #3257

test plan:
semgrep --config semgrep-core/tests/OTHER/rules/metavar_pattern_nested.yaml \
    semgrep-core/tests/OTHER/rules/metavar_pattern_nested.py
    #^ works!

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/rule_schema.yaml
+++ b/semgrep/semgrep/rule_schema.yaml
@@ -16,6 +16,7 @@ definitions:
       - $ref: '#/definitions/pattern-not-regex'
       - $ref: '#/definitions/pattern-where-python'
       - $ref: '#/definitions/metavariable-regex'
+      - $ref: '#/definitions/metavariable-pattern'
       - $ref: '#/definitions/metavariable-comparison'
   pattern-either-content:
     type: array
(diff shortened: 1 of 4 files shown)


semgrep/semgrep/rule_schema.yaml
```diff
@@ -49,6 +50,37 @@ definitions:
     required:
     - metavariable-regex
     additionalProperties: false
+  metavariable-pattern:
```
aryx, line 53: Does this accept also just pattern: like
```
metavariable-pattern:
  metavariable: $OPTS
  pattern: foo(...)
```
?

aryx (approved):
awesome

semgrep/semgrep/rule_schema.yaml
IagoAbal: In semgrep-core it does because I just reuse the same parsing function as for top-level formulas. However in Python I allowed only `patterns` and `pattern-either` to keep it short. I imagined these were the most useful cases. Now that you are asking for it and I give it another thought, I do see useful cases for `pattern`. Let's add it.

PR #3301, opened 2021-06-07
Fixups and test for running perf benchmarks via docker

bpkroth:
A couple of minor fixups and a test for `run-benchmarks` with `--docker`


minusworld:
Thanks @bpkroth !

@brendongo Do you think this would slow down CI significantly?

brendongo:
Are we tracking benchmarks from running in docker right now @minusworld @colleend? The code changes in perf/run-benchmarks lgtm but not sure if actually running the perf benchmarks with --docker is data we want?

minusworld:
I don't need the benchmarking data in docker specifically, but it seems like @bpkroth might be interested in the data. @bpkroth Could you shed some light on your interested in docker benchmarks? Thanks!

bpkroth:
> I don't need the benchmarking data in docker specifically, but it seems like @bpkroth might be interested in the data. @bpkroth Could you shed some light on your interested in docker benchmarks? Thanks!

Mostly I was just looking for a convenient way to execute the benchmarks without installing all of the rest of the semgrep deps in my VM.  I didn't think you all needed/wanted the results `--upload`ed, though (seemed duplicative), but having a test that the tests were still capable of being run through docker seemed useful.

Only trouble I haven't had a chance to track down yet was the failure to parse the json output.

brendongo:
>  but having a test that the tests were still capable of being run through docker seemed useful.

Oh makes sense. Seems fine to run it without uploading to make sure it doesnt break.

colleend:
bye stalebot

aryx:
@brendongo what's the status of this? Can we merge it?

brendongo:
There's a failing test that @bpkroth was going to look into

bpkroth:
Yes, sorry, been OOF.  I will try and look at this a bit next week.

Cheers,
Brian

bpkroth:
CI tests running now.  Passing `--quiet` fixes it locally for me at least so that output skips some of these leading and trailing info logging lines and can be processed by the json parser inside the perf tests for validation:

```
running 1 rules...
{"errors": [], "results": [{"check_id": "rules.exec", "end": {"col": 16, "line": 1}, "extra": {"is_ignored": false, "lines": "exec(\"fortune\");", "message": "call to exec()", "metadata": {}, "metavars": {}, "severity": "ERROR"}, "path": "/targets/hello.js", "start": {"col": 1, "line": 1}}], "time": {"rule_parse_info": [0.0008699893951416016], "rules": [{"id": "rules.exec"}], "targets": [{"match_times": [0.0], "num_bytes": 23, "parse_times": [0.0], "path": "/targets/malformed.js", "run_times": [0.0003001689910888672]}, {"match_times": [9.083747863769531e-05], "num_bytes": 17, "parse_times": [0.0025730133056640625], "path": "/targets/hello.js", "run_times": [0.002977132797241211]}], "total_bytes": 40, "total_time": 0.11453557014465332}}
ran 1 rules on 2 files: 1 findings
```

brendongo (approved):

perf/run-benchmarks
```diff
@@ -223,12 +223,18 @@ def run_semgrep(
         args = [
             "docker",
             "run",
+            "--rm", # cleanup after ourselves
```
brendongo:
```suggestion
            "--rm",  # cleanup after ourselves
```

brendongo: To fix linting issue

brendongo:
Thanks for the contribution @bpkroth sorry took so long to get it merged!

PR #3302, opened 2021-06-07
ci: run public_repo_test in separate job

brendongo:
This PR separates out the public_repo tests from other unit/e2e tests
so those tests can return faster.

This PR also uncomplicates the pytest fixture hacks by just making things functions and
explicitly adding the parametrization instead of fancy pytest things.


aryx (approved):
I don't really understand what's going on here ... maybe @mschwager can better review. But if it helps getting faster in CI yes let's merge this!

.github/workflows/tests.yml
```diff
@@ -154,6 +145,51 @@ jobs:
...
+          path: ~/.cache/semgrep-cache
+          key: semgrep-repo-cache-test
+      - name: create cache dir if not exists
+        run: mkdir -p ~/.cache/semgrep-cache
+      - run: touch ~/.cache/semgrep-cache/ok
+      - name: Test semgrep
+        run: |
+          cd semgrep
+          export PATH=/github/home/.local/bin:$PATH
+          pipenv run pytest -n auto -v --tb=short --durations=0 tests/qa/run_public_repos.py
```
mschwager: Hmm, I like this change! The less we're relying on magic Makefile commands and non-standard flags the better, in my opinion. However, I think this behavior is a bit non-obvious at first. It took me a minute to realize why `pytest ... tests/` wouldn't also run this test (TL;DR because the filename isn't prefixed with `test_`). Personally, if I run `pytest tests/` I would expect this to run every Python test we have.

Here's my thought: would it make more sense to keep conventional naming here with `test_public_repos.py`? We could then do something like `pytest --ignore=tests/qa/test_public_repos.py ... tests/` above, and `pytest ... tests/qa/test_public_repos.py` here. I think that would be most explicit about what's going on. Thoughts?

brendongo: Yeah that does seem more obvious. I'll make that change before merging

PR #3303, opened 2021-06-07
ci: wait 10 min after pypi to publish to homebrew

brendongo:
brew bump-formula-pr depends on pipgrip to update dependencies
and pipgrip seems to not work right after publishing to pypi but seems
to work fine when run a few minutes later. Suspect it is pypi propagating
the latest release. Add a CI job that waits 10 minutes before trying to
publish to homebrew.

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -286,9 +286,18 @@ jobs:
           password: ${{ secrets.pypi_upload_token }}
           skip_existing: true
 
+  sleep-before-homebrew:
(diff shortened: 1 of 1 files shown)


adamberman (approved):

brendongo:
Note: during the sleep we are indeed going to have to pay for the CI minutes spent waiting. We only do this once per release so seems fine considering we run > 30 minutes of tests per PR

PR #3317, opened 2021-06-09
Fix wrong line/column reported on multi-line match in generic

### LINKED ISSUE #3315: Unexpected reported locations in 0.55.0
**Describe the bug**
A code change resulted in different locations reported by Semgrep in 0.55.0. Need to investigate why, as this is not expected.

See https://github.com/returntocorp/semgrep/pull/3312.

https://github.com/returntocorp/semgrep-app/pull/2675/files/208152566bbeacb0e0ac88dd13f308c23fca2fcd#r647864400 is also relevant.

**Environment**

aryx:
This closes https://github.com/returntocorp/semgrep/issues/3315

test plan:
test file included
Also below the line for "end" is now 9, and not anymore 4.
```
semgrep --lang generic -e "$(cat ~/yy/tests/generic/metavar_equality_var.sgrep)" ~/yy/tests/generic/metavar_equality_var.generic  --json | jq
{
  "errors": [],
  "results": [
    {
      "check_id": "-",
      "end": {
        "col": 21,
        "line": 9
      },
```

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -399,7 +399,7 @@ type ('target_content, 'xpattern) xpattern_matcher = {
     'target_content ->
     filename ->
     'xpattern ->
-    (Parse_info.token_location * MV.bindings) list;
+    (Parse_info.token_location * Parse_info.token_location * MV.bindings) list;
 }
 
 (* this will be adjusted later in range_to_pattern_match_adjusted *)
@@ -430,15 +430,15 @@ let (matches_of_matcher :
               |> List.map (fun (xpat, id, pstr) ->
                      let xs = matcher.matcher target_content file xpat in
                      xs
-                     |> List.map (fun (loc, env) ->
+                     |> List.map (fun (loc1, loc2, env) ->
                             (* this will be adjusted later *)
                             let rule_id = fake_rule_id (id, pstr) in
                             {
                               PM.rule_id;
                               file;
-                              range_loc = (loc, loc);
+                              range_loc = (loc1, loc2);
                               env;
-                              tokens = lazy [ info_of_token_location loc ];
+                              tokens = lazy [ info_of_token_location loc1 ];
                             }))
               |> List.flatten)
         in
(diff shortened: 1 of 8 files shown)


semgrep-core/src/engine/Match_rules.ml
```diff
@@ -399,7 +399,7 @@ type ('target_content, 'xpattern) xpattern_matcher = {
     'target_content ->
     filename ->
     'xpattern ->
-    (Parse_info.token_location * MV.bindings) list;
+    (Parse_info.token_location * Parse_info.token_location * MV.bindings) list;
```
IagoAbal: Maybe a comment explaining why this is needed?

IagoAbal: I mean why we need two `token_location` instead of one.

IagoAbal (approved):

semgrep-core/src/engine/Match_rules.ml
aryx: yep, will do.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -430,15 +430,15 @@ let (matches_of_matcher :
...
                      xs
-                     |> List.map (fun (loc, env) ->
+                     |> List.map (fun (loc1, loc2, env) ->
                             (* this will be adjusted later *)
                             let rule_id = fake_rule_id (id, pstr) in
                             {
                               PM.rule_id;
                               file;
-                              range_loc = (loc, loc);
+                              range_loc = (loc1, loc2);
```
IagoAbal, line 444: From the changes to the Comby and Regex matchers it seems that we could still get a single `loc` from `matcher` here, and instead `loc2` could be computed from that `loc`. But I don't know what `pos2` is in `let (pos1, _), (_, pos2) = m.Spacegrep.Match.region in` in the Spacegrep matcher. If `pos2` cannot be computed from `pos1` and `m.Spacegrep.Match.capture.value` then I guess that is the reason why we need both `loc1` and `loc2`.

aryx: It's because in Regex and Comby I use line_col_of_charpos, so I just need the starting charpos and the matched string.
But this line_col_of_charpos has a cost, because it parses the file and compute the line x col for each charpos.
Spacegrep actually already has this information, which is why I reuse it instead of calling line_col_of_charpos there.
Actually Comby can also return a line x col, which would again avoid calls to line_col_of_charpos.

PR #3325, opened 2021-06-09
Sort JSON output in generate_test_matrix.py

mschwager:


IagoAbal:
@aryx @emjin @mjambon I think this is the way to go, unless we want to use a diff tool that understands JSON. Otherwise the generated `test-matrix.json` can be randomly different due to a different sorting of the keys (despite equivalent). If this didn't happen before then I guess we were kind of lucky?

However, once we merge this, the resulting `test-matrix.json` will differ from the one in semgrep-app/frontend/src/assets/cheatsheet.json. Perhaps we want to wait for v0.55.0 before we merge this.

aryx (approved):

PR #3326, opened 2021-06-09
[php] Only coerce stmt-expr patterns into expr patterns in taint mode

IagoAbal:
PR #3288 messed up with the column numbers reported for matches in PHP
(where e.g. the ';' was no longer included in the match). We thus revert
PR #3288 (partially) and only do this transformation where we need it,
which is in taint-mode.

Fixes: 9ec7768caf1 ("[php] parsing: Convert stmt-expr patterns into pure expr patterns (#3288)")

PR checklist:
- [x] changelog is up to date


IagoAbal:
I generated the test-matrix.json locally and compared the results for the PHP tests, and it's all good.

emjin (approved):

PR #3331, opened 2021-06-10
Use logger level as single source of truth instead of verbosity globals

brendongo:
Before this commit we were using DEBUG and QUIET globals to decide when
things should or should not be logged (we cannot rely on just the logging
framework's handlers because tqdm uses prints). This PR uses whatever logging
level is stored by the root logger as the source of truth instead of relying
on the globals.

PR checklist:
- [ ] changelog is up to date


mschwager (approved):
Noice! Now for this pesky `FORCE_COLOR` at some point ;)

PR #3332, opened 2021-06-10
matching: Add support for associative-commutative (AC) matching

### LINKED ISSUE #3198: b && c does not match a && b && c
**Describe the bug**
In PHP, `pattern: "!check1($VAR) && !check2($VAR)"` or even `pattern: <... !check1($VAR) && !check2($VAR) ...>` do not match something like `if (!check3($stuff) && !check1($stuff) && !check2($stuff)) {`, but matches `if (!check1($stuff) && !check2($stuff) && !check3($stuff)) {`

Don't know if it's a regression or not, but I am pretty sure it was working in at least a previous version of semgrep

**To Reproduce**
https://semgrep.dev/s/1n75

**Expected behavior**
Would expect both pattern to match the expression where `check3()` is before the actual statement to match

P1: important to fix or quite annoying

IagoAbal:
Closes #3198

test plan:
make test # tests included

PR checklist:
- [x] changelog is up to date


IagoAbal:
Let me note that `&&` and `||` are not really commutative, e.g. `x > 0 && 2/x > 1`, maybe we could think how to perform associative-only matching which may also play nicer with other features like Autofix?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1241,6 +1244,80 @@ and m_other_argument_operator = m_other_xxx
...
+  | Some aargs_ac, Some bargs_ac when aop =*= bop ->
+      if_config
+        (fun x -> x.ac_matching)
+        ~then_:(m_ac_op tokb aop aargs_ac bargs_ac)
+        ~else_:m_op_default
+  | ___else___ -> m_op_default
+
+(* Associative-Commutative (AC) matching of operators.
+ *
+ * This for example, will successfully match `a && b && c` against `b && a && c`!
```
aryx, line 1269: maybe you can add that "
this will also let match `if (<... b && c ...>) `  against `if(a && b && c)` (which internally was parsed
 as `if ((a && b) && c)` which was not matching before.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1241,6 +1244,80 @@ and m_other_argument_operator = m_other_xxx
...
+      H.ac_matching_nf bop (B.unbracket bargs) )
+  with
+  | Some aargs_ac, Some bargs_ac when aop =*= bop ->
+      if_config
+        (fun x -> x.ac_matching)
+        ~then_:(m_ac_op tokb aop aargs_ac bargs_ac)
+        ~else_:m_op_default
+  | ___else___ -> m_op_default
+
+(* Associative-Commutative (AC) matching of operators.
```
aryx, line 1267: Great comment!

semgrep-core/src/matching/Matching_generic.ml
```diff
@@ -575,6 +575,79 @@ let rec m_list_in_any_order ~less_is_ok f xsa xsb =
...
+
+let m_combs_fold m_combs xs comb_matches : (_ list * tout) list =
+  List.fold_left
+    (fun combs x -> m_combs_bind combs (fun ys' tin -> m_combs x ys' tin))
+    comb_matches xs
+
+(* Tries matching `a` against each `b_i` in `bs`, and for each succesful match
+ * returns `(bs \ b_i, m a b_i)`. This is essentially the combinatorial version
+ * of `or_list`. *)
+let m_combs_or (m : _ matcher) a bs tin : (_ list * tout) list =
```
aryx, line 624: hmm, we should probably use that in the current implementation of all_elem_and_rest_of_list?

aryx (commented):
Very nice.
Did you run our benchmark with and without? Did you see any noticeable regressions?
(we don't have that many rules using A and B, so probably no difference)

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1241,6 +1244,80 @@ and m_other_argument_operator = m_other_xxx
...
+ * implementation with lots of dumb combinatorial search. If it
+ * becomes a perf issue, see paper
+ *
+ *     "Associative-Commutative Matching Via Bipartite Graph Matching"
+ *
+ * by Steven M. Eker.
+ *)
+and m_ac_op tok op aargs_ac bargs_ac =
+  (* partitition aargs_ac to separate metavariables and ellipsis (avars) from
+   * other kinds of expressions (aapps) *)
```
aryx, line 1293: what does apps in aapps mean here?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1241,6 +1244,80 @@ and m_other_argument_operator = m_other_xxx
...
+         | ___else___ -> false)
+  in
+  (* try to match each aapp to a different barg, this is a 1-to-1 matching *)
+  let bs_left =
+    match aapps with
+    (* if there are no aapps we don't want to fail *)
+    | [] -> m_combs_unit bargs_ac
+    | _ -> m_combs_1to1 m_expr aapps bargs_ac
+  in
+  (* try to natch each varible to a different subset of the remaining bs_left,
```
aryx: typo, match not natch, and variable, not varible.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1241,6 +1244,80 @@ and m_other_argument_operator = m_other_xxx
...
+   * ellipsis (`...`) can match an empty subset. *)
+  match avars with
+  | [] -> m_combs_flatten bs_left
+  | _ ->
+      let m_var x bs' =
+        match (x, H.undo_ac_matching_nf tok op bs') with
+        | A.Ellipsis _, None -> return ()
+        | ___var___, None -> fail ()
```
aryx: `___mvar___` maybe?

aryx (approved):
Very very nice!
I have to say I don't fully understand all the m_combs_xxx; each of them requires
some brain power, but they look good.

semgrep-core/src/matching/Generic_vs_generic.ml
IagoAbal: It stands for "application", meaning that these expressions have the shape `f args` where `f` is some constructor/operation (different from `op`), so we can do 1-to-1 matching with them. I could also have called it `non_vars` perhaps. Do you have a preference?

Picking the right names here was almost as difficult as writing the code itself. In particular when it comes to the `m_combs_*` operations.

semgrep-core/src/matching/Generic_vs_generic.ml
IagoAbal: Better indeed!

semgrep-core/src/matching/Matching_generic.ml
IagoAbal: Yes, I can do the refactoring!

IagoAbal:
> Very nice.
> Did you run our benchmark with and without? Did you see any noticeable regressions?
> (we don't have that many rules using A and B, so probably no difference)

I just did, with `perf/run-benchmarks --std-only`, and I saw no meaningful differences. But for sure this is going to be slower matching `&&` and `||` patterns that involve many variables. @emjin do you want to suggest some benchmark suite that I should run?

emjin:
We probably don't have a great benchmark to test this specific case, but you could try creating a couple of rules you suspect might be slower and running them against large files.

IagoAbal:
I found one rule in semgrep-rules that uses this https://github.com/returntocorp/semgrep-rules/blob/develop/python/sqlalchemy/correctness/bad-operator-in-filter.yaml:
```
  patterns:
  - pattern-inside: |
      def $ANY(...):
          ...
          $MODEL.query
  - ...
  - pattern-either:
    - ...
    - pattern: $A and $B
    - pattern: $A or $B
    - ...
```
But I will "stress test" it a bit more.

aryx (approved):
looks good, let's merge this because I need to modify Config_semgrep.ml so I don't want conflicts :)

IagoAbal:
OK I'll open a new PR.

PR #3337, opened 2021-06-11
Add setup for Hack

mjambon:
It compiles and I think all the files are there.
* uses the existing Ast_php AST
* uses tree-sitter to parse patterns as well as targets

I'll let Nicholas figure out how to get to a point where "hello world" parses.

Bizarrely, at the moment semgrep-core exits successfully silently:
```
$ ~/semgrep/semgrep-core $ ./bin/semgrep-core -lang hack -e 'x' <(echo -e "<?hh\necho 'Hello, World';")
```
but it should fail to parse the pattern and the target. I don't think anything gets parsed. Anyway, I'll let Nicholas figure this out.

PR checklist:
- [x] changelog is up to date

Depends on https://github.com/returntocorp/pfff/pull/438

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -43,3 +43,6 @@
 [submodule "semgrep-core/src/tree-sitter-lang/semgrep-ocaml"]
 	path = semgrep-core/src/tree-sitter-lang/semgrep-ocaml
 	url = https://github.com/returntocorp/semgrep-ocaml.git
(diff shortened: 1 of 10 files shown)


aryx (approved):
looks good.
I can move ast_php.ml at some point in semgrep-core/, to ease up development.

PR #3338, opened 2021-06-11
Allow users to disable certain matching features in a rule

aryx:
This will allow to disable for example constant propagation
on a per-rule basis.

I've chosen the 'settings' keyword, but we could change it
for 'preferences', or 'config', whatever.

Still need to update the jsonschema after so it can be used
from the semgrep python wrapper.

test plan:
test file included

PR checklist:
- [ ] changelog is up to date


IagoAbal (approved):
Great! I was looking forward to seeing this implemented.

PR #3339, opened 2021-06-11
matching: Restrict AC matching to avoid a combinatorial explosion

IagoAbal:
Plus, rethink/refactor `m_comb_*` operators.

Fixes: 7d36fc50af4 ("matching: Add support for associative-commutative (AC) matching (#3332)")

test plan:
make test # tests included

PR checklist:
- changelog update not needed


semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1267,56 +1267,99 @@ and m_call_op aop aargs bop tokb bargs =
 (* Associative-Commutative (AC) matching of operators.
  *
  * This for example, will successfully match `a && b && c` against `b && a && c`!
- * It will also match `if (<... b && c ...>) ...` against `if (a && b && c) ...`
+ * It will also match `if (<... b && c ...>) ...` against `if (a && b && c) S`
  * which was not matching before (`a && b && c` is parsed as `(a && b) && c`).
+ * In fact, we do not even need a deep-expr pattern, `if (b && c) ...` will also
```
aryx, line 1272: I'm not sure we want that. I thought when the final rest was non-null we were failing.

IagoAbal:
Now that we will have the `settings` key to enable/disable this kind of features, I think it's better to disable AC-matching by default. It's an useful feature but it can mess up things a bit due to the ranges. What do you think?

aryx:
I think we can enable it for now and wait for bug reports. If we disable it, nobody is gonna use that feature and we will not
get any feedback on it.

IagoAbal:
> I think we can enable it for now and wait for bug reports. If we disable it, nobody is gonna use that feature and we will not
> get any feedback on it.

That's right. I had the idea to add associative-only matching and make that the default. But it would be good indeed to get some user feedback on AC matching first.

IagoAbal:
OK to merge this @aryx?

aryx (approved):
I think I understood better the previous code, but it's late so maybe it's becaus I'm tired.
Let's merge this.

semgrep-core/src/matching/Generic_vs_generic.ml
IagoAbal: Let's say that we have pattern `<... a && b ...>`, and the code is `a && (b && c)`,  the sub-expressions of `a && (b && c)` are itself, `a`, `b && c`, `b`, and `c`. The only chance to match `a && b` is to do it on `a && (b && c)`, and for that we cannot fail when there are left-overs to match. We could of course try to move AC-matching to `SubAST_generic.subexprs_of_expr` by enumerating all partitions, but I think that would be more expensive. Right now AC matching should scale fairly well if there are no variables involved, but probably that won't be the case if implemented in `SubAST_generic`. Otherwise we can e.g. add an optional argument `?deep` to `m_expr` so that we know that we're inside a deep-expr. If we are not inside a deep-expr then we can fail when there are args left to match.

IagoAbal: Or we could restrict AC matching to `<... a && b ...>` (require deep-expr) or `a && b && ...` (require ending `...`) or something like that.

aryx: > we can e.g. add an optional argument ?deep to m_expr so that we know that we're inside a deep-expr. If we are not inside a deep-expr then we can fail when there are args left to match

maybe this. But we can keep it like that for now and see if we get false positives and people complain.

Maybe you can add a comment about this discussion in the code so we remember it.

PR #3340, opened 2021-06-11
core: tests: Add three more tests for metavariable-pattern

IagoAbal:
PR checklist:
- changelog update not needed

### DIFF
--- a/semgrep-core/tests/OTHER/rules/metavar_pattern_dots_mvar.py
+++ b/semgrep-core/tests/OTHER/rules/metavar_pattern_dots_mvar.py
@@ -1,0 +1,17 @@
+# ruleid:test-mvp-dots-mvar
+def f():
+    foo(1)
+    bar(2)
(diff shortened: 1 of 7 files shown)


emjin:
I feel like with metavariable-pattern there's very little reason to have equivalences---it's not hard to express equivalences as a pattern-either now. Especially true if we allowed something like

```
- metavariable-pattern:          
   metavariable: $REQUEST, $REQUEST2
   patterns:           
   - pattern-either:              
      - pattern: request.$W.get(...)     
      - pattern: request.$W[...]
```

aryx:
looks good. The failures are probably because semgrep-rules changed and we need to regenerate some snapshots.

IagoAbal:
Thanks for the hint I'll take a look!

IagoAbal:
Tests passing now 👍

aryx (approved):

PR #3342, opened 2021-06-11
Remove .ocp-indent files

### LINKED ISSUE #2884: Enforce consistent formatting by using ocamlformat instead of ocp-indent on our OCaml repos.
We need to:
* identify the configuration we need (preferably a standard, conservative style with minimal tweaks)
* check that we're getting satisfying results when running ocamlformat on our code
* use it instead of ocp-indent in the pre-commit hook

mschwager:
Fixes #2884.

Made possible by: https://github.com/returntocorp/semgrep/commit/be6984f4f5cf7ff57998f35eaaebb45445f95c74


mjambon:
ok, let's remove it. I'll use a global config for myself instead, [as mentioned here](https://github.com/returntocorp/semgrep/issues/2884#issuecomment-859933327).

mjambon (approved):

mschwager:
> ok, let's remove it. I'll use a global config for myself instead, as mentioned here.

I'm happy to abandon this PR if you'd like to keep these files around. I just thought they weren't used any more.

Are multiple people on the team using these configurations? If so, I'd say we should leave them here. If not, I'd vote for using a local configuration on your machine.

mjambon:
I'll let the team know what to do, in case they're using ocp-indent like me.

PR #3346, opened 2021-06-12
fix: use --debug flag to gate semgrep-core debug output 

brendongo:
Prior to this commit, semgrep-core output was only being printed out when semgrep
was called with `--debugging-json` but not when `--debug` was set. This PR makes
the output of debug info more consistent with the rest of our codebase.

Now `--debugging-json` is only used to add step-by-step pattern match evaluation
logs. Holding off on renaming for now since the OutputSettings class is potentially
used externally.

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -49,6 +49,7 @@ from semgrep.spacegrep import run_spacegrep
 from semgrep.target_manager import TargetManager
 from semgrep.target_manager_extensions import all_supported_languages
 from semgrep.util import debug_tqdm_write
+from semgrep.util import is_debug
 from semgrep.util import partition
 from semgrep.util import progress_bar
 from semgrep.util import SEMGREP_PATH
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/core_runner.py
```diff
@@ -257,16 +257,14 @@ def _run_core_command(
                 "-max_memory",
                 str(self._max_memory),
                 "-json_time",
+                "-debug",
```
brendongo: @mjambon instead of passing -debug conditionally and outputing debug logs if they exist, I switch to always passing debug logs and conditionally outputing. This assumes that there is no performance hit from using -debug in semgrep-core which I don't imagine there is but thought I would double check with you.

semgrep/semgrep/core_runner.py
```diff
@@ -283,13 +281,12 @@ def _extract_core_output(
...
         #
-        if semgrep_error_output != "":
-            name = f"[rule '{rule.id}']"
-            logger.info(
-                f"--- semgrep-core stderr {name} ---\n"
-                f"{semgrep_error_output}"
-                f"--- end semgrep-core stderr {name} ---"
-            )
+        name = f"[rule '{rule.id}']"
+        logger.debug(
```
brendongo, line 288: Logging handler in util.py handles whether or not a debug level log here gets printed to stderr

semgrep/semgrep/core_runner.py
mjambon: I would assume that passing `-debug` can make things significantly slower. Even it's not slow today, it can become slow in the future. I think you should keep it conditional, i.e. make it reflect the semgrep command line, requiring either `--debug` or `--debugging-json` so as to minimize surprises.

brendongo: Makes sense. I returned it to conditionally passing the -debug flag

mschwager (approved):

PR #3351, opened 2021-06-14
Support metavariable inside strings

aryx:
This allows `foo("$VAR")` to match `foo("a string")` where
the metavariable $VAR contains "a string" but without the
enclosing quotes.

See the poll here:
https://r2c-community.slack.com/archives/CK86BJ5DW/p1623223809187200
people seem to like the idea.

test plan:
test file included

PR checklist:
- [x] changelog is up to date


semgrep-core/src/core/Metavariable.ml
```diff
@@ -63,13 +66,16 @@ type mvalue =
...
   | P of AST_generic.pattern
+  | Text of string AST_generic.wrap
   (* Those can be now empty with $...XXX metavariables.
    * coupling: if you add more constructors that allow an empty content,
    * you may need to modify JSON_report.range_of_any to not get
    * some NoTokenLocation exn.
    *)
   | Ss of AST_generic.stmt list
   | Args of AST_generic.argument list
+(* This is to match the content of a string or atom, without the
```
IagoAbal, line 77: I guess this comment applies to `Text` but it is placed after `Args`.

semgrep-core/src/core/Metavariable.ml
```diff
@@ -89,18 +95,21 @@ let mvalue_to_any = function
...
+  match mval with
   | E expr -> Some [ G.exprstmt expr ]
   | S stmt -> Some [ stmt ]
   | Id (id, Some idinfo) -> Some [ G.exprstmt (G.N (G.Id (id, idinfo))) ]
   | Id (id, None) -> Some [ G.exprstmt (G.N (G.Id (id, G.empty_id_info ()))) ]
   | N x -> Some [ G.exprstmt (G.N x) ]
   | Ss stmts -> Some stmts
-  | Args _ | T _ | P _ ->
-      logger#debug "program_of_mvalue: Args | T | P are not yet handled!";
+  | Args _ | T _ | P _ | Text _ ->
```
IagoAbal, line 111: We can convert `Text` into a program:
```
Text s -> Some [ G.exprstmt (G.L (G.String (s, info))) ]
```
Maybe `program_of_mvalue` could be written in terms of `mvalue_to_any` since there is some duplication.

IagoAbal (approved):
Awesome!

semgrep-core/src/core/Metavariable.ml
aryx: yes, I usually put the comments before ...

semgrep-core/src/core/Metavariable.ml
aryx: Well in theory we match the raw text inside a string, so it's not a valid program in the host language. It could be a program in another language, in which case we would do it via the metavariable-pattern: language: we talked about.

semgrep-core/src/core/Metavariable.ml
aryx: oh something got messed up, Text should be after, ocamlformat did something wrong here (or me).

mjambon (approved):
oh this is nice!

ievans (approved):
This is great!

- Should there be tests that verify the bound metavariable doesn't have `"` quotes in it?
- Will there be a corresponding PR with update to the docs?
- What if I want to literally match `foo("$X")`? Is there an escape mechanism? Or do I just use a metavariable-regex on the binding?

aryx:
> This is great!
> 
> * Should there be tests that verify the bound metavariable doesn't have `"` quotes in it?

Yes we should.

> * Will there be a corresponding PR with update to the docs?

I'm adding support for foo("$VAR"), foo(:$VAR), and foo(/$VAR/) and after
that we can update the doc.

> * What if I want to literally match `foo("$X")`? Is there an escape mechanism? Or do I just use a metavariable-regex on the binding?

No escape mechanism. We have the same issue with foo("...").

PR #3352, opened 2021-06-14
[Go] naming: Recognize other common package naming conventions

### LINKED ISSUE #2424: go - equivalent naming import does not work
I was testing this registry rule `go.jwt-go.security.jwt-none-alg` and I noticed that it does not work when the Go import has an alias (like `foo`). I've made a live editor clone here https://semgrep.dev/s/fproulx-boostsecurity:go-equivalent-naming-import???registry=go.jwt-go.security.jwt-none-alg

@minusworld looked at this with me and if we modify the pattern to have a variable in the import, then it breaks the default behavior (https://semgrep.dev/s/ZveL/)

IagoAbal:
Closes #2424

test plan:
make test # tests included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 - C#: Parse attributes for local functions (#3348)
+- Go: Recognize other common package naming conventions (#2424)
 
 ## [0.55.1](https://github.com/returntocorp/semgrep/releases/tag/v0.55.1) - 2021-06-9
 
--- a/semgrep-core/src/analyzing/Naming_AST.ml
+++ b/semgrep-core/src/analyzing/Naming_AST.ml
@@ -582,7 +582,18 @@ let resolve2 lang prog =
           | ImportAs (_, FileName (s, tok), Some (alias, id_info)) ->
               (* for Go *)
               let sid = H.gensym () in
-              let base = (Filename.basename s, tok) in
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 - C#: Parse attributes for local functions (#3348)
+- Go: Recognize other common package naming conventions (#2424)
```
aryx, line 14: an example could be nice. Those changelog entries can be hard to understand and I guess some people just read that to check what is new.

aryx (approved):
Great!

semgrep-core/src/analyzing/Naming_AST.ml
```diff
@@ -582,7 +582,18 @@ let resolve2 lang prog =
           | ImportAs (_, FileName (s, tok), Some (alias, id_info)) ->
               (* for Go *)
               let sid = H.gensym () in
-              let base = (Filename.basename s, tok) in
+              let pkgname =
+                let pkgpath, pkgbase = Common2.dirs_and_base_of_file s in
+                if pkgbase =~ "v[0-9]+" then
```
aryx, line 587: when there's a long list of if elseif, I actually now tend to prefer
match () with
| _ when pkgbase =~ "v[0-9]+" -> ...
| _ when pkgbase =~ ... -> ...
| _ ->

aryx (approved):

PR #3354, opened 2021-06-14
Remove unnecessary/incorrect ruamel comment

mschwager:
After https://github.com/returntocorp/semgrep/pull/2883 this comment is no longer correct.

### DIFF
--- a/semgrep/setup.py
+++ b/semgrep/setup.py
@@ -101,9 +101,6 @@ setuptools.setup(
         "attrs>=19.3.0",
         "colorama>=0.4.3",
         "requests>=2.22.0",
-        # Pin exact version of 'ruamel.yaml' because of unstable API.
-        # Ensure default type is still 'rt' (round-trip) and that RoundTrip
-        # inherits from Safe when upgrading the version.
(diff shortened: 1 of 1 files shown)


semgrep/setup.py
```diff
@@ -101,9 +101,6 @@ def find_executable(env_name, exec_name):
         "attrs>=19.3.0",
         "colorama>=0.4.3",
         "requests>=2.22.0",
-        # Pin exact version of 'ruamel.yaml' because of unstable API.
```
mschwager, line 104: This part is no longer true.

semgrep/setup.py
```diff
@@ -101,9 +101,6 @@ def find_executable(env_name, exec_name):
...
-        # Ensure default type is still 'rt' (round-trip) and that RoundTrip
```
mschwager, line 105: This part is now guaranteed by the `test_default_yaml_type_safe` unit test, so we no longer need to manually ensure this.

mschwager:
Also building off of: https://github.com/returntocorp/semgrep/pull/3088.

aryx (approved):

PR #3363, opened 2021-06-15
Support metavariable for atoms in Ruby and ellipsis in regexp for JS

aryx:
This currently works, but both are specific to Ruby or JS. Next PR
will generalize and make it work for other languages by refactoring
the Regexp and Atom type in the generic AST

test plan:
test file included

PR checklist:
- [ ] changelog is up to date


semgrep-core/src/matching/Matching_generic.ml
```diff
@@ -754,6 +754,19 @@ let m_string_ellipsis_or_metavar_or_default ?(m_string_for_default = m_string) a
...
   | _ -> m_wrap m_string_for_default a b
 
+let m_atom_ellipsis_or_metavar_or_default ?(m_string_for_default = m_string) a b
+    =
+  match fst a with
+  | s when s =~ "^:\\(.*\\)" && MV.is_metavar_name (Common.matched1 s) ->
+      let mvar = Common.matched1 s in
+      let tok = snd a in
+      envf (mvar, tok) (MV.E (B.L (B.Atom b)))
+  | _ -> m_wrap m_string_for_default a b
```
IagoAbal, line 764: Will `:...` match `:atom` ?

semgrep-core/src/matching/Matching_generic.ml
```diff
@@ -754,6 +754,19 @@ let m_string_ellipsis_or_metavar_or_default ?(m_string_for_default = m_string) a
...
+
+let m_regexp_ellipsis_or_metavar_or_default ?(m_string_for_default = m_string) a
+    b =
+  match fst a with "/.../" -> return () | _ -> m_wrap m_string_for_default a b
```
IagoAbal, line 768: Will `/$X/` match `/abc/` and bind `$X` to `abc` ?

IagoAbal (approved):

semgrep-core/src/matching/Matching_generic.ml
aryx: not yet, but in the next diff yes.
This is just refactoring to prepare for next diff.

semgrep-core/src/matching/Matching_generic.ml
aryx: not yet but in next diff yes.

PR #3364, opened 2021-06-15
Convert __makeref, __reftype, __refvalue in C#

Sjord:
E.g.
```
int x = 3;
var xref = __makeref(x);
```

We convert `__reftype(x)` to `typeof(DeRef(x))`. Does that make sense?

These functions are very rarely used in real C# code.

PR checklist:
- [x] changelog is up to date


aryx (approved):
Sounds good.

PR #3366, opened 2021-06-15
metavariable-pattern: Allow matching content under another language

IagoAbal:
test plan:
make test # test included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -121,7 +121,7 @@ and metavar_cond =
    * the "regexpizer" optimizer (see Analyze_rule.ml).
    *)
   | CondRegexp of MV.mvar * regexp
-  | CondPattern of MV.mvar * formula
+  | CondPattern of MV.mvar * Lang.t option * formula
 [@@deriving show, eq]
 
 (*****************************************************************************)
@@ -149,7 +149,8 @@ type formula_old =
 (* extra conditions, usually on metavariable content *)
 and extra =
   | MetavarRegexp of MV.mvar * regexp
-  | MetavarPattern of MV.mvar * formula
+  (* TODO: Generalize `Lang.t option` to `xlang`. *)
+  | MetavarPattern of MV.mvar * Lang.t option * formula
   | MetavarComparison of metavariable_comparison
   | PatWherePython of string
 
@@ -241,7 +242,8 @@ let rewrite_metavar_comparison_strip mvar cond =
 let convert_extra x =
   match x with
   | MetavarRegexp (mvar, re) -> CondRegexp (mvar, re)
-  | MetavarPattern (mvar, formula) -> CondPattern (mvar, formula)
+  | MetavarPattern (mvar, opt_lang, formula) ->
+      CondPattern (mvar, opt_lang, formula)
   | MetavarComparison comp -> (
       match comp with
       (* do we care about strip and base? should not Eval_generic handle it?
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -798,40 +798,6 @@ and filter_ranges env xs cond =
              Eval_generic.eval_bool env e
          | R.CondPattern (mvar, opt_lang, formula) -> (
              (* If anything goes wrong the default is to filter out! *)
-             let lazy_ast_of_string lang str =
-               lazy
-                 (let ext =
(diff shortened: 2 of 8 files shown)


semgrep-core/src/core/Rule.ml
```diff
@@ -149,7 +149,7 @@ type formula_old =
 (* extra conditions, usually on metavariable content *)
 and extra =
   | MetavarRegexp of MV.mvar * regexp
-  | MetavarPattern of MV.mvar * formula
+  | MetavarPattern of MV.mvar * Lang.t option * formula
```
aryx, line 153: We may want xlang here at some point, to allow to use spacegrep, or even comby.

aryx: but for now let's start with just Lang.t

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -796,48 +796,90 @@ and filter_ranges env xs cond =
          | R.CondGeneric e ->
              let env = Eval_generic.bindings_to_env bindings in
              Eval_generic.eval_bool env e
-         | R.CondPattern (mvar, formula) -> (
+         | R.CondPattern (mvar, opt_lang, formula) -> (
              (* If anything goes wrong the default is to filter out! *)
+             let lazy_ast_of_string lang str =
```
aryx: I'm not a big fan of long nested functions. Maybe better to put that as an helper function out of filter_ranges.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -796,48 +796,90 @@ and filter_ranges env xs cond =
...
+                    | [] -> assert false
+                  in
+                  Common2.with_tmp_file ~str ~ext (fun file ->
+                      let { Parse_target.ast; errors; _ } =
+                        Parse_target
+                        .parse_and_resolve_name_use_pfff_or_treesitter lang file
+                      in
+                      (ast, errors)))
+             in
+             let go_eval_formula env xlang formula lazy_ast_and_errors
```
aryx: Same here, because it's difficult to see if it's use enclosed variables (which then are really like using globals).

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -796,48 +796,90 @@ and filter_ranges env xs cond =
...
+                     (* The matched text must be interpreted according to `lang'. *)
+                     let lazy_ast_and_errors =
+                       lazy_ast_of_string lang content
                      in
-                     let env =
-                       { env with pattern_matches = pattern_matches_per_id }
+                     go_eval_formula env env.xlang formula lazy_ast_and_errors
+                       (lazy content)
+                       None
+                 | Some lang, _ ->
```
aryx: We could probably restrict here to G.E (G.L (G.String ...)) or the stuff that spacegrep and comby returns via mval_of_string.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -318,15 +318,34 @@ and parse_extra env x =
...
       | [ ("metavariable", Some (J.String metavar)) ], rest ->
+          let opt_lang, rest' =
+            match find_fields [ "language" ] rest with
+            | [ ("language", Some (J.String s)) ], rest' ->
+                let lang =
+                  match Lang.lang_of_string_opt s with
+                  | None ->
+                      raise
+                        (E.InvalidLanguageException
+                           ("FIME id", spf "unsupported language: %s" s))
```
aryx: FIXME typo

aryx (approved):

semgrep-core/src/engine/Match_rules.ml
IagoAbal: I agree, I was being a bit lazy here. I was actually considering to move the logic for `CondPattern` to separate function. But I'll just pull out these two helpers for now.

semgrep-core/src/engine/Match_rules.ml
IagoAbal: I did, but note that `mval_of_string` tries to interpret the string as an integer and if possible it returns an integer literal instead of a string literal. So those cases we won't be able to match.

semgrep-core/src/engine/Match_rules.ml
IagoAbal: I ended up taking `CondPattern`s logic into a separate function as well.

PR #3374, opened 2021-06-16
Add coupling comment in rule_schema.yaml

aryx:
test plan:
wait for CI to complain if there is a problem

PR checklist:
- [x] changelog is up to date


IagoAbal:
In this particular case, wouldn't be easier if semgrep-app took the rule schema from semgrep-cli? Is there a reason to have two separate schemas?

aryx:
I agree, I asked for it in the linked PR above

PR #3375, opened 2021-06-16
Move spacegrep code to semgrep-core/src

mjambon:
The spacegrep library and the executable are now built in the same dune project as semgrep-core. Thanks to this, there's no longer a need to install the spacegrep library for semgrep-core to use it.

The spacegrep executable still needs to be shipped alongside semgrep-core, for now.

PR checklist:
- [x] changelog is up to date


aryx (approved):
looks good

Issue #3372, opened 2021-06-15
Ruby Regex Matching


colleend:
**Describe the bug**
Ruby regex matching does not match equating a string constant with a regex. 
In
https://semgrep.dev/s/KWj5/ 
the expectation is that the pattern will match both regex expressions.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

aryx:
Not sure if that's what the user wanted, but you can now use metavariable in regexps, and even ..., as in `foo(/.../)`

See https://semgrep.dev/s/grx0/?version=develop

aryx:
Or https://semgrep.dev/s/Q4rZ/?version=develop

PR #3376, opened 2021-06-16
Refactor Regexp construct and allow ellipsis and metavar in regexps

aryx:
Before, the regexp /.../ was represented differently in JS and Ruby
because we were using an imprecise type for the Regexp construct.
This PR fixes the AST_generic part, other PRs will adjust
then the type for ast_js.ml and ast_ruby.ml

test plan:
test file included

PR checklist:
- [x] changelog is up to date


aryx:
This should help https://github.com/returntocorp/semgrep/issues/3372

semgrep-core/src/parsing/pfff/js_to_generic.ml
```diff
@@ -218,9 +218,11 @@ and literal x =
   | String v1 ->
       let v1 = wrap string v1 in
       G.String v1
-  | Regexp v1 ->
+  | Regexp (v1, v2) ->
       let v1 = wrap string v1 in
-      G.Regexp v1
+      (* TODO: *)
```
IagoAbal, line 223: What is left to do exactly?

IagoAbal (approved):

semgrep-core/src/parsing/pfff/js_to_generic.ml
aryx: What is left is to change Ast_js.Regexp type to also include position information for the leading / and closing /
and the modifier, so we avoid those fake_bracket and fake_info below.
Otherwise we can match regexps, but the matched position might be wrong if for example
we do /$XXX/g as a pattern, right now $XXX will be bounded to a Parse_info.t that include the whole
range of /abc/g, not just abc.

PR #3382, opened 2021-06-17
Move corpuses from large to medium

emjin:
Several corpuses that used to take too long to run for medium no longer do, so they can be moved

PR checklist:
- [x] changelog is up to date


colleend (approved):
lgtm!

PR #3383, opened 2021-06-17
Bind pattern-regexp capture groups to metavariables

### LINKED ISSUE #3356: bind pattern-regex capture groups to metavariables
I'd like to propose we extend pattern-regex capture group support to bind the capture groups into metavariables like $1, $2, etc. that could be referenced in the message.

My use case: https://semgrep.dev/s/O8lL/. I’d like $1 and $2 to be bound and usable in the output message.

aryx:
This closes #3356

test plan:
```
$ semgrep-core -lang generic -config
tests/OTHER/rules/regexp_capture_groups.yaml
tests/OTHER/rules/regexp_capture_groups.gem -json |jq
...
      "metavars": {
          "$1": {
            "start": {
              "line": 5,
              "col": 6,
              "offset": 91
            },
            "end": {
              "line": 5,
              "col": 26,
              "offset": 111
            },
            "abstract_content": "google-cloud-storage",
            "unique_id": {
              "type": "AST",
              "md5sum": "f57552512d4701f50efed06c43308f6e"
            }
```

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -44,25 +44,30 @@ let debug_matches = ref false
  *  - visiting code (=~ Match_patterns.ml)
  *  - matching code (=~ Generic_vs_generic.ml)
  *
- * There are also "preprocessing" work before that:
+ * There are also "preprocessing" work before:
  *  - parsing (lexing, parsing) rules, code, patterns
  *  - normalizing (convert to a generic AST)
  *  - naming (but bugs probably)
  *  - SEMI typing (propagating type decls at least and small inference)
  *  - SEMI analyzing (dataflow constant propagation)
- *    but could do much more: deep static analysis using Datalog?
+ *    but we could do much more: deep static analysis using Datalog?
  *
  * TODO
  *  - associate the metavariable-regexp to the appropriate pattern
+ *    to get the right scope (right now we accept the wrong scope but
+ *    this forces to extend some ranges with metavars from other ranges)
  *  - pattern-where-python? use pycaml? works for dlint rule?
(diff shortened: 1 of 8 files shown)


semgrep-core/src/engine/Match_rules.ml
```diff
@@ -557,8 +560,27 @@ let regexp_matcher big_str file (_s, re) =
          let line, column = line_col_of_charpos file charpos in
          let loc2 = { PI.str; charpos; file; line; column } in
 
-         (* TODO? return regexp binded group? $1 $2 etc? *)
-         let env = [] in
+         (* return regexp binded group $1 $2 etc *)
+         let n = Pcre.num_of_subs sub in
+         let env =
+           match n with
+           | 1 -> []
```
IagoAbal, line 567: Should this be `0 -> []` ?

IagoAbal (approved):

semgrep-core/tests/OTHER/rules/regexp_capture_groups.yaml
```diff
@@ -0,0 +1,7 @@
+rules:
+  - id: new-Gemfile-dependency
+    languages:
+      - generic
+    message: New dependency with name $1 and version $2
+    pattern-regex: gem "(.*)", "(.*)"
```
IagoAbal, line 6: Would be cool if one could use `metavariable-pattern` on `$1` etc, but I'm not sure the `$1` will parse as a valid metavariable name? It needs to start with a letter.

aryx: hmm right. Not sure we check that though when we parse metavariable-pattern.

PR #3384, opened 2021-06-17
Update port-language.md

Nicholas-Lin:
Original path of `parsing/dune` is the incorrect dune file. The correct file is at `parsing/tree_sitter/dune`.


emjin:
This is probably failing because you didn't set up authentication. If you click on "Details" in the failing test, it'll point you to instructions for what you need to do.

brendongo:
@emjin this is cause github doesn't pass in secrets for PRs coming from forks. If this change looks good to you we can just merge it

emjin (approved):

PR #3386, opened 2021-06-18
Add advice on filling out CST-to-AST boilerplate

mjambon:
I also applied this advice in one new spot.

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_typescript_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_typescript_tree_sitter.ml
@@ -196,10 +196,8 @@ let rec id_or_nested_id (env : env) (x : CST.anon_choice_type_id_42c0412) :
   | `Id tok -> [ identifier env tok ] (* identifier *)
   | `Nested_id x -> nested_identifier env x
 
-and anon_choice_type_id_42c0412 env x = id_or_nested_id env x
-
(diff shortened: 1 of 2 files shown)


emjin (approved):

semgrep-core/src/parsing/tree_sitter/Parse_typescript_tree_sitter.ml
```diff
@@ -196,10 +196,8 @@ let rec id_or_nested_id (env : env) (x : CST.anon_choice_type_id_42c0412) :
   | `Id tok -> [ identifier env tok ] (* identifier *)
   | `Nested_id x -> nested_identifier env x
 
-and anon_choice_type_id_42c0412 env x = id_or_nested_id env x
```
emjin, line 199: Why was this originally aliased to anon_choice_type_id_42c0412? I would have expected the opposite (we needed to alias the autogenerated name to something better)

mjambon: This was part of my original attempts to improve the situation with these automatic function names. First I thought it would be good to keep the original function around, so I did that for some of the code, until I realized it wasn't useful. So now I just replace the generated function name by a nicer name. The original name is still there, in the type of the argument (`(x : CST.anon_choice_type_id_42c0412)`), which is all we need.

PR #3403, opened 2021-06-22
Merge parsing stats

mjambon:
Same as https://github.com/returntocorp/semgrep/pull/3391 but has a different branch name so it won't try and fail to run the parsing stats. There's a chicken-and-egg issue: we need to merge the code into `develop` so as to have a docker image with the correct semgrep-core in it first. Then we can run the stats: they will run daily on `develop` and with every push on the `parsing-stats` branch.

PR checklist:
- [x] changelog is up to date


.circleci/config.yml
```diff
@@ -81,3 +93,32 @@ workflows:
...
+  scheduled-parsing-stats:
+    triggers:
+      - schedule:
+          # Run at 00:00 every day, UTC.
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only:
+                - develop
+                - main
```
emjin, line 106: Why run on main? (I'm not actually clear on what main is in semgrep-core)

parsing-stats/Makefile
```diff
@@ -0,0 +1,15 @@
+.PHONY: all
+all:
+	./run-all
+
+.PHONY: test
+test:
+	$(MAKE) -C test
```
emjin: Nitpick: maybe `test-lang` would be better?

emjin (approved):

.circleci/config.yml
mjambon: There's no `main` branch in semgrep-core. In the code above, I mean "any branch that looks like the default branch". `main` is the default branch for new repos initialized on GitHub, and hopefully soon the default for `git init` as well. I'm just expecting that some day we'll switch to using `main` instead of `develop`.

parsing-stats/Makefile
mjambon: `test-lang` doesn't work because it's already file but yes, good idea. I'll change the folder name to `test-parsing`.

mjambon: oops sorry, it's `run-lang` that's already a file.

PR #3406, opened 2021-06-23
matching: Disable commutative-matching for && and || by default

### LINKED ISSUE #3399: Commutative operator breaking rules
https://semgrep.dev/s/grJJ

expecting no match. This rule is from https://github.com/dgryski/semgrep-go/blob/master/badnilguard.yml

IagoAbal:
Instead we perform associative-matching only, full AC-matching can be
requested in a per-rule basis.

Closes #3399

test plan:
make test # tests included

PR checklist:
- [x] changelog is up to date


IagoAbal:
It's actually a small amount of changes to the codebase, but I had to update a bunch of tests, and added some more.

mjambon (approved):

PR #3407, opened 2021-06-23
cli: Allow users to disable certain matching features in a rule

IagoAbal:
Follows: eb54e024f0c ("Allow users to disable certain matching features in a rule (#3338)")

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/rule_schema.yaml
+++ b/semgrep/semgrep/rule_schema.yaml
@@ -291,6 +291,9 @@ properties:
         metadata:
           title: Arbitrary structured data for your own reference
           type: object
+        settings:
+          title: Settings object to enable/disable certain matching features in semgrep-core
+          type: object
         pattern:
           title: Return finding where Semgrep pattern matches exactly
           type: string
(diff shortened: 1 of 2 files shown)


semgrep/semgrep/rule_schema.yaml
```diff
@@ -291,6 +291,9 @@ properties:
         metadata:
           title: Arbitrary structured data for your own reference
           type: object
+        settings:
```
emjin, line 294: We should include a test case (maybe `x = 2; x == 2`, and check that with constant prop off this doesn't match)

IagoAbal: Note that eb54e024f0c9f3dcac854e24e37a988bc775f6b4 added that test to semgrep-core/tests.

emjin: Would still be good to have one for semgrep, but that's good

aryx (approved):
I think drew preferred options to settings, but we can do that in another PR
to change it everywhere.

aryx (approved):

PR #3408, opened 2021-06-23
metavariable-pattern: Fix nested language implementation

IagoAbal:
1. Pass the right xlang to `mini_rule_of_pattern`, this fixes a crash
(`Impossible` exception) when using a non-generic pattern within a
generic rule.

2. Adjust parse info after `with_tmp_file`, this fixes the ranges given
of the matches obtained by metavariable-pattern/language (which dumps
the content of the metavariable into a temporary file in order to call
the parsing functions), so that these refer to the correct original file
and position. This alllows to start a patterns op with a pattern-not.

Fixes: 70a20605ae4 ("metavariable-pattern: Allow matching content under another language (#3366)")

test plan:
make test # tests included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -227,9 +227,15 @@ let lazy_force x = Lazy.force x [@@profiling]
 (* Adapters *)
 (*****************************************************************************)
 
-let (mini_rule_of_pattern : R.t -> Pattern.t * Rule.pattern_id * string -> MR.t)
-    =
- fun r (pattern, id, pstr) ->
+(* old: Before we used to pass the entire rule and we got the list of languages
+ * from there. But after metavariable-pattern we can now recursively call
+ * evaluate_formula using a different language than the one used by the rule!
+ * If the rule is a generic one, but the nested pattern formula is not, then
+ * this will raise Impossible... Thus, now we have to pass the language(s) that
+ * we are specifically targeting. *)
+let (mini_rule_of_pattern :
+      R.xlang -> Pattern.t * Rule.pattern_id * string -> MR.t) =
+ fun xlang (pattern, id, pstr) ->
   {
     MR.id = string_of_int id;
     pattern;
@@ -239,7 +245,7 @@ let (mini_rule_of_pattern : R.t -> Pattern.t * Rule.pattern_id * string -> MR.t)
     message = "";
     severity = MR.Error;
     languages =
-      ( match r.R.languages with
+      ( match xlang with
       | R.L (x, xs) -> x :: xs
       | R.LNone | R.LGeneric -> raise Impossible );
     (* useful for debugging timeout *)
@@ -359,8 +365,8 @@ let debug_semgrep config mini_rules equivalences file lang ast =
 (* Evaluating Semgrep patterns *)
 (*****************************************************************************)
 
-let matches_of_patterns config orig_rule equivalences
-    (file, xlang, lazy_ast_and_errors) patterns =
+let matches_of_patterns config equivalences (file, xlang, lazy_ast_and_errors)
+    patterns =
   match xlang with
   | R.L (lang, _) ->
       let (ast, errors), parse_time =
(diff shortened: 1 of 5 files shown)


semgrep-core/src/engine/Match_rules.ml
```diff
@@ -227,9 +227,15 @@ let lazy_force x = Lazy.force x [@@profiling]
...
 
-let (mini_rule_of_pattern : R.t -> Pattern.t * Rule.pattern_id * string -> MR.t)
-    =
- fun r (pattern, id, pstr) ->
+(* old: Before we used to pass the entire rule and we got the list of languages
+ * from there. But after metavariable-pattern we can now recursively call
+ * evaluate_formula using a different language than the one used by the rule!
+ * If the rule is a generic one, but the nested pattern formula is not, then
+ * this will raise Impossible... Thus, now we have to pass the language(s) that
+ * we are specifically targeting. *)
```
emjin, line 235: This is a great comment, thank you

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -239,7 +245,7 @@ let (mini_rule_of_pattern : R.t -> Pattern.t * Rule.pattern_id * string -> MR.t)
     message = "";
     severity = MR.Error;
     languages =
-      ( match r.R.languages with
+      ( match xlang with
```
emjin, line 248: Naming nit: perhaps `xlangs` would be better? So that it is obvious this could be a list of languages

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -359,8 +365,8 @@ let debug_semgrep config mini_rules equivalences file lang ast =
 (* Evaluating Semgrep patterns *)
 (*****************************************************************************)
 
-let matches_of_patterns config orig_rule equivalences
-    (file, xlang, lazy_ast_and_errors) patterns =
+let matches_of_patterns config equivalences (file, xlang, lazy_ast_and_errors)
```
emjin, line 368: I appreciate not having to pass through the original rule through here :)

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -866,19 +872,21 @@ and match_range_with_pattern env r mvar opt_lang formula =
...
           let lazy_ast_and_errors =
-            lazy_ast_of_string lang content (fun () ->
+            lazy_ast_of_string lang tok content (fun () ->
                 spf
                   "rule %s: metavariable-pattern: failed to fully parse the \
                    content of %s"
                   env.rule.id mvar)
           in
-          eval_nested_formula env env.xlang formula lazy_ast_and_errors
+          eval_nested_formula env
```
emjin: I'm not clear on how this works, but would it make more sense to update `env`, and always use `env.xlang` instead of passing through both `env` and `env.xlang` in most cases?

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -902,7 +910,19 @@ and match_range_with_pattern env r mvar opt_lang formula =
                 lazy_content
                 (Some { r with r = mval_range }) ) )
 
-and lazy_ast_of_string lang str warn_msg =
+and lazy_ast_of_string lang base_tok str warn_msg =
+  let fix_toks ast =
```
emjin, line 913: Could use a comment about why this is necessary

emjin (approved):

semgrep-core/src/engine/Match_rules.ml
IagoAbal: I kind of agree given that `xlang` can carry a list of languages, although in most places it is guaranteed to carry a single language. We basically take the argument to `-lang` and pass it around as an `xlang`. For now I also prefer to be consistent with the naming used in other places of the code. We would need to ask @pad here to have the full picture.  It might be relevant to have two separate types, one to represent the languages to which a rule applies, and another to represent the actual language that was selected for a particular file.

semgrep-core/src/engine/Match_rules.ml
IagoAbal: I agree!

semgrep-core/src/engine/Match_rules.ml
IagoAbal: Sure thing!

PR #3416, opened 2021-06-24
Refactor Match_rules somewhat

emjin:
Factorize the repeated code in `check` and what was `evaluate_nested_formula`. This function takes a formula and produces the matching ranges, so we call it `evaluate_formula`. The old `evaluate_formula` is now renamed to `compose_patterns`.

Also, rename the methods that return bools to make it more obvious. Am taking suggestions for names; I just moved around words in the original names

Test plan: this should change no functionality

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -945,7 +945,6 @@ and nested_formula_has_matches env formula lazy_ast_and_errors lazy_content
     evaluate_formula env formula lazy_ast_and_errors lazy_content opt_context
   in
   match final_ranges with [] -> false | _ :: _ -> true
-  [@@profiling]
 
 and evaluate_formula env formula lazy_ast_and_errors lazy_content opt_context =
   let xpatterns = xpatterns_in_formula formula in
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -710,114 +710,7 @@ let matches_of_xpatterns config equivalences
 (* Formula evaluation *)
 (*****************************************************************************)
 
-(* less: use Set instead of list? *)
-let rec (compose_patterns :
-          env -> range_with_mvars option -> R.formula -> range_with_mvars list)
-    =
- fun env ctx_opt e ->
-  match e with
-  | R.Leaf (R.P (xpat, inside)) ->
-      let id = xpat.R.pid in
-      let match_results =
-        try Hashtbl.find_all env.pattern_matches id with Not_found -> []
-      in
-      let kind =
-        match inside with
-        | Some R.Inside -> Inside
-        | None when R.is_regexp xpat -> Regexp
-        | None -> Plain
-      in
-      match_results
-      |> List.map match_result_to_range
-      |> List.map (fun r -> { r with kind })
-  | R.Or xs -> xs |> List.map (compose_patterns env ctx_opt) |> List.flatten
-  | R.And xs -> (
-      let pos, neg, conds = split_and xs in
-
-      (* we now treat pattern: and pattern-inside: differently. We first
-       * process the pattern: and then the pattern-inside.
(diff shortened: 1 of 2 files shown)


semgrep-core/src/engine/Match_rules.ml
```diff
@@ -711,7 +711,7 @@ let matches_of_xpatterns config equivalences
 (*****************************************************************************)
 
 (* less: use Set instead of list? *)
-let rec (evaluate_formula :
+let rec (compose_patterns :
```
IagoAbal: I liked `evaluate_formula` more for this function, IMO `compose_patterns` doesn't capture so well what's really going on. This is actually the function that evaluates the formula. My proposal would be to keep this function as `evaluate_formula` and re-name the other one as `evaluate_formula_on_content` (or similar). Just an opinion, maybe just give it another thought.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -938,21 +939,29 @@ and lazy_ast_of_string lang base_tok str warn_msg =
...
-  (* the following code is very similar to `check' below;
-     * we could maybe factorize things *)
+and nested_formula_has_matches env formula lazy_ast_and_errors lazy_content
+    opt_context =
+  let _, final_ranges =
+    evaluate_formula env formula lazy_ast_and_errors lazy_content opt_context
+  in
+  match final_ranges with [] -> false | _ :: _ -> true
+
+and evaluate_formula env formula lazy_ast_and_errors lazy_content opt_context =
```
IagoAbal: Nice refactoring here, minor comment: I would perhaps move this declaration up closer to `compose_patterns`.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -986,33 +994,23 @@ let check hook default_config rules equivs file_and_more =
...
-                * the formula *)
-               let pattern_matches_per_id =
-                 group_matches_per_pattern_id res.matches
-               in
+               let dummy_matches = Hashtbl.create 1 in
                let env =
                  {
                    config;
-                   pattern_matches = pattern_matches_per_id;
+                   pattern_matches = dummy_matches;
```
IagoAbal, line 1003: It seems cleaner to create `env` in `evaluate_formula` so we avoid passing it an `env` with dummy or unrelated `pattern_matches`. It's not that many arguments I think, because `config` is computed from the `rule` and it's only needed in `evaluate_formula`. And `file`, `xlang` and `lazy_ast_and_errors` could be passed together as `file_and_more` (which is already an argument of `check`).

IagoAbal: And after my PR, I wonder if we should be passing the entire rule in `env` or just the `rule.id` which is the only thing we really use right now!

IagoAbal (approved):
Nice refactoring! I gave you some food for thought but you are welcome to merge any time.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -938,21 +939,29 @@ and lazy_ast_of_string lang base_tok str warn_msg =
          if errors <> [] then pr2 (warn_msg ());
          (ast, errors)))
 
-and eval_nested_formula env formula lazy_ast_and_errors lazy_content opt_context
-    =
-  (* the following code is very similar to `check' below;
-     * we could maybe factorize things *)
+and nested_formula_has_matches env formula lazy_ast_and_errors lazy_content
```
IagoAbal, line 835: Definitely a better name!

semgrep-core/src/engine/Match_rules.ml
emjin: Happy to change it; I wasn't very comfortable with the dummy

emjin: Actually it is looking like a lot---I'll think about it more but merge for now so I can move on

PR #3418, opened 2021-06-25
Prevent spacegrep from skipping files

### LINKED ISSUE #2987: spacegrep error when parsing a markdown file
**Describe the bug**
spacegrep encounters an error when parsing the markdown file in issue https://github.com/returntocorp/semgrep/issues/2985.

**To Reproduce**
1. ❌  This is the original issue with a reduced pattern for easier debugging. This fails: https://semgrep.dev/s/ljGo/
2. ✅  Removing a big chunk of the markdown file works: https://semgrep.dev/s/YgJd/
3. ❌  Removing each line until it fails, this is the point at which it fails: https://semgrep.dev/s/qrWO/
4. ❌  Removing a big chunk from the top half does not work: https://semgrep.dev/s/6rPK/
5. ✅  Extremely minimal file, only two lines: https://semgrep.dev/s/ok3X/
6. ❌  Minimal version with the "offending" line from (3) above: https://semgrep.dev/s/zkZn/

My hypothesis include:
1.  Might be some kind of timeout behavior, and certain lines take a while to parse.
2. Might be related to backticks, which are in abundance and I haven't seen used extensively in spacegrep documents

P2

mjambon:
The `spacegrep` command line can take a single file or folders. It skips files silently based on a heuristic that guess whether the files look like programs, based on average line length. This causes markdown files with very long lines to be skipped. The `--force` option of spacegrep disables this heuristic.

I added test to check the performance on big and/or binary files. It's not that great and will depend on the pattern. Semgrep uses a 10MB limit for all targets, which should be fine for programs with proper indentation but will cause problems on minified files or some binary files that don't contain newlines. Rule authors must ensure that paths are filtered by extension like [this rule](https://github.com/returntocorp/semgrep-rules/blob/2e21f189da7a4b3a2c9f525bd26abf13e5a3e571/generic/dockerfile/best-practice/avoid-apk-upgrade.yaml#L17-L20). If that's already the case in practice, then we're good.

Fixes #2987

PR checklist:
- [x] changelog is up to date


CHANGELOG.md
```diff
@@ -23,6 +23,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   a generic rule
 - metavariable-pattern: Fix parse info when matching content of a metavariable
   under a different language
+- generic mode on Markdown files with very long lines will now work (#2987)
+
+### Changed
+- generic mode: files that don't look like nicely-indented programs
```
IagoAbal, line 29: I couldn't see any check being removed from the code, although perhaps it's a very small change and I'm missing it since there are many other changes to docs etc.

mjambon: It's the `--force` option that was added to the `spacegrep` command line... and the option in question was already implemented by spacegrep.

PR #3419, opened 2021-06-25
Run both normal and manual pre-commit stages in CI

mschwager:
I believe this PR meant we were _only_ running the manual pre-commit stages: https://github.com/returntocorp/semgrep/pull/1468.

If you look at recent PRs they are always skipping all files, i.e. pre-commit is effectively not being run:

- https://github.com/returntocorp/semgrep/runs/2789171649
- https://github.com/returntocorp/semgrep/runs/2824654375
- https://github.com/returntocorp/semgrep/runs/2640845166

Etc...


semgrep/tests/e2e/rules/invalid-rules/invalid-metavariable-regex.yaml
```diff
@@ -9,4 +9,4 @@ rules:
         regex: (boto|boto3)
   message: "Boto3 connection to internal network"
   languages: [python]
-  severity: ERROR
\ No newline at end of file
+  severity: ERROR
```
mschwager, line 12: These missing newlines caused the `invalid-rules` snapshots below to change. I'm not sure how important it is that these stay the same. Note that the snapshot changes are under the top-level `errors` key, not `results`, so this shouldn't affect downstream applications.

brendongo (approved):

PR #3420, opened 2021-06-25
Separate `pattern: $X` in anded patterns to a special field

emjin:
Before getting the matches of a formula, take out any `pattern: $X` or similar that is in a `And` list so that it can be processed separately. When there are other `And` cases, check the metavariable context produced for `$X` instead of getting the matches.

Sorry for the clutter; realized I needed Augmented_ranges to be its own module too late to make a PR for it alone.

Test plan:

`pattern: $X` alone is checked by the python tests. For correctness in the anded case, in `semgrep-core/` run (or `make test`)

```
sc -config tests/OTHER/rules/pattern-x.yaml tests/OTHER/rules/pattern-x.js -lang js
```

The expected results (you can confirm by running semgrep-core) are
```
perf/input/pattern-x.js:17 with rule typescript.react.security.audit.react-css-injection.react-css-injection
             <div style={input}>
perf/input/pattern-x.js:5 with rule typescript.react.security.audit.react-css-injection.react-css-injection
             <div style={input}>
```

To check that this actually worked, you can run

```
sc -config tests/OTHER/rules/pattern-x.js ~/your_root/semgrep/perf/bench/rails/input/rails/actionview/test/ujs/public/vendor/jquery-2.2.0.js -profile -lang js 
```

Before, this will take ~200 seconds, and half of that time will be spent on `uniq_by`. After, this will take ~100 seconds, and only 2 seconds will be spent on `uniq_by`

PR checklist:
- [x] changelog is up to date


mjambon (approved):
One thing I would do is start each file with a comment explaining what it's for.

IagoAbal:
> pattern: $X alone is checked by the python tests.

I am not sure what this means

semgrep-core/src/engine/Augmented_range.mli
```diff
@@ -0,0 +1,21 @@
+type range_kind = Plain | Inside | Regexp [@@deriving show]
```
IagoAbal, line 1: It would be good to add more doc to the module interfaces.

semgrep-core/src/engine/Augmented_range.ml
```diff
@@ -0,0 +1,122 @@
+type range_kind = Plain | Inside | Regexp [@@deriving show]
+
+(* range with metavars *)
+type range_with_mvars = {
```
IagoAbal, line 4: Once we moved this data type to its own module then I would just call it `t` as it is tradition in OCaml.

semgrep-core/src/engine/Selector.ml
```diff
@@ -0,0 +1,130 @@
+module R = Rule
```
IagoAbal, line 1: Should we have an `.mli` for this module?

semgrep-core/src/engine/Selector.ml
```diff
@@ -0,0 +1,130 @@
...
+module RP = Report
+
+type selector = {
+  mvar : M.mvar;
+  pid : int;
+  pstr : string;
+  lazy_matches : RP.times RP.match_result Lazy.t;
+}
+
+type sformula =
```
IagoAbal, line 15: I would name the module and the type a bit different, something derived from "optimized formula". Looking into the future, I imagine that we will analyze the formula and produce a new "optimized" one. So this would not be specific to the `pattern: $X` case.

semgrep-core/src/engine/Selector.ml
```diff
@@ -0,0 +1,130 @@
...
+(*****************************************************************************)
+
+(* this will be adjusted later in range_to_pattern_match_adjusted *)
+let fake_rule_id (id, str) =
+  { PM.id = string_of_int id; pattern_string = str; message = "" }
+
+let match_selector (sel_opt : selector option) : AR.ranges =
+  let get_match (x : RP.times RP.match_result) = x.matches in
+  match sel_opt with
+  | None -> failwith "empty And; no positive terms in And"
```
IagoAbal, line 35: This error message has been taken out of context, so it doesn't make a lot of sense here. Why should we fail with `"empty And; ..."` if `sel_opt` is `None` ? This function could instead return `AR.ranges option` and we could raise the error in `evaluate_formula`.

semgrep-core/src/engine/Selector.ml
```diff
@@ -0,0 +1,130 @@
...
+
+let selector_equal s1 s2 = s1.mvar = s2.mvar
+
+(*****************************************************************************)
+(* Converter *)
+(*****************************************************************************)
+
+let formula_to_sformula match_func formula =
+  let rec formula_to_sformula formula =
+    let selector_from_formula f =
```
IagoAbal, line 78: This function could be pull out of the `let rec formula_to_sformula`, would make it clear that `selector_from_formula` can be understood separately and does not depend on any of the arguments of `let rec formula_to_sformula`.

semgrep-core/src/engine/Selector.ml
```diff
@@ -0,0 +1,130 @@
...
+      | [] -> (selector, acc)
+      | x :: xs ->
+          let selector, acc =
+            match (selector, selector_from_formula x) with
+            | None, None -> (None, x :: acc)
+            | Some s, None -> (Some s, x :: acc)
+            | None, Some s -> (Some s, acc)
+            | Some s1, Some s2 ->
+                if selector_equal s1 s2 then (Some s1, acc) else (None, [])
+            (* This will never be valid *)
```
IagoAbal, line 104: What do you mean by valid here? Do you mean that the pattern-and is not valid?

Maybe I'm missing something important here, but it seems to me that:
```
patterns:
  - pattern: $X
  - pattern: $Y
```
which should match something will become `(None, [])` which matches nothing?

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   under a different language
 
 ### Changed
+- When anded with other patterns, `pattern: $X` will not be evaluated on its own, but will look at the context and find `$X` within the metavariables bound
```
aryx, line 29: you should add "which should be significantly faster"

semgrep-core/src/engine/Augmented_range.ml
aryx: Yep, I agree.
I would call the module maybe Range_with_metavars.ml

semgrep-core/src/engine/Selector.ml
aryx: Not a super fan of the name, Selector.ml.
Also, at some point we want to generalize this to handle different special cases, not just
pattern:$PROP but also pattern: include("foo") ..., and other things we think we can
specialize.

semgrep-core/src/engine/Selector.ml
aryx: yep, I agree.

aryx: Actually the best would be to move this Optimize_formula.ml or Specialize_formula.ml in src/optimizing/
to keep the engine/ clean.

semgrep-core/src/engine/Selector.ml
aryx: Yep, @emjin it's usually better to avoid nested functions. In the end it makes the code a bit more complicated.

aryx (changes requested):
A few comments to address and we can then merge this.

IagoAbal (approved):
Great!

emjin:
Messed up this branch, see https://github.com/returntocorp/semgrep/pull/3435

PR #3422, opened 2021-06-26
Memoize range_of_any for statements

emjin:
Save the range of statements so that range_of_any is much faster.

Test plan:

make test

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/core/ast/AST_generic.ml
+++ b/semgrep-core/src/core/ast/AST_generic.ml
@@ -897,7 +897,11 @@ and stmt = {
      and before matching.
   *)
   (* used in semgrep to skip some AST matching *)
-  mutable s_bf : Bloom_filter.t option; [@equal fun _a _b -> true] [@hash.ignore]
+  mutable s_bf : Bloom_filter.t option;
+      [@equal fun _a _b -> true] [@hash.ignore]
+  mutable s_range :
+    (Parse_info.token_location * Parse_info.token_location) option;
+      [@hash.ignore]
 }
 
 and stmt_kind =
@@ -1976,6 +1980,7 @@ let s skind =
     s_use_cache = false;
     s_backrefs = None;
     s_bf = None;
+    s_range = None;
   }
 
 (*s: function [[AST_generic.basic_field]] *)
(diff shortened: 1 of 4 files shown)


mjambon (approved):

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -897,7 +897,11 @@ and stmt = {
      and before matching.
   *)
   (* used in semgrep to skip some AST matching *)
-  mutable s_bf : Bloom_filter.t option; [@equal fun _a _b -> true] [@hash.ignore]
+  mutable s_bf : Bloom_filter.t option;
+      [@equal fun _a _b -> true] [@hash.ignore]
+  mutable s_range :
+    (Parse_info.token_location * Parse_info.token_location) option;
+      [@hash.ignore]
```
mjambon, line 904: If this field is related to bloom-filter optimization, it should clearly indicate so. A comment + maybe a name containing `bf` would help.

mjambon (approved):

semgrep-core/src/core/ast/AST_generic.ml
emjin: It isn't; `s_range` is just saving the range of the statement

semgrep-core/src/core/ast/Visitor_AST.ml
```diff
@@ -1233,12 +1234,54 @@ let ii_of_any any =
...
-  List.filter Parse_info.is_origintok tokens |> Parse_info.min_max_ii_by_pos
+  List.filter PI.is_origintok tokens |> PI.min_max_ii_by_pos
   [@@profiling]
 
-let range_of_any any =
-  let leftmost_token, rightmost_token = ii_of_any any |> range_of_tokens in
-  ( Parse_info.token_location_of_info leftmost_token,
-    Parse_info.token_location_of_info rightmost_token )
-  [@@profiling]
+let range_of_any any = extract_ranges (fun visitor -> visitor any) [@@profiling]
```
aryx, line 1287: maybe you could add in the .mli that this function will have some side effects and set the s_range fields for any substmt in any.

aryx (approved):

aryx:
Do you have a concrete command one can run to see the difference in performance?

PR #3424, opened 2021-06-27
Fixed metavariable-comparison failure when comparing integers/floats

### LINKED ISSUE #3423: Metavariable comparison fails in mixed integer/float comparisons
**Describe the bug**
The `metavariable-comparison` rule fails when comparing integers and floats. This is particularly problematic when the value of the metavariable can be either/or because then you aren't able to adjust the comparison to match.

**To Reproduce**
Take this example from the docs: https://semgrep.dev/s/L83L

This works as-is, but if you change _any_ of the integers in either the rule or the source to a float then you get an error (see below for screenshot)

**Expected behavior**
I would expect that this comparison would work regardless of typing since the comparison language is python-like according to [the documentation](https://semgrep.dev/docs/writing-rules/rule-syntax/#metavariable-comparison). Since this appears to be a type mismatch issue, I would expect that the integers would be upcast to a float for the comparison

**Screenshots**
Error on the live editor after comparing float to int:
image

**Environment**
Discovered locally running binary v0.51.0 and reproduced on the live editor (link above)

jtimmons:
Fixes #3423 where `metavariable-comparison` was failing when performing operations with mixed integers and floats. This deals with that by upcasting the integers to floats in those mixed cases. Additionally, there was no modulo implementation for floats so I added that as well.

First contrib and not super familiar with OCaml so let me know if I missed something here! Just looked into #3423 a bit after running into it locally and it seemed simple enough so thought I'd give it a shot.

PR checklist:
- [x] changelog is up to date


semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -191,37 +191,54 @@ let rec eval env code =
...
+  | G.And, [ Bool b1; Bool b2 ] -> Bool (b1 && b2)
+  | G.Not, [ Bool b1 ] -> Bool (not b1)
+  | G.Or, [ Bool b1; Bool b2 ] -> Bool (b1 || b2)
   | G.Gt, [ Int i1; Int i2 ] -> Bool (i1 > i2)
-  | G.GtE, [ Int i1; Int i2 ] -> Bool (i1 >= i2)
-  | G.Lt, [ Int i1; Int i2 ] -> Bool (i1 < i2)
-  | G.LtE, [ Int i1; Int i2 ] -> Bool (i1 <= i2)
-  (* todo: we should perform cast and allow to mix Float and Int *)
   | G.Gt, [ Float i1; Float i2 ] -> Bool (i1 > i2)
+  | G.Gt, [ Int i1; Float i2 ] -> Bool (float_of_int i1 > i2)
```
aryx, line 199: We could maybe factorizing code by matching
op, [Int x; Float y] and call recursively eval_op with upcasted Int,
but this is fine too.

aryx (approved):

aryx (approved):

aryx:
Thx a lot!

daghan:
@jtimmons can you fill this [form](https://r2c.dev/swag) so I can send you some contributor swag?

jtimmons:
> @jtimmons can you fill this [form](https://r2c.dev/swag) so I can send you some contributor swag?

Oh awesome, done!

jtimmons:
@aryx: I actually had no idea that the `metavariable-comparison` operator supported arithmetic expressions like that until I got into the code - do you mind if I take a stab at updating [the docs](https://semgrep.dev/docs/writing-rules/rule-syntax/#metavariable-comparison) to include an example?

aryx:
Sure, that would be great!

PR #3425, opened 2021-06-28
metavariable-pattern: Add support for non-Semgrep patterns

### LINKED ISSUE #3411: metavariable-pattern does not work on metavariables bound by a regex
**Describe the bug**
Once we now bind capture groups to metavariables (`$1`, `$2`, ...) we should be able to filter these matches using `metavariable-pattern`, but this does not seem to work yet.

**To Reproduce**
https://semgrep.dev/s/Dwko

IagoAbal:
This allows metavariable-pattern to invoke the Spacegrep, regex, and
Comby engines.

Closes #3411

test plan:
make test # test included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -753,7 +753,7 @@ let rec filter_ranges env xs cond =
              in
              Eval_generic.eval_bool env e)
 
-and range_matches_with_pattern env r mvar opt_lang formula =
+and range_matches_with_pattern env r mvar opt_xlang formula =
   let bindings = r.mvars in
   (* If anything goes wrong the default is to filter out! *)
   match List.assoc_opt mvar bindings with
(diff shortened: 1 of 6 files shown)


semgrep-core/src/engine/Match_rules.ml
```diff
@@ -801,36 +785,57 @@ and range_matches_with_pattern env r mvar opt_lang formula =
...
-           Parse_target.parse_and_resolve_name_use_pfff_or_treesitter lang file
-         in
-         let ast = fix_toks ast in
-         if errors <> [] then pr2 (warn_msg ());
-         (ast, errors)))
+                (Some { r with r = mval_range }) )
+      | Some xlang, MV.Text (content, _tok)
+      | Some xlang, MV.E (G.L (G.String (content, _tok))) ->
+          (* We reinterpret the matched text as `xlang`. *)
+          let ext =
```
aryx, line 792: I don't think the extension really matters.  I don't think Parse_target will reject files with the wrong extension

aryx (approved):

Issue #3410, opened 2021-06-24
Constness analysis is very slow on very large functions


IagoAbal:
**To Reproduce**
```
parsing-stats$ ../semgrep-core/bin/semgrep-core -lang csharp -dump_named_ast lang/csharp/tmp/aspnetcore/src/Http/Routing/perf/Microbenchmarks/Matching/MatcherAzureBenchmarkBase.generated.cs
```

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

mjambon:
Here's how I found it:
```
cd parsing-stats
./run-lang csharp
```
You can run it for other languages and see if you get a similar problem. I just copied the `develop` branch into `parsing-stats`, which triggers a parsing-stats job for all languages in CircleCI. You [can download the log](https://app.circleci.com/pipelines/github/returntocorp/semgrep/4512/workflows/1e996a8e-dcf0-4240-85a7-fca5ee739f40/jobs/6142) if that helps.

aryx:
please copy the offending file in semgrep-core/perf/input/lxxxx.cs so we can quickly reproduce (and maybe add it as a part
of perf regression testing, see semgrep-core/perf/run-perf-suite)

IagoAbal:
> please copy the offending file in semgrep-core/perf/input/lxxxx.cs so we can quickly reproduce (and maybe add it as a part
> of perf regression testing, see semgrep-core/perf/run-perf-suite)

The file is part of a microbenchark for aspnetcore, the problematic function looks like this:
```
        private protected void SetupRequests()
        {
            Requests = new HttpContext[5160];
            Requests[0] = new DefaultHttpContext();
            Requests[0].RequestServices = CreateServices();
            Requests[0].Request.Method = HttpMethods.GetCanonicalizedValue("GET");
            Requests[0].Request.Path = "/account";
            Requests[1] = new DefaultHttpContext();
            Requests[1].RequestServices = CreateServices();
            Requests[1].Request.Method = HttpMethods.GetCanonicalizedValue("POST");
            Requests[1].Request.Path = "/analyze";
            Requests[2] = new DefaultHttpContext();
            Requests[2].RequestServices = CreateServices();
            Requests[2].Request.Method = HttpMethods.GetCanonicalizedValue("GET");
            Requests[2].Request.Path = "/apis";
[...]

PR #3434, opened 2021-06-29
parsing-stats: Set a timeout of 10s per file

IagoAbal:
This was motivated by data-flow const-prop being too slow on:

    aspnetcore/src/Http/Routing/perf/Microbenchmarks/Matching/MatcherAzureBenchmarkBase.generated.cs

But we could have bad perf for any other reason, so it is a good idea to set a timeout.

Related-to: #3410

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/parsing/Test_parsing.ml
+++ b/semgrep-core/src/parsing/Test_parsing.ml
@@ -175,6 +175,7 @@ let test_parse_tree_sitter lang xs =
    semgrep scans.
 *)
 let parsing_common ?(verbose = true) lang xs =
+  let timeout_seconds = 10 in
   let xs = List.map Common.fullpath xs in
   let fullxs =
     Lang.files_of_dirs_or_files lang xs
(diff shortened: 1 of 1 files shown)


semgrep-core/src/parsing/Test_parsing.ml
```diff
@@ -186,8 +187,9 @@ let parsing_common ?(verbose = true) lang xs =
          let stat =
            try
              let res =
-               Parse_target.parse_and_resolve_name_use_pfff_or_treesitter lang
-                 file
+               Common.timeout_function ~verbose:false timeout_seconds (fun () ->
```
aryx, line 190: In the code below that catch the exn, instead of PI.bad_stat file, we could also set the PI.have_timeout to true
in case of a Timeout exn.
It's probably not used by the dashboard metrics, but we may at some point.

aryx (approved):

PR #3440, opened 2021-06-29
analyzing: dataflow: Optimize env manipulation functions

### LINKED ISSUE #3410: Constness analysis is very slow on very large functions

IagoAbal:
Re-implemented union/diff functions in terms of Map.merge. The Map.fold
based implementations re-allocated the entire map `env2` even if `env1`
was empty. If the map is large then this triggers too much GC slowing
down the execution.

Most of the perf gain just required adding:

    if VarMap.is_empty ... then ... else ...

But we went with Map.merge anyways because the implementations are more
concise.

Closes #3410

test plan:
time semgrep-core/bin/semgrep-core -lang csharp -dump_named_ast parsing-stats/lang/csharp/tmp/aspnetcore/src/Http/Routing/perf/Microbenchmarks/Matching/MatcherAzureBenchmarkBase.generated.cs &>/dev/null
  #^ now takes less than a minute

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/analyzing/Dataflow.ml
+++ b/semgrep-core/src/analyzing/Dataflow.ml
@@ -124,50 +124,41 @@ let eq_inout eq io1 io2 =
 
 let (varmap_union : ('a -> 'a -> 'a) -> 'a env -> 'a env -> 'a env) =
  fun union_op env1 env2 ->
-  let acc = env1 in
-  VarMap.fold
-    (fun var x acc ->
-      let x' = try union_op x (VarMap.find var acc) with Not_found -> x in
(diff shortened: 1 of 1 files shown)


aryx (approved):
Excellent! Clearer and faster, love those diffs.

PR #3444, opened 2021-06-30
Use TSX parser for JavaScript

mjambon:
Switching to using tree-sitter's TSX parser for JavaScript instead of tree-sitter-javascript (JSX). This will save effort in maintaining the CST-to-AST mapping.

* Uses an updated pfff: https://github.com/returntocorp/pfff/pull/449
* Uses an updated semgrep-rules: https://github.com/returntocorp/semgrep-rules/pull/1362 (see comment down the page)

PR checklist:
- [x] changelog is up to date


aryx:
exciting!

mjambon:
Two tests are failing tests on semgrep-rules. See https://github.com/returntocorp/semgrep-rules/pull/1362

We're blocked on https://github.com/tree-sitter/tree-sitter-javascript/pull/183 and on the follow-up integration of javascript+typescript into semgrep. I'm doing this next.

mjambon:
This should work now, using the correct version of semgrep-rules. I believe I fixed the test expectations the other day, in this PR which hasn't been merged: https://github.com/returntocorp/semgrep-rules/pull/1362

We have a CI deadlock because:
* semgrep CI tests against the main branch of semgrep-rules, and
* semgrep-rules CI tests against the main branch of semgrep

This would be solved by using a git submodule in semgrep pointing to the correct branch of semgrep-rules. I don't know how this kind of problem has been solved in the past.

aryx (approved):

aryx:
Ok merging this. I tested locally and with the modified semgrep-rules, make test is passing with this version of semgrep-core.
The only CI failures in this job are related to semgrep-rules, so we should be good.

PR #3447, opened 2021-06-30
Cleanup comments

aryx:
test plan:
make test

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/core/il/IL.ml
+++ b/semgrep-core/src/core/il/IL.ml
@@ -88,23 +88,15 @@ module G = AST_generic
 type tok = G.tok [@@deriving show]
 
 (*e: type [[IL.tok]] *)
-
-(* with tarzan *)
-
 (*s: type [[IL.wrap]] *)
 type 'a wrap = 'a G.wrap [@@deriving show]
 
 (*e: type [[IL.wrap]] *)
-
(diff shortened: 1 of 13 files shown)


semgrep-core/src/core/il/IL.ml
```diff
@@ -205,18 +189,18 @@ and var_special = This | Super | Self | Parent
...
  *)
-(*s: type [[IL.exp]] *)
-and exp = { e : exp_kind; (* todo: etype: typ; *)
-                          eorig : G.expr }
+and exp = { e : exp_kind; eorig : G.expr }
 
 (*e: type [[IL.exp]] *)
 (*s: type [[IL.exp_kind]] *)
 and exp_kind =
-  | Lvalue of lval (* lvalue used in a rvalue context *)
```
IagoAbal, line 219: I used to call this `Fetch` (since you're fetching the value contained in the lvalue). I still think `Lvalue` was more appropriate than `Rvalue`, because basically any `exp_kind` is an r-value, and `Lvalue` made clear that it is the r-value obtained from an l-value.

aryx: Ok, I'll rename Fetch. I like Fetch.

PR #3448, opened 2021-06-30
core: perf: Add regression test for #3410

IagoAbal:
On a MacBook Pro M1 (16GB), before #3440 this test would take ~40s,
after #3440 it takes ~1s.

Follows: 94a30efe4bf ("analyzing: dataflow: Optimize env manipulation functions (#3440)")

PR checklist:
- [ ] changelog is up to date


semgrep-core/perf/run-perf-suite
```diff
@@ -87,6 +89,10 @@ def run_tests(
 
         pattern_file = os.path.join(dirname, test.name + ext)
         target_file = os.path.join(dirname, test.target)
+        target_file_gz = os.path.join(dirname, test.target + ".gz")
+
+        if os.path.isfile(target_file_gz):
+            subprocess.check_call(["gunzip", "-f", target_file_gz])
```
aryx: so the uncompressed file will be in perf/input/? Should we ignore it in .gitignore then?

aryx (approved):

semgrep-core/perf/run-perf-suite
IagoAbal: We'll have to ignore one by one, but I will do.

PR #3459, opened 2021-07-02
Add optimization to send multiple rules to semgrep-core

brendongo:
PR checklist:
- [ ] changelog is up to date


brendongo:
Closing for now since with recent improvements in ocaml side, startup time is less of a factor. Will revisit this after API between semgrep and semgrep-core is solidified

Issue #3430, opened 2021-06-28
Slow parsing and matching with spacegrep


mjambon:
Here are 3 slow rules using the generic mode (reported by @emjin ):
* https://github.com/returntocorp/semgrep-rules/blob/develop/python/django/security/audit/xss/template-translate-as-no-escape.yaml
* https://github.com/returntocorp/semgrep-rules/blob/develop/python/django/security/audit/xss/template-translate-no-escape.yaml
* https://github.com/returntocorp/semgrep-rules/blob/develop/python/django/security/audit/xss/template-blocktranslate-no-escape.yaml

What they have in common is this subpattern:
```
foo
...
...
...
...
...
...
...
...
...
...
bar
```
That's `...` repeated 10 times. This is currently the only way to allow an ellipsis to span over up to 100 lines, since a single `...` spans at most 10 lines.

The problem is that the matching engine doesn't do any caching at the moment. When `foo` matches but there's no `bar`, and there are n items following `foo`, each `...` will be matched starting from each possible position. Within the 10-line limit, the cost of matching a single `...` ellipsis is O(n) (which is an actual asymptotic complexity if line length is unbounded). The cost of matching 10 of these in a sequence is O(n^10). If there are 5 tokens per line and there are 10 remaining lines, that gives us a typical n=50. For n=50, the complexity of the matching the pattern is 50^10 times the cost of the matching the tail of the pattern (`bar`), which is a lot.

Proposal: internally collapse successive `...` into one with an increased line span limit. This changes nothing for the user, but the cost of matching 10 `...` on a block of at most 10 lines would become the same a single `...`. In blocks longer then 100 lines, the cost becomes 10*10 instead of 10^10.

mjambon:
@emjin I haven't been able reproduce the slowness on artificial files. We already have an optimization that takes care of treating consecutive `...` as one. Do you have a target that's know to be slow for one of these rules? Preferably over 100 ms.

emjin:
You can use `semgrep/perf/bench/gitlab/input/gitlab/locale/ko/gitlab.po`, which takes half a second to match on my machine. In general, I don't think it takes a super long time on any one file, it just takes a non-negligible amount of time for each file, which adds up.

For the gitlab repo, here is the general summary:
```
Semgrep-core timing summary:
Total CPU time: 3092.0721 File parse time: 2661.0576 Rule parse time: 2.7672 Match time: 340.6150
Slowest 5/36936 files
[...]

PR #3460, opened 2021-07-02
spacegrep: add post-parse/pre-match optimization

mjambon:
This wasn't too hard to add, but also the savings are only about 15% on one of the patterns for django templates (#3430) as shown here:

```
[spacegrep] Run complex pattern on big django template

real	0m0.173s
user	0m0.134s
sys	0m0.039s
[spacegrep] Run complex pattern on big django template - no pre-match optimization

real	0m0.203s
user	0m0.178s
sys	0m0.024s
```

This result is produced by `make test` from spacegrep's base folder. I put 200 KB of input data under version control, which isn't great but I think is still reasonable if we don't add much more.

The conclusion of this is that parsing takes longer than matching in this case, despite taking a pattern from the slowest rules we could find.

Next, we might want to add a pre-parsing optimization: it would be essentially the same thing but would run a grep-like scan on the raw input file.

PR checklist:
- [x] changelog is up to date


semgrep-core/src/spacegrep/src/bin/Spacegrep_main.ml
```diff
@@ -24,6 +24,7 @@ type config = {
   time : bool;
   timeout : int option;
   warn : bool;
+  no_skip_search : bool;
```
IagoAbal, line 27: Although the CLI may expose `--no-skip-search` because you want this enabled by default, in the code I would have gone for `skip_search` since it's simpler to read. Minor comment.

IagoAbal:
> the savings are only about 15% on one of the patterns for django templates 

Does this (may) have any measurable cost on some other benchmarks?

edit: After reviewing the code, I think it probably does not. Although it would be good to stop traversing the document as soon as it is clear that the pattern may match.

IagoAbal:
Would #3461 supersede this optimization?

semgrep-core/src/spacegrep/src/lib/Pre_match.ml
```diff
@@ -0,0 +1,73 @@
+(*
+   Fast check that tells whether a pattern has any chance of matching the
+   document with 'Match.search'.
+*)
+
+(*
+   Check if the pattern contains 'Byte' tokens, since they often don't
+   and the target often does. This is meant as an optimization to not
+   fill the lookup table with many useless entries.
```
IagoAbal: The comment does not align with what I see in the rest of the code, or maybe I am not getting it correctly? The only place where this is used is in:

```ocaml
if check_bytes then Hashtbl.remove chars k
```
So it is not preventing anything from being added too the lookup table.

IagoAbal: Maybe before you were building the lookup table from the document? Then it would make sense. But the lookup table is built from the pattern.

semgrep-core/src/spacegrep/src/lib/Pre_match.ml
```diff
@@ -0,0 +1,73 @@
...
+    | List nodes -> check nodes
+  and check_atom = function
+    | Byte k -> if check_bytes then Hashtbl.remove chars k
+    | Word word ->
+        let k = if case_sensitive then word else lowercase word in
+        Hashtbl.remove words k
+    | Punct k -> Hashtbl.remove chars k
+  in
+  check doc;
+  Hashtbl.length words = 0 && Hashtbl.length chars = 0
```
IagoAbal, line 74: This test is super cheap so we should be making it in `check` before going down into the nodes. Then you don't have to traverse the entire document in every case.

IagoAbal (approved):
It is nice! But I think #3461 would supersede this making it unnecessary. So maybe it's better to just implement #3461?

semgrep-core/src/spacegrep/src/bin/Spacegrep_main.ml
mjambon: I'd rather not because "skip search" sounds like a request to skip the search, but it only means "skip search if possible".

mjambon:
> It is nice! But I think #3461 would supersede this making it unnecessary. So maybe it's better to just implement #3461?

Yes, I added a comment about that. We can remove it when #3461 is done.

semgrep-core/src/spacegrep/src/lib/Pre_match.ml
mjambon: maybe; it was just a first guess, I didn't check the impact but I don't think it's worth it since the whole module will be replaced by #3461.

mjambon:
> > the savings are only about 15% on one of the patterns for django templates
> 
> Does this (may) have any measurable cost on some other benchmarks?
> 
> edit: After reviewing the code, I think it probably does not. Although it would be good to stop traversing the document as soon as it is clear that the pattern may match.

I just added support for printing total parse time and total match time. We can see that for this particular pattern, matching time is significantly reduced (4x). This matters as soon as targets are not reparsed over and over. We'll find out about the global impact once this is merged.

```
13:44:28 martin@mj2 [mj-prematch] ~/semgrep/semgrep-core/src/spacegrep $ time ./bin/spacegrep -p perf-data/django-template.pat -d tmp/django-template.po --time --no-skip-search
parse time: 0.149465 s
match time: 0.037967 s

real	0m0.199s
user	0m0.162s
sys	0m0.037s
13:47:13 martin@mj2 [mj-prematch] ~/semgrep/semgrep-core/src/spacegrep $ time ./bin/spacegrep -p perf-data/django-template.pat -d tmp/django-template.po --time
parse time: 0.156106 s
match time: 0.008484 s

real	0m0.175s
user	0m0.150s
sys	0m0.025s
```

mjambon:
Here's the performance measured for a very simple pattern (pattern = `foo`) on a large file. It's still a 2x speedup on match time.
```
[spacegrep] Run simple pattern on big file
parse time: 0.800288 s
match time: 0.046390 s

real	0m0.871s
user	0m0.750s
sys	0m0.120s
[spacegrep] Run simple pattern on big file - no pre-match optimization
parse time: 0.786020 s
match time: 0.098097 s

real	0m0.913s
user	0m0.761s
sys	0m0.151s
```

PR #3462, opened 2021-07-02
AST_to_IL: Translate Python's with statements (roughly)

### LINKED ISSUE #3402: [Python] constant propagation inside/outside context created by `with` statement
https://semgrep.dev/s/RP8O/

```python
def test1():
    hardcoded = "yo man!"
    with something() as foobar:
        print(hardcoded)
```

```yaml
  patterns:
    - pattern: |
        print(...)
    - pattern-not: |
        print("...")
```

constant propagation does not work for variables outside `with` context

IagoAbal:
Closes #3402

PR checklist:
- [x] changelog is up to date


semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -793,13 +793,102 @@ let rec stmt_aux env st =
       let stmt2 = stmt env stmt2 in
       stmt1 @ stmt2
   | G.DisjStmt _ -> sgrep_construct (G.S st)
+  | G.OtherStmtWithStmt (G.OSWS_With, Some mk_mgr_as_pat, body) ->
```
aryx: mgr?
manager?

aryx (approved):

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: That's right, I made it more clear in the code.

PR #3466, opened 2021-07-02
Trigger semgrep rule from fork

brendongo:
PR checklist:
- [ ] changelog is up to date


brendongo:
Looks like with pull_request_target PRs from forks are able to run semgrep.

PR #3468, opened 2021-07-02
ci: be able to run semgrep linter from forked projects

brendongo:
Redoes #3464

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,7 +1,7 @@
 name: Lint
 on:
   workflow_dispatch:
-  pull_request:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


mjambon (approved):

.github/workflows/lint.yml
```diff
@@ -1,7 +1,7 @@
 name: Lint
 on:
   workflow_dispatch:
-  pull_request:
+  pull_request_target:
```
aryx, line 4: This is annoying because any modifications to lint.yml is not taken into account in the PR.
See the problem I had here: https://github.com/returntocorp/semgrep/pull/3516

PR #3470, opened 2021-07-05
spacegrep: Fix non-portable use of zcat

IagoAbal:
scripts/show-perf was failing on MacOS 11.

Fixes: 5128b066302 ("spacegrep: add post-parse/pre-match optimization (#3460)")

PR checklist:
- [ ] changelog is up to date


IagoAbal:
@mjambon do we need to run `scripts/show-perf` as part of `make test`? Maybe it could be run with a timeout (so we actually check something) and redirecting stderr to `/dev/null` or to some log file.

aryx (approved):

PR #3478, opened 2021-07-06
Revert `pattern: $X` optimization

### LINKED ISSUE #3476: Regression due to `pattern: $X` optimization
**Describe the bug**

We can't (in general) optimize a `pattern: $X` that follows a sequence of `pattern-inside`s by simply extracting the value of `$X` in the `pattern-inside` matches. As long as there is a `...`, `pattern: $X` could be match inside that region of code, and the result would not be equivalent.

**To Reproduce**

The link to the Playground would be https://semgrep.dev/s/65Bq but the example gets rewritten and no longer works...

Config file:
```yaml
rules:
- id: test
  patterns:
    - pattern-inside: |
        f($X, ...)
        ...
    - pattern: |
        $X
  languages: [python]
  severity: WARNING
```

Target file:
```python
#ruleid:test
f(a, b, c)
g(c)
h(b)
#ruleid:test
i(a) # NOT MATCHING
```

emjin:
Closes https://github.com/returntocorp/semgrep/issues/3476

Turn off the `pattern: $X` optimization for now. To perform it properly, we need to get the sub-AST of the range produced by the other patterns in the `and` and run the pattern on that range.

Test plan: make test

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Java varargs are now correctly matched (#3455)
 - Support for partial statements (e.g., `try { ... }`) for Java (#3417)
 - Constant propagation now works inside Python `with` statements (#3402)
+- Revert `pattern: $X` optimization (#3476)
 
 ### Changed
 - Faster matching times for generic mode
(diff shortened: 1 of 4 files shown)


IagoAbal (approved):
I hope that we find another way to implement this.

Not needed if you plan to work on this very soon, but otherwise we should leave a TODO note somewhere.

emjin:
There's a TODO in specialize_formula, but I also hope to work on this after hack week.

IagoAbal:
> There's a TODO in specialize_formula

Ooops, true, lapsus on my side.

> I also hope to work on this after hack week

Great!

PR #3483, opened 2021-07-07
Fixed metavariable name collision during interpolation of message / autofix

### LINKED ISSUE #3458: Generic/autofix Metavariable Name Issue
**Describe the bug**
When applying autofixes, uses metavariable with shortest name instead of the actual variable referenced. 
set $B, autofix will never use a variable named $BA or $BAR, it will fill in that $B first.

**To Reproduce**
https://semgrep.dev/s/xePj

**Expected behavior**
We would expect it to use the variables that I give it, read the whole variable name, not just the first matching characters

**Environment**
Currently local, ultimately it will be a gitlab pipeline involving a docker image

jtimmons:
Fixes #3458 where the string interpolation process used for both `message` display and `autofix` replacement was having issues with name collisions. This was happening if there were two metavariables in which one name was a subset of another name (eg. `$X` vs `$X2`); since the metavariables are replaced one-by-one in no particular order the shorter one could be hit first and replace part of the longer one. In the previous example this would replace `$X` with its value and `$X2` with the value of `$X` with `2` tacked on the end.

This PR sorts the metavariable names in descending length order so that longer names such as `$X2` will always be resolved before the shorter ones. To keep behavior consistent, it also combines the two interpolation functions into one and adds a unit test for them.

Original example snippet from the issue: https://semgrep.dev/s/xePj

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Java varargs are now correctly matched (#3455)
 - Support for partial statements (e.g., `try { ... }`) for Java (#3417)
 - Constant propagation now works inside Python `with` statements (#3402)
+- Metavariable value replacement in message/autofix no longer mixes up short and long names like $X vs $X2 (#3458)
 
 ### Changed
 - Faster matching times for generic mode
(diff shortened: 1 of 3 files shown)


brendongo (approved):
Nice catch! Thanks @jtimmons

Issue #3295, opened 2021-06-07
Implement taint-analysis in semgrep-core


emjin:
Currently, we're going through `--optimizations none` for taint rules. Change this

aryx:
I think we can wait for @IagoAbal to finish tainting 2.0.

PR #3486, opened 2021-07-08
core: Refactor taint mode

IagoAbal:
1. Generalize Rule type to absorb Tainting_rule.
2. Use same parser for search and taint rules.
3. Move tainting code into the engine library.
4. Let -config run taint rules, remove -tainting_rules_file
   and -dump_tainting_rules.
5. Print taint-mode matches to stdout if -json is not given.

NOTE: This requires an update of Semgrep CLI.

test plan:
make test

PR checklist:
- changelog needs no update


IagoAbal:
This should allow us to close #3295, cc @emjin @brendongo.

IagoAbal:
Note that this is still taint-mode v1.0, I will create a separate PR for taint-mode v2.0 when this is merged.

aryx (approved):
Excellent.

PR #3488, opened 2021-07-08
synthesizing: Pattern_from_Targets ExprStmt lists

defreez:
Add support for targets that are ExprStmt lists in Pattern_from_Targets.

For example, two separate targets might be

```
x = req.query.foo;
exec(x);
```

and

```
y = req.query.foo;
exec(y);
```

yielding the pattern

```
$X = req.query.foo;
exec($X);
```

Limitations:
- Targets must be of equal length. Cannot yet generate ellipsis for
statements to equalize length of targets.
- No additional support for statements other than ExprStmt.

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/synthesizing/Pattern_from_Targets.ml
+++ b/semgrep-core/src/synthesizing/Pattern_from_Targets.ml
@@ -175,6 +175,25 @@ let lookup_pattern pattern s = Set.mem (p_any pattern) s
 
 let set_prev env prev' = { env with prev = prev' }
 
+(* Tranposes a list of lists, must be rectangular. *)
+let rec transpose (list : 'a list list) : 'a list list =
+  match list with
+  | [] -> []
+  | [] :: xss -> transpose xss
+  | (x :: xs) :: xss ->
+      (x :: List.map List.hd xss) :: transpose (xs :: List.map List.tl xss)
+
+(* We can't handle lists of statements of unequal size yet.
+ * Check that each target has the same number of statements.
+ *)
+let check_equal_length (targets : 'a list list) : bool =
+  match targets with
+  | [] -> true
+  | _ ->
+      let lengths = List.map List.length targets in
+      let hdlen = List.hd lengths in
+      List.for_all (( == ) hdlen) lengths
+
 (*****************************************************************************)
 (* Pattern generation *)
 (*****************************************************************************)
@@ -423,10 +442,7 @@ let rec generate_patterns_help (target_patterns : pattern_instrs list) =
     pr2 "target patterns";
     show_pattern_sets target_patterns );
   let pattern_children =
-    List.map
-      (fun patterns ->
-        List.map (fun pattern -> get_one_step_replacements pattern) patterns)
-      target_patterns
+    List.map (List.map get_one_step_replacements) target_patterns
   in
   (* Keep only the patterns in each Sn that appear in every other OR *)
   (* the patterns that were included last time, don't have children, and have another replacement to try *)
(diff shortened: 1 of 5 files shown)


emjin:
See https://github.com/returntocorp/semgrep/tree/develop/doc#development-workflow for how to make the lint errors go away.

defreez:
Thanks @emjin I have the pre-commit installed now.

semgrep-core/src/synthesizing/Pattern_from_Targets.mli
```diff
@@ -55,8 +55,42 @@
...
+ * To accomplish this, the metavariable environment is retained between
+ * target sets. The metavariable count and mappings established when
+ * the pattern $X = a is generated for for [x = a; y = a] is passed on to the
+ * starting patterns when the pattern exec($X) is being generated from
+ * [foo(x); foo(y)].
+ *
+ * For now each target must be the same length (no ellipsis support).
  *)
 
 (* limited to expressions for now *)
```
emjin, line 94: No longer true :D

semgrep-core/src/synthesizing/Range_to_AST.ml
```diff
@@ -30,6 +31,9 @@ exception Found of G.any
 
 exception FoundExpr of G.expr
 
+(* Returns range of tokens in AST. *)
+let range_of_ast = V.ii_of_any $ Range.range_of_tokens
```
emjin: @aryx I don't think we tend to use this operator in our codebase?

semgrep-core/src/synthesizing/Synthesizer.ml
```diff
@@ -7,23 +7,33 @@ let range_to_ast file lang s =
     Parse_target.parse_and_resolve_name_use_pfff_or_treesitter lang file
   in
   if errors <> [] then failwith (spf "problem parsing %s" file);
-  let a_opt = Range_to_AST.any_at_range r ast in
+  let a_opt = Range_to_AST.many_at_range r ast in
```
emjin, line 10: Is this a preemptive change? It looks like you later just take one element

semgrep-core/src/synthesizing/Pattern_from_Targets.ml
```diff
@@ -437,15 +453,15 @@ let rec generate_patterns_help (target_patterns : pattern_instrs list) =
...
-(*****************************************************************************)
+let extract_pattern (pats : pattern_instr) : Pattern.t =
+  (fun (_, pattern, _) -> pattern) pats
 
-let generate_patterns config s lang =
-  global_lang := lang;
-  (* Start each target node any as [$X, [ any, fun x -> x ]] *)
+(* Start each target node any as [$X, [ any, fun x -> x ]] *)
+let generate_starting_patterns config (targets : AST_generic.any list list) :
+    pattern_instrs list list =
```
emjin, line 464: Similarly to comment in mli, I don't think this needs to be a `list list`

semgrep-core/src/synthesizing/Pattern_from_Targets.mli
```diff
@@ -55,8 +55,42 @@
...
 val generate_patterns :
-  Config_semgrep.t -> AST_generic.any list -> Lang.t -> Pattern.t list
+  Config_semgrep.t -> AST_generic.any list list -> Lang.t -> Pattern.t list
```
emjin, line 96: I don't think we should ever need this to be a list list. A Pr program is an any, and other statement lists are generally contained within blocks.

emjin (changes requested):
Looks pretty good! I think, instead of just always expecting an `AST_generic.any list list` in generate patterns, it would be better to take programs (and blocks, though you can start by separating them) and compare them in the element-by-element fashion. This will also set you up better for when you have to handle anything nested.

defreez:
Thanks for the review :) Agreed that the `list list` interface is clumsy. I'll update it, hopefully by tomorrow.

aryx (approved):
This looks great.
I think we can merge this.

semgrep-core/src/synthesizing/Range_to_AST.ml
aryx: What do you mean? The $?

aryx: I tend to prefer to put the parameters instead of using https://wiki.haskell.org/Pointfree style.

aryx:
The next steps are to remove those list list like @emjin suggested and start writing some
test infrastructure where we can add some code examples and assert the resulting pattern has a certain form.
Something like in Unit_synthesizer.ml

where you can add a test like

let p1 = "
x = req.query.foo;
exec(x);
"
in
let p2 = "
y = req.query.foo;
exec(y);
"
in
let res = do_the_magic Lang.Java [p1; p2] in
let expected = "
$X = req.query.foo;
exec($X);
"
in
assert_same_ast "it should match" res expected

aryx (approved):

PR #3491, opened 2021-07-08
Join rules

minusworld:
PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   `...` that are added to record patterns, plus allow matching "spread fields"
   (JS `...x`) at any position (#3120)
 - Support globstar (`**`) syntax in path include/exclude (#3173)
+- A new experimental 'join' mode. This mode runs multiple Semgrep rules
+  on a codebase and "joins" the results based on metavariable contents. This
+  creates the ability to ask questions of codebases like "do any 3rd party
+  libraries use a dangerous function, and do I import that library directly?"
+  with several Semgrep rules.
 
 ### Fixed
 - Apple M1: Semgrep installed from HomeBrew no longer hangs (#2432)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Added
+- A new experimental 'join' mode. This mode runs multiple Semgrep rules
+  on a codebase and "joins" the results based on metavariable contents. This
+  lets users ask questions of codebases like "do any 3rd party
+  libraries use a dangerous function, and do I import that library directly?" or
+  "is this variable passed to an HTML template, and is it rendered in that template?"
+  with several Semgrep rules.
+
 ## [0.58.1](https://github.com/returntocorp/semgrep/releases/tag/v0.58.1) - 2021-07-15
 
 ### Fixed
@@ -30,11 +38,6 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   `...` that are added to record patterns, plus allow matching "spread fields"
   (JS `...x`) at any position (#3120)
 - Support globstar (`**`) syntax in path include/exclude (#3173)
-- A new experimental 'join' mode. This mode runs multiple Semgrep rules
-  on a codebase and "joins" the results based on metavariable contents. This
-  creates the ability to ask questions of codebases like "do any 3rd party
-  libraries use a dangerous function, and do I import that library directly?"
-  with several Semgrep rules.
 
 ### Fixed
 - Apple M1: Semgrep installed from HomeBrew no longer hangs (#2432)
--- a/semgrep/semgrep/join_rule.py
+++ b/semgrep/semgrep/join_rule.py
@@ -1,0 +1,296 @@
+import json
+import logging
+import sys
+import tempfile
+from collections import defaultdict
+from enum import Enum
+from functools import reduce
+from pathlib import Path
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Tuple
+
+import attr
+import peewee as pw
+from ruamel.yaml import YAML
+
+import semgrep.semgrep_main
+from semgrep.config_resolver import Config
+from semgrep.config_resolver import resolve_config
+from semgrep.rule import Rule
+
+yaml = YAML()
+
+logger = logging.getLogger(__file__)
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler(stream=sys.stderr)
+handler.setFormatter(
+    logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+)
+logger.addHandler(handler)
+
+# TODO: wire up to main semgrep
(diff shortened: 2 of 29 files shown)


semgrep/semgrep/join_rule.py
```diff
@@ -0,0 +1,404 @@
...
+from semgrep.error import FATAL_EXIT_CODE
+from semgrep.error import Level
+from semgrep.error import SemgrepError
+from semgrep.pattern_match import PatternMatch
+from semgrep.rule import Rule
+from semgrep.rule_match import RuleMatch
+
+yaml = YAML()
+
+logger = logging.getLogger(__file__)
```
brendongo: Let's reuse the same loggers the rest of the code uses: https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/core_runner.py#L43 and instantiated two lines below that

semgrep/semgrep/join_rule.py
```diff
@@ -0,0 +1,404 @@
...
+        k = key(item)
+        dd[k].append(item)
+    return dd
+
+
+def camel_case(s: str) -> str:
+    return "".join(c for c in s.title() if c.isalnum())
+
+
+class Operator(Enum):
```
brendongo: We have another Operator class elseswhere so probably better for sanity to just have this be JoinOperator

semgrep/semgrep/join_rule.py
```diff
@@ -0,0 +1,404 @@
...
+    parser.add_argument("target", default=".")
+    args = parser.parse_args()
+
+    with Path(args.config).open("r") as fin:
+        contents = yaml.load(fin)
+
+    for rule in contents.get("rules", []):
+        matches, errors = main(rule, list(Path(args.target).rglob("*")))
+        print(matches)
+        print(errors)
```
brendongo: Is it worth keeping this?

semgrep/semgrep/join_rule.py
```diff
@@ -0,0 +1,404 @@
...
+        return [], parsed_errors
+
+    # Rename metavariables with user-defined renames.
+    rename_metavars_in_place(results, refs_lookup)
+
+    # Create a model map. This allows dynamically creating DB tables based
+    # on Semgrep's results. There is one table for each rule ID.
+    model_map = create_model_map(results)
+    db.connect()
+    db.create_tables(model_map.values())
```
brendongo, line 425: Should we be deleting the database on exit/error? or are in memory dbs automatically deleted on finish?

semgrep/semgrep/join_rule.py
```diff
@@ -0,0 +1,404 @@
...
+            for condition_string in join_contents.get("on", [])
+        ]
+    except InvalidConditionError as e:
+        return [], [e]
+
+    # Run Semgrep
+    with tempfile.NamedTemporaryFile() as rule_path:
+        yaml.dump({"rules": [rule.raw for rule in config_map.values()]}, rule_path)
+        rule_path.flush()
+        output = semgrep.semgrep_main.invoke_semgrep(
```
brendongo, line 367: In the future we can probably be smarter and run the rules that need to be run along with all the other rules and just pass results to join but this is fine for experimental feature :D

semgrep/semgrep/join_rule.py
```diff
@@ -0,0 +1,404 @@
...
+                    .get("metavars", {})
+                    .items()
+                },
+            )
+
+
+def main(
+    join_rule: Dict[str, Any],
+    targets: List[Path],
+) -> Tuple[List[RuleMatch], List[SemgrepError]]:
```
brendongo, line 306: Can we rename from `main` probably `run_join_rules` and add function level comment that this expects a single dictionary join rule and runs all relevant subrules with semgrep and joins as necessary

semgrep/semgrep/join_rule.py
```diff
@@ -0,0 +1,404 @@
...
+
+    assert isinstance(output, dict)  # placate mypy
+
+    results = output.get("results", [])
+    errors = output.get("errors", [])
+
+    parsed_errors = []
+    for error_dict in errors:
+        try:
+            del error_dict["code"]
```
brendongo, line 392: why do these keys need to be deleted?

semgrep/semgrep/join_rule.py
```diff
@@ -0,0 +1,404 @@
...
+        metavars = set()
+        for finding in findings:
+            metavars.update(finding.get("extra", {}).get("metavars", {}).keys())
+        model_fields = ["path"] + list(metavars)
+        model_class = model_factory(camel_case(name), model_fields)
+        model_map[name] = model_class
+    return model_map
+
+
+def populate_data(
```
brendongo: Would help to have more descriptive function name/comments. Maybe something like `load_results_into_db`.

semgrep/semgrep/semgrep_main.py
```diff
@@ -255,6 +257,12 @@ def main(
             )
             optimizations = "none"
 
+    join_rules, rest_of_the_rules = partition(
+        lambda rule: rule.raw.get("mode") == JOIN_MODE,
```
brendongo: does rule.mode not give us what we want here?

semgrep/semgrep/join_rule.py
minusworld: I assumed they were deleted when the program closed, but I can explicitly `.close()` which definitely removes the in-memory database

semgrep/semgrep/join_rule.py
minusworld: Yah, it was easier to make a different code path using the unofficial API 😅

semgrep/semgrep/join_rule.py
minusworld: This was a hack to reconstitute errors after they've been JSONified as output. `SemgrepError` subclasses define the `level` and `code` as class properties, which means they aren't accepted as arguments when instantiated. They'll still be properly populated because it's using the class properties according to the instantiated class.

I'll add a comment explaining this.

I think a more proper implementation to avoid having to reconstitute errors would be to return the actual result and error objects from the Semgrep API instead of JSON. Or add a flag indicating which is preferred.  But in the meantime, I figured this was... well, it works. 😅

aryx (commented):
Impressive!
The semgrep Python wrapper is back in the game!

minusworld:
> Impressive!
> The semgrep Python wrapper is back in the game!

Hah! I didn't want to resurrect it, but it's where I work fastest 😅

semgrep/semgrep/semgrep_main.py
minusworld: I think I thought that `rule.mode` may not be present if `mode:` wasn't explicitly present in the YAML file, but it looks like it gets automatically populated. I added a new e2e test that checks multiple rule modes in the same file to make sure it works as expected

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   using arbitrary pattern formulas. This provides plenty of flexibility. Note
   that we breaks compatibility with the previous taint-mode format, e.g.
   `- source(...)` must now be written as `- pattern: source(...)`.
+- A new experimental 'join' mode. This mode runs multiple Semgrep rules
```
brendongo: Need to update to move to Unreleased changelog now that 0.58.0 is released

brendongo (approved):

Issue #3297, opened 2021-06-07
Implement in semgrep-core or deprecate pattern-where-python 


emjin:
Currently, we're going through `--optimizations none` for pattern-where-python. Either implement some version of this or deprecate it altogether

Issue #3296, opened 2021-06-07
Implement debugging-json in semgrep-core


emjin:
This will allow us to properly turn on optimizations all

brendongo:
Playground will run with optimizations off unless the debugger is on so not high priority anymore but still necessary before we delete old code paths

emjin:
Still something we want to do, though perhaps in a different form

emjin:
.

aryx:
@zyannes was working on it during his hack week.

aryx:
Solved! https://github.com/returntocorp/semgrep/pull/5898

Issue #3049, opened 2021-05-06
RFC: deprecate and remove user-defined 'equivalences' functionality


mschwager:
https://semgrep.dev/docs/experiments/overview/#equivalences

This functionality has went mostly unused for a year+. In order to reduce our maintenance burden let's deprecate and remove it. We'll always have the git history if we'd like to revive it at a later date.

underyx:
I think this would’ve solved many people’s complaints about how they have to copy-paste variants of the same patterns when writing rules, I’d probably have recommended its use if I knew I about the feature

mschwager:
Some additional context:

> It is a nice feature. I wonder if we can provide approximately the same functionality for users through builtin equivalences and taint? Most of the user-defined equivalences I see in rules are just describing a list of sources and sinks. Of course there are a few that aren't, like dict.get and dict[...], but I think the vast majority are. I think with some user input on how they'd like to describe this type of functionality we could revive it at a later date. The spirit of the functionality is definitely right, I wonder if the ergonomics are a bit off?

> I thought user-defined equivalence would be useful for matching over artihmetic expressions where you can write the same thing in many different ways because of the communtatitivity and associactivity of those operators, but nobody write rules for that :slightly_smiling_face: there's no security bugs in arithmetic expressions.

> Integer overflow! But we don't have many people writing C rules yet :)

> I think this would’ve solved many people’s complaints about how they have to copy-paste variants of the same patterns when writing rules, I’d probably have recommended its use if I knew I about the feature

> Yeah I think the spirit of the functionality is right, but the ergonomics may be a bit off. That and we have quite a few bugs: https://github.com/returntocorp/semgrep/issues?q=is%3Aissue+is%3Aopen+label%3Apattern%3Aequivalence. Really we can only support so much experimental functionality at a time, and after a certain point we should either promote things to more official support, or remove them

aryx:
subscribe

mschwager:
To start, I'm thinking of adding some kind of [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) when we detect `equivalences` use. We could then warn about the deprecation for 2-4 releases, link to this issue in the message, give users some time to voice their concerns, then remove the functionality from the Python interface if there are minimal objections.

aryx:
sounds good to me.

dlukeomalley:
This feature has been deprecated 🙏

PR #3493, opened 2021-07-08
Add deprecation notice if using any experimental features not in optimized semgrep-core

brendongo:
Returns to using `--optimizations none` when running any rule with equivalences.

Informs user that the following features will be deprecated by 0.60.0 if we detect they are using them
- pattern-where-python
- taint-mode
- equivalences
- step-by-step evaluation output

### DIFF
--- a/semgrep/semgrep/semgrep_main.py
+++ b/semgrep/semgrep/semgrep_main.py
@@ -12,6 +12,7 @@ from typing import Tuple
 from typing import Union
 
 import attr
+from colorama import Fore
 
 from semgrep.autofix import apply_fixes
 from semgrep.config_resolver import get_config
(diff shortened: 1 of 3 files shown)


brendongo:
Closes https://github.com/returntocorp/semgrep/issues/3297, https://github.com/returntocorp/semgrep/issues/3296, https://github.com/returntocorp/semgrep/issues/3049

emjin:
lol you literally didn't change anything (and some of these are even on slower baselines)

brendongo:
I made rules with equivalences run without optimizations

brendongo:
We should probably finish https://github.com/returntocorp/semgrep-rules/issues/1365 before getting this in

brendongo:
Going to hold off since https://github.com/returntocorp/semgrep-rules/issues/1365 looks like we still have 4 rules that use `pattern-where-python`

brendongo:
Looks like none of the official rulesets rely on these experimental features anymore so going ahead and adding the deprecation notice

Issue #3199, opened 2021-05-26
Taint Mode Features Request


erwanlr:
**Is your feature request related to a problem? Please describe.**
1. ~~When playing around with the taint mode in PHP, I came across a case where I need to specify in which argument position the source should be (either in `pattern-sinks` or `pattern-sanitizers`) - https://semgrep.dev/s/pljJ~~ There is a trick to do that: https://github.com/returntocorp/semgrep/issues/3199#issuecomment-908635051

2. ~~Furthermore, there are some cases where I need to only do the taint analysis on specific pattern source and sink format, which so far is not possible~~ Possible since semgrep 0.58.0

3. Sometimes I also need to have a source, like a method with a specific first argument or type or argument (like array, or anything non array)

4. ~~When the source and sink are not on the same line, the output only display the matching sink, would be nice to also have the matching source to make it easier to spot the good stuff, or False Positive w/o having to go to the actual source and find it manually.~~ Would also be nice to be able to have a special keyword/way to output the matching source in the message (this is related to my issue of not being able to quickly see the matching source in the output)

**Describe the solution you'd like**
1. ~~Be able to specify where the source should be in the sinks and sanitizers method arguments, for example, a reserved keyword such as `$SOURCE` could be used: `- sink1($SOURCE, ...)`, `- sanitise_method(..., $SOURCE)`~~ See trick at https://github.com/returntocorp/semgrep/issues/3199#issuecomment-908635051
2.
   1. ~~Support patterns in `pattern-sources` and `pattern-sinks`. It seems it's planned for the future, https://github.com/returntocorp/semgrep/issues/3191#issuecomment-848255785, but support of a basic pattern in them would already be a huge deal  ^_^~~ Supported by Semgrep 0.58.0
   2. ~~Support metavariable-regex in the patterns, right now in the source would be more than enough for my need~~ Supported by Semgrep 0.58.0

3. 
   1. Be able have have a source such as `something([...])` match `$v = array('1'); something($v);` and not just `something(array('1'))`, if it could also match things like `$v = array_merge($a1, $a2);` that would be awesome but I don't mind to have to add another source like `something(array_merge(…))` to handle that myself (as long as the propagation is correctly handled)
   2. Related to the above one, but it's opposite: be able to have a source with an argument not being an array for example

4.
   1. ~~Display both the matching source and sink in the output when they are not on the same line, some like~~
       ```php
       $var = $_GET['hello'];
       ...
       echo $var;
       ```
[...]

PR #3495, opened 2021-07-09
taint-mode: Specify sources/sanitizers/sinks using pattern formulas

IagoAbal:
We are no longer restricted to single instruction/expression patterns.

Goodbye fake taint?

Helps #3199

test plan:
make test # tests included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Tainting_generic.ml
+++ b/semgrep-core/src/engine/Tainting_generic.ml
@@ -18,7 +18,6 @@
 module AST = AST_generic
 module V = Visitor_AST
 module R = Rule
-module R2 = Mini_rule
 module PM = Pattern_match
 
 (*****************************************************************************)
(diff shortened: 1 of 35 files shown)


aryx (approved):
Fantastic!

semgrep-core/tests/tainting_rules/go/zip-traversal.yaml
```diff
@@ -0,0 +1,26 @@
...
+        - pattern: $READER.File
+        - pattern-inside: |
+            import "archive/zip"
+            ...
+        - pattern-inside: |
+            $READER, $ERR := zip.OpenReader($ARCHIVE)
+            ...
+    pattern-sinks:
+      - patterns:
+        - pattern: $PATH
```
aryx, line 21: Will this be optimized or the recent revert of emma will prevent the optimization here?

semgrep-core/src/engine/Tainting_generic.ml
```diff
@@ -47,61 +46,40 @@ module DataflowY = Dataflow.Make (struct
...
-  match_pat_eorig pat eorig
-
-(*s: function [[Tainting_generic.match_pat_instr]] *)
-let match_pat_instr pat instr =
-  let eorig = instr.IL.iorig in
-  match_pat_eorig pat eorig
-
-(*e: function [[Tainting_generic.match_pat_instr]] *)
+let any_in_ranges any ranges =
+  let tok1, tok2 = Visitor_AST.range_of_any any in
```
aryx, line 52: Maybe add a comment that says "This is potentially slow. We may need to store
range position in the AST at some point".

semgrep-core/tests/tainting_rules/go/zip-traversal.yaml
IagoAbal: Unfortunately this won't be optimized now, we need to put the optimization back or find some other alternative if we want this to be efficient.

semgrep-core/src/engine/Tainting_generic.ml
IagoAbal: Added!

PR #3506, opened 2021-07-12
matching: Make record matching more consistent

### LINKED ISSUE #3120: [JS] Dictionary matching yield inconsistent results
**Describe the bug**
Matching JS dictionaries yield unexpected results, for example the pattern `{}` is equivalent to `{...}`, the pattern `{...$X}` matches anything and the pattern `{x: $X}` also matches any dictionary with that key.

Note that this is different to what happens in Python. (#1003 could be related)

**To Reproduce**
```js
a = {};
a = {...x};
a = {...x, ...y};
a = {...x, y: z};
a = {...x, z: 3, ...y}
a = {x: 1};
a = {x:1, y:2};
```

https://semgrep.dev/s/2bP8 (note that the test code gets overwritten when you load it up)

**Expected behavior**
- Pattern `{}` should only match the empty dictionary `{}`
- Pattern `{...$X}` should only match dictionaries of the form `{...anotherDict}`
- Pattern `{x:$X}` should only match dictionaries with just the key `x`

**Environment**
Semgrep Playground and 0.48.0 installed via homebrew.

IagoAbal:
1. Allow disabling the implicit ... for record patterns via `options:`.
2. Match "spread fields" (JS `...x`), or any other kind of field, at any
   position; same as we already do for definitions.

Closes #3120

test plan:
make test # tests included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   using arbitrary pattern formulas. This provides plenty of flexibility. Note
   that we breaks compatibility with the previous taint-mode format, e.g.
   `- source(...)` must now be written as `- pattern: source(...)`.
+- New matching option `implicit_ellipsis` that allows to disable the implicit
+  `...` that are added to record patterns, plus allow matching "spread fields"
+  (JS `...x`) at any position (#3120)
 
 ### Fixed
 - Apple M1: Semgrep installed from HomeBrew no longer hangs (#2432)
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   using arbitrary pattern formulas. This provides plenty of flexibility. Note
   that we breaks compatibility with the previous taint-mode format, e.g.
   `- source(...)` must now be written as `- pattern: source(...)`.
+- New matching option `implicit_ellipsis` that allows to disable the implicit
+  `...` that are added to record patterns, plus allow matching "spread fields"
+  (JS `...x`) at any position (#3120)
```
mjambon, line 18: I was once told it's "allows disabling" or "allows one to disable" but not "allows to disable".

mjambon (approved):
:+1:

aryx (approved):

Issue #3062, opened 2021-05-07
Fix dependencies for comby for upgrading to ocaml 4.12


mjambon:
This is for [upgrading semgrep to use ocaml 4.12](https://github.com/returntocorp/semgrep/issues/3027).

The docker build for the opam packages used by semgrep fails with the [following error](https://app.circleci.com/pipelines/github/returntocorp/ocaml-layer/59/workflows/85aa26db-2967-4606-9a65-27ecbd8b47f9/jobs/39):
![image](https://user-images.githubusercontent.com/343265/117510595-34e3a000-af41-11eb-97fc-e9b5a5579fe4.png)

aryx:
The main issue is that pfff is using ppxlib 0.15.0 and it does not work with ocaml 4.12, so I need
to update to ppxlib 0.20.0

Issue #3027, opened 2021-05-01
Upgrade build to use ocaml 4.12


mjambon:
This [homebrew PR for semgrep](https://github.com/Homebrew/homebrew-core/pull/76176) shows a bug that could be avoided by switching to ocaml-4.12.

Building docker images with all the external dependencies on it is [in progress](https://github.com/returntocorp/ocaml-layer/pull/5) and [blocked by a version constraint on `lsp`](https://github.com/returntocorp/semgrep/issues/3026).

Update: now that we use ocamlformat, I added it to the list of packages and it can't be installed (in any version). Opam produces the [following message](https://app.circleci.com/pipelines/github/returntocorp/ocaml-layer/65/workflows/8c920054-7586-4500-a4c9-aace46b789ff/jobs/45):

```
The following dependencies couldn't be met:
  - ocamlformat -> base < v0.12 -> ocaml < 4.08.0
      base of this switch (use `--unlock-base' to force)
  - ocamlformat -> ocaml < 4.12
      base of this switch (use `--unlock-base' to force)
  - ocamlformat -> ocamlformat_support < 0.2 -> ocaml < 4.08.0
      base of this switch (use `--unlock-base' to force)
  - ocamlformat -> ppxlib >= 0.18.0 -> ocaml < 4.12
      base of this switch (use `--unlock-base' to force)
Your request can't be satisfied:
  - No available version of ppxlib satisfies the constraints
```

to do:
- [x] resolve installation of `lsp`
- [x] resolve installation of `ocamlformat`
- [x] [resolve installation of `comby`](https://github.com/returntocorp/semgrep/issues/3062)

mjambon:
The lsp issue was fixed by @aryx by switching to lsp.1.3.0. We now have an [issue involving comby and its dependencies](https://github.com/returntocorp/semgrep/issues/3062).

Issue #3405, opened 2021-06-23
semgrep-core slow startup and/or slow stack trace collection and/or big binaries


mjambon:
[please add details if you have any]

On Linux, with the ocaml 4.10.0, `semgrep-core` takes about 2 seconds of active CPU time to obtain a stack trace (`Printexc.get_backtrace ()`) when an exception is caught. We don't know what causes this. It can be reproduced by adding code the `Main.ml` that simply raises an exception, catches it, calls `Printexc.get_backtrace ()`, and exits.

On MacOS, a startup time of about 2 seconds was observed, but not on Linux. We don't know what might be causing this either or if it has anything to do with the stack trace collection slowness.

We suspect this may have been introduced when we started using a library, possibly something intrusive/invisible like a ppx extension.

Plan: observe what functions get called during the 2 seconds that it takes for `Printexc.get_backtrace ()` to return. The `perf` program might be useful for this.

mjambon:
I'm not sure if this is a problem for typical use of semgrep. It could be if for example `Printexc.get_backtrace ()` is called for many files that fail to parse and that call takes a long time every time. It's possible that it's slow only the first time. More data is needed.

mschwager:
Looks related to https://github.com/returntocorp/semgrep/issues/3397.

aryx:
I'm being hit by this with https://github.com/returntocorp/semgrep/pull/3509
If you update to ppxlib 0.20.0, then suddenly Main.exe take more than 2s with ocaml 4.10.0 (also 4.09.1), but not 4.12.0.

aryx:
This is on Linux.

aryx:
By having a 2s semgrep-core startup time, the CI time goes from 4m to 40min ...

aryx:
See the CI job here: https://github.com/returntocorp/semgrep/pull/3509

aryx:
I don't have the problem with 4.10.2

aryx:
cc @IagoAbal

aryx:
if I strace semgrep-core, it looks like it's waiting 2sec in a call to
wait4(-1, ...)

aryx:
with 4.10.2, there is also this call to wait4, but it returns immediately

aryx:
I did a diff on the 2 strace (one with 4.10.0 and one with 4.10.2) but don't see anything that different. except one wait4 takes 2s and the other 0.1s

aryx:
[...]

PR #3509, opened 2021-07-13
update to ppxlib 0.20.0 and OCaml 4.10.2

aryx:
This is needed to be able to compile pfff (and so semgrep) with
OCaml 4.12.0 (which itself is needed to compile correctly on Homebrew
and recent macOS)

This should help #3027

test plan:
make test

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,6 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Support metavariables and ellipsis inside regexp literals
   (e.g., `foo(/.../)`)
 - Associative-commutative matching for bitwise OR, AND, and XOR operations
-- Add support for $...MVAR in generic patterns
 - Add support for $...MVAR in generic patterns.
 - metavariable-pattern: Add support for nested Spacegrep/regex/Comby patterns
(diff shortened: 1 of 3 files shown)


aryx:
opam switch create 4.12.0
opam install --deps-only .
seems to work now.

aryx:
This closes #3062

aryx:
We have huge perf regression in CI with this ...

aryx:
I think it's because the startup time of semgrep-core jumps to 3s with 4.10. The problem disappear with 4.12,
but we still use 4.10 in ocaml-layer and so in our Dockerfile.

aryx:
@IagoAbal @mjambon if you use this branch and opam install --deps-only .
you should also see a super slow startup time with 4.10.0 (but not 4.10.2).
I have no idea why.
I tried to use perf and strace, but could not find the problem.
If you have some ideas ...

aryx:
I'm adding debugging notes in https://github.com/returntocorp/semgrep/issues/3405

aryx:
I've updated the ocaml-layer docker image to use ocaml 4.10.2 which seems to not have this perf regression.
Crossing fingers!

PR #3512, opened 2021-07-14
Private rule upload script

brendongo:


upload/upload_private_rules.py
```diff
@@ -0,0 +1,121 @@
...
+
+def is_valid(rule: Dict[str, Any]) -> bool:
+    """
+    Returns true if the rule is valid for uploading
+
+    A valid rule definition should:
+    - only have one rule definition
+    - passes schema validation
+    """
+    if len(rule["rules"]) != 1:
```
nbrahms:
```suggestion
    if len(rule.get("rules", [])) != 1:
```

nbrahms: this would still die if rules isn't a list, of course

upload/upload_private_rules.py
```diff
@@ -0,0 +1,121 @@
...
+        sys.exit(1)
+
+    session = requests.Session()
+    session.headers["Authorization"] = f"Bearer {token}"
+
+    request_json = {
+        "definition": rule,
+        # below should always be defined if passed validation
+        "languages": rule["rules"][0]["languages"],
+        "deployment_id": deployment_id,
```
nbrahms, line 74: Do you need a path?

upload/upload_private_rules.py
```diff
@@ -0,0 +1,121 @@
...
+    response = session.post(SEMGREP_REGISTRY_UPLOAD_URL, json=request_json, timeout=30)
+
+    if not response.ok:
+        print(f"Failed to upload rule with status_code {response.status_code}")
+        print(response.text)
+        sys.exit(1)
+    else:
+        created_rule = response.json()
+        print(
+            f"You can find your private rule at {SEMGREP_REGISTRY_VIEW_URL}{created_rule['path']}"
```
nbrahms, line 86: how does this created rule get a path?

nbrahms (approved):

upload/upload_private_rules.py
brendongo: For v0 we're gonna make a simplifying assumption that there's no path and the rule name in the registry will just be orgname:rulename

upload/upload_private_rules.py
brendongo: Server side handles creating during insertion in the format `orgname:rulename`

PR #3516, opened 2021-07-14
update to ocamlformat 0.16.0

aryx:
This is needed if we want to use ppxlib 0.20 (itself needed
if we want to use OCaml 4.12.0, itself needed if we don't
want errors on macos HomeBrew)

test plan:
pre-commit run --all lint-ocaml

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,7 +22,8 @@ jobs:
 
   pre-commit-ocaml:
     runs-on: ubuntu-latest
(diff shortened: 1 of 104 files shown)


aryx:
No idea why the Lint / pre-commit-ocaml (pull_request_target)  fails but not the pull_request variant.
Also the semgrep bench failures are present in previous PRs. This PR just refactor code, so the perf regression
does not come from this PR.

PR #3522, opened 2021-07-14
Pipe semgrep-core stderr to stderr when --debug

### LINKED ISSUE #3220: Show which files are currently processed in --debug mode
context: https://r2c-community.slack.com/archives/CK86BJ5DW/p1622147935097300

right now if semgrep-core segfaults on a file, there is no easy way to know what was the file in the error messages.
People resorts to:
```
Don't know how big the repo is and if you have tried this already, but for places where semgrep was barfing on the file itself, I had reasonable success going with a simple pattern ('$X = $X') and then just trying different files and folders and file extensions in a binary search to quickly zero down the file with the issue. Moving to a simple pattern also meant that debug output was much more tractable
```
which is not good.

The --debug mode should display the rules x files pair being currently processed.

brendongo:
Closes: https://github.com/returntocorp/semgrep/issues/3220

@aryx @mjambon with this PR we pipe semgrep-core errors directly to stderr if --debug is set but because of that we don't capture that error message anymore so are not able to print the error with @mjambon's nicely formatted _fail message. So might be a little more annoying but at least the stderr will always be there even if semgrep-core loops/segfaults

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -9,6 +9,7 @@ from typing import cast
 from typing import Dict
 from typing import IO
 from typing import List
+from typing import Optional
 from typing import Set
 from typing import Tuple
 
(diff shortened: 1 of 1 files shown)


semgrep/semgrep/core_runner.py
```diff
@@ -128,25 +129,11 @@ def _extract_core_output(
         self, rule: Rule, patterns: List[Pattern], core_run: subprocess.CompletedProcess
     ) -> Dict[str, Any]:
         semgrep_output = core_run.stdout.decode("utf-8", errors="replace")
-        semgrep_error_output = core_run.stderr.decode("utf-8", errors="replace")
```
aryx, line 131: So where does the stderr output goes when in normal mode?
Where is it displayed if all this code is removed?

brendongo: Oh my bad. Yes should definitely still keep this code path

brendongo: Hmm good call. Should not have removed all of this

aryx:
BTW in my PR I usually have a 'test plan:' section where I describe how to test the functionality
the PR provides, for example here
'semgrep --debug -e foo /tmp/foo.x should output: xxxxx
`semgrep -e foo /tmp/foo.xxx should just output: yyyy

aryx (approved):

PR #3523, opened 2021-07-15
Switch back to ocp-indent to unblock ocaml upgrade

mjambon:
[DO NOT SQUASH COMMITS WHEN MERGING]

Switching back to ocp-indent for auto-reformatting of ocaml code, due to opam dependency conflicts when using ocaml 4.12.
See https://github.com/returntocorp/ocaml-layer/pull/5#issuecomment-878997912

I left the ocamlformat config in place (.ocamlformat files and ppx annotation for disabling it in some file) because it doesn't hurt and we won't have to fish it out of the git history when we switch back.

PR checklist:
- [x] changelog is up to date


aryx:
This is not needed I think, ocamlformat 0.16 works fine with ocaml 4.12.

PR #3530, opened 2021-07-15
Replace yaml parser

emjin:
Parse yaml rules into the generic AST so that we can have span information.

While doing this, save the path within the yaml file for each pattern (this will allow us to later report errors in playground basic mode).

We parse each generic AST dictionary by reading it into a Hashtbl, which we delete from as we read each field. Originally, we instead gave a list of keys we wanted and split the dictionary list of tuples into a list of values corresponding to the keys and the rest. I don't like that I mutate the Hashtbl, but I prefer my approach because it makes it easier to report specific error messages, and I'm hoping that the function names are clear enough.

PR checklist:
- [x] changelog is up to date


semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -210,200 +236,191 @@ let parse_options json =
...
-  match lang with
-  | R.L (lang, _) -> R.mk_xpat (Sem (H.parse_pattern ~id ~lang s, lang)) s
+let parse_pattern env e =
+  let s =
+    match e with
+    | G.L (String (value, _)) -> value
+    | G.N (Id ((value, _), _)) -> value
+    | _ -> error ("Expected a string value for " ^ env.id)
+  in
+  let start, end_ = Visitor_AST.range_of_any (G.E e) in
```
emjin, line 275: This is for later, but note that `start` and `end_` are currently the same (each pattern is only associated with one token). This might be really annoying to change (we need to compute an accurate `end_`, but the string given to us by the yaml parser has tabs removed). Will include a note to this effect when I make my "add ranges to patterns" PR.

semgrep-core/src/parsing/other/yaml_to_generic.ml
```diff
@@ -148,7 +148,7 @@ let make_scalar _anchor _tag pos value env : A.expr =
     let token = mk_tok pos value env in
     match value with
     | "__sgrep_ellipses__" -> A.Ellipsis (Parse_info.fake_info "...")
-    | "null" | "NULL" | "" | "Null" | "~" -> A.L (A.Null token)
```
emjin, line 151: This is just to merge this PR. I will put `""` back to `Null` and have either a warning or an error when we try to parse a string and get Null in another PR.

semgrep-core/src/cli/Main.ml
```diff
@@ -176,15 +176,6 @@ let mvars = ref ([] : Metavariable.mvar list)
 (*s: constant [[Main_semgrep_core.layer_file]] *)
 (*e: constant [[Main_semgrep_core.layer_file]] *)
 
-(*s: constant [[Main_semgrep_core.keys]] *)
```
IagoAbal, line 179: Very nice moving these functions out of Main!

semgrep-core/src/parsing/other/yaml_to_generic.ml
IagoAbal: I am not very familiar with this code so I don't know exactly why this helps this PR, but I guess @aryx will know.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -43,64 +49,107 @@ module H = Parse_mini_rule
...
-  | `Null -> J.Null
-  | `Bool b -> J.Bool b
-  | `Float f ->
-      (* or use J.Int? *)
-      J.Float f
-  | `String s -> J.String s
-  | `A xs -> J.Array (xs |> List.map yaml_to_json)
-  | `O xs -> J.Object (xs |> List.map (fun (k, v) -> (k, yaml_to_json v)))
+let generic_to_json key ast =
+  let rec generic_to_json_help = function
```
IagoAbal: It's quite common to name this kind of auxiliary functions as `aux`, `loop`, `go`, or something like that.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -43,64 +49,107 @@ module H = Parse_mini_rule
...
+          |> List.map (fun x ->
+                 match x with
+                 | G.Tuple (_, [ L (String (k, _)); v ], _) ->
+                     (k, generic_to_json_help v)
+                 | _ -> error ("Expected key value pair in " ^ key ^ " dict")))
+    | _ -> error "Unexpected generic representation of yaml"
+  in
+  generic_to_json_help ast
+
+let opt_to_list = function None -> [] | Some xs -> xs
```
IagoAbal: We could find another name for this perhaps, it clashes with `Common.opt_to_list` but does something different.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -414,144 +431,110 @@ let parse_languages ~id langs =
...
+    | FT.Config FT.Yaml -> Yaml_to_generic.program file
+    | FT.Config FT.Json ->
+        Json_to_generic.program (Parse_json.parse_program file)
     | FT.Config FT.Jsonnet ->
         Common2.with_tmp_file ~str:"parse_rule" ~ext:"json" (fun tmpfile ->
             let cmd = spf "jsonnet %s -o %s" file tmpfile in
             let n = Sys.command cmd in
             if n <> 0 then failwith (spf "error executing %s" cmd);
-            J.load_json tmpfile)
+            Json_to_generic.program (Parse_json.parse_program tmpfile))
```
IagoAbal, line 562: Do we really need to do the conversion to Generic? What is the advantage?

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -43,64 +49,107 @@ module H = Parse_mini_rule
...
+let opt_to_list = function None -> [] | Some xs -> xs
+
+(*****************************************************************************)
+(* Dict helper methods *)
+(*****************************************************************************)
+
+let yaml_to_dict name rule =
+  match rule with
+  | G.Container (Dict, (_, fields, _)) ->
+      let dict = Hashtbl.create 10 in
```
IagoAbal, line 107: Have you considered using a persistent `Map` here? It should be less prone to subtle errors.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -43,64 +49,107 @@ module H = Parse_mini_rule
...
+  in
+  generic_to_json_help ast
+
+let opt_to_list = function None -> [] | Some xs -> xs
+
+(*****************************************************************************)
+(* Dict helper methods *)
+(*****************************************************************************)
+
+let yaml_to_dict name rule =
```
IagoAbal, line 104: Could the `rule` param be called `obj` or something like that? Or am I mistaken about what that parameter is?

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -43,64 +49,107 @@ module H = Parse_mini_rule
...
+  match rule with
+  | G.Container (Dict, (_, fields, _)) ->
+      let dict = Hashtbl.create 10 in
+      fields
+      |> List.iter (fun field ->
+             match field with
+             | G.Tuple (_, [ L (String (key, _)); value ], _) ->
+                 Hashtbl.add dict key value
+             | _ -> error "Not a valid key value pair");
```
IagoAbal, line 113: Can this happen?

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -43,64 +49,107 @@ module H = Parse_mini_rule
...
+      dict
+  | _ -> error ("each " ^ name ^ " should be a dictionary of fields")
```
IagoAbal: Perhaps the error message should talk about "object" (YAML/JSON terminology) rather than "dictionary of fields" (the way we chose map a JSON object to the Generic AST)?

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -43,64 +49,107 @@ module H = Parse_mini_rule
...
+
+(* Mutates the Hashtbl! *)
+let take dict f key =
```
IagoAbal, line 120: Could this be defined in terms of `take_opt`? Just noticed that `take` removes the mapping from the hash table, but `take_opt` does not, is that intentional?

IagoAbal (approved):
It would take me a while to make sure that this new parser is equivalent to the previous one. I'll trust the tests and I hope we added enough ones to catch any errors!

I added some comments, but to summarize my main concerns are:

1. Do we need to convert `Ast_json` (well, `Ast_js`) to `AST_generic`? It feels a bit weird to convert JS/JSON to Generic, to then read JSON from Generic. Is this just to use `Visitor_AST.range_of_tokens` ? It would be easy to add `range_of_tokens` to `Visitor_js`.
2. I would consider using persistent `Map`s instead of `Hashtbl` here, the code would be more functional and less prone to errors.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -210,200 +236,191 @@ let parse_options json =
...
-      let extra = parse_extra env x in
-      R.PatExtra extra
-
-and parse_formula_new env (x : J.t) : R.formula =
+  | "metavariable-regex", _
+  | "metavariable-pattern", _
+  | "metavariable-comparison", _
+  | "pattern-where-python", _ ->
+      R.PatExtra (parse_extra env key value)
+  | _x -> error "unimplemented"
```
IagoAbal: Maybe `error (spf "unexpected key %s" _x)`?

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -210,200 +236,191 @@ let parse_options json =
...
+      let metavar = take mv_pattern_dict parse_string "metavariable" in
+      let env', opt_xlang =
+        match take_opt mv_pattern_dict parse_string "language" with
+        | Some s ->
+            let xlang = R.xlang_of_string ~id:(Some env.id) s in
+            let env' =
+              {
+                id = env.id;
+                languages = xlang;
+                path = "metavariable-pattern" :: "metavariable" :: env.path;
```
IagoAbal, line 414: Do we need `"metavariable"` in the path here?

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -43,64 +49,107 @@ module H = Parse_mini_rule
...
-(* less: could use a hash to accelerate things *)
-let rec find_fields flds xs =
-  match flds with
-  | [] -> ([], xs)
-  | fld :: flds ->
-      let fld_match = List.assoc_opt fld xs in
-      let xs = List.remove_assoc fld xs in
-      let matches, rest = find_fields flds xs in
-      ((fld, fld_match) :: matches, rest)
+type env = { id : string; languages : R.xlang; path : string list }
```
IagoAbal: Assuming we can always point to concrete line and column, do we need the `path` or is it just to provide yet another way of locating the error?

semgrep-core/src/cli/Main.ml
aryx: yep

semgrep-core/src/parsing/Parse_rule.ml
aryx: done

semgrep-core/src/parsing/Parse_rule.ml
aryx: done. I've renamed optlist_to_list

semgrep-core/src/parsing/Parse_rule.ml
aryx: I tried to explain in the "history:" comment at the top of this file.  Basically we need that to get precise location information.

semgrep-core/src/parsing/Parse_rule.ml
aryx: I've put a todo comment

semgrep-core/src/parsing/Parse_rule.ml
aryx: done

semgrep-core/src/parsing/Parse_rule.ml
aryx: done

aryx:
test plan: semgrep-core -test_rules ~/github/semgrep-rules/ seems to work!

semgrep-core/src/parsing/Parse_rule.ml
emjin: This is for the playground, because with basic mode this is a much easier way to locate the error

semgrep-core/src/parsing/Parse_rule.ml
emjin: Oh, no---this is an error! Thanks for pointing it out, I will fix it later

emjin:
> 1. Do we need to convert `Ast_json` (well, `Ast_js`) to `AST_generic`? It feels a bit weird to convert JS/JSON to Generic, to then read JSON from Generic. Is this just to use `Visitor_AST.range_of_tokens` ? It would be easy to add `range_of_tokens` to `Visitor_js`.

I agree that this is a little silly. On the other hand, maintaining two parsers may be an unnecessary duplication of code? Not sure what the best approach is.

PR #3536, opened 2021-07-16
ci: put only jobs that require secrets behind pull_request_target

brendongo:


.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,17 @@
+name: Lint
+on:
+  workflow_dispatch:
+  pull_request_target:
```
aryx, line 8: I would put a comment here, explaining why we need this instead of the classic pull_target, and also
the caveat I put before which is you can't really modify this file and get the effect immediately.

aryx (approved):

.github/workflows/semgrep.yml
brendongo: Added comment

PR #3542, opened 2021-07-16
core: engine: metavariable-pattern: Fix location info

### LINKED ISSUE #3503: pattern-not-regex doesn't work when under metavariable-pattern
**Describe the bug**

In the following (simplified) structure

```yml
      - metavariable-pattern:
          metavariable: $VALUE
          patterns:
            - pattern-not-regex: ^[0-9]+$
```

It seems like the `pattern-not-regex` is ignored. `pattern-regex` works as expected however.

**To Reproduce**

https://semgrep.dev/s/WOLj

**Expected behavior**

`pattern-not-regex` should work

**Screenshots**

**Environment**

IagoAbal:
File, file content, and Generic AST, must all have consistent location
info. This was being done when using the `language` sub-key but not
otherwise, yet it is still needed e.g. when using `pattern-regex`.

Closes #3503

test plan:
make test # tests included

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/core/ast/Map_AST.ml
+++ b/semgrep-core/src/core/ast/Map_AST.ml
@@ -1099,3 +1099,21 @@ let (mk_visitor : visitor_in -> visitor_out) =
     }
   in
   all_functions
+
+(*****************************************************************************)
+(* Fix token locations *)
+(*****************************************************************************)
+
+let mk_fix_token_locations fix =
+  mk_visitor
+    {
+      default_visitor with
+      kstmt =
+        (fun (k, _) s ->
+          k
+            {
+              s with
+              s_range = Common.map_opt (fun (x, y) -> (fix x, fix y)) s.s_range;
+            });
+      kinfo = (fun (_, _) t -> Parse_info.fix_token_location fix t);
+    }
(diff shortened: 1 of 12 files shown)


semgrep-core/src/core/ast/Map_AST.ml
```diff
@@ -1099,3 +1099,21 @@ let (mk_visitor : visitor_in -> visitor_out) =
...
+let mk_fix_token_locations fix =
+  mk_visitor
+    {
+      default_visitor with
+      kstmt =
+        (fun (k, _) s ->
+          k
+            {
+              s with
+              s_range = Common.map_opt (fun (x, y) -> (fix x, fix y)) s.s_range;
```
aryx, line 1116: good call, I would have forgotten that one.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -647,7 +647,13 @@ and satisfies_metavar_pattern_condition env r mvar opt_xlang formula =
         mvar;
       false
   | Some mval -> (
+      (* We will crete a temporary file with the content of the metavariable,
```
aryx: typo, crete -> create

aryx (approved):

semgrep-core/src/core/ast/Map_AST.ml
IagoAbal: I forgot too, but the tests reminded me about it!

PR #3545, opened 2021-07-19
synthesizing: Remove list list from PfT mli, add tests

defreez:
This a follow-up to #3488 

1. Removes the use of list list in the Pattern from Targets interface. Instead, statement lists become `AST_generic.Ss`

2. Adds test infrastructure for Pattern_from_Targets and a handful of tests.

PR checklist:
- [ ] changelog is up to date


semgrep-core/src/synthesizing/Unit_synthesizer_targets.ml
```diff
@@ -0,0 +1,114 @@
...
+open OUnit
+module R = Mini_rule
+
+let test_path = "../../../tests/OTHER/synthesizing/targets/"
+
+(* Format: file, list of target ranges, expected pattern line. *)
+let stmt_tests =
+  [
+    (* Motivating example. *)
+    ("hello.py", [ "2:0-2:5"; "5:0-5:8" ], "$X(3, ...)");
```
aryx, line 12: An alternative would be to put the content directly. It would then be easier to see
the input / expected output relation.
That would require to calls functions from semgrep-core/parsing/ to correctly parse the code into generic ASTs.

aryx (approved):
Emma is in vacation; She'll be back next week, but
this looks good to me.

semgrep-core/src/synthesizing/Range_to_AST.ml
```diff
@@ -90,16 +97,43 @@ let any_at_range r1 ast =
...
+  match anys with
+  | [] -> None
+  | [ xs ] -> Some xs
+  | G.S _ :: _ -> Some (G.Ss (List.map any_to_stmt anys))
+  | _ ->
+      failwith
+        "Unable to handle ranges that contain multiple expressions. Range must \
+         contain a single expression or one-or-more statements."
+
+let split_any (any : G.any) : G.any list =
```
emjin, line 118: Perhaps `split_stmts` would be a better name for this, since it's unclear what the node is being split on otherwise

emjin:
Thanks for adding these tests!

Issue #3538, opened 2021-07-16
[Internal] investigate why parsing stats circleCI jobs fails often


aryx:
See https://app.circleci.com/pipelines/gh/returntocorp/semgrep?branch=develop

Relevant discussion: https://returntocorp.slack.com/archives/C01NXGX2EHZ/p1626372454313200
@mjambon thinks that:
"
I remember now: some jobs on large files were taking a long time with increasing memory usage. @iago said it was due to something having to do with constant propagation. I haven't checked the progress on this since.
martin  11 hours ago
This takes place during the initialization of the generic AST, which is included in what we run for the "parsing" stats. The symptoms were:
long running time at 100% CPU usage (more than 30 min)
... during which nothing was being printed
... with slowly increasing memory usage.
We found that it was caused by some large (typically generated) function or class bodies in the target file. (edited) "

IagoAbal:
We merged returntocorp/semgrep#3434 and returntocorp/semgrep#3440 that should have helped with this, but I'll take a look.

aryx:
Ok, feel free to add more logging in the parsing stat code, so at least we can better see which files are problematic.

aryx:
Feel free to assign the task back to me if you don't find anything obvious.

IagoAbal:
There may be different causes for the failures, for now I have looked into C and there is a segfault involved:
```
parsing-stats % ../semgrep-core/bin/semgrep-core -lang c -dump_ast lang/c/tmp/lede/package/lean/mt/drivers/mt7603e/src/mt7603_wifi/ap/ap_cfg.c
zsh: segmentation fault  ../semgrep-core/bin/semgrep-core -lang c -dump_ast
```

aryx:
How did you find the name of this file? You've added logging in the cron job or you were able
to reproduce locally?

IagoAbal:
I was able to reproduce it locally, and I am trying to check each language separately with `./run-lang <lang>`. BTW I've been digging into this particular issue with the C parsing stats, seems to be caused by a stack overflow (some Parsing_hacks functions are non tail recursive and this file is somewhat large with 17k+ LOC).

IagoAbal:
We could perhaps generate a log for each language and archive those logs as artifacts.

aryx:
good idea

IagoAbal:
@mjambon in lang/hack/projects.txt we have three different `framework` projects which is causing problems:
```
https://github.com/titon/framework.git
[...]

PR #3549, opened 2021-07-19
parsing-stats: Fix some issues and log output per-lang

IagoAbal:
When a "todo" exception is raised in a tree-sitter parser we tend to
dump the whole CST. These dumps can make the `stats.json` file very
large, which hangs the `-n` test:

    -  if [[ -e "$jsonfile" && -n "$(cat "$jsonfile")" ]]; then
    +  if [[ -s "$jsonfile" ]]; then

Helps #3538

test plan:
Go to https://app.circleci.com/pipelines/gh/returntocorp/semgrep?branch=parsing-stats
and check that the parsing-stats workflow now finishes in under 40
minutes instead of timing out, plus it archives the output of running
each language suite in a separate log file.

PR checklist:
- no need for documentation update
- no need for changelog update

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,6 +55,10 @@ jobs:
           command: |
             cd parsing-stats
             ./run-all --upload
+      - store_artifacts:
+          path: parsing-stats/logs
+      - store_artifacts:
+          path: parsing-stats/results.txt
 
 workflows:
   version: 2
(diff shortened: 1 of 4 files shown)


aryx:
what is the test plan:?

aryx (approved):

IagoAbal:
Ooops, added test plan.

PR #3555, opened 2021-07-20
Use Ls module

mjambon:
Uses https://github.com/returntocorp/pfff/pull/451

PR checklist:
- [x] documentation is up to date
- [x] changelog is up to date

PR #3565, opened 2021-07-21
Improve PR warnings about performance

mjambon:
Improve PR warnings about performance:
- use min of two identical runs for each benchmark (was using the max in one instance)
- show which benchmark produced unexpected results (kinda: shows just the number but at least won't give the impression that it represents an average across all benchmarks)
- show warnings for performance improvements
- show errors in PR comments
- produce only one PR message for all the benchmarks instead of one per benchmark
- fail based on avg(latest/baseline) instead of avg(latest)/avg(baseline), for balance.
- emojis!

I tested locally with:
```
~/semgrep $ python3 ./scripts/compare_perf.py baseline1 baseline2 latest1 latest2 xxx
```
where `baseline1` `baseline2` `latest1` `latest2` contain one number per line:
```
21:20:21 martin@mj2 [mj-slowdown-percent] ~/semgrep $ cat baseline1
1.0
2.0
4.0
99
21:20:27 martin@mj2 [mj-slowdown-percent] ~/semgrep $ cat baseline2
1.02
2.01
7.7
95
21:20:29 martin@mj2 [mj-slowdown-percent] ~/semgrep $ cat latest1
0.99
2.17
3.2
167
21:20:33 martin@mj2 [mj-slowdown-percent] ~/semgrep $ cat latest2
0.98
2.18
3.1
180
```
Output:
```
$ python3 ./scripts/compare_perf.py baseline1 baseline2 latest1 latest2 xxx
Reading baseline1
Reading baseline2
Reading latest1
Reading latest2
[0] 0.980x Baseline: 1.000, Latest: 0.980
[1] 1.085x Baseline: 2.000, Latest: 2.170
[2] 0.775x Baseline: 4.000, Latest: 3.100
[3] 1.758x Baseline: 95.000, Latest: 167.000
Average: 1.149x, Min: 0.775x, Max: 1.758x
Total Baseline: 102.0 s, Latest: 173.25 s
Sending warnings and errors as a PR comment:
🔥 Potential speedup in benchmark #2: -22.5%

🚫 Benchmark #3 is too slow: +75.8%

4 benchmarks, 14.9% slower on average.

  File "./scripts/compare_perf.py", line 147, in <module>
    main()
  File "./scripts/compare_perf.py", line 140, in main
    assert not errors
AssertionError
```

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/scripts/compare_perf.py
+++ b/scripts/compare_perf.py
@@ -56,33 +56,87 @@ def main() -> None:
     baseline_times = zip(
         read_timing(baseline_timing_file_1), read_timing(baseline_timing_file_2)
     )
-    baseline_times = [max(t1, t2) for t1, t2 in baseline_times]
     latest_times = zip(
         read_timing(latest_timing_file_1), read_timing(latest_timing_file_2)
     )
+    # Eliminate "flukes" by taking the fastest time available for each
+    # benchmark.
+    baseline_times = [min(t1, t2) for t1, t2 in baseline_times]
     latest_times = [min(t1, t2) for t1, t2 in latest_times]
 
+    # Accumulators
+    n = 0
     total_baseline = 0.0
     total_latest = 0.0
-    for baseline_time, latest_time in zip(baseline_times, latest_times):
-        print(f"Baseline: {baseline_time}, Latest: {latest_time}")
+    total_rel_dur = 0.0
+    min_rel_dur = 1.0
+    max_rel_dur = 1.0
 
-        # Assert latest time is not more than 20% slower than baseline
-        # or is within a fixed "probably environmental" range
-        assert latest_time < baseline_time * 1.2 or latest_time - baseline_time < 5.0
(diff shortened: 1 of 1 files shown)


mjambon:
Merging this. I want to see how it looks like on that branch: https://github.com/returntocorp/semgrep/pull/3555

emjin:
This is awesome! I like seeing the speedups :D I originally took the max of the baseline vs the min of the new code so that we would have some more leeway, but min of both provides a more equal comparison.

mjambon:
@emjin I think the distribution of times isn't a bell curve, it's more like one sharp peak (a narrow bell curve I guess) + sparse points to its right due to accidental competition from other processes. i.e. there's a hard minimum but no hard maximum. That's why I think the minimum is a reasonable estimate of the ideal performance, even though in general the minimum doesn't make a good estimator in statistics.

PR #3569, opened 2021-07-21
Show benchmark names in the automatic PR warnings

mjambon:
Now we'll see which benchmark fails. This changes the output format of 'run-benchmarks --save-to xxx' to json.

Test plan: run `./scripts/test-compare-perf` and inspect the output (it should fail from a fatal threshold being exceeded)

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/scripts/compare_perf.py
+++ b/scripts/compare_perf.py
@@ -1,6 +1,7 @@
 import json
 import os
 import sys
+from collections import namedtuple
 from pathlib import Path
 from typing import List
 
@@ -36,10 +37,23 @@ def send_comment(message: str, github_token: str, pull_request_number: str) -> N
     r.raise_for_status()
 
 
-def read_timing(filename: Path) -> List[float]:
+Timing = namedtuple("Timing", "name time")
+
+
+def read_timing(filename: Path) -> List[Timing]:
     print(f"Reading {filename}")
-    times = filename.read_text()
-    return [float(t) for t in times.split()]
+    with open(filename) as f:
+        return [Timing(a["name"], a["time"]) for a in json.load(f)]
+
+
+def select_min_timings(al: List[Timing], bl: List[Timing]) -> List[Timing]:
+    res = []
+    for a, b in zip(al, bl):
+        if a.time < b.time:
+            res.append(a)
+        else:
+            res.append(b)
+    return res
 
 
 def main() -> None:
@@ -53,16 +67,14 @@ def main() -> None:
     # Note this only defined in pull requests
     pull_request_number = sys.argv[6] if len(sys.argv) >= 7 else ""
 
-    baseline_times = zip(
+    # Eliminate "flukes" by taking the fastest time available for each
+    # benchmark that ran twice.
+    baseline_times = select_min_timings(
         read_timing(baseline_timing_file_1), read_timing(baseline_timing_file_2)
     )
-    latest_times = zip(
+    latest_times = select_min_timings(
         read_timing(latest_timing_file_1), read_timing(latest_timing_file_2)
     )
-    # Eliminate "flukes" by taking the fastest time available for each
-    # benchmark.
-    baseline_times = [min(t1, t2) for t1, t2 in baseline_times]
-    latest_times = [min(t1, t2) for t1, t2 in latest_times]
 
     # Accumulators
     n = 0
@@ -77,8 +89,10 @@ def main() -> None:
     messages = []
     errors = 0
 
-    for baseline_time, latest_time in zip(baseline_times, latest_times):
-        i = n
+    for baseline, latest in zip(baseline_times, latest_times):
+        name = baseline.name
+        baseline_time = baseline.time
+        latest_time = latest.time
         n += 1
         total_baseline += baseline_time
         total_latest += latest_time
@@ -89,7 +103,7 @@ def main() -> None:
         max_rel_dur = max(max_rel_dur, rel_dur)
 
         print(
-            f"[{i}] {rel_dur:.3f}x "
+            f"[{name}] {rel_dur:.3f}x "
             f"Baseline: {baseline_time:.3f}, Latest: {latest_time:.3f}"
         )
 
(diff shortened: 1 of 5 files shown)


aryx (approved):

aryx:
This PR does not modify any semgrep code, and still we got those messages:
```
fire Potential speedup in benchmark semgrep.bench.coinbase.std: -15.1%

fire Potential speedup in benchmark semgrep.bench.njs-old-dvna.std: -11.2%

fire Potential speedup in benchmark semgrep.bench.t00sh.std: -15.1%

15 benchmarks, 6.8% faster on average.

```
That's noise. We should not get anything.

aryx:
also if the differences is from 2.8s to 3.2, or 2.5, I think it does not mean anything given how little those numbers are.
If the time is less < 3s, we should have a higher barrier (20% difference?) because there's lots of noise otherwise.

scripts/compare_perf.py
```diff
@@ -119,7 +134,7 @@ def main() -> None:
         f"Min: {min_rel_dur:.3f}x, "
         f"Max: {max_rel_dur:.3f}x"
     )
-    print(f"Total Baseline: {total_baseline} s, Latest: {total_latest} s")
+    print(f"Total Baseline: {total_baseline:.3f} s, Latest: {total_latest:.3f} s")
```
aryx, line 151: below there is "See run output for more details."
But in the bot-actions messages I get, where do I find this "run output"?
A link would be great.

scripts/compare_perf.py
```diff
@@ -99,13 +113,14 @@ def main() -> None:
         # or is within a fixed "probably environmental" range
         if latest_time > baseline_time * 1.2 and latest_time - baseline_time > 5.0:
             errors += 1
-            messages.append(f"🚫 Benchmark {i} is too slow: " f"+{perc:.1f}%")
+            messages.append(f"🚫 Benchmark {name} is too slow: " f"+{perc:.1f}%")
         elif rel_dur > 1.1:
             messages.append(
-                f"⚠ Potential non-blocking slowdown in benchmark {i}: " f"+{perc:.1f}%"
+                f"⚠ Potential non-blocking slowdown in benchmark {name}: "
```
aryx: maybe also give the actualy time, because 6% slower of 2.5s means nothing I think.
There are lots of noises in how each benchmark is run.

mjambon: What do you make of this:
```
[semgrep.bench.zulip.std] 0.923x Baseline: 6.290, Latest: 5.804
[semgrep.bench.dropbox.std] 0.968x Baseline: 1.789, Latest: 1.732
[semgrep.bench.coinbase.std] 0.849x Baseline: 11.051, Latest: 9.378
[semgrep.bench.apache.std] 0.989x Baseline: 11.496, Latest: 11.375
[semgrep.bench.0c34.std] 0.992x Baseline: 1.697, Latest: 1.683
[semgrep.bench.lodash.std] 0.959x Baseline: 6.095, Latest: 5.842
[semgrep.bench.njs-old-dvna.std] 0.888x Baseline: 10.246, Latest: 9.099
[semgrep.bench.DVWA.std] 0.901x Baseline: 8.323, Latest: 7.498
[semgrep.bench.juice-shop.std] 0.947x Baseline: 17.133, Latest: 16.232
[semgrep.bench.Vulnerable-Flask-App.std] 0.913x Baseline: 9.370, Latest: 8.555
[semgrep.bench.pallets.std] 0.938x Baseline: 5.542, Latest: 5.198
[semgrep.bench.socketio.std] 0.954x Baseline: 3.794, Latest: 3.618
[semgrep.bench.coolMenu.std] 0.946x Baseline: 2.254, Latest: 2.133
[semgrep.bench.t00sh.std] 0.849x Baseline: 4.027, Latest: 3.419
[semgrep.bench.grpc.std] 0.964x Baseline: 3.343, Latest: 3.224
Average: 0.932x, Min: 0.849x, Max: 1.000x
Total Baseline: 102.451 s, Latest: 94.790 s
```
semgrep.bench.t00sh.std is 3.4 s vs. 4.0 s. This is huge. Also, all the 15 benchmarks were faster than the baseline. It's extremely unlikely that it's due to chance if the benchmarks use the same binary and all run in the same clean environment. But:
* Innocuous differences in the source code can cause huge differences in the performance of the binary. Note that we switched from byte code to native last week.
* There may be some competing jobs running on the GHA instance at the same time.

scripts/compare_perf.py
mjambon: It doesn't seem to be possible to [create a link to the job](https://github.community/t/github-actions-url-to-run/16029) without using the GitHub API, which is overkill. I'll change the message a bit.

mjambon:
> also if the differences is from 2.8s to 3.2, or 2.5, I think it does not mean anything given how little those numbers are.
> If the time is less < 3s, we should have a higher barrier (20% difference?) because there's lots of noise otherwise.

I'm eliminating the silencing of warnings if the time difference is less than 5 seconds, because it just means we run the benchmark for nothing.

If it helps to make benchmarks longer than X seconds (which I doubt), then let's just make sure they're long enough or remove them.

mjambon:
New thresholds:
* individual warning: +20% (was +10%)
* individual speedup warning: -20% (was -10%)
* individual error: +30% (was +20%, silenced if difference was less than 5 seconds)
* global error: +7% (was +6%)

Sample PR comment obtained from command line:

⚠️  Potential non-blocking slowdown in benchmark semgrep.bench.b.std: +20.5% (+0.410 s)

🔥 Potential speedup in benchmark semgrep.bench.c.std: -22.5% (-0.900 s)

🚫 Benchmark semgrep.bench.d.std is too slow: +75.8% (+72.000 s)

4 benchmarks, 17.9% slower on average.

Individual deviations greater than 20% from the baseline are reported. An individual performance degradation of over 30% or an average degradation of over 7% is an error and will block the pull request. See run output for full results ('Show all checks' > 'Tests / semgrep benchmark tests' 'Details').

mjambon:
Merging this so it's available to the work in progress at https://github.com/returntocorp/semgrep/pull/3555

emjin:
We can change the benchmarks that are being run. Also, should we start running fewer benchmarks more times? Maybe an average of 5 would be more stable.

mjambon:
Recent results (possibly since the switch to native code) suggest that some large deviations will persist regardless of how many times we run them (we run them twice already). These variations are unpredictable but reproducible for a specific binary. Last week I saw these unexpected changes (where the source code only changed in ways that shouldn't affect performance):
* all 15 benchmarks were faster than the baseline (around 0.9x)
* 14 benchmarks about as fast as the baseline and 1 benchmark 30-40% slower

I expect to see a lot more oddities. We should only worry if the performance is really worse for all the benchmarks. The overall error threshold is now set at +7% of the baseline (= average degradation is >= 7%). We might have to increase it even more, we'll see.

PR #3574, opened 2021-07-21
Detect duplicate key in YAML dictionary in our semgrep rule

aryx:
This will help avoid silly mistakes.

Test plan:
test file included, also
```
[pad@yrax yy (check_duplicate_field)]$ yy -check_rules ~/yy/tests/OTHER/errors/ignored_metavar_regex.yaml
+ /home/pad/yy/_build/default/src/cli/Main.exe -check_rules /home/pad/yy/tests/OTHER/errors/ignored_metavar_regex.yaml
[0.039  Info       Main.Dune__exe__Main ] loaded log_config.json
[0.040  Info       Main.Dune__exe__Main ] Executed as: /home/pad/yy/_build/default/src/cli/Main.exe -check_rules /home/pad/yy/tests/OTHER/errors/ignored_metavar_regex.yaml
[0.040  Info       Main.Dune__exe__Main ] Version: semgrep-core version: v0.59.0-6-ga525cec1-dirty, pfff: 0.42
[0.040  Info       Main.Check_rule      ] processing /home/pad/yy/tests/OTHER/errors/ignored_metavar_regex.yaml
Uncaught exception:

  Parse_rule.InvalidYaml("duplicate key 'metavariable' in dictionary. You should use multiple metavariable-regex", _)

Raised at Parse_rule.parse_extra in file "src/parsing/Parse_rule.ml", line 498, characters 10-87
```

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Added
+- Detect duplicate keys in YAML dictionaries in semgrep rules when parsing a rule
+  (e.g., detect multiple 'metavariable' inside one 'metavariable-regex')
+
 ### Changed
 - Added precise error location for the semgrep metachecker, to detect for example
   duplicate patterns in a rule
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Added
 - Detect duplicate keys in YAML dictionaries in semgrep rules when parsing a rule
   (e.g., detect multiple 'metavariable' inside one 'metavariable-regex')
+  
 ### Fixed
 - C/C++: Fixed stack overflows (segmentation faults) when processing very large
   files (#3538)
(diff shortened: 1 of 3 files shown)


aryx:
Same noise here, the PR does not modify the semgrep-core matching code, but I get:

```

fire Potential speedup in benchmark 2: -12.2%

15 benchmarks, 3.1% faster on average.

```
We need to fix that, to get real signal from those bench, otherwise we will ignore them when they occasionnaly matters.

aryx:
cc @emjin @mjambon about semgrep-benchmark bot

aryx:
wow, @mjambon look at those 2 bot bench conflicting comments on the same PR ...

IagoAbal:
I agree about the noise of the perf alarms, -+15% in one individual benchmark is just noise. Our benchmarks take ~10s each so 10% is just 1s... What if we raise the threshold to 20% at least (or even 25%)?

Otherwise we could run each benchmark a few more times, remove the fastest and the slowest, and then compute the average. At least run them three times and take the middle value.

I would still keep the threshold at -+10% for the average, because having 15 benchmarks the noise will get somewhat cancelled (would be better with more benchmarks).

IagoAbal:
> Otherwise we could run each benchmark a few more times, remove the fastest and the slowest, and then compute the average. At least run them three times and take the middle value.

The reference values should also be computed that way, maybe they are overly optimistic because it's very common to get these "+10% slowdown" warnings.

mjambon:
> wow, @mjambon look at those 2 bot bench conflicting comments on the same PR ...

They're not conflicting, they're different benchmarks (2 and 4). Since they're very close to the alert threshold, it's not surprising that they don't always show. This is why we have two levels: warning and error. Warnings will always be erratic but if we see a warning that's persistent or close to the error threshold, we know we should fix it.

mjambon (approved):

PR #3585, opened 2021-07-22
Match GHA workflow name with the corresponding config file name

mjambon:
There were two workflows named "Lint", and other workflows whose name didn't obviously match their file name, which I find confusing. This PR sets the workflow names to be exactly the file name so they can be found easily. We could rename them if that makes things clearer, but let's avoid discrepancies between file name and workflow name.

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -1,4 +1,4 @@
-name: Build Tests
+name: build-tests
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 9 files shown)


mjambon:
image

Looks like GHA hasn't picked up the new name for this workflow. "Lint" should now be "semgrep".

Update: oh, it's because it uses `pull_request_target:`, i.e. it still uses the config from the main branch.

aryx (approved):

aryx:
Maybe we need to update the release process which may reference some workflow jobs? 
https://r2c.quip.com/zrklAwLqfrm7/Using-the-Semgrep-Repo

mjambon:
good call. I made a best effort to update the instructions.

PR #3602, opened 2021-07-24
Ignore parsing timeouts for stats purposes

mjambon:
This should improve some of the parsing stats. I'll push this to the `parsing-stats` branch to see what happens.

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/parsing/Test_parsing.ml
+++ b/semgrep-core/src/parsing/Test_parsing.ml
@@ -217,9 +217,14 @@ let parsing_common ?(verbose = true) lang xs =
    Parse files from multiple root folders, each root being considered a
    separate project. Keeping projects separate allows us to spot
    projects with unusual results.
+
+   Timeouts are ignored.
 *)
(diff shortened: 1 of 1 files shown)


semgrep-core/src/parsing/Test_parsing.ml
```diff
@@ -217,9 +217,14 @@ let parsing_common ?(verbose = true) lang xs =
    Parse files from multiple root folders, each root being considered a
    separate project. Keeping projects separate allows us to spot
    projects with unusual results.
+
+   Timeouts are ignored.
 *)
 let parse_project ~verbose lang name files_or_dirs =
-  let stat_list = parsing_common ~verbose lang files_or_dirs in
+  let stat_list =
```
aryx, line 224: I don't think we should do that.
The idea of those parsing-stats (vs the stats we had just for tree-sitter) is to accurately represent
what semgrep can parse end-to-end (not just the tree-sitter part).

PR #3607, opened 2021-07-26
Remove per-pattern running from semgrep python

brendongo:
This PR deprecates the following experimental features:
- pattern-where-python
- taint-mode
- equivalences
- step-by-step evaluation output

This is a first pass at deleting code no longer necessary for per-pattern running. We can still simplify a lot of code in rule.py, and between rule/pattern_match.py but we can do so in a separate PR


emjin:
Woah are these speedups potentially legit?

brendongo:
I suspect it's cause we were using `--optimizations none` when deprecated features were encountered (equivalence rules for example)

semgrep/semgrep/core_runner.py
```diff
@@ -88,61 +79,35 @@ def _raise_semgrep_error_from_json(
         elif error_type == "invalid regexp in rule":
             raise SemgrepError(f'Invalid regexp in rule: {error_json["message"]}')
         elif error_type == "invalid pattern":
-            if self._optimizations == "all":
```
brendongo, line 91: The code path for optimizations == "none" is no longer reachable after deprecating per-pattern running

semgrep/semgrep/core_runner.py
```diff
@@ -88,61 +79,35 @@ def _raise_semgrep_error_from_json(
...
+                line=range.get("line", 0),
+                col=range.get("col", 0),
+                path=range.get("path", []),
+                filename="semgrep temp file",
+            )
+            if error_json["message"] == "Parsing.Parse_error":
+                long_msg = f"Pattern `{s.strip()}` could not be parsed as a {error_json['language']} semgrep pattern"
             else:
-                matching_pattern = next(
-                    (p for p in patterns if p._id == error_json["pattern_id"]), None
```
brendongo, line 115: Per pattern errors are not something that can happen now that we do not run per-pattern

semgrep/semgrep/core_runner.py
```diff
@@ -336,7 +301,6 @@ def _run_rules_direct_to_semgrep_core(
                         cmd = [SEMGREP_PATH] + [
                             "-lang",
                             language.value,
-                            "-fast",
```
brendongo, line 339: `-fast` skips files that do not contain tokens in the pattern. Removing it when optimizations are set to none so parsing tests do not skip parsing files.

semgrep/semgrep/rule.py
```diff
@@ -70,213 +52,61 @@ def __init__(self, raw: YamlTree[YamlMap]) -> None:
...
-        if not isinstance(other, type(self)):
-            return False
-
-        return self._raw == other._raw
-
-    def __hash__(self) -> int:
-        return hash(self.id)
-
-    def has_pattern_where_python(self) -> bool:
+    def _validate_none_language_rule(self) -> None:
```
brendongo, line 68: Rewrote the validate_none_language_rule to not need to parse into BooleanExpression. The check before is equivalent to making sure `pattern` does not exist in a regex rule.

semgrep/semgrep/rule.py
```diff
@@ -70,213 +52,61 @@ def __init__(self, raw: YamlTree[YamlMap]) -> None:
...
-                return BooleanRuleExpression(
-                    operator=OPERATORS.AND_EITHER,
-                    pattern_id=None,
-                    children=list(self._parse_boolean_expression(patterns)),
-                    operand=None,
-                )
+        return self._raw == other._raw
 
-        raise SemgrepError(f"rule with id '{_rule_id}' is missing top-level operator")
+    def __hash__(self) -> int:
```
brendongo, line 108: This was not added/deleted just moved

semgrep/semgrep/evaluation.py
```diff
@@ -1,97 +1,16 @@
-import copy
```
brendongo, line 1: Deleted boolean composition logic from this file now that it is handled in semgrep-core but kept the string interpolation logic

semgrep/semgrep/rule.py
```diff
@@ -70,213 +52,61 @@ def __init__(self, raw: YamlTree[YamlMap]) -> None:
...
             rule_languages.add(Language.TYPESCRIPT)
+            self._raw["languages"] = [r.value for r in rule_languages]
 
         self._languages = sorted(rule_languages, key=lambda lang: lang.value)  # type: ignore
 
         # check taint/search mode
         if self._raw.get("mode") == JOIN_MODE:
             self._mode = JOIN_MODE
         else:
-            self._expression, self._mode = self._build_search_patterns_for_mode(
```
brendongo, line 80: We no longer need to parse the rule in the python side. Most changes to this file are deleting this code path

semgrep/semgrep/rule_match.py
```diff
@@ -136,7 +136,3 @@ def lines(self) -> List[str]:
 
         self._lines_cache[(start_line, end_line)] = result
         return result
-
-    @property
-    def should_fail_run(self) -> bool:
```
brendongo, line 141: This was unused

semgrep/semgrep/spacegrep.py
```diff
@@ -1,163 +0,0 @@
-import json
```
brendongo, line 1: Spacegrep when needed is called from semgrep-core now

perf/bench/r2c-rules/r2c-ci.yml
```diff
@@ -577,8 +577,6 @@ rules:
 - id: java.lang.correctness.no-string-eqeq.no-string-eqeq
   languages:
   - java
-  equivalences:
```
emjin, line 580: To be clear, does this equivalence now happen automatically?

brendongo: I didn't think so but https://github.com/returntocorp/semgrep-rules/pull/1370/files makes it seem like it does.

semgrep/semgrep/core_runner.py
emjin: Nice, makes a lot of sense

semgrep/semgrep/error.py
```diff
@@ -262,27 +258,6 @@ class InvalidPatternError(ErrorWithSpan):
     level = Level.ERROR
 
 
-@attr.s(frozen=True, eq=True)
-class InvalidPatternErrorNoSpan(SemgrepError):
```
emjin, line 266: Yay glad to see this killed :D

emjin (approved):

PR #3608, opened 2021-07-27
Report pattern error ranges to semgrep cli

emjin:
Pass pattern range information through to the JSON output, so that it will be surfaced in error messages in CLI and playground.

Test plan - CLI:

Modify `tests/OTHER/rules/metavar_pattern_lang.yaml` so that the pattern is `bad eval_C("$CODE")` instead of `eval_C("$CODE")`. Then run (in `tests/OTHER/rules`)

`semgrep --config metavar_pattern_lang.yaml metavar_pattern_lang.py`

The error message should say

```
running 1 rules...
semgrep error: invalid pattern
  --> semgrep temp file:8

Pattern `bad eval_C("$CODE")` could not be parsed as a Python semgrep pattern
```

We will need to fix this so that the temp file is retained

Test plan - Playground: https://semgrep.dev/s/BLJR/?version=4cfd20f237a9106b185d56c747215b856bca2e84

Copy the docker hash produced by the github actions run and go to `https://semgrep.dev/editor/?version=[hash]`. Make an error in a pattern and verify that the underline happens in the correct pattern.

Underlining will be somewhat off because of YAML parser limitations.

PR checklist:
- [ ] changelog is up to date


semgrep/semgrep/rule_lang.py
```diff
@@ -131,6 +134,31 @@ def from_string(cls, s: str, filename: Optional[str] = None) -> "Span":
...
+        row_diff = len(lines)
+        col_diff = len(lines[-1])
+        end = Position(line=(row_diff + line - 1), col=(col_diff + col + 1))
+        return Span(
+            start=start,
+            end=end,
+            file=filename,
+            source_hash=src_hash,
+            config_path=tuple(path),
+            config_start=Position(0, 1),
```
emjin, line 161: @brendongo Unclear if this should instead be Position(1, 1)? Or Position(0, 0)?

emjin:
Not sure what I need to do to my pre-commit to surface these errors

aryx:
pre-commit run -a did not fix those?

aryx:
lots of CI errors to fix.

semgrep-core/src/core/Rule.ml
```diff
@@ -40,6 +40,9 @@ type tok = AST_generic.tok [@@deriving show, eq, hash]
 
 type 'a wrap = 'a AST_generic.wrap [@@deriving show, eq, hash]
 
+(* To help report pattern errors in simple mode in the playground *)
```
IagoAbal, line 43: Perhaps clarify that the `string list` represents a path of pattern operators (correct?). You could also use a record here, usually preferable over tuples IMO.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -83,7 +83,14 @@ exception InvalidLanguage of Rule.rule_id * string * Parse_info.t
  * start of the pattern *)
 exception
   InvalidPattern of
-    Rule.rule_id * string * Rule.xlang * string (* exn *) * Parse_info.t
+    Rule.rule_id
```
IagoAbal, line 86: A record type could be nice here too.

emjin: I'll leave this for now because the exception isn't used in that many places and I'd want to change all of them for consistency

aryx:
What's going on with CI? Why some cancelled jobs and skipped checks?

semgrep/semgrep/rule_lang.py
```diff
@@ -114,6 +114,9 @@ class Span:
     file: Optional[str]
     context_start: Optional[Position] = None
     context_end: Optional[Position] = None
+    config_path: Optional[Tuple[Any, ...]] = None
```
brendongo, line 120: What is the difference between file and config_path and start/end and config_start/end?

emjin: config_path is the yaml path to the pattern, config_start/end are the start/end for just the pattern

brendongo: Are there errors that need to be reported on both a target file and rule file? Cant the file be used for config_path?

emjin: No it's the path like ["rules", 0, "patterns", "pattern-either", 1]. We need it for the playground to report errors in basic mode

emjin:
Remaining TODOs:
Show users the snippet of pattern that has an error: https://github.com/returntocorp/semgrep/issues/3635
Report specific range for parse errors in patterns: https://github.com/returntocorp/semgrep/issues/3637
Fix inaccuracies in reported error ranges: https://github.com/returntocorp/semgrep/issues/3638

aryx:
A bit annoying that we have this intermediate temp file. Why not the original file? So we can report errors in the original file?

emjin:
> A bit annoying that we have this intermediate temp file. Why not the original file? So we can report errors in the original file?

If the user sends us a config file with rules in multiple languages, we have to send separate rules to `semgrep-core`.

aryx:
Maybe we should have some #line directive in the yaml like in C with cpp

aryx:
So we can report error on the oroginal file

semgrep/semgrep/rule_lang.py
IagoAbal: Did you get an answer to this?

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/badpattern/error-in-color.txt
```diff
@@ -1,2 +1,6 @@
 running 1 rules...
-[31mRule id: eqeq-is-bad contains a pattern that could not be parsed as a pattern for language Python: `$X == $X 3`[39m
+[31msemgrep error[39m: invalid pattern
+  --> semgrep temp file:5
```
IagoAbal, line 3: Was the name of the file `temp` or is it in fact a temporary file? If we get temporary files then it may be better not to include them in the report for now.

IagoAbal (approved):

IagoAbal:
> > A bit annoying that we have this intermediate temp file. Why not the original file? So we can report errors in the original file?
> 
> If the user sends us a config file with rules in multiple languages, we have to send separate rules to `semgrep-core`.

We should fix that, but in the meantime better not to report errors on a `temp` file, it's going to be confusing rather than helpful IMO.

semgrep-core/src/reporting/JSON_report.ml
```diff
@@ -269,15 +269,37 @@ let json_of_exn e =
...
+      in
       J.Object
         [
           ("pattern_id", J.String pattern_id);
           ("error", J.String "invalid pattern");
           ("pattern", J.String pattern);
           ("language", J.String lang);
           ("message", J.String message);
+          ("range", range_json);
         ]
```
mjambon, line 303: We should switch to using atdgen at some point because it's unclear what type this has and whether it's going to be correct. See for example Semgrep.atd.

mjambon (approved):

Issue #3603, opened 2021-07-26
Parsing error for parameter calls with ()


devd:
**Describe the bug**

Parse error on a ruby file that accepts block as a param and calls it via ->() and the pattern has metavariables.

**To Reproduce**
https://semgrep.dev/s/AwBL/

**Expected behavior**
no parse error

**What is the priority of the bug to you?**

- [ ] P2: regular bug that should get fixed

IagoAbal:
This seems to be due to fake tokens in the Ruby AST.

PR #3613, opened 2021-07-27
core: Avoid crashing due to fake tokens

IagoAbal:
Hopefully one day we can guarantee that every expression has a known
range. But, in the meantime, if we cannot get a range, we should handle
the situation more gracefully.

Let's remove Visitor_AST.range_of_any in favor of the safer
Visitor_AST.range_of_any_opt. This will force us to think what to do
in each situation where range info is not available. For now we just
log an error and move to something else.

Follows: #3603

test plan:
make test

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Eval_generic.ml
+++ b/semgrep-core/src/engine/Eval_generic.ml
@@ -295,12 +295,18 @@ let bindings_to_env xs =
 (* this is for metavariable-regexp *)
 let bindings_to_env_with_just_strings xs =
   xs
-  |> List.map (fun (mvar, mval) ->
+  |> Common.map_filter (fun (mvar, mval) ->
          let any = MV.mvalue_to_any mval in
-         let min, max = Visitor_AST.range_of_any any in
-         let file = min.Parse_info.file in
-         let range = Range.range_of_token_locations min max in
-         (mvar, String (Range.content_at_range file range)))
+         match Visitor_AST.range_of_any_opt any with
+         | None ->
+             (* TODO: Report a warning to the user? *)
+             logger#error "We lack range info for metavariable %s: %s" mvar
+               (G.show_any any);
+             None
+         | Some (min, max) ->
+             let file = min.Parse_info.file in
+             let range = Range.range_of_token_locations min max in
+             Some (mvar, String (Range.content_at_range file range)))
   |> Common.hash_of_list
 
 let eval_bool env e =
--- a/semgrep-core/src/engine/Match_patterns.ml
+++ b/semgrep-core/src/engine/Match_patterns.ml
@@ -167,13 +167,20 @@ let match_rules_and_recurse config (file, hook, matches) rules matcher k any x =
            matches_with_env
            |> List.iter (fun (env : MG.tin) ->
                   let env = env.mv.full_env in
-                  let range_loc = V.range_of_any (any x) in
-                  let tokens = lazy (V.ii_of_any (any x)) in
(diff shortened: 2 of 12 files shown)


IagoAbal:
I converted this PR to a draft, there are still some places where we call `range_of_any` and I would like to replace all of them with `range_of_any_opt` so hopefully we get rid of all these crashes...

semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -295,12 +295,18 @@ let bindings_to_env xs =
...
   xs
-  |> List.map (fun (mvar, mval) ->
+  |> Common.map_filter (fun (mvar, mval) ->
          let any = MV.mvalue_to_any mval in
-         let min, max = Visitor_AST.range_of_any any in
-         let file = min.Parse_info.file in
-         let range = Range.range_of_token_locations min max in
-         (mvar, String (Range.content_at_range file range)))
+         match Visitor_AST.range_of_any_opt any with
+         | None ->
```
emjin, line 301: Can we factorize the error reporting? Maybe make a `no_range_info_error` function, and if we later decide we want to report a warning to the user we can add it there.

aryx (approved):

aryx:
@nbrahms wanted that for a long time I think.

semgrep-core/src/engine/Match_patterns.ml
```diff
@@ -167,13 +167,20 @@ let match_rules_and_recurse config (file, hook, matches) rules matcher k any x =
...
            |> List.iter (fun (env : MG.tin) ->
                   let env = env.mv.full_env in
-                  let range_loc = V.range_of_any (any x) in
-                  let tokens = lazy (V.ii_of_any (any x)) in
-                  let rule_id = rule_id_of_mini_rule rule in
-                  Common.push
-                    { PM.rule_id; file; env; range_loc; tokens }
-                    matches;
-                  hook env tokens));
+                  match V.range_of_any_opt (any x) with
```
emjin, line 170: This pattern is getting rather long now. Can we make a `push_match` function?

aryx:
Note though that we will not get anymore bug reports that forced us to fix code. Now all those errors
will be silently ignored. I wonder if we could surface them back by recording those logger#error in the cloud somehow.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -642,92 +642,107 @@ and satisfies_metavar_pattern_condition env r mvar opt_xlang formula =
...
+                  let content = Range.content_at_range env.file mval_range in
+                  Common2.with_tmp_file ~str:content ~ext:"mvar-pattern"
+                    (fun file ->
+                      (* We don't want having to re-parse `content', but then we
+                       * need to fix the token locations in `mast`. *)
+                      let mast_start_loc =
+                        mval |> MV.ii_of_mval |> Visitor_AST.range_of_tokens
+                        |> fst |> PI.token_location_of_info
+                      in
+                      let fix_loc loc =
```
emjin, line 692: Can we pull some of this out to make this section of code less deeply nested?

emjin (approved):

IagoAbal:
Thanks for the comments! I will take care of them in a separate PR.

> Note though that we will not get anymore bug reports that forced us to fix code. Now all those errors
will be silently ignored. I wonder if we could surface them back by recording those logger#error in the cloud somehow.

Good point we should do something about it.

Issue #3606, opened 2021-07-26
Option for SARIF output to include ignored findings using the suppressions property


ShaylanDias:
**Is your feature request related to a problem? Please describe.**
We would like visibility into Semgrep findings which have been ignored via a `nosemgrep` comment— especially for integration with other tooling which performs requires this information.

**Describe the solution you'd like**
There should be an option for Semgrep findings which are ignored through the use of `nosemgrep` comments to still show up in the SARIF output as a `result` with the `suppressions` property set to an array containing the object `{"kind": "inSource"}`. This marks them as ignored by an inline comment but keeps them in the output so they are still visible to the end user/tooling. This could be disabled by default but enabled through the CLI with a flag `--sarif-include-nosemgrep`.

**Additional context**
The [SARIF spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Toc10541171) details a `suppression` object contained within the `suppressions` property. The idea is to allow analysis tools to convey information on findings which have been "suppressed." The `"kind": "inSource"` attribute specifically refers to if the source code contains flags which remove warnings, which fits `nosemgrep` comments.
<img width="192" alt="Screen Shot 2021-07-27 at 2 40 31 PM" src="https://user-images.githubusercontent.com/38924672/127231245-54fdb44b-568e-4412-8bfe-fa777cda00f8.png">

ShaylanDias:
After looking at the source code briefly I feel fairly confident about taking this on myself and submitting a PR. I just wanted to gauge whether or not it fits within the Semgrep design plan first.

brendongo:
Seems like `suppressedInSource` https://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html#_Toc517436079  would be more appropriate here?

ShaylanDias:
Yes, you're definitely correct—I'll update the post to reflect. I think the idea still stands though.

PR #3616, opened 2021-07-27
Include ignored findings in SARIF output using suppression syntax

ShaylanDias:
In reference to #3606.

This change adds ignored findings to SARIF output using the `suppressions` option for `result` objects.

PR checklist:
- [ ] changelog is up to date


semgrep/semgrep/semgrep_main.py
```diff
@@ -356,6 +357,8 @@ def main(
             for rule_match in rule_matches:
                 if rule_match._is_ignored:
                     num_findings_nosem += 1
+                    if sarif_include_ignores:
```
mschwager: Should we avoid the additional CLI arg and always add these?

brendongo: I'm supportive. It does break the assumption that people may have built around that all findings printed in the output are new findings but it does conform to sarif spec.

brendongo: I think it's not ideal for developers to see suppressed findings as it would be confusing but IMO developers should be shown the plain text output anyway and the sarif is for those who are building out some other tooling. @ShaylanDias mind modifying the PR so that behavior is to always include the suppressed finding in the sarif and for now not including an extra flag to remove it.

ShaylanDias: Makes sense to me since it avoids more complications in the tool— but like you said @brendongo I didn't want to break any assumptions people might have already made. However, if there's consensus that the CLI arg is unnecessary I think it makes sense to remove it.

brendongo (approved):
This looks great @ShaylanDias thanks!

tdonaworth:
Resurfacing this, mostly to understand the logic behind the "always include suppressed findings in Sarif". Just ran up against this issue; and ran the rabbit hole of finding what was going on, - basically was still getting x results found, even though they were `#nosemgrep` commented - tracked it down to the `--sarif` option, etc. 

The issue here, is that GitHub CodeQL has no idea about the `suppressions` property, and basically carries on as if the findings are open and valid, even though they're suppressed in code. So curious - why include it, or is there a way to NOT include it in the sarif output?

wwmoraes:
> Resurfacing this, mostly to understand the logic behind the "always include suppressed findings in Sarif". Just ran up against this issue; and ran the rabbit hole of finding what was going on, - basically was still getting x results found, even though they were `#nosemgrep` commented - tracked it down to the `--sarif` option, etc.
> 
> The issue here, is that GitHub CodeQL has no idea about the `suppressions` property, and basically carries on as if the findings are open and valid, even though they're suppressed in code. So curious - why include it, or is there a way to NOT include it in the sarif output?

My apologies in advance for the necroposting. For the sake of those landing here from web search (like me), here's a solution that filters out suppressed findings from a generated SARIF file using [jq](https://jqlang.org):

```shell
semgrep scan --quiet --sarif 2>/dev/null | jq 'del(.runs[].results[] | select(.suppressions))' > semgrep-scan.sarif
## OR with sponge from moreutils package; useful if you need to merge SARIFs and filter all at once
semgrep scan --quiet --sarif-output-semgrep-scan.sarif 2>/dev/null
jq 'del(.runs[].results[] | select(.suppressions))' semgrep-scan.sarif | sponge semgrep-scan.sarif
```

I've added both `--quiet` and the redirect to silence banners as the CLI output reports all findings including the suppressed ones, which is misleading. If you must see them after generation then either use [sarif-fmt](https://psastras.github.io/sarif-rs/) on the final SARIF file or run another scan without any of the SARIF flags to see only the unsupressed ones.

PR #3617, opened 2021-07-28
Update to LSP 1.7.0

aryx:
test plan:
```
$ semgrep-core -lang ocaml -lsp -e '($E : string ref)' src/cli/Main.ml
src/cli/Main.ml:339
   Common.write_file ~file:tmp (Common.read_file !rules_file)
src/cli/Main.ml:374
   if !use_parsing_cache = "" then f ()
src/cli/Main.ml:402
   let dir = !use_parsing_cache in
...

```

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/typing/LSP_client.ml
+++ b/semgrep-core/src/typing/LSP_client.ml
@@ -21,6 +21,8 @@ open Types
 module PI = Parse_info
 module G = AST_generic
 
+let logger = Logging.get_logger [ __MODULE__ ]
+
 (*****************************************************************************)
 (* Prelude *)
 (*****************************************************************************)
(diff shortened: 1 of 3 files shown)


semgrep-core/src/typing/LSP_client.ml
```diff
@@ -21,26 +21,27 @@ open Types
...
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
 
 let server =
-  match 1 with
-  | 1 -> "/home/pad/.opam/4.09.1/bin/ocamllsp"
+  match 0 with
+  | 0 -> "ocamllsp"
+  | 1 -> "/home/pad/.opam/4.12.0/bin/ocamllsp"
```
IagoAbal, line 44: We should use `Sys.getenv "HOME"` here.

IagoAbal: Ideally we should search for the binary in the `$PATH`.

IagoAbal:
Very cool!

semgrep-core/src/typing/LSP_client.ml
aryx: yes, this is hardcoded for now because it's very experimental.

aryx:
It's cool, but it does not really work :( It can infer local types, but I am not able to get
full type information for entities defined in other files. For example on many files where
ocamlmerlin says it's an AST_generic.expr under the cursor, ocamllsp just says 'a.

PR #3624, opened 2021-07-29
Add OCaml module for filtering files based on name + exec perm + first line

mjambon:
I added a `Guess_lang` module which takes care of filtering target files using whatever criteria we want, including inspecting the contents of the file. For now, this is useful to exclude certain files from the parsing stats, such as the `.d.ts` typescript files that we can't parse. This is the immediate motivation for doing this in OCaml.

Additionally, this also supports shebang line inspection, which will be useful for parsing files without a distinguishing extension. For example, if the language is bash (and some other popular ones including python, ruby, javascript) a file with executable permission and no extension will have its first line inspected. If it's like `#!/bin/bash` or `#!/usr/bin/env bash` or some variation of those, the file is assumed to be a bash script. For using this feature for regular semgrep use, the python wrapper would have to pass those files to semgrep-core, and semgrep-core will have to indicate which of those files were actually accepted or rejected.

Right now, `Guess_lang` is used only for the parsing stats.

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/core/ast/Guess_lang.ml
+++ b/semgrep-core/src/core/ast/Guess_lang.ml
@@ -1,0 +1,187 @@
+(*
+   Guess whether a given file is indeed written in the specified
+   programming language.
+*)
+
+(*
+   Evaluation will be left-to-right and lazy, using the usual (&&) and (||)
+   operators.
+*)
+type test =
+  | And of test * test
+  | Or of test * test
+  | Not of test
+  | Test_path of (string -> bool)
+
+let eval test path =
+  let rec eval = function
+    | And (f, g) -> eval f && eval g
+    | Or (f, g) -> eval f || eval g
+    | Not f -> not (eval f)
+    | Test_path f -> f path
+  in
+  eval test
+
+let string_chop_prefix ~pref s =
+  let len = String.length s in
+  let preflen = String.length pref in
+  if len >= preflen && String.sub s 0 preflen = pref then
+    Some (String.sub s preflen (len - preflen))
+  else None
+
+let has_suffix suffixes =
+  let f path =
+    List.exists (fun suf -> Filename.check_suffix path suf) suffixes
+  in
+  Test_path f
+
+let prepend_period_if_needed s =
+  match s with "" -> "." | s -> if s.[0] <> '.' then "." ^ s else s
+
+(*
+   Both '.d.ts' and '.ts' are considered extensions of 'hello.d.ts'.
+*)
+let has_extension extensions =
+  has_suffix (List.map prepend_period_if_needed extensions)
+
+let has_lang_extension lang = has_extension (Lang.ext_of_lang lang)
+
+let has_an_extension =
+  let f path = Filename.extension path <> "" in
+  Test_path f
+
+let is_executable =
+  let f path =
+    Sys.file_exists path
+    &&
+    let st = Unix.stat path in
+    match st.st_kind with
+    | S_REG ->
+        (* at least some user has exec permission *)
+        st.st_perm land 0o111 <> 0
+    | _ -> false
+  in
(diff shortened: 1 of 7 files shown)


mjambon:
Please ignore the failed "ci/circleci: parsing-stats" job that shouldn't show up here.

Here are the latest stats:

Languages for which the stats just went back to normal are:
* go https://dashboard.semgrep.dev/metric/semgrep.core.go.parse.pct
* ocaml https://dashboard.semgrep.dev/metric/semgrep.core.ocaml.parse.pct
* typescript https://dashboard.semgrep.dev/metric/semgrep.core.typescript.parse.pct

Unclear:
* C# csharp https://dashboard.semgrep.dev/metric/semgrep.core.csharp.parse.pct

For Typescript, it's presumably because we now filter out the `.d.ts` files. For Go and OCaml, I don't know.

Full results:
```
Language: c
no data

Language: cpp
Line count: 19464841
Parsing rate: 0

Language: csharp
Line count: 7118387
Parsing rate: 0.999322880309823

Language: go
Line count: 8295081
Parsing rate: 0.999862086940441

Language: hack
Line count: 35535
Parsing rate: 0

Language: java
Line count: 4757773
Parsing rate: 0.9999957963526213

Language: javascript
Line count: 1432082
Parsing rate: 0.9999881291713743

Language: kotlin
parse error: Invalid numeric literal at line 1, column 2
Line count: 
parse error: Invalid numeric literal at line 1, column 2
Parsing rate: 

Language: lua
no data

Language: ocaml
Line count: 2892335
Parsing rate: 0.887389600443932

Language: php
Line count: 5008994
Parsing rate: 0.9704088285991159

Language: python
Line count: 4631020
Parsing rate: 0.9936480516171383

Language: r
Line count: 266422
Parsing rate: 0

Language: ruby
Line count: 5422701
Parsing rate: 0.9986272523600324

Language: rust
Line count: 4141155
Parsing rate: 0.9873825056053203

Language: typescript
Line count: 6711512
Parsing rate: 0.993034356490758

For the long version, check results.txt.

FAIL 73 c
7329 ocaml
OK 634 cpp
OK 1660 csharp
OK 3691 go
OK 6480 hack
OK 12082 java
OK 18726 javascript
FAIL 25639 kotlin
FAIL 33807 lua
OK 7329 ocaml
OK 18893 php
OK 31410 python
OK 9502 r
OK 27594 ruby
OK 11046 rust
OK 27816 typescript
```

aryx:
For OCaml it's because I've filled all the todos in Parse_ocaml_tree_sitter.ml

aryx (approved):

semgrep-core/src/core/ast/Guess_lang.ml
```diff
@@ -0,0 +1,187 @@
+(*
+   Guess whether a given file is indeed written in the specified
+   programming language.
+*)
+
+(*
+   Evaluation will be left-to-right and lazy, using the usual (&&) and (||)
+   operators.
+*)
+type test =
```
IagoAbal, line 10: Do we need this mini-language?

IagoAbal: Hmm, OK, I get why you added it. You could also have defined some custom combinators to work with `'a -> bool`, e.g.:
```
let (&&$) f g = fun x -> f x && g x 
```

IagoAbal:
Very nice! 

> For now, this is useful to exclude certain files from the parsing stats, such as the .d.ts typescript files that we can't parse. 

Do we want to use this also in `Main` ?

semgrep-core/src/core/ast/Guess_lang.ml
mjambon: yeah, it was one or the other. I like the feel of the `And`, `Or`, etc, probably because they're clearer than punctuation-based operators.

Issue #3623, opened 2021-07-28
[call for ideas] Eliminate stack overflows or troubleshoot them more easily


mjambon:
In OCaml (native code), stack overflows often result in a segfault rather than a nice `Stack_overflow` exception. Any practical solution that mitigates this problem is welcome.

Stack overflows may occur in the normal use of semgrep if the maximum stack size is too small and the input is a bit extreme. Things we can do include:
* Don't use the stack unless necessary (e.g. don't use `Stdlib.List.map`).
* Use `ulimit -s` to lower the maximum stack size during testing and development; raise the maximum stack size in production.
* Count the recursive calls to a few functions that are known for occasionally causing stack overflows (and that can't be avoided), and print a warning when passing a threshold.
* Modify the ocaml runtime so it always raises `Stack_overflow` on stack overflows.
* Find some other trick to detect when the stack size approaches the limit.

mjambon:
I don't know why there's a maximum stack size while there's no maximum heap size. The only clue I found is that for imperative languages like C, a large stack usually is due to an accidental infinite loop. Catching this error would prevent memory exhaustion for all the processes of the OS.

If that's all there is to it, I suggest we increase the maximum stack size to a very large value and only worry about not exhausting physical memory.

mjambon:
I asked the question on [Discuss](https://discuss.ocaml.org/t/is-there-any-value-in-having-a-maximum-stack-size/8214):
> Is there any value in having a maximum stack size?

mjambon:
Main solution for unix systems
--

Set the stack size limit to `unlimited`. This can be done from bash with `ulimit -s unlimited`. From OCaml, this requires a third-party library that provides an ocaml interface to the `getrlimit` and `setrlimit` syscalls. From python: see https://docs.python.org/3/library/resource.html

For Windows, changing the stack limit apparently must be done at compilation/linking time or by editing the binary with a special tool.

If the stack limit can't be raised sufficiently or at all (possibly semgrep-core local testing and/or Windows)
--

1. If we know the stack size limit: extend the current `Gc` alarm to check the stack size and raise `Stack_overflow` if it gets close to the actual limit.
2. If we don't know the stack size limit: print a warning if the stack passes 7 MB.

IagoAbal:
IMO we should just set up a signal handler for `SEGV` and use it to terminate the program more gracefully with a proper message and some debug info.

mjambon:
[...]

PR #3625, opened 2021-07-29
Do something about stack overflows

mjambon:
* Print a warning when the stack size exceeds 7 MB.
* Raise a `Stack_overflow` exception when the stack size exceeds half of the user-specified memory limit (requires `ulimit -s unlimited` or equivalent).

See discussions and plan: https://github.com/returntocorp/semgrep/issues/3623

PR checklist:
- [x] changelog is up to date


semgrep-core/src/cli/Main.ml
```diff
@@ -441,12 +454,31 @@ let run_with_memory_limit limit_mb f =
...
+      else if stack_bytes > stack_warning && not !stack_already_warned then (
+        logger#warning
+          "large stack size: %d bytes. If a crash follows, you should suspect \
+           a stack overflow. Make sure the maximum stack size is set to \
+           'unlimited' or to a value greater than %d bytes so as to obtain an \
+           exception rather than a segfault."
+          stack_bytes mem_limit;
+        stack_already_warned := true)
     in
     let alarm = Gc.create_alarm limit_memory in
```
IagoAbal: > create_alarm f will arrange for f to be called at the end of each major GC cycle, starting with the current cycle or the next one. A value of type alarm is returned that you can use to call delete_alarm.

I guess there is no warranty that the alarm will be called before the stack overflows?

IagoAbal: But it should work for the cases that we have experienced, I believe that previous to the stack overflow you could see increased allocation. Have you tested this?

aryx:
You need to add Memory_limiit.ml

aryx (changes requested):
missing file

mjambon:
From what I can tell, simply using the GC alarm causes a 6x slowdown. We may have not seen it before because we'd not set a GC alarm with the default settings, but now we do.

I made another branch that is the same as `develop` branch except it always create the GC alarm: #3643. Let's see if this affects performance.

aryx:
Ok, for once the semgrep benchmark regression test played its role! Glad it can at least catch those big regressions :)

mjambon:
I think it should work without any performance degradation now.

Stack overflow prevention doesn't work very well because it requires a lot of allocation per function call for the GC alarm to run frequently enough (see the unit tests).

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Fixed
 - Go: Match import module paths correctly (#3484)
 - core: Do not crash when is not possible to compute range info
+- eliminate 6x slowdown when using the '--max-memory' option
```
IagoAbal, line 29: Nice!

IagoAbal:
Fixing the slowdown for `-max_memory` was a very nice side-effect of this PR.

The stack overflow detection I am not sure if it's worth while. It seems to be an issue to be solved in the OCaml compiler, and apparently it only affects ARM. It only works under very specific conditions (this is not discussed in the code or in `-max_stack`...), it must trigger a major collection at the right time. Do we know if it actually works for the stack overflows that we experienced?

I think the right move would be to fix Semgrep so it doesn't get killed if anything happens when analyzing some specific file.

mjambon:
Yeah, `-max_stack` doesn't work well - I had to put a lot of artificial allocation in the `grow_stack` function used in the unit tests otherwise it wouldn't work. I think it's misleading: we could still have segfaults (on those archs) due to stack overflow but not realize it because we expect `-max_stack` to work.

I'm going to do the following:
- remove `-max_stack` completely
- keep the other things this branch fixes (fix the perf issue with `Gc.compact()`, monitor heap+stack memory rather than just heap with `-max_memory`)
- always create a GC alarm so that all our tests and all production code run with it.

PR #3629, opened 2021-07-29
Refactoring: consistently use module G = AST_generic

aryx:
When using semgrep -lsp to find code using type information,
we rely on ocamllsp (and merlin) to accurately print types.
It does not print normalized types though, so if each
file use a different alias, we would need each time a different
pattern.

test plan:
make

PR checklist:
- [x] changelog is up to date


aryx:
The failing CI test is because the semgrep check use the develop docker image, and in the last PR
I've modified the OCaml grammar to not support attributes on let patterns, which are used in semgrep.yml.
With this PR we should allow that back and hopefully will not get any failure at the next PR.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -17,17 +17,16 @@
...
-(* A is the pattern, and B the concrete source code. For now
+(* G is the pattern, and B the concrete source code. For now
  * we both use the same module but they may differ later
  * as the expressivity of the pattern language grows.
  *
  * subtle: use 'b' to report errors, because 'a' is the sgrep pattern and it
  * has no file information usually.
  *)
-module A = AST_generic
 module B = AST_generic
```
IagoAbal, line 27: So we will need a different pattern for this `B`?

PR #3632, opened 2021-07-29
Deduplicate findings that fire on same lines and have same message

### LINKED ISSUE #3614: Double Findings
**Describe the bug**
When we run semgrep in terminal on a local dir, we are getting double findings. Even if there is only one pattern, it gives 2 findings per line, usually. Sometimes, there is only one finding per line, I cannot see a pattern as to when there is double findings vs single. Also, by we I mean myself and another engineer both tried on our machines.
Also, I just updated from semgrep .57 to .59 and it still occurs.

**To Reproduce**
It does not reproduce in playground, but we reproduce it every time we scan.

**Expected behavior**
Single finding per rule or pattern.

**Screenshots**
editedcodeoutSemgrep
Screen Shot 2021-07-26 at 3 17 44 PM

**Environment**
Local scans, does not seem to be occurring in our semgrep CI.

### LINKED ISSUE #3618: Duplicated findings in the output results
**Describe the bug**
When running SemGrep scans I see duplication of results for some rules in my automated scans. 
Looks like this behaviour introduces around version 0.57.0

**To Reproduce**
See two examples:
https://semgrep.dev/s/Gwz1?version=0.60.0 - two similar/duplicated finding for the same line
https://semgrep.dev/s/Gwz1?version=0.56.0 - only one finding reported - correct one.

**Expected behavior**
Rule should not produce duplicated output targeting the same line. 

**Screenshots**
Screenshot 2021-07-28 at 17 52 37

P0-P1 - we are blocked and unable to update to the latest SemGrep versions, but we are reverted to v0.56.0 and able to unblock our CI flow. 

**Environment**
My builds, deployments and semGrep dev: https://semgrep.dev/s/Gwz1

brendongo:
PR checklist:
- [x] changelog is up to date


brendongo:
Closes https://github.com/returntocorp/semgrep/issues/3618, https://github.com/returntocorp/semgrep/issues/3614

semgrep/semgrep/core_runner.py
```diff
@@ -430,3 +429,32 @@ def invoke_semgrep(
...
+
+
+# This will remove matches that have the same range but different
+# metavariable bindings, choosing the last one in the list. We want the
+# last because if there multiple possible bindings, they will be returned
+# by semgrep-core from largest range to smallest. For an example, see
+# tests/e2e/test_message_interpolation.py::test_message_interpolation;
+# specifically, the multi-pattern-inside test
+#
+# Another option is to not dedup, since Semgrep.ml now does its own deduping
```
emjin, line 441: Maybe comment that this option has been done and rejected

emjin (approved):

aryx:
probably need to make regenerate snapshot to make CI works.

aryx (dismissed):
need to regenerate the snapshot I think.

brendongo:
I'll regenerate them

PR #3633, opened 2021-07-29
Improve --test failure output

### LINKED ISSUE #3604: Found mismatch between test and result ids
While testing rules with `semgrep --test`, I occassionally get an error message. I have problems interpreting the error message on a user level.

```
$ semgrep --test python/lang/correctness/
....
  File "/usr/local/bin/semgrep", line 33, in <module>
    sys.exit(load_entry_point('semgrep==0.57.0', 'console_scripts', 'semgrep')())
  File "/usr/local/Cellar/semgrep/0.57.0/libexec/lib/python3.9/site-packages/semgrep/__main__.py", line 20, in main
    cli()
  File "/usr/local/Cellar/semgrep/0.57.0/libexec/lib/python3.9/site-packages/semgrep/cli.py", line 487, in cli
    semgrep.test.test_main(args)
  File "/usr/local/Cellar/semgrep/0.57.0/libexec/lib/python3.9/site-packages/semgrep/test.py", line 462, in test_main
    generate_file_pairs(
  File "/usr/local/Cellar/semgrep/0.57.0/libexec/lib/python3.9/site-packages/semgrep/test.py", line 348, in generate_file_pairs
    tested = {
  File "/usr/local/Cellar/semgrep/0.57.0/libexec/lib/python3.9/site-packages/semgrep/test.py", line 349, in <dictcomp>
    filename: score_output_json(
  File "/usr/local/Cellar/semgrep/0.57.0/libexec/lib/python3.9/site-packages/semgrep/test.py", line 200, in score_output_json
    raise Exception(
Exception: found mismatch between test and result ids - test={'yield-in-init', 'return-in-init'} result={'yield-in-init'}
```

I think this happens when the `return-in-init` matches nothing. However, this message is pretty technical and I don't understand it. It seems semgrep compares two files and these don't match up, but this seems like an implementation detail that is not needed while writing rules.

So I suggest changing the error message, or adding a explicit check, that says something like:

> rule 'return-in-init' did not match any line in the test file 'return-in-init.py'.

mschwager:
Fixes #3604.

### DIFF
--- a/semgrep/semgrep/test.py
+++ b/semgrep/semgrep/test.py
@@ -45,6 +45,8 @@ RULEID = "ruleid"
 TODOOK = "todook"
 OK = "ok"
 
+EXIT_FAILURE = 1
+
 
 def _remove_ending_comments(rule: str) -> str:
     for _, end in COMMENT_SYNTAXES:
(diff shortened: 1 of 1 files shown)


IagoAbal (approved):

semgrep/semgrep/test.py
```diff
@@ -187,23 +189,34 @@ def score_output_json(
...
+    test_lines: Dict[str, Set] = collections.defaultdict(set)
+    for lines in [ruleid_lines, ok_lines]:
+        for file_path, test_annotations in lines.items():
+            test_lines[file_path].update(test_annotations.keys())
+
+    rule_id_mismatch = False
+    if reported_lines:
+        for file_path, test_ids in test_lines.items():
+            reported_ids = set(reported_lines[file_path].keys())
+            if test_ids.symmetric_difference(reported_ids):
```
IagoAbal, line 206: I guess `reported_lines` it's the errors that Semgrep reported, and `test_lines` includes both `ruleid:` as well as `ok:` lines. If we have one rule for which there is no `ruleid:` but there is some `ok:`, then that rule will not be in `reported_lines` but it will be in `test_lines`, then we raise an error. Is my reasoning correct?

minusworld (approved):

semgrep/semgrep/test.py
mschwager: Hmm, that is an interesting edge case. I believe you are correct. IMO it's still okay to error out because it'd be quite odd to have a test case with `ok`'s but no `ruleid`'s. We're also maintaining the existing behavior, so we can consider this error case more closely if a user has concerns :+1:

PR #3643, opened 2021-07-31
Check performance impact of always running the GC alarm

mjambon:
PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -452,9 +452,7 @@ let run_with_memory_limit limit_mb f =
         raise Out_of_memory)
     in
     let alarm = Gc.create_alarm limit_memory in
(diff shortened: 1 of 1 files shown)


aryx:
At least we can disable it by default and when people get a segfault they can turn it on, get something slower but working at least .. or can give us a better bug report than just segfault.

aryx:
Still we should ask on discuss.ocaml how to improve this or get the Stack_overflow working also on macOS reliably.

mjambon:
oh, that code calls `Gc.compact ()` at the end of each run. That seems like a bad idea. I'll check.

mjambon:
yup, that was it. We can call `Gc.compact ()` only when catching an `Out_of_memory` exception, although I'm not sure it's the right thing to do. Perhaps `Gc.full_major ()` would be just fine. It probably doesn't matter.

PR #3647, opened 2021-08-01
Pattern: $X speedup try 2

emjin:
As before, we want to speed up something like

```
  patterns:
  - pattern-either:
    - pattern-inside: |
        import $STYLE from "...";
        ...
    - pattern-inside: |
        $STYLE = $METHOD(...);
        ...
  - pattern-inside: |
      <$EL style={$STYLE} />
  - pattern: $STYLE
```

In this PR, we try to do this by extracting the AST from the final range matched by everything except `pattern: $STYLE`. Currently does not work

Test plan:

For correctness in the anded case, in semgrep-core/ run (or make test)

```
sc -config tests/OTHER/rules/pattern-x.yaml tests/OTHER/rules/pattern-x.js -lang js
```

The expected results (you can confirm by running semgrep-core) are

```
perf/input/pattern-x.js:17 with rule typescript.react.security.audit.react-css-injection.react-css-injection
             <div style={input}>
perf/input/pattern-x.js:5 with rule typescript.react.security.audit.react-css-injection.react-css-injection
             <div style={input}>
```
To check that this actually worked, you can run

```
sc -config tests/OTHER/rules/pattern-x.js ~/your_root/semgrep/perf/bench/rails/input/rails/actionview/test/ujs/public/vendor/jquery-2.2.0.js -profile -lang js
```

We expect this to drop from ~100 s to 2 s

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -729,12 +729,62 @@ and satisfies_metavar_pattern_condition env r mvar opt_xlang formula =
             env.rule_id mvar;
           false)
 
-and nested_formula_has_matches env formula lazy_ast_and_errors lazy_content
+and run_metavariable_on_ranges env ranges selector_opt =
+  let run_metavariable_on_range range =
+    match selector_opt with
+    | None -> [ range ]
+    | Some { S.pattern; pid; pstr; _ } ->
+        let content = Range.content_at_range env.file range.RM.r in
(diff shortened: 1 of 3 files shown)


IagoAbal:
We could fix this PR, but is this approach general enough? What if the range on which we want to run `$X` cannot be parsed into a Generic program? I think we may need something to fall-back in the general case.

PR #3649, opened 2021-08-02
Speed up taint mode by memoizing range computation for expressions

### LINKED ISSUE #3596: Investigate why taint-mode appears to be slower
@minusworld reported that after switching a rule to taint-mode in https://github.com/returntocorp/semgrep-rules/pull/1389, it became about 2x slower.

IagoAbal:
Closes #3596

test plan:
semgrep-core -lang py -profile \
    -config semgrep-rules/python/django/security/injection/open-redirect.yaml \
    semgrep/perf/bench/django/input/django
    #^ This is now as fast as the old fake-taint rule and range_of_any_opt
    # amounts for less than 20% of the total time (vs above 60% before).

PR checklist:
- [x] changelog is up to date


semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -381,7 +381,14 @@ and id_info = {
  * e.g., type information, or constant evaluation, or range, but it
  * would be a bigger refactoring than for stmt.
  *)
-and expr = { e : expr_kind; e_id : int }
+and expr = {
+  e : expr_kind;
+  e_id : int;
+  (* used in semgrep to quickly get the range of a matched statement *)
```
aryx: typo: of a matched expression

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -381,7 +381,14 @@ and id_info = {
...
+  mutable e_range :
+    (Parse_info.token_location * Parse_info.token_location) option;
+      [@hash.ignore]
+}
```
aryx, line 391: Probably need to also define the equal for it. I think we forgot for s_range but it's a bug, we should.

aryx (approved):

aryx:
CI failures, you probably need to make regenerate-tests. We have some debugging test that
do not abstract away the md5sum and so fail each time we change the expression type.
Not sure who wrote the debugging test, but I remember @brendongo or @mschwager wrote some code to not care about those md5sum in the expected json output.

PR #3650, opened 2021-08-02
Add basic support for parsing Hack files

frankeld:
### Primary Changes
Hello 👋 ! This PR includes the first set of CST-to-AST changes to support Hack(lang) in Semgrep. The overwhelming majority of changes are in the Hack parse file. Since we decided to move forward with mapping directly to the generic AST, this PR also includes small updates for that as well. The parse file is still very much a WIP and contains many outstanding TODOs and questions. It also does not yet support Semgrep language extensions.

The parse file changes are too numerous to document explicitly, but the commit messages should be accurate to each language feature introduced. There is also one basic test file - there will be more in subsequent PRs.

### Secondary Changes
- Update sempgrep-hack to the latest release.
- Mirror of the repo removal in https://github.com/returntocorp/ocaml-tree-sitter-languages/pull/6 to remove non-Hack repo from project list.
- Mirror of the clone script change in https://github.com/returntocorp/ocaml-tree-sitter-core/pull/6 to speed up cloning.
- I commented out the last two "framework" git repos. Because there were three repos with the same name, I was getting a script error. I don't know if there's a workaround for this, but feel free to revert the change if so. I can also look into a script modification that will handle name conflicts differently and/or open an issue.

PR checklist:
- [x] changelog is up to date


semgrep-core/src/parsing/tree_sitter/Parse_hack_tree_sitter.ml
```diff
@@ -488,1277 +598,1603 @@ and anon_choice_exp_rep_COMMA_choice_exp_opt_COMMA_e4364bb (env : env)
...
-  in
-  let v2 =
-    match v2 with Some x -> type_arguments env x | None -> todo env ()
+    | `Choice_array x ->
+        G.N (G.Id (collection_type env x, G.empty_id_info ())) |> G.e
   in
+  let v2 = match v2 with Some x -> type_arguments env x | None -> None in
   let v3 = arguments env v3 in
-  todo env (v1, v2, v3)
+  G.Call (v1, v3) |> G.e
```
mjambon, line 1034: Like I mentioned to @aryx the other day, maintenance is much easier if the left-handside variables are given somewhat meaningful names instead of `v1`, `v2` etc. Instead of this:
```ocaml
let v2 = match v2 with Some x -> type_arguments env x | None -> None in
```
I write this:
```ocaml
let type_args = match v2 with Some x -> type_arguments env x | None -> None in
```
It's easy to give reasonable names while their meaning fresh in your mind but it's much harder later.

Here, the return value would be something like
```ocaml
G.Call (f, args) |> G.e
```
so when v3 becomes v4 after some grammar change, it helps reduce confusion.

I've started using [this style systematically for the bash parser](https://github.com/returntocorp/semgrep/blob/47d0935186b25fc9e9c23c1db229a59756f512ea/semgrep-core/src/parsing/tree_sitter/Parse_bash_tree_sitter.ml#L362-L366), after experiencing painful upgrades with javascript and typescript.

mjambon (approved):
Looks great!

mjambon:
The lint job failed because you didn't set up the pre-commit hook. You need to install `pre-commit` and run `pre-commit install` once. It will reformat code for you when you run `git commit`. To fix things up, run `pre-commit run -a` (or study the other options).

aryx:
He probably installed pre-commit, but he might have not installed ocamlformat.0.16.0.
Our pre-commit look for the presence of ocamlformat and if it's not there we just don't run it (I think we did that
to not force people who wants to modify only the Python part of semgrep to install ocaml tools;  I think we might
want to revisit that decision).

Try opam install ocamlformat 0.16.0 (or opam install --deps-only .  from semgrep-core/dev/)

aryx:
Ok I'll merge this and fix the ocamlformat for this time.

semgrep-core/src/parsing/tree_sitter/Parse_hack_tree_sitter.ml
frankeld: Noted! I'll update as I go with future changes. I'll also try to rename the `anon_choice_` functions to be more descriptive.

frankeld:
The issue came from a conflict between some of my VSCode extensions and ocamlformat. I fixed it for the main Parse file (hence the lack of formatting issues there), but forgot to go back and fix it for the target file. Regardless, thanks for updating and it should be good in the future with pre-commit.

PR #3652, opened 2021-08-03
Use ocaml-tree-sitter-core as a submodule

mjambon:
This simplifies the build a little. Also, we no longer build the executables from the submodules because we don't need them.

PR checklist:
- [ ] changelog is up to date


aryx (approved):

aryx:
CI failure: Error: Library "tsort" not found.
I guess we need to do the opam install --deps-only in ocaml-tree-sitter-core in one of our workflow?

aryx:
Also when I use your branch and try to compile from semgrep-core I get:
```
m -f bin
./scripts/make-version > ./src/cli/version.ml
dune build
File "tree-sitter.opam", line 1, characters 0-0:
Warning: Unable to read opam file. Some information about this package such
as its version will be ignored.
Reason: Sys_error("tree-sitter.opam: No such file or directory")
Error: Too many opam files for package "tree-sitter":
- src/ocaml-tree-sitter-core/tree-sitter.opam
- tree-sitter.opam
make: *** [Makefile:20: all] Error 1

```

aryx:
hmm, I just deleted this semgrep-core/tree-sitter.opam symlink. Not sure why it was there.
Also apparently it's not even in the repo. Also weird that git status does not show it. It must be ignored
somewhere but I can't find it in semgrep/.gitignore

aryx:
oh, it was in the semgrep-core/.gitignore.
We should avoid those nested .gitignore honestly.

semgrep-core/dune
```diff
@@ -13,3 +13,125 @@
...
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Don't build executables for the submodules that we only use as libraries.
+; - Saves a tiny bit of build time (about 3-4%).
+; - Keeps the bin/ folder clean.
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(subdir
+   src/pfff
+   (data_only_dirs cli)
```
aryx, line 25: Actually I use the binary from pfff inside semgrep, because it has more options to dump ASTs, like
-dump_ml, -dump_java, -dump_php, which dumps the original ASTs of the language, not the generic AST one.

But I guess I can clone pfff in its own place and use the binary built from there.

aryx:
Weird that the CI failure: Error: Library "tsort" not found. happens only for the osx-release job.
Our docker build should also fail.

mjambon:
> oh, it was in the semgrep-core/.gitignore.
> We should avoid those nested .gitignore honestly.

Why? Here, if you did `git status`, you'd see right away that there's some junk left over from before. Local `.gitignore` files allow moving and removing folders without breaking anything. This is the same design principle as having local `dune` files. If we had everything in a global `.gitignore`, which naturally I would not have thought about editing, we'd also get a build error but `git status` wouldn't tell us that we have junk files that should be removed.

semgrep-core/dune
mjambon: I wasn't sure about pfff. I was mostly annoyed when calling `ocaml-tree-sitter` because the `semgrep-core/bin` was in my PATH so I could call `semgrep-core` without installing. When calling `ocaml-tree-sitter`, it would find `semgrep-core/bin/ocaml-tree-sitter` which is usually not the version I want.

mjambon:
> Weird that the CI failure: Error: Library "tsort" not found. happens only for the osx-release job.
> Our docker build should also fail.

Yeah. Thanks for fixing this & the rest.

It's because dune was complaining about finding two `tree-sitter.opam` files so I removed the one I could remove but broke the installation of dependencies.

Issue #3552, opened 2021-07-20
Fix semgrep-core parsing stats


mjambon:
We don't know why the new parsing stats on semgrep-core are so much worse than those produced by ocaml-tree-sitter-semgrep.

See for example:
* ocaml-tree-sitter stats: https://dashboard.semgrep.dev/metric/semgrep.tree-sitter.c-sharp.parse.pct
* semgrep-core stats: https://dashboard.semgrep.dev/metric/semgrep.core.c.parse.pct This used to receive ocaml-tree-sitter stats, hence the sharp drop.

The semgrep-core stats include the construction of the semgrep generic AST, unlike the ocaml-tree-sitter stats.

aryx:
Dupe of https://github.com/returntocorp/semgrep/issues/3538

PR #3664, opened 2021-08-04
Many fixes to improve the parsing_stat CI job

aryx:
The Lua parsing stat job was failing because it was using
a non-github repo which triggered an RPC error at clone time.

The Kotlin parsing stat job was failing because it was using
CST.dump_tree which output things on stdout, which then messes up
the json.

The C parsing stat job was failing because of some out of memory
fatal errors on huge files. I tried to use Memory_limit
but ran in many issues. It's hard to intercept those Out_of_memory
reliably.

test plan:
./run-lang --upload lua
./run-lang --upload kotlin
./run-lang --upload c

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/parsing/Test_parsing.ml
+++ b/semgrep-core/src/parsing/Test_parsing.ml
@@ -186,12 +186,52 @@ let test_parse_tree_sitter lang xs =
 *)
 let parsing_common ?(verbose = true) lang xs =
   let timeout_seconds = 10.0 in
+  (* Without the use of Memory_limit below, we were getting some
+   * 'Fatal error: out of memory' errors in the parsing stat CI job,
+   * because of some ulimit -v 4000000 in the run-lang script,
+   * which then was aborting the whole job.
+   *
+   * Note that the mem_limit_mb value below must be less than the
+   * ulimit value in run-lang.
+   *
+   * In fact, it is not clear how much less it must be. For example with:
+   * $ ulimit -v 1000000
+   * $ semgrep-core -lang c -parsing_stats -json lang/c/tmp/php-php-src/ext/fileinfo/
+   * I still get a core dump with a "Fatal error: out of memory",
+   * even if I set mem_limit_mb below to 500.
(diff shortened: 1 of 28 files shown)


aryx:
This may help https://github.com/returntocorp/semgrep/issues/3552

aryx:
This may help https://github.com/returntocorp/semgrep/issues/3538

aryx:
@IagoAbal here is an example of a file that uses lots of memory:
semgrep-core -lang c -parsing_stats ~/semgrep/parsing-stats/lang/c/tmp/php-php-src/ext/fileinfo/

I've put a memory limit for now. Dunno if it's because of parsing or naming.

semgrep-core/src/parsing/Test_parsing.ml
```diff
@@ -186,12 +186,52 @@ let test_parse_tree_sitter lang xs =
...
+   *
+   * In fact, it is not clear how much less it must be. For example with:
+   * $ ulimit -v 1000000
+   * $ semgrep-core -lang c -parsing_stats -json lang/c/tmp/php-php-src/ext/fileinfo/
+   * I still get a core dump with a "Fatal error: out of memory",
+   * even if I set mem_limit_mb below to 500.
+   * Seems like the Gc alarm does not trigger, maybe because there is no
+   * major GC cycle triggered before reaching the 1 GB ulimit.
+   *
+   * So the current workaround is to still put a limit below, but
```
IagoAbal: > ... put a limit below, ...

I am not sure what is the reference for "below" here. I mean, below what exactly? Before it was below the `ulimit -v` but now that does not exist. Below the memory available in the system?

aryx:
Another file that use more than 4GB of memor:
/home/pad/github/semgrep/parsing-stats/lang/ocaml/tmp/rescript-lang-rescript-compiler/lib/4.06.1/whole_compiler.ml
however this file is 22MB. Still, a bit sad that we get 2 orders of magnitude more memory than the file to store its AST
in memory.

IagoAbal (approved):
Nice!

PR #3676, opened 2021-08-05
core: Make sure memory-limit test works

IagoAbal:
The GC alarm doesn't trigger so reliably it seems, the test was
failing on my laptop.

PR checklist:
- [x] changelog is up to date


mjambon (approved):

mjambon:
Many workflows are failing, which seems to be due to network errors. I reran just one of them. I don't know if there's an easy way to rerun all the workflows.

PR #3684, opened 2021-08-06
Refactor m_name, factorize more code

aryx:
test plan:
make test

PR checklist:
- [x] changelog is up to date


aryx:
The bench test fails, but it's actually slightly faster on average so I'll merge.

Issue #3547, opened 2021-07-19
Properly communicate errors in semgrep-core to the CLI


IagoAbal:
**Is your feature request related to a problem? Please describe.**
Warnings or errors raised by semgrep-core are quite often not communicated properly to the CLI, so the CLI fails with some generic error that may not be very helpful. For example, in #3531, semgrep-core reports an exception when parsing a piece of Python code, but this is reported as a YAML syntax error by the CLI.

**Describe the solution you'd like**
We should make sure that all warnings or errors raised by semgrep-core will be added to its JSON output when `-json` is given.

aryx:
I'm done with what I wanted to do.
Feel free to take over.

IagoAbal:
Another example of bad error communication is #3603.

emjin:
I've done what I meant to: this now gives a better error message: https://semgrep.dev/s/BLJR/?version=4cfd20f237a9106b185d56c747215b856bca2e84

PR #3686, opened 2021-08-06
core: Better communicate errors to the CLI

IagoAbal:
- Add backtrace to fatal errors to help diagnosing bug reports (Pfff).
- Report rule evaluation errors instead of "silently" logging them.

Helps #3547

test plan:

    % cat fatal.py
    def foo():
    return 1
    }
    % semgrep -l py -e 'x' fatal.py
    running 1 rules...
    semgrep-core reported a fatal error:
    -----
    Fatal Error: (Failure "Lexer_python.top_mode: empty stack")
    Raised at Parse_target.run in file "src/parsing/Parse_target.ml", line 176, characters 17-26
    ...
    -----
    Please file a bug report at https://github.com/returntocorp/semgrep/issues/new/choose
    ...

    % cat warn.yaml
    rules:
    - id: test
      languages: [python]
      patterns:
      - pattern: $X
      - metavariable-pattern:
          metavariable: $Y
          pattern: x
      message: Test
      severity: ERROR
    % cat warn.py
    x
    % semgrep -c warn.yaml -v warn.py
    ...
    semgrep-core reported a matching error
    --> matching internal error: rule test: metavariable-pattern failed because $Y it not in scope, please check your rule
    ...

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -25,6 +25,7 @@ module MV = Metavariable
 module RP = Report
 module S = Specialize_formula
 module RM = Range_with_metavars
+module E = Error_code
 
 let logger = Logging.get_logger [ __MODULE__ ]
 
@@ -111,14 +112,29 @@ type env = {
   pattern_matches : id_to_match_results;
   (* used by metavariable-pattern to recursively call evaluate_formula *)
   file : Common.filename;
-  rule_id : string;
+  rule_id : R.rule_id;
   xlang : R.xlang;
   equivalences : Equivalence.equivalences;
+  (* problems found during evaluation, one day these may be caught earlier by
+   * the meta-checker *)
+  errors : E.error list ref;
 }
 
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
+
+(* Report errors during evaluation to the user rather than just logging them
+ * as we did before. *)
+let error env msg =
+  (* We are not supposed to report errors in the config file for several reasons
+   * (one being that it's often a temporary file anyways), so we report them on
+   * the target file. *)
+  let loc = PI.first_loc_of_file env.file in
+  let s = Printf.sprintf "rule %s: %s" env.rule_id msg in
+  (* TODO: warning or error? MatchingError or ... ? *)
+  let err = E.mk_error_loc loc (E.MatchingError s) in
+  Common.push err env.errors
 
 let (xpatterns_in_formula : S.sformula -> R.xpattern list) =
  fun e ->
@@ -294,7 +310,7 @@ type ('target_content, 'xpattern) xpattern_matcher = {
  * expect a different location if the end part is on a different line
  * (e.g., the semgrep Python wrapper), so I now return a pair.
  *)
-and match_range = Parse_info.token_location * Parse_info.token_location
+and match_range = PI.token_location * PI.token_location
 
 (* this will be adjusted later in range_to_pattern_match_adjusted *)
 let fake_rule_id (id, str) =
@@ -612,7 +628,7 @@ let rec filter_ranges env xs cond =
           * the text representation of the metavar content.
           *)
          | R.CondRegexp (mvar, (re_str, _re)) ->
-             let fk = Parse_info.fake_info "" in
+             let fk = PI.fake_info "" in
              let fki = AST_generic.empty_id_info () in
              let e =
                (* old: spf "semgrep_re_match(%s, \"%s\")" mvar re_str
@@ -644,19 +660,22 @@ and satisfies_metavar_pattern_condition env r mvar opt_xlang formula =
   (* If anything goes wrong the default is to filter out! *)
   match List.assoc_opt mvar bindings with
   | None ->
-      (* TODO: Report a warning to the user? *)
-      logger#error "rule %s: metavariable-pattern: %s not found" env.rule_id
-        mvar;
+      error env
+        (Common.spf
+           "metavariable-pattern failed because %s it not in scope, please \
+            check your rule"
+           mvar);
       false
   | Some mval -> (
       (* We will create a temporary file with the content of the metavariable,
        * then call evaluate_formula recursively. *)
       match MV.range_of_mvalue mval with
       | None ->
-          (* TODO: Report a warning to the user? *)
-          logger#error
-            "rule %s: metavariable-pattern: we lack range info for %s: %s"
-            env.rule_id mvar (MV.show_mvalue mval);
+          error env
+            (Common.spf
+               "metavariable-pattern failed because we lack range info for %s, \
+                please file a bug report"
+               mvar);
           false
       | Some mval_range -> (
           let r' =
@@ -675,11 +694,11 @@ and satisfies_metavar_pattern_condition env r mvar opt_xlang formula =
                *   specify `language: generic` (case `Some xlang` below). *)
               match MV.program_of_mvalue mval with
               | None ->
-                  (* THINK: fatal error instead? *)
-                  logger#error
-                    "rule %s: metavariable-pattern: %s does not bound a \
-                     sub-program"
-                    env.rule_id mvar;
+                  error env
+                    (Common.spf
+                       "metavariable-pattern failed because %s does not bind \
+                        to a sub-program, please check your rule"
+                       mvar);
                   false
               | Some mast ->
                   let content = Range.content_at_range env.file mval_range in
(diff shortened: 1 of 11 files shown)


IagoAbal:
@brendongo you or someone who is more involved in the CLI should check that you like my changes to `core_exception.py` and `errors.py`.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -111,15 +112,29 @@ type env = {
...
 
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
 
+(* Report errors during evaluation to the user rather than just logging them
+ * as we did before. *)
+let error env msg =
+  (* We are not supposed to report errors in the config file for several reasons
+   * (one being that it's often a temporary file anyways, so we report them on
```
emjin: I think you want an end parenthesis after the "anyways"?

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -111,15 +112,30 @@ type env = {
...
 
+(* Report errors during evaluation to the user rather than just logging them
+ * as we did before. *)
+let error env msg =
+  (* We are not supposed to report errors in the config file for several reasons
+   * (one being that it's often a temporary file anyways), so we report them on
+   * the target file. *)
+  let loc = PI.first_loc_of_file env.file in
+  let s = Printf.sprintf "rule %s: %s" env.rule_id msg in
+  (* TODO: warning or error? MatchingError or ... ? *)
```
emjin, line 135: A little confused---it sounds like these used to be silently logged. Will reporting them as errors now cause them to crash Semgrep? Or is this just a cosmetic question?

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -644,19 +660,20 @@ and satisfies_metavar_pattern_condition env r mvar opt_xlang formula =
   (* If anything goes wrong the default is to filter out! *)
   match List.assoc_opt mvar bindings with
   | None ->
-      (* TODO: Report a warning to the user? *)
```
emjin, line 647: Yay!

semgrep/semgrep/core_exception.py
```diff
@@ -112,12 +114,13 @@ def into_semgrep_error(self) -> SemgrepError:
             )
             return SourceParseError(
                 short_msg="parse error",
-                long_msg=f"Could not parse {self._path.name} as {self._language}"
-                + (
-                    f"\n\nsemgrep-core message:\n\n\t{self._extra['message']}"
-                    if self._check_id != "ParseError" and "message" in self._extra
-                    else ""
-                ),
+                long_msg=f"Could not parse {self._path.name} as {self._language}",
```
emjin, line 117: Is `self._extra['message']` ever informative with ParseErrors? Maybe worth leaving it in

emjin (approved):

semgrep-core/src/engine/Match_rules.ml
IagoAbal: No it won't crash Semgrep, they will just be communicated to the user instead of "silently" logged. Now that I'm looking into the test failures, I'm wondering if maybe not all of them should be communicated though!

semgrep/semgrep/core_exception.py
IagoAbal: Note that before there was a `self._check_id != "ParseError"`. In the few tests I did I just got a "Syntax error" message which is not very useful.

semgrep-core/src/engine/Match_rules.ml
IagoAbal: When running Semgrep you will e.g. see:
```
found problems analyzing 1 files; run with --verbose for details or run with --strict to exit non-zero if any file cannot be analyzed cleanly
```
And using `--verbose` you'll get:
```
semgrep-core reported a matching error
  --> matching internal error: rule test: metavariable-pattern failed because $Y it not in scope, please check your rule
```
But any matches that were found and reported previously should continue to be found and reported now.

semgrep/tests/e2e/snapshots/test_check/test_timeout_threshold/error.txt
```diff
@@ -7,4 +7,4 @@ rules:
 Warning: 1 timeout error(s) in targets/equivalence/open_redirect.py when running the following rules: [rules.forcetimeout]
 Semgrep stopped running rules on targets/equivalence/open_redirect.py after 1 timeout error(s). See `--timeout-threshold` for more info.
 ran 3 rules on 1 files: 0 findings
-1 files could not be analyzed; run with --verbose for details or run with --strict to exit non-zero if any file cannot be analyzed
+found problems analyzing 1 files; run with --verbose for details or run with --strict to exit non-zero if any file cannot be analyzed cleanly
```
aryx: we could fix the plural, it's 1 file, not 1 files :)

aryx (approved):

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -111,15 +112,30 @@ type env = {
   pattern_matches : id_to_match_results;
   (* used by metavariable-pattern to recursively call evaluate_formula *)
   file : Common.filename;
-  rule_id : string;
+  rule_id : R.rule_id;
   xlang : R.xlang;
   equivalences : Equivalence.equivalences;
+  (* problems found during evaluation, one day these may be caught earlier by
+   * the meta-checker *)
```
aryx, line 119: yep, exactly, for each of those errors now surfaced to the user we should improve the metachecker to detect them too
statically.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -739,20 +758,21 @@ and satisfies_metavar_pattern_condition env r mvar opt_xlang formula =
                     lazy_ast_and_errors
                     (lazy content)
                     (Some r'))
-          | Some _lang, _mval ->
-              (* THINK: fatal error instead? *)
-              logger#error
-                "rule %s: metavariable-pattern: the content of %s is not text"
-                env.rule_id mvar;
+          | Some _lang, mval ->
+              logger#debug
```
aryx, line 766: why not an error for this one?

aryx (approved):

semgrep-core/src/engine/Match_rules.ml
IagoAbal: Initially I thought it should be an error too, but it can happen easily without being an error, I saw this when running the changes on semgrep-rules. I added this comment:
```ocaml
(* This is not necessarily an error in the rule, e.g. you may be
* matching `$STRING + ...` and then add a metavariable-pattern on
* `$STRING`. This will only work when `$STRING` binds to some text
* but it can naturally bind to other string expressions. *)
```

semgrep/semgrep/output.py
```diff
@@ -307,9 +307,9 @@ def close(self) -> None:
         elif self.semgrep_structured_errors:
             # make a simplifying assumption that # errors = # files failed
             # it's a quite a bit of work to simplify further because errors may or may not have path, span, etc.
-            error_stats = (
-                f"{len(self.semgrep_structured_errors)} files could not be analyzed"
-            )
+            num_errors = len(self.semgrep_structured_errors)
+            plural = "s" if num_errors > 1 else ""
+            error_stats = f"found problems analyzing {num_errors} file{plural}"
             final_error = self.semgrep_structured_errors[-1]
```
brendongo, line 313: I think with this hacky assumption to have final_error be the last element of semgrep_structured_errors and CoreWarning having a 0 return code and https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/__main__.py#L25 I think there might be a scenario when semgrep should be exiting with non-zero exit code but instead has a 0 exit code?

IagoAbal: Good catch, I don't have an example of that but it seems possible to me. I'll give it a try at fixing it.

semgrep/semgrep/error.py
```diff
@@ -342,6 +343,47 @@ def to_dict_base(self) -> Dict[str, Any]:
         }
 
 
+@attr.s(frozen=True, eq=True)
+class CoreWarning(SemgrepError):
+    check_id: str = attr.ib()
+    msg: str = attr.ib()
+
+    code = INVALID_PATTERN_EXIT_CODE
```
IagoAbal, line 351: @brendongo Despite this Semgrep still returns 0 when one of these warnings is reported, is that normal?

IagoAbal: I'll merge but let me know if I should do something about this.

PR #3688, opened 2021-08-06
Don't consider test files in language stats for the repository on Github

ievans:
This should make us show up more obviously as a Ocaml + python repo.


aryx (approved):
Wow nice. I always wanted to fix that :)

aryx:
I wonder how long before we see an update.

PR #3689, opened 2021-08-07
Do not match "..." to interpolated strings

emjin:
This is a quick fix to immediately improve Semgrep's consistency, but it is definitely not a good one and we should change our implementation of interpolated strings.

The problem was that `x = "..."` does not match `x = "str{input}"`, but `"..."` matches `"str{input}"`. This is especially unfortunate because rule writers will use `"..."` to search for constant strings, and there is no other syntactic way (that I could figure out) to do this.

TODO: A better solution might be to create an InterpolatedLiteral expression and have `"str"` in the above example be an `InterpolatedLiteral(String("str", _))` instead of `L(String(str, _))`. This is fairly time consuming and should be thought through more carefully first, though, and I wanted to be able to write my rule.

PR checklist:
- [ ] changelog is up to date


ievans:
@emjin this makes me wonder whether it's actually a consistent idiom to use `"..."` as a string literal. We've entrenched that pretty well so I doubt we'd want to change, but I could see (more in the interactive search case) users expecting that `"..."` would still find `"str{input}"`

IagoAbal:
Could you share what rule did you want to write? We should first discuss whether you can achieve what you want in some other way, even if Semgrep is pre-1.0 software, we should still be careful not to break things lightly. This somewhat reminds me of the "duplicate findings" issue.

> The problem was that x = "..." does not match x = "str{input}", but "..." matches "str{input}". 

I am not sure I see any inconsistency here, it depends on how you view interpolated strings.

IagoAbal (changes requested):
We should discuss the rule that motivates this breaking change.

emjin:
I want to write a pattern like:

```
  patterns:
  - pattern-either:
    - pattern: Dir.$X($STR)
    - pattern: File.$X($STR)
  - metavariable-pattern:
      metavariable: $X
      patterns: 
      - pattern-either:
        - pattern: open
        - pattern: read
  - metavariable-pattern:
      metavariable: $STR
      patterns:
        - pattern: ...
        - pattern-not: "..."
```

(There are many more than just Dir and File, this is just a subset)

So anything like `Dir.read("#{user_input}")` is bad, but `Dir.read("path_to_file")` is fine. I think I could express this as `Dir.$X(...)` but not `Dir.$X("...")` (and I will try to do this after I send this message), but I'd rather not because it doubles the number of patterns I need in the first `pattern-either`. It also confuses me that this would be different, because I think of `metavariable-pattern` as being equivalent to if substituted back the pattern (in this case `...` and not `"..."`) where I have the metavariable.

I also kind of doubt this is a breaking change, because I would be surprised if people have rules that are just `"..."` (`metavariable-pattern` is fairly new). It is definitely worth thinking about, though, particularly @ievans's point. Re: consistency, I personally don't think people view "foo{hello}bar" as ["foo", hello, "bar"]; if I searched for "foo", I would expect exact matches to "foo", not matches to "foobar" and "foo{hello}bar". If people want to search for strings containing "foo", they should use pattern-regex. Of course, this is based on my own intuition, not any real data.

Would love input from @inkz @minusworld @clintgibler @mschwager @colleend

clintgibler:
I don't think almost anyone outside of r2c is using `metavariable-pattern` yet (and we don't use it extensively either), so I think any improvements to `metavariable-pattern` that make it more behave like one would expect is 👍  Better to make any breaking changes now vs later.

Personally, `"..."` to me should match only string literals. That's the way I've always thought we intended it to be used, and have presented it that way at like a dozen meetups.

I **don't** think `"..."` should match strings with variable interpolation. Those should be matched via `"... {$X} ..."` or something like that. 

I personally view `"..."` matching `"str{input}"` as a Semgrep bug.

colleend:
I also believe that `"..."` should only match string literals. A good number of the patterns we write are based off of this assumption, and it makes for a simpler worldview overall.

I don't believe that `"..."` should match variable interpolation -- it would make detecting false positives on some HTML concatenation detection rules harder, and also isn't consistent with what I've viewed `"..."` to match. 

Thanks for opening this up to discussion!

IagoAbal:
@emjin doesn't this do what you want https://semgrep.dev/s/YXn4 ? I simplified it a bit.

emjin:
...huh

I no longer have my original rule, so I can't reproduce what I was doing before to have this problem (to verify if I was very silly). I think I tried the pattern that is currently working for me? But I'm glad it works now!

IagoAbal:
Glad that it works! The Playground keeps changing from when I started using it months ago... now if you write `"..."` it will auto-magically remove the quotes (`"`) and convert that into `...` which is a very different pattern..., so you need to write `'"..."'` instead.

@emjin @clint @colleend

Personally, I would say people sees `f"foo{hello}bar"` as `"foo{hello}bar".format(hello=hello)` (hope I get Python's syntax right). So yeah we should probably represent interpolated strings in a different way, because searching for `"foo"` alone should indeed not return matches on `"foo{hello}bar"`. But I don't think they are a _literal_, they are a nice-looking "printf" expression.

I am not sure I see a problem with `"..."` matching a sub-component of an interpolated string, `"..."` will _not_ match the entire thing _unless_ the entire thing evaluates to a constant string. And you can disable that using `options:` and `constant_propagation: false`. But if you have good examples that show why that is bad please share.

Issue #3660, opened 2021-08-03
git submodules break scan target identification for "generic" language


malexmave:
**Describe the bug**
Running a rule with language "general" on the [OpenHaystack](https://github.com/seemoo-lab/openhaystack) repository results in an error message `Could not open 'Firmware/Microbit_v1/blessed': [Errno 21] Is a directory: 'Firmware/Microbit_v1/blessed'`. The directory "blessed" is empty.

I set a breakpoint on [core_exception.py:106](https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/core_exception.py#L106) and the call stack indicates that the error was raised by semgrep-core, not the python wrapper (as it comes from [core_runner.py:382](https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/core_runner.py#L382), which processes the results of the binary, it seems). As I have no idea about the languages used in the core, I did not investigate it further. The problem also occurs on other rules with the "general" language selected, the rule linked below is simply one of the ones from p/security-audit that triggers it.

In search of a minimum set of files needed to reproduce the issue, I found that if you delete all visible folders (on the root of the repository level) except "Firmware" and all hidden files and folders except ".git", the error still occurs. If you delete or rename the .git folder, the error goes away, and if you restore the .git folder, it appears again - so apparently the core binaries do something inside the .git folder that has an influence on this. This is further corroborated by the fact that deleting all the contents of the .git folder also makes the error disappear, so the mere presence of the folder does not suffice to trigger the bug.

If you put a file in the previously-empty "blessed" folder, the error still occurs.

**To Reproduce**
Clone the OpenHaystack repository, (optionally) delete all visible folders except "Firmware" and all invisible files and folders except .git, and run, for example (with example output provided):
```bash
$ semgrep --config="r/python.django.security.audit.xss.template-blocktranslate-no-escape.template-blocktranslate-no-escape"
using config from https://semgrep.dev/r/python.django.security.audit.xss.template-blocktranslate-no-escape.template-blocktranslate-no-escape. Visit https://semgrep.dev/registry to see all public rules.
downloading config...
running 1 rules...
Could not open 'Firmware/Microbit_v1/blessed': [Errno 21] Is a directory: 'Firmware/Microbit_v1/blessed'
ran 1 rules on 33 files: 0 findings
```

**Expected behavior**
[...]

PR #3690, opened 2021-08-07
Ensure that scanned paths aren't directories

malexmave:
Workaround for issue with git submodules and `git ls-files`, see #3660 for details.

Someone who knows more about semgrep internals should definitely take a closer look at this and ensure that I didn't miss any cases where passing folders into the list of targets is desired - I didn't find any, but I'm new to the codebase :). One could also replace it with an `is_file`, but again, I don't know if there are any non-file things you want to be able to investigate.

Note that at least on my machine, this code has [a bunch of failing `pytest` tests](https://gist.github.com/malexmave/aa61e2acf30de02e838d90049f0a3c52). However, the tests [also fail for the unmodified code](https://gist.github.com/malexmave/1d94f4ff68f5725a8ff57a5d523f5141) on my machine, so I don't quite know what to make of this.

PR checklist:
- [x] documentation is up to date (no changes necessary)
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 - core: Fix parsing of numeric literals in rule files
+- Generic parser no longer tries to open submodule folders as files
 
 ## [0.61.0](https://github.com/returntocorp/semgrep/releases/tag/v0.61.0) - 2021-08-04
 
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -284,8 +284,7 @@ class CoreRunner:
                         targets = [
                             target
                             for target in targets
-                            if target not in max_timeout_files
-                            and not target.is_dir()
+                            if target not in max_timeout_files and not target.is_dir()
                         ]
 
                         # opti: no need to call semgrep-core if no target files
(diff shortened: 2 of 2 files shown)


malexmave:
That's what happens when you just make a quick edit using the GitHub Web UI and forget about pre-commit. Fixed the linter complaint, let's hope the CI run is happy now :).

mjambon:
Great investigation, @malexmave. This is a problem right now for the generic mode but will be a bigger problem for all the scripting languages when we start scanning executable scripts that don't have an extension.

 In your branch, I think the filtering needs to take place earlier, as close as possible to where we run `git ls-files`. I'll give it a try.

brendongo:
Agree with @mjambon I think filtering the directories out in TargetManager._expand_dir during the git ls-files call would be better. TargetManager.get_files should only be actual files. Nice find with the submodules behavior

malexmave:
I figured I'd avoid adding another loop somewhere else and rather "recycle" an existing loop, but your argument about the semantics / expected behavior of the function make a lot more sense than shaving off a few microseconds (if even that). All maintainers of this repo should have write access to the branch, so feel free to make the change yourself if you know where you want the exclusion to happen :). Otherwise I can do it tomorrow (GMT).

mjambon:
@malexmave I took over your branch and made a new PR (https://github.com/returntocorp/semgrep/pull/3701). Sorry about that, I normally try to avoid to take over people's work when they're almost done... It's an area of the code I wanted to get familiar with. Thanks again for the finding.

malexmave:
@mjambon No worries, I'm happy it was helpful and the issue will be fixed. Closing this PR in favor of the other one, then :).

Issue #3576, opened 2021-07-21
Improve PHP parse rate to beta percentage


mschwager:
[PHP currently has](https://semgrep.dev/docs/status/):

- All Semgrep beta features
- A 10+ rule ruleset: https://semgrep.dev/p/phpcs-security-audit
- 97% of 99% beta parse rate

If we can bump up this last 2% parse rate we can promote it to beta!

mschwager:
Some fodder here:

- https://github.com/returntocorp/semgrep/issues/2648
- https://github.com/returntocorp/semgrep/issues/2650
- https://github.com/returntocorp/semgrep/issues/3587
- https://github.com/returntocorp/semgrep/issues/3588
- https://github.com/returntocorp/semgrep/issues/3589
- https://github.com/returntocorp/semgrep/issues/3590
- https://github.com/returntocorp/semgrep/issues/3591
- https://github.com/returntocorp/semgrep/issues/3592

Sjord:
Do you want to improve the built-in PHP parser, or switch to tree-sitter?

mschwager:
> Do you want to improve the built-in PHP parser, or switch to tree-sitter?

Hmm, that's a good question. @aryx will have a much better idea about the best path forward here.

aryx:
I think we should switch to the tree-sitter one. Note that Slack added support for Hack, which is kinda a superset of PHP.
They could have defined tree-sitter-hack as an extension of tree-sitter-php but didn't, because they said Hack and PHP diverged and it was more pain to make Hack an extension. Anyway, tree-sitter-php should be easy to integrate, it's just lots of boilerplate code to write.

Sjord:
The tree-sitter-php grammar doesn't parse string interpolation correctly yet. E.g. `"hello {$planet["a"]}"`. @cfroystad [got a long way with parsing this](https://github.com/tree-sitter/tree-sitter-php/pull/72), but seems to be stuck on some tree-sitter weirdness. He could use some help with this, and this would further improve parsing percentage.

cfroystad:
Thanks for focusing on this!
I'd be very happy to get some help get unstuck on that. The code is a bit messy due to PHP strings almost being a separate language, thus if anyone is familiar with tree-sitter and willing to lend me a hand, I'd be happy to jump on a call for some pair programming if that would make it easier to assist.

aryx:
[...]

PR #3693, opened 2021-08-09
add semgrep-php

aryx:
PR checklist:
- [x] changelog is up to date

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -61,3 +61,6 @@
 [submodule "semgrep-core/src/ocaml-tree-sitter-core"]
 	path = semgrep-core/src/ocaml-tree-sitter-core
 	url = https://github.com/returntocorp/ocaml-tree-sitter-core.git
+[submodule "semgrep-core/src/tree-sitter-lang/semgrep-php"]
+	path = semgrep-core/src/tree-sitter-lang/semgrep-php
+	url = https://github.com/returntocorp/semgrep-php


aryx:
This will help https://github.com/returntocorp/semgrep/issues/3576

aquynh:
do i understand correctly that this means future version will move away from pfff to tree-sitter?

aryx:
Yes.

aquynh:
What are pfff-based languages are missing now, compared with tree-sitter?

aryx:
I think right now the PHP parser in pfff is more complete than the tree-sitter-php, but in the past we've seen the community around tree-sitter being quite active, so I think the tree-sitter-php should soon be more complete than the pfff one.

aquynh:
I mean in principle what pfff cannot do that tree-sitter can, so switching
to tree-sitter is encouraged?

aryx:
The pfff repository contains mostly OCaml code, and most of the grammars are using ocamlyacc or menhir which are LALR(1) parser generators. tree-sitter is using a GLR parser generator which is more general and in principle more powerful.
Moreover, I've found that it's difficult for people to extend grammars written in OCaml, but for tree-sitter many people have manage to contribute. So regarding PHP it's either me who need to modify the PHP grammar in pfff, or people in the tree-sitter community who can modify tree-sitter-php. The later is a bigger set of people.

PR #3697, opened 2021-08-09
adding more options to semgrep sarif formatter

kingbbello:
Added new options to semgrep sarif formatter:

- ability to update shortDescription through metadata field, instead of having duplicates with fullDescription
- ability to add help text through metadata field, as required by [GitHub](https://docs.github.com/en/code-security/secure-coding/integrating-with-code-scanning/sarif-support-for-code-scanning#reportingdescriptor-object) 
- ability to add custom tags other than `cwe` and `owasp`

```
metadata:
      tags: [security, sometag]
      shortDescription: "some short description"
      help: "some help text"
```

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/formatter/sarif.py
+++ b/semgrep/semgrep/formatter/sarif.py
@@ -57,11 +57,11 @@ class SarifFormatter(BaseFormatter):
 
         rule_short_description = rule.metadata.get("shortDescription")
         if rule_short_description:
-            rule_json["shortDescription"]["text"] = rule_short_description
+            rule_json["shortDescription"] = {"text": rule_short_description}
 
         rule_help_text = rule.metadata.get("help")
         if rule_help_text:
-            rule_json["help"] = { "text":  rule_help_text }
+            rule_json["help"] = {"text": rule_help_text}
 
         return rule_json
 
@@ -87,9 +87,9 @@ class SarifFormatter(BaseFormatter):
             owasp = rule.metadata["owasp"]
             result.append(f"OWASP-{owasp}")
 
-        for tags in rule.metadata.get("tags"):
+        for tags in rule.metadata.get("tags", []):
             result.append(tags)
-            
+
         return result
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


semgrep/semgrep/formatter/sarif.py
```diff
@@ -55,6 +55,14 @@ def _rule_to_sarif(rule: Rule) -> Dict[str, Any]:
         if rule_url is not None:
             rule_json["helpUri"] = rule_url
 
+        rule_short_description = rule.metadata.get("shortDescription")
```
brendongo, line 58: We have plans in the next 3-6 months to overhaul the rule schema to include short and long descriptions at the top level as well as tags. For now I think it would be better to not rely on too much of this functionality in the optional meta field of the rule. Can we get away in the interim having the short description and help field just be the rule message as well?

kingbbello: The rule message doesn't give enough information to our developers. We need to provide more context and having everything defined as the message is not sufficient. Can we use this as an interim solution? We can reconsider the approach once the rule schema overhaul is completed.

brendongo: Can you elaborate on what information is missing from a message and how this solves that need? Is there a reason context cannot be added to the message?

brendongo: Messaged out of band: 

> GitHub uses the shortDesription as the title for the alert and fullDescription as the subtitle for the alert. We want to keep the title short and the subtitle a bit longer explaining what exactly the alert is. We’re fine with using the fullDescription as the message but we’d also like to have a separate help text that guides the developers with helpful text and links which is also currently missing. As for the tags, we have tags like security, object-deserialization, external/cwe/cwe-502 etc. that is also not being captured.

> https://docs.github.com/en/code-security/secure-coding/integrating-with-code-scanning/sarif-support-for-code-scanning

brendongo (approved):

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/OSS-1/adding-more-options-to-semgrep-sarif-formatter.

PR #3701, opened 2021-08-10
Ensure we don't accidentally return folders as targets

### LINKED ISSUE #3660: git submodules break scan target identification for "generic" language

mjambon:
This is a take-over of @malexmave's PR https://github.com/returntocorp/semgrep/pull/3690

Fixes #3660

I also added some logging for the tests so we can see what's going on with `pytest -s` (prints captured stdout and stderr even when the test passes).

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Fixed
 - core: Fix parsing of numeric literals in rule files
 - Java: fix the range and autofix of Cast expressions (#3669)
-- Generic mode scanner no longer tries to open submodule folders as files (#3690)
+- Generic mode scanner no longer tries to open submodule folders as files (#3701)
 
 ## [0.61.0](https://github.com/returntocorp/semgrep/releases/tag/v0.61.0) - 2021-08-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 - core: Fix parsing of numeric literals in rule files
+- Generic parser no longer tries to open submodule folders as files
 
 ## [0.61.0](https://github.com/returntocorp/semgrep/releases/tag/v0.61.0) - 2021-08-04
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Fixed
 - core: Fix parsing of numeric literals in rule files
 - Java: fix the range and autofix of Cast expressions (#3669)
-- Generic parser no longer tries to open submodule folders as files (#3690)
+- Generic mode scanner no longer tries to open submodule folders as files (#3690)
 
 ## [0.61.0](https://github.com/returntocorp/semgrep/releases/tag/v0.61.0) - 2021-08-04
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Fixed
 - core: Fix parsing of numeric literals in rule files
 - Java: fix the range and autofix of Cast expressions (#3669)
+- Generic mode scanner no longer tries to open submodule folders as files (#3690)
```
malexmave: You may want to change this reference to #3660 (original issue) or #3701 (new PR).

aryx (approved):

CHANGELOG.md
brendongo:
```suggestion
- Generic mode scanner no longer tries to open submodule folders as files (#3701)
```

brendongo (approved):

CHANGELOG.md
mjambon: thanks for the fix!

PR #3703, opened 2021-08-10
Optimize pattern $X

IagoAbal:
Fixes: bfc4f3da447 ("Revert `pattern: $X` optimization (#3478)")
Fixes: 32e8897523c ("Separate pattern: $X in anded patterns to a special field (#3435)")

test plan:

    % make test # tests included

    % semgrep-core -lang py -config \
            ~/semgrep/semgrep-core/tests/OTHER/rules/pattern-x-1.yaml \
            bench/django/input/django
        #^ now takes ~6 seconds (i.e., almost 10x faster)

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_patterns.ml
+++ b/semgrep-core/src/engine/Match_patterns.ml
@@ -205,7 +205,7 @@ let must_analyze_statement_bloom_opti_failed pattern_strs
 (*****************************************************************************)
 
 (*s: function [[Semgrep_generic.check2]] *)
-let check2 ~hook config rules equivs (file, lang, ast) =
+let check2 ~hook range_filter config rules equivs (file, lang, ast) =
   logger#info "checking %s with %d mini rules" file (List.length rules);
 
   let rules =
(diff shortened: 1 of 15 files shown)


semgrep-core/src/engine/Match_patterns.mli
```diff
@@ -3,11 +3,20 @@
...
   hook:(Metavariable.bindings -> Parse_info.t list Lazy.t -> unit) ->
+  ?range_filter:(Parse_info.token_location * Parse_info.token_location -> bool) ->
   Config_semgrep.t ->
   Mini_rule.rules ->
   Equivalence.equivalences ->
   Common.filename * Lang.t * Target.t ->
   Pattern_match.t list
+(** Match mini-rules (patterns) against a target file.
+  *
+  * @param range_filter A predicate that defines a range of interest when matching
```
emjin: I think (given our conventions) this comment is better in the ml. It's more of an implementation doc

semgrep-core/src/engine/Match_patterns.ml
```diff
@@ -205,7 +205,7 @@ let must_analyze_statement_bloom_opti_failed pattern_strs
 (*****************************************************************************)
 
 (*s: function [[Semgrep_generic.check2]] *)
-let check2 ~hook config rules equivs (file, lang, ast) =
+let check2 ~hook range_filter config rules equivs (file, lang, ast) =
```
emjin, line 216: Nice! I didn't think of this, but it's probably the most idiomatic way to do it with how our code is structured

semgrep-core/src/engine/Specialize_formula.ml
```diff
@@ -25,97 +25,56 @@ type sformula =
...
+          | Some s, None -> (Some s, x :: acc)
+          | None, Some s -> (Some s, acc)
+          | Some s1, Some s2 when selector_equal s1 s2 -> (Some s1, acc)
+          | Some s1, Some _s2 ->
+              (* patterns:
+               * ...
+               * - pattern: $X
+               * - pattern: $Y
+               *)
+              (* TODO: Should we fail here or just reported as a warning? This
```
emjin, line 60: We should definitely catch this with the metachecker. Ideally I think this should be a warning, because while it's probably a mistake, it's a well-formed pattern. There is a hypothetical world where we do more translations of the user's rule so that we enter this state without them writing it. However, this is unlikely and hypothetical, so for now this pattern is always probably a mistake, so if failing is easier to implement I would also be fine with it.

semgrep-core/tests/OTHER/rules/pattern-x-1.yaml
```diff
@@ -0,0 +1,12 @@
+rules:
```
emjin, line 1: I don't know if we already have this, but would be good to have a test for the case that is only `pattern: $X`

emjin (approved):

semgrep-core/src/engine/Match_patterns.mli
IagoAbal: Do you have a pointer to those conventions? Maybe there was a discussion about this when I had not yet joined R2C full-time? The only thing I remember was actually @mjambon raising the issue that we don't have docs for many interfaces, and I believe we had agreed to document them.

IagoAbal: The main question here is what is defined as an implementation detail.

semgrep-core/tests/OTHER/rules/pattern-x-1.yaml
IagoAbal: I will add one!

IagoAbal: Good catch BTW, this is an edge case we should explicitly test for.

semgrep-core/src/engine/Match_patterns.mli
emjin: It's mostly a sense based on PRs. Reading it again I see why this also is an interface doc. Maybe what I'm expressing is more than when I was looking at the check2 function I thought it would be helpful to have a comment over the range filtering logic to explain why this is useful.

IagoAbal: Fair enough! I will add a more implementation-oriented comment to `check2`.

semgrep-core/src/engine/Specialize_formula.ml
IagoAbal: I tried to report the warning/error here and got into some difficulties to do it in a quick and clean way. I'll postpone it but I won't forget about it (-> #3547).

PR #3167, opened 2021-05-21
[Scala] starting alpha support for Scala

aryx:
This closes https://github.com/returntocorp/semgrep/issues/1953

test plan:
test files included

PR checklist:
- [x] changelog is up to date

PR #3704, opened 2021-08-10
[Hack] Add Semgrep extension tests

frankeld:
~⚠️  BUILD WILL FAIL (requires updated semgrep-hack build - see https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/244; once semgrep-hack is updated, I will add another commit to fix the submodule version)~ Done
- Adds many test cases for Semgrep matching
- Updates Parse_hack to support Semgrep extensions
- Misc. parse file updates (type params)

Following https://semgrep.dev/docs/language-support/#maturity-definitions, I think we have reached alpha. If so, I can open another PR to add Hack where relevant. 

PR checklist:
- [ ] changelog is up to date


frankeld:
@mjambon I anticipate the `semgrep unit and e2e tests` failure is because Hack hasn't been added as a supported language in certain files. I will follow along with the python file changes I found [here](https://github.com/returntocorp/semgrep/pull/3167) unless you have any other suggestions.

semgrep/semgrep/target_manager_extensions.py
```diff
@@ -42,6 +43,7 @@
     + C_EXTENSIONS
     + GO_EXTENSIONS
     + RUBY_EXTENSIONS
```
frankeld, line 45: Why isn't PHP here? Are only certain languages added?

mjambon: I don't know either. This file needs some clean-up:
* It looks like `ALL_EXTENSIONS` should be derived from one of the dicts defined below anyway.
* The languages should be ordered alphabetically to make it easier to find a specific language from those lists.

I'd say don't worry about it.

mjambon (approved):
I didn't look into the details. Looks good to me!

Issue #2438, opened 2021-01-21
Add support for semgrep rules that check for specific file path names


rlundinSlack:
**Is your feature request related to a problem? Please describe.**
I am working on creating a custom semgrep ruleset and would like to be able to check the target repository for a specific filepath, directory name, or file name.  For example, in a golang repository, I would like to be able to check for the existence of a "vendor" directory.

**Describe the solution you'd like**
It would be great if I could add a field to the rule's yaml file that looks for a pattern that matches a file path rather than looking for a pattern within the file itself.  For example if I could add the following rule with a "pattern-filepath" field, my specific problem would be solved:

```
rules:
  - id: detect-vendor-directory
    message: "Detected a vendor directory"
    languages: [go]
    severity: INFO
    pattern-filepath: "vendor/"
```

**Describe alternatives you've considered**
@clintgibler mentioned the following existing workaround in the r2c Slack workspace that will work to detect filepaths for the time being.  This workaround helps but is not the most straightforward or simple solution.

```
rules:
  - id: find-go-mod-file
    message: "Detected a go.mod file."
    languages: [generic]
    severity: INFO
    pattern-regex: "^.*"
    paths:
      include:
        - "go.mod"
      exclude:
        - "vendor/"
```

**Additional context**
See r2c Slack workspace's "rules" channel for the original thread re: this feature request.

karoun:
This would be terrific, allowing semgrep to take the place of tools like `repolinter` or `ls-lint`.

DrewDennison:
@mschwager this reminds me of the great work you did using regexs and the boolean logic for `pattern-regex` Do you think it would be feasible to wrap `find` or python's Pathlib in a similar fashion?

Isaac0616:
+1 to this feature, it would be very helpful to my use cases.
In addition, I would love to be able to use regex to match the path.

bertonjulian:
+1 also to this feature (unless there is another way to check this), as we currently use something like this for our rules and it would help to have a rule checking that this file still exists as we want to make sure our rules don't silently fail if devs change the name of the file we are searching in.

```  
paths:
    include:
      - src/apis/authRoutes.ts
```

ievans:
I believe this is mostly solved now with logic under the ‘paths’ key

mschwager:
Following up here, this is the official workaround we use internally:

```yaml
rules:
  - id: found-file
    message: Detected a file
    languages: [generic]
    severity: INFO
[...]

PR #3707, opened 2021-08-10
Fix 'pattern-regex' with completely empty file

### LINKED ISSUE #3705: Crash on empty file with blanket 'pattern-regex'
Using the following `pattern-regex` rule:

```yaml
rules:
- id: any-file
  message: Match any file
  languages:
    - generic
  severity: INFO
  pattern-regex: ^.*
```

With a completely empty file:

```
$ echo -n "" | docker run --rm -v /tmp/test.yaml:/tmp/test.yaml -i returntocorp/semgrep:0.61.0 --config /tmp/test.yaml -
running 1 rules...
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/semgrep/__main__.py", line 20, in main
    cli()
  File "/usr/local/lib/python3.9/site-packages/semgrep/cli.py", line 511, in cli
    semgrep.semgrep_main.main(
  File "/usr/local/lib/python3.9/site-packages/semgrep/semgrep_main.py", line 289, in main
    ignored, returned_errors = rule_match_nosem(rule_match, strict)
  File "/usr/local/lib/python3.9/site-packages/semgrep/semgrep_main.py", line 71, in rule_match_nosem
    if not rule_match.lines:
  File "/usr/local/lib/python3.9/site-packages/semgrep/rule_match.py", line 135, in lines
    result = list(itertools.islice(fd, start_line, end_line))
ValueError: Indices for islice() must be None or an integer: 0 <= x <= sys.maxsize.
```

With a 1-byte file:

```
$ echo -n "a" | docker run --rm -v /tmp/test.yaml:/tmp/test.yaml -i returntocorp/semgrep:0.61.0 --config /tmp/test.yaml -
running 1 rules...
/tmp/tmp89rr7_r3
severity:info rule:tmp.any-file: Match any file
1:a
```

It looks like `rule_match` is expecting 1-indexed line numbers, but we may be receiving 0-indexed:

```python
# Start and end line are one-indexed, but the subsequent slice call is
# inclusive for start and exclusive for end, so only subtract from start
start_line = self.start["line"] - 1
end_line = self.end["line"]
```

mschwager:
Fixes #3705.

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep/semgrep/rule_match.py
+++ b/semgrep/semgrep/rule_match.py
@@ -125,6 +125,10 @@ class RuleMatch:
         start_line = self.start["line"] - 1
         end_line = self.end["line"]
 
+        if start_line == -1 and end_line == 0:
+            # Completely empty file
+            return []
+
         try:
             return self._lines_cache[(start_line, end_line)]
         except KeyError:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - core: Fix parsing of numeric literals in rule files
 - Java: fix the range and autofix of Cast expressions (#3669)
 - Generic mode scanner no longer tries to open submodule folders as files (#3701)
+- `pattern-regex` with completely empty files (#3705)
 
 ### Changed
 - Add backtrace to fatal errors reported by semgrep-core
--- a/semgrep/tests/e2e/rules/pattern-regex-empty-file.yaml
+++ b/semgrep/tests/e2e/rules/pattern-regex-empty-file.yaml
@@ -1,0 +1,7 @@
+rules:
+- id: any-file
+  message: Match any file
+  languages:
+    - generic
+  severity: INFO
+  pattern-regex: ^.*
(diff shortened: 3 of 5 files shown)


semgrep/semgrep/rule_match.py
```diff
@@ -125,6 +125,10 @@ def lines(self) -> List[str]:
         start_line = self.start["line"] - 1
         end_line = self.end["line"]
 
+        if start_line == -1 and end_line == 0:
```
nbrahms, line 128: This smells off to me. Good defensive programming I guess, but semgrep shouldn't return a match here.

brendongo: Should .* not match an empty file?

brendongo (approved):

brendongo:
Going to kick over GithubActions

semgrep/semgrep/rule_match.py
mschwager: > ... but semgrep shouldn't return a match here.

IMO `pattern-regex` is a bit of an outlier, and it seems reasonable for it to return an empty match:

```
In [10]: print(re.search("^.*", ""))                                                                    
<_sre.SRE_Match object; span=(0, 0), match=''>

In [11]: print(re.search("^.+", ""))                                                                    
None
```

This type of rule is also important for our file detection rules (https://github.com/returntocorp/semgrep/issues/2438#issuecomment-891039714). In this case we don't really care if the file is empty or not, only if it exists. So I would vote that we do want to support empty matches, if only for this specific use-case.

PR #3718, opened 2021-08-13
core: Fix fatal error "an equal is already in progress"

IagoAbal:
And added an optimized `Pattern_match.uniq` function to replace
`Common.uniq_by`.

Note that Semgrep returns the same matches as before, but sometimes in
a different order.

This issue was reported by a customer weeks ago but we could not figure
it out without an example. I just happened to hit the same bug when
benchmarking some taint rules.

test plan:

    % make test

    % semgrep-core -timeout 5 -j 1 -profile -lang js \
         -config semgrep-rules/typescript/react/security/audit/react-props-injection.yaml \
         semgrep/parsing-stats/lang/javascript/tmp/mui-org-material-ui/test/bundling/fixtures/next-webpack5

      #^ Try it several times, it no longer triggers a fatal error

Rule react-props-injection contains the following pattern:

    import $PROPS from "...";
    ...

This pattern leads to more than 90k matches in
lang/javascript/tmp/mui-org-material-ui/test/bundling/fixtures/next-webpack5/pages/next-webpack.fixture.js
which causes Common.uniq_by to "explode". If we are unlucky and the
analysis timesout while we are testing the equality of two ASTs, then
`AST_utils.busy_with_equal` is left in the wrong state and all the
subsequent equality tests will lead to a fatal error.

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Generic mode scanner no longer tries to open submodule folders as files (#3701)
 - `pattern-regex` with completely empty files (#3705)
 - `--sarif` exit code with suppressed findings (#3680)
+- Fixed fatal errors when a pattern results in a large number of matches
 
 ### Changed
 - Add backtrace to fatal errors reported by semgrep-core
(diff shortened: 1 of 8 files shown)


aryx (approved):
very nice!

PR #3734, opened 2021-08-19
Allow autofix using the command line

### LINKED ISSUE #840: Allow using autofix via the command line
Sometimes I just want to fix up a pattern once through the codebase. I can search for the problem pattern via the command line but am unable to just fix it as-is

something like 
```
semgrep --lang python --pattern 'self.assertEqual($X,$Y)' --fix 'assert $X == $Y'`
```
**Additional context**

I understand the value of checking in this kind of stuff to a config file. But when doing things like operating on a repo where I have limited control, being able to use `semgrep` is still useful

ievans:
Fixes https://github.com/returntocorp/semgrep/issues/840

```
python -m semgrep -e '$X + 1' --replacement '$X + 2' -l py
...
semgrep/profiling.py
autofix: self._file_num_times_scanned.get(target, 0) + 2
108:            self._file_num_times_scanned.get(target, 0) + 1
...
```

Adding `-a` will apply the patch in place as you'd expect.

PR checklist:
- [x] changelog is up to date


ievans:
Tagging @rtpg, @michaelgruenewald, @msorens, @somebody1234 from the original issue as potentially interested parties. What should the flag be named? I went with `--replacement` but open to suggestion

somebody1234:
er... seems like [autofix is still experimental](https://semgrep.dev/docs/experiments/overview/) (so names are potentially still subject to change?) but it should be named the same as that imo

ievans:
@somebody1234 ironically I got the `--replacement` idea from your comment on #840. @DrewDennison can confirm, but we're planning on promoting autofix from experimental soon -- we intend to keep investing in it. So I think it's ok to put a new flag refrencing it in the CLI. 

Personally I felt like "replacement" was more generic than "fix", which might be what CLI users want.

somebody1234:
i mean... i *am* partial to `--replacement` - but worth nothing the naming in the yaml might need to be reconsidered?

ievans:
OTOH the CLI arguments don't really match the YAML at all (`--expression` vs. `pattern:`). The other reason I liked `replacement` was that we already have the `--autofix` flag which _applies_ replacements. @DrewDennison wdyt?

aryx (approved):
lgtm!

semgrep/semgrep/config_resolver.py
```diff
@@ -485,13 +488,17 @@ def generate_config(fd: IO, lang: Optional[str], pattern: Optional[str]) -> None
 
 
 def get_config(
-    pattern: str, lang: str, config_strs: List[str]
+    pattern: str, lang: str, config_strs: List[str], autofix: Optional[str] = None
```
emjin: Can we make these names consistent? It seems like `fix_expression` is the one we're using in cli

semgrep/semgrep/config_resolver.py
```diff
@@ -68,9 +68,9 @@ def __init__(self, valid_configs: Dict[str, List[Rule]]) -> None:
 
     @classmethod
     def from_pattern_lang(
-        cls, pattern: str, lang: str
+        cls, pattern: str, lang: str, autofix: Optional[str] = None
```
emjin: I think this is also `fixexpression`?

emjin (approved):
Approved but would prefer to reduce the number of names we're using. I count `autofix`, `fix` (from `rules_key`), `replacement`, and `fixexpression`.

PR #3740, opened 2021-08-20
Optimize pattern-inside's ending with `...`

IagoAbal:
This probably won't have a huge perf impact overall due to parsing being
the main bottle neck. However it does help perf and it's a fairly simple
optimization.

test plan:

    % make test

    % semgrep-core -lang js -config ~/semgrep-rules/typescript/react/security/audit/react-css-injection.yaml ~/semgrep/perf/bench/big-js/input/big-js
      #^ Now takes about ~19s (vs ~27s, or ~1.4x faster)
      # -profile reports matching time is ~2s (vs ~10s, or ~5x faster)

    % semgrep-core -profile -lang js \
         -config semgrep-rules/typescript/react/security/audit/react-props-injection.yaml \
         semgrep/parsing-stats/lang/javascript/tmp/mui-org-material-ui/test/bundling/fixtures/next-webpack5
      #^ Now takes about ~0.1s (vs ~1.5s, or ~12x faster)

    % semgrep -j1 --time --max-target-bytes=250000 --config semgrep-rules/typescript/react/security/audit/react-props-injection.yaml semgrep/parsing-stats/lang/javascript/tmp/*
      #^ Match time dropped from ~20s to ~10 (2x faster), but overall
      # there is not much of a speed up since parsing takes ~80% of
      # the time.

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Match_patterns.ml
+++ b/semgrep-core/src/engine/Match_patterns.ml
@@ -109,7 +109,7 @@ let match_sts_sts rule a b env =
             | [] -> env
             | stmt :: _ -> MG.extend_stmts_match_span stmt env
           in
-          GG.m_stmts_deep ~less_is_ok:true a b env))
+          GG.m_stmts_deep ~inside:rule.R.inside ~less_is_ok:true a b env))
   [@@profiling]
 
 (*e: function [[Semgrep_generic.match_sts_sts]] *)
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -137,10 +137,10 @@ let error env msg =
   let err = E.mk_error_loc loc (E.MatchingError s) in
   Common.push err env.errors
 
-let (xpatterns_in_formula : S.sformula -> R.xpattern list) =
+let (xpatterns_in_formula : S.sformula -> (R.xpattern * R.inside option) list) =
  fun e ->
   let res = ref [] in
-  e |> S.visit_sformula (fun xpat -> Common.push xpat res);
+  e |> S.visit_sformula (fun xpat inside -> Common.push (xpat, inside) res);
   !res
 
 let partition_xpatterns xs =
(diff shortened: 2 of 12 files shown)


semgrep-core/src/engine/Match_patterns.ml
```diff
@@ -109,7 +109,7 @@ let match_sts_sts rule a b env =
             | [] -> env
             | stmt :: _ -> MG.extend_stmts_match_span stmt env
           in
-          GG.m_stmts_deep ~less_is_ok:true a b env))
+          GG.m_stmts_deep ~inside:rule.R.inside ~less_is_ok:true a b env))
```
emjin, line 112: Can we put this in the env?

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -137,10 +137,10 @@ let error env msg =
   let err = E.mk_error_loc loc (E.MatchingError s) in
   Common.push err env.errors
 
-let (xpatterns_in_formula : S.sformula -> R.xpattern list) =
+let (xpatterns_in_formula : S.sformula -> (R.xpattern * R.inside option) list) =
```
emjin, line 140: Could we make R.inside part of the R.xpattern type? Or have a type with both?

semgrep-core/src/engine/Match_patterns.ml
emjin: Oh I see `env` is something different. Leaving the comment because I still think this feels a little spaghetti.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1736,6 +1738,17 @@ and m_stmts_deep_uncached ~less_is_ok (xsa : G.stmt list) (xsb : G.stmt list) =
...
   (* dots: '...', can also match no statement *)
   | [ { s = G.ExprStmt ({ e = G.Ellipsis _i; _ }, _); _ } ], [] -> return ()
+  (* inside:
+   * When a pattern-inside (or pattern-not-inside) ends in ... we do not
+   * need to compute all possible endings, it's enough to return the largest
+   * one. If some other pattern P matches inside the largest match here, then
+   * that is enough to either keep P (pattern-inside) or filter it out
+   * (pattern-not-inside).
+   *)
+  | [ { s = G.ExprStmt ({ e = G.Ellipsis _i; _ }, _); _ } ], xb :: bbs
```
emjin, line 1748: Great comment, and optimization. Good to take advantage of knowing that the pattern is for a `pattern-inside`

emjin (approved):

semgrep-core/src/engine/Match_patterns.ml
IagoAbal: Why spaghetti, due to the number of arguments being passed? Eventually we could perhaps pass a record.

semgrep-core/src/engine/Match_rules.ml
IagoAbal: That was the first thing I tried, but that would mean either some ugly hacks (to my taste) in `Parse_rules.ml`, or a cascade of changes in `Rule.ml` and `Parse_rule.ml`. I would probably then change the `Rule.formula_old` type too.  If we do that I would prefer to do it in a separate clean-up PR.

PR #3779, opened 2021-08-30
Switch to Guess_lang for language filtering

mjambon:
This branches switches semgrep-core to doing its own target filtering, in addition to whatever filtering the python wrapper may already be doing.

Uses https://github.com/returntocorp/pfff/pull/464

The next step will be to relax the filtering done by the python wrapper.

PR checklist:
- [x] changelog is up to date


semgrep-core/src/cli/Main.ml
```diff
@@ -946,7 +930,13 @@ let rule_of_pattern lang pattern_string pattern =
 
 (*s: function [[Main_semgrep_core.semgrep_with_one_pattern]] *)
 (* simpler code path compared to semgrep_with_rules *)
-let semgrep_with_one_pattern lang xs =
+(* FIXME: don't use a different a processing logic depending on the output
```
IagoAbal: Small grammar error:
```
  a different
- a
  processing
```

semgrep-core/src/core/ast/Find_target.ml
```diff
@@ -0,0 +1,68 @@
...
+(*
+   Some source files are really huge (> 20 MB) and they cause
+   some annoying 'out of memory' crash that sometimes even the use
+   of mem_limit_mb above does not solve.
+
+   We could configure the size limit based on a per-language basis if we
+   know that a language parser can handle larger files.
+*)
+let filter_by_size _lang paths =
+  let max_bytes = 5_000_000 in
```
IagoAbal: Should this be a CLI option?

semgrep-core/src/core/ast/Find_target.ml
```diff
@@ -0,0 +1,68 @@
...
+      roots
+      |> List.partition (fun path ->
+             Sys.file_exists path && not (Sys.is_directory path))
+    else (roots, [])
+  in
+  let paths = Common.files_of_dir_or_files_no_vcs_nofilter paths in
+  let paths, skipped1 = exclude_files_in_skip_lists paths in
+  let paths, skipped2 = Guess_lang.inspect_files lang paths in
+  let paths, skipped3 = filter_by_size lang paths in
+  let skipped = List.flatten [ skipped1; skipped2; skipped3 ] in
```
IagoAbal: `List.flatten` is not tail-rec, if we scan a very large repos with thousands of files that need to be skipped, this may stack overflow perhaps?

IagoAbal: We have `Common.flatten` which is tail-rec. Here we could even use some `rev_flatten` since you're going to sort `skipped` anyways.

IagoAbal (approved):
Nice!

I didn't see any logic to filter out minimized JS files, I guess that is a problem on its own.

semgrep-core/src/core/ast/Find_target.ml
mjambon: maybe if anyone asks for it. I'm not convinced it's a good hack at all and would rather keep it internal for now.

semgrep-core/src/cli/Main.ml
mjambon: :+1:

semgrep-core/src/core/ast/Find_target.ml
mjambon: yes. Btw, we should automate this with semgrep.

mjambon:
> I didn't see any logic to filter out minimized JS files, I guess that is a problem on its own.

It's definitely something I can and want to do. I just forgot. Meanwhile feel free to merge this PR.

semgrep-core/src/core/ast/Find_target.ml
mjambon: I just created the CLI option.

aryx (approved):

PR #3792, opened 2021-09-01
Free tree-sitter tree when done

emjin:
Integrates changes from https://github.com/returntocorp/ocaml-tree-sitter-core/pull/11

For convenience, all the information in that commit message is here too.

Expose ts_tree_delete and call it when done parsing (at the end of of_ts_tree). Also, remove the fork when calling tree-sitter in tsx (just to check). This should affect all javascript and tsx repos.

Test plan:

Create a yaml file with pattern $X == $Y. Uncomment ts_tree_delete in finalize_tree in semgrep-core/src/ocaml-tree-sitter-core/src/bindings/lib/bindings.c.

Now go to perf/bench/njs-juice (or really any bench)

Run

sc -config /Users/emma/workspace/segfault/metavar.yaml .  -lang js
You should get a segfault.

Checkout this branch

There should be no segfault! You can confirm that the tree is being deleted by printing in octs_tree_delete.

To verify memory leak:

Go to Parse_typescript_tree_sitter (or your language of choice).

Before where it says `let cst = Tree_sitter_tsx.Parse ...`, add

```
          let rec loop () =
          let _cst = Tree_sitter_tsx.Parse.file file in
          loop (); ()
          in
          loop ()
```

Run `sc -config /Users/emma/workspace/segfault/metavar.yaml .  -lang js`. I used the `l50000.js` test in `semgrep-core/perf`, but it shouldn't matter. In another terminal, run top. Verify that memory usage loops. (Important: program should hang)

To be extra sure, checkout develop, put these lines in, look at top. It will not go down (I let it go to 12G+).

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_bash_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_bash_tree_sitter.ml
@@ -1177,7 +1177,7 @@ let parse file =
   H.wrap_parser
     (fun () ->
       Parallel.backtrace_when_exn := false;
-      Parallel.invoke Tree_sitter_bash.Parse.file file ())
+      Tree_sitter_bash.Parse.file file)
     (fun cst ->
       let env = { H.file; conv = H.line_col_to_pos file; extra = () } in
       let bash_ast = program env cst in
@@ -1187,7 +1187,7 @@ let parse_pattern str =
   H.wrap_parser
     (fun () ->
       Parallel.backtrace_when_exn := false;
-      Parallel.invoke Tree_sitter_bash.Parse.string str ())
+      Tree_sitter_bash.Parse.string str)
     (fun cst ->
       let file = "<pattern>" in
       let env = { H.file; conv = Hashtbl.create 0; extra = () } in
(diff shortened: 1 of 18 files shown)


aryx:
awesome!
So it works? No more memory leak (nor segfault)?

IagoAbal:
Why not removing the `invoke` in all the tree-sitter parsers?

emjin:
I will! I just wanted to let you guys look at it (and suggest tests to run, which I will now go do) before I went and deleted all of them. Side-note: really want to be able to do a semgrep-style find-replace in VSCode to change all of them

semgrep-core/src/parsing/tree_sitter/Parse_bash_tree_sitter.ml
```diff
@@ -1177,7 +1177,7 @@ let parse file =
   H.wrap_parser
     (fun () ->
       Parallel.backtrace_when_exn := false;
```
emjin: Oh can we also kill these then?

aryx: You mean the setting of Parallel.backtrace_when_exn? Yes!

aryx (approved):

PR #3795, opened 2021-09-01
Detect and filter out minified files in semgrep-core

mjambon:
This is in addition to the filtering of files with the `.min.js` extension. This will work for all languages, although I don't know if languages other than javascript use this.

Uses https://github.com/returntocorp/pfff/pull/465

PR checklist:
- [x] changelog is up to date


IagoAbal (approved):
Very nice!

brendongo:
Are we only filtering out minified js files or any minified file? And should we have a flag that turns this off?

aryx (approved):

IagoAbal:
> Are we only filtering out minified js files or any minified file? And should we have a flag that turns this off?

@mjambon can correct me if I'm wrong. But it's for filtering out any minimized file, or machine-generated looking file, JS or not. In summary, if the file has _at least_ 1000 bytes, we take a sample of the file and we expect to see a certain ratio of whitespace and line breaks.

Martin has pointed out before that this can be a problem for analyzing Markdown, Asciidoc, etc. Should we do something about it?

IMO we could limit this filtering to files larger than 50k or 100k.

mjambon:
@brendongo 
> Are we only filtering out minified js files or any minified file?

Any language. We could easily make exceptions on a per-language basis, if that's necessary.

> And should we have a flag that turns this off?

Nothing specific to minified files for now. I can add a command-line flag, that's no problem.

More importantly, I think we need a way to enable or disable filtering for a specific target file, on the command line:
* some targets are explicitly requested by the user (of the semgrep CLI) and should not be filtered at all.
* some targets are filtered via a pattern in a rules file e.g. `Dockerfile*` and probably should not be subject to filtering either.
* some targets have no extension, such as scripts, and semgrep-core is the one in charge of finding out if they're in the correct language.

So, semgrep-core should accept two kinds of target files: those subject to filtering and those that bypass filtering. I'm doing this next. Then I'll make semgrep pass scripts (extensionless executable files) to semgrep-core.

mjambon:
@IagoAbal 
> IMO we could limit this filtering to files larger than 50k or 100k.

Maybe you're thinking of performance issues.

One problem with minified files is not their size, it's that they're not source code. Any semgrep finding in a minified file is hard to deal with because it's not source code:
* if semgrep reports a problem, there's no convenient way to fix it. You would have to go to the original authors of that library, who probably know what they're doing anyway and would tell you that their code is fine.
* you can't insert a nosemgrep comment in a minified file, and whatever you do would be overwritten next time the minified library is updated.

ievans:
@mjambon how will the minified file be reported in the output of the python wrapper? Specifically, if I run on a directory that has `foo.min.js` and it is indeed a minified file, will semgrep report "ran 1 rules on 0 files" or "ran 1 rules on 1 files"? I would expect the behavior is the former.

mjambon:
@ievans I haven't done that part yet but I'm thinking of printing this:

```
ran 1 rule on 42 files.
skipped 0 file.
```
or
```
ran 1 rule on 39 files.
skipped 3 files for the following reasons:
  wrong_language: 2
  minified: 1
```

I'm hoping it's clear that "ran 1 rule on 39 files" doesn't include the files that were skipped.

Note that semgrep-core will also do some skipping of certain rules for certain files, but those are performance optimizations. The data is now available to the semgrep wrapper but I don't think we should report that in this summary.

IagoAbal:
> Maybe you're thinking of performance issues.

I was more concerned about false positives, but if you improve the way this is reported by Semgrep CLI then I think it's fine to be more aggressive here.

ievans:
@mjambon that approach sounds good to me! Just for your reference, there are some output mocks @underyx worked on a while back if you're looking for ideas on the output format: https://miro.com/app/board/o9J_lQGE5WM=/?moveToWidget=3074457356225172540&cot=14

mjambon:
@ievans I don't have access to this board: https://miro.com/app/board/o9J_lQGE5WM=/?moveToWidget=3074457356225172540&cot=14

PR #3807, opened 2021-09-02
Hide dangerously flag and remove kwargs

mschwager:
Since `pattern-where-python` has been removed we no longer need this flag. To maintain backwards compatibility, I've only suppressed it, not fully removed it. This means you can still run with the flag, but it's not visible in `-h` and won't actually do anything. I figured the less we have `dangerously_allow_arbitrary_code_execution` floating around in the codebase the better.

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -433,6 +433,11 @@ def cli() -> None:
     if args.dump_ast and not args.lang:
         parser.error("--dump-ast and -l/--lang must both be specified")
 
+    if args.dangerously_allow_arbitrary_code_execution_from_rules:
+        logger.warning(
+            "The '--dangerously-allow-arbitrary-code-execution-from-rules' flag is now deprecated and does nothing. It will be removed in the future."
+        )
+
     output_time = args.time or args.json_time
 
     # set the flags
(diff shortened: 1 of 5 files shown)


semgrep/semgrep/core_runner.py
```diff
@@ -273,11 +273,7 @@ def _run_rules_direct_to_semgrep_core(
                     debug_tqdm_write(f"Running rule {rule.id}...")
                     with tempfile.NamedTemporaryFile(
                         "w", suffix=".yaml"
-                    ) as rule_file, tempfile.NamedTemporaryFile(
-                        "w"
-                    ) as target_file, tempfile.NamedTemporaryFile(
-                        "w"
-                    ) as equiv_file:
+                    ) as rule_file, tempfile.NamedTemporaryFile("w") as target_file:
```
mschwager, line 276: Unrelated - without `equivalences` we no longer need `equiv_file`. No need to open another file every run.

semgrep/semgrep/cli.py
```diff
@@ -418,6 +408,12 @@ def cli() -> None:
         help=argparse.SUPPRESS,
         # help="Legacy pattern recommendation functionality for use in semgrep-app playground",
     )
+    config.add_argument(
+        "--dangerously-allow-arbitrary-code-execution-from-rules",
```
emjin, line 412: What now happens if you run with this flag? I presume nothing?

emjin: I wonder if we should output some kind of message warning users that they are using a deprecated flag.

IagoAbal: I agree with Emma here. I guess we don't want to keep this flag around forever, so we should warn that it is deprecated and will be removed in some future release.

mschwager: Done! I added a warning below :+1:

IagoAbal (approved):

PR #3809, opened 2021-09-02
cli: Improve how semgrep-core fatal errors are reported

### LINKED ISSUE #3774: Improve surfaced semgrep-core messages
Since https://github.com/returntocorp/semgrep/pull/3657 we surface `semgrep-core` messages so users can report and fix them. These are not very good messages. For example:

```
semgrep-core reported a fatal error:
-----
Fatal Error: Parse_info.NoTokenLocation("FakeTokStr")
Raised at Parse_info.token_location_of_info in file "src/pfff/h_program-lang/Parse_info.ml", line 321, characters 28-64
...
Called from Memory_limit.run_with_memory_limit in file "src/system/Memory_limit.ml", line 59, characters 6-62
Called from Dune__exe__Main.iter_files_and_get_matches_and_exn_to_errors.(fun) in file "src/cli/Main.ml", line 630, characters 17-816
-----
Please file a bug report at https://github.com/returntocorp/semgrep/issues/new/choose
```

We should explain what is happening and include the rule/file where this happened so they can file a bug report.

IagoAbal:
1. The error message now specficies rule and target file.
2. Only the key info is displayed in red, the error trace will be
   white/gray, to avoid overwhelming the user with internal details.

Closes #3774

test plan:

    % cat test.yaml
    rules:
    - id: test
        languages:
        - py
        message: Match Found!
        pattern: x
        severity: WARNING
    % cat test.py
    def test():
        return x
        }
    % semgrep -c test.yaml test.py

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Changed
 - Add logging on failure to git ls-files (#3777)
 - Ignore files whose contents look minified (#3795)
+- Display semgrep-core errors in a better way (#3774)
 
 ### Fixed
 - Java: separate import static from regular imports during matching (#3772)
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -673,7 +673,7 @@ and satisfies_metavar_pattern_condition env r mvar opt_xlang formula =
   | None ->
       error env
         (Common.spf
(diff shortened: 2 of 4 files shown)


aryx (approved):

aryx:
test plan could display the new output for the error instead of just giving the command.

IagoAbal:
I wasn't sure it was going to look very different in plain-text, the effect is more evident when the output is colored. I attach here a screenshot (the error trace also looks fine if the terminal's background is dark).

PR #3815, opened 2021-09-03
Fix bugs in `pattern: $X` optimization

IagoAbal:
Fixes: 24e8cabfb30 ("Optimize pattern $X (#3703)")
    
test plan:
make test # test included
    
        % cat test.yaml
        rules:
        - id: test
            languages:
            - py
            message: Match Found!
            patterns:
            - pattern: $X
            - pattern-not: a
            severity: WARNING
        % cat test.py
        def test(x):
            return x
        % semgrep-core -lang py -config test.yaml test.py
        # ^ Now works, whereas previously it resulted in a crash:
        #
        #       Fatal Error: (Failure "empty And; no positive terms in And")

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/engine/Specialize_formula.ml
+++ b/semgrep-core/src/engine/Specialize_formula.ml
@@ -1,16 +1,17 @@
 module R = Rule
 module RM = Range_with_metavars
 module G = AST_generic
-module M = Metavariable
+module MV = Metavariable
 module PM = Pattern_match
 module RP = Report
 
 type selector = {
-  mvar : M.mvar;
+  mvar : MV.mvar;
   pattern : AST_generic.any;
   pid : int;
   pstr : string R.wrap;
 }
+[@@deriving show]
 
 type sformula =
   | Leaf of R.leaf
@@ -20,6 +21,7 @@ type sformula =
    * should always be inside an And to be intersected with "positive" formula.
    *)
   | Not of sformula
+[@@deriving show]
 
 (*****************************************************************************)
 (* Selecting methods *)
@@ -35,7 +37,7 @@ let selector_from_formula f =
   match f with
   | R.Leaf (R.P ({ pat = Sem (pattern, _); pid; pstr }, None)) -> (
       match pattern with
-      | G.E { e = G.N (G.Id ((mvar, _), _)); _ } ->
+      | G.E { e = G.N (G.Id ((mvar, _), _)); _ } when MV.is_metavar_name mvar ->
           Some { mvar; pattern; pid; pstr }
       | _ -> None)
   | _ -> None
--- a/semgrep-core/src/engine/Specialize_formula.ml
+++ b/semgrep-core/src/engine/Specialize_formula.ml
@@ -47,9 +47,12 @@ let split_and :
   xs
   |> Common.partition_either3 (fun e ->
          match e with
+         (* positives *)
+         | R.Leaf (R.P _) | R.And _ | R.Or _ -> Left3 e
+         (* negatives *)
          | R.Not (_, f) -> Middle3 f
-         | R.Leaf (R.MetavarCond (_, c)) -> Right3 c
-         | _ -> Left3 e)
+         (* conditionals *)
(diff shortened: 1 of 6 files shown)


semgrep-core/src/engine/Specialize_formula.ml
```diff
@@ -31,11 +40,22 @@ let selector_equal s1 s2 = s1.mvar = s2.mvar
...
+(* return list of "positive" x list of Not x list of Conds *)
+let split_and :
+    R.formula list -> R.formula list * R.formula list * R.metavar_cond list =
+ fun xs ->
+  xs
+  |> Common.partition_either3 (fun e ->
+         match e with
+         | R.Not (_, f) -> Middle3 f
+         | R.Leaf (R.MetavarCond (_, c)) -> Right3 c
+         | _ -> Left3 e)
```
emjin: imo this is worth enumerating, so that a reader can see what is considered positive. Also, to guard against the off chance we add another negative term.

semgrep-core/src/engine/Specialize_formula.ml
```diff
@@ -31,11 +40,22 @@ let selector_equal s1 s2 = s1.mvar = s2.mvar
 (* Converter *)
 (*****************************************************************************)
 
+(* return list of "positive" x list of Not x list of Conds *)
```
emjin, line 43: I think this comment got garbled? Definitely should have something that specifies this is (positive, negative, conditions)

semgrep-core/src/engine/Specialize_formula.ml
```diff
@@ -31,11 +40,22 @@ let selector_equal s1 s2 = s1.mvar = s2.mvar
...
+let split_and :
+    R.formula list -> R.formula list * R.formula list * R.metavar_cond list =
+ fun xs ->
+  xs
+  |> Common.partition_either3 (fun e ->
+         match e with
+         | R.Not (_, f) -> Middle3 f
+         | R.Leaf (R.MetavarCond (_, c)) -> Right3 c
```
emjin: I feel like we do `R.Leaf (R.MetavarCond (...))` and `R.Leaf (R.P (...))` a lot (at least, in this one file). Very not important but makes me think we would do better to have formula be
```
type formula =
  | P of xpattern * inside option
  | And of tok * formula list (* see Match_rules.split_and() *)
  | Or of tok * formula list
  | Not of tok * formula
  | MetavarCond of tok * metavar_cond
```

emjin: It also feels weird to me to think of MetavarCond as a leaf because you certainly can have more patterns in it, and it doesn't stand alone.

semgrep-core/src/engine/Specialize_formula.mli
```diff
@@ -10,16 +10,24 @@ type selector = {
...
-      (** Invariant: [And (sel_opt, fs)] satisfies
-     * [not (Option.is_some sel_opt) || fs <> []], that is,
-     * we can only select from a non-empty context. *)
+  | And of sformula_and
   | Or of sformula list
   | Not of sformula
 
+and sformula_and = {
+  selector_opt : selector option;
+      (** Invariant: [not (Option.is_some selector_opt) || positives <> []]
```
emjin, line 23: Love this thank you

emjin (approved):

semgrep-core/src/engine/Specialize_formula.ml
IagoAbal: I agree!

semgrep-core/src/engine/Specialize_formula.ml
IagoAbal: I agree too. I think they may have been grouped with the idea that some functions could handle "leafs" uniformly, but actually patterns and conditions are (almost?) always handled differently.

IagoAbal: Hmm, while I agree, if I would do this, I would want to do the same in `Rule.formula`. I leave this one for now.

PR #3826, opened 2021-09-07
Temporarily comment out failing benchmarks

emjin:
PR checklist:
- [x] changelog is up to date

### DIFF
--- a/perf/corpus.py
+++ b/perf/corpus.py
@@ -110,7 +110,8 @@ SMALL_CORPUSES = [
 
 MEDIUM_CORPUSES = [
     # Single rule bench at the origin of the --filter-irrelevant-rules opti
-    Corpus("big-js", "input/semgrep.yml", "input/big-js"),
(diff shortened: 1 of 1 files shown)


mjambon (approved):

aryx (commented):
great!

IagoAbal:
Does the job work now? Perhaps I did something wrong but I tried running locally and I get segfaults.

emjin:
Wait what do you mean locally? (I've never been able to get a segfault on my computer from these benchmarks, if you can that would be very useful for debugging)

emjin:
Ah, actually just now I ran it again and I got one. Unfortunately it is not the same one as the one that happens in CI, but it's still good to be able to produce it. I do see that the benchmarks segfaulted last night (on django, whereas locally I get it on pmd). Maybe I will just comment out everything from django on.

IagoAbal:
Fixed this one: https://github.com/returntocorp/semgrep/pull/3843

PR #3827, opened 2021-09-07
Follow a defined API for semgrep-core return objects

emjin:
This PR cleans up error handling in three main ways:

1. Separates out semgrep-specific error functionality into `Semgrep_error_code.ml`. Going forward, `Semgrep_error_code.ml` should be the single source of truth for all error statuses
2. Generates all JSON via Semgrep_core_response.atd (instead of creating some by hand). This edits the error type in `Semgrep_core_response.atd` so that it always contains the necessary information. See also JSON_report.ml.
3. Merges changes from Brendon's PR to parse the JSON into classes.

Also, it changes `Parse_rule.ml` so that `env` is passed through to all functions possible, setting up for reporting parse errors through the Ocaml code.

For visibility, the error type has changed significantly. It now looks like this:

```
type location = {
  start: position;
  end_ <json name="end">: position;
  lines: string list;
}

(* See Semgrep_error_code *)
type error = {
  ?rule_id: rule_id option;
  error_type: string;
  severity: severity;
  location: location;
  message: string;
  ?details: string option;
  ?yaml_path: string list option;
}
```

Not all errors are now ideal. We still do not report `rule_id` for Timeout, OutOfMemory, and FatalError, even when it is likely available. Such changes will be made in a later PR (see https://github.com/returntocorp/semgrep/issues/3861). This PR mainly concerns reorganizing the code.

Also includes changes from https://github.com/returntocorp/pfff/pull/471

See object via: `sc -config rules.yaml . -lang js -json`

PR checklist:
- [ ] changelog is up to date


emjin:
Will need to coordinate with https://github.com/returntocorp/semgrep/pull/3829 to merge

brendongo (approved):

brendongo:
All tests passing except clabot. Just going to merge this in

semgrep-core/src/core/Semgrep_error_code.ml
```diff
@@ -0,0 +1,247 @@
...
+(* fast = no stack trace *)
+let try_with_print_exn_and_exit_fast file f =
+  try f ()
+  with exn ->
+    let err = exn_to_error file exn in
+    pr2 (string_of_error err);
+    exit 2
+
+(*****************************************************************************)
+(* Helper functions to use in testing code *)
```
IagoAbal, line 196: This doesn't seem very related to error codes?

emjin: It was in Error_code.ml, but you're right, we should probably separate that out.

PR #3831, opened 2021-09-08
Replace argparse with click

nbrahms:
- Sets the stage for further improvements
- Adds a whole bucket of error checking (e.g. you can't now have both --vim and --emacs output)
- Improves help text with sorting, grouping, and better more consistent messages

Also reduce time to run `semgrep --version` from 450 ms to 250 ms.

TODO:
- [x] Add full test coverage


nbrahms:
`semgrep --help` now prints:

```
Usage: semgrep [OPTIONS] [TARGET]...

  Semgrep CLI. Searches TARGET paths for matches to rules or patterns.
  Defaults to searching entire current working directory.

  For more information about Semgrep, go to https://semgrep.dev.

Options:
  -a, --autofix / --no-autofix    Apply autofix patches. WARNING: data loss
                                  can occur with this flag. Make sure your
                                  files are stored in a version control
                                  system. Note that this mode is experimental
                                  and not guaranteed to function properly.
  --enable-metrics / --disable-metrics
                                  Send pseudonymous usage metrics to Semgrep.
                                  If absent, uses the value of the
                                  SEMGREP_SEND_METRICS environment variable;
                                  defaults to no metrics. NOTE: THIS IS
                                  SUBJECT TO CHANGE IN A FUTURE SEMGREP
                                  RELEASE.
  --error / --no-error            Exit 1 if there are findings. Useful for CI
                                  and scripts.
  -l, --lang TEXT                 Parse pattern and all files in specified
                                  language. Must be used with -e/--pattern.
  --optimizations [all|none]      Turn on/off optimizations. Default = 'all'.
                                  Use 'none' to turn all optimizations off.
  --scan-unknown-extensions / --skip-unknown-extensions
                                  If true, explicit files will be scanned
                                  using the language specified in --lang. If
                                  --skip-unknown-extensions, these files will
                                  not be scanned
  --severity [INFO|WARNING|ERROR]
                                  Report findings only from rules matching the
                                  supplied severity level. By default all
                                  applicable rules are run.Can add multiple
                                  times. Each should be one of INFO, WARNING,
                                  or ERROR.
  --strict / --no-strict          Return a nonzero exit code when WARN level
                                  errors are encountered. Fails early if
                                  invalid configuration files are present.
                                  Defaults to --no-strict.
  --use-git-ignore / --no-git-ignore
                                  Skip files ignored by git. Scanning starts
                                  from the root folder specified on the
                                  Semgrep command line. Normally, if the
                                  scanning root is within a git repository,
                                  only the tracked files and the new files
                                  would be scanned. Git submodules and git-
                                  ignored files would normally be skipped.
                                  --no-git-ignore will disable git-aware
                                  filtering. Setting this flag does nothing if
                                  the scanning root is not in a git
                                  repository.
  Configuration options: [mutually_exclusive]
    -c, -f, --config TEXT         YAML configuration file, directory of YAML
                                  files ending in .yml|.yaml, URL of a
                                  configuration file, or Semgrep registry
                                  entry name. See
                                  https://semgrep.dev/docs/writing-rules/rule-
                                  syntax for information on configuration file
                                  format.
    -e, --pattern TEXT            Code search pattern. See
                                  https://semgrep.dev/docs/writing-
                                  rules/pattern-syntax for information on
                                  pattern features.
  Alternate modes:                No search is performed in these modes
    --validate                    Validate configuration file(s). No search is
                                  performed.
    --version                     Show the version and exit.
  Path options: [mutually_exclusive]
    --exclude TEXT                Skip any file or directory that matches this
                                  pattern; --exclude='*.py' will ignore the
                                  following: foo.py, src/foo.py,
                                  foo.py/bar.sh. --exclude='tests' will ignore
                                  tests/foo.py as well as a/b/tests/c/foo.py.
                                  Can add multiple times.
    --include TEXT                Filter files or directories by path. The
                                  argument is a glob-style pattern such as
                                  'foo.*' that must match the path. This is an
                                  extra filter in addition to other applicable
                                  filters. For example, specifying the
                                  language with '-l javascript' might
                                  preselect files 'src/foo.jsx' and
                                  'lib/bar.js'. Specifying one of '--
                                  include=src', '--include=*.jsx', or '--
                                  include=src/foo.*' will restrict the
                                  selection to the single file 'src/foo.jsx'.
                                  A choice of multiple '--include' patterns
                                  can be specified. For example, '--
                                  include=foo.* --include=bar.*' will select
                                  both 'src/foo.jsx' and 'lib/bar.js'. Glob-
                                  style patterns follow the syntax supported
                                  by python, which is documented at
                                  https://docs.python.org/3/library/glob.html
  Performance and memory options: 
    --enable-version-check / --disable-version-check
                                  Checks Semgrep servers to see if the latest
                                  version is run; disabling this may reduce
                                  exit time after returning results.
    -j, --jobs INTEGER            Number of subprocesses to use to run checks
                                  in parallel. Defaults to the number of cores
                                  on the system.
    --max-memory INTEGER          Maximum system memory to use running a rule
                                  on a single file in MB. If set to 0 will not
                                  have memory limit. Defaults to 0.
    --max-target-bytes BYTES      Maximum size for a file to be scanned by
                                  Semgrep, e.g '1.5MB'. Any input program
                                  larger than this will be ignored. A zero or
                                  negative value disables this filter.
                                  Defaults to 1000000 bytes.
    --timeout INTEGER             Maximum time to spend running a rule on a
                                  single file in seconds. If set to 0 will not
                                  have time limit. Defaults to 30 s.
    --timeout-threshold INTEGER   Maximum number of rules that can timeout on
                                  a file before the file is skipped. If set to
                                  0 will not have limit. Defaults to 0.
  Display options: 
    --enable-nosem / --disable-nosem
                                  --enable-nosem enables 'nosem'. Findings
                                  will not be reported on lines containing a
                                  'nosem' comment at the end. Enabled by
                                  default.
    --force-color / --no-force-color
                                  Always include ANSI color in the output,
                                  even if not writing to a TTY; defaults to
                                  using the TTY status
    --max-chars-per-line INTEGER  Maximum number of characters to show per
                                  line.
    --max-lines-per-finding INTEGER
                                  Maximum number of lines of code that will be
                                  shown for each match before trimming (set to
                                  0 for unlimited).
    -o, --output TEXT             Save search results to a file or post to
                                  URL. Default is to print to stdout.
    --rewrite-rule-ids / --no-rewrite-rule-ids
                                  Rewrite rule ids when they appear in nested
                                  sub-directories (Rule 'foo' in
                                  test/rules.yaml will be renamed 'test.foo').
    --time / --no-time            Include a timing summary with the resultsIf
                                  output format is json, provides times for
                                  each pair (rule, target).
  Output formats: [mutually_exclusive]
                                  Uses ASCII output if no format specified.
    --emacs                       Output results in Emacs single-line format.
    --json                        Output results in JSON format.
    --junit-xml                   Output results in JUnit XML format.
    --sarif                       Output results in SARIF format.
    --vim                         Output results in vim single-line format.
  Verbosity options: [mutually_exclusive]
    -q, --quiet                   Only output findings.
    -v, --verbose                 Show more details about what rules are
                                  running, which files failed to parse, etc.
    --debug                       All of --verbose, but with additional
                                  debugging information.
  Test and debug options: 
    --test                        Run test suite.
    --test-ignore-todo / --no-test-ignore-todo
                                  If --test-ignore-todo, ignores rules marked
                                  as '#todoruleid:' in test files.
    --dump-ast / --no-dump-ast    If --dump-ast, shows AST of the input file
                                  or passed expression and then exit (can use
                                  --json).
    --dryrun / --no-dryrun        If --dryrun, does not write autofixes to a
                                  file. This will print the changes to the
                                  console. This lets you see the changes
                                  before you commit to them. Only works with
                                  the --autofix flag. Otherwise does nothing.
  --help                          Show this message and exit.
```

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/bad-three-layers/error-in-color.txt
```diff
@@ -1,13 +1,13 @@
-[31msemgrep error[39m: Invalid rule schema
```
brendongo, line 1: What are these 39m to 0m?

semgrep/tests/e2e/snapshots/test_exclude_include/test_exclude_include/include-and-exclude/results.json
```diff
@@ -1,43 +1,6 @@
...
-      "path": "targets/exclude_include/included/included.js",
-      "start": {
-        "col": 13,
-        "line": 3
-      }
-    }
-  ]
-}
\ No newline at end of file
+Usage: python -m semgrep [OPTIONS] [TARGET]...
```
brendongo: Exclude and include can be used at the same time

semgrep/Pipfile
```diff
@@ -11,4 +11,6 @@ pytest-xdist = "*"
 tox = "*"
 
 [packages]
+click = "~=8.0.1"
```
brendongo, line 14: These need to be put in setup.py for packaging

semgrep/semgrep/cli.py
```diff
@@ -1,449 +1,483 @@
...
         print(__VERSION__)
-        if not args.disable_version_check:
+        if enable_version_check:
+            from semgrep.version import version_check
+
             version_check()
         return
 
-    if args.enable_metrics:
+    # To keep version runtime fast, we defer non-version imports until here
```
brendongo, line 452: <3

brendongo (approved):

semgrep/tests/e2e/snapshots/test_exclude_include/test_exclude_include/include-and-exclude/results.json
nbrahms: According to the docs they can't though?

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/bad-three-layers/error-in-color.txt
nbrahms: This is an artifact of how we were using Colorama, changing which reset token we use now that we are printing with click.

It doesn't change what the user sees... unless they're terminal was already in a different color. But click should now be doing the TM right thing.

semgrep/Pipfile
nbrahms: Ah, in bento we pulled this in automatically. I'll investigate doing that here.

semgrep/semgrep/semgrep_main.py
```diff
@@ -183,10 +185,10 @@ def main(
     configs_obj, errors = get_config(pattern, lang, configs)
     all_rules = configs_obj.get_rules(no_rewrite_rule_ids)
 
-    if severity is None or severity == []:
+    if not severity:
         filtered_rules = all_rules
     else:
-        filtered_rules = [rule for rule in all_rules if rule.severity in severity]
+        filtered_rules = [rule for rule in all_rules if rule.severity in set(severity)]
```
nbrahms: this should be undone

semgrep/tests/e2e/snapshots/test_exclude_include/test_exclude_include/include-and-exclude/results.json
brendongo: Where? `semgrep --pattern '$X==$X' -l ts --include ./semgrep-core/tests/ts/ --exclude ./semgrep-core/tests/ts/misc_ast.ts  .` vs `semgrep --pattern '$X==$X' -l ts --include ./semgrep-core/tests/ts/  .` behaves as I would expect

nbrahms: updated per ^^

PR #3833, opened 2021-09-08
Allow disabling the vardef-assign equivalence

IagoAbal:
test plan:
make test # tests included

PR checklist:
- [x] changelog is up to date


mjambon (approved):

semgrep-core/src/core/ast/Visitor_AST.ml
```diff
@@ -1003,18 +1002,6 @@ let (mk_visitor : visitor_in -> visitor_out) =
     let v_vinit = v_option v_expr v_vinit in
     let v_vtype = v_option v_type_ v_vtype in
     ()
-  and v_vardef_as_assign_expr ventity = function
-    | VarDef ({ vinit = Some _; _ } as vdef) ->
-        (* A VarDef is implicitly a declaration followed by an assignment expression,
-         * so we should visit the assignment expression as well.
-         *
-         * Note that we cannot treat this as a simple equivalence later, as
```
aryx: hmm how did you solve this point?

semgrep-core/src/engine/Match_patterns.ml
```diff
@@ -270,7 +271,25 @@ let check2 ~hook range_filter config rules equivs (file, lang, ast) =
...
+                    * typically do side-effectful things and VarDefs were
+                    * visited twice (as a VarDef and as an Assign).
+                    *
+                    * If we just add an ExprStmt-pattern to stmt_rules,
+                    * the "vardef ==> assign" equivalence in
+                    * Generic_vs_generic.m_stmt will take care of the rest.
+                    * This will lead to some duplicate matches but that
+                    * should be OK (we already deal with duplicates). *)
+                   let s_pattern = s (ExprStmt (pattern, tok)) in
+                   push_with_annotation any s_pattern stmt_rules
```
aryx: hmm but this will work only if the pattern is a single assign.
What if we have a more complex pattern like
```
foo()
...
x  = 1
```

Will this match code like
```
foo()
var x = 1
```
?

aryx (approved):
I'm suprised this does not introduce any CI failure ...
I thought we had many rules doing things like foo()... x = 1;
that we wanted to match foo()... var x = 1;
but maybe not in the end.

semgrep-core/src/engine/Match_patterns.ml
aryx: Oh I got it, it's because we already do that something in m_stmts. It was written in your comment actually.
All good!
I don't even understand why we needed this visitor trick in the first place anymore ,,,

semgrep-core/src/engine/Match_patterns.ml
```diff
@@ -270,7 +271,25 @@ let check2 ~hook range_filter config rules equivs (file, lang, ast) =
...
+                   (* vardef-assign: let `$X = $E` match e.g. `var x = 1;`
+                    *
+                    * old: We used to have this equivalence built into
+                    * Visitor_AST. This was error prone because visitors
+                    * typically do side-effectful things and VarDefs were
+                    * visited twice (as a VarDef and as an Assign).
+                    *
+                    * If we just add an ExprStmt-pattern to stmt_rules,
+                    * the "vardef ==> assign" equivalence in
+                    * Generic_vs_generic.m_stmt will take care of the rest.
```
aryx: hmm maybe we need to use config.Config.vardef_assign in m_stmts too then?

IagoAbal: True, that's right!

semgrep-core/src/core/ast/Visitor_AST.ml
IagoAbal: If I understood this note correctly, it is saying that expression patterns are only tried against expressions, and statement patterns are only tried against statements, and that's why we can't use an equivalence to match `$X = $E` against `var x = 1;`. But that can be solved by matching both `$X = $E` and also `ExprStmt ($X = $E)`. That is, if I understood the note correctly!

semgrep-core/src/engine/Match_patterns.ml
IagoAbal: I was thinking that this could indeed not work for deep expression patterns, e.g. matching `<... x = 1 ...>` against `var x = 1;`, but it seems that that did not work before either.

IagoAbal: Oh I was making a silly mistake when testing this, yeah `<... x = 1 ...>` used to work and now it doesn't.

IagoAbal: Hmmm, I can't think of any other case than `<... $X = $E ...>` now but there could be others. Maybe we can just configure whether we want this equivalence or not when we create the visitor.

aryx: if CI is passing, maybe we should just go for it even though we know it's more limited than before.
The cleaniness of Visitor_AST.ml seems more important to me, especially if there are no semgrep-rules
that fail with the change.
 (not sure why CI has failures now. Your first PR didn't have any).

IagoAbal: Oh the failure it's because I added a new test case and I used tabs instead of spaces in the YAML file.

IagoAbal: I updated this PR and took care of `<... $X = $E ...>`. If the pattern occurs in a place where we expect a statement then it's already going to be parsed as a `ExprStmt`; and if it occurs in a place where we expect an expression, we should not need the equivalence... unless that expression is a `Lambda` or something that contains statements.

I start to like more the alternative of adding a `bool` parameter to `mk_visitor`... I did the work here: https://github.com/returntocorp/semgrep/commit/daca5cceb7929e8e3a932e7219a7fafaf5672132. Anyways `Visitor_AST` still needs to deal with e.g. `Partial*`, so by taking out `vardef-assign` we are not simplifying it that much.

IagoAbal:
I finally went for the more conservative solution, which is to keep the VarDef-Assign equivalence as part of Visitor_AST, but to allow disabling it through an optional `vardef_assign` flag. The default value for this flag is `false`, so no VarDef-Assign equivalence is applied in Visitor_AST unless explicitly requested. In the places where this flag matters I set it to the desired value, in the places where the value does not matter (AFAICT) I let it use the default (=`false`).

The "trick" in Match_patterns should work but I did not feel 100% satisfied with it. And, because we still deal with PartialDefs in Visitor_AST, we were not simplifying it that much by taking VarDef-Assign out of it. So, the safer move for me was to change as little as possible to achieve the goal of making Visitor_AST less error prone when VarDefs are involved.

Ideally we should find a solution that works for both, VarDef-Assign and PartialDefs, at once. Ideally, it should all be done in Generic_vs_generic... but then maybe Match_patterns should just do `any`-to-`any` matching. Let's take it a fresh look at it in the future.

PR #3834, opened 2021-09-08
Switch to Alcotest for unit testing

mjambon:
Uses https://github.com/returntocorp/pfff/pull/468

It looks like this when successful:

Screenshot_20210908_051534

And it looks like this in case of an error:

Screenshot_20210908_052153

PR checklist:
- [x] changelog is up to date


semgrep-core/src/testutil/Testutil.mli
```diff
@@ -0,0 +1,82 @@
...
+   value as its output, it's just easier to fail with 'assert false'.
+
+   In any case, Alcotest will capture the output (stdout, stderr) of each
+   test and put it in its own file so we can consult it later. Don't
+   hesitate to log a lot during the execution of the test.
+*)
+type test = string * (unit -> unit)
+
+(*
+   This extends the name of each test by adding a prefix, using '.'
```
aryx: I thought the separator was '>' in the end, not '.'.

aryx (approved):

emjin:
Looks great! Can we merge?

emjin (approved):

semgrep-core/src/testutil/Testutil.mli
mjambon: thx! I updated the doc.

IagoAbal:
Nice!

PR #3837, opened 2021-09-08
Add type inference for struct literals in Go

### LINKED ISSUE #3622: ($X : TypeName) only matches variables declared with `var x TypeName`
**Describe the bug**

In Go, patterns matching method invocations on variables of a given type work when the variable declared using `var x TypeName` but not when using `x := TypeName{}` or `var x = TypeName{}`.

**To Reproduce**

https://semgrep.dev/s/WO5E

Pattern: `($X : http.Server).ListenAndServe`

Sample input (I would expect all three to be matched):

```
func foo() {
    var a http.Server
    a.ListenAndServe() // match, as expected

    b := http.Server{}
    b.ListenAndServe() // no match, but match expected

    var c = http.Server{}
    c.ListenAndServe() // no match, but match expected
}
```

**Expected behavior**

Any variable of the given type should be matched, not just those declared with `var x TypeName`.

**Environment**

Local

mmcqd:
Closes #3622 
Struct literals in Go like http.Server{} (with or without initialized fields) now have their types inferred 
Test plan: sc -f metavar_typed_struct.sgrep metavar_typed_struct.go -lang go

PR checklist:
- [x] changelog is up to date


mjambon (approved):
I'm not familiar with this code but it looks good to me (in the sense that it's not much code, it has new tests and no regressions). Given that the generic AST is meant to work for all languages, I agree with your comment about not checking if the language is Go. @aryx would know better.

semgrep-core/src/analyzing/Typing.ml
```diff
@@ -33,4 +33,14 @@ let get_resolved_type lang (vinit, vtype) =
       | Some { e = L (Null tok); _ } -> make_type "null" tok
       | Some { e = L (Imag (_, tok)); _ } -> make_type "imag" tok
       | Some { e = N (Id (_, { id_type; _ })); _ } -> !id_type
+      | Some
+          {
+            e = Call ({ e = IdSpecial (New, _); _ }, (_, ArgType tp :: _, _));
+            _;
+          } -> (
+          match lang with
```
IagoAbal: Yeah I wouldn't do the language check. This should e.g. work for Java too. If it matches that pattern I think it's safe toreturn `tp` as the type.

semgrep-core/tests/go/metavar_typed_struct.go
```diff
@@ -0,0 +1,24 @@
...
+	d := http.Server{Addr:":8080"}
+	//ERROR: match
+	d.ListenAndServe()
+
+	var e = http.Server{Addr:":8080",ReadTimeout:10}
+	//ERROR: match
+	e.ListenAndServe()
+
+	var f = http.Client{}
+	f.ListenAndServe()
```
IagoAbal, line 24: Although it really doesn't do anything AFAICT, we tend to annotate these lines with `//OK:` to stress that there should be no match there. Not a big deal.

IagoAbal (approved):
Neat!

I would just remove the language check, as you have pointed out it should work for other languages too.

perf/variant.py
```diff
@@ -26,18 +26,14 @@ def __init__(self, name: str, semgrep_core_extra: str, semgrep_extra: str = ""):
     SemgrepVariant("max-cache", "-opt_max_cache"),
     SemgrepVariant("no-bloom", "-no_bloom_filter"),
     SemgrepVariant("no-gc-tuning", "-no_gc_tuning"),
+    SemgrepVariant("no-filter-irrelevant-rules", "-no_filter_irrelevant_rules"),
```
IagoAbal, line 29: Does this change belong to this PR?

semgrep-core/tests/go/metavar_typed_struct.go
```diff
@@ -0,0 +1,25 @@
...
+
+	d := http.Server{Addr:":8080"}
+	//ERROR: match
+	d.ListenAndServe()
+
+	var e = http.Server{Addr:":8080",ReadTimeout:10}
+	//ERROR: match
+	e.ListenAndServe()
+
+	//OK:
```
IagoAbal: I think the `OK` is the next line.

perf/variant.py
mmcqd: Not really, but I had to make this change to prevent some benchmarks from failing. See @emjin 's comment here https://github.com/returntocorp/semgrep/pull/3844

emjin: I'm going to just merge this, in the future will make separate PR in cases like this

IagoAbal: Totally fine, it just looked strange.

aryx (commented):
This is great!
Thx.

PR #3839, opened 2021-09-08
Remove legacy metavariable interpolation code

brendongo:
We had to keep track of all metavariables because we used to run
semgrep-core over single patterns instead of over entire rules. This
meant that some patterns did not have the metavariable a rule was using.

Now that we run over the entire rule, match object returned by
semgrep-core should have all the metavariables needed for interpolation.

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep/semgrep/evaluation.py
+++ b/semgrep/semgrep/evaluation.py
@@ -1,4 +1,3 @@
-from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Set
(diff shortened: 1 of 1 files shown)


semgrep/semgrep/evaluation.py
```diff
@@ -12,73 +11,35 @@
...
-
-
 def create_output(
     rule: Rule,
     pattern_matches: List[PatternMatch],
     valid_ranges_to_output: Optional[Set[Range]] = None,
 ) -> List[RuleMatch]:
     output = []
-
-    if valid_ranges_to_output is None:
```
brendongo, line 36: This is always None since the only place we call create_output does not set this value. So this and line 52 are essentially no-ops now.

semgrep/semgrep/evaluation.py
```diff
@@ -1,84 +1,41 @@
...
+            fix=fix,
+            fix_regex=rule.fix_regex,
+        )
+        output.append(rule_match)
 
     return sorted(output, key=lambda rule_match: rule_match._pattern_match.range.start)
 
 
 def interpolate_string_with_metavariables(
-    text: str, pattern_match: PatternMatch, propagated_metavariables: Dict[str, str]
```
emjin, line 81: Yaaaaaay

semgrep/semgrep/evaluation.py
```diff
@@ -88,8 +45,4 @@ def interpolate_string_with_metavariables(
             metavariable, pattern_match.get_metavariable_value(metavariable)
         )
 
-    # Sort by metavariable length to avoid name collisions (eg. $X2 must be handled before $X)
```
emjin, line 91: Did this use to just happen twice?

emjin (approved):

emjin:
I'm not super familiar, but what you say seems correct (the metavariable information is always available), this passes tests, and the new code is soooo much cleaner.

PR #3841, opened 2021-09-09
Reduce usage of pattern_match object

brendongo:
Now that we run whole rules in semgrep-core this class is unnecessary. After this PR the only remaining
usage is in converting semgrep-core output to RuleMatch objects. It's a little involved for now to break
that part out so will do that in a separate PR.


semgrep/tests/e2e/snapshots/test_join_rules/test_join_rules/rulesjoin_rulesmultiple-rules.yaml-join_rulesuser-input-with-unescaped-extension/results.json
```diff
@@ -32,6 +32,25 @@
         "lines": "        <li>person_name_full is <b>{{ person_name_full }}</b></li>",
         "message": "The variable '$VAR' is most likely an XSS. This variable originates from user input and is rendered in an unescaped manner. An attacker could control this variable and input scripts onto rendered pages, resulting in all manner of bad juju. The best fix is to make sure your template extensions end in '.html', which automatically escapes rendered variables.",
         "metadata": {},
+        "metavars": {
```
brendongo, line 35: Not sure how my change had this affect on join --json output but seems fine to have metavar info included in extra field of a join finding? @minusworld ?

brendongo:
Handled in https://github.com/returntocorp/semgrep/pull/3847/files

PR #3843, opened 2021-09-09
use_parsing_cache: Handle SIGXFSZ (file size limit exceeded)

IagoAbal:
Closes PA-318

test plan:

    % semgrep-core -lang js -j 1 -e '$X == $X' -fast \
          -use_parsing_cache ./cache \
          ~/semgrep/perf/bench/juice-shop/input/juice-shop/
      #^ no longer gets killed by SIGXFSZ (Apple M1, MacOS 11)
      # and with -j 2 we no longer get a segfault

PR checklist:
- [x] changelog is up to date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -402,7 +402,14 @@ let cache_computation file cache_file_of_file f =
           res)
         else
           let res = f () in
-          Common2.write_value (Version.version, file, res) file_cache;
+          (try Common2.write_value (Version.version, file, res) file_cache
+           with Sys_error err ->
+             (* We must ignore SIGXFSZ to get this exception, see
+              * note "SIGXFSZ (file size limit exceeded)". *)
+             logger#error "Could not write cache file for %s (%s): %s" file
+               file_cache err;
+             (* Make sure we don't leave corrupt cache files behind us. *)
+             if Sys.file_exists file_cache then Sys.remove file_cache);
           res)
 
 let cache_file_of_file filename =
@@ -1458,6 +1465,21 @@ let options () =
 
 (*s: function [[Main_semgrep_core.main]] *)
 let main () =
+  (* SIGXFSZ (file size limit exceeded)
+   * ----------------------------------
+   * By default this signal will kill the process, which is not good. If we
+   * would raise an exception from within the handler, the exception could
+   * appear anywhere, which is not good either if you want to recover from it
+   * gracefully. So, we ignore it, and that causes the syscalls to fail and
+   * we get a `Sys_error` or some other exception. Apparently this is standard
+   * behavior under both Linux and MacOS:
+   *
+   * > The SIGXFSZ signal is sent to the process. If the process is holding or
+   * > ignoring SIGXFSZ, continued attempts to increase the size of a file
+   * > beyond the limit will fail with errno set to EFBIG.
+   *)
+  Sys.set_signal Sys.sigxfsz Sys.Signal_ignore;
+
   profile_start := Unix.gettimeofday ();
 
   let usage_msg =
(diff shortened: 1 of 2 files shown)


semgrep-core/src/cli/Main.ml
```diff
@@ -402,7 +402,23 @@ let cache_computation file cache_file_of_file f =
...
+          let cleanup_cache () =
+            if Sys.file_exists file_cache then Sys.remove file_cache
+          in
+          Sys.set_signal Sys.sigxfsz
+            (Sys.Signal_handle
+               (fun _ ->
+                 logger#error
+                   "Could not write cache file for %s: file size limit exceeded"
+                   file;
+                 cleanup_cache ()));
```
mjambon: I think this signal handler is unsafe because it's going to execute asynchronously, similar to a thread. If you want to write a message to stderr, use `Unix.write` (unbuffered write) instead of the buffered write that `logger#error` presumably uses. In other words, you probably don't want to do that at all. The simplest thing to do is to have the signal handler raise an exception, and have this exception be caught in the normal execution flow (main thread).

Also, registering this signal handler should probably be done once and for all during program initialization (wherever we do these things).

e.g.
```
exception File_size_limit_exceeded
let () =
  Sys.signal_handler Sys.sigxfsz
    (Sys.Signal_handle (fun _signal -> raise File_size_limit_exceeded))
```
and then catch `File_size_limit_exceeded` where you think it's relevant.

IagoAbal: Great feedback, I'll fix that.

IagoAbal: If we raise an exception inside the handler this can pop up anywhere and that is not very useful for the purpose of recovering from this error. It seems that if you simply ignore it, that will cause the sys call to fail, and you get a `Sys_error` instead, from which it is easy to catch and recover from. Let me know what do you think!

semgrep-core/src/cli/Main.ml
```diff
@@ -1458,6 +1463,14 @@ let options () =
 
 (*s: function [[Main_semgrep_core.main]] *)
 let main () =
+  (* SIGXFSZ (file size linit exceeded) by default will kill the process, which
```
aryx: typo, limit not linit.

semgrep-core/src/cli/Main.ml
```diff
@@ -402,7 +402,12 @@ let cache_computation file cache_file_of_file f =
           res)
         else
           let res = f () in
-          Common2.write_value (Version.version, file, res) file_cache;
+          (try Common2.write_value (Version.version, file, res) file_cache
+           with Sys_error err ->
```
aryx, line 406: maybe a comment here that talks about SIGXFSZ?
What is the size limit actually? Why do we reach it? Our ASTs are too big?

IagoAbal: I don't know exactly how this happens but in some shells `ulimit -f` shows me `unlimited` whereas in others shows me `10000` (file blocks). I certainly have not set that limit myself, it's very strange. I get the error on a file whose AST dump has a size of 19,993,228 bytes (so not that large).

semgrep-core/src/cli/Main.ml
mjambon: If that's how it works, let's try this.

mjambon (approved):

semgrep-core/src/cli/Main.ml
emjin: Just to check, do you get this error always or when `ulimit -f` shows you `10000`?

emjin (approved):

semgrep-core/src/cli/Main.ml
IagoAbal: If there is no file size limit then I don't get the error.

IagoAbal:
Initially I thought that we were creating huge files in some cases. It turns out it is not the case. But, somehow, something is setting a file size limit in some of my terminals triggering this problem. I wonder if it is any of our internal scripts, `pipenv`, or some other tool we use. Because that's all I'm running on this laptop.

Anyways, I will still merge this because the changes are minimal, and because it's still a good idea to be defensive, catch errors during writing, and clean up the cache files to avoid corruption. But if you dislike it feel free to revert.

mjambon:
+1 on being defensive

PR #3844, opened 2021-09-09
Fix broken benchmark args

nbrahms:
--optimizations needs an argument; previously it was a no-op, but now semgrep kicks
out a warning and exits

### DIFF
--- a/perf/variant.py
+++ b/perf/variant.py
@@ -28,10 +28,8 @@ SEMGREP_VARIANTS = [
     SemgrepVariant("no-gc-tuning", "-no_gc_tuning"),
     # alternate optimisations
     SemgrepVariant("set_filters", "-set_filter"),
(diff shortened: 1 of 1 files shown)


emjin:
This was causing @mmcqd's PR to fail we made essentially these changes there: https://github.com/returntocorp/semgrep/pull/3837. In hindsight perhaps should have been a separate PR, but it's ready to merge in any case.

brendongo (approved):

emjin:
Closing in favor of change made in https://github.com/returntocorp/semgrep/pull/3837

brendongo:
@emjin i think this made changes that werent in #3837?

emjin:
Same idea; I prefer to just delete the `optimizations` variant because it's now the default and we have made it difficult to change back.

PR #3856, opened 2021-09-13
Refactor nosemgrep into its own file

nbrahms:
This improves maintainability, and allows us to start hacking on semgrep ignore functionality


semgrep/semgrep/ignores.py
```diff
@@ -0,0 +1,106 @@
...
+
+from semgrep.constants import COMMA_SEPARATED_LIST_RE
+from semgrep.constants import NOSEM_INLINE_RE
+from semgrep.error import Level
+from semgrep.error import SemgrepError
+from semgrep.output import logger
+from semgrep.output import OutputHandler
+from semgrep.rule_match import RuleMatch
+from semgrep.types import RuleMatchMap
+
```
mjambon, line 25: What is this module for? Please add a header that explains this module's intended role and scope for any future visitor.

examples:
* https://github.com/python/cpython/blob/1fc41ae8709e20d741bd86c2345173688a5e84b0/Lib/getopt.py#L1-L15
* https://github.com/python/cpython/blob/1fc41ae8709e20d741bd86c2345173688a5e84b0/Lib/gzip.py#L1-L6
* https://github.com/python/cpython/blob/1fc41ae8709e20d741bd86c2345173688a5e84b0/Lib/warnings.py#L1

mjambon (approved):

semgrep/semgrep/ignores.py
nbrahms: left some documentation

PR #3857, opened 2021-09-13
Refactoring AST_generic.function_definition.fbody

aryx:
This refactoring makes fbody more precise and relies less
on empty blocks to encode special function body.

test plan:
make test

PR checklist:
- [x] changelog is up to date


IagoAbal (approved):

IagoAbal:
At some point we may want to rename AST_generic as CST_generic. 😄

aryx:
I think you're right ... it's getting more and more like a CST indeed. There are just the commas missing and a few middle tokens
and some unsugaring with multivariable declaration and that's it.

PR #3858, opened 2021-09-13
Add wrapper script in perf/ for benchmarking a local semgrep vs. released semgrep

minusworld:
command: `./benchmark-against-version.py --against 0.60.0 --local <path-to-local-semgrep> --std-only --dummy

Note that `path-to-local-semgrep` should be the top level directory when you `git clone semgrep`. Pointing to this path should install the most-recently built version of semgrep-core (from the path `semgrep/semgrep/bin/semgrep-core`).

PR checklist:
- [ ] changelog is up to date


perf/benchmark-against-version.py
```diff
@@ -0,0 +1,108 @@
+#! /usr/bin/env python3
+import logging
+import os
+import subprocess
+import sys
+import venv
+from pathlib import Path
+from typing import Optional
+
```
mjambon, line 9: ^ It would be helpful to have a brief explanation of what this program is for at the beginning of the file.

mjambon: If the script is executable, then it shouldn't have a `.py` extension (because it's a command and it doesn't matter to the user in which language it's implemented).

mjambon (approved):

PR #3876, opened 2021-09-14
Return stack trace printing for FatalErrors

brendongo:
PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep/semgrep/core_output.py
+++ b/semgrep/semgrep/core_output.py
@@ -105,6 +105,7 @@ class CoreError:
     message: CoreErrorMessage
     level: Level
     spans: Optional[Tuple[LegacySpan, ...]]
+    details: Optional[str]
 
     @classmethod
     def parse(cls, raw_json: JsonObject) -> "CoreError":
@@ -122,6 +123,8 @@ class CoreError:
             level_str = "WARN"
         level = Level[level_str.upper()]
 
+        details = raw_json.get("details")
+
         # TODO legacy support for live editor pattern parse highlighting
         spans = None
         if "yaml_path" in raw_json:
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/core_output.py
```diff
@@ -122,14 +123,18 @@ def parse(cls, raw_json: JsonObject) -> "CoreError":
             level_str = "WARN"
         level = Level[level_str.upper()]
 
+        details = raw_json.get("details")
```
emjin, line 126: Wait what does this do if details isn't there?

brendongo: Returns None

brendongo: So https://github.com/returntocorp/semgrep/pull/3876#diff-ac75fcfbab642d994ff6593c78b023777d5cd0bb672397b73bbe950f96e26bc6R145 will be used

emjin (approved):
Does this use Iago's nice printing?

brendongo:
Yeah I followed the format in https://github.com/returntocorp/semgrep/pull/3809 but kept it just all red cause white text probably wont show up nicely on white terminals.

emjin:
> Yeah I followed the format in #3809 but kept it just all red cause white text probably wont show up nicely on white terminals.

I remember the text being a light gray, which I thought worked nicely (doesn't overwhelm the person running Semgrep, but they can see if it they need)

IagoAbal:
> Yeah I followed the format in #3809 but kept it just all red cause white text probably wont show up nicely on white terminals.

See https://github.com/returntocorp/semgrep/pull/3809#issuecomment-912363781. I couldn't find gray color defined in that Python library but it seems to do the right thing depending on the color theme of the terminal.

IagoAbal (approved):

brendongo:
I see. I'll change it back to white

PR #3878, opened 2021-09-15
Report rule_id whenever possible

### LINKED ISSUE #3861: Semgrep-core does not report correct rule_id in timeout and maxmemory errors
Print out semgrep-core output of `semgrep --verbose --max-memory 1 --config tests/e2e/rules/long.yaml tests/e2e/targets/equivalence` and you will find that rule_id is 1

Should fix semgrep-core to place correct rule_id. For now, since we only run one rule at a time on semgrep-core we just work around this on the python side.

emjin:
Closes #3861

Catch errors in Match_rules.ml and report the rule_id of the rule being run if not already present in the exception.

PR checklist:
- [ ] changelog is up to date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -362,7 +362,7 @@ let exn_to_error file exn =
   match exn with
   | AST_generic.Error (s, tok) ->
       let loc = PI.unsafe_token_location_of_info tok in
-      E.mk_error_no_rule loc s AstBuilderError
+      E.mk_error loc s AstBuilderError
   | _ -> E.exn_to_error file exn
 
 (*****************************************************************************)
@@ -499,7 +499,7 @@ let filter_files_with_too_many_matches_and_transform_as_timeout matches =
            (* todo: we should maybe use a new error: TooManyMatches of int * string*)
            let loc = Parse_info.first_loc_of_file file in
            let error =
-             E.mk_error id loc
+             E.mk_error ~rule_id:(Some id) loc
                (spf
                   "%d rules result in too many matches, most offending rule \
                    has %d: %s"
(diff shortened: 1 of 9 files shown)


semgrep/tests/e2e/snapshots/test_check/test_timeout/results.json
```diff
@@ -3,7 +3,7 @@
     {
       "code": 2,
       "level": "warn",
-      "message": "Semgrep Core WARN - Timeout: When running rules.forcetimeout on targets/equivalence/open_redirect.py: ",
+      "message": "Semgrep Core WARN - Timeout: When running rules.forcetimeout on targets/equivalence/open_redirect.py: Main.timeout_function:1",
```
brendongo, line 6: Hmm I don't think this is right?

emjin: It's the actual message that's associated with the exception in `semgrep-core`---I can override it if you think it's too confusing for users

semgrep-core/src/cli/Main.ml
```diff
@@ -664,22 +664,18 @@ let iter_files_and_get_matches_and_exn_to_errors f files =
...
                | (Main_timeout _ | Out_of_memory) as exn ->
-                   let error_fun =
-                     match !Match_patterns.last_matched_rule with
-                     | None -> E.mk_error_no_rule
-                     | Some rule ->
-                         logger#info "critical exn while matching ruleid %s"
-                           rule.MR.id;
-                         logger#info "full pattern is: %s"
-                           rule.MR.pattern_string;
-                         E.mk_error rule.MR.id
```
aryx, line 675: why we don't return an mk_error ~rule_id in that case then? Seems like we have a current rule, so we could be more precise?

aryx (approved):

semgrep/tests/e2e/snapshots/test_check/test_timeout/results.json
brendongo: I think we can keep it if it's useful for us to debug but I imagine it's always going to timeout on this main function? If so overriding would be great :D

semgrep-core/src/cli/Main.ml
emjin: This is the mini-rule id, so it's not very useful to a user. Also, if semgrep-core ever times out while running a rule, it should be caught in `Run_rules.check`. If the exception reaches Main, there shouldn't be a rule_id

PR #3894, opened 2021-09-16
[Ruby] Fix false "..." matches

### LINKED ISSUE #3316: [ruby] Constant propagation ignores side-effects on string object
**Describe the bug**
Constant propagation incorrectly flags a variable as constant after it has been modified by a method call.

**To Reproduce**
https://semgrep.dev/s/4y19/

**Expected behavior**
No false positives.

**Environment**

IagoAbal:
- AST_to_IL: Recognize "concat" method and make translation language-dependent.
- Dataflow_constness: Assume that void method calls may update the callee
    
Closes #3316

test plan:
make test # tests included

PR checklist:
- [x] Changelog is up-to-date


IagoAbal:
> ⚠️ Potential non-blocking slowdown in benchmark semgrep.bench.0c34.std: +21.0% (+0.365 s)
> 
> 15 benchmarks, 7.1% slower on average.

I don't see how this PR makes anything 20% slower, this is just noise IMO.

IagoAbal:
Indeed, I re-run the jobs and this time it passed.  🎉

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -192,6 +193,25 @@ let mk_unit tok eorig =
 let add_instr env instr = Common.push (mk_s (Instr instr)) env.stmts
 
 (*e: function [[AST_to_IL.add_instr]] *)
+
+let var_of_exp env tok exp =
```
emjin: Wonder if there's a better name for this, or maybe you could name the return type and annotate the function with it? It's a little confusing to me what the purpose of `(var, lval)` is going to be and thus what this function is.

Of course I am saying that based on a skim of your PR; I'm sure in context it would be clear

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -352,16 +372,37 @@ and assign env lhs _tok rhs_exp eorig =
...
  *)
-and expr_aux env eorig =
+and expr_aux env ?(void = false) eorig =
   match eorig.G.e with
-  | G.Call ({ e = G.IdSpecial (G.Op op, tok); _ }, args) ->
+  | G.Call ({ e = G.IdSpecial (G.Op op, tok); _ }, args) -> (
       let args = arguments env args in
-      mk_e (Operator ((op, tok), args)) eorig
+      if not void then mk_e (Operator ((op, tok), args)) eorig
+      else
```
emjin, line 382: Are there any operators of this kind which return something and also mutate state? (This seems like a useful proxy, I'm more just curious)

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -386,6 +427,33 @@ and expr_aux env eorig =
           add_instr env (mk_i (Assign (lval, opexp)) eorig);
           lvalexp
       | _ -> impossible (G.E eorig))
+  | G.Call
```
emjin, line 432: Curious why you took a different approach here than in the previous. It seems like it could also be reasonable to assume `$X.$Y(...);` mutates `$X` (which I think you do later in dataflow?)

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -352,16 +372,37 @@ and assign env lhs _tok rhs_exp eorig =
...
+        (* The operation's result is not being used, so it may have side-effects.
+         * We then assume this is just syntax sugar for a method call. E.g. in
+         * Ruby `s << "hello"` is syntax sugar for `s.<<("hello")` and it mutates
+         * the string `s` appending "hello" to it. *)
+        match args with
+        | [] -> impossible (G.E eorig)
+        | obj :: args' ->
+            let obj_var, _obj_lval = var_of_exp env tok obj in
+            let method_id = (Parse_info.str_of_info tok, tok) in
+            let method_name = (method_id, -1) in
```
emjin, line 392: Not an immediate concern, but this would be nice as a record. `-1` doesn't mean anything to me

emjin (approved):

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: Do `tmp_var_of_exp` or `mk_tmp_var` sound better?

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: Probably yes. I hope stdlibs do not do that, but users can totally define such kind of operators.

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: I agree. As one can infer from some of my PR comments, I am not fond of tuple types myself, I tend to prefer record types. Maybe a good small task for the next anti-hack week? That second field is the `sid`, `-1` just says that the name has not been resolved.

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: Very good observation/question! In the case of arbitrary void-method calls we just assume that "anything can happen", so we the callee may end up with some arbitrary non-constant value. But `concat` is something we know about, so when we find `a.concat("hello")` we want to know that if `a` was constant it continues to be. Maybe I should add a test for this case too.

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: I renamed it to `mk_aux_var` plus added a comment.

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: Added a test for such case.

PR #3900, opened 2021-09-17
Stop analyzing dead code in Taint

mmcqd:
This PR prevents taint mode from analyzing dead code

Test plan: sc -config tests/tainting_rules/go/goto_dead_code.yaml tests/tainting_rules/go/goto_dead_code.go -lang go

Without this PR the dead code in this test is erroneously matched.

PR checklist:
- [x] Changelog is up-to-date


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Changed
 - Grouped semgrep CLI options and added constraints when useful (e.g. cannot use `--vim` and `--emacs` at the same time)
+- CFGs now track their entry point, and dataflow uses this to build a workset by DFS
```
IagoAbal: In the changelog I would personally only describe what effect this has from the user perspective, which is that unreachable statements are no longer visited by the Dataflow engine.

semgrep-core/src/analyzing/Dataflow.ml
```diff
@@ -46,7 +46,10 @@ module type Flow = sig
 
   type edge
 
-  type flow = (node, edge) Ograph_extended.ograph_mutable
+  type flow = {
+    graph : (node, edge) Ograph_extended.ograph_mutable;
+    entry : int;
```
IagoAbal: Use the `nodei` type alias here rather than `int`.

semgrep-core/src/analyzing/Dataflow.mli
```diff
@@ -113,7 +113,10 @@ module type Flow = sig
 
   type edge
 
-  type flow = (node, edge) Ograph_extended.ograph_mutable
+  type flow = {
+    graph : (node, edge) Ograph_extended.ograph_mutable;
+    entry : int;
```
IagoAbal: I would use `nodei` here too.

semgrep-core/src/analyzing/Dataflow.ml
IagoAbal: Hmm, maybe better to define a type previously in `Dataflow` (outside of `module type Flow`):

```
type ('node, 'edge) cfg = {
     graph : ('node, 'edge) Ograph_extended.ograph_mutable;
     entry : nodei;
}
```
And then refer to this everywhere else, so here it would be:
```
type flow = (node, edge) cfg;
```

IagoAbal:
You should add a "test plan" for this, and a test. You can just take your taint example with `continue` and replace it with a `goto`.

semgrep-core/src/analyzing/CFG_build.ml
```diff
@@ -255,14 +255,21 @@ let (cfg_of_stmts : stmt list -> F.cfg) =
 
   let newi = enteri in
 
-  let state = { g; exiti; labels = Hashtbl.create 2; gotos = ref [] } in
+  let state =
+    {
+      g = { graph = g; entry = enteri };
```
IagoAbal: Do we use the `entry` field anywhere in the CFG construction?

semgrep-core/src/analyzing/Dataflow.ml
```diff
@@ -264,10 +267,24 @@ module Make (F : Flow) = struct
...
+      (fun s (ni, _) -> NodeiSet.add ni s)
+      NodeiSet.empty
+
+  let mk_worklist ({ graph; entry } : F.flow) =
+    let rec aux nodei seen =
+      if NodeiSet.mem nodei seen then seen
+      else
+        let seen = NodeiSet.add nodei seen in
+        let succs = forward_succs { graph; entry } nodei in
+        NodeiSet.fold aux succs seen
```
IagoAbal, line 285: Correct me if I'm wrong, but here you may be visiting the same nodes several times. The `seen` set only knows about the nodes visited in one specific branch of the traversal.

semgrep-core/src/analyzing/Dataflow_tainting.ml
```diff
@@ -249,16 +255,23 @@ let (transfer :
 (* Entry point *)
 (*****************************************************************************)
 
+(* Very silly function. We previously relied on the fact that F.cfg and X.flow were equal types,
+   but because OCaml does not have first class records, this equivalence no longer holds definitionally.
+   Instead we have to manually convert between the types even though they are technically the same
+*)
+let cfg_to_flow ({ graph; entry } : F.cfg) : X.flow = { graph; entry }
```
IagoAbal: As I said in a previous comment, I think we should have a `('node, 'edge) cfg` type in `Dataflow` and everywhere else we just instantiate that type.

semgrep-core/src/analyzing/controlflow_build.ml
```diff
@@ -692,7 +698,7 @@ let (control_flow_graph_of_stmts : parameter list -> stmt list -> F.flow) =
 
   let state =
     {
-      g;
+      g = { graph = g; entry = enteri };
```
IagoAbal: Again, this causes a cascade of changes, and I am not sure it's justified unless we need the `entry` somewhere in the construction of the graph.

IagoAbal (changes requested):
Overall very nice job! But please take a look a the comments and let me know if you disagree!

aryx:
I don't fully understand the change. You can get the entry point of the graph by looking for the Start node. There is even a function for that. What is the point of this change?

IagoAbal:
> I don't fully understand the change. You can get the entry point of the graph by looking for the Start node. There is even a function for that. What is the point of this change?

I agree the point of the PR is not clear, it should have a different message and a different changelog entry, IMO. The point is that currently the dataflow engine visits all nodes in the CFG irrespectively of whether they are reachable or unreachable. We wanted to skip unreachable nodes.

The need for an `entry` field is due to `Dataflow` working on an abstract `node` type, so there is no notion of `Enter` node.

aryx:
Ah, I see. Then your comment should be a comment in the code next to this cfg definition.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Changed
 - Grouped semgrep CLI options and added constraints when useful (e.g. cannot use `--vim` and `--emacs` at the same time)
+- Taint no longer analyzes dead/unreachable code
```
IagoAbal: Need to fix the changelog, this does to belong to 0.65.0.

semgrep-core/src/analyzing/Dataflow.ml
```diff
@@ -226,10 +227,24 @@ module Make (F : Flow) = struct
...
+      (fun s (ni, _) -> NodeiSet.add ni s)
+      NodeiSet.empty
 
   let backward_succs (f : F.flow) n =
-    (f#predecessors n)#fold (fun s (ni, _) -> NodeiSet.add ni s) NodeiSet.empty
+    (f.graph#predecessors n)#fold
+      (fun s (ni, _) -> NodeiSet.add ni s)
+      NodeiSet.empty
+
+  let mk_worklist ({ graph; entry } : F.flow) =
```
IagoAbal, line 240: Could we add here a comment to explain that we are computing the set of reachable nodes in the CFG?

semgrep-core/src/core/il/CFG.ml
```diff
@@ -12,3 +12,16 @@
...
  *)
+
+type nodei = int
+
+(* We include the entry point because the Dataflow engine uses this, and works using an abstract node type,
+   which prevent it from searching for an entry node.
+*)
+type ('node, 'edge) t = {
+  graph : ('node, 'edge) Ograph_extended.ograph_mutable;
+  entry : nodei;
```
IagoAbal, line 23: Minor thing, you could have referred to `Ograph_extended.nodei`.

IagoAbal (approved):
Great!

IagoAbal:
> ⚠️ Potential non-blocking slowdown in benchmark semgrep.bench.lodash.std: +24.7% (+1.128 s)

Noise...

PR #3912, opened 2021-09-20
Add message for out of memory exceptions

emjin:
Distinguish between out of memory errors triggered by the garbage collector and triggered by a user's memory limit. The message will show as either

"Out of memory: Heap space exceeded" or "Out of memory: Exceeded memory limit"

If useful, we can also add specific numbers in the latter messaged ($USED mb > $LIMIT mb). Leaving this for now because doing that requires changing the testing script slightly and unclear if there is demand.

Incorporates changes from https://github.com/returntocorp/pfff/pull/474

Test plan: make test (see new message)

PR checklist:
- [ ] Changelog is up-to-date


semgrep-core/src/core/Semgrep_error_code.ml
```diff
@@ -99,7 +99,10 @@ let exn_to_error ?(rule_id = None) file exn =
       mk_error ~rule_id loc msg Timeout
   | Out_of_memory ->
       let loc = Parse_info.first_loc_of_file file in
-      mk_error ~rule_id loc "Out of memory" OutOfMemory
+      mk_error ~rule_id loc "Heap space exceeded" OutOfMemory
```
IagoAbal, line 102: Just curious about the motivation of this change, isn't "out of memory" a rather standard way of reporting this?

emjin: The error message will now say "Out of memory: heap space exceeded" (vs if you exceed your own memory limit it will say "Out of memory: memory limit reached").

mjambon (approved):

PR #3924, opened 2021-09-22
Add attribute-expression equivalence

### LINKED ISSUE #3489: Expressions forming python decorators aren't matched ('@foo', '@foo()', '@foo.bar()')
Using metavariables in Python `from ... import ...` style imports misses decorator usage of the metavariable.

See here: https://semgrep.dev/s/jk4y

image

IagoAbal:
A named attribute is essentially a function call, and it is useful to
be able to match it as such.

Plus, let a DotAccess of idents match a name.
 
Closes #3489
    
test plan:
make test # tests included

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/core/ast/AST_generic_helpers.ml
+++ b/semgrep-core/src/core/ast/AST_generic_helpers.ml
@@ -65,6 +65,18 @@ let name_of_ids ?(name_typeargs = None) xs =
 
 let name_of_id id = Id (id, empty_id_info ())
 
+let name_of_dot_access e =
+  let ( let* ) = Option.bind in
+  let rec fetch_ids = function
+    | G.N (G.Id (x, _)) -> Some [ x ]
+    | G.DotAccess (e1, _, G.EN (G.Id (x, _))) ->
+        let* xs = fetch_ids e1.e in
+        Some (xs @ [ x ])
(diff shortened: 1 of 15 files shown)


semgrep-core/src/core/ast/AST_generic_helpers.ml
```diff
@@ -65,6 +65,18 @@ let name_of_ids ?(name_typeargs = None) xs =
 
 let name_of_id id = Id (id, empty_id_info ())
 
+let name_of_dot_access e =
```
aryx, line 68: Maybe update the comment in AST_generic.ml about DotAccess and reference name_of_dot_access as temporary hack to solve the issue.

aryx (approved):

emjin:
Documentation will definitely need to be updated

semgrep-core/src/core/ast/AST_generic_helpers.ml
IagoAbal: Done! Rewriting Naming_AST for this and other stuff could be a future tech-debt project. GHC e.g. also uses the "renaming" phase to "fix" things in the AST that cannot be decided during parsing (such as our `DotAccess` vs `IdQualified`).

PR #3925, opened 2021-09-22
Add support for Break and Continue in IL

mmcqd:
Break and Continue can now be translated into the IL and analyzed by taint mode.

Test Plan: make test

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -39,9 +39,17 @@ type env = {
    * usually simple Instr, but can be also If when handling Conditional expr.
    *)
   stmts : stmt list ref;
+  (* When entering a loop, we create a two labels, one to jump to if a Continue stmt is found
```
aryx: typo "a two"

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -39,9 +39,17 @@ type env = {
...
+     and another to jump to if a Goto stmt is found. Since PHP supports breaking an arbitrary
```
aryx: You mean Break here, not Goto?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -711,6 +719,25 @@ let parameters _env params =
...
+  let cont_label = fresh_label env tok in
+  let break_label = fresh_label env tok in
+  let st_env =
+    {
+      env with
+      break_labels = break_label :: env.break_labels;
+      cont_label = Some cont_label;
+    }
+  in
+  (* We add a dummy fixme stmt after each label in case the label ends up
```
aryx: maybe we could fix the CFG builder instead?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -733,16 +760,21 @@ let rec stmt_aux env st =
       ss @ [ mk_s (If (tok, e', st1, st2)) ]
   | G.Switch (_, _, _) -> todo (G.S st)
   | G.While (tok, e, st) ->
+      let cont_label_s, break_label_s, st_env = mk_st_env_and_labels env tok in
```
aryx: if the original 'env' is not used after, I tend to do let env = stuff_that_return_a_new_env env in.
This avoids some typo where you forgot to use the right env.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -39,9 +39,17 @@ type env = {
    * usually simple Instr, but can be also If when handling Conditional expr.
    *)
   stmts : stmt list ref;
+  (* When entering a loop, we create a two labels, one to jump to if a Continue stmt is found
+     and another to jump to if a Goto stmt is found. Since PHP supports breaking an arbitrary
+     number of loops up, we keep a stack of break labels instead of just one
+  *)
+  (* TODO: Bad to mix mutable + unmutable fields? *)
```
aryx: I use mutable in environment only when info computed recursively needs to be used in other branches.
Here I don't think it's the case.

aryx (approved):
Looks good. It's a bit annoying that we create all those labels and todo_fixstmt
for loops when most of the time there will be actually no break or continue in them,
and all those labels introduce noise in the output.

semgrep-core/src/analyzing/AST_to_IL.ml
mmcqd: The original 'env' is used after, and because the translation has side effects (modifying the ref in the env), there are cases where I can't just switch around the order of calls.

semgrep-core/src/analyzing/AST_to_IL.ml
mmcqd: This would be a non-trivial change. The CFG builder currently associates labels to nodes, and doesn't include labels as their own nodes, which is why this happens. @IagoAbal do you think this is worth doing?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## Unreleased
 
 ### Added
+- Added break continue support for taint mode
```
IagoAbal: Maybe better something like:

> Added support for `break` and `continue` in the dataflow engine

The "break continue" phrase may be a bit confusing, and this change also affects constant propagation.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -711,6 +718,25 @@ let parameters _env params =
 (*****************************************************************************)
 (* Statement *)
 (*****************************************************************************)
+
+let mk_st_env_and_labels env tok =
```
IagoAbal: Maybe `mk_break_continue_labels`.

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: [CIL](https://github.com/cil-project/cil) for example annotates statements with labels and it uses [dummy statements](https://github.com/cil-project/cil/blob/936b04103eb573f320c6badf280e8bb17f6e7b26/src/frontc/cabs2cil.ml#L1001) for `goto`s too. Although e.g. for `continue` I think they only add these labels on-demand.

Another option is to change the `IL.node` type and add a `NLabel` node, then change the CFG builder to use these nodes instead.

IagoAbal: I am not sure about pros/cons here, but given that in the IL we have a `Label` statement perhaps we should interpret it as `label: noop`. Then we can try to optimize this in the CFG builder, but create a dummy node by default for each label (in the CFG builder).

I suggest we add a `FIXME` note and create an issue to fix this (in some way) ASAP, but use a separate PR.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -793,17 +827,44 @@ let rec stmt_aux env st =
...
   (* TODO: repeat env work of controlflow_build.ml *)
-  | G.Continue _
-  | G.Break _ ->
-      todo (G.S st)
+  | G.Continue (tok, lbl_ident, _) -> (
+      match lbl_ident with
+      | G.LNone -> (
+          match env.cont_label with
+          | None ->
+              todo (G.S st) (* Continue outside of loop? Should never happen *)
```
IagoAbal: We have `impossible` for this.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -793,17 +827,44 @@ let rec stmt_aux env st =
...
+          | Some lbl -> [ mk_s (Goto (tok, lbl)) ])
+      | G.LId lbl -> [ mk_s (Goto (tok, label_of_label env lbl)) ]
+      | G.LInt _
+      | G.LDynamic _ ->
+          todo (G.S st))
+  | G.Break (tok, lbl_ident, _) -> (
+      match lbl_ident with
+      | G.LNone -> (
+          match env.break_labels with
+          | [] -> todo (G.S st) (* Break outside of loop? Should never happen *)
```
IagoAbal: `impossible`

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -793,17 +827,44 @@ let rec stmt_aux env st =
...
+          | lbl :: _ -> [ mk_s (Goto (tok, lbl)) ])
+      | G.LId lbl -> [ mk_s (Goto (tok, label_of_label env lbl)) ]
+      | G.LInt (i, _) -> (
+          match List.nth_opt env.break_labels i with
+          | None ->
+              todo (G.S st)
```
IagoAbal: `impossible`

IagoAbal:
> Looks good. It's a bit annoying that we create all those labels and todo_fixstmt
> for loops when most of the time there will be actually no break or continue in them,
> and all those labels introduce noise in the output.

Yeah I agree it would be nice to introduce the labels on demand, but it could make the translation more complex too. The good thing is that (at least right now) the labels disappear in the CFG.

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: Hmm, I looked into this more carefully and it may be easier to fix this in `CFG_build.cfg_stmt_list`. Basically, if we are done folding, and we have a bunch of unassigned labels, we just need to create a dummy node, connect `previ` (if any) to this node, and assign the labels to it.

For the dummy node we could use `NTodo` or `NOther` if we extend `other_stmt` with e.g. `Noop`.

mmcqd: Ok yeah this actually was pretty easy. I added `Noop` to `other_stmt`. All the tests still pass with no complaints of unresolved gotos.

IagoAbal (approved):
Nice!

PR #3931, opened 2021-09-24
[Kotlin] finish all todos in Parse_kotlin_tree_sitter.ml

aryx:
Some constructs are still not represented correctly, and we need to
fix that at some point, but at least we will not throw some
exceptions.

This closes PA-401

test plan:
make

PR checklist:
- [x] Changelog is up-to-date


aryx:
@ievans semgrepdep took more than 30min (I cancelled the job). Not sure what's going on.

PR #3932, opened 2021-09-24
[AST_generic] refactor type_parameter for Kotlin/Java/Scala/C++/Rust

aryx:
This will help PA-406
Lots of information was lost previously.

test plan:
make

PR checklist:
- [x] Changelog is up-to-date


emjin:
Is it possible to add one or two test cases? At least, I would appreciate an example in the PR---I don't really understand what each field is for. (Trying to code review more!)

PR #3944, opened 2021-09-27
Add `--config auto`

nbrahms:
Currently, it is difficult to know which rules to run on a codebase. This manifests in three different distinct moments of Semgrep use:

* On first use, a project maintainer would like to configure rules that are germane to the project. For instance, a Python Django / Angular project supporting a production web-app should probably run language-specific correctness, general-security, and injection rules for Python, JavaScript, and Jinja, and framework-specific rules for Django and Angular.
* As the project matures, new languages or frameworks may be introduced. For example, React may be added, or a set of backend services may be ported to Golang for scalability. This should cause the addition of new rules. Similarly, removing or replacing a framework should cause the removal of Semgrep rules that no longer apply (but presumably still consume job run-time).
* As new Semgrep rules are introduced to the Semgrep registry, new project-relevant rules should be added to the configuration.

In order to ease both set-up and maintenance of Semgrep on a project, we introduce --config auto, which causes Semgrep to coordinate with the Semgrep Registry to customize rules for a project. This applies both to setup and maintenance phases of Semgrep use.

Initially, Semgrep will use --config auto mainly to support the third of the above cases. Semgrep will report the project URL to the Semgrep registry in its server request for project-relevant rules. In a future release of Semgrep, it will also use Semgrep’s inventory-audit rules to determine which rules are relevant to a project. I’ve updated documentation throughout (including PRIVACY.md to make this clear to the user).

PR checklist:
- [x] Changelog is up-to-date


nbrahms:
Semgrep output:
```
Auto config mode will use semgrep rules to scan your codebase and uses the Semgrep Registry to generate recommended
rules based on your languages and frameworks.

Logging in to the Semgrep Registry as project 'git@github.com:returntocorp/semgrep.git'...

Using config from https://semgrep.dev/auto. Visit https://semgrep.dev/registry to see all public rules.
```

CHANGELOG.md
```diff
@@ -40,6 +40,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Added
 - HCL (a.k.a Terraform) experimental support
+- Added "automatic configuration" (`--config auto`), which collaborates with
```
brendongo: This needs to be moved to unreleased

semgrep/semgrep/util.py
```diff
@@ -111,6 +114,22 @@ def with_color(color: str, text: str, bold: bool = False) -> str:
     return click.style(text, fg=color)
 
 
+def terminal_wrap(text: str) -> str:
+    from shutil import get_terminal_size
+    import textwrap
```
brendongo, line 119: Do these imports take longer? should we just do them at top level?

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/bad1/error-in-color.txt
```diff
@@ -1,4 +1,4 @@
-[31msemgrep error[0m: Invalid rule schema
+[31m[31msemgrep error[0m: Invalid rule schema
```
brendongo, line 1: Should we have `semgrep error` be colored something that isn't red since the entire error is red now?

semgrep/semgrep/config_resolver.py
```diff
@@ -204,6 +210,118 @@ def _validate(  # type: ignore[misc]
...
+                    "\n"
+                ),
+            )
+            if self.project_url is not None:
+                extra_headers["X-Semgrep-Project"] = self.project_url
+                logger.warning(
+                    with_color(
+                        "green",
+                        terminal_wrap(
+                            f"Logging in to the Semgrep Registry as project '{self.project_url}'...\n"
```
brendongo: Can we just use `Fetching rules from semgrep registry for project : {self.project_url}` instead of logging in language

brendongo (approved):

semgrep/semgrep/config_resolver.py
nbrahms: cc. @ievans . This language is to communicate to the user that we're communicating to the server with a cleartext identity.

semgrep/semgrep/util.py
nbrahms: util is imported on CLI load, so we should minimize its top-level imports

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/bad1/error-in-color.txt
nbrahms: It's actually not, because of the reset characters within the error body.

ievans:
@nbrahms can https://github.com/returntocorp/semgrep/blob/38cb77d7371e918354a7319beea8905c220a7c00/semgrep/semgrep/semgrep_main.py#L163 also be updated to suggest config auto rather than p/ci or p/security?

ievans (approved):

emjin:
> @nbrahms can
> 
> https://github.com/returntocorp/semgrep/blob/38cb77d7371e918354a7319beea8905c220a7c00/semgrep/semgrep/semgrep_main.py#L163
> 
> also be updated to suggest config auto rather than p/ci or p/security?

Perhaps we could just use auto by default if nothing is specified?

ievans:
> Perhaps we could just use auto by default if nothing is specified?

+1 to that, though I was thinking we'd make sure it's working as intended for a bit before doing so

semgrep/semgrep/config_resolver.py
```diff
@@ -204,6 +210,118 @@ def _validate(  # type: ignore[misc]
...
+            )
+
+    def resolve_config(self, config_str: str) -> Dict[str, YamlTree]:
+        """ resolves if config arg is a registry entry, a url, or a file, folder, or loads from defaults if None"""
+        start_t = time.time()
+        extra_headers: Dict[str, str] = {}
+        if config_str == AUTO_CONFIG_KEY:
+            logger.warning(
+                terminal_wrap(
+                    "Auto config mode will use semgrep rules to scan your codebase and uses the Semgrep Registry"
```
ievans:
```suggestion
                    "Auto config uses Semgrep rules to scan your codebase and the Semgrep Registry"
```

semgrep/semgrep/config_resolver.py
```diff
@@ -204,6 +210,118 @@ def _validate(  # type: ignore[misc]
...
+                    " to generate recommended rules based on your languages and frameworks."
+                    "\n"
+                ),
+            )
+            if self.project_url is not None:
+                extra_headers["X-Semgrep-Project"] = self.project_url
+                logger.warning(
+                    with_color(
+                        "green",
```
ievans: why green?

ievans (dismissed):
Can we eliminate the newlines before and after "Logging in to..." and remove the green? cc @adamboros 
Screen Shot 2021-10-08 at 3 35 02 PM

semgrep/semgrep/config_resolver.py
brendongo: Login includes identification and authentication and I just get confused with what authentication i setup on semgrep.

brendongo: I think the suggestion I made above makes it much more clear that a cleartext identity is being communicated to the server

adamboros: Maybe it's the language barrier, but this suggestion does not make it clear for me. If it would say _'from semgrep registry **with** project: {self.project_url}',_ that would indicate that the project is being communicated (for me at least). 

Accessing (the Semgrep Registry as project) might be a better word here.

semgrep/semgrep/config_resolver.py
```diff
@@ -66,42 +70,73 @@
 }
 
 
+class ConfigType(Enum):
+    REGISTRY = 1
+    CDN = 2
+    LOCAL = 3
```
nbrahms:
```suggestion
    REGISTRY = auto()
    CDN = auto()
    LOCAL = auto()
```

semgrep/semgrep/config_resolver.py
```diff
@@ -66,42 +70,73 @@
...
-    def __init__(self, config_str: str) -> None:
+    def __init__(self, config_str: str, project_url: Optional[str]) -> None:
         """
         Mutates MetricManager state!
         Takes a user's inputted config_str and transforms it into the appropriate
         path, checking whether the config string is a registry url or not. If it
         is, also set the appropriate MetricManager flag
         """
+        self._project_url = project_url
+
```
nbrahms:
```suggestion
        self._project_url = project_url
        self._origin = ConfigType.REGISTRY
```

then only override the two places we differ

semgrep/semgrep/config_resolver.py
```diff
@@ -315,11 +337,10 @@ def _rename_rule_ids(valid_configs: Mapping[str, Sequence[Rule]]) -> None:
                     f"{Config._convert_config_id_to_prefix(config_id)}{rule.id or MISSING_RULE_ID}"
                 )
 
-    # the mypy ignore is cause YamlTree puts an Any inside the @staticmethod decorator
     @staticmethod
-    def _validate(  # type: ignore[misc]
+    def _validate(
         config_dict: Mapping[str, YamlTree]
-    ) -> Tuple[Mapping[str, Sequence[Rule]], Sequence[SemgrepError]]:
+    ) -> Tuple[Dict[str, List[Rule]], List[SemgrepError]]:
```
nbrahms: We should revert this change (prefer immutable returns)

semgrep/semgrep/config_resolver.py
```diff
@@ -227,8 +253,8 @@ def from_pattern_lang(
 
     @classmethod
     def from_config_list(
-        cls, configs: Sequence[str]
-    ) -> Tuple["Config", Sequence[SemgrepError]]:
+        cls, configs: Sequence[str], project_url: Optional[str]
+    ) -> Tuple["Config", List[SemgrepError]]:
```
nbrahms: Revert me

emjin (approved):

PR #3945, opened 2021-09-27
Fix CFG builder creating dummy nodes that do not connect to exit node

mmcqd:
If a dummy node created for label with no accompanying statement is the final node in a CFG, it was previously not connected to the exit node. Now it is.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/analyzing/CFG_build.ml
+++ b/semgrep-core/src/analyzing/CFG_build.ml
@@ -238,7 +238,7 @@ and cfg_stmt_list state previ xs =
       let dummyi = state.g#add_node { n = NOther Noop } in
       label_node state (l :: ls) dummyi;
       add_arc (lasti, dummyi) state.g;
-      lasti_opt
+      Some dummyi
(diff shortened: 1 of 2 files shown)


emjin (approved):

semgrep-core/src/analyzing/CFG_build.ml
```diff
@@ -238,7 +238,7 @@ and cfg_stmt_list state previ xs =
       let dummyi = state.g#add_node { n = NOther Noop } in
       label_node state (l :: ls) dummyi;
       add_arc (lasti, dummyi) state.g;
-      lasti_opt
+      Some dummyi
```
IagoAbal, line 241: Ooops

aryx:
I like to have a 'test plan:' section in the PR where you describe a procedure to see that the change works.
In that case it could be a test file and an example of a command like semgrep-core -cfg_generic ... where you say the node is now connected.

PR #3947, opened 2021-09-28
feat(sarif): set name from rule metadata

eliblock:
According to `v2.1.0` of the SARIF spec ([link](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317843)),
the `name` property of a `reportingDescriptor` object may be a:

> localizable string containing an identifier that is understandable to an end user

Update the `name` property to be configurable based on each rule's
`metadata.name` field.

Additionally, for backwards compatibility, properly set the `deprecatedNames`
property when a metadata-sourced name is set.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Attribute-expression equivalence that allows matching expression patterns against
   attributes, it is enabled by default but can be disabled via rule `options:` with
   `attr_expr: false` (#3489)
+- The `name` field of rules in the SARIF output is now set by a rule's `name`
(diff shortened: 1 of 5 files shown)


brendongo:
Hi @eliblock we are planning on adding a name field in the top level of a rule definition in the next month (easy fix on this repo but we have to clean other internal dependencies up elsewhere too). I would rather wait til that work is done than temporarily support an optional field in meta that would do the same thing (but just for sarif). Does that timeline work for you?

eliblock:
@brendongo thank you for the heads up - that timeline works for me!

Would it be possible to add updating the SARIF format to properly surface rule `name`s (once they're defined) to your plan for that feature, if it's not there already?

PR #3948, opened 2021-09-28
optimizing: Analyze_rule: Fix OR simplifications

### LINKED ISSUE #3755: Pattern-either works wrong with metavariable-regex.
**Describe the bug**
1. When you have metavariable-regex in one of pattern-either cases/branches - it is unable to detect this case if there is no second case present in the file, even only in the comments. 

2. How is it possible that **comments** affect rule execution?

**To Reproduce**
First:
https://semgrep.dev/s/YXwD - works as desired at a first glance.
Let`s comment second either pattern
https://semgrep.dev/s/65wW - working fine as well. Fine, we have just commented coule lines
Now **REMOVE** lines for second either patten:
https://semgrep.dev/s/owox - Wrong! First pattern does not trigger anymore.

**Expected behavior**
Both cases/patterns from either expressions picked up by SemGrep. 

**Screenshots**
Looks right, but comments affecting the result:
image

Wrong:
image

**Environment**
observable > v0.61.0

IagoAbal:
When a term in an OR cannot be checked with a regex, we have to assume
that the term holds and thus reduce the entire OR to "true".

Plus, enable -filter_irrelevant_rules for full-rule tests.

Plus, prevent Analyze_rule from crashing when given a "new" formula
where things like `Not (Or ...)` are possible, otherwise e.g. test
tests/OTHER/rules/negation_ajin fails (by `failwith "Not Or"`).

Closes #3755

test plan:
make test # tests included

TL;DR

For example:

    - pattern-either:
        - patterns:
            - pattern: $X.sha1("$RE")
            - metavariable-regex:
                metavariable: $RE
                regex: (?i)md2
        - patterns:
            - pattern: $X.getMd2Digest(...)

This is seen by Analyze_rule as:

    OR
        AND (contains "sha1") (dont-know-how-to-check)
        AND (contains "getMd2Digest")

Sometimes Analyze_rule does not know how to extract a regex from a
pattern, here we represent that with `(dont-know-how-to-check)`. The
"problem" with the `metavariable-regex` here is that Analyze_rule does
not know how to handle the case-insensitive mode `(?i)`.

When converting the regex formula into CNF we get:

    AND
        OR (contains "sha1") (contains "getMd2Digest")
        OR (dont-know-how-to-check) (contains "getMd2Digest")

Then in step1 we try to get rid of these `(dont-know-how-to-check)`. The
bug was caused by incorrectly simplifying this as:

    AND
        OR (contains "sha1") (contains "getMd2Digest")
        OR (contains "getMd2Digest")

That is, treating `(dont-know-how-to-check)` as "false" rather than
"true". The correct simplification is:

    AND
        OR (contains "sha1") (contains "getMd2Digest")

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 - Fix CFG dummy nodes to always connect to exit node
-
+- Fixed bug in semgrep-core's `-filter_irrelevant_rules` causing Semgrep to
+  incorrectly skip a file (#3755)
 
 ## [0.66.0](https://github.com/returntocorp/semgrep/releases/tag/v0.66.0) - 09-22-2021
 
--- a/semgrep-core/src/optimizing/Analyze_rule.ml
+++ b/semgrep-core/src/optimizing/Analyze_rule.ml
@@ -90,6 +90,27 @@ exception EmptyAnd
 exception EmptyOr
 
 (*****************************************************************************)
+(* Utils *)
+(*****************************************************************************)
+
+let ( let* ) = Common.( >>= )
+
+(* NOTE "AND vs OR and map_filter":
+ * We cannot use `Common.map_filter` for `R.Or`, because it has the wrong
+ * semantics. We use `None` to say "we can't handle this", or in other words,
+ * "we assume this pattern can match", or just "true"! So in an AND we can
+ * remove those "true" terms, but in an OR we need to reduce the entire OR to
+ * "true". Therefore, `Common.map_filter` works for AND-semantics, but for
+ * OR-semantics we need `option_map`. *)
+let option_map f xs =
+  List.fold_left
+    (fun acc x ->
+      let* ys = acc in
+      let* y = f x in
+      Some (y :: ys))
+    (Some []) xs
+
+(*****************************************************************************)
 (* Step0: a complex formula to a CNF *)
 (*****************************************************************************)
 (* Transforming a complex formula to a simple CNF formula.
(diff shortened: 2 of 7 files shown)


aryx:
what is this 'cannot-check'?

aryx:
Also the test plan say 'test included' but I don't see any tests in the PR.

aryx (changes requested):
Can you add some tests with the example you put in the PR?

IagoAbal:
> Also the test plan say 'test included' but I don't see any tests in the PR.

Ooops, I forgot to `git add`!

IagoAbal:
> 🚫 Benchmark semgrep.bench.zulip.std is too slow: +52.7% (+4.142 s)
> 
> 🚫 Benchmark semgrep.bench.apache.std is too slow: +150.2% (+20.659 s)
> 
> 🚫 Benchmark semgrep.bench.lodash.std is too slow: +58.9% (+2.980 s)

Will try to check these slow downs, just in case there might be something we can do about them. But if the speed up was caused by incorrectly filtering files we have no other choice... 😞

IagoAbal:
Damn I had a very silly bug.

IagoAbal:
Perhaps we need a small test suite just for -filter_irrelevant_rules.

IagoAbal:
@aryx if everything is green (as I hope) then I believe this is now ready. I fixed a bug that was causing the perf regressions, and I added a couple of tests. For irrelevant rules I had to add a separate test suite (tests/OTHER/irrelevant_rules), where the test fails if either a) Analyze_rule cannot produce a regex prefilter, or b) the regex filter determines that the file is _relevant_ (it is expected to be _irrelevant_). For relevant rules that produce matches we can use tests/OTHER/rules. If we need to test relevant rules that produce no matches then we need to add a new test suite or rethink things a bit.

emjin:
Woah the benchmark tests were useful!!

IagoAbal:
> Woah the benchmark tests were useful!!

They're definitely useful! My main complain is that the warnings fire too often. In the two cases where they flagged a real problem the slow down was 50% or more. Here the apache benchmark was twice as slow, with a +13s delta. But I don't think our setup allows us to reliably detect very subtle regressions that barely affect performance, it's very difficult to distinguish those from noise.

aryx:
great indeed that the benchmark helped finding bugs!

aryx (approved):
This is great! Nice find.

PR #3949, opened 2021-09-28
Allow non-explicit targets to be filtered out by semgrep-core

mjambon:
* Target files discovered in folders by semgrep are now passed to semgrep-core as "filterable targets" via a new option. Semgrep-core will then filter out whatever it thinks should be filtered out, such as for example `.min.js` files.
* Filtering targets by size is now delegated to semgrep-core

Overall these changes result in strictly more target filtering than before.

PR checklist:
- [x] Changelog is up-to-date
- [x] Change has no security implications


semgrep/semgrep/target_manager_extensions.py
```diff
@@ -62,30 +73,64 @@
...
+    Language.PYTHON3: True,
+    Language.PYTHON: True,
+    Language.REGEX: False,  # special
+    Language.RUBY: True,
+    Language.RUST: True,
+    Language.SCALA: True,
+    Language.TYPESCRIPT: True,
+    Language.VUE: False,
+    Language.YAML: False,
 }
```
mjambon, line 134: This is for `lang_is_script`, which is unused for now. I did that earlier, but didn't finish because it's too much for one PR.

semgrep/semgrep/target_manager.py
```diff
@@ -6,27 +6,83 @@
 from pathlib import Path
 from typing import Collection
 from typing import Dict
-from typing import FrozenSet
```
mjambon, line 9: Now everything uses `Set` for simplicity and to avoid unnecessary conversions. I wanted to use `FrozenSet` for everything but there was a set that was being constructed one element at a time and it would have been annoying to change it.

semgrep/semgrep/target_manager.py
```diff
@@ -6,27 +6,83 @@
...
+    elif lang in {"rb", "ruby"}:
+        return RUBY_EXTENSIONS
+    elif lang in {"php"}:
+        return PHP_EXTENSIONS
+    elif lang in {"json", "JSON", "Json"}:
+        return JSON_EXTENSIONS
+    elif lang in {"none", "generic"}:
+        return [FileExtension("*")]
+    else:
+        raise _UnknownLanguageError(f"Unsupported Language: {lang}")
```
mjambon: This is weird. I need to clean it up.

mjambon: At some point I was working with an old version of `target_manager.py`. I think I fixed most of the damage now.

aryx:
any idea about the slowdown? Noise?

semgrep/semgrep/target_manager_extensions.py
```diff
@@ -13,47 +13,58 @@
...
 # pipenv run pytest tests/e2e/test_rule_parser.py --snapshot-update
-PYTHON_EXTENSIONS = [FileExtension(".py"), FileExtension(".pyi")]
-JAVASCRIPT_EXTENSIONS = [FileExtension(".js"), FileExtension(".jsx")]
-TYPESCRIPT_EXTENSIONS = [FileExtension(".ts"), FileExtension(".tsx")]
-JAVA_EXTENSIONS = [FileExtension(".java")]
+#
+# Extensions that are permissible for a given language. Semgrep-core will
+# perform its own filtering to check that the file is indeed in the correct
+# language before attempting full parsing. Additionally, all executable
+# files without an extension will also be passed to semgrep-core.
```
aryx, line 20: What? Is this just for Bash? Should we restrict this to bash then?

aryx (approved):
lgtm

IagoAbal:
Those +10s slowdowns are probably not noise. It's a bit weird, I would expect anything but slowdowns since supposedly we will be analyzing fewer (or worst case, the same) files.

IagoAbal:
We definitely should check what are the before/after lists of files analyzed by semgrep-core on those now slower benchmarks. My wild guess after looking at the PR is that we may be passing files without extension to semgrep-core to analyze that previously were ignored.

semgrep/semgrep/target_manager_extensions.py
IagoAbal: If this wasn't done before, it may be the reason for the slow downs. Perhaps we should not do it by default and require the user to explicitly enable it via a flag?

mjambon: > What? Is this just for Bash? Should we restrict this to bash then?

I'm trying to remove duplicate efforts, with a preference for ocaml over python.

This is for all the languages that can be run as scripts, which is about half of them. What would we be a benefit of doing this only when the requested language is bash?

IagoAbal:
Presumably after commit `4a4490b0d73 ("For now, comment out section that looks for executable scripts")` it seems that we no longer have that perf regression with _coinbase_. It would be nice to know why that slow down... are there so many of such files? Or are they very large? Is analyzing the files that is slow or is the heuristic to check whether we should analyze them?

mjambon:
I figured why the test were failing. I had incorrectly modified a function that was comparing relative paths and absolute paths without realizing it. It turns out we don't need absolute paths, and the test we have don't like absolute paths at all. In order for semgrep to represent targets as absolute paths we'd need to change our testing infrastructure, which may not be worth it. Anyway, I learned a lot and left notes and corrections about the things I discovered along the way.

Also, I commented out the code that would include executable scripts. I'll enable it in another PR when I get back to it.

mjambon:
@IagoAbal
> Presumably after commit 4a4490b0d73 ("For now, comment out section that looks for executable scripts") it seems that we no longer have that perf regression with coinbase. It would be nice to know why that slow down... are there so many of such files? Or are they very large? Is analyzing the files that is slow or is the heuristic to check whether we should analyze them?

I'm not expecting repos to have such a large number of executable scripts that they introduce a significant slowdown. The current approach as implemented in semgrep-core (Guess_lang module) is that given a language, we tell whether a file (executable, no extension) is in that language. In theory, it could be costly. In practice, I'm not sure. Maybe that's the problem we're seeing with the coinbase repo. If we have a problem because of n languages each inspecting the same script to determine if it's a valid target, we could start thinking of a classification mechanism: given a target, determine which language(s) it's in by inspecting it just once, instead of n times.

Anyway, this is now commented out. I'm planning to get back to this next week after I make some progress on bash.

mjambon:
I'm currently re-running the tests locally with `--snapshot-update` because some paths were updated. When the tests pass, this PR should be good to merge.

IagoAbal (approved):

PR #3952, opened 2021-09-28
Replace bench folders with config file

emjin:
Previously, the PA benchmarks were added by creating a new folder, adding a prep file, and then adding the bench run to the list of corpuses. This is a pretty annoying process. In that vein, @minusworld added functionality to run benchmarks from a config file for the security team. This PR migrates the PA benchmarks to also use that config file.

In general, this was straightforward. The exception was the nodejsscan rules. Since these behaved a little exceptionally (we need to clone another repo, we need to cd into a specific folder) it was hard to port this functionality. As a lazy shortcut to be able to merge, I made a special case if you say your config is `njsscan`. It will checkout a specific commit of njsscan and copy the semgrep rules from there. 

This means the benchmarks are not identical in one specific repo: `njs-old-dvna` originally used a different commit from njsscan. Since we have not be monitoring this very closely (we're more interested in how changes in the engine change performance), I decided this was unimportant.

Also performs some cleanup work:
- removes semgrep-core-benchmarks (no longer necessary now -fast is the default)
- removes gitlab benchmarking mode (these were convenient when we cared a lot about our speed in gitlab; now we should just look at our metrics)

PR checklist:
- [ ] Changelog is up-to-date


aryx:
Can we cancel this PR?

emjin:
I'm going to pick this back up today, got busy with the metrics stuff + interviewing

emjin:
Having a hard time getting rid of the semgrep complaints---looks like a semgrep-action bug. I don't think it would happen again if these were merged. See https://returntocorp.slack.com/archives/C6A1GN4P3/p1634595264027800

mjambon (approved):

PR #3958, opened 2021-09-29
tainting: Filter out sanitizers that conflict with sources/sinks

IagoAbal:
On certain codebases, assuming that every function call will always
propagate taint leads to too many false alarms. But fixing this via
a `$F(...)` sanitizer was very cumbersome, because `$F(...)` will
sanitize all sinks and sources, unless you re-enumerate all of them
via `pattern-not`s.

Closes PA-402

test plan:
make test # test included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## Unreleased
 
 ### Added
+
+### Changed
+- taint-mode: Sanitizers that match exactly a source or a sink are filtered out,
+  making it possible to use `- pattern: $F(...)` for declaring that any other
+  function is a sanitizer
+
+### Fixed
+
+## [0.67.0](https://github.com/returntocorp/semgrep/releases/tag/v0.67.0) - 09-29-2021
+
+### Added
 - Added support for break and continue in the dataflow engine
 - Added support for switch statements in the dataflow engine
 
--- a/semgrep-core/tests/tainting_rules/php/opaque_safe.yaml
+++ b/semgrep-core/tests/tainting_rules/php/opaque_safe.yaml
@@ -1,0 +1,14 @@
+rules:
+  - id: tainted
+    languages:
+      - php
+    message: Match
+    mode: taint
+    pattern-sanitizers:
+      - pattern: $F(...)
+    pattern-sinks:
+      - pattern: sink(...)
+    pattern-sources:
+      - pattern: tainted(...)
+    severity: WARNING
+
--- a/semgrep-core/src/engine/Tainting_generic.ml
+++ b/semgrep-core/src/engine/Tainting_generic.ml
@@ -129,8 +129,18 @@ let taint_config_of_rule default_config equivs file ast_and_errors
     |> List.concat
   in
   let sources_ranges = find_ranges spec.sources
-  and sanitizers_ranges = find_ranges spec.sanitizers
   and sinks_ranges = find_ranges spec.sinks in
+  let sanitizers_ranges =
+    find_ranges spec.sanitizers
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -21,6 +20,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Fixed bug in semgrep-core's `-filter_irrelevant_rules` causing Semgrep to
   incorrectly skip a file (#3755)
 - PHP: allows more keywords as valid field names (#3954)
+- taint-mode: It is now possible to tell Semgrep to trust third functions and assume
```
aryx: third-party

semgrep-core/src/analyzing/Dataflow_tainting.ml
```diff
@@ -40,6 +40,7 @@ type mapping = unit Dataflow.mapping
 type fun_env = (Dataflow.var, unit) Hashtbl.t
 
 type config = {
+  opaque_is_safe : bool;
```
aryx: why not third_party_is_safe?

semgrep-core/src/analyzing/Dataflow_tainting.mli
```diff
@@ -6,6 +6,10 @@ type fun_env = (Dataflow.var, unit) Hashtbl.t
   * Note that here [Dataflow.var] is a string of the form "<source name>:<sid>". *)
 
 type config = {
+  opaque_is_safe : bool;
+      (** If [opaque_is_safe] taint tracking will assume that third functions, that
```
aryx: third-party

aryx: but really it's not even third-party, it's just every functions.

aryx (approved):
I feel like we need a better name.

semgrep-core/src/analyzing/Dataflow_tainting.mli
IagoAbal: Yeah I am not entirely happy with any term. I say "opaque" in many places because there are functions like `Concat` that we understand, whereas other functions are "opaque" to Semgrep and those are the ones affected by this option. But then I use "third(-party)" in others because I'm not sure whether users will understand what is "opaque" here.

IagoAbal: What about "other opaque functions", despite being a bit long. Config option could be `other_opaque_funs_are_safe`?

IagoAbal:
I'll try something else tomorrow, maybe to just prevent sinks from being sanitized by:
```
pattern-sanitizers:
- pattern: $F(...)
```

IagoAbal:
Sorry about confusing everyone! I dramatically changed the PR for another solution that should achieve the same. It is much simpler and does not introduce a new `options:`. We simply filter out sanitizers that conflict with sources/sinks. CC @aryx

I am just slightly concerned that if somebody _unintentionally_ declares a sanitizer other than `$F(...)` that conflicts with e.g. a sink, it will be silently filtered out. Maybe we could issue some warning when we filter out sanitizers, but I leave this as a TODO.

semgrep-core/tests/tainting_rules/php/opaque_safe.yaml
```diff
@@ -0,0 +1,14 @@
+rules:
+  - id: tainted
+    languages:
+      - php
+    message: Match
+    mode: taint
+    pattern-sanitizers:
+      - pattern: sanitizer(...)
```
aryx: should it be $F instead here?
to match the example mentioned in the PR?

IagoAbal: That's right I forgot to update it! Test fails otherwise. Fixing it.

PR #3959, opened 2021-09-29
Add support for switch statements in the IL

mmcqd:
Switch statements can now be translated into the IL and analyzed by taint mode. Switch statements with no scrutinee (as in Go) are also supported.

Test plan: make test (see added test files)
PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -719,6 +719,11 @@ let parameters _env params =
 (* Statement *)
 (*****************************************************************************)
 
+(* TODO: What other langauges have no fallthrough? *)
```
aryx: typo, languages

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -753,7 +765,40 @@ let rec stmt_aux env st =
         List.map (stmt env) (st2 |> Common.opt_to_list) |> List.flatten
       in
       ss @ [ mk_s (If (tok, e', st1, st2)) ]
-  | G.Switch (_, _, _) -> todo (G.S st)
+  | G.Switch (tok, scrut_opt, cases_and_bodies) ->
```
aryx: scrut? what does it mean?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -928,6 +973,69 @@ let rec stmt_aux env st =
   | G.OtherStmtWithStmt _ ->
       todo (G.S st)
 
+(* TODO: Maybe this and the following function could be merged *)
+and scrut_and_cases_to_exp env tok scrut_orig scrut cases =
+  (* If there is a scrutinee, the cases are expressions we need to check for equality with the scrutinee  *)
```
aryx, line 960: ah scrutinee. Is it a standard name? I'd rather use expr, or switch_expr

mmcqd: yeah it's definitely a name I've seen before. I can change it

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -719,6 +719,11 @@ let parameters _env params =
 (* Statement *)
 (*****************************************************************************)
 
+(* TODO: What other langauges have no fallthrough? *)
+let no_fallthrough : Lang.t -> bool = function
```
IagoAbal: Maybe `no_switch_fallthrough` ?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -753,7 +765,40 @@ let rec stmt_aux env st =
...
+        match scrut_opt with
+        | Some scrut ->
+            let ss, scrut' = expr_with_pre_stmts env scrut in
+            (ss, scrut_and_cases_to_exp env tok scrut scrut')
+        | None -> ([], cases_to_exp env tok)
+      in
+      let break_label, break_label_s, switch_env =
+        mk_switch_break_label env tok
+      in
+      let rec cases_and_bodies_to_stmts = function
```
IagoAbal: `stmt_aux` is already part of a group of mutually recursive functions, so I would avoid nesting another recursive function in here. We could put it at the same level as `scrut_and_cases_to_exp ` and such.

IagoAbal: Could be nice adding a comment that gives a small example of what kind of transformation is done, like in `python_with_stmt`.

IagoAbal: I guess it's because you want to avoid passing `translate_cases` to it ?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -928,6 +973,69 @@ let rec stmt_aux env st =
   | G.OtherStmtWithStmt _ ->
       todo (G.S st)
 
+(* TODO: Maybe this and the following function could be merged *)
+and scrut_and_cases_to_exp env tok scrut_orig scrut cases =
+  (* If there is a scrutinee, the cases are expressions we need to check for equality with the scrutinee  *)
+  let ss, es =
+    List.fold_right
```
IagoAbal: Does `fold_right` achieve something interesting here? I mean, wrt `fold_left`. Is the order important?

IagoAbal: I'm asking because in OCaml's Stdlib `fold_right` is not tail-rec, and if we encounter some absurdly large switch statement that may cause a stack overflow. Whereas `fold_left` is safe in this respect.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -928,6 +973,69 @@ let rec stmt_aux env st =
...
+            impossible (G.Tk tok)
+        | G.Case (tok, _) -> (ss, fixme_exp ToDo (G.Tk tok) scrut_orig :: es))
+      cases ([], [])
+  in
+  (ss, { e = Operator ((Or, tok), es); eorig = scrut_orig })
+
+and cases_to_exp env tok cases =
+  (* If we have no scrutinee, the cases are boolean expressions, so we Or them together *)
+  let ss, es =
+    List.fold_right
```
IagoAbal: Same here.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -928,6 +973,69 @@ let rec stmt_aux env st =
...
+      (fun case (ss, es) ->
+        match case with
+        | G.Case (_, G.PatLiteral l) ->
+            ( ss,
+              (* TODO: seems bad to make an artificial eorig, but seems to be nothing to use  *)
```
IagoAbal, line 1000: In similar cases we pass the `eorig` as an argument to the function.

IagoAbal: Hmmm, but yeah, we have no scrutinee so... I also don't know what else to do. 😄 Maybe it should be an optional.

IagoAbal (approved):
Very nice!

Issue #3953, opened 2021-09-28
Conditional expression is not matched by deep expression


mjambon:
**Describe the bug**

**To Reproduce**

All 3 methods should match but currently the first one doesn't.

pattern:
```
<... a ...>;
...;
<... b ...>
```

```
public class FooClass{
    public bool Bad1(){
        a;
        if(b){}
    }
    public bool Bad2(){
        a;
        cond = b;
        if(cond){}
    }
    public bool Bad3(){
        a;
        if(true){
            b;
        }
    }
}
```

https://semgrep.dev/s/mjambon:match-conditional

- [ ] P0: blocking your adoption of Semgrep or workflow
- [x] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-417/conditional-expression-is-not-matched-by-deep-expression.

nbrahms:
Fixed in `cf73f8af36301ac7aa362699652608c66eb2a295`

PR #3960, opened 2021-09-29
Match deep ellipsis against statement sub-expressions

nbrahms:
Previously, we only supported deep-ellipsis for matching against
expression sub-expressions.

That is, the following pattern:

    if (cond) {
      <... a ...>
    }

will match:

    if (cond) {
      some_complicated_expression_of(a);
    }

However, we would not match statement sub-expressions. So the pattern:

    <... a ...>
    b

would not match:

    if (a) { ... }
    b

This commit generalizes m_expr_deep to enable use with either expression
or statement sub-expressions (actually to anything for which we can
extract sub-expressions), and then adds a matching branch for
deep-ellipsis against statements.

Closes PA-41.

PR checklist:
- [x] Changelog is up-to-date


mjambon:
Maybe you need to install ocamlformat for the pre-commit hook to run correctly locally. If you don't have it installed, there's no error on purpose.

mjambon (approved):
Awesome.

mjambon:
I did the ocaml reformatting. Merging this when CI succeeds.

mjambon:
I thought this would fix https://github.com/returntocorp/semgrep/issues/3953 (C#) but it doesn't, according to https://semgrep.dev/s/mjambon:match-conditional?version=develop

nbrahms:
Is something up with the Docker building? I copied the test case in this code into your link, and it doesn't match. 🤔

PR #3966, opened 2021-09-30
Make Semgrep work with subshell targets

nbrahms:
E.g.

    semgrep -e 'a' --lang js <(echo 'a')

To do this, we recycle the logic for redirecting stdin to copy subshells
to temporary files.

Note that we _must_ do this in the Python wrapper, as the semgrep-core
subprocess does not have access to these file descriptors (without
explicit re-wiring in the CLI).

Fixes PA-423.

PR checklist:
- [x] Changelog is up-to-date


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Added
 - Added support for break and continue in the dataflow engine
+- Input can be derived from subshells: `semgrep -f ... <(...)`
```
brendongo: Wrong version :D

brendongo (approved):

CHANGELOG.md
nbrahms: It works! (try `semgrep -f p/ci ...`)

PR #3970, opened 2021-10-01
Make Semgrep work with multiple subshells

nbrahms:
Previously (#3966) we added support for reading targets from a subshell.
However, this only worked with a single subshell target.

This was because we were deleting the created temp file prior to running
semgrep-core from the Semgrep CLI.

Here, add ability to support multiple subshells.

PR checklist:
- [x] Changelog is up-to-date


brendongo (approved):

semgrep/foo.js
```diff
@@ -0,0 +1,2 @@
+a
```
IagoAbal, line 1: Was this intended to be committed?

PR #3874, opened 2021-09-14
Add new default metrics collection when pulling from server

nbrahms:
This PR changes the default behavior of Semgrep’s metrics. It adds a new default “auto metrics” mode. Notably, these metrics changes will only affect users who are using the Semgrep.dev Registry.

### Auto metrics mode

* When running with rules from the Registry, Semgrep collects metrics by default, unless `--metrics off` is set.
* When running with a `--config` pointing at a local file that does not identify itself as coming from the Registry, or with `-e/--pattern`, Semgrep does not collect metrics by default, unless `--metrics on` is set.
* The metrics themselves are unchanged and remain the same pseudonymous data described in the Semgrep docs (https://semgrep.dev/docs/metrics/). Code, of course, isn’t and will never be uploaded (in fact there isn’t and will not be any code path in Semgrep that could make that possible).
        
### Disabling metrics
* To disable metrics entirely, run Semgrep with `--metrics off`, or with `SEMGREP_SEND_METRICS` environment variable set to `off`.

### Motivations
* We need to prioritize where to improve Semgrep’s performance and language support. Semgrep added ~13 languages in the past year, and also became committers to the tree-sitter project, which powers parsing. Multi-language work has increased support load and development cost of the engine and rules. If someone contributes a slow rule, it’s difficult to trace the stack (from parser to AST generic engine to rule) to understand which part is making it slow. We have aggressive performance goals for Semgrep (think: in-editor speed) so capturing these metrics helps us understand real-world performance and be data-driven in prioritization.
* We plan to open-source and merge our (currently proprietary) Semgrep CI wrapper, semgrep-action, into the main Semgrep codebase. This requires managing the metrics that semgrep-action already collects (https://github.com/returntocorp/semgrep-action/blob/develop/PRIVACY.md).
* Our decisions around metrics are informed by these principles:
  * Transparency: Collect and use data in a way that is clearly explained to the user and benefits them
  * User control: Put users in control of their data at all times
  * Limited data: Collect what is needed, de-identify where possible, and delete when no longer necessary
* We’re opening a comment period to collect feedback about these changes. We’d deeply appreciate any input on this topic.

### Additional details

r2c users can see https://r2c.quip.com/b2lbAogcYhyF/semgrep-metrics-auto-PRD for
more information.

Closes PA-335.

PR checklist:
- [ ] documentation is up to date
- [ ] changelog is up to date

PR #3973, opened 2021-10-02
feat(config_resolver): Pull packs from versioned CDN

brendongo:
PR checklist:
- [ ] Changelog is up-to-date


semgrep/semgrep/rule_schema.yaml
```diff
@@ -386,8 +386,6 @@ properties:
         pattern-either:
           $ref: '#/definitions/pattern-either-content'
 
-      additionalProperties: false
```
brendongo, line 390: Allow `version` field in rule

semgrep/semgrep/config_resolver.py
```diff
@@ -455,13 +458,177 @@ def saved_snippet_to_url(snippet_id: str) -> str:
...
+
+
 def resolve_config(config_str: str) -> Dict[str, YamlTree]:
     """ resolves if config arg is a registry entry, a url, or a file, folder, or loads from defaults if None"""
     start_t = time.time()
     if config_str in RULES_REGISTRY:
         config = download_config(RULES_REGISTRY[config_str])
     elif is_url(config_str):
         config = download_config(config_str)
+    elif is_pack_id(config_str) and SEMGREP_CDN_BASE_URL:
```
brendongo, line 647: Read from env variable so can act as an escape hatch. We can have user bypass registry cdn resolution and fall back to regular resolution by setting the env var to ""

semgrep/semgrep/config_resolver.py
```diff
@@ -455,13 +458,177 @@ def saved_snippet_to_url(snippet_id: str) -> str:
     return registry_id_to_url(f"s/{snippet_id}")
 
 
+def download_pack_config(ruleset_name: str) -> Dict[str, YamlTree]:
```
brendongo, line 478: Single large function since there are large refactors in https://github.com/returntocorp/semgrep/pull/3944 and https://github.com/returntocorp/semgrep/pull/3874. Will clean up once those are merged in.

semgrep/semgrep/constants.py
```diff
@@ -23,6 +23,8 @@
 if SEMGREP_USER_AGENT_APPEND is not None:
     SEMGREP_USER_AGENT = f"{SEMGREP_USER_AGENT} {SEMGREP_USER_AGENT_APPEND}"
 
+SEMGREP_CDN_BASE_URL = os.environ.get("SEMGREP_CDN_BASE_URL", "https://cdn.semgrep.dev")
```
nbrahms, line 26: Is there a similar set of registry endpoints for staging & dev?

semgrep/semgrep/rule_schema.yaml
nbrahms: mmm I think instead you should just add the version field; this change will prevent us from flagging when people put random trash in their rules, which will then cause us issues when we change the API at a a later date

semgrep/semgrep/constants.py
brendongo: https://staging-cdn.semgrep.dev and https://dev-cdn.semgrep.dev

semgrep/setup.py
```diff
@@ -122,6 +122,7 @@ def find_executable(env_name, exec_name):
     "click-option-group>=0.5.3",
     "requests>=2.22.0",
     "ruamel.yaml>=0.16.0,<0.18",
+    "semantic_version~=2.8.5",
```
nbrahms: If we wanted to reduce the dependency graph (since this is a command-line tool), you could presumably just use the version logic already built into `packaging`.

stubs/semantic_version.pyi
```diff
@@ -0,0 +1,27 @@
+from typing import Any, Iterable, Optional
```
nbrahms: wow this isn't in typeshed yet?

nbrahms (approved):
Nice @brendongo !

Pre-emptively approving, but I would strongly recommend restoring the `allowedProperties` in `rule_schema.yaml`, and instead just adding another property for the `verison`.

semgrep/semgrep/constants.py
nbrahms: 🐧

brendongo:
My reasoning there is we might want to start adding extra fields soon (i.e. human readable name separate from id, rule markdown readme etc.) and if we do older versions of semgrep will fail. For now we always run latest rule published but I think once we add the ability to run a specific version of a ruleset/rule we can put the allowedProperties back to false.

semgrep/setup.py
brendongo: Good idea

semgrep/semgrep/rule_schema.yaml
brendongo: hmm that's a fair point.

brendongo: I guess I'll add the version field and we can punt down making extra fields optional to a later discussion

PR #3986, opened 2021-10-05
Scan scripts

nbrahms:
Currently, Semgrep only scans files if they either:
- match an extension for the language being scanned
- are explicitly passed to Semgrep

However, scripting languages are often used to write extensionless
files, with the script interpreter in a "shebang" line.

This commit updates Semgrep to also scan these scripts, without
requiring them to be explicitly passed to Semgrep. To do this, we also
look for executable files, then read the shebang of each executable
file.

On this repo, this algorithm consumes approximately 50 us per file
(including the time to also find files with matching extensions).

Note that Semgrep actually used to have this behavior prior to Semgrep
0.11, but it was removed when moving file handling into the Python
wrapper. cf. Guess_lang.ml.

PR checklist:
- [ ] Changelog is up-to-date


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - taint-mode: Sanitizers that match exactly a source or a sink are filtered out,
   making it possible to use `- pattern: $F(...)` for declaring that any other
   function is a sanitizer
+- Semgrep now scans executable scripts whose shebang interpreter matches the
+  rule's language
```
mjambon: There's no need to inspect the files for a shebang line because semgrep-core already does it. It's in `Guess_lang`: https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/src/core/Guess_lang.ml

nbrahms: Yes, but without this logic, the file names never get passed to semgrep-core.

semgrep/semgrep/target_manager_extensions.py
```diff
@@ -99,7 +107,7 @@ def all_supported_languages() -> List[str]:
 # create a dictionary for fast lookup and reverse lookup
 _EXTS_TO_LANG: Dict[FileExtension, Language] = {}
 for language in _LANGS_TO_EXTS.keys():
-    for extension in _LANGS_TO_EXTS[language]:
+    for extension in _LANGS_TO_EXTS[language][0]:
```
brendongo, line 113: Can we change the name of the object if it is not just mapping to extensions anymore?

semgrep/semgrep/target_manager.py
```diff
@@ -138,71 +143,111 @@ def _parse_output(output: str, curr_dir: Path) -> FrozenSet[Path]:
...
+                deleted_output = sub_check_output(
+                    ["git", "ls-files", "--deleted"],
+                    cwd=curr_dir.resolve(),
+                    encoding="utf-8",
+                    stderr=subprocess.DEVNULL,
+                )
+                tracked = _parse_output(tracked_output, curr_dir)
+                untracked_unignored = _parse_output(untracked_output, curr_dir)
+                deleted = _parse_output(deleted_output, curr_dir)
+                paths = tracked.union(untracked_unignored).union(deleted)
```
brendongo:
```suggestion
                paths = tracked.union(untracked_unignored).difference(deleted)
```

brendongo: Can we add a test for this too. (Specifically TargetManager should not be returning a file that has been marked by git as deleted)

mjambon (approved):
I'll let you fix the conflicts

semgrep/semgrep/target_manager.py
nbrahms: Where did we end up on porting all the integration tests from Bento? (They tested all these vagaries of functionality)

CHANGELOG.md
emjin: Seems really unfortunate that we do this logic twice

brendongo (changes requested):
Changelog in bad state

brendongo (approved):

semgrep/semgrep/target_manager_extensions.py
brendongo: sigh

CHANGELOG.md
brendongo: yep :(

PR #3998, opened 2021-10-05
Allow invisible rules

nbrahms:
These will be used in conjunction with --config auto (see #3944) to
determine the languages and frameworks of a project, in order to tune
rules on a per-project basis.

Because these are used internally, and are not "findings", we hide their
results from the user.

PR checklist:
- [X] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -247,7 +247,7 @@ type taint_spec = {
 type mode = Search of pformula | Taint of taint_spec [@@deriving show]
 
 (* TODO? just reuse Error_code.severity *)
-type severity = Error | Warning | Info [@@deriving show]
+type severity = Error | Warning | Info | Inventory [@@deriving show]
 
 type rule = {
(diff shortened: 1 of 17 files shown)


minusworld (approved):
Exclusively from curiosity, why were many of the types changed to `Iterable`?

brendongo:
Does this play well with semgrep-action reporting the inventory rules to backend still?

nbrahms:
> Does this play well with semgrep-action reporting the inventory rules to backend still?

Those rules will have to be sent with their (current) `INFO` inventory. This doesn't touch that at all.

nbrahms:
> Exclusively from curiosity, why were many of the types changed to `Iterable`?

As good software practice, it's generally good to use the most generic type needed in input, and return the most specific type possible in output. Here I actually needed to do this, as I had `FrozenSet` objects and `Set` objects, neither of which inherit from the other. Note that both are of type `Collection`, but I didn't use anything about them being a collection (viz., the `in` operator), so I just use the more generic `Iterable` instead.

PR #4001, opened 2021-10-06
Add support for GitHub annotations as formatter output

directionless:
I was trying to use semgrep in github, and I _really_ wanted it's finding as annotations in the PR. This led me down a path of needing to get the dashboard/SaaS setup, and get permissions, and none of it worked, and I was chatted on slack. Etc. 

But, I think there's a much simpler route. According to the github [workflow docs](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-a-notice-message), github will parse specially formatted messages and treat them as annotations. This completely alleviates the need for any kind of permissions wrangling. 

This does run locally, though I have not tested this code as an action. I think there's a lot more plumbing for that, and I'm not sure if this is a direction y'all want to go in. But, I did test a basic PoC version of this, and it appears to work. 

PR checklist:
- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -304,6 +304,11 @@ def __validate_lang(option: str, lang: Optional[str]) -> str:
     is_flag=True,
     help="Output results in vim single-line format.",
 )
+@optgroup.option(
+    "--github",
+    is_flag=True,
+    help="Output results in github annotation format.",
+)
 @optgroup.group("Verbosity options", cls=MutuallyExclusiveOptionGroup)
 @optgroup.option(
     "-q",
@@ -408,6 +413,7 @@ def cli(
     exclude: Optional[Tuple[str, ...]],
     force_color: bool,
     generate_config: bool,
+    github: bool,
     include: Optional[Tuple[str, ...]],
     jobs: int,
     json: bool,
@@ -519,6 +525,8 @@ def cli(
         output_format = OutputFormat.EMACS
     elif vim:
         output_format = OutputFormat.VIM
+    elif github:
+        output_format = OutputFormat.GITHUB
 
     output_settings = OutputSettings(
         output_format=output_format,
--- a/semgrep/semgrep/constants.py
+++ b/semgrep/semgrep/constants.py
@@ -35,6 +35,7 @@ class OutputFormat(Enum):
     SARIF = auto()
     EMACS = auto()
     VIM = auto()
+    GITHUB = auto()
 
     def is_json(self) -> bool:
         return self in [OutputFormat.JSON, OutputFormat.SARIF]
--- a/semgrep/semgrep/formatter/github.py
+++ b/semgrep/semgrep/formatter/github.py
@@ -1,0 +1,33 @@
+from typing import Any
+from typing import FrozenSet
+from typing import Mapping
+from typing import Sequence
+
+from semgrep.constants import CLI_RULE_ID
+from semgrep.error import SemgrepError
+from semgrep.formatter.base import BaseFormatter
+from semgrep.rule import Rule
+from semgrep.rule_match import RuleMatch
+
+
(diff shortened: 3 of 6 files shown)


directionless:
FWIW the testing I did, was to add some steps to my semgrep github actions, that did this:
```
      - uses: ruby/setup-ruby@v1
        if: always()
      - name: annotate
        run: ruby ./semgrep-sarif-helper.rb semgrep.sarif
        if: always()
```

```
$ cat semgrep-sarif-helper.rb 
#!/usr/bin/env ruby

# Annotate via printf
# https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-a-notice-message

require 'json'

def anno(file:, startline:, endline:, title:, message:, level: 'warning')
  puts "::#{level} file=#{file},line=#{startline},endLine=#{endline},title=#{title}::#{message}"
end


sarif = JSON.parse(File.read(ARGV[0]))

sarif['runs'].first['results'].each do |result|
  result['locations'].each do |loc|
    anno(
      file: loc.dig('physicalLocation', 'artifactLocation', 'uri'),
      startline: loc.dig('physicalLocation', 'region', 'startLine'),
      endline: loc.dig('physicalLocation', 'region', 'endLine'),
      title: "semgrep failed for #{result['ruleId']}",
      message: result.dig('message', 'text'),
    )
  end
end
```

And I did get the annotations as expected. 

So I have high confidence in the approach, but don't know enough about semgrep to make this PR really shine. (Formatting, docs, python standards, etc)

brendongo:
This looks good @directionless ! Is there a way we can test the actual annotation step? Or does github just take any stdout that looks like that format and automagically makes am annotation/PR comment for it. 

Can you also add a quick test like https://github.com/returntocorp/semgrep/blob/develop/semgrep/tests/e2e/test_output.py#L114 intention is to catch and breakages to this output format if we do a refactor somewhere and forget to update this output format.

We run pre-commit to do auto formatting so you might want to install that https://semgrep.dev/docs/contributing/contributing-code/#development-workflow or just make the changes necessary to make the lint/pre-commit test pass.

directionless:
> This looks good @directionless ! Is there a way we can test the actual annotation step? Or does github just take any stdout that looks like that format and automagically makes am annotation/PR comment for it.

Yes, my understanding is that any stdout that matches their special format becomes an annotation. That's what I tested with the above ruby. (Which is a test of github functions, not this PR)

I'm not sure what the right way to test this is. Ideally, by building a test `semgrep-action` docker image, and running it in actions. I was hoping y'all had a setup for that. Even a docker image with this PR would probably be enough. 

I did not attempt to document this. Partly because I'm not sure if y'all will like it, or how it may eventually fit with the SaaS and the whole comment flow.

directionless:
TBH it might make more sense in https://github.com/directionless/semgrep-action/blob/develop/src/semgrep_agent/formatter.py (Though ideally, it could be in both). I'd assume these were more linked.

brendongo:
In the next ~3 months we have plans to absorb semgrep-action functionality into semgrep but for now it's still a little separate. If you want the functionality in semgrep-action output I would recommend making a new output format similar to gitlab-json output here: https://github.com/directionless/semgrep-action/blob/develop/src/semgrep_agent/main.py#L342

directionless:
Good to hear everything is moving into `semgrep`. I opened https://github.com/returntocorp/semgrep-action/pull/389 I don't know how y'all want to manage the two. (I'm not going to be offended any way y'all go)

brendongo:
Hey @directionless apologies on the delay release was a little hectic.

I think once we bring in semgrep-action features into semgrep this PR would be great but I think for now, might make sense to take https://github.com/returntocorp/semgrep-action/pull/389/files over the line instead. Will close this for now and continue discussion on that PR

PR #4007, opened 2021-10-06
CI: compare-perf: Reduce number of false positives

IagoAbal:
PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/scripts/compare-perf
+++ b/scripts/compare-perf
@@ -112,6 +112,14 @@ def main() -> None:
         min_rel_dur = min(min_rel_dur, rel_dur)
         max_rel_dur = max(max_rel_dur, rel_dur)
 
+        # Our benchmarks are small and we run them in GitHub-hosted runners,
+        # so we cannot distinguish between small perf deviations and noise.
+        # Without this filter, the warnings fire all the time with a very high
+        # false positive rate.
+        probably_noise = (
+            0.7 <= rel_dur and rel_dur <= 1.3 and abs(latest_time - baseline_time) < 2
+        )
+
         print(
             f"[{name}] {rel_dur:.3f}x "
             f"Baseline: {baseline_time:.3f}, Latest: {latest_time:.3f}"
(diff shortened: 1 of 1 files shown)


brendongo (approved):

scripts/compare-perf
```diff
@@ -112,6 +112,14 @@ def main() -> None:
         min_rel_dur = min(min_rel_dur, rel_dur)
         max_rel_dur = max(max_rel_dur, rel_dur)
 
+        # Our benchmarks are small and we run them in GitHub-hosted runners,
+        # so we cannot distinguish between small perf deviations and noise.
+        # Without this filter, the warnings fire all the time with a very high
+        # false positive rate.
+        probably_noise = (
+            0.7 <= rel_dur and rel_dur <= 1.3 and abs(latest_time - baseline_time) < 2
+        )
```
mjambon, line 121: What's the reasoning here? Why would the absolute time difference be a valid criterion for ignoring these results? You need to explain your statistical model, which just got more complicated. This is why I removed similar code that was there before. This seems just equivalent to partially silencing benchmarks that are too short. If they're too short, then they should be made longer. If there's no such thing as a benchmark that's too short, we shouldn't silence their results. If 2 seconds represents some unpredictable variability that's independent of the total duration, then you should explain where it is believed to come from and why it wouldn't be solved by running each benchmark multiple times. We're already running them twice, we could run them a couple more times. The key is that we expect a strongly one-sided distribution, because on a given host, a task can't finish faster than some minimum but it could go arbitrarily high due to concurrent utilization of the host's resources (CPU, storage access, other?). The minimum duration corresponds to the best case when there's no competition for these resources.

IagoAbal: We run the benchmarks in a rather busy and noisy environment. I don't know how many times exactly we should run the benchmarks to eliminate that noise, but two times is clearly insufficient. Just look at the 10 latest merged PRs, how many got these warnings? And how many were considered meaningful? But if I go and bump that to 4 times, then I am doubling the time that the CI job takes to run. If that's not sufficient, do we then bump it to 8 times?

We are already setting a 20% threshold for warnings and a 30% one for errors, numbers that are as arbitrary as the 2 seconds here. I simply noticed that a) the majority of the false alarms are due <2s absolute differences, b) in all the PRs I looked at for months nobody ever did anything to fix any of such warnings.

Also, I am not so sure about our choice of taking the minimum time of _N_ runs. The approach/reasoning does not sound right to me at first glance. Is there any book/paper/etc on benchmarking supporting it?

IMO this CI job is very useful for catching quite noticeable / _obvious_ performance problems. It is simply not able to distinguish small perf regressions from noise. It is better to use it for what is good at, and have a larger benchmark suite to catch small perf regressions.

PR #4009, opened 2021-10-06
tainting: Avoid having "built-in" sources or sanitizers

IagoAbal:
I am guessing they were used for quick experiments during taint-mode
early days, but right now they have no use since both `pattern-sources`
and `pattern-sanitizers` are required anyways. This probably doesn't
happen very often, but if e.g. `source` happened to be an object or
function in some real-world code, then we were "magically" considering
it as a source of taint!

test plan:
make test # test file included

PR checklist:
- [x] Changelog is up-to-date


mmcqd (approved):
Nice, surprised this never came up before haha.

aryx (commented):
oops

PR #4017, opened 2021-10-07
fix(target_manager): respect rule level path filters

brendongo:
PR checklist:
- [x] Changelog is up-to-date


mjambon (approved):
Thank you Brendon!

PR #4019, opened 2021-10-07
Replace named pipes by regular files to avoid errors when re-reading the file

mjambon:
This now works:
```
$ semgrep-core -lang bash -e '... &' -target <(echo 'ls &')
/tmp/semgrep-core-95ab40-63:1
 ls &
```

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -1307,6 +1307,34 @@ let options () =
     ]
 
 (*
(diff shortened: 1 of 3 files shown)


mjambon:
Note the `-target` flag which is now required to avoid filtering (since this file is not executable and has no `.bash` or `.sh` extension). This is new and probably a design mistake because files passed directly on the command line should not be subject to filtering. I think I'll change it later.

PR #4024, opened 2021-10-08
Fix problem with If body that needs to be a Block + tests

mjambon:
I found out the hard way that a pattern of type `stmts` can't match an `If` body if it's not a `Block`, because only a `Block contains `stmts` (a list of statements).

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/parsing/ast/Bash_to_generic.ml
```diff
@@ -106,12 +106,15 @@ let stmt_or_expr_loc = function
...
-let block : stmt_or_expr list -> stmt_or_expr = function
-  | [ x ] -> x
-  | several ->
-      let loc = list_loc stmt_or_expr_loc several in
-      let stmts = List.map as_stmt several in
-      Stmt (loc, G.s (G.Block (bracket loc stmts)))
+(*
+   Important: do not optimize a singleton [stmt] into just stmt, because
+   it could no longer be matched by semgrep if the pattern is a list
+   of statements.
```
aryx, line 112: I don't understand. If the pattern is a list of statements (so at least 2 elts), why would we want that it
maches a single stmt? 
Also your examples do not contain any pattern with multiple statemnet; it's each time just an expression pattern.

aryx (changes requested):

semgrep-core/src/parsing/ast/Bash_to_generic.ml
mjambon: The pattern `a` (in bash) is currently parsed as a list of statements rather than a single statement, just like `a; b`. The pattern `a` then doesn't match `if b; then a; fi` unless the `a` after `then` is also a list of statements, which wasn't the case. It looks like a question of consistency. Perhaps it gets easier if we have that in mind, I'm not sure. The other thing I'm not sure about is how to enforce consistency and how to catch these mistakes.

What I can do is:
* This PR: have all single statements be actually a Block. This doesn't make much sense in retrospect. Instead, it would make more sense to change the type of the `then` branch to `stmts` instead of `stmt`, but that may be a bad idea for other reasons.
* Ensure that all patterns of the form `Block (open_br, [stmt], close_br)` are simplified to `stmt`.

The second solution seems better to me right now. However, the simplification should only take place if the brackets are fake, as in the case of bash statements, which is dirty compared to directly constructing the tree correctly in the first place. At least we can document the `Block` construct where it is defined in `AST_generic` and explain what can go in there.

I'll make a PR implementing the second solution.

mjambon: See https://github.com/returntocorp/semgrep/pull/4031 for the alternate solution.

PR #4025, opened 2021-10-08
Add optional tracing for the matcher

mjambon:
This is the least intrusive way I found to trace the pairs (pattern node, target node) being matched with the guarantee that it doesn't affect performance during normal operation. I used this during the investigation of #4024 .

When turned on, the output looks like this:
```
----- m_stmt -----
stmt pattern:
ExprStmt(Call(N(...), [...]), ())
~~~~~
stmt target:
ExprStmt(Call(N(...), [...]), ())

----- m_expr -----
expr pattern:
Call(N(Id(..., ...)), [Arg(...)])
~~~~~
expr target:
Call(N(Id(..., ...)), [Arg(...)])

----- m_expr -----
expr pattern:
N(Id((..., ...), {id_resolved=...; id_type=...; id_constness=...; }))
~~~~~
expr target:
N(Id((..., ...), {id_resolved=...; id_type=...; id_constness=...; }))

----- m_expr -----
expr pattern:
Call(N(Id(..., ...)), [Arg(...)])
~~~~~
expr target:
Call(N(Id(..., ...)), [Arg(...)])

----- m_expr -----
expr pattern:
N(Id((..., ...), {id_resolved=...; id_type=...; id_constness=...; }))
~~~~~
expr target:
N(Id((..., ...), {id_resolved=...; id_type=...; id_constness=...; }))

----- m_expr -----
expr pattern:
L(String((..., ...)))
~~~~~
expr target:
L(String((..., ...)))

----- m_literal -----
literal pattern:
String(("a", ()))
~~~~~
literal target:
String(("b", ()))

----- m_expr -----
expr pattern:
Call(N(Id(..., ...)), [Arg(...)])
~~~~~
expr target:
L(String((..., ...)))

----- m_expr -----
expr pattern:
Call(N(Id(..., ...)), [Arg(...)])
~~~~~
expr target:
L(String((..., ...)))

----- m_expr -----
expr pattern:
Call(N(Id(..., ...)), [Arg(...)])
~~~~~
expr target:
N(Id((..., ...), {id_resolved=...; id_type=...; id_constness=...; }))

----- m_expr -----
expr pattern:
Call(N(Id(..., ...)), [Arg(...)])
~~~~~
expr target:
N(Id((..., ...), {id_resolved=...; id_type=...; id_constness=...; }))

```

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -71,6 +71,57 @@ let logger = Logging.get_logger [ __MODULE__ ]
...
+let print_arithmetic_operator_pair =
+  print_pair "arithmetic_operator" Meta_AST.vof_arithmetic_operator
+
+let print_function_definition_pair =
+  print_pair "function_definition" Meta_AST.vof_function_definition
+
+let print_class_definition_pair =
+  print_pair "class_definition" Meta_AST.vof_class_definition
+
+let print_definition_pair = print_pair "definition" Meta_AST.vof_definition
```
aryx: instead of all those print_xxx_pair you can define print_any_pair and then at the call site encapsulate
the xxx in an any (e.g., print_any_pair (T a) (T b)) in m_type.

aryx (approved):

semgrep-core/src/matching/Generic_vs_generic.ml
mjambon: I considered doing that but didn't do it because I want to keep the code under `let m_stmt ... =` as free of distractions as possible. I don't mind having boilerplate elsewhere because it's not hard to maintain. Here's how both solutions look:

```ocaml
and m_stmt a b =
  if trace_matching then print_stmt_pair a b;
```
vs.
```ocaml
and m_stmt a b =
  if trace_matching then print_pair "stmt" (S a) (S b);
```
I can also move the whole `(* Tracing *)` section to another module to get it out of the way. It was more compact before ocamlformat did its thing.

mjambon: ok, so I moved the tracing code to its own module. Calls look like this now:
```ocaml
Trace_matching.(if on then print_stmt_pair a b);
```

aryx:
Do you think ocamlopt optmize the if now that the on global is defined in another module?

mjambon:
I tested this locally in a small program, using `ocamlopt -c -S Foo.ml` without special options. My ocaml setup uses flambda, which maybe we're not using in prod (?). Anyway, for what it's worth here's what I ran:
```
#!/usr/bin/env bash
#
# Test if ocamlopt removes 'if A.x then ...;' when A.x is false.
#

cat > B.ml <<EOF
let main () =
  A.(if on then print_endline "hello");
  ignore (1 + 1)

let () = main ()
EOF

echo 'let on = true' > A.ml
ocamlopt -c -S A.ml
ocamlopt -c -S B.ml
mv -f A.s A.s.on
mv -f B.s B.s.on

echo 'let on = false' > A.ml
ocamlopt -c -S A.ml
ocamlopt -c -S B.ml
mv -f A.s A.s.off
mv -f B.s B.s.off

echo "diff between on=true and on=false"
diff -u B.s.on B.s.off
```
which shows that the code containing `print_endline` is gone :tada: :
```
-	call	camlStdlib__print_endline_1181@PLT
```

PR #4026, opened 2021-10-08
tainting: Always check subexpressions and function call arguments

IagoAbal:
Follows: c31c31895b0 ("tainting: Make sure we always check function call arguments (#3996)")

test plan:
make test # test included

PR checklist:
- [x] Changelog is up-to-date


mmcqd (approved):
Looks good. I feel like relying on short circuiting, especially with no comment mentioning it, was kind of confusing.

IagoAbal:
> Looks good. I feel like relying on short circuiting, especially with no comment mentioning it, was kind of confusing.

The fact that we are inferring "tainted" and checking tainted sinks at the same time also helps making it confusing. Also, at some point I was trying to minimize duplicate findings... We may want to rewrite the traversal in a more sane way.

PR #4040, opened 2021-10-12
remove severity from text output

ievans:
PR checklist:
- [x] Changelog is up-to-date


ievans:
Alternatives/other enhancements: 
* we could order output by severity
* we could support a --severity= flag

PR #4048, opened 2021-10-12
Revert "Allow non-explicit targets to be filtered out by semgrep-core (#3949)"

nbrahms:
This reverts commit 771e6ade18665a44ecd64119064cfc9e72c49ab8.

This caused extra findings as well as potential performance issues;
reverting until we can add more thorough end-to-end tests.

Aside from minor Python `List` -> `Sequence` change to satisfy mypy, this entirely is a --mine conflict resolution


nbrahms:
Aside from minor change to satisfy `mypy`, this entirely is a `--mine` conflict resolution

semgrep/tests/unit/test_target_manager.py
```diff
@@ -766,74 +716,49 @@ def test_explicit_path(tmp_path, monkeypatch):
...
-    targets = TargetManager(
-        [], ["foo/a.py"], [".", "foo/a.py"], False, defaulthandler, False
+    assert foo_a in TargetManager(
+        [], ["foo/a.py"], 0, [".", "foo/a.py"], False, defaulthandler, False
     ).get_files(python_language, [], [])
-    assert foo_a in targets.explicit
-    assert foo_a not in targets.filterable
 
-    # Should respect includes/excludes passed to get_files even if target explicitly passed
-    assert cmp_targets(
```
brendongo, line 795: We want to keep this test. Tests that explicit files still get filtered with rule path includes

brendongo:
Also reverts https://github.com/returntocorp/semgrep/pull/4017 and https://github.com/returntocorp/semgrep/pull/4021

brendongo (approved):
Lgtm. @mjambon is there anything here we relied on in future PRs?

mjambon:
> Lgtm. @mjambon is there anything here we relied on in future PRs?

Not sure what you mean. Yes, we'll need those changes if we want to move forward, with bash in particular. Right now, it's fine to revert, I don't expect immediate breakage due to reverting this merge.

The original changes were supposed to speed up some scans. I believe they do because the median scan time hasn't changed since the deployment, and at least one user is reporting a 50%+ slowdown. If it's not isolated, then it would mean that some scans indeed got faster, as intended. In this scenario, we have a mix of speedups and slowdowns.

This set of changes was meant to be a strict reduction of the number of targets processed by semgrep-core (by declaring some of them as filterable when they were originally all considered explicit).

IagoAbal:
Do we know how this caused the extra findings? Initially the PR was _removing_ the filtering of files based on the extension in some cases, but I believe that was not in the final PR that was merged. So what we had left (supposedly) was a PR that allowed us to filter more files but should not let pass any extra file.

brendongo:
@mjambon yeah was referring to if there would be immediate breakage

@IagoAbal there were extra findings but I think I got most of them with the fix PRs in 4017 and 4021 but I think there were still bugs with handling explicit files. Tbh this was more a revert for performance that some users were seeing and I don't fully understand what is causing it but I guess we'll find out if it works

PR #4062, opened 2021-10-13
chore(ci): Update ref of pre-commit semgrep to 0.69.0

brendongo:
PR checklist:
- [ ] Changelog is up-to-date


semgrep/semgrep/settings.py
```diff
@@ -36,6 +36,7 @@ def __init__(self) -> None:
         if self._path.exists():
             with self._path.open("r") as fd:
                 yaml_file = self._yaml.load(fd)
+                print(yaml_file)
```
brendongo:
```suggestion
```

PR #4073, opened 2021-10-14
Enforce metavariable equality between sources/sinks/sanitizers, and show metavariables in taint mode messages

mmcqd:
We can now write rules that use the same metavariable across sources/sinks/sanitizers, and refer to these metavariables in match messages. Currently if a sink could be matched with multiple possible metavariable assignments it will be matched twice by `semgrep-core`, but `semgrep` will filter out all but one of these matches.

PR checklist:
- [x] Changelog is up-to-date


aryx (changes requested):
I don't want the core to depends on analyzing. There has to be a way to avoid this dependency.
In fact in core/dune we should remove pfff-lang_GENERIC-analyze, it's not currently needed; not sure why it was there in the first place.

aryx:
Maybe you can move Dataflow_tainting.ml in engine/, and add a note that it used to be in analyzing/, but with
metavariable checking the environment now depends on Pattern_match.t, which means it depends on semgrep-core/core.
Then say it can't be in analyzing/ because
we want analyzing/ to be independent of semgrep-core/core because it's used in other projects (e.g., codemap from aryx)

semgrep-core/src/core/Pattern_match.ml
```diff
@@ -116,3 +116,25 @@ let no_submatches pms =
                  Hashtbl.replace tbl k (pm :: ys')));
   tbl |> Hashtbl.to_seq_values |> Seq.flat_map List.to_seq |> List.of_seq
   [@@profiling]
+
+type int_range = int * int * int * int [@@deriving ord]
```
aryx: A comment on what those 4 ints are?
lines? columns? if it's for the left and right part, maybe they need to better grouped.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,286 @@
...
+ * This is step1 for taint tracking support in semgrep.
+ * This was originally in semgrep-core/src/analyze, but it now depends on Pattern_match,
+ * so it was moved to semgrep-core/src/engine
+ *)
+
+(*****************************************************************************)
+(* Types *)
+(*****************************************************************************)
+
+type mapping = PM.Set.t Dataflow.mapping
```
aryx, line 39: So PM.Set.t is a set of pattern matches. How does it related to "a variable is 'tainted'"?
I would expect a bool Dataflow.mapping for "whether a variable is tainted".

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,286 @@
...
+
+let str_of_name ((s, _tok), sid) = spf "%s:%d" s sid
+
+let set_opt_to_set = function
+  | None -> PM.Set.empty
+  | Some x -> x
+
+let ( <::> ) x = Option.map (fun xs -> x :: xs)
+
+let unify_meta_envs env1 env2 =
```
aryx, line 80: comment, what this function is doing? Why we need it?

semgrep-core/src/engine/Dataflow_tainting.mli
```diff
@@ -1,15 +1,16 @@
...
+type fun_env = (Dataflow.var, Pattern_match.Set.t) Hashtbl.t
 (** Set of "tainted" functions in the overall program.
   * Note that here [Dataflow.var] is a string of the form "<source name>:<sid>". *)
 
 type config = {
-  is_source : AST_generic.any -> bool;
-  is_sink : AST_generic.any -> bool;
-  is_sanitizer : AST_generic.any -> bool;
-  found_tainted_sink : AST_generic.any -> unit Dataflow.env -> unit;
+  is_source : AST_generic.any -> Pattern_match.t list;
```
aryx, line 9: Can you add a comment on the meaning of the return type?
Before with bool it was pretty clear "whether or not this expr/stmt/any is a source.
But now why does it return a set of matches? If it's to allow to compare metavariables in different places,
then why we need Pattern_match.t? why not just a metavariable environment?

aryx (changes requested):
Asking for more comments mostly.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,286 @@
...
+
+let set_filter_map f pm_set =
+  PM.Set.fold
+    (fun pm pm_set ->
+      match f pm with
+      | Some pm' -> PM.Set.add pm' pm_set
+      | None -> pm_set)
+    pm_set PM.Set.empty
+
+let ( <|> ) = PM.Set.union
```
IagoAbal: IMO it's better to avoid defining custom operators unless they are going to be used a lot. If I encounter `PM.Set.union` I immediately know what that means, whereas if I read `<|>` I need to go look up what `<|>` means.

semgrep-core/src/engine/Dataflow_tainting.mli
mmcqd: Hmm you're right that we only need the metavariable environment for this, but tracking the pattern matches that originated a variable's taint will make it easier to implement something that displays information about how a sink ended up being tainted. I don't know whether or not it's worth preemptively tracking this information if this PR isn't going to be implementing that feature. It wouldn't be hard to switch to only tracking metavar envs if you think that would be a better idea.

semgrep-core/src/core/Pattern_match.ml
mmcqd: I'm just using builtin compare on the range_locs now. Seemed more straightforward than picking some ints to extract like I was doing here.

semgrep-core/src/core/Pattern_match.ml
```diff
@@ -116,3 +116,20 @@ let no_submatches pms =
...
+(* Silly, but we have to because type declarations are automatically recursive so [type t = t] doesn't work *)
+type pm = t
+
+module Set = Set.Make (struct
+  type t = pm
+
+  (* If the pattern matches are obviously different (have different ranges), this is enough to compare them.
+     If their ranges are the same, compare their metavariable environments. This is not robust to reordering
+     metavariable environments. [("$A",e1);("$B",e2)] is not equal to [("$B",e2);("$A",e1)]. This should be ok
+     but is potentially a source of unexpected behavior.
```
IagoAbal: What kind of unexpected behavior? I guess the worst that can happen is that we get duplicate findings?

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+   if they have a conflicting metavariable assignment, returns None,
+   otherwise returns the union of the environments. We use this to
+   prevent reporting that a sink is tainted if the source of taint and
+   the sink have conflicting metavariable environments
+*)
+let unify_meta_envs env1 env2 =
+  let xs =
+    List.fold_left
+      (fun xs (mvar, mval) ->
+        match List.assoc_opt mvar env2 with
```
IagoAbal, line 86: You can use let-punning https://ocaml.org/manual/bindingops.html if you want to simplify some `option`-handling code, e.g.:
```ocaml
let (let*) = Option.bind in
...
let* xs = xs in
...
| None -> Some ((mvar, mval) :: xs)
```

IagoAbal:
> I don't want the core to depends on analyzing. There has to be a way to avoid this dependency.

Sorry that I missed the initial context. @mmcqd was the `core` that depended on `analyzing`? Or `analyzing` that depended on `core`? I would only expect `Dataflow_tainting` to depend on `Pattern_match.t`.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+    pm_set PM.Set.empty
+
+let ( <|> ) = PM.Set.union
+
+(* Takes a pattern match (should be from a sink) and a set of pattern matches (should be sources),
+   returns a set containing a copy of the sink match for each source match with a metavariable environment
+   that could unify with the sink's environment. The copies will have their environments updated with the
+   successful unification.
+ *)
+let update_meta_envs pm =
```
IagoAbal: For readability I think it's better to declare the other argument that this function expects. Or write down the return type.

IagoAbal: It could be beneficial to use ocamldoc tags (https://ocaml.org/manual/ocamldoc.html#ss:ocamldoc-tags): 
```ocaml
(* @param sink_pm Pattern match of a sink.
   @param src_pms Set of pattern matches corresponding to sources.
   @returns [sink_pm] unified with each [src_pms] when possible. *)
let update_meta_envs sink_pm src_pms = 
  ...
```

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+
+let varmap_update env x data f =
+  match VarMap.find_opt x env with
+  | None -> VarMap.add x data env
+  | Some data' -> VarMap.add x (f data') env
+
+let hashtbl_update env x data f =
+  match Hashtbl.find_opt env x with
+  | None -> Hashtbl.add env x data
+  | Some data' -> Hashtbl.add env x (f data')
```
IagoAbal: You should probably use `Hashtbl.replace` here.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+  let sanitized_pms = config.is_sanitizer (G.E exp.eorig) in
+  match sanitized_pms with
+  | _ :: _ -> PM.Set.empty
+  | [] ->
+      let tainted_pms =
+        check_subexpr exp.e
+        <|> PM.Set.of_list (config.is_source (G.E exp.eorig))
+      in
+      let found = make_tainted_sink_matches sink_pms tainted_pms in
+      if PM.Set.is_empty found then tainted_pms
```
IagoAbal:
```ocaml
if not (PM.Set.is_empty found) then config.found_tainted_sink found env;
tainted_pms
```

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+  in
+  let node = flow.graph#nodes#assoc ni in
+  let out' =
+    match node.F.n with
+    | NInstr x -> (
+        let tainted = check_tainted_instr config fun_env in' x in
+        match (PM.Set.is_empty tainted, IL.lvar_of_instr_opt x) with
+        | true, Some var -> VarMap.remove (str_of_name var) in'
+        | _, Some var ->
+            varmap_update in' (str_of_name var) tainted (PM.Set.union tainted)
```
IagoAbal: I think it could be more readable to just use `VarMap.update` here.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+      VarMap.empty
+  in
+  let node = flow.graph#nodes#assoc ni in
+  let out' =
+    match node.F.n with
+    | NInstr x -> (
+        let tainted = check_tainted_instr config fun_env in' x in
+        match (PM.Set.is_empty tainted, IL.lvar_of_instr_opt x) with
+        | true, Some var -> VarMap.remove (str_of_name var) in'
+        | _, Some var ->
```
IagoAbal: Better write this `| false, Some var ->`, since AFAICT the `_` must be `false`, and the RHS only makes sense when it's `false`.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+            varmap_update in' (str_of_name var) tainted (PM.Set.union tainted)
+        | _ -> in')
```
IagoAbal: This could be a little bit more clear if it were `| _, None -> ...`. It has the nice effect that you can now reorder the clauses without altering the semantics.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+    | NReturn (tok, e) -> (
+        let tainted = check_tainted_return config fun_env in' tok e in
+        match opt_name with
+        | Some var ->
+            hashtbl_update fun_env (str_of_name var) tainted
```
IagoAbal: I would probably inline `hashtbl_update` here, it's the only place where it's used anyways.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,296 @@
...
+              (PM.Set.union tainted);
+            in'
+        | _ -> in')
```
IagoAbal: `| None -> in'`

semgrep-core/src/engine/Dataflow_tainting.mli
IagoAbal: We can just add a short comment. We have future plans for the `Pattern_match.t` (debugging mode and maybe special pretty-printing of taint matches).

semgrep-core/tests/tainting_rules/js/metavar_eq_conditional.js
```diff
@@ -0,0 +1,9 @@
+
+
+var source = cond ? get(A) : get(f(A))
```
IagoAbal: Maybe `get(f(A))` should be `get(f(B))` to make the test more relevant?

IagoAbal: Or maybe I'm not getting right what you're trying to test here?

semgrep-core/src/core/Pattern_match.ml
```diff
@@ -116,3 +116,20 @@ let no_submatches pms =
                  Hashtbl.replace tbl k (pm :: ys')));
   tbl |> Hashtbl.to_seq_values |> Seq.flat_map List.to_seq |> List.of_seq
   [@@profiling]
+
+(* Silly, but we have to because type declarations are automatically recursive so [type t = t] doesn't work *)
+type pm = t
+
+module Set = Set.Make (struct
+  type t = pm
```
IagoAbal: Try `type nonrec t = t`.

semgrep-core/tests/tainting_rules/js/metavar_eq_conditional.js
mmcqd: The idea of this test is that since the sink can either be on $X or f($X), there's two metavariable assignments that would make the sink tainted. Either $X = A or $X = f(A). I wanted to check that semgrep-core would correctly report the tainted sink twice, and it does. Though `make test` can't actually check that the double report happens I think.

semgrep-core/src/core/Pattern_match.ml
mmcqd: Yep. Updated comment to mention this.

semgrep-core/src/core/Pattern_match.ml
```diff
@@ -116,3 +116,17 @@ let no_submatches pms =
                  Hashtbl.replace tbl k (pm :: ys')));
   tbl |> Hashtbl.to_seq_values |> Seq.flat_map List.to_seq |> List.of_seq
   [@@profiling]
+
+module Set = Set.Make (struct
```
aryx, line 125: Maybe a comment like
"This special Set is used in the dataflow tainting code, which manipulate sets of matches (actually we currently care only about the metavariables in the matches) associated to each variables.".

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,291 @@
...
+(* Helpers *)
+(*****************************************************************************)
+
+let str_of_name ((s, _tok), sid) = spf "%s:%d" s sid
+
+let set_opt_to_set = function
+  | None -> PM.Set.empty
+  | Some x -> x
+
+(* @param env1 Metavariable environment.
```
aryx: meh on those @param and @return. I'd rather see a type that says the same thing.
I would keep the @ returns  comment (but would remove the leading @).

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,291 @@
...
+
+let set_filter_map f pm_set =
+  PM.Set.fold
+    (fun pm pm_set ->
+      match f pm with
+      | Some pm' -> PM.Set.add pm' pm_set
+      | None -> pm_set)
+    pm_set PM.Set.empty
+
+(* @param sink_pm Pattern match of a sink.
```
aryx, line 110: ok here the comments are not bad. They don't just paraphrase a type.

semgrep-core/tests/tainting_rules/js/metavar_eq_conditional.js
aryx: ok this comment about "the idea of this test ..." should be put in the code. It's a useful comment.

aryx: Also I think this test is too subtle, and like you say 'make test' just check if something is reported at a line, so it does
not really enforce what you think, so maybe there is a better way to write the test that would enforce what you say.

semgrep-core/tests/tainting_rules/js/metavar_eq_simple.js
```diff
@@ -0,0 +1,19 @@
+
```
aryx, line 1: Again, the test is a bit confusing because the pattern-source is get($X),
but then in the pattern-sink you actually don't care about the source (even though it's passed as an  extra parameter),
you just check for $X, which now make the reader think that $X itself is a source.

Maybe I would not pass the source1, source2, ... extra arguments to sink, they are distracting.

I wonder again if there are better tests. This looks like taint use-cases noone would need.
Why did we implement this check-metavar-equality in the first place? Do we have a simple example showing
the needs for the feature?

aryx (approved):

IagoAbal (approved):

semgrep-core/src/engine/Dataflow_tainting.ml
IagoAbal: Just to say that `env1` is a `"Metavariable environment."` I agree it's not necessary. But if you are going to say something long about a parameter, then it's better to use `@param` as this is a standard tag and helps reading more efficiently. Which is why I suggested `@param` in the first place.

A common style to doc this is also:
```ocaml
(* [unify_meta_envs env1 env1] returns [Some (union env1 env2)] if [env1] and [env2] contain no conflicting metavariable assignments, otherwise [None]. *)
```

semgrep-core/tests/tainting_rules/js/metavar_eq_simple.js
IagoAbal: IIRC this is a feature that @nbrahms pointed out that we were lacking and it made sense to me too. The behavior overall seems consistent with what we do in search mode, although I guess there may be different ways of seeing it.

It also made sense because we want to support referring to any of these metavariables in the message. Once you allow that, it also makes sense to allow referring to a metavariable bound while matching the source, and once you are there then it makes sense to do this kind of "unification" anyways. Also, most of the work was already needed simply to identify the source of taint, which is something that users have asked for.

I agree the tests are somewhat artificial. In this particular case we don't have good motivating examples for the "filtering" aspect of it. Maybe @minusworld has encountered some real rules where this is useful?

aryx: Oh I just saw the demos of matthew at the weekly-demos and now I see that we actually need to pass those source extra arguments ... Again could be nice to add comments in the test to explain those subtle things (basically what matthew said orally during his weekly-demo).

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -0,0 +1,291 @@
...
+   @param env2 Metavariable environment.
+   @returns Some (union [env1] [env2]) if [env1] and [env2] contain no conflicting metavariable assignments, otherwise None. *)
+let unify_meta_envs env1 env2 =
+  let ( let* ) = Option.bind in
+  let xs =
+    List.fold_left
+      (fun xs (mvar, mval) ->
+        match List.assoc_opt mvar env2 with
+        | None ->
+            let* xs = xs in
```
IagoAbal: Note that you can move this up instead of duplicating it in each branch.

semgrep-core/tests/tainting_rules/js/metavar_eq_conditional.js
IagoAbal: If you really want to test that you get two matches on the same sink then you will need to use `pytest` (/semgrep/tests/) and use `$X` in the message so that the findings don't get deduplicated. Otherwise the test doesn't really test what you intend to test.

semgrep-core/tests/tainting_rules/js/metavar_eq_simple.js
minusworld: This example is a bit contrived, but illustrates a case where metavariable equality would be nice. The match on `f2.write(...)` I would expect not to be matched if the metavariables were equivalent: https://semgrep.dev/s/pYg0

semgrep-core/src/engine/Dataflow_tainting.ml
mmcqd: Ok I switched to Iago's suggestion here

semgrep-core/tests/tainting_rules/js/metavar_eq_conditional.js
mmcqd: OK I made this test less subtle and explained how it's supposed to work in a comment. I'm not testing if a single sink is matched twice now. As you say the semgrep-core tests won't actually check for that, and semgrep won't even currently display both matches, so I think it's not essential to test for that right now.

semgrep-core/tests/tainting_rules/js/metavar_eq_simple.js
mmcqd: Added comments explaining the test. @minusworld's example here doesn't actually do what we want even with the PR because that `f2.write` is not between an `f2.write` and an `f2.close`, exactly as the rule specifies. If we change it to this: https://semgrep.dev/s/EvZL, it works even without the PR, because the important metavariable equality enforcement happens entirely within the sink rule. (said this on slack already but wanted to copy here in case someone reads this thread in the future).

mmcqd: Going to mark this resolved and merge, but I created an issue to come up with a less contrived test on Linear (PA-601).

Issue #4080, opened 2021-10-16
-fast skips a target that it shouldn't (bash)


mjambon:
I'm not getting matches with `-fast` for bash:

```
$ cat hello.yaml 
rules:
- id: '-'
  pattern: echo ...
  message: echo ...
  languages:
  - bash
  severity: ERROR

$ cat hello.sh
echo hello

$ semgrep-core -lang bash -config hello.yaml hello.sh
hello.sh:1 with rule -
 echo hello

$ semgrep-core -lang bash -config hello.yaml hello.sh -fast
```
^ The last command should return a match.

Here's the pattern's AST:
```
$ semgrep-core -lang bash -dump_pattern <(echo 'echo ...')
E(
  Call(
    N(
      Id(("!sh_cmd!", ()),
        {id_resolved=Ref(None); id_type=Ref(None); id_constness=Ref(None); })),
    [Arg(L(String(("echo", ())))); Arg(Ellipsis(()))]))
```

I guess the problem is due to the artificial function call `!sh_cmd!` that was introduced, and since it's not found in the raw source code, `-fast` assumes that the pattern won't match anything. Of course, `!sh_cmd!` also exists in the target AST but not in the target source. Note that other shell constructs are translated like this, using other functions of the form `!sh_...!` wherever the generic AST doesn't provide the constructs.

If that's indeed the problem, I suggest that `-fast` should ignore identifiers like `!sh_cmd!`. We'd have to agree on a prefix or pattern for this purpose.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-503/-fast-skips-a-target-that-it-shouldnt-bash.

mjambon:
Maybe we should introduce a new kind of `special` ID:
```
| Implicit of string  (* any name that doesn't occur in the source code *)
```

PR #4081, opened 2021-10-16
Add bash to the languages supported by the semgrep/python CLI

mjambon:
* This is a minimal change that supports only targets with the `.sh` or `.bash` extensions.
* I also added a debug log line for each skipped target, which was useful for debugging and something a user was requesting earlier this week. This is how I found about bug #4080.

semgrep currently won't return matches on bash programs due to #4080, which will be fixed independently.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -487,7 +487,7 @@ let filter_files_with_too_many_matches_and_transform_as_timeout matches =
                       Semgrep_core_response_t.path = file;
                       reason = Too_many_matches;
                       details;
-                      skipped_rule = Some rule_id;
+                      rule_id = Some rule_id;
                     })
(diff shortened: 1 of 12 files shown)


aryx (approved):
Looks good. Not sure what's going on with the CI failures.

mjambon:
> Not sure what's going on with the CI failures.

It's a test that expects an error message that shows the list of available languages. The list of available languages just changed.

mjambon:
The other failed test was due to the new error message in verbose mode. I just improved it to show the rule name. It looks like this now:
```
skipped 'targets/bad/invalid_python.py' [rule assert-eqeq-is-ok]: irrelevant_rule: target doesn't contain some elements required by rule 'assert-eqeq-is-ok'
```

PR #4085, opened 2021-10-18
Introduce new property 'id_hidden' [PA-509]

### LINKED ISSUE #4080: -fast skips a target that it shouldn't (bash)

mjambon:
Fixes #4080

This allows bash to work with -fast/-filter_irrelevant_rules because many hidden/fake function calls are created in the generic AST to support constructs that don't have a dedicated constructor already.

This solution adds an `id_hidden` field to the existing `id_info` record. The main advantage is that it doesn't make things more complicated except where we care about excluding hidden identifiers.

Other solutions that were considered include:
* extending the global list of exclusions implemented in `Pattern.is_special_identifier`. This isn't great because that list is global and an ID that's special in one language may not be special in another.
* adding a new `Hidden` kind of `special` identifier. Either this essentially disables the possibility of equality between a hidden and non-hidden identifier of the same name or it makes matching and other analyses more complicated.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/api/AST_generic_to_v1.ml
+++ b/semgrep-core/src/api/AST_generic_to_v1.ml
@@ -115,7 +115,12 @@ and map_resolved_name_kind = function
 
 and map_id_info x =
   match x with
-  | { G.id_resolved = v_id_resolved; id_type = v_id_type; id_constness = v3 } ->
+  | {
+   G.id_resolved = v_id_resolved;
+   id_type = v_id_type;
+   id_constness = v3;
+   id_hidden = _not_available_in_v1;
+  } ->
       let v3 = map_of_ref (map_of_option map_constness) v3 in
       let v_id_type = map_of_ref (map_of_option map_type_) v_id_type in
       let v_id_resolved =
--- a/semgrep-core/src/core/Pattern.ml
+++ b/semgrep-core/src/core/Pattern.ml
@@ -55,6 +55,8 @@ let is_js lang =
 
 (* This is used in Analyze_pattern.ml to skip
  * semgrep special identifiers.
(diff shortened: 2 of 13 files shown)


aryx (approved):
I think I preferred the Hidden in special, but maybe this is better.
Let's go with it.

PR #4093, opened 2021-10-19
feat(config_resolver): don't use pack cdn for now

brendongo:
cdn registry updating is still WIP so for now, use legacy registry

PR checklist:
- [x] Changelog is up-to-date


mmcqd (approved):
lgtm

PR #4097, opened 2021-10-20
Add a correctness check to the perf comparison

emjin:
We run benchmarks latest against an older version. While we're at it, we may as well make sure the results are correct.

The one annoyance is that if semgrep output changes, we would have to reset the benchmarks. Unfortunately, since we benchmark against a release, we can't just update a snapshot and make it pass. We _can_ mask certain parts of the semgrep output in `./run-benchmarks`, it's just a bit annoying.

The upside is that with these correctness checks, we can add a taint benchmarking pass that also serves as real world taint tests cc @IagoAbal 

I will do a separate documentation PR for all the benchmark changes

Test plan:

The script is run in two stages. First, get four different json output files by running (in `perf/`)

```
./run-benchmarks --config configs/temp.yaml --std-only --save-to bench_temp1.json
./run-benchmarks --config configs/temp.yaml --std-only --save-to bench_temp2.json
```

Next go to `../scripts/compare-perf` and comment out the call to `send_comment`, since you are running it locally, not on a PR.

Then run:

```
../scripts/compare-perf bench_temp1.json bench_temp1.json bench_temp2.json bench_temp2.json 0 0
```

The 0 0 will be ignored, since you commented out `send_comment`. We use `bench_temp1.json` twice because it's the same run, just done again to make timing more accurate.

You should get only timings and no error. If you change the results of `bench_temp1.json`, you should run into an error. The script currently assumes that the results of the first two arguments will always be the same, and ditto for the second.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -206,9 +206,9 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: ${{ matrix.python }}
-      - name: Run 0.65.0 Timing Benchmark
+      - name: Run 0.69.1 Timing Benchmark
         run: |
-          pip3 install semgrep==0.65.0
+          pip3 install semgrep==0.69.1
           semgrep --version
           python3 -m semgrep --version
           export PATH=/github/home/.local/bin:$PATH
--- a/perf/run-benchmarks
+++ b/perf/run-benchmarks
@@ -175,6 +175,18 @@ def standardize_findings(findings: dict, include_time: bool) -> Tuple[dict, dict
     return results, timings
 
 
+def jsonify_findings(findings: dict) -> dict:
+    """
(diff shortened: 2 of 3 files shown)


aryx:
What is the 'Test plan:' for this PR?
How can I test those changes?

emjin:
Converting to draft because this implementation will not work; we cannot make benchmarks pass if we change Semgrep behavior. We should instead save benchmark output to a snapshot and read from that. But then we will need to make it easy to reproduce tests, so we should wrap all of this in some sort of `run-benchmarks-test` script with a `--snapshot-update` option.

emjin:
Still hope to work on this

emjin:
I'll close this for now, but hopefully will take it up again soon-ish

PR #4109, opened 2021-10-20
Recognize Tuple/Array destructuring assignments during constant propagation

### LINKED ISSUE #4087: Constant propagation doesn't understand array destructuring
In the following code, `imageOrNull` is set.

```
function test0() {

    let imageOrNull = null;
    [imageOrNull] = image()

    if (imageOrNull == null) {
        return
    }
}
```

However, semgrep does not realize this, so `$X == $X` matches.

https://semgrep.dev/s/zxAP/

mmcqd:
Fixes #4087 
Previously an assignment like `[x,y] = f()` would not be counted as an assignment to `x` or `y` by constant propagation. Now it is recognized as such by both basic and dataflow based constant propagation.

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -336,7 +344,9 @@ and assign env lhs _tok rhs_exp eorig =
                assign env lhs_i tok1 { e = Fetch lval_i; eorig } eorig)
       in
       (* (E1, ..., En) *)
-      mk_e (Composite (CTuple, (tok1, tup_elems, tok2))) eorig
+      mk_e
+        (Composite (composite_of_container ckind, (tok1, tup_elems, tok2)))
+        eorig
```
emjin, line 354: Add a TODO for the case:

```
[a, b, ...rest] = [10, 20, 30, 40, 50];
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -169,6 +169,27 @@ let var_stats prog : var_stats =
               | AssignOp _ -> incr stat.rvalue
               | _ -> ());
               vout (E e2)
+          | Assign ({ e = Container ((Tuple | Array), (_, es, _)); _ }, _, e2)
```
emjin, line 172: TODO: ugly that we had to add this twice. cc @IagoAbal

emjin (approved):

semgrep-core/src/analyzing/Constant_propagation.ml
IagoAbal: Yes indeed, but we need to do some more work on the dataflow part before we can get rid of the "basic constant propagation" pass.

IagoAbal (approved):

PR #4120, opened 2021-10-21
feat(core_runner): increase stacklimit soft limit

### LINKED ISSUE #3623: [call for ideas] Eliminate stack overflows or troubleshoot them more easily

brendongo:
Try to increase the limit in the child fork process running
semgrep-core to prevent occurences of stack exhaustion

Note set soft limit to (hard limit / 3 etc) because for some reason
setting the soft limit to the hard limit gets a "current limit exceeds maximum limit"
error. If users are still hitting this issue we might want to look into this / increasing 
the multiplier.

Closes #3623

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -53,30 +53,32 @@ def setrlimits_preexec_fn() -> None:
     so all code here runs in a child fork before os switches to semgrep-core binary
     """
     # Get current soft and hard stack limits
-    old_s, h = resource.getrlimit(resource.RLIMIT_STACK)
-    logger.info(f"Existing stack limits: {old_s}, {h}")
+    old_soft_limit, hard_limit = resource.getrlimit(resource.RLIMIT_STACK)
+    logger.info(f"Existing stack limits: Soft: {old_soft_limit}, Hard: {hard_limit}")
 
     # Have candidates in case os unable to set certain limit
-    potential_softlimits = [
+    potential_soft_limits = [
         int(
-            h / 3
+            hard_limit / 3
         ),  # Larger fractions cause "current limit exceeds maximum limit" for unknown reason
-        int(h / 4),
+        int(hard_limit / 4),
         5120000,  # Magic number that seems to work for most cases
-        old_s,
+        old_soft_limit,
     ]
 
     # Reverse sort so maximum possible soft limit is set
-    potential_softlimits.sort(reverse=True)
-    for s in potential_softlimits:
+    potential_soft_limits.sort(reverse=True)
+    for soft_limit in potential_soft_limits:
         try:
-            logger.info(f"Trying to set soft limit to {s}")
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/core_runner.py
```diff
@@ -39,6 +40,45 @@
...
+        old_s,
+    ]
+
+    # Reverse sort so maximum possible soft limit is set
+    potential_softlimits.sort(reverse=True)
+    for s in potential_softlimits:
+        try:
+            logger.info(f"Trying to set soft limit to {s}")
+            resource.setrlimit(resource.RLIMIT_STACK, (s, h))
+            logger.info(f"Set stack limit to {s}, {h}")
```
emjin: Make this message a little better: "hard limit is {h}"

semgrep/semgrep/core_runner.py
```diff
@@ -39,6 +40,45 @@
...
+    - 5120000
+    - stack hard limit / 3
+    - stack hard limit / 4
+    - current existing soft limit
+
+    Note this is intended to run as a preexec_fn before semgrep-core in a subprocess
+    so all code here runs in a child fork before os switches to semgrep-core binary
+    """
+    # Get current soft and hard stack limits
+    old_s, h = resource.getrlimit(resource.RLIMIT_STACK)
```
emjin: Name this "hard_limit" instead? No need for it to be short

semgrep/tests/e2e/test_check.py
```diff
@@ -345,6 +345,48 @@ def test_max_memory(run_semgrep_in_tmp, snapshot):
...
+    when calling semgrep core
+    """
+
+    # long.yaml and equivalence were chosen since they happen to cause
+    # stack exhaustion
+    e2e_dir = Path(__file__).parent
+    targetpath = Path(e2e_dir / "targets").resolve() / "equivalence"
+    rulepath = Path(e2e_dir / "rules").resolve() / "long.yaml"
+
+    # If set hard and soft stack limit we should hit stack overflow
```
emjin: Suggestion "Set the hard as well as the soft stack limit. This should force a stack overflow. If this fails, the test is broken and needs to be fixed."

emjin (approved):

PR #4122, opened 2021-10-22
Refactor main so that check_rules can run semgrep

emjin:
For the metachecker, I want to be able to have files in the metachecking folder call functions like `semgrep_with_rules_file`. Unfortunately, those files currently live in Main, and thus doing so would lead to a circular dependency.

This PR creates a `runner` folder which contains the functions that run semgrep and their dependencies. I also had to move `Check_pattern` to parsing to avoid circular dependency.

PR checklist:
- [x] Changelog is up-to-date


emjin:
Not sure what I should do about the semgrep checks, seems awkward to add (* nosemgrep *) on all those lines

aryx:
You can modify semgrep.yml, see
```
    paths:
      exclude:
       - cli/*.ml
       - scripts/*
       - Test.ml
       - Matching_report.ml
       - Unit_*.ml
       - Test_*.ml

```
you can probably just add runner/*.ml

semgrep-core/src/runner/Parse_with_caching.ml
```diff
@@ -0,0 +1,145 @@
...
+  | Right exn -> raise exn
+  [@@profiling]
+
+let parse_equivalences equivalences_file =
+  match equivalences_file with
+  | "" -> []
+  | file -> Parse_equivalences.parse file
+  [@@profiling]
+
+let parse_pattern lang_pattern str =
```
aryx: Those functions don't use caching, so it's questionable to put them in this file.

aryx:
Why "For the metachecker, I want to be able to have files in the metachecking folder call functions like semgrep_with_rules_file."?
Can't you call the engine already from the metachecking directory?

aryx (approved):
Not completely convinced, but I think martin also wants to take that direction, so ok.

emjin:
> Why "For the metachecker, I want to be able to have files in the metachecking folder call functions like semgrep_with_rules_file."? Can't you call the engine already from the metachecking directory?

I guess I could have done that instead. It would have required duplicating a decent amount of code, though, which I don't prefer. This is the closest to what I wanted to do: run semgrep with the rules we have. Also, Main.ml was extremely long

IagoAbal (approved):

PR #4132, opened 2021-10-22
Check-rules runs metachecks using semgrep

emjin:
Have `-check_rules` take in a file of semgrep metachecks, which it runs on the subsequent rules. This will make it easy to implement `--check-rules` in semgrep.

Remaining TODO: output JSON when passed `--json`, be able to take in a directory for the metachecks

Test plan: make test

PR checklist:
- [x] Changelog is up-to-date


emjin:
Do not merge this until https://github.com/returntocorp/semgrep/pull/4122 is merged

aryx:
What was wrong with running semgrep with the classic --config metacheck_file.yaml /path/to/rules?
Why we need -check_rules to do that?

emjin:
> What was wrong with running semgrep with the classic --config metacheck_file.yaml /path/to/rules? Why we need -check_rules to do that?

Oh...this is what I thought we talked about doing. I still think it's nice to have both under `-check_rules`, then we can return the same kind of error for both so on the python side we don't have to do extra processing

aryx (approved):

Issue #4036, opened 2021-10-11
Correctly Tag Pypi Releases for alpine


brendongo:
We currently tag our linux pypi release with `any` platform since we want the wheel to be downloaded with alpine and manylinux platforms and there was no way at the time to tag an alpine platform.

Since then there is now a musllinux tag https://www.python.org/dev/peps/pep-0656/

We can have the wheel be uploaded as both a manylinux and a musllinux wheel and we should have the following outcomes:
- manylinux (e.g. ubuntu ) will download the manylinux wheel
- alpine will download the musllinux wheel

This frees up the `any` wheel as well and we might want to look into not publishing one so platforms we don't support (e.g. windows) will not run into the current rough edge of semgrep-core binary just not working (with a cryptic invalid binary format error) after they install.

brendongo:
If not too difficult should also look into publishing a macosx_11_0_arm64 wheel (and a macosx 11 intel wheel)

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-474/correctly-tag-pypi-releases-for-alpine.

chasemp:
Just reporting in on desire for this, currently on M1 (ARM) MacBook and pip comes with binaries for x86/AMD unfortunately.  I can use macports but the version there is quite stale.  Thanks for all you folks do.

brendongo:
The macports version is not maintained by us. Would recommend the homebrew version for now til we make time for this

Issue #2402, opened 2021-01-15
Semgrep fails to run on MacOS 11 Big Sur


brendongo:
**Describe the bug**
When running semgrep installed with pip on MacOS 11 get the following error message:
```
 [Errno 8] Exec format error: .../semgrep-core
```
**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?
macOS BigSur

brendongo:
Looks like this occurs because pypi thinks the wheel published for Mac_10_14 is incompatible with MacOS 11. This was fixed in  pip 20.3 (https://github.com/pypa/pip/issues/9138). Updating pip and reinstalling semgrep should fix this.

PR #4166, opened 2021-10-27
feat(release): correctly tag wheels

brendongo:
Add tags for musllinux, manylinux, macosx10.15, macosx11.0 instead of using the any platform tag
in pypi wheels.

Tested locally using test.pypi.org upload

Also removing all install code for spacegrep

PR checklist:
- [ ] Changelog is up-to-date


brendongo:
Closes https://github.com/returntocorp/semgrep/issues/4036 and should also prevent people from running into https://github.com/returntocorp/semgrep/issues/2402 even if they are on older pip versions

PR #4173, opened 2021-10-28
constness: Track constness of names and nothing else

IagoAbal:
When translating `x.f(y)` to IL we did not carry the constness of `x`,
so even if `x` was a constant at that point, we would not recognize it
as such in that specific context.

Initially we tracked constness in the `lval`s to later support field
sensitivity, but this is not yet implemented and just adds confusion.
For now it is pointless given that we only track the constness of
simple variables.

By making this simpler, the `x.f(y)` bug almost fixes itself.

Closes PA-588

test plan:
make test # test included

PR checklist:
- [x] Changelog is up-to-date


IagoAbal:
Oh well, fixing this bug solves a couple of `todoruleid`s in semgrep-rules, so we need to update semgrep-rules too for the tests to pass. I opened a PR for this: returntocorp/semgrep-rules#1578.

You can make Semgrep tests pass with:
```diff
--- a/semgrep/tests/e2e/test_semgrep_rules_repo.py
+++ b/semgrep/tests/e2e/test_semgrep_rules_repo.py
@@ -18,7 +18,7 @@ def _fail_subprocess_on_error(cmd):
 
 def test_semgrep_rules_repo(run_semgrep_in_tmp):
     subprocess.check_output(
-        ["git", "clone", "--depth=1", "https://github.com/returntocorp/semgrep-rules"]
+        ["git", "clone", "--depth=1", "--branch=iago/PA-588", "https://github.com/returntocorp/semgrep-rules"]
     )
 
     _fail_subprocess_on_error(
```

semgrep-core/src/core/il/IL.ml
```diff
@@ -102,7 +102,8 @@ type ident = G.ident [@@deriving show]
  * can become control-flow sensitive? (e.g., DOOP)
  *
  *)
-type name = ident * G.sid [@@deriving show]
+type name = { ident : ident; sid : G.sid; id_info : G.id_info }
```
mmcqd, line 105: We should change the comment here so it no longer mentions "the pair" since names aren't pairs anymore. Maybe "each name is guaranteed to be global and unique"

mmcqd (approved):

PR #4177, opened 2021-10-28
Translate await into IL

mmcqd:
Translate `await` into the IL for taint mode

PR checklist:
- [x] Changelog is up-to-date


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ these metavariables correctly appear in match messages
 - Constant propagation: Tuple/Array destructuring assignments now correctly prevent constant propagation
 - Scala: parse underscore separators in number literals, and parse 'l'/'L' long suffix on number literals
 - Prevent bad settings files from crashing Semgrep (#4164)
+- Taint mode can now handle await
```
IagoAbal: I would say this is not specific to taint mode, applies to constant propagation too or anything that uses the dataflow engine.

IagoAbal (approved):
I am guessing this causes new findings in semgrep-rules so you will have to check that, make a PR to semgrep-rules, and  check that the tests here pass with that PR. Then we can merge this despite the CI failures. I just got the same problem here: returntocorp/semgrep#4173

aryx (approved):
Maybe we should transform all those todo (G.E eorig) in something
that construct a Todo in the IL but at least convert the expressions inside it, so
at least we would call expr env e and get a change to match the code?

aryx:
Hmm, I also have CI errors on semgrep-rules on my PR. Maybe something is broken.

IagoAbal:
In this case at least the test for JS rule `tainted-sql-string` seems to be affected.

aryx:
yep, I just checked for mine and apparently it's a kotlin test that is now failing, so maybe it makes sense and I have a bug
(or the rule needs to be updated).

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -579,11 +579,19 @@ and expr_aux env ?(void = false) eorig =
                 ss_for_e3 @ [ mk_s (Instr (mk_i (Assign (lval, e3)) e3orig)) ]
               )));
       lvalexp
+  | G.Await (tok, e) ->
+      let lval = fresh_lval env (G.fake "await") in
```
nbrahms: Shouldn't we use `tok` instead of `G.fake`? Just want to make sure this wouldn't cause a crash if we tried to convert to a range later.

IagoAbal: True, good catch.

IagoAbal: Otherwise, you could perhaps use `Parse_info.fake_info` to get a "safe fake token". And we should make `AST_generic.fake` an alias of that one.

IagoAbal:
> Maybe we should transform all those todo (G.E eorig) in something
> that construct a Todo in the IL but at least convert the expressions inside it, so
> at least we would call expr env e and get a change to match the code?

Good idea, I'll work on a separate PR to do this.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -579,11 +579,19 @@ and expr_aux env ?(void = false) eorig =
...
+      let lvalexp = mk_e (Fetch lval) eorig in
+      add_stmt env
```
IagoAbal: You can use `add_instr` instead.

CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ these metavariables correctly appear in match messages
 - Constant propagation: Tuple/Array destructuring assignments now correctly prevent constant propagation
 - Scala: parse underscore separators in number literals, and parse 'l'/'L' long suffix on number literals
 - Prevent bad settings files from crashing Semgrep (#4164)
-- Taint mode can now handle await
+- Dataflow: translate taint into the IL
```
IagoAbal:
```
translate
- taint
+ await
into the IL
```

IagoAbal (approved):

IagoAbal:
Assuming that you checked that tests pass after updating semgrep-rules, will merge this now.

PR #4178, opened 2021-10-28
Scala: Parse by name types in function arguments

mmcqd:
We can now parse types like `(=> Int) => Int`

PR checklist:
- [x] Changelog is up-to-date


mjambon (approved):

aryx (commented):
woohoo!

aryx:
Do you know how much it improves our parsing stats for Scala?

mmcqd:
> Do you know how much it improves our parsing stats for Scala?

We went from 89.6 to 90.0, so not massive, but measurable.

PR #4179, opened 2021-10-29
bash: deep ellipsis etc.

mjambon:
* support for deep ellipsis
* simple patterns can now match any command argument rather than having to match the whole command
* no longer completely ignore redirects so as to not treat `a > b` as a simple pattern like `a`

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/parsing/ast/AST_bash.ml
```diff
@@ -375,11 +401,12 @@ and expression =
   | Ansii_c_string of (* $'...' *) string wrap
   | Concatenation of loc * expression list
   | Expr_ellipsis of (* ... *) tok
+  | Expr_deep_ellipsis of (* <... x ...> *) loc * expression bracket
```
aryx, line 404: I usually separate those semgrep constructs from the other regular bash constructs with a comment like
(* semgrep-ext: *) to indicate that those constructors are used only by semgrep.

aryx (approved):

semgrep-core/src/parsing/ast/AST_bash.ml
mjambon: What about having "semgrep" in the name? This is what I did for the tree-sitter rules but not here.

mjambon: See PR: https://github.com/returntocorp/semgrep/pull/4181

PR #4194, opened 2021-11-01
[Java] dataflow: IL: Add partial support for synchronized blocks

### LINKED ISSUE #4150: Constant propagation does not understand synchronized construct
**Describe the bug**
Constant propagation happens even if a variable is reassigned within a `synchronized` block.

**To Reproduce**
https://semgrep.dev/s/rYj0

**Expected behavior**
The returned expression in this function should not be matched.

IagoAbal:
Closes #4150

test plan:
make test # tests included

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -1036,6 +1036,10 @@ let rec stmt_aux env st =
         | _ -> (None, manager_as_pat)
       in
       python_with_stmt env manager opt_pat body
+  (* Java: synchronized (E) S *)
+  | G.OtherStmt (G.OS_Sync, [ G.E objorig; G.S stmt1 ]) ->
```
aryx, line 1040: hmm, it should be a OtherStmtWithStmt then. We should remove G.OS_Sync and move it under OtherStmtWithStmt.

aryx (approved):

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: I will open a separate PR for that.

PR #4202, opened 2021-11-02
Scala: parse case class declaration in block

mmcqd:
Improves our parsing rate from 95.2 to 96.8.
Adds parsing for case class declarations inside blocks, such as:
```
object Foo {
    def f(x : Int) : Int = {
        case class F(x : Int)
        x
    }
}
```

PR checklist:
- [x] Changelog is up-to-date


mmcqd:
Failing because we need to force-merge https://github.com/returntocorp/semgrep-rules/pull/1578

mjambon (approved):

aryx (approved):

PR #4203, opened 2021-11-03
Enable Info log level everywhere when logging is requested

mjambon:
via one of the command-line options that imply some logging. Previously, only the Main module was logging anything unless there was an override in the log config file.

This restores the logging that used to be in Main and disappeared with the creation of Run_semgrep. Additionally, now we start seeing some log messages from other modules.

Depends on https://github.com/returntocorp/pfff/pull/487

test plan: compare
```
semgrep-core -debug -lang js -dump_ast <(echo 'a a')
```
with the same without the `-debug` option:
```
semgrep-core -lang js -dump_ast <(echo 'a a')
```

image

image

PR checklist:
- [x] Changelog is up-to-date


mjambon:
@aryx Do we have to prefix the name of each logger with `Main.`? What was the idea here? I want to remove it, but it would require changing all the `config_log.json` files that people may be using. Happy to do what's necessary if it makes sense.

aryx:
Easy_logging supports hierarchy, so by prefixing all of them with Main, I could put in the log_config.json file:
```
    "loggers": 
        [
            { "name": "Main", "level": "debug",
              "handlers": [ {"cli_err": {"level" : "debug"}} ] 
            },
```
and it was applying to all modules.

aryx (approved):
Cool! Thx a lot.

emjin:
Thank you!!

PR #4208, opened 2021-11-04
Prevent infinite loop when running in --test mode

mjambon:
This prevents an infinite loop/hang when running in `--test` mode by avoiding raising an exception when attempting to create a `SemgrepCoreError` exception. It may only be a problem when running semgrep with `--test`, which is what we use to test the semgrep-rules corpus, but I'm not sure. Before, we would end up with a [GHA timeout after 6 hours](https://github.com/returntocorp/semgrep-rules/runs/4052577405?check_suite_focus=true).

PR checklist:
- [x] Changelog is up-to-date


semgrep/semgrep/error.py
```diff
@@ -88,7 +88,9 @@ class LegacySpan:
     config_path: Tuple[str]
 
 
-@attr.s(auto_attribs=True, frozen=True)
+# Don't use frozen=True because it's violated and raises an exception
+# while we're trying to create a clean SemgrepCoreError exception.
+@attr.s(auto_attribs=True)
 class SemgrepCoreError(SemgrepError):
     code: int
     level: Level
```
mjambon: ^ The real fix is this.

semgrep-core/src/core/Semgrep_error_code.ml
```diff
@@ -106,13 +106,14 @@ let exn_to_error ?(rule_id = None) file exn =
...
       let loc = Parse_info.first_loc_of_file file in
       {
         rule_id;
         typ = FatalError;
         loc;
         msg = Common.exn_to_s exn;
-        details = Some (Printexc.get_backtrace ());
+        details = Some trace;
         yaml_path = None;
       }
```
mjambon, line 118: For those interested in ocaml, what I do here is capture the backtrace as soon as possible after catching the exception. It actually would be best to capture it in the `with` part of the original try-with. The problem is that calling a function before capturing the backtrace can clear the backtrace. This happens if such a function raises and catches exceptions internally.

There are other spots where we may have similar problems and we should fix them. This one just happened to be in my way.

emjin: Ah, makes sense. I should have realized that myself. Maybe we can write a semgrep rule for this

emjin (dismissed):
Approved but Brendon should probably look at the Python change

Edit: I think the final fix will be more involved.

semgrep/semgrep/error.py
aryx: ??? What is going on here? What is this frozen thing in the first place?

emjin:
It looks like the test failures are because without `Frozen` SemgrepCoreError is now unhashable. I don't really understand why we're raising another SemgrepCoreError when we try to make a clean SemgrepCoreError. Shouldn't we try to build a Semgrep Error instead?

semgrep/semgrep/error.py
```diff
@@ -88,7 +88,9 @@ class LegacySpan:
     config_path: Tuple[str]
 
 
-@attr.s(auto_attribs=True, frozen=True)
+# Don't use frozen=True because it's violated and raises an exception
+# while we're trying to create a clean SemgrepCoreError exception.
+@attr.s(auto_attribs=True)
```
emjin: Do you have a reproduction for the bad `--test` run?

semgrep/semgrep/error.py
brendongo: Frozen is to make it hashable. Do you have a repro of this causing an infinite loop?

semgrep/semgrep/error.py
mjambon: I modified semgrep-core until it produced problematic output for semgrep. Here's the function to modify in [`Run_semgrep.ml`](https://github.com/returntocorp/semgrep/blob/20f6c1c4ce52cd9168bd6d406fe5d6e04ca5d431/semgrep-core/src/runner/Run_semgrep.ml#L561-L580):
```
let run_semgrep_with_rules config files_or_dirs =
  let rules_file = config.config_file in
  try
    logger#info "Parsing %s" rules_file;
    let timed_rules =
      Common.with_time (fun () -> Parse_rule.parse rules_file)
    in
    let res, files = semgrep_with_rules config timed_rules files_or_dirs in
    (None, res, files)
  with exn ->
    let trace = Printexc.get_backtrace () in
    logger#info "Uncaught exception: %s\n%s" (Common.exn_to_s exn) trace;
    let res =
      {
        RP.matches = [];
        errors = [ E.exn_to_error "" exn ];
        skipped = [];
        rule_profiling = None;
      }
    in
    (Some exn, res, [])
```
The output poses a problem to semgrep if (and only if?) the `errors` field is not empty. To trigger this, raise an exception to be caught by the try-with:
```ocaml
try
  assert (not true);
  ...
```
I'll try to produce a simpler, fake semgrep-core that just spits out the problematic json output.

mjambon: Unfortunately, replacing `semgrep-core` by this script doesn't cause `semgrep` to hang:
```bash
#!/usr/bin/env bash
#
# Simulate an error in semgrep-core that causes 'semgrep --test' to loop
# forever. Copy it to where semgrep-core is found normally.
# If you're working in the semgrep repo, this is:
#
#   "$(git rev-parse --show-toplevel)/semgrep/bin/semgrep-core
#
# otherwise, it falls back to the first semgrep-core found in your PATH.
#
set -eu

cat <<EOF
{"errors": [{"code": 2, "level": "error", "message": "Semgrep Core ERROR - Fatal error in file .:1\n\t\"Assert_failure src/runner/Run_semgrep.ml:564:4\"\n-----[ BEGIN error trace ]-----\nRaised at Stdlib__string.index_rec in file \"string.ml\", line 115, characters 19-34\nCalled from Sexplib0__Sexp.Printing.index_of_newline in file \"src/sexp.ml\", line 113, characters 13-47\n\n-----[ END error trace ]-----\n", "path": ".", "type": "Fatal error"}], "results": []}
EOF

exit 2
```

emjin: I tried that, but I get:
```
➜  semgrep-core git:(develop) ✗ semgrep --config tests/OTHER/rules/settings.yaml 
Running 1 rules...
Semgrep Core ERROR - Fatal error in file .:1
	"Assert_failure src/runner/Run_semgrep.ml:565:4"
-----[ BEGIN error trace ]-----
Raised at Stdlib__string.index_rec in file "string.ml", line 115, characters 19-34
Called from Sexplib0__Sexp.Printing.index_of_newline in file "src/sexp.ml", line 113, characters 13-47

-----[ END error trace ]-----
```

which is what I would expect

emjin: Ah I forgot to use semgrep --test, let me try again

emjin: Reproduced!

emjin: Failed to reproduce it working without the `frozen=True` ---> nvm

emjin: Edit: I don't think this didn't hang, but is just slow. What command did you run exactly after you made the semgrep-core change? Edit again: I clearly need to be more patient (in my defense I did not think `get_config_test_filenames` would take several minutes)

mjambon: image

where (1) `sg` runs my development version of `semgrep`,

(2) my input folder `toto` has a simple rule with a pattern that matches the target (not sure if necessary to match):

```
$ tree toto
toto
├── toto.bash
└── toto.yaml

0 directories, 2 files

$ cat toto/toto.yaml 
rules:
- id: toto
  message: yo
  languages:
    - bash
  pattern: x
  severity: WARNING

$ cat toto/toto.bash 
x
```

and (3) my semgrep-core contains an `assert (not true);` placed as follows. This is the diff from the `develop` branch:
```
$ git diff
diff --git a/semgrep-core/src/runner/Run_semgrep.ml b/semgrep-core/src/runner/Run_semgrep.ml
index 46b62906..4985ea8c 100644
--- a/semgrep-core/src/runner/Run_semgrep.ml
+++ b/semgrep-core/src/runner/Run_semgrep.ml
@@ -561,6 +561,7 @@ let semgrep_with_rules config (rules, rule_parse_time) files_or_dirs =
 let run_semgrep_with_rules config files_or_dirs =
   let rules_file = config.config_file in
   try
+    assert (not true);
     logger#info "Parsing %s" rules_file;
     let timed_rules =
       Common.with_time (fun () -> Parse_rule.parse rules_file)
```

mjambon:
thanks @emjin !

semgrep/semgrep/error.py
```diff
@@ -90,7 +90,7 @@ class LegacySpan:
 
 # Don't use frozen=True because it's violated and raises an exception
 # while we're trying to create a clean SemgrepCoreError exception.
```
nbrahms: Should we also remove this comment?

PR #4212, opened 2021-11-04
Scala: parse unicode characters in identifiers

mmcqd:
Improves parse rate from 97.2 to 97.8.
We can now parse unicode identifiers.

Test plan: make test (see added test in pfff)

PR checklist:
- [x] Changelog is up-to-date


mmcqd:
CI failure should be fixed by this: https://github.com/returntocorp/semgrep/pull/4216

aryx (approved):

PR #4218, opened 2021-11-05
engine: Make metavariable-comparison work on constant variables

### LINKED ISSUE #3727: Have `metavariable-comparison` work on variables, not just literals

Currently, metavariable-comparison only works on literals (`5`, `"cat"`).

Sometimes, however, it'd be useful to operate on variables, using whatever constant propagation is able to tell us about them.

Something like:

```yaml
rules:
- id: c-comparing-metavars
  patterns:
    - pattern: |
        int $X = $XVAL;
        int $Y = $YVAL;
    - metavariable-comparison:
      metavariable: $XVAL
      comparison: $XVAL < $YVAL  
  message: Trying to compare metavars
  languages: [c]
  severity: WARNING
```

```c
int main() {
    int x = 3;
    int y = 5;
    if (x < y) {
        print("less");
    }
}
```

Ideally metavariable-comparison just "does the right thing," regardless of if the matched metavariable is a literal or variable.

This would allow broader, automatic matching of rules like this, which would currently not work if a non literal was used: [use_of_weak_rsa_key.yaml](https://github.com/returntocorp/semgrep-rules/blob/2b32e459b0e9baf8f41ab881d0bdf6cf9a72799f/go/lang/security/audit/crypto/use_of_weak_rsa_key.yaml
).

I suppose `pattern-where-python` could be used here, but that's been deprecated.

**Additional context**

There's been one or two customer asks for functionality like this, like being able to compare array bounds in C, or using `metavariable-` patterns to reason about the string values that constant propagation can tell us.

More discussion and context: https://returntocorp.slack.com/archives/CHG013EPJ/p1629233756164100?thread_ts=1629203004.157800&cid=CHG013EPJ

IagoAbal:
Closes #3727

test plan:
make test # test included

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/tests/OTHER/rules/metavar_comparison_constness.yaml
+++ b/semgrep-core/tests/OTHER/rules/metavar_comparison_constness.yaml
@@ -1,0 +1,12 @@
+rules:
+  - id: c-comparing-const-vars
+    languages:
+      - c
+    message: Trying to compare constant variables
+    patterns:
+      - pattern: |
+          $X < $Y
+      - metavariable-comparison:
+          comparison: $X < $Y
+          metavariable: $X
+    severity: WARNING
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Constant propagation: In a method call `x.f(y)`, if `x` is a constant then it will be recognized as such
 - Scala: parse `case object` within blocks
 - Go: match correctly braces in composite literals for autofix (#4210)
+- `metavariable-comparison`: if a metavariable binds to a code variable that
+  is known to be constant, then we use that constant value in the comparison (#3727)
+
+### Fixed
+- Constant propagation: In a method call `x.f(y)`, if `x` is a constant then
+  it will be recognized as such
 - Scala: parse typed patterns with variables that begin with an underscore: `case _x : Int => ...`
 - Scala: parse unicode identifiers
 - semgrep-core accepts `sh` as an alias for bash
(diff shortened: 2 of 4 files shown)


semgrep-core/tests/OTHER/rules/metavar_comparison_constness.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - id: c-comparing-const-vars
+    languages:
+      - c
+    message: Trying to compare constant variables
+    patterns:
+      - pattern: |
+          $X < $Y
+      - metavariable-comparison:
+          comparison: $X < $Y
+          metavariable: $X
```
aryx, line 11: this makes the metavariable field a bit useless now, if you can use multiple metavars in comparison.

aryx (approved):
LGTM. You need to rerun the pre-commit in CI manually though. It seems blocked
(or maybe because you have a conflict and CI refuses to kick in?)

emjin (approved):
lgtm

emjin:
> LGTM. You need to rerun the pre-commit in CI manually though. It seems blocked (or maybe because you have a conflict and CI refuses to kick in?)

It's blocked because of the conflict

semgrep-core/tests/OTHER/rules/metavar_comparison_constness.yaml
IagoAbal: We already allowed referencing multiple variables. And indeed I think the only use of `metavariable` here is when we have `strip: true` (the default is `false`) in which case we rewrite `$X` as `int($X)`... and this is probably something the user should write manually. If we get rid of `strip` then we can get rid of `metavariable`.

IagoAbal: Without having to break any rules, we could make `metavariable` optional and require it only when `strip: true`.

emjin: +1 to having the user write `int($X)` manually. It would be extremely frustrating to want to do `int($X) < int($Y)` and not

Issue #3387, opened 2021-06-18
[Autofix] sometimes drops closing bracket in python slices


JDegesysHelm:
**Describe the bug**
The closing bracket is sometimes dropped in the autofix.

**To Reproduce**
https://semgrep.dev/s/juliusdegesys:closingbracketbugrepro
```yaml
rules:
- id: test
  message: Bug Repro
  severity: INFO
  pattern: f($X)
  fix: f($X, 1)
```
Input file:
```python
# These are ok:
f(x[0])
f([0, 1])
f(x(0, 1))
# Closing square bracket of $X gets dropped in autofix below:
f(x[0, 1])
f(x[0, 1, 2])
f(x[0, 1:2])
f(x['a', 'b'])
```
Output:
```
severity:info rule:test: Bug Repro
autofix: f(x[0], 1)
2:f(x[0])
--------------------------------------------------------------------------------
autofix: f([0, 1], 1)
3:f([0, 1])
--------------------------------------------------------------------------------
autofix: f(x(0, 1), 1)
4:f(x(0, 1))
--------------------------------------------------------------------------------
autofix: f(x[0, 1, 1)
6:f(x[0, 1])
--------------------------------------------------------------------------------
autofix: f(x[0, 1, 2, 1)
7:f(x[0, 1, 2])
--------------------------------------------------------------------------------
autofix: f(x[0, 1:2, 1)
8:f(x[0, 1:2])
--------------------------------------------------------------------------------
autofix: f(x['a', 'b', 1)
9:f(x['a', 'b'])
ran 1 rules on 1 files: 7 findings
```

**Expected behavior**
Closing bracket not to be dropped in autofix.
```
severity:info rule:test: Bug Repro
autofix: f(x[0], 1)
2:f(x[0])
--------------------------------------------------------------------------------
autofix: f([0, 1], 1)
3:f([0, 1])
--------------------------------------------------------------------------------
autofix: f(x(0, 1), 1)
4:f(x(0, 1))
--------------------------------------------------------------------------------
autofix: f(x[0, 1], 1)
6:f(x[0, 1])
--------------------------------------------------------------------------------
autofix: f(x[0, 1, 2], 1)
7:f(x[0, 1, 2])
--------------------------------------------------------------------------------
autofix: f(x[0, 1:2], 1)
8:f(x[0, 1:2])
--------------------------------------------------------------------------------
autofix: f(x['a', 'b'], 1)
9:f(x['a', 'b'])
[...]

Issue #2051, opened 2020-11-15
Autofix loses fidelity by adding and subtracting characters


msorens:
**Describe the bug**
Two errors and one conundrum in the attached sandbox:

### The errors
(1) Lost a space from both matches.
(2) Gained an extra closing parenthesis--but only from one of the two matches!

### The conundrum
I understand why my two fragments of code are semantically equivalent and matched by the pattern. However, I was actually trying to write a rule to match just the first one; i.e., I only want to match the explicit `return`, so I could then use autofix to convert to the form of the second one. Can that be done?

<img width="1028" alt="image" src="https://user-images.githubusercontent.com/6817500/99196846-b18c8300-2743-11eb-8358-20f3c0714ff8.png">

**To Reproduce**
https://semgrep.dev/s/GRPq/

**Expected behavior**
Autofix should maintain fidelity.

**What is the priority of the bug to you?**
Is this a P0 (blocking your adoption of Semgrep or workflow), P1 (important to fix or quite annoying), P2 (regular bug that should get fixed)?
P2

**Environment**
v0.31.0 and v0.31.1

brendongo:
@mjambon I'm not too sure but I think our parsed typescript code loses the parenthesis around the returned dictionary on line 2? Is something like this a job for spacegrep?

mjambon:
reopening

emjin:
Reopening--I do mean to take a look at this.

nmote:
The autofix part of this has been fixed by #7686. Autofix should no longer mess up parentheses. I'm closing this because the autofix misbehavior seemed to be the primary issue here. However, feel free to open another issue about the matching equivalence between the two arrow function styles if that is still important to you.

PR #4219, opened 2021-11-05
[Go] do not drop closing paren in Cast in autofix

aryx:
This will help #3387

test plan:
```
$ yy -full_token_info -lang go -dump_ast tests/go/misc_cast_autofix.go
...
ParenExpr(
                                 ({
                                   token=OriginTok(
                                           {str="("; charpos=131; line=5;
                                            column=38;
                                            file="tests/go/misc_cast_autofix.go";
                                            });
                                   transfo=NoTransfo; },
                                  N(
                                    Id(
                                      ("data",
                                       {
                                        token=OriginTok(
                                                {str="data"; charpos=132;
                                                 line=5; column=39;
                                                 file="tests/go/misc_cast_autofix.go";
                                                 });
                                        transfo=NoTransfo; }),
                                      {id_hidden=false;
                                       id_resolved=Ref(None);
                                       id_type=Ref(None);
                                       id_constness=Ref(None); })),
                                  {
                                   token=OriginTok(
                                           {str=")"; charpos=136; line=5;
                                            column=43;
                                            file="tests/go/misc_cast_autofix.go";
                                            });
                                   transfo=NoTransfo; }))));
...
```

So the closing paren is now in the AST so the metavariable will have
a correct range and the autofix will work better

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/core/ast/AST_generic.ml
+++ b/semgrep-core/src/core/ast/AST_generic.ml
@@ -476,6 +476,13 @@ and expr_kind =
   (* less: could be in Special, but pretty important so I've lifted them here*)
   | Ref of tok (* &, address of *) * expr
   | DeRef of tok (* '*' in C, '!' or '<-' in OCaml, ^ in Reason *) * expr
+  (* In some rare cases, we need to keep the parenthesis around an expression
+   * otherwise in autofix semgrep could produce incorrect code. For example,
+   * in Go a cast int(3.0) requires the parenthesis.
+   * alt: change cast to take a expr bracket, but this is used only for Go.
+   * Note that this data structure is really becoming more a CST than an AST.
+   *)
+  | ParenExpr of expr bracket
   (* sgrep: ... in expressions, args, stmts, items, and fields
    * (and unfortunately also in types in Python) *)
   | Ellipsis of tok (* '...' *)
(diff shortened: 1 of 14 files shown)


semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -476,6 +476,13 @@ and expr_kind =
   (* less: could be in Special, but pretty important so I've lifted them here*)
   | Ref of tok (* &, address of *) * expr
   | DeRef of tok (* '*' in C, '!' or '<-' in OCaml, ^ in Reason *) * expr
+  (* In some rare cases, we need to keep the parenthesis around an expression
+   * otherwise in autofix semgrep could produce incorrect code. For example,
+   * in Go a cast int(3.0) requires the parenthesis.
+   * alt: change cast to take a expr bracket, but this is used only for Go.
+   * Note that this data structure is really becoming more a CST than an AST.
+   *)
+  | ParenExpr of expr bracket
```
emjin, line 485: I don't know how I feel about adding this. Right now it seems like we aren't going into the direction of saving useless parentheses? (e.g. `((x + 1)) + 2`) But adding `ParenExpr` instead of adding tokens to `Cast` seems to imply we want to go that direction. 

The generic AST is already kind of a CST, so maybe that is the direction we want to go for autofix, but I'm not convinced yet. It seems to me like our problem with the "string replace" method of autofix is that we rely on storing tokens for expression ranges (instead of saving a range with each AST element). Not that I think that's necessarily the correct approach, either. But I don't think we've made a decision

For now, it looks like adding tokens to `Cast` would also solve this problem and adhere closer to our current philosophy? Though I guess it's annoying because languages like C put the parentheses around the type.

If we do go on with adding a `ParenExpr`, this issue also needs it: https://github.com/returntocorp/semgrep/issues/2051

aryx: I totally agree with you. I am not super convinced either. I think of it as an experiment ...

PR #4222, opened 2021-11-05
Add a test for hang on semgrep --test

emjin:
PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/tests/e2e/snapshots/test_cli_test/test_timeout/results.json
+++ b/semgrep/tests/e2e/snapshots/test_cli_test/test_timeout/results.json
@@ -1,0 +1,7 @@
+{
+  "config_missing_tests": [
+    "rules/cli_test/error/long.yaml"
+  ],
+  "config_with_errors": [],
+  "results": {}
+}
--- a/semgrep/tests/e2e/test_cli_test.py
+++ b/semgrep/tests/e2e/test_cli_test.py
@@ -8,6 +8,20 @@ def test_cli_test_basic(run_semgrep_in_tmp, snapshot):
         target_name="cli_test/basic/",
         output_format=OutputFormat.JSON,
     )
+
+    snapshot.assert_match(
+        results,
+        "results.json",
+    )
+
+
+def test_timeout(run_semgrep_in_tmp, snapshot):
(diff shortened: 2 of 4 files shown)


semgrep/tests/e2e/test_cli_test.py
```diff
@@ -8,6 +8,20 @@ def test_cli_test_basic(run_semgrep_in_tmp, snapshot):
...
+def test_timeout(run_semgrep_in_tmp, snapshot):
+    results, _ = run_semgrep_in_tmp(
+        "rules/cli_test/error/",
+        options=["--test"],
+        target_name="cli_test/error/",
+        output_format=OutputFormat.JSON,
+    )
     snapshot.assert_match(
         results,
         "results.json",
```
mjambon, line 27: Maybe reduce the timeout to speed up the tests?
```
    --timeout INTEGER             Maximum time to spend running a rule on a
                                  single file in seconds. If set to 0 will not
                                  have time limit. Defaults to 30 s.
```

mjambon (approved):

semgrep/tests/e2e/test_cli_test.py
emjin: hm I don't think `--timeout` affects `--test`

emjin: Anyway I think it stack overflows pretty quickly

semgrep/tests/e2e/snapshots/test_cli_test/test_timeout/results.json
```diff
@@ -0,0 +1,7 @@
+{
+  "config_missing_tests": [
+    "rules/cli_test/error/long.yaml"
```
aryx, line 3: Why a timeout on a file should generate a config_missing_tests in the JSON??
In fact the rules/cli_test/error/long.py file is useless, because it's in the rules/ directory, and run_semgrep_in_tmp
will take target relative to e2e/targets/ ...

This test never worked as intended.

aryx: I've fixed it in https://github.com/semgrep/semgrep/pull/10189

PR #4224, opened 2021-11-05
ci(fix): pin pipenv for benchmark tests

brendongo:
Similar to #4223

PR checklist:
- [x] Changelog is up-to-date


colleend (approved):
nice!

PR #4227, opened 2021-11-08
AST_to_IL: Translate more Generic expressions to the IL

IagoAbal:
Plus, allow for "partial" translations in order to help removing some
FP/FN when we encounter unsupported constructs.

Helps #PA-311

test plan:
make test # tests included

PR checklist:
- [x] Changelog is up-to-date


mmcqd (approved):
Nice, looks good to me

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -126,6 +126,12 @@ let var_of_id_info id id_info =
   in
   { ident = id; sid; id_info }
 
+let var_of_name name =
+  match name with
+  | G.Id (id, id_info) -> var_of_id_info id id_info
+  | G.IdQualified { G.name_last = id, _typeargsTODO; name_info = id_info; _ } ->
```
aryx, line 132: FYI this type changed quite a lot with the introduction of C++. There is now a name_middle, and name_top
field.

aryx (approved):
Very nice!

PR #4230, opened 2021-11-08
remove severity from text output

ievans:
Replacement for https://github.com/returntocorp/semgrep/pull/4040

PR checklist:
- [x] Changelog is up-to-date


emjin (approved):
Change looks good. I haven't been following the discussion. What is the reason we decided to remove severity from the output? Where is severity now used?

ievans:
AFAICT severity in the textual output is actively harmful: (1) there are no options to filter by it, which is what you would want if you were trying to find all “high” sev for instance and (2) users sometimes interpret it as semgrep-core error/warning/info

Issue #4234, opened 2021-11-09
Kotlin metavariable equality statements


colleend:
**Describe the bug**
Kotlin metavariable equality statements (in if-else statements) do not match on expected code snippets.

**To Reproduce**
semgrep.dev currently isn't working for Kotlin (keeps giving wrong matches). What I'm running is: 

_sgrep file:_
```
if ($E) {
  $S
} else {
  $S
}
```

_kotlin file:_
```
class Foo {
    fun foo() {
        // ERROR:
        if (x > 2) {
            foo()
            bar()
        } else {
            foo()
            bar()
        }

        if (x > 2) {
            foo()
            bar()
        } else {
            foo()
        }
    }
}
```

**Expected behavior**
I would expect there to be a match on the first if-else statement, but there are currently no matches for this.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [X] P2: regular bug that should get fixed

aryx:
I think it's because $S matches a single statement. You put those $S inside { } so it will only match blocks with one statement.
But you can get $S to match a whole block (which is a single statemnet) with
```
if ($E)
  $S
else
 $S
```

To double check.

colleend:
I see, that makes sense. hmm, changed it to that and still getting no matches 🤔 ,  but 
```
if ($E)
  $S
else $S
```
works. This might be because of the newline parsing issue with a newline after the `else` in `if_expressions` 🤔  (so might be fixed after your PRs to tree-sitter!)

aryx:
Maybe. I'll check!

colleend:
I'm going to close this issue then because it works (and newline is separate issue :D)

PR #4235, opened 2021-11-09
beta tests for kotlin

colleend:
Adding all the `semgrep-core` tests necessary for beta!

* all tests #justworked except for `metavar_equality_stmt`, which i filed this bug: https://github.com/returntocorp/semgrep/issues/4234 for.


brendongo (approved):

semgrep-core/tests/kotlin/metavar_equality_stmt.sgrep
```diff
@@ -0,0 +1,3 @@
+if ($E)
```
aryx, line 1: so it works then?

colleend: yep!

aryx:
Oops, those tests were actually never executed. We forgot code in  tests/Test.ml to run those tests.

aryx:
I actually got some regressions today spotted by semgrep-rules, but not by semgrep-core tests because they were never executed.

colleend:
🤦 totally forgot about that, thanks for adding it in!

PR #4236, opened 2021-11-09
Scala: parse infix type operators with tuple arguments

mmcqd:
Increases parse rate from 98.5% (with https://github.com/returntocorp/semgrep/pull/4231) to 98.9%.
Handles types like this, where `-|` is an infix type operator.
```
(S, *) -| (S => *)
```

test plan: make test (see added test here and in pfff)
PR checklist:
- [x] Changelog is up-to-date


aryx (approved):
Woohoo! almost there! 0.1% remaining

aryx:
hmm conflict because of pfff. Will need to merge develop first.

PR #4237, opened 2021-11-09
fix(test): dont treat todoruleid as false positive

### LINKED ISSUE #4204: testing: Ignore matches or the lack of matches on "todook:" and "todoruleid:" locations
Goal: tolerate false positives and false negatives that are marked as such, so as to allow running a version semgrep on an incompatible version of semgrep-rules.

The `--test-ignore-todo` looks like it should provide just this, but it doesn't work. We still get an exit code of 1.

Repro setup
--

Target file `toto.py`:
```js
# todoruleid:toto
x
```

Rule file, `toto.yaml`:
```yaml
rules:
- id: toto
  message: yo
  languages:
    - python
  pattern: x
  severity: WARNING
```

Plain `--test`
--

Output obtained with plain `--test`:
```
$ semgrep --test .
Running 1 rules...
found false positives on ok'ed lines - file=/home/martin/tmp/todo/toto/toto.py fps={2}
failing due to false positives
$ echo $?
1
```

A failure was expected, so this is fine. I don't know why it's aborting rather than printing a confusion matrix, but it's fine.

Attempt to ignore false positive with `--test-ignore-todo`
--

```
$ semgrep --test --test-ignore-todo .
Running 1 rules...
1 yaml files tested
check id scoring:
(TODO: 1) toto.yaml
	✖ toto                                                         TP: 0 TN: 0 FP: 1 FN: 0
	test: /home/martin/tmp/todo/toto/toto.py, expected lines: [], reported lines: [2]
final confusion matrix: TP: 0 TN: 0 FP: 1 FN: 0
$ echo $?
1
```

I was expecting a success (0) but instead got an error status (1). I believe this is the problem that causes CI errors like this one: 
https://github.com/returntocorp/semgrep-rules/runs/4083075591?check_suite_focus=true

brendongo:
When passing --test-ignore-todo dont return non zero exit code
when a rule fires on a line tagged with todoruleid

Fixes https://github.com/returntocorp/semgrep/issues/4204

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/test.py
+++ b/semgrep/semgrep/test.py
@@ -146,6 +146,9 @@ def score_output_json(
     todo_ok_lines: Dict[str, Dict[str, List[int]]] = collections.defaultdict(
         lambda: collections.defaultdict(list)
     )
+    todo_ruleid_lines: Dict[str, List[int]] = collections.defaultdict(
+        lambda: collections.defaultdict(list)
+    )
     score_by_checkid: Dict[str, List[int]] = collections.defaultdict(
         lambda: [0, 0, 0, 0]
     )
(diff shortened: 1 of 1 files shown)


aryx (approved):

aryx:
I usually like to have a Test plan: section in my PR where I describe a precise command to test the change
and what we should look for in the output. Sometimes it's as simple as
"test file included, just run 'make test'"
but sometimes if it's not covered by a regression tests, I put a semgrep-core command and its output.

PR #4241, opened 2021-11-09
ast: Fix generic representation of synchronize blocks

IagoAbal:
It's assumed that an `OtherStmt` does not have any nested statements
as arguments, for that we have `OtherStmtWithStmt`.

test plan:
make test

PR checklist:
- [x] Changelog is up-to-date


aryx (approved):
Yes! Thanks for cleaning up tech debt.

PR #4242, opened 2021-11-09
[JS] new options `flddef_assign` and `arrow_is_function`

### LINKED ISSUE #4187: Matching JavaScript arrow functions defined on class
**Describe the bug**
There doesn't appear to be a way to match exclusively against arrow functions defined on a class. It's only possible to match against all function definitions, and to match against local arrow functions.

**To Reproduce**
It's possible to detect all function definitions: https://semgrep.dev/s/Jxrj
It's also possible to detect local arrow functions: https://semgrep.dev/s/Pb5p

**Expected behavior**
I'd like to be able to match against arrow functions that are defined on a class, without including standard functions.

This is a P1 for us. We walk the prototype of several classes and wrap all functions in a permission check. We recently discovered that arrow functions do not exist on the prototype and thus aren't subject to this permission check. We now want to prevent arrow functions from being defined on the class.

aryx:
This closes #4187

test plan:
test file included

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Java: Add partial support for `synchronized` blocks in the dataflow IL (#4150)
 - Dataflow: Add partial support for `yield`, `&`, and other expressions
 - Field-definition-as-assignemnt equivalence that allows matching expression
-  patterns against field definitions. It disabled by default but can be
+  patterns against field definitions. It is disabled by default but can be
   enabled via rule `options:` with  `flddef_assign: true` (#4187)
 - Arrows (a.k.a short lambdas) patterns used to match also regular function
   definitions. This can now be disabled via rule `options:` with
   `arrow_is_function: false` (#4187)
+- Javascript variable patterns using the 'var' keyword used to also
+  match variable declarations using 'let' or 'const'. This can now be
+  disabled via rule `options:` with `let_is_var: false`
 
 ### Fixed
 - Constant propagation: In a method call `x.f(y)`, if `x` is a constant then 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Added
 - Java: Add partial support for `synchronized` blocks in the dataflow IL (#4150)
 - Dataflow: Add partial support for `yield`, `&`, and other expressions
+- Field-definition-as-assignemnt equivalence that allows matching expression
+  patterns against field definitions. It disabled by default but can be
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Added
 - Java: Add partial support for `synchronized` blocks in the dataflow IL (#4150)
 - Dataflow: Add partial support for `yield`, `&`, and other expressions
+- Field-definition-as-assignemnt equivalence that allows matching expression
+  patterns against field definitions. It disabled by default but can be
```
IagoAbal:
```
  It
+ is
  disabled by default
```

semgrep-core/src/core/Config_semgrep.atd
```diff
@@ -47,9 +44,18 @@ type t = {
...
+   * (e.g. `class Foo {  x = function() { return; } }`).
+   *)
+  ~flddef_assign <ocaml default="false"> : bool;
   (* expression patterns (e.g. `f($X)`) will match attributes (e.g. `@f(a)`)  *)
   ~attr_expr <ocaml default="true"> : bool;
 
+  (* treat arrows (a.k.a short lambdas) like other functions *)
+  ~arrow_is_function <ocaml default="true">: bool;
+  (* treat let/const as equivalent to var (JS-specific) *)
+  ~let_is_var <ocaml default="true">: bool;
```
IagoAbal, line 57: Not mentioned in the changelog, intentionally?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1642,7 +1642,8 @@ and m_wildcard (a1, a2) (b1, b2) =
 and m_keyword_attribute a b =
   match (a, b) with
   (* equivalent: quite JS-specific *)
-  | G.Var, (G.Var | G.Let | G.Const) -> return ()
+  | G.Var, (G.Var | G.Let | G.Const) ->
```
IagoAbal: If I'm not mistaken, if you have `a = G.Var` and `b = G.Var` and `let_is_var` is disabled, this will fail.
It should be:
```ocaml
| G.Var, G.Var -> return ()
| G.Var, (G.Let | G.Const) -> if_config ...
```

semgrep-core/src/core/Config_semgrep.atd
aryx: It was an old TODO not connected to any issues.
I can add it.

IagoAbal (approved):

semgrep-core/src/matching/Generic_vs_generic.ml
aryx: oh good call!

PR #4243, opened 2021-11-09
analyzing: cfg: Fix `try` blocks ending in `return`

### LINKED ISSUE #4091: Constant propagation happening mistakenly after try/catch
**Describe the bug**
When a variable is reassigned inside of a catch block, which follows a try block ending in returning a statement which might throw an exception, which itself is inside some other block (while,if,etc.), constant propagation mistakenly thinks the variable is never reassigned, and this causes pattern like `$X == $X` to be matched when they should not.

**To Reproduce**
https://semgrep.dev/s/w35B

**Expected behavior**
The returned expression in the first function (`bad`) should not match the pattern `$X == $X`

IagoAbal:
When building the CFG of a `try-catch` we connect the end of the
try-block to the beginning of the catch-block. But, if the CFG of the
try-block ended in `return`, this did not happen, and the catch-block
became dead code, despite something in the try-block could throw an
exception.

Closes #4091

test plan:
make test # tests included

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/tests/java/cp_try_return.java
```diff
@@ -0,0 +1,44 @@
...
+            catch (Throwable t) {
+                // dead code
+                throwable = t;
+            }
+        }
+        //ERROR:
+        return throwable == null;
+    }
+
+   public boolean test3() {
```
aryx, line 40: seems quite similar to test1
Not sure what are the consequences of the extra while(cond) in test1() on the CFG problem of this PR.

aryx (approved):

aryx:
CI issues seems unrelated.
If they are, then you should post to #friction and say those pytest metrics CI failures are annoying.

emjin:
Oh yes please complain about this test it's so annoying. (I believe it makes a network call, which we should instead mock)

semgrep-core/tests/java/cp_try_return.java
IagoAbal: I added some comments to the test case to explain what we are trying to test in each case. But `test3` in particular is just there because it was in the bug report, it is not particularly useful in this case. I added a more meaningful test based on the same pattern in tainting_rules/try_return.

PR #4249, opened 2021-11-10
[AST generic] remove TyBuiltin

aryx:
It's redundant with TyN (Id). Actually there was places
where we were using TyBuiltin and other TyN (Id), so better
to be consistent.

test plan:
$ /home/pad/yy/_build/default/src/cli/Main.exe -diff_pfff_tree_sitter
tests/cpp/parsing/char.cpp

does not show anymore meaningful differences.

PR checklist:
- [x] Changelog is up-to-date


emjin:
Great refactor!

PR #4257, opened 2021-11-11
CLI: fail immediately if any configs are invalid

### LINKED ISSUE #4229: Inconsistent error behavior for configs
If we run semgrep on a folder where one of the configs has an error, we will abort and exit with 7. However, if we run semgrep with those configs individually, we will still run the working config and exit with 4. We should make these behave in the same way.

Note: gitlab would prefer we always fail fast on configs.

The configs tested with:

With error:
```
➜  testing cat error_configs/error_second.yaml 
rules:
- id: unsafe-exec
  pattern: exec(...);
  message: Avoid use of exec; it can lead to a remote code execution.
  languages: [js]
  severity: WARNING
- id: boto3-internal-network
  patterns:
    - pattern-inside: boto3.client(host="...")
    - pattern-regex: '192.168\.\d{1,3}\.\d{1,3}'
  severity: ERROR
```

Without error:
```
➜  testing cat error_configs/no_errors.yaml    
rules:
- id: use-sys-exit
  pattern: exit($X)
  message: |
      Use "sys.exit" over the python shell "exit" built-in. "exit" is a helper
      for the interactive shell and is not be available on all Python implementations.
      https://stackoverflow.com/a/6501134
  languages: [python]
  severity: WARNING
  fix: sys.exit($X)
```

These were the only two files in the folder

```
➜  testing ls error_configs 
error_second.yaml no_errors.yaml
```

Running on the folder:

```
➜  testing semgrep --config error_configs 
semgrep error: Invalid rule schema
  --> error_configs/error_second.yaml:7
7  | - id: boto3-internal-network
8  |   patterns:
9  |     - pattern-inside: boto3.client(host="...")
10 |     - pattern-regex: '192.168\.\d{1,3}\.\d{1,3}'
11 |   severity: ERROR

One of these properties may be invalid: 'patterns'
One of these properties is missing: 'languages', 'message', 'mode'

no valid configuration file found (1 configs were invalid)
➜  testing echo $?                       
7
```

Running with them individually

```
➜  testing semgrep --config error_configs/error_second.yaml --config error_configs/no_errors.yaml
semgrep error: Invalid rule schema
  --> error_configs/error_second.yaml:7
7  | - id: boto3-internal-network
8  |   patterns:
9  |     - pattern-inside: boto3.client(host="...")
10 |     - pattern-regex: '192.168\.\d{1,3}\.\d{1,3}'
11 |   severity: ERROR

One of these properties may be invalid: 'patterns'
One of these properties is missing: 'languages', 'message', 'mode'

Running 1 rules...
python_code.py
severity:warning rule:error_configs.use-sys-exit: Use "sys.exit" over the python shell "exit" built-in. "exit" is a helper
for the interactive shell and is not be available on all Python implementations.
https://stackoverflow.com/a/6501134

autofix: sys.exit(3)
9:    exit(3)
autofix: sys.exit(4)
14:        exit(4)
ran 1 rules on 2 files: 2 findings
➜  testing echo $?
4
```

mmcqd:
Fixes https://github.com/returntocorp/semgrep/issues/4229
Now if at least one config provided to semgrep is invalid, we will fail without running any configs, and exit with 7, which is consistent with behavior when providing a directory as a config.

test plan: make test (all tests still pass)
PR checklist:
- [x] Changelog is up-to-date


semgrep/semgrep/semgrep_main.py
```diff
@@ -156,19 +156,17 @@ def main(
         logger.verbose(
             f"running {len(filtered_rules)} rules from {len(configs_obj.valid)} config{plural} {config_id_if_single} {invalid_msg}".strip()
         )
-
+        if len(errors) > 0:
```
emjin, line 159: Need a test plan, including to make sure that we aren't failing fast on targets

emjin: Actually it's probably worth adding a unit test

emjin (approved):

Issue #4258, opened 2021-11-11
Semgrep 0.72.0 tries to analyze .png files and crashes


IagoAbal:
**Describe the bug**
Semgrep 0.72.0 tries to analyze .png and other binary files and then crashes with `Pcre.Error(BadUTF8)`:
https://r2c-community.slack.com/archives/C0242656NQ2/p1636567860034600. Affects paying customers.

**To Reproduce**
This can be reproduced on this public repo https://github.com/RecordReplay/devtools/, see https://github.com/RecordReplay/devtools/runs/4169052484?check_suite_focus=true#step:4:19

**What is the priority of the bug to you?**

- [x] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

IagoAbal:
cc @mjambon

IagoAbal:
cc @brendongo @mmcqd

jayvdb:
I encountered this on 0.72 with zips, tar.gz, .eot, .ttf, .woff(2).   I was able to workaround it with both `.semgrepignore` and the semgrep app ignore overrides.

IagoAbal:
This happens using the `regex` language.

PR #4264, opened 2021-11-11
Catch and log PCRE errors via wrappers in Pcre_settings

mjambon:
I tried to cover all the cases where we run pcre to obtain matches. We should use the functions in `Pcre_settings` from now on.

The compilation of a regexp (`Pcre_settings.regexp`) may still raise an exception but it's less likely to fail and I think it's ok for now.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/core/Guess_lang.ml
+++ b/semgrep-core/src/core/Guess_lang.ml
@@ -120,12 +120,7 @@ let split_cmd_re = lazy (Pcre_settings.regexp "[ \t]+")
      "#!/usr/bin/env -S bash -e -u" -> ["/usr/bin/env"; "bash"; "-e"; "-u"]
 *)
 let parse_shebang_line s =
-  let matched =
-    try Some (Pcre.exec ~rex:(Lazy.force shebang_re) s) with
-    | Not_found
-    | Pcre.Error BadUTF8 ->
-        None
-  in
+  let matched = Pcre_settings.exec_noerr ~rex:(Lazy.force shebang_re) s in
   match matched with
   | None -> None
   | Some matched -> (
@@ -137,7 +132,8 @@ let parse_shebang_line s =
           match string_chop_prefix ~pref:"-S" arg1 with
           | Some packed_args ->
               let args =
-                Pcre.split ~rex:(Lazy.force split_cmd_re) packed_args
+                Pcre_settings.split_noerr ~rex:(Lazy.force split_cmd_re)
+                  ~on_error:[ packed_args ] packed_args
                 |> List.filter (fun fragment -> fragment <> "")
               in
               Some (arg0 :: args)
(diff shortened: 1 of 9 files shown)


IagoAbal (approved):
Super!

IagoAbal:
This should close #4258. We may want to add a test to semgrep/tests, but I can do that tomorrow morning (CET).

mjambon:
I leave this open in case you want to merge into the release branch rather than `develop`.

Having tests would be good, but as mentioned in Slack, I think useful tests would be to scan a repo with weird files (wrong extensions, binary files, etc.) and expect semgrep to not crash and not take too long to ignore irrelevant files. I didn't have time to do this today.

semgrep-core/src/core/Guess_lang.ml
```diff
@@ -120,12 +120,7 @@ let split_cmd_re = lazy (Pcre_settings.regexp "[ \t]+")
      "#!/usr/bin/env -S bash -e -u" -> ["/usr/bin/env"; "bash"; "-e"; "-u"]
 *)
 let parse_shebang_line s =
-  let matched =
-    try Some (Pcre.exec ~rex:(Lazy.force shebang_re) s) with
-    | Not_found
-    | Pcre.Error BadUTF8 ->
-        None
-  in
+  let matched = Pcre_settings.exec_noerr ~rex:(Lazy.force shebang_re) s in
```
aryx, line 123: We should augment semgrep.yml and add a rule to say to not use Pcre.exec directly.

aryx: I don't like the name of the module though Pcre_settings; sounds like it contains PCRE settings when really
we use it as a wrapper over Pcre.

semgrep-core/src/core/Pcre_settings.mli
```diff
@@ -20,6 +20,99 @@ val regexp :
   string ->
   Pcre.regexp
 
+(*
```
aryx, line 23: What about renaming this file Pcre_noexn.ml?
Or Pcre_result?

mjambon: I don't like the name `Pcre_settings` for this either, it's misleading. Maybe `R2C_pcre`? `R2C` is the namespace we wish we had if OCaml had namespaces.

aryx: nah, not a huge fan of adding R2C here. Maybe PPcre? I saw some people doing LList, MMap for wrappers.

mjambon: I'd much rather have a common prefix because it allows grouping, sorting, and grepping. One letter is fine, but should be always the same letter. The [containers library uses the two letters `CC`](https://github.com/c-cube/ocaml-containers/tree/master/src/core). I think it works well. It's unambiguous, doesn't shadow the original definitions, and it's not too long.

Maybe just `S` for `Semgrep`, giving us `SPcre`.

aryx: Ok, I'm fine with SPcre.
cc @IagoAbal @emjin @mmcqd

PR #4266, opened 2021-11-12
Deduplicate rules by id and behavior

emjin:
If multiple rules have the same id and behavior, run only the first one. Behavior is everything except message, metadata, and severity.

This is a bit of a patch, ideally there will be a consistent story about what we expect from a set of rules passed into semgrep. 

TODO: Add a metacheck to ensure configs don't have duplicates (https://github.com/returntocorp/semgrep/issues/4283)
Consider: enforce the registry doesn't contain two rules with the same behavior
Consider: check that users aren't adding duplicate rules to their rulesets in semgrep-app

Test plan: pytest

PR checklist:
- [x] Changelog is up-to-date


ievans:
@emjin won't this lead to non-deterministic output?

emjin:
Left this issue unfinished. This PR comes with a TODO to add a pre-check that ensures no two rules have the same ruleid but different behavior. We think this is always unintentional and should be blocked. If you think that's sufficiently important, I can add that change to this PR.

Edit: clarified verbally and resolved

dlukeomalley:
> If you think that's sufficiently important, I can add that change to this PR.

I think it's worth making discoverable and/or very clear that rules have been deduplicated. I'd be most comfortable with dedup if it compared a hash of the rule's content, and if that hash isn't the same, run both rules but warn to the user in the CLI output that the same rule id was seen (and we expect them to be unique).

emjin:
We already dedup when rules are exactly identical. In CI, we're still running duplicate rules. The problem is that we annotate rules with the policy they come from via the "ruleset" field in "metadata", thus making the rules not the same. Obviously, this is wrong. The minimum change we must make is to ignore the ruleset field in a comparison. Immediately this solution sets off some flags to me as fragile, however. What if we rename it? What if there are duplicates coming in some other way? I'd much rather have a coherent way of handling duplicates.

An axiom:
We should never run two rules with identical behavior. Even if we wanted to return separate findings, we should only run the rule once

In thinking about same-id-different-content rules, I think the most important distinction is by whether the finding returned would be different or not.

Case: behavior is different

It seems very unlikely someone would _want_ two rules to have identical ids but different behavior. It seems very easy for someone to get two rules with identical ids but different behavior by mistake; they modify a rule without changing the id, but the original rule is still in a policy. I think they would want to know that. This is why I think we should error, though I see the argument that there should be an escape hatch so we don't block people over this. Open to argument

Case: behavior is the same

It still seems unlikely someone would want two identical rules to have different information. If they don't like our message/severity ratings, they will want to override it entirely. If they think there's additional metadata, that metadata (except specifically for ruleset) applies to the behavior. 

For metadata specifically, there is a legitimate argument for it being merged. Something else similar to ruleset could come up. Perhaps someone would want to know that some rule was in multiple policies they run. But I think implementing this would be overly complex. I imagine that in most cases it would be more appropriate for metadata to be merged. However, I'm open to adding a warning

Conclusion:

There are several possible solutions, but we definitely should not be running rules with identical behavior multiple times. I argue for just checking the ruleids as the simplest to implement and (thus) easiest to communicate, then expecting ruleids to imply uniqueness.

mjambon:
Random thoughts:
* We could use the ruleset name or path as a namespace, which would tolerate two rules with the same name if they come from different rulesets (but not exactly like this, see longer comment below).
* Deduplicating rules that behave identically can be done independently from their name. Two identical rules with different names should still be deduplicated and/or produce a warning, no?

If rule names are meant to be globally unique, we should give tools to reassure rule authors that it's the case. Usually, that's some kind of namespacing. I don't think file names or paths should automatically form a namespace unless it's a relative path with respect to some root as in Java. A possibility moving forward is to support and encourage adding a `namespace` field to the rule file. Something like this:
```
namespace: r2c
rules:
  -id: not-using-our-pcre-wrappers
    patterns:
    ...
```
Replace `r2c` by your organization, team, or project name. The default and legacy namespace would be the empty path.

Namespaces would solve the problem of copy-paste-modify without having to rename all the rules.

emjin:
You're right; ultimately I think it should be about behavior if anything. I'm trying to figure out how we want to communicate to users that we ran a rule only once. Maybe for now what we should do is tie it to our output---deduplicate based on the fields we display to the user. I'm concerned that this won't be very robust (what if we add a new field?), but I think we have worse APIs.

For now I will use id and behavior and dedup silently. We will see if we get complaints

mjambon:
I think I would print warnings about duplicate rules or skipped rules at the end of the scan, in a compact fashion. Something like this:
```
Warning: 17 rules are duplicates. See details above.
```
This would be part of the final summary, just before or after printing the findings. The full list would be printed earlier, more or less like this:
```
duplicate rule: NAME found in RULESET1 and RULESET2
duplicate rule: NAME1, NAME2 found in RULESET1, RULESET2, and RULESET3
...
```

Similarly, we should also print warnings about skipped targets, which we haven't done yet. I also think we should print a brief summary at the end, and the full list of skipped targets earlier, possibly only in `--verbose` or `--debug` mode.

semgrep/semgrep/rule.py
```diff
@@ -182,3 +183,11 @@ def full_hash(self) -> str:
         return hashlib.sha256(
             json.dumps(self._raw, sort_keys=True).encode()
         ).hexdigest()
+
+
+def obscure_informational_fields(rule: Rule) -> Rule:
```
aryx: obscure? Do you mean "reset" instead?

IagoAbal:
I would not deduplicate silently here. If the user gave us weird input, either the user did it on purpose and hence we should do as the user commands, or the user made a mistake and then the user needs to know. I would add this check to the meta checker (if it is not there yet) and I would warn the user at the end of the run. If we think this is almost for sure a mistake, then I would warn reasonably "loud" so that it gets noticed.

emjin:
> I would not deduplicate silently here. If the user gave us weird input, either the user did it on purpose and hence we should do as the user commands, or the user made a mistake and then the user needs to know. I would add this check to the meta checker (if it is not there yet) and I would warn the user at the end of the run. If we think this is almost for sure a mistake, then I would warn reasonably "loud" so that it gets noticed.

I think this is not a metacheck and it definitely should be. I'll try to add it by EOW. However, this can easily happen without the user making a mistake or intending this behavior. Users run semgrep-action with rulesets, and two rulesets frequently include the same rule. These rules will be identical except for the "ruleset" field in the "metadata" section.

I did consider only ignoring that specific field, but this feels fragile. What if we change that field? Also, what if two rulesets contain identical rules except for severity? Users probably still only want that rule run once. I'm happy to add a warning, but ultimately I don't think the warning would be very helpful (what action would users take?). The checks imo should be at the registry level and the semgrep-app ruleset creation level. I can make TODOs for those.

IagoAbal:
> I think this is not a metacheck and it definitely should be. I'll try to add it by EOW. However, this can easily happen without the user making a mistake or intending this behavior. Users run semgrep-action with rulesets, and two rulesets frequently include the same rule. These rules will be identical except for the "ruleset" field in the "metadata" section.

Thanks for the context! (Sorry, I did read the previous comments but clearly not carefully enough, my bad.) If it's OK for rulesets to overlap then silently deduplicating makes more sense. Although I think that if we deduplicate by severity, then should not we make sure that we keep the rule with the highest severity and not just a random one?

> I did consider only ignoring that specific field, but this feels fragile. What if we change that field? Also, what if two rulesets contain identical rules except for severity? Users probably still only want that rule run once. I'm happy to add a warning, but ultimately I don't think the warning would be very helpful (what action would users take?). The checks imo should be at the registry level and the semgrep-app ruleset creation level. I can make TODOs for those.

Couldn't we just silently deduplicate but only ignoring the metadata? We could still throw some warnings if running with `--verbose`.

emjin:
>  Couldn't we just silently deduplicate but only ignoring the metadata? We could still throw some warnings if running with --verbose.

I'm happy to deduplicate ignoring only the metadata! I talked to Brendon and it sounds like the situation I was concerned about (someone modifies severity in one ruleset but not the other) can't happen because rulesets contain pointers to rules. I was on the fence about severity and message anyway since they show up in the results.

brendongo:
I think with just removing metadata field for silent deduping we should be fine wrt concerns brought up in this thread. Users should not even notice this change in the output of semgrep.

brendongo (approved):

PR #4268, opened 2021-11-12
Add Xlang module and better error messages for invalid languages

mjambon:
Now, we store the result of the `-lang` option after parsing it instead of keeping it as a string. In general, we now expect a language of type `Xlang.t`, which used to be defined in `Rule.ml` and now has its own module. In practice, we still expect a "normal" semgrep language for most operations, that is `generic` (spacegrep) and `regex` are excluded. The goal is to progressively add support for these as well. For example, scanning for a pattern specified on the command line with `-e` should eventually work with `-lang generic` but doesn't work at the moment.

Testing: we expect invalid languages specified with `-lang` to show the correct list of languages

Case 1: invalid language - `generic` and `regex` are shown in the list
```
$ semgrep-core -lang bf
Uncaught exception:
  
  (Failure
   "unsupported language: bf; supported language tags are: bash, c, c#, c++, cpp, cs, csharp, generic, go, golang, hack, hcl, html, java, javascript, js, json, kotlin, kt, lua, ml, ocaml, php, py, python, python2, python3, r, rb, regex, rs, ruby, rust, scala, sh, terraform, tf, ts, typescript, vue, yaml")
...
```

Case 2: valid extended language that's not supported for the given operation - `generic` and `regex` are not in the list
```
$ semgrep-core -lang regex -e foo <(echo foo)
Uncaught exception:
  
  (Failure
   "unsupported language: regex; supported language tags are: bash, c, c#, c++, cpp, cs, csharp, go, golang, hack, hcl, html, java, javascript, js, json, kotlin, kt, lua, ml, ocaml, php, py, python, python2, python3, r, rb, rs, ruby, rust, scala, sh, terraform, tf, ts, typescript, vue, yaml")
...
```

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -488,7 +488,7 @@ let options () =
       Arg.Set_string config_file,
       " <file> obtain formula of patterns from YAML/JSON/Jsonnet file" );
     ( "-lang",
-      Arg.String (fun s -> lang := Xlang.of_string_opt s),
+      Arg.String (fun s -> lang := Some (Xlang.of_string s)),
       spf " <str> choose language (valid choices:\n     %s)"
         Xlang.supported_xlangs );
     ( "-target_file",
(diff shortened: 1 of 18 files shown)


aryx (approved):
Nice

PR #4269, opened 2021-11-12
Apply language-independent filtering of regex and generic targets

mjambon:
This will filter out regex and generic targets that exceed the usual criteria such as excessive size, binary files etc. like we already do for other languages.

todo: testing

PR checklist:
- [x] Changelog is up-to-date


aryx (approved):

mjambon:
I believe these changes have no effect right now on `semgrep` since it passes individual files to `semgrep-core` which then doesn't filter them because they're passed explicitly. For any filtering to happen in semgrep-core, semgrep needs to distinguish explicit targets from discovered targets, which was the object of PR #3949 but was undone by #4048.

aryx (approved):

PR #4282, opened 2021-11-15
[Internal] make the OCaml library public with public_name

aryx:
This will allow to use semgrep inside other projects

test plan:
see related PR in xxx-semgrep

PR checklist:
- [x] Changelog is up-to-date


IagoAbal (approved):
Not an OPAM expert, but looks fine.

Issue #3880, opened 2021-09-15
Ruby: Taint does not flow into do-end blocks


minusworld:
**Describe the bug**
This taint rule does not appear to flow into the do-end block. I want it to match the `# ruleid` inside the block.
The block looks like this:

```ruby
  # ruleid: avoid-tainted-http-request
  Net::HTTP.start(uri.host, uri.port) do |http|
    # ruleid: avoid-tainted-http-request
    req = Net::HTTP::Get.new uri
    resp = http.request request
  end
```

**To Reproduce**
https://semgrep.dev/s/GwQe

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [x] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

emjin:
cc @IagoAbal @mmcqd

IagoAbal:
Do you need to match both the block and the `Get` inside the block with the exact same rule? Most likely this is the result of taint-mode removing those kind of "duplicate matches", see PR https://github.com/returntocorp/semgrep/pull/3781. If taint finds two matches, and one is strictly contained in the other, it will only report the larger one.

IagoAbal:
I think this is mainly an issue of the Ruby front-end that encodes:
```ruby
Net::HTTP.start(uri.host, uri.port) do |http|
    # ruleid: avoid-tainted-http-request
    req = Net::HTTP::Get.new uri
    resp = http.request request
end
```
in Generic as
```
  Net::HTTP.start(uri.host, uri.port, |http|
    # ruleid: avoid-tainted-http-request
    req = Net::HTTP::Get.new uri
    resp = http.request request)
```
which I don't find very accurate. And that encoding causes that matching the call to `Net::HTTP.start` also matches the entire `do-end` block.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-360/ruby-taint-does-not-flow-into-do-end-blocks.

IagoAbal:
> And that encoding causes that matching the call to `Net::HTTP.start` also matches the entire do-end block.

Well, I didn't know about Ruby blocks and I thought they had different semantics.. sorry if confused things. So it seems that the `do-end` block is a lambda that is passed as an argument to the method that precedes it?

Anyways, perhaps we could instead encode:

```ruby
Net::HTTP.start(uri.host, uri.port) do |http|
    # ruleid: avoid-tainted-http-request
[...]

PR #4285, opened 2021-11-16
[ruby] Use an extra Call node to encode Ruby blocks

IagoAbal:
Passing the body of the block as yet another argument had some
undesirable (?) side effects. For example, given `f(x) { |n| puts n }`,
pattern `f(...)` matched the entire block rather than just `f(x)`, and
`f($X)` did not match anything!

Helps #3880

test plan:
make test

PR checklist:
- [x] Changelog is up-to-date


emjin (approved):

emjin:
We probably need to update the benchmark test version, unless anyone wants to hunt down the regression

PR #4286, opened 2021-11-16
Scala: recognize metavariables in patterns

mmcqd:
Previously a pattern like `case $X => ...` would not match against `case x => x`, because we did not recognize metavariables in patterns.

test plan: make test (see test)

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/pfff/scala_to_generic.ml
+++ b/semgrep-core/src/parsing/pfff/scala_to_generic.ml
@@ -320,6 +320,9 @@ and v_pattern = function
       match v1 with
       | Left lit -> G.PatLiteral lit
       | Right e -> todo_pattern "PatLiteralExpr" [ G.E e ])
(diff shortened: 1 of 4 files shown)


emjin (approved):

semgrep-core/src/parsing/pfff/scala_to_generic.ml
```diff
@@ -320,6 +320,9 @@ and v_pattern = function
       match v1 with
       | Left lit -> G.PatLiteral lit
       | Right e -> todo_pattern "PatLiteralExpr" [ G.E e ])
+  | PatName (Id id, [])
+    when AST_generic_.is_metavar_name (fst (v_varid_or_wildcard id)) ->
```
aryx, line 324: Thanks for using AST_generic_ and not AST_generic here :)

PR #4287, opened 2021-11-16
fix(ci): trick brew to use pip installed pipgrip

### LINKED ISSUE #4213: brew auto-pr is removing ruamel.yaml.clib dependency
**Describe the bug**
When creating an auto homebrew PR, it removes the ruamle.yaml.clib resource

Looks like this is because ruamel only lists clib as a dependency when running python < 3.10
https://sourceforge.net/p/ruamel-yaml/code/ci/default/tree/__init__.py#l16

And in our release script https://github.com/returntocorp/semgrep/blob/develop/.github/workflows/release.yml#L303 we install pipgrip with brew which installs with a 3.10 runtime which means when pipgrip is used to find dependencies of semgrep, it misses clib.

Fix will be to install pipgrip with pip instead so we can run on python < 3.10

brendongo:
brew bump-formula-pr expects pipgrip to be installed using brew so https://github.com/returntocorp/semgrep/pull/4233 did not correctly fix #4213 

Instead we have to install pipgrip with brew and pypi and link where the brew pipgrip executable should be to the pypi installed executable

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -297,15 +297,21 @@ jobs:
     needs: [sleep-before-homebrew]  # Needs to run after pypi released so brew can update pypi dependency hashes
     runs-on: macos-10.15
     steps:
-      - name: Brew update
-        run: brew update
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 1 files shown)


brendongo:
The failing benchmark test is unrelated

PR #4288, opened 2021-11-16
Add more useful scripts

emjin:
PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/scripts/processing-output/duplicate-diff.py
+++ b/scripts/processing-output/duplicate-diff.py
@@ -1,0 +1,50 @@
+import argparse
+import json
+
+# Measure the effect of accidental duplication
+# Should not be useful soon!
+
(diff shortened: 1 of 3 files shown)


scripts/processing-output/duplicate-diff.py
```diff
@@ -0,0 +1,50 @@
+import argparse
+import json
+
+# Measure the effect of accidental duplication
+# Should not be useful soon!
```
IagoAbal, line 5: So is this still useful?

IagoAbal (approved):

scripts/processing-output/duplicate-diff.py
emjin: Yeah, it's useful until we release deduplication

PR #4290, opened 2021-11-17
Scala: translate infix operators as method calls in the generic AST

mmcqd:
`x f y` means `x.f(y)` in scala. We want to be able to match `xs map f` with the pattern `$X.map($F)`.

PR checklist:
- [x] Changelog is up-to-date


aryx (approved):
Nice.
Note that maybe at one point people may want to explicitely find all the
infix variant of the call, in which case we could make it a config option  (See Config_semgrep.atd for examples of such config).

PR #4314, opened 2021-11-19
Feature/pattern addressable PoC

DrewDennison:
Introduces a new key, `r2c-internal-patterns-from`, which can take as its value any config type that can be passed to `--config`. The contents of the key will be replaced with the contents of the `patterns` key in the remote config (which is only allowed to have a single rule and must contain a `patterns` key).

In this initial release, the key is named r2c-internal-project-depends-on to signal that the syntax & behavior for the key is not stable and may be subject to removal or future changes. (We welcome external contributors to try out the key, but keep in mind there's no expectation of stability across releases yet.)

Example values:
* `- r2c-internal-patterns-from: ./safe-document-html-filters.yaml`
* `- r2c-internal-patterns-from: r/kotlin.lang.security.cookie-missing-secure-flag`

PR checklist:
- [ ] Changelog is up-to-date
- [ ] Change has no security implications (otherwise, ping security team) -> pinged


semgrep/semgrep/config_resolver.py
```diff
@@ -339,6 +340,68 @@ def _rename_rule_ids(valid_configs: Mapping[str, Sequence[Rule]]) -> None:
...
+                        for item in inside_wrapper_key:
+                            if len(item) == 1 and 'patterns' in item:
+                                patterns = item['patterns']
+                                new_patterns = replace_patterns(patterns)
+                                item['patterns'] = new_patterns
+                            else:
+                                logger.warning("invalid state not supported yet for patterns-from")
+                    else:
+                        if 'patterns' in rule.raw:
+                            patterns = rule.raw['patterns']
```
brendongo: fyi: Modifying rule in place bypasses rule schema validation as that happens in init time.

semgrep/semgrep/config_resolver.py
```diff
@@ -339,6 +340,68 @@ def _rename_rule_ids(valid_configs: Mapping[str, Sequence[Rule]]) -> None:
...
+            # TODO make copy of patterns and return new_patterns
+            to_replace = {}
+            for idx, item in enumerate(patterns):
+                if 'patterns-from' in item: # TODO handle `patterns-from` not in top level
+                    value = item['patterns-from']
+                    # now resolve them as config objects
+                    config, errors = Config.from_config_list([value]) # TODO validate no errors
+                    if errors:
+                        logger.warning(f"There were errors resolving patterns-from: {errors}")
+                    rules = config.get_rules(no_rewrite_rule_ids)
```
brendongo: Probably fine to resolve in line for now but we probably want something that will precalculate the pattern dependencies and pull them all in at once instead of on demand like this?

brendongo:
IMO clean solution for prime time would be to separate the class responsible for downloading and parsing rules from the class that hydrates pattern-from

We’re gonna have some problems on error reporting cause semgrep-core reporting a rule parse error on line 25 of the rule will not be actually on line 25 of the prehydrated rule.yaml but that’s fine and we probably need to move to a world where we report the broken line and rule id instead of the line number

but i think it’s good enough to merge as is after cleaning up the todos for internal dogfooding

semgrep/semgrep/util.py
```diff
@@ -229,7 +231,9 @@ def dict_mutate_keyvalues(indict: Dict[Any, Any], f: Callable[[Any, Any], Any])
                     dict_mutate_keyvalues(v, f)
             else:
                 # print(f'replaced {key}: {value} -> {f(key, value)}')
-                indict[key] = f(key, value)
+                new_key, new_value = f(key, value)
+                del indict[key]
```
ievans: @brendongo it's ugly

semgrep/semgrep/config_resolver.py
ievans: I implemented that @brendongo

semgrep/semgrep/config_resolver.py
ievans: I fixed this @brendongo

brendongo (approved):
Can we include a unittest for dict_mutate_keyvalues

semgrep/semgrep/util.py
```diff
@@ -204,5 +206,69 @@ def truncate(file_name: str, col_lim: int) -> str:
...
+    For any child of `indict` which is a list value with a dictionary
+    where `predicate(key)` holds true for a value in that dictionary
+    remove the dictionary and replace its entry in the list by splicing in the
+    value of `replacement(key, value)`
+
+    If a child which is a dictionary has a key which matches the predicate, delete
+    the matched key and insert a new key `new_key_name` with the value `replacement(key, value)`.
+
+    """
+    if isinstance(indict, dict):
```
nbrahms, line 239: https://en.wikipedia.org/wiki/Cyclomatic_complexity !

semgrep/semgrep/util.py
```diff
@@ -204,5 +206,69 @@ def truncate(file_name: str, col_lim: int) -> str:
...
+
+        for key, value in list(indict.items()):
+            # print(f'{key}: {value}')
```
nbrahms, line 242: Delete me...

semgrep/semgrep/config_resolver.py
```diff
@@ -348,6 +407,42 @@ def _rename_rule_ids(valid_configs: Mapping[str, Sequence[Rule]]) -> None:
                     f"{Config._convert_config_id_to_prefix(config_id)}{rule.id or MISSING_RULE_ID}"
                 )
 
+    @staticmethod
```
nbrahms, line 410: For ease of future maintainability, can we put:
- `_fetch_remote_configs`
- The above inline code in `get_rules`
- `_find_config_remote_references`
- `_replace_patterns_from`
in a separate file?

semgrep/semgrep/config_resolver.py
```diff
@@ -312,6 +360,17 @@ def get_rules(self, no_rewrite_rule_ids: bool) -> List[Rule]:
             # re-write the configs to have the hierarchical rule ids
             self._rename_rule_ids(configs)
 
+        config_to_patterns_from: Dict[str, List[str]] = {}
+        for config in configs:
```
nbrahms, line 364: Do you want to deduplicate repeated references?

kurt-r2c (commented):
I have gone through this a couple times now and nothing security-related has jumped out at me.  I will leave the engineering correctness to @nbrahms

semgrep/semgrep/config_resolver.py
```diff
@@ -127,7 +133,9 @@ def __init__(self, config_str: str, project_url: Optional[str]) -> None:
         if self.is_registry_url():
             metric_manager.set_using_server_true()
 
-    def resolve_config(self) -> Mapping[str, YamlTree]:
+    def resolve_config(
```
kurt-r2c, line 136: `_make_config_request` doesn't right-pin `semgrep.dev` in `ParseResult.netloc`:
https://github.com/returntocorp/semgrep/pull/4314/files#diff-fa93cdfb09f1a1e0580936c43d7b3fd0bcec220bc1e8163bcd2b54747ec159b5L155

The URL specification allows subdomains to be ridiculously long (something like 255 characters) so the following attacker URL would pass all of the config URL checks:  `https://semgrep.dev.something.else.evil.com/api/agent/deployment` eventually leaking the app deployment token here: https://github.com/returntocorp/semgrep/pull/4314/files#diff-fa93cdfb09f1a1e0580936c43d7b3fd0bcec220bc1e8163bcd2b54747ec159b5L215

kurt-r2c (changes requested):
Found at least one info disclosure vector from Evil Rule Definitions

erwanlr:
That sounds like a great feature, especially for taint mode!

If it could also allow to load multiple rules, e.g:

```
- pattern-either:
  - r2c-internal-patterns-from:
    - ./rules-1.yml
    - ./rules-2.yml
```

as well as be able to include the `r2c-internal-patterns-from` inside a rule which is already a loaded from, eg:

```
rules:
- id: pattern-from-2
  languages: [python]
  severity: INFO
  patterns:
    - r2c-internal-patterns-from: ./pattern-from-1.yml
    - pattern-either:
      - pattern: another_pattern(...)
```

PR #4316, opened 2021-11-19
analyzing: CFG_build: Make sure labels are resolved

### LINKED ISSUE #4265: Java control flow analysis false positive
**Describe the bug**

**To Reproduce**
Running on [this java file](https://github.com/eclipse/jetty.project/blob/b06ccf286d8b72c745883a8218e065cba455dcf4/jetty-websocket/websocket-jetty-common/src/main/java/org/eclipse/jetty/websocket/common/JettyWebSocketFrameHandler.java#L463)

there is a false positive on the $X == $X rule. See https://semgrep.dev/s/DzWP

**Expected behavior**
The rule should not fire on the file or https://semgrep.dev/s/DzWP

**Environment**
semgrep 0.72

IagoAbal:
We did not create a dummy node to apply pending labels to when the
preceeding statement exited the function. We must create the dummy node
and label it, otherwise some `goto` may be left unresolved.

This also fixes the "Could not resolve label: _label" warnings that we
had been observing for a while.

Fixes: 20cd8a784b8 ("Add support for Break and Continue in IL (#3925)")

Closes #4265

test plan:
make test # test included

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/analyzing/CFG_build.ml
```diff
@@ -261,14 +261,28 @@ and cfg_stmt_list state previ xs =
...
-      (* If we had labels at the end of our stmt list, we create a dummy node to assign them to,
-         and connect it to the last node we looked at
-      *)
-      let dummyi = state.g#add_node { n = NOther Noop } in
+  match labels with
+  | l :: ls ->
+      (* If we have labels at the end of our list of stmt, we create a dummy
+       * node to assign them to.
+       *
+       *     lasti -> label: dummy
```
aryx: I've found this 'lasti -> label: dummy' confusing, because actually what this new code does is that it
creates the dummy node even if there is no lasti right?

aryx (approved):

aryx:
Not sure what is going on with CI.

semgrep-core/src/analyzing/CFG_build.ml
IagoAbal: Yeah that's right, I edited the comment to remove that. It was probably not adding much anyways.

mmcqd (approved):

PR #4317, opened 2021-11-19
Fix Stack_overflow when using -filter_irrelevant_rules

### LINKED ISSUE #4305: [PHP] Stack overflow with big list of pattern-inside
See https://r2c-community.slack.com/archives/C018NJRRCJ0/p1637242553127900

aryx:
This closes #4305

test plan:
test file included

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/optimizing/Analyze_rule.ml
+++ b/semgrep-core/src/optimizing/Analyze_rule.ml
@@ -170,7 +170,12 @@ type step0 = LPat of Rule.xpattern | LCond of Rule.metavar_cond
 
 type cnf_step0 = step0 cnf [@@deriving show]
 
-(* reference? https://www.cs.jhu.edu/~jason/tutorials/convert-to-CNF.html *)
+(* reference? https://www.cs.jhu.edu/~jason/tutorials/convert-to-CNF.html
+ * TODO the current code triggers some Stack_overflow on
+ * tests/OTHER/rules/tainted-filename.yaml. I've replaced some List.map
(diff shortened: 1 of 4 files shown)


semgrep-core/src/optimizing/Analyze_rule.ml
```diff
@@ -517,8 +522,14 @@ let regexp_prefilter_of_rule r =
...
-      | R.Taint spec -> regexp_prefilter_of_taint_rule t spec)
+      try
+        match r.mode with
+        | R.Search pf ->
+            let f = R.formula_of_pformula pf in
+            regexp_prefilter_of_formula f
+        | R.Taint spec -> regexp_prefilter_of_taint_rule t spec
+      with
+      (* TODO: see tests/OTHER/rules/tainted-filename.yaml *)
+      | Stack_overflow ->
```
IagoAbal, line 533: `regexp_prefilter_of_taint_rule` already has a top-level `try-with`, maybe one could just add a case for `| Stack_overflow ->` there?

IagoAbal (approved):

semgrep-core/src/optimizing/Analyze_rule.ml
aryx: yes, good idea.

aryx: hmm but then I don't have the rule id to logger the issue
simpler to keep there then.

PR #4321, opened 2021-11-23
fix(nosem): restore disable-nosem functionality that modifies is_ignore field of findings

brendongo:
PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/ignores.py
+++ b/semgrep/semgrep/ignores.py
@@ -41,14 +41,13 @@ def process_ignores(
     :param strict: The value of the --strict flag (affects error return)
     :param disable_nosem: The value of the --disable-nosem flag
     :return:
-    - RuleMatchMap with nosem findings removed if disable_nosem is true
+    - RuleMatchMap: dict from rule to list of findings. Findings have is_ignored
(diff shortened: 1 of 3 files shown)


semgrep/semgrep/ignores.py
```diff
@@ -41,14 +41,13 @@ def process_ignores(
     :param strict: The value of the --strict flag (affects error return)
     :param disable_nosem: The value of the --disable-nosem flag
     :return:
-    - RuleMatchMap with nosem findings removed if disable_nosem is true
+    - RuleMatchMap: dict from rule to list of findings. Findings have is_ignored
+        set to true if there was matching nosem comment found for it.
+        Contains only findings without nosem if disable_nosem set to False
```
emjin, line 46: So many negatives...maybe "contains findings with nosem unless disable_nosem is passed"?

emjin (approved):

semgrep/semgrep/ignores.py
brendongo: I'll follow up in with a refactor from disable_nosem to filter_nosem or something to make it more clear

PR #4333, opened 2021-11-25
Use semgrep-langs submodule for language data

nbrahms:
This should reduce errors when adding new languages or promoting
languages by reducing language data to a single source of truth.

Removes the following language keys, as they can be confusing now, and may
prevent us from adding languages in the future:
`cs`, `ml`, `rb`, `rs`
I've confirmed that none of these are in current use in returntocorp/semgrep-rules.

cf.
- https://academiccommons.columbia.edu/doi/10.7916/D8959VR5
- SML, DML, F#, Lazy ML, PAL and other ML variants that are not OCaml

PR checklist:
- [x] Changelog is up-to-date


brendongo:
Is there a reason to have this json blob be outside this repo?

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/badlanguage/error-in-color.txt
```diff
@@ -5,5 +5,5 @@
 [94m  | [0m               [31m^^^^^^^^^^[0m
 [94m8 | [0m    severity: WARNING
 
-[31munsupported language: intercal. supported languages are: C#, C++, HCL, JSON, Json, Kotlin, Rust, Yaml, bash, c, cpp, cs, csharp, generic, go, golang, hack, hacklang, hcl, html, java, javascript, js, json, kotlin, kt, lua, ml, none, ocaml, php, py, python, python2, python3, rb, regex, rs, ruby, rust, scala, sh, terraform, tf, ts, typescript, vue, yaml[0m
+[31munsupported language: intercal. supported languages are: bash, c, c#, c++, cpp, csharp, generic, go, golang, hack, hcl, html, java, javascript, js, json, kotlin, kt, lua, none, ocaml, php, py, python, python2, python3, r, regex, ruby, rust, scala, sh, terraform, tf, ts, typescript, vue, yaml[0m
```
nbrahms, line 8: Hmm I'm not sure `cs`, `ml`, `rb`, or `rs` make sense here. And I don't think `hacklang` is valid in semgrep-core.

nbrahms: Also this is now case-insensitive.

nbrahms: I'll investigate to see if anyone is already using the four short abbreviations. Confirmed that `hacklang` is invalid in `Lang.ml`.

nbrahms:
> Is there a reason to have this json blob be outside this repo?

Yes, see https://github.com/returntocorp/semgrep-app/pull/4044/

brendongo:
Can we have this json in this repo as source of truth and have that repo pull in that file from here instead of a third repo? (Nonblocking question. Probably fine to have a third repo. Just wary about adding more submodules)

semgrep/semgrep/semgrep_types.py
```diff
@@ -14,97 +17,76 @@
...
-    Keyword arguments;
-    lang_str -- string representing a language (e.g. "C#")
-    span     -- span of language string in the original file (for error reporting),
-                None if resolve was called within semgrep
+
+class Language(str):
+    pass
+
+
+@dataclass(frozen=True)
```
brendongo: Let's not use dataclass since it isnt available in python 3.6

brendongo: We use attr elsewhere in the codebase

nbrahms: 👍

semgrep/semgrep/semgrep_types.py
```diff
@@ -14,97 +17,76 @@
...
+            id=Language(data["id"]),
+            name=data["name"],
+            keys=data["keys"],
+            exts=data["exts"],
+            reverse_exts=data.get("reverse_exts", data["exts"]),
+            shebangs=data.get("shebangs", []),
+        )
+
+
+class Lang:
```
brendongo: Lang is pretty close to Language and might get confusing. What do you think of `Linguist`?

Also looks like most places LANG (which should also be renamed) is used, it is invoked `LANG.data_by_id[language]` should we just expose a `get_definition_for_language` function so it's more readable?

brendongo:
Love this btw. Makes is super simple to add user defined file handling rules :O

nbrahms:
> Can we have this json in this repo as source of truth and have that repo pull in that file from here instead of a third repo? (Nonblocking question. Probably fine to have a third repo. Just wary about adding more submodules)

No, alas. You can't have a partial sub-module.

semgrep/semgrep/semgrep_types.py
nbrahms: I'll just make the class private, since one only every interacts with the singleton anyway

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/badlanguage/error-in-color.txt
nbrahms: This is the current exhaustive list used in semgrep-rules:

`['C#', 'bash', 'c', 'csharp', 'generic', 'go', 'hcl', 'java', 'javascript', 'js', 'json', 'kotlin', 'kt', 'lua', 'ocaml', 'php', 'python', 'regex', 'ruby', 'rust', 'scala', 'sh', 'ts', 'typescript', 'yaml']`

semgrep/semgrep/semgrep_types.py
```diff
@@ -14,97 +18,77 @@
...
-            str_to_language[lang_str] = language
-
-    """ Convert an inputted string representing a language to a Language
-
-    Keyword arguments;
-    lang_str -- string representing a language (e.g. "C#")
-    span     -- span of language string in the original file (for error reporting),
-                None if resolve was called within semgrep
+
+class Language(str):
```
brendongo: Is there a difference in declaring a class this way vs a NewType like above?

nbrahms: I think `NewType` is a type alias, and this is a strict subclass.

nbrahms: Hmm interesting. This way it's a subclass at both type-checking and runtime. `NewType` is only a subclass at type-checking time.

cf. https://docs.python.org/3/library/typing.html

brendongo (approved):

DrewDennison (approved):

semgrep/semgrep/semgrep_types.py
brendongo: Looks like NewType is just a faster helper? Not like python does any runtime typechecking anyway?

nbrahms: Right on. I'll swap it.

PR #4339, opened 2021-11-28
Generate Lang.ml from flat file

nbrahms:
This:
- Reduces the number of code locations we need to update when
  adding or modifying supported languages
- Moves language data to a single source of truth
- Ensures consistency of language data between idioms and codebases

cc. @aryx I saw the note about external libraries, so wanted to get your thoughts here and how to best accomplish this goal while not negatively impacting external consumers too much.

PR checklist:
- [X] Changelog is up-to-date


nbrahms:
cc. @ievans

aryx (approved):
Lgtm.
What about my notes on external libraries? Where do I speak about that?

nbrahms:
> Lgtm. What about my notes on external libraries? Where do I speak about that?

It's in a comment in `src/core/ast/dune`:

> This library used to be in pfff and is still used in projects outside semgrep (e.g., codemap/efuns) which rely on this pfff-lang_... name and rely on small dependencies.

aryx:
ah true ... annoying.

aryx:
What about we do the code generation in the submodule and commit in the submodule repo the generated OCaml, python, typescript code?

PR #4342, opened 2021-11-29
logging: Reduce the size of --debug dumps

IagoAbal:
In a first step, we typically just want `semgrep --debug` to help us
identify which files are causing problems. For this we use the default
INFO level. If a message is logged rarely then it does no harm having it
at the INFO level. But, if it is logged often, then it may just be making
the dump larger without really helping that much in most situations.

This will move some frequently-logged messages to the TRACE level to
reduce the size of --debug dump. My intuition is that these messages are
not particularly useful when one is trying to find "the file" causing
trouble.

test plan:
$ semgrep --time --debug -c p/auto lang/javascript/tmp/* &>log
The size of the log files went down from 838M to 174M (~5x smaller).

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Match_rules.ml
+++ b/semgrep-core/src/engine/Match_rules.ml
@@ -932,7 +932,7 @@ and matches_of_formula config equivs rule_id file_and_more formula opt_context :
       (file, xlang, lazy_ast_and_errors, lazy_content)
       xpatterns
   in
-  logger#info "found %d matches" (List.length res.matches);
+  logger#trace "found %d matches" (List.length res.matches);
   (* match results per minirule id which is the same than pattern_id in
    * the formula *)
   let pattern_matches_per_id = group_matches_per_pattern_id res.matches in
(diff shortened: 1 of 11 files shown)


semgrep-core/src/engine/Match_rules.ml
```diff
@@ -986,8 +986,7 @@ let check_rule r hook default_config pformula equivs file_and_more =
 
 let check hook default_config rules equivs file_and_more =
   let { FM.file; lazy_ast_and_errors; _ } = file_and_more in
-  logger#info "checking %s with %d rules" file (List.length rules);
-  if rules = [] then logger#error "empty rules";
```
aryx, line 990: not sure why you removed the empty rules logger.

aryx (approved):

semgrep-core/src/engine/Match_rules.ml
IagoAbal: Despite being an "error" this happens very often either because we get a taint rule, or because the rule is not relevant for the file. If we don't consider it an error, then it's somewhat redundant with the previous logger. But I can add it back, I would just suggest that then we change the code so that `check` is not called with an empty list of rules.

IagoAbal: We would have to change this in Run_rules.ml
```ocaml
let check hook default_config rules equivs file_and_more =
  let search_rules, taint_rules, skipped_rules =
    filter_and_partition_rules rules file_and_more
  in
  let res_search =
    Match_rules.check hook default_config search_rules equivs file_and_more
  in
```

IagoAbal: I'll merge, but if you want us to do something about this just let me know.

PR #4349, opened 2021-12-01
Improve bloom filter optimization

emjin:
This PR makes two changes:
1. Uses sets instead of bloom filters (relatively minor)
2. Moves the filtering step from `m_list__m_stmt` to only before going deep in `m_stmts_deep` (very important!)

The two are in the same PR right now because I noticed the second one while testing timings for the first. The set filters change had previously been validated over the standard benchmarks (could be run via `-set_filters`). From perf benchmark timings, set filters are roughly the same speed as bloom filters. From spot checking memory (using top and running some of the perf benchmarks), set filters consistently take less memory than bloom filters. Therefore we replace set filters entirely with this PR.

Additionally, this PR fixes a huge regression caused by bloom filters on large files with deep statement matching. On `semgrep --config ~/workspace/testing/perf_tests/nodejsscan.yaml perf/bench/django/input/django`, semgrep goes from 2.4 minutes to 2.8 seconds (compared to 3.3 seconds with `-no_bloom_filter`. This is absolutely my bad I'm very sorry.

The regression: filters only live on each statement, so to get the list of strings for each statement list, we have to union all the strings for the statements. Previously, we would do this each time we went from `x :: xs` to `xs`. We wanted to exclude statement lists, but some files (such as 
`perf/bench/django/input/django/django/contrib/admin/static/admin/js/vendor/jquery/jquery.js`) have 3000 statements.

To achieve both, we now place the check before we choose to flatten the sub-statements of each statement. Each flattening should represent a significantly different state of the statements being matched, while also catching when we don't need to go deeper.

Timings:

`./perf-benchmarks on develop`:

image

This was the original timing, before either opts. Here, bloom/set filters make the time significantly worse for `njs-old-dvna`, which runs nodejsscan rules.

`./perf-benchmarks --filter-variant "std|no-bloom|set-filters"` with the `m_list__m_stmt` filter moved (to see just the effect of set filters):

image

The `njs-old-dvna` slowdown is gone! We can see that set filters are sometimes slower and sometimes faster than bloom filters (std); most deviations are likely random noise. Where the deviations are significant, they are in the set filters' favor.

`./perf-benchmarks --filter-variant "std|no-bloom"` on this branch:

image

Final comparison of having filters (std) vs not. While the two are mostly the same, there are a couple of cases (including, notably, `njs-old-dvna`, which previously did much worse with filters) where having set filters is significantly better. The differences should be most visible with rules such as the `nodejsscan` rules.

Some notes:

We see clearly that we have a need for better benchmarks. Since `nodejsscan` is so fast on `django`, we can probably add several nodejsscan benchmarks. This will fix one of our current blindspots.

PR checklist:
- [x] Changelog is up-to-date


emjin:
Running on `juice-shop`, which used to take hours, is now done in 5 minutes! (Probably could get it faster, it looks like there's strange stuff going on with file parsing)

```
Semgrep-core timing summary:
Total CPU time: 349.0481  File parse time: -0.7740  Rule parse time: 0.9869  Match time: 16.3976
Slowest 5/416 files
...ch/juice-shop/input/juice-shop/frontend/src/assets/private/three.js (833KB): 308.4940
...ce-shop/input/juice-shop/frontend/src/assets/private/dat.gui.min.js ( 48KB): 5.5905
perf/bench/juice-shop/input/juice-shop/server.js                       ( 25KB): 4.0858
perf/bench/juice-shop/input/juice-shop/data/datacreator.js             ( 21KB): 0.8517
perf/bench/juice-shop/input/juice-shop/lib/utils.js                    (  7KB): 0.5194
Slowest 5 rules to run (excluding parse time)
...orkspace.testing.perf_tests.ajinabraham.njsscan.ssrf_node.node_ssrf: run time 34.6650  match time 2.4803
...rkspace.testing.perf_tests.ajinabraham.njsscan.xss_node.express_xss: run time 33.7325  match time 1.4987
...orkspace.testing.perf_tests.ajinabraham.njsscan.regex_dos.regex_dos: run time 33.0647  match time 0.9014
...ce.testing.perf_tests.ajinabraham.njsscan.eval_require.eval_require: run time 32.9352  match time 0.8946
....perf_tests.ajinabraham.njsscan.regex_injection.regex_injection_dos: run time 32.3987  match time 0.2501
ran 122 rules on 416 files: 99 findings
found problems analyzing 22 files; run with --verbose for details or run with
--strict to exit non-zero if any file cannot be analyzed cleanly
semgrep --config ~/workspace/testing/perf_tests/nodejsscan.yaml  --time  357.02s user 31.29s system 112% cpu 5:43.99 total
```

semgrep-core/src/core/Flag_semgrep.ml
```diff
@@ -23,8 +23,6 @@ let filter_irrelevant_rules = ref false
 (* check for identifiers before attempting to match a stmt or stmt list *)
 let use_bloom_filter = ref true
```
aryx, line 29: so this should be called use_set_filter?

aryx: Or you prefer to consistently keep the bloom_filter legacy name?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -169,31 +169,29 @@ let cache_access : tin Caching.Cache.access =
...
+          | false -> patterns_in_any_stmt pats rest (pats_in_stmt pats stmt)
+          | true -> acc)
     in
-    let patterns_all_in_stmts acc x =
-      match acc with
-      | F.No -> Bloom_filter.No
-      | Maybe -> pattern_in_any_stmt x stmts F.No
-    in
-    List.fold_left patterns_all_in_stmts F.Maybe pattern_list
+    patterns_in_any_stmt pattern_strs stmts true
```
aryx, line 194: Maybe can be replaced by List.exists instead of manually iterating and maintaining a boolean accumulator.

aryx (approved):
Awesome. Very nice PR message, with very nice graphs!

IagoAbal:
It's a pity that your PR comment is not part of any of the commit messages. One day, and it could be sooner than one expects, GitHub may not exist anymore and that may be lost. IMO the motivation for this change, that you explained very well in the PR comment, should be recorded with Git too.

emjin:
> It's a pity that your PR comment is not part of any of the commit messages. One day, and it could be sooner than one expects, GitHub may not exist anymore and that may be lost. IMO the motivation for this change, that you explained very well in the PR comment, should be recorded with Git too.

I tried to include relevant information in the comment in Bloom_filter.ml! Let me know what specifics you think are missing there (other than the graphs)

Edit: also I will add the django with njsscan rules benchmark, which will prevent anyone from making my mistake with `m_list__m_stmt` again.

IagoAbal:
Sorry if it's there and I didn't see it (I did only skimmed through the changes), but I'm talking about the history behind this PR. Anyways, my point is that if you felt the need to explain all that in the PR comment, then IMO that's an argument for having the same info in the Git history.

Stuff like
> Additionally, this PR fixes a huge regression caused by bloom filters on large files with deep statement matching.

will not be obvious when someone digs into the Git history in N years.

emjin:
Will be more mindful to put it in the commit next time!

semgrep-core/src/core/Flag_semgrep.ml
emjin: Yeah, I'm used to the legacy name :)

semgrep-core/src/matching/Generic_vs_generic.ml
emjin: Good point 😅

PR #4352, opened 2021-12-01
optimizing: Analyze_rule: Fix segfault on Apple M1

IagoAbal:
In 7e28cd2104b (#4317) we added test example 'tainted-filename.yaml'
that has a very large `pattern-either` causing CNF computation to
"explode". The code was non tail-rec so this example lead to a
stack overflow, and we relied on OCaml's runtime intercepting it and
raising a Stack_overflow exception. Unfortunately this does not work on
Apple M1, no Stack_overflow is raised, and instead we get a segfault.

This rewrites the application of the distributive law in a tail-rec
manner, and instead we check the size of the accumulated formula to
detect a potential blow up

Fixes 7e28cd2104b ("Fix Stack_overflow when using -filter_irrelevant_rules (#4317)")

test plan:
Run this on Apple M1:
$ semgrep-core -lang php -config tests/OTHER/rules/tainted-filename.yaml -filter_irrelevant_rules tests/OTHER/rules/tainted-filename.php
It should no longer segfault.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/optimizing/Analyze_rule.ml
+++ b/semgrep-core/src/optimizing/Analyze_rule.ml
@@ -63,7 +63,7 @@ let logger = Logging.get_logger [ __MODULE__ ]
 (* Types *)
 (*****************************************************************************)
 
-(* Conjonctive normal form (CNF).
+(* Conjunctive normal form (CNF).
  *
  * Why a CNF instead of a DNF (Disjunctive normal form)?
  * Because in the context of producing a regexp, regexps are good at
@@ -88,6 +88,8 @@ exception GeneralPattern
 exception EmptyAnd
 
 exception EmptyOr
+
+exception CNF_exploded
 
 (*****************************************************************************)
 (* Utils *)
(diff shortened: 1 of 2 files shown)


emjin:
Oh this is because I took out the `--set_filter` option. I think if you merge develop it will work?

emjin (approved):
Nice!

aryx:
dummy benchmark error seems unrelated.

semgrep-core/src/optimizing/Analyze_rule.ml
```diff
@@ -198,30 +200,32 @@ let rec (cnf : Rule.formula -> cnf_step0) =
       let zs = Common.map (fun (_t, cond) -> And [ Or [ LCond cond ] ]) conds in
       And (ys @ zs |> Common.map (function And ors -> ors) |> List.flatten)
   | R.Or (_, xs) ->
+      let is_dangerously_large l = List.compare_length_with l 1_000_000 = 1 in
```
aryx, line 203: This result value is ugly, 1 I guess means greater. I always find it ugly in C.

aryx (approved):

aryx:
Ok let's merge this.

semgrep-core/src/optimizing/Analyze_rule.ml
IagoAbal: Totally agree! Actually it's even worse, I just double-checked `compare`s spec and it says that "a positive integer if x is greater than y". So it should be `> 0` to be on the safe side, I'll fix it in the next PR.

PR #4353, opened 2021-12-01
cli: Do not print to stdout when --output is given

IagoAbal:
test plan:
```
$ semgrep --max-target-bytes=64000 -c p/auto -o results.txt parsing-stats/lang/javascript/tmp/mui-org-material-ui
# ^ Does not print any results to stdout, but only writes them to 'results.txt'
```

PR checklist:
- [x] Changelog is up-to-date


IagoAbal:
I will close this in favor of deprecating `--output` (#4354).

IagoAbal:
It seems that not everyone agrees with #4354 so in the meantime I guess it's better to get this fixed.

aryx (approved):

PR #4357, opened 2021-12-02
Feature/project-depends-on PoC

ievans:
This PR adds support for a new key: `project-depends-on`. Under this key, NPM or PyPI dependencies can be specified along with the semver range that the rule should trigger for. `project-depends-on` filters the rule unless one of the children is matched by a lockfile. In this initial release, the key is named `r2c-internal-project-depends-on` to signal that the syntax & behavior for the key is not stable and may be subject to removal or future changes. (We welcome external contributors to try out the key, but keep in mind there's no expectation of stability across releases yet.)

limitations:
* Dependency resolution supports Yarn, npm, and Pip lockfiles only (pipfile.lock, yarn.lock, package-json.lock)
* ~very limited semver support (uses the python parser from `packaging` library, no support for NPM operators like `~`).~ Uses https://packaging.pypa.io/en/latest/specifiers.html which supports almost all NPM operators except `^`

Example:
```
rules:
- id: vulnerable-awscli-apr-2017
  patterns:
    - pattern-either:
      - pattern: boto3.resource('s3', ...)
      - pattern: boto3.client('s3', ...)
    - r2c-internal-project-depends-on:
      - namespace: pypi
        package: awscli
        version: "== 1.11.82"
  message: this version of awscli is subject to a directory traversal vulnerability in the s3 module 1
  languages: [python]
```

PR checklist:
- [x] Documentation is up-to-date (https://github.com/returntocorp/semgrep-docs/pull/402)
- [x] Changelog is up-to-date


aryx (commented):
Nice.

DrewDennison:
Woot!!

semgrep/Pipfile
```diff
@@ -14,3 +14,4 @@ tox = "*"
 click = "~=8.0.1"
 click-option-group = "~=0.5.3"
 semgrep = {editable = true,path = "."}
+packaging = "==21.0"
```
nbrahms: FYSA this may not be available on earlier python versions

semgrep/dependencyparser/models.py
```diff
@@ -0,0 +1,108 @@
...
+        while True:
+            data = fp.read(BLOCKSIZE)
+            if not data:
+                break
+            hash_func.update(data)
+        hexdigest: str = hash_func.hexdigest()
+        return hexdigest
+
+    @classmethod
+    def hash_file(cls, algorithm: str, filepath: str) -> str:
```
nbrahms: Building our own hash function seems wrong. There has to be a library for this.

semgrep/dependencyparser/package_restrictions.py
```diff
@@ -0,0 +1,95 @@
...
+        [
+            ProjectDependsOnEntry(PackageManagers.PYPI, "awscli", "<= 1.1182"),
+            ProjectDependsOnEntry(PackageManagers.PYPI, "colorama", "0.4.4"),
+        ],
+        lockfiles,
+    )
+    print(list(output))
+"""
+
+# main(Path("../../"))
```
nbrahms: remove me

semgrep/dependencyparser/package_restrictions.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@attr.s(auto_attribs=True, frozen=True, eq=True, order=True)
+class ProjectDependsOnEntry:
+    namespace: PackageManagers
+    package_name: str
+    semver_range: str
+
+
+def semver_matches(expression: str, actual_version: str) -> bool:
```
nbrahms, line 38: You should use the individual semver parsers, and instead use the semver comparison operator, vs. doing your own comparisons in python.

E.g. there's a NPM parser in semver.

semgrep/dependencyparser/parse_lockfile.py
```diff
@@ -0,0 +1,157 @@
...
+    To:
+        sha512,
+    """
+    algorithm = s.split("-")[0]
+    rest = s[len(algorithm) + 1 :]
+    decode_base_64 = base64.b64decode(rest)
+    return {algorithm: [base64.b16encode(decode_base_64).decode("ascii").lower()]}
+
+
+def parse_Yarnlock_str(lockfile_text: str) -> Generator[LockfileDependency, None, None]:
```
nbrahms, line 49: yarn lock is a json blob, so we should use a json parser, not string parsing

semgrep/semgrep/dependency_aware_rule.py
```diff
@@ -0,0 +1,98 @@
...
+) -> Generator[ProjectDependsOnEntry, None, None]:
+    for entry in entries:
+        for package_manager, dep_name_and_version_str in entry.items():
+            pm = None
+            if package_manager == "pypi":
+                pm = PackageManagers.PYPI
+            elif package_manager == "npm":
+                pm = PackageManagers.NPM
+            else:
+                raise SemgrepError(f"unknown package manager: {package_manager}")
```
nbrahms: Suggestion: make this a dictionary

semgrep/semgrep/rule_schema.yaml
```diff
@@ -443,47 +446,50 @@ properties:
             - message
             - languages
             - severity
-          oneOf:
+          anyOf:
```
nbrahms: I worry about this extra complexity here. Can we still require a pattern?

semgrep/semgrep/semgrep_main.py
```diff
@@ -193,7 +193,13 @@ def main(
         lambda rule: rule.mode == JOIN_MODE,
         filtered_rules,
     )
-    filtered_rules = rest_of_the_rules
+    dependency_aware_rules, _ = partition(
```
nbrahms: I also worry about this complexity & maintenance cost. Can we downscope this to require all depends-on rules to be runnable?

semgrep/semgrep/semgrep_main.py
```diff
@@ -216,6 +222,21 @@ def main(
             rule_matches_by_rule.update(join_rule_matches_by_rule)
             output_handler.handle_semgrep_errors(join_rule_errors)
 
+    if dependency_aware_rules:
```
nbrahms: And then can we move this logic into the rule evaluation logic?

nbrahms:
As a general request to improve maintainability of Semgrep, can we make sure to separate data from shared code (e.g. forks for various dependency managers, etc.)?

I'd expect, e.g., a shared class with implementations for various package managers.

semgrep/dependencyparser/models.py
nbrahms: StackOverflow seems to disagree with me :(

semgrep/Pipfile
brendongo: Do you need 21.0? semgrep/setup.py already requires "packaging>=20.4",

semgrep/dependencyparser/find_lockfiles.py
```diff
@@ -0,0 +1,17 @@
...
+
+TARGET_LOCKFILE_FILENAMES = ["pipfile.lock", "yarn.lock", "package-lock.json"]
+
+
+@functools.lru_cache(maxsize=None)
+def find_lockfiles(current_dir: Path) -> Generator[Path, None, None]:
+    for entry in os.scandir(current_dir):
+        full_path = Path(os.path.join(current_dir, entry.name))
+        if entry.is_dir():
+            yield from find_lockfiles(full_path)
```
brendongo: Do we need to worry about cycles cause of symlinks?

semgrep/semgrep/rule_schema.yaml
ievans: We don't want to require a pattern, so that semgrep can filter just based on the precense of a dep at a specific version. It might be better to have a `mode: dependencies`?

semgrep/dependencyparser/parse_lockfile.py
ievans: Yarn lock is JSON?

semgrep/Pipfile
ievans: the old version should be just fine

semgrep/dependencyparser/models.py
ievans: sad! we can remove this anyways -- this is vestigial code from my dependency parsing for scarydeps

semgrep/semgrep/semgrep_main.py
ievans: Not without moving the work into the core, which I don't think makes sense -- see comment on https://github.com/returntocorp/semgrep/pull/4357#discussion_r765170208

emjin: Is that a reason we can't move this code into the core eventually? To make our interface cleaner, I was hoping to move to a world where semgrep never opens a rule file.

emjin:
@aryx Would it make sense for us to support "lockfiles" as a language? We have a lot of CAI rules that read `yarn.lock` files too, and they're pretty slow considering they're just string searches. On semgrep-app:

```
| These files are taking the most time. Consider adding them to .semgrepignore or
| ignoring them in your Semgrep.dev policy.
|   - 19.76 s: frontend/yarn.lock
|   - 12.18 s: frontend/src/screens/editor/EditorHome.tsx
|   - 9.23 s: contributing/yarn.lock
```

And here is another use.

aryx:
:wave:

clintgibler:
> @emjin Would it make sense for us to support "lockfiles" as a language? 

This could be useful! Maybe that's what the implementation of `project-depends-on` uses at some point?

semgrep/dependencyparser/models.py
ievans: removed

semgrep/dependencyparser/package_restrictions.py
ievans: @nbrahms I don't see an NPM parser per se, but https://packaging.pypa.io/en/latest/specifiers.html looks great -- assuming that's what you are referring to, I will use it!

semgrep/dependencyparser/find_lockfiles.py
ievans: @brendongo probably. I think I addressed in 89410c1

semgrep/Pipfile
ievans: (I reverted this addition)

semgrep/dependencyparser/parse_lockfile.py
ievans: @nbrahms can you point at a source for that? It doesn't look like valid JSON to me.

ievans:
See https://r2c.quip.com/qnlxABdbSoXD/project-depends-on-syntax for further internal discussion on syntax.

semgrep/semgrep/rule.py
```diff
@@ -189,6 +201,10 @@ def full_hash(self) -> str:
             json.dumps(self._raw, sort_keys=True).encode()
         ).hexdigest()
 
+    @property
+    def has_runable_semgrep_rules(self) -> bool:
```
ievans: @emjin after your PR to the core, I think we can remove this

semgrep/semgrep/rule_schema.yaml
ievans: Resolved by moving under `patterns`

semgrep/semgrep/rule.py
ievans: nope, we still need it

semgrep/semgrep/semgrep_main.py
ievans: after discussion with @emjin and @nbrahms, we agreed we will eventually move this into the core. But for now it will be handled in python. In a later commit, I removed the partition so the same rules are going to the core, and are just post-processed in Python, which should make it a bit cleaner when it is time to do it all in the core.

semgrep/semgrep/rule.py
```diff
@@ -145,6 +147,19 @@ def severity(self) -> RuleSeverity:
     def mode(self) -> str:
         return self._mode
 
+    @property
+    def project_depends_on(self) -> Optional[List[List[Dict[str, str]]]]:
```
brendongo, line 151: Can you add docstring here on what this returns

semgrep/semgrep/rule.py
```diff
@@ -189,6 +204,28 @@ def full_hash(self) -> str:
             json.dumps(self._raw, sort_keys=True).encode()
         ).hexdigest()
 
+    @property
+    def has_runable_semgrep_rules(self) -> bool:
```
brendongo: What does `runable` mean here?  A rule is runable if it contains non-project-depends-on patterns? Maybe `should_run_on_semgrep_core` is more understandable?

semgrep/semgrep/dependency_aware_rule.py
```diff
@@ -0,0 +1,108 @@
...
+            )
+
+
+def run_dependency_aware_rule(
+    matches: List[RuleMatch],
+    rule: Rule,
+    targets: List[Path],
+) -> Tuple[List[RuleMatch], List[SemgrepError]]:
+    """
+    Run a dependency aware rule.
```
brendongo: More detail would be helpful. Maybe would be more helpful to mention takes Matches and rule and filters out the rules that don't match the depends on + what behavior is if rule doesnt have depends-on

semgrep/semgrep/dependency_aware_rule.py
```diff
@@ -0,0 +1,108 @@
...
+            if namespace is None:
+                raise SemgrepError(f"project-depends-on is missing `namespace`")
+            pm = PACKAGE_MANAGER_MAP.get(namespace)
+            if pm is None:
+                raise SemgrepError(
+                    f"unknown package namespace: {namespace}, only {list(PACKAGE_MANAGER_MAP.keys())} are supported"
+                )
+            package_name = entry.get("package")
+            if package_name is None:
+                raise SemgrepError(f"project-depends-on is missing `package`")
```
brendongo, line 47: Maybe worth making `project-depends-on` a constant since we're starting with `r2c-internal-only-project-depends-on`?

semgrep/semgrep/dependency_aware_rule.py
```diff
@@ -0,0 +1,108 @@
...
+
+PACKAGE_MANAGER_MAP = {
+    "pypi": PackageManagers.PYPI,
+    "npm": PackageManagers.NPM,
+}
+
+
+def parse_depends_on_restrictions(
+    entries: List[List[Dict[str, str]]]
+) -> Generator[ProjectDependsOnEntry, None, None]:
```
brendongo, line 29: What is a restriction?

brendongo (approved):
In general could use more comments to help maintainability

semgrep/semgrep/dependency_aware_rule.py
ievans: I think it's ok for now; there's not an existing source of other constants to put it with, since this is destined for the core eventually anyways

PR #4360, opened 2021-12-02
IL: Elaborate the different kinds of origins

IagoAbal:
Up until now we could not assume that an IL expression was semantically
equivalent to its `eorig`. This is useful if we want to annotate Generic
expressions with their constant values (via `e_constness`), or to do
symbolic propagation.

test plan:
1. make test
2. semgrep-core -lang go -dump_il tests/go/dots_for.go
     ^ intermediate expressions generated for i++ now use Related-orig
        making it clear that they don't have a 1-to-1 correspondence with
        some other expression in the Generic sources.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -88,11 +88,11 @@ let log_fixme kind gany =
       log_error opt_tok "Impossible happened during AST-to-IL translation"
 
 let fixme_exp ?partial kind gany eorig =
-  log_fixme kind (G.E eorig);
+  log_fixme kind (any_of_orig eorig);
   { e = FixmeExp (kind, gany, partial); eorig }
 
 let fixme_instr kind gany eorig =
-  log_fixme kind (G.E eorig);
+  log_fixme kind (any_of_orig eorig);
   { i = FixmeInstr (kind, gany); iorig = eorig }
 
 let fixme_stmt kind gany =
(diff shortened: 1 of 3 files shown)


semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -164,23 +164,23 @@ let add_instr env instr = Common.push (mk_s (Instr instr)) env.stmts
...
-let mk_aux_var ?var env tok exp =
+let mk_aux_var ?str env tok exp =
   match exp.e with
   | Fetch ({ base = Var var; offset = NoOffset; _ } as lval) -> (var, lval)
   | _ ->
-      let var = fresh_var ?var env tok in
+      let var = fresh_var ?str env tok in
       let lval = lval_of_base (Var var) in
-      add_instr env (mk_i (Assign (lval, exp)) exp.eorig);
+      add_instr env (mk_i (Assign (lval, exp)) NoOrig);
```
aryx, line 173: not even a related_tok tok here?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -1093,39 +1110,40 @@ and switch_expr_and_cases_to_exp env tok switch_expr_orig switch_expr cases =
...
+  (ss, { e = Operator ((Or, tok), es); eorig = SameAs switch_expr_orig })
 
 and cases_to_exp env tok cases =
   (* If we have no scrutinee, the cases are boolean expressions, so we Or them together *)
   let ss, es =
     List.fold_left
       (fun (ss, es) -> function
-        | G.Case (_, G.PatLiteral l) ->
-            ( ss,
-              (* TODO: seems bad to make an artificial eorig, but seems to be nothing to use  *)
```
aryx, line 1127: nice that this PR fixes those TODOs.

aryx (approved):
I don't fully understand when to use related_tok and NoOrig but this PR seems
to go in the right direction so let's merge it!

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: I will add a comment to IL.ml on how to use `Related` and `NoOrig`, although I don't know what are the right guidelines yet. In this particular case I put a `NoOrig` because I know that the RHS of the assignment already captures the correct `eorig`. If there was no subexpression carrying the origin then I would have added a `related_tok`. I try to avoid too many `related_exp` because it makes `-dump_il` harder to read. Although `related_tok` is lighter in that sense. In this case adding a `related_tok` would be harmless but in general those `tok` will be used by taint-tracking to decide whether something is a source/sanitized/sink.

PR #4364, opened 2021-12-02
Run dummy benchmarks on latest instead of develop

emjin:
We run the dummy benchmarks to make sure semgrep didn't break anything required for the full benchmark suite. Previously, this was done in CircleCI, but there doesn't appear to be a nice way to run on latest with CircleCI. This PR moves the dummy benchmark test run into github actions.

Test plan:
See the run from the `Download artifacts` commit. Since a modification was made to the benchmarks to expect a nonexistent flag, this should fail (and indeed does).

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,18 +18,6 @@ jobs:
   #    semgrep and whatever utilities we added.
   #
   # We use B out of necessity.
-  #
-  # Go through the same steps as the real benchmarks but quickly.
-  dummy-benchmarks-latest:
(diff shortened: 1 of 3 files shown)


brendongo (approved):

aryx:
What was the bug in your previous PR? Why it could not find aemgrep?

emjin:
> What was the bug in your previous PR? Why it could not find aemgrep?

I skipped the download artifacts step by accident

PR #4366, opened 2021-12-02
fix(tests): make test_metrics less flakey

brendongo:
Looks like in CI sometimes the tests running in parallel cause sending the metrics to fail. Intention of the tests is that the code path is going to send metrics.

Instead of testing that metrics were sent and we got 200 back in all the ways metrics should/should not be sent, just test for the fact that we tried to send metrics.

Have a single path for tests that check metrics actually respond with 200. This test could still be flakey but instead of having ~10 tests that rely on network we now just have 1.

Also instead of using r/all only use a single rule to test metrics since the actual rules that run are unimportant for this test

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/tests/e2e/test_metrics.py
+++ b/semgrep/tests/e2e/test_metrics.py
@@ -82,6 +82,7 @@ def test_flags(
         options=[*options, "--debug"],
         env={"SEMGREP_USER_AGENT_APPEND": "testing", **env},
     )
+    print(output)
     assert (
         "Sent pseudonymous metrics" in output
         or "Failed to send pseudonymous metrics" in output
--- a/semgrep/tests/e2e/test_metrics.py
+++ b/semgrep/tests/e2e/test_metrics.py
@@ -52,26 +52,61 @@ def mock_config_request(monkeypatch: MonkeyPatch) -> Iterator[None]:
     "config,options,env,should_send",
     [
         ("rules/eqeq.yaml", [], {}, False),
-        ("r/a", [], {}, True),
+        ("r/python.lang.correctness.useless-eqeq.useless-eqeq", [], {}, True),
         ("rules/eqeq.yaml", ["--metrics", "auto"], {}, False),
-        ("r/a", ["--metrics", "auto"], {}, True),
+        (
+            "r/python.lang.correctness.useless-eqeq.useless-eqeq",
+            ["--metrics", "auto"],
+            {},
+            True,
+        ),
         ("rules/eqeq.yaml", ["--metrics", "on"], {}, True),
-        ("r/a", ["--metrics", "on"], {}, True),
+        (
+            "r/python.lang.correctness.useless-eqeq.useless-eqeq",
+            ["--metrics", "on"],
+            {},
+            True,
+        ),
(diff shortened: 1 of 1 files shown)


brendongo:
Going to add another test that will fail if the network request fails. I think all these cases trying to send metrics is desired test but we should have at least one test that checks that the server actually accepts the metric

semgrep/tests/e2e/test_metrics.py
```diff
@@ -52,31 +52,96 @@ def mock_config_request(monkeypatch: MonkeyPatch) -> Iterator[None]:
...
+        ),
         (
             "rules/eqeq.yaml",
             ["--metrics", "auto"],
             {"SEMGREP_SEND_METRICS": "on"},
             False,
         ),
-        ("r/a", ["--metrics", "auto"], {"SEMGREP_SEND_METRICS": "off"}, True),
+        (
+            "r/python.lang.correctness.useless-eqeq.useless-eqeq",
```
brendongo, line 105: No need to run a full ruleset just to check metrics

semgrep/tests/e2e/test_metrics.py
```diff
@@ -52,31 +52,96 @@ def mock_config_request(monkeypatch: MonkeyPatch) -> Iterator[None]:
...
+        options=[*options, "--debug"],
+        env={"SEMGREP_USER_AGENT_APPEND": "testing", **env},
+    )
+    print(output)
+    # Test that we try to send metrics. Even if it fails sending
+    assert (
+        "Sent pseudonymous metrics" in output
+        or "Failed to send pseudonymous metrics" in output
+        if should_send
+        else "Sent pseudonymous metrics" not in output
```
emjin, line 129: Might as well also test "Failed to send pseudonymous metrics" not in output if not should_send. A user would probably be concerned if they had `--metrics off` and that message.

semgrep/tests/e2e/test_metrics.py
```diff
@@ -52,31 +52,96 @@ def mock_config_request(monkeypatch: MonkeyPatch) -> Iterator[None]:
...
+    "config,options,env,should_send",
+    [
+        ("rules/eqeq.yaml", [], {"SEMGREP_SEND_METRICS": "on"}, True),
+    ],
+)
+def test_flags_actual_send(
+    run_semgrep_in_tmp, mock_config_request, config, options, env, should_send
+):
+    """
+    Test that the server for metrics sends back success
```
emjin, line 143: Just to confirm, this could still fail, but will likely be less flakey because the rulesets are shorter? From the PR, I got the idea that you were going to run only one of the options, but it looks like that didn't change. (Sorry if this comment is too soon I'm excited for this PR)

brendongo: Yep. Updated PR body

semgrep/tests/e2e/test_metrics.py
brendongo: Hmm good call. Both are mutually exclusive right now but might as well.

aryx:
Got another failure, let's merge this!

aryx (approved):

PR #4373, opened 2021-12-03
Track values of all user flags in metrics

mmcqd:
This PR add a `user_flags` field to our metrics, which maps each flag to its value. This provides the value of every single flag, even if the user did not explicitly pass it.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -533,6 +533,10 @@ def cli(
             version_check()
         return
 
+    # Creates a dict mapping all local variables to their arugments
+    # Because no other variables have been defined yet, this contains only the arguments to the function
+    user_flags = locals()
+
     # To keep version runtime fast, we defer non-version imports until here
     import semgrep.semgrep_main
     import semgrep.test
@@ -551,6 +555,8 @@ def cli(
     target_sequence: Sequence[str] = list(target) if target else [os.curdir]
 
     metric_manager.configure(metrics, metrics_legacy)
+    if metric_manager.is_enabled():
+        metric_manager.set_user_flags(user_flags)
 
     if include and exclude:
         logger.warning(
--- a/semgrep/semgrep/metric_manager.py
+++ b/semgrep/semgrep/metric_manager.py
@@ -50,6 +50,7 @@ class _MetricManager:
         self._file_stats: List[Dict[str, Any]] = []
         self._rule_stats: List[Dict[str, Any]] = []
         self._rules_with_findings: Mapping[str, int] = {}
+        self._user_flags: Dict[str, Any] = {}
 
         self._send_metrics: MetricsState = MetricsState.OFF
         self._using_server = False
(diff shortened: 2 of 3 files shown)


semgrep/semgrep/cli.py
```diff
@@ -533,6 +533,10 @@ def cli(
             version_check()
         return
 
+    # Creates a dict mapping all local variables to their arugments
+    # Because no other variables have been defined yet, this contains only the arguments to the function
+    user_flags = locals()
```
brendongo, line 538: I think we're more interested in what user flags were used and less what values they eventually get in run time. So we might want to collect this info one level up?

brendongo: We probably also don't care about what targets user is running on?

semgrep/semgrep/metric_manager.py
```diff
@@ -151,6 +152,12 @@ def set_rules_with_findings(
     ) -> None:
         self._rules_with_findings = {r.full_hash: len(f) for r, f in findings.items()}
 
+    def set_user_flags(self, user_flags: Dict[str, Any]) -> None:
```
brendongo, line 155: If possible can we not have metric_manager store and then send arbitrary dict. It's probably less error prone to be more explicit here.

mmcqd: You mean like having a metric_manager field for each flag?

semgrep/semgrep/cli.py
nbrahms: I think we really only want flags, not options with finite arity. Or at least, we want to be really careful with these, as they could contain customer-sensitive data we'd want to avoid collecting.

semgrep/semgrep/metric_manager.py
```diff
@@ -151,6 +152,12 @@ def set_rules_with_findings(
...
+        self._user_flags = user_flags
+        # Need to turn MetricState objects intro strings so they can be serialized
+        self._user_flags["metrics"] = str(self._user_flags["metrics"])
```
nbrahms, line 158: Suggestion:
```python
self._user_flags = {flag: str(value) for flag, value in self._user_flags.items()}
```

semgrep/semgrep/cli.py
nbrahms: I'd suggest using an allow-list of options we want to collect, to ensure we don't accidentally collect a sensitive option now or in the future.

nbrahms: Also, you'll need to update PRIVACY.md with the additional collected values (together with _why_ we're collecting those data)

nbrahms:
@mmcqd Can you update the PR body to explain _why_ we're collecting these additional metrics?

semgrep/semgrep/metric_manager.py
brendongo: That might be a lot. Maybe what nathan suggested above and an allow-list of options and only copying those over? (Maybe even having a typed object be the arg here with allowed fields?)

emjin: +1 to having a typed arg

emjin: For which metrics: I think you can just copy all the ones we currently have. I don't think any of them are sensitive, as long as we only store whether they are used and not what argument they have

PR #4375, opened 2021-12-04
Add tests for filter-irrelevant rules

emjin:
Modify the irrelevant-rules test to allow files with all extensions to be filtered out. Add tests to ensure files are correctly filtered or not filtered out, including a few examples that have previously caught our mistakes!

Note: `metavariable-pattern` and `metavariable-regex` are currently not covered by Analyze_rule. Example TODO test:

Rule:
```yaml
rules:
- id: hello_todo
  patterns:
  - pattern: $X
  - metavariable-pattern:
      metavariable: $X
      pattern: "string"
  languages: [py]
  severity: WARNING
  message: "todo"
```

Code:
```python
hello
```

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/tests/OTHER/irrelevant_rules/pattern-and-one-general.js
```diff
@@ -0,0 +1,13 @@
+// https://raw.githubusercontent.com/lodash/lodash/master/random.js
+
+require("child")
```
aryx, line 4: maybe a comment here saying note that this is not child___process, which is why this file should not
be processed at all by the rule

aryx: or maybe add the comment in the rule

semgrep-core/tests/OTHER/irrelevant_rules/taint-general-sink.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
```
aryx, line 4: maybe a comment saying even though the sink is too general, at least the source is concrete so we can filter some files.

semgrep-core/tests/OTHER/rules/pattern-either-one-general.yaml
```diff
@@ -0,0 +1,21 @@
+rules:
+- id: python.django.best-practice.upsell_django_environ.use-django-environ
+  patterns:
+  - pattern-not: |
+      ...
+      import environ
+      ...
+  - pattern-either:
+    - pattern: |
+        ...
```
aryx, line 11: seems a very general pattern here.

semgrep-core/tests/OTHER/rules/pattern-either-one-general.py
```diff
@@ -0,0 +1,3 @@
+# tests filter-irrelevant-rules does not filter this out
```
aryx: you forgot 'that'

semgrep-core/tests/OTHER/rules/pattern-either-one-general.yaml
aryx: maybe we should add a comment here that's it's this pattern that forbid filter-irrelevant-rules to apply here?

semgrep-core/tests/bash/added_identifier.yaml
```diff
@@ -0,0 +1,7 @@
+rules:
```
aryx, line 1: why those bash tests?

semgrep-core/tests/Test.ml
```diff
@@ -510,25 +510,35 @@ let test_irrelevant_rule rule_file target_file =
...
-
-let test_irrelevant_rule_file rule_file =
-  Filename.basename rule_file, (fun () ->
-    let target_file =
-      let (d,b,_e) = Common2.dbe_of_filename rule_file in
-      (* TODO: Support other extensions, note that we don't need
-       * to parse the target files! *)
-      let candidate1 = Common2.filename_of_dbe (d,b,"py") in
+  
+let test_irrelevant_rule_file file =
```
aryx: I would have rather use target_file then now, instead of just file.

aryx (approved):

semgrep-core/tests/bash/added_identifier.yaml
emjin: I didn't see a test for the error with -filter_irrelevant_rules that Martin ran into: https://github.com/returntocorp/semgrep/pull/4085

PR #4378, opened 2021-12-06
Gracefully match "$MVAR" against a non-source string

### LINKED ISSUE #4371: Crash when metavariable-regex is applied to the result of constant propagation
**Describe the bug**
`metavariable-regex` may crash if the `metavariable` binds to the result of constant propagation, because the string literals produced by constant folding may have fake tokens.

**To Reproduce**
https://semgrep.dev/s/KXGj

**Expected behavior**
Should be possible to run a regex on a propagated value. 

IagoAbal:
Strings produced by constant folding have fake tokens, and trying
to match them with "$MVAR" patterns resulted in a crash.

Closes #4371

test plan:
make test # test included

PR checklist:
- [x] Changelog is up-to-date


emjin (dismissed):
Putting your test case into `iago-test.yaml` and `iago-test.py` and running `sc -lang python -config iago-test.yaml iago-test.py -json` returns:

```
{"matches":[],"errors":[{"rule_id":"my_pattern_id","error_type":"Internal matching error","severity":"warning","location":{"path":"emma-test.py","start":{"line":1,"col":1,"offset":0},"end":{"line":1,"col":1,"offset":0},"lines":["def test():"]},"message":"NoTokenLocation with pattern return \"$X\", NoTokenLocation with metavar $X, close location = {\"line\":6,\"col\":3,\"offset\":56}"}],"skipped":[],"stats":{"okfiles":0,"errorfiles":1}}
```

(I originally tested https://semgrep.dev/editor?version=3897942c8ba326610d46407e6e48993576a4e57f)

IagoAbal:
> Putting your test case into `iago-test.yaml` and `iago-test.py` and running `sc -lang python -config iago-test.yaml iago-test.py -json` returns:
> 
> ```
> {"matches":[],"errors":[{"rule_id":"my_pattern_id","error_type":"Internal matching error","severity":"warning","location":{"path":"emma-test.py","start":{"line":1,"col":1,"offset":0},"end":{"line":1,"col":1,"offset":0},"lines":["def test():"]},"message":"NoTokenLocation with pattern return \"$X\", NoTokenLocation with metavar $X, close location = {\"line\":6,\"col\":3,\"offset\":56}"}],"skipped":[],"stats":{"okfiles":0,"errorfiles":1}}
> ```

I can't reproduce. Can you attach the exact files that you are running?

> (I originally tested https://semgrep.dev/editor?version=3897942c8ba326610d46407e6e48993576a4e57f)

Could you clarify the connection with the example above?

IagoAbal:
Oh, very sorry, I knew I had to add `-json` but nevertheless I forgot to add the `-json` flag. 😄  Well spotted. This is a more general problem. With `-json` we compute the location info for metavariables, but in this case there is no such location info because the string is the result of constant folding. And I think the only sensible thing to do is to not add any location info.

IagoAbal:
Fixing that in semgrep-core is not very difficult, now the problem is to update Semgrep and make sure it doesn't interact badly with other features like autofix.

Perhaps the simplest solution for now is to restrict `$MVAR`s to only match source strings.

IagoAbal:
Tried the simple approach of not allowing the match, but we can't do that now since there are rules that rely on it. `Normalize_generic` just uses the location of the first string in the concatenation... I wonder how does this interact with autofix. I can do the same. 🤷‍♂️

IagoAbal:
Fixed. I don't know how this may interact with Autofix or something else given that the location info that we are returning is wrong. But it's what we were already doing in `Normalize_generic` anyways.

emjin (approved):
Sounds like this will still be broken for some cases but is an improvement over what we previously had :)

Debated whether or not to approve because I wanted to see what this looked like in the playground with the fake range but my internet is too poor. This is what I had done with `(I originally tested https://semgrep.dev/editor?version=3897942c8ba326610d46407e6e48993576a4e57f)`. I went to the docker check that ran for your last commit, found the docker image, and ran your example in the playground using that as the version.

emjin:
Potentially broken: autofix, metavars in messages

IagoAbal:
The matches have the correct range regardless; but the metavariables may not, if their value was computed by constant folding. That may break thing, for sure. However this is not a problem introduced by this PR, it has been working that way at least since we introduced `"$MVAR"` patterns. Try matching `return "$X"` against `return "foo" + "bar"` and see what range is assigned to `$X`... We should definitely do something about `$MVAR`s having "misleading" location info, but that is out of the scope of this PR, here we just want to fix a crash.

Issue #4374, opened 2021-12-04
Metavariable-pattern broken when applying to yaml


enncoded:
**Describe the bug**
`metavariable-pattern` within a rule for yaml causes a Semgrep Core ERROR.

**To Reproduce**
https://semgrep.dev/s/lejliu:yaml-metavariable-pattern-bug

Here is the relevant error: 
![Screen Shot 2021-12-03 at 17 06 09](https://user-images.githubusercontent.com/14846866/144690799-7b352bb8-ab57-49a0-b18d-8c4a282a07b0.png)

**Expected behavior**
I expected `$X` to match `$Y`

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [X] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

aryx:
@emjin is is related to your recent PR? I don't see it referenced.

emjin:
Thanks for reminding me of this! I meant to link it

PR #4380, opened 2021-12-06
Fix YAML parsing off by one

emjin:
We had previously identified that the end index for a yaml array would be off by one compared to how semgrep expects tokens. However, we only adjusted this for the line and column.

Test plan: make test

PR checklist:
- [x] Changelog is up-to-date


mmcqd (approved):

emjin:
Closed https://github.com/returntocorp/semgrep/issues/4374

Issue #1301, opened 2020-07-17
[Ruby] Typed metavariable handling new, literal arrays


dlukeomalley:
https://semgrep.dev/s/ExwN?version=develop

 Doesn’t catch the Array.new from either casting of syntax

aryx:
Not sure what you want, but your semgrep.live link above has the wrong pattern.
You need $ARR.instance_eval(...) as in https://semgrep.dev/s/K39j/?version=develop

aryx:
if you're talking about the alias of 'foo' created when called Array.new(), which might be an alias for 'Array' (not super familiar with Ruby), but yes it's not handled by semgrep. You need $ARR.instance_eval or we need some kind of typed metavariables for Ruby to handle this case.

aryx:
But the title of this issue is confusing. Maybe "support typed metavariable for Ruby" or "handle type aliases".

dlukeomalley:
> Not sure what you want, but your semgrep.live link above has the wrong pattern.

It's the right pattern for the behavior I _want_ Semgrep to have, but is not right in the context of how Semgrep actually works. 

> we need some kind of typed metavariables

Thanks for the clarification. It makes sense to me now that what I want to query for is "instance_eval calls on objects of type Array", which isn't supported.

@nobletrout - Does it suffice for you to use `$ARR.instance_eval` for now?

aryx:
A nice example of "complex symbolic expression" constant propagation (or typed metavariable in Ruby).

aryx:
keep it

IagoAbal:
I think this is more about typing. To make those two examples work maybe it's not too difficult, but in general we would need a type inferrer for the generic AST. I guess we focus on those two simple cases for now?

aryx:
yep, just make the basic examples work should be enough for now. I hope the work on symbolic expression propagation
can get those for free.

aryx:
:wave:

IagoAbal:
Another thing that would be great to get done in Q4.

emjin:
+1

Sjord:
Links were broken, here are working links:

* https://semgrep.dev/s/ExwN?version=develop
* https://semgrep.dev/s/K39j/?version=develop

The title indicates Ruby, but this would be useful in any object-oriented language. For example, in PHP I want to match calls to `DOMDocument::load(...)`, which would look something like this:

```
$doc = new DOMDocument();
$doc->load('book.xml');
```

[...]

PR #4388, opened 2021-12-07
const-prop: Propagate simple symbolic expresions

### LINKED ISSUE #2783: Expression equivalence under variable definition substitution
Say I want to check whether a project is checking for string equality under lowercase. The following rule seems like it should do the job:

```
pattern-either:
  - pattern: $X.lower() == $Y.lower()
  - pattern: $X.lower() != $Y.lower()
  - pattern: $X.upper() == $Y.upper()
  - pattern: $X.upper() != $Y.upper()
```

However, this will only match those exact expressions. Even a simple semantically equivalent case such as the following would remain undetected:

```python
a = "FOO"
a = a.lower()

if a != "BAR".lower():
    pass
```

Ideally, semgrep should be able to match the above snippet with the previously stated rules by concluding that `a` is equivalent to the expression `a.lower()` since there were no re-assignments or mutations of `a` in between the assignment and the comparison.

Real world cases are likely to get more complicated, with at least some statements in between, which might complicate matters due to mutability concerns in some languages. It still seems like it could work in a significant number of cases, though (for instance, with `lower()` in Python as above, since Python strings are immutable).

Is implementing something like this possible with the current state of semgrep's engine? Or am I perhaps missing an easy workaround? 

## Alternatives

Of course, it would be possible to "unroll" this manually, by wrapping everything in a large `pattern-either` clause and then listing variants of the above patterns with and without an appropriate `pattern-inside` that would detect this simple case, but this gets very cumbersome quickly.

### LINKED ISSUE #2859: Automatically detect deeply-chained calls on objects through assignment
## Is your feature request related to a problem? Please describe.
As described on [Slack](https://r2c-community.slack.com/archives/CK86BJ5DW/p1617137770088500), the pattern
```yaml
- pattern: pandas.DataFrame(...).index.set_value(...)
```
works great on code where this entire chain is present, but does not work if it's broken by assigning to variables, e.g., the above rule can't catch
```python
df = pd.DataFrame(x)
ix = df.index
ix.set_value(a, b, c)
```

## Describe the solution you'd like
It would be wonderful if the pattern `pandas.DataFrame(...).index.set_value(...)` can recognize when the call chain is broken by simple assignment, perhaps in line with #2783?

## Describe alternatives you've considered
The current workaround would be *two* additional rules that use separate `pattern-inside`, specifically,
```yaml
- pattern-inside: |
    $DF = pandas.DataFrame(...)
    ...
    $DF.index.set_value(...)
```
and
```yaml
- pattern-inside: |
    $DF = pandas.DataFrame(...)
    ...
    $IX = df.index
    ...
    $IX.set_value(...)
```
As the length of the chain expands, combinatorially-more separate rules are required 😞.

IagoAbal:
This should allow users to write simpler yet still powerful rules, by
letting the dataflow engine keep track of intermediate assignments.

Closes #2783
Closes #2859

test plan:
make test # tests included

PR checklist:
- [x] Changelog is up-to-date


aryx:
Woohoo!

aryx:
This might fix one of the first semgrep issue from @dlukeomalley :) https://github.com/returntocorp/semgrep/issues/1301

aryx:
There's a few regressions on semgrep-rules, did you see them?

IagoAbal:
> There's a few regressions on semgrep-rules, did you see them?

I took a look at some, but not at all of them. As far as I could tell they are essentially extra matches, expected as a "side effect" of symbolic propagation. For example, if we are looking for `foo.bar()`, and in the code we have:

```
x = foo.bar()
return x
```

With symbolic propagation you have two matches, because `return x` can be seen as `return foo.bar()`.

This is why I preferred to keep it disabled by default. We could perhaps try removing those matches by only matching a variable modulo symbolic propagation when the variable is not the top-level AST being matched against.

IagoAbal:
Ooops, I guess you meant the failing tests. I was talking about running semgrep-rules tests with symbolic propagation enabled by default. Will look into the CI failures!

IagoAbal:
Seems that, in some cases, symbolic propagation causes `Bloom_annotation.annote_program` to segfault. Looking into this.

IagoAbal:
Ooops, due to an existing issue in the constness analysis we can end up introducing a recursive `id_constness` if we find code such as:

```go
func (c *Container) logResourceUsageStats() {
        for true {
                cpuids = append(cpuids)
        }
}
```

I'll open a separate PR.

IagoAbal:
OK I fixed the cycle issue, added a test case, and also added a cycle check just in case. cc @aryx

emjin:
I would be interested in understanding the cycle issue. I looked at your PR to make constant propagation a true must analysis, but didn't get all the details.

semgrep-core/src/core/Pretty_print_AST.ml
```diff
@@ -770,6 +770,23 @@ and pattern env = function
...
 
+let ctype = function
+  | G.Cbool -> "bool"
+  | G.Cint -> "int"
+  | G.Cstr -> "str"
+  | G.Cany -> "???"
+
+let constness env = function
+  | G.NotCst -> "TOP"
+  | G.Sym e -> Printf.sprintf "sym(%s)" (expr env e)
```
emjin, line 781: Just to check: this means `constness` can contain non-constant expressions now, right? Should we rename it?

IagoAbal: > I would be interested in understanding the cycle issue. I looked at your PR to make constant propagation a true must analysis, but didn't get all the details.

There is a test case to detect if it happens again, I can comment it a bit more. But it's just a for-loop:

```go
for true {
     cpuids = append(cpuids)
}
```

In the original example `cpuids` was declared as `var cupids []int` which I presume initializes it as an empty array. But the analysis didn't catch that and it was considered simply uninitialized. And at the moment the analysis started with the assumption that all locals were `_|_` at entry point. So the analysis inferred that at the beginning of the loop `cpuids` was `append(cpuids)`, because in one branch we have `_|_` and in another we have `append(cpuids)`. So we set the `id_constness` of the second occurrence of `cpuids` to `append(cpuids)`, and that is a cycle (both `cpuids` are the exact same occurrence, so they share the same `id_constness` `ref`).

Of course this wouldn't happen if the analysis knew that `cpuids` was actually initialized to `[]`, but it can always happen something like that.

After #4410 we are actually a bit too strict and initialize all mappings so that variables are assumed non-constant (`T`). But this removes the problem with loops, `T` and `append(cpuids)` gives us `T`.

IagoAbal: > Just to check: this means `constness` can contain non-constant expressions now, right? Should we rename it?

Yeah indeed, let's do that. I had thought about calling it "value propagation".

emjin (approved):
Overall LGTM! Probably wait for Pad for any details I might have missed. Appreciate the renamings

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -546,6 +552,10 @@ and m_id_info a b =
 (* experimental! *)
 and m_expr_deep a b =
   m_deep m_expr_deep m_expr SubAST_generic.subexprs_of_expr a b
+  >||> m_with_symbolic_propagation
```
aryx: Maybe a comment on why we do it at this level. Why not in m_expr? why in m_expr_deep?

semgrep-core/tests/OTHER/rules/sym_prop_chain.py
```diff
@@ -0,0 +1,8 @@
+# https://github.com/returntocorp/semgrep/issues/2859
+import pandas as pd
+def test():
+    df = pd.DataFrame(x)
+    ix = df.index
+    # using symbolic propagation we know that `ix` is `pandas.DataFrame(x).index`
+    # ruleid: test
+    ix.set_value(a, b, c)
```
aryx, line 8: :exploding_head:

semgrep-core/tests/OTHER/rules/sym_prop_no_merge.yaml
```diff
@@ -0,0 +1,11 @@
+rules:
+- id: test
+  options:
+    symbolic_propagation: true
+  pattern-either:
+    - pattern: return a.foo
```
aryx, line 6: would be nice though that at some point we allow $X.foo to match!
That is when we do the merge of a.foo and b.foo we generalize to $X.foo on the target side!

aryx (approved):
LGTM. Looks great!
In the future it might be better though to split this PR in multiple steps, because
there's lot of trivial refactoring (renaming id_constness to id_svalue and constness to lvalue,
and also the Pretty_print_generic to Pretty_print_AST) that are distracting.

semgrep-core/src/matching/Generic_vs_generic.ml
IagoAbal: Actually it is not needed here, just in `m_expr`, I fixed it!

semgrep-core/tests/OTHER/rules/sym_prop_chain.py
IagoAbal: 😄

PR #4395, opened 2021-12-07
Set default pre-commit flags

### LINKED ISSUE #4386: Tweak pre-commit hooks to be non-noisy by default
I've been using semgrep in one of my repos for about a year, and it's been really helpful! Thanks for making this tool :)
One thing I have found annoying is that the pre-commit hook emits a lot of extra output that's not related to the actual warnings/errors - mainly timing info and version check info.
Because pre-commit passes each individual filename to semgrep, it must call semgrep multiple times, and so there are many lines of timing/version check info (once per invocation) which usually dwarfs the one or two lines of error/warning output and makes them hard to find.

pre-commit supports specifying `args` to the hook in two places:
- In the consuming repo's `.pre-commit-config.yaml`: I currently do this as a workaround
- In the repo defining the hooks' `.pre-commit-hooks.yaml`: Can you please update this file?

Specifically, can we add `args: ['--disable-version-check', '--quiet']` to https://github.com/returntocorp/semgrep/blob/558bf60c6dcb7290f1935a776bc8cb6d0719f1f4/.pre-commit-hooks.yaml#L13-L16?
https://pre-commit.com/index.html#hooks-args is the docs, 
here's an example of a different hook that sets `args`: https://github.com/pre-commit/mirrors-mypy/blob/5cf22ccb774a8be8f47dfe4c1e8c4f177c608cbf/.pre-commit-hooks.yaml#L6

Note that `pre-commit` already has a [`pre-commit autoupdate` feature](https://pre-commit.com/index.html#pre-commit-autoupdate) that my company uses regularly to update all our pre-commit hooks, which makes the version check a bit redundant.

I haven't gotten approval to contribute to this repo from my company's OSS review process yet, otherwise I would have sent a PR. Would love if one of the semgrep authors or community could fix this instead.

I am doing this in my own repo now, but it would be nice for this to be the default for all pre-commit users.
Based on https://semgrep.dev/docs/cli-usage/, I originally tried `--no-time` instead of `--quiet` to remove the timing info from the output, but it didn't do anything.

**Additional context*
Here's a screenshot that shows how much output is emitted when I've injected a single fake failure into my codebase. The first run is with the default settings, the second is adding `--disable-version-check` and `--quiet` locally.

Note I temporarily downgraded to semgrep `v0.74.0` (second-most-recent release) for this screenshot - we normally use the latest version but that doesn't emit the version check messages.

Screen Shot 2021-12-07 at 10 04 54 AM

brendongo:
Closes https://github.com/returntocorp/semgrep/issues/4386

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -14,3 +14,4 @@
   name: semgrep
   entry: semgrep
   language: python
+  args: ["--disable-version-check", "--quiet", "--skip-unknown-extensions"]


aryx (approved):
Maybe a comment on --disable-version-check about this autoupdate thing?

Also I think it would be better to put all of that under a --pre-commit flag.

aneeshusa (approved):
Thank you for the PR, I appreciate it!

Personally, I don't think a `--pre-commit` flag would be useful as I don't believe it would need to be used anywhere in the world except in this file. Folks using semgrep via pre-commit will get these flags for free from this change so they won't need to specify it. Maybe you can say more about the use case you had in mind for it?

brendongo:
Yeah I think this should be fine for now, users will never need to run with --pre-commit since they will just pull from this settings file

aryx:
Oh i see. Then maybe it is because i dont understand how this file is used by precommit. Brendon could you comment in this file or somewhere how and when this file is used by precommit? Like when you add a repo: in your local precommit file, what is going on, how it works internally.

PR #4396, opened 2021-12-08
Convert named pipes

mjambon:
This ensures target files passed directly on the command line are converted from named pipes to regular files whenever applicable.

test plan: from bash, we should get a match and this match should be reported without raising an exception:
```
$ cat foo.yml
rules:
- id: '-'
  pattern: foo
  message: yay
  languages:
  - python
  severity: ERROR

$ semgrep-core -lang python -config foo.yml <(echo 'foo')
/tmp/semgrep-core-9d4b6d-63:1 with rule -
 foo
```

Previous behavior:
```
$ semgrep-core -lang python -config foo.yml <(echo 'foo')
/dev/fd/63:1:0: Fatal error: (Sys_error "Illegal seek")
Raised at Parse_target.run in file "src/parsing/Parse_target.ml", line 185, characters 17-26
Called from Parse_target.parse_and_resolve_name_use_pfff_or_treesitter in file "src/parsing/Parse_target.ml", line 349, characters 30-60
...
```

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -631,12 +631,8 @@ let main () =
   (* does side effect on many global flags *)
   let args = Common.parse_options (options ()) usage_msg (Array.of_list argv) in
   let args =
-    if !target_file = "" then
-      Common.map S.replace_named_pipe_by_regular_file args
-    else
-      (* no conversion from named pipes in this mode so we don't have to worry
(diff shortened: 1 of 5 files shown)


aryx:
What is your test plan?
What was not working before?

mjambon:
updated PR description.

PR #4401, opened 2021-12-08
Update output for timing to use profiling_times

emjin:
Also handle errors properly when outputting

PR checklist:
- [x] Changelog is up-to-date


brendongo (approved):
I'm worried this will slowdown semgrep-action since that runs with `--time`

semgrep/semgrep/formatter/text.py
```diff
@@ -164,8 +184,21 @@ def _build_text_timing_output(
...
+        total_time = time_data["profiling_times"]["total_time"]
+        config_time = time_data["profiling_times"]["config_time"]
+        core_time = time_data["profiling_times"]["core_time"]
+        ignores_time = time_data["profiling_times"]["ignores_time"]
+
+        yield f"\nSemgrep timing summary:"
+
+        yield f"Total time: {total_time:.4f} Config time: {config_time:.4f} Core time: {core_time:.4f} Ignores time: {ignores_time:.4f}"
+
+        if run_errors > 0 or parse_errors > 0 or match_errors > 0:
```
brendongo: Users will also be confused with new `"errors"` being talked about in their output

PR #4402, opened 2021-12-08
Stop reporting absolute file paths when .semgrepignore is present

mmcqd:
This PR fixes a bug where the presence of a `.semgrepignore` file would cause target file paths to be reported as absolute instead of relative. It also adds many more tests for the `.semgrepignore` feature.

PR checklist:
- [x] Changelog is up-to-date


semgrep/semgrep/ignores.py
```diff
@@ -81,9 +81,12 @@ def _survives(self, path: Path) -> bool:
         return True
 
     def filter_paths(self, paths: Iterable[Path]) -> FrozenSet[Path]:
-        abspaths = (p.absolute() for p in paths if p.exists())
         return frozenset(
-            p for p in abspaths if self._survives(p) or p.samefile(self.base_path)
+            p
+            for p in paths
+            if p.exists()
+            and self._survives(p.absolute())
```
IagoAbal: If I'm not mistaken this is parsed as `(A and B) or C`, but shouldn't it be `A and (B or C)` ?

mmcqd: you're right, fixed.

brendongo (approved):

PR #4404, opened 2021-12-08
Go: Allow newline after ellipsis standing for key-value pair list

### LINKED ISSUE #4392: Newlines sometimes break patterns in Go
**Describe the bug**
The pattern `&http.Transport{...}` works but 
```
&http.Transport{
  ...
}
```
produces an error

**To Reproduce**
https://semgrep.dev/s/KXj9

**Expected behavior**
The two patterns should be equivalent

mmcqd:
Fixes #4392

Previously automatic semicolon insertion would cause the pattern
```
&http.Transport{
  ...
}
```
to become
```
&http.Transport{
  ...;
}
```
which caused a parse error. This PR adds a special case to the Go parser to handle this scenario.

PR checklist:
- [x] Changelog is up-to-date


emjin (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+## Fixed
+- Go: fixed bug where using an ellipsis to stand for a list of key-value pairs w  would sometimes cause a parse error
```
IagoAbal, line 8: s/pairs w  would/pairs would/

aryx:
Regarding the tests, I'm not sure 'make test' actually use those tests under tests/go/semgrep/.

aryx:
Ah yes it does. It's test by Unit_typing_generic.ml (not sure if it's the right place for such parsing-pattern tests).

PR #4408, opened 2021-12-09
Add a correctness check to the perf comparison

emjin:
We run benchmarks latest against an older version. While we're at it, we may as well make sure the results are correct.
    
With the `--save-findings-to` flag, we can now save findings to a file. The `compare-bench-findings` script takes the findings file as an argument and compares the findings against a snapshot. If there are any that are different, it will error.
    
To update the snapshot, run `./run-benchmarks --small-only --std-only --snapshot-update`.
    
With these correctness checks, we can add a taint benchmarking pass that also serves as real world taint tests
    
Test plan:
    
The script is run in two stages. First, get findings:
```
./run-benchmarks --small-only --std-only --save-findings-to findings_temp.js
```

Then run:
```    
../scripts/compare-bench-findings findings_temp.json
```
    
You should get:
```
Reading snapshots/benchmark_findings.json
Reading findings.json
```
    
Next, to test that it actually detects differences, open `findings.json` and change the value of `start` for two of the findings. Run `compare-bench-findings` again. You should get two errors.

PR checklist:
- [x] Changelog is up-to-date


IagoAbal:
> To update the snapshot, run `./run-benchmarks --small-only --std-only --snapshot-update`.

IMO we should have a script or `make` target for this, like `regenerate-tests` in semgrep/Makefile.

aryx (approved):

PR #4409, opened 2021-12-09
Scala: Translate definitions using patterns to generic AST, add PatEllipsis

### LINKED ISSUE #4346: Scala annotations sometimes not matched
**Describe the bug**
Scala annotations that match a pattern are not detected by semgrep.

**To Reproduce**
https://semgrep.dev/s/zxBk

**Expected behavior**
All annotations should be matched 

mmcqd:
Closes #4346
This PR does several things, all of which were needed to make the added tests pass. If it needs to be broken up into multiple PRs for simplicity that can be done.
Things this PR does:
* Translate Scala definitions like `val List(x,y,z) = List(1,2,3)` or `val (x,y) = (1,2)` to the generic AST
* Add `PatEllipsis` to the Scala AST, allowing for semgrep patterns like `val List(...) = ...`
* Add a case to the generic AST matcher for `PatConstructor(name,[PatEllipsis _])`, which ought to match any pat constructor with the same name and any number or arguments, but previously did not

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -2158,6 +2158,9 @@ and m_pattern a b =
       m_ident a1 b1 >>= fun () -> m_id_info a2 b2
   | G.PatLiteral a1, B.PatLiteral b1 -> m_literal a1 b1
   | G.PatType a1, B.PatType b1 -> m_type_ a1 b1
+  | G.PatConstructor (a1, [ G.PatEllipsis _ ]), B.PatConstructor (b1, _)
```
aryx, line 2161: First, if you look at the comment at the top of this file (maybe we should improve it), the pattern is the first
parameter (the axxx) and the target source code the second (the bxxxx), so PatEllipsis can't appear in the bxxx.

Second, this will handle just List(...), but what about List(a, ..., b), or List(a, ...), etc. You probably want to use m_list_with_dots and imitate what we do for handling container elements (see m_container_ordered_elements).

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -2158,6 +2158,9 @@ and m_pattern a b =
...
+  | G.PatConstructor (a1, _), B.PatConstructor (b1, [ B.PatEllipsis _ ]) ->
+      m_name a1 b1
   | G.PatConstructor (a1, a2), B.PatConstructor (b1, b2) ->
       m_name a1 b1 >>= fun () -> (m_list m_pattern) a2 b2
```
aryx, line 2165: so really you want to replace (m_list m_pattern) with an m_constructor_arguments
that will use m_list_with_dots and m_pattern.

aryx (changes requested):

aryx:
Ok I'll merge this otherwise I get conflicts in pfff for another PR, but the code should be fixed in another PR.

PR #4410, opened 2021-12-09
dataflow: Make const-prop a true must-analysis

IagoAbal:
And only use reachable nodes when computing the input mapping of a node.

test plan:
make test

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/analyzing/Dataflow_constness.ml
```diff
@@ -280,22 +280,27 @@ and eval_concat env args =
 (* Transfer *)
 (*****************************************************************************)
 
-(* FIXME: This takes "Bottom" as the default constness of a variable.
```
aryx, line 283: I would have removed the FIXME but kept most of the comment, especially the example.

semgrep-core/src/analyzing/Dataflow_constness.ml
```diff
@@ -280,22 +280,27 @@ and eval_concat env args =
...
+  let node = flow.graph#nodes#assoc ni in
+  match node.F.n with
+  | Enter -> enter_env
+  | _else -> (
+      let pred_envs =
+        CFG.predecessors flow ni
+        |> Common.map (fun (pi, _) -> mapping.(pi).D.out_env)
+      in
+      match pred_envs with
+      | [] -> VarMap.empty
```
aryx, line 330: why can't we simplify and just call fold_left with VarMap.empty as the accumulator?
Why we need the 3 cases? union_env will not work as intended then?

semgrep-core/src/core/il/CFG.ml
```diff
@@ -15,12 +15,34 @@
...
+        (graph#successors nodei)#fold
+          (fun s (ni, _) -> NodeiSet.add ni s)
+          NodeiSet.empty
+      in
+      NodeiSet.fold aux succs seen
+  in
+  { graph; entry; reachable = aux entry NodeiSet.empty }
+
+(* Predecessors of a node (that can be reached from the entry node). *)
+let predecessors cfg nodei : (nodei * 'node) list =
```
aryx, line 46: I would then add a successor function too, just for symmetry.

semgrep-core/tests/python/constprop_dataflow.py
```diff
@@ -1,28 +1,42 @@
...
+        # ERROR:
+        foo(a)
     # ERROR:
-    v = a
+    foo(a)
 
+def test3(c):
+    if c:
+        x = "hi"
+    # Must analysis, and `x`` is only coonstant in one path!
```
aryx: typo, coonstant

aryx (approved):

semgrep-core/src/analyzing/Dataflow_constness.ml
IagoAbal: Good point, will do.

PR #4413, opened 2021-12-09
Allow arbitary ellipsis patterns in pattern constructors

mmcqd:
Now semgrep patterns like `val List(...,x,...) = ...` will correctly match against definitions like `val List(a,x,b) = List(1,2,3)`

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -2158,11 +2158,13 @@ and m_pattern a b =
       m_ident a1 b1 >>= fun () -> m_id_info a2 b2
   | G.PatLiteral a1, B.PatLiteral b1 -> m_literal a1 b1
   | G.PatType a1, B.PatType b1 -> m_type_ a1 b1
-  | G.PatConstructor (a1, [ G.PatEllipsis _ ]), B.PatConstructor (b1, _)
-  | G.PatConstructor (a1, _), B.PatConstructor (b1, [ B.PatEllipsis _ ]) ->
-      m_name a1 b1
   | G.PatConstructor (a1, a2), B.PatConstructor (b1, b2) ->
-      m_name a1 b1 >>= fun () -> (m_list m_pattern) a2 b2
+      (* List(...) ought to match List(), so less_is_ok *)
```
aryx: that's not what less_is_ok mean. It means that if the pattern is empty, e.g., List(), is it ok that it
matches a non-empty list like List(a,b). Is it ok to have less elements in the pattern than in the target.
Here I don't think so (in fact most calls to m_list_with_dots have less_is_ok:false), so less_is_ok should
be false.

aryx: less_is_ok is a bad name for one of the principle of Semgrep which is that it's ok sometimes to not specify everything.
For example 'int a;' will also match 'static int a;' or 'final int a;' The absence of thing is ok (I'm open to a better name for this principle).

aryx (approved):

PR #4415, opened 2021-12-10
Add hack to allow matches on Unicode data with -fast

### LINKED ISSUE #4336: Python: Can't match unicode snowman literals in strings
**Describe the bug**

I'm trying to match a string, in python, which contains a literal snowman character (`☃`). I can't find a way to do so.

**To Reproduce**

https://semgrep.dev/s/craigds:unicode-snowman

**Expected behavior**
I expected to just be able to use the string itself

```
semgrep --lang=python --pattern '"Test ☃"'
```

It didn't match. Neither did any of the other things I tried, e.g.:

* `"Test \x{FE0F}"`
* `"Test \u2603"`

**Environment**

* semgrep 0.70.0 via homebrew on MacOS 11.6
* Doesn't work on semgrep.dev either though, see link above

mjambon:
This "fixes" https://github.com/returntocorp/semgrep/issues/4336 in the sense that the "filter irrelevant rules" optimization no longer skips files that it shouldn't. The core issue is that we use a hack that replaces any non-ascii byte by the character `Z`. Because of this, we have false positives, but that's an older problem that will be fixed another day.

I added a test that will prevent regressions.

Closes #4336

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/optimizing/Analyze_rule.ml
```diff
@@ -457,15 +457,31 @@ let eval_and p (And xs) =
          if not v then logger#trace "this Or failed: %s" (Common.dump (Or xs));
          v)
 
+(*
+   At least for some languages, '"🚀"' is replaced by 'ZZZ' in the pattern
+   but not in the raw target file that we inspect as part of this optimization.
+   Because of this, we don't require a match in the target file is the
```
emjin:
```suggestion
   Because of this, we don't require a match in the target file if the
```

semgrep-core/tests/OTHER/rules/unicode_string.py
```diff
@@ -0,0 +1,10 @@
+# Both 🎅 and 🐻 are UTF-8-encoded using 3 bytes.
```
emjin, line 1: Might be good to also have a test case that a keyword like "buzz" doesn't also set this off.

emjin (approved):
Whoo!!

Issue #4419, opened 2021-12-13
[PHP] "$MVAR" patterns not working


IagoAbal:
Reporting on behalf of Kurt Boberg (https://r2c-community.slack.com/archives/C02DM47JHS4/p1639265066009600).

**Describe the bug**

PHP does not support `"$M"` pattens, these are parsed as interpolated strings.

PHP support is still experimental so this may be seen as a feature request.

**To Reproduce**

https://semgrep.dev/s/GyR1

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

PR #4421, opened 2021-12-13
metavariable-comparison: Fix order of re.match arguments

IagoAbal:
The pattern is the first argument, the string is the second!
See https://docs.python.org/3/library/re.html#re.match

Hoping that this won't break many rules since I expect most users will
be using `metavariable-regex` for this.

Related to #4419 (potential workaround)

test plan:
make test # tests updated

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Eval_generic.ml
+++ b/semgrep-core/src/engine/Eval_generic.ml
@@ -169,13 +169,13 @@ let rec eval env code =
                 FN (Id (("match", _), _)) );
           _;
         },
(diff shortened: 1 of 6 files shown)


emjin:
Do we still need to expose this? If few people are using it, maybe we can deprecate instead?

IagoAbal:
> Do we still need to expose this? If few people are using it, maybe we can deprecate instead?

That's a good discussion to have, although in the meantime I would just fix it. We could maybe take out the changelog entry if we don't want to give it publicity.

emjin (approved):

IagoAbal:
I did delete the changelog entry so we avoid somewhat encouraging users to use this.

PR #4423, opened 2021-12-14
Add a default set of semgrepignore patterns

mmcqd:
Now the file `semgrep/templates/.semgrepignore` will be used as a set of patterns to ignore by default if no `.semgrepignore` file is found in the directory where Semgrep is being run. This PR also moves the `.semgrepignore` filtering code to the `filtered_files` function, so that it is cached on a per language basis.

PR checklist:
- [x] Changelog is up-to-date


mmcqd:
Seems like this causes some significant slowdowns.

mmcqd:
Failing because the `templates` dir doesn't get included in the docker image. Could be solved by just storing the default `.semgrepignore` as a string?

emjin:
hm these slowdowns are worse than I realized. Most of these repos are pretty small. Can you time some larger repos?

emjin:
Whoo the benchmark tests work, always good to see. If you can't figure out in like 5 minutes how to fix the benchmark tests, ping me to document this better.

emjin (approved):

semgrep/semgrep/semgrep_main.py
```diff
@@ -63,21 +63,21 @@ def notify_user_of_work(
...
-            )
+    if not semgrepignore_path.is_file():
+        logger.verbose("using default .semgrepignore")
+        semgrepignore_path = TEMPLATES_DIR / IGNORE_FILE_NAME
     else:
-        logger.verbose("no .semgrepignore found")
-        file_ignore = None
+        logger.verbose("using path ignore rules from user provided .semgrepignore")
+
+    with open(semgrepignore_path) as f:
```
brendongo:
```suggestion
    with semgrepignore_path.open() as f:
```
Path objects expose an open helper :D

semgrep/semgrep/semgrep_main.py
```diff
@@ -63,21 +63,21 @@ def notify_user_of_work(
...
     semgrepignore_path = Path(workdir / IGNORE_FILE_NAME)
-    if semgrepignore_path.is_file():
-        logger.verbose("using path ignore rules from .semgrepignore")
-        with open(semgrepignore_path) as f:
-            file_ignore: Optional[FileIgnore] = FileIgnore(
-                base_path=workdir,
-                patterns=Parser(workdir).parse(f),
-            )
+    if not semgrepignore_path.is_file():
+        logger.verbose("using default .semgrepignore")
```
brendongo:
```suggestion
        logger.verbose("No .semgrepignore found. Using default ignores rules.")
```
Thoughts?

brendongo (approved):
Some nits.

semgrep/semgrep/semgrep_main.py
ievans: can we provide a pointer to it? maybe a link to docs?

emjin:
Collllleeeeeeeeeeen

Edit: nvm I can't read

semgrep/semgrep/semgrep_main.py
mmcqd: Will update this and the docs later with info about the default.

mmcqd: For now just made Brendon's change

Issue #4428, opened 2021-12-14
Autofix overwrites wrong line with multiple multiline matches (second example)


finnigja:
**Describe the bug**
When using autofix to manipulate Terraform configuration files, I found that fixes for multiple issues within the same file would not be written to the correct line locations.

This appears to be similar to an already reported issue - https://github.com/returntocorp/semgrep/issues/3388 - but I'm reporting it here because having a separate test case may help with development of a fix.

**To Reproduce**
https://semgrep.dev/s/chair6:terraform-multiline-autofix-bug

Input file:
```
resource "aws_instance" "example1" {
  ami           = "ami-005e54dee72cc1d01"
  instance_type = "t2.micro"
}

resource "aws_instance" "example2" {
  ami           = "ami-005e54dee72cc1d02"
  instance_type = "t2.micro"
}
```

Rule file:
```
rules:
- id: ec2-instance-metadata-options
  languages:
  - terraform
  message: EC2 instance does not set metadata options
  severity: WARNING
  patterns:
  - pattern-inside: |
      resource "aws_instance" "$RESNAME" {
      ...
      }
  - pattern-not-inside: |
      resource "aws_instance" "..." {
        ...
        metadata_options {
          ...
        }
        ...
      }
  fix-regex:
    regex: (.*)\}
    replacement: |
      \1
        metadata_options {
          http_tokens = "required"
        }
      }
```

Output without `--autofix`:
```
$ semgrep -c ec2-instance-metadata-options.yml main.tf
Running 1 rules...
main.tf
rule:ec2-instance-metadata-options: EC2 instance does not set metadata options
autofix: s/(.*)\}/\1
  metadata_options {
    http_tokens = "required"
  }
}
/g
1:resource "aws_instance" "example1" {
2:  ami           = "ami-005e54dee72cc1d01"
3:  instance_type = "t2.micro"
4:}
--------------------------------------------------------------------------------
autofix: s/(.*)\}/\1
  metadata_options {
    http_tokens = "required"
  }
}
/g
6:resource "aws_instance" "example2" {
7:  ami           = "ami-005e54dee72cc1d02"
8:  instance_type = "t2.micro"
9:}
ran 1 rules on 1 files: 2 findings
```

Output with `--autofix`:
```
$ semgrep --autofix -c ec2-instance-metadata-options.yml main.tf
Running 1 rules...
successfully modified 1 file.
main.tf
rule:ec2-instance-metadata-options: EC2 instance does not set metadata options
[...]

Issue #3577, opened 2021-07-21
interspersed autofix of multiple expressions on same line


gordonwoodhull:
Similar to #3388, if fixing multiple expressions in the same line, the autofixes can get mixed up and duplicated, resulting in garbled code.

Here is the repro: https://semgrep.dev/s/gordonwoodhull:wrap-strings

On semgrep.dev it produces the correct individual suggestions. However, when running semgrep 0.59 on the command line with `--autofix`, it looks like it confuses the output positions between the two fixes.

The rule attempts to replace the expression pattern `(String $S)` with the fix `wrap($S)`.

On input

```java
        return "a" + "b";
```

it produces the output

```java
        return wrap("wrap("b") + "b";
```

emjin:
I assume this is also P1 for you?

gordonwoodhull:
Yes. I would say it is P1 because we can still use Semgrep for a lot of things but this makes a small percentage of our fixes impossible.

(We are trying to do some pretty ambitious things with autofix! Thank you for the amazing tool.)

emjin:
Glad to hear that :D Autofix is still an experimental feature, but we're really excited about what people can do with it and hope we can put it on our roadmap soon.

gordonwoodhull:
I am planning to work on this.

gordonwoodhull:
Outline of solution in https://github.com/returntocorp/semgrep/issues/3388#issuecomment-967756894.

Hope to contribute soon.

ievans:
Confirmed fixed in 0.77! Playground is confused now but semgrep CLI is solid.

gordonwoodhull:
Thanks @chair6! Took a glance at your solution and it looks equivalent and maybe simpler than what I did.

Will check for edge cases soon.

Issue #3388, opened 2021-06-18
Autofix overwrites wrong line with multiple multiline matches


JuliusDegesys:
**Describe the bug**
Incorrect chunks of the file are auto-fixed. Searching for `assert_eq(True, $X)` and replacing with `assert $X` will mangle the output when auto-fixing multiple multi-line matches.

**To Reproduce**
https://semgrep.dev/s/juliusdegesys:asserteqbug
Input file:
```python
assert_eq(
    True, "a"
)
assert_eq(
    True, "b"
)
x = "abc"
```
Rule:
```yaml
rules:
    - id: assert_eq-true
      message: Change assert_eq(True, x) to assert x
      severity: INFO
      languages:
      - python
      pattern: assert_eq(True, $ACTUAL)
      fix: assert $ACTUAL
```
The matches look correct, and so do the auto-fixes. But when you "Apply all fixes" or apply the first fix followed by the second, the following output is generated:
```python
assert "a"
assert_eq(
    True, "b"
assert "b"
```

**Expected behavior**
```python
assert "a"
assert "b"
x = "abc"
```

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [X] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?
```
pip install semgrep
$ semgrep --version
0.56.0
```

colleend:
Hi! Thanks for filing this issue -- we'll look into it as soon as we get the time to. As a heads up, autofix is an alpha feature so it is not as well supported as other GA features!

gordonwoodhull:
Understand 100% that autofix is in alpha, however I thought I would add that autofix also gets confused when applying multiple changes to the same line.

Here is the repro: https://semgrep.dev/s/gordonwoodhull:wrap-strings

On semgrep.dev it produces the correct individual suggestions. However, when running on the command line with `--autofix`, it looks like it confuses the two output positions.

The rule attempts to replace the expression pattern `(String $S)` with the fix `wrap($S)`.

On input

```java
        return "a" + "b";
```

it produces the output

```
        return wrap("wrap("b") + "b";
```

colleend:
Thanks for commenting about this issue! Would you mind filing a separate autofix issue with the contents of the comment? 🙏

gordonwoodhull:
It’s an important issue - please don’t close.

[...]

PR #4429, opened 2021-12-14
Address several autofix output issues by adding per-file line/column offset tracking

finnigja:
This change addresses several autofix-related issues by adding per-file line/column offset tracking, and using those offsets when making edits to files.

There is no change in usage or documentation; it just addresses several edge cases within the existing autofix implementation (multi-line fixes, multiple-fixes-in-one-line fixes) that were not cleanly handled until now.

This should address several open issues; https://github.com/returntocorp/semgrep/issues/4428, https://github.com/returntocorp/semgrep/issues/3577, and https://github.com/returntocorp/semgrep/issues/3388.

It adds several new e2e tests with rules/targets based on the issues above, and passes the existing tests in `tests/e2e/test_autofix.py` on my local machine.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/autofix.py
+++ b/semgrep/semgrep/autofix.py
@@ -22,6 +22,9 @@ class Fix:
 
 
 class FileOffsets:
+    """
+    Used to track state when applying multiple fixes to a single file/line.
+    """
     def __init__(self, line_offset: int, col_offset: int, active_line: int):
         self.line_offset = line_offset
         self.col_offset = col_offset
--- a/semgrep/semgrep/autofix.py
+++ b/semgrep/semgrep/autofix.py
@@ -1,5 +1,6 @@
 import re
 from pathlib import Path
+from typing import Dict
 from typing import List
 from typing import Set
 from typing import Tuple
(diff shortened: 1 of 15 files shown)


ievans (approved):
@chair6 thanks for a terrific first contribution. This looks good to me. I think a docstring in code about what FileOffsets represents might be helpful, as it wasn't obvious to me at first, but as I understand it maintains the state the is necessary to track maintain correct behavior when applying multiple fixes to a single file/line.

finnigja:
Thanks for the review, @ievans!  Yup, you got it.. I've added the docstring.

semgrep/semgrep/autofix.py
```diff
@@ -20,28 +21,52 @@ def __init__(self, fixed_contents: str, fixed_lines: List[str]):
         self.fixed_lines = fixed_lines
 
 
+class FileOffsets:
+    """
+    Used to track state when applying multiple fixes to a single file/line.
```
emjin:
```suggestion
    Used to track state when applying multiple fixes to the same lines in a file.
```
```suggestion
    Used to track state when applying multiple fixes to a single file/line.
```

semgrep/semgrep/autofix.py
```diff
@@ -20,28 +21,52 @@ def __init__(self, fixed_contents: str, fixed_lines: List[str]):
...
+    """
+
+    def __init__(self, line_offset: int, col_offset: int, active_line: int):
+        self.line_offset = line_offset
+        self.col_offset = col_offset
+        self.active_line = active_line
```
emjin, line 38: I think this assumes that fixes will happen in order (we will never come back and re-fix a line). This means that if one rule modifies lines 3 and 7, then another rule modifies lines 3 and 5, we'll still have a problem with line 3 (file parsing is cached, which means they won't be reparsed after the autofix is applied). We should add a comment documenting this limitation.

(If not sorry for misunderstanding!)

emjin (approved):
This looks awesome! Made a very minor clarification to your comment. Also, I believe this won't work if two rules make a change to the same line, so we should just make a note of that.

Thank you so much for this contribution! A huge improvement for autofix

semgrep/semgrep/autofix.py
finnigja: 👍 , I've addressed this suggestion in the followup commit. I think "applying multiple fixes to the same or subsequent lines in a single file" is accurate.

semgrep/semgrep/autofix.py
finnigja: That makes sense.. I've extended the docstring to mention this limitation.

brendongo (approved):

PR #4430, opened 2021-12-14
simplifying test file and refactoring

colleend:
This PR:
* refactors test file
* removes the functionality to compute matrix of tn, fn, tp, fp
* removes todo output functionality
* improves error output for ruleid and rule mismatches

_To Test:_
* run `semgrep --test --strict --quiet .` in `semgrep-rules`.

_Test Output:_
For directories that pass:
Screen Shot 2021-12-13 at 5 59 37 PM

For directories with errors:
Screen Shot 2021-12-15 at 5 29 15 PM

_Note:_
* this will not affect the output of the `semgrep_runs_output.tar.gz` file used to highlight the results in the `semgrep-app` registry. This is because `semgrep_runs_output.tar.gz` output is based on the `results` dictionary, which is not touched by these changes. So our later upgrades to this file will not cause the registry to have no highlights 👍 

Link for more thorough test file output outline: https://r2c.quip.com/hESGAvfkaHIv/Proposal-for-test-output

### DIFF
--- a/semgrep/semgrep/test.py
+++ b/semgrep/semgrep/test.py
@@ -68,41 +68,6 @@ def normalize_rule_ids(line: str) -> Set[str]:
     return set(filter(None, [rule.strip() for rule in rules_clean]))
 
 
-def compute_confusion_matrix(
-    reported: Set[Any], expected: Set[Any], oked: Set[Any]
-) -> Tuple[int, int, int, int]:
-    true_positives = len(expected.intersection(reported))
-    false_positives = len(reported - expected)
-    true_negatives = len(oked)
-    false_negatives = len(expected - reported)
-
-    return (true_positives, true_negatives, false_positives, false_negatives)
-
-
-def _test_compute_confusion_matrix() -> None:
-    tp, tn, fp, fn = compute_confusion_matrix(set([1, 2, 3, 4]), set([1]), set())
-    assert tp == 1
-    assert tn == 0
-    assert fp == 3
-    assert fn == 0
-
-    tp, tn, fp, fn = compute_confusion_matrix(
-        set([1, 2, 3, 4]), set([1, 2, 3, 4]), set([1])
-    )
-    assert tp == 4
-    assert tn == 1
-    assert fp == 0
-    assert fn == 0
-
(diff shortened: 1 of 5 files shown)


emjin:
What do you mean by todo output functionality?

colleend:
I removed all the code that reports TODOs (mentioned in the second bullet point of the quip). I'm planning on removing the `--test-ignore-todo` flag soon, but want to do that in another PR because there's a lot of different files involved 👍

semgrep/semgrep/test.py
```diff
@@ -176,24 +185,25 @@ def score_output_json(
...
-                        )
-                if (not ignore_todo and todo_rule_in_line) or ok_in_line:
-                    for rule_id in rule_ids:
-                        ok_lines[test_file_resolved][rule_id].append(effective_line_num)
-                if ignore_todo and todo_ok_in_line:
-                    for rule_id in rule_ids:
-                        todo_ok_lines[test_file_resolved][rule_id].append(
-                            effective_line_num
-                        )
+                if todo_rule_in_line or rule_in_line:
```
minusworld, line 188: Trying to understand this behavior here: Does this mean that `todo` lines will still be considered as part of the final set of rules? Will there be any errors if `todo` lines match? Don't match?

minusworld: Ah, if I read the thread correctly this will get removed at a later time

minusworld:
Also, I noticed in the image with failed tests that the X's weren't on the same column. Is that a terminal resize glitch or will they display that way normally? Could it be put on the same column?

semgrep/semgrep/test.py
colleend: yep! sorry for the late reply -- there's no errors if `todo` lines match currently, and the `todo` lines aren't considered as the final set of rules (they're later removed from the reported and expected sets of match line numbers). So currently this is still completely functional 👍

colleend:
Ah yep, I fixed that in one of the later commits! I'll update the screenshot (I also added a new line between the errors so that it's more clear)

minusworld (approved):

PR #4443, opened 2021-12-15
Support process-subst for -config in semgrep-core

aryx:
This allows to use semgrep-core with rules from the registry.
This will be useful when extending semgrep-core to process
a set of rules in different language at once, to test things.

Test plan:
semgrep-core -lang ocaml -config <(curl https://semgrep.dev/c/p/ocaml) tests/ocaml/

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/Parse_rule.ml
+++ b/semgrep-core/src/parsing/Parse_rule.ml
@@ -21,6 +21,8 @@ module G = AST_generic
 module PI = Parse_info
 module Set = Set_
 
+let logger = Logging.get_logger [ __MODULE__ ]
+
 (*****************************************************************************)
 (* Prelude *)
 (*****************************************************************************)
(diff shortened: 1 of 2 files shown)


mjambon (approved):

mjambon:
I don't know what's up with the CLA bot. Check your commit's email address?

brendongo:
Reopened PR to retrigger CLA bot :D

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -814,8 +815,10 @@ let parse file =
             let n = Sys.command cmd in
             if n <> 0 then failwith (spf "error executing %s" cmd);
             Json_to_generic.program (Parse_json.parse_program tmpfile))
+    | FT.Config FT.Yaml -> Yaml_to_generic.program file
     | _ ->
-        failwith
-          (spf "wrong rule format, only JSON/YAML/JSONNET are valid:%s:" file)
+        logger#error "wrong rule format, only JSON/YAML/JSONNET are valid";
```
IagoAbal, line 820: If we expect this to happen when using a named pipe, then I don't think this should be logged as an error, but rather as info.

aryx: true, but also I don't expect people to abuse semgrep-core to do what semgrep is supposed to do, so using named pipes for -config is really not the usual case.

Issue #4327, opened 2021-11-23
Support for Dockerfile


minusworld:
**Describe the solution you'd like**

Support for dockerfile.

[This rule](https://github.com/returntocorp/semgrep-rules/issues/1596) has a false positive due to comments in a Dockerfile. Others have reported similar experience. Dockerfile support would help eliminate these FPs.

**Additional context**
Priority: Low (Not urgent)

mjambon:
I was working on dockerfile support before I broke my arm. Maybe there'll be some preliminary support in 2-3 weeks, i.e. whenever I get back to full-time work.

minusworld:
It's no rush - mainly just wanted to put it here for accounting purposes 👍

PR #4451, opened 2021-12-21
Initial Dockerfile support

mjambon:
All dockerfile constructs are now supported. Ellipsis or metavariables are supported in the following places:
* whole dockerfile instructions
* in bash code
* within dockerfile arrays e.g. `RUN [ "sudo", ... ]`

What's left is to support ellipses and metavariables in the various instructions that are not argv or shell commands e.g. `ENV ... HOME=$DIR ...`.

The only controversial thing I think is the introduction of `ArgRequired` and `ArgOptional` which is used to allow matching optional attributes of dockerfile commands translated into the `Call` construct. See the examples `from.dockerfile`, `from2.dockerfile` and the accompanying patterns `from.sgrep` and `from2.sgrep`.

Main things to review:
- [ ] introduction of `ArgOptional`
- [ ] new `Loc` module
- [ ] capture of raw backtrace alongside exception using `Error of exn * Printexc.raw_backtrace`
- [ ] representation of shell commands in dockerfile AST
- [ ] hack for getting correct locations after parsing embedded bash commands 
- [ ] test cases `*.dockerfile`/`*.sgrep`

PR checklist:
- [x] Changelog is up-to-date


ievans:
related: https://github.com/returntocorp/semgrep/issues/4327

semgrep-core/src/parsing/tree_sitter/Parse_dockerfile_tree_sitter.ml
```diff
@@ -0,0 +1,609 @@
...
+  let v2 = token env v2 (* "=" *) in
+  let v3 = string env v3 in
+  todo env (v1, v2, v3)
+
+(* hack to obtain correct locations when parsing a string extracted from
+   a larger file. *)
+let shift_locations (str, tok) =
+  let line (* 0-based *) = max 0 (PI.line_of_info tok - 1) (* 1-based *) in
+  let column (* 0-based *) = max 0 (PI.col_of_info tok) in
+  String.make line '\n' ^ String.make column ' ' ^ str
```
mjambon, line 327: ^ hack to obtain correct locations when parsing embedded bash

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -790,11 +790,23 @@ and xml_body =
 (* brackets can be fake '()' for OCaml/Ruby *)
 and arguments = argument list bracket
 
+(*
+   ArgRequired = argument must exist in both pattern and target.
+   ArgOptional = still matches if the argument exists in the target but
```
aryx, line 795: maybe a comment saying that this is currently used just for Bash and Dockerfile, and maybe explain why we need that with an example.

aryx: I would have preferred a separate ArgKwdOptional maybe.

semgrep-core/src/parsing/ast/Loc.mli
```diff
@@ -0,0 +1,79 @@
+(*
+   Representation of the location of a piece of code by a pair of tokens.
+
+   This is used so far in intermediate, language-specific ASTs.
```
aryx, line 4: Maybe explain why we need that? Why only Bash/Dockerfile are using that?
Which advantages does that bring over using Parse_info.t?

semgrep-core/tests/dockerfile/add-chown.dockerfile
```diff
@@ -0,0 +1,6 @@
+ADD file1 file2 /mydir/
```
aryx, line 1: usually the semgrep tests files in semgrep-core/tests/xxx/ follow certain conventions. See the script that build the cheatsheet. I would recomment to rename lots of those files misc_xxx.dockerfile to not clutter the different test files that exhibit important features of semgrep (e.g., metavar_args.dockerfile, dots_args.dockerfile, etc.).

aryx (approved):
Looks mostly good.

semgrep-core/src/core/ast/AST_generic.ml
mjambon: A dedicated ArgKwdOptional also seemed best to me initially but it seemed to make the pattern-matching implementation trickier. Let me see if I can make this work.

semgrep-core/tests/dockerfile/add-chown.dockerfile
mjambon: Sure, but each docker command has its own syntax and needs to be tested. They're all converted to functions in the generic AST. I think I'll use the standard test names for the `RUN` commands since they are most commonly used.

semgrep-core/src/core/ast/AST_generic.ml
mjambon: ok, I changed it, it works fine after all. See upcoming PR.

PR #4467, opened 2022-01-04
use latest pfff, and make it compatible with pad's codegraph

aryx:
I can now use 'make index' in semgrep-core with the latest codegraph

PR checklist:
- [x] Changelog is up-to-date


aryx:
seems like we got some new parse errors in public_repos during christmas.

PR #4471, opened 2022-01-05
Dockerfile support - fixes for previous PR

mjambon:
Addresses comments left by @aryx in #4451.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/api/AST_generic_v1.atd
+++ b/semgrep-core/src/api/AST_generic_v1.atd
@@ -466,6 +466,9 @@ type argument = [
   | Arg of expr (* can be Call (IdSpecial Spread, Id foo) *)
   (* keyword argument *)
   | ArgKwd of (ident * expr)
+  (* optional keyword argument - can occur in target code without being
+     specified in the pattern *)
+  | ArgKwdOptional of (ident * expr)
   (* type argument for New, instanceof/sizeof/typeof, C macros *)
   | ArgType of type_
   | ArgOther of (string * any list)
--- a/semgrep-core/src/parsing/ast/Loc.mli
+++ b/semgrep-core/src/parsing/ast/Loc.mli
@@ -1,12 +1,40 @@
 (*
-   Representation of the location of a piece of code by a pair of tokens.
+   Representation of the location of a piece of code by the first and
+   the last token of the region. The token type is the one used in the
+   generic AST.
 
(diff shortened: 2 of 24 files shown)


semgrep-core/src/api/AST_generic_v1.atd
```diff
@@ -466,6 +466,9 @@ type argument = [
   | Arg of expr (* can be Call (IdSpecial Spread, Id foo) *)
   (* keyword argument *)
   | ArgKwd of (ident * expr)
+  (* optional keyword argument - can occur in target code without being
```
aryx, line 469: Actually in theory we should not change those _v1.atd because it's supposed to be a fixed AST.
Nobody uses this _v1 for now, so no big deal, but the spirit of those _vxxx.atd is to be fixed. 
See the toplevel comment in this file: "DO NOT MODIFY THIS FILE ..."

semgrep-core/src/parsing/ast/Loc.mli
```diff
@@ -1,12 +1,40 @@
...
+   generic AST.
 
-   This is used so far in intermediate, language-specific ASTs.
+   This is used so far in intermediate, language-specific ASTs. A location is
+   stored with each node of the AST, similarly to how it's done in Camlp4's
+   OCaml AST where it used to work well. I (Martin) encourage others to use
+   the same approach in newer code. I also think it would be an improvement
+   to use this in the generic AST, but that would need a lot of refactoring.
+
+   What it simplifies:
```
aryx, line 12: fantastic comment. Thx!

aryx (approved):

semgrep-core/src/api/AST_generic_v1.atd
mjambon: oh yes, I wanted to ask about that. What we should we do when we run into a new construct? The simplest would be to raise an exception. What do you think?

mjambon: e.g. when exporting the generic AST to json we'd have this:
```ocaml
| ArgKwdOptional _ ->
    raise (Unsupported_construct "ArgKwdOptional")
```

aryx: Right now I raise/failwith because anyway this code is not really used, but in theory we should try to convert to something similar (here we could convert to an ArgKwd, better than nothing)

Issue #4445, opened 2021-12-15
Rust parse error `!` was unexpected


rc-mattschwager:
**Describe the bug**

Getting a Rust parse error: `'!' was unexpected`.

Unfortunately it doesn't tell me which `!` was unexpected, so I've included the whole file.

**To Reproduce**

https://semgrep.dev/s/Pgpq

**Expected behavior**

Successfully parses.

**Screenshots**

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?

aryx:
This works now: https://semgrep.dev/s/Pgpq?version=develop

Issue #4444, opened 2021-12-15
Rust parse error on 'allow' attribute


rc-mattschwager:
**Describe the bug**

Parse error on `allow` attribute: https://stackoverflow.com/questions/25877285/how-to-disable-unused-code-warnings-in-rust

Similar to https://github.com/returntocorp/semgrep/issues/3762, but not exactly the same.

**To Reproduce**

https://semgrep.dev/s/38Ab

**Expected behavior**

Successfully parses.

**Screenshots**

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?

emjin:
👋

rc-mattschwager:
Hi from the outside! I'm giving Semgrep a go on some Rust projects :)

aryx:
Issue 4444! I'm sure this pleases @rc-mattschwager who likes magic numbers :)

PR #4472, opened 2022-01-05
Use latest tree-sitter-rust

aryx:
This actually introduces some parsing regressions from 99.03%
to 98.85%
But as I explained in
https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/277
there is no much choice. We need to start from the latest
tree-sitter-rust going forward.

test plan:
```
$ cd parsing-stats; ./run-lang rust
lang/rust/stats.json:3
   "global": {
     "name": "*",
     "parsing_rate": 0.9885293607901287,
     "line_count": 4413965,
     "error_line_count": 50631,
     "file_count": 32991,
     "error_file_count": 2815
   },
```

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/parsing-stats/run-lang
+++ b/parsing-stats/run-lang
@@ -121,6 +121,8 @@ main() {
     ./upload-parsing-rates lang/"$lang"/stats.json
   fi
   echo "stats available in lang/$lang/stats.json"
+  #alt: cat lang/$lang/stats.json | jq .global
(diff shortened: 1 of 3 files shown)


aryx:
The problem with #4444 was because tree-sitter-rust of ruin0x11 didn't allow
inner_attribute in block.
By switching back to the official tree-sitter-rust we didn't have the problem anymore.

aryx:
This closes also https://github.com/returntocorp/semgrep/issues/4445

Issue #4473, opened 2022-01-05
[Internal] cleanup xfail_repo in public_repos.py, lots of parse errors


aryx:
We should try to fix all those parse errors.
This starts to add up.

aryx:
Looks like most of the issues are with javascript parsing errors.

PR #4474, opened 2022-01-05
Fix CI failure, add more xfail()

aryx:
One regression was because a repository does not exist anymore (returntocorp/check-buffer-rules)
one was because of a real parse error in returntocorp/semgrep-rules
=> I've created https://github.com/returntocorp/semgrep-rules/pull/1698
the two other are JS parse errors
=> I've created https://github.com/returntocorp/semgrep/issues/4473

test plan:
pipenv shell
pytest  -v --tb=short tests/qa/test_public_repos.py

PR checklist:
- [x] Changelog is up-to-date


aryx:
See https://github.com/returntocorp/semgrep/runs/4713621483?check_suite_focus=true for the previous failures

PR #4483, opened 2022-01-06
sym-prop: Fix matching with deep expression patterns

### LINKED ISSUE #3207: [Java] Symbolic expression propagation
**Describe the bug**
Hello team,
Can semgrep expand the expression（Similar to three address codes）？Or is there any other tool that can help me expand the expression in Java? thx :D

**To Reproduce**

Code A and code B are semantically the same.  For code B, I just expanded the expression in code A.
code A:
```java
public final ZipEntry getNextEntry() {
        String name;
        ZipEntry nextEntry = super.getNextEntry();
        if (nextEntry == null || (name = nextEntry.getName()) == null || (!name.contains("../") && !name.contains("..\\"))) {
            return nextEntry;
        }
        throw new SecurityException(":" + nextEntry.getName());
    }
```
code B:
```java
public final ZipEntry getNextEntry() {
        String name;
        ZipEntry nextEntry = super.getNextEntry();
        if (nextEntry == null)
            return nextEntry;
        ZipEntry c = nextEntry;
        name = c.getName();
        if (name == null)
            return nextEntry;
        
        boolean b1 = !name.contains("../");
        boolean b2 = !name.contains("..\\");
        if (b1 && b2)
            return nextEntry;

        throw new SecurityException(":" + nextEntry.getName());
    }
```
The following rule matches code B well, but not code a. 
```
rules:
- id: ZipEntry
  patterns:
    - pattern: $Y = $X.getName();
    - pattern-inside: |
        (ZipEntry $X) = ...;
        ...
        (String $Y) = $X.getName();
        ...
        $Z = <... $Y.contains("...") ...>;
        if (<... $Z ...>) {
          ...
        }

        ...
  languages: 
    - java
  message: |
    match
  severity: WARNING
```

**Expected behavior**

For Java, can semgrep or other tools split expressions into unary or binary expressions?

IagoAbal:
Let's say you want to match pattern:

    return <... name.contains("...") ...>;

against this code:

    boolean b1 = !name.contains("../");
    boolean b2 = !name.contains("..\\");
    return b1 && b2;

It is not enough to compute the subexpressions of `b1 && b2` and then try
matching them with `name.contains("...")`. This does not work. We need to
consider symbolic values *while* computing the subexpressions, so that
`name.contains("../")` and `name.contains("..\\")` are enumerated.

Also allow special ids to be symbolically propagated.

Closes #3207

test plan:
make test # test included

changelog:
Symbolic propagation didn't exist in 0.76.0, but may have ended up there
by mistake when merging the release-0.76.0 branch. The feature never made
it into a release announcement, so I'm adding it for the next one.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/analyzing/Dataflow_svalue.ml
+++ b/semgrep-core/src/analyzing/Dataflow_svalue.ml
@@ -305,6 +305,7 @@ let rec is_symbolic_expr expr =
   match expr.G.e with
   | G.L _ -> true
   | G.N _ -> true
+  | G.IdSpecial _ -> true
   | G.DotAccess (e, _, FN _) -> is_symbolic_expr e
   | G.ArrayAccess (e1, (_, e2, _)) -> is_symbolic_expr e1 && is_symbolic_expr e2
   | G.Call (e, (_, args, _)) ->
--- a/semgrep-core/src/matching/SubAST_generic.ml
+++ b/semgrep-core/src/matching/SubAST_generic.ml
@@ -93,8 +93,11 @@ let subexprs_of_stmt_kind = function
 let subexprs_of_stmt st = subexprs_of_stmt_kind st.s
 
 (* used for deep expression matching *)
-let subexprs_of_expr e =
+let subexprs_of_expr with_symbolic_propagation e =
   match e.e with
+  | N (Id (_, { id_svalue = { contents = Some (Sym e1) }; _ }))
+    when with_symbolic_propagation ->
+      [ e1 ]
   | L _
   | N _
   | IdSpecial _
@@ -159,6 +162,10 @@ let subexprs_of_expr e =
       []
   | DisjExpr _ -> raise Common.Impossible
   [@@profiling]
+
+(* Need this wrapper because [@@profiling] has the side-effect of removing labels. *)
+let subexprs_of_expr ?(symbolic_propagation = false) e =
+  subexprs_of_expr symbolic_propagation e
 
 (* used for deep statement matching *)
 let substmts_of_stmt st =
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Added
 - Pre-alpha support for Dockerfile as a new target language
+- Semgrep is now able to symbolically propagate simple definitions. E.g., given
+  an assignment `x = foo.bar()` followed by a call `x.baz()`, Semgrep will keep
+  track of `x`'s definition, and it will successfully match `x.baz()` with a
+  pattern like `foo.bar().baz()`. This feature should help writing simple yet
+  powerful rules, by letting the dataflow engine take care of any intermediate
+  assignments. Symbolic propagation is still experimental and it is disabled by
+  default, it must be enabled in a per-rule basis using `options:` and setting
+  `symbolic_propagation: true`. (#2783, #2859, #3207)
 
 ### Fixed
 - Rust: inner attributes are allowed again inside functions (#4444) (#4445)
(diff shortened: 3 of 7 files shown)


semgrep-core/src/analyzing/Dataflow_svalue.ml
```diff
@@ -305,6 +305,7 @@ let rec is_symbolic_expr expr =
   match expr.G.e with
   | G.L _ -> true
   | G.N _ -> true
+  | G.IdSpecial _ -> true
```
aryx, line 308: Why IdSpecial was added there? I'm just curious which example we were not handling before.
Do we need all IdSpecial or only a subset?

semgrep-core/src/matching/SubAST_generic.ml
```diff
@@ -93,8 +93,11 @@ let subexprs_of_stmt_kind = function
 let subexprs_of_stmt st = subexprs_of_stmt_kind st.s
 
 (* used for deep expression matching *)
-let subexprs_of_expr e =
+let subexprs_of_expr with_symbolic_propagation e =
   match e.e with
+  | N (Id (_, { id_svalue = { contents = Some (Sym e1) }; _ }))
```
aryx, line 98: Shouldn't we allow also when id_svalue is a Lit?

aryx: Actually do we still need Lit now that it can be represented by Sym (Lit ...)?
Or it's still nice to have this special case encoded specially (but then we need to add the invariant somewhere
that we never can have Sym (L ...)

semgrep-core/src/matching/SubAST_generic.ml
```diff
@@ -160,6 +163,10 @@ let subexprs_of_expr e =
   | DisjExpr _ -> raise Common.Impossible
   [@@profiling]
 
+(* Need this wrapper because [@@profiling] has the side-effect of removing labels. *)
```
aryx, line 166: If you feel like you want to learn ppx and metaprogramming, you can fix it here :) 
https://github.com/returntocorp/pfff/blob/develop/ppx/ppx_profiling.ml

aryx (approved):
Awesome!

aryx:
Ok I've fixed the merge conflict with the builtin conflict resolver UI of github.

semgrep-core/src/analyzing/Dataflow_svalue.ml
IagoAbal: Should have made this more clear, but the example I added from #3207 is one case, since we have `boolean b1 = !name.contains("../")` and the `!` is an `IdSpecial(Op)`. I think in general we want all the `IdSpecial` so that e.g. we can propagate `x = this.y`.

semgrep-core/src/matching/SubAST_generic.ml
IagoAbal: I actually want. 😄 I had read some articles about it but never actually wrote any PPX extension. I just add this to my to-dos.

semgrep-core/src/matching/SubAST_generic.ml
IagoAbal: Good point, that invariant is supposed to hold (you never have `Sym (Lit ...)` so it should be documented and maybe even checked at some places.

We could add `Lit` here too but I believe it is not needed. This is because if `x` equals to `1`,  then as long as `x` is enumerated by `subexprs_of_expr` we are fine.  You can always match `1` against `x` (if `x` is known to be equals to `1`). The problem is when `x` is equals to `!y`, and the pattern is `<... y ...>`. In that case, if only `x` is enumerated,  then `y` cannot be matched directly against `!y`. We need to say that `!y` is a subexpression of `x` so that a recursive call on `!y` will finally yield `y`.

Issue #3428, opened 2021-06-28
Generic mode should have a way to ignore comments


nbrahms:
**Is your feature request related to a problem? Please describe.**

Let's say a generic file has code like:

```
# WARNING: Do not use
#  my_terrible_api(bad_input)

ok_api(good_input)
```

This will be caught by rules that, e.g., check for `my_terrible_api(...)`.

**Describe the solution you'd like**
Semgrep should have some mechanism to ignore content in comments. Three ideas:
1. Blanket ignore content after common comment sentinels `#`, `//`, `/*` et c.
2. Map file extensions to the the comments for that extension (so generic mode starts being a little less generic).
3. Push the work to the user with a `generic-comment-token` setting, configured on a per-rule basis:
```
settings:
  generic-comment-token: '#'
```

IagoAbal:
I believe Comby does this via idea 2, and we support Comby patterns. I don't know if that works or if we still want to add this feature to Generic mode too. We have `options:` now so we could use idea 3.

IagoAbal:
@returntocorp/pa any comments on this? I don't know how often this is a problem, but I would be happy with solution 3.

aryx:
cc @mjambon

aryx:
Option 3 seems fine to me.
But in many cases people should anyway switch to the real parser instead of using generic mode.
Generic mode is a last resort solution.
We now have JSON, YAML, terraform, and martin just added recently dockerfile, so where do we still need the generic mode?

emjin:
yarn.lock files use generic

mjambon:
I think we could implement a comment stripper which would replace comments by whitespace before proper spacegrep parsing.

mjambon:
@aryx 
> We now have JSON, YAML, terraform, and martin just added recently dockerfile, so where do we still need the generic mode?

I'm pretty sure there are quite a few more formats that we don't support. For example, I had to fix a problem for someone who was using it on markdown files.

aryx:
But does markdown and yarn.lock require a comment stripper?

mjambon:
We're not targetting markdown and yarn.lock specifically, we're targetting all the file formats that people might use. I'm confident that there will always be formats for which we don't have a dedicated parser.

(but yeah, markdown does supports html-style comments and yarn.lock supports shell-style comments)

aryx:
Totally agree, I love generic mode, I just was wondering how important is to handle comments in the rules we need to write.
[...]

PR #4485, opened 2022-01-07
Configurable comment support in spacegrep

mjambon:
I added configurable support for comments to spacegrep. This is not integrated into semgrep, for which we would need to read options from a yaml rule file.

This works as a preprocessing pass on the source file that replaces comments by spaces, keeping newlines intact. For each preprocessing pass, we support at most one kind of end-of-line comments and one kind of multiline comments.

This is the first step to ignore comments in generic mode (#3428).

test plan:
* `make test` (unit tests)
* exercise the `spacegrep` command line to make sure it's there:
```
$ cd semgrep-core/src/spacegrep
$ make
$ echo $'hello # ignore this\nworld!!!\nend' | ./bin/spacegrep 'hello world' --comment-style shell
hello # ignore this
world!!!
```
(highlights region from `hello` to `world`)

This shows the tokenization (and the AST which here is flat):
```
$ echo $'hello # ignore this\nworld!!!\nend' | ./bin/spacecat --comment-style shell
hello
world
!
!
!
end
```

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/spacegrep/src/lib/Comment.ml
+++ b/semgrep-core/src/spacegrep/src/lib/Comment.ml
@@ -4,29 +4,18 @@
 
 open Printf
 
-type delimiters = { start : string; end_ : string }
+type comment_syntax = End_of_line of string | Multiline of string * string
 
-type style = {
-  name : string;
-  start : string option;
-  delimiters : delimiters option;
-}
+type style = comment_syntax list
 
-let style ?start ?delimiters name =
-  let delimiters =
-    match delimiters with
-    | None -> None
-    | Some (start, end_) -> Some { start; end_ }
-  in
-  { name; start; delimiters }
+let shell_style = [ End_of_line "#" ]
 
-let shell_style = style ~start:"#" "shell"
+let c_style = [ Multiline ("/*", "*/") ]
 
-let c_style = style ~delimiters:("/*", "*/") "c"
+let cpp_style = End_of_line "//" :: c_style
 
-let cpp_style = style ~start:"//" ~delimiters:("/*", "*/") "cpp"
-
-let predefined_styles = [ c_style; cpp_style; shell_style ]
+let predefined_styles =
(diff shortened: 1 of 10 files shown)


aryx:
What is the 'Test plan:' ?

semgrep-core/src/spacegrep/src/lib/Comment.ml
```diff
@@ -0,0 +1,163 @@
...
+   Strip comments and the like as a preprocessing phase.
+*)
+
+open Printf
+
+type delimiters = { start : string; end_ : string }
+
+type style = {
+  name : string;
+  start : string option;
```
aryx: Why not Start of string | Delimited of string * string ?
This would avoid those extra style functions as you could build directly the value.

aryx (approved):

aryx:
your unit tests are failing.

mjambon:
I thought the problem was due to ocamlformat and I thought I fixed it but it's due to some other pre-commit hook that removes trailing whitespace.

mjambon:
> What is the 'Test plan:' ?

```
make test
```
as always.

The rest is command-line usage of spacegrep and spacecat, neither of which are used in production. We'll have tests for the semgrep/semgrep-core integration.

semgrep-core/src/spacegrep/src/lib/Comment.ml
aryx: What about this comment? An ADT seems simpler no?

mjambon: I want to allow at most one kind of each comment (end-of-line and multine) because I think it makes the configuration in yaml more natural. We could still use a list of variants internally, though. Let me see.

mjambon: ok, I switched to variants resulting in the following savings:
```
 3 files changed, 38 insertions(+), 77 deletions(-)
```

aryx (approved):

PR #4488, opened 2022-01-07
Sort semgrep-core match results

mjambon:
This adds a pass to sort the results just before export. The issue was [this](https://semgrep.dev/s/08Zv):

image
^ the matches presented here are actually where the `$SOURCE` metavariable matches, while the whole match is at the same location.

This may break a bunch of tests due to the json output of semgrep-core being reordered. Meanwhile, testing can be done manually with this:
```
$ semgrep-core -lang json -e '[...,$X,...]' <(echo $'["first",\n"second"]') -json | jq .matches[].extra.metavars
{
  "$X": {
    "start": {
      "line": 1,
      "col": 2,
      "offset": 1
    },
    "end": {
      "line": 1,
      "col": 9,
      "offset": 8
    },
    "abstract_content": "\"first\"",
    "unique_id": {
      "type": "AST",
      "md5sum": "061635551a453cba667bcafb772a1f53"
    }
  }
}
{
  "$X": {
    "start": {
      "line": 2,
      "col": 1,
      "offset": 10
    },
    "end": {
      "line": 2,
      "col": 9,
      "offset": 18
    },
    "abstract_content": "\"second\"",
    "unique_id": {
      "type": "AST",
      "md5sum": "2d63ad48d9657782f7af8782745a44c8"
    }
  }
}
```
^ the `line` fields should be 1 then 2 as in the above.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/core/Semgrep_core_response_util.ml
+++ b/semgrep-core/src/core/Semgrep_core_response_util.ml
@@ -8,7 +8,7 @@ open Semgrep_core_response_t
 let compare_position (a : position) b = Int.compare a.offset b.offset
 
 let compare_location (a : location) b =
-  let c = String.compare a.path a.path in
+  let c = String.compare a.path b.path in
   if c <> 0 then c
   else
     let c = compare_position a.start b.start in
--- a/semgrep-core/src/core/Semgrep_core_response_util.ml
+++ b/semgrep-core/src/core/Semgrep_core_response_util.ml
@@ -1,0 +1,74 @@
+(*
+   Utilities for working with the types defined in Semgrep_core_response.atd
+   (Semgrep_core_response_t module)
+*)
+
+open Semgrep_core_response_t
+
+let compare_position (a : position) b = Int.compare a.offset b.offset
+
+let compare_location (a : location) b =
+  let c = String.compare a.path a.path in
+  if c <> 0 then c
+  else
+    let c = compare_position a.start b.start in
+    if c <> 0 then c else compare_position a.end_ b.end_
+
+let compare_metavar_value (a : metavar_value) (b : metavar_value) =
+  let c = compare_position a.start b.start in
+  if c <> 0 then c else compare_position a.end_ b.end_
+
+(* Generic list comparison. The input lists must already be sorted according
+   to 'compare_elt'.
+
+   [1] < [2]
+   [1] < [1; 2]
+   [1; 2] < [2]
+*)
+let rec compare_sorted_list compare_elt a b =
+  match (a, b) with
+  | [], [] -> 0
+  | [], _ :: _ -> -1
+  | _ :: _, [] -> 1
+  | a :: aa, b :: bb ->
+      let c = compare_elt a b in
+      if c <> 0 then c else compare_sorted_list compare_elt aa bb
+
+(*
+   Order the metavariable bindings by location first, then by name.
+   (could go the other way too; feel free to change)
+*)
+let compare_metavar_binding (name1, mv1) (name2, mv2) =
+  let c = compare_metavar_value mv1 mv2 in
+  if c <> 0 then c else String.compare name1 name2
+
+(* Assumes the metavariable captures within each match_extra are already
+   sorted. *)
+let compare_match_extra (a : match_extra) (b : match_extra) =
+  let c = compare_sorted_list compare_metavar_binding a.metavars b.metavars in
+  if c <> 0 then c else compare a.message b.message
+
+(*
+   While the locations of the matches are already in correct order, they
+   come in a reverse order when looking at the metavariables that they
+   match. This function makes a best a effort to return the results
+   in a natural order.
+*)
+let compare_match (a : match_) (b : match_) =
+  let c = compare_location a.location b.location in
+  if c <> 0 then c else compare_match_extra a.extra b.extra
+
+let sort_metavars (metavars : (string * metavar_value) list) =
+  List.stable_sort compare_metavar_binding metavars
+
+let sort_extra (extra : match_extra) =
+  { extra with metavars = sort_metavars extra.metavars }
+
+let sort_match_list (matches : match_ list) : match_ list =
+  let matches =
+    Common.map (fun x -> { x with extra = sort_extra x.extra }) matches
+  in
+  List.stable_sort compare_match matches
+
+let sort_match_results (res : match_results) : match_results =
+  { res with matches = sort_match_list res.matches }
--- a/semgrep-core/src/reporting/JSON_report.ml
+++ b/semgrep-core/src/reporting/JSON_report.ml
@@ -214,6 +214,7 @@ let match_results_of_matches_and_errors files res =
     stats = { okfiles = count_ok; errorfiles = count_errors };
     time = res.RP.rule_profiling |> Common.map_opt json_time_of_profiling_data;
   }
+  |> Semgrep_core_response_util.sort_match_results
   [@@profiling]
 
 let json_of_profile_info profile_start =
--- a/semgrep/tests/e2e/snapshots/test_check/test_nested_pattern_either_rule/results.json
+++ b/semgrep/tests/e2e/snapshots/test_check/test_nested_pattern_either_rule/results.json
@@ -13,25 +13,7 @@
         "lines": "    nested_patterns_func('foo', 1)",
         "message": "test",
         "metadata": {},
-        "metavars": {
-          "$X": {
-            "abstract_content": "'foo'",
-            "end": {
-              "col": 31,
-              "line": 2,
-              "offset": 48
-            },
-            "start": {
-              "col": 26,
-              "line": 2,
-              "offset": 43
-            },
-            "unique_id": {
-              "md5sum": "<masked in tests>",
-              "type": "AST"
-            }
-          }
-        },
+        "metavars": {},
         "severity": "WARNING"
       },
       "path": "targets/basic/nested-patterns.js",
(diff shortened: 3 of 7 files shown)


mjambon:
The tests now pass.

semgrep-core/src/reporting/JSON_report.ml
```diff
@@ -214,6 +214,7 @@ let match_results_of_matches_and_errors files res =
     stats = { okfiles = count_ok; errorfiles = count_errors };
     time = res.RP.rule_profiling |> Common.map_opt json_time_of_profiling_data;
   }
+  |> Semgrep_core_response_util.sort_match_results
```
aryx, line 217: Maybe we should sort the matches instead of the Semgrep_core_respond.match_results?

semgrep-core/src/core/Semgrep_core_response_util.ml
```diff
@@ -0,0 +1,74 @@
+(*
+   Utilities for working with the types defined in Semgrep_core_response.atd
+   (Semgrep_core_response_t module)
+*)
+
+open Semgrep_core_response_t
+
+let compare_position (a : position) b = Int.compare a.offset b.offset
+
+let compare_location (a : location) b =
```
aryx, line 10: can't we use deriving ord to automate some of this boilerplate?

semgrep-core/src/reporting/JSON_report.ml
aryx: I'm just wondering why we sort so late, why not doing it earlier?

semgrep/tests/e2e/snapshots/test_check/test_nested_pattern_either_rule/results.json
```diff
@@ -53,25 +35,7 @@
...
-              "line": 3,
-              "offset": 78
-            },
-            "unique_id": {
-              "md5sum": "<masked in tests>",
-              "type": "AST"
-            }
-          }
-        },
+        "metavars": {},
```
aryx, line 38: who those metavars disappear from results.json?

aryx (changes requested):

semgrep-core/src/core/Semgrep_core_response_util.ml
```diff
@@ -0,0 +1,74 @@
...
+  let c = String.compare a.path a.path in
```
IagoAbal: There may be a bug here, it should probably be `String.compare a.path b.path`, I think that `c` is always going to be `0`. Maybe a Semgrep rule to add for OCaml?

aryx: ohoh, good one, more reason to use deriving ord to automatically generate those.

mjambon: ah, it's like the semgrep rule `$X == $X` which I thought was completely useless. 😅

semgrep-core/src/reporting/JSON_report.ml
mjambon: It was the easiest way to sort the results and this guarantees that the results are sorted regardless of what produced them. But yes, it's a good idea, in general, to produce results in the least surprising order when possible (not just for "final" results).

semgrep-core/src/core/Semgrep_core_response_util.ml
mjambon: I don't think anyone added this to atdgen. I'm sure something can be done about it. I still don't know the details about ppx syntax, but it would be a matter of supporting annotations like
```ocaml
type bar = {
  foo: foo;
} <ocaml ppx="[@@deriving compare]">

type bar = [ A | B of foo ]
```
In the above, the type definitions are mutually recursive but they don't agree on the ppx annotation that the whole recursive type definition should have. It's probably no big deal, just a matter of choosing a convention. Maybe we could just require all the individual atd definitions that are merged into one recursive ocaml definition to have the same exact ppx annotations (since merging them correctly may be more difficult).

Anyway, we don't have this at the moment.

semgrep/tests/e2e/snapshots/test_check/test_nested_pattern_either_rule/results.json
mjambon: oops, I didn't notice these were gone. I still don't know what's happening. `semgrep-core` returns correct results, just in a different order.

semgrep-core/src/core/Semgrep_core_response_util.ml
aryx: Oh I see. Yes, it would be great to add this to atdgen, I had the need for a deriving show recently in another .atd. Could you put a TODO in a comment in the code, so we remember why we have to write this boilerplate manually.

aryx (approved):

semgrep/tests/e2e/snapshots/test_check/test_nested_pattern_either_rule/results.json
mjambon: ~~I think~~ I understand. semgrep "deduplicates" the matches it receives from semgrep-core. It considers duplicates matches that occur at the same location, regardless of what the metavariables captured, and this is order-dependent.

The rule file in the above is
```
rules:
- id: test
  message: test
  severity: WARNING
  languages:
  - ts
  - js
  patterns:
  - pattern-either:
    - pattern: nested_patterns_func('foo', ...)
    - pattern: nested_patterns_func('bar', ...)
  - pattern-either:
    - pattern: nested_patterns_func($X, 1, ...)
    - pattern: nested_patterns_func($X, 2, ...)
```
It can be simplified to (`repro2.yaml`):
```
rules:
- id: test
  message: test
  severity: WARNING
  languages:
  - ts
  - js
  patterns:
  - pattern-either:
    - pattern: f(...)
    - pattern: f($X)
```
and the following target (`repro2.js`):
```
f(x)
```
semgrep-core returns 2 matches, one without a metavariable and one with `$X = 'x'`. Because they match at the same location, the first match is kept by semgrep and the second one is discarded.

Here are the old results without sorting in semgrep-core:
```
$ semgrep-core -config repro2.yaml repro2.js -json | ydump
{
  "matches": [
    {
      "rule_id": "test",
      "location": {
        "path": "repro2.js",
        "end": { "line": 1, "col": 5, "offset": 4 },
        "lines": []
      },
      "extra": {
        "message": "test",
        "metavars": {
          "$X": {
            "start": { "line": 1, "col": 3, "offset": 2 },
            "end": { "line": 1, "col": 4, "offset": 3 },
            "abstract_content": "x",
            "unique_id": {
              "type": "AST",
              "md5sum": "52a71841547db9cbd91c34927fc21f01"
            }
          }
        }
      }
    },
    {
      "rule_id": "test",
      "location": {
        "path": "repro2.js",
        "end": { "line": 1, "col": 5, "offset": 4 },
        "lines": []
      },
      "extra": { "message": "test", "metavars": {} }
    }
  ],
  "errors": [],
  "skipped": [],
}

$ semgrep -c repro2.yaml repro2.js --json | ydump
Running 1 rules...
{
  "errors": [],
  "results": [
    {
      "check_id": "test",
      "end": { "col": 5, "line": 1, "offset": 4 },
      "extra": {
        "is_ignored": false,
        "lines": "f(x)",
        "message": "test",
        "metadata": {},
        "metavars": {
          "$X": {
            "abstract_content": "x",
            "end": { "col": 4, "line": 1, "offset": 3 },
            "start": { "col": 3, "line": 1, "offset": 2 },
            "unique_id": {
              "md5sum": "52a71841547db9cbd91c34927fc21f01",
              "type": "AST"
            }
          }
        },
        "severity": "WARNING"
      },
      "path": "repro2.js",
      "start": { "col": 1, "line": 1, "offset": 0 }
    }
  ]
}
```
and the results now with sorting in semgrep-core:
```
$ semgrep-core -config repro2.yaml repro2.js -json | ydump
{
  "matches": [
    {
      "rule_id": "test",
      "location": {
        "path": "repro2.js",
        "end": { "line": 1, "col": 5, "offset": 4 },
        "lines": []
      },
      "extra": { "message": "test", "metavars": {} }
    },
    {
      "rule_id": "test",
      "location": {
        "path": "repro2.js",
        "end": { "line": 1, "col": 5, "offset": 4 },
        "lines": []
      },
      "extra": {
        "message": "test",
        "metavars": {
          "$X": {
            "start": { "line": 1, "col": 3, "offset": 2 },
            "end": { "line": 1, "col": 4, "offset": 3 },
            "abstract_content": "x",
            "unique_id": {
              "type": "AST",
              "md5sum": "52a71841547db9cbd91c34927fc21f01"
            }
          }
        }
      }
    }
  ],
  "errors": [],
  "skipped": [],
}

$ semgrep -c repro2.yaml repro2.js --json | ydump
Running 1 rules...
{
  "errors": [],
  "results": [
    {
      "check_id": "test",
      "end": { "col": 5, "line": 1, "offset": 4 },
      "extra": {
        "is_ignored": false,
        "lines": "f(x)",
        "message": "test",
        "metadata": {},
        "metavars": {},
        "severity": "WARNING"
      },
      "path": "repro2.js",
      "start": { "col": 1, "line": 1, "offset": 0 }
    }
  ]
}
```

mjambon: I don't know if we should do anything about this. It looks like proper behavior to me.

aryx: ok, sounds good to me. Thanks for investigating.

semgrep-core/src/core/Semgrep_core_response_util.ml
aryx: I guess you can add $M_.compare $Y $Y in the local semgrep.yml

mjambon: added to semgrep-rules: https://github.com/returntocorp/semgrep-rules/pull/1714

semgrep/tests/e2e/snapshots/test_check/test_nested_pattern_either_rule/results.json
IagoAbal: Not something that concerns this PR, but maybe Semgrep's deduplication should be deterministic and not depend on the order in which semgrep-core outputs the findings. cc @brendongo

PR #4493, opened 2022-01-10
new interfaces/ directory to share component interfaces

aryx:
PR checklist:
- [x] Changelog is up-to-date


interfaces/Parsing_stats.atd
```diff
@@ -0,0 +1,40 @@
+(*
+   Output parsing stats for a collection of projects to json.
+   This is used by semgrep-core -parsing_stats -json
+*)
+
+type t = {
+  language: string; (* should be compatible with Lang.t *)
+  global: project_stats;
+  projects: project_stats list;
+}
```
mjambon, line 10: It's usually best to default list fields to the empty list when it makes sense. Here, it would be:
```
  ~projects: project_stats list;
```
`~` means the field is optional and there's a default value. This type (a list) has an implicit default value so there's no need to provide it.

interfaces/rule_schema.yaml
```diff
@@ -0,0 +1,496 @@
+$id: https://raw.githubusercontent.com/returntocorp/semgrep/develop/semgrep/semgrep/rule_schema.yaml
+$schema: http://json-schema.org/draft-07/schema#
+$comment: '!!if you modify this file, you probably also need to update the one in semgrep-app:server/semgrep_app/rule_schema.yaml!!'
```
emjin, line 3: The semgrep-core rules parser doesn't use this schema right now; we should document that. Also, it might be reasonable to build the semgrep-core rules parser using this schema, which would be better.

mjambon (approved):

emjin (approved):
This is nice, I didn't realize you could create a module from a file path in that way. Would be good to add a README in the interfaces folder explaining what these interfaces mostly are for

interfaces/rule_schema.yaml
emjin: (I would be happy to do that at some point)

interfaces/Parsing_stats.atd
aryx: But here we use this .atd for output information, so for sure the projects list should never be empty.

aryx:
@emjin "I didn't realize you could create a module from a file path in that way" what are you talking about?

emjin:
> @emjin "I didn't realize you could create a module from a file path in that way" what are you talking about?

image

Probably used the wrong words (this is an atd file) --- this is what I was surprised by.

aryx:
> > @emjin "I didn't realize you could create a module from a file path in that way" what are you talking about?
> 
> image
> 
> Probably used the wrong words (this is an atd file) --- this is what I was surprised by.

Still not sure what you mean, but yes this is a symbolic link towards ../../../interfaces/. In the long term it would be better to avoid those symlinks and have a dune file in this top interfaces/ directory, but that would require to put the dune-project at the top too, which had problems in the past.

aryx:
the errors seems unrelated. I get read timeouts from the registry, and otherwise semgrepdep seems confused by the presence of symbolic links.
I'll merge.

mjambon:
This broke the build. I fixed it here but unfortunately GHA is still not working: https://github.com/returntocorp/semgrep/pull/4510

image

PR #4500, opened 2022-01-11
Pass JSON data to -targets

aryx:
This is part of  https://r2c.quip.com/Xk8ZAGJPBVlf/Semgrep-core-separation

The end goal is to pass the list of targets with more information attached to them such
as the language of the target, and the rule ids to use for this target, so all the
file targeting work can be done Python-side instead of currently spread over
Python and OCaml right now.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/interfaces/Input_to_core.atd
+++ b/interfaces/Input_to_core.atd
@@ -1,6 +1,6 @@
 (*
-   Type definitions for semgrep-core's input and command-line flags,
-   meant for the python wrapper, 'semgrep'.
+   Type definitions for semgrep-core's JSON input, meant for the
+   python wrapper, 'semgrep'.
 
    There are other very important form of inputs which are not specified here:
     - The rule syntax and schema (see rule_schema.yaml; only the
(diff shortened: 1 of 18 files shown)


aryx:
I get those errors:```
semgrep/semgrep/target_manager.py:322: error: Argument 1 to "frozenset" has incompatible type "List[Union[str, _PathLike[str]]]"; expected "Iterable[Path]"  [arg-type]
semgrep/semgrep/target_manager.py:339: error: Argument 1 to "frozenset" has incompatible type "List[Union[str, _PathLike[str]]]"; expected "Iterable[Path]"  [arg-type]
Found 2 errors in 1 file (checked 1 source file)
```
with mypy, anyone know how to fix it? I didn't even change target_manager.py

aryx:
Also why I get those mypy errors locally, but not in CI?
cc @nbrahms @brendongo

mjambon (approved):

emjin (commented):
oops didn't mean to press enter

emjin (approved):
LGTM

Issue #4490, opened 2022-01-10
python.lang.maintainability.useless-literal.useless-literal-dict doesn't work with multiple bytes unicode key


sileht:
**Describe the bug**

python.lang.maintainability.useless-literal.useless-literal-dict doesn't work with multiple bytes unicode key

**To Reproduce**
This code:
```
OPERATOR_ALIASES = {
    ":": "=",
    "==": "=",
    "≠": "!=",
    "≥": ">=",
    "≤": "<=",
}
```
wrongly report:

````
L1: key `"≠": ` is uselessly assigned twice
L1: key `"≥": ` is uselessly assigned twice
L1: key `"≤": ` is uselessly assigned twice
```

**Expected behavior**

Should report nothing

**Screenshots**


**What is the priority of the bug to you?**

- [X] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**

binary downloaded with brew or pip.

Issue #4480, opened 2022-01-06
[Python] restore GA level, current parsing stats are at 99.2%


aryx:
See https://dashboard.semgrep.dev/metric/semgrep.core.Python.parse.pct
we are at 99.2%.
I've recentrly added tree-sitter-python to ocaml-tree-sitter-semgrep and it's at 99.925%
so there is good hope. We just need to switch to tree-sitter for python too.

aryx:
Also we need to add more languages to https://dashboard.semgrep.dev/
now that we have more GA languages.
See here: https://semgrep.dev/docs/language-support/

aryx:
@brandonspark are we there now with tree-sitter-python and all the TODOs fixed?

brandonspark:
@aryx No, sadly not, we are at 99.7%.

There is a single TODO that I left in the code, which I have now fixed in #6014, but I highly doubt that it's why.

I'm pretty sure the fact that we are failing to parse is a result of people passing off fake meta-Python files as Python, which borks the parser. I don't really know if there's something we can do about that.

aryx:
Well we can look at the projects listed in the projects.txt and if 1 project really abuses and have fake meta-python files
we could just skip it (and use another popular Python project instead) and get better stats.

brandonspark:
The two projects are:
```
https://github.com/python/cpython
https://github.com/huggingface/transformers
```

Deleting just one is not enough to get us to 99.9%.

I have investigated both. `huggingface/transformers` uses some weird shenaniganry which looks like metaprogramming.

`cpython` just has code using this weird `except*` and `!` inside of f-strings, which as far as I have checked is not real Python code. Since it also has code _testing_ this syntax, I am under the impression it's a language extension or something, but I'm not sure. `cpython` also has python files which are just tests for the Python grammar, meaning that it has code which does stuff like having a decorator `@False or null`, which is nonsense that would crash any real Python program.

I am down to delete both. Thoughts?

aryx:
Go for it. Just put them in comments in the projects.txt file with your notes on why we skip them

aryx:
It is totally ok to skip cpython. We do the same for other languages where usually the compiler for a lang contains many test cases not representative of real code

aryx:
Got back to 100% Closing!

PR #4506, opened 2022-01-11
[Python] start Parse_python_tree_sitter.ml

aryx:
This might help https://github.com/returntocorp/semgrep/issues/4490
and more generally https://github.com/returntocorp/semgrep/issues/4480

test plan:
make

PR checklist:
- [x] Changelog is up-to-date


mjambon:
ok, I'm not touching the unicode hack in pfff, then.

mjambon (approved):

aryx:
You can touch the unicode hack in pfff if it's a quick fix. I'm independently trying to gradually support Python via tree-sitter, but this will take a little bit of time.

mjambon:
I didn't get to the point where I understand or remember the unicode issue. I see there's no problem with unicode when using tree-sitter parsers, so I believe you that it may be simple. I'll take a look.

PR #4514, opened 2022-01-12
Some cleanup, and change -config to -rules

aryx:
PR checklist:
- [x] Changelog is up-to-date


aryx:
The reasoning behind changing -config to -rules is  partly discussed here:
https://returntocorp.slack.com/archives/C01NXGX2EHZ/p1642002809117500
TL;DR what semgrep-core takes as a parameter is not a "configuration" in the semgrep sense (which can be a URL, a directory of rules, a single file rule, etc.), it's just a flat list of rules in a file, hence the rename to -rules.

emjin (approved):

PR #4517, opened 2022-01-12
Add support for YAML anchors and aliases

### LINKED ISSUE #3677: Add support for anchors in YAML
YAML uses anchors and aliases to allow people to avoid repeating code. Currently, Semgrep essentially discards information about anchors.

From https://r2c-community.slack.com/archives/C022H9MSTKN/p1627286903025100:

If we have

```
---
production:
  assets: &anchor
    "*.html":
      - some_dangerous_things_we_want_to_check
    "*.js":
      *anchor
```

In building rules that check for lines that are in assets -> $ROUTE, we can know that `some_dangerous_things_we_want_to_check` will be checked for the "*.html": which also covers for "*.js": .

But if we have

```
---
anchor: &anchor
  - some_dangerous_things_we_want_to_check
production:
  assets: &anchor
    "*.html":
      *anchor
    "*.js":
      *anchor
```

Then we cannot create a rule that checks for some_dangerous_things_we_want_to_check

See the desired: https://semgrep.dev/s/WOE2

emjin:
Closes https://github.com/returntocorp/semgrep/issues/3677

Aliases will be expanded to the expression at the anchor. To support reporting at the correct range, we also save the original string for the anchor and modify the visitor to ignore tokens in the aliased expression.

Test plan: make test

PR checklist:
- [x] Changelog is up-to-date


emjin:
We could probably promote YAML to Beta (or GA?) after this

aryx (commented):
Could we handle those aliases as local variable definitions, and then let
the symbolic propagation engine report the matches?

emjin:
Oh I forgot we have symbolic propagation! That's probably better :D

Will revisit and figure out how to make it work, this is fine for now and probably less likely to have bugs

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -563,6 +563,9 @@ and m_expr a b =
   Trace_matching.(if on then print_expr_pair a b);
   match (a.G.e, b.G.e) with
   (* the order of the matches matters! take care! *)
+  (* alias: match on the expr inside *)
+  | G.Alias (_alias, a1), _ -> m_expr a1 b
```
emjin, line 567: Alternatively perhaps this one should only match when both are aliases

IagoAbal: Whether it uses the dataflow engine or not, you could require `symbolic_propogation` to be enabled. Then, if it's not enabled, both sides would have to be an `Alias` for it to match.

semgrep-core/src/parsing/other/yaml_to_generic.ml
```diff
@@ -168,65 +171,80 @@ let do_parse env =
...
+  (match anchor with
+  | Some anchor -> Hashtbl.add env.anchors anchor node_expr
+  | None -> ());
+  node_expr
+
+let make_alias anchor pos env =
+  let t = mk_tok pos anchor env in
+  match Hashtbl.find_opt env.anchors anchor with
+  | Some (expr, _p) -> (G.e (G.Alias ((anchor, t), expr)), pos)
+  | None -> raise UnrecognizedAlias
```
IagoAbal: I would use an `option` or `result` type here to be on the safe side. E.g. using `result` and polymorphic variants you could write this as:
```ocaml
let make_alias anchor pos env =
   let t = mk_tok pos anchor env in
   match Hashtbl.find_opt env.anchors anchor with
   | Some (expr, _p) -> Ok (G.e (G.Alias ((anchor, t), expr)), pos)
   | None -> Error `UnrecognizedAlias
```

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -478,6 +478,14 @@ and expr_kind =
...
   | Ref of tok (* &, address of *) * expr
   | DeRef of tok (* '*' in C, '!' or '<-' in OCaml, ^ in Reason *) * expr
+  (* For YAML aliases
+     TODO a better solution would be to use symbolic propagation
+     This is a little tricky because YAML is a highly nested expression
+     and anchors can be nested. We can get around this by extracting
+     the aliases as VarDefs in the beginning and then using them later.
+     Revisit when symbolic propagation is more stable
+  *)
+  | Alias of string wrap * expr
```
IagoAbal, line 488: My only objection is that the expressions we have are pretty generic (used by many languages) and this one is very specific to YAML. Don't know if this would be stretching it too much, but maybe we could encode it in some other way, and then match it specially for YAML?

IagoAbal: E.g. for Go we encode `:=` as `AssignOp(Eq)`, and in the Generic-to-IL translation we decode it into a regular assignment.

semgrep-core/src/api/AST_generic_to_v1.ml
```diff
@@ -283,6 +283,7 @@ and map_expr x : B.expr =
       let t = map_tok t in
       let v1 = map_expr v1 in
       `DeRef (t, v1)
+  | Alias (_alias, v1) -> map_expr v1
```
aryx: we usually do a failwith here, for construct not supported.

semgrep-core/src/parsing/other/yaml_to_generic.ml
```diff
@@ -56,13 +56,16 @@ type env = {
...
   parser : S.parser;
+  anchors : (string, G.expr * E.pos) Hashtbl.t;
   mutable last_event : (E.t * E.pos) option;
 }
 
 (*****************************************************************************)
 (* Helper functions *)
 (*****************************************************************************)
 
+exception UnrecognizedAlias
```
aryx: usually I try to add the location in the error, so of Parse_info.t,
so when we catch it somewhere we can report a location.

aryx (changes requested):
Looks good.
You probably need to rebase though and fix those CI errors.

aryx (approved):

PR #4529, opened 2022-01-15
feat(cli): add autocomplete to command line options

### LINKED ISSUE #3373: Semgrep commands are autocomplete-able
Autocomplete was a delightful feature in our past project [Bento](https://github.com/returntocorp/bento). It aided discoverability of commands and made the tool fast to use. This ticket is a request to provide this functionality for Semgrep, allowing me to tab autocomplete and see relevant flags and their help text as I go.

The AWS CLI, Bento, and others do a good job of this!

Not providing this feature. The tool is still useful, but can be challenging to discover it's capabilities and sub options.

ajbt200128:
Closes #3373

Uses [click](https://click.palletsprojects.com/en/8.0.x/shell-completion/) for autocompleting any command options, and their respective paramaters. Will also autocomplete possible public rulesets, i.e. `semgrep --config p/<tab><tab>` will return all public and visible rulesets, and `semgrep --config p/java<tab><tab>` will return `p/java` and `p/javascript` for autocomplete options.

PR checklist:
- [ ] Changelog is up-to-date


underyx:
Austin you are the best! Wow!

brendongo:
@ajbt200128 does click do caching of autocomplete for configs? I would worry that this would need to make lots of network requests and be slow

ajbt200128:
Currently it does not, network requests are only made when a user is specifically trying to autocomplete a ruleset, which seemed acceptable to me as even with the network overhead, it's still less time/friction than opening up and searching the semgrep registry. I can add a cache file to ~/.semgrep, or if you think somewhere else makes more sense I can add it there. My only question would be what cadence to update the cache on?

semgrep/semgrep/commands/scan.py
```diff
@@ -138,11 +210,12 @@ def convert(
 @click.option(
     "--severity",
     multiple=True,
-    type=click.Choice(["INFO", "WARNING", "ERROR"]),
+    type=click.Choice([e.value for e in RuleSeverity]),
     help=(
         "Report findings only from rules matching the supplied severity level. By default all applicable rules are run."
-        "Can add multiple times. Each should be one of INFO, WARNING, or ERROR."
+        "Can add multiple times. Each should be one of INFO, WARNING, ERROR, or INVENTORY."
```
brendongo: I think for now we want to keep INVENTORY severity internal instead of something users will ever specify

brendongo (approved):

brendongo:
> it's still less time/friction than opening up and searching the semgrep registry

Yeah makes sense

emjin:
Hm I've checked out your branch but I can't reproduce the autocomplete...is there something I need to do? cc @brendongo I'm probably being dumb

brendongo:
You need to let the shell know about the autocomplete :D https://click.palletsprojects.com/en/7.x/bashcomplete/#activation

brendongo:
You need to let the shell know an autocompletion is available: https://click.palletsprojects.com/en/7.x/bashcomplete/#activation

`eval "$(_SEMGREP_COMPLETE=source_zsh semgrep)"`


We will probably want to update docs or create a `semgrep install-autocomplete` command :D 


EDIT: hmm doesnt seem to work for me @ajbt200128 how do we reproduce autocomplete behavior locally?

ajbt200128:
My bad, I forgot to mention I have a documentation PR [here](https://github.com/returntocorp/semgrep-docs/pull/404). As far as an install command goes, I considered it, but personally I don't prefer things to just modify my bash profile automatically. If you think it makes sense though, I definitely can add one, or we could meet in the middle and have the command simply print out something like:

 `Add this line to your [insert detected shell profile] then restart the shell`

 I'm not 100% sure why you're not able to reproduce autcomplete, but it could be that it only works if semgrep is installed (`python -m semgrep` will not work), or that we have different shells, and there is some issue with click.

semgrep/semgrep/commands/scan.py
brendongo: @ajbt200128 mind updating the PR so INVENTORY is not mentioned as an option :D

PR #4535, opened 2022-01-18
Add support for nosemgrep comments on the line before the match

### LINKED ISSUE #3521: Can't suppress with // nosemgrep because of autoformatter
I can’t use //nosemgrep because my autoformatter moves it to its own line, where it is ignored by semgrep. 

To reproduce: https://semgrep.dev/s/340G
Desired behavior: I'd like all of the examples in the above snippet to be suppressed by the `nosemgrep` annotation, including:
* where `nosemgrep` is on its own line before the target line
* where `nosemgrep` is inside the matched range.

This is currently blocking me from merging my PR.

cc @mschwager @minusworld curious if you have thoughts

mjambon:
A nosemgrep comment will cause a match to be ignored if one of these conditions is met:
* the special comment is alone on the line just before the match (new)
* the special comment is on the first line of the match (unchanged)

Note that a nosemgrep comment in the middle of a match (i.e. not on the first line) still won't disable this match. The source code says:
```
    # Only consider the first line of a match. This will keep consistent
    # behavior on where we expect a 'nosem' comment to exist. If we allow these
    # comments on any line of a match it will get confusing as to what finding
    # the 'nosem' is referring to.
```

Documentation PR: https://github.com/returntocorp/semgrep-docs/pull/407

Closes #3521

PR checklist:
- [x] Changelog is up-to-date


semgrep/semgrep/nosemgrep.py
```diff
@@ -15,6 +15,7 @@
 
 from semgrep.constants import COMMA_SEPARATED_LIST_RE
 from semgrep.constants import NOSEM_INLINE_RE
+from semgrep.constants import NOSEM_PREVIOUS_LINE_RE
```
aryx, line 24: @brendongo those long list of import are really tedious to write.
Why not import semgrep.constants as constants, and then we can simply use constants.NOSEM_PREVIOUS_LINE_RE so we don't need to list all those constants?

aryx (commented):
Love this!

semgrep/semgrep/nosemgrep.py
mjambon: pycharm inserts the imports for me and lets me know if they become unused, so it's really not a problem for me.

brendongo (approved):

Issue #4377, opened 2021-12-06
hardcoded color in output


timhemel:
**Describe the bug**
If I run semgrep with a ruleset, part of the output is colored "black" (for some value of black), which happens to be the same color as my terminal background (SolarizedDark color scheme). 

**To Reproduce**
Steps to reproduce the behavior, ideally a link to https://semgrep.dev:
Install a Solarized Dark color scheme in your terminal and run `semgrep -f xyz.yaml`

**Expected behavior**
I expect output to use the current foreground color.

**Screenshots**

The responsible piece of code is in `semgrep/semgrep/formatter/text.py`:

``` python
        line = (
            line[:start_color]
            + with_color(
                "bright_black", line[start_color : end_color + 1]
            )  # want the color to include the end_col
            + line[end_color + 1 :]
        )
```

I am afraid the only proper way to solve this is to implement color schemes in semgrep.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?
Local build.

ievans:
@adamboros are you still working on this?

adamboros:
@ievans Yes, I want to push my changes this week

ievans:
Fixed in #4541 :)

PR #4541, opened 2022-01-19
style: Change colors and indentation of CLI output

adamboros:
PR checklist:
- [ ] Changelog is up-to-date

**Errors and warnings**
Before/after
image
Changes
- Begin/End error trace separator changed to `=`. Also Begin has 2 more `=`, so full length is the same for both lines.
- Added short explanation for error and where to get help.
- Used bgcolor for marking errors, so it shows up right in every terminal. In terminal without colors `[ ]` brackets will be shown (for those with color it won't be visible)
- Moved newline characters (\n), so each error, the useful error message and the error trace is better separated from each other
- in util.py extended `with_color` with new variables

To do
- [x] Mention 'error' and 'warn' only once. It should be `[ERROR] Semgrep CORE`
- [ ] For warn it should be `[WARNING] Semgrep CORE` with same red styling as errors
- [x] Make sure all errors show up with the new style
- [x] Grammar check new help line
- [x] Check if colored brackets can be refactored for a cleaner code

**Results output**
Before/after
image

Changes
- Changed colors and tested new ones in 5 different themes and 3 different terminal softwares. Focused using colors only if necessary and to create a visual rhythm.
- For file name moved away from green as it has a success/repair meaning to it. Used cyan as its readable in every theme and helps separating results.
- For autofix used green color and a fast-forward char to imply action.
- Added new lines and indents to create white-space to increase scannability.
- For line numbers switched from `:` to `┆`
- Before file name there is a space to match the space in Error tags and for premium feel.

To do
- [ ] Fix wrap as it's not using util.py, but import click itself to work. `Terminal_wrap` does not seem to handle indents.
- [x] Separator dashes should start where the rule name/`▶▶` starts.
- [ ] The seperator should end 3-5 characters before the wrap. The text wrap for the rule message should be
- [ ] >>    for smaller windows (<100) at least 5 spaces away from the window border
- [ ] >>    for larger windows (100<) about 15 spaces away from border
- [ ] Indentations should match the current one in the code. Equal tab groups: (filename) // (rule name, autofix, separator) // (rule message, details link, line numbers)
- [ ] There is a double new line under each result, but in the code they add up from two different places.
- [ ] If possible the file name should be bold as well. But in click bolding changes colors to their bright version, especially visible in solarized themes. Only do this, if there is an easy way to do bolding without color change.

Current bugs this PR fixes:
Unreadable/hard to read text in solarized themes and overall scannability
#4377

image

### DIFF
--- a/semgrep/semgrep/constants.py
+++ b/semgrep/semgrep/constants.py
@@ -10,7 +10,7 @@ RULES_KEY = "rules"
 ID_KEY = "id"
 CLI_RULE_ID = "-"
 SEMGREP_URL = os.environ.get("SEMGREP_URL", "https://semgrep.dev/")
-PLEASE_FILE_ISSUE_TEXT = "An error occurred while invoking the semgrep engine; please help us fix this by creating an issue at https://github.com/returntocorp/semgrep"
+PLEASE_FILE_ISSUE_TEXT = "An error occurred while invoking the Semgrep engine. Please help us fix this by creating an issue at https://github.com/returntocorp/semgrep"
 
 DEFAULT_SEMGREP_CONFIG_NAME = "semgrep"
 DEFAULT_CONFIG_FILE = f".{DEFAULT_SEMGREP_CONFIG_NAME}.yml"
--- a/semgrep/semgrep/constants.py
+++ b/semgrep/semgrep/constants.py
@@ -2,6 +2,7 @@ import os
 import re
 from enum import auto
 from enum import Enum
+from types import SimpleNamespace
 from typing import Type
 
 from semgrep import __VERSION__
@@ -116,3 +117,20 @@ MAX_CHARS_FLAG_NAME = "--max-chars-per-line"
 DEFAULT_MAX_CHARS_PER_LINE = 160
 ELLIPSIS_STRING = " ... "
 DEFAULT_MAX_TARGET_SIZE = 1000000  # 1 MB
+
+
+class Colors(Enum):
+    # these colors come from user's terminal theme
+    foreground = 0
+    bright = 15  # also referred to as "bold color"
+    white = 7
+    black = 256
+    cyan = "cyan"  # for filenames
+    green = "green"  # for autofix
+    yellow = "yellow"  # TODO: benchmark timing output?
+    red = "red"  # for errors
+    bright_blue = "bright_blue"  # TODO: line numbers?
+
+    # these colors ignore user's terminal theme
+    forced_black = 16  # #000
+    forced_white = 231  # #FFF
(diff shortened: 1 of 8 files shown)


semgrep/semgrep/error.py
```diff
@@ -162,13 +162,18 @@ def _stack_trace(self) -> str:
         """
         if self.error_type == "Fatal error":
             error_trace = self.details or "<no stack trace returned>"
-            return f"\n-----[ BEGIN error trace ]-----\n{error_trace}\n-----[ END error trace ]-----\n"
+            return f"\nSemgrep failed to run this rule.\nWe don't know yet why this happened, but happy to help. Please file a GitHub issue or report it on Slack.\n\n====[ BEGIN error trace ]====\n{error_trace}=====[ END error trace ]=====\n"
```
ievans: Can we make this more actionable by providing a link? Modify/re-use PLEASE_FILE_ISSUE_TEXT from semgrep/semgrep/constants.py?

semgrep/semgrep/formatter/text.py
```diff
@@ -237,10 +238,10 @@ def _build_text_output(
             )
 
             if fix:
-                yield f"{with_color('bright_blue', 'autofix:')} {fix}"
+                yield f"{with_color('green', '     ▶▶ Autofix')} {fix}"
```
ievans: @adamboros What do you think about having autofix _after_ the issue? I was always confused by seeing it show up before the line

semgrep/semgrep/error.py
adamboros: Definitely, I didn't know about

semgrep/semgrep/formatter/text.py
adamboros: @ievans is there any case the autofix shows something different for each snippet on the same finding? Cause might be useful to get rid of the repetition as well if not.

adamboros:
Restructured the error messages a bit and cleaned up the code, current error output:

image

semgrep/semgrep/formatter/text.py
ievans: I think it can be different, because it can use metavars

adamboros:
Restructured the separator and fixed a bug where line numbers shifted indentation

image

semgrep/semgrep/constants.py
```diff
@@ -2,6 +2,7 @@
 import re
 from enum import auto
 from enum import Enum
+from types import SimpleNamespace
```
brendongo: Can this be removed?

underyx (approved):

IagoAbal:
For the next time, I think you should have updated the changelog. Unfortunately I missed this when preparing the release notes.

PR #4542, opened 2022-01-19
Allow overriding .semgrepignore with an environment variable

mmcqd:
If the environment variable `SEMGREP_R2C_INTERNAL_EXPLICIT_SEMGREPIGNORE` is set, then Semgrep will use its value as the path to use to find semgrepignore patterns, ignoring any existing `.semgrepignore` file. This is meant to be used by semgrep-action, to allow including ignored files from semgrep-app, without disrupting a user's existing `.semgrepignore` file.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Add an experimental key for internal team use:
   `r2c-internal-project-depends-on` that allows rules to filter based on the
   presence of 3rd-party dependencies at specific version ranges.
+- FOR INTERNAL USE: if the environment variable SEMGREP_R2C_INTERNAL_EXPLICIT_SEMGREPIGNORE
+  is set, then its value will be used as the path to find semgrepignore patterns, overriding
+  any existing .semgrepignore file
 
 ### Changed
 - CLI: parse errors (reported with `--verbose`) appear once per file, 
--- a/semgrep/tests/e2e/snapshots/test_ignores/test_internal_explicit_semgrepignore/results.json
+++ b/semgrep/tests/e2e/snapshots/test_ignores/test_internal_explicit_semgrepignore/results.json
@@ -1,0 +1,205 @@
+{
+  "errors": [],
+  "results": [
+    {
+      "check_id": "rules.eqeq-bad",
+      "end": {
+        "col": 15,
+        "line": 1,
+        "offset": 14
+      },
+      "extra": {
+        "is_ignored": false,
+        "lines": "var x = 0 == 0",
+        "message": "useless comparison",
+        "metadata": {},
+        "metavars": {
+          "$X": {
+            "abstract_content": "0",
+            "end": {
+              "col": 10,
+              "line": 1,
+              "offset": 9
+            },
+            "start": {
+              "col": 9,
+              "line": 1,
+              "offset": 8
+            },
+            "unique_id": {
+              "md5sum": "<masked in tests>",
+              "type": "AST"
+            }
+          }
+        },
+        "severity": "ERROR"
+      },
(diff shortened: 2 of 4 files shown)


emjin:
We should add a unit test. Also---though I don't know if we have the infra for this rn---a semgrep-action test that verifies that app-ignores are respected. This bit us already last time; probably it will bite future devs

brendongo (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Add an experimental key for internal team use:
   `r2c-internal-project-depends-on` that allows rules to filter based on the
   presence of 3rd-party dependencies at specific version ranges.
+- FOR INTERNAL USE: if the environment variable SEMGREP_R2C_INTERNAL_EXPLICIT_SEMGREPIGNORE
```
aryx, line 15: We actually don't want to add it in the changelog. The changelog is something to read for users.
When I do internal refactoring, I don't add anything in the changelog; users do not care about it.
You can add something in https://semgrep.dev/docs/contributing/semgrep-contributing/ maybe.
Also I don't see the utitlity of this feature. Why anybody want to use that?

semgrep/tests/e2e/snapshots/test_ignores/test_internal_explicit_semgrepignore/results.json
```diff
@@ -0,0 +1,205 @@
+{
```
aryx, line 1: This is part of the reason I don't like our python tests based on those snapshots. They are very long,
and not super readable. I far prefer what we have in semgrep-core where basically the test file
contain its own assertion by having the ruleid: or MATCH: special comment, so when you look at the test code,
you also see what we're supposed to return.

semgrep/tests/e2e/test_ignores.py
```diff
@@ -51,3 +51,19 @@ def test_file_not_relative_to_base_path(tmp_path, monkeypatch, snapshot):
     )
     stdout, _ = process.communicate("a")
     snapshot.assert_match(_clean_output_json(stdout), "results.json")
+
+
+def test_internal_explicit_semgrepignore(run_semgrep_in_tmp, tmp_path, snapshot):
+
+    (tmp_path / ".semgrepignore").symlink_to(
```
aryx, line 58: looking at the test, it's hard to see that it works because I need to read an extra file for the code, an extra file for the rule, and then the snapshot to see if it match. There has to be a more declarative way to do.

CHANGELOG.md
mmcqd: My bad, I can make a PR removing it from the changelog, or you can just delete it when you do the release. The utility of the feature, which I mention in the PR description but probably in not enough detail, is that semgrep-action needs to read an extra set of ignore rules from semgrep-app, and append these to whatever is being ignored by a user's .semgrepignore file. But we don't actually want to edit a user's .semgrepignore file, since that's likely to cause confusion. They run semgrep-action and suddenly new stuff is getting ignored by the CLI. So this lets semgrep-action create a temporary file of ignore rules, without editing or renaming the user's .semgrepignore, and force the CLI to follow the new temporary file.

semgrep/tests/e2e/test_ignores.py
mmcqd: I agree that the e2e tests are hard to read. The tests in `unit/` are closer to what you want, since they're just testing individual bits of code (specifically `unit/test_target_manager.py` has a bunch of straightforward "assert output = expected"). It's not entirely clear to me how we'd make e2e tests better, because by definition they have to involve actually calling Semgrep and then looking at the whole output, which contains a bunch of text. The test I added should have had a comment on it though.

PR #4556, opened 2022-01-22
Dockerfile: add support for metavariables where argument expansion is already supported

mjambon:
Dockerfiles allow for variable expansion using the `$FOO` or `${FOO}` syntax. These variables are declared with the `ARG` instruction and their value can be set on the command line at build time. Like for Bash, in a Semgrep pattern `$FOO` will instead stand for a Semgrep metavariable that can match anything, while `${BAR}` is the expansion of the Docker variable `BAR` and `${$VAR}` is the expansion of any Docker variable.

`make test` checks that this works.

Note that we want metavariables in even more locations. This will come in a future PR.

PR checklist:
- [x] Changelog is up-to-date


aryx (approved):
We need to document somewhere this kind of escaping mechanism which
allows metavariables to match expansion variables. We're lucky Docker
support those 2 syntaxes to have this tricky encoding.

mjambon:
@aryx I think we need some form of official/easy-to-find documentation for each language supported by Semgrep. I wrote a section about this in our internal "backlog work" document.

Maybe the best implementation would be to improve the current cheatsheet that appears in the live editor. It would have to allow for extra examples that make sense only for a specific language.

PR #4560, opened 2022-01-23
Send all rules to semgrep-core

emjin:
Instead of sending the rules one at a time to semgrep-core, send them all, thus only invoking semgrep-core once. We expect this to cause a significant speed up for diff scans (or other scans that run many rules on only a few bytes of target).
    
This also requires changing the return object. Matches and errors are still returned in a flat list as before, with the rule id annotated. The profiling information had to be grouped differently. A few fields are lost (run time is now known for each file, not each file and each rule). Also, the progress bar is lost.
    
TODOs:

Fix by next release:
- [ ] This is certainly going to break connections with semgrep-action, perhaps also anything that receives metrics?
- [x] No progress bar
- [x] A timeout anywhere stops all rules from running
- [x] Rename final semgrep output JSON fields

Fix eventually:
- [ ] This PR makes minimal changes necessary to send all the rules. In the future, we will probably want to refactor the code to operate file-wise (which is how results are outputted).
- [ ] This PR does not solve the error reporting problem for rules. Rules are still parsed twice, and the rules sent to semgrep-core are reindented. To change that, we will need to change pattern-ref and also allow semgrep-core to take multiple rules files (or else perform a transformation on the Python side)

Additionally, this PR makes it easy to reproduce semgrep-core runs from semgrep runs with the flag `--dump-command-for-core` (or `-d`), which outputs the exact command semgrep sends semgrep-core, saving the rules and targets files.

PR checklist:
- [x] Changelog is up-to-date


aryx:
Why do we need the spacegrep binary?

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -358,15 +358,12 @@ let iter_targets_and_get_matches_and_exn_to_errors config f targets =
 (* File targeting and rule filtering *)
 (*****************************************************************************)
 
-let rules_for_xlang xlang rules =
```
emjin: @aryx I saw in another PR you exposed this function. I deleted it since I didn't see it used anywhere else, but I can put it back

aryx: yes please, put it back, it's used in deep-semgrep
(and it could be useful also I think when we run semgrep-core without -targets but with simply -lang)

aryx:
Note that you didn't ask for any reviewer yet, so when someone does that I consider the PR in the draft state, in which case I don't  review, so let me know when you're ready for the the PR to be reviewed.

aryx:
(I guess you wait for fixing the CI errors first, which may take some time ...)

aryx:
That's lots of speedup!

semgrep-core/src/engine/Match_tainting_rules.ml
```diff
@@ -173,17 +175,20 @@ let lazy_force x = Lazy.force x [@@profiling]
 (*****************************************************************************)
 
 let check_bis ~match_hook (default_config, equivs)
-    (taint_rules : (Rule.rule * Rule.taint_spec) list) file lang ast =
+    (taint_rule : Rule.rule * Rule.taint_spec) file lang ast =
+  (* @Iago I went and modified this to work one rule at a time *)
```
emjin, line 179: @IagoAbal made some changes here so this runs one rule at a time, possibly there are smarter things to do but I just made the dumbest possible change

aryx:
Boom!

semgrep/tests/unit/test_metric_manager.py
```diff
@@ -133,49 +133,41 @@ def test_timings(snapshot) -> None:
...
+        profiling_data.set_file_times(targets[0], target_a_time, 0.4)
+        profiling_data.set_file_times(targets[1], target_b_time, 1.4)
+        profiling_data.set_rules_parse_time(0.09)
 
         metric_manager.set_run_timings(profiling_data, targets, rules)
 
     assert metric_manager._rule_stats == [
         {
             "ruleHash": "720c14cd416c021bc45d6db0689dd0eb54d1d062bf9f446f85dae0cb5d1438c0",
-            "parseTime": 0.05,
```
emjin, line 162: parseTime and runTime for a rule are no longer metrics we capture

semgrep/tests/e2e/snapshots/test_output/test_sarif_output_with_nosemgrep_and_error/results.sarif
```diff
@@ -75,6 +75,23 @@
                 "text": "useless comparison operation `$X == $X` or `$X != $X`; possible bug?"
               }
             },
+            {
```
emjin, line 78: Unclear whether this should appear twice. Seems to be same rule, different path. Maybe I need to fix some deduping logic?

semgrep/tests/e2e/snapshots/test_ignores/test_file_not_relative_to_base_path/results.json
```diff
@@ -1,27 +1,6 @@
 {
   "errors": [],
   "results": [
-    {
```
emjin, line 4: I think the new behavior for deduping is better

semgrep/tests/e2e/snapshots/test_check/test_subshell_input/results.json
```diff
@@ -1,27 +1,6 @@
 {
   "errors": [],
   "results": [
-    {
```
emjin, line 4: I think the new behavior for deduping is better

semgrep/tests/e2e/snapshots/test_check/test_stdin_input/results.json
```diff
@@ -1,27 +1,6 @@
 {
   "errors": [],
   "results": [
-    {
```
emjin, line 4: I think the new behavior for deduping is better

semgrep/tests/e2e/snapshots/test_check/test_multiple_configs_different_origins/results.json
```diff
@@ -1,48 +1,6 @@
 {
   "errors": [],
   "results": [
-    {
```
emjin: I think the new behavior for deduping is better

semgrep/tests/e2e/snapshots/test_check/test_multi_subshell_input/results.json
```diff
@@ -1,48 +1,6 @@
 {
   "errors": [],
   "results": [
-    {
```
emjin, line 4: I think the new behavior for deduping is better

semgrep/tests/e2e/snapshots/test_check/test_max_memory/results.json
```diff
@@ -34,9 +32,7 @@
           2.022
         ],
         "path": "targets/equivalence/open_redirect.py",
-        "run_times": [
-          2.022
-        ]
+        "run_times": 2.022
```
emjin, line 35: TODO rename fields like this one? Probably a separate PR

interfaces/Output_from_core.atd
```diff
@@ -181,20 +181,26 @@ type skip_reason = [
 
 type time = {
   targets: target_time list;
-  ?rule_parse_time: float option; (* not used in spacegrep *)
+  rules: string list;
```
aryx, line 187: why do we return the list of rules here? Maybe a comment explaining why in the .atd?

semgrep-core/src/engine/Match_search_rules.ml
```diff
@@ -260,6 +262,7 @@ let matches_of_patterns ?range_filter config file_and_more patterns =
   in
   match xlang with
   | Xlang.L (lang, _) ->
+      (* TODO can we only force this once? *)
```
aryx: What is this TODO?

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -369,6 +369,13 @@ let rules_for_xlang xlang rules =
          | Xlang.L (x, _empty), Xlang.L (y, ys) -> List.mem x (y :: ys)
          | (Xlang.LRegex | Xlang.LGeneric | Xlang.L _), _ -> false)
 
+module Rule_table = Map.Make (String)
+
+let mk_rule_table rules =
```
aryx, line 372: I usually use Common.hash_of_list for that instead of instantiating each time a module.

semgrep-core/tests/e2e/target
```diff
@@ -1,4 +1,5 @@
 [ {"path": "tests/e2e/targets/basic.py",
-   "language": "python"
+   "language": "python",
+   "rule_ids": ["-"]
```
aryx, line 3: Is this "-" documented in Input_to_core.atd?

semgrep/tests/e2e/snapshots/test_check/test_timeout_threshold/error_2.txt
```diff
@@ -5,10 +5,7 @@ rules:
 - rules.forcetimeout2
 - rules.forcetimeout3
 No .semgrepignore found. Using default .semgrepignore rules. See the docs for the list of default ignores: https://semgrep.dev/docs/cli-usage/#ignoring-files
-Warning: 2 timeout error(s) in targets/equivalence/open_redirect.py when running the following rules: [forcetimeout,
-forcetimeout2]
-Semgrep stopped running rules on targets/equivalence/open_redirect.py after 2 timeout error(s). See `--timeout-
-threshold` for more info.
+Warning: 1 timeout error(s) in targets/equivalence/open_redirect.py when running the following rules: [forcetimeout]
```
aryx: looks better :)

semgrep/tests/e2e/snapshots/test_check/test_timeout_threshold/results.json
```diff
@@ -17,11 +17,7 @@
       "ignores_time": 2.022,
       "total_time": 2.022
     },
-    "rule_parse_info": [
-      2.022,
-      2.022,
-      2.022
-    ],
+    "rule_parse_info": 2.022,
```
aryx, line 20: looks better too :)

aryx (approved):
That's a long PR. Looks mostly good. We can merge this and iterate from there.

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -386,8 +393,8 @@ let xtarget_of_file config xlang file =
     lazy_ast_and_errors;
   }
 
-let targets_of_config (config : Runner_config.t) :
-    In.targets * Out.skipped_target list =
+let targets_of_config (config : Runner_config.t) (rule_ids : Rule.rule_id list)
```
aryx: maybe instead of rule_ids call it all_rule_ids_when_no_target_file?

brendongo:
Didn't get to this but will review on Monday

semgrep/semgrep/commands/scan.py
```diff
@@ -533,6 +533,7 @@ def convert(
     hidden=True
     # help="WARNING: allow rules to run arbitrary code (pattern-where-python)",
 )
+@click.option("--dump-command-for-core", "-d", is_flag=True, hidden=True)
```
brendongo, line 545: For internal stuff might we be better served by an env var like `SEMGREP_DUMP_CORE_COMMAND`

semgrep/semgrep/rule.py
```diff
@@ -199,6 +200,10 @@ def __repr__(self) -> str:
     def rename_id(self, new_id: str) -> None:
         self._id = new_id
 
+    @property
+    def orig_id(self) -> str:
```
brendongo: Why do we need orig_id?

semgrep/semgrep/core_output.py
```diff
@@ -73,16 +75,16 @@ class CoreMatch:
...
     path: Path
     start: CoreLocation
     end: CoreLocation
     extra: Dict[str, Any]
     metavars: CoreMetavars
 
     @classmethod
-    def parse(cls, raw_json: JsonObject) -> "CoreMatch":
-        rule_id = RuleId(raw_json["rule_id"])
+    def parse(cls, rule_table: Dict[RuleId, Rule], raw_json: JsonObject) -> "CoreMatch":
```
brendongo, line 86: So original intention of the objects in this class is to only handle parsing of semgrep-core output. Any information that we need to combine core output with (in this case rules) should be handled in a different class (i.e. CoreMatch is just core output in object form and RuleMatch takes a Rule and a CoreMatch).

semgrep/semgrep/core_runner.py
```diff
@@ -285,113 +348,89 @@ def _run_rules_direct_to_semgrep_core(
...
+
+                if self._optimizations != "none":
+                    cmd.append("-fast")
+
+                stderr: Optional[int] = subprocess.PIPE
+                if is_debug():
+                    cmd += ["-debug"]
+                    stderr = None
+
+                core_run = sub_run(
```
brendongo, line 380: To prevent code duplication ( and make core command dump actually accurate to what is being run )probably better to print it here instead of duplicating work

semgrep/semgrep/core_runner.py
```diff
@@ -272,6 +261,80 @@ def get_files_for_language(
             ) from ex
         return list(targets)
 
+    def _get_targets(
+        self, rules: List[Rule], target_manager: TargetManager, all_targets: Set[Path]
+    ) -> List[Dict[str, Any]]:
```
brendongo, line 270: Add comment on what this function does

semgrep/semgrep/core_runner.py
```diff
@@ -272,6 +261,80 @@ def get_files_for_language(
...
+
+        for rule in rules:
+            for language in rule.languages:
+                targets = self.get_files_for_language(language, rule, target_manager)
+
+                for target in targets:
+                    all_targets.add(target)
+                    t = (target, language)
+                    if t in target_info:
+                        target_info[t].append(RuleId(rule.orig_id))
```
brendongo: idk if it's safe to use orig_id when elsewhere in the code we consistently use id?

semgrep/semgrep/core_runner.py
```diff
@@ -272,6 +261,80 @@ def get_files_for_language(
...
+                    else:
+                        target_info[t] = [RuleId(rule.orig_id)]
```
brendongo: Also potentially cleaner if in the inner for loop you construct a list and after that add it to target_info instead of constantly updating in place.

brendongo: Do we also need to worry about name collisions with orig_id

semgrep/semgrep/core_output.py
```diff
@@ -196,53 +198,61 @@ def parse(cls, raw_json: JsonObject) -> "CoreSkipped":
...
+        per_rule_timings = [
+            CoreRuleTiming.parse(rule_table, timing)
+            for timing in raw_json["rule_times"]
+        ]
+        run_time = raw_json["run_time"]
+        return cls(target, per_rule_timings, run_time)
 
 
 @attr.s(auto_attribs=True, frozen=True)
 class CoreTiming:
```
brendongo, line 236: Similarly to comment above, IMO CoreTiming should just be timing json returned by semgrep-core and it is up to ProfileManager class to combine list of Rules and CoreTiming

semgrep-core/src/engine/Match_search_rules.ml
emjin: I was wondering if we could restructure where we choose to parse the file now

semgrep/semgrep/rule.py
emjin: Need it to replicate previous behavior

semgrep/semgrep/commands/scan.py
emjin: Env vars are so annoying for when you want to flip back and forth though

semgrep/semgrep/core_runner.py
emjin: saaaaaaad but fair

semgrep/semgrep/core_runner.py
emjin: Not certain I see how this works. My outside loop is iterating rule-wise but I'm building a dictionary target-wise

semgrep/semgrep/core_output.py
emjin: Leaving as is for now, this is a more minimal change from the original

semgrep/tests/e2e/snapshots/test_check/test_multi_subshell_input/results.json
brendongo: Yeah. Though not sure why it wasn't deduping this before?

semgrep/tests/e2e/snapshots/test_check/test_multiple_configs_different_origins/results.json
brendongo: This one is a bug since they have different check_ids

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/badpattern/error-in-color.txt
```diff
@@ -2,5 +2,5 @@ Running 1 rules...
 [31m[22m[24m[31m[41m[22m[24m[[0m[38;5;231m[41m[1m[24mERROR[0m[31m[41m[22m[24m][0m Semgrep Core — Pattern parse error
 An error occurred while invoking the Semgrep engine. Please help us fix this by creating an issue at https://github.com/returntocorp/semgrep
 
-In rule eqeq-is-bad: Invalid pattern for Python
+In rule rules.syntax.eqeq-is-bad: Invalid pattern for Python
```
emjin, line 5: @brendongo Here is an illustrative change

emjin: I'm a bit concerned because full paths can be quite long

semgrep/tests/e2e/snapshots/test_check/test_timeout_threshold/error_2.txt
```diff
@@ -5,10 +5,8 @@ rules:
 - rules.forcetimeout2
 - rules.forcetimeout3
 No .semgrepignore found. Using default .semgrepignore rules. See the docs for the list of default ignores: https://semgrep.dev/docs/cli-usage/#ignoring-files
-Warning: 2 timeout error(s) in targets/equivalence/open_redirect.py when running the following rules: [forcetimeout,
-forcetimeout2]
-Semgrep stopped running rules on targets/equivalence/open_redirect.py after 2 timeout error(s). See `--timeout-
```
brendongo, line 10: Do we still keep old behavior of not running a rule after it times out 2 times?

semgrep/tests/e2e/snapshots/test_output/test_sarif_output_with_nosemgrep_and_error/results.sarif
brendongo: I think this is fine. I think before we were adding only rules that found results but I think it's more correct anyway to have all rules run in the sarif output

semgrep/tests/e2e/snapshots/test_check/test_timeout_threshold/error_2.txt
emjin: Hmmm I think timeouts now crash all semgrep. We might need to change that

semgrep/tests/e2e/snapshots/test_output/test_sarif_output_with_nosemgrep_and_error/results.sarif
brendongo: Probably add to changelog to that effect

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/badpattern/error-in-color.txt
brendongo: Seems fine to have more info. if anything, easier to find where the offending rule is now and rules from semgrep.dev shouldnt get any change in behavior

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/badpattern/error.json
```diff
@@ -3,14 +3,14 @@
...
-      "rule_id": "eqeq-is-bad",
+      "message": "[ERROR] Semgrep Core \u2014 Pattern parse error\nAn error occurred while invoking the Semgrep engine. Please help us fix this by creating an issue at https://github.com/returntocorp/semgrep\n\nIn rule rules.syntax.eqeq-is-bad: Invalid pattern for Python\n",
+      "rule_id": "rules.syntax.eqeq-is-bad",
       "spans": [
         {
           "config_end": {
             "col": 24,
             "line": 4,
-            "offset": 60
+            "offset": 73
```
brendongo, line 13: Are the offset changes here due to rewriting the rule yaml file?

semgrep/tests/e2e/snapshots/test_semgrep_core_parse_error/test_rule_parser__failure__error_messages/invalid_python.py/error.txt
```diff
@@ -4,7 +4,6 @@ rules:
 - rules.assert-eqeq-is-ok
 - rules.eqeq-is-bad
 No .semgrepignore found. Using default .semgrepignore rules. See the docs for the list of default ignores: https://semgrep.dev/docs/cli-usage/#ignoring-files
-skipped 'targets/bad/invalid_python.py' [rule assert-eqeq-is-ok]: irrelevant_rule: target doesn't contain some elements required by rule 'assert-eqeq-is-ok'
```
brendongo, line 7: Do we still report files skipped due to irrelevant rule?

semgrep/tests/e2e/snapshots/test_rule_parser/test_rule_parser__failure__error_messages/badpattern/error.json
emjin: Yes, `rules.syntax.` is 13 characters

emjin: Yes, `rules.syntax.` is 13 characters

semgrep/tests/e2e/snapshots/test_semgrep_core_parse_error/test_rule_parser__failure__error_messages/invalid_python.py/error.txt
emjin: Yes; this is because, having failed to parse it once, we now don't try to filter irrelevant rules

semgrep/semgrep/commands/scan.py
brendongo: yes but with this we need to send this flag all the way down increasing the interface of lots of functions for an internal thing

PR #4571, opened 2022-01-25
feat(baseline): Enable semgrep to perform baseline scans

brendongo:
This commit introduces the `--baseline-commit` flag. When given a commit hash (or valid git identifies like `HEAD^1`),
`semgrep` will only show findings that exist now but did not exist at said hash.

Semgrep accomplishes this by performing an initial scan,  checking out the older state, performing another scan, and
then removing findings from the former that appear in the latter. It considers findings the same if they have the same rule_id, path, and syntactic context (the lines that match the semgrep rule). Because of this we avoid refiring on findings when they simply have been moved around the file. However a limitation of this is that if a finding matches on a large section of lines (with a `...`, for example) some file changes will still show up as a `new` finding.

This behavior also interacts with inline ignores. Table below summarizes what we think it means for the 9 possible cases. `semgrep --baseline-commit` will only show findings that are either `New` or `Unignored`.

|                         |                                  |                                       |                                                      |
| ------------- | ------------------- | ---------------------- | ------------------------------- |
|                        | Finding in Baseline  | No Finding in Baseline | Finding in Baseline with nosem |
| Finding in Head  | Unchanged  | New  |  Unignored |
| No Finding in Head  | Fixed  | Unchanged | Fixed |
| Finding in Head with Nosem | Ignored | New and Ignored| Unchanged |

PR checklist:
- [ ] Changelog is up-to-date


aryx:
Haha, the license/cla got blocked on you :)

brendongo:
Will need to open an issue to handle renames as baseline equivalent

semgrep/semgrep/core_output.py
```diff
@@ -354,7 +354,6 @@ def convert_to_rule_match(match: CoreMatch, rule: Rule) -> RuleMatch:
                 start=match.start,
                 end=match.end,
                 extra=match.extra,
-                lines_cache={},
```
brendongo, line 357: These were unused so got rid of them as part of refactor to make RuleMatch.lines usable with baseline scanning

semgrep/semgrep/core_runner.py
```diff
@@ -411,7 +398,7 @@ def invoke_semgrep(
             errors,
             all_targets,
             profiling_data,
-        ) = self._run_rules_direct_to_semgrep_core(rules, target_manager, profiler)
```
brendongo, line 415: profiler was unused

semgrep/semgrep/git.py
```diff
@@ -0,0 +1,332 @@
+import subprocess
```
brendongo, line 1: Mostly copies from https://github.com/returntocorp/semgrep-action/blob/5cb31fb22ea0fffbae63b5ad0986d42caa2b562c/src/semgrep_agent/targets.py with necessary changes to work in semgrep-cli

semgrep/semgrep/semgrep_main.py
```diff
@@ -140,6 +141,105 @@ def invoke_semgrep(
     return json.loads(output_handler._build_output())  # type: ignore
 
 
+def run_rules(
```
brendongo, line 144: Refactored into function so could be reused to scan head and baseline

brendongo:
Still missing e2e tests but unit tests already included

semgrep/semgrep/git.py
```diff
@@ -0,0 +1,332 @@
...
+            raise Exception(
+                f"Found files that are untracked by git but exist in {self._base_commit}",
+                "Running a baseline scan will cause changes to be overwritten, so aborting.",
+                f"Please commit or stash your untracked changes in these paths: {overlapping_paths}.",
+            )
+
+    @contextmanager
+    def baseline_context(self) -> Iterator[None]:
+        """
+        Runs a block of code on files from the current branch HEAD.
```
emjin: I'm extremely confused about what this description means

semgrep/semgrep/semgrep_main.py
```diff
@@ -244,52 +345,47 @@ def main(
...
+        profiling_data,
+    ) = run_rules(filtered_rules, target_manager, core_runner, output_handler)
+    profiler.save("core_time", core_start_time)
+    output_handler.handle_semgrep_errors(semgrep_errors)
 
-        for rule in dependency_aware_rules:
+    # Run baseline if needed
+    if baseline_handler:
+        logger.info(f"Running baseline scan with base set to: {baseline_commit}")
+        with baseline_handler.baseline_context():
```
emjin: Does this need to be in a try-catch? Or are errors here handled somewhere else? If so please update main docstring

emjin (approved):
Approved assuming we will add more tests

semgrep/semgrep/semgrep_main.py
brendongo: Good call

xmo-odoo:
Question about this PR: does it have some sort of cache which can be pre-populated with the baseline scans in order to avoid the extra validations?

e.g. let's say there's a bunch of PR against a base branch, by default `--baseline-commit` would have to perform a full scan of the base branch in all PRs in order to compute a baselined diagnostic report. But with some setup and caching the project could compute and store the base branch's report (probably useful anyway as a statistics / tracking device), then retrieve that from the per-PR jobs and compute things from that baseline directly.

PR #4575, opened 2022-01-26
Correctly raise Timeout in the presence of multiples rules

aryx:
Emma found this problem while passing all the rules from
semgrep to semgrep-core.

test plan:
semgrep-core -rules ~/force_timeout.yaml ~/force_timeout.py -timeout 2
-lang python
now correctly raise a timeout instead of continuing forever.

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Match_search_rules.ml
+++ b/semgrep-core/src/engine/Match_search_rules.ml
@@ -989,15 +989,5 @@ let check ~match_hook default_config rules file_and_more =
   |> List.map (fun (r, pformula) ->
          let rule_id = fst r.R.id in
          Common.profile_code (spf "real_rule:%s" rule_id) (fun () ->
-             try
-               check_rule r match_hook default_config pformula file_and_more
-               (* TODO: why do we intercept errors here? We already do that in
-                * Run_semgrep.ml. *)
-             with exn when not !Flag_semgrep.fail_fast ->
-               {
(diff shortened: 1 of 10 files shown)


semgrep-core/src/engine/Match_search_rules.ml
```diff
@@ -989,15 +989,5 @@ let check ~match_hook default_config rules file_and_more =
...
          let rule_id = fst r.R.id in
          Common.profile_code (spf "real_rule:%s" rule_id) (fun () ->
-             try
-               check_rule r match_hook default_config pformula file_and_more
-               (* TODO: why do we intercept errors here? We already do that in
-                * Run_semgrep.ml. *)
-             with exn when not !Flag_semgrep.fail_fast ->
-               {
-                 RP.matches = [];
-                 errors = [ E.exn_to_error ~rule_id:(Some rule_id) file exn ];
```
emjin, line 999: Hm I think it might have been to attach the rule_id, we'll probably have some failing tests

aryx: We can do the same trick I did in Run_semgrep.ml where there's a global that remember the last mini_ruleid and set it in Match_search_rules.ml and use it in Run_semgrep.ml

emjin (approved):

PR #4577, opened 2022-01-26
Dockerfile: Add support for metavariables and ellipses in the remaining places

mjambon:
Metavariables and anonymous ellipses are now supported everywhere I planned to support them except for one tricky case.

This leaves 3 known bugs or limitations, which I left in the `tests/dockerfile/todo` folder. I fixed two of them and they're pending approval in tree-sitter-dockerfile (https://github.com/camdencheek/tree-sitter-dockerfile/pull/13, https://github.com/camdencheek/tree-sitter-dockerfile/pull/14). The third issue is about supporting metavariables for the keys in the ENV command. It has tricky conflicts so I'll just leave it there until somebody shows interest in the feature.

Note the sheer number of test cases. This was necessary (and caught some bugs) since each dockerfile instruction has its own syntax.

I'm leaving the documentation checkbox empty because I think we need language-specific documentation for Dockerfile, but we don't have this in place for any of the other languages. Setting this up is beyond the scope of this PR.

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/tests/dockerfile/arg-metavar.dockerfile
```diff
@@ -0,0 +1,7 @@
+# MATCH:
+ARG foo
+
+# MATCH:
+ARG bar
+
+ARG foo=42
```
aryx, line 7: I would have expected a match there too. That's usually how things work for other language where a metavariable can match a var, but also any expression.
Maybe in the context of an ARG it makes sense to restrict a little bit more things. We'll see how user react.

aryx (approved):

Issue #4228, opened 2021-11-08
Missing detection for try method value


LewisArdern:
**Describe the bug**
Semgrep currently fails to detect values within a try statement https://semgrep.dev/s/0ykj
```
try ($VALUE) { body }
```
In java semgrep translates it to try { body } in the generic AST and completely forgets about $VALUE.

Semgrep should detect the $VALUE of the try not just the body. There are a lot of cases where code examples in Java show this, and it would be important to detect the use.

**To Reproduce**
https://semgrep.dev/s/0ykj line 51

**Expected behavior**
It should detect the value within line 51.

**Screenshots**
N/A

**What is the priority of the bug to you?**

- [ ] P1: important to fix or quite annoying

**Environment**
https://semgrep.dev/s/0ykj

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-668/missing-detection-for-try-method-value.

IagoAbal:
We could perhaps encode this combining `WithUsingResource` and `Try`?

LewisArdern:
I'm seeing this hasn't been fixed, we're running into missing catches on production code we're trying to scan 

Here's another example: 
* https://semgrep.dev/s/6gjB

See: https://github.com/returntocorp/semgrep/issues/717 which said it was fixed, and well i think it hasnt been

emjin:
Should match now! https://semgrep.dev/s/0ykj?version=develop. Fix will go out with the release tomorrow

PR #4595, opened 2022-02-01
Support matching resources in Java try-with-resources

### LINKED ISSUE #717: Java: semgrep does not match the 'try-with-resources' pattern.
**Describe the bug**
The 'try-with-resources' pattern was added in Java 8: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

This example I was using as a test case uses this pattern, but semgrep doesn't pick it up:
https://find-sec-bugs.github.io/bugs.htm#OBJECT_DESERIALIZATION

**To Reproduce**
This pattern should match both cases:
https://semgrep.live/GD1

It looks like it just gets skipped in the AST altogether:
image

emjin:
Closes #717, #4228

Translate try with resources to WithUsingResource and Try. This allows people to find problems in how the resources are acquired.

Test plan: make test

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/parsing/pfff/java_to_generic.ml
```diff
@@ -421,6 +421,19 @@ and arguments v : G.argument list G.bracket = bracket (list argument) v
...
+  | Left v ->
+      let ent, v = var_with_init v in
+      G.DefStmt (ent, G.VarDef v) |> G.s
+  | Right e -> G.ExprStmt (expr e, unsafe_fake "") |> G.s
+
+and resources (t1, v, t2) =
+  match v with
+  | [] -> G.Block (t1, [], t2) |> G.s
+  | [ v ] -> resource v
+  | _ :: _ -> G.Block (t1, list resource v, t2) |> G.s
```
IagoAbal: I think you could merge the cases `[]` and `_ :: _`. Do we need a special case for `[ v ]` ?

semgrep-core/src/parsing/pfff/java_to_generic.ml
```diff
@@ -421,6 +421,19 @@ and arguments v : G.argument list G.bracket = bracket (list argument) v
 
 and fix_op v = H.conv_incr v
 
+and resource (v : resource) : G.stmt =
+  match v with
+  | Left v ->
+      let ent, v = var_with_init v in
+      G.DefStmt (ent, G.VarDef v) |> G.s
+  | Right e -> G.ExprStmt (expr e, unsafe_fake "") |> G.s
```
IagoAbal: Can easily avoid `unsafe_fake` here by passing a token to `resource`, which we have readily available in `resources`.

semgrep-core/src/parsing/pfff/java_to_generic.ml
emjin: I wasn't sure if it would be better to avoid the unnecessary Block, can take it out

IagoAbal (approved):
Nice! Thanks for taking care of this @emjin! I would appreciate if you get rid of the `unsafe_fake` before merging, we should avoid unsafe fake tokens unless it's really a hassle to use safe ones.

semgrep-core/src/parsing/pfff/java_to_generic.ml
IagoAbal: I would keep the translation trivial unless there is a good reason to save that `Block`. If you prefer to save the `Block` I would merge `[]` and `_ :: _` since they are the same. But these are just minor observations anyways. It's your call!

semgrep-core/src/parsing/pfff/java_to_generic.ml
```diff
@@ -421,6 +421,15 @@ and arguments v : G.argument list G.bracket = bracket (list argument) v
...
 and fix_op v = H.conv_incr v
 
+and resource t (v : resource) : G.stmt =
+  match v with
+  | Left v ->
+      let ent, v = var_with_init v in
+      G.DefStmt (ent, G.VarDef v) |> G.s
+  | Right e -> G.ExprStmt (expr e, t) |> G.s
+
+and resources (t1, v, t2) = G.Block (t1, list (resource t2) v, t2) |> G.s
```
nbrahms: What happens to this code:
```
    try (
        java.util.zip.ZipFile zf =
             new java.util.zip.ZipFile(zipFileName);
        java.io.BufferedWriter writer = 
            java.nio.file.Files.newBufferedWriter(outputFilePath, charset)
    ) {
        // Enumerate each entry
        for (java.util.Enumeration entries =
                                zf.entries(); entries.hasMoreElements();) {
            // Get the entry name and write it to the output file
            String newLine = System.getProperty("line.separator");
            String zipEntryName =
                 ((java.util.zip.ZipEntry)entries.nextElement()).getName() +
                 newLine;
            writer.write(zipEntryName, 0, zipEntryName.length());
        }
    }
```
(found on the Java documentation at https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html)

nbrahms: E.g., can we have a test that matches against multiple resources?

emjin: It's a block of resources (the semicolon token is lost, may be a problem one day). I added a test to cover it

nbrahms (approved):
lgtm

emjin:
Error because of https://github.com/returntocorp/semgrep-rules/pull/1763, I forget what the preferred way to resolve the cycle is

Edit: answer is merge PR with errors to develop branch of semgrep-rules

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 - Handle missing target files without raising an exception (#4462)
+- Match resources in Java try-with-resources statements (#4228)
```
IagoAbal, line 22: Just spotted this when I was going to do the release, but note that you added the entry in the wrong place.

PR #4598, opened 2022-02-01
ci(tests): Commit snapshot changes on failure

underyx:
PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -167,7 +167,8 @@ jobs:
       - name: Comment about any snapshot updates
         if: ${{ failure() && steps.snapshot-commit.outputs.pushed == 'true' }}
         run: |
-          echo ":camera_flash: The pytest shapshots changed in your PR" >> /tmp/message.txt
+          echo ":camera_flash: The pytest shapshots changed in your PR." >> /tmp/message.txt
+          echo "Please carefully review these changes and make sure they are intended:" >> /tmp/message.txt
           echo >> /tmp/message.txt
           echo "1. Review the changes at https://github.com/returntocorp/semgrep/commit/${{ steps.snapshot-commit.outputs.commit_long_sha }}" >> /tmp/message.txt
           echo "2. Accept the new snapshots with" >> /tmp/message.txt
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,12 +4,12 @@ on:
   workflow_dispatch:
   pull_request:
     paths-ignore:
-      - '**.md'
+      - "**.md"
   push:
     branches:
       - develop
     paths-ignore:
-      - '**.md'
+      - "**.md"
 
 jobs:
   build-core:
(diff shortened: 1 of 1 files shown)


underyx:
The last snapshot commit (https://github.com/returntocorp/semgrep/commit/c1183b454e53026deff5ba27f3c5bf9b2aaf1693) included some additional build artifact files.

So I changed the job to limit which directory is [passed to `git add` like documented](https://github.com/EndBug/add-and-commit#inputs): `+ add: semgrep/tests/e2e/snapshots`. I'm not gonna set up another round of testing this with a fake snapshot failure and than changing the PR back cause it'd be another 40 minutes :weary:

.github/workflows/tests.yml
```diff
@@ -137,14 +140,43 @@ jobs:
...
+          add: semgrep/tests/e2e/snapshots
+          default_author: github_actions
+          message: "Update pytest snapshots"
+          new_branch: snapshot-updates-${{ github.run_id }}-${{ github.run_attempt }}
+      - name: Comment about any snapshot updates
+        if: ${{ failure() && steps.snapshot-commit.outputs.pushed == 'true' }}
+        run: |
+          echo ":camera_flash: The pytest shapshots changed in your PR" >> /tmp/message.txt
+          echo >> /tmp/message.txt
+          echo "1. Review the changes at https://github.com/returntocorp/semgrep/commit/${{ steps.snapshot-commit.outputs.commit_long_sha }}" >> /tmp/message.txt
```
nbrahms, line 173:
```suggestion
          echo "Please carefully review these changes and make sure they are intended:" >> /tmp/message.txt
          echo "1. Review the changes at https://github.com/returntocorp/semgrep/commit/${{ steps.snapshot-commit.outputs.commit_long_sha }}" >> /tmp/message.txt
```

emjin (approved):
Looks good to me! Are there any security implications associated with giving write access?

brendongo (approved):
Like how it automates generating but requires manual work to get changes in the PR

.github/workflows/tests.yml
```diff
@@ -137,14 +140,43 @@ jobs:
...
+        with:
+          add: semgrep/tests/e2e/snapshots
+          default_author: github_actions
+          message: "Update pytest snapshots"
+          new_branch: snapshot-updates-${{ github.run_id }}-${{ github.run_attempt }}
+      - name: Comment about any snapshot updates
+        if: ${{ failure() && steps.snapshot-commit.outputs.pushed == 'true' }}
+        run: |
+          echo ":camera_flash: The pytest shapshots changed in your PR" >> /tmp/message.txt
+          echo >> /tmp/message.txt
```
underyx:
```suggestion
          echo ":camera_flash: The pytest shapshots changed in your PR." >> /tmp/message.txt
          echo "Please carefully review these changes and make sure they are intended:" >> /tmp/message.txt
          echo >> /tmp/message.txt
```

.github/workflows/tests.yml
underyx: applied your suggestion with slightly different spacing!

underyx:
> Looks good to me! Are there any security implications associated with giving write access?

Thanks for calling this out @emjin! Actually by default [the token has write access to _everything_](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token). So by listing the two write permissions here, we're making it more restrictive.

PR #4534, opened 2022-01-17
Add core_runner.pyi

aryx:
This is an experiment is starting to use real interface files to
define Python modules. This is similar to the .mli defining
the interface of .ml files in OCaml.

test plan:
mypy --config-file ../mypy.ini semgrep/

PR checklist:
- [x] Documentation is up-to-date
- [x] Changelog is up-to-date
- [x] Change has no security implications (otherwise, ping security team)

PR #4604, opened 2022-02-02
New semgrep-core -pycheck, to check Python _.pyi interface

aryx:
This is a followup of
https://github.com/returntocorp/semgrep/pull/4534
This is an experiment is starting to use real interface files to
define Python modules. This is similar to the .mli defining
the interface of .ml files in OCaml.

test plan:
cd semgrep/semgrep
make check

PR checklist:
- [x] Changelog is up-to-date


aryx:
This is also a good way to dogfood the graph_code_AST.ml name resolving (for Python at least).
Next steps are to add a deadcode checker so we can start to remove dead Python code.

semgrep-core/src/experiments/pycheck/Pycheck.ml
```diff
@@ -0,0 +1,217 @@
...
+(* Simple Python interface checker.
+ *
+ * The term "interface" above is used in the original sense of the word,
+ * that is the exposed API of a module (not an interface in the Java sense).
+ *
+ * IMHO, the most basic software engineering practice is to define
+ * interfaces and have the compiler (or whatever tool) making sure this
+ * interface is respected. This comes from David Parnas in the 70's.
+ * Unfortunately, Python does not support interface, so I had to make a tool
+ * to overcome this limimation by offering the possibility to Python
```
underyx:
```suggestion
 * to overcome this limitation by offering the possibility to Python
```

semgrep-core/src/experiments/pycheck/dune
```diff
@@ -0,0 +1,19 @@
+; this code could at some point live outside semgrep (e.g., in codegraph
+; or codecheck), so do add too many dependencies, and especiall do not
```
underyx:
```suggestion
; or codecheck), so don't add too many dependencies, and especially do not
```

underyx:
Hmmm I don't have any context on why and what we're doing here, but I'm approaching it with a bit of fear :D

1. I see that there are docstrings in the .pyi files. Are they the same docstrings we have in the .py files? Will we need to manually keep the documentation in sync?
2. Am I understanding it right that if I wanted to change the function signature of something from `str` to `Optional[str]` now I'd have to edit two separate files? That's gonna be somewhat annoying :/ Especially if I'm trying to update something via the github.com editor to save time, or via PR comment suggestions.
3. What's going to tell me if I have a mismatch between the foo.py and foo_.pyi? Will I usually have to wait until pre-commit or CI time for me to realize I've been doing something wrong?

aryx:
I can explain more during the next PA standup. But for some context:
From https://en.wikipedia.org/wiki/Interface_(computing) 
"A key principle of design is to prohibit access to all resources by default, allowing access only through well-defined entry points, i.e., interfaces.[7] Software interfaces provide access to computer resources (such as memory, CPU, storage, etc.) of the underlying computer system; direct access (i.e., not through well-designed interfaces) to such resources by software can have major ramifications—sometimes disastrous ones—for functionality and stability.[citation needed]

Interfaces between software components can provide constants, data types, types of procedures, exception specifications, and method signatures. Sometimes, public variables are also defined as part of an interface.[8]

The interface of a software module A is deliberately defined separately from the implementation of that module. The latter contains the actual code of the procedures and methods described in the interface, as well as other "private" variables, procedures, etc. Another software module B, for example the client to A, that interacts with A is forced to do so only through the published interface. One practical advantage of this arrangement is that replacing the implementation of A with another implementation of the same interface should not cause B to fail—how A internally meets the requirements of the interface is not relevant to B, which is only concerned with the specifications of the interface. (See also Liskov substitution principle.)[citation needed]"

aryx:
regarding your questions.
1. Right now they are the same, but really they should be different. The one in the .pyi are meant to the user of the function/class, and should not be in the implemetation code but in the interface.
2. yes you'll have to edit 2 files.
3. make check will run semrep-core -pycheck which will make sure that the interface boundaries are respected (other modules can't use entities not defined in the interface file), and that the interface is a subset of the implementation file.
4.

underyx:
> 1. Right now they are the same, but really they should be different. The one in the .pyi are meant to the user of the function/class, and should not be in the implemetation code but in the interface.

Well this will be uncomfortable, because when I'm using the function/class myself, I'll be typically looking at the function/class and not the interface. This is because editors can jump to definition of a function, and bring in docstrings of a function as hover tooltips, and stuff like that. I don't think I can configure my editor to understand where the interface is defined.

> 2. yes you'll have to edit 2 files.

Sad 😞 

> 3. make check will run semrep-core -pycheck which will make sure that the interface boundaries are respected (other modules can't use entities not defined in the interface file), and that the interface is a subset of the implementation file.

Normally if I'm violating type definitions, my editor would immediately highlight the offending code. We're also losing this capability then.

aryx:
> > 1. Right now they are the same, but really they should be different. The one in the .pyi are meant to the user of the function/class, and should not be in the implemetation code but in the interface.
> 
> Well this will be uncomfortable, because when I'm using the function/class myself, I'll be typically looking at the function/class and not the interface. This is because editors can jump to definition of a function, and bring in docstrings of a function as hover tooltips, and stuff like that. I don't think I can configure my editor to understand where the interface is defined.

I see. Then keep the docstring in the .py. You can ignore this _.pyi if you want. This is an experiment.

> 
> > 2. yes you'll have to edit 2 files.
> 
> Sad disappointed
> 
> > 3. make check will run semrep-core -pycheck which will make sure that the interface boundaries are respected (other modules can't use entities not defined in the interface file), and that the interface is a subset of the implementation file.
> 
> Normally if I'm violating type definitions, my editor would immediately highlight the offending code. We're also losing this capability then.

No you still gets that. Like I said this is just an experiment. You can ignore those _.pyi if you want and everything will work as before. For now, I can be the only one who update those _.pyi file and the only one who run 'make check'.

aryx:
For now just consider it as an effort to document some modules and some dogfooding for deep-semgrep as
the code behind pycheck relies on the same Graph_code_AST.ml to correctly resolve names across files.

PR #4605, opened 2022-02-02
optimizing: Make bloom filter consider import file names

IagoAbal:
Previously, Bloom annotation did not extract any meaningful string from
a pattern such as `import { $X } from 'v8'`, making it 10x slower or
more than another pattern such as `import { foo } from 'v8'`, even in a
file where there was no `import ... from 'v8'` whatsoever.

This should fix a perf problem affecting a customer.

test plan:
$ semgrep -l ts --time -e 'import { $X } from "v8"' semgrep/parsing-stats/lang/typescript/tmp/*
 #^ This now runs in ~14s, about twice as fast as previously

PR checklist:
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/optimizing/Analyze_pattern.ml
+++ b/semgrep-core/src/optimizing/Analyze_pattern.ml
@@ -86,8 +86,10 @@ let extract_strings_and_mvars ?lang any =
               when str <> "..."
                    && (not (Metavariable.is_metavar_name str))
                    && (* deprecated *) not (Pattern.is_regexp_string str) ->
-                (* Semgrep can match "foo" against "foo/bar", so we just overapproximate taking the sub-strings. *)
-                Common.split "/\\|\\\\" str
+                (* Semgrep can match "foo" against "foo/bar", so we just
+                 * overapproximate taking the sub-strings, see
+                 * Generic_vs_generic.m_module_name_prefix. *)
+                Common.split {|/\|\\|} str
                 |> List.iter (fun s -> Common.push s strings);
                 k x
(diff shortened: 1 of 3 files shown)


semgrep-core/src/optimizing/Analyze_pattern.ml
```diff
@@ -73,6 +77,20 @@ let extract_strings_and_mvars ?lang any =
...
+          (fun (k, _) x ->
+            match x with
+            | { d = ImportFrom (_, FileName (str, _), _, _); _ }
+            | { d = ImportAs (_, FileName (str, _), _); _ }
+            | { d = ImportAll (_, FileName (str, _), _); _ }
+              when str <> "..."
+                   && (not (Metavariable.is_metavar_name str))
+                   && (* deprecated *) not (Pattern.is_regexp_string str) ->
+                (* Semgrep can match "foo" against "foo/bar", so we just overapproximate taking the sub-strings. *)
+                Common.split "/\\|\\\\" str
```
aryx: wow, that's lots of \

aryx (approved):

semgrep-core/src/optimizing/Analyze_pattern.ml
IagoAbal: 😅 yeah indeed, I changed it to use raw strings.

IagoAbal:
The CI failures (which didn't happen before) are due to `tests/e2e/test_check.py::test_auto_config`. I don't think these changes should have affected that test.

PR #4616, opened 2022-02-04
Add ignore log to output

underyx:
Closes PA-757
Closes PA-756

PR checklist:
- [x] Changelog is up-to-date


semgrep/semgrep/target_manager.py
```diff
@@ -73,6 +84,196 @@ def converted_pipe_targets(targets: Sequence[str]) -> Iterator[Sequence[str]]:
...
+    rule_size_limit: Dict[str, Set[Path]] = Factory(lambda: defaultdict(set))
+
+    @property
+    def size_limited_paths(self) -> Set[Path]:
+        """
+        All paths that were skipped because they were larger than the max_target_bytes.
+
+        This is needed because sometimes a file is not targeted at first
+        when we initially filter files that are too large,
+        but then a rule includes it.
```
brendongo: Is this actually what happens? Shouldn't a rule include only ever results in a smaller or equal number of files. i.e. given set A, passing A through rule_includes should not results in a set with elements not in A.

semgrep/semgrep/target_manager.py
```diff
@@ -73,6 +84,196 @@ def converted_pipe_targets(targets: Sequence[str]) -> Iterator[Sequence[str]]:
...
+
+        if not skip_fragments:
+            return "no files were skipped"
+
+        message = "skipped: " + ", ".join(skip_fragments)
+
+        message += "\nfor a detailed list of skipped files, run semgrep with the --verbose flag\n"
+        return message
+
+    def yield_verbose_lines(self) -> Iterator[Tuple[Literal[0, 1, 2], str]]:
```
brendongo, line 161: Is the intention to have this class also handle the json output?

semgrep/semgrep/target_manager.py
```diff
@@ -84,6 +285,8 @@ class TargetManager:
     If skip_unknown_extensions is False then targets with extensions that are
     not understood by semgrep will always be returned by get_files. Else will discard
     targets with unknown extensions
+
+    TargetManager not to be confused with https://jobs.target.com/search-jobs/store%20manager
```
brendongo, line 275: :P

semgrep/semgrep/target_manager.py
brendongo: IMO would be cleaner if the thing outputing handled this code. That way we wouldnt need this class to have a pointer to it's parent.

semgrep/semgrep/target_manager.py
```diff
@@ -391,9 +602,20 @@ def filtered_files(self, lang: Language) -> FrozenSet[Path]:
         explicit_files, _ = partition_set(lambda p: p.is_file(), files)
 
         targets = self.expand_targets(directories, lang, self.respect_git_ignore)
-        targets = self.filter_includes(targets, self.includes)
-        targets = self.filter_excludes(targets, [*self.excludes, ".git"])
-        targets = self.filter_by_size(targets, self.max_target_bytes)
+        targets = self.filter_includes(
```
brendongo: I wonder if it might be less cognitive load to instead of having the decorator and specifying a removal_log, just do it inline? 
i.e.
```
targets, filtered_out = self.filter_includes(self.includes, targets)
log_removed_paths(filtered_out, CLI_INCLUDES)
```

Much more obvious than having to know to pass a removal_log into a decorator that will silently do nothing if removal_log is not specified

brendongo (changes requested):

semgrep/semgrep/target_manager.py
underyx: My thoughts that led to deciding against that were: it would more than double the amount of symbols copy-pasted here (and the other method that does filtering as well), so updating this would get a lot more error-prone

underyx: Basically trying to mitigate the last line effect https://pvs-studio.com/en/blog/posts/cpp/0260/

brendongo: IMO that's fine and those errors should be caught in tests anyway. Trying to decrease the complexity in reading the code. With the decorator someone has to go to the function declaration of filter_includes and the function declaration of the decorator. With the suggested approach it is clear even without loading that state that we are filtering out things and logging. the removed things.

brendongo: re optimizing for reading: https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds

and zen of python `Explicit is better than implicit`

brendongo: Alternatively we can modify the filter_includes and filter_excludes to always perform the log behavior and change the function name to reflect that.

CHANGELOG.md
```diff
@@ -5,13 +5,19 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ## Unreleased
 
 ### Added
+
```
brendongo: Also can we not include all these changes to the changelog file

semgrep/semgrep/target_manager.py
underyx: I made this assumption when seeing that we do a second size filter pass after applying the rule includes. Good question though, let me check.

brendongo: Looks like this is necessary for explicit files

mypy-tests.ini
```diff
@@ -31,6 +31,9 @@ disallow_untyped_decorators = False
 # Implicit optional
 no_implicit_optional = True
 
+# allow functions with no return statement since assert_foo helpers are common with pytest
+warn_no_return = False
```
brendongo, line 35: I would just add a dummy return tbh. (but also i don't think mypy runs on our tests anyway)

brendongo (approved):

mypy-tests.ini
underyx: I'm worried that the mypy error message will lead people astray if they're not experienced with python. We had a comparison function something like:

```py
def _cmp(a, b):
  return a > b

def test_foo():
  assert _cmp(1, 2)
```

but then pytest will just say "False is not True", while you can have pytest smartly explain "a = 1 is not larger than b = 2" if you do this:

```py
def _assert_cmp(a, b):
  assert a > b

def test_foo():
  _assert_cmp(1, 2)
```

I'm worried that people new to python might have the intuition to write the second style, but the mypy message requiring to return something would leave them with to options:

1. unknowingly downgrade to the former style
2. write a weird `assert a > b; return None`

and considering the strangeness of the second option I think most would pick option one here.

semgrep/tests/e2e/snapshots/test_baseline/test_some_intersection/baseline_error.txt
```diff
@@ -1,2 +1,6 @@
 Running baseline scan with base set to: baseline-commit
+
+skipped: all .gitignored files, 1 files larger than 1.0 MB
```
underyx, line 3: @brendongo any idea why there's a 1+ MB large file in this test?

mypy-tests.ini
brendongo: yeah ok makes sense

semgrep/tests/e2e/snapshots/test_baseline/test_some_intersection/baseline_error.txt
brendongo: hmmm there def shouldnt be

brendongo: Ah I think this is some issue with TargetManager caching things. Bug with --baseline-commit but since that is experimental non blocking for this PR. Will follow up with issue + fix

brendongo: Tracked https://linear.app/r2c/issue/PA-844/semgrep-baseline-commit-file-size-filtering-bug

PR #4619, opened 2022-02-04
feat(core_runner): parse output stream from core to update progress

brendongo:
With https://github.com/returntocorp/semgrep/pull/4560 we send all rules to semgrep-core at once. Before that change
we updated progress on python side with each successive rule sent to core.

Changes in this commit have us read semgrep-core output as it comes in instead of blocking on semgrep-core
returning before starting to process. We have semgrep-core output to the stream to let semgrep know that progress
is being made and semgrep-cli updates the progress bar as necessary.

This PR also lays the groundwork to stream findings from semgrep-core

PR checklist:
- [x] Changelog is up-to-date


semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -493,6 +498,7 @@ let semgrep_with_rules config (rules, rules_parse_time) =
    *)
   let skipped = new_skipped @ res.skipped in
   let errors = new_errors @ res.errors in
+  (* Print done so Python knows to expect it *)
```
brendongo:
```suggestion
```

semgrep/semgrep/core_runner.py
```diff
@@ -99,6 +101,140 @@ def uniq_error_id(
...
+        Asynchronously process stdout of semgrep-core
+
+        Updates progress bar one increment for every "." it sees from semgrep-core
+        stdout
+
+        When it sees non-"." output it saves it to self._stdout
+        """
+        # appease mypy. stream is only None if call to create_subproccess_exec
+        # sets stdout/stderr stream to None
+        assert stream
```
underyx, line 151:
```suggestion
        # appease mypy. stream is only None if call to create_subprocess_exec
        # sets stdout/stderr stream to None
        if stream is not None:
            raise RuntimeError("subprocess was created without stream")
```

semgrep/semgrep/core_runner.py
```diff
@@ -99,6 +101,140 @@ def uniq_error_id(
...
+        while True:
+            # blocking read if buffer doesnt contain any lines or EOF
+            line_bytes = await stream.readline()
+
+            # readline returns None on EOF
+            if not line_bytes:
```
underyx, line 157:
```suggestion
            if line_bytes is None:
```
I assume in some situation it could return `b""` and we don't wanna break then? or if we update this code to do some preprocessing on the line, like `line_bytes.strip()` or something

semgrep/semgrep/core_runner.py
```diff
@@ -99,6 +101,140 @@ def uniq_error_id(
...
+            else:
+                self._stdout += line
+
+    async def _core_stderr_processor(
+        self, stream: Optional[asyncio.StreamReader]
+    ) -> None:
+        """
+        Asynchronously process stderr of semgrep-core
+
+        Basically works sychronnously and combines output to
```
underyx:
```suggestion
        Basically works synchronously and combines output to
```

semgrep/semgrep/core_runner.py
```diff
@@ -99,6 +101,140 @@ def uniq_error_id(
...
+        stderr to self._stderr
+        """
+        assert stream
```
underyx:
```suggestion
        if stream is not None:
            raise RuntimeError("subprocess was created without stream")
```

semgrep/semgrep/core_runner.py
```diff
@@ -99,6 +101,140 @@ def uniq_error_id(
...
+        while True:
+            # blocking read if buffer doesnt contain any lines or EOF
+            line_bytes = await stream.readline()
+
+            # readline returns None on EOF
+            if not line_bytes:
```
underyx, line 184:
```suggestion
            if line_bytes is None:
```
same as above

semgrep/semgrep/core_runner.py
```diff
@@ -99,6 +101,140 @@ def uniq_error_id(
...
+
+            line = line_bytes.decode("utf-8")
+            self._stderr += line
+
+    async def _stream_subprocess(self) -> int:
+        process = await asyncio.create_subprocess_exec(
+            *self._cmd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            limit=1024 * 1024 * 1024,
```
underyx: [The default is 64KiB](https://docs.python.org/3/library/asyncio-stream.html#:~:text=The%20returned%20reader%20and%20writer,is%20set%20to%2064%20KiB.).

Are we should 1 GiB is safe here? it's way higher than I'd expect, and based on the naming, I don't think this is the limit for maximum transferred data anyway

semgrep/semgrep/core_runner.py
brendongo: I think it has to return None to mark EOF otherwise it was intentionally an empty line

brendongo: Ah yes you're right

underyx (approved):
only got nitpicks, everything seems right otherwise!

semgrep/semgrep/core_runner.py
brendongo: changed to if stream is None:

semgrep/semgrep/core_runner.py
brendongo: the default was way too small for some cases of semgrep-core output since I think asyncio here blocks until a newline but the json output could be pretty large

semgrep/semgrep/core_runner.py
brendongo: Double checked and docs say it returns empty buffer on EOF https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readline

PR #4639, opened 2022-02-08
use non-foreground color for higlighting

### LINKED ISSUE #4637: Semgrep missing highlight in text output mode
**Describe the bug**
Semgrep doesn't highlight the matched range when printing textual output on the CLI

**To Reproduce**
Run semgrep on the CLI with some findings

**Expected behavior**

Semgrep used to highlight the matched range with a different color/bold, as described in https://github.com/returntocorp/semgrep/issues/4400
image

**Environment**
v 0.81.0

**What will fixing this bug enable for you?**
Causes confusion when identical lines are printed that have different matched ranges but no visual differentiators

ievans:
Fixes https://github.com/returntocorp/semgrep/issues/4637

v.80:
Screen Shot 2022-02-08 at 10 28 29 AM

v.81:
Screen Shot 2022-02-08 at 10 28 23 AM

this PR:
Screen Shot 2022-02-08 at 10 31 18 AM

@underyx @adamboros Is there a brighter code we can use / way to get more contrast? It is still very subtle to me.

PR checklist:
- [ ] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,12 +20,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Required minimum version of python to run semgrep now 3.7 instead of EOL 3.6
 - Bloom filter optimization now considers `import` module file names, thus
   speeding up matching of patterns like `import { $X } from 'foo'`
+- Indentation is now removed from matches to conserve horizontal space
 
 ### Fixed
 
 - Typescript: Patterns `E as T` will be matched correctly. E.g. previously
   a pattern like `v as $T` would match `v` but not `v as any`, now it
   correctly matches `v as any` but not `v`. (#4515)
+- Highlighting has been restored for matching code fragments within a finding
 
 ## [0.81.0](https://github.com/returntocorp/semgrep/releases/tag/v0.81.0) - 02-02-2022
 
--- a/semgrep/semgrep/formatter/text.py
+++ b/semgrep/semgrep/formatter/text.py
@@ -1,4 +1,5 @@
 import itertools
+import textwrap
 from itertools import groupby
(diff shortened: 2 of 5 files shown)


underyx:
@ievans we removed the bright color on the main branch, because it just makes text invisible on light themes, such as here:

image

underyx (approved):

brendongo (approved):

ievans:
the root cause was: "the problem with colors.bright is that for some themes it matches the background color"

we debated a few proposals here:

1. uncompromising static orange or adaptive red a la ripgrep
2. semgrep classic bright lettering (from v.80)

the first approaches have a lot in common with other tools as seen in this Slack screenshot:
Screen Shot 2022-02-08 at 10 16 24 PM

but ultimately @adamboros  decided against the first because:
* often semgrep is going to want to highlight “the entire snippet” whereas this is uncommon for most of the examples we provided above
* the absolute snippet length is short generally so less contrast is needed
* the bright color has a  "warning feeling to it"

underyx:
CleanShot 2022-02-09 at 09 59 05@2x

We also had an eng-wide poll with low participation. The poll showed gifs of each option:

CleanShot 2022-02-08 at 10 58 07

CleanShot 2022-02-08 at 11 22 04

CleanShot 2022-02-08 at 11 20 22

We learned in the process that the third option didn't look right on my gif
- first because bold styles were disabled in my iTerm config
- then because the terminal font I use (Cozette) does not have a bold style available

underyx:
However, in the process @nbrahms pointed out bento used to remove pointless indentation so we snuck that in here to save space https://github.com/returntocorp/semgrep/pull/4639/commits/98fb846ac627b1621ca7797b08f822dc9f95174a

PR #4645, opened 2022-02-09
[Codegraph] resolve method calls with self

aryx:
Test plan:
make in tests/python/graph_code
use codegraph and see the edge between bar and foo in self.py

PR checklist:

- [x] Changelog is up-to-date


aryx:
same error that previous PR with deprecated Cmdliner function. I'll force merge, it's not related to this PR.

PR #4648, opened 2022-02-09
core: pcre: Prevent Pcre engine from 'hanging" Semgrep

IagoAbal:
The PCRE engine does not allow setting timeouts and the existing PCRE
settings set very generous limits. Given the right regex and file, the
PCRE matching can take a long time to run, and `Common.set_timeout`
won't work as PCRE is calling a C library! As a result, Semgrep appeared
to hang.

Here we lower the match-attempts and recurson limits for the PCRE engine,
hoping to avoid such "hangs". For now, when the limits are reached, we
return as if the regex did not match, but we should probably report these
situations as timeout.

test plan:
$ semgrep-core -lang regex -rules perf/input/semgrep_targets.yaml perf/input/semgrep_targets.txt
 #^ now returns almost instantly whereas previously it would take a long time
$ cd perf && ./run-perf-suite # test included

PR checklist:

- [x] Changelog is up-to-date


ievans:
I wonder how https://github.com/BurntSushi/ripgrep deals with this problem

mjambon (approved):

semgrep-core/src/utils/Unit_pcre_settings.ml
```diff
@@ -4,15 +4,15 @@
 
 let test_match_limit_ok () =
   let rex = SPcre.regexp "(a+)+$" in
-  match SPcre.pmatch ~rex "aaaaaaaaaaaaaaaaaaa!" with
+  match SPcre.pmatch ~rex "aaaaaaaaaaaaaaaaa!" with
```
aryx, line 7: you needed to change the test to pass the limit? the longer aaa would exhaust  the 1M limit we set?

aryx (approved):
Need to fix the merge conflict and also merge develop to get CI to pass, but otherwise looks good.
Thanks a lot for fixing that and spending time on debugging the issue!

semgrep-core/src/utils/Unit_pcre_settings.ml
IagoAbal: Yes indeed. I'm hoping this only affects synthetic cases like this or to extreme cases like with `semgrep_targets.txt`.

IagoAbal:
> I wonder how https://github.com/BurntSushi/ripgrep deals with this problem

Good point, this seems to not be a problem for ripgrep and grep. 🤔

IagoAbal:
I'll merge this because it's better than the current situation. But I think @ievans made a good observation, perhaps there is something wrong with ocaml-pcre or the way we use it, we should investigate this!

PR #4649, opened 2022-02-09
fix(baseline): files not existing in baseline

brendongo:
Closes  PA-844

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -405,7 +405,7 @@ class CoreRunner:
             ) from ex
         return list(targets)
 
-    def _create_core_run_planner(
+    def _plan_core_run(
         self, rules: List[Rule], target_manager: TargetManager, all_targets: Set[Path]
     ) -> List[Dict[str, Any]]:
         """
@@ -475,10 +475,8 @@ class CoreRunner:
             target_file_name, "w+"
         ) as target_file:
 
-            targets_with_rules = self._create_core_run_planner(
-                rules, target_manager, all_targets
-            )
-            target_file.write(json.dumps(targets_with_rules))
+            plan = self._plan_core_run(rules, target_manager, all_targets)
+            target_file.write(json.dumps(plan))
             target_file.flush()
 
             yaml = YAML()
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -513,7 +513,7 @@ class CoreRunner:
                 print(" ".join(cmd))
                 sys.exit(0)
 
-            runner = StreamingSemgrepCore(cmd, len(targets_with_rules))
+            runner = StreamingSemgrepCore(cmd, len(plan))
             returncode = runner.execute()
 
             # Process output
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -390,9 +390,8 @@ class CoreRunner:
             }
             profiling_data.set_file_times(t.target, rule_timings, t.run_time)
 
-    @staticmethod
     def get_files_for_language(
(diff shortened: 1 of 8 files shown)


brendongo:
Closes https://linear.app/r2c/issue/PA-844/semgrep-baseline-commit-file-size-filtering-bug

semgrep/semgrep/core_runner.py
```diff
@@ -406,7 +405,7 @@ def get_files_for_language(
             ) from ex
         return list(targets)
 
-    def _get_targets(
+    def _create_core_run_planner(
```
underyx:
```suggestion
    def _plan_core_run(
```

the "create planner" name implies that this function returns a callable that can plan the run, but the return value is actually the plan itself

semgrep/semgrep/core_runner.py
```diff
@@ -390,9 +390,8 @@ def _add_match_times(
             }
             profiling_data.set_file_times(t.target, rule_timings, t.run_time)
 
-    @staticmethod
     def get_files_for_language(
-        language: Language, rule: Rule, target_manager: TargetManager
+        self, language: Language, rule: Rule, target_manager: TargetManager
```
underyx, line 394: I assume this can be changed back to what it was?

semgrep/semgrep/core_runner.py
```diff
@@ -476,7 +475,9 @@ def _run_rules_direct_to_semgrep_core(
             target_file_name, "w+"
         ) as target_file:
 
-            targets_with_rules = self._get_targets(rules, target_manager, all_targets)
+            targets_with_rules = self._create_core_run_planner(
+                rules, target_manager, all_targets
+            )
             target_file.write(json.dumps(targets_with_rules))
```
underyx:
```suggestion
            plan = self._plan_core_run(rules, target_manager, all_targets)
            target_file.write(json.dumps(plan))
```
this is in case you accept the above suggestion

semgrep/semgrep/semgrep_main.py
```diff
@@ -385,14 +385,31 @@ def main(
...
+                        max_target_bytes=max_target_bytes,
+                        targets=target,
+                        respect_git_ignore=respect_git_ignore,
+                        skip_unknown_extensions=skip_unknown_extensions,
+                        file_ignore=get_file_ignore(),
+                    )
+                except FilesNotFoundError as e:
+                    # This means a file existed in head but not
+                    # in baseline context which is fine
+                    pass
```
underyx: Maybe add a debug log?

semgrep/semgrep/semgrep_main.py
```diff
@@ -385,14 +385,31 @@ def main(
         logger.info(f"Running baseline scan with base set to: {baseline_commit}")
         try:
             with baseline_handler.baseline_context():
+                # Need to reinstantiate target_manager since
+                # filesystem has changed
+                try:
+                    baseline_target_manager = TargetManager(
```
underyx, line 391: Maybe this should be a method on the TargetManager? Something like `target_manager.reset()`?

underyx: although reset() implies that it's an in-place operation so that's not ideal

underyx (approved):
all comments are optional!

semgrep/semgrep/semgrep_main.py
brendongo:
```suggestion
                    logger.debug(f"File not found in baseline: {e}")
```

semgrep/semgrep/core_runner.py
brendongo: Shouldn't be exposed out of this class anyway

PR #4650, opened 2022-02-09
Remove Match from Generic AST

mmcqd:
Everything expressible with `Match` is expressible with `Switch`. Specifically:
```
Match (tok, e, [ (pat, expr) ; ... ] 
=
Switch (tok, Some (Cond e), [ CaseAndBody ([ Case (fake "case", pat) ], stmtexpr expr) ; ... ])
```
The `fake "case"` can be replaced with a real token when possible.
This simplifies the generic AST and will make supporting patterns like `$X match { ... }` in Scala or `match $X with ...` in OCaml very simple.

test plan: make test (this does not add new functionality, so we just want all old tests to pass)
PR checklist:
- [x] Changelog is up-to-date


mmcqd:
Not sure what these errors are, they aren't coming from code that I touched.

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -897,10 +897,6 @@ and stmt_kind =
       tok (* 'switch', also 'select' in Go, or 'case' in Bash *)
       * condition option
       * case_and_body list (* TODO brace *)
-  (* todo: merge Match with Switch?
-   * In Scala and C# the match is infix (after the expr)
```
aryx, line 901: maybe add a comment on Switch saying this covers also the match of OCaml/Scala/C# and also the note about that it's infix in Scala/C#
You can even add the nice comment in your PR about the translation of Match (tok, e, (pat, exp)) -> Switch ...

semgrep-core/src/parsing/pfff/ml_to_generic.ml
```diff
@@ -362,11 +365,15 @@ and expr e =
...
   | Function (t, xs) ->
-      let xs = list match_case xs in
+      let xs =
+        list (fun a -> a |> match_case |> match_case_to_switch_case) xs
+      in
       let id = ("!_implicit_param!", t) in
       let params = [ G.Param (G.param_of_id id) ] in
       let body_stmt =
-        G.Match (t, G.N (G.Id (id, G.empty_id_info ())) |> G.e, xs) |> G.s
+        G.Switch
```
aryx, line 370: you can use H2.name_of_id (where H2 is AST_generic_helpers).

semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml
```diff
@@ -1003,7 +1003,7 @@ and switch_expression_arm (env : env)
   in
   let _v3 = token env v3 (* "=>" *) in
   let v4 = expression env v4 in
-  (v2, v4)
+  G.CasesAndBody ([ G.Case (fake "case", v2) ], G.exprstmt v4)
```
aryx: sometimes when those expressions appear in multiple xxx_to_generic.ml, I add them to AST_generic_helpers.ml or even AST_generic.ml (e.g., AST_generic.interpolated), so that way if we decide at some point to change again, we just need to modify one function.

semgrep-core/src/parsing/tree_sitter/Parse_rust_tree_sitter.ml
```diff
@@ -1444,8 +1444,12 @@ and map_expression_ending_with_block (env : env)
   | `Match_exp (v1, v2, v3) ->
       let t = token env v1 (* "match" *) in
       let expr = map_expression env v2 in
-      let actions = map_match_block env v3 in
-      let st = G.Match (t, expr, actions) |> G.s in
+      let actions =
+        map_match_block env v3
+        |> List.map (fun (pat, e) ->
+               G.CasesAndBody ([ G.Case (G.fake "case", pat) ], G.exprstmt e))
```
aryx: like here, could be a G.case_of_pat_and_expr pat e

aryx (approved):
Need to merge develop first to fix the CI I think and then should be good to go.

PR #4652, opened 2022-02-10
Update usage of attrs to next-gen API

underyx:
Depends on #4653

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/dependencyparser/models.py
+++ b/semgrep/dependencyparser/models.py
@@ -5,7 +5,7 @@ from typing import Dict
 from typing import List
 from typing import Optional
 
-from attrs import define
+from attrs import frozen
 
 KNOWN_HASH_ALGORITHMS: Dict[str, Callable] = {
     "sha256": hashlib.sha256,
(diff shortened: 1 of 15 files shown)


brendongo (approved):

semgrep/dependencyparser/models.py
```diff
@@ -19,7 +19,7 @@ class PackageManagers(Enum):
     PYPI = "pypy"
 
 
-@attr.s(auto_attribs=True, frozen=True, eq=True, order=True)
+@define(auto_attribs=True, frozen=True, eq=True, order=True)
```
aryx: why not using simply @dataclass? Isn't it equivalent to this long @define?

underyx: here: https://www.attrs.org/en/stable/why.html#data-classes

underyx: primarily

> There is a long list of features that were sacrificed for the sake of simplicity and while the most obvious ones are validators, converters, [equality customization](https://www.attrs.org/en/stable/comparison.html#custom-comparison), or [extensibility](https://www.attrs.org/en/stable/extending.html) in general

PR #4659, opened 2022-02-11
Add paths to json output

underyx:
Closes PA-838

PR checklist:

- [ ] Changelog is up-to-date


semgrep/semgrep/output.py
```diff
@@ -362,7 +362,16 @@ def _post_output(self, output_url: str, output: str) -> None:
...
     ) -> str:
-        extra: Dict[str, Any] = {}
+        # Extra, extra! This just in! 🗞️
+        # The extra dict is for blatantly skipping type checking and function signatures.
+        # - The text formatter uses it to store settings
+        # - But the JSON formatter uses it to store additional data to directly output
+        extra: Dict[str, Any] = {
+            "paths": {
+                "scanned": [str(path) for path in sorted(self.all_targets)],
+                "skipped": "<add --verbose for a list of skipped paths>",
```
brendongo: skipped value doesnt have consistent type. can even make this optional and not include this field when `--verbose` not set?

brendongo (approved):

semgrep/semgrep/output.py
underyx: Yeah, this is again a question of… should we make this JSON communicate better with humans, or with machines? My thoughts were:

- **why I like this line**: As a user of a CLI tool, I'd never guess that `--verbose` affects more than the output to stderr. So if I run `semgrep --json` to see what data is available, if the `"skipped"` key is not there, I'd just assume the data is not available. Put simply, changing the JSON output's schema based on flags is surprising. And if we work around this by saying `"skipped": None`, that would still look misleading even if technically correct (being a null value instead of an empty list).
- **why I'm not that worried about inconsistency**: The type is consistent based on what flags you use. And any tooling will almost certainly be consistent about whether it passes `--verbose` or not. For instance, semgrep-agent always passes it. Furthermore, if a tool doesn't pass `--verbose`, this is a dummy field in the JSON output, so nobody would access and depend on its type.
- **why I'm still a bit worried about inconsistency**: If we ever end up committing to a jsonschema for our JSON output, this will be a weird exception there, and will make life more difficult for any dev who tries to work off of the jsonschema.

Now I'm thinking a nice compromise would be making the output this:

```json
"paths": {
  "_comment": "run with --verbose to also get the skipped paths in JSON",
  "scanned": ["foo.py", "bar.py"]
}
```

I'm curious what @DrewDennison, @ievans, and @adamboros would prefer.

brendongo: I don't feel strongly enough to block this change cause of this. IMO we can work with this and cross the bridge for consistency once we commit to a jsonschema for output

ievans: @underyx I like your compromise solution myself

PR #4660, opened 2022-02-11
[HCL] Fix CST-to-Generic translation of heredocs

### LINKED ISSUE #4582: [Terraform] Can't match YAML documents inside Terraform strings with metavariable-pattern
**Describe the bug**
I have a rule in which I'm trying to match YAML inside a Terraform string. Examples are in the next section.
I have tried many combinations of patterns, but it seems like it just isn't getting picked up.
I'm still able to match YAML documents in other languages, though.

**To Reproduce**
Case I want to match: https://semgrep.dev/s/GyGq
Full rule I'm trying to write with JSON examples that work: https://semgrep.dev/s/Ry6q
YAML working with metavariable-pattern in Python: https://semgrep.dev/s/AyE2

IagoAbal:
Heredocs are parsed as a list of tokens without any newline or
whitespace. But heredocs may represent code that we want to analyze with
metavariable-regex or metavariable-pattern, so we must parse these
strings correctly. Ideally, we should preserve these newlines and
whitespace already during parsing... but for now we just "restore" them
using token location info.

Closes #4582

test plan:
make test # test included

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_hcl_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_hcl_tree_sitter.ml
@@ -85,16 +85,38 @@ let map_numeric_lit (env : env) (x : CST.numeric_lit) : literal =
 
 let map_template_literal (env : env) (xs : CST.template_literal) :
     string wrap option =
-  (* each character is a separate template_literal_chunk, so we need
-   * to merge them in a single string.
-   *)
-  let xs = List.map (str env (* template_literal_chunk *)) xs in
   match xs with
   | [] -> None
-  | (s1, t1) :: ys ->
-      let str = List.map fst ys |> String.concat "" in
-      let toks = List.map snd ys in
-      Some (s1 ^ str, PI.combine_infos t1 toks)
+  | ((x_pos, x_str) as x) :: xs ->
+      (* each character is a separate template_literal_chunk, so we need
+       * to merge them in a single string. *)
+      let module Loc = Tree_sitter_run.Loc in
+      let base_col = x_pos.Loc.start.column in
+      let rec concat_chunks acc_str prev_row prev_col xs =
+        match xs with
+        | [] -> acc_str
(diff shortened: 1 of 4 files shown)


semgrep-core/src/parsing/tree_sitter/Parse_hcl_tree_sitter.ml
```diff
@@ -85,16 +85,38 @@ let map_numeric_lit (env : env) (x : CST.numeric_lit) : literal =
...
-   *)
-  let xs = List.map (str env (* template_literal_chunk *)) xs in
   match xs with
   | [] -> None
-  | (s1, t1) :: ys ->
-      let str = List.map fst ys |> String.concat "" in
-      let toks = List.map snd ys in
-      Some (s1 ^ str, PI.combine_infos t1 toks)
+  | ((x_pos, x_str) as x) :: xs ->
+      (* each character is a separate template_literal_chunk, so we need
```
emjin: This is an old comment, but just noting that I'm confused by what "each character is a separate template_literal_chunk" means

semgrep-core/src/parsing/tree_sitter/Parse_hcl_tree_sitter.ml
```diff
@@ -85,16 +85,38 @@ let map_numeric_lit (env : env) (x : CST.numeric_lit) : literal =
...
+      let rec concat_chunks acc_str prev_row prev_col xs =
+        match xs with
+        | [] -> acc_str
+        | x :: xs ->
+            let x_loc, x_str = x in
+            let last_col =
+              if x_loc.Loc.start.row > prev_row then base_col else prev_col
+            in
+            let acc_str =
+              (* we need to respect newlines and spaces so that HEREDOCs are
```
emjin: This is great, might be better as a function comment so a reader can understand what map_template_string needs to do without looking into the code

emjin (approved):
Thanks!

PR #4662, opened 2022-02-11
Report parse errors even when run with --strict

### LINKED ISSUE #1922: Parse errors not shown when run with --strict
When running with `--strict`, the suffix about parse errors removed (although the exit code flag is set).

```
$ semgrep --config "p/python" .
...
ran 27 rules on 66 files: 8 findings
2 files could not be analyzed; run with --verbose for details or run with --strict to exit non-zero if any file cannot be analyzed

$ semgrep --strict --config "p/python" .
...
ran 27 rules on 66 files: 8 findings
```

So with --strict the only way to know there was a parse error is with --verbose.
Otherwise, the only hint is the exit code (which makes CI failing) that something is going on:

image

nmote:
Fixes #1922

PR checklist:

- [x] Changelog is up-to-date


semgrep/tests/e2e/snapshots/test_nosemgrep/test_nosem_rule__invalid_id/error.txt
```diff
@@ -4,3 +4,4 @@ skipped: all .gitignored files
 for a detailed list of skipped files, run semgrep with the --verbose flag
 
 ran 1 rules on 1 files: 1 findings
+found problems analyzing 0 file; run with --verbose for details
```
nmote, line 7: This "0 file" thing is a little suspicious and I haven't looked into it. It seems likely that Semgrep would already display this message without the `--strict` flag, so in that sense I suspect it is unrelated to this change. I don't know enough about Semgrep to assert that as fact, though. I'd be happy to dig in a little if reviewers think this is important.

nmote:
I assume that this change is minor enough that no documentation needs to be updated, so I checked that box.

semgrep/tests/e2e/snapshots/test_nosemgrep/test_nosem_rule__invalid_id/error.txt
nbrahms: I think in this case the error is from the configuration, rather than any files, hence the `0 file` report.

nbrahms: Maybe we want to remove the "analyzing 0 file" part of this output (although that seems out of the scope of this work per se).

semgrep/semgrep/output.py
```diff
@@ -248,13 +248,16 @@ def _final_raise(self, ex: Optional[Exception], error_stats: Optional[str]) -> N
...
             else:
-                if self.settings.strict:
-                    raise ex
+                message = f"{error_stats}; run with --verbose for details"
+                if not self.settings.strict:
+                    message += " or run with --strict to exit non-zero if any file cannot be analyzed cleanly"
                 logger.info(
                     terminal_wrap(
-                        f"{error_stats}; run with --verbose for details or run with --strict to exit non-zero if any file cannot be analyzed cleanly"
+                        message
```
nbrahms: I think maybe pre-commit got skipped here?

nbrahms (approved):
LGTM. Just needs a pass through `black`.

underyx:
> LGTM. Just needs a pass through `black`.

For your reference @nmote, following these instructions will format the code correctly to pass our automated checks: https://semgrep.dev/docs/contributing/contributing-code/#development-workflow

semgrep/tests/e2e/snapshots/test_nosemgrep/test_nosem_rule__invalid_id/error.txt
brendongo: I agree out of scope for this PR but we should probably fix before releasing this change so we don't have this confusing message

semgrep/semgrep/output.py
nmote: Yeah, sorry. I missed that section of the contributing instructions. @emjin pointed them out in my other PR but I haven't yet circled back to this PR to double-check it. I'll get that sorted out.

nmote:
This isn't stale, just waiting to get merged. I made the necessary changes some time ago but I guess I should have commented too.

PR #4664, opened 2022-02-12
--validate checks metavariables used by metavariable-x

### LINKED ISSUE #3305: Undetected typos / errors in metavariable names
When using metavariable-regex with an incorrect metavariable, no error is triggered:

```yml
    patterns:
      - pattern: |
          var $UTIL = require("something_or_other");
          ...
          $UTIL.$FUNC(...)
      - metavariable-regex:
          metavariable: $METHOD
          regex: (a|b|c)
```
Note the `metavariable` doesn't match anything defined in the `pattern`. I did not.

Now I do not know how "missing" metavariables are modelled internally by semgrep, or how `metavariable-regex` is modelled, so it's possible that there's a way to check that the metavariable is *not* matched and thus the feature makes sense, but it seems a bit odd.

emjin:
Closes https://github.com/returntocorp/semgrep/issues/3305

When metavariable-pattern or metavariable-regex is used on a metavariable that is not valid (not in any pattern-and clause or all pattern-or clauses), emit an error.
    
Also, remove the redundant "duplicate pattern" and "unsatisfiable" checks that are already covered by semgrep-rule-lints. We previously could not because otherwise there would have been no test that the OCaml checks are being used.
    
Test plan: `semgrep --validate --config tests/OTHER/metachecks/unused-metavariable.yaml`
    
TODO:
  - [ ] fix --validate message for semgrep-cli
  - [ ] add --validate tests to semgrep-cli
  - [ ] improve message for the unused metavariable check to be more helpful

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/tests/OTHER/errors/unsatisfiable_formula.yaml
```diff
@@ -2,7 +2,6 @@ rules:
 - id: test-template
   patterns:
   - pattern: eval(...)
-  #ERROR: unsatsfiable
```
emjin, line 5: @aryx I was really confused about why the other files in this folder that had error annotations weren't causing tests to fail? I understand why this and duplicate_pattern were

emjin:
Failing tests because some metavariable cases aren't considered. It might be better to extract metavariables by looking at the original pattern strings and looking for tokens starting with `$` since I'm afraid there will be too many cases

semgrep-core/src/metachecking/Check_rule.ml
```diff
@@ -80,66 +78,80 @@ let error env t s =
...
-              |> List.iter (fun y ->
-                     if equal_formula (Not (t, x)) y then
-                       let tx, ty = (R.tok_of_formula x, R.tok_of_formula y) in
-                       let kind = R.kind_of_formula x in
-                       error env ty
-                         (spf "Unsatisfiable formula with %s at line %d" kind
-                            (PI.line_of_info tx)));
-              aux xs
+    | P ({ pat = _pat; pstr = pstr, _; pid = _pid }, _) ->
+        let words = Str.split (Str.regexp "[^a-zA-Z0-9_$]") pstr in
```
emjin, line 91: This is pretty jank. Originally I traversed the tree and extracted metavariables, but I found that this wouldn't work with generic mode, and adding that (and other potential parsing modes we could add) seemed annoying. On the flip side, if we ever change our metavariable syntax, this is fragile. Not sure what the best approach is

mjambon: I'm not completely following but based on your comment, it would be reasonable to expect any matcher that returns bound metavariables to be able to tell us in advance which metavariables it would bind upon success. Generic mode uses metavariables similarly to semgrep. pattern-regex also does (or should) have the ability to bind metavariables to matched substrings ("groups" in regex jargon). In the future, we could have also other analyzers (the new `metavariable-analysis`) that could return a collection of bound metavariables. This collection should be known statically without resorting to special hacks.

A matcher (semgrep, generic, regex, ...) would have to follow a module signature that might look something like:
```ocaml
type target = ... (* file or a string *)

type search_results = ... (* list of matches with bound metavariables or errors *)

module type Matcher = sig
  type param
  type parsed_target
  val metavariables : param -> string list
  val parse : target -> parsed_target
  val search : param -> parsed_target -> search_results
end
```

mjambon (approved):

adamkvitek:
Hi there @emjin,

Just asking to understand more about the possible changelog entry here: "_When metavariable-pattern or metavariable-regex is used on a metavariable that is not valid (not in any pattern-and clause or all pattern-or clauses)_" - What is pattern-and clause or all pattern-or clauses?

"_remove the redundant "duplicate pattern" and "unsatisfiable" checks that are already covered by semgrep-rule-lints._" What were these checks? They are not used anymore, right?

emjin:
> Hi there @emjin,
> 
> Just asking to understand more about the possible changelog entry here: "_When metavariable-pattern or metavariable-regex is used on a metavariable that is not valid (not in any pattern-and clause or all pattern-or clauses)_" - What is pattern-and clause or all pattern-or clauses?

Ah, I should have said `patterns` (which performs an and of patterns) and `pattern-either` (which performs an or of patterns) instead

> "_remove the redundant "duplicate pattern" and "unsatisfiable" checks that are already covered by semgrep-rule-lints._" What were these checks? They are not used anymore, right?

No functionality has changed here

aryx (commented):
Woohoo!

Issue #3938, opened 2021-09-25
Equivalence between go string literals and raw string literals


emjin:
In go, backtick strings are "raw strings" (unescaped) (see https://golang.org/ref/spec#String_literals). In particular, they are always constant literals, and users may want to match them using "...".

Example:

Rule:

```
rules:
- id: "gosec.G201"
  patterns:
  - pattern-inside: |
      $Q := fmt.Sprintf("$QUERY", ...)
      ...
  - metavariable-regex:
      metavariable: "$QUERY"
      regex: "select"
  - pattern-either:
    - pattern: "$DB.Query($Q,...)"
  message: SQL query construction using format string
  severity: "WARNING"
  languages: ["go"]
```

With string literal:

```
package main

import (
    "database/sql"
    "fmt"
)

func test(conn *sql.DB, someVar string) {
    query := fmt.Sprintf("select region from %s", someVar)
    conn.Query(query)
}
```

With raw string literal:

```
package main

import (
    "database/sql"
    "fmt"
)

func test(conn *sql.DB, someVar string) {
    query := fmt.Sprintf(`select region from %s`, someVar)
    conn.Query(query)
}
```

With the equivalence, both of these would match that rule

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-409/equivalence-between-go-string-literals-and-raw-string-literals.

IagoAbal:
Might be similar to #3881.

IagoAbal:
@emjin What about https://semgrep.dev/s/jwKY ?

Internally we represent:
```
`select region from %s`
```
as the string literal:
```
"`select region from %s`"
```
And the `regex` in `metavariable-regex` only matches starting from the beginning of the string.

IagoAbal:
Alternatively, we could represent Go raw strings using `EncodedString "r"` as we do for Python, then it would constant propagation that makes them equivalent to string literals. Unless we want these "encoded strings" to be equivalent to strings regardless of constant propagation.

emjin:
Hm, I think making them equivalent by constant propagation would be fine? Check the linked issue #3930 for context

emjin:
Actually, I just saw your Python raw string PR. I think this is exactly the same kind of construct!

nmote:
I'm taking a look at this.

IagoAbal:
Planning to open a PR @nmote ?

nmote:
Yes, once I have a little more time to prepare it

[...]

PR #4671, opened 2022-02-14
Treat Go raw string literals like ordinary string literals

nmote:
This addresses #3938, though this solution is a bit hacky so future work should be done to change the grammar so that this isn't necessary. I've posted more details in that issue.

PR checklist:

- [x] Changelog is up-to-date


nmote:
I have two questions for reviewers:
1. Assuming this is acceptable at all, is there some utility I should use for manipulating the underlying string in a token, rather than pulling it apart myself?
2. It looks to me like the interpreted string literal code includes the locations of delimiters in the resulting location for the AST node, so am I correct in assuming that it's okay not to manipulate the location for the raw string literal here so that it omits the locations of the delimiters? I hope so, because making manual changes to the location would be very error-prone.

semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
```diff
@@ -51,8 +51,18 @@ let rev = false
...
 
-(* TODO: need remove enclosing `` *)
-let raw_string_literal env tok = str env tok
+(* TODO: Update grammar so that the leading and trailing backticks are tokenized
+ * separately, the way interpreted string literals are:
+ * https://github.com/tree-sitter/tree-sitter-go/blob/0fa917a7022d1cd2e9b779a6a8fc5dc7fad69c75/grammar.js#L839-L843
+ * *)
+let raw_string_literal env tok =
+  let (loc, s) = tok in
+  (* Remove leading and trailing backticks. The grammar guarantees that raw
```
emjin, line 60: Probably better to code defensively and add the check, since this depends on behavior of an external library

emjin (approved):
Thank you for taking this issue! This looks good to me! Leaving the tokens is great; that means autofix won't get messed up. Tagging @aryx or @IagoAbal to take a look in case I've missed some nicer way to do this.

emjin:
Re the failing pre-commit check: https://semgrep.dev/docs/contributing/contributing-code/#development-workflow

semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
nmote: Fair enough, will do. I'll probably add a test that exercises an empty raw string and do a bounds-check with a `failwith`. That way any update that changes this behavior will get us a test failure. Sound good?

nmote:
Thanks for the link; I missed that section. I'll get the formatting sorted out.

nmote:
It looks like the pre-commit job needs approval from a maintainer. I did get it set up locally and it changed the formatting, so I think it should be green once it runs.

semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
```diff
@@ -51,8 +51,20 @@ let rev = false
...
+(* TODO: Update grammar so that the leading and trailing backticks are tokenized
+ * separately, the way interpreted string literals are:
+ * https://github.com/tree-sitter/tree-sitter-go/blob/0fa917a7022d1cd2e9b779a6a8fc5dc7fad69c75/grammar.js#L839-L843
+ * *)
+let raw_string_literal env tok =
+  let loc, s = tok in
+  (* Remove leading and trailing backticks. The grammar guarantees that raw
+   * string literals will always have leading and trailing backticks, so this
+   * String.sub call should be safe. Let's check just to be sure. *)
+  if String.length s < 2 || s.[0] <> '`' || s.[String.length s - 1] <> '`' then
```
aryx: you can open Common and then do
if not (s =~ "^`\\(.*\\)`$")
then failwith "..."
else
  let str = Common.matched1 s in
  ...

Regexps are easier than those String.length, sub functions.

aryx (approved):
looks good! Thanks for fixing that.

semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
```diff
@@ -51,8 +51,20 @@ let rev = false
...
+    failwith "Found unexpected raw string literal without delimiters";
+  let s = String.sub s 1 (String.length s - 2) in
+  let tok = (loc, s) in
+  str env tok
```
IagoAbal, line 67: We could simply add `EncodedString` here if we wanted to. The difference is that the way this is encoded, a pattern:
```
`foo`
```
will match a code expression `"foo"`, they are encoded as equivalent. For Python, Semgrep can match pattern `"foo"` against `r"foo"` via constant folding, but not the other way around.

I am not arguing in favor or against any approach, but I think we should be consistent.

nmote: That makes sense, and it sounds like something that should be done. I'd prefer to leave it for another PR, though, if that's okay with you.

nmote:
Thanks @aryx and @IagoAbal for the comments. I'll have time to make changes on Thursday.

semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
IagoAbal: I think the change is so small that it's better to do it in this PR, something like this should work:

```ocaml
Call(IdSpecial(EncodedString "r", tok), (tok,[str env tok],tok))
```

Otherwise soon enough somebody will complain about why this behaves "this way" in Python and "this other way" in Go.

nmote: Okay, I will take a stab at that on Thursday. I only have an hour left on the time allotted for this project, though, so if I encounter any issues while making or testing this change it could put me over a bit.

IagoAbal: Well, it may not be that small, because `string_literal` is not supposed to return an expression.

nmote: I've looked into this and if I understand correctly, this would require a change to the pfff AST for Go. Right now, when we translate from the tree-sitter CST to the pfff AST, all string literals are represented by the `String` variant of the `literal` type defined in `semgrep-core/src/pfff/lang_go/parsing/ast_go.ml`. We'd need another variant there to differentiate between kinds of string literals, so that we could then choose `EncodedString` rather than `String` in `go_to_generic.ml` when we encounter a raw string literal.

Unfortunately, I don't currently have the time to make such a change. I hope that you can agree that as-is, this PR is still a step forward and is worth merging, but I understand if you'd rather wait until someone is able to make your requested changes.

nmote:
I've made the changes requested by @emjin and @aryx, but unfortunately don't have the time to make the changes requested by @IagoAbal (details in the unresolved comment thread above). I still need a maintainer to approve some of the CI jobs, but I think they should all be green.

aryx:
Ok I'll merge this. Sorry for our late reply, it was hack week last week so we got super busy.

aryx (approved):

semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
IagoAbal: Yeah that sounds right! I think @aryx wants to merge anyways. No big deal for me. I just thought it's better to fix it consistently with current behavior for other languages, unless this is blocking some customer (which is not AFAIK). Good work anyways!

PR #4672, opened 2022-02-15
Entropy and/or ReDoS analysis

mjambon:
I created the plumbing to support analysis on a string captured by a metavariable. There are now two blank modules `Entropy` and `ReDoS` that need to implement a useful analysis.

Here's a sample rule for testing the `entropy` analysis:
```yaml
rules:
  - id: test-entropy
    patterns:
      - pattern: $A = $B
      - metavariable-analysis:
          analyzer: entropy
          metavariable: $B
    message: found high-entropy string
    languages:
      - python
    severity: ERROR
```

Test with:
```
$ semgrep -c entropy.yaml <(echo 'x = "ad;fkjasdfasdfasdf"')
Running 1 rules...

 /tmp/tmpe6zl8076/dev_fd_63 
     test-entropy
        found high-entropy string


          1┆ x = "ad;fkjasdfasdfasdf"

skipped: all .gitignored files
for a detailed list of skipped files, run semgrep with the --verbose flag

```
(right now, if you make the string short enough, there won't be a match)

PR checklist:

- [ ] Changelog is up-to-date


semgrep-core/src/core/Rule.ml
```diff
@@ -135,6 +135,8 @@ and metavar_cond =
    * the "regexpizer" optimizer (see Analyze_rule.ml).
    *)
   | CondRegexp of MV.mvar * regexp
+  | CondEntropy of MV.mvar
```
aryx: Better to have CondAnalysis then, and CondAnalysisKind

aryx (approved):

semgrep-core/src/core/Rule.ml
mjambon: ok :+1:

kurt-r2c (approved):
Entropy stuff works great 👍

mjambon:
Documentation and changelog are **not** up-to-date but I'll do this later. One thing at a time.

semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -305,31 +305,32 @@ let bindings_to_env (config : Config_semgrep.t) xs =
   in
   { mvars; constant_propagation }
 
+let text_of_binding mvar mval =
```
aryx, line 308: i usually suffix with _opt when something return an option.

semgrep-core/src/engine/Eval_generic.mli
```diff
@@ -35,5 +35,8 @@ val parse_json : Common.filename -> env * code
 (* for MetavarCond *)
 val bindings_to_env : Config_semgrep.t -> Metavariable.bindings -> env
 
+(* For entropy analysis and other string analyzers *)
```
aryx, line 38: I would maybe add that the mvar parameter is used just for debugging/logging purpose.

semgrep-core/src/engine/Eval_generic.ml
mjambon: I don't know, the type contains that info already, and there's no alternate implementation to disambiguate from. I read `text_of_binding` as "that's the function to use to convert a binding to text", whereas `text_of_binding_opt` hints that there's something special about it, which isn't the case.

PR #4695, opened 2022-02-17
suggest semgrep login in certain cases

ievans:
we'll suggest to the user that more rules are available under semgrep login, for runs where there are 0 findings, with 1+ rules and 1+ files scanned, and config pulled from the registry

PR checklist:

- [x] Changelog is up-to-date


TinMarkovic (approved):
LGTM

PR #4699, opened 2022-02-18
Support Java, Go, Ruby, and Rust in project-depends-on patterns

mmcqd:
This PR adds support for 4 new languages to `r2c-internal-project-depends-on` patterns.
* Ruby
    * Dependencies scanned from `Gemfile.lock`
* Rust
    * Dependencies scanned from `cargo.lock`  
* Go
    * Dependencies scanned from `go.sum`. Pseudo versions and `+incompatible` versions are ignored, meaning that `v3.0.1-0.20171022003610-9aa49832a739+incompatible` is treated as just `3.0.1`. If a dependency is listed both as and as not a `go module`, then both dependencies are included separately
* Java
    * Dependencies scanned from `pom.xml`. Maven does not actually create lockfiles, so instead we scan the dependency manifest, and include dependencies which are listed with an exact version. References to the `<properties>` field via `${property}` are supported.

This PR also adds `defusedxml` as a new semgrep dependency, to allow the safe parsing of XML files. This is necessary for supporting Java.

Test plan: make test (added e2e tests for each new language)

See PR to update docs: https://github.com/returntocorp/semgrep-docs/pull/465

PR checklist:

- [x] Changelog is up-to-date


.pre-commit-config.yaml
```diff
@@ -45,6 +45,7 @@ repos:
           - click~=8.0
           - packaging~=21.0
           - requests~=2.22
+          - defusedxml~=0.7.1
```
aryx, line 48: I remember there was some discussions on security vulnerabilities in this library. Or was it for another dependency?
Very meta :)

semgrep/dependencyparser/find_lockfiles.py
```diff
@@ -6,7 +6,15 @@
 from typing import Optional
 from typing import Set
 
-TARGET_LOCKFILE_FILENAMES = ["pipfile.lock", "yarn.lock", "package-lock.json"]
+TARGET_LOCKFILE_FILENAMES = [
+    "pipfile.lock",
```
aryx: suggestion: Maybe we could add in comment the language this covers? Just to get a sense of how many languages this feature supports.

semgrep/dependencyparser/models.py
```diff
@@ -7,16 +7,22 @@
 
 from attrs import frozen
 
-KNOWN_HASH_ALGORITHMS: Dict[str, Callable] = {
+KNOWN_HASH_ALGORITHMS: Dict[str, Optional[Callable]] = {
```
aryx: suggestion: Could you add a comment saying what this list is for? Why do we need to run those hashing functions?

semgrep/dependencyparser/find_lockfiles.py
aryx: Or maybe we should just associate the function to parse them in a dictionary so we don't have to do those if elif later.

semgrep/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awarelog4shell.yaml-dependency_awarelog4shell.java/results.json
```diff
@@ -0,0 +1,102 @@
+{
+  "errors": [],
+  "paths": {
+    "_comment": "<add --verbose for a list of skipped paths>",
```
aryx, line 4: Those elements are annoying in test files. Why do we have those _comment?

aryx (approved):
Fantastic!

.pre-commit-config.yaml
mmcqd: The discussion was about the builtin python XML library, which has several security vulnerabilities. `defusedxml` is the library that the python docs recommend for safe XML parsing.

semgrep/dependencyparser/models.py
mmcqd: We don't need to run them as far as I can tell. I was just keeping the code how @ievans set it up. I think we could probably just replace the functions with strings.

semgrep/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awarelog4shell.yaml-dependency_awarelog4shell.java/results.json
mmcqd: I don't know. When I made the PR I got that github-actions request to accept the snapshot changes, and I think it added those.

semgrep/dependencyparser/find_lockfiles.py
mmcqd: Yeah this would be nicer

PR #4700, opened 2022-02-18
Implement ReDoS vulnerability prediction

mjambon:
This implements `metavariable-analysis` with the `redos` analyzer. A regexp found in a program is considered vulnerable if it can be fed some input that will take matching a really long time to complete (O(exp(input length)).
* we have unit tests that run such regexps. It uses `Pcre` which has a protection against this and raises an exception. This test takes about 3 seconds on my old laptop.
* not all vulnerable regexps are detected.
* hopefully false positives will be uncommon in practice.

To do:
- [ ] end-to-end tests

Uses https://github.com/returntocorp/pfff/pull/505

PR checklist:

- [ ] Changelog is up-to-date


aryx (approved):

mjambon:
Merging this. I'll add the appropriate tests and documentation later.

PR #4703, opened 2022-02-18
fix(metrics): dont raise if both metrics value are the same

dee-see:
Hello team!

I was copy-pasting a few snippets that had the `--disable-metrics` switch and because I also set `SEMGREP_SEND_METRICS` to `off` the commands failed. It makes sense not to allow different values to be set, but I felt like it was harmless to set the same value in two different ways so I'm proposing this change.

Do I need a changelog modification here?

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Kotlin: store trailing lambdas in the AST (#4741)
 - Autofix: Semgrep no longer errors during `--dry-run`s where one fix changes the line numbers in a file that also has a second autofix.
 - Performance regression when running with --debug (#4761)
+- Allow metrics flag and metrics env var at the same time if both are set to the same value (#4703)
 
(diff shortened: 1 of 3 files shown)


dee-see:
Marked as draft because I did this change from the GitHub UI and didn't test if it actually works yet

dee-see:
Alright it seems to be working as expected.

brendongo (commented):
This looks good can you add this as a test case in test_metrics and update the docs?

dee-see:
@brendongo :wave: I have 2 questions for you

1. I can't run the tests locally, everything fails and for such a simple test I figured I'd just let the CI tell me if it's right but it seems like the tests aren't run on commit? Or am I just not looking in the right place?
2. Can you point me to the documentation that needs to be updated please? I can't find it

Thanks!

brendongo:
Sorry was on a trip and dropped the ball on this. This test looks great. You probably need to install the semgrep-core binary https://semgrep.dev/docs/contributing/semgrep-contributing/#getting-the-semgrep-core-binary but that seems overkill for this simple fix so I can just take this over the line. Thanks for the contribution @dee-see!

brendongo (approved):

PR #4704, opened 2022-02-18
Get the modified function in a diff

emjin:
Script to support SCA rules generation

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/experiments/synthesizing/Synthesizer.mli
```diff
@@ -5,12 +5,14 @@ val synthesize_patterns :
...
 (* "start row:start col-end row:end col"+ file ((range list)@[file]) -> pattern list *)
 (* Generates a pattern from several target ranges that matches each target.
-  Returns the detected language, the list of targest, and the Pattern list.
+   Returns the detected language, the list of targest, and the Pattern list.
 *)
 
 val generate_pattern_from_targets :
   Config_semgrep.t -> string list -> Lang.t * AST_generic.any list * Pattern.t
 
+val synthesize_from_diff_list : string -> unit
```
mjambon: You should document what's not obvious about the function. Here this would be the format or origin of the string argument, and also ideally explain what it's for.

semgrep-core/src/experiments/synthesizing/Test_synthesizing.mli
```diff
@@ -3,3 +3,5 @@ val expr_at_range : string -> Common.filename -> unit
 val synthesize_patterns : string -> Common.filename -> unit
 
 val generate_pattern_choices : string list -> unit
+
+val locate_function_from_diff : string -> unit
```
mjambon: same remark as above, we should document new interfaces.

semgrep-core/src/experiments/synthesizing/Pattern_from_Diff.ml
```diff
@@ -0,0 +1,128 @@
+module In = Input_to_core_j
```
mjambon: I wouldn't capitalize `diff` in the filename.

semgrep-core/src/core/Range.mli
```diff
@@ -19,6 +19,8 @@ val ( $<=$ ) : t -> t -> bool
 (* disjoint *)
 val ( $<>$ ) : t -> t -> bool
 
+val range_of_line_spec : string -> Common.filename -> t
```
mjambon, line 23: It would be great to explain what kind of string format is expected as input. Same for the old function below (`range_of_linecol_spec`).

interfaces/Input_to_core.atd
```diff
@@ -42,3 +42,18 @@ type target = {
...
+
+(*****************************************************************************)
+(* Synthesizing from diffs *)
+(*****************************************************************************)
+
+type diff_file = {
+  filename : string;
+  diffs : string (* start_line-end_line *) list;
+  url : string
+}
```
mjambon, line 55: * why not `(int * int) list` for the list of (start_line, end_line)?
* is line-numbering 0-based or 1-based?
* what's the `url` field?

mjambon (approved):

semgrep-core/src/experiments/synthesizing/Pattern_from_Diff.ml
```diff
@@ -0,0 +1,128 @@
...
+module Out = Output_from_core_j
+module R = Range
+module E = Entity_code
+module G = Graph_code
+module Set = Set_
+
+(*****************************************************************************)
```
aryx: all this code needs to move to deep-semgrep now that Graph_code_AST has been moved there.
There is a src/experiments/ there too!

aryx (commented):
Also not sure what's going on with all the reindentation in unrelated files.

interfaces/Input_to_core.atd
emjin: Thanks for calling out this and all the other places where I didn't document!

semgrep-core/src/experiments/synthesizing/Pattern_from_Diff.ml
emjin: I'm going to merge this first but will do

PR #4705, opened 2022-02-19
fix off by one in text formatter output highlighting

### LINKED ISSUE #4400: [Python] semgrep always highlights one extra character
**Describe the bug**
When semgrep highlights parts of the shown lines to demonstrate which parts matched the given expression, it always highlights one more character than needed.

**To Reproduce**
Create a file named `test.py` with the following contents:

```python
import sys
sys.exit(0)
```

Then run `semgrep -l python -e sys.exit test.py`. It outputs the following:

```
test.py
2:sys.exit(0)
```

The following segment is highlighted: `sys.exit(`.

**Expected behavior**
The left parenthesis should not be highlighted.

**Screenshots**
image

**Environment**
semgrep 0.76.1 (official binary) running on Python 3.6.

ievans:
Fixes #4400 

I'm suspicious of how simple this fix is, but manual testing seems to indicate we've just been adding an additional character to our output highlight for a long time.

I added tests for this along with a lot of our other text output in https://github.com/returntocorp/semgrep/pull/4706

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - The environment variable used by Semgrep login changed from `SEMGREP_LOGIN_TOKEN` to `SEMGREP_APP_TOKEN`
 - A new subcommand `semgrep publish` allows users to upload private, unlisted, or public rules to the Semgrep Registry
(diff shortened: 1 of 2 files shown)


emjin (approved):
Could you also add a test for this so we don't break it again? 

Edit: Just reread the PR comment

PR #4706, opened 2022-02-19
Add tests for color output

ievans:
To prevent regressions like https://github.com/returntocorp/semgrep/issues/4637

PR checklist:

- [ ] Changelog is up-to-date


ievans:
We can remove the `force_color` option from some of these if we want; I don't know if testing highlighting on error output is particularly useful, and it does clutter the diffs.

underyx (approved):

PR #4709, opened 2022-02-19
use the target as cwd when evaluating git origin

### LINKED ISSUE #4103: --config=auto detects wrong project URL
**Describe the bug**
The --config=auto mode detects the project of the current working directory, not the project in the target repository. 

**Expected behavior**
--config=auto should detect the project URL of the target specified, and possibly enforce that only a single  path target is supplied.

ievans:
Fixes https://github.com/returntocorp/semgrep/issues/4103

Needs careful review by @brendongo though

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/commands/scan.py
+++ b/semgrep/semgrep/commands/scan.py
@@ -750,7 +750,10 @@ def scan(
                 )
             else:
                 resolved_configs, config_errors = semgrep.config_resolver.get_config(
-                    pattern, lang, config or [], project_url=get_project_url()
+                    pattern,
+                    lang,
+                    config or [],
+                    project_url=get_project_url(target_sequence),
                 )
 
                 # Run metachecks specifically on the config files
--- a/semgrep/semgrep/project.py
+++ b/semgrep/semgrep/project.py
@@ -4,6 +4,7 @@ Utility for identifying the URL of the current git project
 import subprocess
 from typing import cast
 from typing import Optional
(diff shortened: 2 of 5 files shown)


semgrep/semgrep/commands/scan.py
```diff
@@ -750,7 +750,10 @@ def scan(
                 )
             else:
                 resolved_configs, config_errors = semgrep.config_resolver.get_config(
-                    pattern, lang, config or [], project_url=get_project_url()
+                    pattern,
+                    lang,
+                    config or [],
+                    project_url=get_project_url(target_sequence),
```
ievans, line 756: @brendongo this is the biggest thing I'm not sure about. What should the behavior be here for the scan subcommand?

ievans:
@brendongo pls review?

semgrep/semgrep/project.py
```diff
@@ -12,17 +13,29 @@
...
+        target = targets[0]
+
     try:
         return cast(
             str,
             sub_check_output(
                 ["git", "ls-remote", "--get-url"],
                 encoding="utf-8",
                 stderr=subprocess.DEVNULL,
+                cwd=target,
```
brendongo, line 38: I don't think this is right? Does this work if target is not a directory?

ievans: you're right, it throws `NotADirectoryError` if it's a directory

PR #4712, opened 2022-02-22
Remove additional `--update` for apk in Dockerfile

PeterDaveHello:
There is no need to use `--update` with `--no-cache` when using `apk` on
Alpine Linux, as using `--no-cache` will fetch the index every time and
leave no local cache, so the index will always be the latest without
temporary files remain in the image.

PR checklist:

- [ ] Changelog is up-to-date


underyx:
Thank you very much @PeterDaveHello! Very nice find.

underyx (approved):

Issue #4708, opened 2022-02-19
And a --apply flag that lets you easily apply for a job at r2c


clintgibler:
**Is your feature request related to a problem? Please describe.**
The problem is that there are awesome people out there who don't yet work at r2c.

**Describe the solution you'd like**
A simple implementation might just hit the Lever API and display our [currently live](https://jobs.lever.co/returntocorp/) list of jobs and a link to them, e.g:

```
We love that you're using Semgrep would you like to... work on it full time?! 😄 

############
# Security #
############

* Senior Security Researcher: https://jobs.lever.co/returntocorp/c0e3a208-9a76-4fc6-9a03-dadc3ebc6f38
* Senior Security Engineer: https://jobs.lever.co/returntocorp/9ac6d4bc-a939-4bf9-8bf0-983774cb23d1

###############
# Engineering #
###############

* Software Engineer, Program Analysis: https://jobs.lever.co/returntocorp/a0185a91-bc99-428c-9441-515489212738
* ...
```

**Describe alternatives you've considered**
Lever, LinkedIn, Twitter, smoke signals, hiring blimps to fly over hacker conferences flying the Semgrep logo, etc.

**Use case**
Hiring a world-class team and taking Semgrep to the moon 🚀 

**Additional context**
@adamberman had this excellent idea and I wanted to write it down.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-891/and-a-apply-flag-that-lets-you-easily-apply-for-a-job-at-r2c.

PR #4716, opened 2022-02-22
added --apply flag that posts current job postings at r2c

0xDC0DE:
PR checklist:

https://github.com/returntocorp/semgrep-docs/commit/96d5931bbcd8b1f52256c7d810267273abbd6fe6
- [ ] Changelog is up-to-date

Implemented https://github.com/returntocorp/semgrep/issues/4708

### DIFF
--- a/semgrep/semgrep/commands/scan.py
+++ b/semgrep/semgrep/commands/scan.py
@@ -152,6 +152,14 @@ CONTEXT_SETTINGS = {"max_content_width": 90}
 @click.argument("target", nargs=-1, type=click.Path(allow_dash=True))
 @click.help_option("--help", "-h", help=("Show this message and exit."))
 @click.option(
+    "--apply",
+    is_flag=True,
+    help=(
+        "Print a list of job postings at r2c."
+    ),
+    # hidden=True, # Not sure if this should be hidden
+)
+@click.option(
     "-a",
     "--autofix/--no-autofix",
     is_flag=True,
@@ -553,6 +561,7 @@ CONTEXT_SETTINGS = {"max_content_width": 90}
 )
 def scan(
     *,
+    apply: bool,
     autofix: bool,
     baseline_commit: Optional[str],
     config: Optional[Tuple[str, ...]],
@@ -630,6 +639,12 @@ def scan(
             from semgrep.version import version_check
 
             version_check()
+        return
+
+    if apply:
+        from semgrep.job_postings import print_job_postings
+        
+        print_job_postings()
         return
 
     # To keep version runtime fast, we defer non-version imports until here
--- a/semgrep/semgrep/commands/scan.py
+++ b/semgrep/semgrep/commands/scan.py
@@ -155,7 +155,7 @@ CONTEXT_SETTINGS = {"max_content_width": 90}
     "--apply",
     is_flag=True,
     help=("Print a list of job postings at r2c."),
-    # hidden=True, # Not sure if this should be hidden
+    hidden=True,
 )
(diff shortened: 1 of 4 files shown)


semgrep/semgrep/job_postings.py
```diff
@@ -0,0 +1,48 @@
...
+
+    print(
+        "We love that you're using Semgrep. Would you like to... work on it full time?! 😄 "
+    )
+    print("")
+    # iterate over teams alphabetically
+    for team in sorted(postings_dict):
+        team_character_count = len(team)
+
+        print_n_hashtags(team_character_count + 4)
```
underyx:
```suggestion
        print((team_character_count + 4) * "#")
```

I think the separate function is not adding much value over writing this.

semgrep/semgrep/job_postings.py
```diff
@@ -0,0 +1,48 @@
...
+        print("#", team, "#")
+        print_n_hashtags(team_character_count + 4)
+        print("")
+
+        for posting in postings_dict[team]:
+            print("*", posting["title"] + ":", posting["url"])
+
+        print("")
```
underyx:
```suggestion
        print("")
        print("PS: Just wanted to apply autofixes? Use --autofix instead of --apply")
```

semgrep/semgrep/job_postings.py
```diff
@@ -0,0 +1,48 @@
...
+    for posting in postings:
+        team = posting["categories"]["team"]
+        if team not in postings_dict:
+            postings_dict[team] = []
+
+        title = posting["text"]
+        url = posting["hostedUrl"]
+        postings_dict[team].append({"title": title, "url": url})
+
+    print(
```
underyx: I think we use `click.echo(..., err=True)` usually for system messages 🤔 I'm really not sure if we want to enforce that though.

semgrep/semgrep/commands/scan.py
```diff
@@ -151,6 +151,12 @@ def convert(
 @click.command(context_settings=CONTEXT_SETTINGS)
 @click.argument("target", nargs=-1, type=click.Path(allow_dash=True))
 @click.help_option("--help", "-h", help=("Show this message and exit."))
+@click.option(
+    "--apply",
+    is_flag=True,
+    help=("Print a list of job postings at r2c."),
+    # hidden=True, # Not sure if this should be hidden
```
underyx:
```suggestion
    hidden=True,
```

yeah I think this makes most sense as a hidden easter egg 👍

underyx (approved):
I'm gonna go ahead and approve this since this is totally isolated from the rest of semgrep and it's not like maintainability matters a whole lot here 👍 So all the comments I left are optional (and/or educational :D).

brendongo:
Can we have this as a separate sub command instead of a flag in scan?

aryx:
Love this :) Actually I think it should be displayed when we do --help. I kinda like that it's buried inside other regular flags :)

semgrep/semgrep/commands/scan.py
```diff
@@ -151,6 +151,12 @@ def convert(
 @click.command(context_settings=CONTEXT_SETTINGS)
 @click.argument("target", nargs=-1, type=click.Path(allow_dash=True))
 @click.help_option("--help", "-h", help=("Show this message and exit."))
+@click.option(
+    "--apply",
+    is_flag=True,
+    help=("Print a list of job postings at r2c."),
+    hidden=True,
```
aryx, line 158: why hidden? :)

PR #4726, opened 2022-02-24
metavar-comparison with strip now extracts the metavar's digits

underyx:
PR checklist:

- [x] Documentation is up-to-date — https://github.com/returntocorp/semgrep-docs/pull/453
- [x] Changelog is up-to-date


CHANGELOG.md
```diff
@@ -31,6 +31,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - PHP: Constant propagation now has built-in knowledge of `escapeshellarg` and
   `htmlspecialchars_decode`, if these functions are given constant arguments,
   then Semgrep assumes that their output is also constant
+- When a `metavariable-comparison` has `strip: true` set,
```
IagoAbal: Better rebase and fix the changelog, otherwise when we merge this PR this entry will go under `0.83.0` which is already released.

semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -158,9 +158,20 @@ let rec eval env code =
       match v with
       | Int _ -> v
       | String s -> (
-          match int_of_string_opt s with
-          | None -> raise (NotHandled code)
-          | Some i -> Int i)
+          (* This code is meant to emulate int(string) in Python.
+           * But! There is one difference:
+           * when called as int("12px"), it will still return 12 instead of erroring.
```
IagoAbal: What the spec of `metavariable-comparison` says?

> The metavariable-comparison operator compares metavariables against a basic [Python comparison](https://docs.python.org/3/reference/expressions.html#comparisons) expression.

This breaks that spec and it will produce extra findings... and some users may not be happy about that. Rather than changing what `int()` does, I would prefer adding a different function `r2c_int` or something.

Ideally we should just expand the subset of Python that we support. But if that is a hassle, then I think the second best solution is to add `r2c_xyz` functions.

IagoAbal (changes requested):

semgrep-core/src/engine/Eval_generic.ml
IagoAbal: Or, maybe, we pretend there is an `r2c` module being imported, and you can refer to the functions as `r2c.my_int`?

underyx: Good point! Thanks for bringing this up, I wasn't sure what level of faithfulness we were looking for but this makes perfect sense. Since it's common in python to define helper functions without a namespace, I didn't bother with a virtual `r2c.` module. Also, I figured it's more clean two have two composable functions instead of a regular `int(...)` and a reimplementation of `int(...)` with different behavior. So I ended up adding a `digits(value: str) -> str` function and documented it like this: https://github.com/returntocorp/semgrep-docs/pull/453/files#diff-27ce63caf629155835158f784d2556151d037d7018d5618da4eab9cce3195162R208-R214

semgrep-core/src/core/Rule.ml
```diff
@@ -316,7 +316,7 @@ let kind_of_formula = function
 (* Converters *)
 (*****************************************************************************)
 
-(* Substitutes `$MVAR` with `int($MVAR)` in cond. *)
+(* Substitutes `$MVAR` with `int(digits($MVAR))` in cond. *)
```
IagoAbal, line 319: This preserves the meaning of `int` but it breaks the spec of the `strip:` key. I would avoid that too. You could perhaps add a different key. (Although I personally think `metavariable-comparison` should just take a Python expression and nothing else, but that's just my opinion anyways.)

semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -155,13 +155,33 @@ let rec eval env code =
...
           | Some i -> Int i)
       | __else__ -> raise (NotHandled code))
+  (* Custom digits() operator
+   * This is a virtual builtin helper function provided by Semgrep.
+   * It returns the first group of digits in a value.
+   * It also keeps underscores, and leading minus signs.
+   * The returned value will always cleanly pass through Python's int().
+   * This is useful e.g. to write rules that work with generic captures in CSS.
+   *)
+  | G.Call ({ e = G.N (G.Id (("digits", _), _)); _ }, (_, [ Arg e ], _)) -> (
```
IagoAbal, line 173: Nice! If you don't want to have an `r2c` module then I would at least prefix it with `r2c_` to avoid (future) potential clashes with Python's stdlib.

semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -155,13 +155,33 @@ let rec eval env code =
...
+      let v = eval env e in
+      logger#trace "called digits(%s)" (show_value v);
```
IagoAbal, line 175: Probably want to remove this from the PR?

semgrep-core/src/engine/Eval_generic.ml
IagoAbal: What if we added support for Python's `rstrip` ? You can then write your example as `$PIXELS.rstrip("px") > 1`. It's more explicit about what you are expecting (the `px` suffix) and it's pure Python.

underyx: :one: starting off with a nitpick: the correct methods would be `.removeprefix()` and `.removesuffix()`, since the spec of `rstrip()` says it takes a sequence of characters and it removes all occurrences of any of them from the right of the string. So it'd also convert `"12pxxxpxpxpxppxp"` to `"12"`. And `12xp` to `12`. Neither of those are likely desired (imagine trying to match a distance in meters, but getting `"1000mm".rstrip("m") == "1000"`).

:two: With that aside, our rule syntax spec already has a `strip: true` convenience feature, which I was hoping to just make smarter, instead of making everyone increase the complexity of their Python comparisons. Basically I'd prefer if people could write

```yaml
strip: true
comparison: $X == $Y
```

instead of

```yaml
comparison: int($X.removesuffix("px")) == int($Y.removesuffix("px"))
```

in their rules. We can make `strip` do the right thing with int & digits cause they take no parameters other than the metavar. But if we have `.removesuffix()` instead, that is not possible.

:three: For the font-size case, you can specify the size as both integer values and strings, see excerpt from semgrep.dev code:

CleanShot 2022-02-28 at 08 57 51@2x
CleanShot 2022-02-28 at 08 58 09@2x

If we tried to keep 100% purity, `$PIXELS_INTEGER.removesuffix("px")` would throw an error, since this method does not makes sense and is not available on integers. So you'd end up having to add a `pattern-either`, then typed metavariable matching, and duplicate your metavar-comparison. What's worse is, without an IDE that can show you hints about types, this issue would be near-impossible to find out about and then debug. So considering the lack of tooling available to write correct metavar-comparison snippets… I think we're better off keeping it simple and just handling this for users who turn on `strip: true`.

semgrep-core/src/core/Rule.ml
underyx: > it breaks the spec of the strip: key. I would avoid that too. 

Is there any reason for that? We're still on 0-ver and can introduce breaking changes. This is also a very, very obscure feature (perhaps specifically because most usages would need strip to strip more characters than just quotes). I reviewed the rules our production instance has and we have four usages of `metavariable-comparison` with `strip: true`:

1. the rule we embedded in the documentation to demo the feature
2. a test rule from an external user that never worked; they thought they could compare against a regex capture group by writing `$1`
3. a test rule by Isaac which I never worked, because strip only works on str/int metavars, but he was matching a function name with it
4. my CSS rule

Basically this feels like a pretty straightforward, harmless, and useful spec update to me.

aryx:
What's the use case for this new behavior? Who needs that feature? From the example it looks like it's mostly
for JS and CSS where the syntax to specify units is quite loose. Seems dangerous to me to get strip to suddenly
remove not just the quotes but also some suffix or prefix. For example the rule bence wrote use the $PIXEL metavariable name, but what happens if the code contains a = "12cm", or "1inch"? This is not a pixel anymore, so the rule should explicitely check that it's using pixels.

aryx:
But maybe the rule/pattern syntax we currently have could be improved to make it easier to match those fontsize dimensions.

aryx:
Maybe we could have a special language to match dimensions, so we could use metavariable-pattern to refine the constraints we want on the fontsize value. A bit like adding support for SQL would be nice to use metavariable-pattern in python or C on strings that contains SQL statements.

aryx:
I wonder if we need to parse those dimensions in languages other than JS and CSS.

aryx:
Some languages such as F# have even special types for those dimensions, so they can even do stuff like. '1 cm >= 3 inch'
so metavariable-comparison could also be extended to be aware of those dimensions, so you could
just do metavariable-comparison: comparison: $FONTSIZE > 1 px

semgrep-core/src/core/Rule.ml
IagoAbal: This is why I would have appreciated having a discussion about when it's acceptable to break things. Indeed we are pre-1.0, so users can expect we break things sometimes. However, we also have a distinction between GA, beta, and experimental features. Breaking things also has a cost in terms of reputation and such.

IMO it is OK to break experimental features (e.g. join-mode or previously taint-mode). IMO it is also OK to change something when we made an obvious mistake and later we realize that there is a better way. For example, we had `pattern-where-python:` which was a mistake, and we removed support for it and instead added `metavariable-comparison` (and other `metavariable-xyz` operators).

In this case I don't see why the current spec of `strip: true` is a mistake in any way. In fact I can see how some rule authors may not like it. It is not obvious at all to me that everyone wants "123abc" to be interpreted as `123` in every use of `metavariable-comparison:`.

underyx:
> What's the use case for this new behavior? Who needs that feature? From the example it looks like it's mostly for JS and CSS where the syntax to specify units is quite loose. Seems dangerous to me to get strip to suddenly remove not just the quotes but also some suffix or prefix. For example the rule bence wrote use the $PIXEL metavariable name, but what happens if the code contains a = "12cm", or "1inch"? This is not a pixel anymore, so the rule should explicitely check that it's using pixels.

my rule has a metavariable-regex to filter to the correct unit

aryx:
Do you have a link to the full rule you want to write? I wonder if there's a better way to solve your problem.

underyx:
> Do you have a link to the full rule you want to write? I wonder if there's a better way to solve your problem.

https://semgrep.dev/s/returntocorp:metavar-comparison-example

aryx:
For your case I think it would be simpler to have:
```
rules:
  - id: metavar-comparison-example
    patterns:
      - pattern-regex: "font-size: ([0-9]+)px;"
      - metavariable-comparison:
        - metavariable: $1
        - comparison: $1 > 2
    message: "Semgrep found a match: $1"
    languages:
      - none
    severity: WARNING
```
which currently does not work, but that we could extend to make it work.

aryx:
Actually the rule that I wrote make spin forever the playground ...

underyx:
That would be an okay solution… but it doesn't cover other cases I'd also like to match such as

```tsx
const BAD_SIZE = "13px"
const OTHER_BAD SIZE = 14

<p style={{ fontSize: BAD_SIZE }} />
<p style={{ fontSize: OTHER_BAD_SIZE }} />
```

I don't know if metavariable-comparison would currently understand the dataflow here, but this would clearly never work with regex anyway.

aryx:
But does the rule you wrote that use 'generic' handle constant propagation?

underyx:
> But does the rule you wrote that use 'generic' handle constant propagation?

Not the one on the editor, but this example of another thing I'd like to be able to match is a valid TSX snippet.

aryx:
Do you have the exact rule you want to write (that does not use generic mode but TSX)?
Because you also rely on metavariable-regexp to make sure it has a "px", which would not work either I guess.

I think what you want to check is cool, and something we should work on, I just think 'strip: true' is not the right solution for that.

underyx:
@aryx here's the typescript one: https://semgrep.dev/orgs/returntocorp/editor/s/returntocorp:font-sizes-should-be-multiples-of-four

aryx:
So do you agree that because it could strip other dimensions, e.g., const FOO = "2cm", it
seems bad to have strip arbitrarily remove not just quotes but also characters before and after the digit?
We want to enforce more constraints.

aryx:
Now, how would you write the check in python about the content of $PIXELS?
In OCaml I would write:  '$PIXELS =~ "\"?([0-9]+)(px)?\"?" && int_of_string (Common.matched1 $PIXELS) > 2'

semgrep-core/src/engine/Eval_generic.ml
IagoAbal: Note that
```
strip: true
comparison: $X == $Y
```
does not work as you suggest, you need to pick _one_ metavariable to _strip_.

aryx: Another good reason to redesign metavariable-comparison and make strip() (and other useful functions) regular functions you can use on any metavariables in the comparison expression.

aryx: I've created this new issue: https://github.com/returntocorp/semgrep/issues/4764

IagoAbal:
I like the `pattern-regex:` solution, although it's true that it won't work with const-prop. But we could make it work somehow. It could be useful in other cases. (For the record, `metavariable-comparison:` did not work with constant propagation either, until a few months ago.)

**update:** Right now you could match the pattern `fontSize: "$SIZE"` then use `metavariable-pattern:` on `$SIZE`. That should work.

aryx:
nah I don't think we can make pattern-regex: work with constant prop because it does not even work on the generic AST.
It just sees a giant string.

PR #4735, opened 2022-02-25
Add `focus-metavariable:` operator

### LINKED ISSUE #4453: Limit the result to the function argument
For example, if I use the pattern `requests.get(..., verify=False, ...)` to match a function with a specific argument, it will report the entire function call like:
```
requests.get(
  "https//www.example.com",
  params={
    "a": "a",
    "b": "b",
    "c": "c",
  },
  headers={
    "X-REQUEST-ID": "123456",
  },
  timeout=10,
  verify=False,
)
```
Therefore, any changes to other arguments would be reported as a new finding in Semgrep CI.
This might be the desire behavior in some cases, but if a function call (or class initialization) has many arguments, it often reports irrelevant changes.

A feature to optionally report new findings for changes in only part of a pattern (e.g. argument of a function call).
Throwing out some ideas here.
1. Use special syntax to limit the scope. For example, `requests.get(..., $[[verify=False]], ...)`.
2. A new option to limit the result to a metavariable.
  Something like
```
patterns:
   - pattern-inside: 'requests.get(..., verify=$X, ...)'
   - pattern: $X
```
Doesn't work in general because assume `$X` is `False`, other argument with the value `False` would also be reported.
However, maybe we can have a new field to tell semgrep to only return a metavariable.
For example,
```
pattern: 'requests.get(..., verify=$X, ...)'
result-metavariable: $X
```
3. Similar to option 2, but use a special metavariable.
For example, `pattern: 'requests.get(..., verify=$RESULT, ...)'`
4. Add a new option similar to option 2, but semgrep will return the same finding and only limit the scope in Semgrep CI.
For example,
```
pattern: 'requests.get(..., verify=$X, ...)'
ci-context: $X
```

IagoAbal:
This operator allows to focus matching on a given metavariable, within a
`patterns:` operator.

Closes #4453

test plan:
make test # test included

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -109,8 +109,13 @@ type formula =
    * (see tests/OTHER/rules/negation_exact.yaml)
    *)
   | P of xpattern (* a leaf pattern *) * inside option
-  (* see Match_rules.split_and() *)
-  | And of tok * formula list * (tok * metavar_cond) list
+  (* see Specialize_formula.split_and() *)
+  | And of
+      tok
+      * formula list
+      * (tok * metavar_cond) list
+      * (* focus-metavariable:'s *)
+      (tok * MV.mvar) list
   | Or of tok * formula list
   (* There are currently restrictions on where a Not can appear in a formula.
    * It must be inside an And to be intersected with "positive" formula.
@@ -152,7 +157,10 @@ type formula_old =
   | Pat of xpattern
   (* pattern-not: *)
   | PatNot of tok * xpattern
+  (* metavariable-xyz: *)
   | PatExtra of tok * extra
+  (* focus-metavariable: *)
+  | PatFocus of tok * MV.mvar
   (* pattern-inside: *)
   | PatInside of xpattern
   (* pattern-not-inside: *)
@@ -294,12 +302,12 @@ let rec visit_new_formula f formula =
   | P (p, _) -> f p
   | Not (_, x) -> visit_new_formula f x
   | Or (_, xs)
-  | And (_, xs, _) ->
+  | And (_, xs, _, _) ->
       xs |> List.iter (visit_new_formula f)
 
 (* used by the metachecker for precise error location *)
 let tok_of_formula = function
-  | And (t, _, _)
+  | And (t, _, _, _)
   | Or (t, _)
   | Not (t, _) ->
       t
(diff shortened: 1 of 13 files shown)


semgrep-core/src/core/Rule.ml
```diff
@@ -109,8 +109,13 @@ type formula =
    * (see tests/OTHER/rules/negation_exact.yaml)
    *)
   | P of xpattern (* a leaf pattern *) * inside option
-  (* see Match_rules.split_and() *)
-  | And of tok * formula list * (tok * metavar_cond) list
+  (* see Specialize_formula.split_and() *)
+  | And of
```
aryx: maybe we can introduce a record at this point.

aryx (approved):

ievans (commented):
@IagoAbal can we hold this for a bit and try to get some external user feedback on it? It is really not obvious to me what the operator does, even reading through the example, which concerns me as a long-time Semgrep user. Do you have a docs PR that I could share with some users?

Also I feel like a test for the JSON output of this would be good

IagoAbal:
> @IagoAbal can we hold this for a bit and try to get some external user feedback on it? It is really not obvious to me what the operator does, even reading through the example, which concerns me as a long-time Semgrep user. Do you have a docs PR that I could share with some users?

No docs PR yet, I will work on one. But given a `focus-metavariable: $X` operator, whatever your patterns match it will be intersected with the range of what `$X` matches. So you can "focus" the match on a specific metavariable.

ievans:
I think the main thing I'm missing is what's different from adding a `pattern: $X`? (just performance or are there other implications)

Isaac0616:
Not sure about `focus-metavariable` but only adding `pattern: $X` doesn't quite work because, for example,
```
patterns:
   - pattern-inside: 'requests.get(..., verify=$X, ...)'
   - pattern: $X
```
will match every `False` in `request.get(url, verify=False, foo=False, bar=False)`.

aryx:
II guess we can add poning's examples in the docs, so the difference are clear.

IagoAbal:
Oh OK, I can add a simpler test case that exemplifies this in a more obvious way.

```
patterns:
    - pattern-inside: foo($X, ...)
    - focus-metavariable: $X
    - pattern: 1
```

matches

```
foo (1,  # here
     2)  # not here
```

but it does not match

```
foo (2,
     1)
```

also it would match:

```
foo (1,  # here
     1)  # not here
```

whereas `pattern: $X` would have matched all the occurrences of `1`.

PR #4750, opened 2022-03-04
chore(docker): pin primary build by sha hash

ievans:
PR checklist:

- [x] Changelog is up-to-date n/a


aryx:
Why do we need the extra precision @ievans ?

aryx (approved):

Issue #4539, opened 2022-01-19
Add `-u` flag (potentially repeatable) to disable filtering


xmo-odoo:
Copy of https://github.com/returntocorp/semgrep/issues/1964#issuecomment-1011849002 on the recommendation of @ievans.

semgrep has support for `.gitignore` files and its own `.semgrepignore` (possibly `.ignore` as well? Unsure).

It does have a flag to skip `.gititgnore` (i don't think it has one for `.semgrepignore` as yet? but `--no-git-ignore` can be a bit of a ~~mouthful~~ fingerful, and it's unclear what other hiding behaviour semgrep will have in the future (e.g. hidden files).

Here `ripgrep` has individual flags for ignore sources (e.g. `--no-hidden`, `--binary`, and a slew of `--no-ignore-$X` related to various categories of ignore files).

But it also has a repeatable `--unrestricted` (`-u`) flag which allows bypassing those ignore categories conveniently:

> Reduce the level of "smart" searching. A single -u won't respect .gitignore (etc.) files (--no-ignore). Two -u flags will additionally search hidden files and directories (-./--hidden). Three -u flags will additionally search binary files (--binary).
>
> 'rg -uuu' is roughly equivalent to 'grep -r'.

This is extremely convenient at the command-line (using the tool interactively) when trying to maximise the search area e.g, you don't find what you expected, you can just pop in a `-u` or 3, obviously at a cost (since it's probably going to trawl through a lot more things).

I think semgrep could benefit from this.

aryx:
@mmcqd recently added support for .semgrepignore in semgrep itself (it was only in the agent before). Maybe he has some feedback on this feature request.

xmo-odoo:
> @mmcqd recently added support for .semgrepignore in semgrep itself (it was only in the agent before). Maybe he has some feedback on this feature request.

Makes sense, this feature suggestion is, in a way, a followup to that change.

xmo-odoo:
![animal 3](https://user-images.githubusercontent.com/7571158/152739463-9ecc8d56-137f-4d26-a480-618be9425d14.png)

mjambon:
To ignore `.semgrepignore` files, use `--x-ignore-semgrepignore-files` or `--x-semgrepignore-filename` (subject to change).

This is part of our [Semgrepignore v2](https://semgrep.dev/docs/semgrepignore-v2-reference) implementation that is the default behavior since Semgrep 1.117.0. It works like Gitignore.

PR #4759, opened 2022-03-08
add a semgrepignore so the cheatsheet generation works

ievans:
@returntocorp/pa I'm unsure how the generation was working since we added .semgrepignore? Is it still part of the release process?

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/scripts/generate_cheatsheet.py
+++ b/scripts/generate_cheatsheet.py
@@ -224,6 +224,11 @@ def run_semgrep_on_example(
         pattern_text = open(config_arg_str).read()
         config.write(_config_to_string(_single_pattern_to_dict(pattern_text, lang)))
         config.flush()
+
+        with open(".semgrepignore", "w") as ignore_out:
+            ignore_out.writelines("")
+            ignore_out.flush()
+
         cmd = ["semgrep", "--strict", "--json", f"--config={config.name}", code_path]
         print(">>> " + " ".join(cmd))
         output = subprocess.run(  # nosemgrep: python.lang.security.audit.dangerous-subprocess-use.dangerous-subprocess-use


aryx:
According to https://r2c.quip.com/zrklAwLqfrm7/Using-the-Semgrep-Repo it's actually not part of the release anymore. @nbrahms removed it because it was not automated.
(latest version of the document is here: https://docs.google.com/document/d/1vz13vs7CmTfHAs6UUbpZk3IZm5lxwo9Xxo8xxMkwchU/edit#heading=h.4p1gca9pu4r5

scripts/generate_cheatsheet.py
```diff
@@ -224,6 +224,11 @@ def run_semgrep_on_example(
         pattern_text = open(config_arg_str).read()
         config.write(_config_to_string(_single_pattern_to_dict(pattern_text, lang)))
         config.flush()
+
+        with open(".semgrepignore", "w") as ignore_out:
```
aryx, line 228: Maybe a comment explaining why, that is if there's no .semgrepignore then semgrep uses one by default which skip tests/ directories?

aryx: Or is it because we have a .semgrepignore that skip semgrep-core/tests?

aryx: Also where this .semgrepignore is created? In which directory? Shouldn't we delete it after?

aryx (changes requested):

scripts/generate_cheatsheet.py
ievans: @aryx we can hold off on this and just wait for https://github.com/returntocorp/semgrep/issues/4539. I'll close the PR till then

PR #4761, opened 2022-03-08
fix(core_runner): performance regression with --debug 

brendongo:
When semgrep was run with --debug we were seeing a significant performance
regression than without. The root cause was in core_runner we were appending
stdout and stderr to a string which was running in O(n^3) for n line length.

With this PR we append to a list O(1) and do a single join at the end
removing the perf regression

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -147,6 +147,8 @@ class StreamingSemgrepCore:
 
         When it sees non-"." output it saves it to self._stdout
         """
+        stdout_lines: List[str] = []
+
         # appease mypy. stream is only None if call to create_subproccess_exec
         # sets stdout/stderr stream to None
         assert stream
@@ -156,6 +158,7 @@ class StreamingSemgrepCore:
 
             # readline returns empty when EOF
             if not line_bytes:
+                self._stdout = "".join(stdout_lines)
                 break
 
             line = line_bytes.decode("utf-8")
@@ -163,7 +166,7 @@ class StreamingSemgrepCore:
                 if self._progress_bar:
                     self._progress_bar.update()
             else:
-                self._stdout += line
+                stdout_lines.append(line)
 
     async def _core_stderr_processor(
         self, stream: Optional[asyncio.StreamReader]
@@ -174,6 +177,8 @@ class StreamingSemgrepCore:
         Basically works synchronously and combines output to
         stderr to self._stderr
         """
+        stderr_lines: List[str] = []
+
         if stream is None:
             raise RuntimeError("subprocess was created without a stream")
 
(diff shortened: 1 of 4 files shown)


semgrep/tests/e2e/test_performance.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    Verify that running semgrep with --debug does not result in
+    performance slowdown wrt running without --debug
+    """
+    start_without_debug = time.time()
+    run_semgrep_in_tmp("rules/long_message.yaml", target_name="simple.yaml")
+    time_without_debug = time.time() - start_without_debug
+
+    # If this fails the rule was probably optimized out
+    assert time_without_debug > 2
```
underyx: **suggestion**: move advice to pytest error output
```suggestion
    assert time_without_debug > 2, (
        "runtime was suspiciously low, was the rule optimized out?"
    )
```

semgrep/tests/e2e/test_performance.py
```diff
@@ -0,0 +1,26 @@
...
+    run_semgrep_in_tmp(
+        "rules/long_message.yaml",
+        target_name="simple.yaml",
+        options=["--debug", "--time"],
+    )
+    time_with_debug = time.time() - start_with_debug
+
+    # There was a slowdown of more than 10% in reality the time should be basically the same
+    # but we account for variablity in runtime here
+    assert time_with_debug - time_without_debug < 0.1 * time_without_debug
```
underyx: **suggestion**: use pytest.approx for better readability
```suggestion
    from pytest import approx
    assert time_with_debug == approx(time_without_debug, rel=0.1), (
        "adding --debug slowed runtime by more than 10%"
    )
```

underyx: (if you accept this you'll need to move the import to top level)

underyx (approved):

semgrep/semgrep/core_runner.py
```diff
@@ -183,10 +188,11 @@ async def _core_stderr_processor(
 
             # readline returns empty when EOF
             if not line_bytes:
+                self._stderr = "".join(stderr_lines)
```
underyx, line 191: **thought**: we could skip calling "".join() after every line
Maybe it'd speed this up even more if we saved the lines on an instance attribute, and moved the "".join() call to the `@property`.

underyx: Ah, i see, it does actually run only once, when we encounter EOF. Ignore this comment :D

semgrep/tests/e2e/test_performance.py
brendongo: it did not but w/e

aryx (approved):

PR #4767, opened 2022-03-08
Support custom string interpolators

### LINKED ISSUE #4655: Scala String Interpolator Rules Do Not Match Against
**Describe the bug**
Scala 

**To Reproduce**
https://semgrep.dev/s/PYpe

**Expected behavior**
The rule should not match against `s""` or `sql""`, only `asdasdawkdpokasd""`. 

**Screenshots**

**Environment**

**Use case**
What will fixing this bug enable for you? Write a rule to detect unsafe String interpolation in Slick SQL queries (e.g. `sql"#$unsafe"`) as well as other String interpolation related vulnerabilities.

mmcqd:
Closes #4655

Adds support in the generic AST for scala's custom string interpolators. Now we can distinguish between `a"str"` and `b"str"`.

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/parsing/pfff/Python_to_generic.ml
```diff
@@ -199,7 +199,8 @@ let rec expr env (x : expr) =
       |> G.e
   | InterpolatedString xs ->
       G.Call
-        ( G.IdSpecial (G.ConcatString G.FString, unsafe_fake "concat") |> G.e,
+        ( G.IdSpecial (G.ConcatString (G.FString "f"), unsafe_fake "concat")
```
emjin, line 203: A comment to give context would be good here (Python strings are always `f"..."`, corresponding to Scala's `abc"..."`)

emjin (approved):

semgrep-core/src/parsing/pfff/Python_to_generic.ml
mmcqd: done

semgrep-core/src/parsing/pfff/Python_to_generic.ml
```diff
@@ -199,7 +199,9 @@ let rec expr env (x : expr) =
       |> G.e
   | InterpolatedString xs ->
       G.Call
-        ( G.IdSpecial (G.ConcatString G.FString, unsafe_fake "concat") |> G.e,
+        (* Python interpolated strings are always of the form f"...", we need to support arbitary strings in the generic AST because Scala has custom interpolators *)
```
IagoAbal, line 202: Shouldn't we avoid such long lines in the code? @mjambon do you know if ocamlformat can do this for us?

interfaces/AST_generic_v1.atd
```diff
@@ -420,7 +420,7 @@ type concat_string_kind = [
    * and some semgrep users may want to explicitely match only f-strings,
    * which is why we record this information here.
   *)
-  | FString (* Python *)
+  | FString of string(* Python *)
```
IagoAbal, line 423: Are we allowed to modify this AST?

```
 * DO NOT MODIFY THIS FILE. The "API" defined here should be stable
 * across multiple versions of Semgrep. If you really want to change
 * things, you'll have to define an AST_generic_v2.atd
```

IagoAbal: @aryx could you confirm whether this is OK?

aryx: Yes we're not supposed to modify this file because there is a _v1 so it's supposed to be the same forever.
Nobody use those (supposidely stable) exported ASTs so it's also fine.
But usually I don't modify and put some raise Todo in the AST_generic converter to AST_generic_v1

semgrep-core/src/experiments/api/AST_generic_to_v1.ml
```diff
@@ -411,7 +411,7 @@ and map_special x =
 and map_of_interpolated_kind = function
   | InterpolatedConcat -> `InterpolatedConcat
   | SequenceConcat -> `SequenceConcat
-  | FString -> `FString
+  | FString v1 -> `FString v1
```
aryx, line 414: Here I would just discard v1, or raise a Todo

aryx (commented):
Great!

PR #4772, opened 2022-03-09
project-depends-on: scan yarn.lock dependencies without hashes, and run rules with pattern-inside at their leaves

mmcqd:
This PR does two things:
1. Reads dependencies from `yarn.lock` files even if they have no hash. This happens sometimes and does not violate `yarn.lock` rules, and we allow this behavior in other lockfiles
2. Sends rules with only `pattern-inside` at their leaves to be run by `semgrep-core`. `pattern-inside` with no `pattern` is a valid pattern, but we previously considered it "un-runnable" when executing `project-depends-on` patterns, so we wouldn't even try running the full rule and would just check the dependency.

I combined these fixes because they were both needed to make a test pass. They can be split into two PRs if necessary.

PR checklist:

- [x] Changelog is up-to-date


adamberman (approved):

semgrep/semgrep/rule.py
```diff
@@ -219,7 +219,8 @@ def should_run_on_semgrep_core(self) -> bool:
 
         def has_runnable_rule(d: Dict[str, Any]) -> bool:
             for k in d:
-                if k in RuleValidation.PATTERN_KEYS:
+                # `pattern-inside` is valid without an accompanying `pattern`
```
aryx, line 222: I'm not sure we want to accept this. Usually people use pattern-inside: with a pattern:.
If you just have pattern-inside: everywhere, you can simply use pattern: instead.

mmcqd: You're right, but semgrep currently accepts this, so it would be weird if `project-depends-on` rules had different behavior.

PR #4777, opened 2022-03-10
feat(ci): subcommand that integrates with semgrep.dev

brendongo:
This PR adds `semgrep ci` as a subcommand for semgrep. This command takes primarily the same arguments as `semgrep scan` except it takes the configuration of rules and excludes on semgrep.dev (using credentials saved in `semgrep login`), runs semgrep rules over files, reports back findings for semgrep.dev to display, and exits with the appropriate exit code given online configuration and which rules triggered findings

PR checklist:

- [ ] Changelog is up-to-date


brendongo:
@underyx the semgrep-app communication part is mostly done. Have to move over meta.py, add tests, and update the syntactic_id hash to match semgrep-agent hash.

semgrep/semgrep/commands/app_scan.py
```diff
@@ -0,0 +1,122 @@
...
+        sys.exit(1)
+    if not Authentication.is_valid_token(token):
+        click.echo(
+            "API token not valid. Try to run `semgrep logout` and `semgrep login` again.",
+            err=True,
+        )
+        sys.exit(1)
+
+    scan_handler = ScanHandler(token)
+    metadata = {"repository": "returntocorp/semgrep"}
```
underyx: left here accidentally?

semgrep/semgrep/commands/app_scan.py
```diff
@@ -0,0 +1,122 @@
...
+        patterns.extend(DEFAULT_EXCLUDE_PATTERNS)
+
+    for pattern in patterns:
+        yield from ["--exclude", pattern]
+
+
+@click.command(context_settings=CONTEXT_SETTINGS)
+@click.pass_context
+@scan_options
+def app_scan(ctx: click.Context, *args: Any, **kwargs: Any) -> None:
```
underyx:
```
$ semgrep scan
$ semgrep app_scan
```
Doesn't sound like a usual subcommand name to me. Update to `semgrep ci`?

semgrep/semgrep/commands/app_scan.py
```diff
@@ -0,0 +1,122 @@
...
+        exit_code = scan_handler.fail_open_exit_code(metadata["repository"], 1)
+        if exit_code == 0:
+            click.echo(
+                "Fail Open is configured for this repository on semgrep.dev so exiting with code 0",
+                err=True,
+            )
+        sys.exit(exit_code)
+
+    # Set config to rules if default of (,) or is --config policy
+    if not kwargs.get("config") or kwargs.get("config") == ("policy",):
```
underyx: I don't think `semgrep app_scan --config=p/ci` should be supported. This currently errors out in the agent: https://github.com/returntocorp/semgrep-action/blob/678eff1a4269ed04b76631771688c8be860ec4e9/src/semgrep_agent/main.py#L299-L306

semgrep/semgrep/semgrep_app.py
```diff
@@ -0,0 +1,220 @@
...
+            lambda match: match.extra.get("is_ignored"),
+            all_matches,
+        )
+
+        # if self.scan.autofix:
+        #     fields_to_omit.remove("fixed_lines")
+
+        findings = {
+            # send a backup token in case the app is not available
+            "token": os.getenv("GITHUB_TOKEN"),
```
underyx, line 161:
```suggestion
```
I'm gonna assume this is not useful anymore since we haven't recommended exposing the GITHUB_TOKEN for 1+ year.

semgrep/semgrep/commands/app_scan.py
```diff
@@ -0,0 +1,122 @@
...
+        kwargs["config"] = (scan_handler.scan_rules_url,)
+
+    # Append ignores configured on semgrep.dev
+    kwargs["exclude"] = (*kwargs["exclude"], *scan_handler.ignore_patterns)
```
underyx:
```suggestion
    kwargs["exclude"] = (*kwargs["exclude"], *yield_exclude_args(scan_handler.ignore_patterns))
```

Updated cause I think this missing the default tests/ exclusion.

Will also need the update to `yield_exclude_args` I suggested above.

semgrep/semgrep/commands/app_scan.py
```diff
@@ -0,0 +1,122 @@
...
+        yield pattern
+
+
+def yield_exclude_args(requested_patterns: Sequence[str]) -> Iterable[str]:
+    patterns = [*yield_valid_patterns(requested_patterns), *ALWAYS_EXCLUDE_PATTERNS]
+    if Path(IGNORE_FILE_NAME).is_file():
+        patterns.extend(DEFAULT_EXCLUDE_PATTERNS)
+
+    for pattern in patterns:
+        yield from ["--exclude", pattern]
```
underyx:
```suggestion
    yield from patterns
```

even better if this is updated with a new name and to return a Sequence.

underyx (changes requested):
Requesting changes as commented above.

semgrep/semgrep/commands/app_scan.py
brendongo: semgrep-agent does support it as long as APP_TOKEN is not defined. So you can use policies or you can use rulesets

underyx: Right, but in Semgrep CLI that's just a regular `semgrep scan --config foo --baseline-commit bar`, no?

I think these two are equivalent:
1. an agent scan with an app token and a --config value set
2. a `semgrep app_scan` with a --config value set

And because [the 1. case errors out in agent](https://github.com/returntocorp/semgrep-action/blob/678eff1a4269ed04b76631771688c8be860ec4e9/src/semgrep_agent/main.py#L299-L306), I think the 2. case should error out in CLI.

brendongo: Oh I see. yes I agree

semgrep/semgrep/commands/app_scan.py
brendongo: Placeholder for now while I work on https://github.com/returntocorp/semgrep/pull/4820

semgrep/semgrep/commands/app_scan.py
brendongo: Updated

semgrep/semgrep/commands/app_scan.py
brendongo: Done

semgrep/semgrep/commands/app_scan.py
brendongo: Oh lol these two functions aren't even being used :D

brendongo: will remove

brendongo: Oh they should be used. Will clean up then

semgrep/semgrep/meta.py
```diff
@@ -0,0 +1,416 @@
+import json
```
brendongo, line 1: Largely copied from semgrep-agent except I moved away from gitpython https://github.com/returntocorp/semgrep/commit/b9da079cb19bcda828e6cde7224f2bcfdb0883f7

semgrep/semgrep/commands/ci.py
```diff
@@ -132,14 +132,22 @@ def fix_head_if_github_action(metadata: GitMeta) -> Iterator[None]:
...
-    envvar=["SEMGREP_AUDIT_ON"],
+    envvar="SEMGREP_AUDIT_ON",
     multiple=True,
     type=str,
     hidden=True,
 )
+@click.option(
+    "--app-url",
+    default="https://semgrep.dev",
+    envvar="SEMGREP_APP_URL",
```
brendongo, line 144: Can we have semgrep-agent handle the translation between this ENV var and SEMGREP_URL env var that semgrep expects?

semgrep/semgrep/commands/ci.py
```diff
@@ -183,21 +205,31 @@ def ci(
...
-    metric_manager.configure(MetricsState.ON, MetricsState.ON)
+    metric_manager.configure(metrics, metrics_legacy)
+    scan_handler = None
 
-    # Check that we have valid api token
     token = Authentication.get_token()
-    if not token:
+    if not token and not config:
+        # Not logged in and no explicit config
         logger.info("run `semgrep login` before using `semgrep ci`")
```
underyx: `" or set --config manually"`

semgrep/semgrep/commands/ci.py
```diff
@@ -183,21 +205,31 @@ def ci(
...
         sys.exit(INVALID_API_KEY_EXIT_CODE)
-    if not Authentication.is_valid_token(token):
+    elif not token and config:
+        # Not logged in but has explicit config
+        logger.info(f"Running `semgrep ci` without API token but with configs {config}")
+    elif token and config:
+        # Logged in but has explicit config
         logger.info(
-            "API token not valid. Try to run `semgrep logout` and `semgrep login` again.",
+            "Cannot run `semgrep ci` while logged in and with explicit config. Use semgrep.dev to configure policy."
```
underyx: `Use semgrep.dev to configure rules to run`

underyx (approved):

PR #4787, opened 2022-03-10
When matching a class, also try its parents

emjin:
Test plan: see https://github.com/returntocorp/deep-semgrep/pull/22

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -374,6 +374,21 @@ let rec m_name a b =
       m_name a (B.Id (idb, B.empty_id_info ()))
       >||> (* try this time a match with the resolved entity *)
       m_name a (H.name_of_ids dotted)
+      >||>
+      (* Try the parents *)
+      let parents =
+        match !hook_find_possible_parents with
+        | None -> []
+        | Some f -> f dotted
+      in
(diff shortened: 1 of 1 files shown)


semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -374,6 +374,20 @@ let rec m_name a b =
       m_name a (B.Id (idb, B.empty_id_info ()))
       >||> (* try this time a match with the resolved entity *)
       m_name a (H.name_of_ids dotted)
+      >||>
```
aryx, line 377: This may introduce a slowdown when we do this in m_name. Maybe we should do it only when we know we are in a type context where subtyping can be used (new Foo, raise Bar, class X extends Bla, etc.).

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -374,6 +374,20 @@ let rec m_name a b =
...
+      (* Try the parents *)
```
aryx, line 378: next time please add a (* deep: *) comment so we know all the places relevant to deep semgrep.
(It's usually the same that the places where we use !hook_xxx, but still, I like tagging those places).

aryx (approved):

PR #4792, opened 2022-03-11
AST_to_IL: Translate "expression blocks" (StmtExpr)

IagoAbal:
This should help running data-flow analysis on e.g. Scala.

Closes #PA-947

test plan:
make test # test included

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -624,6 +624,13 @@ and expr_aux env ?(void = false) e_gen =
   | G.DeepEllipsis _
   | G.DotAccessEllipsis _ ->
       sgrep_construct (G.E e_gen)
+  | G.StmtExpr { s = G.Block (_, block, _); _ } -> (
+      (* See 'AST_generic.stmt_to_expr' *)
+      match List.rev block with
+      | { s = G.ExprStmt (e, _); _ } :: rev_sts ->
+          rev_sts |> List.rev |> List.concat_map (stmt env) |> add_stmts env;
+          expr env e
+      | __else__ -> todo (G.E e_gen))
   | G.StmtExpr _st -> todo (G.E e_gen)
   | G.OtherExpr ((str, tok), xs) ->
       let es =
@@ -740,8 +747,7 @@ and record env ((_tok, origfields, _) as record_def) =
 (*****************************************************************************)
 (* Exprs and instrs *)
 (*****************************************************************************)
-
-let lval_of_ent env ent =
+and lval_of_ent env ent =
   match ent.G.name with
   | G.EN (G.Id (id, idinfo)) -> lval_of_id_info env id idinfo
   | G.EN name -> lval env (G.N name |> G.e)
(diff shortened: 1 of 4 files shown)


semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -761,20 +767,21 @@ let lval_of_ent env ent =
       | [] -> raise Impossible
       | x :: _ -> fresh_lval env x)
 
-(* just to ensure the code after does not call expr directly *)
-let expr_orig = expr
+(* just to ensure the code after does not call expr directly
+ * TODO: Remove this if we can't figure out why it was done in the first place... *)
```
aryx: I think the idea is that I wanted to prevent the code below to call expr directly; when you process stmts, you should
call expr_with_pre_stmts, not expr, which is why I kinda hide 'expr' by changing its type, so each use of
expr_orig below should be suspect.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -761,20 +767,21 @@ let lval_of_ent env ent =
...
+and expr_orig env ?void e = expr env ?void e
 
-let expr () = ()
+and dummy_expr () = ()
 
-let expr_with_pre_stmts env ?void e =
-  ignore (expr ());
+and expr_with_pre_stmts env ?void e =
```
aryx, line 770: Why we suddenly need to make all those functions mutually recursive?

semgrep-core/tests/OTHER/rules/df_scala_expr_block.scala
```diff
@@ -0,0 +1,27 @@
...
+
+@Singleton
+class AuthController @Inject()(val controllerComponents: ControllerComponents) extends BaseController{
+  
+  def test1(input: String) = Action { implicit request: Request[AnyContent] =>
+    // ruleid: test
+    Ok(input)
+  }
+
+  def test2(input: String) = Action {
```
aryx, line 17: Those Action { <stuff>; Ok(input) }  are represented in the generic AST as a StmtExpr (Block) ?
Where is Action in the AST then?

Those blocks are actually lambdas without any parameter I think in Scala.

aryx (approved):

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: Before we didn't account for `StmtExpr` so `expr` did not depend on `stmt`. But now `expr` and `stmt` are mutually recursive, so all those utility functions that are used by `stmt_aux` need to be part of the same `let-rec`.

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: Aha! I have the feeling that I knew but I had forgotten, thanks for pointing it out. Now being `expr` and `stmt` mutually recursive it would be more cumbersome to ensure that, but I will add a comment at least.

semgrep-core/tests/OTHER/rules/df_scala_expr_block.scala
IagoAbal: Yep indeed, they are represented as:

```
Call(
    N(Id(("Action", ...), ...)),
    [Arg(
        StmtExpr(
        Block(
            [ExprStmt(
            Call(
                N(
                Id(("Ok", ...), ...)),
                [Arg(
                    N(
                    Id(("input", ...), ...
                        )))]), ())])))])
```

I had a similar thought, that it could be fixed also by changing the translation. But I was not sure about the actual semantics in Scala, and anyways we had to handle `StmtExpr` at some point.

aryx: ok sounds good!

IagoAbal:
Tests pass after applying this patch (see returntocorp/semgrep-rules#1886):
```diff
diff --git a/semgrep/tests/e2e/test_semgrep_rules_repo.py b/semgrep/tests/e2e/test_semgrep_rules_repo.py
index 08444f1e9b3..b3af1a960dc 100644
--- a/semgrep/tests/e2e/test_semgrep_rules_repo.py
+++ b/semgrep/tests/e2e/test_semgrep_rules_repo.py
@@ -19,7 +19,7 @@ def _fail_subprocess_on_error(cmd):
 
 def test_semgrep_rules_repo(run_semgrep_in_tmp):
     subprocess.check_output(
-        ["git", "clone", "--depth=1", "https://github.com/returntocorp/semgrep-rules"]
+        ["git", "clone", "--depth=1", "--branch=iago/semgrep-4792", "https://github.com/returntocorp/semgrep-rules"]
     )
 
     # Remove subdir that doesnt contain rules
```

PR #4793, opened 2022-03-11
semgrep/setup.py: exclude tests from packaging

### LINKED ISSUE #2633: pypi dist installs a package called `tests`
**Describe the bug**

After I `pip install semgrep`, I have a package called `tests` in my virtualenv. 

I'm aware of the debate about whether or not to include tests in your sdist, and take no particular view on that, but if you're going to do it, please don't put them in the top level of the dist. I feel like cluttering the package namespace in this way is incorrect.

**To Reproduce**

```
$ pip install semgrep
$ python -c "import tests.conftest"

  File "<string>", line 1, in <module>
  File "/home/rav/work/synapse/env3/lib/python3.8/site-packages/tests/conftest.py", line 13, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

**Expected behavior**

semgrep's tests should not be installed in the root of the package namespace.

P2

priv-kweihmann:
Fixes #2633

Signed-off-by: Konrad Weihmann <kweihmann@outlook.com>

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Changed
+
+- Removed `tests` from published python wheel
+
 ## [0.85.0](https://github.com/returntocorp/semgrep/releases/tag/v0.85.0) - 2022-03-16
 
 ### Added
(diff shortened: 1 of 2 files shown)


ievans:
thanks for this contribution! @brendongo can you review?

brendongo:
I think running `SEMGREP_SKIP_BIN=1 python3 setup.py bdist_wheel` still generates a wheel with the tests folder for me

priv-kweihmann:
@brendongo you're right :disappointed: - Didn't saw that the tests actually land in a subpackage - latest commit you address that

brendongo:
Would `*/tests/*` work? Wondering how to avoid a file with the name test in it being ignored. Or is this only for top level package names and not files?

brendongo:
On second thought seems fine as is. Thanks for the contribution @priv-kweihmann

brendongo (approved):

PR #4796, opened 2022-03-11
Do not cut (use >!>) when matching deep expressions

aryx:
This closes PA-953
This probably introduces some important perf regressions.

test plan:
test file included

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -331,14 +331,17 @@ let m_deep (deep_fun : G.expr Matching_generic.matcher)
...
     ~else_:
-      ( first_fun a b >!> fun () ->
-        (* less: could use a fold *)
-        let rec aux xs =
-          match xs with
-          | [] -> fail ()
-          | x :: xs -> deep_fun a x >||> aux xs
-        in
-        b |> sub_fun |> aux )
+      (* bugfix: this used to be a >!>, but this does not work!
```
aryx: I need to add more comments. I just put the PR out so people can look at it and evaluate the perf regressions.

semgrep/tests/e2e/rules/long.yaml
```diff
@@ -7,10 +7,6 @@ rules:
       ...
       <... $B ...>
       ...
```
aryx, line 9: This is what we can handle now ... just 2 deep-expr with a metavariable. This is because before the $B itself was matching greedily, but we actually need to match all possible subexpressions.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 
+- Fixed Deep expression matching and metavariables interaction. Semgrep will
+  not stop anymore at the first match and will enumarate all possible matchings
+  if a metavariable is used in a deep expression pattern
+  (e.g., `<... $X ...>`). This can introduce some performance regressions.
```
IagoAbal, line 17: If the old behavior is preferred by some we could always make it optional.

aryx: true, but I think the old behavior is just wrong. If $X was bounded before, then it's the same, because the first match will fail and force to go deeper, but otherwise it will not explore all possibilities.

PR #4802, opened 2022-03-14
taint-mode: Enable deep taint analysis

IagoAbal:
This makes possible to implement deep (i.e., interfile interprocedural)
taint analysis. But it does *NOT* implement it! Deep taint-mode is a
feature reserved for Deep Semgrep.

Helps PA-899

test plan:
make test

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Match_tainting_rules.mli
+++ b/semgrep-core/src/engine/Match_tainting_rules.mli
@@ -6,15 +6,7 @@ val check_rule :
   Xtarget.t ->
   Report.rule_profiling Report.match_result
 
-(* Deep Semgrep *)
-val check_def_with_rules :
-  Lang.t ->
-  Dataflow_tainting.config list ->
-  string ->
-  AST_generic.function_definition ->
-  unit
-
-(* Deep Semgrep *)
+(* It could be a private function, but it is also used by Deep Semgrep *)
 val taint_config_of_rule :
   Config_semgrep_t.t ->
   Equivalence.equivalences ->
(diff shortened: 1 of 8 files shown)


aryx:
CI failures @IagoAbal

aryx (approved):

aryx:
at some point we need to move those deep-semgrep functions in the other repository.

semgrep-core/src/engine/Match_tainting_rules.mli
```diff
@@ -5,3 +5,25 @@ val check_rule :
   Rule.taint_spec ->
   Xtarget.t ->
   Report.rule_profiling Report.match_result
+
+(* Deep Semgrep *)
```
aryx: Can this be moved in the other repo?

semgrep-core/src/engine/Match_tainting_rules.mli
```diff
@@ -5,3 +5,25 @@ val check_rule :
...
+val check_def_with_rules :
+  Lang.t ->
+  Dataflow_tainting.config list ->
+  string ->
+  AST_generic.function_definition ->
+  unit
+
+(* Deep Semgrep *)
+val taint_config_of_rule :
```
aryx, line 10: Same? Can this be moved in the other repo?

IagoAbal: This one no because it is needed by Semgrep. Before it was an internal function, and now it's exported to be used by Deep Semgrep, I'll clarify that.

semgrep-core/src/engine/Match_tainting_rules.mli
IagoAbal: Good point, I think that it can be removed from here indeed. I'll try doing that.

Issue #4693, opened 2022-02-17
Semgrep throws an error with a missing separator


DrewDennison:
**Describe the bug**

```
{
      "code": 2,
      "level": "error",
      "message": "Error while running rules: Separator is not found, and chunk exceed the limit",
      "type": "SemgrepError"
}
```

**To Reproduce**
Reporting on behalf of a community user: https://semgrep.dev: https://r2c-community.slack.com/archives/C0242656NQ2/p1645089315876449

**Expected behavior**
Semgrep should produce an understandable error message to the user

**Screenshots**

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?

**Use case**
What will fixing this bug enable for you?

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-886/semgrep-throws-an-error-with-a-missing-separator.

agrawalsmart7:
Anyupdates on this? I am facing the same error.

emjin:
We're currently trying to reproduce this error. If you could provide a reproduction on code you can share with us (an open source repo, your own code you're comfortable sending directly) that would be a big help!

aryx:
@emjin  someone gave me the last.log showing the problem but I didn't yet have the time to investigate

aryx:
the error string is not in our repo, so it’s probably a python builtin error

[yoann (r2c)](https://app.slack.com/team/UL1MB92F2)  [7 days ago](https://r2c-community.slack.com/archives/C0242656NQ2/p1646651745456959?thread_ts=1646632804.900049&cid=C0242656NQ2)

https://stackoverflow.com/questions/55457370/how-to-avoid-valueerror-separator-is-not-found-and-chunk-exceed-the-limit

aryx:
I've looked at the log reported by the user, but I don't see anything from semgrep-core in it. The problem is that
semgrep-python does not log and show stderr as soon as it arrives.

brendongo:
User on slack confirmed that above commit fixed their issue.

PR #4804, opened 2022-03-14
debug(cli): log stderr from semgrep-core ASAP (POSTPONED AFTER 0.85)

aryx:
semgrep was waiting for semgrep-core to finish to print its stderr,
but in case of weird problems (e.g.,
https://github.com/returntocorp/semgrep/issues/4693 ) it was too late.

test plan:
add a Unix.sleep 10 in Main.ml and run
`semgrep -e '$X == $X' -l js misc_bang.rs --debug`
and see that the logging is displayed ASAP (not after 10s)

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -195,7 +195,9 @@ class StreamingSemgrepCore:
                 break
 
             line = line_bytes.decode("utf-8")
-            # log stderr ASAP
+            # log stderr ASAP. This seems to introduce some perf
+            # regressions when using --debug which went from 10% slowdown
+            # to 30%, but it's more important to log ASAP!
             logger.debug(line.rstrip("\n"))
             stderr_lines.append(line)
 
--- a/semgrep/semgrep/core_runner.py
+++ b/semgrep/semgrep/core_runner.py
@@ -182,16 +182,21 @@ class StreamingSemgrepCore:
         if stream is None:
             raise RuntimeError("subprocess was created without a stream")
 
+        logger.debug("--- semgrep-core stderr ---")
+
         while True:
             # blocking read if buffer doesnt contain any lines or EOF
             line_bytes = await stream.readline()
 
             # readline returns empty when EOF
             if not line_bytes:
+                logger.debug("--- end semgrep-core stderr ---")
                 self._stderr = "".join(stderr_lines)
                 break
 
             line = line_bytes.decode("utf-8")
+            # log stderr ASAP
+            logger.debug(line.rstrip("\n"))
             stderr_lines.append(line)
 
     async def _stream_subprocess(self) -> int:
@@ -277,18 +282,6 @@ class CoreRunner:
             core_stderr = (
                 "<semgrep-core stderr not captured, should be printed above>\n"
             )
-
-        # By default, we print semgrep-core's error output, which includes
-        # semgrep-core's logging if it was requested via --debug.
-        #
-        # If semgrep-core prints anything on stderr when running with default
-        # flags, it's a bug that should be fixed in semgrep-core.
-        #
-        logger.debug(
-            f"--- semgrep-core stderr ---\n"
-            f"{core_stderr}"
-            f"--- end semgrep-core stderr ---"
-        )
 
         if returncode != 0:
             output_json = self._parse_core_output(
--- a/semgrep/tests/e2e/test_performance.py
+++ b/semgrep/tests/e2e/test_performance.py
@@ -25,7 +25,7 @@ def test_debug_performance(run_semgrep_in_tmp):
     from pytest import approx
 
     # this used to be 0.1 for 10% but logging stderr ASAP in
-    # core_runner.py can slow down things up to 30%
+    # core_runner.py seems to slow down things a lot
     assert time_with_debug == approx(
-        time_without_debug, rel=0.3
-    ), "adding --debug slowed runtime by more than 30%"
+        time_without_debug, rel=0.5
+    ), "adding --debug slowed runtime by more than 50%"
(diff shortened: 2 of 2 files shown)


semgrep/semgrep/core_runner.py
```diff
@@ -182,16 +182,21 @@ async def _core_stderr_processor(
...
             # readline returns empty when EOF
             if not line_bytes:
+                logger.debug("--- end semgrep-core stderr ---")
                 self._stderr = "".join(stderr_lines)
                 break
 
             line = line_bytes.decode("utf-8")
+            # log stderr ASAP
+            logger.debug(line.rstrip("\n"))
             stderr_lines.append(line)
```
IagoAbal, line 202: Do we still need to build this potentially huge list of `stderr_lines`?

aryx: Probably not. But this gives the option to _extract_core_output() to do something else with those stderr.
I'm ok with removing it. Brendon knows better.

aryx: Actually CI fails because of FAILED tests/e2e/test_performance.py::test_debug_performance - AssertionError...
Not sure why it's slower now.

aryx: or maybe because the snapshot needs to be updated. the order of the logging may have changed.

IagoAbal: If we need to do something with the log later then perhaps we should be writing it to a temporary file instead of keeping it in memory.

brendongo: Reran tests to see if regression was a blip

brendongo: I suspect logger might be causing this to run synchronously or something, causing a perf regression?

aryx: weird, the github-actions below says there's a potential speedup instead ...

semgrep/tests/e2e/test_performance.py
```diff
@@ -24,6 +24,8 @@ def test_debug_performance(run_semgrep_in_tmp):
```
aryx, line 24: BTW @brendongo was it intended to run this python rule on the target file simple.yaml?
There will never be a match there.

brendongo: Yeah it doesn't actually need to run semgrep-core this test is verifying that the debug output doesnt cause a slowdown and since the rule has a really long message it would cause a slowdown if we handle stderr improperly. 

IMO we shouldn't allow the 30% slowdown

semgrep/semgrep/core_runner.py
brendongo: The perf test with the speed up doesn't run with `--debug` which is where this is slowing down

brendongo:
Let's hold off on merging this while semgrep-agent changes are in flight so we don't cause regression here. Once we finish transition I think we wont have to keep --debug performant. But it would still be nice for `--debug` to not have a performance penalty. Will look into ways to have this functionality without regression

semgrep/semgrep/core_runner.py
brendongo: tbh it might just be that logging ~10000 lines just does add 4 seconds :D

aryx:
Ok, but without this it can be annoying to debug certain issues, because the log will not contain anything (e.g., the recent issue about separate and chunk size exceeded).

semgrep/semgrep/core_runner.py
IagoAbal: Printing a lot of stuff to stdout/err does slowdown a program quite significantly.

aryx: The old behavior was also printing stuff on stderr, but it was I guess faster because it was doing it all at once, whereas now we do it as soon as possible (which I prefer for debugging purpose).

IagoAbal: That's right, the old way there is buffering under the hood. What about writing the log to a file? That would buffer the writes, so it would be fast, and we would have logs even if something goes wrong.

aryx: it is already always logging to a file, in ~/.semgrep/last.log. But if you use --debug it also print on stderr
and that's where things are slow.

aryx:
can we merge this @brendongo now that 0.85 was released?

semgrep/semgrep/core_runner.py
IagoAbal: Isn't it enough to get that `last.log` file? Or is there a difference between what you get from `last.log` and from `--debug`?

aryx: the content of last.log and --debug should be the same (well last.log contains all the logs, so despite its name is actually not just the log of the last command)
But in both cases, if something weird happened between semgrep-core and semgrep, neither --debug nor last.log contained stuff logged in semgrep-core.

IagoAbal: Oh I see, of course. 🙂I was thinking that right now, given that we send everything to `semgrep-core`, we probably don't need Semgrep to do anything with the debugging log. Semgrep-core could just write the log directly to a file like `~/.semgrep/core.log`. And Semgrep could always read this file instead of keeping a potentially huge list of log entries in memory.

aryx:
@brendongo ping? Can we merge it?

brendongo:
I'll take a look again next week

brendongo:
I think we don't need this. #4693 was failing on https://github.com/returntocorp/semgrep/pull/4804/files#diff-966f6b0e3edf7d627dd5f5b958cfb2dc72819660ab2414c9e87108ad9447f5dfR197 so this PR would not have printed anything anyway. I think that was a rare case and only failures on that line will cause us to not log semgrep-core stderr. That issue was unrelated to semgrep-core stderr output anyway and would not have helped debug the issue. It's definitely not worth the performance regression. We could curtail the amount of stuff semgrep-core sends to stderr?

PR #4806, opened 2022-03-14
Rework printed plan output

underyx:
PR checklist:

- [x] Changelog is up-to-date


semgrep/tests/e2e/snapshots/test_check/test_max_memory/error.txt
```diff
@@ -36,10 +36,10 @@ Files skipped:
    • <none>
 
 
-skipped: all .gitignored files
-for a detailed list of skipped files, run semgrep with the --verbose flag
+Some files were skipped.
+  Scan was limited to files tracked by git.
```
ievans, line 40: To me this is notably worse than the previous message, which explained how I could affect the outcome (edit the gitignore). My suggestion:

```
Skipped some files:
* skipped files not tracked by git (using .gitignore)
```

semgrep/tests/e2e/snapshots/test_check/test_max_memory/error.txt
```diff
@@ -36,10 +36,10 @@ Files skipped:
...
 
-ran 1 rules on 1 files: 0 findings
+Ran 1 rule on 1 file: 0 findings.
```
ievans, line 42: Why the period? :P

semgrep/tests/e2e/snapshots/test_check/test_terminal_output/error.txt
```diff
@@ -4,9 +4,14 @@ Using configs only from local files (like --config=xyz.yml) does not enable metr
 
 More information: https://semgrep.dev/docs/metrics
 [0m
-Running 4 rules...
 
-skipped: all .gitignored files
-for a detailed list of skipped files, run semgrep with the --verbose flag
+Scanning with multiple kinds of rules.
```
ievans:
```suggestion
Scanning across multiple languages:
```
& then no need for a nested "Summary" header IMO?

semgrep/tests/e2e/snapshots/test_check/test_max_memory/error.txt
underyx: Most other lines are full sentences & I like the style consistency.

semgrep/tests/e2e/snapshots/test_baseline/test_all_intersect/baseline_error.txt
```diff
@@ -1,6 +1,13 @@
...
-for a detailed list of skipped files, run semgrep with the --verbose flag
+Switching repository to baseline commit 'baseline-commit'.
+  Will report findings introduced by these commits:
+    * 1c6ece2 second
 
-ran 1 rules on 1 files: 0 findings
+Scanning 1 file.
+
+Some files were skipped.
+  Scan was limited to files changed since baseline commit.
```
ievans, line 11: Why doesn't this also have:
`For a full list of skipped files, run semgrep with the --verbose ...`

underyx: Because when we get the list of files to scan from `git diff`, we save the effort of walking through the filesystem, and thus don't even know what other files are in the directory.

So verbose output couldn't give you any more information here.

semgrep/tests/e2e/snapshots/test_check/test_max_memory/error.txt
underyx: My original issue here was that I couldn't find a simple way to word

> Skipped files that are unchanged since your baseline commit.

"skipped" and "unchanged" is kind of a double negative which never doesn't hurt understanding.

Then I just made the `git ls-files` message the same as the `git diff` message.

brendongo: I think you use `Scan was limited to files changed since baseline commit.` for the git diff message now right? Would be supportive of isaacs suggestion here.

brendongo (approved):
We can cc technical writers to do a pass but imo can be iterative here

semgrep/tests/e2e/snapshots/test_check/test_max_memory/error.txt
underyx: Yeah, and I figured the diff and ls-files messages being similar would be useful, but I might be ok with not keeping that.

I do wanna workshop the message more though.

```
Some files were skipped.
  Skipped files not tracked by git (using .gitignore).
```

- Bit misleading cause we don't actually use .gitignore, we use ls-files
- Other than being misleading about the internal behavior, it can also be incorrect in practice: a gitignored file will still be scanned if it was force-tracked

underyx: so actually I disagree with

> worse than the previous message, which explained how I could affect the outcome (edit the gitignore)

in fact I think it gave you false confidence, because

1. you cannot skip a file by just .gitignore-ing it, you also need to untrack it
2. you cannot unskip a file by removing the .gitignore entry, you also need to track it

semgrep/semgrep/core_runner.py
```diff
@@ -249,6 +253,77 @@ def execute(self) -> int:
...
+
+    @property
+    def file_count(self) -> int:
+        return len(self)
+
+    def split_by_lang_label(self) -> Dict[str, "Plan"]:
+        result: Dict[str, Plan] = collections.defaultdict(Plan)
+        for task in self:
+            label = (
+                "<multilang>"
```
underyx, line 276: @dlukeomalley's reasons against doing this:

> we are introducing language into our output that doesn’t match the underlying rules. I wonder if instead we accurately communicate the specified “language” of the rules running, so folks know that they have (perhaps less good) generic rules or are indeed running simple ol’ regex . I’m not strong on this one, just a flag that multilang is new. Perhaps generic would be better renamed to multilang to capture what it does :shrug:

my reasons in favor:

- merging regex/generic can save a line of output; in typical usages such as our common recommendation of `p/ci + p/secrets` this'd often mean 2 lines instead of 3.
- regex and generic will always have the same file count which makes the extra line even less useful
- If we list "python, js, regex", that can be misinterpreted as "we're scanning this many of your regex patterns for correctness"
- `--config=auto` users will likely totally not care nor they should need to know what underlying tech is yielding results

underyx:
Merging this now. If reviewers are still interested, please create follow-up issues for:

- wording git ls-files usage as "skipped files" – @ievans 
- splitting up <multilang> – @dlukeomalley 
- generic format review changes – @adamboros

PR #4809, opened 2022-03-14
Allow passing an explicit lockfile target via environment variable

mmcqd:
This PR adds support for passing an explicit lockfile target (either directory to search or file) to Semgrep via the `SEMGREP_LOCKFILE_PATH` environment variable. This is intended as a temporary workaround for using `project-depends-on` rules in repos with many lockfiles, and it should be removed once we have automatic logic to handle this.

PR checklist:

- [x] Changelog is up-to-date


semgrep/dependencyparser/find_lockfiles.py
```diff
@@ -13,16 +13,21 @@
 
 @functools.lru_cache(maxsize=None)
```
underyx: **question**: can generators be cached?
I know it's not part of this PR, but does function caching actually work on generator functions? Has anyone called this function twice with the same parameters?

semgrep/semgrep/dependency_aware_rule.py
```diff
@@ -65,12 +66,18 @@ def run_dependency_aware_rule(
...
     top_level_target: Path = (
         targets[0] if len(top_level_target_rooted) == 0 else top_level_target_rooted[-1]
     )
-    # TODO fix this; run on the top-level of all the targets
+
+    lockfile_target_opt = getenv("SEMGREP_LOCKFILE_PATH")
+    if not lockfile_target_opt:
+        lockfile_target = top_level_target
+    else:
+        lockfile_target = Path(lockfile_target_opt)
```
underyx:
```suggestion
    lockfile_target = Path(getenv("SEMGREP_LOCKFILE_PATH", top_level_target))
```

even for just a week simpler code might help you here 🤷

underyx (approved):

semgrep/dependencyparser/find_lockfiles.py
mmcqd: just checked, and no, generators cannot be cached in the normal way lol. If you call this twice with the same arguments you get the same pointer. I'll just remove it for now, since we only call it once anyway, and I don't think it will ever be called with the same arguments recursively while it runs. If there ends up being a need for it, it's possible to do this using a custom decorator that uses `itertools.tee`.

PR #4811, opened 2022-03-15
fix(async): Set maximum bytes to read from core-stdout

brendongo:
Before this change we were trying to read the entire json blob
returned by semgrep-core in a single go. This was potentially causing
a LimitOverrunError.

This PR modifies the core-stdout processing to instead read in chunks
of `1024*1024*512` bytes

PR checklist:

- [ ] Changelog is up-to-date


brendongo:
Might fix https://github.com/returntocorp/semgrep/issues/4693 but currently unable to reproduce so cannot write test for it

aryx (approved):

Issue #4770, opened 2022-03-08
Allow 'CMD ...' to match any CMD instruction


mjambon:
**Describe the bug**

`CMD ...` right now only matches a CMD with shell code. It should also match `CMD ["foo"]` but doesn't.

**To Reproduce**

Go to this URL and **press Enter to add a trailing newline at the end of the pattern** (bug #4771):
https://semgrep.dev/s/mjambon:docker-cmd-catchall

**Expected behavior**

This should match all the CMD instructions.

**Screenshots**

![image](https://user-images.githubusercontent.com/343265/157340814-a6bed7c9-65d1-491e-93fd-570e30528511.png)

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [x] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**

any

**Use case**

Write dockerfile rules with an explosion of the number of patterns as the number of CMD catch-alls increases. This will avoid writing a pattern with `CMD ...` and another one with `CMD [ ... ]`.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-943/allow-cmd-to-match-any-cmd-instruction.

mjambon:
What we want is:
* `CMD ...` matches any CMD instruction (shell or argv) rather than just shell code
* `CMD ... ;` matches only shell code (already works, see https://semgrep.dev/s/mjambon:docker-cmd-shell). Alternatively, `CMD { ... }` might be a better recommendation (the braces would have to be ignored which isn't the case right now).
* `CMD [ ... ]` matches only argv (already works)

ajbt200128:
@mjambon Checkout the PRs above, fixed this issue, and also fixed a similar issue for `ENTRYPOINT` and `RUN`, and added relevant tests

mjambon:
@ajbt200128 thank you for looking into this, but I don't think it helps. The parser has no difficulty parsing `CMD ...` at the moment because `...` is a valid [semgrep-extended] shell program. I think the question is how it's translated into semgrep's generic AST. We want it to be translated into something like "call the CMD function on one argument that is the Ellipsis". This code is in the semgrep repo, somewhere under `semgrep-core/parsing`. It's getting late now but I'm planning to look into this next week.

ajbt200128:
[...]

PR #4813, opened 2022-03-15
Hack to identify dockerfile commands like 'CMD ...'

mjambon:
Implements https://github.com/returntocorp/semgrep/issues/4770

This doesn't touch the tree-sitter grammar.

It works via a regexp but should eventually use the extended tree-sitter grammar (see https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/293) since it's a little clearer and it's the usual way to go about it. The issue right now is that there have been changes in tree-sitter-dockerfile, so we'd have to upgrade. This can be done later.

I marked the documentation as up-to-date under the assumption that everything is intuitive. It would be good to document that `CMD ...;` can be used to match any shell command and not `CMD ["ls"]`.

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/parsing/ast/AST_dockerfile.ml
```diff
@@ -78,6 +78,7 @@ type string_array = array_elt list bracket
    which changes the shell to an unsupported shell (i.e. not sh or bash).
 *)
 type argv_or_shell =
+  | Ellipsis of tok
```
aryx, line 81: Can you add a (* semgrep-ext: *) comment before so we can quickly know which parts of the AST are semgrep extensions and which parts are original docker constructs?

semgrep-core/tests/dockerfile/cmd-shell.sgrep
```diff
@@ -0,0 +1 @@
+CMD ...;
```
aryx, line 1: do you need the semicolon at the end? CMD ... should also work right?

aryx (approved):

semgrep-core/tests/dockerfile/cmd-shell.sgrep
mjambon: No, because here we want to match only shell commands and specifically not match the argv array syntax such as `CMD ["ls"]`. If we use a plain `CMD ...`, both shell commands and argv arrays will match. This is the novelty of this branch.

mjambon: `CMD ... #` also works (but might change since it requires the comment to be treated as a shell comment rather than a dockerfile comment). I don't want users to think about these subtleties but I think it's a good thing that the most common case is now the most obvious (`CMD ...` matches any `CMD`).

semgrep-core/src/parsing/ast/AST_dockerfile.ml
mjambon: I forgot again. The approach I used elsewhere is to name the constructor `Semgrep_`something.

Issue #2232, opened 2020-12-10
Add support for Swift


cpholguera:
It would be great to have Swift support, along with Kotlin support this can allow for very comprehensive static analysis of mobile apps.

underyx:
I gave you a +1 as thanks for dropping by on our issue tracker :D

traw:
[Tree-sitter-swift](https://github.com/tree-sitter/tree-sitter-swift) might be of some help

fabiopires:
Hi @aryx - not sure if you can give an update on this one. Any thoughts on when this will be added?

Thank you!

aryx:
It's not on our roadmap yet I think. We don't have any customers yet asking for it. Also the tree-sitter-switft above
has not been updated in 2 years, so this is problematic because we need an up-to-date tree-sitter grammar to get good parsing stats.
Note that people can help. In the last year a few external people added support for language like Rust, C#, Hack.

Hupka:
Good morning! 

Recently the Tree-Sitter-Swift-Project has been moved and is now actively maintained in a different place: https://github.com/alex-pinkus/tree-sitter-swift

How much effort would it be to include support for Swift? What would be most of the effort?

aryx:
Most of the effort is writing the Parse_swift_tree_sitter.ml which converts the Concrete syntax tree of Swift as returned by tree-sitter-swift, to the generic AST used internally in semgrep. See for example https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/src/parsing/tree_sitter/Parse_csharp_tree_sitter.ml which is the converter written for C#.

aryx:
Before writing this Parse_swift_tree_sitter.ml, there are lots of small steps to do as documented here: https://semgrep.dev/docs/contributing/adding-a-language/

PR #4817, opened 2022-03-15
Add semgrep-swift submodule

nmote:
I did some copy/paste from the recent Solidity change to get this building.

I found that some of the generated mapper functions were unused. For now, I've just disabled the unused declaration warning in this file, but I intend to follow up on that. Since I'm unfamiliar with the process here, I figured that it might be possible that as I fill in other cases I'll need to use the currently-unused functions. If a reviewer is confident that I won't need these functions in the future, I'll be happy to delete them in this PR and address the TODO.

Partially addresses #2232

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -79,3 +79,6 @@
 [submodule "semgrep-core/src/tree-sitter-lang/semgrep-elixir"]
 	path = semgrep-core/src/tree-sitter-lang/semgrep-elixir
 	url = https://github.com/returntocorp/semgrep-elixir
+[submodule "semgrep-core/src/tree-sitter-lang/semgrep-swift"]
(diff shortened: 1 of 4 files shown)


aryx (approved):

aryx:
Looks good. Yes you can probably remove the unused functions. ocaml-tree-sitter sometimes "inline" the calls to token instead of using the wrapper function (and add a comment about it just before the call to token).

Otherwise, I usually like to have a 'Test plan:' section in the commit message where I explain how I tested the change. Most of the time it can be just "unit tests included; make test". Here it can be just "make", but sometimes it can be the output of semgrep-core -lang swift -dump_ast foo.swift, just to show that the PR does what it's supposed to do.

Issue #4465, opened 2021-12-30
Support XDG Base Directory Specification


itaranto:
Please support the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) on Linux/*nix systems.

Right now `semgrep` puts its config file under `~/.semgrep/`, the correct location should be `$XDG_CONFIG_HOME/semgrep/` with a fallback to `~/.config/semgrep/`.

It doesn't need to be perfect at first, but moving `semgrep` files from the `$HOME` directory would be a good start.  
It's 2021, the standard has been for a while now, and users don't like having junk in their home directories.

Thanks.

brendongo:
I think #4818 handles this

itaranto:
> I think #4818 handles this

This doesn't support the XDG spec properly, to do that, the fallback should be `$HOME/.config/semgrep` not `$HOME/.semgrep`.

@tpetr

itaranto:
Something like this:

```python
if "XDG_CONFIG_HOME" in os.environ and Path(os.environ["XDG_CONFIG_HOME"]).is_dir():
    USER_DATA_FOLDER = Path(os.environ["XDG_CONFIG_HOME"]) / "semgrep"
else:
    USER_DATA_FOLDER = Path.home() / ".config" / "semgrep"
```

tpetr:
Whoops, thanks for catching this. I can make a follow up PR in the near future.

PR #4818, opened 2022-03-15
use XDG_CONFIG_HOME for USER_DATA_FOLDER if set

tpetr:
This PR updates `semgrep` to use `$XDG_CONFIG_HOME` from the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) as its user data folder if the env var is set. (re: https://github.com/returntocorp/semgrep/issues/4465)

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/constants.py
+++ b/semgrep/semgrep/constants.py
@@ -19,7 +19,11 @@ DEFAULT_CONFIG_FOLDER = f".{DEFAULT_SEMGREP_CONFIG_NAME}"
 
 DEFAULT_TIMEOUT = 30  # seconds
 
-USER_DATA_FOLDER = Path.home() / ".semgrep"
+if "XDG_CONFIG_HOME" in os.environ and Path(ox.environ["XDG_CONFIG_HOME"]).is_dir():
+    USER_DATA_FOLDER = Path(os.environ["XDG_CONFIG_HOME"]) / "semgrep"
+else:
+    USER_DATA_FOLDER = Path.home() / ".semgrep"
+
 USER_LOG_FILE = Path(os.environ.get("SEMGREP_LOG_FILE", USER_DATA_FOLDER / "last.log"))
 SETTINGS_FILE = "settings.yml"
 SEMGREP_SETTING_ENVVAR_NAME = "SEMGREP_SETTINGS_FILE"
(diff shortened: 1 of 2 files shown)


semgrep/semgrep/settings.py
```diff
@@ -53,7 +53,7 @@ def __init__(self) -> None:
     @staticmethod
     def get_path_to_settings() -> Path:
         """
-        Uses ~/.semgrep/settings.yaml unless SEMGREP_SETTINGS_FILE is set
+        Uses {$XDG_CONFIG_HOME/semgrep || ~/.semgrep}/settings.yaml unless SEMGREP_SETTINGS_FILE is set
```
tpetr, line 56: I'm not in love with my update to this docstring -- open to feedback here

semgrep/semgrep/constants.py
```diff
@@ -19,7 +19,11 @@
 
 DEFAULT_TIMEOUT = 30  # seconds
 
-USER_DATA_FOLDER = Path.home() / ".semgrep"
+if "XDG_CONFIG_HOME" in os.environ and Path(ox.environ["XDG_CONFIG_HOME"]).is_dir():
```
brendongo:
```suggestion
if "XDG_CONFIG_HOME" in os.environ and Path(os.environ["XDG_CONFIG_HOME"]).is_dir():
```

tpetr: 🤦‍♂️ Thanks!

semgrep/semgrep/settings.py
brendongo: Seems fine :D

brendongo (approved):

brendongo:
Thanks for the contribution @tpetr !

tpetr:
hat tip

tpetr:
@brendongo do you think this change will be an annoyance for Linux users who have a settings.yml file in the old location? I’d be happy to put some additional thought into a migration strategy (doc update or more code changes) if you think it’s worthwhile

brendongo:
I think they will see the metrics notification again but otherwise the functionality that uses the settings.yml file is not widely used yet so I think annoyance will be minimal

itaranto:
This doesn't support the XDG spec properly, to do that, the fallback should be `$HOME/.config/semgrep` not `$HOME/semgrep`.

PR #4820, opened 2022-03-16
Get necessary info from environment

brendongo:
PR checklist:

- [ ] Changelog is up-to-date


brendongo:
https://github.com/returntocorp/semgrep/pull/4820/commits/b9da079cb19bcda828e6cde7224f2bcfdb0883f7 shows diff with semgrep-agent meta.py

semgrep/semgrep/meta.py
```diff
@@ -31,72 +26,109 @@ class GitMeta:
     cli_baseline_ref: Optional[str] = None
     environment: str = field(default="git", init=False)
 
-    @cachedproperty
+    @property
```
brendongo, line 29: None of these variables are called more than once so no need to cache (there are like 2 that are called but they arent slow so w/e)

semgrep/semgrep/meta.py
```diff
@@ -360,16 +407,5 @@ def generate_meta_from_environment(
     elif os.getenv("GITLAB_CI") == "true":
         return GitlabMeta(baseline_ref)
 
-    else:  # nosem
-        if not baseline_ref:
```
brendongo: local env is not used

PR #4824, opened 2022-03-16
Enable the latest cmdliner (>= 1.1.0)

mjambon:
This silences the deprecation alert for code that uses cmdliner in semgrep-core and in the ocaml-tree-sitter-core submodule.

Uses https://github.com/returntocorp/ocaml-tree-sitter-core/pull/28

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/spacegrep.opam
+++ b/semgrep-core/spacegrep.opam
@@ -14,7 +14,7 @@ depends: [
   "alcotest"
   "atdgen"
   "ANSITerminal"
-  "cmdliner" {= "1.0.4"}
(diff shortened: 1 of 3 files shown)


aryx (approved):
Not sure what's going on with the CI failures, but this looks good to me.

mjambon:
One error is because I accidentally upgraded ocamlformat on my machine (when upgrading cmdliner for testing).

The other one is a failed pytest test that succeeds on my machine. Let's what happens after a re-run.

PR #4832, opened 2022-03-17
Add syntactic ID to RuleMatches

underyx:
PR checklist:

- [ ] Changelog is up-to-date


semgrep/Pipfile
```diff
@@ -11,17 +11,16 @@ pytest = "*"
...
 # - semgrep/setup.py lists dependencies
 # - semgrep/Pipfile lists type hint packages for dev env
 # - .pre-commit-config.yaml's mypy hooks also list type hint packages
-
 # type hints
 types-all = "*"
 types-setuptools = "*"
 types-colorama = "~=0.4.0"
 types-jsonschema = "~=3.2"
+semgrep = {path = "."}
```
brendongo: Is this right?

semgrep/semgrep/semgrep_main.py
```diff
@@ -191,33 +191,13 @@ def remove_matches_in_baseline(
...
-        # remove elements in head_matches that appear in baseline_matches
-        # because the above non-uniqueness of id means we need to remove
-        # a match 1:1 (i.e. if match with id X appears 3 times in head but
-        # 2 times in baseline) this function needs to return an object with one
-        # match with id X.
-        for head_match in head_matches:
-            for idx in range(len(baseline_matches)):
-                if head_match.is_baseline_equivalent(baseline_matches[idx]):
-                    baseline_matches.pop(idx)
-                    num_removed += 1
```
brendongo, line 215: num_removed printed below is not updated anymore after code change

underyx: oh great catch! github copilot knew this was missing actually, so I just had to accept its suggestion :D

CleanShot 2022-03-17 at 14 46 55@2x

semgrep/semgrep/rule_match.py
```diff
@@ -51,39 +64,55 @@ def parse(cls, raw_json: JsonObject) -> "CoreLocation":
...
+
+    # None means we didn't check; ignore status is unknown
+    is_ignored: Optional[bool] = field(default=None)
+
+    # derived attributes
+    lines: List[str] = field(init=False, repr=False)
+    lines_hash: str = field(init=False, repr=False)
+    previous_line: str = field(init=False, repr=False)
+    syntactic_context: str = field(init=False, repr=False)
+    cli_unique_key: Tuple = field(init=False, repr=False)
```
brendongo, line 107: Could use docstring on cli_unique_key vs ci_unique_key vs syntactic_id

underyx: 100% agree. Adding now.

semgrep/semgrep/rule_match.py
```diff
@@ -192,19 +160,98 @@ def _get_previous_line(self) -> str:
...
+
+    def add(self, rule_match: RuleMatch) -> None:
+        """
+        Add finding, even if the same (rule, path, code) existed.
+        This is used over regular `.add` to increment the finding's index
+        if it already exists in the set, thereby retaining multiple copies
+        of the same (rule_id, path, line_of_code) tuple.
+        """
+        while rule_match.ci_unique_key in self._seen_ci_keys:
+            rule_match = evolve(rule_match, index=rule_match.index + 1)
```
brendongo: Should we not just keep a mapping from ci_unique_key without the index -> number of things with that key so this operation is constant time instead of O(num collisions)?

underyx: makes sense, this code I copied from semgrep-agent and then added more stuff to it, not realizing there's an easier way like you describe

underyx: I'll just use a collections.Counter

semgrep/semgrep/core_output.py
```diff
@@ -367,24 +345,22 @@ def convert_to_rule_match(match: CoreMatch) -> RuleMatch:
...
+            findings[match.rule].add(rule_match)
 
         # Sort results so as to guarantee the same results across different
         # runs. Results may arrive in a different order due to parallelism
         # (-j option).
-        findings = {
-            rule: order_rule_matches(matches) for rule, matches in findings.items()
-        }
-        return findings
+        ordered_unique_findings: Dict[Rule, OrderedRuleMatchList] = {}
```
brendongo, line 357: Nice

semgrep/semgrep/core_output.py
```diff
@@ -298,30 +300,6 @@ def rule_matches(self, rules: List[Rule]) -> Dict[Rule, List[RuleMatch]]:
 
         For now assumes that all matches encapsulated by this object are from the same rulee
         """
-        # This will remove matches that have the same range but different
```
brendongo, line 301: You should probably keep some of this docstring so it's clear why they need to be in reverse

brendongo (approved):

semgrep/tests/e2e/snapshots/test_output/test_output_format/--gitlab-secrets/results.out
```diff
@@ -10,7 +10,7 @@
       },
       "confidence": "High",
       "cve": "targets/basic/stupid.py:4eb070c4173775f0145392a730b9f22018e5e44638f8d085bd48e7cd18fec7da:rules.eqeq-is-bad",
-      "id": "d96aa79c-6c81-fd1c-7ba1-490c1941ab40",
+      "id": "39bb28b2-c30a-7673-d4de-9be1f0093192",
```
brendongo, line 13: Is this a sufficent test for us that the syntactic_id doesn't change? Might be worth adding a specific test for that?

semgrep/semgrep/core_output.py
underyx: I'm confused by this comment and actually completely ignored it while reimplementing sorting/deduping. The comment makes it sound like it tries to deduplicate findings so that the largest bound metavariable is kept, but the `uniq_id` function actually includes the interpolated messages. So with two findings where one is bound to a "smaller" metavariable and the other is bound to a "larger" metavariable, there is no decision to be made about which one to discard: we keep them both because their messages are different.

semgrep/semgrep/rule_match.py
```diff
@@ -83,17 +83,18 @@ class RuleMatch:
...
 
+    metadata: Dict[str, Any] = field(repr=False, factory=dict)
+    extra: Dict[str, Any] = field(repr=False, factory=dict)
+
     # We call rstrip() for consistency with semgrep-core, which ignores whitespace
     # including newline chars at the end of multiline patterns
-    fix: Optional[str] = field(converter=rstrip)
-    fix_regex: Optional[Dict[str, Any]]
+    fix: Optional[str] = field(converter=rstrip, default=None)
+    fix_regex: Optional[Dict[str, Any]] = None
```
underyx: added defaults to make unit testing easier

PR #4833, opened 2022-03-18
Identify strings made of repeated characters as having low entropy

mjambon:
This should work but I want to add tests to make sure it does. The issue is that the semgrep matcher returns quoted strings e.g. `"xxxxxxxxxxxx"` and the entropy analysis is supposed to be smart and eliminate the quotes. This is necessary for the string to be recognized as a repeated character.

Edit: added tests and fixed implementation accordingly.

PR checklist:

- [x] Tests
- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/tests/OTHER/rules/entropy_python.yaml
+++ b/semgrep-core/tests/OTHER/rules/entropy_python.yaml
@@ -1,0 +1,11 @@
+rules:
+  - id: entropy-python
+    patterns:
+      - pattern: $A = $B
+      - metavariable-analysis:
+          analyzer: entropy
+          metavariable: $B
+    message: found high-entropy string
+    languages:
+      - python
+    severity: ERROR
--- a/semgrep-core/tests/OTHER/rules/redos_python.yaml
+++ b/semgrep-core/tests/OTHER/rules/redos_python.yaml
@@ -1,0 +1,18 @@
+rules:
+  - id: redos-python
+    patterns:
+      - pattern: re.match($PAT, ...)
+      - metavariable-analysis:
+          analyzer: redos
+          metavariable: $PAT
+    message: |
+      Found a regex potentially vulnerable to regex denial-of-service
+      attacks (ReDoS). Consider the use of possessive quantifiers to
+      disable backtracking where it isn't necessary e.g. instead of '[a-z]*',
+      write '[a-z]*+'.
+    metadata:
+      references:
+        - https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS
+    languages:
+      - python
+    severity: ERROR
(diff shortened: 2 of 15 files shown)


semgrep-core/tests/OTHER/rules/entropy_python.yaml
```diff
@@ -0,0 +1,11 @@
+rules:
+  - id: entropy-python
+    patterns:
+      - pattern: $A = $B
```
aryx, line 4: if you use $A = "$B" then in $B you should get the raw content of the string without the enclosing quotes.

semgrep-core/tests/OTHER/rules/redos_python.yaml
```diff
@@ -0,0 +1,18 @@
+rules:
+  - id: redos-python
+    patterns:
+      - pattern: re.match($PAT, ...)
```
aryx, line 4: same here, "$PAT"

aryx, line 4: same here, "$PAT"

semgrep-core/tests/OTHER/rules/entropy_python.yaml
mjambon: I was getting quotes because I was extracting the string data using the wrong function. I simplified it now, it's all good. I left the hack that performs some unescaping, though. It's not great but doing unescaping correctly will require a lot of work. I left a todo note about this in `AST_generic.ml`.

mjambon:
@kurt-r2c if you find a link to the original issue about entropy analysis, we should add it to `CHANGES.md`, otherwise I think it's no big deal. This is what I wrote:
```
- Entropy analysis: strings made of repeated characters such as
  `'xxxxxxxxxxxxxx'` are no longer reported has having high entropy (#4833)
```

mjambon:
mmm, a test fails with generic mode when binding a metavariable to something that looks like an int, which ends up being represented as an `Int` kind of value:
```ocaml
(Metavariable.E
   { e = (L (Int ((Some 123456789012), ()))); e_id = 0; e_range = None })
```
Trying to figure out how this works.

mjambon:
The new behavior is to unquote the captured string value if possible. If it's not a string, use the original source code fragment.

mjambon:
Getting what looks like a network error (twice in a row). The URL being reported (`https://semgrep.dev/c/auto`) works for me from home. I'm going to re-run the job again.

https://github.com/returntocorp/semgrep/runs/5610188578?check_suite_focus=true
```
--- stdout from semgrep process ---
{"errors": [{"code": 2, "level": "error", "message": "Failed to download config from https://semgrep.dev/c/auto: ('Connection aborted.', RemoteDisconnected('Remote end closed\nconnection without response'))", "type": "SemgrepError"}, {"code": 7, "level": "error", "message": "invalid configuration file found (1 configs were invalid)", "type": "SemgrepError"}], "paths": {"_comment": "<add --verbose for a list of skipped paths>", "scanned": []}, "results": []}

--- end semgrep stdout ---
```

kurt-r2c:
@mjambon we were tracking this internally via Linear: RULES-446.  There's no associated GitHub issue.  The statement as-is should be fine.

PR #4838, opened 2022-03-18
Fix symlinks being scanned

### LINKED ISSUE #4827: Symlink handling differs when scanning directories vs. specific files
**Describe the bug**
Semgreps behavior towards symlinked files differs between the directory-scanning and specific-file-scanning mode: A symlinked file can be scanned when scanning a directory it is contained in, but when targeting a scan directly on that file, semgrep claims it does not exist.

**To Reproduce**

Place the following in a file called test.py
```python
def danger2(request):
    # ruleid: command-injection-os-system
    image = request.POST['image']
    os.system("./face-recognize %s --N 24" % image)
```

Run the following command:
```bash
$ ln -s test.py test2.py
```

Scan the entire folder - both files are scanned:
```bash
$ semgrep -c p/ci .
Fetching rules from https://semgrep.dev/registry.
Scanning 2 files with 60 python rules.

Findings:

  test.py 
     python.django.security.injection.command.command-injection-os-system.command-injection-os-
     system
        Request data detected in os.system. This could be vulnerable to a command injection and
        should be avoided. If this must be done, use the 'subprocess' module instead and pass the
        arguments as a list. See https://owasp.org/www-community/attacks/Command_Injection for more
        information.
        Details: https://sg.run/Gen2

          6┆ image = request.POST['image']
          7┆ os.system("./face-recognize %s --N 24" % image)


  test2.py 
     python.django.security.injection.command.command-injection-os-system.command-injection-os-
     system
        Request data detected in os.system. This could be vulnerable to a command injection and
        should be avoided. If this must be done, use the 'subprocess' module instead and pass the
        arguments as a list. See https://owasp.org/www-community/attacks/Command_Injection for more
        information.
        Details: https://sg.run/Gen2

          6┆ image = request.POST['image']
          7┆ os.system("./face-recognize %s --N 24" % image)

Some files were skipped.

Ran 135 rules on 2 files: 2 findings.
```

Scan only the symlink - not found:
```bash
$ semgrep -c p/ci test2.py
Fetching rules from https://semgrep.dev/registry.
File not found: test2.py
```

**Expected behavior**
I expect the behavior to be consistent - either symlinks can be scanned, or they can't, but the behavior should not differ based on if I am giving the folder as a parameter, or the exact path to the symlink itself.

- [x] P1: important to fix or quite annoying (because it messes with an integration test)

**Environment**
Mac OS, semgrep 0.85.0, installed via Homebrew.

underyx:
PR checklist:

- [x] Changelog is up-to-date

Closes https://github.com/returntocorp/semgrep/issues/4827

### DIFF
--- a/semgrep/semgrep/target_manager.py
+++ b/semgrep/semgrep/target_manager.py
@@ -393,7 +393,7 @@ class Target:
         return frozenset(
             match
             for match in self.path.glob("**/*")
-            if not match.is_dir() and match.is_file()
+            if match.is_file() and not match.is_symlink()
         )
 
     @lru_cache(maxsize=None)
--- a/semgrep/tests/unit/targeting/test_target_manager.py
+++ b/semgrep/tests/unit/targeting/test_target_manager.py
@@ -57,6 +57,10 @@ def assert_path_sets_equal(a: Collection[Path], b: Collection[Path]):
     """
     Assert that two sets of path contain the same paths
     """
+    for elem in (*a, *b):
+        assert (
+            not elem.is_symlink()
+        ), f"{elem} is a symlink so we cannot determine if it's the same as its counterpart in the other set"
     a_abs = {elem.resolve() for elem in a}
     b_abs = {elem.resolve() for elem in b}
     assert a_abs == b_abs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   - Two findings are now identical after a nosemgrep comment is added
   - Findings are now different if the same code triggered them on different lines
 
+### Fixed
+
+- Symlinks found in directories are skipped from being scanned again.
+  This is a fix for a regression introduced in 0.85.0.
+
 ## [0.85.0](https://github.com/returntocorp/semgrep/releases/tag/v0.85.0) - 2022-03-16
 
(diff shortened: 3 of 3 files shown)


semgrep/semgrep/target_manager.py
```diff
@@ -393,7 +393,7 @@ def files_from_filesystem(self) -> FrozenSet[Path]:
         return frozenset(
             match
             for match in self.path.glob("**/*")
-            if not match.is_dir() and match.is_file()
```
underyx, line 396: The `not is_dir` conditions seems redundant with `is_file`

semgrep/tests/unit/targeting/test_target_manager.py
```diff
@@ -57,6 +57,10 @@ def assert_path_sets_equal(a: Collection[Path], b: Collection[Path]):
     """
     Assert that two sets of path contain the same paths
     """
+    for elem in (*a, *b):
+        assert (
+            not elem.is_symlink()
+        ), f"{elem} is a symlink so we cannot determine if it's the same as its counterpart in the other set"
     a_abs = {elem.resolve() for elem in a}
     b_abs = {elem.resolve() for elem in b}
```
underyx, line 65: `.resolve()` resolves not just relative paths to absolute, but also follows symlinks. This is why our symlink test case didn't catch the regression in 0.85.

mmcqd (approved):

PR #4845, opened 2022-03-20
What happens if we obey the ocamlformat deprecation

mjambon:
ocamlformat >= 0.20.0 deprecates the use of `break = all`. This results on a bunch of warning in 0.21.0. We could probably live with that but the problem is that the feature will be completely removed in a future release.

This PR is a preview of the reformatting occurring if we use what I think is the closest supported option (`break = fit-or-vertical`). See https://github.com/ocaml-ppx/ocamlformat/issues/1662

PR checklist:

- [x] Changelog is up-to-date


mjambon:
The diff was against the wrong branch. The intended diff is: https://github.com/returntocorp/semgrep/pull/4846

PR #4846, opened 2022-03-20
What happens if we obey the ocamlformat deprecation (2)

mjambon:
ocamlformat >= 0.20.0 deprecates the use of `break = all`. This results on a bunch of warning in 0.21.0. We could probably live with that but the problem is that the feature will be completely removed in a future release.

This PR is a preview of the reformatting occurring if we use what I think is the closest supported option (`break = fit-or-vertical`). See https://github.com/ocaml-ppx/ocamlformat/issues/1662

PR checklist:

- [x] Changelog is up-to-date


aryx:
I'm fine with not breaking all the cases. I like it actually, results in less lines of code.

mjambon:
@IagoAbal are you ok with this formatting style? I don't have a strong opinion. What I most dislike is to have to consult everyone and make those massive reformattings during weekends.

IagoAbal:
> @IagoAbal are you ok with this formatting style? I don't have a strong opinion. What I most dislike is to have to consult everyone and make those massive reformattings during weekends.

I am OK with it! I am happy with any style that keeps code reasonably easy to read. I truly appreciate not having to worry about this any more thanks to ocamlformat, so I am not going to be very picky about it. 😄

PR #4857, opened 2022-03-22
taint-mode: Attach trace to findings

IagoAbal:
Helps PA-314
Helps PA-315

test plan:
make test

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Dataflow_tainting.ml
+++ b/semgrep-core/src/engine/Dataflow_tainting.ml
@@ -51,6 +51,10 @@ end)
 (*****************************************************************************)
 
 type var = Dataflow_core.var
+
+(* TODO: Given that the analysis is path-insensitive, the trace should capture
+ * all potential paths. So a set of tokens seems more appropriate than a list.
+ *)
 type trace = G.tok list [@@deriving show]
 
 type deep_match = PM of Pattern_match.t | Call of G.expr * trace * deep_match
@@ -60,11 +64,19 @@ type source = deep_match [@@deriving show]
 type sink = deep_match [@@deriving show]
 type arg_pos = int [@@deriving show]
 
+type source_to_sink = {
+  source : source;
+  trace : trace;
+  sink : sink;
+  merged_env : Metavariable.bindings;
+}
+[@@deriving show]
+
 type finding =
-  | SrcToSink of source * trace * sink * Metavariable.bindings
-  | SrcToReturn of source * trace
+  | SrcToSink of source_to_sink
+  | SrcToReturn of source * trace * tok
   | ArgToSink of arg_pos * trace * sink
-  | ArgToReturn of arg_pos * trace
+  | ArgToReturn of arg_pos * trace * tok
 [@@deriving show]
 
 type taint_orig = Src of source | Arg of arg_pos [@@deriving show]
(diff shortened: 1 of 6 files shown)


semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -51,19 +51,24 @@ end)
...
+type source = deep_match [@@deriving show]
+type sink = deep_match [@@deriving show]
+type arg_pos = int [@@deriving show]
 
 type finding =
-  | SrcToSink of source * sink * Metavariable.bindings
-  | SrcToReturn of source
-  | ArgToSink of arg_pos * sink
-  | ArgToReturn of arg_pos
+  | SrcToSink of source * trace * sink * Metavariable.bindings
```
emjin: Some of these are pretty big tuples, and they're all different types; might be worth making them records

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -334,12 +341,20 @@ let check_function_signature env fun_exp args_taint =
...
         (fun_sig
         |> List.filter_map (function
-             | SrcToReturn dm ->
-                 let dm = Call (eorig, dm) in
-                 Some (Taint.singleton (Src dm))
-             | ArgToReturn i -> taint_of_arg i
-             | ArgToSink (i, sink) ->
-                 let sink = Call (eorig, sink) in
+             | SrcToReturn (dm, trace) ->
+                 let dm = Call (eorig, trace, dm) in
```
emjin, line 360: Nit: why is this "trace, dm" but SrcToReturn "dm, trace"? If we don't make it a record, would be better to have an order on the args

semgrep-core/src/engine/Match_tainting_rules.ml
```diff
@@ -293,9 +304,10 @@ let check_rule rule match_hook (default_config, equivs) taint_spec xtarget =
...
-    skipped_targets = [];
-    profiling = { RP.rule_id = fst rule.Rule.id; parse_time; match_time };
-  }
+  ( {
+      RP.matches;
+      errors;
+      skipped_targets = [];
+      profiling = { RP.rule_id = fst rule.Rule.id; parse_time; match_time };
+    },
+    debug_taint )
```
emjin, line 313: It's a bit ugly to tack it on at the end like this. How bad would it be to add this as a field in the previous record? In the future we will probably also want to send debugging information for the normal patterns too (what debugging_json used to be) so it would be nice to have this set up well.

emjin (approved):
Exciting! Just left some minor comments, you can ignore them. When you actually add the taint debugging information to the JSON, let's just try to make that code path nice.

semgrep-core/src/engine/Dataflow_tainting.ml
IagoAbal: There is a meaning to it, but I understand it's not clear. In `SrcToReturn` it is the trace _from_ the source `dm` to the `return` (it's missing something to identify the `return`... but I'll add a `tok` to fix that!). In `Call` is the trace from the function in `eorig` _to_  the match `dm`. So I tried to put the `trace` in between the "start" and "end" of the "path".

semgrep-core/src/engine/Dataflow_tainting.ml
IagoAbal: Yeah, maybe the `finding` type needs to be re-designed, but in the meantime I added a record just for the `SrcToSink` constructor. I think it helps more in that case because `source` and `sink` are just type aliases, so there is a real possibility of mixing them up.

semgrep-core/src/engine/Match_tainting_rules.ml
IagoAbal: Totally agree! In this PR I just did the work needed inside the dataflow engine.  There should be another PR that actually carries this info all the way to `semgrep-core`'s output which I think it's a better time to tackle this issue. I left a TODO note.

PR #4861, opened 2022-03-22
const-prop: Handle conditional expressions in Generic

### LINKED ISSUE #4301: FP: string constant inference not correct for non-literal-require
**Describe the bug**
https://semgrep.dev/s/0y95 should not trigger for the third case. Full string inference is hard but this is pretty basic and common and one line. The ? is to decide whether you are in prod/dev mode, e.g..

IagoAbal:
This is a quick-fix. The constant value of the conditional expression is
known during constant propagation, but we only annotate variable
occurences in the Generic AST. Eventually we should add `e_svalue` to
`expr` and annotate all expressions with their constant value.

Closes #4301

test plan:
make test # test included

PR checklist:

- [x] Changelog is up-to-date


mmcqd:
CI failure seems unrelated to this change?

mmcqd (approved):

PR #4870, opened 2022-03-23
Add support for reviewdog output format

mdeous:
This PR adds support for [Reviewdog](https://github.com/reviewdog/reviewdog) diagnostic format ([spec](https://github.com/reviewdog/reviewdog/tree/master/proto/rdf)), allowing to pass semgrep output to reviewdog in order to have comments and/or change suggestions made to a pull-request from the CI. 

I've modified the unit tests to support that option, let me know if additional tests should be created.

PR checklist:

- [x] Documentation is up-to-date ([PR](https://github.com/returntocorp/semgrep-docs/pull/476))
- [x] Changelog is up-to-date


brendongo:
Hi Thanks for the contribution @mdeous!

It looks like Reviewdog accepts vim output format as a way to ingest results. Is it possible to use the `--vim` output instead of having this as a special output format?

mdeous:
@brendongo It seems that the vim output doesn't include fixes, so passing it to Reviewdog would allow to submit comments to a PR, but it would be missing the changes suggestions.

brendongo:
Would maybe SARIF work? We're hesitant to add yet another output format. https://github.com/reviewdog/errorformat/pull/66

mdeous:
@brendongo This seems to be an implementation of SARIF as an output format of their `errorformat` lib when used as a CLI tool, but unfortunately it's not in Reviewdog's supported [input formats](https://github.com/reviewdog/reviewdog#input-format), that would have been perfect otherwise. There is a reference to SARIF as an alternative to rdformat [in the rdformat spec](https://github.com/reviewdog/reviewdog/blob/1ccbebf6b17523333dcae9678bac96978018d89a/proto/rdf/README.md#the-static-analysis-results-interchange-format-sarif), but it doesn't look like it's been implemented. Implementing this was the only was we found to make semgrep and reviewdog work together.

I understand if you don't want to add another output format, but I still think it could be a valuable addition as reviewdog is compatible with a bunch of CI platforms. Just let me know what you want to do.

ievans:
@mdeous thanks so much for taking the time to make a PR! I do think our preference would be to re-use the SARIF output, but I also want to be respectful of your contribution. Since many other codescanning tools already support SARIF, that might be the better way to go though. What are your thoughts?

@brendongo maybe for less-common output formats, we should have a --output-type flag or similar. I know we wanted to keep `--json` as a top-level, but tiering here might make sense?

mdeous:
@ievans No worries, I understand your point (and I understand if you don't wish to include another output format, that's the risk of open-source contributions, it's ok!), I would definitely prefer to use SARIF as well as it's clearly more standard than rdformat, but we faced a few issues while trying to use SARIF:
- we found no (maintained) tool that would take a SARIF file and post PR suggestions from it
- it seems that the SARIF outputed by semgrep doesn't include the [fixes section](https://docs.oasis-open.org/sarif/sarif/v2.1.0/csprd01/sarif-v2.1.0-csprd01.html#_Ref530139075), making it impossible to use it to generate change suggestions

Another option might be to add support for fixes in semgrep SARIF output, then the problem would remain that there is no tool to make PR suggestions from it so it would require additional development, but maybe it would be a better long-term alternative.

brendongo:
@ievans it's less the flag pollution that I am worried about and more that changes to the base output.py sometimes requires changes to be made in all the formats we support + it's another thing that we have to make sure doesnt break 😄. It might be time to think about how to do something extenion-eque where other output formats can plugin to non-officially supported formatters.

ievans:
@mdeous based on the feedback from @brendongo I think I'm going to close this PR. Very sorry that you put all this work in and it didn't get merged :( Idk if you have ideas on how to make this less likely to happen in the future--one idea creating a github issue proposing the work beforehand, can help optimization so we have the discussion up front. 

Let us know if there's anything else we can support you with on the SARIF path--I've gone ahead and created https://github.com/returntocorp/semgrep/issues/5029 for now!

brendongo:
@mdeous FYI we have added fixes to the sarif output with https://github.com/returntocorp/semgrep/issues/5029

PR #4872, opened 2022-03-23
Update snapshot

emjin:
A false positive fixed by https://github.com/returntocorp/semgrep/pull/4867 was reflected in the benchmark tests. Update them now to include the correction

PR checklist:

- [x] Changelog is up-to-date


aryx (approved):
Those /emma/workspace paths are weird, but so were the /underyx/r2c so ...

emjin:
Yeah the errors field doesn't get compared; we should probably just block it out but not important

PR #4877, opened 2022-03-24
taint-mode: Propagate metavariables correctly

### LINKED ISSUE #4464: In taint mode, variables in `pattern-inside` are not exposed to messages and join mode
**Describe the bug**
In taint mode, variables used in `pattern-inside` are not available to the message and rules in join mode.

**To Reproduce**
https://semgrep.dev/s/bX6W
You can see "VAR: a FUNC: $FUNC" in the match message.

**Expected behavior**
"VAR: a FUNC: doSmth" in the match message.

**Details**
This happens only in taint mode. In search mode, if you rename `pattern-sources` to `pattern-either`, variables in `pattern-inside` *are* exposed to messages, see https://semgrep.dev/s/NwOy: "VAR: a FUNC: doSmth".

This is important for me as this eliminates a major use case of taint + join mode for me.

IagoAbal:
Even when they are bound by a pattern-inside!

Closes #4464

test plan:
make test # test included

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   constant will also be considered constant (#4301)
 - Constant propagation now recognizes operators `++` and `--` as side-effectful
   (#4667)
+- Taint mode now propagates metavariables that are bound by a `pattern-inside`.
+  In particular, these metavariables are now available to the rule message. (#4464)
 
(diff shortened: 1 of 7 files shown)


mjambon (approved):

aryx (approved):

aryx:
I think we got some regressions with this in semgrep-rules.
This was merged without all the checks; probably because it was during the time that github was down and was not
running all CI actions.

PR #4879, opened 2022-03-24
Add dryrun option to semgrep ci

brendongo:
PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/commands/ci.py
+++ b/semgrep/semgrep/commands/ci.py
@@ -164,6 +164,14 @@ def fix_head_if_github_action(metadata: GitMeta) -> Iterator[None]:
     """,
     envvar="SEMGREP_RULES",
 )
+@click.option(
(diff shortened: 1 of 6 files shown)


semgrep/semgrep/commands/ci.py
```diff
@@ -164,6 +164,14 @@ def fix_head_if_github_action(metadata: GitMeta) -> Iterator[None]:
     """,
     envvar="SEMGREP_RULES",
 )
+@click.option(
+    "--dryrun",
```
underyx:
```suggestion
    "--dry-run",
    "dryrun",
```

this would keep consistent with
- `kubectl apply --dry-run`
- `npm install --dry-run`
- `poetry install --dry-run`

I added `"dryrun"` just so you can apply the suggestion without changing the variable name below.

PR #4886, opened 2022-03-25
[AST_generic] refactor New, lift it up from Special

aryx:
New is too important for naming, typing, and other program
analysis to be put inside IdSpecial without easily
accessible the class it applies to.

test plan:
make test

PR checklist:

- [x] Changelog is up-to-date


aryx:
The remaining CI failures in semgrep-rules seem to come from a regression existing already on develop, because
of https://github.com/returntocorp/semgrep/pull/4877

PR #4901, opened 2022-03-28
taint-mode: Stop unifying source-sink metavariables

### LINKED ISSUE #4464: In taint mode, variables in `pattern-inside` are not exposed to messages and join mode

IagoAbal:
At least it will not be the default. Forcing unification of source-sink
metavariables turned out to be confusing even for r2c's security team!
But we never had a real use for it, we just wanted to expose the
metavariables to the rule message, which we can achieve in other ways.
Here we chose to do the union of both metavariable sets, but prefer
the sink ones in case of collision.

And restore most of "taint-mode: Propagate metavariables correctly (#4877)"
We still have to propagate metavariables correctly!

Closes #4464

test plan:
make test # test included

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,28 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Changed
+
+- taint-mode: We no longer force the unification of metavariables between
+  sources and sinks by default. It is not clear that this is the most natural
+  behavior; and we realized that, in fact, it was confusing even for experienced
+  Semgrep users. Instead, each set of metavariables is now considered independent.
+  The metavariables available to the rule message are all metavariables bound by
+  `pattern-sinks`, plus the subset of metavariables bound by `pattern-sources`
+  that do not collide with the ones bound by `pattern-sinks`. We do not expect
+  this change to break many taint rules because source-sink metavariable
+  unification had a bug (see #4464) that prevented metavariables bound by a
+  `pattern-inside` to be unified, thus limiting the usefulness of the feature.
+  Nonetheless, it is still possible to force metavariable unification by setting
+  `taint_unify_mvars: true` in the rule's `options`.
+
 ### Fixed
 
 - `-` is now parsed as a valid identifier in Scala
 - `new $OBJECT(...)` will now work properly as a taint sink (#4858)
 - JS/TS: `...{$X}...` will no longer match `str`
+- taint-mode: Metavariables bound by a `pattern-inside` are now available to the
+  rule message. (#4464)
 
 ## [0.86.5](https://github.com/returntocorp/semgrep/releases/tag/v0.86.5) - 2022-03-28
 
--- a/semgrep-core/src/engine/Dataflow_tainting.ml
+++ b/semgrep-core/src/engine/Dataflow_tainting.ml
@@ -127,6 +127,7 @@ type config = {
   is_source : G.any -> PM.t list;
   is_sink : G.any -> PM.t list;
   is_sanitizer : G.any -> PM.t list;
+  unify_mvars : bool;
   handle_findings :
     var option -> finding list -> Taint.t Dataflow_core.env -> unit;
 }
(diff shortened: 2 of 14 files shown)


IagoAbal:
Note that CI failure is expected until returntocorp/semgrep-rules#1941 is merged.

CHANGELOG.md
```diff
@@ -41,6 +49,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   constant will also be considered constant (#4301)
 - Constant propagation now recognizes operators `++` and `--` as side-effectful
   (#4667)
+- taint-mode: Metavariables bound by a `pattenr-inside` are now available to the
```
nmote:
```suggestion
- taint-mode: Metavariables bound by a `pattern-inside` are now available to the
```

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -184,28 +185,49 @@ let report_findings env findings =
...
 
+let sink_biased_union_mvars source_mvars sink_mvars =
+  let source_mvars' =
+    List.filter
+      (fun (mvar, _) -> not @@ List.mem_assoc mvar sink_mvars)
+      source_mvars
+  in
+  Some (source_mvars' @ sink_mvars)
+
+let merge_source_sink_mvars env =
```
emjin: The names of these functions are a little too compact to serve as descriptions. I think if I didn't already have context, even with the comments already here, I would be a bit confused about what's going on. Would be nice to have a function-level comment stating that we're merging the metavariables bound in the source pattern and the sink pattern.

Also, what happens to metavariables in sanitizers?

semgrep-core/src/engine/Match_tainting_rules.ml
```diff
@@ -159,13 +160,9 @@ let taint_config_of_rule default_config equivs file ast_and_errors
...
                not
-                 (List.exists
-                    (fun rng' ->
-                      rng'.Range_with_metavars.r = rng.Range_with_metavars.r)
-                    sinks_ranges
+                 (List.exists (fun rng' -> rng'.RM.r = rng.RM.r) sinks_ranges
                  || List.exists
-                      (fun rng' ->
-                        rng'.Range_with_metavars.r = rng.Range_with_metavars.r)
+                      (fun rng' -> rng'.RM.r = rng.RM.r)
```
emjin: I know you're just making a small change here, but I think using the alias for `Range_with_metavars` really makes this alphabet soup. Can we use a more descriptive name for `rng`?

emjin (approved):
I kind of think it's going to be confusing either way, but we can try this. Maybe the real move is to forbid users from using the same metavariable in their sources and sinks unless they include an "I know what I'm doing I swear" option 😆

IagoAbal:
> I kind of think it's going to be confusing either way, but we can try this. Maybe the real move is to forbid users from using the same metavariable in their sources and sinks unless they include an "I know what I'm doing I swear" option 😆

If done from the beginning, I agree that it could have been a good approach! It's just a little bit inconvenient having to write `$FUNC1`, `$FUNC2`, and so on... but for sure it leaves no room for confusion.

Another option was to make it work similar-ish to join mode, and have a field to choose what variables to unify. You could have a qualifier to refer to each metavariable, as in `sources.$X`. After all, `pattern-sources` and `pattern-sinks` are two separate queries that are run independently. And this is not just an implementation detail!

I guess `taint_unify_mvars: true` somehow means "I know what I'm doing I swear"? 😄

semgrep-core/src/engine/Match_tainting_rules.ml
IagoAbal: I replaced it with `range`.

semgrep-core/src/engine/Dataflow_tainting.ml
IagoAbal: I believe we were also meant to unify source-sanitizer metavariables, but it seems that we never did. I added a `TODO` comment in a previous PR.

PR #4902, opened 2022-03-28
Convert Swift hello world to generic AST

nmote:
I've done just the bare minimum here to convert a "hello world" program in Swift to Semgrep's generic AST. To get it building with my changes, I had to add some more todos. In places where I was unsure about best practices, I looked to other languages for examples, but I'd still appreciate any pointers if there are better ways to do things.

Test plan:

New automated test (I verified that without these changes it fails), plus the following manual test:

```
$ semgrep-core -lang swift -dump_ast hello-world.swift
Pr(
  [ExprStmt(
     Call(N(Id(("print", ()), {id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); })),
       [Arg(L(String(("hello world", ()))))]), ())])
```

Partially addresses #2232

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
@@ -22,9 +22,8 @@ module H2 = AST_generic_helpers
 
 type env = unit H.env
 
-let token (env : env) (_tok : Tree_sitter_run.Token.t) =
-  failwith "not implemented"
-
+let token = H.token
+let str = H.str
 let blank (env : env) () = failwith "not implemented"
 let todo (env : env) _ = failwith "not implemented"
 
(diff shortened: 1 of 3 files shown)


mmcqd (approved):
Slightly confused about all the added `todo`s, because it seems they're mostly added in places where `todo` was already going to end up being called. I assume they're just there for your own benefit while working? Otherwise looks good! Excited that we can translate "hello world".

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1350,9 +1352,9 @@ and map_directly_assignable_expression (env : env)
   match x with
   | `Simple_id x -> map_simple_identifier env x
   | `Navi_exp x -> map_navigation_expression env x
-  | `Call_exp x -> map_call_expression env x
+  | `Call_exp x -> map_call_expression env x |> todo env
```
mmcqd, line 1355: question: why is `todo` here? It looks like you implemented `map_call_expression` as part of this PR

nmote:
I added the todos to satisfy the typechecker. There were a variety of places where, as you noted, `todo` will end up being called anyway. However, there were places where the addition of type annotations or actual implementations led to type errors that did not previously exist. I don't specifically recall the `todo` call that you asked about inline, but I'm guessing that when I implemented `map_simple_identifier` (called two lines above) and `map_call_expression`, their types did not match and I chose to add a `todo` instead of solving that issue right then.

mmcqd:
Makes sense 👍

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1426,9 +1428,11 @@ and map_expr_hack_at_ternary_call_suffix (env : env)
     (x : CST.expr_hack_at_ternary_call_suffix) =
   map_value_arguments env x
 
-and map_expression (env : env) (x : CST.expression) =
+and map_expression (env : env) (x : CST.expression) : G.expr =
   match x with
-  | `Simple_id x -> map_simple_identifier env x
+  | `Simple_id x ->
+      let id = map_simple_identifier env x in
+      G.N (G.Id (id, G.empty_id_info ())) |> G.e
```
aryx, line 1435: You can use H2.name_of_id
where module H2 = AST_generic_helpers (H is usually used for Parse_tree_sitter_helpers.ml)

aryx (approved):
yeah!

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
nmote: Addressed in #4912

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -3030,28 +3037,22 @@ let map_global_declaration (env : env) (x : CST.global_declaration) =
...
-  let v2 =
-    List.map
-      (fun (v1, v2) ->
-        let v1 = map_top_level_statement env v1 in
-        let v2 = (* semi *) token env v2 in
-        todo env (v1, v2))
-      v2
+let map_source_file (env : env) ((_shebang, stmts) : CST.source_file) : G.any =
+  let stmts =
+    List.map (fun (stmt, semi) -> map_top_level_statement env stmt semi) stmts
```
nbrahms, line 3053: Just FYI I think we've run into memory issues with `List.map`; you may want to `List.rev List.rev_map`

IagoAbal: If this is about stack overflows, we have `Common.map`.

PR #4903, opened 2022-03-28
Remove commented-out semgrepdep action

nmote:
This was commented out in https://github.com/returntocorp/semgrep/commit/6c447ca489bd287c0453b0a61226fc4907bd7197.

Now, every time I put up a pull request, I get an email like this because the workflow file exists without defining any jobs:
image

It seems likely that this can be safely removed since it has not been missed since it was disabled in January.

PR checklist:

- [x] Changelog is up-to-date


emjin (approved):
Let's do it so we don't normalize failure, and ask Nathan about it next standup when he's back

PR #4907, opened 2022-03-28
Scala: translate `new` more simply when possible to enable tainting

### LINKED ISSUE #4858: Scala class constructor calls do not work as taint sinks
**Describe the bug**
Scala class constructor calls do not work as taint sinks

**To Reproduce**
https://semgrep.dev/s/ezzer17:constructor-taint-bug

**Expected behavior**
Match detected on attached playground

P1: important to fix or quite annoying

**Use case**
Taint analysis with class constructors as sinks

mmcqd:
Closes #4858
This changes the translation of `new Obj(args)` in Scala to use `Ast_generic.New` instead of `Ast_generic.AnonClass`. In this simple case where we are just creating a new object without multiple parent classes/mixins and not creating an actual anonymous class, it is simpler to use `New`. This fixes #4858 because `New` is actually analyzed by taint mode, but `AnonClass` is not.

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/parsing/pfff/scala_to_generic.ml
```diff
@@ -410,11 +410,26 @@ and v_expr e : G.expr =
...
       let v1 = v_function_definition v1 in
       G.Lambda v1 |> G.e
-  | New (v1, v2) ->
-      let _v1 = v_tok v1 and v2 = v_template_definition v2 in
-      let cl = G.AnonClass v2 |> G.e in
-      (* TODO: transform in New if v_template_definition is simple *)
-      G.Call (cl, fb []) |> G.e
+  | New (v1, v2) -> (
+      let v1 = v_tok v1 and v2 = v_template_definition v2 in
+      match v2 with
```
aryx, line 415: you could have done also the match on the AST_Scala.v2 instead of v2 after the call to v_template_definition.

semgrep-core/tests/tainting_rules/scala/new_object_sink.yaml
```diff
@@ -0,0 +1,13 @@
+rules:
+  - id: tainted-obj
+    mode: taint
+    pattern-sinks:
+      - patterns:
+          - pattern: new Example(...)
+    pattern-sources:
```
aryx, line 7: I would put the sources before the sinks in the file.

aryx (approved):

PR #4908, opened 2022-03-29
project-depends-on: create dependency only matches even without a pattern match

mmcqd:
This PR changes the behavior of the `r2c-internal-project-depends-on` key in the following ways:
1. Given a rule with both some `pattern` key and an `r2c-internal-project-depends-on` key, if a dependency match is found, but the pattern does _not_ match (the vuln is not reachable), semgrep _will_ report a match, imitating the behavior that would occur if there had been no pattern at all. The fact that this was a dependency match but not a pattern match will be reported, see next point. This behavior has been requested by customers.
2. All matches from rules with the `r2c-internal-project-depends-on` key will include two new keys in the `extra` field of semgrep's JSON output:
    * `dependency_match_only`: will be `true` if the scenario in (1) occurs, and `false` otherwise (there was a pattern match _and_ a dependency match)
    * `dependency_matches`: will be a list of dependency matches of the following form:
        ```
        [
                  {
                    "dependency_pattern": {
                      "namespace": "cargo",
                      "package_name": "adler",
                      "semver_range": "<= 1.0.3"
                    },
                    "found_dependency": {
                      "allowed_hashes": {
                        "sha256": [
                          "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
                        ]
                      },
                      "name": "adler",
                      "namespace": "cargo",
                      "resolved_url": null,
                      "version": "1.0.2"
                    },
                    "lockfile": "cargo.lock"
                  }
         ]
        ```

Ideally the reported file in the case of (1) would be the offending lockfile. This is currently not possible because of the poor lockfile targeting situation, but should be doable soon, once that's implemented.

test plan: make test

A test that previously did not fire now (correctly) fires, and a comment has been updated to reflect this, so this functionality is now tested by existing tests.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/dependencyparser/models.py
+++ b/semgrep/dependencyparser/models.py
@@ -23,7 +23,7 @@ class PackageManagers(str, Enum):
     MAVEN = "maven"
 
 
-@dataclass(eq=True, order=True)
+@dataclass(eq=True, order=True, frozen=True)
 class LockfileDependency:
     name: str
     version: str
--- a/semgrep/dependencyparser/models.py
+++ b/semgrep/dependencyparser/models.py
@@ -1,10 +1,9 @@
+from dataclasses import dataclass
 from enum import Enum
 from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Set
-
-from attrs import frozen
 
 KNOWN_HASH_ALGORITHMS: Set[str] = {
     "sha256",
@@ -15,7 +14,7 @@ KNOWN_HASH_ALGORITHMS: Set[str] = {
 }
 
 
-class PackageManagers(Enum):
+class PackageManagers(str, Enum):
     NPM = "npm"
     PYPI = "pypi"
     GEM = "gem"
@@ -24,7 +23,7 @@ class PackageManagers(Enum):
     MAVEN = "maven"
 
 
-@frozen(eq=True, order=True)
+@dataclass(eq=True, order=True)
 class LockfileDependency:
     name: str
     version: str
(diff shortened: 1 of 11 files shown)


semgrep/dependencyparser/models.py
```diff
@@ -24,7 +23,7 @@ class PackageManagers(Enum):
     MAVEN = "maven"
 
 
-@frozen(eq=True, order=True)
+@dataclass(eq=True, order=True)
```
mmcqd: This `dataclass` change was to make JSON serialization possible without doing it manually

semgrep/dependencyparser/models.py
```diff
@@ -24,7 +23,7 @@ class PackageManagers(Enum):
     MAVEN = "maven"
 
 
-@frozen(eq=True, order=True)
+@dataclass(eq=True, order=True, frozen=True)
```
mmcqd, line 26: This `dataclass` change was to make JSON serialization possible without doing it manually

semgrep/semgrep/dependency_aware_rule.py
```diff
@@ -110,7 +100,33 @@ def run_dependency_aware_rule(
                 depends_on_entries, find_and_parse_lockfiles(lockfile_target)
             )
         )
-        final_matches = [] if len(output) == 0 else matches
+        output_for_json = [
```
aryx, line 103: Would be great to use ATD and atdpy to specify the JSON output formally.

semgrep/dependencyparser/models.py
aryx: Even better would be to use atdpy!
That way this can also be consumed in a type-safe way on the other side (semgrep-app backend)

aryx (approved):

aryx:
I would love the use of atdpy here (but probably better in a different PR).

PR #4909, opened 2022-03-29
Run tests from quick to slow, in parallel

mjambon:
Impact of this PR:
* faster when running locally with `make test`. Doesn't change the duration of CI.
* fails faster on average when running `make test`.

I haven't made any single test faster.

I tagged each pytest test with one of `quick`, `kinda_slow`, or `slow` markers. The makefile for running `make test` will run the tests from quick to slow so as to fail faster on average.

This makefile, which isn't used in CI, wasn't using the `-n` option to run multiple jobs in parallel. It now does, except for the quick tests that end up being slower with this option due to some startup overhead.
* `quick`: 725 tests, 7 seconds
* `kinda_slow`: 255 tests, 5 minutes
* `slow`: 164 tests, 35 minutes

I also added a check to ensure that all the `test_`* functions are categorized.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -144,6 +144,7 @@ jobs:
           unset CI
           unset "${!GITHUB_@}"
 
+          make check-markers
           pipenv run pytest --ignore=tests/qa/test_public_repos.py -n auto -vv --tb=short --durations=0 --snapshot-update tests/
       - name: Generate cheatsheets for semgrep.dev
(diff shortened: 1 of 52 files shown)


mjambon:
> 🚫 Benchmark semgrep.bench.0c34.std is too slow: +34.7% (+0.505 s)

Bad benchmark. Re-running.

aryx (approved):
Nice!

PR #4914, opened 2022-03-29
JS/TS: do not match `...${$X}...` with non-interpolated strings

mmcqd:
Previously the pattern
```
`...${$X}...`
```
would match the expression
```
`str`
```
This PR fixes that.

PR checklist:

- [x] Changelog is up-to-date


emjin (approved):
Nice!

PR #4924, opened 2022-03-30
Specify required packages argument

tino:
The latest setuptools requires this, even though it is empty. Otherwise
it won't install with a `error: Multiple top-level packages discovered
in a flat-layout` exception.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,4 +7,5 @@ setup(
     name="semgrep_pre_commit_package",
     version="0.0.0",
     install_requires=["semgrep==0.86.5"],
+    packages=[],
 )


ievans:
Thanks so much @tino. Surprised our test suite missed this, do you have a reproduction we can look at / can be added to tests?

underyx:
> Thanks so much @tino. Surprised our test suite missed this, do you have a reproduction we can look at / can be added to tests?

@ievans note that I think we're not testing this setup.py, as it's just a dummy file for the pre-commit hook to work. The real setup.py is under semgrep/.

underyx:
I think I see the relevant breaking change here: https://github.com/pypa/setuptools/blob/main/CHANGES.rst#breaking-changes

Looks good, merging!

underyx (approved):

underyx:
Here's a sample failure: https://github.com/returntocorp/semgrep/runs/5762565672?check_suite_focus=true

tino:
```
git clone ...
cd semgrep
python -m venv venv/
source venv/bin/activate
pip install setuptools -U  # should be v 58 or higher atm, I think 57 introcuded it
python setup.py install 
```

The last command will fail without this change.

Edit: commented without reloading :).

Thanks for merging!

brandoncazander:
We just ran into this issue ourselves when using semgrep with pre-commit, but it was only affecting developers that were newly spinning up their development environments so it was a bit hard to track down.

Looking forward for this one to be released, and thanks for the fix!

underyx:
@brandoncazander FYI we'll be releasing a new version of semgrep next week, but until then, we tagged `0.86.5-tmp-fix-pre-commit` so you can already update your configs to that.

brandoncazander:
> @brandoncazander FYI we'll be releasing a new version of semgrep next week, but until then, we tagged `0.86.5-tmp-fix-pre-commit` so you can already update your configs to that.

Thank you!

PR #4927, opened 2022-03-30
Random finding appeared in some agent scans with lots of complex merge commits

underyx:
PR checklist:

- [ ] Changelog is up-to-date


underyx:
needs tests

semgrep/semgrep/git.py
```diff
@@ -108,27 +108,44 @@ def _get_git_status(self) -> GitStatus:
...
+            "--name-status",
+            "--no-ext-diff",
+            "-z",
+            "--diff-filter=ACDMRTUXB",
+            "--ignore-submodules",
+            "--relative",
+            self._base_commit,
+        ]
+        try:
+            # nosemgrep: python.lang.security.audit.dangerous-subprocess-use.dangerous-subprocess-use
```
nbrahms, line 124: @minusworld This looks like a false positive...

semgrep/semgrep/git.py
```diff
@@ -271,34 +288,33 @@ def baseline_context(self) -> Iterator[None]:
 
         Raises CalledProcessError if any calls to git return non-zero exit code
         """
-        status = self.status
+        self.status
```
nbrahms: Is this dead code?

semgrep/semgrep/git.py
underyx: I reported via issue feedback

semgrep/semgrep/git.py
underyx: yep, good catch! removed

semgrep/semgrep/git.py
```diff
@@ -108,27 +108,44 @@ def _get_git_status(self) -> GitStatus:
...
+            raw_output = subprocess.run(
+                [*status_cmd, "--merge-base"],
                 timeout=GIT_SH_TIMEOUT,
                 capture_output=True,
                 encoding="utf-8",
                 check=True,
             ).stdout
-        )
+
+        except subprocess.CalledProcessError as exc:
```
brendongo, line 133: Seems like this would only fix cases when we hit this exception of multiple merge bases but findings appearing would mean that the former command exited with exit code zero right?

brendongo (approved):

PR #4929, opened 2022-03-30
Update pre-commit hooks

underyx:
PR checklist:

- [ ] Changelog is up-to-date


brendongo (approved):
Should we just tag latest release or does pre-commit not work well with mutable tags (for semgrep)

underyx:
> Should we just tag latest release or does pre-commit not work well with mutable tags

not sure, this PR can just wait till the next semgrep release

brendongo:
Gonna update this to pin to a temp tag with the setuptools fix since it wasnt in 0.86.5

PR #4930, opened 2022-03-31
Make it possible to run any command in docker image

underyx:
PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-if [[ "$*" =~ "--config" || "$1" == "publish" || "$1" == "ci" ]]; then
+if [[ "$*" =~ "--config" || "$*" == "--help" || "$1" == "publish" || "$1" == "ci" ]]; then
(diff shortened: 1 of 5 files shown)


entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+set -e
+
+if [[ "$*" =~ "--config" ]] || [[ "$1" == "publish" ]]; then
```
brendongo: what about `ci` and `--help`

brendongo:
Some flavor for this PR: we need to make modifications to the entrypoint of the docker image to allow the image to be used as a base image in CI (github/gitlab/circleci etc) these runners run their own scripts to inject the git project in the right place and assume that the entrypoint for docker images is a shell. Prior to this change our entry point was `semgrep` which was causing issues since essentially the arguments being passed by these CI envs were bash commands. This change has us do some guessing to see if the user assumed a `semgrep` entry point.

brendongo (approved):

PR #4931, opened 2022-03-31
Use atdpy

mjambon:
This requires the latest version of `atdpy` as specified in the opam file used to install dev dependencies. The Python code `output_from_core.py` is generated when calling `make`  (= `make build`) from the project root or when running `make build` from the Python project root (`/semgrep/`).

This branch changes a tiny bit of the Python code that interprets the output from semgrep-core. There's more to do but I want to merge this first. It will get increasingly faster as there will be less and less work to do on atdpy.

PR checklist:

- [x] Changelog is up-to-date


mjambon:
After adding an argument to `black` in the pre-commit config so as to ignore a file, this error happens: https://github.com/returntocorp/semgrep/runs/5764698462?check_suite_focus=true

interfaces/Output_from_core.atd
```diff
@@ -207,7 +225,8 @@ type target_time = {
   run_time: float;
 }
 
-type rule_times = {
+type rule_times
+  <python decorator="dataclass(frozen=True)"> = {
```
aryx, line 229: maybe this dataclass(frozen=True) should be the default in atdpy? so we don't need to specify it each time?

semgrep/semgrep/output_from_core.py
```diff
@@ -283,7 +283,7 @@ def to_json_string(self, **kw: Any) -> str:
         return json.dumps(self.to_json(), **kw)
 
 
-@dataclass
+@dataclass(frozen=True)
```
aryx, line 286: So do we want to commit the generated code in the repository? Simpler for people?

aryx (approved):
Awesome start!

interfaces/Output_from_core.atd
mjambon: Yes, maybe we could do that. Maybe other options would make sense by default as well.

There's the issue of overriding them vs. extending them. The default decorator is `@dataclass` but if the user specifies another `@dataclass(...)` decorator, it replaces the default. Now if the default is no longer a simple `@dataclass`, it gets tricky for the user to do the right thing.

semgrep/semgrep/output_from_core.py
mjambon: I tried removing it but I got pre-commit linter errors, due to an attempt to resolve names and the absence of the generated file that provides those names. Not getting this error would mean running atdpy before running the pre-commit linters, which seemed too complicated.

interfaces/Output_from_core.atd
```diff
@@ -223,10 +242,12 @@ type rule_times = {
 (* Synthesizing from diffs (see locate_patched_functions in Synthesizing.mli) *)
 (*****************************************************************************)
 
-type cve_result = {
+type cve_result
+  <python decorator="dataclass(frozen=True)"> = {
```
aryx, line 246: can we make this the default?

.pre-commit-config.yaml
```diff
@@ -21,6 +21,7 @@ repos:
     rev: 22.3.0
     hooks:
       - id: black
+        args: [--exclude=/semgrep/semgrep/output_from_core.py]
```
aryx, line 24: I don't think this works. I've regenerated a few times this file and it got reformatted by some tools (black I guess).

interfaces/Output_from_core.atd
mjambon: please open an issue: https://github.com/ahrefs/atd/issues

aryx: done: https://github.com/ahrefs/atd/issues/289

PR #4933, opened 2022-03-31
fix: use new endpoint for deployments

brendongo:
Tested locally using `semgrep scan --config policy`

PR checklist:

- [ ] Changelog is up-to-date


brendongo:
Tested locally and want to update develop docker image to merging

PR #4938, opened 2022-04-01
add: use github enterprise url if available

brendongo:
PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   Nonetheless, it is still possible to force metavariable unification by setting
   `taint_unify_mvars: true` in the rule's `options`.
 
+### Added
+
+- `semgrep ci` uses "GITHUB_ENTERPRISE_URL" as repository url if it is available
+
 ### Fixed
 
 - `-` is now parsed as a valid identifier in Scala
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Added
 
-- `semgrep ci` uses "GITHUB_ENTERPRISE_URL" as repository url if it is available
+- `semgrep ci` uses "GITHUB_SERVER_URL" to generate urls if it is available
 
 ### Fixed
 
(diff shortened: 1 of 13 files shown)


semgrep/tests/e2e/snapshots/test_ci/test_full_run/github-enterprise-autofix/meta.json
```diff
@@ -0,0 +1,22 @@
+{
+    "meta": {
+        "semgrep_version": "<sanitized version>",
+        "repository": "project_name/project_name",
+        "repo_url": "https://some.enterprise.url.com",
+        "branch": "some/branch-name",
+        "ci_job_url": "https://some.enterprise.url.com/actions/runs/35",
```
brendongo: @zyannes is this URL in the correct format for a GHE action run?

semgrep/semgrep/meta.py
```diff
@@ -202,7 +202,9 @@ def repo_name(self) -> str:
 
     @property
     def repo_url(self) -> Optional[str]:
-        if self.repo_name:
+        if "GITHUB_ENTERPRISE_URL" in os.environ:
+            return os.getenv("GITHUB_ENTERPRISE_URL")
+        elif self.repo_name:
             return f"https://github.com/{self.repo_name}"
```
zyannes: How about using the `GITHUB_SERVER_URL` environment variable (https://docs.github.com/en/actions/learn-github-actions/environment-variables#about-environment-variables) instead of introducing a new environment variable?
```suggestion
        if self.repo_name:
            github_url = os.getenv("GITHUB_SERVER_URL", "https://github.com")
            return f"{github_url}/{self.repo_name}"
```

brendongo: Sg!

Updated PR

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Added
+
+- `semgrep ci` uses "GITHUB_ENTERPRISE_URL" as repository url if it is available
```
zyannes: Nit: this no longer applies following the latest commit

CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   Nonetheless, it is still possible to force metavariable unification by setting
   `taint_unify_mvars: true` in the rule's `options`.
 
+### Added
+
+- `semgrep ci` uses "GITHUB_SERVER_URL" to generate urls if it is available
```
zyannes, line 30: Nit: this no longer applies following the latest commit

zyannes (approved):

PR #4954, opened 2022-04-05
project-depends-on: associate each lockfile to the files that depend on it

mmcqd:
This PR changes the behavior of the `r2c-internal-project-depends-on` key so that a pattern match + dependency match finding (aka a _reachable_ dependency finding) will only be reported if the file containing the pattern match actually depends on (lol) the lockfile containing the dependency match. The definition of _depends_ here is the following:

`foo.js` depends on `yarn.lock` if `yarn.lock` is the closest `NPM` lockfile to `foo.js` going _only up_ the directory tree.
As an example, given the following directory structure, and assuming `==>` means "depends on"
```
yarn.lock
A.js
folder1/
  B.js
folder2/
  yarn.lock
  C.js
```

`./A.js ==> ./yarn.lock`
`./folder1/B.js ==> ./yarn.lock`
`./folder2/C.js ==> ./folder2/yarn.lock`

Checking these dependencies is accomplished via a trie of lockfile paths, and a method which takes a file path and finds the lockfile that shares the longest common prefix with this path.

This PR also makes some general architectural/performance edits, like only parsing lockfiles associated to the languages of the `project-depends-on` rules that are being run.

Note: having written this, I'm beginning to think this all ought to be in OCaml. It feels a bit weird to be doing so much work (especially parsing) in python, while at the same time sending whole rules to be run by semgrep-core. If we want to add dependency autofix to lockfiles, we'll need to start tracking location info during lockfile parsing, at which point porting the parsers to OCaml seems like the obvious choice.

PR checklist:

- [x] Changelog is up-to-date


semgrep/semgrep/cli.py
```diff
@@ -14,7 +14,6 @@ def cli() -> None:
     """
     To get started quickly, run `semgrep scan --config auto`
     """
-    pass
```
mmcqd, line 17: This seems to be caused by a change in the pre-commit requirements?

brendongo: Fixed on develop so you can rebase if ever

semgrep/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awareansi-html.yaml-dependency_awareansi.js/results.json
```diff
@@ -35,7 +35,7 @@
             "lockfile": "yarn.lock"
           }
         ],
-        "fingerprint": "34dbc50eb98e00ba192ad6f3e9716450",
+        "fingerprint": "e87e6dc13255fff5bd134556c84b5a22",
```
brendongo, line 38: @mmcqd did we expect the fingerprints to change?

semgrep/dependencyparser/find_lockfiles.py
```diff
@@ -1,29 +1,70 @@
...
-            if entry.is_file() and entry.name.lower() in TARGET_LOCKFILE_FILENAMES:
-                yield full_path
+@dataclass
+class Node:
+    children: Dict[str, "Node"]
+    val: Optional[Dict[Path, List[LockfileDependency]]]
+
+
+class DependencyTrie:
+    def __init__(self) -> None:
```
brendongo, line 49: Class level and function level comments for exposed functions appreciated

semgrep/semgrep/dependency_aware_rule.py
```diff
@@ -56,76 +57,74 @@ def parse_depends_on_yaml(
...
+            for match in matches:
+                if lockfile_path in (dep_trie.find_dependencies(match.path) or {}):
                     match.extra["dependency_match_only"] = False
                     match.extra["dependency_matches"] = output_for_json
+                    reachable.append(match)
+                else:
+                    matches_remaining.append(match)
+            matches = matches_remaining
+            if not reachable:
+                dummy_match = RuleMatch(
```
brendongo: What is this dummy_match for?

brendongo (approved):
Non-blocking comments (mostly docs/clarification needed)

semgrep/semgrep/dependency_aware_rule.py
mmcqd: `dummy_match` was a bad name. If the rule had no pattern matches, but there was a dependency match, we want to create a match that points to the offending lockfile. Then the `dependency_match_only` field  added to `extra` specifies that this is what happened.

semgrep/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awareansi-html.yaml-dependency_awareansi.js/results.json
mmcqd: I assumed it had to do with changes you guys made between the time this branch was created and now (since I merged develop a couple times). I'm not sure specifically why they changed/if it's expected.

underyx: well if fingerprinting changed on develop and you merged that into your branch, you'd have also merged the updated fingerprints into your branch 🤔

mmcqd: Actually, I was wrong, this is expected. The target mentioned in the finding changed for all of the findings where the fingerprint changed. Because now dependency only matches are reported as being from the offending lockfile, not from the toplevel. I assume changing that should also change the fingerprint, right?

PR #4961, opened 2022-04-05
Update README.md

ievans:
Don't advertise p/r2c-ci

It's replaced with p/ci anyways and is now strictly worse than config=auto

PR checklist:

- [x] Changelog is up-to-date


s-santillan (approved):

adamboros:
Btw, just wanted to note that within our documentation we rarely show now the usage pattern for rulesets. But I think it's fine since rulesets are mainly found on places like the Registry, where users can copy the command to run.

PR #4962, opened 2022-04-06
Add support for inline rules in join mode

minusworld:
PR checklist:

- [x] Changelog is up-to-date


brendongo (commented):
Code LGTM cc: @ievans on rule syntax

brendongo (approved):
Changelog needs to be updated since we are currently releasing. Maybe just remove it for now and we can add it back in later :( 

Double checking that inline rules will not actually fire and only the join rule gets bubbled up

PR #4968, opened 2022-04-06
chore(tail-recursion): Assert efficient tail-recursion in List ops

nbrahms:
This commit:
- Updates semgrep.yml to check for use of List.map (which is not
  tail-recursive), and map operations piped to List.flatten
- Fixes all such occurrences in our code base

PR checklist:

- [x] Changelog is up-to-date


nmote (approved):
LGTM!

PR #4969, opened 2022-04-06
Rules with invalid patterns fail fast

emjin:
In 0.86.5, semgrep-core started collecting the invalid rules to send at the end. This is nice if we want to fail open, but we haven't implemented that in semgrep yet. This means that rules with bad patterns are getting skipped and then later causing Not_found to be raised when the target looks for the rule. For now, just throw the exception but leave the infrastructure to fail open.

Also changes the code that gets the rules for each target to expect that some might not be found. This assumes that any rule that is not in the table is invalid.

Test plan:

rule_file.yaml
```
rules:
- id: println
  pattern: System.out.println(...);
  message: Using println
  languages: [java]
  severity: WARNING
- id: console_log
  pattern: console.l..);
  message: Using console.log
  languages: [js]
  severity: WARNING
```

Run `semgrep --config rule_file.yaml .`, expect a crash

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/runner/Run_semgrep.ml
+++ b/semgrep-core/src/runner/Run_semgrep.ml
@@ -522,7 +522,7 @@ let semgrep_with_rules config ((rules, invalid_rules), rules_parse_time) =
       RP.matches;
       errors;
       skipped_targets;
-      invalid_rules;
+      skipped_rules = invalid_rules;
       final_profiling = res.RP.final_profiling;
     },
     targets |> List.map (fun x -> x.In.path) )
--- a/semgrep-core/src/runner/Run_semgrep.ml
+++ b/semgrep-core/src/runner/Run_semgrep.ml
@@ -452,11 +452,16 @@ let targets_of_config (config : Runner_config.t)
  * It takes a set of rules and a set of targets and
  * recursively process those targets.
  *)
-let semgrep_with_rules config ((rules, skipped_rules), rules_parse_time) =
-  (* if there are no rules but just skipped rules, better to return an exn *)
-  (match (rules, skipped_rules) with
+let semgrep_with_rules config ((rules, invalid_rules), rules_parse_time) =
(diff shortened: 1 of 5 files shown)


brendongo (approved):

brendongo:
A test would be appreciated so when we put this back we wont have this failure :D

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -452,11 +452,16 @@ let targets_of_config (config : Runner_config.t)
...
+  (* Return an exception
+     - always, if there are no rules but just invalid rules
+     - when users want to fail fast, if there are valid and invalid rules *)
+  (* TODO right now there is no option to not fail fast *)
+  (match (rules, invalid_rules) with
   | [], [] -> ()
   | [], err :: _ -> raise (Rule.InvalidRule err)
+  | _, err :: _ (* TODO fail fast when only when strict? *) ->
+      raise (Rule.InvalidRule err)
   | _ -> ());
```
IagoAbal, line 465: I don't think you need this case, but instead of `| [], [] -> ()` you could just write `| _, [] -> ()`.

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -471,8 +476,10 @@ let semgrep_with_rules config ((rules, skipped_rules), rules_parse_time) =
            let file = target.In.path in
            let xlang = Xlang.of_string target.In.language in
            let rules =
-             Common.map
-               (fun r_id -> Hashtbl.find rule_table r_id)
+             (* Assumption: find_opt will return None iff a r_id
```
IagoAbal, line 479: IMO we should check the assumption and not just document it. In this case it's probably cheap to do so (we don't expect skippe_rules to be large)? If the assumption check fails, we can call `logger#error`.

PR #4970, opened 2022-04-07
chore(cli): PA-928- remove the include/exclude paths for the individual rule…

ghost:
Removed the paths included by rules with exclude/include patterns from --verbose output.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/target_manager.py
+++ b/semgrep/semgrep/target_manager.py
@@ -208,6 +208,20 @@ class FileTargetingLog:
         else:
             yield 2, "<none>"
 
+        yield 1, "Skipped by --include patterns:"
+        if self.cli_includes:
+            for path in self.cli_includes:
+                yield 2, with_color(Colors.cyan, str(path))
+        else:
+            yield 2, "<none>"
+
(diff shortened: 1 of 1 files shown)


semgrep/semgrep/target_manager.py
```diff
@@ -208,20 +208,6 @@ def yield_verbose_lines(self) -> Iterator[Tuple[Literal[0, 1, 2], str]]:
...
-        else:
-            yield 2, "<none>"
-
-        yield 1, "Skipped by --exclude patterns:"
-        if self.cli_excludes:
-            for path in self.cli_excludes:
-                yield 2, with_color(Colors.cyan, str(path))
-        else:
-            yield 2, "<none>"
-
```
underyx: I think you removed the wrong lines here. This looks like logs about the global CLI flags `--include` and `--exclude`. The stuff we wanted to get rid of is the per-rule patterns which are printed for each rule below under `        for rule_id in self.rule_ids_with_skipped_paths:`.

underyx (approved):

PR #4972, opened 2022-04-07
allow single rule

### LINKED ISSUE #4784: update schema to allow a single rule in a config .yaml without a `rules:` superkey
The new editor has a hacked schema that doesn't require nesting under `rules`. Semgrep should just support this natively, as the trend is towards only allowing a single rule in a file anyways.

Update semgrep schema so that this is a valid rule:

```
id: python-no-prints-in-prod
pattern: old_print($X)
message: Use logging.debug() instead of old_print()
languages:
  - python
```

Fix the editor so that it doesn't automagically include `rules:` @underyx 

**Use case**
Easy to copy / paste between editor, playground, local files. Less friction for users.

**Additional context**

ievans:
Closes https://github.com/returntocorp/semgrep/issues/4784

TBH, I'm not sure this is a good idea. A lot of backend logic assumes rules are nested under `rules`. Try running `semgrep -e '"rules"' -l py` on semgrep-app to get a sense: 127 results.

Also would be nice to DRY out the rule-schema but my knowledge of jsonschema is lacking.

PR checklist:

- [ ] Changelog is up-to-date


ievans:
closing this out as I don't think it's a good idea

PR #4973, opened 2022-04-07
Upgrade Scala to GA

mmcqd:
Scala support has unofficially been GA for a while (everything our customers + security team have requested is parseable/working and response time is quick), but this PR officially upgrades it, and adds a couple missing features to take it all the way there. Things this PR does:
* Removes many parsing stats repos that contained unsupported features like XML and Scala 3, and replaces them with others. Parsing rate is now 99.83, which is below 99.9, but most of our other GA languages are also slightly below 99.9. Total lines of scala has gone from ~4 million to ~2 million, because I removed some large repos like the Scala compiler. I think this is acceptable. Our other GA languages have 4+ million lines, but I'm confident that there is a lot less Scala written than Python.
* Adds `...` method chaining
* Adds typed metavariables (syntax : `$X : Int`)
* Adds tests for all GA features (except regex strings, which have been deprecated by `metavariable-regex`)
* Fixes a bug where unicode character literals would not parse
* Fixes a bug where multiple annotated type parameters would fail to parse

PR checklist:

- [x] Changelog is up-to-date


mmcqd:
Not really sure why that pytest snapshot changed, but it seems to have become more correct. The rule that errors is in fact meant to be for JavaScript, not TypeScript.

EDIT: it's changed back now. I still haven't touched anything to do with JS/TS...

mjambon (approved):

CHANGELOG.md
```diff
@@ -135,6 +138,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   constant will also be considered constant (#4301)
 - Constant propagation now recognizes operators `++` and `--` as side-effectful
   (#4667)
+- Scala: unicode characters now parse
+- Scala: mulitple annotated type parameters now parse (`def f[@an A, @an B](x : A, y : B) = ...`)
```
mjambon:
```suggestion
- Scala: multiple annotated type parameters now parse (`def f[@an A, @an B](x : A, y : B) = ...`)
```

PR #4977, opened 2022-04-07
Dockerfile: parse exposed ports as integers

### LINKED ISSUE #4875: [Docker] Port value of `EXPOSE` is parsed as a String, not an integer.
**Describe the bug**
The value of the port after the `EXPOSE` keyword is parsed as a string.

It is my assumption that this is why the following rule with `metavariable-comparison` does not work:
https://semgrep.dev/s/pieterdc1:dockerfile-invalid-port

**To Reproduce**
image

**Use case**
Porting dockerfile rules from `generic` to `dockerfile`

mmcqd:
`EXPOSE 12345` now parses `12345` as an integer instead of a string. 
`EXPOSE 12345/udp` parses `12345` as an integer and `/udp` as a string.

Fixes #4875.

test plan: make test

PR checklist:

- [x] Changelog is up-to-date


mjambon (approved):
looks good to me

PR #4983, opened 2022-04-08
SCA: allow generic rules, don't fail if no lockfiles are found

mmcqd:
This PR fixes two bugs:
* If no lockfiles for a language in a rule were found, semgrep would fail
* Generic mode rules using `r2c-internal-project-depends-on` always failed

PR checklist:

- [x] Changelog is up-to-date


semgrep/tests/e2e/test_dependency_aware_rules.py
```diff
@@ -31,3 +37,23 @@ def test_dependency_aware_rules(run_semgrep_in_tmp_no_symlink, snapshot, rule, t
         run_semgrep_in_tmp_no_symlink(rule, target_name=target)[0],
         "results.json",
     )
+
+
+# Quite awkward. To test that we can handle a target whose toplevel parent
+# contains no lockfiles for the language in our rule, we need to _not_ pass in
+# a target that begins with "targets", as that dir contains every kind of lockfile
```
brendongo, line 44: Why does `targets` contain every kind of lockfile?

mmcqd: Because of `targets/dependency_aware`

brendongo: Doesnt SCA only ever move up the tree anyway so running semgrep in `targets/basic` is not affected by lock files in `targets/dependency_aware`?

mmcqd: Not exactly, though perhaps it ought to work that way. Currently we go to the top-most parent of the first semgrep target, then recursively traverse it for all lockfiles for the languages in the rules. Target files get associated to the nearest lockfile up the tree, but this bug occurred when there were no lockfiles _at all_, not just not lockfiles associated to the target.

brendongo: ah i see

brendongo (approved):

semgrep/tests/e2e/test_dependency_aware_rules.py
brendongo: I guess that functionality is out of scope of this PR

PR #4986, opened 2022-04-08
Guard unix-only module load in unix check

### LINKED ISSUE #4947: ModuleNotFoundError: No module named resource - Windows 10
**Describe the bug**
Can't start Semgrep Windows 10

**To Reproduce**
1. pip install semgrep
2. semgrep

**Expected behavior**
Start application, but doesn't

**Screenshots**
image

**Environment**
Yes
**Use case**
Start semgrep

nbrahms:
This was causing us to hard-fail on Windows 100% of the time.

(Probably) fixes #4947.

Test-plan: just automated CI; we don't have any Windows test environments.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - Moved description of parse/internal errors to the "skipped" section of output
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


brendongo:
Idk if there's any point in fixing this part of the code to work with Windows when we don't make any guarantees / even check that semgrep-core even runs in Windows

PR #4998, opened 2022-04-11
add some more invalid test cases

ievans:
Adds some tests for rule writing issues I encountered with especially bad / confusing error messages.

PR checklist:

- [x] Changelog is up-to-date


semgrep/tests/e2e/test_rule_parser.py
```diff
@@ -90,3 +90,38 @@ def test_rule_parser_cli_pattern(run_semgrep_in_tmp, snapshot):
...
+    snapshot.assert_match(excinfo.value.stderr, "error.txt")
+
+
+@pytest.mark.kinda_slow
+def test_rule_parser_error_invalid_key_name_text(run_semgrep_in_tmp, snapshot):
+    with pytest.raises(CalledProcessError) as excinfo:
+        run_semgrep_in_tmp(
+            f"rules/syntax/invalid-patterns-key.yml",
+            output_format=OutputFormat.TEXT,
+            force_color=True,
```
underyx, line 125: Hmm. Forcing color lets us test that any output formatting is consistent. But it also makes reviewing the snapshots a lot more difficult (I like looking at snapshots for instance in a PR, taking a step back, and just thinking about whether I like the output.)

Not a blocking comment, but you might wanna consider 1) not forcing color, 2) running two assertions with only one of them forcing color, or 3) which is my favorite, snapshotting first with color, then running the output through some ANSI sanitizer library, and snapshotting again.

underyx (approved):

semgrep/tests/e2e/test_rule_parser.py
ievans: Agreed @underyx . I'm hoping to follow this up at some point with a PR that makes these error messages a lot clearer, will do textual snapshots then!

PR #4999, opened 2022-04-11
Fix Ruby regexp matching in the presence of escape characters

### LINKED ISSUE #4991: [ruby] certain backslashes break regex matching
**Describe the bug**
Certain character-class sequences in regex break regex matching

**To Reproduce**
https://semgrep.dev/s/WDZE

**Expected behavior**
Regexes match

nmote:
Fixes #4991

The tree sitter parser returns each escape sequence as a distinct node within the string. `Parse_ruby_tree_sitter.ml` converts escape sequence nodes into `StrChars` nodes, indistinguishable from those generated from ordinary string content. However, they remain separate nodes.

This generalizes the conversion of these nodes to generic AST by combining multiple `StrChars` nodes together rather than falling back. If there is interpolation, we still fall back the way we did before. I've added a test case to illustrate this.

Test plan:

* Additional automated tests
* Manual inspection of the output of `semgrep-core -lang ruby -dump_ast regexp_literal.rb` to verify that the string contents of the Regexp nodes match the original string contents of the regex literal.

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/parsing/pfff/ruby_to_generic.ml
```diff
@@ -436,12 +436,18 @@ and literal x =
...
+  | Regexp ((l, xs, r), opt) ->
+      let rec f strs toks = function
+        | [ StrChars (s, t) ] ->
+            let str = String.concat "" (s :: strs) in
+            let tok = PI.combine_infos t toks in
+            G.L (G.Regexp ((l, (str, tok), r), opt))
+        | StrChars (s, t) :: tl -> f (s :: strs) (t :: toks) tl
+        | _ ->
+            (* TODO *)
+            string_contents_list (l, xs, r)
```
nbrahms: Is this typed enough that this can be
```ocaml
| [] -> ...
```
?

nbrahms (approved):
lgtm, just one type exhaustion question

semgrep-core/src/parsing/pfff/ruby_to_generic.ml
nmote: Are you asking whether `[]` can ever flow into this function? My guess is no, but I'm not sure. Either way, it will have the same result as it previously did.

nmote: Oh, I think I understand. You're asking whether I can replace the wildcard match with a match against `[]`. No, we could also encounter `StrExpr` instead of `StrChars` at any point in the list.

nmote: I've updated it to be an exhaustive match without using the wildcard. Should be clearer now.

nmote:
I'm guessing the benchmark issue was spurious since I have all green after my unrelated update. Merging.

IagoAbal:
@nmote I love the detail in your PR comments. Let me just suggest that those notes could be written as part of the commit message. When you open a PR with a single commit, the commit message automatically becomes the PR comment. And as a bonus, we keep those valuable comments in the Git history, they could be very useful in the future while doing archeology.

nmote:
Sounds good, I'll write this kind of thing up in the commit message in the future.

Issue #4866, opened 2022-03-23
[Docker] environment variables can not be matched using metavariables in the ENV statement


0xDC0DE:
**Describe the bug**
I want to find the value of a environment variable declared with ENV
```
pattern: ENV ...
```
matches the line on my local version, but not in the playground. 
https://semgrep.dev/s/pieterdc1:dockerfile-env-ellipse

<img width="1649" alt="image" src="https://user-images.githubusercontent.com/3809983/159712676-d7f20809-537b-4220-a392-31c17228ee06.png">

```
pattern: ENV $KEY=$VALUE
```
doesn’t match in either place

https://semgrep.dev/s/pieterdc1:dockerfile-env-value
<img width="1651" alt="image" src="https://user-images.githubusercontent.com/3809983/159712901-98f55835-89e5-4394-9e61-1921482a9948.png">

I also found that this is not parseable when using `semgrep-core -dump_ast`
<img width="1019" alt="image" src="https://user-images.githubusercontent.com/3809983/159712507-dc2c0687-7b17-4cb5-b06e-7162ffde082f.png">

**Expected behavior**
The most intuitive would be for environment values to be evaluated, When writing a rule that checks for the directory I was hoping this would just work:
https://semgrep.dev/s/pieterdc1:dockerfile-workdir

Otherwise, I would expect to be able to capture the environment variable name and value into two separate metavariables so I can check 
1) where the variable is used 
2) what the value of the variable is 
Something similar to this rule:
https://semgrep.dev/s/pieterdc1:dockerfile-workdir-env

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [X] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Use case**
Porting r2c rules from generic to dockerfile parser

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1032/[docker]-environment-variables-can-not-be-matched-using-metavariables.

mjambon:
> I also found that this is not parseable when using `semgrep-core -dump_ast`

To parse a pattern, use `-dump_pattern` instead of `dump_ast`.

mjambon:
@pieterdc1 I'll let you report this to the app team since it appears to be an issue with the live editor, possibly because the target file is named wrong (given the presence of the `include` filters).

0xDC0DE:
@mjambon 

[...]

PR #5002, opened 2022-04-12
Add constant propagation to Dockerfile, support constant propagation for metavariable regex

mmcqd:
This PR does three things, if need be it can be separated into multiple PRs.

1. Dockerfile now has constant propagation. Specifically `ENV` variables will be propagated as long as they are not shadowed (which is the only way for them to be non-constant).
2. To make (1) more useful `EXPOSE 123` will no longer parse 123 as an integer, but instead will treat it as a string, reversing one part of #4977. This is needed because `ENV` variables have their values parsed as strings. Handling ints specially here would require handling them specially in several more places. It's not worth it since the issue #4875 can be fixed using `strip: true` anyway.
3. The `metavariable-regex` key now supports an additional `constant-propagation` key, which takes a boolean, and will use constant propagation on the metavariable in question before running the regex if set to true. It's default value is false.

These combine to solve #4866, which will now work as long as the test does not shadow `dirpath` and makes use of `constant-propagation: true`

test plan: make test (tests added)

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -267,6 +280,25 @@ and eval_op op values code =
 
 (* when called from the new semgrep-full-rule-in-ocaml *)
 
+let text_of_binding mvar mval =
```
mmcqd, line 283: This just got moved, it's not a new function

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Scala support is now officially GA
   - Ellipsis method chaining is now supported
   - Type metavariables are now supported
+- `metavariable-regex` now supports and optional `constant-propagation` key.
```
IagoAbal: typo s/and optional/an optional/

semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -496,23 +511,45 @@ let (terraform_stmt_to_vardefs : item -> (ident * expr) list) =
...
+      ({ e = Call ({ e = N (Id (("ENV", _), _)); _ }, (_, args, _)); _ }, _) ->
+      args
+      |> Common.map_filter (function
+           (* A shadowed variable cannot be constant, just ignore it *)
+           | ArgKwd (lbl, e) ->
+               if Hashtbl.mem seen lbl then None
+               else (
+                 Hashtbl.add seen lbl ();
+                 Some (lbl, e))
+           | _ -> (* Should never actually happen *) None)
```
IagoAbal: `logger#error` then?

semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -496,23 +511,45 @@ let (terraform_stmt_to_vardefs : item -> (ident * expr) list) =
...
 
+let (dockerfile_stmt_to_vardefs : item -> (ident * expr) list) =
+ fun st ->
+  let seen = Hashtbl.create 10 in
+  match st.s with
+  | ExprStmt
+      ({ e = Call ({ e = N (Id (("ENV", _), _)); _ }, (_, args, _)); _ }, _) ->
+      args
+      |> Common.map_filter (function
+           (* A shadowed variable cannot be constant, just ignore it *)
```
IagoAbal, line 522: I don't know too much about Dockerfiles, but can't they be constants? Or is it just that you're not sure about what's their exact value (but you know it's one of a few options) ?

semgrep-core/src/core/Rule.ml
```diff
@@ -144,7 +144,7 @@ and metavar_cond =
    * update: this is also useful to keep separate from CondEval for
    * the "regexpizer" optimizer (see Analyze_rule.ml).
    *)
-  | CondRegexp of MV.mvar * regexp
+  | CondRegexp of MV.mvar * regexp * bool (* constant propagation? *)
```
IagoAbal, line 147: It could be nice having a record type for this, although at this point I think a tuple is still fine.

semgrep-core/tests/OTHER/rules/compare-exposed-port.yaml
```diff
@@ -20,4 +20,6 @@ rules:
           - pattern: EXPOSE $PORT
           - metavariable-comparison:
               metavariable: $PORT
-              comparison: $PORT > 65535
\ No newline at end of file
+              comparison: $PORT > 65535
+              strip: true
+              base: 10
```
IagoAbal: `base: 10` here is actually ignored by semgrep-core.

semgrep/semgrep/rule_schema.yaml
```diff
@@ -190,6 +190,8 @@ definitions:
             type: string
           regex:
             type: string
+          constant-propagation:
```
IagoAbal, line 193: Remember to copy this rule schema to Semgrep-app, it seems that they are still using their own copy.

semgrep/tests/e2e/test_metavariable_regex_const_prop.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+
+@pytest.mark.kinda_slow
```
IagoAbal, line 4: I'm curious about what makes this test "kinda slow".

semgrep-core/tests/dockerfile/env-const-prop.dockerfile
```diff
@@ -0,0 +1,5 @@
+
+ENV FOO=0
+
+# MATCH:
+EXPOSE $FOO
```
IagoAbal, line 5: Is it guaranteed that `$FOO` here can only be the port number, or could `$FOO` be the string `22/tcp` ?

semgrep-core/src/engine/Match_search_rules.ml
```diff
@@ -667,8 +667,12 @@ let rec filter_ranges env xs cond =
              in
 
              let env =
-               Eval_generic.bindings_to_env_with_just_strings (fst env.config)
-                 bindings
+               if const_prop then
```
IagoAbal, line 670: I wonder if this option should be independent of the global `constant_propagation` option, maybe we should And them? What should we do if the rule options say `constant_propagation: false` but `metavariable-regex` has `constant-propagation: true`?

semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -304,33 +336,45 @@ let bindings_to_env (config : Config_semgrep.t) xs =
...
-          let file = min.Parse_info.file in
-          let range = Range.range_of_token_locations min max in
-          Some (Range.content_at_range file range))
+let string_of_binding mvar mval =
+  Option.bind (text_of_binding mvar mval) @@ fun x -> Some (mvar, String x)
 
 (* this is for metavariable-regexp *)
-let bindings_to_env_with_just_strings (config : Config_semgrep.t) bindings =
-  let ( let* ) = Option.bind in
+let bindings_to_env_just_strings_const_prop (config : Config_semgrep.t) xs =
```
IagoAbal: Just curious about the `bindings` -> `xs` renaming, isn't `bindings` a more meaningful name?

semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -304,33 +336,45 @@ let bindings_to_env (config : Config_semgrep.t) xs =
...
+  let constant_propagation = config.constant_propagation in
   let mvars =
-    bindings
+    xs
     |> Common.map_filter (fun (mvar, mval) ->
-           let* text = text_of_binding mvar mval in
-           Some (mvar, String text))
+           let try_bind_to_exp e =
```
IagoAbal, line 349: This looks like copy-pasted from `bindings_to_env`, can we refactor?

IagoAbal (approved):
Looks good! But if time allows I think there are moderately important details to address.

semgrep-core/src/analyzing/Constant_propagation.ml
mmcqd: Yeah, I was going off of behavior like this in python, where once a variable gets shadowed, constant propagation gives up everywhere: https://semgrep.dev/s/R5Eq. But I see that it works inside a function: https://semgrep.dev/s/ALO2. I tried turning `ENV x=0 y=1` into a block of `ExprStmt (Assign ...)`, but this made it difficult to do matching correctly. I could spend more time trying to make it work if we think it's important.

semgrep/tests/e2e/test_metavariable_regex_const_prop.py
mmcqd: Nothing in particular, in fact it probably isn't slow. But as far as I can tell every other test in the repo is either marked `kinda_slow` or `slow`, so I just went with that.

semgrep-core/tests/dockerfile/env-const-prop.dockerfile
mmcqd:
```
ENV FOO=22/tcp
EXPOSE $FOO
```
is valid, yes.

semgrep-core/src/engine/Match_search_rules.ml
mmcqd: I think `And`ing them makes sense

semgrep-core/src/engine/Eval_generic.ml
mmcqd: you're right, it wasn't purposeful. Just happened while I was shuffling things are around and making some names more consistent. I should have made the other `xs` into `bindings` instead of this

semgrep-core/src/engine/Eval_generic.ml
mmcqd: yeah, it is. Definitely ugly. Initially I tried to combine them but the resulting function felt pretty hard to read. I'll give it another go before merging

semgrep-core/src/engine/Match_search_rules.ml
mmcqd: Actually that's technically already happening, becuse `bindings_to_env_just_strings_const_prop` uses `constant_propagation` from the `config`, but it's definitely confusing to see that.

semgrep-core/src/analyzing/Constant_propagation.ml
mmcqd: Actually, even when I translate `ENV` to assignments, I get the python toplevel behavior. Is there something special that would need to be done here?

mmcqd: Ok, it's because of the dataflow phase of const-prop not running on dockerfile due to lack of function definitions. I think I can make this work.

semgrep-core/tests/dockerfile/env-const-prop.dockerfile
IagoAbal: I am not double checking the code so I may be plain wrong, but aren't we trying to split the argument passed to `EXPOSE` into _port_ and _protocol_? Wouldn't that classify `$FOO` as _port_ when it actually contains both?

semgrep-core/src/analyzing/Constant_propagation.ml
IagoAbal: To be clear I was just trying to clarify if the comment was accurate. I think it's totally fine if you leave the code as it is but update the comment just to say that we don't handle those cases to keep it simple. In the dataflow const-prop indeed we handle them by using `Cst Cstr` which means "a constant of type string", which are useful in case the pattern is just `"..."`.

semgrep-core/src/engine/Eval_generic.ml
IagoAbal: No worries, sometimes it's better to have duplicate code. I just wanted to point out that possibility of refactoring, but you did consider it already. 👍

semgrep-core/tests/dockerfile/env-const-prop.dockerfile
mmcqd: you're right. I'm not entirely sure what to do about that. We could entirely reverse the port/protocol split, but then something like `EXPOSE $X/$Y` wouldn't work. Of course that doesn't currently work anyway...

semgrep-core/src/analyzing/Constant_propagation.ml
mmcqd: Makes sense. I realized it was actually not too bad to do the translation using `DefStmt` without messing up matching.  That makes shadowing actually work. I'm going to close this and open a new PR.

mmcqd:
Closing in favor of https://github.com/returntocorp/semgrep/pull/5011, which can handle variable shadowing

Issue #4917, opened 2022-03-29
Go parsing failure "Found unexpected raw string literal without delimiters"


stevendpclark:
**Describe the bug**
While running semgrep 0.86.5, or any 0.86.x version we get the following error 
```
semgrep --error --include '*.go' --config "p/ci" .
Fetching rules from https://semgrep.dev/registry.
Scanning 1 file with 19 go rules.
[ERROR] Semgrep Core — Fatal error
An error occurred while invoking the Semgrep engine. Please help us fix this by creating an issue at https://github.com/returntocorp/semgrep

At line test.go:1: (Failure "Found unexpected raw string literal without delimiters")

====[ BEGIN error trace ]====
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Parse_go_tree_sitter.raw_string_literal in file "src/parsing/tree_sitter/Parse_go_tree_sitter.ml", line 69, characters 4-69
Called from Parse_go_tree_sitter.expression in file "src/parsing/tree_sitter/Parse_go_tree_sitter.ml", line 694, characters 46-68
Called from Parse_go_tree_sitter.expression_list in file "src/parsing/tree_sitter/Parse_go_tree_sitter.ml", line 586, characters 11-28
Called from Parse_go_tree_sitter.const_spec in file "src/parsing/tree_sitter/Parse_go_tree_sitter.ml", line 1009, characters 15-37
Called from Parse_go_tree_sitter.statement in file "src/parsing/tree_sitter/Parse_go_tree_sitter.ml", line 771, characters 25-44
Called from Parse_go_tree_sitter.source_file.(fun) in file "src/parsing/tree_sitter/Parse_go_tree_sitter.ml", line 1354, characters 19-35
Called from Stdlib__list.map in file "list.ml", line 92, characters 20-23
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Stdlib__list.map in file "list.ml", line 92, characters 32-39
Called from Parse_go_tree_sitter.source_file in file "src/parsing/tree_sitter/Parse_go_tree_sitter.ml", line 1350, characters 2-369
Called from Parse_tree_sitter_helpers.wrap_parser in file "src/parsing/tree_sitter/Parse_tree_sitter_helpers.ml", line 157, characters 13-29
Called from Parse_target.run_parser in file "src/parsing/Parse_target.ml", line 100, characters 18-24
[...]

PR #5010, opened 2022-04-12
Go: parse multiline raw string literals

mmcqd:
We were using a faulty regex to remove the backticks from raw string literals. The faulty regex did not match multiline strings.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
@@ -65,7 +65,12 @@ let raw_string_literal env tok =
   (* Remove leading and trailing backticks. The grammar guarantees that raw
    * string literals will always have leading and trailing backticks, so this
    * String.sub call should be safe. Let's check just to be sure. *)
-  if not (String.get s 0 = '`' && String.get s (String.length s - 1) = '`') then
+  if
+    not
+      (String.length s >= 2
+      && String.get s 0 = '`'
+      && String.get s (String.length s - 1) = '`')
+  then
     failwith @@ "Found unexpected raw string literal without delimiters: " ^ s;
   let s = String.sub s 1 (String.length s - 2) in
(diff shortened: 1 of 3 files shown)


semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
```diff
@@ -65,9 +65,9 @@ let raw_string_literal env tok =
   (* Remove leading and trailing backticks. The grammar guarantees that raw
    * string literals will always have leading and trailing backticks, so this
    * String.sub call should be safe. Let's check just to be sure. *)
-  if not Common.(s =~ "^`\\(.*\\)`$") then
-    failwith "Found unexpected raw string literal without delimiters";
-  let s = Common.matched1 s in
+  if not (String.get s 0 = '`' && String.get s (String.length s - 1) = '`') then
```
nmote: This is almost exactly what I wrote in #4671 before I was asked in code review to change it to the regex :laughing:

I think it would be a good idea to check that the string has at least length 2, as well. It always should be, but it would be better to fail with an intelligible error message if that's not the case.

It would still be nice to update the tree sitter grammar to parse the delimiters separately, like it does for other string literals. Then we wouldn't have to use this hack at all. I wrote up more details in #3938.

nmote (approved):

semgrep-core/src/parsing/tree_sitter/Parse_go_tree_sitter.ml
mmcqd: Well, that's unfortunate lol. Updated with your length check suggestion. And yes, changing the tree-sitter grammar would definitely be better.

aryx:
This closed https://github.com/returntocorp/semgrep/issues/4917

aryx:
This closes https://github.com/returntocorp/semgrep/issues/5167

PR #5011, opened 2022-04-12
Add constant propagation to Dockerfile, support constant propagation for metavariable regex

mmcqd:
This PR does three things, if need be it can be separated into multiple PRs.

1. Dockerfile now has constant propagation. Specifically `ENV` variables will be propagated. This is done by translating `ENV` declarations into actual variable definitions in the generic AST. Shadowing is handled. This is what differentiates this from https://github.com/returntocorp/semgrep/pull/5002.

2. To make (1) more useful `EXPOSE 123` will no longer parse 123 as an integer, but instead will treat it as a string, reversing one part of #4977. This is needed because `ENV` variables have their values parsed as strings. Handling ints specially here would require handling them specially in several more places. It's not worth it since the issue #4875 can be fixed using `strip: true` anyway.

3. The `metavariable-regex` key now supports an additional `constant-propagation` key, which takes a boolean, and will use constant propagation on the metavariable in question before running the regex if set to true. It's default value is false.

These combine to solve #4866, which will now work as long as the test does not shadow `dirpath` and makes use of `constant-propagation: true`

semgrep-app rule schema update: https://github.com/returntocorp/semgrep-app/pull/4986

test plan: make test (tests added)

PR checklist:

- [x] Changelog is up-to-date


mmcqd:
Converted to draft while waiting for progress to be made towards preventing rule schema changes from causing outages for customers

semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -630,16 +648,26 @@ let propagate_basic a b =
...
-            let inputs, xs = AST_to_IL.function_definition lang def in
-            let flow = CFG_build.cfg_of_stmts xs in
-            let mapping = Dataflow_svalue.fixpoint lang inputs flow in
-            Dataflow_svalue.update_svalue flow mapping);
-      }
-  in
-  v (Pr ast)
+  match lang with
+  | Lang.Dockerfile ->
+      (* Dockerfile has no functions. The whole file is just a single scope *)
```
IagoAbal, line 658: This is clever!

emjin (approved):

PR #5020, opened 2022-04-13
core: Log rule and file affected by a memory warning

IagoAbal:
Helps PA-705

test plan:
```
% cat oom.sgrep
<... $A ...>;
...
<... $B ...>;
...
<... $C ...>;
% semgrep-core -lang js -f oom.sgrep perf/input/l1000.js -json -debug
...
[0.133  Warning    Main.Memory_limit    ] [-e/-f on l1000.js] large stack size: 356408 bytes. If a crash follows, you should suspect a stack overflow. Make sure the maximum stack size is set to 'unlimited' or to a value greater than 0 bytes so as to obtain an exception rather than a segfault.
[0.979  Warning    Main.Memory_limit    ] [-e/-f on l1000.js] large heap size: 553 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[2.724  Warning    Main.Memory_limit    ] [-e/-f on l1000.js] large heap size: 1164 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[6.302  Warning    Main.Memory_limit    ] [-e/-f on l1000.js] large heap size: 1652 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[16.315 Warning    Main.Memory_limit    ] [-e/-f on l1000.js] large heap size: 3361 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
...
```

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/runner/Run_semgrep.ml
+++ b/semgrep-core/src/runner/Run_semgrep.ml
@@ -293,7 +293,12 @@ let iter_targets_and_get_matches_and_exn_to_errors config f targets =
          let res, run_time =
            Common.with_time (fun () ->
                try
-                 Memory_limit.run_with_memory_limit
+                 let get_context () =
+                   match !Rule.last_matched_rule with
+                   | None -> file
+                   | Some rule_id -> spf "%s on %s" rule_id file
+                 in
+                 Memory_limit.run_with_memory_limit ~get_context
                    ~mem_limit_mb:config.max_memory_mb (fun () ->
                      (* we used to call timeout_function() here, but this
                       * is now done in Match_rules because we now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Added
 
+### Changed
+
+- Semgrep-core now logs the rule and file affected by a memory warning.
+
+### Fixed
+
+### Added
+
 - Scala support is now officially GA
   - Ellipsis method chaining is now supported
   - Type metavariables are now supported
--- a/semgrep-core/src/system/Memory_limit.ml
+++ b/semgrep-core/src/system/Memory_limit.ml
@@ -18,7 +18,8 @@ let default_heap_warning_mb = 400
    for detailed explanations.
 
 *)
-let run_with_memory_limit ?(stack_warning_kb = default_stack_warning_kb)
+let run_with_memory_limit ?get_context
+    ?(stack_warning_kb = default_stack_warning_kb)
     ?(heap_warning_mb = default_heap_warning_mb) ~mem_limit_mb f =
   if stack_warning_kb < 0 then
     invalid_arg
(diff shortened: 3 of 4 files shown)


semgrep-core/src/engine/Match_rules.ml
```diff
@@ -87,60 +88,64 @@ let check ~match_hook ~timeout ~timeout_threshold default_config rules xtarget =
...
-                         errors =
-                           [
-                             E.mk_error ~rule_id:(Some rule_id) loc "" E.Timeout;
-                           ];
-                         skipped_targets = [];
-                         profiling = RP.empty_rule_profiling r;
-                       }))
+           let context = spf "rule: %s target: %s" (fst r.Rule.id) file in
+           Memory_limit.run_with_memory_limit ~context
+             ~mem_limit_mb:max_memory_mb (fun () ->
```
nbrahms: (Non-blocking suggestion): It's probably about time to lift this inner function to an external definition :D

I was able to make sense of this change with `-w 1`, but it may be easier for later devs to grok if this is split out.

semgrep-core/src/engine/Match_rules.mli
```diff
@@ -12,6 +12,7 @@ val check :
     (string -> Metavariable.bindings -> Parse_info.t list Lazy.t -> unit) ->
   timeout:float ->
   timeout_threshold:int ->
+  max_memory_mb:int ->
```
nbrahms: Suggestion: seems like we're getting to the point that these options should be some auxiliary record 🤷

nbrahms (approved):

semgrep-core/src/engine/Match_rules.ml
emjin: If I'm understanding your suggestion correctly, I think I disagree. I find this easier to read, since this code is logically part of check; we're just wrapping it with the memory limit.

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -293,26 +293,24 @@ let iter_targets_and_get_matches_and_exn_to_errors config f targets =
...
-                     logger#trace "done with %s" file;
-                     v)
+                 (* we used to call timeout_function() here, but this
+                  * is now done in Match_rules because we now
+                  * timeout per rule, not per file since semgrep-python
+                  * pass all the rules to semgrep-core.
+                  *
+                  * old: timeout_function file config.timeout ...
+                  *)
+                 f target |> fun v ->
```
emjin: Are there any memory intensive operations that are not contained within rule/file matching? I'm mostly thinking of file parsing as a possibility?

Edit: File parsing happens lazily so we're ok (thanks for reminding me :D); I think I'm convinced that nothing that happens before check_rules is significant, so this is fine

emjin (approved):
Yay! I'm a little bit concerned if we might be missing any place where the memory could exceed the limit but not be caught, if some user is relying on --max-memory. Also, can you make sure to also test it with `semgrep`?

aryx (approved):

Issue #1849, opened 2020-10-16
Return "value propagation" in JSON output


mschwager:
Building on our [constant propagation](https://github.com/returntocorp/semgrep/blob/develop/docs/pattern-features.md#constants), it would be useful to also return the value the constant holds.

Consider the following rule and code:

```yaml
rules:
- id: value-propagation
  message: Test value propagation
  severity: ERROR
  languages: [python]
  pattern: print($X)
```

```python
CONST = 42

def func():
    print(CONST)
```

Running with `--json` gives the following output:

```
$ python -m semgrep --config /tmp/test.yaml /tmp/test.py --json | python -m json.tool
running 1 rules...
ran 1 rules on 1 files: 1 findings
{
    "results": [
        {
            "check_id": "tmp.value-propagation",
            "path": "/tmp/test.py",
            "start": {
                "line": 4,
                "col": 5
            },
            "end": {
                "line": 4,
                "col": 17
            },
            "extra": {
                "message": "Test value propagation",
                "metavars": {
                    "$X": {
                        "start": {
                            "line": 4,
                            "col": 11,
                            "offset": 34
                        },
                        "end": {
                            "line": 4,
                            "col": 16,
                            "offset": 39
                        },
                        "abstract_content": "CONST",
                        "unique_id": {
                            "type": "id",
                            "value": "CONST",
                            "kind": "Global",
                            "sid": 1
                        }
                    }
                },
                "metadata": {},
                "severity": "ERROR",
                "lines": "    print(CONST)"
            }
        }
    ],
    "errors": []
}
```

The `abstract_content` key tells us the literal code token in the metavariable. It'd be very useful to also receive the value of that token if constant propagation occurred. In this case, something like `"abstract_value": 42`. I can see both the code token and its value being useful to the caller, so we'll likely want a new key in the output.

This will be useful for things like `metavariable-regex`, `metavariable-comparison`, and other applications that build on the JSON output of Semgrep.

aryx:
abstract_content does not contain the code token :) it contains the name of the expression, here the constant named 'CONST'

aryx:
if you do print (($X : int_literal)) maybe you will get the right value in $X.

IagoAbal:
Is this issue still relevant once we are planning to handle rules inside semgrep-core?

aryx:
Well we still need to handle that correctly in semgrep-core, to modify how we do metavariable-regex, and metavariable-comparison to possibly use the propagated constant.

[...]

Issue #4976, opened 2022-04-07
metavariable-comparison shows variable verses integer value 


LewisArdern:
**Describe the bug**
The metavariable-comparison correctly detects that the $VALUE < NUMBER but in the message it fails to output the number just the variable. Many people use the message to capture this type of information for encryption and other factors, so its pretty important that we have the expected value inside the message.

**To Reproduce**
* Many pattern-insides https://semgrep.dev/s/xPjp
* With symbolic - https://semgrep.dev/s/XW74
* direct issue - https://semgrep.dev/s/pieterdc1:insufficient-rsa-key-size-1

**Expected behavior**
The message should contain the integer and not the variable.

**What is the priority of the bug to you?**

- [x] P1: important to fix or quite annoying

**Use case**
It will enable our customers to have a better experience using semgrep to ensure their data is correct in the message, where most people use it for sharing or displaying information in their own pipelines.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1122/metavariable-comparison-shows-variable-verses-integer-value.

nbrahms:
Thanks for the report @LewisArdern !

Issue #4692, opened 2022-02-17
Make clear when a finding is making use of constant-propagation 


IagoAbal:
(This affects customers, more info in the Linear issue.)

**Is your feature request related to a problem? Please describe.**
When findings make use of constant-propagation, and especially when it is based on a false constant (constant-propagation isn't perfect), it may not be obvious why there is a finding.

**Describe the solution you'd like**
Semgrep should indicate that it has made assumptions about certain variables having certain constant values.

**Describe alternatives you've considered**
The simplest solution I can think of is to record these assumptions in `tin` during matching in `Generic_vs_generic`. They can be added to the `Pattern_match.t`, and passed to Semgrep CLI. The CLI could then append these assumptions to the finding's message. This should be relatively simple.

A step further would be to record _how_ we inferred that assumption, but that would be some work.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-884/make-clear-when-a-finding-is-making-use-of-constant-propagation.

nbrahms:
cc. @ievans @adamboros who may have thoughts

adamboros:
Maybe it's implied and I can't interpret it, but what are you planning to pass to the CLI? Is it the value, the name of the variable or maybe even the line where Semgrep made the assumption?

IagoAbal:
I would say the variable and its assumed value, e.g. `x = 1`. We could also add what exact occurrence of `x` that ware referring to.

adamboros:
Sounds good, I think the exact occurrence would be great cause that will help the CLI team to present this at just the right place (instead of just amending it at the end of each finding where it's bound to be harder to decipher for users). Thanks Iago! cc. @underyx @brendongo

PR #5028, opened 2022-04-13
Report the propagated value of a metavariable

emjin:
Users want to access the original propagated value of a metavariable. Semgrep-core now reports the svalue of a metavariable in addition to its literal value to make it possible for users to see the metavariable in the JSON.

Previously, this PR would have also introduced the syntax `$$X` to use the propagated value in the message if available, but that is tabled for now pending rule versioning.

PR checklist:

- [x] Changelog is up-to-date


emjin:
Merge https://github.com/returntocorp/semgrep/pull/5042 first

emjin:
The changes this causes in the tests:

```
        "message": "log4j logger.error tainted argument",
        "message": "log4j LogManager.getLogger(com.example.log4shell.log4j.class).error tainted argument",
```

```
        "message": "'password' is using the empty string as its default and is being used to set\nthe password on 'user'. If you meant to set an unusable password, set\nthe default value to 'None' or call 'set_unusable_password()'.\n",
        "message": "'password' is using the empty string as its default and is being used to set\nthe password on 'self.model(email=self.normalize_email(email))'. If you meant to set an unusable password, set\nthe default value to 'None' or call 'set_unusable_password()'.\n",
```

```
        "message": "useless comparison operation `SUPER_LONG_CONSTANT_TO_TRIGGER_A_LINE_BREAK == SUPER_LONG_CONSTANT_TO_TRIGGER_A_LINE_BREAK` or `SUPER_LONG_CONSTANT_TO_TRIGGER_A_LINE_BREAK != SUPER_LONG_CONSTANT_TO_TRIGGER_A_LINE_BREAK`; possible bug?",
        "message": "useless comparison operation `0 == 0` or `0 != 0`; possible bug?",
```

Some of them make more sense than others. I suspect we need a different solution---either new syntax (`*METAVARIABLE` to use the underlying value in the message) or some other representation (the propagated value in brackets)

semgrep-core/src/reporting/JSON_report.ml
```diff
@@ -101,8 +101,39 @@ let range_of_any_opt startp_of_match_range any =
...
+        Some
+          {
+            ST.svalue_start = start;
+            svalue_end = end_;
+            svalue_abstract_content = metavar_string_of_any any;
+          }
+    | None -> None
+  in
+  match metavar with
+  | E { e = N (Id (_, id_info)); _ } -> (
```
IagoAbal, line 135: Is `metavar` here really a metavariable? To my knowledge only code variables will have a relevant `id_svalue`.

interfaces/Output_from_core.atd
```diff
@@ -106,6 +107,13 @@ type metavar_value
   unique_id: unique_id;
 }
 
+type svalue_value
+  <python decorator="dataclass(frozen=True)"> = {
+  svalue_start: position;
```
IagoAbal: I think these two positions should be optional. We don't try very hard to assign positions to propagated constants so I guess this will filter out many cases.

semgrep-core/src/reporting/JSON_report.ml
```diff
@@ -101,8 +101,39 @@ let range_of_any_opt startp_of_match_range any =
...
+          any_to_svalue_value any
+      | Some (Sym x) -> any_to_svalue_value (E x)
+      | Some (Cst _) -> None
+      | Some NotCst -> None
+      | None -> None)
+  | _ -> None
+
 let metavars startp_of_match_range (s, mval) =
   let any = MV.mvalue_to_any mval in
+  pr2 (show_any any);
```
IagoAbal: Probably want to remove this.

semgrep-core/src/reporting/JSON_report.ml
IagoAbal: Maybe rename it to `mvalue`?

IagoAbal:
I believe this would truly close #1849 and it's also somewhat related to #4692 (this one needs more work).

IagoAbal:
By the way, do we have a Linear issue with details about the users/customers requesting this and the use case?

emjin:
Oops sorry, this is related to https://linear.app/r2c/issue/PA-1122/metavariable-comparison-shows-variable-verses-integer-value, aka https://github.com/returntocorp/semgrep/issues/4976

underyx:
As for the syntax, SQL uses `$$` as a multiline string marker. also, doubling a special character is sometimes used as a way to escape the character. So if you're still looking to alternative, maybe this is like dereferencing in C or whatever languages have that feature? So `*$X` is another option.

I don't think `$$` is bad tho.

underyx:
Also if this is only available in message output, some other more readable implementations would be:

- `message: Found $X.resolve()`
- `message: Found $X.resolve`
- `message: Found resolve($X)`

semgrep-core/src/reporting/JSON_report.ml
```diff
@@ -101,6 +101,42 @@ let range_of_any_opt startp_of_match_range any =
       let startp, endp = position_range min_loc max_loc in
       Some (startp, endp)
 
+let metavar_string_of_any any =
```
IagoAbal, line 104: Maybe add a `TODO` to handle the cases where there is no origin tokens? E.g. in `x = 1; y = x + 1; ...` we have `y = 2` but there is no source location for `2`.

semgrep/semgrep/output_from_core.py
```diff
@@ -474,6 +474,88 @@ def to_json_string(self, **kw: Any) -> str:
         return json.dumps(self.to_json(), **kw)
 
 
+@dataclass(frozen=True, order=True)
```
IagoAbal, line 498: I assume this is being autogenerated by ATD?

semgrep/tests/e2e/rules/message_interpolation/propagated-constant.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - id: test_rule_id
+    patterns:
+      - pattern-inside: |
+          class $CLASS(...):
+            ...
+      - pattern: |
+            $X == $X
+    severity: INFO
+    languages: [python]
+    message: equal comparison of $X ($$X) in $CLASS
```
IagoAbal: You're using `$$X` here, but if I understood correctly you wanted remove that feature from this PR.

semgrep/tests/e2e/test_message_interpolation.py
```diff
@@ -1,44 +0,0 @@
-import pytest
```
IagoAbal, line 1: Do we want to remove these tests?

IagoAbal (approved):

semgrep/tests/e2e/test_message_interpolation.py
emjin: Not all of them, but indeed I should have removed `propagated-constant.yaml`; I missed it

PR #5032, opened 2022-04-14
SCA: respect .semgrepignore when scanning lockfiles

mmcqd:
Fixes [SCA-27](https://linear.app/r2c/issue/SCA-27/sca-finds-findings-in-yarnlock-files-in-node-modules)

Lockfiles which would be ignored by `.semgrepignore` are now actually ignored. Ultimately this process should be integrated with `TargetManager`, and we should report the number of lockfiles we scan and ignore, just like with targets.

test plan: see added test, where we scan `targets/dependency_aware` for lockfiles with `bad-yarn-lib`, but correctly ignore the only one present in `node_modules/`

PR checklist:

- [x] Changelog is up-to-date


semgrep/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awaremonorepo.yaml-dependency_awaremonorepo/results.json
```diff
@@ -14,12 +14,12 @@
     {
       "check_id": "rules.dependency_aware.js-sca",
       "end": {
-        "col": 6,
-        "line": 1,
-        "offset": 5
+        "col": 0,
```
mmcqd, line 17: Ok yeah this test is definitely non-determenistic and bad. It's unrelated to this PR and being ignored by CI right now so nothing is breaking. Going to figure out why this is happening when I'm back from vacation.

brendongo (approved):

brendongo (approved):

PR #5033, opened 2022-04-14
Remove conflicting spacegrep license

nmote:
The LICENSE at the root of the repository should be referred to instead

PR checklist:

- [x] Changelog is up-to-date


nmote:
I chose not to update the changelog since I'm not sure how relevant this is to users, since I believe the license at the root was intended to be the license to begin with, but let me know if I should.

mjambon (approved):

PR #5034, opened 2022-04-14
[Swift] Convert member and prefix expressions to the Generic AST

nmote:
There are three unusual things in this PR that I would like to draw attention to:
* `Thing<Int>.foo` is parsed as two comparison expressions. I implemented the mapper for comparison expressions so that this example would pass, even though the ambiguity should almost certainly be resolved differently. I added a TODO to the test file.
* I added two uses of `OtherExpr`:
  * The first is for implicit member expressions, which are mentioned in the test plan. I don't believe that `.foo` should be treated like a custom operator, since it semantically is a `DotAccess`, just without an explicit target.
  * The second is for an explicit member expression where the target is a type, such as `Dictionary<Int, Int>.foo`. Perhaps there's a way to convert that into a legitimate expression, but it wasn't obvious to me.

Test plan:

The automated tests, plus manual inspection of the Generic AST produced for each of these constructs. For example:

`.foo;`
results in
```
Pr(
  [ExprStmt(
     DotAccess(OtherExpr(("Implicit", ()), []), (),
       FN(
         Id(("foo", ()),
           {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))), ())])
```

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/tests/swift/parsing/expressions.swift
+++ b/semgrep-core/tests/swift/parsing/expressions.swift
@@ -68,3 +68,36 @@ UserType<Int>(bar);
 
 [Int] { x in y };
 [Thing<Int>] { x in y };
+
+// Navigation expressions:
+
+5.foo;
+Int.foo;
+// TODO This is parsed as two comparison expressions. That's most likely wrong.
+Thing<Int>.foo;
+Thing<Int, String>.foo;
+// TODO This is parsed as a navigation expression where the first part is an
+// expression rather than an array type. Figure out if that's the correct
+// resolution to the ambiguity.
+[Int].5;
+[Thing<Int>].5;
+// TODO This is parsed as a navigation expression where the first part is an
+// expression rather than a dict type. Figure out if that's the correct
+// resolution to the ambiguity.
+[Int: Int].42;
+[Int: Thing<Int>].42;
+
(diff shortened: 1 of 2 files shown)


semgrep-core/tests/swift/parsing/expressions.swift
```diff
@@ -68,3 +68,36 @@ UserType<Int>(bar);
...
+
+++5;
+--5;
+-5;
++5;
+!5;
+&5;
+~5;
+.foo;
+.5;
```
nmote, line 100: I believe GitHub is right that this is a syntax error. The suffix for an implicit member expression should always be an identifier. However, currently the grammar allows this.

nmote:
Unfortunately it looks like I've found another construct that leads the Swift parser into an infinite loop in CI but not on my machine... Back to draft until I figure out which one it is.

aryx (approved):

nmote:
Rebased over #5048 which addresses the hang

nmote:
The benchmark issue seems to be spurious. It also happened on https://github.com/returntocorp/semgrep/pull/5049, which only changed comments.

PR #5044, opened 2022-04-18
fix: workaround for setting git safe.directory

brendongo:
After git was updated to 2.35.2 with a security fix, we were seeing
https://github.com/actions/checkout/issues/766 since the user running semgrep
in docker is not the user who copied the repository. Suggested workaround of
calling `git config --global --add safe.directory "$GITHUB_WORKSPACE"` was used

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/cli.py
+++ b/semgrep/semgrep/cli.py
@@ -1,4 +1,7 @@
 #!/usr/bin/env python3
+import os
+import subprocess
+
 import click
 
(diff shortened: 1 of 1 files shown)


underyx (approved):

brendongo:
Workaround works: https://github.com/returntocorp/semgrep-app/runs/6068054904?check_suite_focus=true

PR #5048, opened 2022-04-19
Update semgrep-swift to latest

nmote:
See https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/298

Test plan: Automated tests

PR checklist:

- [x] Changelog is up-to-date


nmote:
I rebased #5034 atop this and it has fixed the infinite looping behavior we observed in CI. However, I'm still puzzled about *why* it got into the state where it looped, and why it only happened in CI. We should make sure to test the constructs in question to make sure they are parsed correctly, when we get to that point.

aryx (approved):

Issue #4281, opened 2021-11-14
License confusion


nightwatchcyber:
The licensing for semgrep CLI (not the rules) seems to be LGPL as indicated here:
https://github.com/returntocorp/semgrep/blob/develop/LICENSE
https://pypi.org/project/semgrep/
https://semgrep.dev/docs/faq/

However, within the source there are other licenses that appear such as AGPL, Apache and GPL, and some of this code is shipped with semgrep (such as spacegrep):
https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/src/spacegrep/LICENSE
https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/src/engine/Test_comby.ml
https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/tests/java/parsing/Ghidra2.java

Can someone clarify what the actual license is?

Thanks

nbrahms:
@ievans @mjambon @aryx can y'all work together to resolve these discrepancies?

ievans:
Hi @nightwatchcyber, 

The source for semgrep itself (this repository) is LGPL-2.1 licensed, but as you mention we embed many different sub-repositories which have different licenses. More permissive licenses such as Apache and MIT are well-understood to not pose a problem in such a project, but others, for instance AGPL, do. Fortunately our company (r2c) is the copyright holder by virtue of being the original author of spacegrep, so we are able to license it for use in Semgrep under LGPL even though the "standalone" license is AGPL.

This is admittedly not at all explicit, and there are a few tasks I'm tracking already in terms of double-checking on the dependency licenses and making our own license selection consistent. I will add this to the queue, but if it is a time-sensitive issue for you or your team feel free to email me (I'm isaac@ the r2c domain name).

ievans:
This is in progress!

tapdiego-amzn:
Hi, wondering if there's any news on this?

ievans:
@tapdiego-amzn we provided an update on our community Slack a few weeks ago--tl;dr still a work in progress

ievans:
@aryx going to re-open until I've had a chance to verify

ievans:
@aryx @tapdiego-amzn @nightwatchcyber we found a few license headers that indeed appeared to contradict the overall license and tracked those down with the original contributors. Thanks for the pings. I can verify we are now fully LGPL for redistributed Semgrep.

[...]

PR #5096, opened 2022-04-27
Use latest pfff with only LGPL code

aryx:
Also remove linking exception clause for now.

test plan:
git grep GPL in semgrep-core/src/pfff/ and check if
remaining GPL (vs LGPL) files

PR checklist:

- [x] Changelog is up-to-date


aryx:
This will help https://github.com/returntocorp/semgrep/issues/4281

aryx:
CI failures are unrelated

PR #5098, opened 2022-04-27
[refactor] reduce core_output.py a bit more

aryx:
test plan:
make test

PR checklist:

- [x] Changelog is up-to-date


aryx:
After that I'll go after removing CoreMatch (and just reuse core.Match),
and the other simple wrappers in this file.

aryx:
A difference between CoreMatch and core.Match is that the second one just a rule_id instead of a rule.
@emjin do you think we can just replace CoreMatch by core.Match and pass around the rule_table in code using CoreMatch
so they can access the Rule from the rule_id?

aryx:
merging this @emjin , so you can start from there to remove also CoreTimingXxxx

PR #5101, opened 2022-04-28
Switch to RuleId instead of Rule for profiling.py

aryx:
This is part of the work to reduce core_output.py
by using directly types from output_from_core.py

test plan:
make test

PR checklist:

- [x] Changelog is up-to-date


aryx:
This PR is needed to simplify CoreRuleTiming in core_output.py down the road.

PR #5116, opened 2022-04-29
feat(ci): print all output by default

### LINKED ISSUE #5069: Flag to make audit findings visible in semgrep ci output
`semgrep ci` hides the findings from audit column rules. While for PR scans it makes sense to hide these, for full scans users might want to export the finding data (--json)

Expose a `--show-audit-findings` flag that shows the non-blocking audit rules in the output of `semgrep ci`

** Additional Context **
Right now semgrep.dev rule board has three columns: audit, non-blocking notify, blocking. For now `semgrep ci` only includes the blocking findings in the output. We will want to have non-blocking notify always visible in the output and this flag will be to control the visibility of the audit column.

brendongo:
Previously we were only printing out findings from blocking
rules in semgrep ci. This change prints out all findings by
default.

PR checklist:

- [x] Changelog is up-to-date


brendongo:
Closes https://github.com/returntocorp/semgrep/issues/5069

semgrep/semgrep/output.py
```diff
@@ -329,7 +324,7 @@ def output(
             else:
                 if output:
                     try:
-                        print(output, file=self.stdout)
```
brendongo, line 332: @underyx This was causing findings to not appear in test_ci stdout (it was bypassing CliRunner.invoke's stdout capture). We were checking findings with the network capture part of the test but thought would be nice to have test for --json output and regular output with this PR. Looks like we never use the stdout stderr args of OutputHandler anyway so I think this is fine

PR #5124, opened 2022-05-02
tainting: Optimize the test that checks for taint labels

IagoAbal:
To know that e.g. an expression is a source of taint, we check whether
its range is in a list of pre-computed source ranges. This test may be
performned millions of times for each taint rule, so any small cost ends
up slowing things down noticeably.

test plan
% make test
% semgrep-core -lang java -rules taint.yaml stress-test-monorepo/repos/intellij-community
    ^ at least ~30s or 1.2x faster after this change, and running N taint
    ^ rules would multiply the cost by N!
    ^ where taint.yaml is just a dummy taint rule:
```
rules:
  - id: test-rule
    languages:
      - java
    message: Test
    mode: taint
    pattern-sources:
      - pattern: user_input
    pattern-sinks:
      - pattern: sink(...)
    severity: WARNING
```

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Match_tainting_rules.ml
+++ b/semgrep-core/src/engine/Match_tainting_rules.ml
@@ -98,9 +98,10 @@ let any_in_ranges rule any rwms =
    * the AST at some point. *)
   match Visitor_AST.range_of_any_opt any with
   | None ->
-      logger#debug
-        "Cannot compute range, there are no real tokens in this AST: %s"
-        (G.show_any any);
+      if any <> G.Anys [] then
+        logger#info
+          "Cannot compute range, there are no real tokens in this AST: %s"
+          (G.show_any any);
       []
   | Some (tok1, tok2) ->
       let r = Range.range_of_token_locations tok1 tok2 in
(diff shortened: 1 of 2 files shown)


semgrep-core/src/engine/Match_tainting_rules.ml
```diff
@@ -98,9 +98,10 @@ let any_in_ranges rule any rwms =
    * the AST at some point. *)
   match Visitor_AST.range_of_any_opt any with
   | None ->
-      logger#debug
-        "Cannot compute range, there are no real tokens in this AST: %s"
-        (G.show_any any);
+      if any <> G.Anys [] then
```
aryx, line 105: there's a few places where we don't have a token, see Metavariable.ml

  (* Those can be now empty with $...XXX metavariables.
   * coupling: if you add more constructors that allow an empty content,
   * you may need to modify JSON_report.range_of_any to not get
   * some NoTokenLocation exn.
   *)
  | Ss of AST_generic.stmt list
  | Args of AST_generic.argument list
  | Params of AST_generic.parameter list
  | Xmls of AST_generic.xml_body list


so I guess you could check for those too.

or maybe this can never happen.

aryx (approved):

semgrep-core/src/engine/Match_tainting_rules.ml
IagoAbal: Good question. I added a clarification comment. This is because I convert `orig` into `any` just for the purpose of checking for taint labels, and `NoOrig` becomes `Any []`, so there are a lot of those pointless warnings.

PR #5127, opened 2022-05-02
Fix mypy error with most recent requests version

underyx:
The latest requests typings don't allow setting a header to None
(even though the functionality still works).
This commit removes the only usage of this feature.

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/app/metrics.py
+++ b/semgrep/semgrep/app/metrics.py
@@ -10,6 +10,7 @@ from typing import Sequence
 from urllib.parse import urlparse
 
 import click
+import requests
 
 from semgrep.app import app_session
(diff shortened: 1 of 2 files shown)


brendongo (approved):

nbrahms (dismissed):
This seems to be changing the functionality (viz., we're now sending auth tokens).

IMO the desired behavior is to _not_ send auth tokens when auth is unnecessary.

brendongo:
> IMO the desired behavior is to not send auth tokens when auth is unnecessary.

@nbrahms isnt that what this PR is doing?

underyx:
not needed after the changes in https://github.com/returntocorp/semgrep/pull/5288

PR #5133, opened 2022-05-03
put back fingerprint: and path: in results.json

aryx:
This allows to sanity check that those JSON follows the ATD spec
for semgrep JSON output

test plan:
make test

PR checklist:

- [x] Changelog is up-to-date


semgrep/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awaremonorepo.yaml-dependency_awaremonorepo/results.json
```diff
@@ -14,12 +14,12 @@
     {
       "check_id": "rules.dependency_aware.js-sca",
       "end": {
-        "col": 6,
-        "line": 1,
-        "offset": 5
+        "col": 0,
```
aryx, line 17: @mmcqd not sure what's going on here, but it looks mostly like a reordering of the JSON. It looks mostly equivalent otherwise. Not sure why the modification in conftest.py (not removing the path: and fingerprint: field) has an impact on this.

PR #5135, opened 2022-05-03
Move Semgrep_scan_output_v1.atd to shared repo semgrep-interfaces

aryx:
test plan:
make test

PR checklist:

- [x] Changelog is up-to-date


semgrep/semgrep/app/metrics.py
```diff
@@ -254,7 +255,9 @@ def send(self) -> None:
                 METRICS_ENDPOINT,
                 json=self.as_dict(),
                 # metrics ingestion shouldn't see auth tokens
-                headers={"Authorization": None},
+                # TODO: weird, we get a mypy error only in CI, and only
```
aryx, line 258: @brendongo I have no idea why I need this cast() here, and why it's needed only for CI.
Locally I don't need it. also I have no idea why I need to add this cast only now after I've added submodules=true
above in the pre-commit workflow.

PR #5150, opened 2022-05-04
tainting: Add hack to detect some side-effectful updates

IagoAbal:
And fix -dfg_tainting to work on class methods.

test plan:
make test # test added

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Changed
+
+- Taint analysis now assumes that void methods may update the callee object, so
+  if the arguments are tainted, the callee will also be tainted. This e.g. makes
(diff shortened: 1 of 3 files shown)


mjambon (approved):

aryx:
regressions in semgrep rules: FAILED tests/e2e/test_semgrep_rules_repo.py::test_semgrep_rules_repo - Failed...

IagoAbal:
This hack doesn't work so well for taint-mode. 😞

IagoAbal:
In constant propagation we use this trick to conservatively remove constant info. This doesn't cause false positives.

But this PR uses the same trick for taint-mode, where it is adding taint to objects. If the inference is wrong, then you do cause false positives. The cases I found in semgrep-rules were not too serious (e.g. `out.println(tainted)`), but it's ugly.

It's better to work on taint propagators as I already implemented a PoC during hack week!

PR #5173, opened 2022-05-09
Fix issue with Python version breaking homebrew release

spencerdrak:
PR checklist:

- [x] Changelog is up-to-date


.github/workflows/release.yml
```diff
@@ -302,15 +302,23 @@ jobs:
     name: Update on Homebrew-Core
     needs: [sleep-before-homebrew] # Needs to run after pypi released so brew can update pypi dependency hashes
     runs-on: macos-10.15
-    strategy:
-      matrix:
-        python: ["3.9.12"]
     steps:
       - name: Setup Python
         uses: actions/setup-python@v1
```
underyx:
```suggestion
        uses: actions/setup-python@v3
```

.github/workflows/release.yml
```diff
@@ -302,15 +302,23 @@ jobs:
...
         with:
-          # Needs to be < 3.10 see https://github.com/returntocorp/semgrep/issues/4213
-          python-version: ${{ matrix.python }}
+          # Pin to python 3.9 for pipgrip issues: see https://github.com/returntocorp/semgrep/issues/4213
+          python-version: '3.9.*'
```
underyx:
```suggestion
          python-version: '3.9.x'
```

I'm sure that worked as well, but this format matches the documentation at https://github.com/actions/setup-python

.github/workflows/release.yml
```diff
@@ -302,15 +302,23 @@ jobs:
...
+        id: python-version
+        run: |
+          # older versions of python pass the version output on stderr, redirect this to stdout
+          # Input: Python 3.9.12
+          # Output: 3.9.12
+          # sed -n: only print match
+          # [Pp]ython : consume initial output up to first digit
+          # [0-9][0-9]*.[0-9][0-9]*.[0-9][0-9]*: match a pattern of three groups of one or more digits and dots, using basic RE
+          PYTHON_VERSION=$(python --version 2>&1 | sed -n 's|[Pp]ython \([0-9][0-9]*.[0-9][0-9]*.[0-9][0-9]*\)*$|\1|p')
+          echo ::set-output name=python_version::"$PYTHON_VERSION"
```
underyx: Seems like we can actually get rid of all this code cause setup-python outputs the version 🎉 https://github.com/actions/setup-python/issues/235

So something like `${{ steps.setup-python.outputs.python-version }}` should work already if we set a step ID for it.

.github/workflows/release.yml
spencerdrak: change applied

.github/workflows/release.yml
spencerdrak: change applied

.github/workflows/release.yml
spencerdrak: change applied

underyx (approved):

PR #5174, opened 2022-05-09
fix: Set all directories as safe for Git in Docker

underyx:
This fixes crashes that happen when running Semgrep in Git on a
directory that is not owned by the current user.

PR checklist:

- [x] Changelog is up-to-date


brendongo (approved):

underyx:
@returntocorp/security FYI this change has security implications

PR #5179, opened 2022-05-10
Fix docker badge not being flat-square in README.md

Jasperhino:
PR checklist:

- [x] Changelog is up-to-date


underyx (approved):
thanks :D that's better 😌

PR #5194, opened 2022-05-11
refactor: Use click context to keep app session & metrics around

underyx:
PR checklist:

- [ ] Changelog is up-to-date


semgrep/tests/e2e/test_ci.py
```diff
@@ -2,10 +2,8 @@
 import re
 import shutil
 import subprocess
-import sys
 from pathlib import Path
 from textwrap import dedent
-from unittest import mock
```
underyx, line 8: I installed `pytest-mock` so instead of ever importing mock, we should use the `mocker` fixture instead.

semgrep/tests/e2e/test_ci.py
```diff
@@ -92,14 +89,15 @@ def git_tmp_path_with_commit(monkeypatch, tmp_path):
...
+    )
 
-import contextlib
+    yield (repo_base, base_commit, head_commit)
 
 
-@contextlib.contextmanager
-def ci_mocks(base_commit, autofix):
+@pytest.fixture(autouse=True)
+def automocks(mocker):
```
underyx, line 103: I made this an autouse fixture, so instead of using it as a context manager, these mocks get applied to every test case in this file automatically.

semgrep/tests/e2e/test_ci.py
```diff
@@ -135,29 +133,20 @@ def ci_mocks(base_commit, autofix):
...
-                    with mock.patch.object(
-                        ScanHandler, "autofix", mock.PropertyMock(return_value=autofix)
-                    ):
-                        with mock.patch.object(AppSession, "post"):
-                            yield
+    mocker.patch.object(ConfigPath, "_make_config_request", return_value=file_content)
+    mocker.patch.object(
+        ScanHandler, "_get_deployment_details", return_value=(DEPLOYMENT_ID, "org_name")
+    )
+    mocker.patch("semgrep.app.auth.is_valid_token", return_value=True)
```
underyx, line 143: When we patch a member of a module (like here, it's a function defined on the top level). We can use this form instead of `mocker.patch.object` so that we don't need to import. It's kinda annoying that this way IDEs will not automatically update import paths, so maybe this is worse though.

semgrep/tests/e2e/test_ci.py
```diff
@@ -135,29 +133,20 @@ def ci_mocks(base_commit, autofix):
...
+    mocker.patch.object(AppSession, "post")
+
+
+@pytest.fixture(params=[True, False], ids=["autofix", "noautofix"])
+def autofix(request, mocker):
+    mocker.patch.object(ScanHandler, "autofix", request.param)
```
underyx, line 149: These same parameters were duplicated across multiple test functions. So here I extracted them to a fixture, and just by requesting the `autofix` fixture pytest will automatically parametrize a test function with autofix/noautofix for us.

This happened to change the test IDs though (local-autofix to autofix-local) which is why all those snapshot files are renamed.

semgrep/tests/e2e/test_ci.py
```diff
@@ -92,14 +89,15 @@ def git_tmp_path_with_commit(monkeypatch, tmp_path):
         ["git", "rev-parse", "HEAD"], encoding="utf-8"
     ).strip()
 
-    yield (repo_base, base_commit, head_commit)
-
+    mocker.patch.object(
+        GitlabMeta, "_fetch_branch_get_merge_base", return_value=base_commit
+    )
```
underyx, line 97: In this case we had a fixture that generates test data, and I figured we might as well use this as the standard place where that data is mocked into the code!

semgrep/tests/e2e/test_ci.py
```diff
@@ -271,208 +256,187 @@ def test_full_run(tmp_path, git_tmp_path_with_commit, snapshot, env, autofix):
...
+        .replace(head_commit[:7], "<sanitized head_commit>")
+        .replace(base_commit, "<sanitized base_commit>")
+        .replace(__VERSION__, "<sanitized semgrep_version>")
+    )
+    sanitized_output = re.sub(
+        r"python 3\.\d+\.\d+", "python <sanitized_version>", sanitized_output
+    )
+    snapshot.assert_match(sanitized_output, "output.txt")
 
+    post_calls = AppSession.post.call_args_list  # type: ignore
```
underyx, line 286: 1. Instead of keeping the return value of the mock call around as a variable, We can directly access the mocked object under the `semgrep.` namespace.
2. Instead of checking `.mock_calls`, which contains all attribute access, etc. (this is why we had that comment saying `Return a different mock so post_mock only contains direct calls to session.post`), we can access `.call_args_list` which contains only the direct calls.
3. This should work on python 3.7 according to docs, so I removed the `skipif` decorator.
4. The caveat here is that mypy doesn't like us accessing the attributes that mock added. I didn't find a nice solution yet.

brendongo (approved):

PR #5196, opened 2022-05-11
Automate PR opening/templates

spencerdrak:
This allows auto-creation of the release branch PR, plus automatically creates the template. 

NOTE: The PR templating here gets really weird due to limitations on GitHub's REST API. You can't set a default template, so we left the original template where it is. Additionally, the REST API doesn't let you specify templates, so instead we have to read the file, serialize it as JSON, and send it as the body of the PR. 

PR checklist:

    - will be updated once changes are live
- [x] Changelog is up-to-date - N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release_pr_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release_pr_template.md
@@ -1,0 +1,12 @@
+# Release Checklist
+
+- [ ] Updated CHANGELOG with latest changes
+- [ ] Updated latest release and release date in `CHANGELOG.md`
+- [ ] Tests/CI passes on this Pull Request
+- [ ] Announce on community slack that release is in progress
+- [ ] Build Release Candidate finished successfully (`release` workflow)
+- [ ] Publish Github Release Page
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE/release_pr_template.md
```diff
@@ -0,0 +1,12 @@
+# Release Checklist
```
brendongo, line 1: Does github use the files here for actual PR templates when we open it manually in the UI?

brendongo (approved):

.github/PULL_REQUEST_TEMPLATE/release_pr_template.md
spencerdrak: _Sometimes._ The stuff in there can be automatically put in via the `template` query param, but that query param is only respected when using the "web" API. The REST API ignores it, and if you use the web api, it brings you to the "finishing editing this PR and click submit screen", so not useful for our purposes. That's why `.github/pull_request_template.md` still exists, because that's the default one to use when others aren't specified.

PR #5213, opened 2022-05-13
Remove equivalence_eq from the semgrep-core testsuite

aryx:
Those tests were causing regressions with DeepSemgrep because
the test code in DeepSemgrep was running the full engine
whereas the test code in semgrep was running just the
Match_patterns check patterns with some special
equivalences about eq. Semgrep itself was not using
those equivalances (which are a deprecated feature).

test plan:
make test

PR checklist:

- [x] Changelog is up-to-date


aryx:
I've updated semgrep-docs too: https://github.com/returntocorp/semgrep-docs/pull/536

PR #5214, opened 2022-05-13
AST_to_IL: Translate if-then-else as expressions

### LINKED ISSUE #4965: Taint mode does not handle conditional assignment in Ruby correctly
**Describe the bug**
Ruby supports conditional assignment:

```ruby
foo = if hash[:key]
        hash[:key]
      else
        default_value
      end
```

**To Reproduce**
A minimal example is here: https://semgrep.dev/s/ALK6
This functionality is required to port a Brakeman rule correctly based on this set of test cases: https://semgrep.dev/s/returntocorp:check-symbol-dos

**Expected behavior**
Taint should propagate through conditional assignment.

**Screenshots**
N/A

- [x] P1: important to fix or quite annoying - this is blocking completion of a Brakeman check for Ruby < 2.2 / Rails < 5.0

**Environment**
semgrep.dev

**Use case**
Porting Brakeman rules, Ruby behavior correctness

IagoAbal:
We supported if-then-else statements, but not expressions as in OCaml,
Ruby, etc.

Closes #PA-1118
Closes #4965

test plan:
make test # test included

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -798,12 +798,30 @@ and stmt_expr env ?e_gen st =
   match st.G.s with
   | G.ExprStmt (eorig, _) -> expr env eorig
   | G.If (tok, cond, st1, opt_st2) ->
+      (* if cond then e1 else e2
+       * -->
+       * if cond {
+       *   tmp = e1;
+       * }
+       * else {
+       *   tmp = e2;
+       * }
+       * tmp
+       *
+       * TODO: Look at RIL (used by Diamondblack Ruby) for insiration,
+       *       see https://www.cs.umd.edu/~mwh/papers/ril.pdf.
+       *)
       let ss, e' = cond_with_pre_stmts env cond in
       let e1 = stmt_expr env st1 in
       let e2 =
         match opt_st2 with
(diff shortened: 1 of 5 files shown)


aryx:
FYI the ruby parser in pfff comes from the diamondruby project, which defined also a Ruby IL (see pfff/lang_ruby/analyze/il_ruby.ml) that does some of those transformations. FYI here was the paper about it: https://www.cs.umd.edu/~mwh/papers/ril.pdf

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -796,6 +789,41 @@ and record env ((_tok, origfields, _) as record_def) =
...
 
+and stmt_expr env ?e_gen st =
+  let todo () =
+    match e_gen with
+    | None -> todo (G.E (G.e (G.StmtExpr st)))
+    | Some e_gen -> todo (G.E e_gen)
+  in
+  match st.G.s with
+  | G.ExprStmt (eorig, _) -> expr env eorig
+  | G.If (tok, cond, st1, opt_st2) ->
```
aryx, line 800: again maybe a comment like   IF <cond> <then> <else> --> ???
to get an overview of the transformation scheme.

aryx (approved):

PR #5217, opened 2022-05-13
fix: Include fixed_lines field in JSON output when --autofix --dryrun

mmcqd:
Fixes a regression where this field was being left out of the output JSON

PR checklist:

- [x] Changelog is up-to-date


aryx (approved):

aryx:
Weird that it was not in the snapshot before

mmcqd:
> Weird that it was not in the snapshot before

it's because those results were generated without `--autofix --dryrun`, so none of the autofix related code got run at all.

PR #5233, opened 2022-05-17
Add info to version.py

mjambon:
PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/app/version.py
+++ b/semgrep/semgrep/app/version.py
@@ -1,3 +1,11 @@
+"""Ask the Semgrep App server about the latest Semgrep version
+
+This module is for pinging the app to ask for the latest Semgrep release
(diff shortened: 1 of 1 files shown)


IagoAbal (approved):

semgrep/semgrep/app/version.py
```diff
@@ -1,3 +1,11 @@
+"""Ask the Semgrep App server about the latest Semgrep version
+
+This module is for pinging the app to ask for the latest Semgrep release
+so we can print a message prompting the user to upgrade if they have
+an outdated version.
+"""
+# TODO: for predictable test output, add a flag to avoid making actual
```
underyx, line 7: --disable-version-check is supposed to do that I think for this file specifically, but not for other requests

Issue #3539, opened 2021-07-16
[Internal] break mutual dependency between semgrep CI and semgrep-rules CI


aryx:
As martin wrote:
We have a CI deadlock because:

semgrep CI tests against the main branch of semgrep-rules, and
semgrep-rules CI tests against the main branch of semgrep
This would be solved by using a git submodule in semgrep pointing to the correct branch of semgrep-rules.

See for example this semgrep PR https://github.com/returntocorp/semgrep/pull/3444 mutually dependent on
this semgrep-rules PR https://github.com/returntocorp/semgrep-rules/pull/1362#event-5029680119 where currently both fails in CI because they are mutually dependent.

aryx:
This is how we handled in the past, but it was not ideal:

"This mutual dependency issue happens rarely, so usually what I do is that I make the PR to semgrep, and in the test plan I show how a modified semgrep-rules pass with the modified semgrep
then I force merge the semgrep PR if only the semgrep-rules CI job fails
then I wait an hour that the docker develop image has been updated
and I make the PR to semgrep-rules"

aryx:
As suggested by martin, it would be better to make semgrep-rules a submodule and remove the test that rely on the online version of semgrep-rules and instead use the rules in the submodule.

aryx:
cc @mschwager @brendongo @minusworld

aryx:
:wave:

aryx:
One thing we can do during the release process, as Iago suggested, is to update the semgrep-rules submodule to
use the latest semgrep-rules. We can also run a check to transform certain todoruleid: in regular ruleid: when the engine
recognizes them correctly.

minusworld:
I'm also happy to work with you all on this - sorry about all the trouble from last week.

From our standpoint, our main desire is that rules that make it all the way to registry are able to run on the latest released version of Semgrep that's in all the package managers. i.e., we don't want someone to try and run a rule that won't work on the most recently available version. This is what `test-latest` does on semgrep-rules, but is also what caused the issue last week.

If a submodule of semgrep-rules sounds like a huge pain, maybe now is a decent time to include a version field in semgrep rules.

In any case, I'm happy to lend help to resolve this.

minusworld:
semgrep-rules now officially releases off of a release branch. This means that changes can be made to `develop` and will be reflected in semgrep's tests without actually releasing rules that use unreleased features. Does this sound like it resolves this issue?

aryx:
[...]

PR #5241, opened 2022-05-17
feat!: Make dot access ellipsis match field access

### LINKED ISSUE #5231: TS/JS - dot ellipsis dot should also match field access chain (not just method calls chain)
**Describe the bug**
Probably because of lambdas' but the current assignment does not work with arbitrary ellipsis, which is very important to capture values where a value may be assigned. 

**To Reproduce**
Does not work:
https://semgrep.dev/s/90Dy
Works:
https://semgrep.dev/s/r3OJ

**Expected behavior**
Should capture on Arbitrary ellipsis 

**Use case**
TS/React to 🚀 🌔

nmote:
Previously, `foo. ... .bar` would not match `foo.baz.bar`, but would
match `foo.baz().bar`. I found this confusing, and it seems that I am
not alone (#5231).

Test plan:

Automated tests. Before this change, the following examples in the added
test did not match:

```
foo.baz.bar;
foo.a.b.c.bar;
foo.a.b().c.bar;
```

Now, they all do. In addition, there is a new match in the changed Hack
test.

Also ran `semgrep --test --strict --test-ignore-todo --quiet $PWD`  after making the change in https://github.com/returntocorp/semgrep-rules/tree/nmote/dots-field-chaining and it passed (that was previously the command that failed).

Fixes #5231

PR checklist:

- [x] Changelog is up-to-date


nmote:
I have two questions for reviewers:
* This is showing a 5.8% average regression in the benchmarks I noticed this because the previous commit I pushed had a 7+% regression, which triggered a failure. Two recent commits on `develop` have a 2-3% regression, so I suspect that this commit does in fact come with a performance penalty, probably in the vicinity of 3-5%. It makes sense, since it's doing more work when a rule contains a dot access ellipsis, but I wanted to call this out to reviewers and make sure that this is acceptable.
* This changes the output of the `check-send-file.rb` test case in the `semgrep-rules` repo. This looks legit to me, but I'm not sure how to proceed. Should we just merge this despite the failure and then update the `semgrep-rules` repo separately?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -126,15 +126,21 @@ let rec obj_and_method_calls_of_expr e =
   match e.G.e with
```
aryx, line 126: maybe rename obj_and_dot_accesses_of_expr?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -126,15 +126,21 @@ let rec obj_and_method_calls_of_expr e =
...
   | B.Call ({ e = B.DotAccess (e, tok, fld); _ }, args) ->
       let o, xs = obj_and_method_calls_of_expr e in
-      (o, (fld, tok, args) :: xs)
+      (o, (fld, tok, Some args) :: xs)
+  | B.DotAccess (e, tok, fld) ->
+      let o, xs = obj_and_method_calls_of_expr e in
+      (o, (fld, tok, None) :: xs)
   | _ -> (e, [])
 
 let rec expr_of_obj_and_method_calls (obj, xs) =
```
aryx: maybe rename expr_of_obj_and_dot_accesses?

semgrep-core/tests/js/dot_access_ellipsis.js
```diff
@@ -0,0 +1,14 @@
+// MATCH:
```
aryx: there's some kinds of convention on how to name those tests (you can see generate_cheatsheet.py for some hints).
but basically the beginning of the filename should repreresent a big feature of semgrep, e.g., dots_... metavar_... equivalence_... so here you probably want dots_field_chaining.js

aryx (approved):

aryx:
regarding perf, it's not the only PR with some issues, so I don't think it's due to this PR; there's a little more job done, but not a lot.

regarding semgrep-rules, the best is to run make tests in semgrep-rules on a modified branch, with an updated semgrep-core in your PATH (do make install), to check that everything works fine.

aryx:
It's a known problem: https://github.com/returntocorp/semgrep/issues/3539
I actually created a thread on slack to try to make semgrep-rules a submodule in semgrep-core/tests/
so we can be more sure that everything works together.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -126,15 +126,21 @@ let rec obj_and_method_calls_of_expr e =
   match e.G.e with
   | B.Call ({ e = B.DotAccess (e, tok, fld); _ }, args) ->
       let o, xs = obj_and_method_calls_of_expr e in
-      (o, (fld, tok, args) :: xs)
+      (o, (fld, tok, Some args) :: xs)
+  | B.DotAccess (e, tok, fld) ->
```
aryx, line 130: nice, I wrote the original code, and I'm surprised the fix was so simple :) I imagined worse.

nmote:
So in the short term, how should I approach the issue with `semgrep-rules`? I've updated the PR description with the test I conducted, but that won't stop our CI from turning red when this PR is merged.

aryx:
I'll try to add semgrep-rules as a submodule (as well as the test infra to use the tests in there), 
so then you can merge develop, modify locally semgrep-rules with the required modifications, so we can review and get confident that all of this can work together.

nmote:
I've updated the `semgrep-rules` submodule to my branch with the updated test, and `make test` passes in `semgrep-core` for me locally (it didn't before).

aryx:
Great! What I would do is open a PR at the same time with this branch you have in semgrep-rules to the semgrep-rules repo (and reference this PR here to say it's ok, so we get also a review from seceng on whether this looks legitimate).

aryx:
I'll do it.

PR #5243, opened 2022-05-17
SCA: support Gradle and Poetry lockfiles

mmcqd:
This PR:
1. Adds support for `gradle.lockfile` and `poetry.lock` files. `poetry.lock` files are scanned as part of the `pypi` namespace. `gradle.lockfiles` are scanned as part of a new `gradle` namespace. It also fixes a small bug where the set of namespaces from a set of rules would contain many duplicates, which did not affect correctness but could affect performance.
2. Fixes a bug where collecting all the namespaces from a rule would result in many duplicates
3. Fixes a bug where `cargo.lock` was used instead of `Cargo.lock`

PR checklist:

- [x] Changelog is up-to-date


mjambon (approved):

aryx:
wow, we start to handle lots of stuff here!

PR #5245, opened 2022-05-17
fix(ci): Prefer GITHUB_HEAD_REF over GITHUB_REF

underyx:
For pull_request_target events, GITHUB_REF is the base branch name.
What's useful for us is the PR ref that triggered the job though.
See https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/meta.py
+++ b/semgrep/semgrep/meta.py
@@ -331,6 +331,31 @@ class GithubMeta(GitMeta):
         pr_title = self.glom_event(T["pull_request"]["title"])
         return str(pr_title) if pr_title else None
 
+    @property
+    def branch(self) -> Optional[str]:
+        """This branch name gets used for tracking issue state over time on the backend.
+
+        The head ref is in GITHUB_HEAD_REF and the base ref is in GITHUB_REF.
+
+        Event name            GITHUB_HEAD_REF -> GITHUB_REF
+        ---------------------------------------------------
+        pull_request        - johnny-patch-1  -> refs/pulls/123/merge
+        pull_request_target - johnny-patch-1  -> main
+        push/schedule/etc.  - <unset>         -> main
+
+        This code originally always sent GITHUB_REF.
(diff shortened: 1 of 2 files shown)


semgrep/semgrep/meta.py
```diff
@@ -339,7 +364,6 @@ def initialize_repo(self) -> None:
     def to_dict(self) -> Dict[str, Any]:
         return {
             **super().to_dict(),
-            "branch": os.getenv("GITHUB_REF"),
```
chmccreery, line 342: We can delete this because the super().to_dict() already has `"branch": self.branch`, right?

chmccreery (approved):
Wowow A+ documentation, thank you!!

semgrep/semgrep/meta.py
underyx: correct!

semgrep/semgrep/meta.py
```diff
@@ -331,6 +331,31 @@ def pr_title(self) -> Optional[str]:
...
+    def branch(self) -> Optional[str]:
+        """This branch name gets used for tracking issue state over time on the backend.
+
+        The head ref is in GITHUB_HEAD_REF and the base ref is in GITHUB_REF.
+
+        Event name            GITHUB_HEAD_REF -> GITHUB_REF
+        ---------------------------------------------------
+        pull_request        - johnny-patch-1  -> refs/pulls/123/merge
+        pull_request_target - johnny-patch-1  -> main
+        push/schedule/etc.  - <unset>         -> main
```
underyx:
```suggestion
        push/schedule/etc.  - <unset>         -> refs/heads/main
```
```suggestion
        pull_request_target - johnny-patch-1  -> refs/heads/main
        push/schedule/etc.  - <unset>         -> refs/heads/main
```

PR #5249, opened 2022-05-18
Experiment: Add support for defining "metatypes" to use in rules

emjin:
This is an experiment based on a conversation with security. We allow
users to pass in a file (like equivalences, but specifically for types)
to define a metatype. When they use this metatype in a rule, we also
check whether any of the metatype's subtypes are present in the code.
In addition, we still check for the metatype, to allow the name to be
reused.

Test plan: [Note: do I need to add programmatic tests? I notice we don't run tests for our other experiments]

Define three files:

```
➜  metatypes cat log_rule.yaml
rules:
- id: secret-log
  pattern: (LoggerType $LOGGER).log(secret);
  message: Don't log secrets!
  languages:
  - java
  severity: ERROR
➜  metatypes cat app.java
public class Test {

    public static void main(String[] args) {
        Logger log1 = new Logger();
        Log4j.Logger log2 = new Log4j.Logger();
        MyLogger log3 = new MyLogger();
        SafeLogger log4 = new SafeLogger();

        String secret = "Keep this secret";

        // MATCH:
        log1.log(secret);
        log1.log("innocuous");

        // TODO
        log2.log(secret);
        log2.log("innocuous");

        // MATCH:
        log3.log(secret);
        log3.log("innocuous");

        // Ok:
        log4.log(secret);
        log4.log("innocuous");
    }

}
➜  metatypes cat metatypes.yaml
types:
- LoggerType:
    - Logger
    - Log4j.Logger
    - MyLogger
```

```
➜  metatypes sc -lang java -rules log_rule.yaml app.java
```

```
➜  metatypes sc -lang java -rules log_rule.yaml app.java -metatypes metatypes.yaml
app.java:11 with rule secret-log
         log1.log(secret);
app.java:17 with rule secret-log
         log3.log(secret);
```

For the semgrep portion:

Run:

```
➜  metatypes semgrep -f log_rule.yaml app.java
Scanning 1 file.

Ran 1 rule on 1 file: 0 findings.
```

Now put the `metatypes.yaml` file in `~/.semgrep`. Rename it to `~/.semgrep/r2c-internal-experiment-metatypes-file.yaml`

Then run:
```
➜  metatypes semgrep -f log_rule.yaml app.java
Scanning 1 file.

Findings:

  app.java 
     secret-log
        Don't log secrets!

         12┆ log1.log(secret);
          ⋮┆----------------------------------------
         20┆ log3.log(secret);

Ran 1 rule on 1 file: 2 findings.
```

PR checklist:

- [ ] Changelog is up-to-date


aryx:
I'd rather use jsonnet for those kinds of stuff, but if seceng think that this special case has values for customers, I'm fine with it.
Maybe it should come with a more general way to "override" the content of some rules. I know people sometimes want to refine the error message of a rule.

emjin:
@aryx It would be great if we could merge this! Then security can test it with community/customers

mjambon (approved):

mjambon:
> Test plan: [Note: do I need to add programmatic tests? I notice we don't run tests for our other experiments]

For what it's worth, I'd treat experimental code like any other piece of code (tests, maintainability, documentation). In particular, it's useful to document experimental features as such rather than having people discover them and using them without knowing their status.

aryx (approved):

emjin:
Makes sense! I'll circle back by mid-June and either add tests + docs or delete the code

IagoAbal:
How is this going to be used in combination with the registry? Is it just for CLI users writing custom rules that they keep locally?

I'm also in favor of jsonnet for this kind of stuff to be honest. We still fall in "YAML traps" nowadays, and jsonnet has a pretty decent syntax.

I'm nitpicking here but I'm not entirely convinced about the name "meta types". Maybe I'm missing something, but I think this could be used for any kind of identifier. They are more like pattern templates/parameters. Preferably there should be a special syntax for those parameters, so that when you read a pattern it's clear what is what.

emjin:
The idea was that paying users could make their rules far more specific/powerful without having to understand semgrep's syntax at all. They could just enter their types for each category (I think I should actually have called them critical types), and the rules would become tailored to them. Ideally, the app would walk them through the types to add during onboarding. 

I don't know if that approach would actually make sense though, because it depends on the kinds of rules that would benefit from this. That's why I just made this a hack and have been slow about adding tests/documentation; I'm still not certain it's worth it, and security hasn't had time to test it yet. It's on their roadmap for this quarter, so I will re-evalute after that. (I know I said that last time but I mean it this time 😅)

PR #5258, opened 2022-05-18
perf: Speed up rule deduplication

underyx:
PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/rule.py
+++ b/semgrep/semgrep/rule.py
@@ -1,6 +1,5 @@
 import hashlib
 import json
-from copy import deepcopy
 from typing import Any
 from typing import cast
 from typing import Dict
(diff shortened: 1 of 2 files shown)


semgrep/semgrep/rule.py
```diff
@@ -247,6 +246,7 @@ def has_runnable_rule(d: Dict[str, Any]) -> bool:
 
 
 def rule_without_metadata(rule: Rule) -> Rule:
-    rule2 = deepcopy(rule)
```
nbrahms, line 250: what are the implications of now doing a shallow copy?

underyx: Nothing! We only ever use this to create a read-only dictionary key.

underyx: I'd like to make `Rule` a frozen attrs class to make sure it's immutable. But that's much more refactoring work.

ievans (approved):

emjin:
Oops, this was me. Thanks for fixing it!

PR #5261, opened 2022-05-19
perf: Speed up Python layer in a dozen ways

underyx:
PR checklist:

- [ ] Changelog is up-to-date


semgrep/semgrep/core_runner.py
```diff
@@ -561,6 +561,7 @@ def _run_rules_direct_to_semgrep_core(
         rule_file_name = (
             RULE_SAVE_FILE
             if dump_command_for_core
+            # we dump JSON formatted rules to .yaml because core's JSON parser can't parse them
```
aryx: what is the semgrep-core error?

aryx: I think semgrep-core can parse rules as .yaml, or .json (and also .jsonnet)

underyx: @aryx letting core parse as JSON causes a pattern parse error. My guess would be that backslash escaping in strings is not parsed correctly by the JSON parser.

aryx: Do you have an example showing the issue. We actuallu use @mjambon json parser

underyx: I think if you download p/default and dump it as JSON, it'll just fail and you can debug from there

mjambon: @underyx can you please post the json data that causes the parsing error?

underyx: [rules.json.zip](https://github.com/returntocorp/semgrep/files/8733776/rules.json.zip)

This is it:

```
❯ semgrep-core -rules /tmp/rules.json
/tmp/rules.json:205:21: "Pattern parse error": Invalid pattern for Python
```

underyx: That line is

```
"pattern": "itsdangerous.jws.JSONWebSignatureSerializer(..., algorithm_name=\"none\", ...)"
```

which is why I assumed that core's JSON parser doesn't parse `\"` within strings correctly. The same file works if I rename it to `.yaml` though.

mjambon: thanks, @underyx.

@pad are we not using the yaml parser for this, not a json-only parser? It's very suspicious that it would not parse this correctly, though. This is line 205:
```
          "pattern": "itsdangerous.jws.JSONWebSignatureSerializer(..., algorithm_name=\"none\", ...)"
```

mjambon: Here's a smaller example with the same error:
```
{
  "rules": [
    {
      "id": "insecure-itsdangerous-use",
      "message": "The Python third-party 'itsdangerous' module used with 'none' signing algorithm",
      "languages": [
        "python"
      ],
      "severity": "WARNING",
      "metadata": {
        "source_rule_url": "https://github.com/dlint-py/dlint/blob/master/docs/linters/DUO137.md",
        "category": "security",
        "technology": [
          "itsdangerous"
        ],
        "license": "Commons Clause License Condition v1.0[LGPL-2.1-only]"
      },
      "pattern-either": [
        {
          "pattern": "itsdangerous.jws.JSONWebSignatureSerializer(..., algorithm_name=\"none\", ...)"
        },
        {
          "pattern": "itsdangerous.JSONWebSignatureSerializer(..., algorithm_name=\"none\", ...)"
        }
      ]
    }
  ]
}
```

```
semgrep-core -rules rules.json
rules.json:20:21: "Pattern parse error": Invalid pattern for Python
```

mjambon: This works:
```
{
  "rules": [
    {
      "id": "x",
      "message": "",
      "languages": ["python"],
      "severity": "WARNING",
      "pattern": "'hello'"
    }
  ]
}
```
->
```
```

The problem is with double-quoted strings:
```
{
  "rules": [
    {
      "id": "x",
      "message": "",
      "languages": ["python"],
      "severity": "WARNING",
      "pattern": "\"hello\""
    }
  ]
}
```
->
```
rules.json:8:17: "Pattern parse error": Invalid pattern for Python
```

mjambon: Two findings:
1. OCaml's yaml package parses `"\"hello\""` correctly.
2. Passing the pattern directly to semgrep-core with `-e` also works as expected.

1:
```
utop # Yaml.of_string {|{
  "rules": [
    {
      "id": "x",
      "message": "",
      "languages": ["python"],
      "severity": "WARNING",
      "pattern": "\"hello\""
    }
  ]
}|};;
- : Yaml.value Yaml.res =
Ok
 (`O
    [("rules",
      `A
        [`O
           [("id", `String "x"); ("message", `String "");
            ("languages", `A [`String "python"]);
            ("severity", `String "WARNING"); ("pattern", `String "\"hello\"")]])])

```

2:
```
$ semgrep-core -lang python -e '"hello"' <(echo '"hello"')
/tmp/semgrep-core-552fe4-63:1
 "hello"
```

I haven't looked at the semgrep-core source code yet.

mjambon: Same error with more logging:
```
rules.json:8:17: "Pattern parse error": Invalid pattern for Python:
--- pattern ---
\"hello\"
--- end pattern ---
Pattern error: Parse_info.Lexical_error (unrecognized symbol: \, /tmp/tmp-2268658-8712a8.py:1:0 "\\")
```

mjambon: I located the bug, in the custom yaml parser where the string literal should be evaluated but isn't. I'll file a proper bug report.

mjambon: I created https://github.com/returntocorp/semgrep/issues/5268 and https://github.com/returntocorp/semgrep/issues/5269

aryx:
nice!

semgrep/semgrep/core_runner.py
aryx: BTW @underyx when you see those kinds of bugs in semgrep-core, it's better to tell us than writing some comments in the code (I happened to review this diff by chance, but otherwise I would have missed the bug we have in semgrep-core).

underyx: @aryx I was gonna, don't worry

aryx: Got it. Btw i've tried to fix it so maybe we can retry this json thing at some point.

semgrep/semgrep/core_runner.py
```diff
@@ -479,9 +479,8 @@ def _add_match_times(
         profiling_data: ProfilingData,
         timing: core.CoreTiming,
     ) -> None:
-        targets = [Path(t.path) for t in timing.targets]
+        [Path(t.path) for t in timing.targets]
```
nbrahms: Probable bug: does this code do anything?

semgrep/semgrep/profiling.py
```diff
@@ -14,43 +14,24 @@ class Semgrep_run(NamedTuple):
...
-            for rule in rules
-            for target in targets
-        }
-
-        self._rule_match_times = {rule: 0.0 for rule in rules}
-        self._rule_bytes_scanned = {rule: 0 for rule in rules}
-
-        self._file_match_times = {target: 0.0 for target in targets}
-        self._file_num_times_scanned = {target: 0 for target in targets}
+        self._file_parse_time: Dict[Path, float] = defaultdict(float)
```
nbrahms, line 24: nice 👌

semgrep/semgrep/metrics.py
```diff
@@ -287,7 +287,6 @@ def send(self) -> None:
                 # TODO: weird, we get a mypy error only in CI, and only
                 # after I've enabled submodules for pre-commit
                 headers=cast(Dict[str, str], {"Authorization": None}),
```
nbrahms: Probable regression:
IIRC, this timeout was here to prevent Semgrep from appearing "hung" in the user CLI.

I think, instead, we want `retries=1`.

Hopefully we have a commit message that accompanies this original change.

.semgrepignore
```diff
@@ -24,10 +24,6 @@
 # To ignore a file with a literal ':' character, escape it with
 # a backslash, e.g. "\:foo".
 
-# Ignore git items
```
nbrahms, line 27: Please change: Can you add a commit message explaining this change?

My guess is that, because we use `git ls-tree` or the like, this is unnecessary, but it would be nice to be very clear in the commit message.

semgrep/semgrep/ignores.py
```diff
@@ -51,8 +51,14 @@ def _survives(self, path: Path) -> bool:
         """
         Determines if a single Path survives the ignore filter.
         """
+        path_is_dir = path.is_dir()
+        path_is_relative_to_base = path_is_relative_to(path, self.base_path)
+        if path_is_relative_to_base:
+            path_relative_to_base = str(path.relative_to(self.base_path))
+        else:
+            path_relative_to_base = ""
```
nbrahms, line 58: Non-blocking suggestion:
Should this just be `return False`?

semgrep/semgrep/ignores.py
```diff
@@ -1,14 +1,16 @@
 import fnmatch
 import os
 import re
+from functools import lru_cache
```
nbrahms, line 4: Need info: We should double-check that these caching changes don't break the backend runner service (maybe we just need to make the caching configurable?)

semgrep/semgrep/ignores.py
```diff
@@ -89,15 +89,15 @@ def _survives(self, path: Path) -> bool:
...
 
-    def filter_paths(self, *, candidates: Iterable[Path]) -> FilteredFiles:
-        kept, removed = partition(
-            candidates,
-            lambda path: path.exists()
-            and (
-                self._survives(path.absolute())
-                or path.absolute().samefile(self.base_path)
-            ),
+    @lru_cache(maxsize=100_000)
```
nbrahms: Non-blocking question:
This maxsize seems effectively infinite, in which case, wouldn't we rather just use an in-memory data-structure?

nbrahms (changes requested):
See some potential issues to address

semgrep/semgrep/metrics.py
underyx: @nbrahms we already have a default timeout of 30s that comes from the app session. I think the timeout of 2s might be excessively low and might cause missed metrics (considering that most of my request timings were around 2.5s, and that's from our gigabit office internet right next to the AWS datacenter.)

I'd be ok with a middle ground timeout, but since at this point the results have been printed already, I thought it'd be fine to wait as long as for any other request, and thus end up with fewer special cases.

semgrep/semgrep/ignores.py
underyx: `@lru_cache` itself uses an in-memory data-structure; if you're suggesting implementing custom code for caching in a dict we maintain ourselves, I just prefer not doing that as this is less code. I could also set maxsize to `None` but I like setting a specific number to communicate "we thought about this and yes, we can cache all values that will ever show up"

semgrep/semgrep/metrics.py
nbrahms: We should find the original history; the intention was to prefer dropping metrics over blocking the command line for more than 2 seconds. (cc. @ievans ; I'm not sure what the ideal number is, but intuitively it should be far lower than 30 s)

semgrep/semgrep/ignores.py
nbrahms: Right, the trap one gets into with a large size is that, when the access pattern has on the order of 100k new values before a cache hit, that the performance can drop once the list size crosses the threshold.

When I've worked with these cache patterns before, it's been more useful to, e.g., sort the input list, and use a much smaller cache size.

semgrep/semgrep/target_manager.py
```diff
@@ -426,6 +426,23 @@ def files(self) -> FrozenSet[Path]:
...
+@lru_cache(maxsize=100_000)
+def get_shebang_line(path: Path) -> Optional[str]:
+    if not os.access(str(path), os.X_OK | os.R_OK):
+        return None
+
+    with path.open() as f:
+        return f.readline(MAX_CHARS_TO_READ_FOR_SHEBANG).rstrip()
+
+
+@lru_cache(maxsize=10_000)
```
nbrahms: Question: why is this 10k and all the others are 100k?

nbrahms:
> 📸 The pytest shapshots changed in your PR. Please carefully review these changes and make sure they are intended:
> 
> 1. Review the changes at [4eff950](https://github.com/returntocorp/semgrep/commit/4eff950ba09571a0f8e81df107acf7af9cb3e979)
> 2. Accept the new snapshots with
>    ```
>    git fetch origin && git cherry-pick 4eff950b && git push
>    ```

Just to check my sanity, there should be no snapshot changes, right?

semgrep/semgrep/ignores.py
underyx: Although in this case if someone is scanning 100,000 files surely they don't care about the couple hundred milliseconds of overhead added by caching-and-evicting compared to the actual time spent scanning? And especially compared to the slowdown caused by not having any caching.

I also can't think of a way of pre-sorting the entries with the biggest time loss.

semgrep/semgrep/ignores.py
underyx: Good point. Many of them use the `self` instance as a cache key btw so that should make them isolated per scan. But we should test this.

semgrep/semgrep/ignores.py
underyx: Uhh, I think not? The code below seems like it can return true even if path is not relative to base. But it's late so I gotta re-check next week.

semgrep/semgrep/metrics.py
underyx: what if we launch a new `semgrep` subprocess to send off the metrics so that the scan process can exit immediately?

semgrep/semgrep/core_runner.py
underyx: thanks for pointing this out! pre-commit removed the variable assigned as it became unused

underyx:
> Just to check my sanity, there should be no snapshot changes, right?

Correct. That change was expected from a commit that I split off into a separate PR. See discussion there: https://github.com/returntocorp/semgrep/pull/5272

semgrep/semgrep/target_manager.py
underyx: As I was profiling on a real life project, this function had an order of magnitude fewer calls than others. I tried to set all cache limits to be around 100x the unique calls I saw on returntocorp/semgrep, so that we cache everything but have a safety check in place in case there's an unexpected combinatorical explosion.

semgrep/semgrep/ignores.py
underyx: @nbrahms how can I test this? 🤔

.semgrepignore
underyx: done!

semgrep/semgrep/metrics.py
nbrahms: yup, just need that daemonized

.semgrepignore
underyx: done!

semgrep/semgrep/target_manager.py
nbrahms: Great, can we leave a comment to this effect?

nbrahms (approved):
Approved pending figuring out the metrics blocking / retries

PR #5266, opened 2022-05-19
tainting: Allow variables to be sanitized by side-effect

IagoAbal:
Closes PA-1360

test plan:
make test # one test added

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,18 +16,18 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   safe data, this was not recognized by the taint engine. Also, if `taint(x)`
   occurred inside e.g. an `if` block, any occurrence of `x` outside that block
   was not considered tainted. Now, if you specify that the code variable itself
-  is a taint source, the taint engine will handle this as expected and it will
-  not suffer from the aforementioned limitations. We believe that this change
-  should not break existing taint rules, but please report any regressions that
-  you may find.
+  is a taint source (using `focus-metavaraible`), the taint engine will handle
+  this as expected, and it will not suffer from the aforementioned limitations.
+  We believe that this change should not break existing taint rules, but please
+  report any regressions that you may find.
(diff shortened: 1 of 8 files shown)


semgrep-core/tests/OTHER/rules/taint_sanitizer_var.py
```diff
@@ -0,0 +1,11 @@
+def ok():
+    x = source
+    # `x` is sanitized by side effect!
+    sanitize(x)
```
aryx, line 4: would be good to also add test examples that shows the limitations you mention in the changelog no longer applies here.
I guess the old solution was to do  pattern-inside: sanitize($X) AND pattern: $X 
and that was sanitizing every $X, so some code with some if and reassign to x later to source would
help show that the limitations no longer applies with focus-metavariable.

aryx (approved):

CHANGELOG.md
```diff
@@ -20,6 +20,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   not suffer from the aforementioned limitations. We believe that this change
   should not break existing taint rules, but please report any regressions that
   you may find.
+- taint-mode: Let's say that e.g. `sanitize(x)` sanitizes `x` by side-effect.
+  Previously, we had to rely on a trick that declared that _any_ occurrence of
+  `x` inside `sanitize(x); ...` was sanitized. If `x` later overwritten with
+  tainted data, the taint engine would still regard `x` as safe. Now, if you
+  specify that the code variable itself is sanitized, the taint engine will
```
aryx: you could add "(thanks to the new 'focus-metavariable:' feature introduced in 0.9??)"

aryx (approved):

semgrep-core/tests/OTHER/rules/taint_sanitizer_var.py
IagoAbal: Good point!

PR #5270, opened 2022-05-19
perf: Request registry rules as JSON instead of YAML

underyx:
PR checklist:

- [x] Changelog is up-to-date
- [x] Merge https://github.com/returntocorp/semgrep-app/pull/5242

### DIFF
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -15,7 +15,6 @@ from typing import Sequence
 from typing import Tuple
 
 import requests
-from requests import JSONDecodeError
 from ruamel.yaml import YAML
 from ruamel.yaml import YAMLError
 
@@ -460,7 +459,7 @@ def parse_config_string(
         ]
         # we pretend it came from YAML so we can keep later code simple
         return {config_id: YamlTree.wrap({"rules": rule_definitions}, EmptySpan)}
-    except JSONDecodeError:
+    except json.decoder.JSONDecodeError:
         pass
     try:
         data = parse_yaml_preserve_spans(contents, filename)
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -1,3 +1,4 @@
+import json
 import os
 import time
 from collections import OrderedDict
@@ -14,6 +15,7 @@ from typing import Sequence
 from typing import Tuple
 
 import requests
+from requests import JSONDecodeError
 from ruamel.yaml import YAML
 from ruamel.yaml import YAMLError
 
@@ -32,6 +34,7 @@ from semgrep.error import SemgrepError
 from semgrep.error import UNPARSEABLE_YAML_EXIT_CODE
 from semgrep.rule import Rule
 from semgrep.rule import rule_without_metadata
+from semgrep.rule_lang import EmptySpan
 from semgrep.rule_lang import EmptyYamlException
 from semgrep.rule_lang import parse_yaml_preserve_spans
 from semgrep.rule_lang import Span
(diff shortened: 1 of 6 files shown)


semgrep/semgrep/config_resolver.py
```diff
@@ -439,6 +438,16 @@ def parse_config_string(
             f"Empty configuration file {filename}",
             code=UNPARSEABLE_YAML_EXIT_CODE,
         )
+    try:
+        rule_definitions = [
+            rule
+            for rule_model in json.loads(contents)
+            for rule in rule_model["definition"]["rules"]
+        ]
+        # we pretend it came from YAML so we can keep later code simple
```
underyx, line 447: Despite this, parsing local YAML rules is now broken somehow. Will have to come back here and fix that.

semgrep/semgrep/config_resolver.py
```diff
@@ -195,22 +197,12 @@ def _load_config_from_local_path(self) -> Dict[str, YamlTree]:
...
-                "text/x-yaml",
-                "text/yaml",
-                "text/vnd.yaml",
-            ]
-            if content_type and any(ct in content_type for ct in yaml_types):
-                return r.content.decode("utf-8", errors="replace")
-            else:
-                raise SemgrepError(
-                    f"unknown content-type: {content_type} returned by config url: {self._config_path}. Can not parse"
-                )
```
underyx, line 213: Removed this because we later just do

```
try:
  # parse as json
except:
  pass
try:
  # parse as yaml
except:
  raise
```

anyway so type assertions here seem unnecessary.

semgrep/semgrep/config_resolver.py
```diff
@@ -441,31 +433,34 @@ def parse_config_at_path(
...
 ) -> Dict[str, YamlTree]:
     if not contents:
         raise SemgrepError(
-            f"Empty configuration file {filename}",
-            code=UNPARSEABLE_YAML_EXIT_CODE,
+            f"Empty configuration file {filename}", code=UNPARSEABLE_YAML_EXIT_CODE
         )
+    try:
+        # we pretend it came from YAML so we can keep later code simple
+        data = YamlTree.wrap(json.loads(contents), EmptySpan)
```
underyx, line 448: As far as understand passing `EmptySpan` (which I took from `Rule.from_json`) will just make this tree unusable in pretty error messages?

semgrep/semgrep/rule_lang.py
```diff
@@ -284,7 +284,7 @@ def unroll(self) -> YamlValue:
             return {str(k.unroll()): v.unroll() for k, v in self.value.items()}
         elif isinstance(self.value, YamlTree):
             return self.value.unroll()
-        elif isinstance(self.value, str) or isinstance(self.value, int):
+        elif isinstance(self.value, (str, int)) or self.value is None:
```
brendongo, line 287: Is value nullable?

semgrep/semgrep/config_resolver.py
brendongo: Should we add a test? Maybe it blows up?

brendongo (approved):

semgrep/semgrep/rule_lang.py
underyx: well according to the JSON and YAML specs, yes, cause `null` is a legit value :D

we just happened to have no rules that include a `null` value so far (and this is still the case)

but a rules with

```yaml
metadata:
  foo: null
```

would make semgrep crash today I think

semgrep/semgrep/config_resolver.py
underyx: well we have lots of E2E tests that already test it, no? 🤔

PR #5271, opened 2022-05-19
perf: Cache registry responses

underyx:
PR checklist:

- [ ] Changelog is up-to-date
- [ ] Merge https://github.com/returntocorp/semgrep-app/pull/5242

### DIFF
--- a/semgrep/semgrep/app/session.py
+++ b/semgrep/semgrep/app/session.py
@@ -1,14 +1,17 @@
 import os
+from datetime import timedelta
 from typing import Any
 from typing import Optional
 from typing import Set
 
 import requests
+import requests_cache
 import urllib3
 from attrs import define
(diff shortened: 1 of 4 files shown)


semgrep/semgrep/app/session.py
```diff
@@ -43,7 +46,10 @@ def __str__(self) -> str:
         return result
 
 
-class AppSession(requests.Session):
+REQUESTS_CACHE_PATH = USER_DATA_FOLDER / "requests-cache.sqlite"
```
underyx, line 49: TODOs with the whole caching concept:

- [ ] Make it possible to disable cache
- [ ] Make it possible to change cache path (maybe USER_DATA_FOLDER is already configurable enough?)
- [ ] Clean up expired entries when running

PR #5272, opened 2022-05-20
perf: Pass rules to core as JSON

underyx:
PR checklist:

- [ ] Changelog is up-to-date


semgrep/semgrep/core_runner.py
```diff
@@ -578,8 +579,7 @@ def _run_rules_direct_to_semgrep_core(
             target_file.write(json.dumps(plan.to_json()))
             target_file.flush()
 
-            yaml = YAML()
-            yaml.dump({"rules": [rule._raw for rule in rules]}, rule_file)
+            rule_file.write(json.dumps({"rules": [rule._raw for rule in rules]}))
```
underyx: This changes the pattern parse error locations/spans reported from core. But that might be okay, considering that with the YAML load-dump round trip we've probably already been changing the line numbers inadvertently. It's especially okay for registry rules where the user doesn't see the original rule file anyway.

So maybe we we pretty-format the JSON with indentation, and the patterns are still readable when we print it out in an error message, it's okay to change the source YAML to JSON here?

underyx:
CleanShot 2022-05-24 at 12 56 14@2x

CLI output remains the same in the invalid pattern case

underyx:
CleanShot 2022-05-24 at 13 06 42@2x

seems like the error ranges aren't used on the playground either

brendongo (approved):
I'm a little suspicious that we don't use the ranges to highlight the exact line of the error in the rule. If so, there's a lot of code we can delete now :D

underyx:
CleanShot 2022-05-24 at 19 33 09@2x

timed these things with p/default, turns out sort_keys=True is so cheap we might as well make the tests deterministic with it

Issue #5269, opened 2022-05-19
Expose unescaped string literals in generic AST


mjambon:
**Is your feature request related to a problem? Please describe.**

The analysis of string contents in all the target languages supported by Semgrep is wrong whenever the strings contain escape sequences. This is almost the case with regexps, which we'd like to analyze correctly. Common examples include `\"` and `\\` which often stand for `"` and `\`.

**Describe the solution you'd like**

Have each language parser provide an unescaped/evaluated version of string literals, in addition to the original escaped contents.
* Source: `"\"hello"`
* Escaped contents: `\"hello`
* Unescaped contents: `"hello`

**Describe alternatives you've considered**

none

**Use case**

* parsing Semgrep rules formatted as JSON (#5268)
* analysis of the contents of string literals in various languages, such as regexps

**Additional context**

See proposed change to the generic AST:

```ocaml
  (* String literals:
     The token includes the quotes (if any) but the string value excludes them.
     The value is the escaped string content. For example,
     The escaped content of the Python string literal '\\(\\)' is \\(\\).
     The unescaped content would be \(\).
     TODO: use bracket instead of wrap, ', ", or even """
     TODO: expose the unescaped contents if known, so that we could analyze
     string contents correctly.
     An incremental change could be:

     | String of (string * string option) bracket
                  ^^^^^^   ^^^^^^
                  escaped  unescaped
  *)
```

https://github.com/returntocorp/semgrep/blob/f208b3acd0792123ab9dae30f46ee96dbb8b3d60/semgrep-core/src/core/ast/AST_generic.ml#L556-L570

Perhaps a better representation would use a record and two required fields:

```
| String of { esc: string; unesc: string } bracket
```

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1365/expose-unescaped-string-literals-in-generic-ast. _Note_: this link is for r2c use only and is not accessible publicly.

PR #5273, opened 2022-05-20
Print details when bad pattern is found in rule

mjambon:
This shows more details when a bad pattern is encountered. Today, it revealed a bug in the YAML parser. The bug would be fixed by #5269 (which is a bit of work).

Test plan:
```
$ cat rules.json
{
  "rules": [
    {
      "id": "x",
      "message": "",
      "languages": ["python"],
      "severity": "WARNING",
      "pattern": "\"hello\""
    }
  ]
}
rules.json:8:17: "Pattern parse error": Invalid pattern for Python:
--- pattern ---
\"hello\"
--- end pattern ---
Pattern error: Parse_info.Lexical_error (unrecognized symbol: \, /tmp/tmp-2277679-0736a8.py:1:0 "\\")
```

PR checklist:

- [x] Changelog is up-to-date


aryx (approved):

aryx:
need to update the snapshot.

Why we don't have this problem when the rule is stored as YAML? Because YAML does not escape those strings?

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -113,6 +115,7 @@ let generic_to_json env (key : key) ast =
                  match x.G.e with
                  | G.Container
                      (G.Tuple, (_, [ { e = L (String (k, _)); _ }; v ], _)) ->
+                     (* BUG: need to unescape the string k *)
```
aryx: Actually we don't need to do it when the generic AST comes from Yaml_to_generic.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -104,7 +104,9 @@ let generic_to_json env (key : key) ast =
     | G.L (Bool (b, _)) -> J.Bool b
     | G.L (Float (Some f, _)) -> J.Float f
     | G.L (Int (Some i, _)) -> J.Int i
-    | G.L (String (s, _)) -> J.String s
+    | G.L (String (s, _)) ->
+        (* BUG: need to unescape the string s *)
```
aryx: same, we don't need to do anything here.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -127,7 +130,12 @@ let generic_to_json env (key : key) ast =
 
 let read_string_wrap e =
   match e with
-  | G.L (String (value, t)) -> Some (value, t)
+  | G.L (String (value, t)) ->
+      (* BUG: need to unescape the string s e.g. {|\"x\"|} -> {|"x"|}
```
aryx: same, the bug is not here.

The bug is in Json_to_generic.program which should take an option to request to unescape the string, because we are in the context of parsing a rule in which case we want all the strings to get their "final" value.

aryx (changes requested):

mjambon:
> Why we don't have this problem when the rule is stored as YAML? Because YAML does not escape those strings?

No idea. If we're going to put efforts into improving the YAML parser, we should contribute to the original `yaml` package or fork it for everyone's benefit.

mjambon:
@aryx [yaml_to_generic](https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/src/parsing/other/yaml_to_generic.ml) looks fine to me. Is there any weird reason why it's not converted first to a json-like AST with location info and only later to the generic AST? (rather than depending on the generic AST for parsing ordinary yaml)

semgrep-core/src/parsing/Parse_rule.ml
mjambon: These comments are meant to indicate that when we add a field containing the unescaped string, that's the one we should use here.

mjambon: I updated the comments.

aryx (approved):

PR #5274, opened 2022-05-20
Compile regexps found in rules in multiline mode

### LINKED ISSUE #5280: ^ and $ should match at the beginning and end of each line
`pattern-regex` currently compiles regexps in the default mode rather than the multiline mode. The multiline mode is more suitable to scanning multiline input, as `^` and `$` will then be used to match the beginning and the end of lines, respectively, rather than only the beginning and end of the input file. Matching the beginning and end of input can still be done with `\A` and `\Z`, if desired.

The main issue is that without multiline mode, matching beginning and end of line is expert-level, solutions being `(^|(?<=\n))` (beginning of line) and `($|(?=\n))` (end of line).

Request: switch to multiline mode

Motivation: https://github.com/returntocorp/semgrep/issues/5148#issuecomment-1125583613

mjambon:
With this change, `^` and `$` will now match the beginning and end of each line rather than just the beginning and end of the target file. This breaks backward compatibility but it should be for the best and hopefully will affect very few rules.

Closes #5280

PR checklist:

- [x] Changelog is up-to-date


mjambon:
I'll update the documentation at https://github.com/returntocorp/semgrep-docs/edit/main/docs/writing-rules/rule-syntax.md once this change is approved.

semgrep/tests/public_repos.py
```diff
@@ -271,13 +271,18 @@ def xfail_repo(repo, *, reason=""):
...
     #        reason="MatchTimeout error but happens only in CI",
     #    ),
-    xfail_repo(
-        {
-            "repo": "https://github.com/home-assistant/home-assistant",
-            "languages": ALL_LANGUAGES,
-        },
-        reason="https://github.com/returntocorp/semgrep/issues/5092",
-    ),
+    #
```
aryx: yes I also had weird behavior with this project.

aryx (approved):
not sure what's going on with CI failures but this LGTM.

semgrep/tests/e2e/targets/basic/regex-multiline-mode.txt
```diff
@@ -0,0 +1,8 @@
+hi
+
+# ERROR: test-regex-multiline-mode
```
aryx: you could have used semgrep-core/tests/OTHER/rules/ for this test instead of doing it via a pytest.

aryx (approved):

PR #5275, opened 2022-05-20
[internal] semgrep-core accepts again rules stored in .JSON

### LINKED ISSUE #5268: Fix parsing of quoted strings in rules stored in the JSON format
**Describe the bug**

semgrep-core's YAML parser used to parse rules interprets double-quoted strings incorrectly. Note that JSON is a subset of YAML, therefore semgrep-core should accept pure JSON rules without errors.

[Original comment](https://github.com/returntocorp/semgrep/pull/5261#pullrequestreview-978126917)

**To Reproduce**

```
$ cat rules.json
{
  "rules": [
    {
      "id": "x",
      "message": "",
      "languages": ["python"],
      "severity": "WARNING",
      "pattern": "\"hello\""
    }
  ]
}
rules.json:8:17: "Pattern parse error": Invalid pattern for Python
```

Surprisingly to me, the following variant works:
```
$ cat rules.yaml 
rules:
- id: x
  message: x
  languages: [python]
  severity: WARNING
  pattern: "\"hello\""
$ semgrep-core -lang python -rules rules.yaml
```

**Expected behavior**

no error output, exit 0

**Environment**

any

The `semgrep` wrapper works around the bug by passing rules to semgrep-core in YAML syntax rather than strict JSON, presumably managing to avoid double-quoted strings. It's possible but not certain that this workaround is completely effective. If not, the bug would affect some semgrep rules, resulting in unparseable rules and in particular, unparseable semgrep patterns.

**Use case**

Requires `semgrep` to reformat rules in a special way so as to work around the bug.

aryx:
This closes #5268

test plan:
test file included

PR checklist:

- [x] Changelog is up-to-date


mjambon:
great, thank you!

semgrep-core/src/parsing/pfff/json_to_generic.ml
```diff
@@ -65,13 +65,20 @@ let expr x =
                        | Right t -> G.Ellipsis t |> G.e)
                 in
                 G.Container (G.Dict, (lp, zs, rp)) |> G.e
+            | G.L (G.String (s, t)) when unescape_strings ->
+                (* TODO: use the proper JSON unescaping convention, not
+                 * the OCaml one, which is what Scanf.unescape does.
+                 * Reuse Yojson code?
+                 *)
+                G.L (G.String (Scanf.unescaped s, t)) |> G.e
```
aryx, line 73: @mjambon this may require your deeper understanding of JSON than me to do the right thing.

mjambon: oh, I missed this comment earlier. This code is incorrect since it evaluates ocaml string literals rather than json ones. We should be able to use yojson's `read_string` function.

Here's code to read a JSON string literal:

```ocaml
let read_json_string json =
  Yojson.Safe.read_string (Yojson.init_lexer ~buf:(Bi_outbuf.create 128) ()) (Lexing.from_string json)
```

```
# read_json_string {|"\"hello\u0000"|};;
- : string = "\"hello\000"
```
It parses the whole string literal including the quotes. There's no public function to parse the only contents. An easy but slightly inefficient way is:

```ocaml
let unescape_json_string s =
  read_json_string ("\"" ^ s ^ "\"")
```

mjambon: PR in progress: https://github.com/returntocorp/semgrep/pull/5286

PR #5279, opened 2022-05-20
tainting: Pass taint findings to the match hook

IagoAbal:
This will at least facilitate debugging taint findings in DeepSemgrep.

Helps PA-1310

Also, fixed a couple of typos in the changelog.

test plan:
make test

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   safe data, this was not recognized by the taint engine. Also, if `taint(x)`
   occurred inside e.g. an `if` block, any occurrence of `x` outside that block
   was not considered tainted. Now, if you specify that the code variable itself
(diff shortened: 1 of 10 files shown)


aryx (approved):
no test plan? :)

IagoAbal:
> no test plan? :)

Ooops, I did not imagine this would break any test. It's quite weird actually. I did test it with `make` alone...

IagoAbal:
So it seems that `Pattern_match.equal` doesn't use `AST_utils` because `Metavariable.equal_mvalue` is automatically derived by `@@deriving`. I fixed this locally for now, but I think we may need a custom `Metavariable.equal_mvalue`?

_Update:_ Actually we have this custom `equal_mvalue`... we have two, one for structural equality and another for referential equality, under a different submodule each. Maybe we could make `equal_mvalue` default to `Structural.equal_mvalue`? Same for `equal_bindings`.

PR #5287, opened 2022-05-23
Pretty SCA Results

mmcqd:
This PR adds nice CLI output for SCA findings.

Now if semgrep output contains SCA findings, they are sorted into Reachable, Unreachable, and First-Party (the non-SCA findings), and displayed like this:

Screen Shot 2022-05-23 at 10 24 58 AM

An SCA summary is available at the top of the results.

If there are no SCA rules, or no SCA findings, output is unchanged. So `Findings` only becomes `First-Party Findings`, if there are SCA findings.

This also updates semgrep output for SCA findings to include the full path to the lockfile where dependency matches were found, and not just the name of the lockfile.

PR checklist:

- [x] Changelog is up-to-date


semgrep/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awareruby-sca.yaml-dependency_awaresca.rb/results.json
```diff
@@ -30,7 +30,7 @@
               "resolved_url": null,
               "version": "1.21.0"
             },
-            "lockfile": "Gemfile.lock"
+            "lockfile": "targets/dependency_aware/Gemfile.lock"
```
brendongo, line 33: Did you intend to change the path of the lockfile in the json output here @mmcqd

mmcqd: I did. Should have included it in the PR description, my bad. Updating it

semgrep/semgrep/formatter/text.py
```diff
@@ -396,12 +406,16 @@ def format(
...
-            rule_matches,
-            extra.get("color_output", False),
-            extra["per_finding_max_lines_limit"],
-            extra["per_line_max_chars_limit"],
-        )
+        reachable = []
+        unreachable = []
+        first_party = []
+        for match in rule_matches:
+            if "dependency_match_only" not in match.extra:
```
brendongo, line 414: Is this a safe check?

Either way we should add an e2e test for this new output

mmcqd: How could it be unsafe? Every `rule_match` has an `extra` field. Will definitely add an e2e test though

semgrep/semgrep/formatter/text.py
```diff
@@ -413,8 +427,60 @@ def format(
...
-        return "\nFindings:\n" + matches_output
+        if (reachable or unreachable) and first_party:
+            first_party_output = self._build_text_output(
+                first_party,
+                extra.get("color_output", False),
+                extra["per_finding_max_lines_limit"],
+                extra["per_line_max_chars_limit"],
+            )
+            findings_output.append(
+                "\nFirst-Party Findings:\n" + "\n".join(first_party_output)
```
brendongo, line 470: I'm not fully sold on "First-Party Findings" Does just "Findings" cause confusion?

mmcqd: I thought it seemed slightly weird since "Findings" sounds like it ought to include _all_ findings to me. But I don't feel very strongly, happy to change it back

brendongo: I don't feel strongly either. Let's keep for now and see if it needs changing later

brendongo (approved):
Just needs a test for SCA rule output

PR #5288, opened 2022-05-23
perf: Reduce time taken by metrics collection

underyx:
PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/app/session.py
+++ b/semgrep/semgrep/app/session.py
@@ -1,6 +1,6 @@
 import os
-from pathlib import Path
 import subprocess
+from pathlib import Path
 from typing import Any
 from typing import Optional
 from typing import Set
--- a/semgrep/semgrep/app/session.py
+++ b/semgrep/semgrep/app/session.py
@@ -1,4 +1,6 @@
 import os
+from pathlib import Path
+import subprocess
 from typing import Any
 from typing import Optional
 from typing import Set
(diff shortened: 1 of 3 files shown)


underyx:
~gotta set up daemonized subprocess stuff here~

never mind, not possible

semgrep/semgrep/app/session.py
```diff
@@ -31,13 +33,43 @@ class UserAgent:
...
+                ],
+                cwd=Path(__file__).parent,
+                stderr=subprocess.DEVNULL,
+                encoding="utf-8",
+            ).strip()
+            # If we installed semgrep in `.venv/` in the semgrep-docs repo,
+            # this function would report semgrep-docs commit SHAs in the user agent.
+            # This is why we verify the origin URL, and why check with .endswith()
+            if remote_url.replace(".git", "").endswith("/semgrep"):
+                result.add(f"sha/{sha}")
```
underyx, line 64: I thought this'd be handy to be able to recognize when weird metrics come from faulty commits, and to filter out metrics from development in general. But it's highly optional.

semgrep/semgrep/metrics.py
```diff
@@ -280,14 +280,11 @@ def send(self) -> None:
         self._sent_at = datetime.now().astimezone().isoformat()
 
         try:
-            r = app_session.post(
+            r = requests.post(
```
underyx, line 283: We dropped the retry policy by no longer using the app session.

semgrep/semgrep/app/session.py
```diff
@@ -31,13 +33,43 @@ class UserAgent:
...
+        except subprocess.CalledProcessError:
+            pass
+
+        return result
 
     def __str__(self) -> str:
         result = f"{self.name}/{self.version}"
-        for note in sorted(self.tags):
+        for note in sorted(self.tags, key=lambda x: ("/" in x, x)):
```
brendongo, line 72: is note supposed to be a Tuple(bool, str)

underyx: nope, this is just the sorting key; essentially what this does is:

1. first, order all notes to the front that have no `/` in them
2. then, order the rest alphabetically

underyx: so this leads to this ordering: `(Docker) (command/scan) (sha/123987)`

underyx: CleanShot 2022-05-24 at 15 39 20@2x

sample in practice

semgrep/semgrep/app/session.py
brendongo: eh seems special case-y. Is it difficult to filter it out server side?

semgrep/semgrep/app/session.py
brendongo: I see

semgrep/semgrep/app/session.py
underyx: without this we don't have any markers that'd let us know if a metric was sent from one of our devs

brendongo (approved):
Would rather not have the semgrep url filtering but don't feel too strongly

semgrep/semgrep/app/session.py
underyx: so it's impossible to filter; but we can also just decide not to care about filtering

PR #5297, opened 2022-05-24
fix: Attempt a match between Id and IdQualified

nmote:
It's possible for DeepSemgrep to match in this case when the pattern is
an `Id` and the source is an `IdQualified`, but a parent of the source
name ultimately resolves to a matching `Id`.

At some point we'll likely have to add another case so that
`IdQualified` can match `Id`.

Test plan: Automated tests

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -374,6 +374,20 @@ let m_with_symbolic_propagation f b =
  * TODO: remove MV.Id and use always MV.N?
  *)
 let rec m_name a b =
+  let try_parents dotted =
+    let parents =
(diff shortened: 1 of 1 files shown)


aryx (approved):
Not super convinced, but let's try this.

nmote:
Is there something about this that I could clarify?

aryx:
No it's just that I could not imagine easily an example where this would trigger, where this would be useful.

nmote:
Makes sense. Since it's motivated by an example in a different repo, it's not at all obvious. I hope that now that you've reviewed that PR, it's a little clearer?

aryx:
yep, clearer now :)

PR #5301, opened 2022-05-24
Fixed discrepency between CI and Scan outputs

### LINKED ISSUE #5007: Missing "suppressions" in SARIF output for semgrep ci
**Describe the bug**
I used to run scans with semgrep-agent in diff mode, with the SARIF output, and it supported "suppressions" in the SARIF output.
It seems that the `semgrep ci` new method is not reporting to me the issues, I see in the scan logs that it finds all issues, then only reports a part of it, and it's not mentioned in the SARIF file.
I also checked the `--enable-nosem` and it's enabled by default according to the cli doc, so I guess this means "no sem enabled, will report them" ? but still nothing in the SARIF output

**To Reproduce**
I guess it's simple enough to create a file to be scanned and add a comment with // nosemgrep, the tricky part might be scanning on 2 branches though, not sure if this is the bug root cause.
```
semgrep 0.87.0 on python 3.10.4
    <multilang> | 44 rules × 2 files
           java | 53 rules × 1 file 

  Current version has 4 findings.

Switching repository to baseline commit '20f88xxxxxx'.

....

Scanning 1 file.
Returning to original head revision 63686xxxxxx

Some files were skipped.
  Scan was limited to files changed since baseline commit.
....
 Ran 295 rules on 1 file: 2 findings.
Ran 295 blocking rules, 0 audit rules, and 0 internal rules used for rule recommendations.
```
**Expected behavior**
The SARIF output of a diff-aware scan should still produce suppressed findings inside

**Screenshots**

**Environment**
I'm running semgrep in a github action, without the api token, reporting directly SARIF in a PR comment

**Use case**
Just improve the scan details provided in the comment

ajbt200128:
CI command would filter out ignored matches, whether or no nosemgrep was
enabled, or the output format dictated that they should not be filtered out.

Also added e2e tests for possible CI outputs.

PR checklist:

- [ ] Changelog is up-to-date


semgrep/semgrep/commands/ci.py
```diff
@@ -391,19 +393,23 @@ def ci(
     blocking_matches_by_rule: RuleMatchMap = {}
     nonblocking_matches_by_rule: RuleMatchMap = {}
     cai_matches_by_rule: RuleMatchMap = {}
+
+    # Since we keep nosemgrep disabled for the actual scan, we have to apply
+    # that flag here
+    keep_ignored = not enable_nosem or output_handler.formatter.keep_ignores()
```
brendongo, line 399: Nice! Helpful comment 👍

semgrep/tests/e2e/test_ci.py
```diff
@@ -338,19 +339,27 @@ def test_config_run(tmp_path, git_tmp_path_with_commit, snapshot, autofix):
     snapshot.assert_match(sanitized_output, "output.txt")
 
 
-def test_outputs(tmp_path, git_tmp_path_with_commit, snapshot, autofix):
+@pytest.mark.kinda_slow
+@pytest.mark.parametrize(
+    "format",
+    ["--json", "--gitlab-sast", "--gitlab-secrets", "--sarif", "--emacs", "--vim"],
```
brendongo, line 345: 🔥

brendongo: Should we also add using/disabling the `enable-nosem` flag to the tests?

brendongo (approved):

PR #5304, opened 2022-05-25
[C++] apply const to the type, not the decl

### LINKED ISSUE #5300: Const keyword is not compatible with semgrep pattern-matching in C++
**Describe the bug**

It appears that the ability to search for a const variable is not possible in Semgrep C++.

**To Reproduce**
https://semgrep.dev/s/chakirbelhaj:x1xl

**Expected behavior**
Semgrep detecting const variables in c++.

**Screenshots**

**Environment**

**Use case**
For a particular use case, we are trying to detect variables that are not reassigned AND are not defined as a const. With this, we intend to force that "read-only" variables are defined as a const.

aryx:
This closes #5300

test plan:
test file included
make test

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - semgrep-core: you can use again rules stored in JSON instead of YAML (#5268)
 - Python: adds support for parentheses around `with` context expressions
   (e.g., `with (open(x) as a, open(y) as b): pass`) (#5092)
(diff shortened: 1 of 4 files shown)


ChakirBelhaj:
When will this change be available on the playground?

aryx:
it's available now if you select the "use develop docker image".
See https://semgrep.dev/s/chakirbelhaj:x1xl?develop=true

aryx:
it should make the next release next week (semgrep 0.95)

PR #5310, opened 2022-05-25
fix(dataflow,scala): Correctly translate return when it's an expression

### LINKED ISSUE #4975: [scala] Functions do not work as taint sinks
**Describe the bug**
Simple functions do not work as taint syncs, even though they should.

**To Reproduce**
https://semgrep.dev/s/ezzer17:simple-function-sink-2
If you remove the `return`, it works

**Expected behavior**
Rule has one match

- P1: important to fix or quite annoying

kopecs:
Fixes #4975.

The underlying issue that we current do not translate return from the generic AST to IL when `return` in the source language may be an expression (e.g., Rust, Scala). This PR adds a case to handle this.

PR checklist:

- [x] Changelog is up-to-date

Test plan: test included for both implict & explict return in Scala.


semgrep-core/tests/OTHER/rules/taint_scala_return.scala
```diff
@@ -0,0 +1,9 @@
+def no_ret(x: Int) : Int = {
```
mmcqd: This test should go in `semgrep-core/tests/tainting_rules/scala`

mmcqd (approved):
Great job! Fixing tainting bugs isn't easy

semgrep-core/tests/tainting_rules/scala/return_expr.scala
```diff
@@ -0,0 +1,9 @@
+def no_ret(x: Int) : Int = {
```
IagoAbal: No need to change, but just FYI `tests/tainting_rules` is a leftover from when `semgrep-core -rules` could not parse taint rules and we had a different switch for running them. Nowadays I try to put taint tests under `tests/OTHER/rules/` and I just prefix them with `taint_`.

IagoAbal (approved):
Very nice! Could you please add this https://semgrep.dev/s/ezzer17:simple-function-sink-2 (the example provided in #4975) also as a test case?

PR #3200, opened 2021-05-26
Support for $...BODY to match over list of stmts

aryx:
PR checklist:
- [x] changelog is up to date

Issue #3170, opened 2021-05-21
$...BODY not working for Go


DrewDennison:
https://semgrep.dev/s/L0bG is not matching for @dgryski but if you use `...` in place of `$...BODY` it matches. He needs to capture the function body for autofix.

See: https://r2c-community.slack.com/archives/C018NJRRCJ0/p1621623452029300 for more context

aryx:
This works now: https://semgrep.dev/s/L0bG?version=develop
Might want to use gofmt on the autofixed code though ...

PR #5313, opened 2022-05-25
fix(matching): Allow ellipsis metavariable to match statements not in blocks

### LINKED ISSUE #3771: $...BODY not working for Python
**Describe the bug**
Similar issue to https://github.com/returntocorp/semgrep/issues/3170 but in Python

**To Reproduce**
https://semgrep.dev/s/DwPy/

kopecs:
Fixes #3771.

It seems we can currently match statements when they appear in certain blocks, but not otherwise. For instance (Java),

```
class $C {
    void foo(...) {
        if (...) $...BODY;
    }
}
```

doesn't match

```
class Foo {
    void foo() {
        if (true) { int x = 10; return; }
    }
}
```

but would after this PR.

For reviewers: The discussion on #3771, also mentioned #3170 as a similar issue. The PR resolving that issue (#3200) modified some functions like `m_list__m_stmt_uncached`. Do I need to modify additional functions beyond `m_stmt` here for robustness, or does this already handle all cases? I wasn't able to break it on my machine.

PR checklist:

- [x] Changelog is up-to-date

Test plan: Includes tests for Python (original issue language) and Java.


mmcqd (changes requested):
This doesn't seem to be properly capturing the whole statement in the metavariable. If I run your python test with
```
pattern: |
  class $C:
     $...BODY
message: $...BODY
```
Each class should have one match and the message should print the full contents of the class. Instead it produces multiple matches for each class, and `$...BODY` contains only part of the statements in the class.

mmcqd:
Actually I might be totally wrong. I guess this should behave the same as ellipsis. How do those behave in the case I described?

kopecs:
> This doesn't seem to be properly capturing the whole statement in the metavariable. If I run your python test with
> 
> ```
> pattern: |
>   class $C:
>      $...BODY
> message: $...BODY
> ```
> 
> Each class should have one match and the message should print the full contents of the class. Instead it produces multiple matches for each class, and `$...BODY` contains only part of the statements in the class.

~~I can't reproduce this in the manner you describe. I was getting duplicate matches (which, wrt your other comment, I don't get with `...`), but they seem to contain the whole body on the examples I tested.~~ After some more trial & error I got this to reproduce.

Looking at this in more depth, I should probably just stop going deeper once I hit this case if its a metavar ellipsis, à la normal ellipsis (but also associating the matched content).

kopecs:
The duplicate matches seem related to the couple cases following this, which seem to prevent the same thing happening for `...`, but it seems like this is just a hacky fix based on the comments:

https://github.com/returntocorp/semgrep/blob/f2b0064dceb9563c92c4e1ac9013d904a54f704e/semgrep-core/src/matching/Generic_vs_generic.ml#L2035-L2054

Edit: I actually think this is related to the field thing mentioned below, but it'll take me a while to go through and see. Seems like there's also potentially some special-casing for handling python stuff there too.

kopecs:
It looks like for a python class we treat any statements in it as fields instead. This makes sense for reasonable python code, where everything in a class is just a method definition or similar, but because we're order agnostic for fields it might cause issues for classes containing statements like the one in the test.

kopecs:
Closing with explanation in the comment on https://github.com/returntocorp/semgrep/issues/3771#issuecomment-1150208966. There is a separate, somewhat unrelated change in here I'll probably put in a separate PR about the Java statement matching.

PR #5324, opened 2022-05-26
Ajbt200128/5222 ruby normalize extra parameters parenthesis

### LINKED ISSUE #5222: Ruby: normalize extra parameters parenthesis
**Describe the bug**
In Ruby, given the semgrep pattern `method("...")`, semgrep will match `method("foo")` but not the equivalent `method ("foo")`

You could extend the rule to account for that specific case, but I think semgrep should normalize the code to not consider redundant characters. The pattern `method("...")` should match `method "..."`, `method("...")`, `method ("...")`, `method (("..."))`, `method(("..."))`, ... as they are all equivalent method calls with a single string argument.

**To Reproduce**

https://semgrep.dev/s/x9Oy

```
rules:
- id: ruby-method-call
  pattern: method("...")
  languages: [ruby]
  severity: WARNING
```

Test cases:
```
# ruleid: ruby-method-call
method("adsf") # pass

# ruleid: ruby-method-call
method "adsf" # pass

# ruleid: ruby-method-call
method ("adsf") # fail

# ruleid: ruby-method-call
method (("adsf")) # fail

# ruleid: ruby-method-call
method(("adsf")) # fail
```

**Expected behavior**
All test cases pass

**Screenshots**

**Environment**

**Use case**
- Prevent false negatives. Semgrep should be able to find the targeted code even if the code is formatted poorly.

ajbt200128:
When parsing ruby arguments, we don't take into account the fact that `method(1)` is equivalent to `method((1))` when matching, i.e. even though you can surround an argument w/ a bunch of parentheses and semantically it is the same, we don't recognize that when actually parsing the ruby tree sitter syntax tree, or when matching in general for any language that this is possible. To fix this, I modified `ruby_to_generic.ml` to make sure this is caught, and 

Also, added parentheses equivalence (see changelog and 0a17b0a4fa0cf46829a6baf6021c525d5863519f) 

PR checklist:

- [X] Changelog is up-to-date


semgrep-core/src/parsing/pfff/ruby_to_generic.ml
```diff
@@ -177,6 +177,11 @@ and argument arg : G.argument =
   | Arg (Binop (Atom (_tk, AtomSimple id), (Op_ASSOC, _v2), v3)) ->
       let e3 = expr v3 in
       G.ArgKwd (id, e3)
+  (* Ruby allows surrounding single arguments with useless parenthesis *)
+  (* Ex: method(1,2,3) is equivalent to method((1),(2),(3))*)
+  | Arg (S (Block (l, [ S (Block (l2, [ e ], r2)) ], r))) ->
```
mmcqd, line 182: You've only handled the case with 1 or 2 extra sets of parentheses. What if there were more? You'll need to deal with this recursively

mmcqd (changes requested):
So `ParenExpr` is a bit tricky... It's used in a very limited fashion right now, and the potential issue with it is that, even though you can consider `ParenExpr e` to be equal to `e` for the purposes of semgrep matching,  you have to do this manually when pattern matching on expression in *ocaml*. So now `match e with Foo -> ....` has to be `match e with Foo | ParenExpr Foo -> ...`. But, you don't seem to have broken any test cases, so perhaps it's fine. The alternative would be to just drop the extra tokens all together, and turn `Arg (Block e)` into `Arg e`, but this messes up ranges/autofix... Maybe we could do some weird token merging business... cc @aryx for more thoughts.

mmcqd:
Would be good to have a more descriptive PR description (what was the bug, how did you fix it, in high level terms), and to include a line like "Closes #5222", so the issue gets automatically closed when this is merged.

semgrep-core/src/parsing/pfff/ruby_to_generic.ml
ajbt200128: Meant to do that but it totally slipped my mind, will update this

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Added
 
 - Sarif output format now includes `fixes` section
+- Patterns will now match against code with extra parentheses. See Ruby patterns change
+  below for an example
+
+### Changed
+
+- Ruby patterns will now match arguments with extra parentheses around arguments, or with a space
```
emjin, line 15: Nit: I think it would be more useful to give the example of `method("...")` matching `method ("...")`. I imagine I would see this and think "Why would I ever write `method(("foo"))`?"

ajbt200128:
> So `ParenExpr` is a bit tricky... It's used in a very limited fashion right now, and the potential issue with it is that, even though you can consider `ParenExpr e` to be equal to `e` for the purposes of semgrep matching, you have to do this manually when pattern matching on expression in _ocaml_. So now `match e with Foo -> ....` has to be `match e with Foo | ParenExpr Foo -> ...`. But, you don't seem to have broken any test cases, so perhaps it's fine. The alternative would be to just drop the extra tokens all together, and turn `Arg (Block e)` into `Arg e`, but this messes up ranges/autofix

Yea I also had the same conclusion re: dropping the tokens, it didn't seem like that was a worthwhile tradeoff for this fix. I'm not sure I understand what you mean regarding manual matching in OCaml. The way I understand it is since this is only modifying the match expression function, it shouldn't modify pattern matching anywhere else, since all we're doing is shedding a `ParenExpr`. I'm not disagreeing, I just don't think I understand OCaml/the codebase enough to get what you're saying, and would appreciate some more explanation :)

mmcqd:
There's two places it would matter. Consider this ocaml code that matches on an expression
```
match e with
  | L _ -> a
  | N _ -> b
  | _ -> c
```
It matches on either a literal or a name, and in every other case does something else. If `ParenExpr e` is the same as `e`, then this code is now wrong, because `ParenExpr (L ...)` should go to the first case, but instead it goes to the third. The other case would be anytime we're matching deeper than one layer of constructors. (Omitting tokens/record stuff so it's easier to read)
```
match e with
  | Call (_,[ Arg (L _) ]) -> a
  ....
```
Even if we recursively handle the `ParenExpr` case in this match expression, `Call (_, [ Arg (ParenExpr (L _)) ])` isn't going to go to the first case, even though we'd want it to.

Like I said, nothing is broken by this PR, so this may not really be a big deal. It's just something @aryx and I discussed a couple weeks ago.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -730,6 +730,11 @@ and m_expr a b =
    *)
   | G.Ellipsis _a1, _ -> return ()
   | G.DeepEllipsis (_, a1, _), _b -> m_expr_deep a1 b
+  (* equivalence: extra parentheses equivalence!
+   * must be before literal match below, to allow extra paren in the pattern to still match code without parens
```
emjin, line 734: I'm confused about this comment: aren't you allowing `e` to match `(e)` (aka pattern matches _code_ with extra parens)?

semgrep-core/src/parsing/pfff/ruby_to_generic.ml
```diff
@@ -177,6 +177,11 @@ and argument arg : G.argument =
   | Arg (Binop (Atom (_tk, AtomSimple id), (Op_ASSOC, _v2), v3)) ->
       let e3 = expr v3 in
       G.ArgKwd (id, e3)
+  (* Ruby allows surrounding single arguments with useless parenthesis *)
```
emjin, line 180: I did a little investigating, and it's more complicated than this. In Ruby you can omit parentheses in a function call, so that `f(x)` and `f x` are equivalent. The problem now is that when someone writes `f (x)`, they want whitespace to be irrelevant: they mean `f(x)`. But if `f x` = `f(x)`, then `f (x) = f((x))`, and `f((x)) != f(x)`. This is because `f(x if false) != f((x if false))` (see https://stackoverflow.com/questions/17036166/why-do-i-need-extra-parentheses-for-conditional-modifier for why). So whitespace is _not_ irrelevant; `f (x) != f(x)`

I'm not entirely sure what I think is the right solution here tbh; I _think_ using `ParenExpr` (to preserve parentheses for autofix) is still best, but it is a little weird since by the time this gets to Semgrep we don't care if it's `f x`, `f(x)`, or `f((x))`. Another option is to specifically change tree-sitter-ruby to parse `f (x)` as equivalent to `f(x)` (I don't _think_ there are any cases of valid code where that would matter). I don't prefer it mostly because I'm not comfortable enough with Ruby to say whether there would be unintended effects.

emjin: At the very least, the requested changes is to explain all this in the comment :D

emjin (changes requested):

emjin:
I really should have looked into this issue more before giving it to you; I'm now unsure if we should fix this at all.

emjin:
Ok, updated recommendation: we should try to check how Ruby parses `f (x)` vs `f(x)`. If Ruby considers these equivalent, then we should update the tree-sitter parser accordingly.

ajbt200128:
@mmcqd we came to the conclusion that there's too much nuance to this issue, and there's no clear solution, I'll write up what we discovered and close the issue

PR #5325, opened 2022-05-26
refactor(metrics): Rewrite as class with TypedDict payload

underyx:
Why? I'll admit, this is quite experimental. Here are the reasons I
expect this to be an improvement:

- Previously, we'd have to repeat ourselves a lot to set a value:
  we would have to define and type-annotate an intermediary data
  storage attribute, and then translate the private attribute into the
  payload dict's format when sending. Now, we just type annotate the
  dict, and the setter can directly drop it in the payload.
  - This indirection, I think, led to the data schema being subideal in
    some cases, as the data-providing function was disconnected from
    what the metrics payload ended up looking like.
  - Removing the indirection also revealed that we could reorder the
    data setting method calls in semgrep_main, group them nicer, and
    change some of their function signatures to be more consistent.
- Previously, all fields we tracked were in a flat attribute structure,
  now we can namespace them without thinking about attribute names (such
  as errors.count and rules.count).
- Previously, data was serialized as early as it was added on Metrics.
  Now we keep rich data such as datetimes around as late as possible,
  and only convert them when serializing to JSON
  - This also deduplicates some code and makes sure that e.g. all our
    datetimes are in a consistent format
- Previously, you could find out the payload structure from the
  `def as_dict() -> Dict[str, Any]` method.
  Now you can find out the payload structure in the TypedDict structure.
  With the old version it was easier to see the structure of what keys
  are available, but it was harder to tell what type each field is.

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/app/session.py
+++ b/semgrep/semgrep/app/session.py
@@ -127,7 +127,7 @@ class AppSession(requests.Session):
         self.token = auth.get_token()
 
         metrics = get_state().metrics
-        metrics.add_sanitized_token(self.token)
+        metrics.add_token(self.token)
 
     def request(self, *args: Any, **kwargs: Any) -> requests.Response:
         kwargs.setdefault("timeout", 30)
--- a/semgrep/semgrep/app/session.py
+++ b/semgrep/semgrep/app/session.py
@@ -126,8 +126,7 @@ class AppSession(requests.Session):
 
         self.token = auth.get_token()
 
-        metrics = get_state().metrics
-        metrics.set_is_authenticated(bool(self.token))
+        get_state().metrics.payload["environment"]["isAuthenticated"] = bool(self.token)
 
     def request(self, *args: Any, **kwargs: Any) -> requests.Response:
         kwargs.setdefault("timeout", 30)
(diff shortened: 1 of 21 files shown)


underyx:
Another way of thinking about this is, the data layers you have to design are:
- old way: actual data, stored data, transported data
- new way: actual data, transported data

And with the old way it's up to the developer to verify that "stored data" and "transported data" are identical.

semgrep/semgrep/app/session.py
```diff
@@ -127,7 +127,7 @@ def authenticate(self) -> None:
         self.token = auth.get_token()
 
         metrics = get_state().metrics
-        metrics.set_is_authenticated(bool(self.token))
+        metrics.add_sanitized_token(self.token)
```
brendongo: makes it seem like self.token is santized?

underyx: good point, my thinking was "add token after sanitizing it", but I see how I need a new naming scheme to make this unambiguous 🤔

brendongo (approved):

brendongo:
Do we have tests that make sure metrics sent to backend dont unexpectedly change?

underyx:
> Do we have tests that make sure metrics sent to backend dont unexpectedly change?

not yet, will add before merging!

underyx:
Tested that nothing changed by doing

```console
$ git checkout develop
$ git checkout bence/new-datapoints -- semgrep/tests
$ pipenv run pytest -k test_metrics_payload
```

which succeeded.

PR #5326, opened 2022-05-27
Fix Flaky Shouldafound Tests

spencerdrak:
This PR addresses flakiness in the shouldafound tests by addressing two items:
- We shouldn't rely on the presence or absence of external tooling for testing to pass
- We shouldn't rely on the ordering of parametrized tests to determine what snapshots look like.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep/semgrep/commands/shouldafound.py
+++ b/semgrep/semgrep/commands/shouldafound.py
@@ -53,8 +53,9 @@ def shouldafound(
     Report a false negative in this project. "path" should be the file in which you expected the vulnerability to be found.
     """
     if not email:
-        email = _get_git_email()
-        if email is None:
+        try:
+            email = _get_git_email()
+        except subprocess.CalledProcessError:
             click.echo(
                 "Could not parse git email. Please use the --email flag to report this false negative.",
                 err=True,
@@ -93,14 +94,11 @@ def shouldafound(
         sys.exit(2)
 
 
-def _get_git_email() -> Optional[str]:
-    try:
-        email = (
-            subprocess.check_output(["git", "config", "user.email"]).decode().strip()
-        )
-        return email
-    except Exception:
-        return None
+def _get_git_email() -> str:
+    """
+    :raises CalledProcessError: If not in a git environment / email not configured
+    """
+    return subprocess.check_output(["git", "config", "user.email"]).decode().strip()
 
 
 def _make_shouldafound_request(data: JsonObject) -> Optional[str]:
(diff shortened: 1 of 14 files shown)


semgrep/semgrep/commands/shouldafound.py
```diff
@@ -53,13 +53,8 @@ def shouldafound(
...
     if not email:
-        try:
-            email = (
-                subprocess.check_output(["git", "config", "user.email"])
-                .decode()
-                .strip()
-            )
-        except Exception:
+        email = _get_git_email()
+        if email == None:
```
underyx:
```suggestion
        if email is None:
```

underyx (approved):

semgrep/tests/e2e/test_shouldafound.py
```diff
@@ -95,6 +95,8 @@ def test_shouldafound_findings_output(
         "get_no_findings_msg",
         return_value=message,
     )
+    mocker.patch.object(notifications, "possibly_notify_user", return_value=None)
```
underyx:
```suggestion
    mocker.patch.object(scan.notifications, "possibly_notify_user", return_value=None)
```

semgrep/tests/e2e/test_shouldafound.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 from click.testing import CliRunner
 
+from semgrep import notifications
```
underyx:
```suggestion
```

semgrep/semgrep/commands/shouldafound.py
```diff
@@ -98,6 +93,16 @@ def shouldafound(
...
 
 
+def _get_git_email() -> Optional[str]:
+    try:
+        email = (
+            subprocess.check_output(["git", "config", "user.email"]).decode().strip()
+        )
+        return email
+    except Exception:
+        return None
```
nbrahms: What you have definitely works. This may be a bit simpler:

```suggestion
def _get_git_email() -> str:
    ```
    :raises CalledProcessException: If not in a git environment / email not set
    ```
    return subprocess.check_output(["git", "config", "user.email"]).decode().strip()
```

nbrahms (approved):
Just one suggestion

semgrep/semgrep/commands/shouldafound.py
brendongo: @nbrahms do we need extra handling for the exception or is default behavior a nice message

brendongo (approved):

semgrep/semgrep/commands/shouldafound.py
spencerdrak: addressed nathan's suggestion, resolving.

Issue #3941, opened 2021-09-27
[PHP] Use tree-sitter-php to parse PHP code instead of pfff


aryx:
Pad already done the job of integrating tree-sitter-php in ocaml-tree-sitter-semgrep. We just need now
to finish Parse_php_tree_sitter.ml in semgrep-core/src/parsing/tree-sitter/ and implement all those
todos.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-412/[php]-use-tree-sitter-php-to-parse-php-code-instead-of-pfff.

cfroystad:
Great! I hope to find some time to complete tree-sitter-php in the not too distant future for our mutual benefit.

Note that there are some AST inconsistencies in the master branch I hope to resolve before the next release and that the [First-class callable syntax](https://wiki.php.net/rfc/first_class_callable_syntax) of PHP 8.1 might (depending on how you extend the grammar) interfere with Semgrep.

aryx:
:wave:

Sjord:
[This comment](https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml#L28-L32) suggests to map tree-sitter CST to PHP AST, which can then be converted to generic AST. Does this make sense? Wouldn't it be better to map to generic AST immediately?

Currently php_to_generic uses both AST and CST types. E.g. [BoolTy is a CST type](https://github.com/returntocorp/semgrep/blob/develop/semgrep-core/src/parsing/pfff/php_to_generic.ml#L86). But this type is tied to pfff CST, and not to tree-sitter CST. Do we need to map from tree-sitter CST to pfff CST, or should BoolTy be moved to PHP AST, or should we just go directly to generic AST?

aryx:
I would like to remove cst_php.ml from pfff, and only get an ast_php.ml there.
Then, when there is an existing ast_xxx.ml (like the one in pfff), I prefer to translate the tree-sitter CST to this ast_xxx.ml first, and then
separately have a xxx_to_generic.ml.
This is debatable. I usually prefer to separate concerns, but there's a case to be make sometimes to go directly
to the generic AST.
Recently we sometimes went directly to the generic AST (as for Swift, or C#). But recently some people
preferred first to define a proper AST for the language (e.g., AST_bash.ml, AST_dockerfile.ml) and split the process in 2.

Sjord:
[...]

PR #5330, opened 2022-05-27
[PHP] Use latest semgrep-php fix with fix for echo

aryx:
This will help https://github.com/returntocorp/semgrep/issues/3941

The parsing-stats for PHP in ocaml-tree-sitter-semgrep are now at
99.97%!

However, we are back at 92% for the parsing-stats in semgrep-core
for PHP (weird, because we were at 96% last week). Probably
some TODOs started to trigger more because of the parsing
fix.

test plan:
test file included
make test

PR checklist:

- [x] Changelog is up-to-date


aryx:
new stats for PHP after cd parsing-stats/; ./run-lang php
```
stats available in lang/php/stats.json
lang/php/stats.json:3
   "global": {
     "name": "*",
     "parsing_rate": 0.925674482007057,
     "line_count": 5303885,
     "error_line_count": 394214,
     "file_count": 34978,
     "error_file_count": 925
   },
```

PR #5331, opened 2022-05-27
Ignore the var modifier

Sjord:
Since it isn't a modifier. E.g. in

```
class A {
    var $a;
}
```

We can ignore `var` and add no modifiers.

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
@@ -323,8 +323,7 @@ let map_namespace_use_group_clause (env : env)
 
 let map_modifier (env : env) (x : CST.modifier) : A.modifier =
   match x with
-  | `Var_modi tok ->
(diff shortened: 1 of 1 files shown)


semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
```diff
@@ -323,8 +323,7 @@ let map_namespace_use_group_clause (env : env)
 
 let map_modifier (env : env) (x : CST.modifier) : A.modifier =
   match x with
-  | `Var_modi tok ->
-      (* pattern [vV][aA][rR] *) todo env tok (* TODO add to AST *)
+  | `Var_modi tok -> (* pattern [vV][aA][rR] *) failwith "not a modifier"
```
aryx, line 326: I usually prefer a raise Impossible, or assert false explaining why this branch can never be executed
(here because it's filtered in map_modifiers for example).

aryx (approved):

PR #5343, opened 2022-05-27
fix(naming,php): Handle PHP `use` without explicit renaming

### LINKED ISSUE #3964: [PHP] Import aliasing for PHP
semgrep easily tackles import aliasing when import has a name:

```php
<?php
use My\Full\Classname as Another;

$obj = new Another();
```
pattern:
```
new My\Full\Classname()
```
https://semgrep.dev/s/GOpB

works perfectly, but in a situation with imports without a name it does not work:

```php
<?php
use My\Foo\Bar;

$obj2 = new Bar();
```

pattern:
```
new My\Foo\Bar()
```
https://semgrep.dev/s/5De6/

kopecs:
Fixes #3964.

PR checklist:

- [x] Changelog is up-to-date


Sjord:
I don't really like this approach, where the AST is modified to get certain functionality from semgrep. In `use My\Foo\Bar;`, there is no alias specified, so I think the AST should have no alias, not a fake `Bar` as alias. Parse_php_tree_sitter.ml creates PHP-specific AST, so I think it should lean as to how PHP actually works. Specifying a fake alias in php_to_generic.ml is a little bit more acceptable to me, to make the translation between PHP and semgrep generic AST.

aryx:
I agree with @Sjord .

kopecs:
Yeah I'm not a huge fan of it either. I realised afterwards that just using `ImportFrom` makes a lot more sense (assuming it functions like I want it to here), so I was just going to change this over to using that once I start working on this more.

semgrep-core/src/parsing/pfff/php_to_generic.ml
```diff
@@ -163,9 +163,25 @@ let rec stmt_aux = function
...
+  | NamespaceUse (t, v1, v2) -> (
+      let v1 = qualified_ident v1 in
+      match v2 with
+      | Some x ->
+          [
+            G.DirectiveStmt
+              (G.ImportAs (t, G.DottedName v1, Some (alias x)) |> G.d)
+            |> G.s;
+          ]
+      | None -> (
```
mmcqd, line 176: Would be nice to have a comment like 
```
(* `use Foo\Bar\Baz` brings `Baz` into scope *)
```
giving an example of what this does.

mmcqd (approved):

mmcqd:
Oh and needs a changelog entry before you merge

PR #5353, opened 2022-05-30
fix(php): Handle match expressions

nmote:
I modeled the output after that for OCaml, which uses a `Switch` node to represent pattern matching.

Test plan:

`test.php`:

```
<?php

match($foo) {
  Foo::Baz,
  Foo::Bar => 5,
  default => 4
};
```

`semgrep-core -tree_sitter_only -lang php -dump_ast test.php`:

```
Pr(
  [ExprStmt(
     StmtExpr(
       Switch((),
         Some(Cond(
                N(
                  Id(("$foo", ()),
                    {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); })))),
         [CasesAndBody(
            [Case((),
               OtherPat(("ExprToPattern", ()),
                 [E(
                    DotAccess(
                      N(
                        Id(("Foo", ()),
                          {id_info_id=2; id_hidden=false; id_resolved=Ref(
                           None); id_type=Ref(None); id_svalue=Ref(None); })), (),
                      FN(
                        Id(("Baz", ()),
                          {id_info_id=3; id_hidden=false; id_resolved=Ref(
                           None); id_type=Ref(None); id_svalue=Ref(None); }))))]));
             Case((),
               OtherPat(("ExprToPattern", ()),
                 [E(
                    DotAccess(
                      N(
                        Id(("Foo", ()),
                          {id_info_id=4; id_hidden=false; id_resolved=Ref(
                           None); id_type=Ref(None); id_svalue=Ref(None); })), (),
                      FN(
                        Id(("Bar", ()),
                          {id_info_id=5; id_hidden=false; id_resolved=Ref(
                           None); id_type=Ref(None); id_svalue=Ref(None); }))))]))],
            ExprStmt(L(Int((Some(5), ()))), ())); CasesAndBody([Default(())], ExprStmt(L(Int((Some(4), ()))), ()))])),
     ())])
```

PR checklist:

- [x] Documentation is up-to-date
- [x] Changelog is up-to-date
- [x] Change has no security implications (otherwise, ping security team)

PR #5354, opened 2022-05-31
fix(php): Ignore use list

nmote:
Test plan:

`test.php`:

```
<?php

class Foo {
  use Bar1, Bar2 { f as g; }
}
```

Observe differences only with IDs when running:
`semgrep-core -lang php -diff_pfff_tree_sitter test.php`

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
@@ -2353,8 +2353,15 @@ and map_use_declaration (env : env) ((v1, v2, v3, v4) : CST.use_declaration) :
   in
   let v4 =
     match v4 with
-    | `Use_list x -> map_use_list env x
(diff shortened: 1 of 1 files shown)


nmote:
cc @Sjord

In concert with https://github.com/returntocorp/semgrep/pull/5353, this brings the PHP parse rate to 99.93%

Sjord:
I think with the C# parser we changed `todo` at some point to just return nothing instead of raise an error. Is that also a good idea here? Then these functions would remain marked as "todo", but won't stop the parsing process.

aryx (approved):

nmote:
@Sjord that seems pretty reasonable to me, but I'll defer to @aryx.

aryx:
I usually prefer to return an  OtherExpr, or OtherStmt, so at least the tokens of the constructs are somewhere in the AST,
but skipping them is also fine in the short term; better than raise Todo ...

PR #5357, opened 2022-05-31
Remove string quotes

Sjord:
E.g. in `$a = "hello"`, put `hello` in A.String instead of `"hello"`. This
conforms to how the pfff parser does it.

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
@@ -209,7 +209,11 @@ let map_visibility_modifier (env : env) (x : CST.visibility_modifier) =
 
 let map_string__ (env : env) (x : CST.string__) =
   match x with
-  | `Str tok -> (* string *) A.String (_str env tok)
+  | `Str tok ->
+      (* string *)
+      let value, tok = _str env tok in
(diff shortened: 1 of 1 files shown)


semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
```diff
@@ -209,7 +209,11 @@ let map_visibility_modifier (env : env) (x : CST.visibility_modifier) =
 
 let map_string__ (env : env) (x : CST.string__) =
   match x with
-  | `Str tok -> (* string *) A.String (_str env tok)
+  | `Str tok ->
+      (* string *)
+      let value, tok = _str env tok in
```
aryx, line 214: not sure what is this _str, but usually we prefix something with an underscore when it's not used and it's ok
(to not raise a warning).

aryx (approved):

semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
Sjord: Right. `_str = Parse_tree_sitter_helpers.str` was already in the boilerplate code. The underscore was probably there because it wasn't used. I started using it and I didn't think any more about it.

PR #5367, opened 2022-05-31
Support PHP enum when converting to generic AST

Sjord:
PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/semgrep-core/tests/php/backed_enum.php
+++ b/semgrep-core/tests/php/backed_enum.php
@@ -1,0 +1,11 @@
+<?php
+enum Suit: string
+{
+    case Hearts = 'H';
+    case Diamonds = 'D';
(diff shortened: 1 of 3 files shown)


semgrep-core/tests/php/backed_enum.php
```diff
@@ -0,0 +1,11 @@
+<?php
```
aryx, line 1: usually when the PR is only related to parsing, we just add the file in semgrep-core/tests/php/parsing/
Actually you'll see there was a semgrep-core/tests/php/parsing/TODO/enum.php in there.
No need to write a xxx.sgrep for it (unless the .sgrep actually tries to match an element in the construct 
that is now correctly parsed, for example here one case of the enum).

aryx (approved):

PR #5368, opened 2022-05-31
fix(deep): Don't attempt metavariable matches against parent classes

nmote:
Test plan:

Automated tests

`test.rb`:

```
class Bar < Foo::Baz
end
```

`test.sgrep`:

```
class $CLS < Foo::Baz
  ...
end
```

`$ deep-semgrep --lang ruby -f test.sgrep .`

Before, it printed three matches at the same location. Now, it prints
only one. The existing automated test runner doesn't fail on duplicate
matches, so it wouldn't be straightforward to write an automated test
for this.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -405,12 +405,18 @@ let rec m_name a b =
                       _sid );
               };
             _;
-          } ) ) ->
+          } ) ) -> (
       m_name a (B.Id (idb, B.empty_id_info ()))
       >||> (* try this time a match with the resolved entity *)
       m_name a (H.name_of_ids dotted)
-      >||> (* Try the parents *)
-      try_parents dotted
+      >||>
+      (* Try the parents *)
+      match a with
+      (* If we're matching against a metavariable, don't bother checking
+       * parents. It will only cause duplicate matches that can't be deduped,
+       * since the captured metavariable will be different. *)
+      | G.Id ((str, _tok), _info) when MV.is_metavar_name str -> fail ()
+      | _ -> try_parents dotted)
   | G.Id (a1, a2), B.Id (b1, b2) ->
       (* this will handle metavariables in Id *)
       m_ident_and_id_info (a1, a2) (b1, b2)


emjin (approved):
This is great, thanks!

nmote:
For the record, I'm not totally sure this is the right approach. Maybe someone could write a more complicated rule in which it would be useful for a metavariable to match parent classes. If that turns out to be the case, we could consider ignoring the metavariable binding list when deduping matches for display, instead of doing this. I'm having trouble imagining what such a rule would look like, though, so I think this is fine for now.

emjin:
I think I would be pretty surprised actually if my metavariable matched parent classes (though maybe I'm just suffering a failure of imagination), so I think this is better. We'll see as people write more DeepSemgrep rules :)

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -405,12 +405,18 @@ let rec m_name a b =
...
+          } ) ) -> (
       m_name a (B.Id (idb, B.empty_id_info ()))
       >||> (* try this time a match with the resolved entity *)
       m_name a (H.name_of_ids dotted)
-      >||> (* Try the parents *)
-      try_parents dotted
+      >||>
+      (* Try the parents *)
+      match a with
+      (* If we're matching against a metavariable, don't bother checking
```
aryx, line 415: It's fine for now, but in theory the metavariable could be used another time in the same pattern, or in another pattern of the rule, and would match a parent in the other pattern and then this would fail because here we didn't allow the metavar to match also the parent.
Just FYI.
Maybe in practice this does not happen and we prefer to avoid the confusing duplicates.

IagoAbal: Triggering this should be as easy as adding a `metavariable-pattern` on the metavariable in question, that tries to match against some parent:

```
metavariable-pattern:
  metavariable: $MVAR
  pattern: parent
```

IagoAbal:
```ruby
rules:
- id: test
  message: Test
  languages: [ruby]
  severity: WARNING
  patterns:
  - pattern: |
      class $CLS
        ...
      end
  - metavariable-pattern:
      metavariable: $CLS
      pattern: Foo
```

This doesn't produce any matches on the _test.rb_ file from the PR test plan. Should it match?

IagoAbal: Another example could be:

```yaml
rules:
- id: test
  message: Test
  languages: [ruby]
  severity: WARNING
  patterns:
  - pattern: |
      class $B < $A
        ...
      end
  - metavariable-pattern:
      metavariable: $A
      pattern: Foo
```

only matching `Bar < Foo` but not matching `Baz < Bar` ?

```ruby
class Bar < Foo
end

class Baz < Bar
end
```

nmote: Thanks for the counterexamples! I'll look into this and most likely end up reverting this commit.

PR #5370, opened 2022-05-31
feat: Use `value($X)` to interpolate the value of a metavar

emjin:
Currently, you can use `$X` in a message to interpolate the variable captured
by a metavariable named $X, but there is no way to access the underlying value.
However, sometimes that value is more important than the captured variable.
This PR will introduce the syntax `value($X)` to interpolate the underlying
propagated value if it exists (if not, it will just use the variable name).

Other options considered:
- `$$X` -- decided that `value($X)` is more explicit
- use an options flag instead of adding syntax -- the benefit of this is that
this is more in line with how we have made similar modifications to access
the underlying value, but ultimately this is more unintuitive in a message
context and would have put a lot of distance between the message and the
option. Since messages are pretty separate from the pattern and since this
syntax contains all the information it needs to convey to a reader what is
happening, it seems ok to not have consistency.

Test plan: included tests

This also adds back a test I accidentally deleted

PR checklist:

- [x] Changelog is up-to-date


mjambon (approved):
It took me a while to process the description of the feature. I think an example would help (is there a limit to how much we can put into changelog entries?). Maybe I would show a template like `blah $X (value($X))` and then an actual message `blah n (42)`.

PR #5377, opened 2022-06-01
Add generic mode options

### LINKED ISSUE #2255: spacegrep: Memoize + get rid of the 10-line limitation in ellipsis

### LINKED ISSUE #3428: Generic mode should have a way to ignore comments

### LINKED ISSUE #5148: pattern-not-regex does not filter as expected in generic mode
**Describe the bug**
I have https://semgrep.dev/s/kw5O, where I am trying to remove findings that begin with a comment character using `pattern-not-regex`. In the past, the behavior of `pattern-not-regex` was to filter out anything that overlapped with the result of `pattern-not-regex`, but that appears not to be the case here.

This is blocking https://github.com/returntocorp/semgrep-rules/pull/2066. Desired completion date would be end-of-month.

**Environment**

**Use case**
This is blocking https://github.com/returntocorp/semgrep-rules/pull/2066.

### LINKED ISSUE #5211: generic mode: Offer an option to set the maximum number of lines matched by an ellipsis '...'
There is a default limit of 10 lines to avoid bad performance. However, the performance isn't always bad and sometimes it's useful to:
- match a large number of lines, say 200
- match at most one line

The current solution for matching more than 10 lines is to repeat `...` e.g. `... ... ... ... ...` for a limit of 50 lines. There's no solution to matching fewer than 10 lines.

The proposed feature is to offer an option in the semgrep rule that sets this match limit.

The one-line limit would be useful for configuration formats that are line-oriented and don't use indentation to delimit blocks.
 See for example https://github.com/returntocorp/semgrep-rules/pull/2066.

mjambon:
The following new options can now be specified in a Semgrep rule:
* `generic_ellipsis_max_span`: for controlling how many newlines an ellipsis can match (#5211)
* `generic_comment_style`: for removing comments in a processing steps (#3428)

Uses https://github.com/returntocorp/pfff/pull/525

Closes #5211 
Closes #3428 
Closes #5148
Closes #2255

PR checklist:

- [x] Changelog is up-to-date


mjambon:
Marked as a draft because I haven't written the documentation yet. The implementation should already work.

mjambon:
Documentation PR: https://github.com/returntocorp/semgrep-docs/pull/558

mjambon:
I'll force-merge this because the tests were passing before I imported the changes from the main branch and the failure is unrelated to the branch changes.

PR #5384, opened 2022-06-01
Update release flow

ajbt200128:
PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,11 +1,14 @@
 # Create Draft Release
 name: release
 
+env:
+  TEST_RELEASE_SUFFIX: -test-release
+
 on:
   push:
     branches:
       # Sequence of patterns matched against refs/tags
-      - "**-test-release"
+      - ${{ format('**{0}',env.TEST_RELEASE_SUFFIX) }}
     tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -289,11 +292,39 @@ jobs:
           password: ${{ secrets.pypi_upload_token }}
           skip_existing: true
 
+  upload-wheels-test:
```
spencerdrak: If we can just replicate the `Public to Pypi` step, then that would be ideal - not sure if GHA supports conditional steps, however.

.github/workflows/release.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     name: Create the Github Release
     runs-on: ubuntu-latest
     needs: [upload-wheels, release-ubuntu-16-04]
+    if: ${{ !contains(github.head_ref,'-test-release') }}
```
spencerdrak: we should also abstract this logic into a variable. Couple ways to do this, but doing so at the top of the workflow would be great.

brendongo (approved):
@underyx and I were just talking about how the testing job should do this :D Nice!

spencerdrak (approved):

PR #5416, opened 2022-06-03
fix(junit_xml_output): failure content should not be a python list of string

### LINKED ISSUE #5223: junit-xml should join the stacktrace
**Describe the bug**
When producing a `--junit-xml` output, it show the stacktrace a a list of strings rather than a big strings. Maybe, it's only a `"\n".join(...)` away.

**To Reproduce**
`--junit-xml` :)

**Expected behavior**
The list of str should one str

**Screenshots**
Screenshot 2022-05-16 at 15-47-47 jenkins _ Pre-pricing_Tarantino _ test_semgrep _ #4

**Environment**
```
Running `semgrep ci` without API token but with configs ('p/golang', 'p/secrets', 'p/security-audit')
Scan environment:
  versions    - semgrep 0.92.1 on python 3.10.4
```

**Use case**
Wider adoption of semgrep.

bmahe:
The JUnit xml output would serialize the failure messages as a python list of strings.
This change addresses this problem by grouping the lines together.

Fixes #5223

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   the imported name (#3964)
 - Constant propagation is now faster and memory efficient when analyzing
   large functions with lots of variables.
+- The JUnit XML output should serialize the failure messages as a single
+  string instead of a python list of strings.
 
 ## [0.94.0](https://github.com/returntocorp/semgrep/releases/tag/v0.94.0) - 2022-05-25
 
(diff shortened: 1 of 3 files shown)


bmahe:
There is an issue with the order of the XML attributes not being the same on different platforms. I will investigate and fix it

greut:
> I will investigate and fix it

it was probably the sole reason for the `_etree_to_dict` function to exist.

brendongo (approved):

Issue #5421, opened 2022-06-03
Invalid SARIF: additionalProperty "fixes" exists in instance when not allowed


sim642:
**Describe the bug**
SARIF output produces invalid SARIF and thus fails to be uploaded to GitHub Security:
```
Error: Unable to upload "semgrep.sarif" as it is not valid SARIF:
- instance additionalProperty "fixes" exists in instance when not allowed
```

**To Reproduce**
I have the following GitHub actions workflow that fails with the above error at the upload step:
```yaml
jobs:
  semgrep:
    runs-on: ubuntu-latest

    container:
      image: returntocorp/semgrep

    steps:
      - name: Checkout code
        uses: actions/checkout@v3

      - name: Run semgrep
        run: semgrep scan --sarif --output=semgrep.sarif

      - name: Upload SARIF file to GitHub Advanced Security Dashboard
        uses: github/codeql-action/upload-sarif@v2
        with:
          sarif_file: semgrep.sarif
        if: always()
```

**Expected behavior**
SARIF file should be valid and successfully uploadable to GitHub Security, as it previously was.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [x] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
Docker image in GitHub Actions.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1433/invalid-sarif-additionalproperty-fixes-exists-in-instance-when-not. _Note_: this link is for r2c use only and is not accessible publicly.

sim642:
Semgrep generates a SARIF file, which begins with:
```json
{
    "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
    "fixes": [],
    "runs": [
```

According to SARIF 2.1.0, the `fixes` property is allowed for individual `result` objects, not at the top level:
https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317668.

Also that "$schema" link is a 404.

kristof-mattei:
Reading about the scheme some more it seems that `fixes` should be a child of a `result` object. However in https://github.com/returntocorp/semgrep/commit/e8a3ad7db5c64827c160b7d83ca52aaf6a367211 it was introduced as a top-level attribute.

From the docs:

[...]

PR #5425, opened 2022-06-03
Emit fixes local to the result, not as an aggregate

kristof-mattei:
Fix for https://github.com/returntocorp/semgrep/issues/5421

Testing https://github.com/returntocorp/semgrep/blob/5dbf6dd47ab4fda6e9dc8dec42b90fce24d1fd82/semgrep/tests/e2e/snapshots/test_output/test_sarif_output_with_autofix/results.sarif against https://github.com/oasis-tcs/sarif-spec/blob/master/Documents/CommitteeSpecifications/2.1.0/sarif-schema-2.1.0.json validates 100%.

I don't want to use that url though as the $schema in the files we generate, as that is not the spec. 

PR checklist:

- [x] Changelog is up-to-date
   didn't update as the prefix 'fixes' is still under 'unreleased': https://github.com/returntocorp/semgrep/blob/develop/CHANGELOG.md#added


semgrep/semgrep/formatter/sarif.py
```diff
@@ -67,7 +72,7 @@ def _rule_match_to_sarif_fix(rule_match: RuleMatch) -> Optional[Mapping[str, Any
                                 "endLine": rule_match.end.line,
                                 "endColumn": rule_match.end.col,
                             },
-                            "insertedContent": {"text": fixed_lines},
+                            "insertedContent": fixed_lines,
```
ajbt200128: Why remove text? See some examples [here](https://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html#_Toc517436085 ), it seems like it makes sense to keep it

semgrep/semgrep/formatter/sarif.py
```diff
@@ -185,15 +190,6 @@ def format(
...
-        rule_matches, rule_matches_fixes = tee(rule_matches)
-        fixes = [
-            self._rule_match_to_sarif_fix(rule_match)
-            for rule_match in rule_matches_fixes
-        ]
-        # Filter out rule matches w/no fixes
-        fixes = list(filter(None, fixes))
-
         output_dict = {
             "$schema": "https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json",
```
ajbt200128: Can we update this broken link to https://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/schemas/sarif-schema.json ?

semgrep/tox.ini
```diff
@@ -6,7 +6,7 @@ envlist = py37, py38, py39
 allowlist_externals = pipenv
 commands =
     pipenv install --sequential --dev
-    pipenv run pytest --ignore=tests/qa/test_public_repos.py -n auto -vv --tb=short
+    pipenv run pytest --ignore=tests/qa/test_public_repos.py -n auto -vv --tb=short {posargs}
```
ajbt200128: Please remove this line, we do want to run these QA tests, and this isn't relevant to the scope of the issue

ajbt200128 (changes requested):

semgrep/semgrep/formatter/sarif.py
kristof-mattei: Schema validation fails if we do.

semgrep/semgrep/formatter/sarif.py
kristof-mattei: Will do.

semgrep/tox.ini
kristof-mattei: Will do.

semgrep/semgrep/formatter/sarif.py
kristof-mattei: Update: https://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/schemas/sarif-schema.json it is in this one.

But it is not in this one: https://github.com/oasis-tcs/sarif-spec/blob/master/Documents/CommitteeSpecifications/2.1.0/sarif-schema-2.1.0.json

kristof-mattei: Let me see if we can check which validation GitHub uses.

kristof-mattei:
@ajbt200128 trying with https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/schemas/sarif-schema-2.1.0.json, which is the latest version.

ajbt200128:
> @ajbt200128 trying with https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/schemas/sarif-schema-2.1.0.json, which is the latest version.

Great, if it works, make sure to regenerate the tests so the output contains the new schema url :)

kristof-mattei:
@ajbt200128 had to join `text` into a single string as the scheme doesn't allow for `[]`. Ran tests, merged in changes. LMK.

ajbt200128:
> @ajbt200128 had to join `text` into a single string as the scheme doesn't allow for `[]`. Ran tests, merged in changes. LMK.

Good catch, can you escape the newlines though instead of removing them? That'd be in accordant w/ the json specs. After that we should be good

kristof-mattei:
@ajbt200128 I didn't remove them. `text` is an array when it comes from `semgrep`, and then they are joined with `\n`:
https://github.com/returntocorp/semgrep/pull/5425/files#diff-9a6ebaf633a3e083e4a7792fb44df39b96d443dc8641055c8cabc5f50bf8f3baR75

ajbt200128:
@kristof-mattei my bad, I misread, but either way, they should be joined with escaped new lines, as per the json spec

kristof-mattei:
@ajbt200128 they are escaped in the rendered output:

```
❯ python3
Python 3.10.4 (main, Mar 23 2022, 20:25:24) [GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> foo = ["hello", "world"]
>>> bar = { "foo": "\n".join(foo) }
>>> bar
{'foo': 'hello\nworld'}
>>> json.dumps(bar)
'{"foo": "hello\\nworld"}'
>>>
```

ajbt200128:
great!

ajbt200128 (approved):
Looks good!

aryx:
@ajbt200128 can you push this on the finish line?

michael-schwarz:
It would be nice if this could be merged soon, it's quite annoying to have the CI jobs fail on every commit because of this. :smile:

PR #5431, opened 2022-06-03
Don't scan the file system (to create test suites) on module initialization

mjambon:
What this fixed: We can now use `dune utop` from any location, not just from `semgrep-core/tests/`.

This is a problem with `dune utop` for example because it seems it will initialize all the modules on startup. This wasn't a problem in semgrep-core because only library modules that are referenced from non-library modules get initialized in executables. It's not a good idea to rely on this behavior anyway. It is better to assume that any top-level module might be initialized by the program that contains it.

Along the way, I found out that the test suite `Unit_metachecking.tests` was not executed. I left it as a TODO in `Test.ml` since one of its tests fails.

PR checklist:

- [x] Changelog is up-to-date


aryx (commented):
Thx for fixing that

PR #5442, opened 2022-06-06
build: Make Dockerfile cacheable

underyx:
PR checklist:

- [ ] Changelog is up-to-date


.github/workflows/build-tests.yml
```diff
@@ -1,107 +0,0 @@
-name: build-tests
```
brendongo, line 1: what happened to these?

underyx: moved to tests.yml!

- the events were essentially the same anyway
- I got to rename these jobs to match the same naming scheme as tests.yml
- the workflow diagram is also easier to understand now that all jobs are on the same page
- this will let us consolidate the dependencies a bit nicer going forward

semgrep/semgrep/commands/scan.py
```diff
@@ -766,11 +764,9 @@ def scan(
     possibly_notify_user()
 
     # change cwd if using docker
-    try:
+    if not targets:
```
brendongo, line 768: is adjust_for_docker guaranteed to succeed if targets is not null?

semgrep/semgrep/commands/scan.py
```diff
@@ -819,19 +815,17 @@ def scan(
     # The 'optional_stdin_target' context manager must remain before
     # 'managed_output'. Output depends on file contents so we cannot have
     # already deleted the temporary stdin file.
-    with converted_pipe_targets(target_sequence) as target_sequence:
+    with converted_pipe_targets(targets) as targets:
```
brendongo: seems iffy to shadow the variable like this

underyx: yeah I was scared to change the logic though in case we were previously somehow depending on the shadowing :/

semgrep/semgrep/commands/scan.py
underyx: no; but the try-except here was just re-raising the exception anyway! i.e. we always wanted to crash if that function fails

Dockerfile
```diff
@@ -99,5 +86,12 @@ RUN chmod +x /entrypoint.sh
 # Let the user know how their container was built
 COPY dockerfiles/semgrep.Dockerfile /Dockerfile
 
+COPY --from=semgrep-core /semgrep/semgrep-core/_build/default/src/cli/Main.exe /usr/local/bin/semgrep-core
+
+ENV SEMGREP_IN_DOCKER=1 \
```
brendongo, line 91: why does this need to be set at the end instead of with all the other env vars?

brendongo (approved):

semgrep/semgrep/commands/scan.py
brendongo: fair. can fix later if ever

Dockerfile
underyx: I do it so that we can decide to mess with these env vars and reuse all of the build cache from before

aryx:
@underyx can you explain what it means to mahe the "Dockerfile cacheable" and what are the implications
for people who want to modify the Dockerfile. Is there things they should avoid?

PR #5444, opened 2022-06-07
ci: Make python test jobs faster

underyx:
PR checklist:

- [ ] Changelog is up-to-date


.github/workflows/tests.yml
```diff
@@ -29,11 +27,26 @@ jobs:
...
       - name: Post-checkout fixes
         run: ./.github/post-checkout
+      - name: Remove broken symlinks
+        run: find -L . -type l -exec rm {} \;
+      - id: core-cache
+        name: Check if this semgrep-core binary is cached
+        uses: actions/cache@v3
+        with:
+          path: ocaml-build-artifacts.tgz
+          key: "${{ runner.os }}-core-binary-${{ hashFiles('semgrep-core/**') }}-${{ hashFiles('.github/workflows/**') }}"
```
underyx:
```suggestion
          key: "${{ runner.os }}-core-binary-${{ hashFiles('semgrep-core/**') }}-${{ hashFiles('.gitmodules') }}-${{ hashFiles('.github/workflows/**') }}"
```

semgrep/Makefile
```diff
@@ -45,7 +45,7 @@ kinda-quick-tests:
 .PHONY:
 regenerate-tests:
 	pipenv run pytest -n auto \
-	  --snapshot-update --ignore=tests/qa/test_public_repos.py \
+	  --snapshot-update --allow-snapshot-deletion \
```
brendongo, line 48: big!

semgrep/semgrep/terminal.py
```diff
@@ -19,8 +19,21 @@ class Terminal:
...
+        """Call this when semgrep is invoked as a CLI as opposed to a library."""
+        self.configure(quiet=False)
+
     def configure(
-        self, *, verbose: bool, debug: bool, quiet: bool, force_color: bool
+        self,
+        *,
+        verbose: bool = False,
+        debug: bool = False,
+        quiet: bool = True,
```
brendongo, line 35: should quiet be default true?

.github/workflows/tests.yml
```diff
@@ -17,23 +17,37 @@ jobs:
...
       - name: Post-checkout fixes
         run: ./.github/post-checkout
+      - name: Remove broken symlinks
+        run: find -L . -type l -exec rm {} \;
+      - id: core-cache
+        name: Check if this semgrep-core binary is cached
+        uses: actions/cache@v3
+        with:
+          path: ocaml-build-artifacts.tgz
+          key: "${{ runner.os }}-core-binary-${{ hashFiles('semgrep-core/**', '.github/workflows/tests.yml') }}-${{ steps.submodule-status.outputs.versions-hash }}"
```
brendongo: should we add changes to root Makefile to the cache key?

semgrep/semgrep/terminal.py
underyx: @brendongo yes, so that we don't log anything to stderr by default when semgrep is used as a library. There's an init_for_cli method that we call right away when a click command is invoked, which sets quite=False (until the proper settings are configured a few lines later.)

.github/workflows/tests.yml
```diff
@@ -268,46 +250,36 @@ jobs:
...
+        if: failure()
         run: |
           # the commit step that follows will fail to fetch the pfff submodule
           # (perhaps because of the github token's permissions)
           # so we disable recursive fetching
           git config fetch.recurseSubmodules false
       - name: Commit snapshot updates
         id: snapshot-commit
-        if: ${{ failure() && github.event_name == 'pull_request' }}
+        if: failure() && github.event_name == 'pull_request'
```
brendongo: Might be worth adding a guard not to do this if they dont have permissions (from a fork) so the failing tests dont get eclipsed by the error message saying they dont have perms

brendongo: Can do in follow up

.github/workflows/tests.yml
underyx:
```suggestion
          key: "${{ runner.os }}-core-binary-${{ hashFiles('semgrep-core/**', '.github/workflows/tests.yml', 'scripts/install-alpine-semgrep-core') }}-${{ steps.submodule-status.outputs.versions-hash }}"
```

Not the makefile cause it's not used for building here. But it's a good point that the install script should be part of this.

semgrep/skip_list.txt
```diff
@@ -1,7 +0,0 @@
-# This configuration file is used by semgrep-core -pycheck (as well
```
brendongo, line 1: did we not need this anymore?

.github/workflows/tests.yml
underyx:
```suggestion
        if: failure() && github.event_name == 'pull_request' && (github.actor != 'dependabot[bot]' && !(github.event.pull_request.head.repo.full_name != github.repository))
```

semgrep/tests/conftest.py
```diff
@@ -167,10 +141,11 @@ def _run_semgrep(
     strict: bool = True,
     quiet: bool = False,
     env: Optional[Dict[str, str]] = None,
-    fail_on_nonzero: bool = True,
+    assert_exit_code: Union[None, int, Set[int]] = 0,
     settings_file: Optional[str] = None,
     force_color: Optional[bool] = None,
     assume_targets_dir: bool = True,  # See e2e/test_dependency_aware_rule.py for why this is here
+    force_metrics_off: bool = True,
```
brendongo, line 148: Do we have a test that verifies metrics works now that everything might have it as off?

semgrep/skip_list.txt
underyx: I think this was for the `foo_.pyi` interface files which we removed a while ago.

semgrep/tests/conftest.py
underyx: `unit/test_metrics` and `e2e/test_metrics` are pretty extensive and I had to set `force_metrics_off=False` to get those tests to pass

brendongo: Nvm i see the test_metrics sets some to on

brendongo (approved):

Issue #2186, opened 2020-12-04
Better message for partial parsing errors


aryx:
**Is your feature request related to a problem? Please describe.**

Tree-sitter has some powerful error-recovery mechanism where a mistake in a file (or a programming language
feature not yet handled by the grammar) does not cause the whole file to fail; Instead tree-sitter tries its best
to parse what it can and to return a valid AST semgrep can work on.

Unfortunately, currently semgrep wrongly says it was not able to parse 1 file, even though it finds match in it.

See for example https://semgrep.dev/s/clintgibler:c-ifdef-mid-function-call?version=develop
if you copy the source in a file and run in the command-line semgrep you'll get:
```
(semgrep) pad@yrax:~$ semgrep -l c -e 'SOME_FUNC(...)' semgrep/tests/c/misc_ifdef_strings.c 
semgrep/tests/c/misc_ifdef_strings.c
5:  SOME_FUNC(
6:    setexpr,
7:    "some text",
8:    "more text\n"
9:    "other text"
10:#ifdef CONFIG_REGEX
11:    "\n"
12:    "even more text"
13:#endif
14:  );
ran 1 rules on 1 files: 1 findings
1 files could not be analyzed; run with --verbose for details or run with --strict to exit non-zero if any file cannot be analyzed
```

**Describe the solution you'd like**

The last line should be instead
```
1 file could not be fully analyzed; run with ...
```

Maybe it could be useful to return a different kind of error when we know it was a partial parse error.
Maybe the JSON returned by semgrep could be a PartialParseError instead of ParseError.

Still, an easy fix is probably to just change the message to "could not be _fully_ analyzed".
If a file can not parse at all, the sentence is still true :)

**Describe alternatives you've considered**

**Additional context**

aryx:
cc @mjambon again.

aryx:
Ping

mjambon:
Oops, sorry, my email filters were set up incorrectly so this got buried. I'll look into it.

dan5382:
@nbrahms any update on this or link to a related ticket?

ievans:
@dan5382 why did this get closed?

IagoAbal:
The message has already changed since this issue was opened, now Semgrep will say:

> found problems analyzing 1 file; run with ...

If you then pass `--verbose` you can see the parse error.

```
% cat foo.js 
x = x + 1

foo {

% semgrep -l js -e 'x' --verbose foo.js
semgrep warn: parse error
[...]

PR #5446, opened 2022-06-07
Return the skipped tokens (tree-sitter) information to semgrep-cli

aryx:
test plan:
make test

```
$ semgrep-core -l ts -e '$X == $X' skip_tokens.ts  -json
+ /home/pad/yy/_build/default/src/cli/Main.exe -l ts -e '$X == $X' skip_tokens.ts -json
.
{"matches":[],"errors":[{"rule_id":"-e/-f","error_type":["PartialParsing",[{"path":"skip_tokens.ts","start":{"line":2,"col":15,"offset":0},"end":{"line":2,"col":16,"offset":1}},{"path":"skip_tokens.ts","start":{"line":9,"col":15,"offset":0},"end":{"line":9,"col":16,"offset":1}}]],"severity":"warning","location":{"path":"skip_tokens.ts","start":{"line":2,"col":15,"offset":0},"end":{"line":2,"col":16,"offset":1}},"message":"`)` was unexpected"}],"skipped":[],"stats":{"okfiles":0,"errorfiles":1}}
```
PR checklist:

- [x] Changelog is up-to-date


aryx:
This will also help https://github.com/returntocorp/semgrep/issues/2186

semgrep-core/src/core/Xtarget.ml
```diff
@@ -27,5 +27,5 @@ type t = {
   lazy_content : string lazy_t;
   (* This is valid only for xlang = Xlang.L ..., not for LRegex|LGeneric *)
   lazy_ast_and_errors :
-    (AST_generic.program * Semgrep_error_code.error list) lazy_t;
+    (AST_generic.program * Parse_info.token_location list) lazy_t;
```
IagoAbal, line 30: Don't we want to allow for any kind of error here? We could always filter the ones that are interesting for the CLI later. I am trying to update DeepSemgrep and we have:

```
    try
    let { Parse_target.ast; skipped_tokens; _} = 
          Parse_target.parse_and_resolve_name
          lang file in
    file, (ast, skipped_tokens)
    with
    | exn -> file, ([], [exn_to_error file exn])
```
What to do with that `exn` then?

aryx: ok, I'll try to update deep-semgrep and see if we need to go revert some code in semgrep.

aryx: Here the only errors we were getting were always skipped_tokens, so I preferred to be precise.
The parsing function can also return exns (e.g., ParseError, LexicalError, etc.), but once we get an AST,
the only possible thing attached to it are skipped_tokens.

PR #5450, opened 2022-06-07
fix(cli): Small reword of the `semgrep cli` output.

bmahe:
The output of `semgrep ci` should be clear it is exiting with error code 0 when there are findings but none of them being blockers

Here is an example of the new output:
```
  Found 5 findings (0 blocking) from 1374 rules.
  Uploading findings to Semgrep App.
```

PR checklist:

- [ ] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


underyx (approved):

bmahe:
The only failing check is unrelated to this PR and due to a docker login command, which expects to be run in the main repo. Since everything else is green and this pr got reviewed and approved, I will merge it.
Thanks again @underyx !

PR #5453, opened 2022-06-07
fix: scanning wrong thing in github actions

brendongo:
GitHub Actions exposes a file with Event Webhook information and we relied on the sha of
the base to calculate the mergebase of the compare head and the base head. According to
GHA documentation, the event[pull_request][base][sha] is actually frozen at PR open time (though
investigations show it does get updated at least when you merge base into head but does not update
if any new PRs are merged onto base or if base is pulled into head through a third branch).

Because of this, any merges from base to compare branch will result in use not using the updated
merge-base. This results in semgrep reporting findings that were introduced in the base branch
after PR creation time.

PR checklist:

- [ ] Changelog is up-to-date


semgrep/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
```diff
@@ -0,0 +1,48 @@
...
+=== end of stdout - plain
+
+=== stderr - plain
+Scan environment:
+  versions    - semgrep <MASKED> on python <MASKED>
+  environment - running in environment github-actions, triggering event is pull_request
+  semgrep.dev - authenticated as org_name
+
+Fixing git state for github action pull request
+Not on head ref: 17b3114fb3d41437abc2a3e6499db84ad921e6f0; checking that out now.
```
brendongo, line 29: These commits should be stable based on how they were created so no need to mask them

brendongo: Using f7ee31 shows that we now use the correct merge-base. Older versions of semgrep would use 191a3ac

semgrep/semgrep/meta.py
```diff
@@ -96,12 +96,8 @@ def commit_sha(self) -> Optional[str]:
...
         return git_check_output(["git", "rev-parse", "HEAD"])
 
-    @property
-    def head_ref(self) -> Optional[str]:
-        return None
-
     @cachedproperty
-    def base_commit_ref(self) -> Optional[str]:
+    def merge_base_ref(self) -> Optional[str]:
         return self.cli_baseline_ref
```
underyx, line 101: might it be confusing that here it's actually whatever was passed on the CLI, and not necessarily a merge base ref?

semgrep/semgrep/meta.py
```diff
@@ -163,7 +156,7 @@ def to_dict(self) -> Dict[str, Any]:
             "pull_request_id": self.pr_id,
             "pull_request_title": self.pr_title,
             "scan_environment": self.environment,
-            "is_full_scan": self.base_commit_ref == None,
+            "is_full_scan": self.merge_base_ref == None,
```
underyx:
```suggestion
            "is_full_scan": self.merge_base_ref is None,
```

semgrep/semgrep/meta.py
```diff
@@ -215,25 +208,89 @@ def commit_sha(self) -> Optional[str]:
...
+        Split out shallow fetch so we can mock it away in tests
+        """
+        subprocess.run(
+            [
+                "git",
+                "fetch",
+                "origin",
+                "--depth",
+                "1",
+                "-fu",
```
underyx: how dare you speak to me like that 😤

underyx:
```suggestion
                "--depth=1",
                "--force",
                "--update-head-ok",
```

semgrep/semgrep/meta.py
```diff
@@ -250,9 +307,10 @@ def _find_branchoff_point(self, attempt_count: int = 0) -> str:
                     "git",
                     "fetch",
                     "origin",
+                    "-fu",
```
underyx:
```suggestion
                    "--force",
                    "--update-head-ok",
```

semgrep/semgrep/meta.py
```diff
@@ -263,7 +321,15 @@ def _find_branchoff_point(self, attempt_count: int = 0) -> str:
                 f"Base branch fetch: args={process.args}, stdout={process.stdout}, stderr={process.stderr}"
             )
             process = subprocess.run(
-                ["git", "fetch", "origin", "--depth", str(fetch_depth), self.head_ref],
+                [
+                    "git",
+                    "fetch",
+                    "origin",
+                    "-fu",
```
underyx:
```suggestion
                    "--force",
                    "--update-head-ok",
```

semgrep/semgrep/meta.py
```diff
@@ -215,25 +208,89 @@ def commit_sha(self) -> Optional[str]:
...
+        """
+        # Fetch latest
+        self._shallow_fetch_branch(branch_name)
+        branch_rev_parse = subprocess.run(
+            ["git", "rev-parse", branch_name],
+            encoding="utf-8",
+            check=True,
+            timeout=GIT_SH_TIMEOUT,
+            capture_output=True,
+        ).stdout.rstrip()
```
underyx:
```suggestion
        ).stdout.strip()
```

if this works as well, let's use `strip()` like elsewhere in this file, otherwise we'll forever be wondering why `rstrip` is needed for this one method 😅

underyx:
```suggestion
        branch_rev_parse = git_check_output(
            ["git", "rev-parse", branch_name]
        )
```

underyx: ^we should use this helper for every git command we run at the start of `Meta` initialization to print the nice git error message we have

underyx (approved):

semgrep/semgrep/meta.py
brendongo: Hmm you right. I'll clean up in follow up

PR #5469, opened 2022-06-08
ci: Skip core build step if we have a binary cached

underyx:
PR checklist:

- [ ] Changelog is up-to-date


underyx:
draft because actions/cache doesn't work as expected: https://github.com/actions/cache/issues/817

PR #5470, opened 2022-06-08
Add pattern based id

ajbt200128:
PR checklist:

- [ ] Changelog is up-to-date

Of note, is that we still need to update the reference to semgrep interfaces, when that corresponding PR is merged


semgrep/semgrep/rule.py
```diff
@@ -248,6 +250,47 @@ def has_runnable_rule(d: Dict[str, Any]) -> bool:
...
+                for k in RuleValidation.PATTERN_KEYS:
+                    next_raw = raw.get(k)
+                    if next_raw is not None:
+                        patterns += get_subrules(next_raw)
+                        # We want to make sure if this a join, we actually include
+                        # the "on" field
+                        if k == "join" and "on" in next_raw:
+                            patterns += get_subrules(next_raw["on"])
+                return patterns
+            elif isinstance(raw, list):
```
ajbt200128, line 273: should sort list here :)

semgrep/semgrep/rule.py
```diff
@@ -248,6 +250,47 @@ def has_runnable_rule(d: Dict[str, Any]) -> bool:
...
+        Used to calculate a pattern based ID, works through DFS of all
+        PATTERN KEYS, and additionally, if the key is 'join' we include the 'on' field also
+        """
+
+        def get_subrules(raw: Union[AnyStr, Dict, List]) -> str:
+            patterns = ""
+            if isinstance(raw, str):
+                return raw
+            elif isinstance(raw, dict):
+                for k in RuleValidation.PATTERN_KEYS:
```
ajbt200128: Do we feel confident in the order of this being stable? This is part of how we navigate the tree, so maybe we should sort it here, but it's also a constant, so maybe we should just manually sort it or leave a comment?

brendongo: I think we should sort it ourselves. After 3.7 sets should by default be ordered but can't hurt to sort the 10 elements here in case changes are made to PATTERN_KEYS

semgrep/semgrep/rule_match.py
```diff
@@ -246,6 +256,23 @@ def get_syntactic_id(self) -> str:
         hash_bytes = int.to_bytes(hash_int, byteorder="big", length=16, signed=False)
         return str(binascii.hexlify(hash_bytes), "ascii")
 
+    @pattern_based_id.default
+    def get_pattern_based_id(self) -> str:
```
aryx: Need a comment explaining why this is better than simply using the rule ID.

Still think this is a bad name since it's actually analyzing multiple patterns in a formula, so formula_based_id would be better, but then it does not accomodate for the fact that we also using the metavars content, so it's a match_based_id.

ajbt200128: So do we want to switch the name to match_based_id? I'm ok with whatever

ajbt200128:
blocked by [semgrep-interfaces#6](https://github.com/returntocorp/semgrep-interfaces/pull/6)

semgrep/semgrep/rule_match.py
ajbt200128: Going to go ahead and rename it

semgrep/semgrep/rule_match.py
```diff
@@ -324,8 +359,11 @@ class RuleMatchSet(Set[RuleMatch]):
     It also automagically adds the correct finding index when adding elements.
     """
 
-    def __init__(self, __iterable: Optional[Iterable[RuleMatch]] = None) -> None:
+    def __init__(
+        self, rule: Rule, __iterable: Optional[Iterable[RuleMatch]] = None
```
brendongo, line 363: Hmm I'm a little worried about associating a Rule to a RuleMatchSet. On one hand it makes sense that deduplication only matters for matches with the same rule anyway but it was nice that this data-structure worked even if the matches you put in it were from different rules.

Should a RuleMatch just maintain a reference to the Rule it matched with so we can create the match_formula_string without keeping state in RuleMatchSet. Thoughts @underyx?

semgrep/tests/conftest.py
```diff
@@ -85,8 +85,6 @@ def _clean_output_json(output_json: str) -> str:
             p = r.get("path")
             if p and "/tmp" in p:
                 r["path"] = "/tmp/masked/path"
-                # the fingerprint contains the path too
-                r["extra"]["fingerprint"] = "0x42"
```
brendongo, line 96: Nice! Ok to remove since with the change in this PR fingerprint should be stable

semgrep/semgrep/rule_match.py
brendongo: Looks like @ajbt200128 you had the same idea as me: https://github.com/returntocorp/semgrep/pull/5470/files#diff-235785e37f4fac53689eae23436d86080f6e09721a4904ed674be882e4f38505R80-R81

semgrep/semgrep/rule_match.py
```diff
@@ -246,6 +256,30 @@ def get_syntactic_id(self) -> str:
...
+        if self.extra.get("metavars") is not None:
+            metavars = self.extra["metavars"]
+            for metavar in metavars:
+                match_formula_str = match_formula_str.replace(
+                    metavar, metavars[metavar]["abstract_content"]
+                )
+        match_id = (match_formula_str, path, self.rule_id)
+        match_id_str = str(match_id)
+        return (
+            f"{hashlib.blake2b(str.encode(match_id_str)).hexdigest()}_{str(self.index)}"
```
brendongo, line 280: Is there a reason we're picking this hash function?

ajbt200128: Yes, https://www.blake2.net/ it's faster than md5, sha-1,2,3 AND cryptographic!

ajbt200128: It's used by openssl, wireguard, and the linux kernel too

semgrep/semgrep/rule_match.py
ajbt200128: Yea, I figured this made sense since "set" means deduping, and I'd actually disagree that it's nice it work across different rules, I actually think that doesn't make sense. If we have to dedup across rules that doesn't make sense to me. If two rules have two matches that are considered duplicates, wouldn't the rules be duplicates? As far as just keeping a rule reference, we can do that but we still would rely on RuleMatchSet for the index, which the new hash needs, so it felt safer to couple the logic together.

brendongo (approved):

semgrep/semgrep/rule_match.py
brendongo: it's less that we would use the datastructure to dedup findings that were duplicates and more that: before associating a rule to the rulematchset, the rulematchset could contain matches from any rule and them being not duplicates would just work.

brendongo: Actually can we add a check to make sure we are not adding a rulematch whos rule_id doesnt match the RuleMatchSet's rule

brendongo (approved):

semgrep/semgrep/rule_match.py
underyx: I don't fully grasp this change but I will say that this class is supposed to be a native Python `set` just with extra features, so we shouldn't change the function signature from the original `def __init__(self, __iterable)`. If we need to add more stuff, I'd replace this with a type that keeps data in a set in an instance variable.

ajbt200128: Will do that

PR #5473, opened 2022-06-09
Dataflow: translate spread operators in record expressions to IL

mmcqd:
Closes [PA-1464](https://linear.app/r2c/issue/PA-1468)
Now expressions like `{...foo}` or `{a:"a",...foo}` are translated into the IL, so they are affected by taint. We still do not handle the case were a spread operator is used in the last position of an _lvalue_ record (`let {a:"a",...foo} = e`).

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/core/il/IL_lvalue_helpers.ml
```diff
@@ -56,7 +56,13 @@ let rec lvals_of_exp e =
   | Composite (_, (_, xs, _))
   | Operator (_, xs) ->
       lvals_of_exps xs
-  | Record ys -> lvals_of_exps (ys |> Common.map snd)
+  | Record ys ->
+      lvals_of_exps
+        (ys
+        |> Common.map @@ function
```
aryx, line 62: not a huge fan of those @@

aryx (approved):
What I don't understand is why the code before was not doing the right thing. I thought when there was a construct we were not handling, at least we were extracting all the expressions in it,
and generating some FixmeExp so at least the visitor would get to that 'source' buried inside '...source'.

IagoAbal:
I can't find traceability to the GitHub and Linear tickets that this PR is closing, neither in the commit message nor in the changelog entry.

**Updated with suggestion:** So in this case it appears there is no GitHub issue, but we could still add "Closes PA-1468" to the commit message, and if you used the branch name `matthew/pa-1468` then Linear will automatically insert a link and update the status of the ticket.

mmcqd:
@IagoAbal my bad, I should have linked the Linear issue/used the branch name.

semgrep-core/src/core/il/IL_lvalue_helpers.ml
mmcqd: I think `Common.map @@ function ...` looks nicer than `Common.map (function ...)`, but I don't feel too strongly about it.

IagoAbal:
> What I don't understand is why the code before was not doing the right thing. I thought when there was a construct we were not handling, at least we were extracting all the expressions in it,
> and generating some FixmeExp so at least the visitor would get to that 'source' buried inside '...source'.

So we have such a mechanism, the `FixmeExp` node takes an optional "partial" translation.  That said, right now it's only used for `OtherExpr`. In principle we could be using it in other places, but we haven't done that yet. The problem is that you need to identify a subexpression that you can correctly translate, and in many cases when a translation fails we don't have such a subexpression at hand (although we could try to extract them).

semgrep-core/tests/OTHER/rules/taint_spread_record.js
```diff
@@ -0,0 +1,2 @@
+//ruleid: test
+sink({...source})
```
IagoAbal: What would `{...source}` be, some kind of conversion from one record type to another record type where the field names are shared?

semgrep-core/tests/OTHER/rules/taint_spread_record.yaml
```diff
@@ -0,0 +1,12 @@
+rules:
```
IagoAbal: I would also add the original example from PA-1468.

IagoAbal (approved):
I guess what we really want to do with the `Spread` is to "resolve" it? But that's certainly more work so this is a good first step. I'm just in doubt whether the `Spread` constructor that we added to `IL.field` should be called something else (`FixmeSpread` or something?) but never mind me.

semgrep-core/tests/OTHER/rules/taint_spread_record.js
mmcqd: Should have had more examples, `{...source}` would be identical to `source` actually, but you can also do something. like `{...foo, ...bar}`, where this is a record that has all the key/value pairs of both `foo` and `bar`

mmcqd:
> I guess what we really want to do with the `Spread` is to "resolve" it? But that's certainly more work so this is a good first step. I'm just in doubt whether the `Spread` constructor that we added to `IL.field` should be called something else (`FixmeSpread` or something?) but never mind me.

In general it won't be possible to resolve it I think, since you can have code like `(a,b) => {...a, ...b}`, where the key/value pairs are completely unknown.

PR #5482, opened 2022-06-09
refactor: Rewrite test_ci, test_dep_aware, test_shouldafound

### LINKED ISSUE #5163: Add environment variable for disabling version checks
Right now I have to pass `--disable-version-check` manually a lot of the time and in many places because I don't want semgrep to worry about keeping itself up to date (I have other things to take care of that, such as Renovatebot).

It should be possible to disable version checks once by setting an environment variable. Something like `SEMGREP_VERSION_CHECKS=off`.

Continuing to do what I'm already doing - pass the existing flag everywhere.

**Use case**
Unnecessary network traffic is cut down. This also makes it more palatable to run in corporate environments, where these version checks may simply always fail.

underyx:
- Use run_semgrep fixture in some tests modules to disable metrics
- Added new utilities that create nicer snapshots with less code; snapshots now include the command that ran, the exit code, and the output streams with and without color codes.
- Fixed `get_no_findings_msg` making version check request with `--disable-version-check` (This is the only change within semgrep)
- run_semgrep output sanitization is refactored and now always runs every sanitizer implicitly (as well as takes more sanitizers on a per-snapshot basis)
- JSON indentation is now 2 spaces everywhere (previously it was a mixture of 2 and 4)
- the result of run_semgrep is now a dataclass instead of a tuple, where stuff like `.stdout` is accessible
- run_semgrep can now take a STDIN string which will be fed to the semgrep CLI execution
- run_semgrep no longer takes settings_file as a param, instead it can be passed as an env var
- run_semgrep now works properly with any subcommand, because it doesn't always pass a target, doesn't require an output format, and doesn't set flags like `--disable-version-check` by default; instead it disables this stuff via env vars
- run_semgrep now mocks the version cache directory for each execution

Closes https://github.com/returntocorp/semgrep/issues/5163


semgrep/semgrep/app/version.py
```diff
@@ -154,9 +154,9 @@ def version_check() -> None:
 
 def get_no_findings_msg() -> Optional[str]:
     """
-    Gets and returns the latest no_findings message from the backend, using cache if possible.
+    Gets and returns the latest no_findings message from the backend from cache.
     """
-    latest_version_object = _get_latest_version()
+    latest_version_object = _get_latest_version(allow_fetch=False)
```
brendongo, line 163: why is it ok to not fetch here?

semgrep/semgrep/commands/scan.py
```diff
@@ -297,6 +297,7 @@ def convert(
         "--enable-version-check/--disable-version-check",
         is_flag=True,
         default=True,
+        envvar="SEMGREP_ENABLE_VERSION_CHECK",
```
brendongo, line 300: Add to changelog?

brendongo (approved):

PR #5492, opened 2022-06-10
fix: Errors regarding using /src in docker image

### LINKED ISSUE #5479: (docker) semgrep working directory location changed without notice
**Describe the bug**
We run semgrep in CI using the `returntocorp/semgrep` docker image (latest tag).

Today, CI started failing with the following error:

```
WARNING: unable to find a config; path `sep-semgrep-rules` does not exist (since you are running in docker, you cannot specify arbitrary paths on the host; they must be mounted into the container)
invalid configuration file found (1 configs were invalid)
7
```

Which was quite surprising. All previous semgrep versions (including 0.96.0) worked perfectly fine.

Upon local testing, it seems that the working directory must've changed in the 0.97.0 docker image. We were using a relative location for the `config` directory (e.g. `--config semgrep-rules`) rather than `/src/semgrep-rules`). If I switch the `--config` location to be an absolute path, things work fine. 

**To Reproduce**

Create working directory with semgrep rules in `semgrep-rules`. Run:

`docker run --rm -v "${PWD}:/src" returntocorp/semgrep semgrep --config=semgrep-rules`

(That command will work fine in all prior semgrep versions, but will break in 0.97.0)

**Expected behavior**
Ideally, the behavior wouldn't have stopped working at all.

I understand not wanting users to rely on relative paths (not sure why we were), but this should've been noted in the documentation (https://semgrep.dev/docs/getting-started/).

In addition, the behavior change should've first been noted as a deprecation warning when running similar to what was done for the entrypoint change recently made.

**Environment**
semgrep docker image running on MacOS and GitLab CI.

**Use case**
At this point nothing since we had to change our config anyway. Reverting the change quickly may prevent other users/customers from having CI failures.

Updating the documentation to specify that you should not use relative paths will likely prevent future inconveniences by users.

### LINKED ISSUE #5488: PermissionError: [Errno 13] Permission denied: '/src'
when i run  "docker run -v "${PWD}:/src" returntocorp/semgrep semgrep --config=auto”

then error occuered:
```shell
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/semgrep/__main__.py", line 20, in main
    cli()
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/semgrep/cli.py", line 588, in cli
    semgrep.config_resolver.adjust_for_docker()
  File "/usr/local/lib/python3.9/site-packages/semgrep/config_resolver.py", line 443, in adjust_for_docker
    os.chdir(SRC_DIRECTORY)
PermissionError: [Errno 13] Permission denied: '/src'
```

underyx:
This reimplements https://github.com/returntocorp/semgrep/pull/5442/commits/6ef4d854f2251a0c46672647406a868d500be26a
in a safer way.

The WORKDIR /src is readded to the Dockerfile,
and we now check if it contains any files instead of its existence.

Fixes https://github.com/returntocorp/semgrep/issues/5479
Fixes https://github.com/returntocorp/semgrep/issues/5488

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/scripts/validate-docker-build.sh
+++ b/scripts/validate-docker-build.sh
@@ -59,6 +59,6 @@ echo "if 1 == 1: pass" \
     | docker run -i "$image" semgrep -l python -e '$X == $X' - \
     | grep -q "1 == 1"
 
-mkdir foo
-echo "if 1 == 1: pass" > foo/bar.py
-docker run -v foo:/src -i "$image" semgrep -l python -e '$X == $X' | grep -q "1 == 1"
+TEMP_DIR=$(mktemp -d)
+echo "if 1 == 1: pass" > "${TEMP_DIR}/bar.py"
+docker run -v "${TEMP_DIR}:/src" -i "$image" semgrep -l python -e '$X == $X' | grep -q "1 == 1"
--- a/scripts/validate-docker-build.sh
+++ b/scripts/validate-docker-build.sh
@@ -58,3 +58,7 @@ docker run "$image" publish --help
 echo "if 1 == 1: pass" \
     | docker run -i "$image" semgrep -l python -e '$X == $X' - \
     | grep -q "1 == 1"
+
+mkdir foo
+echo "if 1 == 1: pass" > foo/bar.py
+docker run -v foo:/src -i "$image" semgrep -l python -e '$X == $X' | grep -q "1 == 1"
--- a/semgrep/semgrep/config_resolver.py
+++ b/semgrep/semgrep/config_resolver.py
@@ -413,11 +413,15 @@ def adjust_for_docker() -> None:
     """change into this folder so that all paths are relative to it"""
     env = get_state().env
     if env.in_docker and not env.in_gh_action:
-        if not env.src_directory.exists():
+        try:
+            # check if there's at least one file in /src
+            next(env.src_directory.iterdir())
+        except (NotADirectoryError, StopIteration):
             raise SemgrepError(
                 f"Detected Docker environment without a code volume, please include '-v \"${{PWD}}:{env.src_directory}\"'"
             )
-        os.chdir(env.src_directory)
+        else:
+            os.chdir(env.src_directory)
 
 
 def indent(msg: str) -> str:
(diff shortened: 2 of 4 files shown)


semgrep/semgrep/config_resolver.py
```diff
@@ -413,11 +413,15 @@ def adjust_for_docker() -> None:
...
+        try:
+            # check if there's at least one file in /src
+            next(env.src_directory.iterdir())
+        except (NotADirectoryError, StopIteration):
             raise SemgrepError(
                 f"Detected Docker environment without a code volume, please include '-v \"${{PWD}}:{env.src_directory}\"'"
             )
-        os.chdir(env.src_directory)
+        else:
+            os.chdir(env.src_directory)
```
underyx, line 424: should we just remove this? the WORKDIR should be setting it already anyway

scripts/validate-docker-build.sh
```diff
@@ -58,3 +58,7 @@ docker run "$image" publish --help
 echo "if 1 == 1: pass" \
     | docker run -i "$image" semgrep -l python -e '$X == $X' - \
     | grep -q "1 == 1"
+
+mkdir foo
+echo "if 1 == 1: pass" > foo/bar.py
+docker run -v foo:/src -i "$image" semgrep -l python -e '$X == $X' | grep -q "1 == 1"
```
underyx: This should verify that docker mount scans work

semgrep/semgrep/config_resolver.py
brendongo: I think potentially?

brendongo (approved):

PR #5493, opened 2022-06-10
deep: taint-mode: Always report taint findings on the sink match

IagoAbal:
Helps PA-1484
Follows: 1d5db279508 ("deep: taint-mode: Record call chain to reach sink (#4836)")

We experimented with matching always code in the function being analyzed,
but that seems to be more confusing for people. Semgrep users expect
matches on the sink, and then the call trace should explain them how the
taint is reaching the sink.

Besides, this also caused some "duplicate" findings that introduced
regressions in semgrep-rules, e.g. if every `request` is tainted:

    def f(request):
      # This is reported
      # ruleid: test
      sink(request)

    def g(request):
      # This is also reported !!!!! Regression ?????
      # `f` is called with `g`'s request which is also tainted, and that
      # taint reaches the `sink` too...
      f(request)

Whereas the user seems to expect matches only on the `sink`'s line.
The test suite reports a regression because it doesn't expect that extra
match.

(Note that this only concerns DeepSemgrep, in the good-old Semgrep this
issue doesn't apply because our FOSS taint tracking is intraprocedural!)

test plan:
make test

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Match_tainting_rules.ml
+++ b/semgrep-core/src/engine/Match_tainting_rules.ml
@@ -206,28 +206,26 @@ let taint_config_of_rule default_config equivs file ast_and_errors
       sinks = sinks_ranges;
     } )
 
-let pm_of_finding file finding =
+let pm_of_finding finding =
   match finding with
-  | T.SrcToSink { source = _; tokens = _; sink; merged_env } -> (
-      match sink with
-      | T.PM sink_pm -> Some { sink_pm with env = merged_env }
-      | T.Call (fun_call, _trace, _) -> (
-          let code = G.E fun_call in
-          match V.range_of_any_opt code with
-          | None ->
-              logger#error
-                "Cannot report taint finding because we lack range info";
-              None
-          | Some range_loc ->
-              let tokens = lazy (V.ii_of_any code) in
-              Some
-                {
-                  PM.rule_id = (T.pm_of_trace sink).rule_id;
-                  file;
-                  range_loc;
-                  tokens;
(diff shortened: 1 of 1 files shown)


mmcqd:
So the expectation would be there is only one match, but the call trace should report that there are two possible ways for this sink to get tainted input?

mmcqd (approved):

semgrep-core/src/engine/Match_tainting_rules.ml
```diff
@@ -206,28 +206,26 @@ let taint_config_of_rule default_config equivs file ast_and_errors
...
-              Some
-                {
-                  PM.rule_id = (T.pm_of_trace sink).rule_id;
-                  file;
-                  range_loc;
-                  tokens;
-                  env = merged_env;
-                }))
+  | T.SrcToSink { source = _; tokens = _; sink; merged_env } ->
+      (* We always report the finding on the sink that gets tainted, the call trace
```
aryx, line 212: great comment!

aryx (approved):

IagoAbal:
> So the expectation would be there is only one match, but the call trace should report that there are two possible ways for this sink to get tainted input?

Yes, that would be a reasonable way of doing it. Although even if you get two different findings, as long as they are firing on the same line, that at least won't cause a regression in our test suites.

aryx:
yep, works great!! Far easier to debug taint rules.
Used now in deep-semgrep! https://github.com/returntocorp/deep-semgrep/pull/186

PR #5499, opened 2022-06-13
Enable using information about whether functions return constants

emjin:
This will help DeepSemgrep

PR checklist:

- [ ] Changelog is up-to-date


semgrep-core/src/analyzing/Dataflow_svalue.mli
```diff
@@ -1,5 +1,7 @@
 type mapping = AST_generic.svalue Dataflow_core.mapping
+type constness_type = Constant | NotAlwaysConstant [@@deriving show]
```
emjin, line 6: Could reuse G.Cst? But might be nice to separate this out

semgrep-core/src/analyzing/Constant_propagation.mli
```diff
@@ -12,4 +12,8 @@ val constant_propagation_and_evaluate_literal :
  *)
 (* !Note that this assumes Naming_AST.resolve has been called before! *)
 val propagate_basic : Lang.t -> AST_generic.program -> unit
+
+val propagate_dataflow_one_function :
```
aryx, line 19: Could be nice to document the parameter, especially what are those IL.name list?

semgrep-core/src/analyzing/Dataflow_svalue.mli
aryx: then maybe good to add a comment saying you know about G.Cst, but that you deliberately defined another one.

aryx (approved):

semgrep-core/src/analyzing/Dataflow_svalue.ml
```diff
@@ -51,10 +53,77 @@ let warning tok s =
...
 
+let hook_constness_table_of_functions = ref None
+
+let result_of_function_call_is_constant lang f args =
+  let check_f f_name =
+    match !hook_constness_table_of_functions with
+    | Some constness_f -> (
+        match constness_f f_name with
+        | Some Constant ->
+            logger#info "%s is always constant" f_name;
```
IagoAbal: Better use `logger#trace` or else the output of `semgrep --debug` becomes very large, and most of the times this info won't be of much use. Same for the other two uses of `logger#info` below.

semgrep-core/src/analyzing/Dataflow_svalue.ml
```diff
@@ -51,10 +53,77 @@ let warning tok s =
...
+            true
+        | Some NotAlwaysConstant ->
+            logger#info "%s is not always constant" f_name;
+            false
+        | None ->
+            logger#info "%s has not been read" f_name;
```
IagoAbal: Nitpicking here but what does `"not been read"` mean in this context? That we have not yet processed it? I was expecting a message more like `"we have no information about the constness of %s"`.

IagoAbal (approved):
LGTM!

semgrep-core/src/analyzing/Dataflow_svalue.ml
emjin: I was thinking of it as "we haven't read through the function body", but your wording makes sense

PR #5501, opened 2022-06-13
bugfix: metavariable-pattern: and language: php now work

### LINKED ISSUE #5443: metavariable-pattern not flagging with PHP code
**Describe the bug**
Metavariable-pattern is flagging PHP code with 'generic' mode, but not with 'php' mode.

**To Reproduce**
For instance, this doesn't flag: https://semgrep.dev/s/WRD2

But this does: https://semgrep.dev/s/02Wj

**Expected behavior**
Would expect it to flag with php as the language.

**Priority**
How important is this to you?
- [ ] P0: blocking me from making progress
- [X] P1: this will block me in the near future
- [ ] P2: annoying but not blocking me

aryx:
This closes #5443

test plan:
test file included
make test

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Metavariable_pattern.ml
+++ b/semgrep-core/src/engine/Metavariable_pattern.ml
@@ -20,6 +20,22 @@ module G = AST_generic
 
 let logger = Logging.get_logger [ __MODULE__ ]
 
+(*****************************************************************************)
+(* Prelude *)
+(*****************************************************************************)
+
+(*****************************************************************************)
+(* Helpers *)
+(*****************************************************************************)
+
+let adjust_content_for_language (xlang : Xlang.t) (content : string) : string =
+  match xlang with
+  | Xlang.L (Lang.Php, _) -> "<?php " ^ content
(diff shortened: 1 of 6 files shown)


aryx:
This closes PA-1448

semgrep-core/src/engine/Metavariable_pattern.ml
```diff
@@ -20,6 +20,22 @@ module G = AST_generic
...
+(* Prelude *)
+(*****************************************************************************)
+
+(*****************************************************************************)
+(* Helpers *)
+(*****************************************************************************)
+
+let adjust_content_for_language (xlang : Xlang.t) (content : string) : string =
+  match xlang with
+  | Xlang.L (Lang.Php, _) -> "<?php " ^ content
```
IagoAbal, line 33: I don't know PHP, but I'm curious if this works if there is a double `<?php` ? (That is, if `<?php` is already in the `content`).

PR #5510, opened 2022-06-13
fix(js): Parse `New` as a separate construct from `Call`

nmote:
See the corresponding `pfff` commit for background.

Note that this changes the output on a `semgrep-rules` test. It looks
like this is because the `Call` pattern also happened to match `New`
expressions, since `New` expressions were represented as `Call`s. I'm
not sure how to address this. The new behavior seems correct, but it is
also clear that the rule author intended the rule to match both
constructor calls and function calls. I would appreciate reviewer
feedback on this issue.

Corresponding semgrep-rules PR:
https://github.com/returntocorp/semgrep-rules/pull/2130

Test plan: Automated tests

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Fixed a non-deterministic crash when matching a large number of regexes (#5277)
 - Fixed issue when running in GithubActions that caused semgrep to report on
   files not changed in the PR (#5453)
+- JS/TS: `$X()` no longer matches `new Foo()`, for consistency with other languages (#5510)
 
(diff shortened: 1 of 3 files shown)


aryx:
you need to make a PR on semgrep-rules too, and link it here, so we can see what are the required changes.

aryx:
I think it's ok to have a function call pattern (like $FUNC(...)) not matching a new expressions.

nmote:
I think the failure is due to the job that pulls down a copy of semgrep-rules rather than using the submodule. I think this is good to go!

nmote:
The corresponding semgrep-rules PR is merged so this should be good to go.

mmcqd (approved):

PR #5511, opened 2022-06-14
PHP: Support named arguments in patterns + refactor arguments in AST

### LINKED ISSUE #5508: Semgrep does not support named parameters in PHP
**Describe the bug**
PHP parsing does not support named parameters in functions [(supported in PHP 8+)](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments)

in PHP 8 and higher, the following is valid PHP:

```php
setcookie("name","value",0,secure:true);
```

Currently, this fails to parse with an "invalid PHP" error in Semgrep as of 0.97.0

```
[ERROR] Pattern parse error in rule untitled_rule:
 Invalid pattern for PHP:
--- pattern ---
setcookie(...,secure:true)

--- end pattern ---
Pattern error: Parsing.Parse_error
Engine(Pattern parse error)
...
```

**To Reproduce**
https://semgrep.dev/s/ykv8

**Expected behavior**
This should parse - it's valid PHP for PHP 8 and above.

**Screenshots**
N/A

**Environment**
semgrep.dev, CLI version 0.97.0

**Use case**
This is a requirement for T1 PHP rule coverage.

mmcqd:
Fixes https://github.com/returntocorp/semgrep/issues/5508 / [PA-1497](https://linear.app/r2c/issue/PA-1497/semgrep-does-not-support-named-parameters-in-php)
This PR:
1. Updates the `pfff` PHP parser to support named arguments (`f(x:true)`).
2.  Refactors the representation of arguments in `ast_php.ml`. Previously we had the weird situation where function arguments were wrapped in special constructors in `cst_php.ml`, then unwrapped to plain expressions in `ast_php.ml`, and the re-wrapped in `php_to_generic.ml`. Now they are also wrapped in `ast_php.ml`, so their representation is uniform.
3. Updates the tree-sitter parser to parse named arguments as `ArgKwd`. Previously the label was just dropped.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/parsing/pfff/php_to_generic.ml
+++ b/semgrep-core/src/parsing/pfff/php_to_generic.ml
@@ -284,14 +284,14 @@ and expr e : G.expr =
       let v1 = expr v1 and v2 = expr v2 in
       G.DotAccess (v1, t, G.FDynamic v2)
   | New (v0, v1, v2) ->
-      let v1 = expr v1 and v2 = list expr v2 in
+      let v1 = expr v1 and v2 = list argument v2 in
       let t = H.expr_to_type v1 in
-      G.New (v0, t, fb (v2 |> List.map G.arg))
+      G.New (v0, t, fb v2)
   | NewAnonClass (_tTODO, args, cdef) ->
       let _ent, cdef = class_def cdef in
-      let args = list expr args in
+      let args = list argument args in
       let anon_class = G.AnonClass cdef |> G.e in
-      G.Call (anon_class, fb (args |> List.map G.arg))
+      G.Call (anon_class, fb args)
   | InstanceOf (t, v1, v2) ->
       let v1 = expr v1 and v2 = expr v2 in
       G.Call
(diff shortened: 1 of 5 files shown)


mmcqd:
Forgot to use the Linear generated branch name :/

mjambon (approved):

semgrep-core/src/parsing/pfff/php_to_generic.ml
```diff
@@ -435,9 +435,12 @@ and match_ = function
...
       G.CasesAndBody ([ G.Default tok ], G.ExprStmt (e, G.sc) |> G.s)
 
-and argument e =
-  let e = expr e in
-  G.arg e
+and argument = function
+  | Arg e -> expr e |> G.arg
+  | ArgRef (tok, e) -> G.Ref (tok, expr e) |> G.e |> G.arg
+  | ArgUnpack (tok, e) ->
+      G.OtherExpr (("Unpack", tok), [ G.E (expr e) ]) |> G.e |> G.arg
```
Sjord: Shouldn't this be `Call (IdSpecial Spread, Id foo)`? See [this comment](https://github.com/returntocorp/semgrep/blob/3849f73dccdda2b5ff461fc2bff9cf17572df3b3/semgrep-core/src/core/ast/AST_generic.ml#L838).

Sjord (approved):

aryx (approved):

semgrep-core/src/parsing/pfff/php_to_generic.ml
mmcqd: It should be, yep

PR #5519, opened 2022-06-14
Update workflow to allow checks to run without open/close PR

spencerdrak:
PR checklist:

- [x] Changelog is up-to-date

Changes were testing in a separate repo and workflow succeeded.


.github/workflows/start-release.yml
```diff
@@ -61,10 +61,27 @@ jobs:
           git add --all
           git commit -m "chore: Bump version to ${SEMGREP_RELEASE_NEXT_VERSION}"
           git push --set-upstream origin ${{ steps.release-branch.outputs.release-branch }}
+      - name: Get JWT for r2c-argo GitHub App
+        id: jwt
+        uses: docker://public.ecr.aws/y9k7q4m1/devops/cicd:latest
```
brendongo, line 66: This seems sketchy

underyx: would you be willing to consider using github actions workflows to build and github container registry to store public images? we do this already at
- https://github.com/returntocorp/ocaml-layer/blob/master/.github/workflows/release.yml ➡️ https://github.com/orgs/returntocorp/packages/container/package/ocaml-layer
- https://github.com/returntocorp/nginx-docker ➡️ https://github.com/orgs/returntocorp/packages/container/package/nginx
- https://github.com/returntocorp/git-cliff-docker ➡️ https://github.com/orgs/returntocorp/packages/container/package/git-cliff

underyx: but also wow I cannot tell at all from just reading what this step does

how come there's no `run` instruction? 😶

.github/workflows/start-release.yml
```diff
@@ -61,10 +61,27 @@ jobs:
...
+          ISSUER: ${{ secrets.SEMGREP_CI_APP_ID }} # semgrep-ci GitHub App id
+          PRIVATE_KEY: ${{ secrets.SEMGREP_CI_APP_KEY }}
+      - name: Get token for semgrep-ci GitHub App
+        id: token
+        run: |
+          TOKEN="$(curl -X POST \
+          -H "Authorization: Bearer ${{ steps.jwt.outputs.jwt }}" \
+          -H "Accept: application/vnd.github.v3+json" \
+          "https://api.github.com/app/installations/${{ secrets.SEMGREP_CI_APP_INSTALLATION_ID }}/access_tokens" | \
+          jq -r .token)"
```
underyx, line 78: I think this can be replaced with something like

`gh api app/installations/${{ secrets.SEMGREP_CI_APP_INSTALLATION_ID }}/access_tokens --jq .token`

spencerdrak: will look into this

.github/workflows/start-release.yml
spencerdrak: So this is an image that we (@cgdolan ) created to grab the JWT for us and we use in other places in the repo. I'll see about pushing this to another repo. I think ideally we push to a single repo, though, and use that, so if we migrate here, we'll have to migrate other jobs as well. I'm going to just place this on hold until I chat with him more about that image.

brendongo (approved):

.github/workflows/start-release.yml
spencerdrak: chatted with @underyx - this isn't working as expected, so we'll wave off.

.github/workflows/start-release.yml
spencerdrak: chatted with @brendongo - cleared.

PR #5521, opened 2022-06-14
Parse booleans as constants

### LINKED ISSUE #5509: Boolean constant propagation does not behave correctly in PHP
**Describe the bug**
Constant propagation with boolean values does not behave as advertised in PHP.

A rule like 

**To Reproduce**
To illustrate the issue, see the following (correct) Python example:
https://semgrep.dev/s/kPvX

Identical behavior in PHP is not matched as expected:
https://semgrep.dev/s/w2vx

**Expected behavior**
Boolean constant propagation is consistent across languages.

**Screenshots**

**Environment**
semgrep.dev, CLI 0.97.0

**Use case**
Required for T1 PHP vuln coverage & PHP GA.

emjin:
Closes https://github.com/returntocorp/semgrep/issues/5509

This allows constant propagation to happen on booleans in PHP by parsing
booleans as constants instead of Id. It adds a new construct to the AST for
bool. This was decided after reading the manual description of boolean, which
does not specify it as equivalent to an int value (see comment).

Test plan: make test

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
```diff
@@ -291,6 +291,18 @@ let map_integer env tok =
   let value = int_of_string value in
   A.Int (Some value, tok)
 
+let map_boolean env tok =
+  let bool_of_string value =
+    let true_regexp = Str.regexp "['T''t']['R''r']['U''u']['E''e']" in
+    let false_regexp = Str.regexp "['F''f']['A''a']['L''l']['S''s']['E''e']" in
+    if Str.string_match true_regexp value 0 then true
+    else if Str.string_match false_regexp value 0 then false
```
nmote: I like to avoid regexes where possible, since they come with performance penalties, occasional bits of surprising behavior, and generally are not super readable. I'd prefer to replace this with something like the following:

```suggestion
    let canonicalized_value = String.lowercase_ascii value in
    if canonicalized_value = "true" then true
    else if canonicalized_value = "false" then false
```

semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
```diff
@@ -291,6 +291,18 @@ let map_integer env tok =
...
+    else failwith ("Not a valid PHP boolean: " ^ value)
```
nmote, line 301: Is there something less severe than a `failwith` that would be appropriate here?

nmote (approved):
Overall looks good, just a stylistic suggestion and a question inline.

semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
mmcqd: I agree, that regex seems overpowered for this task

semgrep-core/src/parsing/tree_sitter/Parse_php_tree_sitter.ml
emjin: I don't think so :( I'll fix it later if it turns out there is

PR #5528, opened 2022-06-14
Change rule type to use polymorphic variants

kopecs:
Instead of using a normal variant for the mode field this changes the
rule type (`Rule.rule` -> `'mode Rule.rule_info`; `Rule.rule` is now an
alias for this with `'mode` as either a polymorphic variant for search
or taint) to be parameterised over a type variable `'mode` which is
instantiated to be specific polymorphic variants depending on the mode
of the rule.

This allows us to have functions which take/return rules where the modes
are restricted to a subset of all possible modes. Having this capability
is useful for if/when we have additional modes and, e.g., we want to
have a function take both search and taint rules but not rules having
other modes.

This just does the initial refactoring to change the main rule type and
some low-hanging fruit around the
`Match_{search,tainting}_mode.check_rule` types. There is probably some
additional refactoring which can be done to avoid redundant matches.

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/core/Rule.ml
```diff
@@ -226,20 +216,34 @@ and paths = {
...
 
+type search_rule = [ `Search of common * pformula ] [@@deriving show]
+type taint_rule = [ `Taint of common * taint_spec ] [@@deriving show]
+type rule = [ search_rule | taint_rule ] [@@deriving show]
+
 (* alias *)
 type t = rule [@@deriving show]
 type rules = rule list [@@deriving show]
 
+let id (`Search (c, _) | `Taint (c, _)) = c.id
```
mmcqd: Having to write all these helpers and keep them updated is pretty annoying, wonder if there's a ppx deriving extension that will do this?

semgrep-core/src/core/Rule.ml
```diff
@@ -226,20 +216,34 @@ and paths = {
...
+let fix_regexp (`Search (c, _) | `Taint (c, _)) = c.fix_regexp
+let paths (`Search (c, _) | `Taint (c, _)) = c.paths
+let metadata (`Search (c, _) | `Taint (c, _)) = c.metadata
+
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
 
-let partition_rules rules =
+let partition_rules (rules : rules) =
```
mmcqd: Don't need to fix it in this PR but this function seems kind of pointless now, since it basically just renames constructors

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -95,25 +95,25 @@ let check ~match_hook ~timeout ~timeout_threshold default_config rules xtarget =
...
+             let rule_id = R.id r |> fst in
              Rule.last_matched_rule := Some rule_id;
              Common.profile_code (spf "real_rule:%s" rule_id) (fun () ->
                  let match_result =
                    timeout_function r file timeout (fun () ->
                        (* dispatching *)
-                       match r.R.mode with
-                       | Search pformula ->
+                       match r with
+                       | `Search _ as r ->
```
mmcqd: You don't need `as r` here because `r` is already bound to the rule

semgrep-core/src/metachecking/Check_rule.ml
```diff
@@ -171,13 +171,13 @@ let check_formula env (lang : Xlang.t) f =
 (*****************************************************************************)
 
 let check r =
-  let rule_id = fst r.id in
+  let rule_id = R.id r |> fst in
   (* less: maybe we could also have formula_old specific checks *)
-  match r.mode with
-  | Rule.Search pf ->
+  match r with
+  | `Search (_, pf) as r ->
```
mmcqd: same deal with the `as r`

mmcqd (commented):
Will wait on others for approval, but generally looks good to me. I think this doesn't make much of a difference right now, but I can imagine the usefulness once we add your Extract mode and we start needing non singleton subsets of rule modes as arguments.

semgrep-core/src/engine/Match_rules.ml
kopecs: You do, because otherwise the compiler doesn't know we're in the specific variant.

semgrep-core/src/metachecking/Check_rule.ml
kopecs: Ditto

semgrep-core/src/core/Rule.ml
kopecs: I think it's still useful, because it lets us go from a mixed list to two unmixed lists (N-many for N modes).

mmcqd: actually just kidding I did not know what `partition` does lol

semgrep-core/src/core/Rule.ml
kopecs: I don't think there would be, and if we have some fields only shared for a couple modes then we'll probably have to do something like this anyway unfortunately. I don't think it'll be too burdensome though.

IagoAbal: It may be worth considering having a `common : rule -> common` function, and then write all the functions below in terms of that, e.g. `let message rule = (common rule).message`.

kopecs: My only concern there would be wrt how "common" some of the data search and taint share would be to other modes (e.g., `severity`, `fix`)

aryx:
meh, not convinced.

semgrep-core/src/core/Rule.ml
aryx: We could also make a polymorphic rule_
type 'mode rule_ = {
   id: ...
   message:
   mode: 'mode;
 }
and then define the mode = `Tainting of ... | `Search of ...
and then you can type rule = mode rule_
type tainting_rule = [`Tainting of ...] rule_
?

IagoAbal: I agree with @aryx here it's worth trying this. I believe subtyping also applies when using polymorphic variants as type parameters.

kopecs: > I believe subtyping also applies when using polymorphic variants as type parameters.

I couldn't find docs which mention variance explicitly, but a couple examples in the repl seemed to be fine, so I'm thinking the same.

Going to implement it like suggested above and see how much better/worse it is.

aryx (approved):

aryx:
better

PR #5529, opened 2022-06-14
test: Fix syntax error in misc_naming_recursion test

nmote:
I don't see anything in #2615 indicating that this syntax error was
intentional. The test still passes after this change.

PR checklist:

- [x] Changelog is up-to-date


aryx (approved):

aryx:
weird, why CI was working before?

nmote:
Not sure, but it failed when I tried to run DeepSemgrep on these tests which clued me in.

PR #5520, opened 2022-06-14
fix: Invalid_argument "index out of bounds"

nmote:
Fixes #5277

When scanning a repository with a large number of regex rules, Semgrep
creates numerous tmp files. For each one, we create a lookup table to go
from offset -> line/col, and cache it keyed on filename.

This cache does not get properly invalidated when Semgrep is done with
the tmp file, even though the tmp file does get removed. The next time a
tmp file is requested, there is a chance that the randomly-generated
filename will collide with a previous one. Semgrep will rely on the
cached lookup table for the previous file with that filename, and chaos
ensues. This appears to only be an issue in practice when a very large
number of regex matches are attempted while scanning a repo.

This commit applies a quick, but hacky fix. We simply reach into the
cache and remove the entry at the same time that we are done with the
tmp file. I'm not aware of the context around this area of the system so
I don't know how feasible it would be to remove the cache, avoid using
tmp files altogether, or pass the lookup table for a file around
explicitly. It would probably be good to do one of those things at some
point, but for now I think we should fix the issue quickly.

Test plan:

Automated tests

At the following commits:

semgrep: 3c78f29ef4aebd4983caf70abd87a400a7446e1d
terraform-provider-aws: 1f270b150c411268fdf456bd31e004a62a8808db

Run `while python -m semgrep -c .semgrep-service-name0.yml ; do echo; done` in the `terraform-provider-aws` repo.

Observe a crash in about 5 or fewer minutes.

Update semgrep to this commit, and run the command again for 1 hour.
Observe no crash.

PR checklist:

- [x] Documentation is up-to-date
- [x] Changelog is up-to-date
- [x] Change has no security implications (otherwise, ping security team)

PR #5530, opened 2022-06-14
refactor: Improve encapsulation of Xpattern_matcher cache

nmote:
As suggested in review of #5520.

Test plan: Follow test plan of #5520 but only run for 15 minutes.

PR checklist:

- [x] Changelog is up-to-date


nmote:
ocamlformat made some sweeping changes here, but `git show -w` confirms that the changes here are only what I intended.

aryx (approved):

aryx:
ocamlformat is great, but sometimes I'm a bit sad when I can't use git blame because it messes up badly the blame.
Fortunately I don't need that often git blame, so I'm ok with this cons of ocamlformat.

PR #5534, opened 2022-06-15
fix(python): Correct generic ast translation for context expressions (`with`)

### LINKED ISSUE #5513: Pattern match not applied inside Python "with" statement
**Describe the bug**
semgrep doesn't match inside a with statement

**To Reproduce**
https://semgrep.dev/s/J17o

**Expected behavior**
A pattern of "foo()" should match in a block of code like this:

```
with foo():
    pass
```

P2

**Environment**
official docker image / semgrep.dev

**Use case**
catching deprecated function used inside with statements

kopecs:
Closes #5513 / Fixes PA-1501. We previously translated the context expression for a with statement where the result was not bound as an empty list. This PR changes that to be a list containing the context expression itself.

Test plan: included

PR checklist:

- [x] Changelog is up-to-date


kopecs:
@aryx it looks like we used to have something similar (wasn't a list previously) before #4100. Would there be any specific reason this got changed there I should be aware of for this?

kopecs:
Looks like CI is failing due to a difference in matches on benchmarks. Seems like we're getting an extra true positive against zulip [here](https://github.com/zulip/zulip/blob/829f9272d2c4299a0c0a37a09802248d8136c0a8/zerver/management/commands/sync_ldap_user_data.py#L57), though it isn't clear to me why we would miss this previously as we should've been looking at the body of the `with` statement in any case (maybe improvements to const prop? --- not sure if that could affect this PR).

aryx:
good catch (and good blame). I probably did the mistake as it was part of some big refactoring.

aryx (approved):

aryx:
weird indeed we get an extra finding. There is a with transaction.atomic() above the finding, but it's hard to see
how it relates.
Anyway, lgtm.

PR #5540, opened 2022-06-15
fix: Match TyN pattern against TyExpr (N ...) source

nmote:
See the changed test for the goal. In JS, constructors can be arbitrary
expressions, so they are parsed into `TyExpr`. However, that leads to
missed matches.

The obvious solution is to desugar `TyExpr (N ...)` to `TyN` while
constructing the AST. I first tried to do this by modifiying
`AST_generic_helpers.expr_to_type` to do this desugaring, but I ran into
the regression test added in
https://github.com/returntocorp/semgrep/pull/4682.

It would be nice if there were a point where we could do this
desugaring, rather than having to handle it specifically in matching.
I'm not sure where it would make sense to do so, though.

Test plan: Automated tests

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1730,7 +1730,9 @@ and m_type_ a b =
   match (a.t, b.t) with
   (* this must be before the next case, to prefer to bind metavars to
    * MV.Id (or MV.N) when we can, instead of the more general MV.T below *)
-  | G.TyN a1, B.TyN b1 -> m_name a1 b1
+  | G.TyN a1, B.TyN b1
+  | G.TyN a1, B.TyExpr { e = N b1; _ } ->
```
mmcqd, line 1745: Could add a comment about why this is needed

mmcqd (approved):

nmote:
Oh, also I need to update the changelog

underyx:
@nmote security team is fixing the registry state which caused your test failure. Please don't apply the snapshot changes & feel free to merge with pytest failing.

nmote:
@underyx thanks, I was just looking into that. Wasn't going to apply without figuring out what was going on.

PR #5542, opened 2022-06-15
fix(join): Fix for taint rules in inline join rule

minusworld:
PR checklist:

- [x] Changelog is up-to-date


semgrep/semgrep/rule_schema.yaml
```diff
@@ -216,8 +229,6 @@ definitions:
             type: string
           regex:
             type: string
-          constant-propagation:
```
minusworld: This should not have been removed @minusworld

underyx (approved):

kurt-r2c (approved):

PR #5547, opened 2022-06-15
tainting: Add experimental taint-propagators feature

### LINKED ISSUE #4509: [JS] Taint does not propagate through `.push(...)` for arrays
**Describe the bug**
It appears that taint mode stops propagating data at `.push(...)` in array objects.

**To Reproduce**
This link has 3 examples. The third does not work, but should.
https://semgrep.dev/s/bXKp

IagoAbal:
Closes PA-1361
Closes #4509

test plan:
make test # added three tests

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   `SEMGREP_ENABLE_VERSION_CHECK=0`
 - Dataflow: spread operators in record expressions (e.g. `{...foo}`) are now translated into the Dataflow IL
 - An experimental LSP daemon mode for semgrep. Try it with `semgrep lsp --config auto`!
+- taint-mode: New experimental `taint-propagators` feature that allows to specify
+  arbitrary patterns for the propagation of taint by side-effect. In particular,
+  this allows to specify how taint propagates through side-effectful function calls.
+  For example, you can specify that when tainted data is added to an array then the
+  array itself becomes tainted. (#4509)
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


aryx (approved):
This is quite complicated. I didn't have the time to really understand the new code, but let's
merge this to avoid conflicts.

What are the alternatives to taint propagator?
What if we consider that each time you do    foo.bar(stuff), then if stuff is tainted then the object foo becomes tainted? We would get too many FPs after?

aryx:
Also I guess we want somewhere to discuss that this taint-propagators would work well with pattern-ref:, because we really
want to factorize in one place all the different way standard API functions behave regarding tainting no?

IagoAbal:
> This is quite complicated. I didn't have the time to really understand the new code, but let's merge this to avoid conflicts.

I'll give it another look and try to add some more comments!

> What are the alternatives to taint propagator? What if we consider that each time you do foo.bar(stuff), then if stuff is tainted then the object foo becomes tainted? We would get too many FPs after?

I tried to do exactly that as a quick hack in returntocorp/semgrep#5150 but then we got a bunch of FPs in semgrep-rules due to code like `out.println(tainted)` (`out` becomes tainted and then that causes an FP).

aryx:
Ok then good to add those examples and the pr that was implementing that in comment

IagoAbal:
> Also I guess we want somewhere to discuss that this taint-propagators would work well with pattern-ref:, because we really
> want to factorize in one place all the different way standard API functions behave regarding tainting no?

Indeed, I think there is an "pressing" need for `pattern-ref` or similar solution.

IagoAbal:
> Ok then good to add those examples and the pr that was implementing that in comment

👍 Preparing a follow-up PR.

Issue #4785, opened 2022-03-10
Semgrep crashes when working directory doesn't exist


underyx:
My shell was in a temporary directory that has been deleted since I cd'd in there. Running semgrep failed with this:

```
❯ semgrep -e 'echo' -l bash ~/r2c/semgrep/.github/post-checkout
Traceback (most recent call last):
  File "/opt/homebrew/bin/semgrep", line 33, in <module>
    sys.exit(load_entry_point('semgrep==0.84.0', 'console_scripts', 'semgrep')())
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/semgrep/__main__.py", line 20, in main
    cli()
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/semgrep/commands/scan.py", line 816, in scan
    ) = semgrep.semgrep_main.main(
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/semgrep/semgrep_main.py", line 298, in main
    all_rules = configs_obj.get_rules(no_rewrite_rule_ids)
  File "/opt/homebrew/Cellar/semgrep/0.84.0/libexec/lib/python3.10/site-packages/semgrep/config_resolver.py", line 312, in get_rules
    self._rename_rule_ids(configs)
[...]

PR #5548, opened 2022-06-15
Warn if in nonexistent dir

ajbt200128:
PR checklist:

- [X] Changelog is up-to-date


cli/src/semgrep/semgrep_main.py
```diff
@@ -52,7 +52,13 @@
 
 def get_file_ignore() -> FileIgnore:
     TEMPLATES_DIR = Path(__file__).parent / "templates"
-    workdir = Path.cwd()
+    try:
+        workdir = Path.cwd()
+    except FileNotFoundError:
+        workdir = Path.home()
+        logger.warn(
+            f"Current working directory does not exist! Using {workdir} as working directory"
```
brendongo: Maybe clarify that we are looking for .semgrepignore in workdir not that we are scanning workdir which is what this might sound like

cli/src/semgrep/config_resolver.py
```diff
@@ -307,7 +307,10 @@ def _safe_relative_to(a: Path, b: Path) -> Path:
     @staticmethod
     def _convert_config_id_to_prefix(config_id: str) -> str:
         at_path = Path(config_id)
-        at_path = Config._safe_relative_to(at_path, Path.cwd())
+        try:
+            at_path = Config._safe_relative_to(at_path, Path.cwd())
+        except FileNotFoundError:
+            pass
```
brendongo, line 311: Should we just be passing? No need to raise?

ajbt200128: Not sure why we would raise, this function is completely optional, there's an option to turn of rule name rewriting

cli/src/semgrep/semgrep_main.py
ajbt200128: Yea makes sense

cli/src/semgrep/config_resolver.py
brendongo: yeah but if it's turned on might be preferable to fail loudly than have the rules not have the prefix?

ajbt200128: It seems it's just a sorting/output difference, personally I'd rather semgrep run and have the output be slightly less organized than fail completely

brendongo: I guess that's fair. I was worried about that making it onto app but those rules dont have their path rewritten anyway.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
   files not changed in the PR (#5453)
 - JS/TS: `$X()` no longer matches `new Foo()`, for consistency with other languages (#5510)
 - JS/TS: Typed metavariables now match constructor calls (e.g. `($X: C)` matches `new C()`. (#5540)
+- Fixed bug where semgrep would crash in nonexistent directory
```
brendongo: Put in the wrong version btw

ajbt200128: Good catch

brendongo (approved):

aryx:
can we merge this?

IagoAbal:
> PA-951 Semgrep crashes when working directory doesn't exist

I would have appreciated adding "Closes PA-951" even if it feels a bit redundant due to the branch name, because then the link to the Linear ticket is already available in the Git log and it's one level of indirection less.

IagoAbal:
Does this close #4785 ? A reference to #4785 in the commit log would also have been appreciated. For the record, I'm checking the Git log against the changelog while preparing the release. These references are useful to have. 😃

PR #5551, opened 2022-06-16
Ensure we print "scan was limited to" when appropriate

ajbt200128:
This is totally hacky, I will admit. There's no way around manually checking if each target is a git repo, and therefore no way besides this to check if we skipped some files. We could cache the results of `files_from_git_ls` so we don't rerun git, but that also feels somewhat heavy handed. If we believe this is too hackish and there's no better way, and we don't want to merge this I will not argue

PR checklist:

- [X] Changelog is up-to-date


brendongo:
Can we not just check that we are in a valid git repo before printing https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/target_manager.py#L139 ?

ajbt200128:
> Can we not just check that we are in a valid git repo before printing https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/target_manager.py#L139 ?

No, each target could be a git repo, so we could be in say `~/` which isn't a git repo and run `semgrep --config auto ~/git_repo`. Basically what I'm doing is checking to see if any of the targets are git repos and if so print it. Since we respect the .gitignore of each target it wouldn't be accurate to only decide to print if git files were ignored if the CWD is a repo or not

brendongo:
Ah. That makes sense. Can you leave that (re: each target might be a .git) as inline for this block of code for future reference

brendongo (approved):

PR #3379, opened 2021-06-16
Merge changelog using union conflicts by default

nbrahms:
This prevents frequent conflict resolution of CHANGELOG.md.

PR #5577, opened 2022-06-20
Stop using union-merge for CHANGELOG.md

IagoAbal:
Union-merges are helpful in some cases... but they also cause the
changelog to become incorrect with virtually every release PR that is
merged. Unfonrtunately union-merge is quite dumb. Perhaps it's better
to deal with the conflicts manually but do it well.

Reverts: ffba1572570 ("Merge changelog using union conflicts by default (#3379)")

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,13 @@
-# Prevent frequent conflict resolution in the changelog by
-# asking git to always take both sides of a branch in the
-# changelog. As a downside, two edits to an existing line
-# in the changelog will show up as two lines in the merged
-# output. We should remove these duplicates at release time.
-CHANGELOG.md merge=union
+# This causes the changelog to become incorrect with virtually
+# every release PR that is merged. Unfortunately union-merges
+# are quite dumb. Perhaps it's better to deal with the conflicts
+# manually but do it well.
+# # Prevent frequent conflict resolution in the changelog by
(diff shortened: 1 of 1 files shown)


nmote:
This seems good. I don't think this will address the issue where changes get attributed to a previous release, though. Here's what I've noticed:

* I make a change to the changelog under the "Unreleased" header.
* A release gets made, changing "unreleased" to "v0.whatever"
* I rebase/merge, and since the release didn't change anything near the line that I added, Git doesn't make me resolve conflicts, and I never notice that I've retroactively updated a previous changelog.

I try to pay attention to this, but I suspect that it has slipped through the cracks, if not for me, then for others.

I still think this PR is a net positive, but I also think it's worth considering the issue above.

nmote (approved):

IagoAbal:
Yep, you are probably right. This seems like a problem that should affect every project where many people work via PRs. There must be some good solution(s)! Maybe https://about.gitlab.com/blog/2018/07/03/solving-gitlabs-changelog-conflict-crisis/ ? Time ago Bence pointed to a Python tool that implemented GitLab's solution (or something very similar). Basically, each changelog entry is a separate file in some 'unreleased/' folder, and as part of the release process you run a script that collects those files and generates a proper changelog. cc @spencerdrak

Issue #5432, opened 2022-06-04
internal: enable the test suite Unit_metachecking.tests


mjambon:
I found out during https://github.com/returntocorp/semgrep/pull/5431 that the test suite for metachecking wasn't executed. We need to [comment out a line](https://github.com/returntocorp/semgrep/blob/0b7bf3fa27bdf02b7e494d56e1b5429830b33f25/semgrep-core/tests/Test.ml#L69) and fix the test that's failing.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1436/internal-enable-the-test-suite-unit-metacheckingtests. _Note_: this link is for r2c use only and is not accessible publicly.

PR #5581, opened 2022-06-20
Avoid duplicate matches due to naming equivalences

### LINKED ISSUE #5569: Semgrep-core sending import resolution as metavariable value in python
**Describe the bug**
Semgrep reports two findings on the same match because the metavariables are different but the second metavariable value looks like in import resolution or similar

**To Reproduce**
Rule:
```
rules:
  - id: some-rule
    pattern: |
      $MODEL.query.$METHOD(...)
    message: >
      Any message that contains $MODEL
    severity: WARNING
    languages:
      - python
```
Target:
```
from foo.bar import Baz

class C:
    def f(self) -> None:
        Baz.query.filter()
```

https://semgrep.dev/s/02wq (Note two matches instead of 1)

image

**Environment**
Semgrep 0.98.0

**Use case**
Reduce duplicate rules

IagoAbal:
Closes PA-1520
Closes #5569

Related-to: fbd33396c64 ("fix(deep): Don't attempt metavariable matches against parent classes (#5368)")

test plan:

    % cat 5569.py
    from foo.bar import Baz

    class C:
        def f(self) -> None:
            Baz.query.filter()
    % bin/semgrep-core -lang py -e '$MODEL.query.$METHOD(...)' 5569.py
    5569.py:5
            Baz.query.filter()
    # ^^^ just one match rather than two

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 - Inline join mode rules can now run taint-mode rules
 - Python: correctly handle `with` context expressions where the value is not
   bound (#5513)
+- Avoid duplicate findings due to naming equivalences (#5569)
 
 ## [0.98.0](https://github.com/returntocorp/semgrep/releases/tag/v0.98.0) - 2022-06-15
 
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -682,7 +682,7 @@ and m_expr a b =
                         _sid );
                 };
               _;
-            } )) ) ->
+            } )) ) -> (
       (* We used to force to fully qualify entities in the pattern
        * (e.g., with org.foo(...)) but this is confusing for users.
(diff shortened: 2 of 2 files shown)


IagoAbal:
I'm aware of the test failures and I'll investigate.

IagoAbal:
I was applying the same quick-fix as in #5432, but thankfully in this case we had a test to show us that the fix is flawed!

Consider matching:
```yaml
    - patterns:
      - pattern: $REDIRECT(..., $FORMAT, ...)
      - metavariable-pattern:
          metavariable: $REDIRECT
          pattern-either:
            - pattern: django.shortcuts.redirect
            - pattern: django.http.HttpResponseRedirect
```
against:
```python
from django.http import HttpResponse
from django.http import HttpResponseRedirect
from django.shortcuts import redirect
from django.utils.http import is_safe_url


def unsafe(request):
    # ruleid: open-redirect
    url = request.headers.get("referrer")
    print("something")
    return redirect(url)
```
If we just don't try matching `$REDIRECT` against the fully-qualified `redirect` then the `metavariable-pattern` operator will fail because it does expect this fully qualified name!

Closing the PR...

PR #5621, opened 2022-06-23
feat: add `extract` as a new mode

kopecs:
This PR adds an additional mode `extract` which allows the extraction of source code in a different language from the containing file. Namely, an extract rule with `languages: [ X ]` and `dest-language: Y` will run the pattern against files of language X and produce new targets in language Y which search & taint rules will be run against. Currently extract rules will not run against the result of extract rules.

PR checklist:

- [x] Changelog is up-to-date

Test plan:

You can manually test the changes here with this small example:

`test.html`:
```html
<!DOCTYPE html>
<html>
    <body>
        <h2>Demo JavaScript in Body</h2>

        <p id="demo">A Paragraph</p>

        <button type="button" onclick="myFunction()">Try it</button>

        <script>
            function myFunction() { document.getElementById("demo").innerHTML = "Paragraph changed."; }
        </script>

    </body>
</html>
```

`rules.yaml`:
```
rules:
  - id: extract-javascript-from-html
    mode: extract
    pattern: |
        <script>$...BODY</script>
    extract: $...BODY
    languages:
      - html
    paths:
      include:
        - '*.html'
    dest-language: javascript
  - id: get-elem-id
    mode: search
    pattern: document.getElementById(...)
    languages:
      - javascript
    message: "Found match"
    severity: INFO
```

then the command `semgrep --config rules.yaml test.html` should produce:

```
Scanning 1 file.

Findings:

  test.html
     get-elem-id
        Found match

         11┆ function myFunction() { document.getElementById("demo").innerHTML = "Paragraph changed."; }

Ran 2 rules on 1 file: 1 finding.
```


cli/src/semgrep/core_runner.py
```diff
@@ -126,6 +126,7 @@ class StreamingSemgrepCore:
 
     This behavior is assumed to be that semgrep-core:
     - prints a "." on a newline for every file it finishes scanning
+    - prints a number on a newline for any extra targets produced during a scan
```
aryx, line 131: Why those things are part of this diff?
Did you change things for the extractor or it's just you use the opportunity to better document things?

aryx: What is an extra target? How semgrep-core can return extra target?

semgrep-core/src/core/Rule.ml
```diff
@@ -196,6 +196,18 @@ type taint_spec = {
 }
 [@@deriving show]
 
+type extract_spec = {
+  pformula : pformula;
+  (* TODO: determine if we need reductions beyond concat; maybe
+               "separate"?
+   * combine_extract: extract_mode;
+   *)
+  (* TODO: really want Lang.t | Generic *)
```
aryx: I don't see a use for using generic as a dst_lang but who knows.
Good to keep as a TODO for now.

semgrep-core/src/engine/Match_rules.ml
```diff
@@ -118,7 +118,12 @@ let check ~match_hook ~timeout ~timeout_threshold default_config rules xtarget =
                              Match_tainting_mode.check_rule { r with mode }
                                match_hook default_config xtarget
                            in
-                           res)
+                           res
+                       | `Extract extract_spec ->
+                           (* TODO: need another module for extract? *)
```
aryx: not unless things get really different between search and extract. Anyway your Extract.ml seems good enough for now to try to isolate all extractor specific stuff in a separate module.

semgrep-core/src/extracting/Extract.ml
```diff
@@ -0,0 +1,142 @@
+module In = Input_to_core_j
```
aryx, line 16: A toplevel copyright and a toplevel Prelude section explaining what is the
goal of this module.

semgrep-core/src/extracting/Extract.mli
```diff
@@ -0,0 +1,22 @@
+(* should only be extract rules *)
```
aryx: I guess you can remove the comment now that you enforce Rule.extract_rule below.

Then you can replace by a general comment explaining the goal of the function.

semgrep-core/src/extracting/Extract.ml
```diff
@@ -0,0 +1,142 @@
...
+      ~timeout ~timeout_threshold
+      (Config_semgrep.default_config, [] (* no equiv *))
+      (erules :> Rule.rules)
+      xtarget
+  in
+  (* mode for combination *)
+  (* match combine with
+     | "separate" -> *)
+  res.Report.matches
+  (* want Common.filter_map? *)
```
aryx: Common.ml was created a long time ago. If List now has filter_map, better to use it.
There's just Common.map we really prefer over List.map because it's more efficient.

semgrep-core/src/extracting/Extract.mli
```diff
@@ -0,0 +1,22 @@
...
+val extract_nested_lang :
+  match_hook:
+    (string -> Metavariable.bindings -> Parse_info.t list Lazy.t -> unit) ->
+  timeout:float ->
+  timeout_threshold:int ->
+  Rule.extract_rule list ->
+  Xtarget.t ->
+  string list ->
```
aryx: what is this string list?

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -463,28 +463,82 @@ let targets_of_config (config : Runner_config.t)
  * It takes a set of rules and a set of targets and
  * recursively process those targets.
```
aryx: maybe add a comment about the extraction step. How does it influence the regular matching process of semgrep_with_rules.

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -463,28 +463,82 @@ let targets_of_config (config : Runner_config.t)
  * It takes a set of rules and a set of targets and
```
aryx: maybe you can adjust with "and a set of targets (derived from [config])"

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -463,28 +463,82 @@ let targets_of_config (config : Runner_config.t)
...
  * recursively process those targets.
  *)
-let semgrep_with_rules config ((rules, invalid_rules), rules_parse_time) =
+let semgrep_with_rules config extractors
+    ((rules, invalid_rules), rules_parse_time) =
   (* Return an exception
```
aryx, line 558: Maybe you can add before the comment "We want to be tolerant to errors in rules. This is useful
when rules are using advanced features not present in older version of semgrep-core, but
still want this old semgrep-core to process the other (old-compatible) rules".

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -463,28 +463,82 @@ let targets_of_config (config : Runner_config.t)
...
+        | _ -> erules_parse_time)
+    | None -> 0.
+  in
+
   (match (rules, invalid_rules) with
   | [], [] -> ()
   | [], err :: _ -> raise (Rule.InvalidRule err)
   | _, err :: _ (* TODO fail fast when only when strict? *) ->
       raise (Rule.InvalidRule err)
   | _ -> ());
```
aryx: I guess you can replace the last pattern with | _, [] -> ())

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -463,28 +463,82 @@ let targets_of_config (config : Runner_config.t)
...
 
+  let rule_ids = Common.map (fun r -> fst r.R.id) rules in
+
+  let extracted_targets, extract_result_map =
```
aryx, line 579: move this code out of semgrep_with_rules in an helper function.
Maybe add a (* *** *) Extractor helpers section.
semgrep_with_rules() is already long, we don't want to make it too much longer.

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -512,11 +566,14 @@ let semgrep_with_rules config ((rules, invalid_rules), rules_parse_time) =
...
-           res)
+           Hashtbl.find_opt extract_result_map file
+           |> Option.fold ~some:(fun f -> f res) ~none:res)
   in
   let res =
-    RP.make_final_result file_results rules config.report_time rules_parse_time
+    RP.make_final_result file_results rules config.report_time
+      (erules_parse_time +. rules_parse_time)
   in
+  RP.show_final_result res |> pr2;
```
aryx: this call to pr2 seems wrong.
use logger if needed.

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -546,14 +603,54 @@ let semgrep_with_raw_results_and_exn_handler config =
    * semgrep-core -rules <(curl https://semgrep.dev/c/p/ocaml) ...
    *)
   let rules_file = replace_named_pipe_by_regular_file rules_file in
+  let extractors_file =
+    if config.extractors_file <> "" then
+      Some (replace_named_pipe_by_regular_file config.extractors_file)
```
aryx: I doubt we want to define extractor rule via named pipe. 
I think this replace_named_pipe_... was added so you could do in the shell semgrep  -e foo  $(<echo foo>) or something like that, but for -pp I don't think we need that feature.

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -546,14 +603,54 @@ let semgrep_with_raw_results_and_exn_handler config =
...
+    else None
+  in
   try
+    logger#linfo
+      (lazy (spf "Parsing %s:\n%s" rules_file (read_file rules_file)));
+    let timed_extractors =
```
aryx: again move in helper function in the 'Extract helper' section to not clutter the original function.

aryx (dismissed):

aryx:
Please provide a test plan: section in the PR indicating how to test this PR.
You can add an example of target file, extractor rule, like in tests/OTHER/extractor, and indicate the command to run.

cli/src/semgrep/core_runner.py
kopecs: This corresponds to a change in `Run_semgrep.ml`: https://github.com/returntocorp/semgrep/blob/524c85af3e8aeb866d80fabcbb0af09ab8fe2daa/semgrep-core/src/runner/Run_semgrep.ml#L520-L522

An "extra target" in this context would be the temporary files containing extracted text. If we want these to get a entry in the progress bar on the python side, we need some way to communicate the number of targets back to python. Right now the cli just calculates it ahead of calling core, so this gives us a mechanism for that.

semgrep-core/src/engine/Match_rules.ml
kopecs: Okay, sounds good.

semgrep-core/src/extracting/Extract.ml
kopecs: I wasn't sure if it made sense to implement the corresponding functionality in a tail recursive way or not. If we don't have it right now would you just rely on the stdlib implementation?

semgrep-core/src/runner/Run_semgrep.ml
kopecs: Ah this is a debug print that slipped through, thanks for the catch!

semgrep-core/src/runner/Run_semgrep.ml
kopecs: Will do.

semgrep-core/src/extracting/Extract.mli
kopecs: This is a list of rule ids we're interested in running. We need this information when generating the additional targets as each target has a list of rule ids to run against it. Normally this would be generated for us by the cli, but in this case because we're creating a new input in core we don't have that.

There's probably a type alias I can swap this to, and when I add docs I'll make it more clear.

semgrep-core/src/runner/Run_semgrep.ml
kopecs: Hmm actually wouldn't the same logic as in the comment apply, like if you wanted to run with `-pp <(curl $URL)`?

aryx: Oh you re right. Might be good to add (* this allows -pp (curl ...)

semgrep-core/src/core/Rule.ml
kopecs: I think the use-case would be for things like django templates, where you have a DSL which can appear in specific parts of the file only (although using generic for the whole file might be fine for those cases).

cli/src/semgrep/core_runner.py
```diff
@@ -163,33 +164,51 @@ async def _core_stdout_processor(
...
         # appease mypy. stream is only None if call to create_subproccess_exec
         # sets stdout/stderr stream to None
         assert stream
 
-        # Start out reading two bytes at a time (".\n")
-        bytes_to_read = 2
         while True:
             # blocking read if buffer doesnt contain any lines or EOF
-            line_bytes = await stream.read(n=bytes_to_read)
+            line_bytes = await stream.readline()
```
brendongo: I think we went from readlint to bytewise reading because we were filling up the pipe.

brendongo: Relevant: https://github.com/returntocorp/semgrep/issues/4693

I think we could keep track of where we are wrt total and pre-emptively breakout instead of trying to read the last line which is a giant json blob?

kopecs: Issue would be when `read` returns less than `n` bytes; per our conversation I'll just switch the termination condition to being based on the number of files scanned.

aryx:
When do you think this will be ready for review and merge? (It's still in draft right now).

kopecs:
Largely good (I can do the tests either with this or in a separate PR), I just want to get the validation working well. The current error messages would be a decently poor regression (imo); it seems like it might make sense to upgrade the version of JSON schema we're using so we can use [`dependentRequired`](https://json-schema.org/draft/2020-12/json-schema-validation.html#name-dependentrequired)---as it stands the validation schema is a bit of a mess.

emjin:
Can you include some form of test plan? That can be just pasting the contents of two short files with instructions for what to name them and what command to run. It would be helpful to see just one example of the output, maybe a second to show how bad the error messages become.

kopecs:
I'll update the original comment with that. The error messages I'm referring to there are mostly about the json schema validation. You can see the [pytest snapshot diff](https://github.com/returntocorp/semgrep/commit/90278d46f0838a31eff319e3cacb3345d0694c84) for that---basically it just dumps a bunch of keys that don't make sense (cf. behaviour with taint mode keys, as those don't show up).

cli/Pipfile
```diff
@@ -29,5 +29,21 @@ pytest-split = "*"
...
+# If we change to Poetry or another system which allows multiple version of
+# python these could likely be removed.
+#
+# Sourced from
+# <https://github.com/python-jsonschema/jsonschema/blob/afc22f09e74d696ab00be8a711bbc5c2a15327b7/pyproject.toml#L39-L41>
+#
+# These _should not_ be added to setup.py
+importlib_metadata = "*"
+typing_extensions = "*"
+importlib_resources = ">=1.4.0"
```
mmcqd, line 45: Will defer to @brendongo on the okayness of this lol

semgrep-core/src/extracting/Extract.ml
```diff
@@ -0,0 +1,181 @@
...
+let extract_nested_lang ~match_hook ~timeout ~timeout_threshold
+    (erules : Rule.extract_rule list) xtarget rule_ids =
+  let erule_table = mk_rule_table erules in
+  let res =
+    Match_rules.check ~match_hook ~timeout ~timeout_threshold
+      (Config_semgrep.default_config, [] (* no equiv *))
+      (erules :> Rule.rules)
+      xtarget
+  in
+  (* mode for combination *)
```
mmcqd, line 67: Probably want to delete this or explain what it is

semgrep-core/src/extracting/Extract.ml
```diff
@@ -0,0 +1,181 @@
...
+           Common.find_some_opt
+             (fun (x, mvar) ->
+               match
+                 Hashtbl.find_opt erule_table m.Pattern_match.rule_id.id
+               with
+               | None -> None
+               | Some r ->
+                   let (`Extract { Rule.extract; _ }) = r.mode in
+                   if x = extract then Some (r, Some mvar) else Some (r, None))
+             m.Pattern_match.env
```
mmcqd, line 82: This would be easier to read if it was computed using a helper function. It's hard to patch `match <a bunch of stuff> with ...`

semgrep-core/src/extracting/Extract.ml
```diff
@@ -0,0 +1,181 @@
...
+                 Metavariable.mvalue_to_any extract_mvalue
+                 |> Visitor_AST.range_of_any_opt
+                 |> Option.map (fun (start_loc, end_loc) ->
+                        let end_len = String.length end_loc.Parse_info.str in
+                        ( ( start_loc.Parse_info.charpos,
+                            (* subtract 1 because lines are 1-indexed, so the offset is
+                             * one less than the current line *)
+                            start_loc.line - 1,
+                            start_loc.column ),
+                          (start_loc.charpos, end_loc.charpos + end_len) ))
```
mmcqd, line 100: Same here. In general I think this function is a little monolithic. Giving names to intermediate results computed by helpers would make it clearer

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1359,7 +1359,7 @@ and m_attrs a b =
 and m_xml_bodies a b =
   match (a, b) with
   (* dots: *)
-  | [ XmlText ("...", _) ], _ -> return ()
+  | [ XmlText (s, _) ], _ when String.trim s = "..." -> return ()
```
mmcqd, line 1362: What's this for?

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -506,36 +506,87 @@ let targets_of_config (config : Runner_config.t)
...
+        Extract.extract_nested_lang ~match_hook ~timeout:config.timeout
+          ~timeout_threshold:config.timeout_threshold extractors xtarget
+          rule_ids
+      in
+      (* Print number of extra targets so Python knows *)
+      if config.output_format = Json && extract_targets <> [] then
+        pr (string_of_int (List.length extract_targets));
+      extract_targets)
+    extract_targets
+  |> Fun.flip
```
mmcqd: I would just use a lambda

mmcqd (approved):
Basically just style comments, which I know you've largely fixed in another branch. Feel free to just resolve them if they're fixed there.

semgrep-core/src/extracting/Extract.ml
kopecs: I have this fleshed out on the other one.

semgrep-core/src/matching/Generic_vs_generic.ml
kopecs: This is to make ellipses in XML/HTML tags a little easier to work with, so that e.g.,
```
<tag>
  ...
</tag>
```
still hits the above case.

Should I leave this in and add a changelog entry, remove it, or do you have another suggestion?

cli/Pipfile
kopecs: @underyx suggested this as a solution to pipenv's lack of support for multiple python versions.

underyx: could you leave a comment explaining the need for this tho?

kopecs: Is the above comment not sufficient? I'm not sure what else specifically to add.

cli/setup.py
```diff
@@ -125,7 +125,7 @@ def find_executable(env_name, exec_name):
     "ruamel.yaml>=0.16.0,<0.18",
     "tqdm~=4.46",
     "packaging~=21.0",
-    "jsonschema~=3.2",
+    "jsonschema~=4.6.1",
```
brendongo:
```suggestion
    "jsonschema~=4.6",
```
So minor version bumps are allowed?

cli/src/semgrep/core_runner.py
```diff
@@ -163,33 +166,50 @@ async def _core_stdout_processor(
...
 
         # appease mypy. stream is only None if call to create_subproccess_exec
         # sets stdout/stderr stream to None
         assert stream
 
         # Start out reading two bytes at a time (".\n")
-        bytes_to_read = 2
+        get_input: Callable[
+            [asyncio.StreamReader], Coroutine[Any, Any, bytes]
+        ] = lambda s: s.readexactly(2)
```
brendongo, line 185: Oooo shiny ✨

cli/setup.py
kopecs: Sounds good.

cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
```diff
@@ -132,6 +132,7 @@ Running semgrep-core with command:
 <MASKED>
 <MASKED>
 <MASKED>
+<MASKED>
```
brendongo, line 135: @kopecs did you add any lines that print to stderr? Probably innocuous but might be worth checking what this is exactly https://github.com/returntocorp/semgrep/blob/develop/cli/tests/e2e/test_ci.py#L534

cli/Pipfile
brendongo: Looks sufficient! Feel free to merge and we can add any comment that is missing later if ever

cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
kopecs: Appears to be the extra line from here: https://github.com/returntocorp/semgrep/blob/3395c61a76c77542f2c2238e1c3b001c21459d5c/semgrep-core/src/runner/Run_semgrep.ml#L522-L524

cli/Pipfile
underyx: oh sorry I missed that one; I thought `importlib_resources` was the only line relevant here so I expected the comment directly above this line

the one we have above this block is very well written!

semgrep-core/src/matching/Generic_vs_generic.ml
kopecs: Added changelog entry.

aryx (commented):
this is great!

PR #5623, opened 2022-06-24
Update PRIVACY.md

natalia-r2c:
Make explicit that IP addresses are collected as a metric for semgrep runs.


PRIVACY.md
```diff
@@ -231,6 +233,7 @@ This is a sample blob of the aggregate metrics described above:
     "sent_at": "2021-05-10T21:05:09+00:00",
     "event_id": "222bcccd-9dc2-4d10-ac3a-5692460e77ee",
     "anonymous_user_id": "5f52484c-3f82-4779-9353-b29bbd3193b6",
+    "clientIP": "0.0.0.0",
```
underyx:
```suggestion
```

It's not actually part of the payload that's sent I think. But we do see the IP address in the data pipeline. Maybe we should still keep it here though since some people might look at this blob to see all the data. But it's technically wrong, so I'm not sure 😅

natalia-r2c: Ok, I will remove from the blob to not introduce confusion. It is made explicit elsewhere.

underyx (approved):

Issue #5472, opened 2022-06-08
Dockerfile/Bash matching bug: pattern `RUN b` should not match `RUN a && b`


mjambon:
**Describe the bug**

Pattern `RUN b` should not match `RUN a && b`.

**To Reproduce**

Actual behavior, which is incorrect (or at least very surprising):
```
$ semgrep-core -lang bash -e '{ b; }' <(echo '{ a && b; }')
/tmp/semgrep-core-68b2b8-63:1
 { a && b; }
```
Expected behavior: no matches

Note that the following variant returns no matches as expected:
```
$ semgrep-core -lang bash -e '{ b; }' <(echo '{ a; b; }')
```

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Environment**

any

**Use case**

Originally reported by @p4p3r on behalf of a customer:

> I’m trying to update a dockerfile rule to address a customer’s negative sentiment and I’m not sure whether I’m hitting a parsing bug or I’m just matching the wrong way: https://semgrep.dev/s/Gppq
> Based on the highlighted tokens it looks to me like the `&&` within the RUN command is “breaking” the shell script into separate scripts? My goal is to match `cd` but only when it’s immediately after the `RUN` keyword.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1470/dockerfilebash-matching-bug-pattern-run-b-should-not-match-run-a. _Note_: this link is for r2c use only and is not accessible publicly.

mjambon:
I was able to get the same behavior for JavaScript:
```
$ semgrep-core -lang js -f <(echo 'a;') <(echo 'a && b;')
/tmp/semgrep-core-d757bc-62:1
 a && b;
```
^ shouldn't match!

Pattern AST:
```
$ semgrep-core -lang js -dump_pattern <(echo 'a;')
S(
  ExprStmt(
    N(
      Id(("a", ()),
        {id_info_id=1; id_hidden=false; id_resolved=Ref(None); 
         id_type=Ref(None); id_svalue=Ref(None); })), ()))
```

Target AST:
```
$ semgrep-core -lang js -dump_ast <(echo 'a && b;')
Pr(
  [ExprStmt(
     Call(IdSpecial((Op(And), ())),
       [Arg(
          N(
            Id(("a", ()),
[...]

PR #5630, opened 2022-06-25
Make statement patterns anchored

mjambon:
This is for https://github.com/returntocorp/semgrep/issues/5472

So far, I added test cases for:
* JavaScript
* Bash
* Dockerfile

The following change makes the tests above pass:
```
   | G.ExprStmt (a1, a2), B.ExprStmt (b1, b2) ->
-      m_expr_deep a1 b1 >>= fun () -> m_tok a2 b2
+      m_expr a1 b1 >>= fun () -> m_tok a2 b2
```
but a bunch of other tests fail. I haven't looked into why they're failing and I'm not sure how this should be resolved.

PR checklist:

- [ ] Changelog is up-to-date


mjambon:
I'm handing this off to @aryx because I'm going on vacation.

aryx:
"a bunch of other test fails", actually 122 tests fail, so it introduces too many regressions.
Including some of the tests you wrote before @mjambon like 
```
FROM debian
...
RUN apt-get ...
```
that you want to match
```
# ERROR:
FROM debian:testing
RUN echo hello
RUN apt-get update && apt-get install -y fortune
```

aryx:
I'll make an option so one can disable this behavior on a per-rule basis.

PR #5649, opened 2022-06-28
feat: log stderr on git failures

brendongo:
PR checklist:

- [ ] Changelog is up-to-date


cli/src/semgrep/meta.py
```diff
@@ -43,6 +43,11 @@ def git_check_output(command: Sequence[str]) -> str:
                     (fix with `git config --global --add safe.directory $(pwd)`)
 
                 Try running the command yourself to debug the issue.
+
+                Command failed with exit code: {e.returncode}
+                -----
+                Command failed with output:
+                {e.stderr}
```
underyx: Could we move the stderr near the top of the message? It will be difficult to find the reasons & recommended next step if the command dumps a lot of lines at the end. Please also make sure that `'{command_str}'` and `Try running the command yourself to debug the issue.` stay close to each other.

cli/src/semgrep/meta.py
```diff
@@ -323,20 +328,15 @@ def _find_branchoff_point(self, attempt_count: int = 0) -> str:
...
-                f"Base branch fetch: args={process.args}, stdout={process.stdout}, stderr={process.stderr}"
+                ]
             )
 
+            # Note that head must be fetched by commit not branch name since if the head
+            # is from a fork repo, the branch name doesn't exist in the base context
             logger.debug(
                 f"Trying to fetch branch {self._head_branch_ref} as commit from origin as head branch tip commit"
             )
-            process = subprocess.run(
```
underyx, line 339: should we write a semgrep rule for `subprocess.$METHOD(["git", ...], ...)`?

underyx: cuz I just pointed this change out in @spencerdrak's PR as well an hour ago

cli/src/semgrep/git.py
```diff
@@ -333,6 +289,7 @@ def baseline_context(self) -> Iterator[None]:
             # Note that we have no good way of detecting this issue without inspecting the checkout output
             # message, which means we are fragile with respect to git version here.
             logger.debug("Running git reset to return original context")
+            # nosem: use-git-check-output-helper
```
underyx, line 292: why?

underyx (approved):

cli/src/semgrep/git.py
brendongo: We use the returncode here for some logic :D

PR #5653, opened 2022-06-28
infra: ensure release scripts update versions

spencerdrak:
PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/scripts/release/bump
+++ b/scripts/release/bump
@@ -65,6 +65,9 @@ update_setup_py() {
   sed_in_place \
     's/^    install_requires=\["semgrep==.*"\],$/    install_requires=["semgrep=='"$release"'"],/g' \
     "$target"
+  sed_in_place \
+    's/^    version=".*",$/    version="'"$release"'",/g' \
+    "$target"
   git add "$target"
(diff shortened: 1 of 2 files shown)


scripts/release/bump
```diff
@@ -65,6 +65,9 @@ update_setup_py() {
   sed_in_place \
     's/^    install_requires=\["semgrep==.*"\],$/    install_requires=["semgrep=='"$release"'"],/g' \
     "$target"
+  sed_in_place \
+    's/^    version=".*",$/    version="'"$release"'",/g' \
+    "$target"
```
underyx, line 70: how come this used to work until 0.100.0? did this code change?

underyx (approved):

scripts/release/bump
spencerdrak: I just checked the log and originally, it was set to `0.0.0` - when we released `0.100.0`, it was updated to `0.100.0` and automation was never put around that.

spencerdrak: Now, it should stay up to date.

PR #5655, opened 2022-06-28
Plumb spans for partial parsing errors

nbrahms:
Spans were emitted from the core for partial parsing errors, but never
wired up so they appeared in CLI output. This change makes that work.

Some nuances:
- Previously, `semgrep --verbose --no-time` would still emit timing
  information; this commit removes the toggling of timing information
  with the verbose flag. Now, neither `semgrep --verbose` nor
  `semgrep --verbose --no-time` will emit timing information.

Fixes PA-1254.

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/cli/src/semgrep/core_output.py
+++ b/cli/src/semgrep/core_output.py
@@ -5,6 +5,7 @@ json output into a typed object
 The precise type of the response from semgrep-core is specified in
 https://github.com/returntocorp/semgrep/blob/develop/interfaces/Output_from_core.atd
 """
+import dataclasses
 from dataclasses import replace
 from typing import Dict
 from typing import List
@@ -27,6 +28,14 @@ from semgrep.verbose_logging import getLogger
 logger = getLogger(__name__)
 
 
+def _core_location_to_error_span(location: core.Location) -> out.ErrorSpan:
+    return out.ErrorSpan(
+        file=location.path,
+        start=out.PositionBis(line=location.start.line, col=location.start.col),
+        end=out.PositionBis(line=location.end.line, col=location.end.col),
+    )
+
+
 def core_error_to_semgrep_error(err: core.CoreError) -> SemgrepCoreError:
 
     # Hackily convert the level string to Semgrep expectations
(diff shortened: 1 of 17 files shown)


cli/src/semgrep/core_output.py
```diff
@@ -27,6 +28,14 @@
 logger = getLogger(__name__)
 
 
+def _core_location_to_error_span(location: core.Location) -> out.ErrorSpan:
```
brendongo, line 31: Should core location export this functionality?

brendongo: Or ErrorSpan expose some classmethod instantiator that takes a core.Location object?

cli/src/semgrep/output.py
```diff
@@ -311,7 +311,15 @@ def output(
                 for err in self.semgrep_structured_errors
                 if SemgrepError.semgrep_error_type(err) == "SemgrepCoreError"
             ]
-            paths = {Path(err.core.location.path) for err in semgrep_core_errors}
+            paths = {
```
brendongo: What is this counting?

cli/src/semgrep/core_output.py
nbrahms: core.* is generated by ATD, so I don't think there's a good way to edit it.

I'm sure we could get fancy with ErrorSpan, but since this code was only used within this file it made sense just to crowbar it into a private function.

cli/src/semgrep/output.py
nbrahms: Theoretically the number of lines that are skipped (but practically some number between 0 and "All", depending on which error comes last).

cli/tests/e2e/snapshots/test_cli_test/test_cli_test_verbose/results.txt
```diff
@@ -6,17 +6,3 @@ Findings:
         Basic test
 
           2┆ print([1m[x.xxx == 1[0m)
-
```
nbrahms, line 9: This is still covered by `test_cli_test.py::test_cli_test_time`

cli/src/semgrep/output.py
brendongo: Can you rename the variable to make that more clear? Can we also rename the instance variable this gets sent to

cli/src/semgrep/core_output.py
brendongo: sg we can move if ever gets used more

cli/src/semgrep/target_manager.py
```diff
@@ -106,7 +106,7 @@ class FileTargetingLog:
     cli_includes: Set[Path] = Factory(set)
     cli_excludes: Set[Path] = Factory(set)
     size_limit: Set[Path] = Factory(set)
-    failed_to_analyze: Set[Path] = Factory(set)
+    failed_to_analyze: Mapping[Path, Optional[int]] = Factory(dict)
```
brendongo: Can we add comment on what this is mapping / make the variable name more clear what it is mapping file to?

cli/src/semgrep/output.py
brendongo: And a comment explaining the edge cases  / why this is only theoretical and not actual number of lines that are skipped

nbrahms: I got rid of all those edge cases in 06baab9

brendongo (approved):

PR #5659, opened 2022-06-29
Feat: LSP Server Bonanza

ajbt200128:
PR checklist:

- [x] Changelog is up-to-date


cli/src/semgrep/lsp/convert.py
```diff
@@ -14,47 +25,107 @@ def findings_severity_to_diagnostic(severity: RuleSeverity) -> DiagnosticSeverit
...
+    return flatten(
+        [
+            list(map(rule_match_to_diagnostic, rule_match_list))
+            for rule_match_list in rule_map.values()
+        ]
+    )
+
+
+# get assocaitated files for a rule match
+def rule_to_files(rule: Rule, target_manager: TargetManager) -> List[str]:
```
underyx:
```suggestion
def rule_to_files(rule: Rule, target_manager: TargetManager) -> List[str]:
    """Get associated files for a rule match."""
```

^ there are a lot of docs automation tools that expect functions to be documented in this format. These docstrings are also made available when e.g. hovering over a call to `rule_to_files` in vscode, or as `rule_to_files.__doc__` in an interpreter

underyx: please update the other functions as well according to this if you have time!

ajbt200128:
Closes https://github.com/returntocorp/semgrep-vscode/issues/6

ajbt200128:
Closes https://github.com/returntocorp/semgrep-vscode/issues/14

cli/src/semgrep/lsp/convert.py
```diff
@@ -11,50 +25,147 @@
...
-    end_col = finding["end"]["col"] - 1
-    lines = content.split("\n")
-    if start_line == end_line:
-        match_source = lines[start_line][start_col:end_col]
-    else:
-        middle_lines = "\n".join(lines[start_line + 1 : end_line])
-        match_source = (
-            lines[start_line][start_col:] + middle_lines + lines[end_line][:end_col]
+def metavar_to_inlay(metavar: str, info: MetavarValue) -> JsonObject:
+    """Convert a Semgrep metavar to an inlay hint"""
```
brendongo, line 35: Can you add pointer to documentation on inlay hint format?

brendongo: Probably helpful to link to all the formats used here to help future people who touch this code

cli/src/semgrep/lsp/server.py
```diff
@@ -103,58 +168,206 @@ def m_initialize(
             },
         }
 
-    def m_initialized(self, **_kwargs: JsonObject) -> None:
-        log.info("Semgrep LSP initialized")
+    def m_initialized(self) -> None:
```
brendongo, line 172: Is there a reason all the methods are `m_` prefixed?

brendongo: Ah we're overriding some methods of the base class?

ajbt200128: Yes, it's how the lsp library pylsp_jsonrpc works https://github.com/python-lsp/python-lsp-jsonrpc/blob/develop/pylsp_jsonrpc/dispatchers.py#L19

cli/src/semgrep/lsp/metrics.py
```diff
@@ -0,0 +1,64 @@
...
+        self.rules_seen_last_map[rule_id] = opened
+        self.rule_seen_closed_map[rule_id] = closed + self.rule_seen_closed_map.get(
+            rule_id, 0
+        )
+
+    def send(
+        self, state: MetricsState, project_url: Optional[str], token: Optional[str]
+    ) -> None:
+        """Send the metrics to the backend."""
+        metrics = Metrics()
```
brendongo, line 54: It seems potentially fragile to use the Metrics class to send metrics when that class is pulling double duty as both the thing that keeps track of what it is sending to the metrics server and the thing that does the sending. Might be worthwhile to pullout the Sending logic to a MetricsSender class or something and using that. We want to avoid in case the Metrics class is modified with specific behavior someone else would expect only happens for metrics but they are unaware we also use it here in LSP metrics.

brendongo:
Also should update PR title

brendongo (approved):
My comments are mostly nits feel free to merge

cli/src/semgrep/lsp/convert.py
ajbt200128: Yea I'll add it everywhere

cli/src/semgrep/lsp/metrics.py
ajbt200128: I think that it's ok to keep the sending logic coupled to the metrics. The metrics class right now is mostly just a sender already. Idt any of the data is mutated once added to the class beyond some basic cleaning, so I don't see I guess what a dedicated sender class would add. https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/metrics.py#L152-L154

PR #5660, opened 2022-06-29
Fix location error with unicode parsing

### LINKED ISSUE #5641: Metavariable regex rule matching against YAML files fails to match after occurrence of unicode characters
**Describe the bug**
We've create a minimal test case that is supposed to match against all values that are set to "test":
```yaml
patterns:
  - pattern-either:
      - pattern: |
          $KEY: $VALUE
  - metavariable-regex:
      metavariable: $VALUE
      regex: test
```

It seems that after the first occurrence of a unicode character, the remaining matches are no longer flagged. One can move the unicode character from a value to a key, up or down in the target file, and can see that the boundary between where matches are correctly found and not found also shifts.

**To Reproduce**
A yaml file with unicode characters, but not using metavariable regex: https://semgrep.dev/playground/s/R4KO

A yaml file with unicode characters: https://semgrep.dev/playground/s/5e9A

A yaml file without unicode characters: https://semgrep.dev/playground/s/Gp4p

**Expected behavior**
All lines in the YAML file match, even after the unicode characters

**Use case**
We have a number of YAML files that contain unicode characters and this is diminishing our confidence in the accuracy of results using metavariable regex in those files.

emjin:
This is a hacky solution (see comment) because the YAML parser counts charpos
differently than Parse_info.

Closes #5641 

Test plan: make test

PR checklist:

- [x] Changelog is up-to-date


semgrep-core/src/parsing/Parse_pattern.ml
```diff
@@ -130,7 +130,7 @@ let parse_pattern lang ?(print_errors = false) str =
         let res = Parse_hack_tree_sitter.parse_pattern str in
         extract_pattern_from_tree_sitter_result res print_errors
     (* use adhoc parser (neither pfff nor tree-sitter) *)
-    | Lang.Yaml -> Yaml_to_generic.any str
+    | Lang.Yaml -> Yaml_to_generic.yaml_any str
```
aryx: this is not consistent with what we do for the other languages. I'd rather keep the any() and program() entry points.

semgrep-core/src/parsing/other/yaml_to_generic.mli
```diff
@@ -1,2 +1,4 @@
 val program : Common.filename -> AST_generic.program
 val any : string -> AST_generic.any
+val yaml_program : Common.filename -> AST_generic.program
```
aryx: I'd rather just the original signatures and just do the work internally.

aryx (approved):

semgrep-core/src/parsing/Parse_pattern.ml
emjin: The problem is that the yaml parser is also used for parsing rules, so if we replace unicode in `Yaml_to_generic.any` it will affect all languages, most of which don't treat unicode like this. I can process the file before passing it to Yaml_to_generic first instead, but then `program` would have to take a string. It's easier to give Yaml its own entry point, just like every other language. I'll rename the entry points (and add a comment) to make this more clear

aryx: oh ok. Not sure I understood but yes adding a comment would help!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 ### Fixed
 
 - `semgrep ci`: CI runs in GitHub Actions failed to checkout the commit assoociated with the head branch, and is fixed here.
+- Yaml: location information is fixed for unicode characters (#5660)
```
brendongo: This needs to be updated to a newer release :(

PR #5661, opened 2022-06-29
Scala: allow ellipsis in for loop headers

### LINKED ISSUE #5650: Scala for loops parse error
I expected that `for($EXPR2; $IDENT <- $GEN; $EXPR) {yield ... }`
Or `for(...; $IDENT <- $GEN; ...) {yield ... }`

would parse as patterns but they don't as of 0.100: https://semgrep.dev/s/NoNx

mmcqd:
Fixes #5650 

This PR adds support for Scala patterns like `for (...; $X <- $Y if $COND; ...) { ... }`. These are supported by adding a new `MultiForEach` for loop header to the generic AST. In the AST to IL translation, this construct is unrolled into nested for loops. Extending the generic AST to treat Scala's for loop headers as a list was more straightforward than attempting to do some funky matching involving big nested for loops.

This PR also deletes the `GIf` constructor for enumerators (for loop headers) from the Scala AST. As far as I can tell this was never actually produced by the parser, as guards are always parsed as directly part of a generator (`x <- y`). The Scala docs agree with me: https://scala-lang.org/files/archive/spec/2.13/06-expressions.html#for-comprehensions-and-for-loops. There is no mention of these guards that are not part of a generator.

test plan: make test

PR checklist:

- [x] Changelog is up-to-date


aryx:
great PR comment!

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -1052,16 +1052,25 @@ and for_header =
...
 
+and for_each =
+  pattern
+  * tok (* 'in' Python, 'range' Go, 'as' PHP, '' Java, '<-' Scala *)
+  * expr (* pattern 'in' expr *)
+
+and multi_for_each =
+  | FE of for_each
+  | FECond of for_each * tok * expr
+  | FEllipsis of tok
```
aryx, line 1072: I usually add a semgrep-ext: comment before those constructs that are Semgrep specific.

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -1052,16 +1052,25 @@ and for_header =
...
   | ForEllipsis of (* ... *) tok
 
+and for_each =
+  pattern
+  * tok (* 'in' Python, 'range' Go, 'as' PHP, '' Java, '<-' Scala *)
+  * expr (* pattern 'in' expr *)
+
+and multi_for_each =
+  | FE of for_each
+  | FECond of for_each * tok * expr
```
aryx, line 1071: maybe in comment you can put an example, like x <- xs if x > 2

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -2262,9 +2266,23 @@ and m_for_header a b =
...
+  m_pattern a1 b1 >>= fun () ->
+  m_tok at bt >>= fun () -> m_expr a2 b2
+
+and m_multi_for_each a b =
+  match (a, b) with
+  | G.FE a1, B.FE b1 -> m_for_each a1 b1
+  | G.FECond (a1, at, a2), B.FECond (b1, bt, b2) ->
+      m_for_each a1 b1 >>= fun () ->
+      m_tok at bt >>= fun () -> m_expr a2 b2
+  | G.FEllipsis _, _ -> return ()
```
aryx, line 2283: I don't think this can can happen given the definition of m_list_with_dots, but I guess it does not hurt.

semgrep-core/src/parsing/pfff/scala_to_generic.ml
```diff
@@ -562,24 +562,7 @@ and v_stmt = function
...
-                      G.If (g_tok, G.Cond g_e, stmt, None) |> G.s)
-                    guards stmt )
-              |> G.s
-          | `GIf guards ->
-              List.fold_right
-                (fun (g_tok, g_e) stmt ->
-                  G.If (g_tok, G.Cond g_e, stmt, None) |> G.s)
-                guards stmt)
-        v2 v3
+      G.For (v1, G.MultiForEach v2, v3) |> G.s
```
aryx, line 565: I usually try to avoid adding a construct in AST_generic when it's useful only for one language, but
maybe it's ok for this one indeed.

aryx (approved):
Great work!

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 ## Unreleased
 
+### Added
+
+- Scala: ellipsis are now allowed in for loop headers (#5650)
```
IagoAbal, line 9: During the "changelog meeting" this week, Bruno said he is now the one who is looking at the changelog to help producing the release notes, and that he thinks we don't generally write good changelog entries (that allow him to do that job easily). I think he will have problems understanding the purpose of this one.

IagoAbal: You can ask him directly, but my understanding is that he would like to know that this change is to allow Scala users to use ellipsis to write patterns matching a sequence of nested for loops. He actually emphasized a lot the _what_, _why_, and _how_ but I'm not sure how that would be instantiated in this particular case.

aryx: I think an example would be nice "(e.g., you can write a pattern like for(...; $X <- bar(); ...) { ... }".
Otherwise I feel the associated issue (#5650) mentioned in the changelog can help answer the what/why/how.

PR #5662, opened 2022-06-29
feat: Expose taint traces in semgrep-core text output

nmote:
This isn't plumbed through to the CLI yet, so it's not user-facing. But,
it's a step in that direction. The raw semgrep-core output is mostly
only used by Semgrep developers.

Instead of using the taint trace data separately provided to the match
hook (which isn't the codepath that JSON output uses), this change makes
it part of the `Pattern_match.t`. This involved defining a separate type
for taint traces that is more appropriate for match reporting, and
avoids a circular dependency. This will allow us to fairly easily add it
to the JSON output too, which is consumed by the CLI and the App.

Because this changes how the taint trace is accessed from the match
hook, a corresponding change to DeepSemgrep (which consumes taint trace
data there) will be required.

Test plan:

Run semgrep-core on the `tainting_rules/js/eslint_obj_inj.js` test:

```
eslint_obj_inj.js:4 with rule tainting
     return o[a]
  * Taint comes from:
    eslint_obj_inj.js:2
         var a = x
  * These intermediate values are tainted: a
  * This is how taint reaches the sink:
    eslint_obj_inj.js:4
         return o[a]
eslint_obj_inj.js:10 with rule tainting
     var z = o[b]
  * Taint comes from:
    eslint_obj_inj.js:8
         var b = baz(0)
  * These intermediate values are tainted: b
  * This is how taint reaches the sink:
    eslint_obj_inj.js:10
         var z = o[b]
eslint_obj_inj.js:21 with rule tainting
     return o[c]
  * Taint comes from:
    eslint_obj_inj.js:17
             c = x
  * These intermediate values are tainted: c
  * This is how taint reaches the sink:
    eslint_obj_inj.js:21
         return o[c]
```

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/core/Pattern_match.ml
+++ b/semgrep-core/src/core/Pattern_match.ml
@@ -36,6 +36,27 @@
  * from a pattern:, we should not merge them!
  *)
 
+type tainted_tokens = Parse_info.t list [@@deriving show]
+
+(* Simplified version of Taint.source_to_sink meant for finding reporting *)
+type taint_call_trace =
+  (* A direct match *)
+  | Toks of Parse_info.t list
+  (* An indirect match through a function call *)
+  | Call of {
+      call_toks : tainted_tokens;
+      intermediate_toks : tainted_tokens;
+      call_trace : taint_call_trace;
+    }
+[@@deriving show]
+
+type taint_trace = {
+  source : taint_call_trace;
+  tokens : tainted_tokens;
+  sink : taint_call_trace;
+}
+[@@deriving show]
+
 type t = {
   (* rule (or mini rule) responsible for the pattern match found *)
   rule_id : rule_id; [@equal fun a b -> a.id = b.id]
@@ -48,6 +69,8 @@ type t = {
   tokens : Parse_info.t list Lazy.t; [@equal fun _a _b -> true]
   (* metavars for the pattern match *)
   env : Metavariable.bindings;
+  (* Lazy since construction involves forcing lazy token lists. *)
+  taint_trace : taint_trace Lazy.t option; [@equal fun _a _b -> true]
 }
 
 (* This is currently a record, but really only the rule id should matter.
--- a/semgrep-core/src/engine/Match_rules.mli
+++ b/semgrep-core/src/engine/Match_rules.mli
@@ -8,12 +8,7 @@ exception File_timeout
    This can raise File_timeout.
 *)
 val check :
-  match_hook:
-    (string ->
-    Metavariable.bindings ->
-    Parse_info.t list Lazy.t ->
-    Taint.finding option ->
-    unit) ->
+  match_hook:(string -> Pattern_match.t -> unit) ->
   timeout:float ->
   timeout_threshold:int ->
   Config_semgrep.t * Equivalence.equivalences ->
--- a/semgrep-core/src/engine/Match_tainting_mode.ml
+++ b/semgrep-core/src/engine/Match_tainting_mode.ml
@@ -22,6 +22,7 @@ module PM = Pattern_match
 module RM = Range_with_metavars
 module RP = Report
 module T = Taint
+module PI = Parse_info
 
 let logger = Logging.get_logger [ __MODULE__ ]
 
@@ -329,9 +330,28 @@ let taint_config_of_rule default_config equivs file ast_and_errors
       sinks = sinks_ranges;
     } )
 
+let rec convert_taint_call_trace = function
+  | Taint.PM pm ->
+      let toks = Lazy.force pm.PM.tokens |> List.filter PI.is_origintok in
+      PM.Toks toks
+  | Taint.Call (expr, toks, ct) ->
+      PM.Call
+        {
+          call_toks = V.ii_of_any (G.E expr) |> List.filter PI.is_origintok;
+          intermediate_toks = toks;
+          call_trace = convert_taint_call_trace ct;
+        }
+
+let taint_trace_of_src_to_sink source tokens sink =
+  {
+    Pattern_match.source = convert_taint_call_trace source;
+    tokens;
+    sink = convert_taint_call_trace sink;
+  }
+
 let pm_of_finding finding =
   match finding with
-  | T.SrcToSink { source = _; tokens = _; sink; merged_env } ->
+  | T.SrcToSink { source; tokens; sink; merged_env } ->
       (* We always report the finding on the sink that gets tainted, the call trace
        * must be used to explain how exactly the taint gets there. At some point
        * we experimented with reporting the match on the `sink`'s function call that
@@ -347,8 +367,11 @@ let pm_of_finding finding =
        * for the injection bug... but most users seem to be confused about this. They
        * already expect Semgrep (and DeepSemgrep) to report the match on `sink(x)`.
        *)
+      let taint_trace =
+        Some (lazy (taint_trace_of_src_to_sink source tokens sink))
+      in
       let sink_pm = T.pm_of_trace sink in
-      Some { sink_pm with env = merged_env }
+      Some { sink_pm with env = merged_env; taint_trace }
   | T.SrcToReturn _
   (* TODO: We might want to report functions that let input taint
    * go into a sink (?) *)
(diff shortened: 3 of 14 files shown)


semgrep-core/src/core/Pattern_match.ml
```diff
@@ -36,6 +36,27 @@
  * from a pattern:, we should not merge them!
  *)
 
+type tainted_tokens = Parse_info.t list [@@deriving show]
+
+(* Simplified version of Taint.source_to_sink meant for finding reporting *)
+type taint_call_trace =
+  (* A direct match *)
+  | Toks of Parse_info.t list
```
aryx, line 44: why not tainted_tokens here?

semgrep-core/src/engine/Match_rules.mli
```diff
@@ -8,12 +8,7 @@ exception File_timeout
    This can raise File_timeout.
 *)
 val check :
-  match_hook:
-    (string ->
-    Metavariable.bindings ->
-    Parse_info.t list Lazy.t ->
-    Taint.finding option ->
-    unit) ->
+  match_hook:(string -> Pattern_match.t -> unit) ->
```
aryx, line 11: nice, cleaner!

semgrep-core/src/core/Pattern_match.ml
```diff
@@ -36,6 +36,27 @@
...
+  (* An indirect match through a function call *)
+  | Call of {
+      call_toks : tainted_tokens;
+      intermediate_toks : tainted_tokens;
```
aryx, line 48: what are those intermediate_toks?

aryx: maybe an example would help?
like with code like  foo(source);   and separately  function foo(x) { sink(source) } 
what would be the taint_call_trace?

aryx: and what would be those intermediate_toks?

semgrep-core/src/engine/Match_tainting_mode.ml
```diff
@@ -420,22 +443,8 @@ let check_fundef lang fun_env taint_config opt_ent fdef =
   check_stmt ?name ~in_env lang fun_env taint_config
     (H.funcbody_to_stmt fdef.G.fbody)
 
-(* TODO: Pass a hashtable to cache the CFG of each def, otherwise we are
```
aryx, line 423: why we don't need this code anymore?

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -511,11 +541,12 @@ let semgrep_with_rules config ((rules, invalid_rules), rules_parse_time) =
...
            let xtarget = xtarget_of_file config xlang file in
-           let match_hook str env matched_tokens _opt_taint_finding =
+           let match_hook str match_ =
              if config.output_format = Text then
-               let xs = Lazy.force matched_tokens in
-               print_match ~str config.match_format config.mvars env
-                 Metavariable.ii_of_mval xs
+               let xs = Lazy.force match_.Pattern_match.tokens in
+               print_match ~str config.match_format config.mvars
+                 match_.Pattern_match.env Metavariable.ii_of_mval xs
```
aryx, line 548: at this point maybe we should just pass the whole match_ to print_match and let print_match internally
access its different elements.
We could also pass the whole config instead of passing those 2 config arguments..

semgrep-core/src/runner/Run_semgrep.mli
```diff
@@ -49,6 +49,7 @@ val print_match :
   Metavariable.bindings ->
   (Metavariable.mvalue -> Parse_info.t list) ->
   Parse_info.t list ->
+  Pattern_match.taint_trace Lazy.t option ->
```
aryx, line 52: maybe pass the whole Pattern_match.t?

aryx (approved):
great!
The code looks simpler too (especially the type interfaces).

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -74,8 +74,37 @@ let replace_named_pipe_by_regular_file path =
...
+  if format <> Matching_report.Normal then ()
+  else
+    let (lazy { Pattern_match.source; tokens; sink }) = taint_trace in
+    pr "  * Taint comes from:";
+    print_taint_call_trace ~format ~spaces:4 source;
+    if tokens <> [] then
+      pr
+        (spf "  * These intermediate values are tainted: %s"
+           (string_of_toks tokens));
+    pr "  * This is how taint reaches the sink:";
```
IagoAbal, line 103: This is needed in DeepSemgrep because we are inter procedural, so the taint may go through several function calls before reaching the sink. But in Semgrep this will never happen so I think it's just redundant.

semgrep-core/src/engine/Match_tainting_mode.ml
IagoAbal: When adding the taint trace to DeepSemgrep, this was a hack that I used to keep track of the taint `finding` associated with the reported pattern match. But now the pattern match already carries the taint trace.

semgrep-core/src/core/Pattern_match.ml
nmote: The `tainted_tokens` type is defined in `Taint.ml`. Referencing it here would lead to a circular dependency. I could have moved it or duplicated it here but I don't think that it makes it any clearer.

semgrep-core/src/runner/Run_semgrep.ml
nmote: Good to know. Maybe we should expose this printing logic so that DeepSemgrep can use it. Anyway, I'm reluctant to not handle a case just because we are pretty sure it won't happen.

semgrep-core/src/core/Pattern_match.ml
nmote: @IagoAbal would probably be better at answering this than me. They are modeled after this: https://github.com/returntocorp/semgrep/blob/e3e87a473a8603acad6c7b5fb40edcd47fb6c610/semgrep-core/src/tainting/Taint.ml#L24

According to a separate comment from @IagoAbal on this PR, this probably can't actually happen in Semgrep in practice, and only occurs in DeepSemgrep.

semgrep-core/src/core/Pattern_match.ml
IagoAbal: I believe Yoann asks why you are not using your type alias `tainted_tokens` instead of `Parse_info.t list`. And this `Parse_info.t list` is not the same as the "tainted tokens". This list of tokens represents a `Pattern_match.t`, whereas `tainted_tokens` are intermediate tokens (right now just variables) that are tainted.

semgrep-core/src/core/Pattern_match.ml
IagoAbal: Right now it could be called `intermediate_vars` 🤔 those tokens are from intermediate variables that get tainted.

semgrep-core/src/core/Pattern_match.ml
nmote: Clearly, I have already forgotten what code I wrote here :laughing:

After a second look here, I don't think that there was any particular reason that I didn't use `tainted_tokens` here. I think @IagoAbal's justification for this decision makes sense, though.

semgrep-core/src/runner/Run_semgrep.ml
IagoAbal: You can still handle it for the "just in case", but instead maybe you can do it conditionally. If `sink` is not a `Call`, then printing it just makes the output more verbose.

semgrep-core/src/core/Pattern_match.ml
IagoAbal: Maybe you can add a type alias `pattern_match_tokens` and use it in the places where you are replacing a `Pattern_match.t` ?

semgrep-core/src/runner/Run_semgrep.ml
nmote: Since end users don't use semgrep-core directly, this is basically only visible to us so I don't think it's worth spending much time trying to polish this output.

IagoAbal: I use it a lot so I do care, but I can fix it myself too, so I will open a PR!

nmote: Sounds good! I'm making a few changes this morning based on feedback here. If you could hold off for a few hours it might help us avoid merge conflicts.

IagoAbal: If you are working on these comments I guess it's also not a big deal for you to apply this diff?
```diff
@@ -100,8 +100,11 @@ let print_taint_trace ~format taint_trace =
       pr
         (spf "  * These intermediate values are tainted: %s"
            (string_of_toks tokens));
-    pr "  * This is how taint reaches the sink:";
-    print_taint_call_trace ~format ~spaces:4 sink
+    match sink with
+    | Pattern_match.Toks _ -> ()
+    | _call ->
+        pr "  * This is how taint reaches the sink:";
+        print_taint_call_trace ~format ~spaces:4 sink
 
 let print_match ?str match_format mvars mvar_binding ii_of_any
     tokens_matched_code taint_trace =
```
Otherwise I don't think it will cause major conflicts.

PR #5664, opened 2022-06-30
fix: merge-base incorrect when shallow fetch depth is insufficient

brendongo:
As an optimization, semgrep uses shallow fetches. Because of this, it is possible with a shallow enough
fetch depth that the merge-base we calculate is not the best possible / most recent merge-base.

In particular the test included in this PR demonstrates a case where we do not calculate the correct one:
if the base branch gets a large number of commits on top of it but has a merge request that short circuits loading
a very old commit that head can also reach.

We expose an envvar SEMGREP_MIN_FETCH_DEPTH that sets a minimum fetch depth
when fetching commits to calculate fetch depth. For now default value is set to 0 and we suspect
setting it to 200 should mean we calculate the correct merge-base 99% of the time. Not setting that
as default for now to investigate impact of large fetch depth on scan time.

For more info on shallow fetch see https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/ 

For more info on repro see #CLI-223

Live tested on https://github.com/returntocorp/test-merge/runs/7138341402?check_suite_focus=true

PR checklist:

- [x] Changelog is up-to-date


cli/src/semgrep/env.py
```diff
@@ -135,3 +136,8 @@ def in_agent_default(self) -> bool:
     @shouldafound_no_email.default
     def shouldafound_no_email_default(self) -> bool:
         return "SEMGREP_SHOULDAFOUND_NO_EMAIL" in os.environ
+
+    @fetch_depth_min.default
+    def fetch_depth_min_default(self) -> int:
+        value = os.getenv("SEMGREP_MIN_FETCH_DEPTH", "200")
```
brendongo:
```suggestion
        value = os.getenv("SEMGREP_MIN_FETCH_DEPTH", "0")
```

cli/tests/e2e/test_ci.py
```diff
@@ -560,6 +560,141 @@ def test_github_ci_bad_base_sha(
...
+    """
+    commits = defaultdict(list)
+    foo = git_tmp_path / "foo.py"
+    bar = git_tmp_path / "bar.py"
+    baz = git_tmp_path / "baz.py"
+
+    subprocess.run(["git", "checkout", "-b", "foo"])
+    foo.open("a").write(f"foo == 5\n")
+    commits["foo"].append(_git_commit(1, add=True))
+    subprocess.run(["git", "show", "-s", "--format=%ct", "b903231925961ac9d787ae53ee0bd15ec156e689"])
```
brendongo:
```suggestion
```

brendongo:
With included test and live test (https://github.com/returntocorp/test-merge/runs/7138341507?check_suite_focus=true shows no change and https://github.com/returntocorp/test-merge/runs/7138341402?check_suite_focus=true shows correct scan when envvar is set) I feel confident this will allow some users to solve their issue without impacting other users while we check on impact of fetch-depth

underyx:
I just realized this can already be set as an option on the checkout step btw. Should we still add it as a Semgrep feature? Or just recommend

```yaml
- run: actions/checkout
  with:
    fetch-depth: 100
```

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - Scala: ellipsis are now allowed in for loop headers, so you can write patterns
   like `for (...; $X <- $Y if $COND; ...) { ... }` to match nested for loops. (#5650)
+- `SEMGREP_MIN_FETCH_DEPTH` with default value 0 to set the minimum fetch depth
+  when calculating merge-base (in GithubActions pull requests). Setting to a high
+  value like 100 should solve occurances of semgrep calculating a stale merge-base
+  causing semgrep to report findings not touched in PRs (#5664)
```
underyx:
```suggestion
- The `SEMGREP_MIN_FETCH_DEPTH` environment variable
  which lets you set how many commits `semgrep ci` fetches from the remote at the minimum
  when the Git repository was cloned shallow.
  Having more commits available helps Semgrep determine what changes came from the current pull request,
  fixing issues where Semgrep would report findings that weren't touched in a given pull request.
  This value is set to 0 by default (#5664)
```

underyx (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - Scala: ellipsis are now allowed in for loop headers, so you can write patterns
   like `for (...; $X <- $Y if $COND; ...) { ... }` to match nested for loops. (#5650)
+- The `SEMGREP_MIN_FETCH_DEPTH` environment variable
+  which lets you set how many commits `semgrep ci` fetches from the remote at the minimum
+  when the Git repository was cloned shallow.
+  Having more commits available helps Semgrep determine what changes came from the current pull request,
+  fixing issues where Semgrep would report findings that weren't touched in a given pull request.
+  This value is set to 0 by default (#5664)
```
brendongo:
```suggestion
- The `SEMGREP_MIN_FETCH_DEPTH` environment variable
  which lets you set how many commits `semgrep ci` fetches from the remote at the minimum
  when the Git repository was cloned shallow.
  Having more commits available helps Semgrep determine what changes came from the current pull request,
  fixing issues where Semgrep would report findings that weren't touched in a given pull request.
  This value is set to 0 by default. Only applies to GithubActions (#5664)
```

brendongo:
@underyx I think that might set the default fetch and if so when we set `--depth` it would not respect the default

edit: but i guess doesnt matter if it was already fetched

Also reading the doc for that it sounds like it pulls that much depth for all branches and tags?

brendongo:
I think we eventually want to make the default higher so I think having this controlled by semgrep will avoid us relying on the user to set the right arg on checkout step.

CHANGELOG.md
underyx: maybe we should put GHA in the env var name, didn't realize this was the case.

brendongo: yeah good idea

brendongo: Done

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ This project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html
 
 - Scala: ellipsis are now allowed in for loop headers, so you can write patterns
   like `for (...; $X <- $Y if $COND; ...) { ... }` to match nested for loops. (#5650)
+- The `SEMGREP_GHA_MIN_FETCH_DEPTH` environment variable which lets you set how many
+  commits `semgrep ci` fetches from the remote at the minimum when calculating the merge-base in GithubActions.
```
underyx:
```suggestion
  commits `semgrep ci` fetches from the remote at the minimum when calculating the merge-base in GitHub Actions.
```

underyx (approved):

PR #5677, opened 2022-07-04
refactor: Simplify print_match

nmote:
Based on review feedback in
https://github.com/returntocorp/semgrep/pull/5662

This makes `Run_semgrep.print_match` take a `Runner_config.t` and
`Pattern_match.t` as arguments instead of taking several of those
structures' assorted fields as individual arguments.

This also involved changing `Match_patterns.check` to take a hook that
takes a `Pattern_match.t` instead of a couple of individual fields.

Test plan: Automated tests

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/engine/Unit_engine.ml
+++ b/semgrep-core/src/engine/Unit_engine.ml
@@ -284,10 +284,9 @@ let regression_tests_for_lang ~with_caching files lang =
              Common.save_excursion Flag_semgrep.with_opt_cache with_caching
                (fun () ->
                  Match_patterns.check
-                   ~hook:(fun _env matched_tokens ->
+                   ~hook:(fun { Pattern_match.tokens = (lazy xs); _ } ->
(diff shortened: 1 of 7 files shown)


IagoAbal (approved):

semgrep-core/src/engine/Unit_engine.ml
```diff
@@ -284,10 +284,9 @@ let regression_tests_for_lang ~with_caching files lang =
              Common.save_excursion Flag_semgrep.with_opt_cache with_caching
                (fun () ->
                  Match_patterns.check
-                   ~hook:(fun _env matched_tokens ->
+                   ~hook:(fun { Pattern_match.tokens = (lazy xs); _ } ->
```
aryx, line 287: what? I didn't know you could use 'lazy' in a pattern :) nice.

nmote: I think this is actually considered more idiomatic than `Lazy.force` and I use it where I can!

PR #5678, opened 2022-07-04
core: cli: Simplify taint trace output

IagoAbal:
Unless the sink is a call then there is no need to print it.

Follows: 6fcc672dc24 ("feat: Expose taint traces in semgrep-core text output (#5662)")

test plan:

    % cat tests/tainting_rules/python/tainting.py
    def foo():
    a = source1()
    if True:
        b = a
        b = sanitize1()
    else:
        b = a
    #ERROR:
    sink1(b)
    % bin/semgrep-core -lang py -rules tests/tainting_rules/python/tainting.yaml tests/tainting_rules/python/tainting.py
    tests/tainting_rules/python/tainting.py:9 with rule tainting
    sink1(b)
    * Taint comes from:
        tests/tainting_rules/python/tainting.py:2
        a = source1()
    * These intermediate variables are tainted: a, b

PR checklist:

- [x] Changelog is up-to-date

### DIFF
--- a/semgrep-core/src/runner/Run_semgrep.ml
+++ b/semgrep-core/src/runner/Run_semgrep.ml
@@ -85,7 +85,7 @@ let rec print_taint_call_trace ~format ~spaces = function
       Matching_report.print_match ~format ~spaces call_toks;
       if intermediate_toks <> [] then
         pr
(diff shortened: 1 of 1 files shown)


IagoAbal:
Ooops, closing in favor of #5679.

PR #5683, opened 2022-07-05
Fix parsing issue to cut CI scan speeds in half

ajbt200128:
PR checklist:

- [X] Changelog is up-to-date

### DIFF
--- a/cli/src/semgrep/config_resolver.py
+++ b/cli/src/semgrep/config_resolver.py
@@ -447,6 +447,7 @@ def parse_config_string(
     try:
         # we pretend it came from YAML so we can keep later code simple
         data = YamlTree.wrap(json.loads(contents), EmptySpan)
+        return {config_id: data}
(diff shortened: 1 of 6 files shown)


cli/src/semgrep/config_resolver.py
```diff
@@ -447,6 +447,7 @@ def parse_config_string(
     try:
         # we pretend it came from YAML so we can keep later code simple
         data = YamlTree.wrap(json.loads(contents), EmptySpan)
+        return {config_id: data}
```
brendongo, line 450: Do we need need to do line 455 below before returning?

ajbt200128: No, the lines below handle the case of if contents is a yaml file , which is why it took so long as L449 was unused. Basically we did all the work for the json stuff but then never returned it and parsed it to yaml and back.

underyx (approved):

PR #5687, opened 2022-07-06
Pass down the parent metavariable contex to metavariable-pattern

### LINKED ISSUE #5060: Allow metavariable-pattern to inherit the variable environment from the parent
Currently when you write a taint rule, you are limited from being able to really dig deep into the 'source' through metavariable-patterns e.g. I want to know if the source is the start of a query, after a variable, or is after a literal string, etc.

If we allowed metavariable-pattern to inherit the variable environment from the parent formula, then we could easily include the source into our validation cases.

 if you could write <... $SOURCE ...> inside a metavariable-pattern and know exactly you are working with the source, and not any $VARIABLE

Example of failing due to current metavariable-pattern: https://semgrep.dev/s/LewisArdern:express-open-redirect 

We should allow in taint-mode for the metavariable-pattern to inherit the variable environment from the parent formula, then we could easily include the source into our validation cases.

If you look at https://semgrep.dev/s/LewisArdern:express-open-redirect  if we had the ability to access the source through something like <... $SOURCE ...> we could potentially ensure that we do not introduce false positives, currently the pattern-nots are not correct, and will introduce false negatives and false positives. 

Hacky hacky https://semgrep.dev/s/rNKW (not useful for most people)

**Use case**
It will enable rule-writers to have a better chance of removing false positives from their rules because you can really ensure the value comes from the source we are capturing

**Additional context**

The reason why  "..."+$REQ matches 

res.redirect("https://google.com/"+req.query.url)

Is because $REQ is in metavariable-pattern so it's not the same mvar used in pattern-sources and it's binding to req.query.url.
But if you try to match "..."+$REQ.query this does not work because $REQ.query cannot match req.query.url , it's the same as trying to match $REQ against req.query and .query against .url, .query and .url cannot be matched.

emjin:
Closes #5060

When a metavariable-pattern is analyzed, retain the metavariable context from its parent

PR checklist:


mmcqd (approved):
Should we make some kind of announcement if we're making a breaking change? I definitely think the change is correct and any rule relying on things not working like this is probably bad style, but it would probably be better not to unexpectedly break anyone's rules.

PR #5693, opened 2022-07-06
Create automatic version compare script

nbrahms:
Previously we had `?version=...` in the Semgrep.dev playground. That's
been removed due to its maintenance overhead. This script should replace
that for internal use with Semgrep and rule development.

Closes CLI-252.

PR checklist:


IagoAbal (approved):
Thanks! I'm sure we can take it from here.

PR #5697, opened 2022-07-07
Field Sensitive Tainting

mmcqd:
Closes PA-1278

This PR makes taint analysis sensitive to individual fields in records/objects in a limited way. Essentially, it adds _per variable_ field sensitivity and not _per object in memory_ field sensitivity. There is no alias analysis involved. A summary of new behavior

 1. Taintedness/cleaness is tracked individually for each field of each variable
 2. If a field `a.b.c` is marked tainted, it does not affect whether `a` and `a.b` are marked tainted. This should reduce FPs at the cost of more FNs. This could be considered a *breaking* change, but hopefully for the best. 
 3. If a field `a.b.c` is marked clean, then `a.b.c` and any previously tainted field of which `a.b.c` is a _prefix_ is marked clean. So `a.b.c.d` will now be clean, even if had been previously tainted.
 4. When checking the taintedness of some dot expression `a.b.c.d`, we use the most up to date knowledge of its fields. That is, if we know only that `a.b` is tainted, then `a.b.c.d` is tainted. But if we know that `a.b.c` was marked clean after `a.b` was tainted, then `a.b.c.d` is clean.

The following technical changes are needed to enable this behavior:
1. IL lvals now have their offsets represented as _backwards lists_. So `a.b.c` is `{base = a, rev_offset = [.c ; .b]}`. This allows even nested field access lvals to be represented as single variables.
2. We differentiate between a variable with _unknown taint_ and one with _known lack of taint_. To see why this is needed, consider:
    ```
    a = source
    a.b = safe
    sink(a.b)
    ```
    If `a.b = safe` simply removed the string corresponding to `a.b` from the set of tainted variables, it would do nothing. We need to specifically remember the fact that `a.b` is clean!

PR checklist:

### DIFF
--- a/semgrep-core/src/core/il/IL_lvalue_helpers.ml
+++ b/semgrep-core/src/core/il/IL_lvalue_helpers.ml
@@ -13,30 +13,15 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the file
  * license.txt for more details.
  *)
-open Common
 open IL
 
 (*****************************************************************************)
 (* Prelude *)
 (*****************************************************************************)
-(* Lvalue/Rvalue helpers working on the IL
- *)
 
-(*****************************************************************************)
-(* Helpers *)
-(*****************************************************************************)
-
-let lval_of_instr_opt x =
-  match x.i with
-  | Assign (lval, _)
-  | AssignAnon (lval, _)
-  | Call (Some lval, _, _)
-  | CallSpecial (Some lval, _, _) ->
-      Some lval
-  | Call _
-  | CallSpecial _ ->
-      None
-  | FixmeInstr _ -> None
+let compare_name x y =
+  let ident_cmp = String.compare (fst x.ident) (fst y.ident) in
+  if ident_cmp <> 0 then ident_cmp else Int.compare x.sid y.sid
 
 let rexps_of_instr x =
   match x.i with
@@ -76,7 +61,7 @@ and lvals_in_lval lval =
       (function
         | Index e -> lvals_of_exp e
         | _ -> [])
-      lval.offset
+      lval.rev_offset
   in
   base_lvals @ offset_lvals
 
@@ -88,43 +73,54 @@ let rlvals_of_instr x =
   lvals_of_exps exps
 
 (*****************************************************************************)
-(* API *)
+(* Public *)
 (*****************************************************************************)
 
-let dotted_lvars_of_lval = function
-  | { base = Var name; offset } -> (
-      let id, tok = name.ident in
-      let str_of_name name = Common.spf "%s-%d" (fst name.ident) name.sid in
-      let dot_strs =
-        List.fold_right
-          (fun o s ->
-            match (s, o) with
-            | Some (s :: ss), Dot name ->
-                let x = s ^ "." ^ str_of_name name in
-                Some (x :: s :: ss)
-            | Some [], Dot name -> Some [ "." ^ str_of_name name ]
-            (* We only care about tracking taint through fields
-             * So if we hit a non-Dot offset, we just give up
-             *)
-            | Some _, Index _
-            | None, _ ->
-                None)
-          offset (Some [])
-      in
-      match dot_strs with
-      | None -> []
-      | Some dot_strs ->
-          let add_dots dots = { name with ident = (id ^ ";" ^ dots, tok) } in
-          Common.map add_dots dot_strs)
-  | _ -> []
+let lval_is_var_and_dots { base; rev_offset } =
+  match base with
+  | Var _ ->
+      rev_offset
+      |> List.for_all (function
+           | Dot _ -> true
+           | Index _ -> false)
+  | VarSpecial _
+  | Mem _ ->
+      false
+
+let lval_is_dotted_prefix lval1 lval2 =
+  let eq_name x y = compare_name x y = 0 in
+  let rec offset_prefix os1 os2 =
+    match (os1, os2) with
+    | [], _ -> true
+    | _ :: _, [] -> false
+    | Index _ :: _, _
+    | _, Index _ :: _ ->
+        false
+    | Dot a :: os1, Dot b :: os2 -> eq_name a b && offset_prefix os1 os2
+  in
+  match (lval1, lval2) with
+  | { base = Var x; rev_offset = ro1 }, { base = Var y; rev_offset = ro2 } ->
+      eq_name x y && offset_prefix (List.rev ro1) (List.rev ro2)
+  | _ -> false
+
+let lval_of_instr_opt x =
+  match x.i with
+  | Assign (lval, _)
+  | AssignAnon (lval, _)
+  | Call (Some lval, _, _)
+  | CallSpecial (Some lval, _, _) ->
+      Some lval
+  | Call _
+  | CallSpecial _ ->
+      None
+  | FixmeInstr _ -> None
 
 let lvar_of_instr_opt x =
-  let* lval = lval_of_instr_opt x in
-  match lval with
-  | { base = Var base_name; _ } ->
-      let dots = dotted_lvars_of_lval lval in
-      Some (base_name, dots)
-  | _ -> None
+  match lval_of_instr_opt x with
+  | Some { base = Var x; _ } -> Some x
+  | Some _
+  | None ->
+      None
 
 let rlvals_of_node = function
   | Enter
@@ -142,3 +138,26 @@ let rlvals_of_node = function
   | NOther _
   | NTodo _ ->
       []
+
+module LvalOrdered = struct
+  type t = lval
+
+  let compare lval1 lval2 =
+    (* Right now we only care about comparing lvals of the form `x.a_1. ... . a_n,
+       so it's OK if `Stdlib.compare` may not be the ideal comparison function for
+       the remaining cases. *)
+    match (lval1, lval2) with
+    | { base = Var x; rev_offset = ro1 }, { base = Var y; rev_offset = ro2 } ->
+        let name_cmp = compare_name x y in
+        if name_cmp <> 0 then name_cmp
+        else
+          List.compare
+            (fun offset1 offset2 ->
+              match (offset1, offset2) with
+              | Dot a, Dot b -> compare_name a b
+              | Index _, _
+              | _, Index _ ->
+                  Stdlib.compare offset1 offset2)
+            ro1 ro2
+    | _, _ -> Stdlib.compare lval1 lval2
+end
--- a/semgrep-core/src/core/il/IL_lvalue_helpers.ml
+++ b/semgrep-core/src/core/il/IL_lvalue_helpers.ml
@@ -72,9 +72,11 @@ and lvals_in_lval lval =
     | _else_ -> []
   in
   let offset_lvals =
-    match lval.offset with
-    | Index e -> lvals_of_exp e
-    | __else_ -> []
+    List.concat_map
+      (function
+        | Index e -> lvals_of_exp e
+        | _ -> [])
+      lval.offset
   in
   base_lvals @ offset_lvals
 
@@ -89,13 +91,40 @@ let rlvals_of_instr x =
 (* API *)
 (*****************************************************************************)
 
+let dotted_lvars_of_lval = function
+  | { base = Var name; offset } -> (
+      let id, tok = name.ident in
+      let str_of_name name = Common.spf "%s-%d" (fst name.ident) name.sid in
+      let dot_strs =
+        List.fold_right
+          (fun o s ->
+            match (s, o) with
+            | Some (s :: ss), Dot name ->
+                let x = s ^ "." ^ str_of_name name in
+                Some (x :: s :: ss)
+            | Some [], Dot name -> Some [ "." ^ str_of_name name ]
+            (* We only care about tracking taint through fields
+             * So if we hit a non-Dot offset, we just give up
+             *)
+            | Some _, Index _
+            | None, _ ->
+                None)
+          offset (Some [])
+      in
+      match dot_strs with
+      | None -> []
+      | Some dot_strs ->
+          let add_dots dots = { name with ident = (id ^ ";" ^ dots, tok) } in
+          Common.map add_dots dot_strs)
+  | _ -> []
+
 let lvar_of_instr_opt x =
   let* lval = lval_of_instr_opt x in
-  match lval.base with
-  | Var n -> Some n
-  | VarSpecial _
-  | Mem _ ->
-      None
+  match lval with
+  | { base = Var base_name; _ } ->
+      let dots = dotted_lvars_of_lval lval in
+      Some (base_name, dots)
+  | _ -> None
 
 let rlvals_of_node = function
   | Enter
--- a/semgrep-core/src/core/il/IL_lvalue_helpers.mli
+++ b/semgrep-core/src/core/il/IL_lvalue_helpers.mli
@@ -1,2 +1,3 @@
-val lvar_of_instr_opt : IL.instr -> IL.name option
+val dotted_lvars_of_lval : IL.lval -> IL.name list
+val lvar_of_instr_opt : IL.instr -> (IL.name * IL.name list) option
 val rlvals_of_node : IL.node_kind -> IL.lval list
--- a/semgrep-core/src/core/il/IL_lvalue_helpers.mli
+++ b/semgrep-core/src/core/il/IL_lvalue_helpers.mli
@@ -1,3 +1,27 @@
-val dotted_lvars_of_lval : IL.lval -> IL.name list
-val lvar_of_instr_opt : IL.instr -> (IL.name * IL.name list) option
+(** Lvalue/Rvalue helpers working on the IL *)
+
+val lval_is_var_and_dots : IL.lval -> bool
+(** Test whether an lvalue is of the form x.a_1. ... .a_n *)
+
+val lval_is_dotted_prefix : IL.lval -> IL.lval -> bool
+(** [lval_is_dotted_prefix lval1 lval2] tests whether [lval1] is of the
+    form x.a_1. ... .a_n, and whether [lval2] is an extension of [lval1],
+    that is, x.a_1. ... .a_n.o_1 ... . o_M. *)
+
+val lval_of_instr_opt : IL.instr -> IL.lval option
+(** If the given instruction stores its result in [lval] then it is
+    [Some lval], otherwise it is [None]. *)
+
+val lvar_of_instr_opt : IL.instr -> IL.name option
+(** If the given instruct stores its result in an lvalue of the form
+    x.o_1. ... .o_N, then it is [Some x], otherwise it is [None]. *)
+
 val rlvals_of_node : IL.node_kind -> IL.lval list
+(** The lvalues that occur in the RHS of a node. *)
+
+(** Useful to instantiate data strutures like Map and Set. *)
+module LvalOrdered : sig
+  type t = IL.lval
+
+  val compare : t -> t -> int
+end
--- a/semgrep-core/src/tainting/Dataflow_tainting.ml
+++ b/semgrep-core/src/tainting/Dataflow_tainting.ml
@@ -23,6 +23,7 @@ module PM = Pattern_match
 module R = Rule
 module LV = IL_lvalue_helpers
 module T = Taint
+module Lval_env = Taint_lval_env
 module Taints = T.Taint_set
 
 let logger = Logging.get_logger [ __MODULE__ ]
@@ -65,14 +66,6 @@ type a_propagator = {
   var : var;
 }
 
-type taint_info = Tainted of Taints.t | MarkedClean
-
-let equal_taint_info a b =
-  match (a, b) with
-  | Tainted taints1, Tainted taints2 -> Taints.equal taints1 taints2
-  | MarkedClean, MarkedClean -> true
-  | _ -> false
-
 type config = {
   filepath : Common.filename;
   rule_id : string;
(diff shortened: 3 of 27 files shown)


IagoAbal:
We have some failures in semgrep-rules tests although that may just be a matter of fixing test expectations.

There is also:
> FAILED tests/e2e/test_performance.py::test_debug_performance - AssertionError...

IagoAbal:
> If a field a.b.c is marked tainted, then a, a.b and a.b.c are all marked tainted

That seems to trade off more FPs for fewer FNs. Is that what we want? What about having an `option:` for it, being the default to minimize FPs?

> One major short-coming, other than the lack of alias analysis, is that when a variable is marked tainted because one of it's fields was tainted, we do not remember what fields caused the taint.

IMO this ☝️ is yet another reason for not doing that kind of taint propagation by default. This will cause even more FPs.

Doing proper field sensitive analysis is tricky in general, so I think our main goal should be to remove FPs. It's nice to also report taint that goes through fields, but we only need to support the most trivial cases.

semgrep-core/src/core/il/IL_lvalue_helpers.mli
```diff
@@ -1,2 +1,3 @@
-val lvar_of_instr_opt : IL.instr -> IL.name option
+val dotted_lvars_of_lval : IL.lval -> IL.name list
+val lvar_of_instr_opt : IL.instr -> (IL.name * IL.name * IL.name list) option
```
IagoAbal: Maybe a comment explaining what all those `IL.name`s are.

semgrep-core/src/core/il/IL_lvalue_helpers.ml
```diff
@@ -90,13 +92,39 @@ let rlvals_of_instr x =
...
-  match lval.base with
-  | Var n -> Some n
-  | VarSpecial _
-  | Mem _ ->
-      None
+  match lval with
+  | { base = Var name; _ } -> (
+      let dots = dotted_lvars_of_lval lval in
+      match dots with
+      | dotted_name :: _ -> Some (name, dotted_name, dots)
```
IagoAbal: This may be obvious later on as I review the other changes, but right here I don't understand why we need to return `dotted_name` separately here. Isn't it already included in `dots`? We could have a short explanation here.

semgrep-core/src/core/il/IL_lvalue_helpers.mli
```diff
@@ -1,2 +1,3 @@
-val lvar_of_instr_opt : IL.instr -> IL.name option
+val dotted_lvars_of_lval : IL.lval -> IL.name list
```
IagoAbal: This deserves a comment too I think, it could just be an example input-output pair.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -68,20 +75,22 @@ type config = {
...
+    var option -> T.finding list -> taint_info Dataflow_core.env -> unit;
 }
 
-type mapping = Taints.t Dataflow_core.mapping
+type mapping = taint_info Dataflow_core.mapping
 
 (* HACK: Tracks tainted functions intrafile. *)
 type fun_env = (var, PM.Set.t) Hashtbl.t
-type var_env = Taints.t VarMap.t
+type field_env = (var, var) Hashtbl.t
```
IagoAbal: Useful to have a comment here explaining what is this `var -> var` mapping.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -58,6 +58,13 @@ type overlap = float
 type propagator_id = var
 type propagator_from = propagator_id
 type propagator_to = propagator_id
+type taint_info = Tainted of Taints.t | MarkedClean
```
IagoAbal: Ideally we didn't need `MarkedClean` but I don't have any better proposal right now, I'll think on it.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -349,24 +375,50 @@ let rec check_tainted_expr env exp : Taints.t * var_env =
...
     | Index e -> check env e
-    | NoOffset
-    | Dot _ ->
-        (Taints.empty, env.var_env)
+    | Dot _ -> (Taints.empty, env.var_env)
   in
   let check_subexpr exp =
     match exp.e with
-    | Fetch { base = VarSpecial (This, _); offset = Dot fld; _ } ->
+    | Fetch { base = VarSpecial (This, _); offset = Dot fld :: _; _ } ->
```
IagoAbal: Maybe the pattern here should be `[Dot fld]` instead of `Dot  fld :: _`?

semgrep-core/src/core/il/IL_lvalue_helpers.ml
```diff
@@ -90,13 +92,39 @@ let rlvals_of_instr x =
...
+      let str_of_name name = Common.spf "%s-%d" (fst name.ident) name.sid in
+      let dot_strs =
+        List.fold_right
+          (fun o s ->
+            match (s, o) with
+            | Some (s :: ss), Dot name ->
+                let x = s ^ "." ^ str_of_name name in
+                Some (x :: s :: ss)
+            | Some [], Dot name -> Some [ "." ^ str_of_name name ]
+            | _ -> None)
```
IagoAbal: I think that rather than `_` here we should write `Some _, Index _` and `None, _`. And explain why we handle `Some _, Index _` in that particular way.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -349,24 +375,50 @@ let rec check_tainted_expr env exp : Taints.t * var_env =
...
           |> PM.Set.elements |> T.taints_of_pms
         in
         (taints, env.var_env)
-    | Fetch { base; offset; _ } ->
-        let base_taints, var_env = check_base env base in
-        let offset_taints, var_env = check_offset { env with var_env } offset in
-        (Taints.union base_taints offset_taints, var_env)
+    | Fetch ({ base; offset; _ } as lval) -> (
+        let dotted_lvars = IL_lvalue_helpers.dotted_lvars_of_lval lval in
+        let x =
```
IagoAbal: `x` isn't a very meaningful name here! 😄

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -349,24 +375,50 @@ let rec check_tainted_expr env exp : Taints.t * var_env =
...
+          match dotted_lvars with
+          | _ :: _ as ns ->
+              let rec go = function
```
IagoAbal: This all seems a bit overcomplicated. Instead of running a `map` and then `go`, why not` go` just calls `VarMap.find_opt` itself and returns the first one that has some `taint_info` ?

IagoAbal: Also, it's worth explaining what we are doing. Given `x1. ... . xn` we're trying to find the longest `x1. ... . xm` for which we have some taint info.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -349,24 +375,50 @@ let rec check_tainted_expr env exp : Taints.t * var_env =
...
           Hashtbl.find_opt env.fun_env (str_of_name fld)
           |> Option.value ~default:PM.Set.empty
           |> PM.Set.elements |> T.taints_of_pms
         in
         (taints, env.var_env)
-    | Fetch { base; offset; _ } ->
-        let base_taints, var_env = check_base env base in
-        let offset_taints, var_env = check_offset { env with var_env } offset in
-        (Taints.union base_taints offset_taints, var_env)
+    | Fetch ({ base; offset; _ } as lval) -> (
```
IagoAbal: This code is quite long now it may be worth having a separate `check_lval` ?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -349,24 +375,50 @@ let rec check_tainted_expr env exp : Taints.t * var_env =
...
+              |> Common.map (fun x ->
+                     VarMap.find_opt (str_of_name x) env.var_env)
+              |> go
+          | [] -> None
+        in
+        match x with
+        | Some MarkedClean -> (Taints.empty, env.var_env)
+        | Some (Tainted _)
+        | None ->
+            let var_taints, var_env =
```
IagoAbal: Could we move this `let` before the `match x with` so that we don't need the `failwith "impossible"` ?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -349,24 +375,50 @@ let rec check_tainted_expr env exp : Taints.t * var_env =
...
+              match x with
+              | Some (Tainted t) -> (t, env.var_env)
+              | None -> (Taints.empty, env.var_env)
+              | Some _ -> failwith "impossible"
+            in
+            let base_taints, var_env = check_base { env with var_env } base in
+            let offset_taints, var_env =
```
IagoAbal: If `a.b.c` is tainted, and `a.b` is also tainted, shouldn't we perhaps union the taints of `a.b` and `a.b.c` ?

IagoAbal: But if we don't do this, and the lval is of the form `a.b.c`, then maybe we shouldn't be checking `a` for taint?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -536,6 +588,38 @@ let check_tainted_return env tok e : Taints.t * var_env =
...
+
+let is_dotted_prefix clean_var new_var =
+  parse_str clean_var |> function
+  | None -> false
+  | Some (id0, dots0, sid0) -> (
+      parse_str new_var |> function
+      | None -> false
+      | Some (id1, dots1, sid1) ->
+          String.equal id0 id1 && String.equal sid0 sid1
+          && is_prefix
```
IagoAbal: `String.starts_with` ?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -536,6 +588,38 @@ let check_tainted_return env tok e : Taints.t * var_env =
...
+  | _ -> None
+
+let rec is_prefix prefix str =
+  match (prefix, str) with
+  | c0 :: prefix, c1 :: str -> Char.equal c0 c1 && is_prefix prefix str
+  | [], _ -> true
+  | _ :: _, [] -> false
+
+let is_dotted_prefix clean_var new_var =
+  parse_str clean_var |> function
```
IagoAbal:
```ocaml
let (let*) = Option.bind in
let* x = parse_str clean_var in
let* y = parse_str new_var in
...
```
I think you can even define `let*` to return `false` by default. It may be cleaner?

IagoAbal: Otherwise a `match ... with ...` may be more standard in OCaml. Even `match parse_str clean_var, parse_str new_var with | None, _ | _, None ->. false | ...`

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -557,34 +641,52 @@ let input_env ~enter_env ~(flow : F.cfg) mapping ni =
...
         let var_env' =
           match LV.lvar_of_instr_opt x with
           | None -> var_env'
-          | Some var ->
+          | Some (_, var, _) ->
               (* We call `check_tainted_var` here because the assigned `var`
                * itself could be annotated as a source of taint. *)
               check_tainted_var { env with var_env = var_env' } var |> snd
         in
         match (Taints.is_empty taints, LV.lvar_of_instr_opt x) with
```
IagoAbal: We're calling `LV.lvar_of_instr_opt x` for a second time here, maybe we should put this in a `let` earlier on?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -557,34 +641,52 @@ let input_env ~enter_env ~(flow : F.cfg) mapping ni =
...
         (* Instruction returns safe data, remove taint from `var`. *)
-        | true, Some var -> VarMap.remove (str_of_name var) var_env'
+        | true, Some (_, var, _) ->
+            let var = str_of_name var in
+            VarMap.fold
```
IagoAbal: I think doing it this way may cause too much allocation? Basically we're creating a new `VarMap`. Couldn't we just add to the existing `VarMap` whatever is needed? Also, unless `var` is of the form `a.b....` I think we should still do the regular `VarMap.remove`.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -557,34 +641,52 @@ let input_env ~enter_env ~(flow : F.cfg) mapping ni =
...
+              (fun str taint map ->
+                if is_dotted_prefix var str then VarMap.add str MarkedClean map
+                else VarMap.add str taint map)
+              (VarMap.add var MarkedClean var_env')
+              VarMap.empty
         (* Instruction returns tainted data, add taints to `var`. *)
-        | false, Some var -> add_taint_to_var_in_env var_env' var taints
+        | false, Some (base_name, _, dots) ->
+            List.fold_left
+              (fun var_env var -> add_taint_to_var_in_env var_env var taints)
```
IagoAbal: As I said, I think in this case we should just taint the `last dots` or `base_name` if `dots = []`. But given `a.b.c = user_input` I don't think we should be tainting `a` and `a.b`, just `a.b.c`. Due to aliasing, it could happen that tainting `a.b.c` should taint `a` itself, but I think we don't need to handle those cases, better to report fewer FPs.

semgrep-core/src/tainting/Dataflow_tainting.mli
```diff
@@ -10,6 +10,7 @@ type overlap = float
 type propagator_id = var
 type propagator_from = propagator_id
 type propagator_to = propagator_id
+type taint_info = Tainted of Taint.taints | MarkedClean
```
IagoAbal: From a memory perspective, it may be more efficient to have:
```ocaml
type var_env = {
  taints : Taint.taints;
  clean : VarSet.t;
}
```
Otherwise we are adding an extra word overhead for every variable in the map. In general it doesn't matter, except when we encounter some large function or machine generated code. Although we shouldn't get that code to be analyzed in the first place, but we still do...

IagoAbal (approved):
Great work! Looking forward to seeing what do we get out of this feature. Hopefully fewer FPs!

semgrep-core/src/tainting/Dataflow_tainting.ml
mmcqd: I don't think so. When we produced lots of intermediate values, this would have worked on the IL generated by a Generic AST that had multiple offsets that _ended_ in a `Dot`. So we should replicate that.

semgrep-core/src/tainting/Dataflow_tainting.ml
mmcqd: I never actually used this 😅 . Deleting it

semgrep-core/src/core/il/IL_lvalue_helpers.ml
mmcqd: It was in fact not needed

semgrep-core/src/tainting/Dataflow_tainting.ml
mmcqd: I think it's not so bad now that I refactored a bit

semgrep-core/src/tainting/Dataflow_tainting.ml
mmcqd: Was this added after `4.12`? I need to upgrade haha

semgrep-core/src/tainting/Dataflow_tainting.ml
mmcqd: The `VarMap` is persistent, so we make a new one everytime we add anything right? There's no way to in place update. I agree that we can just do the regular remove when we don't have a `a.b. ...`

semgrep-core/src/tainting/Dataflow_tainting.mli
mmcqd: I switched to remove variables in the non-dot case, so I think your suggestion would now be equivalent to what we have

semgrep-core/src/tainting/Dataflow_tainting.ml
mmcqd: hmmm, I am not totally sure...

semgrep-core/src/tainting/Dataflow_tainting.ml
mmcqd: I agree that it feels not so elegant. But I couldn't come up with a better solution.

mmcqd:
> That seems to trade off more FPs for fewer FNs. Is that what we want? What about having an `option:` for it, being the default to minimize FPs?

I removed this feature, but left the implementation in a comment. It should be very easy to add if it starts getting requested.

mmcqd:
The failing semgrep_rules test in indeed something I had to update, and did so to the submodule as part of this PR. semgrep-rules PR is here: https://github.com/returntocorp/semgrep-rules/pull/2219

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: That's a great point!

IagoAbal: > That's a great point!

Wait a second! Previously, if there were multiple offsets like `this.x.y` then the `base` would not have been `VarSpecial (This, _)` right? Instead we would have an auxiliary variable `_tmp` assigned to `this.x`, and the only field access that would be handled by this _case_ would be `this.x`. Then `_tmp.y` would be handled in some other way. So I think this didn't use to work that way.

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: `VarMap` is persistent indeed, so every time we operate on it we get a "new" `VarMap`. However, under the hood, I would hope there will a good amount of sharing going on. I am aware that balancing may mess that up, but it shouldn't occur every time you add something. However with the `fold` it's guaranteed that we re-allocate the entire `VarMap` every time. 🤔 We've had timeout issues before with minimized JS files or large autogenerated files, that the CLI lets pass through, and the dataflow engine was super slow due to memory allocations. So that's why I'm concerned about this even before we experience any (new) problem.

IagoAbal: See PR #3440 for example.

semgrep-core/src/tainting/Dataflow_tainting.mli
IagoAbal: It's not really the same thing, I would say they are unrelated issues. So previously we had a mapping from `var` to `Taint.taints`. Now the mapping is from `var` to `taint_info`. So we are wrapping `Taint.taints` with an extra constructor `Tainted`. That extra constructor means memory overhead, and it's an overhead we will pay for every `var` in the environment. So N `var`s means N times that overhead, and in the worst case scenario you would have to multiply N by the number of mappings that we have, and we have two for each node in the CFG.

WIth
```ocaml
type var_env = {
  taints : Taint.taints;
  clean : VarSet.t;
}
```
the overhead is fixed wrt the number of variables in the mapping. Although it's still related to the number of nodes in the CFG.

That said no need to do anything about this now, I can think on this myself later on. I just wanted to explain better my comment. But correct me if you think my reasoning is wrong here.

Hope you don't take it bad that I'm nitpicking on these memory details!

IagoAbal:
> The failing semgrep_rules test in indeed something I had to update, and did so to the submodule as part of this PR. semgrep-rules PR is here: [returntocorp/semgrep-rules#2219](https://github.com/returntocorp/semgrep-rules/pull/2219)

You need to update the `semgrep-rules` dependency in `semgrep-core` to your branch. If that makes the tests pass, then we can merge.

semgrep-core/src/tainting/Dataflow_tainting.mli
IagoAbal: I wasn't thinking right when I proposed the solution. Although the problem is real. I need to think if there is another way.

IagoAbal:
I just want to first generalize `Dataflow_core` so that we can track dataflow facts on arbitrary keys rather than having to use strings, which in this PR is forcing us to do some ugly stuff (which I'm afraid won't be very  good for perf in extreme cases).

IagoAbal (changes requested):
Needs a clean up.

IagoAbal:
@mmcqd could you review the latest version?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -627,47 +613,60 @@ let input_env ~enter_env ~(flow : F.cfg) mapping ni =
...
-        (* Instruction returns tainted data, add taints to `var`. *)
-        | false, Some var -> add_taint_to_var_in_env var_env' var taints
-        (* There is no variable being assigned, presumably the Instruction
-         * returns 'void'. *)
-        | _, None -> var_env')
+        match (Taints.is_empty taints, opt_lval) with
+        (* Instruction returns safe data, remove taint from lval. *)
+        | true, Some lval when LV.lval_is_var_and_dots lval ->
+            Lval_env.clean lval lval_env'
+        | true, Some { base = Var x; _ } ->
```
mmcqd, line 650: Do we want arrays to have different behavior than structs? If `x.y = safe` cleans `x.y` then why should `x[i] = safe` clean `x`?

semgrep-core/tests/OTHER/rules/field_sensitive3.js
```diff
@@ -0,0 +1,13 @@
+// If `a.b.c` is tainted, then all its prefixes are also tainted
```
mmcqd: Should probably change this comment since it's no longer true

semgrep-core/tests/OTHER/rules/field_sensitive4.js
```diff
@@ -0,0 +1,16 @@
+// If `a.b` is marked clean, then anything that is prefix of `a.b` should become clean
```
mmcqd: Should say "anything that has `a.b` as a prefix", or maybe "any suffix of `a.b`"

semgrep-core/src/tainting/Dataflow_tainting.ml
mmcqd: I guess the alternative is to have no cleaning for arrays at all, which is also inconsistent, so perhaps this is better.

IagoAbal: Yep, IIRC not cleaning `x` after `x[i] = safe` made some of our tests fail (due to false positives).

IagoAbal (approved):
🎉

PR #5706, opened 2022-07-08
chore(test-autofix): Prevent source pollution in test

nbrahms:
Previously, tests would intermittently fail due to pollution of
`cli/tests/e2e/targets`. This was caused by `test_autofix` creating
a temporary file within this path at test run time.

Issue was reproduced using `chmod -R 0400 cli/tests/e2e/targets`, then
running `pytest`.

This commit updates test_autofix to run entirely in a copied file-system
context, thereby avoiding any pollution of the test sources. As a bonus,
the test_autofix code has been drastically simplified and runs 2x
faster.

I've also renamed `run_semgrep_no_symlink` to
`run_semgrep_on_copied_files` for developer clarity.

This commit also updates snapshots to reflect the actual output of
semgrep when `--dryrun` is ommitted from its invocation. (Note that the
output with `--dryrun` is still captured in a corresponding set of
snapshots).

No changelog needed with this commit.

PR checklist:

### DIFF
--- a/cli/tests/e2e/snapshots/test_autofix/test_autofix/rulesautofixtwo-autofixes.yaml-autofixtwo-autofixes.txt-not-dryrun/results.json
+++ b/cli/tests/e2e/snapshots/test_autofix/test_autofix/rulesautofixtwo-autofixes.yaml-autofixtwo-autofixes.txt-not-dryrun/results.json
@@ -17,9 +17,6 @@
       "extra": {
         "fingerprint": "070c76f072fcc5c05b90e7e6f2ed676961207fdae73acfd5c78253a8c2294126a6f19168fbbc00272bb3ca882b71f84df59a71277082f6235f5132effdf037a6_0",
         "fix": "one",
-        "fixed_lines": [
-          "one"
-        ],
         "is_ignored": false,
         "lines": "one\ntwo",
         "message": "This rule changes the line numbers for the other rule's match",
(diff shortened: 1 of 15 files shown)


cli/tests/e2e/snapshots/test_autofix/test_autofix/rulesautofixtwo-autofixes.yaml-autofixtwo-autofixes.txt-not-dryrun/results.json
```diff
@@ -44,9 +41,6 @@
       "extra": {
         "fingerprint": "36e519fa759ca74dea55c0aef65e0e3e081cdfeb4370a8b8a2546e77f65952ead0e128ca7b79e10645edaac3006b670ff0e122e96ee1569dca4105167a8b6b7e_0",
         "fix": "four",
-        "fixed_lines": [
```
mmcqd, line 47: We've lost all the `fixed_lines` in autofix output. Semgrep app uses those.

mmcqd (changes requested):

cli/tests/e2e/snapshots/test_autofix/test_autofix/rulesautofixtwo-autofixes.yaml-autofixtwo-autofixes.txt-not-dryrun/results.json
nbrahms: @mmcqd we actually haven't, they were just duplicated. See the squash commit message:

> This commit also updates snapshots to reflect the actual output of
semgrep when `--dryrun` is ommitted from its invocation. (Note that the
output with `--dryrun` is still captured in a corresponding set of
snapshots).

mmcqd: ack, reading is hard

mmcqd (approved):

PR #5721, opened 2022-07-12
Only include skipped_targets in result when --debug is passed

emjin:
Skipped_targets takes a huge amount of memory and isn't even read by core_output
when debug isn't set (see `core_output.py`). Frankly I have questions whether it's
useful at all in this form, but let's start by not including it in the final
result if `--debug` isn't passed.

On angular/angular, reduces final physical memory used by 1G.

I'd like to restructure our code so that skipped targets aren't being saved at all when
--debug isn't passed, but that's more work so fixing this part first.

PR checklist:

### DIFF
--- a/semgrep-core/src/core/Report.ml
+++ b/semgrep-core/src/core/Report.ml
@@ -175,7 +175,9 @@ let make_final_result results rules ~debug ~report_time ~rules_parse_time =
 
   (* These fields take a lot of space and aren't always necessary *)
   let skipped_targets =
-    if debug then results |> Common.map (fun x -> x.skipped_targets) |> List.flatten else []
+    if debug then
+      results |> Common.map (fun x -> x.skipped_targets) |> List.flatten
+    else []
   in
   let file_times = results |> Common.map (fun x -> x.profiling) in
   let final_profiling =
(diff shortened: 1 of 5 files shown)


semgrep-core/src/core/Report.ml
```diff
@@ -169,11 +169,15 @@ let collate_rule_results :
     profiling = { file; rule_times = profiling };
   }
 
-let make_final_result results rules ~report_time ~rules_parse_time =
+let make_final_result results rules ~debug ~report_time ~rules_parse_time =
   let matches = results |> Common.map (fun x -> x.matches) |> List.flatten in
   let errors = results |> Common.map (fun x -> x.errors) |> List.flatten in
+
+  (* These fields take a lot of space and aren't always necessary *)
```
IagoAbal: `... and aren't necessary except when running in debugging mode` ?

IagoAbal (approved):
The changes to Core look good! I would just make sure that our assumptions are correct and that CLI will never request `skipped_targets` unless `--debug` is given. We could also add a comment in the CLI's code, in case CLI devs one day decide to change this?

emjin (commented):
Clarify comment

Edit: huh I don't know what I expected to happen when I tried that box

emjin:
Failing test is build docker image because of github actions so I'm merging

PR #5722, opened 2022-07-12
Support not blocking triage ignored issues

ajbt200128:
PR checklist:

### DIFF
--- a/changelog.d/cli-162.added
+++ b/changelog.d/cli-162.added
@@ -1,0 +1,1 @@
+Semgrep CI will now not block builds on triage ignored issues
--- a/changelog.d/cli-162.added
+++ b/changelog.d/cli-162.added
@@ -1,1 +1,1 @@
-Semgrep CI will now not block builds on triage ignored issues
+`semgrep ci` will now not block builds on triage ignored issues
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -32,6 +32,8 @@ class ScanHandler:
         self._autofix = False
         self.dry_run = dry_run
         self._dry_run_rules_url: str = ""
+        self._skipped_syntactic_ids: List[str] = []
+        self._skipped_match_based_ids: List[str] = []
 
     @property
     def autofix(self) -> bool:
@@ -39,6 +41,20 @@ class ScanHandler:
         Seperate property for easy of mocking in test
         """
         return self._autofix
+
+    @property
+    def skipped_syntactic_ids(self) -> List[str]:
+        """
+        Seperate property for easy of mocking in test
+        """
+        return self._skipped_syntactic_ids
+
+    @property
+    def skipped_match_based_ids(self) -> List[str]:
+        """
+        Seperate property for easy of mocking in test
+        """
+        return self._skipped_match_based_ids
 
     def _get_deployment_details(self) -> Tuple[Optional[int], Optional[str]]:
         """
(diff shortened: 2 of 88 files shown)


changelog.d/cli-162.added
```diff
@@ -0,0 +1 @@
+Semgrep CI will now not block builds on triage ignored issues
```
brendongo:
```suggestion
`semgrep ci` will now not block builds on triage ignored issues
```

cli/tests/e2e/snapshots/test_ci/test_nosem/noautofix---enable-nosem/output.txt
```diff
@@ -22,6 +22,10 @@ Findings:
           ⋮┆----------------------------------------
          11┆ y == y
           ⋮┆----------------------------------------
+         23┆ b == b # Triage ignored by syntactic_id
```
brendongo, line 25: shouldn't these not show up here since they are ignored?

cli/src/semgrep/app/scans.py
```diff
@@ -156,7 +174,12 @@ def report_findings(
             for match in matches_of_rule
         ]
         new_ignored, new_matches = partition(
-            all_matches, lambda match: bool(match.is_ignored)
+            all_matches,
+            lambda match: bool(
+                match.is_ignored
+                or match.syntactic_id in self.skipped_syntactic_ids
```
brendongo: Is it correct to count the matches skipped cause of triage as a `new_ignored`

cli/tests/e2e/snapshots/test_ci/test_nosem/noautofix---enable-nosem/output.txt
ajbt200128: No, this test does not use the CI for rules, it uses the registry. So if there's no response from the app then there's no issues to triage ignore
No login token:
https://github.com/returntocorp/semgrep/blob/7e330b3f29b942a03237262e94d668037fcdf9a6/cli/tests/e2e/test_ci.py#L758-L766

cli/src/semgrep/commands/ci.py
```diff
@@ -385,6 +385,16 @@ def ci(
                 match for match in matches if not match.is_ignored or keep_ignored
             ]
         else:
+            if scan_handler:
```
brendongo, line 389: I think we should add a comment somewhere around here that mentions that blocking findings that have been triaged in UI count as non blocking findings

brendongo: Is that what is happening?

cli/src/semgrep/app/scans.py
ajbt200128: In the linear ticket and after talking with vivek they expect them to just go to the ignored endpoint which is anything in `new_ignored` I guess the name is a misnomer

cli/src/semgrep/commands/ci.py
ajbt200128: They don't count as non blocking findings they're counted as ignored findings i.e. `//nosemgrep` which are still reported to the backend just not the CLI output

brendongo (approved):

cli/src/semgrep/app/scans.py
ajbt200128: But then also new_matches is too as the matches could have been found before. I think `new` here means just to send to the app

PR #5724, opened 2022-07-12
chore(test): Remove additional test target pollution

nbrahms:
Like #5706, except captures the other test file where this pollution
occurred.

I used `rg NamedTemporaryFile\(.*targets.*\)` to make sure this doesn't
occur anywhere else.

No need for changelog.

PR checklist:

### DIFF
--- a/cli/tests/e2e/test_match_based_id.py
+++ b/cli/tests/e2e/test_match_based_id.py
@@ -34,23 +34,20 @@ def test_id_change(run_semgrep_on_copied_files, tmp_path, rule, target, expect_c
 
     suffix = Path(target).suffix
 
-    # Target file prior to edit
-    before_target = tmp_path / "targets" / "match_based_id" / "before" / target
-    # Target file after edit
-    after_target = tmp_path / "targets" / "match_based_id" / "after" / target
     # Static file path (necessary to obtain static id)
     static_target = tmp_path / "targets" / ("_match_based_id" + suffix)
 
-    def run_on_target(target):
-        shutil.copy(target, static_target)
-        before_results, _ = run_semgrep_on_copied_files(
+    def run_on_target(subpath):
+        source_target = tmp_path / "targets" / "match_based_id" / subpath / target
+        shutil.copy(source_target, static_target)
+        results, _ = run_semgrep_on_copied_files(
             rule,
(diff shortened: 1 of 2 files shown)


ajbt200128 (approved):
LGTM

cli/tests/e2e/test_match_based_id.py
```diff
@@ -30,36 +30,24 @@ def test_duplicate_matches_indexing(run_semgrep_in_tmp, snapshot):
...
-        tf.seek(
-            0
-        )  # Seek to beginning since Semgrep will be reading from it. Just in case.
-        after_results, _errors = run_semgrep_in_tmp(
+def test_id_change(run_semgrep_on_copied_files, tmp_path, rule, target, expect_change):
+
+    suffix = Path(target).suffix
+
+    # Static file path (necessary to obtain static id)
+    static_target = tmp_path / "targets" / ("_match_based_id" + suffix)
```
kopecs: I don't entirely follow how this aligns with the description I'm seeing for the `target_name` parameter

> `:param target_name: which directory within ./e2e/targets/ to scan`

but I'm not that familiar with the Python codebase.

ajbt200128: Hm I'm not sure why the description says which directory, it should be which directory or file

nbrahms: Fair; that description is probably incomplete; the path must be _within_ `<cwd>/e2e/targets`, but does not necessarily need to be specified _relative to_ `<cwd>/e2e/targets`.

Note that in this case, the current working directory is `tmp_path`; I specified the target as an absolute path so that things will bomb out if, for whatever reason, the working directory changes.

What's the best way to document this so it's clear?

kopecs: Maybe something like

:param target_name: which file or directory to scan. This must either be a relative path or an absolute path of a file contained in ./e2e/targets

If I understand you correctly?

nbrahms: sgtm; I updated this and other documentation

kopecs: lgtm then

spencerdrak (approved):

PR #5725, opened 2022-07-13
tainting: Add experimental taint labels feature

IagoAbal:
It provides the ability to attach labels to different kinds of taint.
Both sources and sinks can retrict what labels are present in the data
that passes through them in order to apply. This allows to write more
complex taint rules that previously required ugly workarounds. It also
allows us to write certain classes of typestate analyses.

Closes PA-1362

test plan:
make test # added 7 tests

PR checklist:

### DIFF
--- a/changelog.d/pa-1362.added
+++ b/changelog.d/pa-1362.added
@@ -1,6 +1,6 @@
 Add experimental support for _taint labels_, that is the ability to attach labels to
 different kinds of taint. Both sources and sinks can retrict what labels are present
 in the data that passes through them in order to apply. This allows to write more
-complex taint rules that previously required ugly workarounds. Taint labels also
-allows us to write certain classes of typestate analyses (e.g., check that a file
+complex taint rules that previously required ugly workarounds. Taint labels are also
+useful for writing certain classes of typestate analyses (e.g., check that a file
 descriptor is not used after being closed).
--- a/changelog.d/pa-1362.added
+++ b/changelog.d/pa-1362.added
@@ -1,0 +1,6 @@
+Add experimental support for _taint labels_, that is the ability to attach labels to
+different kinds of taint. Both sources and sinks can retrict what labels are present
+in the data that passes through them in order to apply. This allows to write more
+complex taint rules that previously required ugly workarounds. Taint labels also
+allows us to write certain classes of typestate analyses (e.g., check that a file
+descriptor is not used after being closed).
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -167,15 +167,30 @@ type pformula = New of formula | Old of formula_old [@@deriving show, eq]
 (* Taint-specific types *)
 (*****************************************************************************)
 
-type sanitizer_spec = {
+type taint_source = {
+  formula : pformula;
+  label : string;  (** The label to attach to the data *)
+  requires : AST_generic.expr;
+      (** A Boolean formula over taint labels, the expression that
+       is being checked as a source must satisfy this in order to the
+       label to be produced. Note that with 'requires' a taint source
+       behaves a bit like a propagator ... *)
+}
+[@@deriving show]
+
+type taint_sanitizer = {
   not_conflicting : bool;
       (** If [not_conflicting] is enabled, the sanitizer cannot conflict with
     a sink or a source (i.e., match the exact same range) otherwise
     it is filtered out. This allows to e.g. declare `$F(...)` as a sanitizer,
     to assume that any other function will handle tainted data safely.
     Without this, `$F(...)` would automatically sanitize any other function
-    call acting as a sink or a source. *)
-  pformula : pformula;
+    call acting as a sink or a source.
+
+    THINK: In retrospective, I'm not sure this was a good idea. We should add
+    an option to disable the assumption that function calls always propagate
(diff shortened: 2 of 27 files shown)


changelog.d/pa-1362.added
```diff
@@ -0,0 +1,5 @@
+Add experimental support for _taint labels_, that is the ability to attach labels to
+different kinds of taint. This allows to write more complex taint rules that
+previously required ugly workarounds. Taint labels also allows us to write certain
```
aryx: typo: us -> useful

IagoAbal:
Documentation ticket: [PA-1686](https://linear.app/r2c/issue/PA-1686)

semgrep-core/src/core/Rule.ml
```diff
@@ -167,15 +167,30 @@ type pformula = New of formula | Old of formula_old [@@deriving show, eq]
 (* Taint-specific types *)
 (*****************************************************************************)
 
-type sanitizer_spec = {
+type taint_source = {
+  formula : pformula;
+  label : string;  (** The label to attach to the data *)
+  requires : AST_generic.expr;
+      (** A Boolean formula over taint labels, the expression that
```
aryx: please do not use "formula" as we already use it for the formula/combination of patterns.

semgrep-core/src/core/Rule.ml
```diff
@@ -167,15 +167,30 @@ type pformula = New of formula | Old of formula_old [@@deriving show, eq]
...
+       is being checked as a source must satisfy this in order to the
+       label to be produced. Note that with 'requires' a taint source
+       behaves a bit like a propagator ... *)
```
aryx: Maybe you can add here that the expressions will be evaluated at matching time by Eval_generic?

semgrep-core/src/core/Rule.ml
```diff
@@ -167,15 +167,30 @@ type pformula = New of formula | Old of formula_old [@@deriving show, eq]
 (* Taint-specific types *)
 (*****************************************************************************)
 
-type sanitizer_spec = {
+type taint_source = {
+  formula : pformula;
+  label : string;  (** The label to attach to the data *)
```
aryx: Should it be an option? Or do you have a default value for it?
In that case maybe the default value should be defined in this file.

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -810,6 +812,8 @@ let parse_extract_reduction ~id (s, t) =
 (* Sub parsers taint *)
 (*****************************************************************************)
 
+let builtin_label_SOURCE = "__SOURCE__"
```
aryx: I would move that to Rule.ml

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -160,23 +171,71 @@ let merge_source_sink_mvars env source_mvars sink_mvars =
...
+  taints |> Taints.to_seq
+  |> Seq.filter_map (fun (t : T.taint) ->
+         match t.orig with
+         | Src src ->
+             let _, ts = T.pm_of_trace src in
+             Some ts.Rule.label
+         | Arg _ -> None)
+  |> LabelSet.of_seq
+
+let rec eval_label_requires ~labels e =
```
aryx, line 185: maybe add a (* coupling: if you modify the code here, you probably need to modify Parse_rule.parse_require_expression *)

semgrep-core/tests/OTHER/rules/taint_labels2.yaml
```diff
@@ -0,0 +1,17 @@
+rules:
+  - id: tainting
+    mode: taint
+    languages:
+      - python
+    message: |
+      This confirms taint mode works.
+    pattern-sinks:
+      - requires: TAINTED and not CLEANED
```
aryx, line 9: Does that mean sanitizer are now useless? Or they are still useful as a quicker way to express
source/sanitize/sink instead of having to invent labels and requires expressions?

semgrep-core/tests/OTHER/rules/taint_labels5.yaml
```diff
@@ -0,0 +1,23 @@
...
+              $FILE.close()
+          - focus-metavariable: $FILE
+      - label: reopened
+        requires: closed
+        patterns:
+          - pattern: |
+              $FILE = open(...)
+          - focus-metavariable: $FILE
+    pattern-sinks:
+      - requires: closed and not reopened
```
aryx, line 20: noice!

aryx (approved):

semgrep-core/tests/OTHER/rules/taint_labels2.yaml
IagoAbal: I am not sure to be honest, I think we'll have to gain some experience in using labels. Sanitizers do allow you to _completely remove_ taint from data, although I think that can be simulated with labels too. I could have translated (internally) `pattern-sanitizers` to `patter-sources` with a `"__SANITIZED__"` label, and then rewrite the `requires` of sinks as `(...) not __SANITIZED__`. But `not-conflicting` sanitizers cannot be simulated that way. Although I think we should replace `not-conflicting` with some `options:` because they are a bit confusing to use sometimes.

PR #5730, opened 2022-07-14
Reduce memory usage when not necessary

emjin:
This PR discards information that the user won't see much earlier in the process;
 in most cases as soon as possible. This is a change from previously, where we 
collected the information and only decided at the end what to include in the result. 
That extra information represents the majority of memory used by large repos.

In angular/angular with p/default, running with full verbosity (debug) results
in 3.34 GB used on exit, whereas running with only time information uses 1.67 GB
and running with no debug information uses 1.52 GB. (In future PRs, we expect to
be able to drop the 1.52 GB to below 1 GB)

Test plan: Manually compared `-json_time` by running on a small set of files and
confirmed that all the fields are still there.

PR checklist:


emjin:
Not quite ready for review but if people don't like this approach I'd like to know sooner rather than later

semgrep-core/src/core/Report.ml
```diff
@@ -1,9 +1,79 @@
...
-(* TODO: We could use atdgen to specify those to factorize type definitions
+(********************************************************************************
+ * Full result information
+ *
+ * In addition to the results (matches + errors), we also may report extra
+ * information, such as the skipped targets or the profiling times. Many
+ * of the types here are created to collect profiling information in as
+ * well-typed a manner as possible. Creating a type for practically every
+ * stage that reports matches is annoying, but prevents us from relying on
+ * dummy values or unlabeled tuples.
```
IagoAbal, line 9: Why not just an `option` type? Does that work with `atdpy`?

semgrep-core/src/core/Report.ml
```diff
@@ -1,9 +1,79 @@
...
+ * read, or by a previous result. In this way we ensure that the fields
+ * stored in the result are determined by the arguments passed by the user.
+ *
+ * Alternatives considered to the extra field:
+ * - storing the information but just ommitting it in the final result (I
+ *   tried this but it still uses too much memory)
+ * - using option types within the result types for field and a global
+ *   config to decide which fields are in use
+ *
+ * I considered the latter, but I'm not a fan of mix-and-match option types.
```
IagoAbal, line 35: You can also use `option` types without giving the user the choice of picking exactly what info they want.

semgrep-core/src/core/Report.ml
```diff
@@ -1,9 +1,79 @@
...
+type debug_mode = MDebug | MTime | MNo_info [@@deriving show]
+
+type 'a debug_info =
+  (* -debug: save all the information that could be useful *)
+  | Debug of {
+      skipped_targets : Output_from_core_t.skipped_target list;
+      profiling : 'a;
+    }
+  (* -json_time: save just profiling information; currently our metrics record this *)
+  | Time of 'a
```
IagoAbal: Isn't this the same as the `profiling` field above? If so, maybe call both "time" or call both "profiling"?

semgrep-core/src/core/Report.ml
IagoAbal: Sorry, I see it's answered below. 😄

semgrep-core/src/core/Report.ml
IagoAbal: I guess you can also repeat field names in other constructors and then do `Time of { profiling : 'a }` ?

semgrep-core/src/core/Report.ml
```diff
@@ -1,9 +1,79 @@
...
+  (* save nothing else *)
+  | No_info
+[@@deriving show]
+
+let mode = ref MNo_info
```
IagoAbal, line 72: Don't we have the config available somewhere when we are going to build a `Report`?

semgrep-core/src/core/Report.ml
IagoAbal: I like the approach you took, although the part of "giving the user the choice" kind of confused me, because I think it's an orthogonal issue.

IagoAbal:
> Not quite ready for review but if people don't like this approach I'd like to know sooner rather than later

I haven't reviewed the PR sure carefully but overall I personally like the approach. Main criticism I have is that we should work towards removing global variables, and in this case I'm not convinced that adding one saves us that much. We should have the `config` available at the points where we build a `Report`, so we just need to pass it.

Also, if `atdgen` supports this, it would be nice if it was also clear in the interface when e.g. `skipped_targets` isn't available.

IagoAbal:
> In angular/angular with p/default, running with full verbosity (debug) results
in 3.34 GB used on exit, whereas running with only time information uses 1.67 GB
and running with no debug information uses 1.52 GB. 

It's clear that we want to skip `skipped_targets` because the memory reduction is about 50%, but is it worthwhile to do the same for the profiling info? It seems to have little effect even on a large repo, and it does add extra complexity.

semgrep-core/src/core/Report.ml
emjin: That's fair, I'll see if I can get rid of the global cleanly

emjin:
> > In angular/angular with p/default, running with full verbosity (debug) results
> > in 3.34 GB used on exit, whereas running with only time information uses 1.67 GB
> > and running with no debug information uses 1.52 GB.
> 
> It's clear that we want to skip `skipped_targets` because the memory reduction is about 50%, but is it worthwhile to do the same for the profiling info? It seems to have little effect even on a large repo, and it does add extra complexity.

I'm not sure I'd call 150 MB "little effect". Remember that at least 800 GB of the 1.52 is because we're wasteful in how we create `targets`, and I'm personally suspicious of the remaining 700 GB. Also, I feel like the trend is definitely that we're getting bigger customers running on larger monorepos. I'd rather do the work now to remove json_time

semgrep-core/src/core/Report.ml
```diff
@@ -54,140 +123,198 @@ type final_result = {
...
+        logger#debug
+          "Mismatch between mode and result while creating final result";
+        empty_file_profiling
   in
-  let file_times = results |> Common.map (fun x -> x.profiling) in
-  let final_profiling =
-    if report_time then Some { rules; rules_parse_time; file_times } else None
+  let extra =
+    let mk_profiling () =
+      let file_times = results |> Common.map (fun x -> get_profiling x) in
```
IagoAbal: `Common.map get_profiling`, I think we don't need the `fun x -> ...`. Just in case it was unintentional, maybe you're fond of that style then no problem.

semgrep-core/src/core/Report.ml
emjin: The reason I said that is because if we were to give the user choice, then we would need to make every combination of options possible. In that case, I think it would make more sense to use option types.

semgrep-core/src/core/Report.ml
emjin: haha I totally didn't notice that :D Thanks for pointing it out

semgrep-core/src/core/Report.ml
emjin: It would be pretty annoying (would have to pass the config into a lot of places). Would you prefer it if I moved it to Flag_semgrep so at least all our globals were in the same place?

emjin:
(Tests now failing for an unrelated reason)

Edit: resolved!

aryx (approved):

PR #5732, opened 2022-07-14
fix: Changelog was not being processed properly

spencerdrak:
Ensures that the precommit hook runs on the changelog before it is committed. @underyx and I identified that the issue with the `+` character was a very rare (unlucky) occurrence, and is likely not worth our effort to fix right now. I've noted it. 

We also will be updating docs in another PR to ensure that all devs understand what should go in these changelog entries and how they're interpreted.

No changelog entry needed.

PR checklist:


underyx:
> this should just stop the wonky formatting we were seeing with certain chars (like +) in our changelog.

I don't think this was a prettier issue. According to the markdown spec, `+` is a valid list bullet point character: https://github.github.com/gfm/#bullet-list-marker. So regardless of what prettier does, the rendering would be messed up here https://github.com/returntocorp/semgrep/blob/v0.104.0/CHANGELOG.md#01040---2022-07-13

At least the prettier autofix made it clear we had this rendering problem.

underyx:
maybe we can do a regex replace of `^([-+*])` to `\$1` in the entries?

that'd escape bullet point characters at the start of line

brendongo (approved):

PR #5733, opened 2022-07-14
chore: hide internal findings in count

niklasmannhardt:
PR checklist:

### DIFF
--- a/cli/src/semgrep/output.py
+++ b/cli/src/semgrep/output.py
@@ -352,14 +352,19 @@ class OutputHandler:
                         ) from ex
 
         if self.filtered_rules:
-            fingerprint_matches, regular_matches = partition(
+            _, regular_matches = partition(
                 self.rule_matches,
                 lambda m: m.severity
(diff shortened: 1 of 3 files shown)


brendongo (dismissed):
Lets add to the changelog that reported rule count ignores inventory and experiment rules

ievans:
I worry this change could be pretty confusing to the user, and also make us appear extra slow. Is there any way to discover that this is happening in the default output? Does --verbose or --debug show these rules?

Are these rules still reported anyways else? E.g. in the "N rules x M files" header?

raghavjain3:
@ievans @brendongo perhaps we can have a line like "Additionally, scanned with X rules to detect technologies in this project"?

emjin:
> Are these rules still reported anyways else? E.g. in the "N rules x M files" header?

@ievans I would expect that they are unless it's been explicitly changed in this PR, because that's produced from uses the target matrix sent to semgrep-core

PR #5736, opened 2022-07-15
feat(ci): publish linux/arm64 image

wilsonehusin:
ref: https://github.com/docker/build-push-action/blob/42863b1282f7046f6c1ec9222965bbceafae3bb8/docs/advanced/multi-platform.md

PR checklist:

- [x] Documentation is up-to-date
- [x] `changelog.d/<issue>.<type>` is a file with the _what_, _why_, and _how_ of the change.
  - \<issue> is `pa-312` (Linear ticket), `gh-1234` (GitHub issue), or `new-gizmo` (unique semantic name)
  - \<type> is `added`, `changed`, `fixed`, or `infra`.
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure on any of this, please see the [Contribution guidelines](https://semgrep.dev/docs/contributing/how-to-contribute/)!

Edit: 

Closes #5704

PR #5747, opened 2022-07-19
feat(ci): publish linux/arm64 container for dev builds

wilsonehusin:
PR checklist:

  - This is supposed to be part of #5736, so I don't think it warrants another changelog entry — happy to do it if maintainers desire so.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -76,6 +76,8 @@ jobs:
     name: build docker image
     runs-on: ubuntu-latest
     steps:
+      - name: Set up QEMU # required for multiarch
+        uses: docker/setup-qemu-action@v2
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -147,7 +151,7 @@ jobs:
         with:
           push: true
           file: dockerfiles/semgrep-dev.Dockerfile
-          platforms: ${{ matrix.platform }}
+          platforms: linux/amd64,linux/arm64
```
wilsonehusin, line 154: I'm not sure if this means anything internally to maintainers, but from what I understand the current `{{ matrix.platform }}` usage here would result to empty string because I don't see any `strategy.matrix.platform` block. Let me know if there is a more desirable entry here.

brendongo:
@wilsonehusin #5736 will release the image on our next official release. I don't think we need to publish an arm image on every commit here.

wilsonehusin:
sounds good!

brendongo:
Thanks again for the contribution!

PR #5748, opened 2022-07-19
Reduce size of targets file

emjin:
Rule_ids are often quite long. Instead of listing out every rule_id each time,
map targets to indices and list all the rules once at the end.
    
This reduces physical memory consumption from ~1.5GB to ~800MB with -j 1.
(on angular/angular).
With -j 12, memory consumption is reduced from ~1GB per process to ~300GB
per process.
    
Test plan: make test

PR checklist:

### DIFF
--- a/cli/src/semgrep/core_runner.py
+++ b/cli/src/semgrep/core_runner.py
@@ -314,6 +314,8 @@ class Plan:
 
     def __init__(self, mappings: List[Task], rule_ids: List[str]):
         self.target_mappings = TargetMappings(mappings)
+        # important: this is a list of rule_ids, not a set
+        # target_mappings relies on the index of each rule_id in rule_ids
         self.rule_ids = rule_ids
 
     def split_by_lang_label(self) -> Dict[str, "TargetMappings"]:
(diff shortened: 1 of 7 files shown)


emjin:
Branched off https://github.com/returntocorp/semgrep/pull/5730 and would like it to be merged before I rebase onto develop

cli/src/semgrep/core_runner.py
```diff
@@ -699,7 +717,7 @@ def _run_rules_direct_to_semgrep_core(
                 print(" ".join(cmd))
                 sys.exit(0)
 
-            runner = StreamingSemgrepCore(cmd, len(plan))
+            runner = StreamingSemgrepCore(cmd, len(plan.target_mappings))
```
brendongo: nit: Usually less mental load and easier to make changes in the future if users of a plan don't need to know how Plan class keeps data internally. So would have plan expose a num_targets function/property

cli/src/semgrep/core_runner.py
```diff
@@ -291,21 +291,34 @@ def execute(self) -> int:
...
+    """
+    Saves and displays knowledge of what will be run
+
+    to_json: creates the json passed to semgrep_core - see Input_to_core.atd
+    log: outputs a summary of how many files will be scanned for each file
+    """
+
+    def __init__(self, mappings: List[Task], rule_ids: List[str]):
+        self.target_mappings = TargetMappings(mappings)
+        self.rule_ids = rule_ids
```
brendongo, line 319: Probably should add a comment that rule_ids being a list is important

brendongo (approved):

PR #5749, opened 2022-07-19
feat(extract): allow `generic` as language

### LINKED ISSUE #5741: Support `generic` as a destination language for extract mode
There are several use cases where no specific language we support would be the correct destination for an extract mode rule. Examples of this include:

- Django templates; extracting their DSL
- SQL embedded in lots of places (at least until we have robust SQL support)
- Inline asm

One possible issue with the current design is ensuring that relevant generic rules run on the extracted code. The differences in patterns may be sufficient for this, but we should consider if we need an alternative system to ensure your Django filter rule only matches on relevant extracted code, and not also on other configuration files, etc.. where the same text might appear.

kopecs:
PR checklist:

Closes #5741 / Fixes PA-1648

Like I said in #5741, I don't think this is the best final design, and if there's more interested in this feature then we may need to handle `paths:` with extracted targets. That said, the filtering logic is a layer above where we do the extraction, and just allowing this should just enable latent functionality, so I don't see why not.

### DIFF
--- a/semgrep-core/src/parsing/Parse_rule.ml
+++ b/semgrep-core/src/parsing/Parse_rule.ml
@@ -748,6 +748,15 @@ and parse_extra (env : env) (key : key) (value : G.expr) : Rule.extra =
       R.MetavarComparison { R.metavariable; comparison; strip; base }
   | _ -> error_at_key env key ("wrong parse_extra field: " ^ fst key)
 
+let parse_extract_dest ~id lang : Xlang.t =
+  match lang with
+  | ("none" | "regex"), _ -> LRegex
+  | "generic", _ -> LGeneric
+  | s, t -> (
+      match Lang.lang_of_string_opt s with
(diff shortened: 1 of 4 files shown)


semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -748,6 +748,15 @@ and parse_extra (env : env) (key : key) (value : G.expr) : Rule.extra =
       R.MetavarComparison { R.metavariable; comparison; strip; base }
   | _ -> error_at_key env key ("wrong parse_extra field: " ^ fst key)
 
+let parse_extract_dest ~id lang : Xlang.t =
+  match lang with
+  | ("none" | "regex"), _ -> LRegex
+  | "generic", _ -> LGeneric
+  | s, t -> (
+      match Lang.lang_of_string_opt s with
```
emjin: Nit: could use parse_language here

emjin (approved):

PR #5752, opened 2022-07-19
fix(rust): correctly handle scoped idents

### LINKED ISSUE #5717: Rust `use` declarations only include last element
**Description**
I have written a simple custom rule for Rust language. Running the custom rule in Semgrep CLI matches lines of code extraneously that does not match the expected output for the custom rule.

**To Reproduce**
python3 -m venv /path/to/venv
source /path/to/venv/bin/activate
python -m pip install semgrep
semgrep --config=prng.yml prng.rs

prng.yml:
```
rules:
- id: rand-prng
  patterns: 
    - pattern: use rand::$X;
  message:
    The product uses a Pseudo-Random Number Generator (PRNG) in a security context, but the PRNG's algorithm is not cryptographically strong. 
  languages: [rust]
  severity: ERROR
```
prng.rs:
```
use rand::FromEntropy;
use rand::prng::XorShiftRng;
use mini_redis::{client, Result};
use rand::prng::ChaChaRng;
use rand::prng::Hc128Rng;

pub async fn send(&self, message: Message) -> Result<()> {

}
```

**Expected behavior**
It is expected that custom rule id: rand-prng would not match:
```
use mini_redis::{client, Result};
```

Actual behaviour:
Screenshot 2022-07-12 at 4 49 12 PM

**Environment**
 MacBook Pro, 2021, Apple M1 Max
macOS Monterey 12.4
Python 3.9.13

**Use case**
Fixing this bug enables adoption of semgrep.

brandonspark:
What:
Currently, matching is broken for Rust scoped identifiers, because all but the last component of a scoped identifier is ignored. This PR properly deals with creating a name for such scoped identifiers.

For instance, the AST for the following Rust program:
```rust
let x = a::x::collect
```

looked like
```
Pr(
  [DefStmt(
     ({
       name=EDynamic(
              OtherExpr(("LetPat", ()),
                [P(
                   PatId(("x", ()),
                     {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))]));
       attrs=[]; tparams=[]; },
      VarDef(
        {
         vinit=Some(N(
                      Id(("collect", ()),
                        {id_info_id=4; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); })));
         vtype=None; })))])
```

After this PR, it looks like:
```
Pr(
  [DefStmt(
     ({
       name=EDynamic(
              OtherExpr(("LetPat", ()),
                [P(
                   PatId(("x", ()),
                     {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))]));
       attrs=[]; tparams=[]; },
      VarDef(
        {
         vinit=Some(N(
                      IdQualified(
                        {name_last=(("collect", ()), None); name_top=None; 
                         name_middle=Some(QDots([(("a", ()), None); (("x", ()), None)])); 
                         name_info={id_info_id=3; id_hidden=false; id_resolved=Ref(
                                    None); };
                         })));
         vtype=None; })))])
```

Why:
We should support this functionality, because it breaks matching.

How:
This code picks up on a leftover TODO, which was to dispatch on a variable which was declared, but not used. This involves extending the name in the case of a longer identifier, and also possibly adding a type argument to the name.

Fixes #5717 

Test plan: `make test`

PR checklist:


semgrep-core/src/parsing/tree_sitter/Parse_rust_tree_sitter.ml
```diff
@@ -2480,7 +2480,7 @@ and map_return_expression (env : env) (x : CST.return_expression) : G.expr =
 
 and map_scoped_identifier_name (env : env)
     ((v1, v2, v3) : CST.scoped_identifier) : G.name =
-  let _either_optTODO =
+  let _either_opt =
```
kopecs: This could probably get renamed to something more descriptive (and without an underscore in front).

semgrep-core/tests/rust/scoped_idents.rs
```diff
@@ -0,0 +1,2 @@
+let x = A::f;
```
kopecs:
```suggestion
// ERROR: match
let x = A::f;
```

kopecs: We need an annotation to tell the test harness what matches we expect / don't.

kopecs (commented):
Could you add some tests like the ones from https://github.com/returntocorp/semgrep/issues/5717
that verify the naming with `use` works as part of this too?

An additional example with an alias (i.e., a `use` with `as`) would be good too.

semgrep-core/src/parsing/tree_sitter/Parse_rust_tree_sitter.ml
```diff
@@ -2495,10 +2495,30 @@ and map_scoped_identifier_name (env : env)
...
   let _colons = token env v2 (* "::" *) in
-  let ident = ident env v3 in
+  let last_id = ident env v3 in
   (* pattern (r#)?[a-zA-Zα-ωΑ-Ωµ_][a-zA-Zα-ωΑ-Ωµ\d_]* *)
   (* TODO: use either_opt *)
-  H2.name_of_id ident
+  match prefix_info with
+  | Some (Left (Id (first_id, _))) -> H2.name_of_ids [ first_id; last_id ]
+  | Some (Left (IdQualified ({ name_last; name_middle; _ } as q_info))) ->
+      let new_name_middle =
```
aryx: maybe there could be a H2.add_xxx_to_name

aryx (approved):
Ideally we would also like to not have those DefStmt EDynamic EotherPat but
just a DefStmt with EN when we have rust code like 'let x = ...'.

Also usually the tests in semgrep-core/tests/<lang>/ follow some conventions that are used to generate documentation (see scripts/generate-cheatsheet), with names like dots_xxx.sgrep, metavar_xxx.sgrep. In that case for parsing fixes, it's maybe better to use misc_scoped_ident.rs

changelog.d/gh-5717.fixed
```diff
@@ -0,0 +1 @@
+Rust: Fixed a bug with matching for scoped identifiers
```
IagoAbal, line 1: Same here, if you just summarized the what/why of your PR comment here it would make a great changelog entry. Then Bruno and the technical writers take these changelog entries and they produce some release notes, so they appreciate the extra details.

semgrep-core/src/core/ast/AST_generic_helpers.ml
```diff
@@ -128,6 +128,26 @@ let name_of_ids xs =
...
 
+let add_suffix_to_name suffix name =
+  match name with
+  | Id (first_id, _) -> name_of_ids [ first_id; suffix ]
+  | IdQualified ({ name_last; name_middle; _ } as q_info) ->
+      let new_name_middle =
+        let middle_quals =
+          match name_middle with
+          | Some (QDots qualifiers) -> qualifiers
+          | Some (QExpr _) -> raise Impossible
```
kopecs: If this is a more generic routine we may want to handle this case, in the event that someone working on Ruby, etc...

PR #5755, opened 2022-07-19
fix(scala): allowed percolation of name information from class parameters

### LINKED ISSUE #5506: Metavariables do not work for object reference in Scala
**Describe the bug**
Metavariables do not work for object reference in Scala

**To Reproduce**
The link is attached https://semgrep.dev/s/zojz

**Expected behavior**
Based on the link, there should be two matches for the pattern, line 3 and line 8.

**Screenshots**

**Environment**

**Use case**
It allows us to find usage of methods in code base.

brandonspark:
**What:**
Currently, during the naming stage, class parameters are not properly added to the scope. This means that certain identifiers with type information attached to them are not actually known to Semgrep to have those types, if introduced via class parameters. 

**Why:**
We should allow this, because otherwise Semgrep cannot fulfill certain typed metavariable queries.

For instance, in Scala:
```scala
class ExampleClass(val x: TypeName) {
}
```

The identifier `x` should be known to have type `TypeName`, but currently does not.

**How:**
I added some code which adds these parameters to the scope. I used `with_new_function_scope`, even though this is a class scope, since when looking through the code, it seems that this is just a general scoping thing which is undone later. Let me know if that's not the right thing to do, though.

Fixes #5506.

Test plan: `make test`

PR checklist:

### DIFF
--- a/changelog.d/gh-5506.fixed
+++ b/changelog.d/gh-5506.fixed
@@ -1,1 +1,10 @@
 Scala/others: Added a fix allowing percolation of name information from class parameters
+
+For example, classes which take in arguments like the following in Scala:
+```scala
+class ExampleClass(val x: TypeName) {
+}
+```
+
+do not properly enter the context. So in our analysis, we would not know that the identifier
+`x` has type `TypeName`, within the body of `ExampleClass`.
--- a/changelog.d/gh-5506.fixed
+++ b/changelog.d/gh-5506.fixed
@@ -1,0 +1,1 @@
+Scala/others: Added a fix allowing percolation of name information from class parameters
--- a/semgrep-core/src/naming/Naming_AST.ml
+++ b/semgrep-core/src/naming/Naming_AST.ml
@@ -457,7 +457,9 @@ let resolve lang prog =
           (* todo: we should first process all fields in the class before
            * processing the methods, as some languages may allow forward ref.
            *)
-          with_new_context InClass env (fun () -> k x));
+          let class_params = params_of_parameters env x.cparams in
+          with_new_context InClass env (fun () ->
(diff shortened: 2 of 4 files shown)


semgrep-core/src/naming/Naming_AST.ml
```diff
@@ -457,7 +457,9 @@ let resolve lang prog =
           (* todo: we should first process all fields in the class before
            * processing the methods, as some languages may allow forward ref.
            *)
-          with_new_context InClass env (fun () -> k x));
+          let class_params = params_of_parameters env x.cparams in
+          with_new_context InClass env (fun () ->
+              with_new_function_scope class_params env.names (fun () -> k x)));
```
aryx, line 463: maybe add a (* TODO? maybe we need a with_new_class_scope to handle this? abusing with_new_function_scope for now *)

aryx (approved):
Great!

Would be nice to also add this feature in deep-semgrep where we have different code to perform naming (basically each time we touch Naming_AST.ml (which BTW is a ugly hack), we need to also update in deep-semgrep Naming_SAST.ml (which is far cleaner)).

IagoAbal:
Great PR comments!

changelog.d/gh-5506.fixed
```diff
@@ -0,0 +1 @@
+Scala/others: Added a fix allowing percolation of name information from class parameters
```
IagoAbal, line 1: You could elaborate the what/why more here. You could refer to the same example you used in the PR comment:
```Scala
class ExampleClass(val x: TypeName) {
}
```
But you don't need to repeat yourself, if something is explained in the changelog entry you could omit details in the PR comment.

PR #5757, opened 2022-07-20
fix(scala): allow single metavariables in imports

### LINKED ISSUE #5219: Scala: metavariable-regex does not work for import statements
**Describe the bug**
Given a sample Scala as shown below:
```scala
import javax.servlet.{Filter, FilterChain, ServletRequest, ServletResponse}

class SsoFilter(
                 @Value("${sitesso.cookie.secure}") siteSsoCookieSecure: Boolean
               ) extends Filter {
                   override def doFilter(request: ServletRequest, response: ServletResponse, chain: FilterChain): Unit = {
                       val context = new ServletRequestContext(request.asInstanceOf[HttpServletRequest], response.asInstanceOf[HttpServletResponse])
                   }
               }
```
the following rule throws `[ERROR] Pattern parse error in rule scala-bug:
 Invalid pattern for ScalaEngine(InternalError)`
```yaml
id: scala-bug
severity: ERROR
message: "HTTP Request Filter: $BASE_CLASS->$FILTER_METHOD"
languages: [scala]
patterns:
- metavariable-regex:
      metavariable: $PACKAGE
      regex: (?:org\.springframework\.web\.filter\.(?:\*|OncePerRequestFilter)|javax\.ws\.rs\.container\.(?:\*|ContainerRequestFilter)|javax\.servlet\.(?:\*|Filter)|org\.springframework\.web\.servlet\.(?:\*|HandlerInterceptor))
- pattern: import $PACKAGE
```
However the following rule does work 
```yaml
id: scala-bug
severity: ERROR
message: "HTTP Request Filter: $BASE_CLASS->$FILTER_METHOD"
languages: [scala]
patterns:
- metavariable-regex:
    metavariable: $FILTER_METHOD
    regex: ^(?:filter|doFilter)$
- metavariable-regex:
    metavariable: $BASE_CLASS
    regex: ^(?:ContainerRequestFilter|Filter)$
- pattern-either:
    - pattern: |
        class $CLASS_NAME(...) extends $BASE_CLASS {
        ...
        def $FILTER_METHOD(...): Unit = {
        ...
        }
        }
```
**To Reproduce**
Use this playground URL: https://semgrep.dev/playground/s/dpnishant:scala-bug reproduce the bug. Compare the result with this playground URL: https://semgrep.dev/playground/s/dpnishant:scala-http-filter

Interestingly the same pattern works in `Java` language example: https://semgrep.dev/playground/s/dpnishant:audit-java-http-filter

**Expected behavior**
metavariable-regex should ideally match patterns in the import statements too as it does elsewhere.

**Screenshots**
Not available. Please use the Playground URLs for validation.

**Environment**
Using semgrep 0.92.0 on macOS (via Homebrew)

**Use case**
Better semantic coverage for Scala language

brandonspark:
Currently blocked by https://github.com/returntocorp/pfff/pull/541.

**What:**
Scala `import` parsing is currently broken, and both does not conform to the specification, and does not handle metavariables properly. This manifests in Semgrep patterns such as `import $X` being rejected as invalid patterns at parse time. 

**Why:**
We should fix this because the above import pattern is useful to be able to do, particularly when `$X` is a metavariable regex or something.

**How:**
Via [scala-lang](https://www.scala-lang.org/files/archive/spec/2.11/13-syntax-summary.html), we obtain:
Screen Shot 2022-07-19 at 10 38 23 PM

Unrolling this once, we see that we get something like:
```
StableId ::= id
           | StableId '.' `id`
           | [id '.'] `this` `.` id
           | [id '.'] `super` [ClassQualifier] '.' id
```

Basically, a `StableId` starts as either a single `id`, or a `id.this.id2`, or a `id.super [class qualifier]`, and then sees some number of identifiers. Notably, because we have:
Screen Shot 2022-07-19 at 10 43 17 PM

in the case of an `import`, it must be at least one.

This means that something like `import X` is not actually syntactically permissible Scala. This is why, currently, `import $X` fails. But this is still something that would be nice to be able to do, so I went ahead and changed things so that we can support it.

Essentially, I've made it such that `import_expr`s are either what they normally are understood to be, _or_ they can be a metavariable. This makes it so that we don't parse things like `import x`, but we can parse `import $X`.

The previous code also permitted non-dots following an identifier, and I don't think that should actually be possible.

Fixes #5219.

**Test plan:**
So, ideally I'd like to write a ton of tests showing that things which do not conform to the Scala grammar do not parse. Unfortunately, the parsing tests seem to be more about what things _do_ parse, so I don't actually know how to do this. Feedback appreciated.

In the meantime, the test plan can be to try testing dumping patterns on the following `.sgrep` patterns. I've verified the following using `sc -dump_pattern test.sgrep -lang scala`:
| example | status |
|:-|:-|
| `import this.a` | FAILS |
| `import this.a._` | SUCCEEDS |
| `import $X` | SUCCEEDS |
| `import $X._` | SUCCEEDS |
| `import a.$X._` | SUCCEEDS |
| `import a` | FAILS |
| `import a.this.b` | FAILS |
| `import a.this.b._` | SUCCEEDS |

Not sure how to carve this in stone, so to speak, however. Let me know what you think.

PR checklist:

### DIFF
--- a/changelog.d/gh-5219.fixed
+++ b/changelog.d/gh-5219.fixed
@@ -1,0 +1,1 @@
+Scala: Fixed erroneous parsing for `import` in Semgrep patterns.
--- a/changelog.d/gh-5219.fixed
+++ b/changelog.d/gh-5219.fixed
@@ -1,1 +1,1 @@
-Scala: Fixed erroneous parsing for `import` in Semgrep patterns.
+Scala: Allow metavariables in `import` patterns
--- a/semgrep-core/src/parsing/pfff/scala_to_generic.ml
+++ b/semgrep-core/src/parsing/pfff/scala_to_generic.ml
@@ -121,10 +121,19 @@ let v_dotted_name_of_stable_id (v1, v2) =
   let id = id_of_simple_ref v1 in
   id :: v2
 
-let rec v_import_expr tk (v1, v2) =
-  let module_name = G.DottedName (v_dotted_name_of_stable_id v1) in
-  let v2 = v_import_spec v2 in
-  v2 tk module_name
+let rec v_import_expr tk import_expr =
+  match import_expr with
+  | Left id ->
+      [
+        {
+          G.d = G.ImportAll (tk, FileName id, Parse_info.fake_info tk "");
+          d_attrs = [];
+        };
+      ]
+  | Right (v1, v2) ->
+      let module_name = G.DottedName (v_dotted_name_of_stable_id v1) in
+      let v2 = v_import_spec v2 in
+      v2 tk module_name
 
 and v_import_spec = function
   | ImportId v1 ->
--- a/semgrep-core/src/parsing/pfff/scala_to_generic.ml
+++ b/semgrep-core/src/parsing/pfff/scala_to_generic.ml
@@ -124,12 +124,7 @@ let v_dotted_name_of_stable_id (v1, v2) =
 let rec v_import_expr tk import_expr =
   match import_expr with
   | Left id ->
-      [
-        {
-          G.d = G.ImportAll (tk, FileName id, Parse_info.fake_info tk "");
-          d_attrs = [];
-        };
-      ]
+      [ { G.d = G.ImportFrom (tk, DottedName [], id, None); d_attrs = [] } ]
   | Right (v1, v2) ->
       let module_name = G.DottedName (v_dotted_name_of_stable_id v1) in
       let v2 = v_import_spec v2 in
--- a/semgrep-core/tests/scala/import_metavariable.scala
+++ b/semgrep-core/tests/scala/import_metavariable.scala
@@ -1,2 +1,11 @@
 // MATCH:
 import A.B, C.D
+
+// MATCH:
+import this.x.foo
+
+// MATCH:
+import a.this.b.c
+
+// MATCH:
+import x.y.z
--- a/semgrep-core/tests/scala/import_metavariable.scala
+++ b/semgrep-core/tests/scala/import_metavariable.scala
@@ -1,0 +1,1 @@
+import A.B, C.D
--- a/semgrep-core/tests/scala/import_metavariable.scala
+++ b/semgrep-core/tests/scala/import_metavariable.scala
@@ -1,1 +1,2 @@
+// MATCH:
 import A.B, C.D
--- a/semgrep-core/tests/scala/import_metavariable.sgrep
+++ b/semgrep-core/tests/scala/import_metavariable.sgrep
@@ -1,0 +1,1 @@
+import $X


changelog.d/gh-5219.fixed
```diff
@@ -0,0 +1 @@
+Scala: Fixed erroneous parsing for `import` in Semgrep patterns.
```
nmote: This is pretty vague. Maybe something like "Allow metavariables in `import` patterns?"

semgrep-core/src/parsing/pfff/scala_to_generic.ml
```diff
@@ -121,10 +121,19 @@ let v_dotted_name_of_stable_id (v1, v2) =
...
-let rec v_import_expr tk (v1, v2) =
-  let module_name = G.DottedName (v_dotted_name_of_stable_id v1) in
-  let v2 = v_import_spec v2 in
-  v2 tk module_name
+let rec v_import_expr tk import_expr =
+  match import_expr with
+  | Left id ->
+      [
+        {
+          G.d = G.ImportAll (tk, FileName id, Parse_info.fake_info tk "");
```
nmote: I think this should probably be `DottedName` not `FileName`. The case below always uses `DottedName` and I bet the discrepancy here will lead to a failure to match. Sure, we can now parse `import $X` as a pattern, but that doesn't help much if it never matches any source. We can have a `DottedIdent` that contains only a single identifier.

semgrep-core/tests/scala/import_metavariable.scala
```diff
@@ -0,0 +1 @@
+import A.B, C.D
```
nmote: Related to my comment above, I think this test would fail if Semgrep actually matched on this line. If you expect a match you need to precede it with `// MATCH:` or similar.

nmote (changes requested):

semgrep-core/src/parsing/pfff/scala_to_generic.ml
nmote: You might also run into issues with using `ImportAll` exclusively here, since `v_import_spec` chooses between `ImportFrom` and `ImportAll` when converting to the generic AST...

semgrep-core/src/parsing/pfff/scala_to_generic.ml
```diff
@@ -121,10 +121,14 @@ let v_dotted_name_of_stable_id (v1, v2) =
...
   id :: v2
 
-let rec v_import_expr tk (v1, v2) =
-  let module_name = G.DottedName (v_dotted_name_of_stable_id v1) in
-  let v2 = v_import_spec v2 in
-  v2 tk module_name
+let rec v_import_expr tk import_expr =
+  match import_expr with
+  | Left id ->
+      [ { G.d = G.ImportFrom (tk, DottedName [], id, None); d_attrs = [] } ]
```
nmote, line 127: Hmm, there's an unenforced invariant that `dotted_ident`s have at least one element. Using `ImportAll` instead of `ImportFrom` wouldn't require you to create a malformed `dotted_ident`. I assume you tried this and it didn't match, though?

Take a look at `Normalize_generic.ml` and `m_directive` in `Generic_vs_generic.ml`. See if you can structure this in such a way that:
1. No malformed dotted_idents are created
2. `import $X` matches various imports that Scala allows (and add tests)

You might need to tweak some of the matching code to make this happen. I haven't looked into this very closely, but I'm happy to help talk things over if you would like.

nmote: Actually, it looks like `ImportAll` does not get normalized, and there is a comment that it should not get normalized. I'm not sure what the best way forward here is. Maybe look to see what is done for Java, which is very similar?

brandonspark: It seems to be consistent with what Java does:
```java
import $X;
```
translates to
`Pr([DirectiveStmt({d_attrs=[]; d=ImportFrom((), DottedName([]), ("$c", ()), None); })])`

So it looks like `DottedName`s can be empty anyways.

nmote (approved):
Okay, well since there's precedent I think this is fine for now. I guess the comment next to `dotted_ident` is more aspirational than an actual invariant :stuck_out_tongue:

I still think it would be nice to see some tests that illustrate exactly which import constructs this can actually match. You could just add a few variations to `import_metavariable.scala`.

PR #5769, opened 2022-07-20
metavariable-comparison: Allow comparing any AST node as a string

IagoAbal:
And add basic support for Python's  `str`.

Closes PA-1659

test plan:
make test # added one test

PR checklist:

### DIFF
--- a/changelog.d/pa-1659.changed
+++ b/changelog.d/pa-1659.changed
@@ -1,6 +1,6 @@
 `metavariable-comparison` now also works on metavariables that cannot be evaluated
 to simple literals. In such cases, we take the string representation of the code
-bound by the metariable. The way to access this string representation is via
+bound by the metavariable. The way to access this string representation is via
 `str($MVAR)`. For example:
 
 ```
--- a/changelog.d/pa-1659.changed
+++ b/changelog.d/pa-1659.changed
@@ -1,0 +1,14 @@
+`metavariable-comparison` now also works on metavariables that cannot be evaluated
+to simple literals. In such cases, we take the string representation of the code
+bound by the metariable. The way to access this string representation is via
+`str($MVAR)`. For example:
+
+```
+- metavariable-comparison:
+    metavariable: $X
+    comparison: str($X) == str($Y)
+```
+
+Here `$X` and `$Y` may bind to two different code variables, and we check whether
+these two code variables have the same name (e.g. two different variables but both
+named `x`).
--- a/semgrep-core/src/engine/Eval_generic.ml
+++ b/semgrep-core/src/engine/Eval_generic.ml
@@ -137,6 +137,7 @@ let value_of_lit ~code x =
   (* big integers or floats can't be evaluated (Int (None, ...)) *)
   | G.Int (Some i, _t) -> Int i
   | G.Float (Some f, _t) -> Float f
+  | G.Null _ -> Int 0
   | _ -> raise (NotHandled code)
 
 let rec eval env code =
@@ -188,6 +189,10 @@ let rec eval env code =
   | G.Container (G.List, (_, xs, _)) ->
       let vs = Common.map (eval env) xs in
       List vs
+  (* Emulate Python str just enough *)
+  | G.Call ({ e = G.N (G.Id (("str", _), _)); _ }, (_, [ G.Arg e ], _)) ->
+      let v = eval env e in
+      eval_str env ~code v
   (* Emulate Python re.match just enough *)
   | G.Call
       ( {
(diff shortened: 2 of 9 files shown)


changelog.d/pa-1659.changed
```diff
@@ -0,0 +1,14 @@
+`metavariable-comparison` now also works on metavariables that cannot be evaluated
+to simple literals. In such cases, we take the string representation of the code
+bound by the metariable. The way to access this string representation is via
```
aryx: typo, metavariable.

changelog.d/pa-1659.changed
```diff
@@ -0,0 +1,14 @@
...
+`str($MVAR)`. For example:
+
+```
+- metavariable-comparison:
+    metavariable: $X
```
aryx, line 8: it's a bit weird/useless to have to specify metavariable:. Why $X and not $Y?
I think it was done back in the day by Matt Schwager but I think we could just remove it (or make
it optional to remain backward compatible).

semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -137,6 +137,7 @@ let value_of_lit ~code x =
   (* big integers or floats can't be evaluated (Int (None, ...)) *)
   | G.Int (Some i, _t) -> Int i
   | G.Float (Some f, _t) -> Float f
+  | G.Null _ -> Int 0
```
aryx: ? why this? why not add Null to the value?

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -364,15 +364,21 @@ let rec filter_ranges env xs cond =
...
+                         |> G.e);
                      ],
                      fk ) )
                |> G.e
              in
 
              let env =
                if const_prop && (fst env.config).constant_propagation then
-                 Eval_generic.bindings_to_env_just_strings_const_prop bindings
+                 Eval_generic.bindings_to_env (fst env.config) bindings
```
aryx, line 384: nice way to reuse str here so we can get rid of bindings_to_env_just_strings_const_prop!

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -364,15 +364,21 @@ let rec filter_ranges env xs cond =
                    ( fk,
                      [
                        G.Arg (G.L (G.String (re_str, fk)) |> G.e);
-                       G.Arg (G.N (G.Id ((mvar, fk), fki)) |> G.e);
+                       G.Arg
```
aryx: maybe add a comment before 'let e' = ...' giving the same of the code in concrete syntax, so 
re.match(str(...),...)

aryx: It's hard to infer back from the AST what is the actual code we try to generate for Eval_generic.

aryx (approved):

semgrep-core/src/engine/Eval_generic.ml
IagoAbal: We have this rule in the registry:
```yaml
rules:
  - id: c-string-equality
    patterns:
      - pattern: (char *$X) == (char *$Y)
      - metavariable-comparison:
          metavariable: $X
          comparison: $X != 0
      - metavariable-comparison:
          metavariable: $Y
          comparison: $Y != 0
    message: Using == on char* performs pointer comparison, use strcmp instead
    fix: strcmp($X, $Y) == 0
    languages: [c]
    severity: ERROR
```
and without this it was introducing a regression here:
```c
    // ok:c-string-equality
    if (NULL == s) {
      return 1;
    }
```
Because when `NULL` was not evaluated the `metavariable-comparison` just reduced to `false`, but evaluating `NULL` as `"NULL"` would make `$X != 0` reduce to `true`.

Since `NULL` is actually defined as `(void*)0` I figured out this was an OK solution. Although maybe that should be only when `-lang c`.

If we wanted to restore the exact same behavior here we could, basically comparing `AST _` with anything directly could be reduced to `false`, you would have to use `str()` to do any comparison.

IagoAbal: I did this instead:
```ocaml
and eval_op op values code =
  match (op, values) with
  | _op, [ AST _; _]
  | _op, [ _; AST _] ->
      (* To compare `AST` values one needs to explicitly use the `str()` function!
       * Otherwise we would introduce regressions. *)
      raise (NotHandled code)
```

changelog.d/pa-1659.changed
IagoAbal: Done!

PR #5775, opened 2022-07-21
fix(taint): Top level envs into functions

brandonspark:
**What**:
Currently, we have an issue where top-level declarations are not taken into account in the control flow graph, when doing the dataflow analysis for tainting. 

**Why:**
This is a pretty nice thing to be able to do, because it allows us to analyze examples like the following:
https://semgrep.dev/playground/s/brandon-wu:test_xss-copy

This stems from the code in `Match_tainting_mode.ml` doing two analyses - one which traverses each function definition (with an empty environment), and one which walks through the top-level of the program. Notably, each function body analysis is divorced from the code that came before it.

**How:**
I took the code which does an empty analysis on a function, and made it an argument to the transfer function in `Dataflow_tainting.ml`. This allows the analysis of a function node in the control flow graph, using the already-computed in-set from the dataflow analysis. Since this has no effect on the out-set, it simply exists to report any tainted patterns within the function body, and should not effect the actual values passed around in the analysis.

The reason why it is taken in as an argument (instead of just referenced as a variable) is because the function-checking function must take in the language that it is doing the analysis on, which is not already an input to the transfer function, as far as I can tell.

Making an extra argument to the transfer function is not an ideal solution, but it's the first one I thought of. Moreover, I'm not sure if a nicer one exists, because it seems like we would need to reify the function statement within the IL CFG in a better way, so as to leverage the existing transfer function (which should only ever do "one step" of the analysis, so to speak). However, this would require taking care of scoping (i.e., the computed out-set at the end of the function would need to be reset to whatever it was at the beginning of the function in the CFG).

Open to suggestions.

**Test plan:** TODO

**Fixes [PA-1656](https://linear.app/r2c/issue/PA-1656/js-taint-not-flowing-through-async-or-function)**

PR checklist:

### DIFF
--- a/changelog.d/gh-5219.fixed
+++ b/changelog.d/gh-5219.fixed
@@ -1,0 +1,1 @@
+Scala: Fixed erroneous parsing for `import` in Semgrep patterns.
--- a/changelog.d/pa-1656.fixed
+++ b/changelog.d/pa-1656.fixed
@@ -1,0 +1,14 @@
+
+Allowed tainted variables from top-level environments to spread into function bodies.
+
+For instance, a given the following example in Javascript:
+```js
+x = source
+
+function foo () {
+  sink(source)
+}
+```
+
+If we are looking for a `sink($SINK)`, before this Pr it would not be caught, because
+function bodies are all analyzed separately, with empty contexts.
--- a/semgrep-core/src/engine/Match_tainting_mode.ml
+++ b/semgrep-core/src/engine/Match_tainting_mode.ml
@@ -379,7 +379,7 @@ let pm_of_finding finding =
   | T.ArgToReturn _ ->
       None
 
-let check_fundef lang fun_env taint_config opt_ent fdef =
+let check_fundef lang init_env fun_env taint_config opt_ent fdef =
   let name =
     let* ent = opt_ent in
     let* name = AST_to_IL.name_of_entity ent in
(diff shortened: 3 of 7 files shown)


changelog.d/gh-5219.fixed
```diff
@@ -0,0 +1 @@
+Scala: Fixed erroneous parsing for `import` in Semgrep patterns.
```
IagoAbal, line 1: Is this a changelog entry that is missing from another PR or is it a mistake?

changelog.d/pa-1656.fixed
```diff
@@ -0,0 +1,14 @@
+
+Allowed tainted variables from top-level environments to spread into function bodies.
+
+For instance, a given the following example in Javascript:
```
IagoAbal, line 4:
```diff
- a
  given
```

IagoAbal:
> https://semgrep.dev/playground/s/brandon-wu:test_xss-copy

These playground links aren't immutable to they are not great for using in PR comments and such.

IagoAbal:
I think I made a mistake when assessing PA-1656. I filled that ticket quickly simply not to forget about it, and I just blamed it on AST-to-IL, but this is probably just another example for PA-223. You figured that out well.

This may not be a good first task actually, but you are still welcome to work on it.

Regarding the solution, I would be more fond of adding a global variable environment that is built by the visitor and passed down to each call to `check_fundef`. We could visit each global statement like `source = a`, and use the transfer function in `Dataflow_tainting` to tell us how a global statement transforms the environment. It's probably not a lot of work. Do you want to give it a try?

If I'm not mistaken this solution, while it should be valid, it will re-analyze every function in the file twice, and I think we should avoid that.

brandonspark:
Working on a better solution to this. See #5787

PR #5780, opened 2022-07-21
feat(fail-open): [CLI-254] Support fail-open in CI, default to OFF

zyannes:
PR checklist:

### DIFF
--- a/changelog.d/cli-254.added
+++ b/changelog.d/cli-254.added
@@ -1,1 +1,1 @@
-Support fail-open in CI (default to OFF)
+Support fail-open in CI: adds --suppress-errors/--no-suppress-errors (defaults to disabled)
--- a/changelog.d/cli-254.added
+++ b/changelog.d/cli-254.added
@@ -1,0 +1,1 @@
+Support fail-open in CI (default to OFF)
--- a/changelog.d/cli-254.added
+++ b/changelog.d/cli-254.added
@@ -1,1 +1,1 @@
-Support fail-open in CI: adds --suppress-errors/--no-suppress-errors (defaults to disabled)
+Support fail-open in CI: adds --suppress-errors/--no-suppress-errors (defaults to --no-suppress-errors)
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -3,7 +3,7 @@ import sys
 import time
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Iterable
+from typing import Any, Iterable
 from typing import Iterator
 from typing import List
 from typing import Optional
@@ -11,6 +11,8 @@ from typing import Sequence
 from typing import Tuple
 
 import click
+from click.shell_completion import CompletionItem
+from semgrep.fail_open import FailOpenState
 
 import semgrep.semgrep_main
 from semgrep.app import auth
@@ -43,6 +45,40 @@ ALWAYS_EXCLUDE_PATTERNS = [".semgrep/", ".semgrep_logs/"]
 
 # These patterns are excluded via --exclude unless the user provides their own .semgrepignore
 DEFAULT_EXCLUDE_PATTERNS = ["test/", "tests/", "*_test.go"]
+
+
+class FailOpenStateType(click.ParamType):
+    name = "fail_open_state"
+
+    def get_metavar(self, param: click.Parameter) -> str:
+        return "[on|off]"
+
+    def shell_complete(
+        self, context: click.Context, param: click.Parameter, incomplete: str
+    ) -> List[Any]:
+        return [
+            CompletionItem(e) for e in ["on", "off"] if e.startswith(incomplete)
+        ]
+
+    def convert(
+        self,
+        value: Any,
+        param: Optional["click.Parameter"],
+        ctx: Optional["click.Context"],
+    ) -> Any:
+        if value is None:
+            return None
+        if isinstance(value, str):
+            lower = value.lower()
+            # Support setting via old environment variable values 0/1/true/false
+            if lower == "on" or lower == "1" or lower == "true":
+                return FailOpenState.ON
+            if lower == "off" or lower == "0" or lower == "false":
+                return FailOpenState.OFF
+        self.fail("expected 'on', or 'off'")
+
+    
+FAIL_OPEN_STATE_TYPE = FailOpenStateType()
 
 
 def yield_valid_patterns(patterns: Iterable[str]) -> Iterable[str]:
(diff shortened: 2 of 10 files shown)


cli/src/semgrep/commands/ci.py
```diff
@@ -145,6 +181,19 @@ def fix_head_if_github_action(metadata: GitMeta) -> Iterator[None]:
...
+    type=FAIL_OPEN_STATE_TYPE,
+    help="""
+        Configures how scan status is sent to the fail-open server.
+        If 'on', fail-open are always sent.
+        If 'off', fail-open are disabled altogether and not sent.
+        If absent, the SEMGREP_SEND_FAIL_OPEN environment variable value will be used.
+        If no environment variable, defaults to 'off'.
+    """,
+    envvar="SEMGREP_SEND_FAIL_OPEN",
+)
```
underyx: @raghavjain3 I thought we decided on a less jargon-y name for this option. Did we reverse that decision?

cli/src/semgrep/commands/ci.py
```diff
@@ -145,6 +181,19 @@ def fix_head_if_github_action(metadata: GitMeta) -> Iterator[None]:
...
+@click.option(
+    "--fail-open",
+    "fail_open",
+    type=FAIL_OPEN_STATE_TYPE,
+    help="""
+        Configures how scan status is sent to the fail-open server.
+        If 'on', fail-open are always sent.
+        If 'off', fail-open are disabled altogether and not sent.
+        If absent, the SEMGREP_SEND_FAIL_OPEN environment variable value will be used.
+        If no environment variable, defaults to 'off'.
```
underyx: Keep in mind this text shows up when someone runs `semgrep ci --help` so we should reword this with that in mind. E.g. even as a semgrep maintainer I don't understand what a "fail-open server" is and what it means if "fail-open are always sent".

cli/src/semgrep/commands/ci.py
spencerdrak: I think we are going with `--no-suppress-errors` when we flip the default behavior, but not sure what we'd wanna call this.

cli/src/semgrep/fail_open.py
```diff
@@ -0,0 +1,87 @@
...
+class FailOpenState(Enum):
+    """
+    Configures metrics upload.
+
+    ON - Fail-open always sent
+    OFF - Fail-open never sent
+    """
+
+    ON = auto()
+    OFF = auto()
```
underyx: Could we just use a `bool`?

cli/src/semgrep/commands/ci.py
underyx: Reading through the code this looks like a "crash reporting server" and the setting is that "crash reports are always sent"?

cli/src/semgrep/commands/ci.py
zyannes: Yes. Renaming to `--no-suppress-errors` to match what is mentioned in the doc

cli/src/semgrep/fail_open.py
zyannes: Updated!

cli/src/semgrep/commands/ci.py
zyannes: Updated in latest commit

changelog.d/cli-254.added
```diff
@@ -0,0 +1 @@
+Support fail-open in CI: adds --suppress-errors/--no-suppress-errors (defaults to disabled)
```
ievans: What does "disabled" mean in this context?

ievans:
Will this globally catch all possible errors from semgrep-core and all Python exceptions? Just trying to understand the scope of failures captured by the flag.

cli/src/semgrep/commands/wrapper.py
```diff
@@ -50,6 +50,9 @@ def wrapper(*args: Any, **kwargs: Any) -> NoReturn:
             metrics.add_exit_code(exit_code)
             metrics.send()
 
+            error_handler = get_state().error_handler
```
spencerdrak, line 53: Will this still fire even if something goes wrong with the metrics handling?

cli/src/semgrep/env.py
```diff
@@ -50,6 +50,13 @@ class Env:
     between multiple invocations.
     """
 
+    fail_open_url: str = field(
+        default=EnvFactory(
+            ["FAIL_OPEN_URL", "SEMGREP_FAIL_OPEN_URL"],
+            "https://fail-open.semgrep.dev/failure",
```
spencerdrak: Either we need to update the TF module or this needs to be configured to `https://fail-open.prod.semgrep.dev` since the TF module always adds the env name.

spencerdrak (dismissed):

cli/src/semgrep/commands/wrapper.py
zyannes: Yes, it will always fire. The metrics handler has its own `suppress_errors` function wrapper that silences any errors.

cli/src/semgrep/env.py
zyannes: Updating to `https://fail-open.prod.semgrep.dev/failure`

changelog.d/cli-254.added
zyannes: "disabled" in this context means that errors are not suppressed by default. Rephrasing to 
```
Support fail-open in CI: adds --suppress-errors/--no-suppress-errors (defaults to --no-suppress-errors)
```

zyannes:
> Will this globally catch all possible errors from semgrep-core and all Python exceptions? Just trying to understand the scope of failures captured by the flag.

@ievans yes, as long as `--suppress-errors` is set, it will catch any exceptions and forward to the fail-open endpoint. The intention here (at least initially) is to publish any "semgrep CI crash" to the fail-open endpoint so we can have visibility into errors that would normally block CI scans if `--suppress-errors` was not set.

cli/src/semgrep/env.py
```diff
@@ -50,6 +50,13 @@ class Env:
     between multiple invocations.
     """
 
+    fail_open_url: str = field(
+        default=EnvFactory(
+            ["FAIL_OPEN_URL", "SEMGREP_FAIL_OPEN_URL"],
```
brendongo: Do we need the non semgrep prefixed envvar? Also the default url seems ominous :D fail-open and failure

cli/src/semgrep/error_handler.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Returns whether scan status should be sent.
+        """
+        return self.suppress_errors
+
+    def send(self, exit_code: int) -> int:
+        """
+        Send scan status to the fail-open server.
+
+        Will if is_enabled is True
```
brendongo: Missing info on what this will do if enabled :D

brendongo (approved):

cli/src/semgrep/env.py
zyannes: No, removing in next commit!

cli/src/semgrep/error_handler.py
zyannes: Oops, copy/pasted from metrics.py, which is also lacking a description. Updating docstring in latest commit

brendongo (approved):

PR #5785, opened 2022-07-21
fix(const prop): toplevel const prop

brandonspark:
**What:**
Currently, constant propagation at the top level in a program is not behaving as intended.

**Why:**
We should fix this because it makes examples like [this](https://semgrep.dev/s/A9QL) not work properly, but permits examples like [this](https://semgrep.dev/s/D9Qd).

**How:**
Firstly, we notice that the second link above works properly, meaning that dataflow is being done properly for function definitions. Within the code of `Constant_propagation.ml`, this is done via the `propagate_dataflow_one_function` function, which is only called on function definitions.

So this fix just regards the entire top-level AST as the inside of a function, and then runs the working `propagate_dataflow_one_function` on it. I don't believe this should repeat work, because the overall dataflow engine doesn't know about the inside of functions.

I first tried running `propagate_basic` on the entire AST, which looks promising, but it didn't work. I'm not entirely sure why.

**Test plan:** `make test`

**Fixes [PA-223](https://linear.app/r2c/issue/PA-223/constant-propagation-only-runs-inside-functions)**.

PR checklist:

### DIFF
--- a/semgrep-core/src/analyzing/Constant_propagation.ml
+++ b/semgrep-core/src/analyzing/Constant_propagation.ml
@@ -688,6 +688,8 @@ let propagate_dataflow lang ast =
       in
       (* Hack to do dataflow on the entire AST.
          Just think of it as a function body for a function which takes no arguments.
+         
+         Shouldn't repeat work, because the overall dataflow engine doesn't know about the inside of functions. (To be changed?)
       *)
       let stmts = List.concat_map (AST_to_IL.stmt lang) ast in
       let prog_flow = CFG_build.cfg_of_stmts stmts in
--- a/semgrep-core/src/analyzing/Constant_propagation.ml
+++ b/semgrep-core/src/analyzing/Constant_propagation.ml
@@ -686,4 +686,10 @@ let propagate_dataflow lang ast =
                 propagate_dataflow_one_function lang inputs flow);
           }
       in
+      (* Hack to do dataflow on the entire AST.
+         Just think of it as a function body for a function which takes no arguments.
+      *)
+      let stmts = List.concat_map (AST_to_IL.stmt lang) ast in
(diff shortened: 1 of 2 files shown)


emjin:
Re the failing tests: you can look at the commit, but if you ever want to run the python tests yourself, follow instructions here: https://semgrep.dev/docs/contributing/semgrep-contributing/#testing

semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -686,4 +686,12 @@ let propagate_dataflow lang ast =
                 propagate_dataflow_one_function lang inputs flow);
           }
       in
+      (* Hack to do dataflow on the entire AST.
+         Just think of it as a function body for a function which takes no arguments.
+         
+         Shouldn't repeat work, because the overall dataflow engine doesn't know about the inside of functions. (To be changed?)
+      *)
+      let stmts = List.concat_map (AST_to_IL.stmt lang) ast in
```
IagoAbal, line 694: Isn't this done in a better way with your new proposal?

~Here it seems that if we had:~

```
x = 1;
function f() {
  x = 2;
}
foo(x);
function g() {
  x = 3;
}
bar(x);
```
~this would be interpreted as~
```
x = 1;
x = 2;
foo(x); # would incorrectly infer that this is `foo(2)`
x = 3;
bar(x); # would incorrectly infer that this is `bar(3)`
```
~am I correct?~

IagoAbal: Sorry, I was wrong, indeed this should work. I got confused by the `List.concat_map`, but this should be equivalent to `AST_to_IL.stmt lang (G.stmt1 ast)`. However I still think your new proposal would cover this already, or doesn't it?

brandonspark:
Superseded by #5787.

PR #5787, opened 2022-07-22
feat(dataflow): reify complex constructs in dataflow engine

brandonspark:
**What:**
Currently, our set-up for dataflow has a "second-class" notion of functions, classes, and modules (complex constructs which may contain smaller statements). That is to say, when an AST is translated to the IL, each of these constructs is contained within its own (un-translated) `MiscStmt` node. This is preserved in a `NOther` node within the CFG, when converting the list of IL stmts to a CFG, which notably does not interact interestingly with the dataflow's transfer function.

The reason why this is problematic is because the dataflow engine only knows about things at the granularity of nodes. The transfer function can only be applied on a per-node basis, in-sets and out-sets can only be computed at the granularity of nodes, so on and so forth. This means that, if there is a "function node" in the CFG, the dataflow engine doesn't know that it contains a "list of statements", in intention. It just knows that it needs to apply the transfer function once.

----------------------------------

**Why:**
This produces a lot of annoying code, such as `Constant_propagation.ml:662` and `Match_tainting_mode.ml:383`, where we need to do separate analyses for the top-level and each function. It tends to produce a distinction between our analyses on functions and the rest of the program. This distinction can lead to examples like [this one](https://semgrep.dev/s/PK8N), where the analysis works when the code is contained within a function, but not when it's at the top level, which is a distinction that need not exist. Moreover, we can fail use-cases which require information from the top-level to be brought into classes or modules: https://semgrep.dev/s/4KPg.

While this approach still has a distinction between functions and other parts of the program, it is done "automatically" via the dataflow engine. This means that we do not need to use the visitor AST to recurse into the program, but can instead trust dataflow to do it instead.

----------------------------------

**How:**
I propose a refinement of the general dataflow engine, which is specialized to work on IL CFGs. This dataflow is located in `Dataflow_prog.ml`, and operates on CFGs whose nodes are either normal IL nodes, as understood prior to this PR, but also special nodes for functions, classes, and modules. These nodes are special in that they contain the CFG for the body of the construct that it describes.

This enables the new dataflow engine to do its normal analysis throughout the master CFG, but as soon as it reaches a special node, it does the subsidiary analysis on the smaller CFG, using a function which modifies the environment based on the kind of node. Since these analyses are "local" to the scope of the construct, the in-set and out-set are maintained to be just the entering environment, so as to not allow escape, for instance of tainted variables within the scope of a class.

Code was shifted around elsewhere to accommodate this.

----------------------------------

**Important Changes:**
Some important things were changed in this PR. Most notably:
- CFG building works differently, and now there are special nodes for functions, classes, and modules.
- There may be edges from the start node to unreachable function nodes.
- CFG building requires the building of multiple CFGs. Dataflow analysis using `Dataflow_prog` may involve dataflow on multiple CFGs.
- There is a new file `Dataflow_prog.ml`.
- `Dataflow_svalue` and `Dataflow_tainting` now make use of this new dataflow procedure, and they no longer use `Visitor_AST`.

----------------------------------

**Test plan:** `make test`

----------------------------------

**Solves:** [PR-223](https://linear.app/r2c/issue/PA-223/constant-propagation-only-runs-inside-functions)

----------------------------------

PR checklist:

### DIFF
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -594,8 +594,8 @@ and expr_aux env ?(void = false) e_gen =
       (* TODO: we should have a use def.f_tok *)
       let tok = G.fake "lambda" in
       let lval = fresh_lval env tok in
-      let fdef = function_definition_helper env fdef in
-      add_instr env (mk_i (AssignAnon (lval, Lambda fdef)) eorig);
+      let fdef2 = function_definition_helper env fdef in
+      add_instr env (mk_i (AssignAnon (lval, Lambda (fdef, fdef2))) eorig);
       mk_e (Fetch lval) eorig
   | G.AnonClass def ->
       (* TODO: should use def.ckind *)
@@ -1011,6 +1011,12 @@ and mk_switch_break_label env tok =
   in
   (break_label, [ mk_s (Label break_label) ], switch_env)
 
+(* TODO: non module structs *)
+and module_stmt env = function
+  | G.ModuleAlias _ -> []
+  | ModuleStruct (_, stmts) -> List.concat_map (stmt env) stmts
+  | OtherModule _ -> []
+
 and stmt_aux env st =
   match st.G.s with
   | G.ExprStmt (eorig, tok) ->
@@ -1029,6 +1035,9 @@ and stmt_aux env st =
   | G.DefStmt (_, G.ClassDef { cbody = _, fields, _; _ }) ->
       (* TODO: deal with class parameters *)
       [ mk_s (ClassStmt (List.concat_map (fun (G.F s) -> stmt env s) fields)) ]
+  | G.DefStmt (_, G.ModuleDef m) ->
+      (* TODO: deal with class parameters *)
+      [ mk_s (ModuleStmt (module_stmt env m.mbody)) ]
   | G.DefStmt def -> [ mk_s (MiscStmt (DefStmt def)) ]
   | G.DirectiveStmt dir -> [ mk_s (MiscStmt (DirectiveStmt dir)) ]
   | G.Block xs -> xs |> G.unbracket |> Common.map (stmt env) |> List.flatten
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -1036,7 +1036,6 @@ and stmt_aux env st =
       (* TODO: deal with class parameters *)
       [ mk_s (ClassStmt (List.concat_map (fun (G.F s) -> stmt env s) fields)) ]
   | G.DefStmt (_, G.ModuleDef m) ->
-      (* TODO: deal with class parameters *)
       [ mk_s (ModuleStmt (module_stmt env m.mbody)) ]
   | G.DefStmt def -> [ mk_s (MiscStmt (DefStmt def)) ]
   | G.DirectiveStmt dir -> [ mk_s (MiscStmt (DirectiveStmt dir)) ]
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -594,7 +594,7 @@ and expr_aux env ?(void = false) e_gen =
       (* TODO: we should have a use def.f_tok *)
       let tok = G.fake "lambda" in
       let lval = fresh_lval env tok in
-      let fdef = function_definition env fdef in
+      let fdef = function_definition_helper env fdef in
       add_instr env (mk_i (AssignAnon (lval, Lambda fdef)) eorig);
       mk_e (Fetch lval) eorig
   | G.AnonClass def ->
@@ -1023,6 +1023,12 @@ and stmt_aux env st =
       let ss, e' = expr_with_pre_stmts env e in
       let lv = lval_of_ent env ent in
       ss @ [ mk_s (Instr (mk_i (Assign (lv, e')) (Related (G.S st)))) ]
+  | G.DefStmt (ent, G.FuncDef fdef) ->
+      let _, body = function_definition env.lang fdef in
+      [ mk_s (FuncStmt { ent; fdef; body }) ]
+  | G.DefStmt (_, G.ClassDef { cbody = _, fields, _; _ }) ->
+      (* TODO: deal with class parameters *)
+      [ mk_s (ClassStmt (List.concat_map (fun (G.F s) -> stmt env s) fields)) ]
   | G.DefStmt def -> [ mk_s (MiscStmt (DefStmt def)) ]
   | G.DirectiveStmt dir -> [ mk_s (MiscStmt (DirectiveStmt dir)) ]
   | G.Block xs -> xs |> G.unbracket |> Common.map (stmt env) |> List.flatten
@@ -1462,7 +1468,7 @@ and python_with_stmt env manager opt_pat body =
 (* Defs *)
 (*****************************************************************************)
 
-and function_definition env fdef =
+and function_definition_helper env fdef =
   let fparams = parameters env fdef.G.fparams in
   let fbody = function_body env fdef.G.fbody in
   { fparams; frettype = fdef.G.frettype; fbody }
@@ -1471,7 +1477,7 @@ and function_definition env fdef =
 (* Entry points *)
 (*****************************************************************************)
 
-let function_definition lang def =
+and function_definition lang def =
   let env = empty_env lang in
   let params = parameters env def.G.fparams in
   let body = function_body env def.G.fbody in
--- a/semgrep-core/src/analyzing/CFG_build.ml
+++ b/semgrep-core/src/analyzing/CFG_build.ml
@@ -267,10 +267,10 @@ let rec cfg_stmt : Lang.t -> state -> F.nodei option -> stmt -> cfg_stmt_result
   (* Any DefStmts which are FuncDefs are reified as proper Func nodes
      within the CFG, which contain their own smaller CFGs.
   *)
-  | MiscStmt (DefStmt (_, G.FuncDef fdef)) ->
+  | MiscStmt (DefStmt (ent, G.FuncDef fdef)) ->
       let _, stmts = AST_to_IL.function_definition lang fdef in
       let cfg = cfg_of_stmts lang stmts in
-      let newi = state.g#add_node { F.n = Func { fdef; cfg } } in
+      let newi = state.g#add_node { F.n = Func { fdef; cfg; ent } } in
       state.g |> add_arc_from_opt (previ, newi);
       CfgFirstLast (newi, Some newi)
   | MiscStmt x ->
--- a/semgrep-core/src/analyzing/CFG_build.ml
+++ b/semgrep-core/src/analyzing/CFG_build.ml
@@ -270,6 +270,9 @@ let rec cfg_stmt : Lang.t -> state -> F.nodei option -> stmt -> cfg_stmt_result
   | FuncStmt { fdef; ent; body } -> (
       let cfg = cfg_of_stmts lang body in
       let newi = state.g#add_node { F.n = NFunc { fdef; cfg; ent } } in
+      (* If there is no previous node, we add an edge from the entrance node.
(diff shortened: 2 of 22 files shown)


semgrep-core/src/analyzing/CFG_build.ml
```diff
@@ -230,51 +233,66 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
+      let newi = state.g#add_node { F.n = Reg (F.NThrow (tok, e)) } in
       state.g |> add_arc_from_opt (previ, newi);
       (match state.try_catches_opt with
       | None -> state.g |> add_arc (newi, state.exiti)
       | Some catchesi -> state.g |> add_arc (newi, catchesi));
       CfgFirstLast (newi, None)
+  (* Any DefStmts which are FuncDefs are reified as proper Func nodes
+     within the CFG, which contain their own smaller CFGs.
+  *)
+  | MiscStmt (DefStmt (_, G.FuncDef fdef)) ->
```
IagoAbal: If we are going to build the CFG here, then I think it's better to change the IL so that we add a `DefStmt` to `IL.stmt_kind`. Then AST_to_IL would do the translation.

semgrep-core/src/core/il/IL.ml
```diff
@@ -351,10 +356,15 @@ and node_kind =
   | NTodo of stmt
 [@@deriving show { with_path = false }]
 
+and augmented_node_kind =
```
IagoAbal: What do we gain by having an `augmented_node_kind`? Why not just adding `Func` to `node_kind`?

IagoAbal:
Nice proposal! I like the overall approach of having a `Dataflow_prog` that is specialized to do a file-level analysis that takes globals into consideration, while still being intra-procedural. BTW, what to do with class declarations? Should we consider them already?

Initially, I was thinking of using a "fold" over the list of statements in an IL program, but this may not be sufficient since there could be nested function definitions, and (nested) class definitions. Your solution is recursive so it deals with this issue pretty well.

However, as I mentioned in the inline comments, I am less fond of doing the AST-to-IL translation for functions during the CFG construction. I think that should be done earlier in the `AST_to_IL` module. Also I'm not sure about the benefits of introducing `augmented_node_kind`. That said, these are details so my comments there do not invalidate the overall strategy.

In addition, I think we should be very careful with the assumptions that we make based on global statements. In languages where variables are mutable references, the fact that there is a statement `x = "bar"` right before a function declaration `function foo() { ... }`, it does _not_ mean that we can assume that `x` is equals to `"bar"` inside `foo`'s body!

semgrep-core/src/analyzing/Dataflow_prog.ml
```diff
@@ -0,0 +1,64 @@
...
+      'a DC.mapping =
+   fun ~eq ~init ~trans ~flow ~get_input_env ~forward ->
+    ProgDataflow.fixpoint ~eq ~init
+      ~trans:(fun mapping ni ->
+        let { IL.n } = flow.graph#nodes#assoc ni in
+        match n with
+        | IL.Reg _ -> trans mapping ni
+        | Func { cfg = flow; _ } ->
+            (* TODO: change mapping here depending on the function inputs *)
+            fixpoint ~eq ~init:mapping ~trans ~flow ~get_input_env ~forward
```
IagoAbal: We may need to have a function as input that is called on the `mapping`, just in case the client of `Dataflow_prog.fixpoint` wants to do something with the `mapping` of each function analyzed? This would be the case of `Dataflow_svalue` right now. Although perhaps we could change `Dataflow_svalue.transfer` so that it updates the `id_svalue` in-place.

semgrep-core/src/analyzing/CFG_build.ml
```diff
@@ -44,6 +44,8 @@ type label_key = string * G.sid
  *)
 type state = {
   g : (F.node, F.edge) Ograph_extended.ograph_mutable;
+  (* We keep this so we can connect to unreachable nodes. *)
```
aryx, line 48: Can you give an example of why we need enteri?
(like we did for exiti with 'return' as an example).

semgrep-core/src/analyzing/CFG_build.ml
```diff
@@ -127,8 +129,9 @@ type cfg_stmt_result =
    *)
   | CfgFirstLast of F.nodei * F.nodei option
 
-let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
- fun state previ stmt ->
+let rec cfg_stmt : Lang.t -> state -> F.nodei option -> stmt -> cfg_stmt_result
```
aryx: Why not add Lang.t in state (which we kinda use like an environment)?
So we don't need to thread it all around (we already thread state all around).

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -1462,7 +1476,7 @@ and python_with_stmt env manager opt_pat body =
 (* Defs *)
 (*****************************************************************************)
 
-and function_definition env fdef =
+and function_definition_helper env fdef =
```
aryx, line 1479: thx for fixing that; I also don't like overwrite previous def with the same name

semgrep-core/src/analyzing/CFG_build.ml
```diff
@@ -255,6 +264,32 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
       | Some catchesi -> state.g |> add_arc (newi, catchesi));
       CfgFirstLast (newi, None)
+  (* Any DefStmts which are FuncDefs are reified as proper Func nodes
+     within the CFG, which contain their own smaller CFGs.
+  *)
+  | FuncStmt { fdef; ent; body } -> (
+      let cfg = cfg_of_stmts lang body in
+      let newi = state.g#add_node { F.n = NFunc { fdef; cfg; ent } } in
+      (* If there is no previous node, we add an edge from the entrance node.
+         Why? This is so that the dataflow can run on unreachable functions.
```
aryx, line 272: great comment. Might be useful to copy it next to enteri, so better understand the need for enteri.

semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -672,18 +672,9 @@ let propagate_dataflow lang ast =
...
-              (fun (_k, _) def ->
-                let inputs, xs = AST_to_IL.function_definition lang def in
-                let flow = CFG_build.cfg_of_stmts xs in
-                propagate_dataflow_one_function lang inputs flow);
-          }
-      in
-      v (Pr ast)
+      let xs = AST_to_IL.stmt lang (G.stmt1 ast) in
+      let flow = CFG_build.cfg_of_stmts lang xs in
+      propagate_dataflow_one_function lang [] flow
```
aryx, line 680: nice! Maybe rename propagate_dataflow_one_function to propagate_dataflow_program?

semgrep-core/src/core/il/IL.ml
```diff
@@ -275,7 +275,7 @@ and call_special =
    | StringAccess of string (* for records/hashes *)
 *)
 and anonymous_entity =
-  | Lambda of function_definition
+  | Lambda of G.function_definition * function_definition
```
aryx, line 278: Why do we need to keep the original G.function_definition here?

semgrep-core/src/core/il/IL.ml
```diff
@@ -302,6 +302,13 @@ and stmt_kind =
       * (name * stmt list) list (* catches *)
       * stmt list (* finally *)
   | Throw of tok * exp (* less: enforce lval here? *)
+  | FuncStmt of {
+      fdef : G.function_definition;
```
aryx, line 306: why not fdef: function_definition here?

aryx (commented):
This looks very nice to me and cleanup lots of hacks indeed.

semgrep-core/src/analyzing/CFG_build.ml
brandonspark: Suppose we have a program like the following:
```js
function foo () {
  return 2;
  
  function bar() {
    sink(source);
  }
}
```

It's my understanding that this is valid Javascript code.

If we try to analyse this program without using the enter node, then we have parts of the CFG which are unreachable by regular dataflow analysis. This means that taint analysis does not run on these functions. Right now, since Semgrep visits _all_ function nodes via `Visitor_AST`, this means we are losing functionality that Semgrep currently supports.

This easy fix (or hack) remedies that situation by simply making all nodes reachable in the CFG.

semgrep-core/src/analyzing/Constant_propagation.ml
brandonspark: With the way that it's being exposed for DeepSemgrep, I think it may be best to keep the name the same.

One side note is that both before and currently, `_inputs` is being ignored completely. I'm not sure if that's supposed to be used, or if DeemSemgrep is assuming that it is used...

semgrep-core/src/core/il/IL.ml
brandonspark: It's needed so that later on, when we enter into the scope of the lambda, we can properly add the arguments into the environment. The original code operated only on `G.function_definition`, and I didn't understand it well enough to adapt it, so I just copied it here. Is it that bad to carry both?

semgrep-core/src/core/il/IL.ml
brandonspark: This is as above, because we need to carry the `G.function_definition` down into the IL.

semgrep-core/src/analyzing/Constant_propagation.ml
IagoAbal: Inputs were used previously, but right now I think they could be dropped.

semgrep-core/tests/OTHER/rules/prop_into_class.java
```diff
@@ -0,0 +1,7 @@
+static String secret = "secret"
+
+public class App
+{
+    // ruleid: prop-into-class 
+    String foo = sink(secret)
```
IagoAbal, line 6: Where did this test come from? What are `static` variables declared outside classes? 🤔

semgrep-core/tests/OTHER/rules/const_prop_toplevel.py
```diff
@@ -0,0 +1,10 @@
+# ruleid: const_prop_toplevel
+x = "ls"
+
+foo = "ls"
+# ruleid: const_prop_toplevel
+x = foo
+
+foo = "ls"
+# ruleid: const_prop_toplevel
+x = foo
```
IagoAbal, line 10: Maybe we need to reconsider this, because this test does not justify all the code that is being changed here. In `Match_tainting_mode` we have this hack:
```
        let xs = AST_to_IL.stmt lang (G.stmt1 ast) in
        let flow = CFG_build.cfg_of_stmts xs in
        Dataflow_tainting.fixpoint ~fun_env taint_config flow |> ignore)
```
And I wonder if a similar hack could be used for constant propagation to achieve something very similar to what this PR currently achieves? Or am I wrong here?

semgrep-core/src/analyzing/CFG_build.ml
aryx: Could be nice to add some of this explanation in the code, as a comment.

semgrep-core/src/core/il/IL.ml
aryx: Maybe add a (* TODO? we probably should drop the G.function_definition and refactor the code elsewhere
to operate on a function_definition instead of G.function_definition *)

semgrep-core/src/core/il/IL.ml
aryx: My guess is that we didn't consider doing interprocedural in semgrep for a long time, so lots of stuff like you said
was hardcoded for handling one function, and so this is why probably we had functions taking G.function_definition
instead of function_definition, but after your PR we can probably refactor the code to operate more often on a function_definition so we can drop those G.function_definition.

IagoAbal:
I think a good target for this PR is to also replace `Constant_propagation.propagate_basic`. If we can get rid of that while maintaining the same features, then I think the changes would be justified. Then we would have a general mechanism that would work for both const-prop and taint.

PR #5802, opened 2022-07-25
feat(metrics): report ignored-by-language file extensions to metrics

bkettle:
Closes APP-1354. This PR adds new reporting of file extensions that do not match any enabled language except `generic`. File extension here is defined as `os.path.splitext` defines it:

> Split the pathname path into a pair (root, ext) such that root + ext == path, and the extension, ext, is empty or begins with a period and contains at most one period.

This does not report anything for files without an extension, including hidden files without an extensions such as `.semgrepignore`. 

For each file extension, it reports the number of files that were ignored with that extension.

PR checklist:

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -109,6 +109,7 @@ Semgrep reports data that indicate how useful a run is for the end user; e.g.
 - Number of ignored findings
 - Pseudoanonymized hashes of the rule definitions that yield findings
 - The [Semgrep features used](#feature-usage) during the scan
+- File extensions of unscanned files
 
 ### Pseudoanonymization
 
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -170,6 +170,7 @@ r2c will:
 |             | Rule hashes with findings               | Map of rule hashes to number of findings                               | Understand which rules are providing value to the user; diagnose high false-positive rates | `{"7c43c962dfdbc52882f80021e4d0ef2396e6a950867e81e5f61e68390ee9e166": 4}`                                                                                                             | Object         |
 |             | Total Findings                          | Count of all findings                                                  | Understand if rules are super noisy for the user                                           | 7                                                                                                                                                                                     | Number         |
 |             | Total Nosems                            | Count of all `nosem` annotations that tell semgrep to ignore a finding | Understand if rules are super noisy for the user                                           | 3                                                                                                                                                                                     | Number         |
+|             | Unsupported Extensions                  | List of file extensions that did not match language and number of each | Understand demand by language for new rules and language supprt                            | `{".txt": 3, ".erb": 2, ".java": 5}`                                                                                                                                                  |
 
 ### Anonymous user ID
 
@@ -269,6 +270,7 @@ This is a sample blob of the aggregate metrics described above:
         "numFindings": 7,
         "numIgnored": 3,
         "features": ["language/python", "option/deep", "option/no-git-ignore", "key/metavariable-comparison"]
+        "unsupportedExts": {".txt": 3, ".erb": 2, ".java": 5}
     }
 }
 ```
--- a/cli/src/semgrep/metrics.py
+++ b/cli/src/semgrep/metrics.py
@@ -101,15 +101,12 @@ class ValueSchema(ValueRequiredSchema, total=False):
     numFindings: int
     numIgnored: int
     ruleHashesWithFindings: Dict[str, int]
+    unsupportedExts: Dict[str, int]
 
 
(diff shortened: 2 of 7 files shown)


cli/src/semgrep/metrics.py
```diff
@@ -297,6 +303,17 @@ def add_targets(self, targets: Set[Path], profiling_data: ProfilingData) -> None
...
 
+    @suppress_errors
+    def add_ignores(self, ignored: Set[Path]) -> None:
+        ignored_ext_freqs: Dict[str, int] = {}
+        for path in ignored:
+            ext = os.path.splitext(path)[1]
+            if not ext:
+                continue
+            ignored_ext_freqs[ext] = ignored_ext_freqs.get(ext, 0) + 1
+        print("Ignored Extensions Metrics:", ignored_ext_freqs)
```
underyx: I assume this is just for debugging

cli/src/semgrep/metrics.py
```diff
@@ -297,6 +303,17 @@ def add_targets(self, targets: Set[Path], profiling_data: ProfilingData) -> None
         self.payload["performance"]["totalBytesScanned"] = total_bytes_scanned
         self.payload["performance"]["numTargets"] = len(targets)
 
+    @suppress_errors
+    def add_ignores(self, ignored: Set[Path]) -> None:
+        ignored_ext_freqs: Dict[str, int] = {}
```
underyx: might wanna use https://docs.python.org/3/library/collections.html#collections.Counter for simpler code

cli/src/semgrep/metrics.py
bkettle: ya oops, gone now

cli/src/semgrep/semgrep_main.py
```diff
@@ -441,6 +441,7 @@ def main(
     profiler.save("total_time", rule_start_time)
 
     metrics = get_state().metrics
+    metrics.add_ignores(target_manager.ignore_log.unsupported_lang_paths)
```
underyx: although the `if metrics.is_enabled:` thing doesn't make too much sense… couldn't we keep this next to all the other metrics.add calls anyway for the sake of consistency?

PRIVACY.md
```diff
@@ -109,6 +109,7 @@ Semgrep reports data that indicate how useful a run is for the end user; e.g.
 - Number of ignored findings
 - Pseudoanonymized hashes of the rule definitions that yield findings
 - The [Semgrep features used](#feature-usage) during the scan
+- File extensions of unscanned files
```
underyx, line 112: there's also a sample JSON blob code snippet, and a detailed breakdown table in this file; could you update both?

cli/src/semgrep/semgrep_main.py
bkettle: Oh shoot this was for debugging too. Fixed now.

cli/src/semgrep/metrics.py
bkettle: So much simpler now! Thanks :)

PRIVACY.md
bkettle: Just updated. To make the changes more straightforward, I moved the `unsupportedExts` key under `values`. Let me know if you think this doesn't make sense and I can put it back---I thought it seemed more consistent with the existing metrics though.

brendongo:
Can you add tests that verify this is working correctly

PRIVACY.md
```diff
@@ -268,6 +270,7 @@ This is a sample blob of the aggregate metrics described above:
         "numFindings": 7,
         "numIgnored": 3,
         "features": ["language/python", "option/deep", "option/no-git-ignore", "key/metavariable-comparison"]
+        "unsupportedExts": {".txt": 3, ".erb": 2, ".java": 5}
```
emjin, line 273: This is a nit but we do support `.java`. Even though this is just an example, I could see someone reading this and being confused. Shouldn't be too hard to dig up a language we don't support

bkettle:
Closing in favor of #5805 --- I misunderstood some of the requirements and thought it would make more sense to open a new PR given the differences.

PRIVACY.md
bkettle: One thing with this feature is that even if we do support a language, if the user doesn't have any rules selected that scan it, it will show up here---so this could be a real output. But I am happy to change it to be a different language. However, I realized that I misunderstood part of the requirements so am transitioning this to be in the scan metadata---I think that makes this section irrelevant. Sorry to waste your reviewing time

PR #5805, opened 2022-07-26
feat(app-metrics): track unsupported file extensions for app users

bkettle:
Closes APP-1354. This PR adds new reporting of file extensions that do not match any enabled language except generic. Note that this will still return file extensions of languages that _semgrep_ supports if none of the enabled rules support that language. File extension here is defined as `os.path.splitext` defines it:

> Split the pathname path into a pair (root, ext) such that root + ext == path, and the extension, ext, is empty or begins with a period and contains at most one period.

This does not report anything for files without an extension, including hidden files without an extensions such as .semgrepignore. Unlike the previous version #5802, these metrics are added as part of Scan metadata (only for App users). 

PR checklist:

### DIFF
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -1,6 +1,7 @@
 # Handle communication of findings / errors to semgrep.app
 import json
 import os
+from collections import Counter
 from pathlib import Path
 from typing import Any
 from typing import Dict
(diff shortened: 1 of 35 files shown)


cli/src/semgrep/app/scans.py
```diff
@@ -204,6 +212,7 @@ def report_findings(
                 "findings": len(new_matches),
                 "errors": [error.to_dict() for error in errors],
                 "total_time": total_time,
+                "unsupported_exts": ignored_ext_freqs,
```
brendongo: Probably want to convert to some json format since it looks like it prints out a counter object in the tests below

brendongo (approved):

PR #5806, opened 2022-07-26
feat(SCA): Revamp SCA findings generation

mmcqd:
This PR changes the way that SCA finding generation works. It also drastically speeds up SCA PR scans. Features of the new strategy:
* One SCA finding per vulnerable dependency. If one rule matches multiple dependencies in one lockfile, that will produce multiple findings. This still needs to be codified in the typed interface
* No findings in files that were not targeted. If `foo.py` depends on `Pipfile.lock`, and `foo.py` is targeted but `Pipfile.lock` is not, then we can produce reachable findings in `foo.py` but not non-reachable findings in `Pipfile.lock`. If `Pipfile.lock` is included in our targets then we can produce non-reachable findings inside of it
* No massive single scan for lockfiles. For reachable findings, we do a local search up the directory tree for the nearest lockfile, and memoize the process for speed. For non-reachable findings, we filter update the language filtering capacity of `TargetManager` to allow us to also filter out everything but lockfiles.
The other thing this PR does is:
* `baseline_commit` scans now only target files with findings. Previously we accidentally allowed the baseline scan to target the whole repo

PR checklist:


changelog.d/sca-127.changed
```diff
@@ -0,0 +1,11 @@
...
+- One SCA finding per vulnerable dependency. If one rule matches multiple dependencies in one lockfile,
+  that will produce multiple findings. This still needs to be codified in the typed interface
+- No findings in files that were not targeted. If foo.py depends on Pipfile.lock,
+  and foo.py is targeted but Pipfile.lock is not, then we can produce reachable findings
+  in foo.py but not non-reachable findings in Pipfile.lock. If Pipfile.lock is included in
+  our targets then we can produce non-reachable findings inside of it
+- No massive single scan for lockfiles. For reachable findings, we do a
+  local search up the directory tree for the nearest lockfile, and memoize
+  the process for speed. For non-reachable findings, we filter update the language
+  filtering capacity of TargetManager to allow us to also filter out everything but lockfiles.
```
underyx: Remember, this goes in the changelog, which our users read, so I wouldn't reference classes in our code like TargetManager.

cli/src/semdep/find_lockfiles.py
```diff
@@ -110,3 +114,30 @@ def make_dependency_trie(
                 deps = list(parse_lockfile_str(lockfile.read_text(), lockfile))
                 dep_trie.insert(lockfile, deps, namespace)
     return dep_trie
+
+
+@lru_cache(maxsize=None)
```
underyx:
```suggestion
@lru_cache(maxsize=1000)
```

is this called once per lockfile pattern? we might wanna set this to like 100x the expected calls 1) in case a bug would result in a memory leak and 2) as sort of documentation about how much we expect this to be called.

cli/src/semdep/package_restrictions.py
```diff
@@ -35,14 +34,13 @@ def semver_matches(expression: str, actual_version: str) -> bool:
 # compare vulnerable range to version in lockfile
 def dependencies_range_match_any(
     search_for_ranges: List[ProjectDependsOnEntry],
-    lockfile_path: Path,
     have_deps: List[LockfileDependency],
-) -> Generator[Tuple[ProjectDependsOnEntry, LockfileDependency, Path], None, None]:
+) -> Generator[Tuple[ProjectDependsOnEntry, LockfileDependency], None, None]:
```
underyx:
```suggestion
) -> Iterator[Tuple[ProjectDependsOnEntry, LockfileDependency]]:
```

I forget how generators are ever useful but I think you typically use them when you want to communicate some more input to it while it's iterating

cli/src/semgrep/dependency_aware_rule.py
```diff
@@ -54,61 +55,33 @@ def parse_depends_on_yaml(
...
-    matches: List[RuleMatch],
-    rule: Rule,
-    dep_trie: DependencyTrie,
-) -> Tuple[List[RuleMatch], List[SemgrepError]]:
-    """
-    Run a dependency aware rule. These rules filters the results based on the precense or absence
-    of dependencies. Dependencies are determined by searching for lockfiles under the first entry
-    in the `targets` argument.
-    """
+def generate_non_reachable_sca_findings(
```
underyx:
```suggestion
def generate_unreachable_sca_findings(
```

optionally we could make this consistent with phrasing elsewhere

cli/src/semdep/find_lockfiles.py
```diff
@@ -110,3 +114,30 @@ def make_dependency_trie(
...
+def search_path(p: Path, lockfile_pattern: str) -> Optional[Path]:
```
brendongo: Can you add function level comments on what these functions do. For this specifically probably worth clarifying that it only searches direct children of Path. 

You can prefix function name with `_` to suggest it is internal; we probably don't need this function outside of find_single_lockfile

Is there a reason to use glob over just checking `p / lockfile_pattern` exists else None? I don't think we can ever have multiples unless you allow case sensitivity.

cli/src/semdep/find_lockfiles.py
```diff
@@ -110,3 +114,30 @@ def make_dependency_trie(
...
+        return lockfiles[0]
+    else:
+        return None
+
+
+def find_single_lockfile(
+    p: Path, ecosystem: PackageManagers
+) -> Optional[Tuple[Path, List[LockfileDependency]]]:
+    parents = list(p.parents)
+    for path in parents:
```
brendongo: p.parents is immutable sequence so you can change to 
```suggestion
    for path in p.parents:
```

cli/src/semdep/find_lockfiles.py
```diff
@@ -110,3 +114,30 @@ def make_dependency_trie(
...
+        for lockfile_pattern in NAMESPACE_TO_LOCKFILES[ecosystem]:
+            lockfile = search_path(path, lockfile_pattern)
```
brendongo: See above comment on using `(path / lockfile_pattern).exists()`

cli/src/semdep/find_lockfiles.py
```diff
@@ -110,3 +114,30 @@ def make_dependency_trie(
...
+            if lockfile:
+                return lockfile, list(
+                    parse_lockfile_str(lockfile.read_text(encoding="utf8"), lockfile)
```
brendongo: is utf-8 a fair assumption for lockfiles?

cli/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awareno-pattern.yaml-dependency_aware/results.txt
```diff
@@ -21,6 +21,9 @@ SEMGREP_USER_AGENT_APPEND="pytest" SEMGREP_SETTINGS_FILE="<MASKED>" SEMGREP_VERS
 === stderr - plain
 Nothing to scan.
 
+Some files were skipped or only partially analyzed.
+  Scan skipped: 1 files matching .semgrepignore patterns
```
brendongo, line 25: Is this output expected?

cli/tests/e2e/targets/dependency_aware/lodash/yarn.lock
```diff
@@ -0,0 +1,24 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
brendongo, line 1: Is this comment accurate?

cli/src/semgrep/target_manager.py
```diff
@@ -540,7 +547,7 @@ def globfilter(self, candidates: Iterable[Path], pattern: str) -> List[Path]:
         return cast(List[Path], result)
 
     def filter_by_language(
-        self, language: Language, *, candidates: FrozenSet[Path]
+        self, language: Union[Language, PackageManagers], *, candidates: FrozenSet[Path]
```
brendongo, line 565: Probably need to update some comments with new behavior

cli/src/semgrep/semgrep_main.py
```diff
@@ -172,27 +171,42 @@ def run_rules(
...
+                rule_matches_by_rule[rule] = dep_rule_matches
+                output_handler.handle_semgrep_errors(dep_rule_errors)
+                (
+                    dep_rule_matches,
+                    dep_rule_errors,
+                    targeted_lockfiles,
+                ) = generate_non_reachable_sca_findings(
+                    rule, target_manager, matched_lockfiles
+                )
+                rule_matches_by_rule[rule].extend(dep_rule_matches)
```
brendongo, line 206: why is this extend?

cli/src/semgrep/semgrep_main.py
```diff
@@ -172,27 +171,42 @@ def run_rules(
...
+                output_handler.handle_semgrep_errors(dep_rule_errors)
+                all_targets.union(targeted_lockfiles)
```
brendongo, line 208: Do you also need to union with `matched_lockfiles`?

brendongo (approved):

brendongo:
Removing wait-for-me since matthew is out

cli/src/semdep/find_lockfiles.py
mmcqd: It's potentially called multiple times per lockfile pattern. If we have a yarn file in `A/` and we try to find the yarn file for `A/foo.js` and `A/B/bar.js`, then this function will get called on `A/` and `Yarn.lock` twice. So it should get called around `n * d` times, where `n` is the number of files with semgrep-core findings from reachability rules, and `d` is the average distance between a file and its lockfile in the directory tree.

cli/src/semdep/find_lockfiles.py
mmcqd: No there's no reason not to just check if the file exists. That seems like the better thing to do.

cli/src/semdep/find_lockfiles.py
mmcqd: Not actually sure

cli/src/semgrep/semgrep_main.py
mmcqd: As opposed to what? Append? We want to include the list of non-reachable findings generated by the reachability rules in the list of findings associated with the rule.

cli/src/semgrep/semgrep_main.py
mmcqd: Maybe... If we scan a lockfile as part of reachable result, but the lockfile itself was not a target of the scan, should we include it in the scanned files?

cli/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awareno-pattern.yaml-dependency_aware/results.txt
mmcqd: yes. That test runs on the whole set of targets, and it ignores `node_modules` (which it should). The new output is because lockfiles are getting ignored properly now. Before it was an ad-hoc check that didn't update this info when printing results.

cli/src/semdep/find_lockfiles.py
mmcqd: gonna go with yeah until it doesn't work lol

cli/src/semgrep/semgrep_main.py
brendongo: oh I see yes. I think makes sense to not include it

cli/src/semdep/find_lockfiles.py
mmcqd: Ok I just got rid of this function and am not bothering with caching because checking if a file exists is very fast

PR #5811, opened 2022-07-26
chore(docs): record memory/perf insights in Run_semgrep

emjin:
Since I'm closing off the memory investigation for now, I don't want to lose
the insights we gained. I'm recording them for now in a README in runner, as
well as some other notes that might be useful. This is not meant to be
exhaustive.

PR checklist:


kopecs (approved):

aryx:
what about putting it in semgrep-core/doc/?
Better chance to be found I think.

semgrep-core/src/runner/README.md
```diff
@@ -0,0 +1,90 @@
+# Notes
```
IagoAbal: I think this is not that long that it could go in `Run_semgrep.mli`.

IagoAbal: Also we had a document (maybe in Quip) with the "current perf issues", we could have one in Notion too, and maybe it's a better place.

emjin: I ended up putting it in the mli file; I like having it close to the code. Not sure I'd think to check Notion or keep it up to date

IagoAbal (approved):

PR #5815, opened 2022-07-27
feat(metrics): add language parse statistics

kopecs:
Closes PA-1678

PR checklist:

---

This PR adds an additional piece of metics information---by language parse rates. This consists of a language aggregated object which reports the targeted files & byte counts, and the amount of files we were able to parse without error and the number of bytes parsed.

The purpose of this is to help drive parsing improvements in the future by better understanding which of our parsers have poor performance.


kopecs:
Is the existing test coverage for metrics sufficient, or do we have tests on a per-metric basis?

semgrep-core/src/core/Report.ml
```diff
@@ -208,7 +208,7 @@ let collate_results init_extra unzip_extra base_case_extra final_extra results =
   let matches, errors, (skipped_targets, profiling) = unzip_results results in
   {
     matches = List.flatten matches;
-    errors = List.flatten errors;
+    errors = List.flatten errors |> List.sort_uniq compare;
```
emjin: Add a comment in `Semgrep_output_v0.atd` that we guarantee errors are sorted and deduped in core (though matches may not be), and then leave a corresponding comment here pointing users to the `Semgrep_output_v0.atd` comment

emjin: I would also add in the comment here (`Report.ml`) that we dedup errors more aggressively to be consistent with parse errors

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -134,7 +134,9 @@ let (group_matches_per_pattern_id : Pattern_match.t list -> id_to_match_results)
   h
 
 let error_with_rule_id rule_id (error : E.error) =
-  { error with rule_id = Some rule_id }
+  match error.typ with
+  | PartialParsing _ -> error
```
emjin, line 139: Could use a comment that this is to be consistent with parse errors

semgrep-core/src/core/Report.ml
emjin: Oh, one more thing---`collate_results` is run many times, since it's used by both `collate_rule_results` and `collate_pattern_results`. It's worth estimating the complexity of `List.sort_uniq` vs using a set

cli/src/semgrep/parsing_data.py
```diff
@@ -0,0 +1,88 @@
...
+        registered from the original plan with `add_targets`.
+        """
+        path = err.location.path
+        (lang, no_error_yet) = self._file_info[path]
+        lang_parse_data = self._parse_errors_by_lang[lang]
+        if no_error_yet:
+            lang_parse_data.targets_with_errors += 1
+
+        # Total errors for the whole file
+        if isinstance(
```
emjin, line 65: I'd be more comfortable if this were encoded in the interface file. The situation I'm imagining is: say someone turns `ParseError` recoverable. How would they know to change this?

emjin:
I don't know about metrics, but this does need a test to run semgrep on files with various kinds of errors (ideally exercise all five and ideally use files that are actually errors so that hopefully this test doesn't need to be changed later when an error disappears). It should enforce that the parse percentage reported is x%.

semgrep-core/src/core/Report.ml
kopecs: In terms of that last point, do the existing docs we have for perf cover what I would need to do to test that?

cli/src/semgrep/parsing_data.py
kopecs: Encoded in the sense of just a comment on each of the variants in the atd file documenting what we use it for, or are you envisioning something else?

semgrep-core/src/core/Report.ml
emjin: They should; tldr is `./run-benchmarks` with `ci-large-repos.yaml`. If that doesn't make any sense even after reading docs, poke me to update them

cli/src/semgrep/parsing_data.py
emjin: I don't have an option I really like. At minimum a bidirectional comment. Another possibility is to include it as part of the error type, the way `core_severity` is. I think the lack of testing is part of what's making me skittish, so I'll give it some more thought

kopecs: Including it with the error type feels more cumbersome to me, as it provides an additional field which has to get updated to be kept in a valid state (don't want to end up with `is_parse_error` as `true` but have the `error_kind` be `Timeout`).

Added a comment, and I'll increase test coverage.

kopecs:
Looks like CI failure is due to a lagging pr that got merged for semgrep-interfaces.

brendongo (approved):

PR #5821, opened 2022-07-27
fix(debug): make dump-il more cohesive

brandonspark:
**What:**
Currently, `dump-il` uses `Visitor_AST` to separately visit each function definition, and then print the IL stmts contained within. Notably, this means that it misses top-level statements, including the functions themselves.

**Why:**
If you use `dump-il` on the following program:
```py
x = "ls"

def foo():
  x = 5

def bar():
  x = True
```

it will display something like:
```
(S
   { s =
     (Block
        ((),
         [{ s =
            (ExprStmt (
               { e =
                 (Assign (
                    { e =
                      (N
                         (Id (("x", ()),
                            { id_resolved = ref ((Some (LocalVar, 2)));
                              id_type = ref (None); id_svalue = ref (None);
                              id_hidden = false; id_info_id = 2 }
                            )));
                      e_id = 0; e_range = None },
                    (),
                    { e = (L (Int ((Some 5), ()))); e_id = 0; e_range = None
                      }
                    ));
                 e_id = 0; e_range = None },
               ()));
            ...)
==>
(Ss
   [{ s =
      (Instr
         { i =
           (Assign (
              { base =
                (Var
                   { IL.ident = ("x", ()); sid = 2;
                     id_info =
                     { id_resolved = ref ((Some (LocalVar, 2)));
                       id_type = ref (None); id_svalue = ref (None);
                       id_hidden = false; id_info_id = 2 }
                     });
                offset = NoOffset },
              { e = (Literal (Int ((Some 5), ())));
                eorig = ...
                }
              ));
           iorig = ...
      }
     ])
(S
   { s =
     (Block
        ((),
         [{ s =
            (ExprStmt (
               { e =
                 (Assign (
                    { e =
                      (N
                         (Id (("x", ()),
                            { id_resolved = ref ((Some (LocalVar, 3)));
                              id_type = ref (None); id_svalue = ref (None);
                              id_hidden = false; id_info_id = 4 }
                            )));
                      e_id = 0; e_range = None },
                    (),
                    { e = (L (Bool (true, ()))); e_id = 0; e_range = None }));
                 e_id = 0; e_range = None },
               ...)
==>
(Ss
   [{ s =
      (Instr
         { i =
           (Assign (
              { base =
                (Var
                   { IL.ident = ("x", ()); sid = 3;
                     id_info =
                     { id_resolved = ref ((Some (LocalVar, 3)));
                       id_type = ref (None); id_svalue = ref (None);
                       id_hidden = false; id_info_id = 4 }
                     });
                offset = NoOffset },
              { e = (Literal (Bool (true, ())));
                eorig = ...
                }
              ));
           iorig = ...
      }
     ])
```
This is pretty misleading in my opinion, and looks like all the `x` assigns exist in the same top-level or something. The distinction between these functions is not super clear, and in any case, the actual function nodes themselves are completely missing, despite the fact that the IL really looks something like `DefStmt (x) -> MiscStmt (foo) -> MiscStmt (bar)`.

I think it would be nicer for debugging purposes to not obfuscate the IL's form.

**How:**
I changed `dump-il` such that we no longer use `Visitor_AST`, and instead just convert the AST to a series of IL statements, and then print the statements.

Also, I added a new `dump-il-functions` command, which will explicitly go and print out (as before) all top-level functions. It has effectively the same functionality as what `dump-il` currently is. It also prints out function names, if possible.

**Test plan:**
Try it out yourself.

**PR checklist:**

### DIFF
--- a/changelog.d/dump-il.fixed
+++ b/changelog.d/dump-il.fixed
@@ -1,1 +1,0 @@
-Changed it such that `dump-il` prints out all the top-level IL statements of a program, rather than individually printing out all of the statements in each function.
--- a/changelog.d/dump-il.fixed
+++ b/changelog.d/dump-il.fixed
@@ -1,0 +1,1 @@
+Changed it such that `dump-il` prints out all the top-level IL statements of a program, rather than individually printing out all of the statements in each function.
--- a/semgrep-core/src/experiments/datalog/Datalog_experiment.ml
+++ b/semgrep-core/src/experiments/datalog/Datalog_experiment.ml
@@ -70,25 +70,8 @@ let dump_il file =
   let lang = List.hd (Lang.langs_of_filename file) in
   let ast = Parse_target.parse_program file in
   Naming_AST.resolve lang ast;
-
-  let v =
-    V.mk_visitor
-      {
-        V.default_visitor with
-        V.kfunction_definition =
-          (fun (_k, _) def ->
-            let s =
-              AST_generic.show_any (G.S (H.funcbody_to_stmt def.G.fbody))
-            in
-            pr2 s;
-            pr2 "==>";
-
-            let _, xs = AST_to_IL.function_definition lang def in
-            let s = IL.show_any (IL.Ss xs) in
-            pr2 s);
-      }
-  in
-  v (G.Pr ast)
+  let xs = AST_to_IL.stmt lang (G.stmt1 ast) in
+  List.iter (fun stmt -> pr2 (IL.show_stmt stmt)) xs
   [@@action]
 
 (*****************************************************************************)
--- a/semgrep-core/src/experiments/datalog/Datalog_experiment.ml
+++ b/semgrep-core/src/experiments/datalog/Datalog_experiment.ml
@@ -62,19 +62,6 @@ module D = Datalog_fact
 type env = { facts : Datalog_fact.t list ref }
 
 (*****************************************************************************)
-(* Dumper *)
-(*****************************************************************************)
-
-(* mostly a copy paste of pfff/lang_GENERIC/analyze/Test_analyze_generic.ml *)
-let dump_il file =
-  let lang = List.hd (Lang.langs_of_filename file) in
-  let ast = Parse_target.parse_program file in
-  Naming_AST.resolve lang ast;
-  let xs = AST_to_IL.stmt lang (G.stmt1 ast) in
-  List.iter (fun stmt -> pr2 (IL.show_stmt stmt)) xs
-  [@@action]
-
-(*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
 let add env x = Common.push x env.facts
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -242,6 +242,58 @@ let dump_ast ?(naming = false) lang file =
           |> String.concat "\n");
         Runner_exit.(exit_semgrep False)))
 
+(* mostly a copy paste of pfff/lang_GENERIC/analyze/Test_analyze_generic.ml *)
+let dump_il file =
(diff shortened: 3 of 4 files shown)


semgrep-core/src/experiments/datalog/Datalog_experiment.ml
```diff
@@ -70,25 +70,8 @@ let dump_il file =
...
-            pr2 "==>";
-
-            let _, xs = AST_to_IL.function_definition lang def in
-            let s = IL.show_any (IL.Ss xs) in
-            pr2 s);
-      }
-  in
-  v (G.Pr ast)
+  let xs = AST_to_IL.stmt lang (G.stmt1 ast) in
+  List.iter (fun stmt -> pr2 (IL.show_stmt stmt)) xs
```
IagoAbal: Doesn't this require #5787 to be merged first to work as you describe it?

brandonspark: No, this is just a general change to print out the top-level statements in the IL. This is something that isn't done either way.

#5787 just makes new, shinier nodes at the top-level of the IL.

changelog.d/dump-il.fixed
```diff
@@ -0,0 +1 @@
+Changed it such that `dump-il` prints out all the top-level IL statements of a program, rather than individually printing out all of the statements in each function.
```
aryx: In general we do not add stuff useful only for us internally in the changelog (e.g., refactoring, improving the build system, or stuff related to the semgrep-core command-line program since users are not supposed to use semgrep-core directly but use semgrep instead.

aryx (approved):

semgrep-core/src/experiments/datalog/Datalog_experiment.ml
```diff
@@ -70,25 +70,8 @@ let dump_il file =
...
-            pr2 s;
-            pr2 "==>";
-
-            let _, xs = AST_to_IL.function_definition lang def in
-            let s = IL.show_any (IL.Ss xs) in
-            pr2 s);
-      }
-  in
-  v (G.Pr ast)
+  let xs = AST_to_IL.stmt lang (G.stmt1 ast) in
```
aryx: Oh, this function should not be in Datalog_experiment.ml. Not sure why it ended up there, but probably we want to put it in Main.ml next to the other dumper functions.

aryx (approved):

semgrep-core/src/experiments/datalog/Datalog_experiment.ml
IagoAbal: Without #5787, `AST_to_IL.stmt` is not going to translate `DefStmt`s with function definitions to IL, correct? So when you `IL.show_stmt` those statements, it will dump the Generic rather than the IL.

brandonspark: Without #5787, `AST_to_IL.stmt` still translates `DefStmt`s which are function definitions, it's just that it's translated to a more "degenerate" form. cf. 
image

I suppose what you mean by this is that the new `MiscStmt` is not really a "translation", since it just keeps the generic AST representation around. It's true that this will not show the translated version, since they're kept the same, but I think that that can only be a good thing, as it's representative of the true form of the data, which is just an encapsulated `DefStmt`.

IagoAbal: It's fine with me to have the program-level dump that you propose, but we still need to be able to dump the IL of the individual functions in the program in some other way. Because if taint doesn't work one of the first things I do is to check the IL translation of the function(s) involved.

IagoAbal (approved):
Nice! I would prefer `-dump_il_all` for dumping the entire file and `-dump_il` to continue dumping the individual functions. Simply because in my experience what I need most of the times is to see the IL of some function. But feel free to merge and I could change this later on if needed.

PR #5822, opened 2022-07-27
feat(semgrep-core cli): added implicit rules key

brandonspark:
**What:**
`semgrep-core` with the `-rules` option takes in a `.yaml` file, which must have a `rules: ...` at the top level. This PR just makes it so that the `rules: ...` is inserted implicitly if it is not found.

**Why:**
This is annoying, because sometimes I want to copy-paste from the Semgrep playground, which does not require a `rules: ...` (and in fact, does not allow it). This leads to precious seconds of developer time being wasted by me having to write `rules:` on the first line, and the indent the rest.

**How:**
If it doesn't match the `rules: ...`, but it otherwise looks like the format of the Semgrep playground, parse it anyways.

**Test plan:** Try it on the CLI. You can do `sc -rules test.yaml test.py -lang python`, for instance, with a `.yaml` file that looks like:
```
id: java-private-property-track
message: The XML parser $PARSER is not securely configured.
severity: ERROR
languages:
  - java
patterns:
  - pattern: |
      $FACTORY = javax.xml.parsers.DocumentBuilderFactory.newInstance()
  - pattern-inside: |
      $FACTORY = ...;
      ...
      $PARSER = $FACTORY.newDocumentBuilder();
```

PR checklist:

### DIFF
--- a/changelog.d/implicit-rules-sc-core.added
+++ b/changelog.d/implicit-rules-sc-core.added
@@ -1,0 +1,1 @@
+Changed `semgrep-core` so that it can now be run with `-rules` on `.yaml` files which do not have a top-level `rules: ...` key. This means you can now copy paste from the playground editor directly into a `.yaml` file for use with `semgrep-core`.
--- a/semgrep-core/src/parsing/Parse_rule.ml
+++ b/semgrep-core/src/parsing/Parse_rule.ml
@@ -980,7 +980,9 @@ let parse_generic ?(error_recovery = false) file ast =
         (* it's also ok to not have the toplevel rules:, anyway we never
          * used another toplevel key
          *)
+        | G.Container (G.Dict, (l, _, _)) -> (l, [ e ])
         | G.Container (G.Array, (l, rules, _r)) -> (l, rules)
+        (* otherwise, it's bad, and complain *)
         | _ ->
             let loc = PI.first_loc_of_file file in
             yaml_error (PI.mk_info_of_loc loc)


ievans:
This is great but note that gobs of backend registry code assumes that a .yaml file has a `rules` key. If we start standardizing this, we might quickly run into issues there. @brendongo 

It also means that we are breaking the Semgrep rule schema. Viz my previous attempt https://github.com/returntocorp/semgrep/pull/4972

brendongo:
@ievans looks like this is just convenience function for semgrep-core for development purposes :D

brendongo (approved):

aryx:
nice!

changelog.d/implicit-rules-sc-core.added
```diff
@@ -0,0 +1 @@
+Changed `semgrep-core` so that it can now be run with `-rules` on `.yaml` files which do not have a top-level `rules: ...` key. This means you can now copy paste from the playground editor directly into a `.yaml` file for use with `semgrep-core`.
```
aryx, line 1: hmm the absence of toplevel rules: was already accepted I think (see the comment in the code).
But you had to have an array, so something like

- id: rule1
  message: xxx
  ...
- id: rule2
  message: yyy
  ...

What you added is when you have only 1 rule, to not even have to put it in an array with the leading - but just
id: rule1
message: bla
...

IagoAbal, line 1: Unless this means Semgrep CLI will also accept `.yaml` files without a top `rules:` element, this should not go into the changelog, since semgrep-core is internal only.

PR #5830, opened 2022-07-28
chore: enforce submodule commits to be on default branch

kopecs:
PR checklist:

---

This PR adds a new github actions workflow to enforce that the major submodules (i.e., [pfff](https://github.com/returntocorp/pfff), [semgrep-langs](https://github.com/returntocorp/semgrep-langs), [semgrep-interfaces](https://github.com/returntocorp/semgrep-interfaces), and [ocaml-tree-sitter-core](https://github.com/returntocorp/ocaml-tree-sitter-core); excludes the various auto-updated semgrep-$LANG submodules) point to a commit on the default branch (either `develop` or `main`).

This has the benefit of avoiding possible merge conflicts which can arise when a commit makes in onto develop here. Moreover, it avoids the potential issue of commits in the upstream submodule being gc'd.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -511,20 +511,24 @@ jobs:
       - name: Check semgrep-interfaces
         uses: jtmullen/submodule-branch-check-action@v1
         with:
-          path: "${GITHUB_WORKSPACE}/cli/src/semgrep/semgrep_interfaces"
+          # Looks like this uses the value from
+          # git config --file .gitmodules --name-only --get-regexp path
+          # (per https://github.com/jtmullen/submodule-branch-check-action/blob/6c6f99f42640ce5247e478880be431940ad3c515/entrypoint.sh#L94)
+          # instead of the actual path?
+          path: cli/src/semgrep/semgrep-interfaces
           branch: main
       - name: Check semgrep-langs
         uses: jtmullen/submodule-branch-check-action@v1
         with:
-          path: "${GITHUB_WORKSPACE}/cli/src/semgrep/lang"
+          path: cli/src/semgrep/lang
           branch: main
       - name: Check ocaml-tree-sitter-core
         uses: jtmullen/submodule-branch-check-action@v1
         with:
-          path: "${GITHUB_WORKSPACE}/semgrep-core/src/ocaml-tree-sitter-core"
+          path: semgrep-core/src/ocaml-tree-sitter-core
           branch: main
       - name: Check pfff
         uses: jtmullen/submodule-branch-check-action@v1
         with:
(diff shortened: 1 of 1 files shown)


kopecs:
It seems like there's an upstream issue: https://github.com/jtmullen/submodule-branch-check-action/issues/23

We should be able to work around this by renaming the submodule, but it might be easier to just fix it there.

Edit: Resolved upstream

aryx:
Not a big fan. I like to be able to make a PR1 on a submodule X, then in semgrep make another PR2 that references this PR branch and then get all reviewed and accepted and just merge. With your proposal, before merging PR2 I will have to reference the new merged PR1 from PR2, wait for another CI run, and finally being able to merge.

aryx:
I never saw this "oh the branch got gc'ed and so everything broke".
Another option would be to not squash and instead merge, so the original commit reference remains, even when the branch is merged, because the commit is still in the main branch.

IagoAbal:
I think I pushed for this but the reason wasn't only the possibility of those branches getting GCed. Recently I had to upgrade DeepSemgrep to the latest Semgrep and it was quite messy for me because DeepSemgrep depended on a feature branch, and before that feature branch was merged, something else got merged that conflicted with DeepSemgrep. It took me a while to realize what was going on and fix it.

You should be able to open the PRs as usual. Simply you need to merge PR1 before PR2, and yes you will have to re-run a check. But that check should re-run super quickly I guess? Because the only thing that it does it is to check that the dependency refers to a branch that has been merged into `develop`.

kopecs:
Yeah, I think merge conflict avoidance is the bigger deal than the risk of a commit randomly disappearing. I was having some trouble with semgrep_interfaces, which spurred this, because we relied on a commit which existed only on a feature branch, but which had been (I presume) squash-merged. So when I went to bump the version to include commits which were merged into the default branch I ran into a pretty nasty merge conflict (and my understanding is that this isn't an isolated experience).

I think this is more helpful as a guardrail that anything else. Like Iago says, you might need to re-run one check after merging, but this will help us ensure that develop here is always in a consistent state, and you can always rely on a commit not on the default branch while you're working on a feature branch here.

nmote:
I don't have a strong opinion here. The main thing is that I think we should make sure we don't end up with submodules that point to commits that could be garbage collected. GitHub does not seem to actually ever GC anything, but that's an implementation detail that we should not be relying on. Disabling rebase+merge and squash+merge in repos used as submodules should be enough to ensure this in practice, as long as people remember not to rebase manually in those repos.

I kind of like the flexibility that the current setup gives, but I recognize that it can sometimes lead to confusing results.

aryx:
Let's close it for now.
No consensus reached.

PR #5835, opened 2022-07-29
Fix offset not carrying over with multiple autofixes

0xDC0DE:
This causes a bug when multiple fixes are applied in the same file that change the number of lines.
This bug is most easily reproduced with fixes that reduce the number of lines, as that can cause the current implementation to read a line that no longer exists.

Reproduce on the develop or release version with this rule:

```
rules:
- id: broken-method-fix
  languages:
  - python
  message: |
    Use `sys.exit` over the python shell `exit` built-in. `exit` is a helper
    for the interactive shell and is not be available on all Python implementations.
    https://stackoverflow.com/a/6501134
  patterns:
    - pattern: $REQUEST_FUNC($...ARGS)
    - pattern-not: $REQUEST_FUNC(..., _injected=$N, ...)
    - metavariable-pattern:
        metavariable: $REQUEST_FUNC
        patterns:
          - pattern: requests.post
  fix: $REQUEST_FUNC($...ARGS, _injected=True)
  severity: ERROR
```

and following test code:
```
import requests

def _do_stuff():
    return requests.post(
        headers=auth_headers,
        url=self.url,
        endpoints="endpoint",
    )


def _do_other_stuff():
    return requests.post(
        headers=auth_headers,
        url=self.url,
        endpoints="endpoint",
    )


def _do_other_stuff2():
    return requests.post(
        headers=auth_headers,
        url=self.url,
        endpoints="endpoint"
    )
```

This will throw an error. Error can be avoided by adding extra padding of empty lines to the bottom of the file. But the fix is still applied to the wrong lines.

This is fixed with this PR.


ievans:
looks great @pieterdc1 , can tests be added too?

DrewDennison (approved):

brendongo:
~~I thought this would have been addressed by https://github.com/returntocorp/semgrep/pull/4429/files~~ nvm

brendongo:
Lgtm but can you add the reproduction as a test so we dont regress :D

brendongo (approved):

PR #5856, opened 2022-08-02
Add support for wildcard matching interpolated regexps in Ruby

### LINKED ISSUE #5147: [ruby] Regex interpolations not caught by wildcard matches
**Describe the bug**
Regex interpolations should match wildcard regex patterns.

**To Reproduce**
https://semgrep.dev/s/1J9l

mjambon:
Fixes https://github.com/returntocorp/semgrep/issues/5147

It took me a while to figure out what's going on with string matching. The result is that regexp content within `//` is handled like interpolated strings (`""` in Ruby), with additional support for `/.../` and `/$X/` which both match any regexp, whether it's a constant or an interpolated string.

`/.../` or `/$X/` will match `/hello #{name}/`, but `"..."` or `"$X"` still don't match the interpolated string `"hello #{name}"` (if this isn't desired, let me know).

PR checklist:

  - \<type> is `added`, `changed`, `fixed`, or infra`.


semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -602,6 +603,18 @@ and literal =
   (* Go, Python *)
   | Ratio of string wrap
 
+(*
+(*
+   A regexp is a list of template fragment. For example, /a*/ has one fragment
+   and the Ruby template /hello #{name}/ has two fragments.
+*)
+and regexp_fragment =
```
aryx: we might want actually to do that also for String instead of the ugly Interpolated construct we currently have
(see AST_generic.interpolated function).

aryx: We might also want to have a StringLiteral of string wrap bracket | RegexpLiteral of string wrap bracket * string wrap option for the easy cases, and then String of string_or_regexp_fragment list bracket | Regexp of string_or_regexp_fragment list bracket * string wrap option

aryx: And maybe at some point we can get rid of StringLiteral and RegexpLiteral if we feel the special case is not very useful (but for constant propagation they are useful at least).

aryx (changes requested):

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -592,7 +594,6 @@ and literal =
                   escaped  unescaped
   *)
   | String of string wrap
-  | Regexp of string wrap bracket (* // *) * string wrap option (* modifiers *)
```
IagoAbal, line 595: Won't this make const-prop to stop working as we discussed in Slack?

semgrep-core/src/core/ast/AST_generic.ml
mjambon: What I did today is split regexps into literals and templates. Templates are still a list of expressions because it's simpler than having a new type for just the kind of expression a template may hold. We can always improve this later.

semgrep-core/src/core/ast/AST_generic.ml
mjambon: Yes. To avoid this problem, I restored this line and we now have separate expression node just for regexp templates.

mjambon:
Here's what we have now:
* `/.../` will match both regexp literals and regexp templates
* `/$X/` will match only regexp literals, binding the metavariable to the regexp string
* `/$...X/` is not implemented but would capture the contents of any regexp (template or literal)

semgrep-core/src/core/ast/Pretty_print_AST.ml
```diff
@@ -766,12 +766,6 @@ and literal env l =
       | Lang.R
       | Lang.Swift ->
           "\"" ^ s ^ "\"")
-  | Regexp ((_, (s, _), _), rmod) -> (
```
aryx, line 769: Why this code was removed?

semgrep-core/src/matching/SubAST_generic.ml
```diff
@@ -179,6 +179,7 @@ let subexprs_of_expr with_symbolic_propagation e =
           xml_body
   (* currently skipped over but could recurse *)
   | Constructor _
+  | RegexpTemplate _
```
aryx, line 182: Seems like something we should recurse on.
We do for Interpolated strings I think (represented by Calls and IdSpecial).

semgrep-core/src/naming/Naming_AST.ml
```diff
@@ -340,8 +340,9 @@ let get_resolved_type lang (vinit, vtype) =
       | Some { e = L (Float (_, tok)); _ } -> make_type "float" tok
       | Some { e = L (Char (_, tok)); _ } -> make_type "char" tok
       | Some { e = L (String (_, tok)); _ } -> make_type string_str tok
-      | Some { e = L (Regexp ((_, (_, tok), _), _)); _ } ->
```
aryx, line 343: Why did you remove the Regexp case? This one is also a regexp type.

semgrep-core/src/parsing/pfff/dune
```diff
@@ -10,6 +10,7 @@
    dyp
 
    commons commons_core
+   semgrep_core
```
aryx, line 13: I'd rather not have semgrep_core dependencies here (see the top comment in this file).

semgrep-core/src/parsing/pfff/ruby_to_generic.ml
```diff
@@ -448,24 +463,24 @@ and literal x =
...
-        | StrChars (s, t) :: tl -> f (s :: strs) (t :: toks) tl
-        | StrExpr _ :: _
-        | [] ->
-            (* TODO *)
-            string_contents_list (l, xs, r)
+            (("Subshell", l), [ G.E (interpolated_string (l, xs, r) |> G.e) ]))
+  | Regexp ((l, xs, r), opt) -> (
+      let literal_or_template =
+        match concatenate_literal_fragments xs with
+        (* /.../ matches any regexp literal or template *)
```
aryx, line 473: You could return a Regexp with a string literal "..." and recognize specially this in Generic_vs_generic.
/.../ is usually represented like this in other languages.

semgrep-core/tests/ruby/regexp_metavar.rb
```diff
@@ -0,0 +1,26 @@
...
+# MATCH:
+/\\/
+
+# MATCH:
+/\z/
+
+# MATCH:
+/\a\z/
+
+# to match this regexp template, use /.../ (implemented) or /$...X/ (TODO)
```
aryx, line 25: yep, makes sense.

semgrep-core/src/core/ast/Pretty_print_AST.ml
mjambon: I forgot to restore it after my initial series of changes and there was a catch-all case `| x ->` which prevented my mistake from being detected.

semgrep-core/src/matching/SubAST_generic.ml
mjambon: yes. I added this with a test.

semgrep-core/src/naming/Naming_AST.ml
mjambon: This is also something I forgot to restore.

semgrep-core/src/parsing/pfff/ruby_to_generic.ml
mjambon: I'd rather move away from using strings to represent ellipses (or metavariables for the matter).

PR #5864, opened 2022-08-02
feat(swift): added swift labeled statements for parsing

brandonspark:
**What:**
Added ability to parse Swift labeled statements to Generic AST.

This entails also being able to parse a bunch of statements, notably:
- `for-in` statements
- `while` statements
- `if` statements
- `repeat-while` statements
- `switch` statements
- `do` statements

This PR adds functionality for doing that.

**Why:**
These are common constructs in the Swift language and should be a good step towards increasing our parse rate.

**How:**
So, notably, Swift has a bunch of random things that I decided to just make `OtherPat`s and the like. For instance:
- `try` and `await` patterns
- `?` patterns
- `var` and `let` in conditions

Let me know if I handled these strangely.

Here's some examples of `OtherPat`:
```swift
for try await x in collection {
}
```
parses to
```
Pr(
  [For((),
     ForEach(
       (OtherPat(("Await", ()),
          [P(
             OtherPat(("Try", ()),
               [P(
                  PatId(("x", ()),
                    {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))]))]),
        (),
        N(
          Id(("collection", ()),
            {id_info_id=2; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); })))),
     Block([]))])
```

This is because the `x` is a pattern which is ranging over the collection, and the `try` and `await` tokens seem to be modifying the pattern. So I stuck them in an `OtherPat`.

Here's another example:
```swift
if async let x = true {} else {}
```
parses to
```
Pr(
  [If((),
     Cond(
       LetPattern(
         OtherPat(("async", ()),
           [P(
              OtherPat(("let", ()),
                [P(
                   PatId(("x", ()),
                     {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))]))]),
         L(Bool((true, ()))))), Block([]), Some(Block([])))])
```

This is because this `if` is interesting, in that it binds a variable in the conditional. I was looking for an `expr` which seemed like it did that, and the closest thing I could find was `LetPattern`, so I flagged it under that.

Similarly, `let` and `async` seem to be modifiers on the pattern, so I just put them there.

Finally:
```swift
switch foo {
  case let x where true == true:
    return 2
  case var x:
    return 2
}
```
parses to
```
Pr(
  [Switch((),
     Some(Cond(
            N(
              Id(("foo", ()),
                {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); })))),
     [CasesAndBody(
        [Case((),
           PatWhen(
             OtherPat(("let", ()),
               [P(
                  PatId(("x", ()),
                    {id_info_id=2; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))]),
             Call(IdSpecial((Op(Eq), ())), [Arg(L(Bool((true, ())))); Arg(L(Bool((true, ()))))])))],
        Block([Return((), Some(L(Int((Some(2), ())))), ())]));
      CasesAndBody(
        [Case((),
           OtherPat(("var", ()),
             [P(
                PatId(("x", ()),
                  {id_info_id=3; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))]))],
        Block([Return((), Some(L(Int((Some(2), ())))), ())]))])])
```

This is pretty much the same thing.

**Test plan:**
`make test`

PR checklist:

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
@@ -85,8 +85,8 @@ let map_function_modifier (env : env) (x : CST.function_modifier) =
 let map_binding_pattern_kind (env : env) (x : CST.binding_pattern_kind) :
     (string * PI.t) list =
   match x with
-  | `Var tok -> (* "var" *) [ ("Var", token env tok) ]
-  | `Let tok -> (* "let" *) [ ("Let", token env tok) ]
+  | `Var tok -> (* "var" *) [ str env tok ]
+  | `Let tok -> (* "let" *) [ str env tok ]
 
 let map_possibly_async_binding_pattern_kind (env : env)
     ((v1, v2) : CST.possibly_async_binding_pattern_kind) =
@@ -94,7 +94,7 @@ let map_possibly_async_binding_pattern_kind (env : env)
     match v1 with
     | Some tok ->
         (* async_modifier *)
-        [ ("async", token env tok) ]
+        [ str env tok ]
     | None -> []
   in
   async @ map_binding_pattern_kind env v2
--- a/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
@@ -2487,18 +2487,18 @@ and map_subscript_declaration (env : env)
 and map_switch_entry (env : env) ((v1, v2, v3, v4, v5) : CST.switch_entry) =
   let v1 =
     match v1 with
-    | Some x -> map_modifiers env x
-    | None -> todo env ()
+    | Some x -> map_modifiers env x |> todo env
+    | None -> ()
   in
-  let v2 =
+  let case =
     match v2 with
     | `Case_switch_pat_opt_where_kw_exp_rep_COMMA_switch_pat (v1, v2, v3, v4) ->
-        let v1 = (* "case" *) token env v1 in
-        let v2 = map_switch_pattern env v2 in
-        let v3 =
+        let case_tok = (* "case" *) token env v1 in
+        let pat_init = map_switch_pattern env v2 in
+        let add_pat_where pat =
           match v3 with
-          | Some x -> map_where_clause env x
-          | None -> todo env ()
+          | Some x -> G.PatWhen (pat, map_where_clause env x)
+          | None -> pat
         in
         let v4 =
           Common.map
@@ -2508,29 +2508,31 @@ and map_switch_entry (env : env) ((v1, v2, v3, v4, v5) : CST.switch_entry) =
               todo env (v1, v2))
             v4
         in
-        todo env (v1, v2, v3, v4)
-    | `Defa_kw tok -> (* default_keyword *) token env tok
+        let pat = pat_init |> add_pat_where in
+        G.Case (case_tok, pat)
+    | `Defa_kw tok -> (* default_keyword *) G.Default (token env tok)
   in
   let v3 = (* ":" *) token env v3 in
-  let v4 = map_statements env v4 in
+  let stmt = G.Block (G.fake_bracket (map_statements env v4)) in
+  (* For now, don't deal with fallthrough. *)
   let v5 =
     match v5 with
-    | Some tok -> (* "fallthrough" *) token env tok
-    | None -> todo env ()
+    | Some tok -> (* "fallthrough" *) token env tok |> todo env
+    | None -> ()
   in
-  todo env (v1, v2, v3, v4, v5)
+  G.CasesAndBody ([ case ], stmt |> G.s)
 
 and map_switch_pattern (env : env) (x : CST.switch_pattern) =
   map_binding_pattern_with_expr env x
 
 and map_switch_statement (env : env)
     ((v1, v2, v3, v4, v5) : CST.switch_statement) =
-  let v1 = (* "switch" *) token env v1 in
-  let v2 = map_expression env v2 in
+  let switch_tok = (* "switch" *) token env v1 in
+  let expr = map_expression env v2 in
   let v3 = (* "{" *) token env v3 in
   let v4 = Common.map (map_switch_entry env) v4 in
   let v5 = (* "}" *) token env v5 in
-  todo env (v1, v2, v3, v4, v5)
+  G.Switch (switch_tok, Some (G.Cond expr), v4) |> G.s
 
 and map_expr_hack_at_ternary_binary_suffix (env : env)
     (x : CST.expr_hack_at_ternary_binary_suffix) =
--- a/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
@@ -981,19 +981,18 @@ and map_capture_list_item (env : env) (x : CST.capture_list_item) =
       todo env (v1, v2, v3)
 
 and map_catch_block (env : env) ((v1, v2, v3, v4) : CST.catch_block) =
-  let v1 = (* catch_keyword *) token env v1 in
-  let v2 =
-    match v2 with
-    | Some x -> map_binding_pattern_no_expr env x
-    | None -> todo env ()
+  let catch_tok = (* catch_keyword *) token env v1 in
+  let pat =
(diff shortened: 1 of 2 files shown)


semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -82,21 +82,22 @@ let map_function_modifier (env : env) (x : CST.function_modifier) =
...
 
-let map_binding_pattern_kind (env : env) (x : CST.binding_pattern_kind) : G.tok
-    =
+let map_binding_pattern_kind (env : env) (x : CST.binding_pattern_kind) :
+    (string * PI.t) list =
   match x with
-  | `Var tok -> (* "var" *) token env tok
-  | `Let tok -> (* "let" *) token env tok
+  | `Var tok -> (* "var" *) [ ("Var", token env tok) ]
+  | `Let tok -> (* "let" *) [ ("Let", token env tok) ]
```
nmote: you probably want the `str` helper function here

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -82,21 +82,22 @@ let map_function_modifier (env : env) (x : CST.function_modifier) =
...
 let map_possibly_async_binding_pattern_kind (env : env)
     ((v1, v2) : CST.possibly_async_binding_pattern_kind) =
-  let v1 =
+  let async =
     match v1 with
-    | Some tok -> (* async_modifier *) token env tok |> todo env
-    | None -> ()
+    | Some tok ->
+        (* async_modifier *)
+        [ ("async", token env tok) ]
```
nmote: same here

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -82,21 +82,22 @@ let map_function_modifier (env : env) (x : CST.function_modifier) =
   | `Post tok -> (* "postfix" *) token env tok
   | `Prefix tok -> (* "prefix" *) token env tok
 
-let map_binding_pattern_kind (env : env) (x : CST.binding_pattern_kind) : G.tok
-    =
+let map_binding_pattern_kind (env : env) (x : CST.binding_pattern_kind) :
+    (string * PI.t) list =
```
nmote: should this just be an identifier? I think it's the same type

nmote (changes requested):
Thanks! Could you add a section to the test plan with the resulting AST from some of these constructs? I'm especially interested in seeing the `OtherPat` results. I'll take a closer look once that's done.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
brandonspark: how do you mean? I can't seem to find an identifier type that looks like that - I might be looking in the wrong place

nmote: Try `G.ident`

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -982,19 +981,18 @@ and map_capture_list_item (env : env) (x : CST.capture_list_item) =
...
-  in
-  let v4 = map_function_body env v4 in
-  todo env (v1, v2, v3, v4)
+  let catch_tok = (* catch_keyword *) token env v1 in
+  let pat =
+    match (v2, v3) with
+    (* Similar to how Python does it: *)
+    | None, None -> G.PatUnderscore (Parse_info.fake_info catch_tok "_")
+    | Some v2, None -> map_binding_pattern_no_expr env v2
+    | None, Some _ -> raise Common.Impossible
```
nmote, line 993: Could you add a justification for why this case is impossible?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1240,40 +1248,48 @@ and map_expression (env : env) (x : CST.expression) : G.expr =
...
+          G.OtherPat (("Try", map_try_operator env x), [ G.P pat ])
+      | None -> pat
+    in
+    let add_pat_await pat =
+      match v3 with
+      | Some tok (* "await" *) ->
+          G.OtherPat (("Await", token env tok), [ G.P pat ])
+      | None -> pat
+    in
+    pat_init |> add_pat_ty |> add_pat_where |> add_pat_try |> add_pat_await
```
nmote, line 1278: I think this is a good candidate for simplification via `let*` (as syntactic sugar for `Option.bind`) but I'm also fine with it as-is.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1503,11 +1519,13 @@ and map_key_path_postfixes (env : env) (x : CST.key_path_postfixes) =
       let v3 = (* "]" *) token env v3 in
       todo env (v1, v2, v3)
 
-and map_labeled_statement (env : env) ((v1, v2) : CST.labeled_statement) =
+and map_labeled_statement (env : env) ((v1_orig, v2) : CST.labeled_statement) =
```
nmote: change this back to `v1`?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -2038,19 +2067,22 @@ and map_universally_allowed_pattern (env : env)
...
-        | None -> todo env ()
-      in
       let v3 = (* dot_custom *) token env v3 in
-      let v4 = map_bound_identifier env v4 in
-      let v5 =
-        match v5 with
-        | Some x -> map_tuple_pattern env x
-        | None -> todo env ()
+      let id = map_bound_identifier env v4 in
+      let id_info = G.empty_id_info () in
```
nmote, line 2076: There should be helper functions we can use to avoid having to construct this manually. I don't remember what they are but let me know if you have trouble finding them and I can track them down.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -2038,19 +2067,22 @@ and map_universally_allowed_pattern (env : env)
...
+      let pat_init = G.PatId (id, id_info) in
+      let add_pat_args name pat =
+        match Option.map (map_tuple_pattern env) v5 with
+        | None -> pat
+        | Some (G.PatTuple (_, pats, _)) -> G.PatConstructor (name, pats)
+        | _ -> raise Common.Impossible
```
nmote: Might it be possible to refactor so that this isn't necessary? Maybe `map_tuple_pattern` can return just the contents of a `PatTuple`?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -2339,16 +2372,24 @@ and map_raw_str_interpolation (env : env)
...
+    in
+    G.Block (left, stmts, right) |> G.s
   in
-  let v4 = (* "}" *) token env v4 in
   let v5 = (* "while" *) token env v5 in
-  let v6 = map_if_condition_sequence_item env v6 in
+  let expr =
+    match map_if_condition_sequence_item env v6 with
+    | Cond expr -> expr
+    | _ -> raise Common.Impossible (* for now *)
```
nmote: Explain this?

nmote (approved):
Looks good! Just a few minor things to address inline.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
brandonspark: It seems I can't bind both to a single identifier with `name_of_id` because OCaml does not actually treat constructor arguments as tuples without explicit parens. It's very strange.

PR #5867, opened 2022-08-02
Quick-fix for regression in 0.107.0 after taint labels PR

IagoAbal:
Follows: 8f26fabe29a ("tainting: Add experimental taint labels feature (#5725)")

Helps CSE-116
Helps PA-1724

After PR #5725 (taint labels), we started to get reports of crashes like:

    (Invalid_argument "output_value: abstract value (Custom)")
    Raised by primitive operation at Parmap.marshal in file "src/parmap.ml", line 146, characters 10-48
    Called from Parmap.spawn_many.loop in file "src/parmap.ml", line 168, characters 8-23
    Called from Parmap.run_many in file "src/parmap.ml" (inlined), line 190, characters 16-45
    Called from Parmap.simplemapper in file "src/parmap.ml", line 208, characters 2-678
    Called from Run_semgrep.semgrep_with_rules in file "src/runner/Run_semgrep.ml", line 596, characters 4-1023
    Called from Run_semgrep.semgrep_with_raw_results_and_exn_handler in file "src/runner/Run_semgrep.ml", line 667, characters 21-58

This seems to happenn when Parmap tries to do the marshalling of the
taint trace. But it happens even with a fake empty taint trace! And it is
"fixed" if we force the taint-trace's lazy thunk before marshalling it
(presumably that is why it only crashes with -json, since in text mode
the match and trace are printed right away). It needs -j 2 or higher to
be reproduced, so that Parmap is used.

We do not yet understand why it happens, but e.g. in the added test, the
crash doesn't happen if we comment out the `metavariable-regex`. Taint
labels shouldn't be affecting the result returned by Semgrep-core in
these cases, and the match and the taint trace look identical when
compared with the commit before 8f26fabe29a.

Also, we now test that the result of `Match_rules.check` can be
marshalled. This does not seem to affect the runtime of `make test`.

test plan:
make test # added one test

PR checklist:

### DIFF
--- a/changelog.d/pa-1724.fixed
+++ b/changelog.d/pa-1724.fixed
@@ -1,0 +1,4 @@
+Quick fix for a regression introduced in 0.107.0 (presumably by taint labels)
+that could cause some taint rules to crash Semgrep with:
+
+    Invalid_argument "output_value: abstract value (Custom)"
--- a/semgrep-core/src/engine/Match_tainting_mode.ml
+++ b/semgrep-core/src/engine/Match_tainting_mode.ml
@@ -413,7 +413,18 @@ let pm_of_finding finding =
        * already expect Semgrep (and DeepSemgrep) to report the match on `sink(x)`.
        *)
       let taint_trace =
-        Some (lazy (taint_trace_of_src_to_sink source tokens sink))
+        Some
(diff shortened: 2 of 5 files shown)


nmote (approved):
looks reasonable as a stopgap!

PR #5870, opened 2022-08-03
feat(swift): most of swift parsing

brandonspark:
**What:**
Added functionality for parsing Swift enums into the generic AST, along with modifiers, optional types, and literal patterns.

Then I also added most other things.

**Why:**
I pulled up a random Swift repo on Github, and the first thing that failed was that we couldn't parse enums. It's probably a healthy reason for why our parse rate is so abysmal - imagine trying to parse a random OCaml project without `type`.

I also went through more repos and found we couldn't parse them because of heavy use of modifiers like `public`, `private`, decorators, etc. I also found some other miscellaneous things like optional type sugar and casing on literals, so I fixed those too.

**How:**

----------------------------------------------------------------------------------------------------

**Enums:**

I went and handled the TODOs.

It's worth noting that Swift has both C-style enums where each variant is just an alias for some base-type constant, _and_ ML-style sum types. The real terrible thing going on here is that they both use the `enum` syntax.

For raw-style enums, I've mostly followed how Java does them, in that each `enum` entry defines an `EnumEntryDef`. For instance:
```swift
enum Foo: Int {
  case a = 1
  case b
}
```
parses to
```
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("Foo", ()),
                {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[KeywordAttr((EnumClass, ()))]; tparams=[]; },
      ClassDef(
        {ckind=(Class, ());
         cextends=[({t_attrs=[];
                     t=TyN(
                         Id(("Int", ()),
                           {id_info_id=2; id_hidden=false; id_resolved=Ref(
                            None); id_type=Ref(None); id_svalue=Ref(None); }));
                     },
                    None)];
         cimplements=[]; cmixins=[]; cparams=[];
         cbody=[F(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("a", ()),
                               {id_info_id=3; id_hidden=false; id_resolved=Ref(
                                None); }));
                      attrs=[]; tparams=[]; },
                     EnumEntryDef({ee_args=Some([Arg(L(Int((Some(1), ()))))]); ee_body=None; }))));
                F(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("b", ()),
                               {id_info_id=4; id_hidden=false; id_resolved=Ref(
                                None); }));
                      attrs=[]; tparams=[]; },
                     EnumEntryDef({ee_args=None; ee_body=None; }))))];
         })))])
```

Notably, this allows intermingling of regular fields with the actual enums themselves. This is intended.

What we're doing right now is that we just parse each as an `EnumEntryDef` anyways, even though it's not raw-style, and actually works as part of a sum type. This lets it intermingle with the other fields in the class.

Here's an example of parsing a union-style enum:
```
enum Foo {
  case a(Int)
  case b
}
```
goes to
```
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("Foo", ()),
                {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[KeywordAttr((EnumClass, ()))]; tparams=[]; },
      ClassDef(
        {ckind=(Class, ()); cextends=[]; cimplements=[]; cmixins=[]; cparams=[
         ];
         cbody=[F(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("a", ()),
                               {id_info_id=2; id_hidden=false; id_resolved=Ref(
                                None); }));
                      attrs=[]; tparams=[]; },
                     EnumEntryDef(
                       {
                        ee_args=Some([ArgType(
                                        {t_attrs=[];
                                         t=TyRecordAnon((Class, ()),
                                             [F(
                                                DefStmt(
                                                  ({name=OtherEntity(("AnonTupleField", ()), []); attrs=[]; tparams=[]; },
                                                   FieldDefColon(
                                                     {vinit=None;
                                                      vtype=Some({t_attrs=[
                                                                  ];
                                                                  t=TyN(
                                                                    Id((
                                                                    "Int", ()),
                                                                    {id_info_id=3; id_hidden=false;
                                                                    id_resolved=Ref(
                                                                    None); id_type=Ref(
                                                                    None); id_svalue=Ref(
                                                                    None); }));
                                                                  });
                                                      }))))]);
                                         })]);
                        ee_body=None; }))));
                F(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("b", ()),
                               {id_info_id=4; id_hidden=false; id_resolved=Ref(
                                None); }));
                      attrs=[]; tparams=[]; },
                     EnumEntryDef({ee_args=None; ee_body=None; }))))];
         })))])
```

----------------------------------------------------------------------------------------------------

Here's some decisions I made for other stuff:

**Modifiers:** 

For modifiers, here's the dump on the choices I made. Consider a:
```swift
infix postfix prefix mutating nonmutating private public open fileprivate
internal private(set) public(set) open(set) fileprivate(set) internal(set)
weak unowned unowned(safe) unowned(unsafe) final override convenience required
static dynamic optional class inout @escaping @autoclosure lazy
@attr
func foo () { }
```
which is a common code pattern that Swift developers write in their everyday code.

This parses as:
```
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("foo", ()),
                {id_info_id=18; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[NamedAttr((),
                Id(("Infix", ()),
                  {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]);
              NamedAttr((),
                Id(("Postfix", ()),
                  {id_info_id=2; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]);
              NamedAttr((),
                Id(("Prefix", ()),
                  {id_info_id=3; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]);
              NamedAttr((),
                Id(("Mutating", ()),
                  {id_info_id=4; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]);
              NamedAttr((),
                Id(("Nonmutating", ()),
                  {id_info_id=5; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]); KeywordAttr((Private, ())); KeywordAttr((Public, ()));
              KeywordAttr((Public, ())); KeywordAttr((Public, ())); KeywordAttr((Protected, ()));
              OtherAttribute(("Set", ()), [At(KeywordAttr((Private, ())))]);
              OtherAttribute(("Set", ()), [At(KeywordAttr((Public, ())))]);
              OtherAttribute(("Set", ()), [At(KeywordAttr((Public, ())))]);
              OtherAttribute(("Set", ()), [At(KeywordAttr((Public, ())))]);
              OtherAttribute(("Set", ()), [At(KeywordAttr((Protected, ())))]);
              NamedAttr((),
                Id(("Weak", ()),
                  {id_info_id=6; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]);
              NamedAttr((),
                Id(("Unowned", ()),
                  {id_info_id=7; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]);
              NamedAttr((),
                Id(("UnownedSafe", ()),
                  {id_info_id=8; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]);
              NamedAttr((),
                Id(("UnownedUnsafe", ()),
                  {id_info_id=9; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ]); KeywordAttr((Final, ())); KeywordAttr((Override, ()));
              NamedAttr((),
                Id(("Convenience", ()),
                  {id_info_id=10; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }),
                []);
              NamedAttr((),
                Id(("Required", ()),
                  {id_info_id=11; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }),
                []); KeywordAttr((Static, ()));
              NamedAttr((),
                Id(("Dynamic", ()),
                  {id_info_id=12; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }),
                []); KeywordAttr((Optional, ()));
              NamedAttr((),
                Id(("Class", ()),
                  {id_info_id=13; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }),
                []);
              NamedAttr((),
                Id(("Inout", ()),
                  {id_info_id=14; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }),
                []);
              NamedAttr((),
                Id(("Escaping", ()),
                  {id_info_id=15; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }),
                []);
              NamedAttr((),
                Id(("Autoclosure", ()),
                  {id_info_id=16; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }),
                []); KeywordAttr((Lazy, ()));
              NamedAttr((),
                Id(("attr", ()),
                  {id_info_id=17; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }),
                [])];
       tparams=[]; },
      FuncDef({fkind=(Function, ()); fparams=[]; frettype=None; fbody=FBStmt(Block([])); })))])
```

The important thing to note is that for most of these, we don't have a matching `KeywordAttr` that describes them, so I just made them `NamedAttr`s.

For the access modifiers `public`, `private`, `fileprivate`, `open`, and `internal`, it's a little more complicated. Those modifiers are described [here](https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html).

`open` and `public` allow code to be used throughout any file, and only differ in subclassing behavior, so I've made them both `Public`. I figure that we don't really care about name resolution for subclassing at the moment.

`fileprivate` only allows access within the current file, and since Semgrep only does single-file analysis right now, I marked that as `Public` too.

`internal` only allows access from within the defining module, and I wasn't sure of how that interacts with subclassing. Because of that, by analogy to Java, I just marked it as `Protected` behavior to me.

`private` is the most private setting, so it's `Private`. It doesn't allow usage outside of the current declaration.

Also worth noting is that each of these access levels has a suffix `(set)` that may be added, to denote that it is only a modifier on the setting privileges of these variables. I decided to just inject it into an `OtherAttr`, which wraps the particular access level attribute.

It's worth noting that there isn't really anything distinguishing a Swift attribute (something like `@attr`) from a non-Swift attribute which is named `attr`.

----------------------------------------------------------------------------------------------------

**Switches and Patterns:**

Here's also an example of a `switch` using a `where` clause, along with a literal pattern:
```swift
switch true {
  case 1:
    var x = 2
  case true where true == true, 2:
    var x = 2
}
```
goes to
```
Pr(
  [Switch((), Some(Cond(L(Bool((true, ()))))),
     [CasesAndBody([Case((), PatLiteral(Int((Some(1), ()))))],
        Block(
          [DefStmt(
             ({
               name=EN(
                      Id(("x", ()),
                        {id_info_id=2; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
               attrs=[NamedAttr((),
                        Id(("Var", ()),
                          {id_info_id=1; id_hidden=false; id_resolved=Ref(
                           None); id_type=Ref(None); id_svalue=Ref(None); }), [
                        ])];
               tparams=[]; },
              VarDef({vinit=Some(L(Int((Some(2), ())))); vtype=None; })))]));
      CasesAndBody(
        [Case((),
           PatDisj(
             PatWhen(PatLiteral(Bool((true, ()))),
               Call(IdSpecial((Op(Eq), ())), [Arg(L(Bool((true, ())))); Arg(L(Bool((true, ()))))])),
             PatLiteral(Int((Some(2), ())))))],
        Block(
          [DefStmt(
             ({
               name=EN(
                      Id(("x", ()),
                        {id_info_id=4; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
               attrs=[NamedAttr((),
                        Id(("Var", ()),
                          {id_info_id=3; id_hidden=false; id_resolved=Ref(
                           None); id_type=Ref(None); id_svalue=Ref(None); }), [
                        ])];
               tparams=[]; },
              VarDef({vinit=Some(L(Int((Some(2), ())))); vtype=None; })))]))])])
```

I just used `PatLiteral` and `PatWhen` here. I also turned all the comma-separated patterns into `PatDisj`s, because they seem to be disjunctive patterns.

It's also interesting to note that the tree-sitter parser for Swift is wrong. It only allows `where` clauses on the first disjunctive pattern, but it's actually possible to have them on all of them. Swift playground confirms this. This makes me mildly upset, get off my lawn, grrr.

----------------------------------------------------------------------------------------------------

**Question Mark Types:**

I also implemented support for optional type syntactic sugar. So for instance:
```swift
var x : Int? = 2
```
goes to
```
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("x", ()),
                {id_info_id=2; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[NamedAttr((),
                Id(("Var", ()),
                  {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }), [
                ])];
       tparams=[]; },
      VarDef(
        {vinit=Some(L(Int((Some(2), ()))));
         vtype=Some({t_attrs=[];
                     t=TyQuestion((),
                         {t_attrs=[];
                          t=TyN(
                              Id(("Int", ()),
                                {id_info_id=3; id_hidden=false; id_resolved=Ref(
                                 None); }));
                          });
                     });
         })))])
```

(I used `TyQuestion`, which was conveniently existent)

----------------------------------------------------------------------------------------------------
**Closures**:

Probably the next most important thing I did was for closures.

So for instance, the closure
```
{ @foo [ self, x, y = 3]
(self, name, externalname realname : inout @escaping @autoclosure Int ) -> Int in
return 2
};
```
parses to

```
Pr(
  [ExprStmt(
     Lambda(
       {fkind=(LambdaKind, ());
        fparams=[OtherParam(("Self", ()), []);
                 Param(
                   {pname=Some(("name", ())); pdefault=None; ptype=None; pattrs=[
                    ];
                    pinfo={id_info_id=6; id_hidden=false; id_resolved=Ref(
                           Some((Parameter, -1))); id_type=Ref(None); id_svalue=Ref(
                           None); };
                    });
                 Param(
                   {pname=Some(("realname", ())); pdefault=None;
                    ptype=Some({
                                t_attrs=[NamedAttr((),
                                           Id(("Inout", ()),
                                             {id_info_id=7; id_hidden=false; id_resolved=Ref(
                                              None); }), []);
                                         NamedAttr((),
                                           Id(("Escaping", ()),
                                             {id_info_id=8; id_hidden=false; id_resolved=Ref(
                                              None); }), []);
                                         NamedAttr((),
                                           Id(("Autoclosure", ()),
                                             {id_info_id=9; id_hidden=false; id_resolved=Ref(
                                              None); }), [])];
                                t=TyN(
                                    Id(("Int", ()),
                                      {id_info_id=10; id_hidden=false; id_resolved=Ref(
                                       None); }));
                                });
                    pattrs=[NamedAttr((),
                              Id(("Inout", ()),
                                {id_info_id=7; id_hidden=false; id_resolved=Ref(
                                 None); }), []);
                            NamedAttr((),
                              Id(("Escaping", ()),
                                {id_info_id=8; id_hidden=false; id_resolved=Ref(
                                 None); }), []);
                            NamedAttr((),
                              Id(("Autoclosure", ()),
                                {id_info_id=9; id_hidden=false; id_resolved=Ref(
                                 None); }), [])];
                    pinfo={id_info_id=11; id_hidden=false; id_resolved=Ref(
                           Some((Parameter, -1))); id_type=Ref(None); id_svalue=Ref(
                           None); };
                    })];
        frettype=Some({t_attrs=[];
                       t=TyN(
                           Id(("Int", ()),
                             {id_info_id=12; id_hidden=false; id_resolved=Ref(
                              None); }));
                       });
        fbody=FBStmt(
                OtherStmtWithStmt(OSWS_Closure,
                  [Pa(
                     Param(
                       {pname=Some(("self", ())); pdefault=None; ptype=None; pattrs=[
                        ];
                        pinfo={id_info_id=2; id_hidden=false; id_resolved=Ref(
                               Some((Parameter, -1))); id_type=Ref(None); id_svalue=Ref(
                               None); };
                        }));
                   Pa(
                     Param(
                       {pname=Some(("x", ())); pdefault=None; ptype=None; pattrs=[
                        ];
                        pinfo={id_info_id=3; id_hidden=false; id_resolved=Ref(
                               Some((Parameter, -1))); id_type=Ref(None); id_svalue=Ref(
                               None); };
                        }));
                   Pa(
                     Param(
                       {pname=Some(("y", ())); pdefault=Some(L(Int((Some(3), ())))); ptype=None; pattrs=[
                        ];
                        pinfo={id_info_id=5; id_hidden=false; id_resolved=Ref(
                               Some((Parameter, -1))); id_type=Ref(None); id_svalue=Ref(
                               None); };
                        }))], Block([Return((), Some(L(Int((Some(2), ())))), ())])));
        }), ())])
```

This closure supports attributes, capture lists, arguments with attributes, and closure signatures. 

Notably, we ignore the external name for an argument, and also the capture list needed somewhere to go, so I put it in an `OtherStmtWithStmt`.  I added a new `other_stmt_with_stmt_operator` `OSWS_Closure`, and put the capture list as a list of parameters in there. Then the statement enclosed is just the body of the lambda.

----------------------------------------------------------------------------------------------------

**FINAL NOTES:**

Functions whose names are operators don’t have any special names, they’re just identifiers whose names are the same as that of the operator.

++ and — are parsed to Incrdecr ops, with default fixity of postfix. There's no reason for this, but I just needed to put something.

`get`, `set`, and `modify` methods have names which are just those keywords.

I and’ed all the sequences of conditions together.

A tuple type is a TyRecordAnon, the entries of which are FieldDefColons. This is because Swift tuples are allowed to have both entries which have names and do not have names.

Type protocol conforming constraints and equal type constraints are just named attributes on the type.

Things which are OtherExprs:
- #fileID
- #filePath
- #column
- #function
- #dsohandle
- #colorLiteral
- #fileLiteral
- #imageLiteral
- Getter
- setter

A large amount of modifiers are currently handled as just NamedAttributes.
These include:
- Infix
- Postfix
- Prefix
- Weak
- Unowned
- UnownedSafe
- UnownedUnsafe
- Mutating
- Nonmutating
- Class
- Inout
- @escaping
- @autoclosure
- Set
- Indirect
- Type
- Protocol
- Dynamic

Usages of keyword attributes (from Swift to generic AST) are:
- Var -> Mutable
- Let -> Const
- Async -> Async
- Final -> Final
- Lazy -> Lazy
- Visibility modifiers, which warrant further inspection. See Parse_swift_tree_sitter.ml:374
- Optional -> optional
- Static -> static

I also added in a random thing which does reporting for the place in code where TODOs are tripped, and prints them out to the `.log` when running the parse statistics, and sorts them by most frequent. This helps in doing more targeted generic AST conversion, to make sure that we handle the TODOs that are most common.

-----------------------------------------------------------------------------------------------

**Test plan:**
Look at the above, and check whether I'm doing stuff correctly.

Also, `make test`.

PR checklist:

### DIFF
--- a/semgrep-core/src/parsing/Test_parsing.ml
+++ b/semgrep-core/src/parsing/Test_parsing.ml
@@ -29,64 +29,9 @@ let logger = Logging.get_logger [ __MODULE__ ]
 (* Helpers *)
 (*****************************************************************************)
 
-module SMap = Map.Make (String)
-module IMap = Map.Make (Int)
-
-type mode = Searching | Finding
-
-(* Global store, because I don't want to pass it through a bunch of func calls.
- *)
-let store = ref []
-
-let process_exn () =
-  let rec process mode (lines : string list) =
-    match lines with
-    | [] -> "failure"
-    | line :: rest -> (
-        (* Sue me. *)
-        let tokens = Common.split " " line in
-        match (mode, tokens) with
-        | Searching, "Called" :: "from" :: _ -> process Finding rest
-        | Searching, _ -> process Searching rest
-        | ( Finding,
-            "Called" :: "from" :: funcname :: _in :: _file :: _filename
-            :: "line" :: linenum :: _ ) ->
-            funcname ^ ":" ^ linenum
-            (* There should be at least one `Called from` after the first... *)
-        | Finding, _ -> failwith "bad")
-  in
-  let res =
-    Printexc.get_backtrace ()
-    |> Common.split (Str.quote "\n")
-    |> process Searching
-  in
-  store := res :: !store;
-  ()
-
 let print_exn file e =
   let trace = Printexc.get_backtrace () in
-  process_exn ();
   pr2 (spf "%s: exn = %s\n%s" file (Common.exn_to_s e) trace)
-
-let report_counts () =
-  let counts =
-    List.fold_left
-      (fun map x ->
-        SMap.update x
-          (function
-            | None -> Some 1
-            | Some x -> Some (x + 1))
-          map)
-      SMap.empty !store
-    |> SMap.to_seq
-    |> Seq.map (fun (x, y) -> (y, x))
-    |> IMap.of_seq |> IMap.to_rev_seq |> List.of_seq
-  in
-  pr2 "\nTODO statistics:";
-  List.fold_left
-    (fun acc (count, filename) -> acc ^ Common.spf "\n%s -> %d" filename count)
-    "" counts
-  |> pr2
 
 let dump_and_print_errors dumper (res : 'a Tree_sitter_run.Parsing_result.t) =
   (match res.program with
@@ -472,7 +417,6 @@ let parse_projects ~verbose lang project_dirs =
 
 let parsing_stats ?(json = false) ?(verbose = false) lang project_dirs =
   let stat_list = parse_projects ~verbose lang project_dirs in
-  report_counts ();
   if json then print_json lang stat_list
   else
     let flat_stat = List.concat_map snd stat_list in
--- a/semgrep-core/src/parsing/Test_parsing.ml
+++ b/semgrep-core/src/parsing/Test_parsing.ml
@@ -29,9 +29,64 @@ let logger = Logging.get_logger [ __MODULE__ ]
 (* Helpers *)
 (*****************************************************************************)
 
+module SMap = Map.Make (String)
+module IMap = Map.Make (Int)
+
+type mode = Searching | Finding
+
+(* Global store, because I don't want to pass it through a bunch of func calls.
+ *)
+let store = ref []
+
+let process_exn () =
+  let rec process mode (lines : string list) =
+    match lines with
+    | [] -> "failure"
+    | line :: rest -> (
+        (* Sue me. *)
+        let tokens = Common.split " " line in
+        match (mode, tokens) with
+        | Searching, "Called" :: "from" :: _ -> process Finding rest
+        | Searching, _ -> process Searching rest
+        | ( Finding,
+            "Called" :: "from" :: funcname :: _in :: _file :: _filename
+            :: "line" :: linenum :: _ ) ->
+            funcname ^ ":" ^ linenum
+            (* There should be at least one `Called from` after the first... *)
+        | Finding, _ -> failwith "bad")
+  in
+  let res =
+    Printexc.get_backtrace ()
+    |> Common.split (Str.quote "\n")
+    |> process Searching
+  in
+  store := res :: !store;
+  ()
+
 let print_exn file e =
   let trace = Printexc.get_backtrace () in
+  process_exn ();
   pr2 (spf "%s: exn = %s\n%s" file (Common.exn_to_s e) trace)
+
+let report_counts () =
+  let counts =
+    List.fold_left
+      (fun map x ->
+        SMap.update x
+          (function
+            | None -> Some 1
+            | Some x -> Some (x + 1))
+          map)
+      SMap.empty !store
+    |> SMap.to_seq
+    |> Seq.map (fun (x, y) -> (y, x))
+    |> IMap.of_seq |> IMap.to_rev_seq |> List.of_seq
+  in
+  pr2 "\nTODO statistics:";
+  List.fold_left
+    (fun acc (count, filename) -> acc ^ Common.spf "\n%s -> %d" filename count)
+    "" counts
+  |> pr2
 
 let dump_and_print_errors dumper (res : 'a Tree_sitter_run.Parsing_result.t) =
   (match res.program with
@@ -417,6 +472,7 @@ let parse_projects ~verbose lang project_dirs =
 
 let parsing_stats ?(json = false) ?(verbose = false) lang project_dirs =
   let stat_list = parse_projects ~verbose lang project_dirs in
+  report_counts ();
   if json then print_json lang stat_list
   else
     let flat_stat = List.concat_map snd stat_list in
--- a/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
@@ -459,14 +459,15 @@ let map_precedence_group_attribute (env : env)
 
 let map_tuple_type_item_identifier (env : env)
     ((v1, v2, v3) : CST.tuple_type_item_identifier) =
+  (* I don't really know why we permit underscores here. *)
   let v1 =
     match v1 with
-    | Some tok -> (* "_" *) token env tok
-    | None -> todo env ()
+    | Some tok -> (* "_" *) token env tok |> todo env
+    | None -> ()
   in
   let v2 = map_simple_identifier env v2 in
   let v3 = (* ":" *) token env v3 in
-  todo env (v1, v2, v3)
+  v2
 
 let map_referenceable_operator (env : env) (x : CST.referenceable_operator) =
   match x with
@@ -579,57 +580,54 @@ let rec map_annotated_inheritance_specifier (env : env)
   let v2 = map_inheritance_specifier env v2 in
   v2
 
-and map_enum_entry_suffix (env : env) (x : CST.enum_entry_suffix) =
+and map_enum_entry_suffix_union (env : env) (ident : G.ident)
+    (x : CST.enum_entry_suffix) : G.type_ =
   match x with
   | `Enum_type_params (v1, v2, v3) ->
       let v1 = (* "(" *) token env v1 in
-      let v2 =
+      let init_tok = v1 in
+      let fields =
         match v2 with
         | Some (v1, v2, v3, v4) ->
-            let v1 =
-              match v1 with
-              | Some x -> map_tuple_type_item_identifier env x
-              | None -> todo env ()
+            let mk_field id_opt ty expr_opt =
+              let name =
+                match id_opt with
+                | Some x ->
+                    let ident = map_tuple_type_item_identifier env x in
+                    G.EN (G.Id (ident, G.empty_id_info ()))
+                | None -> G.OtherEntity (("AnonTupleField", init_tok), [])
+              in
+              let ty = map_type_ env ty in
+              let ent = { G.name; attrs = []; tparams = [] } in
+              G.DefStmt (ent, G.FieldDefColon { vinit = None; vtype = Some ty })
             in
-            let v2 = map_type_ env v2 in
-            let v3 =
-              match v3 with
-              | Some (v1, v2) ->
-                  let v1 = (* eq_custom *) token env v1 in
-                  let v2 = map_expression env v2 in
-                  todo env (v1, v2)
-              | None -> todo env ()
-            in
-            let v4 =
+            let field_first = mk_field v1 v2 v3 in
+            let field_rest =
               Common.map
                 (fun (v1, v2, v3, v4) ->
                   let v1 = (* "," *) token env v1 in
-                  let v2 =
-                    match v2 with
-                    | Some x -> map_tuple_type_item_identifier env x
-                    | None -> todo env ()
-                  in
-                  let v3 = map_type_ env v3 in
-                  let v4 =
-                    match v4 with
-                    | Some (v1, v2) ->
-                        let v1 = (* eq_custom *) token env v1 in
-                        let v2 = map_expression env v2 in
-                        todo env (v1, v2)
-                    | None -> todo env ()
-                  in
-                  todo env (v1, v2, v3, v4))
+                  mk_field v2 v3 v4)
                 v4
             in
-            todo env (v1, v2, v3, v4)
-        | None -> todo env ()
+            Common.map (fun x -> G.F (x |> G.s)) (field_first :: field_rest)
+        | None -> []
       in
       let v3 = (* ")" *) token env v3 in
-      todo env (v1, v2, v3)
+      G.TyRecordAnon ((G.Class, v1), (v1, fields, v3)) |> G.t
+  | `Equal_sign_exp (v1, v2) -> raise Common.Impossible
+
+and map_enum_entry_suffix_raw (env : env) (ent : G.entity)
+    (x : CST.enum_entry_suffix) : G.stmt =
+  match x with
+  | `Enum_type_params (v1, v2, v3) -> raise Common.Impossible
   | `Equal_sign_exp (v1, v2) ->
       let v1 = (* eq_custom *) token env v1 in
-      let v2 = map_expression env v2 in
-      todo env (v1, v2)
+      let exp = map_expression env v2 in
+      let defkind =
+        G.EnumEntryDef
+          { ee_args = Some (G.fake_bracket [ G.arg exp ]); ee_body = None }
+      in
+      G.DefStmt (ent, defkind) |> G.s
 
 and map_type_casting_pattern (env : env) (x : CST.type_casting_pattern) =
   match x with
@@ -1168,58 +1166,99 @@ and map_else_options (env : env) (x : CST.else_options) =
   | `Blk x -> map_block env x
   | `If_stmt x -> map_if_statement env x
 
-and map_enum_class_body (env : env) ((v1, v2, v3) : CST.enum_class_body) =
+and map_enum_class_body (is_raw : bool) (enum_ident : G.ident) (env : env)
+    ((v1, v2, v3) : CST.enum_class_body) =
   let v1 = (* "{" *) token env v1 in
-  let v2 =
-    Common.map
-      (fun x ->
-        match x with
-        | `Enum_entry x -> map_enum_entry env x
-        | `Type_level_decl x -> map_type_level_declaration env x)
-      v2
-    |> List.concat_map (fun entries ->
-           Common.map (fun entry -> G.F entry) entries)
-  in
   let v3 = (* "}" *) token env v3 in
-  (v1, v2, v3)
+  (* If it's raw, we expect initialized fields, and not constructors.
+     Vice versa for the opposite.
+  *)
+  let fields =
+    if is_raw then
+      let fields =
+        Common.map
+          (fun x ->
+            match x with
+            | `Enum_entry x -> map_enum_entry_raw env x
+            | `Type_level_decl x -> map_type_level_declaration env x)
+          v2
+        |> List.concat_map (fun fields -> Common.map (fun x -> G.F x) fields)
+      in
+      fields
+    else
+      (* Get all of the enum variants, and put them at the front as an OrType declaration.
+     *)
+      let collect_entries entries =
+        List.fold_right
+          (fun entry (variants, decls) ->
+            match entry with
+            | Either.Left new_variants -> (new_variants @ variants, decls)
+            | Right new_decls -> (variants, new_decls @ decls))
+          entries ([], [])
+      in
+      let or_type_elems, stmts =
+        Common.map
+          (fun x ->
+            match x with
+            | `Enum_entry x -> Either.Left (map_enum_entry_union env x)
+            | `Type_level_decl x ->
+                Either.Right (map_type_level_declaration env x))
+          v2
+        |> collect_entries
+      in
+      let type_def = { G.tbody = OrType or_type_elems } in
+      (G.DefStmt (G.basic_entity enum_ident, G.TypeDef type_def) |> G.s)
+      :: stmts
+      |> Common.map (fun x -> G.F x)
+  in
+  (v1, fields, v3)
 
-and map_enum_entry (env : env) ((v1, v2, v3, v4, v5, v6, v7) : CST.enum_entry) =
-  let v1 =
+and map_enum_entry_raw (env : env)
+    ((v1, v2, v3, v4, v5, v6, _v7) : CST.enum_entry) : G.stmt list =
+  let modifiers =
     match v1 with
-    | Some x -> map_modifiers env x
-    | None -> todo env ()
+    | Some x -> map_modifiers env x |> todo env
+    | None -> []
   in
   let v2 =
     match v2 with
-    | Some tok -> (* "indirect" *) token env tok
-    | None -> todo env ()
+    | Some tok -> (* "indirect" *) token env tok |> todo env
+    | None -> []
   in
-  let v3 = (* "case" *) token env v3 in
-  let v4 = map_simple_identifier env v4 in
-  let v5 =
+  let mk_variant id suffix_opt =
+    let ent = G.basic_entity (map_simple_identifier env v4) in
     match v5 with
-    | Some x -> map_enum_entry_suffix env x
-    | None -> todo env ()
+    | Some x -> map_enum_entry_suffix_raw env ent x
+    | None ->
+        G.DefStmt (ent, G.EnumEntryDef { ee_args = None; ee_body = None })
+        |> G.s
   in
-  let v6 =
-    Common.map
-      (fun (v1, v2, v3) ->
-        let v1 = (* "," *) token env v1 in
-        let v2 = map_simple_identifier env v2 in
-        let v3 =
-          match v3 with
-          | Some x -> map_enum_entry_suffix env x
-          | None -> todo env ()
-        in
-        todo env (v1, v2, v3))
-      v6
+  let variant_first = mk_variant v4 v5 in
+  let variants_rest = Common.map (fun (v1, v2, v3) -> mk_variant v2 v3) v6 in
+  variant_first :: variants_rest
+
+and map_enum_entry_union (env : env)
+    ((v1, v2, v3, v4, v5, v6, _v7) : CST.enum_entry) : G.or_type_element list =
+  let modifiers =
+    match v1 with
+    | Some x -> map_modifiers env x |> todo env
+    | None -> []
   in
-  let v7 =
-    match v7 with
-    | Some tok -> (* ";" *) token env tok
-    | None -> todo env ()
+  let v2 =
+    match v2 with
+    | Some tok -> (* "indirect" *) token env tok |> todo env
+    | None -> []
   in
-  todo env (v1, v2, v3, v4, v5, v6, v7)
+  let mk_variant id suffix_opt =
+    let ident = map_simple_identifier env v4 in
+    match v5 with
+    | Some x ->
+        G.OrConstructor (ident, [ map_enum_entry_suffix_union env ident x ])
+    | None -> G.OrConstructor (ident, [])
+  in
+  let variant_first = mk_variant v4 v5 in
+  let variants_rest = Common.map (fun (v1, v2, v3) -> mk_variant v2 v3) v6 in
+  variant_first :: variants_rest
 
 and map_expression (env : env) (x : CST.expression) : G.expr =
   match x with
@@ -1792,6 +1831,18 @@ and construct_class_def :
   in
   G.ClassDef definition_kind
 
+(* TODO?: probably not complete, but that's OK *)
+and is_raw_type env s =
+  match s with
+  | "Int"
+  | "Double"
+  | "String"
+  | "Float"
+  | "UInt"
+  | "Character" ->
+      true
+  | _ -> false
+
 and map_modifierless_class_declaration (env : env)
     (x : CST.modifierless_class_declaration) =
   match x with
@@ -1840,13 +1891,30 @@ and map_modifierless_class_declaration (env : env)
       let v2 = (* "enum" *) token env v2 in
       let attrs = [ G.KeywordAttr (G.EnumClass, v2) ] in
       let v3 = map_simple_identifier env v3 in
+      (* If there are any tparams, then this must be a raw-value style enum. *)
       let tparams =
         match v4 with
         | Some x -> map_type_parameters env x
         | None -> []
       in
+      (* Basically, we can tell if we have a raw type if it's a certain kind of
+         base type like integers, floating point values, or whatever.
+         It's just kind of a pain to extract that information.
+      *)
+      let is_raw =
+        match v5 with
+        | None -> false
+        (* Yeah, I know, but this should cover the use cases that matter.
+         *)
+        | Some (_, (([], `User_type ((bid, None), [])), [])) ->
+            let s, _ = map_bound_identifier env bid in
+            is_raw_type env s
+        | _ -> false
+      in
       let entity = G.basic_entity ~attrs ~tparams v3 in
-      let classdef = construct_class_def env v2 v5 v6 v7 map_enum_class_body in
+      let classdef =
+        construct_class_def env v2 v5 v6 v7 (map_enum_class_body is_raw v3)
+      in
       G.DefStmt (entity, classdef) |> G.s
 
 and map_modifierless_function_declaration (env : env)
--- a/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
@@ -80,13 +80,13 @@ let map_function_modifier (env : env) (x : CST.function_modifier) =
   match x with
   | `Infix tok ->
       (* "infix" *)
-      G.unhandled_keywordattr ("Infix", token env tok)
+      G.unhandled_keywordattr (str env tok)
   | `Post tok ->
       (* "postfix" *)
-      G.unhandled_keywordattr ("Postfix", token env tok)
+      G.unhandled_keywordattr (str env tok)
   | `Prefix tok ->
       (* "prefix" *)
-      G.unhandled_keywordattr ("Prefix", token env tok)
+      G.unhandled_keywordattr (str env tok)
 
 let map_binding_pattern_kind (env : env) (x : CST.binding_pattern_kind) :
     G.ident list =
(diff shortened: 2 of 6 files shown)


nmote:
I'm not clear on the reason for splitting up parsing into two different modes. Could you provide specifics on the undesirable outcome that would occur if we were to parse enum declarations on a best effort basis without having this context?

brandonspark:
> I'm not clear on the reason for splitting up parsing into two different modes. Could you provide specifics on the undesirable outcome that would occur if we were to parse enum declarations on a best effort basis without having this context?

Well, doing it on a best effort basis results in messier code, I feel.

The problem is that depending on which mode we're in, we both modes support something like `case a;`. Now, in a union-style enum, this is simply a constructor `a` which takes no arguments. In a raw-style enum, this is instead a case of the enum which is initialized to one plus the previous raw value enum variant, or a default value if it is the first.

Semantics aside, syntactically this presents an issue because we don't know if we should parse it as an `EnumEntryDef`, or if we should go save it and put it inside of an `OrType`. We could scout ahead and look at the enum's contents to see whether it is used as a raw style or union-style, but that doesn't stop us from partitioning our code's behavior here. We could simply parse however we feel like, but if we start parsing it as a raw style, for instance, once we see a `case a(Int)` we're going to need to go back and remove the previous instances of `EnumEntryDef` that we put.

It just seems cleaner and like the least amount of code duplication to split the code flow early based on a clear indicator that we're given.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1635,85 +1841,91 @@ and map_lambda_literal (env : env) ((v1, v2, v3, v4, v5) : CST.lambda_literal) :
     (* Fake brackets here since the brackets delimit the lambda expression as a
      * whole, not just the statements *)
     (* TODO consider using `in` and the closing bracket as the delimiters *)
-    G.FBStmt (G.Block (G.fake_bracket stmts) |> G.s)
+    G.FBStmt
+      (G.OtherStmtWithStmt
+         ( G.OSWS_Closure,
+           Common.map (fun x -> G.Pa x) captures,
+           G.Block (G.fake_bracket stmts) |> G.s )
+      |> G.s)
```
nmote, line 1863: Makes sense. Could you add an explanation similar to what you said in the PR summary as a comment here?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1608,22 +1818,18 @@ and map_lambda_literal (env : env) ((v1, v2, v3, v4, v5) : CST.lambda_literal) :
     G.expr =
   let v1 = (* "{" *) token env v1 in
   (* TODO include captures *)
```
nmote: Remove this TODO

semgrep-core/src/parsing/Test_parsing.ml
```diff
@@ -29,10 +29,65 @@ let logger = Logging.get_logger [ __MODULE__ ]
 (* Helpers *)
 (*****************************************************************************)
 
+module SMap = Map.Make (String)
```
nmote: The changes to this file seem unrelated. Could you move these into a separate PR so that they can be reviewed separately?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -78,9 +78,15 @@ let map_bitwise_binary_operator (env : env) (x : CST.bitwise_binary_operator) =
...
-  | `Prefix tok -> (* "prefix" *) token env tok
+  | `Infix tok ->
+      (* "infix" *)
+      G.unhandled_keywordattr ("Infix", token env tok)
+  | `Post tok ->
+      (* "postfix" *)
+      G.unhandled_keywordattr ("Postfix", token env tok)
+  | `Prefix tok ->
+      (* "prefix" *)
+      G.unhandled_keywordattr ("Prefix", token env tok)
```
nmote: Instead of writing out the string yourself, `str env tok` should get you the pair of string and parse info that you need here.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -116,11 +139,20 @@ let map_assignment_and_operator (env : env) (x : CST.assignment_and_operator) =
...
+      G.unhandled_keywordattr ("Weak", token env tok)
+  | `Unow_7c8c304 tok ->
+      (* "unowned" *)
+      G.unhandled_keywordattr ("Unowned", token env tok)
+  | `Unow_e455cde tok ->
+      (* "unowned(safe)" *)
+      G.unhandled_keywordattr ("UnownedSafe", token env tok)
+  | `Unow_8fda70e tok ->
+      (* "unowned(unsafe)" *)
+      G.unhandled_keywordattr ("UnownedUnsafe", token env tok)
```
nmote: Look into the `str` helper for this too

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -139,9 +171,25 @@ let map_optionally_valueful_control_keyword (env : env)
...
-  | `Cont tok -> (* "continue" *) token env tok |> todo env
-  | `Brk tok -> (* "break" *) token env tok |> todo env
-  | `Yield tok -> (* "yield" *) token env tok |> todo env
+  | `Cont tok -> (
+      let tok = (* "continue" *) token env tok in
+      match expr with
+      | Some { G.e = N (Id (id, _)); _ } ->
+          G.Continue (tok, LId id, semi) |> G.s
+      (* Need to have an identifier to continue to *)
+      | Some { G.e = _; _ } -> raise Common.Impossible
```
nmote, line 181: Can you provide a stronger justification for why this is impossible? Maybe the grammar is overly permissive and Swift doesn't actually allow this, in which case a link to the spec or at least a specific code snippet that Swift rejects in practice?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -139,9 +171,25 @@ let map_optionally_valueful_control_keyword (env : env)
...
+      | None -> G.Continue (tok, LNone, semi) |> G.s)
+  | `Brk tok -> (
+      let tok = (* "break" *) token env tok in
+      match expr with
+      | Some { G.e = N (Id (id, _)); _ } -> G.Break (tok, LId id, semi) |> G.s
+      (* Need to have an identifier to continue to *)
+      | Some { G.e = _; _ } -> raise Common.Impossible
```
nmote, line 189: Same here

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -165,9 +213,15 @@ let map_boolean_literal (env : env) (x : CST.boolean_literal) =
...
-  | `Requ tok -> (* "required" *) token env tok
+  | `Over tok ->
+      (* "override" *)
+      G.attr G.Override (token env tok)
+  | `Conv tok ->
+      (* "convenience" *)
+      G.unhandled_keywordattr ("Convenience", token env tok)
+  | `Requ tok ->
+      (* "required" *)
+      G.unhandled_keywordattr ("Required", token env tok)
```
nmote: `str` helper here too

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -176,36 +230,66 @@ let map_try_operator (env : env) (x : CST.try_operator) =
...
 let map_mutation_modifier (env : env) (x : CST.mutation_modifier) =
   match x with
-  | `Muta tok -> (* "mutating" *) token env tok
-  | `Nonm tok -> (* "nonmutating" *) token env tok
+  | `Muta tok ->
+      (* "mutating" *)
+      G.unhandled_keywordattr ("Mutating", token env tok)
+  | `Nonm tok ->
+      (* "nonmutating" *)
+      G.unhandled_keywordattr ("Nonmutating", token env tok)
```
nmote: `str` here too

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -176,36 +230,66 @@ let map_try_operator (env : env) (x : CST.try_operator) =
...
+  | `Dyna tok ->
+      (* "dynamic" *)
+      G.unhandled_keywordattr ("Dynamic", token env tok)
+  | `Opt tok ->
+      (* Seems close enough to other Optional field uses.*)
+      (* "optional" *)
+      G.attr G.Optional (token env tok)
+  | `Class tok ->
+      (* "class" *)
+      G.unhandled_keywordattr ("Class", token env tok)
```
nmote: and here

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -176,36 +230,66 @@ let map_try_operator (env : env) (x : CST.try_operator) =
...
-  | `ATau tok -> (* "@autoclosure" *) token env tok
+  | `Inout tok ->
+      (* "inout" *)
+      G.unhandled_keywordattr ("Inout", token env tok)
+  | `ATes tok ->
+      (* "@escaping" *)
+      G.unhandled_keywordattr ("Escaping", token env tok)
+  | `ATau tok ->
+      (* "@autoclosure" *)
+      G.unhandled_keywordattr ("Autoclosure", token env tok)
```
nmote: and here

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -510,8 +632,12 @@ let map_non_constructor_function_decl (env : env)
   let v2 =
     match v2 with
     | `Simple_id x -> map_simple_identifier env x
-    | `Refe_op x -> map_referenceable_operator env x |> todo env
-    | `Bitw_bin_op x -> map_bitwise_binary_operator env x |> todo env
+    | `Refe_op x ->
+        map_referenceable_operator env x |> fst |> fun x -> (PI.str_of_info x, x)
```
nmote: We probably shouldn't be using `str_of_info` here. Maybe you can pass the original string up from `map_referenceable_operator` instead? Once again, the `str` function will let you extract that from the original CST token.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -510,8 +632,12 @@ let map_non_constructor_function_decl (env : env)
...
+    | `Bitw_bin_op x ->
+        (* Maybe come back and do this better? *)
+        let _, tok = map_bitwise_binary_operator env x in
+        (PI.str_of_info tok, tok)
```
nmote: Same here

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -527,9 +653,10 @@ let map_getter_specifier (env : env) ((v1, v2, v3) : CST.getter_specifier) =
     | Some x -> map_getter_effects env x |> todo env
     | None -> ()
   in
-  (v2, `Get)
+  v2
 
 let map_availability_argument (env : env) (x : CST.availability_argument) =
+  (* This does not seem important semantically. *)
```
nmote: If you're going to make this claim, could you justify it?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -579,128 +710,165 @@ let rec map_annotated_inheritance_specifier (env : env)
...
-      todo env (v1, v2, v3)
+      (* This is only kind of correct. It's actually a subclassing thing. *)
+      PatTyped (v1, v3)
 
 and map_computed_getter (env : env) ((v1, v2, v3) : CST.computed_getter) =
   let v1 = Common.map (map_attribute env) v1 in
   let v2 = map_getter_specifier env v2 in
+  (* Yes, I know, but I want them in the same order.
+   *)
+  let attrs = v1 @ [ G.unhandled_keywordattr ("Get", v2) ] in
```
nmote: `str`

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -579,128 +710,165 @@ let rec map_annotated_inheritance_specifier (env : env)
...
       let v3 = map_type_ env v3 in
-      todo env (v1, v2, v3)
+      (* This is only kind of correct. It's actually a subclassing thing. *)
+      PatTyped (v1, v3)
 
 and map_computed_getter (env : env) ((v1, v2, v3) : CST.computed_getter) =
   let v1 = Common.map (map_attribute env) v1 in
   let v2 = map_getter_specifier env v2 in
+  (* Yes, I know, but I want them in the same order.
+   *)
```
nmote: What?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -579,128 +710,165 @@ let rec map_annotated_inheritance_specifier (env : env)
...
+          G.fbody = v3;
+        } )
+  |> G.s
 
 and map_computed_modify (env : env) ((v1, v2, v3) : CST.computed_modify) =
   let v1 = Common.map (map_attribute env) v1 in
   let v2 = map_modify_specifier env v2 in
-  let v3 =
+  (* Yes, I know, but I want them in the same order.
+   *)
```
nmote: Again, what?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -579,128 +710,165 @@ let rec map_annotated_inheritance_specifier (env : env)
...
+  let attrs = v1 @ [ G.unhandled_keywordattr ("Modify", v2) ] in
```
nmote: `str`

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -579,128 +710,165 @@ let rec map_annotated_inheritance_specifier (env : env)
...
-  todo env (v1, v2, v3)
+      in
+      G.Block (v1, getters_setters, v3) |> G.s
 
 and map_computed_setter (env : env) ((v1, v2, v3, v4) : CST.computed_setter) =
   let v1 = Common.map (map_attribute env) v1 in
   let v2 = map_setter_specifier env v2 in
-  let v3 =
+  (* Yes, I know, but I want them in the same order.
+   *)
```
nmote: Same

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -579,128 +710,165 @@ let rec map_annotated_inheritance_specifier (env : env)
...
+  let attrs = v1 @ [ G.unhandled_keywordattr ("Set", v2) ] in
```
nmote: `str`

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1168,58 +1361,50 @@ and map_else_options (env : env) (x : CST.else_options) =
...
 
-and map_enum_class_body (env : env) ((v1, v2, v3) : CST.enum_class_body) =
+and map_enum_class_body (enum_ident : G.ident) (env : env)
+    ((v1, v2, v3) : CST.enum_class_body) =
   let v1 = (* "{" *) token env v1 in
-  let v2 =
+  let v3 = (* "}" *) token env v3 in
+  (* If it's raw, we expect initialized fields, and not constructors.
+     Vice versa for the opposite.
+  *)
```
nmote: Is this comment out of date?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1361,29 +1550,41 @@ and map_function_value_parameters (env : env)
...
+  |> G.s
+
+and combine_conds cond conds =
+  List.fold_left
+    (fun acc x ->
+      G.Call
+        ( G.IdSpecial (G.Op G.And, G.fake "&&") |> G.e,
+          G.fake_bracket [ G.Arg acc; G.Arg x ] )
+      |> G.e)
+    cond conds
```
nmote, line 1586: We've discussed this separately, but could you add a comment here justifying this desugaring?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1556,41 +1761,46 @@ and map_labeled_statement (env : env) ((v1, v2) : CST.labeled_statement) =
   | None -> v2
   | Some ident -> G.Label (ident, v2) |> G.s
 
+(* This is a misleading name, it has nothing to do with the lambda's type... *)
```
nmote: What is it instead?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1635,85 +1841,91 @@ and map_lambda_literal (env : env) ((v1, v2, v3, v4, v5) : CST.lambda_literal) :
...
     | `Opt_simple_id_simple_id_COLON_opt_param_modifs_poss_impl_unwr_type
         (v1, v2, v3, v4, v5) ->
-        let v1 =
+        (* TODO As with `map_parameter`, we only support internal names right now. *)
+        let _v1 =
           match v1 with
-          | Some x -> map_simple_identifier env x
-          | None -> todo env ()
+          | Some x ->
+              let _todo = map_simple_identifier env x in
```
nmote: Just pipe this to `todo env` so that it doesn't get overlooked if we encounter this?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -908,18 +1095,24 @@ and map_block (env : env) ((v1, v2, v3) : CST.block) =
...
   in
+  (* I'm not sure this does anything. "class" is already a possible modifier
+     from the `map_modifiers` - it's a property modifier.
+  *)
   let v2 =
     match v2 with
-    | Some tok -> (* "class" *) token env tok |> todo env
+    | Some tok ->
+        (* "class" *)
+        let _todo = token env tok in
```
nmote: Pipe this to `todo env` so that it doesn't get overlooked if we do encounter it?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -1822,52 +2035,61 @@ and map_modifierless_class_declaration (env : env)
...
-        | None -> ()
-      in
       let v2 = (* "enum" *) token env v2 in
-      let attrs = [ G.KeywordAttr (G.EnumClass, v2) ] in
+      let attrs =
+        let indirect_kw =
+          match v1 with
+          | Some tok ->
+              (* "indirect" *)
+              [ G.unhandled_keywordattr ("Indirect", token env tok) ]
```
nmote: `str`

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -2702,17 +2952,26 @@ and map_type_constraint (env : env) (x : CST.type_constraint) =
       let v2 = map_identifier env v2 in
       let v3 = (* ":" *) token env v3 in
       let v4 = map_possibly_implicitly_unwrapped_type env v4 in
-      todo env (v1, v2, v3, v4)
+      let base_type = G.ArgType (G.TyN (H2.name_of_ids v2) |> G.t) in
+      let conformed_protocol = G.ArgType v4 in
+      G.NamedAttr
+        ( v3,
+          H2.name_of_id ("ConformsTo", v3),
```
nmote: Probably better to use the string in the original program text via `str` even though it's just a colon.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -2702,17 +2952,26 @@ and map_type_constraint (env : env) (x : CST.type_constraint) =
...
       let v3 =
         match v3 with
         | `Equal_sign tok -> (* eq_custom *) token env tok
         | `Eq_eq tok -> (* eq_eq_custom *) token env tok
       in
       let v4 = map_type_ env v4 in
-      todo env (v1, v2, v3, v4)
+      G.NamedAttr
+        ( v3,
+          H2.name_of_id ("EqualTypes", v3),
```
nmote: Same here

nmote (changes requested):
Overall this looks good! I do have several requests which I made inline that I'd like you to address before merging, but this is great progress towards experimental Swift support!

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
brandonspark: yes!

nmote (approved):
Nice! Looking forward to seeing this bump the Swift parsing stats!

PR #5872, opened 2022-08-03
Prototype a utility for pattern parse stats

emjin:
It can be frustrating to write rules because it's hard to get a pattern that
even works. Sometimes the pattern extension isn't supported (e.g. `$X. ...$Y`).
Sometimes you don't understand what is required for a pattern (and sometimes
we allow partial expressions only for specific languages). Sometimes there's
just a pattern parse error. This makes it hard for users to write rules and
especially for new users to learn the tool.

One idea I had was to at least enforce that the patterns that are supposed
to parse, do parse. The ultimate goal of this utility would be to generate
valid patterns from a file's AST, including ones that require substituting
ellipses or other semgrep extensions, and running them as rules to see how
many parse. This would be a lot of work though and would generate a LOT of
patterns unless we found a solution (one idea: only generate one for each
AST construct?).

In the meantime, this is a quick experiment that just extracts expressions
from a target AST and tries to use those as patterns. This already takes a
fair amount of time because currently a single bad pattern makes Semgrep stop
trying to parse a file, so I run each pattern as an individual rule. Also, a
significant proportion of expressions that don't parse are missing tokens; if
the expression were correctly dumped it would probably parse. That might also
be a useful thing to measure, though!

Anyway, wanted to put up the PR so people could know what I was talking about
and give ideas. It's entirely possible to me that this doesn't make sense as
an approach. I didn't even bother to put up a file-level comment yet. It was
cool for me to explore though and just see what our rates looked like.

To run:

```
semgrep-core -lang <LANG> -pattern_parsing_stats <file/dir>
```

e.g.

```
semgrep-core -lang python -pattern_parsing_stats cli/src/semgrep/commands/
...
Success rate: 1689 / 2008 (84%), 38 didn't match, 281 failed to parse
```

(The patterns are expected to match the code they're taken from, in addition to parsing)

PR checklist:

### DIFF
--- a/semgrep-core/src/cli/Main.ml
+++ b/semgrep-core/src/cli/Main.ml
@@ -208,6 +208,7 @@ let json_of_v (v : OCaml.v) =
 
 let dump_v_to_format (v : OCaml.v) =
   match !output_format with
+  | No_output -> ""
   | Text -> OCaml.string_of_v v
   | Json -> J.string_of_json (json_of_v v)
 
@@ -382,6 +383,12 @@ let all_actions () =
           Test_parsing.parsing_stats
             (Xlang.lang_of_opt_xlang !lang)
             ~json:(!output_format = Json) ~verbose:true xs) );
+    ( "-pattern_parsing_stats",
+      " <files or dirs> generate pattern parsing statistics",
+      Common.mk_action_n_arg (fun xs ->
+          Pattern_parse_stats.get_pattern_parse_stats mk_config
+            (Xlang.lang_of_opt_xlang !lang)
+            xs) );
     (* the dumpers *)
     ( "-dump_extensions",
       " print file extension to language mapping",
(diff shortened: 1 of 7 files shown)


emjin:
Looking for thoughts to decide if I should close the PR or merge it

nmote:
This seems like a good thing to me. Ideally we'd run this over some corpus on a regular basis and record the results somewhere. If it's just going to get merged and then forgotten about I don't think it's worth the maintenance burden, but it could be a great tool to help drive up pattern parse rates.

aryx:
My main concern is that for many languages we still use the pfff parser for the pattern, and it's actually ok to me if we don't allow any kind of expressions in the pattern because people rarely use the latest feature of a language when looking for bugs.
So this tool would report lots of issues that I'm not sure are worth fixing.

IagoAbal:
I pretty much agree with what Nat said here. I add that if we truly want to make writing custom rules easy, then we should definitely be monitoring these issues. I believe this a relatively common pain suffered by the SR team. To address Yoann's concern, maybe we could skip languages for which we use the Pfff parsers. Note however that it's not just a problem with the latest features, e.g. Lewis recently reported that the pattern `switch (...) { ... }` doesn't work for JS.

emjin:
Sounds good! I'll leave this open for now and come back later to see if I can make it more efficient and get some regular runs set up. This will also be better to come back to once we start improving autofix, since it tests pattern parsing given autofix works.

PR #5877, opened 2022-08-03
fix: allow matching of simple identifier to metavariable

brandonspark:
**What:**
Currently, we are having an issue where we cannot match a metavariable to a decorator in Python. See PA-1700.

The root of this issue is that components of qualified identifiers are matched via `m_ident`, since they do not possess `id_info`s, and are instead just `ident`s. `m_ident` automatically assigns an `id_info` of `None` when matching against a metavariable, which results in a failure in `envf`, and `equal_ast_binded_code`.

**Why:**
SR would like us to be able to pass this test: https://semgrep.dev/s/9pJz

**How:**
I added a case to `equal_ast_binded_code` that permits matching two `Id`s, where one does not have an `id_info` and the other does, so long as they are the same literal string. To my complete astonishment, this actually passes the tests, and broke nothing.

I'm unsure if there is a reason for there to be an `Id` without an `id_info` other than originating from a simple `Id`. cc @aryx maybe?

**Alternatives considered:**
Something else we could do is just change `AST_generic` such that things like components to qualified identifiers, or `NamedAttributes`, possess `id_info`s and are full-fledged names in their own right. Since this is rather an involved change, wanted to propose this first.

**Test plan:**
`make test`

**Fixes:** PA-1700

PR checklist:


changelog.d/pa-1700.fixed
```diff
@@ -0,0 +1,2 @@
+All: Applied a fix so that qualified identifiers can match to metavariables. Notably, this
```
IagoAbal: They could match already no? The issue is about unification of code variables.

semgrep-core/src/matching/Matching_generic.ml
```diff
@@ -251,6 +251,7 @@ let rec equal_ast_binded_code (config : Config_semgrep.t) (a : MV.mvalue)
         MV.E { e = B.L b_lit; _ } )
       when config.constant_propagation ->
         G.equal_literal a_lit b_lit
+    | MV.Id ((s1, _), None), MV.Id ((s2, _), Some _) -> s1 = s2
```
IagoAbal, line 260: This may deserve a comment to explain why we need this case. It could also include a reference to the ticket.

IagoAbal (approved):

IagoAbal:
This would also help us handle this example https://semgrep.dev/s/NWPp by @Denis-Podgurskii.

PR #5905, opened 2022-08-10
Parse function-like language constructs as specials

Sjord:
Primarily to reduce the differences between tree-sitter and pfff parsing.

PR checklist:

### DIFF
--- a/changelog.d/gh-5382.changed
+++ b/changelog.d/gh-5382.changed
@@ -1,0 +1,1 @@
+Parse several built-in PHP functions in the same way in pfff and tree-sitter. This makes it possible to match exit, eval, empty and isset, even if the pattern is parsed with pfff and the PHP file with tree-sitter.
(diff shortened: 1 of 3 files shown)


Sjord:
https://github.com/returntocorp/pfff/pull/544

Sjord:
I tried [creating automatic tests for this](https://github.com/Sjord/semgrep/blob/test_equivalency/semgrep-core/src/parsing/Unit_parsing.ml#L44-L56), which would compare the AST of tree-sitter and pfff. However, naively comparing the ASTs with `pfff_ast <> ts_ast` didn't work. Perhaps because of differing `s_id` and `id_info_id` values? The semgrep-core `-diff_pfff_tree_sitter` option also runs into that.

The functions `empty`, `eval`, `exit` and `isset` used to differ between pfff and tree-sitter. This PR makes the two parser exhibit the same behavior. This can be tested with `semgrep-core -lang php -diff_pfff_tree_sitter funclikes.php`, where `funclikes.txt` looks something like this:

```
<?php

empty($a);
eval($a);
exit($a);
isset($a);
```

I chose to leave out `echo` and `print`, because these don't require parentheses and are currently already parsed the same.

aryx (approved):

Issue #5854, opened 2022-08-01
pattern-not can negate a formula


emjin:
**Is your feature request related to a problem? Please describe.**
Pattern-not can only have one pattern, not a formula

**Describe the solution you'd like**
Pattern-not takes a formula

**Describe alternatives you've considered**
Rewrite rule

**Use case**
Make it less annoying to negate a rule

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1723/pattern-not-can-negate-a-formula. _Note_: this link is for r2c use only and is not accessible publicly.

IagoAbal:
Can we do this with the rule schema used by CLI? Or do we need a `pattern-neg`?

kopecs:
This is possible with the rule schema; it would be something like:

```yml
  pattern-not:
    type: object
    properties:
      pattern-not:
        title: Do not return finding where Semgrep pattern matches exactly
        oneOf:
          - type: string
          - anyOf: <similar to https://github.com/returntocorp/semgrep-interfaces/blob/develop/rule_schema.yaml#L5-L23>
    required:
      - pattern-not
```

IagoAbal:
Related to #5840

IagoAbal:
We should try doing this, it's probably cheap nowadays.

IagoAbal:
Closed by #6474

Issue #5840, opened 2022-08-01
`pattern-not` can't declare metavariables?


zhengow:
**Describe the bug**
<img width="1384" alt="image" src="https://user-images.githubusercontent.com/38389586/182140342-4c196443-0973-4191-851d-f73f0698335d.png">

```
id: untitled_rule
message: Semgrep found a match
languages: [python]
severity: WARNING
patterns: 
    - pattern: ...
    - pattern-not: $X = 1
    - metavariable-pattern:
        metavariable: $X
        pattern: ...
```

```
a = 1
b = 1
c = 1
```

**To Reproduce**
Steps to reproduce the behavior, ideally a link to https://semgrep.dev:

**Expected behavior**

**Screenshots**

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?

**Use case**
What will fixing this bug enable for you?

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1705/pattern-not-cant-declare-metavariables. _Note_: this link is for r2c use only and is not accessible publicly.

IagoAbal:
@zhengow no unfortunately that's not currently allowed, but if you have a more concrete rule to write let us know, you can usually write it in some other way.

Issue #4647, opened 2022-02-09
Enable metavariable-* clauses with pattern-not* clauses


minusworld:
**Is your feature request related to a problem? Please describe.**
`metavariable-*` clauses do not seem to work with `pattern-not*` clauses. 

**Describe the solution you'd like**
In particular, I want to NOT match on line 73, so I tried this: https://semgrep.dev/s/wXp6
However, the rule has no findings at all. If I remove the metavariable-regex clause, it works, but will match any if statement which isn't exactly what I want: https://semgrep.dev/s/xXxj

**Describe alternatives you've considered**

**Use case**
I've had to work around `metavariable-*` not working with `pattern-not` for a while; this time I am unsure of another way to solve the issue without making several copies of the rule.

Priority: P1/Medium (annoying bug that should get fixed)

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-849/enable-metavariable-clauses-with-pattern-not-clauses.

emjin:
This is intentional; metavariable-xyz operators can only work with metavariables that occur in positive patterns. The best solution would probably be to make `pattern-not-inside` accept a formula. Something like

```
pattern-not-inside:
  - pattern: |
      if $COND:
        ...
      ...
  - metavariable-regex:
      metavariable: $COND
      regex: (i?).*nan.*
```

I think ^ is definitely worth considering, but we're probably not going to prioritize it.

cokeBeer:
Any update ?

Issue #4484, opened 2022-01-06
[Terraform] metavariable-regex does not match properly when used in conjunction with pattern-not


enncoded:
**Describe the bug**
`metavariable-regex` used in conjunction with `pattern-not` does not produce any matches.

**To Reproduce**
<https://semgrep.dev/s/lejliu:mv-regex-pattern-not-test>

**Expected behavior**
`$KEY` should match `aws_kms_key.test.arn` in the first test and `bad` in the second test.

However, `metavariable-regex` does work on `pattern`, as seen here <https://semgrep.dev/s/lejliu:mv-regex-pattern-test>

**What is the priority of the bug to you?**
- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

inkz:
Similar issue but for Python code: https://semgrep.dev/s/abuammar:pattern-not-inside
from our user: https://r2c-community.slack.com/archives/C018NJRRCJ0/p1643230967051000

PR #5916, opened 2022-08-12
feat!: new pattern operators

### LINKED ISSUE #1489: RFC: Simplify pattern matching API

kopecs:
**What:** 
You know what they say, the most important thing about a language is its concrete syntax.

Alright, so this is the big one. This PR concerns the changing of the rule syntax from its previous form to a more standardized, disciplined form, which hopes to be less confusing to users, more intuitive to use, and enable a few features along the way.

-----------------------------------------------------------------------------------

**Why:**
Why should we do this, I hear you asking?

So, the old syntax is somewhat verbose. Needing to remember concrete syntax such as `patterns`, `pattern-either`, including very long names for keys such as `metavariable-comparison`, `metavariable-analysis`, etc. Additionally, the necessity of patterns to fall under a `pattern` key means that we often have quite drastic duplication of `pattern`, such as underneath a `patterns`. Moreover, `pattern` and `patterns` remain dubiously visually distinct.

There are a few issues already discussing this change:
https://github.com/returntocorp/semgrep/issues/1489
https://github.com/returntocorp/semgrep/issues/5840
https://github.com/returntocorp/semgrep/issues/4647
https://github.com/returntocorp/semgrep/issues/4484
https://github.com/returntocorp/semgrep/issues/5854

The first is a proposal to switch to this new syntax, the rest are problems that are related to the fact that we do not have it.

There are two main claims to be had here: one, this syntax will make adoption of Semgrep easier for users. Two, it will add genuine new features which make Semgrep easier/more intuitive to use.

-----------------------------------------------------------------------------------

**How:**

**A New Syntax**
We have changed to a new concrete syntax.
Notably, there are only a few constructs. An artist's rendition of the new rule syntax's EBNF is pictured below:
image
It fits on a whiteboard!

**Base Patterns**

Notably, we have `inside`, `not`, `and`, `or`, and `regex` as "base" patterns. `not` can also now take in entire formulas, rather than being constrained to "base patterns". `pattern` is around too, but now it's pretty much optional.

This means we can now write rules like the following:
```
id: ey-look-an-id
match:
  and:
  - "a"
  - "b"
message: ey look a message
languages:
  - python
severity: WARNING
```

The pattern appears underneath a `match:`, which signals that we are using the new syntax. 

**Backwards Compatibility**

We remain fully backwards compatible, so a top-level `patterns`, `pattern`, or whatever else was allowed before will trigger a parse via the old syntax, into a common formula language Otherwise, we use the new one.

**`pattern` Is Optional**

The nice thing is that standalone patterns may directly be given into the list of a combinator pattern, such as `and`. Previously, you would have to write 
```
patterns:
- pattern: "a"
- pattern: "b"
- pattern: "c"
- pattern: "d"
```
and now you can write
```
and:
- "a"
- "b"
- "c"
- "d"
```

**Where-clauses**

Additionally, we support more explicit modifiers on patterns. Previously, things like `focus-metavariable` had to be combined with other patterns underneath a `patterns`. This was effectively the same as just enforcing some constraint on the intersection of the other patterns given to the `patterns`, but this meant that it had to be explicitly checked for being under a `patterns`, as well as just generally being harder to intuitively think about.

Now, we support syntax like the following:
```
match:
  and:
  - "a"
  - "b"
  where:
  - comparison: $A > 5
  - focus: $A
  - metavariable: "A"
    pattern: "A"
```

The `where` (which must exist in a two-dictionary with some other pattern) now directly modifies the other pattern it is shared with. Additionally, we use terser keys such as `comparison` and `focus`. `metavariable-comparison` also used to demand that you explicitly name what the metavariable is, despite the fact that it is in the `comparison`, so we eliminated that as well.

**`not` as a Parent Pattern**
`not` is now allowed to take in a pattern, instead of a string to match. This is huge for the expressivity of the rules language, since before we had to deal with `pattern-not`, `pattern-not-inside`, and others. Not only is this cumbersome syntactically, but there were some things we could not express, cf. the above issues.

**Migration**

There is also a translation command you can run via `sc -translate_rules`, which can help users migrate their rulebases to the new syntax automatically. We could also use it to migrate the Registry.

-----------------------------------------------------------------------------------

**Concerns:**
- Documentation does not exist yet
- Translation from old-style rules to new-style rules is strictly programmatic and may result in rules which are more complicated than a human may write.

-----------------------------------------------------------------------------------

Closes #1489

PR checklist:

- [ ] Documentation is up-to-date (TODO)


semgrep-core/tests/OTHER/rules/metavar_regex.yaml
```diff
@@ -5,8 +5,8 @@ rules:
   match:
     and:
```
underyx, line 6: @kopecs in imperative languages I see these names for equivalent operations on arrays:

| Language                | AND operation                    | OR operation                     | Notes                                                                                                 |
|-------------------------|----------------------------------|----------------------------------|-------------------------------------------------------------------------------------------------------|
| Python                  | `all(predicate for i in [a, b])` | `any(predicate for i in [a, b])` |                                                                                                       |
| JavaScript / TypeScript | `[a, b].every(predicate)`        | `[a, b].some(predicate)`         |                                                                                                       |
| Java                    | `[a, b].allMatch(predicate)`     | `[a, b].anyMatch(predicate)`     | (this is with something called Stream? I dunno java)                                                  |
| Ruby                    | `[a, b].all? {predicate}`        | `[a, b].any? {predicate}`        |                                                                                                       |
| PHP                     |                                  |                                  | has nothing lol                                                                                       |
| Golang                  | `EveryBy({a, b}, predicate)`      | `SomeBy({a, b}, predicate)`     | also has nothing, but [a popular utility belt library](https://github.com/samber/lo) uses these names |
| Kotlin                  | `arrayOf(a, b).all(predicate)`   |  `arrayOf(a, b).any(predicate)`  |                                                                                                       |
| Scala                   | `Vector(a, b).forall(predicate)` | `Vector(a, b).exists(predicate)` |                                                                                                       |

In all of these languages, the `and` and `or` keywords are used only as binary operators.

Since these are the top eight programming languages our users scan (multiple hundreds of times more popular than all other languages combined), would you be open to renaming `and/or` in our rule schema to `all/any`?

mmcqd: I like `all` and `any`. They appear in Haskell too. They also read more nicely with `match`. 
```
match:
  any:
```
reads like "match any of these patterns"

brendongo: +1 to all/any

underyx: I also found the original discussion when any/all were being added to Python where this naming vs or/and was discussed:

https://mail.python.org/archives/list/python-dev@python.org/thread/K33F5RHPA6UTPU3SIYUMTVM4B7P3JUGC/#N5EBG3QLC4WS3H4Z4WLRFEBXHI4QLQYA

https://www.artima.com/forums/flat.jsp?forum=106&thread=98196

kopecs: Yeah, we talked a little bit about names, and I'm not in love with the exact wording we've got in here.

> in imperative languages I see these names for equivalent operations on arrays:

I don't think any of the examples you have here are analogous---they're all about applying one fixed predicate to array elements, whereas this is essentially a combinator on an array of predicates.

> In all of these languages, the and and or keywords are used only as binary operators.

In any event, I don't think anyone is particularly confused if these are binary operations or not. I've not seen any confusion in our current syntax relating to this, and `pattern-either` is far more indicative of only being a binary operator imo; more broadly, for associative operations I don't think this is a concern---I've never heard the response to "add these 3 things together" be "wait a second! addition is a binary operation, so you can't add 3 things".

> would you be open to renaming and/or in our rule schema to all/any?

and/or vs all/any is mostly the same to me, with maybe a slight preference for the former because I can't read good and they're a bit more visually distinct. If others have a strong preference, happy to change it.

The issue I have with both of these options is that they're not super reflective of the actual operation performed: `and`/`patterns`/whatever doesn't actually (despite what our documentation claims) perform a boolean and operation, nor does it check that "all" of the patterns match exactly---it results in the smallest range in the target file where all the child patterns match some range which is contained/contained by all of the other combined ranges. This is less of an issue for `or`/`patterns-either`/whatever, because its just choice.

underyx: >and/or vs all/any is mostly the same to me, maybe a slight preference for the former because I can't read good and they're a bit more visually distinct. If others have a strong preference, happy to change it.

every/some are a visually distinct option, but those words sound more obscure and arbitrary to me, plus it's extra characters

I also wanna mention ElasticSearch's query DSL, which is the closest reference I can come up with, since our rule syntax is sort of a query DSL as well. They use what I think is a frankly absurd set of keywords: [filter/must/should/must_not](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html). I wouldn't really wanna copy this, it was always annoying me as a user of ES. To be fair, they have an extra requirement of also using this for scoring results, which might have to do with the weird keywords, but still.

For the record I never had issues with misreading any/all in python code; they're not all that similar in fixed width font. Observe, a ridiculous illustration, to semantically satiate you:

CleanShot 2022-08-17 at 18 54 19@2x

It is worthwhile to think about accessibility for dyslexic users though. 💡 I wonder if it'd be a worthwhile addition to our LSP server to add `&&` and `||` as inlay hints between the array elements?

underyx: Meta comment: I dislike the idea that we're calling this bikeshedding. I think this is a serious language design question that's worth discussing if we're gonna release a Semgrep v1.

underyx: And if a PR trying to make our pattern operator names better is not the right place for discussing better pattern operator names, I'm not sure what is :D

underyx: [worldssmallestviolin.avi](https://www.youtube.com/watch?v=rAlTOfl9F2w)

emjin: I agree that this is important; I don't think we should decide by talking and theorizing. In terms of actual feedback, I was there when Spencer was interviewed and he understood `and/or` without any trouble. Obviously we need some more data points for both options to conclude whether one is better.

underyx: My point though is not that all/any is more understandable than and/or, but that it's more readable (reduces cognitive load while reading and writing rules). So we'd need to design the experiment around measuring cognitive load, and the anecdote about spencer is not too relevant.

kopecs: Can you disambiguate what you see as the difference between "understandable" and "readable" here @underyx? I don't understand how you're differentiating them with respect to cognitive load.

Some other options to throw in the mix, just so we don't collapse to just these two ideas too early:
- For `patterns`:
  - `common` (because we're really getting the range common to all the patterns)
  - `intersect` (a bit literal, but we're \~kinda~ doing an intersection, with the constraint that ranges are contained within another)
- For `pattern-either`:
  - `choice` or `choose` (common in parser combinator libraries, e.g., [Parsec](https://hackage.haskell.org/package/parsec-3.1.15.1/docs/Text-Parsec-Combinator.html#v:choice); [racket's Parsack](https://docs.racket-lang.org/parsack/index.html#%28def._%28%28lib._parsack%2Fmain..rkt%29._choice%29%29); [chumsky](https://docs.rs/chumsky/latest/chumsky/primitive/fn.choice.html)), 
  - `union` (the produced set of ranges is the union of the results of the children patterns)

underyx: sure, what I meant is:

- **I expect `all/any` and `and/or` are equally understandable:**  
  99 of 100 readers would correctly explain what a rule does with either option (and this is what we tested on spencer)
- **I expect `all/any` is more readable than `and/or`:**  
  the average time taken to correctly explain what a rule does would be shorter in the `all/any` case compared to `and/or`. I also expect the gap between the two options expands with every level of nesting that the rule has.

underyx: To illustrate with some fake made-up data, I think if we had read access to profiling data from a test subject's brain, if we showed someone a 20-line rule which has four levels of nesting of `and`, `or`, `or`, `and` operators nested within each other, they'd comprehend the rule in 3600ms, while if we showed the same rule with `all`, `any`, `any`, `all`, they'd comprehend it in 3500ms instead.

I also think that a language that's easier to read will often also be easier to write.

semgrep-core/src/core/Rule.ml
```diff
@@ -55,15 +55,16 @@ type 'a loc = {
  *
  * less? enforce invariant that Not can only appear in And?
  *)
-type formula =
+and formula =
```
aryx: Why this and here? Why does it need to be mutually recursive with the loc type?

cli/src/semgrep/core_runner.py
```diff
@@ -834,7 +849,9 @@ def validate_configs(self, configs: Tuple[str, ...]) -> Sequence[SemgrepError]:
 
         parsed_errors = []
 
-        with tempfile.NamedTemporaryFile("w", suffix=".yaml") as rule_file:
+        with tempfile.NamedTemporaryFile(
+            "w", suffix=".yaml", delete=False
```
aryx: was this 'delete=False' used just for debugging?

cli/src/semgrep/core_runner.py
```diff
@@ -852,6 +869,8 @@ def validate_configs(self, configs: Tuple[str, ...]) -> Sequence[SemgrepError]:
                 + list(configs)
             )
 
+            print(f"{' '.join(cmd)}")
```
aryx: Same, why this new print?

semgrep-core/src/core/Rule.ml
```diff
@@ -77,6 +78,7 @@ type formula =
  * See also split_and().
  *)
 and conjunction = {
+  conj_tok : tok;
```
aryx: I'd rather keep it attached to the And, to be more symetric with how we represent Or of tok * ...

semgrep-core/src/core/Rule.ml
```diff
@@ -85,11 +87,6 @@ and conjunction = {
   focus : (tok * MV.mvar) list;
 }
 
-(* todo: try to remove this at some point, but difficult. See
```
aryx, line 88: Can you put back this comment next to the Inside case in formula?

aryx: We still want to get rid of this Inside in the long term.

semgrep-core/src/engine/Dataflow_tainting.ml
```diff
@@ -216,7 +216,7 @@ let findings_of_tainted_sink env taints (sink : T.sink) : T.finding list =
   let ( let* ) = Option.bind in
   let labels = labels_in_taint taints in
   let sink_pm, ts = T.pm_of_trace sink in
-  let req = eval_label_requires ~labels ts.requires in
```
aryx: Why the moving of the files to another directory?

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -496,62 +497,99 @@ and nested_formula_has_matches env formula opt_context =
...
+          *  (match pos with
+          *  | [] -> failwith "empty And; no positive terms in And"
+          *  | start::pos ->
+          *     let res = evaluate_formula env start in
+          *    let res = pos |> List.fold_left (fun acc x ->
+          *      intersect_ranges acc (evaluate_formula env x)
+          * ...
+      *)
+
+      (* This code used to live in Specialize_formula.
```
aryx: It would be better to move this code outside this function, or maybe in another file still, because the enclosing function code gets really really long otherwise.
I generally try to avoid nested functions. It makes the code more difficult to understand, and it's hard to know
if the nested functions are pure or are using enclosing vars.

semgrep-core/src/metachecking/Check_rule.ml
```diff
@@ -294,3 +293,153 @@ let stat_files fparser xs =
                     let s = Semgrep_prefilter_j.string_of_formula f in
                     pr2 (spf "regexp: %s" s)));
   pr2 (spf "good = %d, no regexp found = %d" !good !bad)
+
+let rec expr_to_string expr =
```
aryx: What is this for?
We already have some AST_generic.show_expr function.

semgrep-core/src/metachecking/Check_rule.ml
```diff
@@ -294,3 +293,153 @@ let stat_files fparser xs =
...
+  let { AST_generic.e_range; _ } = expr in
+  match e_range with
+  | Some (start, end_) ->
+      let source_file = open_in_bin start.file in
+      let extract_size = end_.charpos - start.charpos in
+      seek_in source_file start.charpos;
+      really_input_string source_file extract_size
+  | None -> failwith "invalid source/sink requires"
+
+and translate_metavar_cond cond : [> `O of (string * Yaml.value) list ] =
```
aryx: Looks like all this code is there to support -translate_rules.
This should be put in a separate file then.

semgrep-core/src/metachecking/Check_rule.mli
```diff
@@ -19,3 +19,6 @@ val check_files :
 (* -stat_rules *)
 val stat_files :
   (Common.filename -> Rule.t list) -> Common.filename list -> unit
+
+val translate_files :
```
aryx: Yes definitely should be in a different file.

semgrep-core/tests/OTHER/rules/metavar_cond.yaml
```diff
@@ -5,6 +5,8 @@ rules:
   match:
     and:
     - foo($ARG)
```
aryx: But then isn't it weird that you have a 'and' with just one element in it?

semgrep-core/tests/OTHER/rules/metavar_cond.yaml
```diff
@@ -5,6 +5,8 @@ rules:
...
-    - where: re.match(".*bar.*", $ARG)
+    where:
+    - metavariable: $ARG
```
aryx, line 8: I prefer my solution that generalizes where, because metavariable-regex: is redundant
when you support regular expression matching in metavariable-comparison:.
However I can see how it's more backward compatible with the way we were doing it before, so maybe it's
better your way.

aryx (changes requested):

aryx:
Overall I like the approach I think.

semgrep-core/tests/OTHER/rules/metavar_cond.yaml
brandonspark: This doesn't need to be a singular `and`. We could just use a `pattern`. I dunno why it's like that here.

semgrep-core/tests/OTHER/rules/metavar_cond.yaml
brandonspark: If I'm understanding you correctly, if you don't like the explicit naming of the metavariable, it could just be:
```
where:
- comparison: re.match(".*bar.*", $ARG)
```

We could also translate old-style `metavariable-regex` to that, and maintain backwards-compatibility.

aryx: yep

aryx:
There are lots of PRs in this PR:
 (1): getting rid of formula_old in Rule.ml and the conversion functions and move that in Parse_rule.ml where we can do the conversion directly (the reason formula_old was in Rule.ml was because I was not sure we would want to switch to formula_new at first, and I thought it was nice to also save in Rule.ml what actually people write (the old syntax)
 (2): refactor in formula_new where Inside is and move it out of P in its own construct
 (3): propose a new concrete syntax (revised what I proposed 1 year and a half ago but with inside: as a possible modifier
      and where: that generalize the metavariable-xxx but still allow those xxx under).
 (4): make taint a formula instead of a mode (fortunately this has been moved in a separate PR now)
 (5): add a -translate_rules to assist in converting rules from the old to new syntax
 (6): get rid of Specialize_formula sformula type
 (7): move around some files from tainting/ to engine/ (not sure why it's needed)

That's a lot. I understand it was a hack week project and all of those things are inter-related (and probably at first you didn't anticipate you would do certain things), but next time would be great if the PRs could be split in different parts
(note that I am guilty myself of those big PRs).

aryx:
This might close also https://github.com/returntocorp/semgrep/issues/4764

semgrep-core/tests/OTHER/rules/metavar_regex.yaml
IagoAbal: +1 to `all` and `any`.

aryx (approved):

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -757,71 +832,192 @@ and parse_extra (env : env) (key : key) (value : G.expr) : Rule.extra =
...
+      let cond =
+        (* if strip=true we rewrite the condition and insert Python's `int`
+            * function to parse the integer value of mvar. *)
+        match strip with
+        | Some true -> R.rewrite_metavar_comparison_strip cond
+        | _ -> cond
+      in
+      Left (t, R.CondEval cond)
+  | Cfocus ->
+      (* focus: ...
```
IagoAbal, line 915: Shouldn't you have filled the `...` here?

semgrep-core/tests/OTHER/rules/metavar_regex.yaml
mjambon: Isn't the main reason for Python using `all` and `any` that `and` and `or` are reserved keywords? This is possibly the case also in some of the other languages mentioned above. This surely contributes to the popularity of `all` and `any`. (anyway, using all/any is fine to me)

PR #5925, opened 2022-08-16
chore(fail-open): Update fail-open message body

zyannes:
PR checklist:

### DIFF
--- a/cli/src/semgrep/error_handler.py
+++ b/cli/src/semgrep/error_handler.py
@@ -34,7 +34,7 @@ class ErrorHandler:
             self.suppress_errors = suppress_errors
 
     def push_request(self, method: str, url: str, **kwargs: Any) -> None:
-        self.payload = {"method": method, "url": url, **kwargs}
+        self.payload = {"method": method, "url": url}
 
     def append_request(self, **kwargs: Any) -> None:
         self.payload = {**self.payload, **kwargs}
@@ -56,6 +56,7 @@ class ErrorHandler:
         Status will only be emitted if is_enabled is True
         """
         from semgrep.state import get_state  # avoiding circular import
+        from semgrep.app import auth  # avoiding circular import
 
         state = get_state()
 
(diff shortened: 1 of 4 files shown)


cli/src/semgrep/error_handler.py
```diff
@@ -72,13 +73,17 @@ def send(self, exit_code: int) -> int:
             f"Sending to fail-open endpoint {state.env.fail_open_url} since fail-open is configured to {self.suppress_errors}"
         )
 
-        headers = self.payload.get("headers", {})
-        headers["User-Agent"] = str(state.app_session.user_agent)
-        self.payload["headers"] = headers
+        token = auth.get_token()
+        headers = {
+            "User-Agent": str(state.app_session.user_agent),
+            "Authorization": f"Bearer {token or ''}",
```
spencerdrak, line 79: will this token be captured even if we fail early in the process?

cli/tests/unit/test_error_handler.py
```diff
@@ -35,35 +50,38 @@ def test_send_nominal(error_handler, mocker: MockerFixture) -> None:
     exit_code = error_handler.send(FATAL_EXIT_CODE)
 
     expected_payload = {
-        "headers": {"User-Agent": expected_user_agent},
         "method": "get",
         "url": "https://semgrep.dev/api/agent/deployments/current",
     }
 
+    expected_headers = {
```
spencerdrak, line 57: Are these the headers as SQS sees them or as they are in the body of the requet?

cli/src/semgrep/error_handler.py
zyannes: Yes. The token is pulled from an env variable (most commonly for CI) or a `.semgrep/settings.yml` file. So regardless of the failure, the token should be available (or worst case just returns `None`)

cli/tests/unit/test_error_handler.py
zyannes: These are actual headers that will be forwarded to the SQS.

brendongo (approved):

PR #5927, opened 2022-08-16
fix: make symbol prop matching less redundant

brandonspark:
**What**:
Currently, symbolic propagation results in redundant matches, which leads to it being disabled by default.

Consider the following Python program:

```py
def foo():
  x = g(5)
  f(x)
```

If we wanted to match this to the following rule:

```yaml
id: i-cant-be-bothered-to-think-of-a-name-for-this-rule-so-ill-put-this-except-ironically-this-ends-up-significantly-longer-than-if-i-had-just-not-bothered-to-type-this-at-all
options:
  symbolic_propagation: true
pattern: |
  g(5)
message: a match was found or something idk im not an engineer
languages:
  - python
severity: WARNING
```

we get the following matches:

```
test2.py:3 with rule i-cant-be-bothered-to-think-of-a-name-for-this-rule-so-ill-put-this-except-ironically-this-ends-up-significantly-longer-than-if-i-had-just-not-bothered-to-type-this-at-all
   f(x)
test2.py:2 with rule i-cant-be-bothered-to-think-of-a-name-for-this-rule-so-ill-put-this-except-ironically-this-ends-up-significantly-longer-than-if-i-had-just-not-bothered-to-type-this-at-all
   x = g(5)
```

Notably, the first match (on the third line of the program) is extraneous, because it's not particularly useful to match to every instance of x, because we already produced a match at the time that we declared x. Every future match to x is just reiterating information we already know.

However, consider the following case:

```yaml
id: i-cant-be-bothered-to-think-of-a-name-for-this-rule-so-ill-put-this-except-ironically-this-ends-up-significantly-longer-than-if-i-had-just-not-bothered-to-type-this-at-all
options:
  symbolic_propagation: true
pattern: |
  f(g(5))
message: a match was found or something idk im not an engineer
languages:
  - python
severity: WARNING
```
with program
```py
def foo():
  x = g(5)
  // MATCH: 
  f(x)
```
In this case, we do care to use the propagated definition of x - this is a useful match to produce. The distinction is that in the latter case, we had to "do work" to get to the identifier x. The issue that we should not care to produce matches to propagated identifiers if they are at the "top-level", that is, not resulting from decomposing on the target pattern. 

So if we start out with something of the form x ~ <pat>, we should always fail - even if we produce a match, this would inevitably be reiterating some information we already knew at the time of binding x. If we have something like f(x) ~ f(<pat>), it's totally fair game to use the definition of x, though. More generally, any time that we recurse into a code snippet (taking a layer off the pattern), then we should want to use the symbolic propagation information.

**Why:**
We should avoid redundant matches so that we can have symbolic propagation be the default, since it's quite handy to have.

**How:**
I added a wrapper for `m_expr`, effectively allowing it to either take in an `is_root : bool` argument or not. Almost all existing cases where `m_expr` is called is with `is_root` set to `false` - this is because we usually reach `m_expr` from recursing on an existing pattern, meaning we are not at the root.

Notably exceptions include `m_expr_deep`, which just finds a list of sub-expressions, and matches to all of them. This is done with the original pattern, meaning it should be considered to be at the root of the pattern.

Also, `m_expr` was exposed, and I have changed it so that both `m_expr_root` is exposed and used, since from outside the matching engine itself, this is used at the top-level of a pattern.

**Fixes:** PA-1772

PR checklist:

### DIFF
--- a/changelog.d/pa-1772.fixed
+++ b/changelog.d/pa-1772.fixed
@@ -1,4 +1,5 @@
-Changed symbolic propagation such that "redundant" matches are no longer reported as findings. For instance:
+Changed symbolic propagation such that "redundant" matches are no
+longer reported as findings. For instance:
 
 ```py
 def foo():
@@ -6,7 +7,8 @@ def foo():
   f(x)
 ```
 
-If we are looking for the pattern `g(5)`, we should not match on line 3, since we will match on line 2 anyways,
-and this is just repeating information that we already know.
+If we are looking for the pattern `g(5)`, we should not match on line 3,
+since we will match on line 2 anyways, and this is just repeating information that
+we already know.
 
 This patch changes it so that we do not match on line 3 anymore.
--- a/changelog.d/pa-1772.fixed
+++ b/changelog.d/pa-1772.fixed
@@ -1,0 +1,12 @@
+Changed symbolic propagation such that "redundant" matches are no longer reported as findings. For instance:
+
+```py
+def foo():
+  x = g(5)
+  f(x)
+```
+
+If we are looking for the pattern `g(5)`, we should not match on line 3, since we will match on line 2 anyways,
+and this is just repeating information that we already know.
+
+This patch changes it so that we do not match on line 3 anymore.
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -364,9 +364,12 @@ let m_deep (deep_fun : G.expr Matching_generic.matcher)
       in
       b |> sub_fun |> aux)
 
-let m_with_symbolic_propagation f b =
+let m_with_symbolic_propagation ~is_root f b =
   if_config
-    (fun x -> x.Config.constant_propagation && x.Config.symbolic_propagation)
+    (* If we are not at the root, then we permit recursing into substituted values. *)
+      (fun x ->
+      x.Config.constant_propagation && x.Config.symbolic_propagation
+      && not is_root)
     ~then_:
       (match b.G.e with
       | G.N (G.Id (_, { id_svalue = { contents = Some (G.Sym b1) }; _ })) ->
@@ -647,14 +650,14 @@ and m_id_info a b =
 (* Expression *)
 (*****************************************************************************)
 
-(* Deep expression matching extracts sub-expressions without decomposing on the pattern,
-   meaning that matching those sub-expressions should be considered as the root.
-*)
 and m_expr_deep a b tin =
   let symbolic_propagation = tin.config.Config.symbolic_propagation in
   let subexprs_of_expr =
     SubAST_generic.subexprs_of_expr ~symbolic_propagation
   in
+  (* Deep expression matching extracts sub-expressions without decomposing on the pattern,
+     meaning that matching those sub-expressions should be considered as the root.
+  *)
   m_deep m_expr_deep m_expr_root subexprs_of_expr a b tin
 
 (* possibly go deeper. When someone writes a pattern like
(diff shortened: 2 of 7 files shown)


changelog.d/pa-1772.fixed
```diff
@@ -0,0 +1,12 @@
+Changed symbolic propagation such that "redundant" matches are no longer reported as findings. For instance:
```
IagoAbal: I would appreciate hard wrapping very long lines, otherwise it's a bit of a hassle to review them. I don't know if GitHub could be configured to do soft wrapping? But anyways, it's also going to look nicer for those who read the raw changelog.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -647,12 +647,15 @@ and m_id_info a b =
 (* Expression *)
 (*****************************************************************************)
 
+(* Deep expression matching extracts sub-expressions without decomposing on the pattern,
```
IagoAbal: I would move this comment down to where you call `m_expr_root`.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -670,18 +673,36 @@ and m_expr_deep a b tin =
  *   - x = <expr>,
  *   - <call>(<exprs).
  * see SubAST_generic.subexprs_of_expr_implicit
+ *
+ * Also matches sub-expressions without decomposing, meaning it should match as the root.
```
IagoAbal: Same here.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -670,18 +673,36 @@ and m_expr_deep a b tin =
...
 and m_expr_deep_implict a b tin =
   let symbolic_propagation = tin.config.Config.symbolic_propagation in
   let subexprs_of_expr =
     SubAST_generic.subexprs_of_expr_implicit ~symbolic_propagation
   in
-  m_deep m_expr_deep_implict m_expr subexprs_of_expr a b tin
-
+  m_deep m_expr_deep_implict m_expr_root subexprs_of_expr a b tin
+
+(* This is allows the matching of expressions, where we are considered to not be
```
IagoAbal: "This is allows" ?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -670,18 +673,36 @@ and m_expr_deep a b tin =
...
+and m_expr_root a b = m_expr_inner ~is_root:true a b
+
+(* A wrapper for `m_expr`, which allows an `is_root` flag.
+   Any subsequent calls to `m_expr` will set it to false.
+*)
 (* coupling: if you add special sgrep hooks here, you should probably
  * also add them in m_pattern
  *)
-and m_expr a b =
+and m_expr_inner ~is_root a b =
```
IagoAbal: Nothing wrong with your solution really, just wanted to point out that you could have used an optional argument as in `m_expr ?(is_root=false)`. Did you avoid optional arguments for some reason in particular? Pure curiosity.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -752,7 +773,9 @@ and m_expr a b =
       (match H.name_of_dot_access a with
       | None -> fail ()
       | Some a1 -> m_name a1 b1)
-      >||> m_with_symbolic_propagation (fun b1 -> m_expr a b1) b
```
IagoAbal, line 755: May be a bit cleaner to pass `is_root` to `m_with_symbolic_propagation` and then use it in the condition of the `if_config` there.

semgrep-core/tests/OTHER/rules/symbol_prop_redundancy.py
```diff
@@ -0,0 +1,5 @@
+# the function is only necessary because propagation only runs inside of functions at the moment
+def foo():
+  # ruleid: symbol_prop_redundancy:
```
IagoAbal: The test runner probably doesn't care, but I guess the `:` at the end was a typo?

IagoAbal:
Nice work! Left some comments I would appreciate you to look into. I'll also request a review from Yoann.

IagoAbal:
This may be breaking a semgep-rules tests, my guess is that we just need to fix the text expectations. Can you also look into it?

semgrep-core/src/matching/Generic_vs_generic.ml
brandonspark: the reason is that im silly and i started out with an optional argument, wrapped it with two functions which made the optional argument pointless, then eliminated the optional argument

brandonspark: instead of just keeping the optional argument and keeping the name the same

IagoAbal (approved):
Awesome!

IagoAbal:
@aryx may have requests once he reviews the PR, but I think we are good to merge 👍

aryx (commented):
Fantastic!

PR #5928, opened 2022-08-16
refactor: use new pure GET and POST endpoints on semgrep-app

chmccreery:
PR checklist:

Closes APP-2001, APP-2051


cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
```diff
@@ -15,15 +15,15 @@ Findings:
         useless comparison to 5
 
          ▶▶┆ Autofix ▶ bar == 2
-          1┆ bar == 2
+          1┆ bar == 5
```
chmccreery, line 18: Autofix used to be "on" for this test because it was mocked and thus its value was truthy, even though by default it should have been False. The changes in this test and others like it are thus expected.

cli/src/semgrep/config_resolver.py
```diff
@@ -192,15 +198,20 @@ def _load_config_from_local_path(self) -> Dict[str, YamlTree]:
...
             self._config_path,
             headers={"Accept": "application/json", **self._extra_headers},
         )
-        if r.status_code == requests.codes.ok:
-            return r.content.decode("utf-8", errors="replace")
+        if resp.status_code == requests.codes.ok:
+            try:
+                rules_dict = resp.json()["rule_config"]
+                assert rules_dict is not None
+                return pyyaml.dump(rules_dict, Dumper=SafeDumper)  # type: ignore
```
underyx: moving my slack message here:

> by the way I see you're doing this to write our rules: pyyaml.dump(rules_dict)
do you think we need to use YAML here? if we just json.dump instead then maybe we can skip all the trouble of adding a dependency

underyx: JSON is also faster to dump and load, by the way. And YAML is compatible with everything JSON supports, so even if somewhere down do line we yaml.load the same contents, it'll still work. But I think everything downstream will load rules as JSON and fall back to YAML only if needed.

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-buildkite/results.txt
```diff
@@ -44,11 +44,10 @@ Findings:
 Scan environment:
   versions    - semgrep <MASKED> on python <MASKED>
   environment - running in environment buildkite, triggering event is pull_request
-  semgrep.dev - authenticated as org_name
+  server      - https://semgrep.dev
 
 Fetching configuration from semgrep.dev
-Adding ignore patterns configured on semgrep.dev as `--exclude` options: ()
```
chmccreery, line 60: This line also should never have been in these snapshots. It was only because the ignores were equal to a mocked-value which then evaluated truthy because it wasn't None or [].

cli/src/semgrep/config_resolver.py
chmccreery: Ok I looked into this and tried to use `json.dumps` instead. Unfortunately, it seems like all the schema validation stuff is done in Yaml and breaks with JSON. All the happy-path cases work fine with JSON, but the tests for bad_config fail because _validate() does not return any errors even if a required field like 'pattern' is missing from the JSON. We could try to re-write this to be JSON-compatible but that feels like scope-creep to me, so I'd rather merge with this extra dependency for now.

cli/Pipfile
```diff
@@ -47,3 +46,5 @@ importlib_resources = ">=1.4.0"
 [packages]
 semgrep = {editable = true,path = "."}
 jsonschema = "==4.6.1"
+pyyaml = "*"
```
brendongo: Should remove these from here

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines/meta.json
```diff
@@ -19,5 +19,10 @@
     "scan_environment": "azure-pipelines",
     "is_full_scan": true,
     "is_sca_scan": false
-  }
+  },
+  "policy": [
```
brendongo, line 23: Do we need to add this to meta? Aren't policies going to always be audit, comment, block?

cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
```diff
@@ -38,15 +38,11 @@ Trying to fetch branch foo from origin as base branch
...
-Using https://semgrep.dev<MASKED>
-Adding ignore patterns configured on semgrep.dev as `--exclude` options: ()
+Authenticated as org_name
 semgrep version <MASKED>
-Fetching rules from https://semgrep.dev<MASKED>
-trying to download from https://semgrep.dev<MASKED>
-finished downloading from https://semgrep.dev<MASKED>
-loaded 1 configs in<MASKED>
-running 5 rules from 1 config remote-url_0
+running 5 rules from 1 config rules-yaml-string
```
brendongo: Is this expected to go from fetching from registry to string?

cli/src/semgrep/app/scans.py
```diff
@@ -58,25 +86,64 @@ def skipped_match_based_ids(self) -> List[str]:
...
 
-        Returns None if api_token is invalid/doesn't have associated deployment
+    @property
+    def rules(self) -> str:
         """
+        Seperate property for easy of mocking in test
+        """
+        return self._rules
+
+    def get_scan_config_from_app(self, url: str) -> Dict[str, Any]:
```
brendongo: prefix with _ if we are not exporting this as part of the API of this class

cli/src/semgrep/app/scans.py
```diff
@@ -58,25 +86,64 @@ def skipped_match_based_ids(self) -> List[str]:
...
+        body = response.json()
+        if isinstance(body, dict):
+            return body
         else:
-            return None, None
+            raise Exception(
+                f"API server at {state.env.semgrep_url} returned type '{type(response.json())}'. Expected a dictionary."
+            )
+
+    def get_scan_config(self, meta: Dict[str, Any]) -> None:
```
brendongo: idk about naming this get when it doesn't return anything. More of an `initialize` or `init` function

cli/src/semgrep/config_resolver.py
brendongo: Hmm that's weird. Looks like this function returns a string so it shouldn't matter if that string was generated by yaml or json library?

cli/src/semgrep/config_resolver.py
```diff
@@ -226,6 +237,23 @@ def from_pattern_lang(
         valid, errors = cls._validate(config_dict)
         return cls(valid), errors
 
+    @classmethod
+    def from_rules_yaml(cls, config: str) -> Tuple["Config", Sequence[SemgrepError]]:
+        config_dict: Dict[str, YamlTree] = {}
+        errors: List[SemgrepError] = []
+
+        try:
+            resolved_config_key = "rules-yaml-string"
```
brendongo: what's this?

cli/src/semgrep/config_resolver.py
chmccreery: Well it returns different strings - one looks like yaml (with - and \n all over the place) and the other looks like json (with { } and [ ])

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines/meta.json
chmccreery: Oh yeah I wanted to ask y'all about that! This got added to meta when I did a snapshot-update and I had no idea why 😅  Can you shed some light on what this is for and how to make it work without it?

cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
chmccreery: Yes, because it used to download from a rules.yaml endpoint and now it doesn't need to do that because it gets the rules from the same endpoint where it gets deployment id and name.

cli/src/semgrep/config_resolver.py
chmccreery: A hack! 😅  It seems like the current config_resolver expects a dictionary so that each config can be accessed by a unique key. To fit that schema and avoid re-writing most of config_resolver, we just picked a key that would indicate what this config is (a stringified rules.yaml file).

cli/src/semgrep/app/scans.py
chmccreery: What do you think of `fetch_and_init_scan_config`? I think `init` sounds like it's just like formatting or doing something with data it already has, so fetch at least indicates hopefully that it's pulling that data in from elsewhere (semgrep-app).

cli/src/semgrep/app/scans.py
chmccreery: done

cli/Pipfile
chmccreery: done

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines/meta.json
brendongo: It looks like you added it here: https://github.com/returntocorp/semgrep/pull/5928/files#diff-7c7cda58065ec85eebe55f7c9b79daed7e22b02b2ee553bb3379317a0ec4e505R162

brendongo: Oh no that's different.

brendongo: No I take it back it's definitely that line. Looks like you are sending this info in the body of the request?

cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
brendongo: Can we have it say semgrep.dev or something? rules-yaml-string will probably confuse users

cli/src/semgrep/config_resolver.py
brendongo: So seems like backend is returning rule_config as a json blob, then we parse that json as yaml and then dump it as a string. Should backend return it as a yaml string instead?

chmccreery: It can't, because it is now returning all of the GET information at once. So it returns the rules yaml along with a bunch of other things that have always been returned as JSON. Does that make sense?

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines/meta.json
chmccreery: Oh, yes we are. And we need to send that info in the body of the request so that the scan object on semgrep-app has that information.

cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
chmccreery: Yep! Technically this could be re-used for something besides app down the road, but changed for now

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines/meta.json
chmccreery: This tells semgrep-app which policies were used to perform the scan. As we work to make semgrep ci work even if/when semgrep-app is down, we needed to separate all "get" logic from all "post" logic. While the "post" _could_ look up what policies are attached to the repo in question, that information could be out-of-date by the time the POST actually goes through.

cli/src/semgrep/config_resolver.py
underyx: for the record, a blob "with { } and [ ]"  is still valid YAML and there's no reason it should fail YAML validation 🤔

brendongo: @chmccreery that makes sense but no reason the value of a specific key in a json blob has to be json vs string. Can you link to where this is happening I can take a look

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines/meta.json
brendongo: Wouldn't the info sent by cli be strictly more stale than doing a lookup on the backend? Also for 99% of users wont this always be block,comment,audit?

chmccreery: It will be more stale in terms of what the user has configured, but possibly inaccurate in terms of recording what rules/policies the scan actually used, since the scan is using the rules that came from the initial GET request.

chmccreery:
Planning to merge on Tuesday of next week! Afaik this is good to go once https://github.com/returntocorp/semgrep-app/pull/5914 is on production

cli/src/semgrep/config_resolver.py
```diff
@@ -192,15 +196,20 @@ def _load_config_from_local_path(self) -> Dict[str, YamlTree]:
...
         )
-        if r.status_code == requests.codes.ok:
-            return r.content.decode("utf-8", errors="replace")
+        if resp.status_code == requests.codes.ok:
+            try:
+                rule_config = resp.json()["rule_config"]
+                assert isinstance(rule_config, str)
+                return rule_config
+            except BaseException:  # catch JSONDecodeError, AssertionError, etc.
+                return resp.content.decode("utf-8", errors="replace")
```
brendongo, line 209: Should we be returning the response content here or should we just raise?

cli/src/semgrep/config_resolver.py
```diff
@@ -192,15 +196,20 @@ def _load_config_from_local_path(self) -> Dict[str, YamlTree]:
...
             headers={"Accept": "application/json", **self._extra_headers},
         )
-        if r.status_code == requests.codes.ok:
-            return r.content.decode("utf-8", errors="replace")
+        if resp.status_code == requests.codes.ok:
+            try:
+                rule_config = resp.json()["rule_config"]
+                assert isinstance(rule_config, str)
+                return rule_config
+            except BaseException:  # catch JSONDecodeError, AssertionError, etc.
```
brendongo: It's recommended to use Exception instead of BaseException so we don't catch things like keyboard interrupts or other signals

brendongo:
```suggestion
            except Exception:  # catch JSONDecodeError, AssertionError, etc.
```

cli/src/semgrep/config_resolver.py
chmccreery: I think this is needed for backwards compatibility

brendongo (approved):

PR #5929, opened 2022-08-16
feat: parsing todo statistics

brandonspark:
**What:**
It's real rough when you're trying to add a new language to Semgrep. There's like a billion TODOs around, and you gotta fix all of them. The thing is though, although we can get parsing statistics as to what files we have difficulty parsing, and our percentage of parse rate, they aren't actually directed towards where in the translation process that we failed; they just count the frequency of failure on a repository basis.

This PR adds parsing TODO statistics, which count up the places where `todo` was called from within the parser, and then reports them in descending order. That way, you can focus on TODOs which are more prevalent, and make changes which have maximal impact in getting parse rate up.

**Why**:
We should be able to write translations in a more directed fashion, prioritizing those parts of the language which are most common.

**How:**
I parse the exception backtrace for each failed test, for the first instance of a non-`todo`-named function. I save the line number too. Then, I concatenate them together, and order them in descending order, and print to the relevant `stats.log` file. This hooks into the `./run-lang <lang>` parse rate estimator.

So, for instance, with our current 2% parse rate Swift parser, I ran the `./run-lang swift` command, and then the end of the `stats.log` that is produced has the following:

```
TODO statistics:

Parse_swift_tree_sitter.map_non_local_scope_modifier:302, -> 133
Parse_swift_tree_sitter.map_attribute:790, -> 26
Parse_swift_tree_sitter.map_value_argument:2931, -> 10
Parse_swift_tree_sitter.map_call_expression:927, -> 9
Parse_swift_tree_sitter.map_property_declaration:2281, -> 4
Parse_swift_tree_sitter.map_binding_pattern_with_expr:890, -> 3
Parse_swift_tree_sitter.map_type_constraints:2719, -> 2
Parse_swift_tree_sitter.map_type_arguments:2687, -> 1
```
This tells us that the largest source of parse errors is on line 302 in `map_non_local_scope_modifier`, tripping a whole 133 times. This means that we should fix that one next, in order to get parse rate up.

PR checklist:

### DIFF
--- a/semgrep-core/src/parsing/Test_parsing.ml
+++ b/semgrep-core/src/parsing/Test_parsing.ml
@@ -29,9 +29,64 @@ let logger = Logging.get_logger [ __MODULE__ ]
 (* Helpers *)
 (*****************************************************************************)
 
+module SMap = Map.Make (String)
+module IMap = Map.Make (Int)
+
+type mode = Searching | Finding
+
+(* Global store, because I don't want to pass it through a bunch of func calls.
+ *)
+let store = ref []
+
+let process_exn () =
+  let rec process mode (lines : string list) =
+    match lines with
+    | [] -> "failure"
+    | line :: rest -> (
+        (* Sue me. *)
+        let tokens = Common.split " " line in
+        match (mode, tokens) with
+        | Searching, "Called" :: "from" :: _ -> process Finding rest
+        | Searching, _ -> process Searching rest
+        | ( Finding,
+            "Called" :: "from" :: funcname :: _in :: _file :: _filename
+            :: "line" :: linenum :: _ ) ->
+            funcname ^ ":" ^ linenum
+            (* There should be at least one `Called from` after the first... *)
+        | Finding, _ -> failwith "bad")
+  in
+  let res =
+    Printexc.get_backtrace ()
+    |> Common.split (Str.quote "\n")
+    |> process Searching
+  in
+  store := res :: !store;
+  ()
+
 let print_exn file e =
   let trace = Printexc.get_backtrace () in
+  process_exn ();
   pr2 (spf "%s: exn = %s\n%s" file (Common.exn_to_s e) trace)
+
+let report_counts () =
+  let counts =
+    List.fold_left
+      (fun map x ->
+        SMap.update x
+          (function
+            | None -> Some 1
+            | Some x -> Some (x + 1))
+          map)
+      SMap.empty !store
+    |> SMap.to_seq
+    |> Seq.map (fun (x, y) -> (y, x))
+    |> IMap.of_seq |> IMap.to_rev_seq |> List.of_seq
+  in
+  pr2 "\nTODO statistics:";
+  List.fold_left
+    (fun acc (count, filename) -> acc ^ Common.spf "\n%s -> %d" filename count)
+    "" counts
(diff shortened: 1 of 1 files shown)


nmote (changes requested):
Nice, this will help quite a bit. I used a similar tactic when working to get PHP parse errors up, but it was a little more ad-hoc.

One request, though -- could you either change it so that it only prints this information when JSON output was not requested, or double-check that it won't interfere with whatever consumes that info? This is printed on stderr and the JSON is on stdout, it looks like, so it should be fine, but I'd be happier if you took a quick look at how the output here is consumed.

brandonspark:
> Nice, this will help quite a bit. I used a similar tactic when working to get PHP parse errors up, but it was a little more ad-hoc.
> 
> One request, though -- could you either change it so that it only prints this information when JSON output was not requested, or double-check that it won't interfere with whatever consumes that info? This is printed on stderr and the JSON is on stdout, it looks like, so it should be fine, but I'd be happier if you took a quick look at how the output here is consumed.

It looks like the `run-lang` script only ever calls `semgrep-core` with the `-json` flag, so if I make it so that it only prints when JSON is not requested, it will never print.

image

It looks like stderr is specifically piped to the `.log`, and `.json` for the stdout. I don't believe that they should interfere, given that we are using `print_endline` for `print_json` and `pr2` for `report_counts`. Let me know if I should look into it more, though.

nmote (approved):
Sounds good, thanks for checking into it!

semgrep-core/src/parsing/Test_parsing.ml
```diff
@@ -29,10 +29,67 @@ let logger = Logging.get_logger [ __MODULE__ ]
...
+(* This function just processes the exception backtrace to find the first instance where a function
+   that is not called "todo" was called.
+   It then records the line number and name of the function, adding it to a global store.
+*)
+let process_exn () =
+  let rec process (lines : string list) =
+    match lines with
+    | [] -> "failure"
+    | line :: rest -> (
+        (* Sue me. I don't want to write a real lexer and parser. *)
```
aryx, line 48: I usually use a regexp.
if s =~ "Called from \\([^ ]+\\) in [^ ]+ line \\([0-9]+\\)"
then
  let (funcname, line) = Common.matched2 s in
 ...

semgrep-core/src/parsing/Test_parsing.ml
```diff
@@ -29,10 +29,67 @@ let logger = Logging.get_logger [ __MODULE__ ]
...
+  ()
+
 let print_exn file e =
   let trace = Printexc.get_backtrace () in
+  process_exn ();
   pr2 (spf "%s: exn = %s\n%s" file (Common.exn_to_s e) trace)
 
+(* This function collects all the function name and line number pairs, and then
+   sorts them in descending order of frequency.
+   I'm using maps to do this so that I can achieve O(log n) insertion per increment.
```
aryx, line 70: I usually just use a Hashtbl and then Common.hash_to_list |> Common.sort_by_value |> Common.take_safe 10 |> ...

aryx (commented):
This is great! We should probably update https://semgrep.dev/docs/contributing/adding-a-language/
and let external contributors (or a new PA hire, or me in 6 months) know about this!

PR #5934, opened 2022-08-18
tainting: Do not analyze named function definitions twice

IagoAbal:
We were visiting named function definitions twice, one time with `kdef`,
and a second time with `kfunction_definition`!

Closes PA-1778

test plan:
% semgrep --metrics=off -c p/java lang/java/tmp/elastic-elasticsearch
  #^ This is now ~1.3x faster

PR checklist:

### DIFF
--- a/semgrep-core/src/core/ast/Visitor_AST.ml
+++ b/semgrep-core/src/core/ast/Visitor_AST.ml
@@ -100,11 +100,9 @@ let (mk_visitor :
       ?vardef_assign:bool ->
       ?flddef_assign:bool ->
       ?attr_expr:bool ->
-      ?fundef_as_anon:bool ->
       visitor_in ->
       visitor_out) =
- fun ?(vardef_assign = false) ?(flddef_assign = false) ?(attr_expr = false)
-     ?(fundef_as_anon = true) vin ->
+ fun ?(vardef_assign = false) ?(flddef_assign = false) ?(attr_expr = false) vin ->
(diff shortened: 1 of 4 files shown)


semgrep-core/src/core/ast/Visitor_AST.ml
```diff
@@ -100,9 +100,11 @@ let (mk_visitor :
       ?vardef_assign:bool ->
       ?flddef_assign:bool ->
       ?attr_expr:bool ->
+      ?fundef_as_anon:bool ->
       visitor_in ->
       visitor_out) =
- fun ?(vardef_assign = false) ?(flddef_assign = false) ?(attr_expr = false) vin ->
+ fun ?(vardef_assign = false) ?(flddef_assign = false) ?(attr_expr = false)
+     ?(fundef_as_anon = true) vin ->
```
aryx: hmm all the default value are set to false, so maybe we should make this more consistent and set it to false too?

aryx (approved):

semgrep-core/src/core/ast/Visitor_AST.ml
IagoAbal: I agree, although it was a bit more cumbersome to make it `false` by default since one needs to review every use of `mk_visitor` to check where we should enable this (which was the default before). So this made me reconsider the solution, and I think the one I pushed now it's simpler and still seems to give good perf.

PR #5940, opened 2022-08-18
fix(python): Include start and end tokens of Python f strings

### LINKED ISSUE #2995: semgrep autofix breaks format strings
**Describe the bug**
autofix opens up format string quotes, but doesn't put them back together

**To Reproduce**
( I'm not linking to  https://semgrep.dev because that exposes another bug in the playground! If you want to see for yourself, try https://semgrep.dev/s/v0vY/ - and switch to the advanced tab. You'll find that it has broken the indentation of `fix`, making it inoperable )

pr.yml:
```
rules:
- id: print-is-bad
  pattern: print($E)
  message: Please use our logging framework instead
  fix: log($E)
  languages:
  - python3
  severity: ERROR
```

pr.py:
```
print(f"abc")
```

run: 
```
 $ semgrep -f pr.yml pr.py
running 1 rules...
pr.py
severity:error rule:print-is-bad: Please use our logging framework instead
1:print(f"abc")
autofix: log(abc)
```

Notice that the autofix line is missing the quotes around the string.

**Expected behavior**

The autofix line should look like:

```
autofix: log(f"abc")
```

**Screenshots**

P1 (important to fix or quite annoying)

**Environment**
Running off a pip installed semgrep from today
```
$ semgrep --version
0.48.0
```

nmote:
This ensures that the start and end tokens are included in the AST node
representing f strings. It's generally good practice to make sure that
all of the tokens that make up a specific construct are included, but in
particular this addresses an issue with autofix that led to the quotes
and `f` of an f string getting dropped.

Fixes #2995

Test plan: Automated tests

PR checklist:

### DIFF
--- a/changelog.d/gh-2995.fixed
+++ b/changelog.d/gh-2995.fixed
@@ -1,0 +1,1 @@
+Fix an incorrect autofix application when the fix includes Python f strings
--- a/cli/tests/e2e/rules/autofix/python-ranges.yaml
+++ b/cli/tests/e2e/rules/autofix/python-ranges.yaml
@@ -1,0 +1,7 @@
+rules:
+- id: range-test
+  message: change it
+  pattern: foo($E)
+  fix: bar($E)
+  languages: [python]
+  severity: ERROR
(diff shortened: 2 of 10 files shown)


nmote:
Converting to draft until I have a chance to address the test failure

brandonspark (approved):
more tokens! *MORE TOKENS!*

(looks great to me!)

aryx:
This is actually an issue I wanted to bring in the autofix meeting. A big source of problems would be solved
by using Pretty_print_AST and AST-based replacement. But another big source of programs is that
the range of the matched pattern is not good because not all the tokens are in the AST and once the range is not
good, even if you pretty print the fix correctly, you'll still have a dangling comma or parenthesis or quote or whatever that will
mess things up.

nmote:
Yep, for sure -- this is something I'm worried about too. However, this is a problem with autofix no matter what approach we take. To make autofix reliable we'll have to address this range issue.

PR #5942, opened 2022-08-18
fix(extract): regression in mk_rule_table

kopecs:
Due to a changes in for memory consumption the logic for making the
rule table in Run_semgrep changed. However, this relies on the
assumption that rule_pairs and list_of_rule_ids should contain the exact
same rules/rule_ids. When running with extract mode rules, this is
intentionally not the case.

This commit fixes this, by filtering out any unpaired rule_ids present
in these lists. These filtered rule_ids comprise the extract mode rules
which were used to target that file, but which should not be run against
it in search mode.

PR checklist:

### DIFF
--- a/semgrep-core/src/runner/Run_semgrep.ml
+++ b/semgrep-core/src/runner/Run_semgrep.ml
@@ -428,6 +428,10 @@ let rules_for_xlang xlang rules =
          | Xlang.L (x, _empty), Xlang.L (y, ys) -> List.mem x (y :: ys)
          | (Xlang.LRegex | Xlang.LGeneric | Xlang.L _), _ -> false)
 
+(* Creates a table mapping rule id indicies to rules. In the case that a rule
+ * id is present and there is no correpsonding rule, that rule is simply
+ * omitted from the final table.
+ *)
 let mk_rule_table rules list_of_rule_ids =
   let rule_pairs = Common.map (fun r -> (fst r.R.id, r)) rules in
   let rule_table = Common.hash_of_list rule_pairs in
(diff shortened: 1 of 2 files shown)


semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -432,9 +432,15 @@ let mk_rule_table rules list_of_rule_ids =
...
   let id_pairs =
-    List.mapi
-      (fun i rule_id -> (i, Hashtbl.find rule_table rule_id))
-      list_of_rule_ids
+    list_of_rule_ids
+    |> List.mapi (fun i x -> (i, x))
+    (* We filter out rules here if they don't exist, because we might have a
+     * rule_id for an extract mode rule, but extract mode rules won't appear in
+     * rule pairs, because they won't be in the table we make for search
+     * because we don't want to run them at this stage.
```
emjin, line 444: Let's make the assumption explicit (all search/taint rule ids are in this table). There might be another comment related to the rule_table which needs to be updated?

emjin (approved):

semgrep-core/src/runner/Run_semgrep.ml
kopecs: > Let's make the assumption explicit (all search/taint rule ids are in this table).

In terms of just documenting this for the usage here? https://github.com/returntocorp/semgrep/blob/58cac3cff1c0989582fceab63014f3e63c086e0b/semgrep-core/src/runner/Run_semgrep.ml#L600

PR #5949, opened 2022-08-20
fix(python): use tree-sitter python parser

brandonspark:
**What:**
Currently, we are solely using `pfff` for parsing Python. This PR is about using `tree-sitter` as a back-up, in the hopes of eventual migration.

**Why:**
`pfff` is harder to maintain, because we have to add new nonterminals ourselves whenever the grammar is updated. `tree-sitter` is maintained by others and sets a standard, and offers benefits like partial parsing besides. Also, our `pfff` parse rate is 99.1%, which is not reaching our promised metric for GA.

**How:**
I added `tree-sitter` as a secondary parser (should `pfff` fail), and then made the necessary changes to get the `tree-sitter` parser to a good parsing rate.

Before and after of the `tree-sitter` parse rate:
image
image

However, due to the fact that `tree-sitter` runs whenever `pfff` fails, we now have a single test that is failing, namely `foo.py`. This is because `tree-sitter` shockingly does not reject programs which fail Python's white-space stringency, and `pfff` does.

I deemed that this is not terribly important, given that we should only be accepting valid programs anyways, so I adhered to the time-honored software engineering tradition of deleting a test when it doesn't work. Let me know if that's a bad thing to do.

**Test plan:**
`make test`

Closes PA-1739.

PR checklist:


semgrep-core/src/parsing/Parse_target.ml
```diff
@@ -305,12 +305,11 @@ let rec just_parse_with_lang lang file =
   | Lang.Python
   | Lang.Python2
   | Lang.Python3 ->
-      let parsing_mode = lang_to_python_parsing_mode lang in
```
aryx, line 308: Why did you remove the parsing_mode argument?

aryx (approved):
Nice!

semgrep-core/src/parsing/Parse_target.ml
brandonspark: oops, good catch, that was an accident

PR #5957, opened 2022-08-22
fix: Cast YAML timestamp to string

zyannes:
PR checklist:


changelog.d/app-2157.fixed
```diff
@@ -0,0 +1 @@
+When rule YAML metadata contains a date or datetime it is parsed as a timestamp causing a validation error. This overrides the ruamel loader so that timestamps are loaded as strings.
```
underyx:
```suggestion
When a YAML rule file had a string that contained an ISO timestamp, that would be parsed as a datetime object, which would then be rejected by Semgrep's rule schema validator. This is now fixed by keeping strings that contain an ISO timestamp as strings.
```

Note that this line is going straight into semgrep's changelog, so referring to the pull request as "This" would be confusing for our users.

underyx (approved):

changelog.d/app-2157.fixed
zyannes: Good catch! Thanks for updating it!

brendongo (approved):

PR #5960, opened 2022-08-22
Add semgrep_bridge_python and semgrep_bridge_core shared libraries.

smcpeak:
This PR adds shared libraries allowing the semgrep Python CLI to directly invoke the OCaml core.

See `semgrep-core/src/cli/cli-bridge/bridge_design.txt` for details.

The PR also includes a few miscellaneous small changes that are unrelated other than having been done as part of the same work.

Test plan: Copying `bridge_design.txt`, revised (2022-08-24) section Testing here:

> As this work is preliminary to building a Snowflake UDF, there is not meant to be any user-visible consequence of using the bridge module.  However, its existence can be confirmed in the log file, as the command after "Running semgrep-core with command" will mention semgrep_bridge_python.so instead of semgrep-core.
>
> The main method of testing is to simply run the main test suite ("make test", etc.) with the bridge active (SEMGREP_USE_BRIDGE=require).


smcpeak:
I just updated the PR with changes to satisfy the code formatters and, in the github workflow actions, install `python3-config` where evidently needed.

aryx:
Can you add a 'Test plan:' section in your PR to explain how to test your change.
Sometimes this can be as simple as "unit tests included; make test", but in your
case it could be how how can test the new code path where the CLI communicate directly with semgrep-core.

semgrep-core/src/cli-bridge/bridge_design.txt
```diff
@@ -0,0 +1,369 @@
...
+adapted to also locate semgrep_bridge_python.so.  That file then looks
+for semgrep_bridge_core.so in whatever directory it was loaded from by
+taking advantage of the "-Wl,-rpath=$ORIGIN" linker option.
+
+
+Analysis runs in a forked child
+-------------------------------
+
+Prior to invoking the bridge module, core_runner.py forks a child
+process in which to invoke the bridge.  (Note: This is fork without
```
aryx, line 78: and fork is allowed in those map-reduce environments?

semgrep-core/src/cli-bridge/bridge_design.txt
```diff
@@ -0,0 +1,369 @@
...
+progress messages means it would take a large number of them getting
+queued in the OS pipe buffer to hit that buffer limit (typically 8kiB),
+which is in turn what would be required for one of the write calls to
+not write all of its data atomically.  Even then, since the usual
+messages are all two bytes long and the pipe buffer's length is even,
+splitting a message would need a "$n\n" messge where n >= 10 to be among
+those enqueued.
+
+There are several ways to solve this, but doing so is outside the scope
+of the bridge module project, so I merely note this problem.
```
aryx, line 140: thx for noting this problem!

aryx (commented):
This looks great!
I didn't have the time yet to deeply review, but this looks great.

semgrep-core/src/cli-bridge/bridge_design.txt
smcpeak: See last paragraph in that section: "It is unknown whether ... are allowed to invoke fork."  :)  Basically, this fork is needed for things to work in normal usage, and I've got a contingency plan in case it is not allowed in a UDF.

smcpeak:
I'll add a test plan section and look into the latest CI failures.

cli/src/semgrep/semgrep_core.py
```diff
@@ -35,18 +43,97 @@ def compute_executable_path(exec_name: str) -> Optional[str]:
...
     def path(cls) -> str:
+        """
+        Return the path to the semgrep binary, either the Python module
+        or the stand-alone program.  Raise Exception if neither is
+        found.
+        """
         if cls._SEMGREP_PATH_ is None:
-            cls._SEMGREP_PATH_ = compute_executable_path("semgrep-core")
+            # Try the module first (unless disabled).
+            if not os.getenv("SEMGREP_DISABLE_BRIDGE"):
```
aryx: I'd rather make the fork/exec be the default code path, and switch to the bridge only when some environment variables are set (SEMGREP_USE_BRIDGE=1).

smcpeak: Done in 4130f06e5a7807001855d20dcd2767ec028e96e8.

.github/workflows/release.yml
```diff
@@ -143,6 +143,8 @@ jobs:
           github_cache_dir="/__w"
           sudo mkdir -p "$github_cache_dir"
           sudo chmod -R 777 "$github_cache_dir"
+      - name: Install python3-config
+        run: sudo apk add python3-dev
```
brendongo, line 147: why was this needed?

cli/src/semgrep/fork_subprocess.py
```diff
@@ -0,0 +1,225 @@
...
+    # It does not work with arbitrary files.
+    fd_file = os.fdopen(fd, "rb", closefd=False)
+
+    # Add 'fd_file' to the set of descriptors monitored by 'loop' (and
+    # remove it upon EOF).
+    #
+    # Although 'connect_read_pipe' is 'async', it does not really block;
+    # it will succeed or fail immediately, although notification takes a
+    # detour through the 'call_soon' queue.
+    await loop.connect_read_pipe(
```
brendongo, line 56: is there a maximum capacity for the read buffer we should worry about?

cli/src/semgrep/core_runner.py
```diff
@@ -235,35 +261,121 @@ async def _core_stderr_processor(
...
+        return process.wait()
+
     def execute(self) -> int:
         """
         Run semgrep-core and listen to stdout to update
         progress_bar as necessary
 
         Blocks til completion and returns exit code
         """
+        open_and_ignore("/tmp/core-runner-semgrep-BEGIN")
```
brendongo, line 377: what are these for?

brendongo (approved):

.github/workflows/release.yml
smcpeak: `python3-config` is called during the build, by `semgrep-core/src/cli-bridge/dune`.  Within this job, the `scripts/install-alpine-semgrep-core` script runs the build.  I'd be happy to put this info into a comment, but I refrained from doing so here (other than what the name says) since there aren't explanations of other steps.

cli/src/semgrep/fork_subprocess.py
smcpeak: Not for `connect_read_pipe` itself, but yes once the pipe starts being used.  This is handled in `core_runner.py` by passing `limit=INPUT_BUFFER_LIMIT` in a manner parallel to the pathway that uses `asyncio.create_subprocess_exec`.  There is an existing test (`test/e2e/test_performance.py`) that fails if the limit is not set properly.

cli/src/semgrep/core_runner.py
smcpeak: My comment on `open_and_ignore` was meant to explain this:

> Attempt to open 'fname' simply so a record of having done so will be seen by 'strace'.

I was using `strace` to test which files were opened by `semgrep-core`.  To do that, I wanted a marker in the `strace` output of where `semgrep-core` starts and ends in order to easily distinguish its calls from those of the Python CLI.

aryx (approved):

PR #5964, opened 2022-08-23
AST_to_IL: Fix translation of `"some string".concat(E)` exprs

IagoAbal:
Because `"some string"` is not a valid l-value the whole thing was
translated into a Fixme node, and taint wouldn't propagate correctly.

Helps PA-1787

test plan:
make test # added one test

PR checklist:

### DIFF
--- a/changelog.d/pa-1787.fixed
+++ b/changelog.d/pa-1787.fixed
@@ -1,0 +1,3 @@
+taint-mode: Fixed the translation from Generic to IL for expressions like
+`"some string".concat(x)`. Previously, when `x` was tainted, the `concat`
+expression was not recognized as tainted and this caused false negatives.
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -502,15 +502,16 @@ and expr_aux env ?(void = false) e_gen =
         args ) ->
       (* obj.concat(args) *)
       (* NOTE: Often this will be string concatenation but not necessarily! *)
-      let obj' = lval env obj in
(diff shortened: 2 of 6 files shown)


brandonspark (approved):
looks pretty cool to me!

side note: for my edification, what would be the ramification of adding a case to `lval` which permits strings? would that break things? i was somewhat concerned that `expr` would be too permissive w.r.t. what can appear as the `obj` to a `concat` call, but realistically that probably wouldn't be a valid program in most languages anyways

IagoAbal:
> side note: for my edification, what would be the ramification of adding a case to `lval` which permits strings? would that break things? i was somewhat concerned that `expr` would be too permissive w.r.t. what can appear as the `obj` to a `concat` call, but realistically that probably wouldn't be a valid program in most languages anyways

What we do for Ruby is a hack, we translate `obj.concat(x)` as `obj = concat(obj, x)` because the IL's `concat` is pure, but Ruby's one is not. We should have used `expr` to translate `obj` before, `obj` could be any expression.

I'm not sure how we would translate `"some string"` as an l-value. In the IL strings are just immutable literals so they cannot be assigned. We could just create a fresh variable initialized to that string perhaps.

But the "right way" I guess would be to realize that these expressions are actually objects, hence they're pointers, and we would use `expr` as we do now, and the l-value would be made by wrapping that pointer expression with `Mem` (aka "dereference"). Probably not worth it going down that route right now, assuming that's the "right way" anyways (I haven't put much thought into it).

Ruby's strings are more complicated than in other languages, for now we just fix it with hacks when we get a report. Eventually we may need to look at the big picture and design a proper solution.

PR #5970, opened 2022-08-24
dataflow: Make the environment a parameter of the fixpoint function

IagoAbal:
In some cases like in PR #5697 we want to have a richer environment than
just a mapping from strings to X. This refactoring allows for arbitrary
environments to be used in dataflow analyses.

test plan:
make test

PR checklist:


aryx (approved):
I'm not sure what kind of other environment we need beside Dataflow_var_env, but
why not.
I was a bit scared by having another level of abstractions it would make the code even
more complicated but in the end it does not look too bad.

PR #5974, opened 2022-08-24
Feature: Exclude rules by id with CLI flag

### LINKED ISSUE #2530: Ability to exclude rules from rulesets on command line
I want to explore additional rulesets to add to my semgrep scans in CI. Before enabling a new ruleset in CI, however, I want to see the results on the whole codebase. So I just came across `p/r2c-bug-scan` that I had not noticed before. I have a convenience entry in my makefile that sort of approximates the .semgrepignore list that CI honors, like so:
```
SEMGREP_IGNORE := --exclude third_party --exclude *_test.go --exclude *.pb.go --exclude *.pb.*.go --exclude *.bindata.go --exclude *.spec.ts --exclude coverage
semgrep-test/%:
	semgrep --config "p/$(@F)" $(SEMGREP_IGNORE)
```

Thus, from the command-line I can run this:
```
make semgrep-test/r2c-bug-scan
```

That gave me 38 findings, but most of them were a single rule that does not really seem useful:
```
components/compliance-service/inspec/cli.go
severity:warning rule:go.lang.correctness.permissions.file_permission.incorrect-default-permission: Expect permissions to be `0600` or less for ioutil.WriteFile()
61:	err = ioutil.WriteFile(filename, content, 0644)

components/compliance-service/reporting/util/zip.go
severity:warning rule:go.lang.correctness.permissions.file_permission.incorrect-default-permission: Expect permissions to be `0600` or less for os.Chmod()
121:			err = os.Chmod(name, 0777)

components/config-mgmt-service/cmd/chef-automate-collect/commands/automate_config.go
severity:warning rule:go.lang.correctness.permissions.file_permission.incorrect-default-permission: Expect permissions to be `0600` or less for ioutil.WriteFile()
396:		err := os.Mkdir(configDir, 0700)

etc...
```

As I said, that is not useful to me, so I know that when I add the ruleset to my policy on the semgrep dashboard, I will exclude that rule. 

Which brings us to my point here--would be great if I could do the same thing on the command-line, so I could filter out that rule and run my "evaluation scan" again, something like this: 
```
SEMGREP_IGNORE_FILES := --exclude third_party --exclude *_test.go --exclude *.pb.go --exclude *.pb.*.go --exclude *.bindata.go --exclude *.spec.ts --exclude coverage
SEMGREP_IGNORE_RULES := --exclude-rule go.lang.correctness.permissions.file_permission.incorrect-default-permission
semgrep-test/%:
	semgrep --config "p/$(@F)" $(SEMGREP_IGNORE_FILES) $(SEMGREP_IGNORE_RULES)
```
(No need to be fluent in `make` -- the request here is simply to provide an `--exclude-rule` command-line option.)

qequ:
Update cli to add flag `--exclude-rule`  and add functionality to filter rules by id's

Fixes #2530

PR checklist:


brendongo:
Thanks for your contribution @qequ we're still in discussions for philosophically if we want this functionality in semgrep and the future of rulesets (p/python, p/secrets, etc.) in general so this PR might be in limbo for a bit

dgutson:
@brendongo I understand but we have a very concrete problem that this PR solves. How we are working around the problem? Downloading the rules and removing those we don't want. Awful hack.
What about considering this PR as a short term solution while the real and consistent one reaches consensus? I notice that this has been unresolved for a while already so nothing makes me believe it will be solved in the short term, whereas I think @qequ's fix is a good intermediate point that will help people like us.
As Voltaire said: perfect is the enemy of good :)

brendongo:
@dgutson the other alternative is to use the rule board functionality that is available for free on semgrep.dev which is less hacky than the rule download and solution. It is very hard to remove functionality once people start relying on it in CI.

dgutson:
@brendongo I understand your concern, especially considering that this is a quite popular request by looking at the original issue.
However, why would you be removing functionality? IIUC what you are discussing is the implementation, but the functionality would remain the same regardless whether it is implemented as @qequ 's flag change, or by " to merge semgrep-action code into semgrep CLI" as explained by @ievans in https://github.com/returntocorp/semgrep/issues/2530#issuecomment-990392607
That being said, I think that the user interface (CLI) shouldn't change, so users shouldn't be impacted.
What I am missing?

brendongo:
That pre-supposes the solution we want to build to exclude rules is a command line flag. It could be using something akin to a requirements.txt/package-lock.json for example. We could also make it easier to make custom rule packs. We could also axe rule packs all together. etc.

brendongo:
Moreover, something also under consideration: if someone configures a rule board on semgrep.dev and uses `semgrep ci` to run it, should cli flag excluded rules be respected? Cause now we get to a world where the state of rules being run necessitates looking at how CI was setup instead of the ruleboard being the source of truth to what rules run.

dgutson:
> That pre-supposes the solution we want to build to exclude rules is a command line flag. It could be using something akin to a requirements.txt/package-lock.json for example. We could also make it easier to make custom rule packs. We could also axe rule packs all together. etc.

I don't see those solutions as mutually exclusive. If you provide such richness of features, the better!

dgutson:
> Moreover, something also under consideration: if someone configures a rule board on semgrep.dev and uses `semgrep ci` to run it, should cli flag excluded rules be respected? Cause now we get to a world where the state of rules being run necessitates looking at how CI was setup instead of the ruleboard being the source of truth to what rules run.

There are use cases not compatible:
1. off-line deployments of semgrep
2. configuration under version control, which requires to be textual (eg gitlab yaml CI files). This enables workflows with approvals and reviews
3. documentation and reuse of configuration files.

dgutson:
My point @brendongo is, the feature request is 561 days old as of today and has 35 comments, and counting. What reason we have to think that it won't be aging more and more? I don't see any indication that something will change. I understand "the" solution and "the" discussion. But, realistically, I don't see it happening. Let's be practical :)

aryx:
@qequ did you manage to sign the CLA?

aryx:
Oh you did, not sure what's going on with the CLA assistant failure on "Alvaro Frias Garay"

qequ:
Yes I have @aryx but the machine from which I have made the commits has slight different information git configuration (full name, email, etc) and Github detects it as I was two different people but all commits are from the account  @qequ, the one that already signed the CLA.

@aryx Is it possible to bypass the CLA sign check for "Alvaro Frias Garay"?

aryx:
I think it's fine.

qequ:
@aryx I've been able to sign the CLA with the other mail too :+1:

aryx:
lint/pre-commit is failing apparently.
Did you run pre-commit run -a before submitting?
(and pre-commit install the first time I think).

qequ:
@aryx applied pre-commit fixes :+1:

brendongo:
> I don't see those solutions as mutually exclusive. If you provide such richness of features, the better!

We can agree to disagree here: but our thinking is multiple ways of accomplishing the same outcome leads to confusing APIs and special cases that require mental load to untangle. Introducing one way to accomplish something then switching the solution up is also not a great experience.

> What reason we have to think that it won't be aging more and more?

There isn't really. In fact, I explicitly call out that this might sit for a bit. But you can trust that the maintainers are interested in the continued adoption of this tool and we are aware of the sharp edges in rule packs.

aryx (approved):
LGTM.

aryx:
Sorry this took so long to merge @qequ and @dgutson
Should be part of the next release!

PR #5975, opened 2022-08-25
fix(swift): parse expression patterns as expressions

brandonspark:
**What:**
Expressions are currently parsed as singleton statements in Swift.

**Why:**
This means that patterns which are just expressions do not register as expressions, and thus do not recurse more deeply to match against nested expressions.

This means that we fail test cases like the following:
https://semgrep.dev/s/wJlx

**How:**
Made standalone statements that are just `ExprStmt`s parse to expression patterns.

This also means that this wasn't specific to the `try` case -- we'd fail this one, too:
https://semgrep.dev/s/OLbe

Closes PA-1811

PR checklist:

- [X]  Documentation is up-to-date


nmote:
Could you add a test or two for this?

Also, this is out of scope for this PR, but I wonder why this transformation isn't done for all patterns after parsing. It seems language-independent, and @LewisArdern mentioned that he's seen similar issues with other languages too. That makes me suspect that this particular thing has been overlooked elsewhere. @aryx, is there a reason this isn't done later in the pipeline for all patterns instead?

brandonspark:
> Could you add a test or two for this?
> 
> Also, this is out of scope for this PR, but I wonder why this transformation isn't done for all patterns after parsing. It seems language-independent, and @LewisArdern mentioned that he's seen similar issues with other languages too. That makes me suspect that this particular thing has been overlooked elsewhere. @aryx, is there a reason this isn't done later in the pipeline for all patterns instead?

Done!

aryx:
Yes, maybe we could factorize this thing in Parse_pattern.ml.
Some languages have explicit semicolons, so sometimes an ExprStmt is actually a stmt, so in that case
if someone wrote foo();, you don't want to transform that in a pattern for just foo()
But still we can probably factorize things a bit more.

nmote:
Maybe we could check if the semicolon is a real token, or perhaps if its text is nonempty, and if so skip the extraction?

nmote (approved):

PR #5976, opened 2022-08-25
fix(swift): attribute implicit semicolon locations to previous line

brandonspark:
**What:**
Currently, the Swift tree-sitter parser inserts fake semicolons whenever there is an implicit one. Strangely, however, it inserts them at the beginning of the next line. This borks things w.r.t. Semgrep, because now it means that matches which match an entire expression with an implicit semicolon at the end now match the expression, and the following line.

See https://github.com/alex-pinkus/tree-sitter-swift/issues/179 for more.

**Why:**
We should fix this so that our range reporting is more accurate to the actual finding produced. Otherwise, people might think we matched two lines instead of one.

**How:**
So, there was a time where I considered hacks my enemy. Now, it looks like I have become Death, destroyer of worlds, because this is really hacky.

Actually, the above sentence is no longer true, but it's funny, so I'm keeping it.

Basically, for anytime we map a semicolon token, we check if it contains nullary data as a string. This should only happen if it has absurd location data, which should only happen if it's one of the aforementioned implicit semicolons. If so, we insert a fake semicolon, which does not cause problems for the range data.

For finding the pattern `foo` in the following program:
```
foo
bar
```
before this change, Semgrep reported:
```
test.swift:1
 foo
 bar
```

After this change, Semgrep reports:
```
test.swift:1
 foo
```

Looks like a win to me.

**Alternatives Considered:**
So, I thought about doing something at the `tree-sitter` level. But I really don't understand what's going on over there, so I decided to just do it here.

Closes PA-1810

PR checklist:

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
@@ -1951,15 +1951,30 @@ and map_local_declaration (env : env) (x : CST.local_declaration) : G.stmt list
 and map_semi (env : env) (semi : CST.semi option) : G.sc =
   match semi with
   | None -> G.sc
-  | Some semi -> token env semi
+  | Some ((loc, str) as semi) -> (
+      let tok = token env semi in
+      match (loc.Tree_sitter_run.Loc.start.column, env.extra) with
+      (* If it's a program, and this semicolon is at the beginning of the line, that's
+         kind of strange. Let's just move it to the charpos occurring immediately before it
+         (the position of the newline on the previous line)
+      *)
+      | 0, Program ->
+          Parse_info.fix_token_location
+            (fun tloc ->
+              let charpos = tloc.charpos in
+              (* Only if this is not the first thing in the program, though. *)
+              if charpos >= 0 then
+                let line, column =
+                  Parse_info.full_charpos_to_pos_large env.file (charpos - 1)
+                in
+                { tloc with charpos = charpos - 1; line; column }
+              else tloc)
+            tok
+      | _ -> tok)
 
 and map_local_statement (env : env) (x : CST.local_statement)
     (semi : CST.semi option) : G.stmt list =
-  let semi =
-    match semi with
-    | Some semi -> token env semi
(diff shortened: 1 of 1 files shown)


brandonspark:
Note: I couldn't figure out how to get Semgrep to stop crashing when an implicit semicolon is the last thing in the program. Hopefully that doesn't happen very often, or I'll keep thinking.

nmote:
This would also be a major hack, but what if instead of inspecting the location of the semicolon, we inspect the contents and if it contains only whitespace (which hopefully the implicit semicolons do?) we drop it and use `G.sc` instead? Would that address the crashing at the end of the file issue?

@alex-pinkus is amenable to changing the grammar so that explicit and implicit semicolons can be distinguished from each other, so if we prove this approach out with a hack, we have a path towards making it less hacky in the future. What do you think?

alex-pinkus:
Yep, just let me know if you need me to separate those out and I'll update the grammar.

Really sorry for causing this hack to be necessary, by the way. I've spent literal months trying to get the whitespace-handling logic to behave similarly to how the real compiler does it, where there's so much nonsense around comments and ambiguous expressions that make me suspect the official rules boil down to "do what we think the programmer wants, no matter what it takes."

brandonspark:
> This would also be a major hack, but what if instead of inspecting the location of the semicolon, we inspect the contents and if it contains only whitespace (which hopefully the implicit semicolons do?) we drop it and use `G.sc` instead? Would that address the crashing at the end of the file issue?
> 
> @alex-pinkus is amenable to changing the grammar so that explicit and implicit semicolons can be distinguished from each other, so if we prove this approach out with a hack, we have a path towards making it less hacky in the future. What do you think?

Actually, this solution is pretty good. It almost feels like it's not even a hack. It fixes the crashing issue, too.

Updated!

nmote (approved):
Nice, looks great! This turned out simpler than I imagined.

nmote:
Could you update the summary to reflect the new implementation?

@alex-pinkus, thanks, and no apologies needed. I think it would be nice to have explicit and implicit semicolons separated out in the parse tree, but this hack isn't actually that bad so I wouldn't consider it urgent.

PR #5980, opened 2022-08-25
feat(js): Add switch case pattern parsing

nmote:
This also begins to fall back on the tree-sitter parser to parse
patterns, rather than extending the pfff parser to support this. For
now, the only semgrep constructs that the tree sitter parser can handle
are metavariables (because they are valid JS syntax already), ellipsis,
and switch cases. We will likely want to extend that, because in its
current state, using switch cases as top level patterns will restrict
users from using more advanced Semgrep constructs in the pattern.

We will also likely want to add support for `default:` as a pattern. The
current behavior is confusing: `default:` does parse as a top level
pattern, but it parses as a label. Because it's a valid program on its
own, we'll have to resort to some hack to interpret it as a default
switch case and match accordingly.

See related PRs for context:
- https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/338
- https://github.com/returntocorp/pfff/pull/547

Test plan: Automated tests

PR checklist:


nmote:
To simplify updates to the pfff submodule, this should most likely be merged after https://github.com/returntocorp/semgrep/pull/5949

nmote:
The failing semgrep scan appears to be due to https://github.com/returntocorp/semgrep/pull/5960

aryx:
Fixed by https://github.com/returntocorp/semgrep/pull/5982

semgrep-core/src/parsing/Parse_pattern.ml
```diff
@@ -98,8 +98,21 @@ let parse_pattern lang ?(print_errors = false) str =
...
-        let any = Parse_js.any_of_string str in
-        Js_to_generic.any any
+        (* This is a simplified version of run_either in Parse_target.ml. We
+         * don't need most of the logic there when we're parsing patterns, so it
+         * doesn't make sense to reuse it. However, we should abstract this out
+         * if we start using fallback parsers for patterns in other languages.
+         * *)
+        let js_ast =
+          try Parse_js.any_of_string str with
+          | exn -> (
```
aryx, line 109: You need to add 
          | Timeout _ as e -> Exception.catch_and_reraise e
otherwise you could swallow a Timeout.
We should write a semgrep rule for that, but we probably need to improve the try/with semgrep pattern for OCaml.

aryx (approved):

semgrep-core/src/parsing/Parse_pattern.ml
nmote: https://github.com/returntocorp/semgrep/pull/5999

PR #5989, opened 2022-08-26
feat!: propagate arbitrary expressions to singly assigned variables

brandonspark:
**What:**
Currently, constant propagation only allows the propagation of _constants_. Symbolic propagation is a thing, but it doesn't work at the level of globals and through scopes. We should identify more cases where we know that a particular identifier is safe to consider set to a given constant.

**Why:**
This will allow test cases like the following to pass:
https://semgrep.dev/s/nNbg

**How:**
I notice that in our flow-insensitive pass for constant propagation, we only associate an identifier to a constant so long as it is declared as `const`, `final`, or it has only been assigned to once (and is in Javascript).

I don't really see the point of the Javascript condition, so I killed that.

In addition, I made it so that this would also happen even if the expression that it was assigned to was an arbitrary expression. Now, this expression may itself contain references to variables, which we shouldn't transport, due to it being out of scope. However, these references must either be themselves propagated constant values, or they are not, in which case they are unknown. In this case, it causes no erroneous matches to just propagate with these identifiers without `svalue`, which is understood to be an identifier whose value we do not know at this time.

This lets us match the case above. In addition, patterns may be written with the _name_ of the variable that they contain, if the exact value is not known. For example, trying to match `f(g(x))` matches:
```
int x = 2;

int y = g(x);

// ruleid: find-propagated-non-constant-exp 
int h = f(y);

x = 3;

int main() {
  // ruleid: find-propagated-non-constant-exp 
  int z = f(y)
}
```
because in both cases, the propagated value at `h` and `z`'s assignment is known to be `f(g(x))`, for some `x` without an `svalue`. This is OK though - although we cannot know what `x` is at those times, we do know that at some point, `f(g(x))` is in fact what is evaluated, for some value of `x`.

**Test plan:**
`make test`

Closes PA-1821.

PR checklist:


IagoAbal:
> I don't really see the point of the Javascript condition, so I killed that.

Presumably because Javascript has keywords to specify what is a constant so we don't need to make guesses and we can just follow what the devs have annotated. That said, it's probably safe to just use our poor-man's const inference, although I'm not a JS expert.

changelog.d/pa-1821.added
```diff
@@ -0,0 +1 @@
+Symbolic propagation: Added propagation of all variables which are assigned to a single time.
```
IagoAbal: To be more precise, this affects to global variables and class attributes, because local variables were already propagated even if not assigned just once. That makes it more clear what you are going to get out of this change.

semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -563,17 +562,25 @@ let propagate_basic lang prog =
...
+              VarDef { vinit = Some e; _ } )
           (* note that some languages such as Python do not have VarDef.
            * todo? should add those somewhere instead of in_lvalue detection?*)
             -> (
               match Hashtbl.find_opt stats (H.str_of_ident id, sid) with
               | Some stats ->
-                  if
-                    H.has_keyword_attr Const attrs
-                    || H.has_keyword_attr Final attrs
-                    || (!(stats.lvalue) = 1 && is_js env)
```
IagoAbal, line 575: @aryx do you think it's OK to remove `is_js env` here? Intuitively, it should be. But we handle so many languages and our const inference may not be perfect, that I'm a bit concerned we could start producing FPs in some places. Although our test suite seems to pass.

semgrep-core/src/analyzing/Constant_propagation.ml
```diff
@@ -563,17 +562,25 @@ let propagate_basic lang prog =
...
+                   || !(stats.lvalue) = 1
+                  then
+                   match e.e with
+                   | L literal -> add_constant_env id (sid, Lit literal) env
+                   (* For any other expression, it is OK to propagate it symbolically so long as
+                      the lvalue is only assigned to once.
+                      Although we may propagate expressions with identifiers in them, those identifiers
+                      will simply not have an `svalue` if they are non-propagated as well.
+                   *)
+                   | _ -> add_constant_env id (sid, Sym e) env);
```
IagoAbal: At least for now I prefer that we restrict this to `Dataflow_svalue.is_symbolic_expr` same way as the dataflow analysis does.

IagoAbal (approved):
Nice!

IagoAbal:
```
FAILED tests/qa/test_public_repos.py::test_semgrep_on_repo[gh:home-assistant/home-assistant]
[184](https://github.com/returntocorp/semgrep/runs/8043434904?check_suite_focus=true#step:11:185)
```
Perhaps also need to update test expectations here?

semgrep-core/src/analyzing/Constant_propagation.ml
IagoAbal: Actually returntocorp/semgrep-rules#2358 is already an example where we are flagging something that we shouldn't, isn't it? Just because a class attribute is assigned once _inside the class itself_, it doesn't mean it cannot be assigned from outside. And const-prop is meant to be a _must_ analysis.

IagoAbal (changes requested):
Actually we need to restrict `VarDef` single assignments. I agree that `is_js env` is perhaps not the best condition to put there, and I'm not sure why we allow that in JS only, but simply removing it is also not OK. We would start reporting many constants that are not really constants in e.g. Java and other OO languages.

brandonspark:
> Actually we need to restrict VarDef single assignments. I agree that is_js env is perhaps not the best condition to put there, and I'm not sure why we allow that in JS only, but simply removing it is also not OK. We would start reporting many constants that are not really constants in e.g. Java and other OO languages.

I'm not quite sure I understand the condition we need to restrict things for. Could you give me an example of something that would break?

IagoAbal:
> > Actually we need to restrict VarDef single assignments. I agree that is_js env is perhaps not the best condition to put there, and I'm not sure why we allow that in JS only, but simply removing it is also not OK. We would start reporting many constants that are not really constants in e.g. Java and other OO languages.
> 
> I'm not quite sure I understand the condition we need to restrict things for. Could you give me an example of something that would break?

We just need to put back the `is_js env` in `|| (!(stats.lvalue) = 1 && is_js env)`. In general, e.g. for Java class attributes as shown by the test that failed in semgrep-rules, that a variable is assigned just once in one file doesn't mean it's a constant. Here we make certain assumptions about when that's OK. We can do that in some cases, but not in all cases. Simply removing `is_js env` is not OK, because we will get new const-prop false positives that we didn't get before. I would like to replace `is_js env` with something better, for sure, but it's probably outside the scope of this PR, let's deal with one problem at a time.

cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
```diff
@@ -111,15 +111,6 @@ Running semgrep-core with command:
       "pattern": "$X == 4",
       "severity": "ERROR"
     },
-    {
```
IagoAbal, line 114: Could you explain this?

IagoAbal (approved):
Nice work! I'm not sure why the changes in cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt but other than it looks perfect to me! Feel free to merge!

cli/tests/e2e/snapshots/test_ci/test_github_ci_bad_base_sha/results.txt
brandonspark: yeah I'm just hoping this is CI being weird and nondeterministic

PR #5993, opened 2022-08-26
feat!: add taint patterns, remove taint mode

brandonspark:
Should be reviewed strictly after #5916.

**What:**
Currently, taint mode lives as something separate from search mode. We can only write rules which are taint rules, but no rules which use taint as a sub-formula to a larger pattern.

**Why:**
This simplifies the writing of some rules.

**How:**
Removed taint mode, and added a new pattern which is dedicated to taint. Taint can appear as a subpattern to other larger patterns. So for instance, a valid rule is:
```
id: new_taint_rule
match:
  and:
  - taint:
      sources:
      - pattern: "source"
      - pattern: "source2"
      sinks:
      - pattern: "sink(...)"
      - pattern: "sink2(...)"
  - inside: |
    def $FUNC(...){
      ...
    }
languages: [python]
severity: WARNING
```

**Taint as a Sub-Pattern**
`taint` is now its own pattern, which may appear as a subpattern to other parent patterns. This allows taint analysis to occur in a local sense, so that you could run taint only within a particular `inside` pattern, for instance.

Well, in reality, taint dataflow is going to run everywhere, but the findings can be local.

**Backwards Compatibility and Migration:**
As with #5916, the new syntax is optional, and full backwards compatibility is maintained with the old taint mode syntax. The `-translate_rules` command has been updated to be able to translate old-style taint mode rules to the new syntax.

**Concerns:**
- Dataflow runs on every single function for every taint pattern, at the moment.
PR checklist:

- [] Documentation is up-to-date


IagoAbal (changes requested):
Haven't properly reviewed this yet, but it needs to wait for #5697 to be merged.

cli/src/semgrep/core_runner.py
```diff
@@ -852,6 +869,8 @@ def validate_configs(self, configs: Tuple[str, ...]) -> Sequence[SemgrepError]:
                 + list(configs)
             )
 
+            print(f"{' '.join(cmd)}")
```
IagoAbal: Should this be here?

cli/src/semgrep/core_runner.py
```diff
@@ -476,7 +476,22 @@ def _parse_core_output(
                 # Killed by signal 11 (segmentation fault), this could be a
                 # stack overflow that was not intercepted by the OCaml runtime.
                 soft_limit, _hard_limit = resource.getrlimit(resource.RLIMIT_STACK)
-                tip = f" Semgrep exceeded system resources. This may be caused by\n\n    1. Stack overflow. Try increasing the stack limit to `{soft_limit}` by running `ulimit -s {soft_limit}` before running Semgrep.\n    2. Out of memory. Try increasing the memory available to your container (if running in CI). If that is not possible, run `semgrep` with `--max-memory $YOUR_MEMORY_LIMIT`.\n    3. Some extremely niche compiler/c-bindings bug. (We've never seen this, but it's always possible.)\n\nYou can also try reducing the number of processes Semgrep uses by running `semgrep` with `--jobs 1` (or some other number of jobs). If you are running in CI, please try running the same command locally."
```
IagoAbal: This seems totally unrelated.

semgrep-core/src/cli/Main.ml
```diff
@@ -515,6 +515,9 @@ let all_actions () =
       " <metachecks file> <files or dirs>",
       Common.mk_action_n_arg (Check_rule.check_files mk_config Parse_rule.parse)
     );
+    ( "-translate_rules",
```
IagoAbal: Unrelated?

cli/src/semgrep/core_runner.py
brandonspark: No. It was deleted in the previous incarnation of this PR (the first wave of changes), and I hadn't gotten around to updating this one yet.

cli/src/semgrep/core_runner.py
brandonspark: Same here

semgrep-core/src/cli/Main.ml
brandonspark: Probably same here

PR #5995, opened 2022-08-28
Add semgrep_bridge_python and semgrep_bridge_core shared libraries.

smcpeak:
This commit is a revision of the reverted b57cb30b48 with fixes for
compatibility with macos.

* cli.py: cli: Adjust "type: ignore" for 'commands'.

At run time I can see that 'ctx.command.commands' is a dictionary.
However, that attribute is not present in the Click documentation, so
mypy is unhappy without this intermediate assignment and warning
suppression.

I do not know how to reconcile the conflict between this usage and the
documentation, nor why I see a mypy complaint that evidently the CI
build does not.

* spacegrep/scripts/show-perf: Read from /dev/urandom, not /dev/random.

/dev/random blocks when the kernel thinks it is low on entropy,
which can cause the test to hang, particularly on VMs.  This test
does not require cryptographically secure randomness, so urandom is
fine.

* semgrep_main.py: Log the command line.

Particularly when debugging tests that invoke semgrep, it is helpful
to have the command line captured.

* qa/public_repos.py: Disable dropbox/changes.

Nested four levels deep in node_modules, the file outro.js consists
of one line, "}.call(this));", which does not parse.

* core_runner.py: When failing to parse output JSON, print the exception.

This can help debug the "internal error".

* semgrep_core.py: compute_executable_path: Catch not found exception.

In my testing, if I remove a file that path() is attempting to load,
it raises FileNotFoundError, rather than returning a path for which
is_file() fails.  Since the intent of this code is to recover, we now
catch and log the exception.

* Add semgrep_bridge_python and semgrep_bridge_core shared libraries.

See semgrep-core/src/cli/cli-bridge/bridge_design.txt for details.

* tests.yml:test-core, Dockerfile: Install python3-dev.

Building semgrep_bridge_python.so needs python3-config.

* install-alpine-semgrep-core: Include the bridge .so libraries.


smcpeak:
The difference between this PR and the reverted PR #5960 is visible as:

https://github.com/smcpeak/semgrep/commit/fa84b02a44ce42bf74d08bbe3442187eff3f80db

aryx:
@spencerdrak is there a way to manually trigger the macos CI actions (which are currently skipped)?

aryx:
Or I guess I can merge and wait to see if some other jobs kick in and fail ...

aryx (approved):

PR #5999, opened 2022-08-29
chore: reraise-timeout Semgrep rule

nmote:
As suggested in #5980

This rule has 100 findings. Based on some spot checking, this seems to
be working as intended. However, I'm not sure whether we actually want
to enforce that for every `try`, we catch and reraise `Timeout`. Some of
the findings are on simple things like
`try Hashtbl.find ... with Not_found -> ...`. It would be annoying to
have to catch and reraise `Timeout` there, but maybe worth it to have
the guarantees of this rule? I'd love to hear what others think.

If people would like to move forward with this, I can go through and fix
the existing findings before committing this.

PR checklist:

### DIFF
--- a/semgrep.yml
+++ b/semgrep.yml
@@ -147,6 +147,18 @@ rules:
     paths:
       include:
         - cli/src/*
+  - id: reraise-timeout
+    patterns:
+      - pattern: try ... with | $X -> ...
+      - pattern-not: try ... with | Timeout _ as e -> ...
+      - pattern-not: try ... with | Timeout $X -> ...
+    message: >-
+      Catch and re-raise Timeout exceptions to avoid swallowing them
+    languages: [ocaml]
+    severity: ERROR
+    paths:
+      include:
+        - semgrep-core/src/*
 # not ready yet
 #  - id: no-exit-in-semgrep
 #    pattern: |


aryx:
For the Not_found we should not have to add a case. I'm more worried about code that
really swallow like try ... with exn -> ...

semgrep.yml
```diff
@@ -147,6 +147,18 @@ rules:
     paths:
       include:
         - cli/src/*
+  - id: reraise-timeout
+    patterns:
+      - pattern: try ... with | $X -> ...
```
aryx, line 152: I think we want to mainly detect the 'try with' where we "swallow" an exn with a general pattern
like
try ... with | exn -> ...
That's where it's dangerous because the Timeout would not pop up. With 'try ... with Not_found -> ... it's fine. It will still pop up.

aryx: But maybe we can be ugly and check that the $X is an ident with metavariable-regex:?
Still I'm not sure it would catch code like try foo with | Not_found -> bla | OtherExn -> bla | exn -> dangerous.

aryx (changes requested):

semgrep.yml
nmote: Oh right, good point. I'll see if I can come up with a pattern that will match what we want.

nmote:
Unfortunately I think you were right that the current OCaml try patterns are not powerful enough to express what we want here. In particular, I haven't found a way to write a single pattern that matches a `try` expression with any number of cases. Closing for now.

aryx:
Good use-case to dogfood semgrep and improve our OCaml support :)

PR #6009, opened 2022-08-30
Experiment with jsonnet

aryx:
Jsonnet  support in the CLI!

test plan:
```
$ cat ~/github/jsonnet-rules/examples-policy/myrule.jsonnet 
{ rules: [
// comments are ok (C++ or Shell syntax)
{
  # no need for “id”:
  id: "my-rule",
  # multilines strings
  pattern: |||
      foo($X)
      ... 
      bar($X)
   |||,
  message: "you should not do that",
  languages: ["python"],
  severity: "ERROR", # trailing commas are ok
}
]}
$ semgrep --config ~/github/jsonnet-rules/examples-policy/myrule.jsonnet ~/github/jsonnet-rules/examples-policy/files/
Scanning 1 file.

Findings:

  /home/pad/github/jsonnet-rules/examples-policy/files/test.py 
     home.pad.github.jsonnet-rules.examples-policy.my-rule
        you should not do that

          2┆ foo(1)
          3┆ stuff()
          4┆ bar(1)

Ran 1 rule on 1 file: 1 finding.

```

PR checklist:


aryx:
@brendongo any feedback on the python refactoring emma did?

cli/src/semgrep/config_resolver.py
```diff
@@ -69,12 +72,15 @@
 }
 
 
+ConfigInfo = namedtuple("ConfigInfo", "config_id, contents, filename config_url")
```
brendongo: Probably missing a comma?

cli/src/semgrep/config_resolver.py
```diff
@@ -120,18 +126,11 @@ def __init__(self, config_str: str, project_url: Optional[str] = None) -> None:
             state.metrics.is_using_registry = True
             state.metrics.add_registry_url(self._config_path)
 
-    def resolve_config(self) -> Mapping[str, YamlTree]:
-        """resolves if config arg is a registry entry, a url, or a file, folder, or loads from defaults if None"""
-        start_t = time.time()
-
+    def load_config(self) -> List[ConfigInfo]:
```
brendongo: Function level comment appreciated

cli/src/semgrep/config_resolver.py
brendongo: Also if you make this a class instead of a named tuple we get type checking so might be worth.

cli/src/semgrep/config_resolver.py
```diff
@@ -145,31 +144,27 @@ def _nice_semgrep_url(self, url: str) -> str:
             return url.replace("/c/", "/")
         return url
 
-    def _download_config(self) -> Mapping[str, YamlTree]:
+    def _download_config(self) -> List[ConfigInfo]:
```
brendongo: Since this only ever returns one download config I would make it return ConfigInfo and have callers wrap in a list if necessary.

cli/src/semgrep/config_resolver.py
```diff
@@ -145,31 +144,27 @@ def _nice_semgrep_url(self, url: str) -> str:
...
         """
         config_url = self._config_path
         logger.debug(f"trying to download from {self._nice_semgrep_url(config_url)}")
         try:
-            config = parse_config_string(
+            config = ConfigInfo(
                 "remote-url",
                 self._make_config_request(),
-                filename=f"{config_url[:20]}...",
+                f"{config_url[:20]}...",
```
brendongo: I'm assuming this line was used somehow in parse_config_string for some deduplication logic? If we arent using parse_config_string here anymore should we just remove this?

cli/src/semgrep/config_resolver.py
```diff
@@ -215,8 +210,67 @@ def _make_config_request(self) -> str:
     def is_registry_url(self) -> bool:
         return self._origin == ConfigType.REGISTRY
 
+
+def read_config_at_path(loc: Path, base_path: Optional[Path] = None) -> ConfigInfo:
```
brendongo: Do these need to be top level functions or can they be hidden inside relevant classes?

cli/src/semgrep/config_resolver.py
```diff
@@ -215,8 +210,67 @@ def _make_config_request(self) -> str:
...
+    for (config_id, contents, filename, config_url) in loaded_config_infos:
+        try:
+            config.update(parse_config_string(config_id, contents, filename))
+        except InvalidRuleSchemaError as e:
+            if config_url:
+                notice = f"\nRules downloaded from {config_url} failed to parse.\nThis is likely because rules have been added that use functionality introduced in later versions of semgrep.\nPlease upgrade to latest version of semgrep (see https://semgrep.dev/docs/upgrading/) and try again.\n"
+                notice_color = with_color(Colors.red, notice, bold=True)
+                logger.error(notice_color)
+                raise e
+            else:
```
brendongo, line 267: Probably worth printing some nice error message if config_url is None?

cli/src/semgrep/config_resolver.py
```diff
@@ -450,17 +504,29 @@ def indent(msg: str) -> str:
...
-    if base_path:
-        config_id = str(loc).replace(str(base_path), "")
-
-    return parse_config_string(config_id, loc.read_text(), str(loc))
+def import_callback(_base: str, path: str) -> Tuple[str, str]:
+    logger.debug(f"import_callback for {path}")
+
+    # On the fly conversion from yaml to json.
+    # Can now do 'local x = import "foo.yml";'
+    if path and path.split(".")[-1] == "yml":
```
brendongo: Might also be yaml :D

brendongo: Is it possible to make the arguments Path objects instead of str. pathlib also exposes this function https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix

brendongo: When is import_callback called? Can we not just have jsonnet generate json so we dont need to convert from yaml to json?

emjin: import_callback implements the function jsonnet uses for its import_callback. I'm not certain what types it expects, so I'm hesitant to change these arguments. And in fact I know that in some cases we will expect the path to be something like `p/python`. The function is called when jsonnet resolves the imports, which is why need this conversion. I'll document these things better

cli/src/semgrep/config_resolver.py
emjin: I ended up not doing this for now because we didn't previously. We can add that in another PR if we want

cli/src/semgrep/config_resolver.py
emjin: I wasn't sure. These are just moved; I didn't know why they used to be top level functions. If they can be moved I might do so in another PR.

emjin (approved):

PR #6010, opened 2022-08-30
fixed(sym prop): added usage of sym prop as last resort

### LINKED ISSUE #6003: Unsupported ellipsis in symbolic propagation option
**Describe the bug**
The ellipsis does not work in symbolic propagation. This issue could be solved by using taint mode. But does semgrep has a simple way to resolve such conditions based on symbolic propagation?

**To Reproduce**
https://semgrep.dev/s/7l6Q

brandonspark:
**What:**
Currently, when matching against a `N` (which may contain a propagated `svalue`), we allow some cases to reach `fail`, without trying to match with symbolic propagation (if enabled) first.

This means that we basically give up without using all the information we can.

**Why:**
We currently fail a test case due to this. See https://semgrep.dev/playground/s/7l6Q, where basically we end up matching a `DotAccessEllipsis` to a `N`, and then give up because `DotAccessEllipsis` was not explicitly stated as something which can unpack what is inside of a `N`.

**How:**
Added a case where if anything matches to an `N`, right before the failing case, we will try to match using symbolic propagation.

I don't see why this shouldn't be here, because if symbolic propagation is on, we might as well use it. Furthermore, there are several cases (for specific pattern variants) where we do essentially this, such as `Call`, `New` and `ArrayAccess`. Due to the fact that we've specifically whitelisted variants that can do this, we missed this case.

I added a catch-all case, but I would be down to delete the `Call`, `New`, and `ArrayAccess` cases, and fold them under this one. I just haven't done it yet in this PR, in case that's something we shouldn't do. Let me know.

**Test plan:**
`make test`

Closes PA-1829.
Closes #6003.

PR checklist:

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -1007,6 +1007,8 @@ and m_expr ?(is_root = false) a b =
   | G.OtherExpr (a1, a2), B.OtherExpr (b1, b2) ->
       m_todo_kind a1 b1 >>= fun () -> (m_list m_any) a2 b2
   | G.N (G.Id _ as a), B.N (B.IdQualified _ as b) -> m_name a b
+  | _, G.N (G.Id _) ->
+      m_with_symbolic_propagation ~is_root (fun b1 -> m_expr a b1) b
   | G.Container _, _
   | G.Comprehension _, _
   | G.Record _, _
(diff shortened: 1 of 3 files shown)


IagoAbal:
> I added a catch-all case, but I would be down to delete the Call, New, and ArrayAccess cases, and fold them under this one. I just haven't done it yet in this PR, in case that's something we shouldn't do. Let me know.

I will review now but if you are quite sure this is strictly more general and no tests break then for sure go ahead and open a new PR after this one! 🚀

IagoAbal (approved):
Nice! TBH right now I can't see a reason why I didn't just do this. I guess I was just being overly careful. Maybe also because initially just did this for `Call` and then kept adding more cases without realizing that we could just do it for _any_ case. Anyways, great job!

Issue #6002, opened 2022-08-29
.class property cannot be parsed in Java


backcover7:
**To Reproduce**
https://semgrep.dev/s/dzA0

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1828/class-property-cannot-be-parsed-in-java. _Note_: this link is for r2c use only and is not accessible publicly.

IagoAbal:
As far as I can tell the issue is that in Pfff `Hello.class` is represented this way:
```
E(
  OtherExpr(("ClassLiteral", ()),
    [T(
       {t_attrs=[]; 
        t=TyN(
            Id(("Hello", ()),
              {id_info_id=1; id_hidden=false; id_resolved=Ref(None); 
               id_type=Ref(None); id_svalue=Ref(None); }));
        })]))
```
And in our Tree Sitter front-end is represented this way:
```
DotAccess(
N(
Id(("Hello", ()),
    {id_info_id=7; id_hidden=false; id_resolved=Ref(
    None); id_type=Ref(
    None); id_svalue=Ref(
    None); })), (),
FN(
Id(("class", ()),
    {id_info_id=8; id_hidden=false; id_resolved=Ref(
    None); id_type=Ref(
    None); id_svalue=Ref(
    None); })))
```

aryx:
@emjin maybe that's the kind of bug we could detect automatically.
Note that semgrep-core -diff_pfff_tree_sitter also shows the difference.

aryx:
Tree-sitter is wrong. I'm not sure why because there is a grammar rule for class_literal, but it does not seem to trigger.

aryx:
Oh someone opened an issue about it: https://github.com/tree-sitter/tree-sitter-java/issues/121

aryx:
This looks like the better fix: https://github.com/tree-sitter/tree-sitter-java/pull/125#pullrequestreview-1111528606

Issue #5558, opened 2022-06-16
Java constructors with modifiers patern not detected


vladiavramov:
**Describe the bug**
When trying to match a pattern to a constructor with modifier inside a class (e.g. public House) the system returns a parse error.

**To Reproduce**
Run the example. (https://semgrep.dev/s/B1R6)
If it's run without the public modifier the patten is captured.

**Expected behavior**
The expected behaviour is for the pattern to be matched.

**Screenshots**

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Environment**
Tested through the Playground, but otherwise used through CodeGrade.

**Use case**
The bug fix would help in auto-grading code for correct usage of modifiers.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1516/java-constructors-with-modifiers-patern-not-detected. _Note_: this link is for r2c use only and is not accessible publicly.

backcover7:
This bug also blocks me when writing rules. Is there any ETA for this bug fix?

nmote:
@backcover7 I have what I believe to be a reasonable fix almost ready for review. It's currently blocked by a tangentially-related problem. I think the fix will be available either in next week's release or the week after.

backcover7:
> @backcover7 I have what I believe to be a reasonable fix almost ready for review. It's currently blocked by a tangentially-related problem. I think the fix will be available either in next week's release or the week after.

@nmote Thank you for the notification. Would you mind linking to a valid playground case here when the fix is applied?

nmote:
Sure, will do

nmote:
A little late to the game here, but this is out in v0.112.x. I should have a fix for #6051 shortly.

PR #6011, opened 2022-08-31
refactor: Generic machinery to try multiple pattern parsers

nmote:
In #5980, I introduced logic to first try the pfff parser when parsing
patterns for JS/TS, and then the tree sitter parser, rather than relying
solely on the pfff parser as we had been doing previously. Now I find
myself wanting to do this for Java as well, so it is time to make this
logic more generic.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/semgrep-core/src/parsing/Parse_pattern.ml
+++ b/semgrep-core/src/parsing/Parse_pattern.ml
@@ -44,6 +44,46 @@ let extract_pattern_from_tree_sitter_result
                pr2 (Tree_sitter_run.Tree_sitter_error.to_string ~color:true err));
       failwith "error parsing the pattern"
 
+type 'ast parser =
+  | Pfff of (string -> 'ast)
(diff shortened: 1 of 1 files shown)


aryx (approved):

aryx:
Regarding your java use case, is it to solve: https://github.com/returntocorp/semgrep/issues/6002 ?

nmote:
No, it's for https://github.com/returntocorp/semgrep/issues/5558. It looks complicated to make that work with an LR(1) parser, and in fact there's already a hack in place in the pfff parser that is apparently intended to address a shift/reduce conflict with the top level constructor rule and something else. I think there may be another shift/reduce conflict that is getting resolved arbitrarily, but I don't get any warnings about it when building.

After spending some time looking into it, I decided it would be easier to fix by extending the tree-sitter grammar, which does a much better job of reporting conflicts and since it's GLR there's no need for hacks when the grammar is not LR(1) but is unambiguous.

aryx:
Good call.

PR #6016, opened 2022-09-01
fix!: multiline tokens range reporting

### LINKED ISSUE #5750: Python matches on multiline strings report ending line number incorrectly
**Describe the bug**
When a match matches a python multiline string:

```python
x = """
    abc
    def
"""
```

The end line number is reported as being the same as the start line number, when in reality the start and end line numbers should be different.

**To Reproduce**
The behavior shows up here: https://semgrep.dev/playground/s/Jg2y

A simpler example can be seen by running `semgrep -e "\$X = \"\$MULTILINE\"" -l py match.py --json` on the string `x = ...` above. The match is reported as starting on line 1 and ending on line 1, but it should really end on line 4.

**Expected behavior**
Line numbers are reported correctly.

**Screenshots**
Output of above "simple example" includes:

```
"end": {
        "col": 28,
        "line": 1,
        "offset": 27
      }
...
"start": {
        "col": 1,
        "line": 1,
        "offset": 0
      }
"""
```

**Environment**
Semgrep.dev, homebrew binary

**Use case**
More predictable behavior and easier debugging in semgrep app!

brandonspark:
**What:**
Currently, multiline tokens (like multiline strings) do not report their ranges correctly.

**Why:**
This causes bugs in the playground, such as the following: 
https://semgrep.dev/playground/s/Jg2y

**How:**
I added a helper function which "flips" a token. For the most part, we use tokens to denote the data of the start of the token - the line, column, and charpos all correspond to that beginning, with the `str` field denoting the actual contents.

There are two places in our code where we are instead interested in the _end_ of a token (specifically, the line and column of the end of a token). This is when we start reporting ranges.

The code currently makes an incorrect assumption that the line and column of a token that it receives is correctly on the same line as the end of the token. This is not the case, however, for all tokens. This causes weird ranges.

The fix was to make a helper that just "flips" a `token_location`. It does so by reading the contents of the string, and by counting the number of newlines, it can synthesize the end line and column from the beginning line and column.

**Alternatives Considered**
I considered using `charpos_to_pos`, but it is expensive, so it would need to be called at the file-level to maximize performance.

Unfortunately, the amount of logic between when Semgrep is considering things as files, and when it is considering the specific tokens it needs to flip, is about as wide as the Gulf of Mexico. I attempted this, but gave up due to the sheer amount of refactoring it required.

This solution is slightly ugly, but it works.

**Test plan:**
`make test`

Closes #5750 (still thinking of you, Ben!)

PR checklist:

### DIFF
--- a/cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_rule_parser__failure__error_messages/settings1/error.txt
+++ b/cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_rule_parser__failure__error_messages/settings1/error.txt
@@ -39,7 +39,7 @@ Files skipped:
 
   [1m[24mSkipped by analysis failure due to parsing or internal Semgrep error[0m
 
-   • [36m[22m[24mtargets/bad/invalid_python.py (1 lines skipped)[0m
+   • [36m[22m[24mtargets/bad/invalid_python.py (2 lines skipped)[0m
 
 
 Some files were skipped or only partially analyzed.
--- a/cli/tests/e2e/rules/multiline.yaml
+++ b/cli/tests/e2e/rules/multiline.yaml
@@ -1,0 +1,28 @@
+rules:
+  - id: assign-multiline-comment
+    message: found an assignment of a multiline comment
+    severity: WARNING
+    languages: [python]
+    pattern: $X = """$MULTILINE"""
+  - id: eqeq-is-bad
+    patterns:
+      - pattern-not-inside: |
+          def __eq__(...):
+              ...
+      - pattern-not-inside: assert(...)
+      - pattern-not-inside: assertTrue(...)
+      - pattern-not-inside: assertFalse(...)
+      - pattern-either:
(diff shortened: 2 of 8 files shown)


aryx:
You need a separate PR for the pfff changes.

aryx (approved):

cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_rule_parser__failure__error_messages/settings1/error.txt
```diff
@@ -39,7 +39,7 @@ Files skipped:
 
   [1m[24mSkipped by analysis failure due to parsing or internal Semgrep error[0m
 
-   • [36m[22m[24mtargets/bad/invalid_python.py (1 lines skipped)[0m
+   • [36m[22m[24mtargets/bad/invalid_python.py (2 lines skipped)[0m
```
IagoAbal, line 42: Do we understand why this changes?

PR #6019, opened 2022-09-01
Simplify Dockerfile and fix Arch Linux failure

### LINKED ISSUE #5715: Permission denied when build semgrep docker image
It bug was discussed into slack: 
https://r2c-community.slack.com/archives/CK86BJ5DW/p1657044622433309

**Describe the bug**
When build image locally I get `The command '/bin/sh -c scripts/install-tree-sitter-lib' returned a non-zero code: 1`

**To Reproduce**
```
git clone
git checkout v0.102.0 -b v0.102.0-build
git submodule update --init --recursive
docker build -t semgrep:v0.102 ./
```
I get this logs:
```
docker build -t semgrep:v0.102 ./
Sending build context to Docker daemon  233.6MB
Step 1/39 : FROM returntocorp/ocaml:alpine-2022-06-09@sha256:99b453a838c9d94414991c0fd7be4711aa1bcc120f576e0f0653c7b921ea9718 as semgrep-core
 ---> dbbb69867321
Step 2/39 : USER root
 ---> Using cache
 ---> f2cb532c2adb
Step 3/39 : RUN apk add --no-cache pcre-dev python3 &&     pip install --no-cache-dir pipenv==2022.6.7
 ---> Using cache
 ---> 9f41d9d356f4
Step 4/39 : USER user
 ---> Using cache
 ---> 8df4905d08d4
Step 5/39 : ENV OPAMYES=true
 ---> Using cache
 ---> 4d7368292b47
Step 6/39 : WORKDIR /semgrep/semgrep-core/src/ocaml-tree-sitter-core
 ---> Using cache
 ---> 519cf6d4c198
Step 7/39 : COPY --chown=user semgrep-core/src/ocaml-tree-sitter-core/ .
 ---> 09f62ec5ff4e
Step 8/39 : RUN scripts/install-tree-sitter-lib
 ---> Running in dd48363b4f2c
mkdir: cannot create directory 'downloads': Permission denied
The command '/bin/sh -c scripts/install-tree-sitter-lib' returned a non-zero code: 1
```
I try to build on different worstastions. 
**Expected behavior**
Build of image is completed.

When I researched this issue, I found that build was broken on v0.97.0 version. Before it all is build, after and v0.97.0 is not builded.  

- [`X` ] P1: important to fix or quite annoying

**Environment**
local build. Arch linux system. 

**Use case**
I have some diffs with semgrep code formatting. I can't update semgrep version for my projects.

aryx:
This closes #5715

The main issue was that WORKDIR does not honor the USER
directive on arch linux (it seems to do on macOS), so
the ocaml-tree-sitter-core directory itself (the '.') was
owned by root which prevented configure to create files
(e.g., tree-sitter-config.ml) in it.

By simplifying the Dockerfile and doing the COPY of
the parent dir, the ocaml-tree-sitter-core directory
is now owned by the user and so configure works.

I've also split some RUN command in 2 because they
were hard to read and I prefer to separate different
concerns in different RUN commands.

test plan:
docker build .

PR checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,6 +55,9 @@ RUN apk add --no-cache --virtual=.run-deps bash git git-lfs openssh
 COPY cli ./
 
 RUN apk add --no-cache --virtual=.build-deps build-base
+
+# DL3013: Pin versions in pip.
+# hadolint ignore=DL3013
 RUN SEMGREP_SKIP_BIN=true pip install /semgrep
 # running this pre-compiles some python files for faster startup times
 RUN semgrep --version
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,8 +58,8 @@ COPY cli ./
 RUN apk add --no-cache --virtual=.build-deps build-base
 RUN SEMGREP_SKIP_BIN=true pip install /semgrep
 # running this pre-compiles some python files for faster startup times
-RUN semgrep --version && apk del .build-deps && mkdir -p /tmp/.cache
-
+RUN semgrep --version
+RUN apk del .build-deps && mkdir -p /tmp/.cache
 # those files are not needed anymore
 RUN rm -rf /semgrep
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,40 +21,21 @@ FROM returntocorp/ocaml:alpine-2022-06-09@sha256:99b453a838c9d94414991c0fd7be471
 USER root
 # for ocaml-pcre now used in semgrep-core
 # TODO: update root image to include python 3.9
-RUN apk add --no-cache pcre-dev python3 python3-dev &&\
-     pip install --no-cache-dir pipenv==2022.6.7
+RUN apk add --no-cache pcre-dev python3 python3-dev
+RUN pip install --no-cache-dir pipenv==2022.6.7
 
 USER user
 
 ENV OPAMYES=true
 
-WORKDIR /semgrep/semgrep-core/src/ocaml-tree-sitter-core
-COPY --chown=user semgrep-core/src/ocaml-tree-sitter-core/ .
-RUN ./configure \
- && ./scripts/install-tree-sitter-lib
-
-WORKDIR /semgrep/semgrep-core/src/pfff
-COPY --chown=user semgrep-core/src/pfff/*.opam .
-WORKDIR /semgrep/semgrep-core/src/ocaml-tree-sitter-core
-COPY --chown=user semgrep-core/src/ocaml-tree-sitter-core/*.opam .
-WORKDIR /semgrep/semgrep-core
-COPY --chown=user semgrep-core/*.opam .
-
-RUN opam install --deps-only \
-     /semgrep/semgrep-core/src/pfff \
-     /semgrep/semgrep-core/src/ocaml-tree-sitter-core \
-     /semgrep/semgrep-core
-
+# TODO: why copy the current dir to /semgrep and run the command from there?
+COPY --chown=user . /semgrep
 WORKDIR /semgrep
-COPY --chown=user semgrep-core/ ./semgrep-core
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -21,40 +21,21 @@ FROM returntocorp/ocaml:alpine-2022-06-09@sha256:99b453a838c9d94414991c0fd7be471
...
-WORKDIR /semgrep/semgrep-core
-COPY --chown=user semgrep-core/*.opam .
-
-RUN opam install --deps-only \
-     /semgrep/semgrep-core/src/pfff \
-     /semgrep/semgrep-core/src/ocaml-tree-sitter-core \
-     /semgrep/semgrep-core
-
+# TODO: why copy the current dir to /semgrep and run the command from there?
+COPY --chown=user . /semgrep
```
mjambon, line 32: I'm not sure why there were different partial COPY instructions instead of just one. Normally, this is done to exclude some content. For this, it's best to rely on the `.dockerignore` mechanism, though. So, lacking a justification for the previous complexity, it's good that it's now simpler.

aryx: Exactly. I could understand than for the final docker image (The python part), we might want to optimize and not copy everything, but anyway for the semgrep-core docker image we build in step1, we just care about the semgrep-core binary, so we don't care about the rest of the docker image.

mjambon:
> By simplifying the Dockerfile and doing the COPY of
> the parent dir, the ocaml-tree-sitter-core directory
> is now owned by the user and so configure works.

Why? Is this a bug in some component used on Arch Linux? If you understand what makes a difference, it would be good to add a comment about it.

Edit: the issue is when WORKDIR needs to create the directory if it doesn't exist. The ownership of this directory is implementation-dependent (or buggy, whatever), so we should run `RUN mkdir -p` beforehand to ensure it's owned by the current user.

Dockerfile
mjambon: Sometimes, we want to exclude random junk to increase determinism, but I don't see how that would be a valid argument here.

mjambon (approved):

aryx:
hmm weird, the docker image is far bigger now: https://hub.docker.com/layers/returntocorp/semgrep/pr-6019/images/sha256-e4f406c8d33526702671b0d61f644192ddef828e478c994573d114bdcdfbfac7?context=explore
127MB vs 64MB before

aryx:
We should probably add a check in scripts/docker-validate-xxx.sh to make sure the size does not grow up too much.

aryx:
I'll revert for now.
Investigating a bit more, looks like @underyx did lots of changes in #5442 that I was reverting here, but I was probably wrong
because the image got far bigger.

Dockerfile
```diff
@@ -73,19 +54,24 @@ ENV PIP_DISABLE_PIP_VERSION_CHECK=true \
...
+
+# DL3013: Pin versions in pip.
 # hadolint ignore=DL3013
-RUN apk add --no-cache --virtual=.build-deps build-base && \
-     SEMGREP_SKIP_BIN=true pip install /semgrep && \
-     # running this pre-compiles some python files for faster startup times
-     semgrep --version && \
-     apk del .build-deps && \
-     mkdir -p /tmp/.cache
+RUN SEMGREP_SKIP_BIN=true pip install /semgrep
```
aryx, line 61: @mjambon maybe that's why the docker image grew from 64MB to 127MB? That's why we need to "consolidate"
and put the RUN commands together?
cc @brendongo and @underyx

underyx: Yep. File deletions only reduce image size if they're in the same RUN instruction that created them.

aryx: wow, really? Why? So how can we save space by deleting all /semgrep/tests?
Be more precise at the COPY cli ./?
and ignore tests/ subdir?

PR #6020, opened 2022-09-01
remove old version of semgrep before running

spencerdrak:
Because we now re-use infra, the `brew install semgrep` wasn't doing anything - leading to a de-sync and failing brew tests. This is sub-optimal (ideally we don't change our workflows to suit the infra) but we don't have a good way (or really any way) to hook into the workflow and do this on the runner. 

Maybe we ought to investigate docker-izing the runner, but even then, we have a) figure out how to run the container not in the linux VM that docker uses by default (very difficult, in my estimation) and b) figure out how to spawn a new one on each run (not trivial).

PR checklist:

### DIFF
--- a/.github/workflows/homebrew-core-head.yml
+++ b/.github/workflows/homebrew-core-head.yml
@@ -21,6 +21,8 @@ jobs:
         run: brew install semgrep --HEAD
       - name: Check installed correctly
         run: brew test semgrep --HEAD
+      - name: Clean up semgrep installation
+        run: brew uninstall semgrep
 
   notify-failure:
     needs: [brew-build]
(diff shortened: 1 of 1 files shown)


.github/workflows/homebrew-core-head.yml
```diff
@@ -11,6 +11,10 @@ jobs:
     name: Build Semgrep via Brew from `returntocorp/semgrep:develop`
     runs-on: ["self-hosted", "macOS", "X64"]
     steps:
+      - name: Uninstall semgrep
+        # This is sub-optimal - our workflows shouldn't have to conform to their environment.
+        # However, on the runner side, we can't hook into the workflow run to clean up after.
+        run: brew uninstall semgrep || true
```
zyannes, line 17: Perhaps we should also run this as a cleanup step at the end of the job. That way the assumption can be that on a nominal run it resets the environment back to start.

zyannes (approved):

.github/workflows/homebrew-core-head.yml
spencerdrak: I like that idea.

PR #6021, opened 2022-09-01
test infra: do not assert_match for test_baseline.py

aryx:
The git output can vary across versions, we should not assert
on its exact content.

test plan:
make test
now works on my Arch Linux (with git 2.37.1)

PR checklist:


nmote (approved):
Interesting, I wonder if this is why I always have failures on the main `semgrep` test suite. I haven't spent much time looking into it -- instead when I do have to change the python side, I just run the test suite that should be affected by my change and let CI handle the rest. It's been working okay so far, but it would be nice if all tests passed locally.

aryx:
Yes, I used to be tolerant to those failures too before "ok, 3 failures in 'make tests', I think it's the usual one", but it's dangerous. Far better to have it work locally too, not just in CI.

PR #6022, opened 2022-09-01
fix(java): `public Foo() {}` as a pattern

### LINKED ISSUE #5558: Java constructors with modifiers patern not detected

nmote:
The pfff parser rejects `public Foo() {}` as a pattern, even though the
grammar appears to allow it and even though `Foo {}` (without privacy
modifiers) is parsed correctly. There is already a hack in place to
address an LR(1) conflict for this rule, and I suspect that it has gone
wrong somehow. Instead of spending time refining the hack, I decided to
use tree-sitter as a fallback to parse Java patterns.

This updates semgrep-java to pull in
https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/339,
which implements basic Semgrep pattern grammar extensions and top-level
constructors.

This also updates semgrep-rules to pull in
https://github.com/returntocorp/semgrep-rules/pull/2366.

Fixes #5558

Test plan: Automated tests

PR checklist:

### DIFF
--- a/changelog.d/gh-5558.fixed
+++ b/changelog.d/gh-5558.fixed
@@ -1,0 +1,1 @@
+Fix parsing of Java constructors with privacy modifiers as patterns, e.g. the pattern `public Foo() { }`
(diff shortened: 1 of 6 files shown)


brandonspark (approved):
looks good! love tree-sitter migration

aryx (approved):

PR #6023, opened 2022-09-01
fix!: allowed other matching to use m_todo_kind

brandonspark:
**What:**
Currently, `OtherPat` matching, when matching the contained `todo_kind`, uses `m_other_pattern_operator`, which is a wrapper for `m_other_xxx`, a physical equality check. Physical equality on tokens is going to return false a lot of the time, because they contain sensitive location data. We should use `m_todo_kind` instead, like `m_expr` does for matching `OtherExpr`s.

**Why:**
This causes silly matching bugs, like https://semgrep.dev/s/ZL97

**How:**
Updated to use `m_todo_kind`.

I also then updated a few more instances of `m_other_xxx` for a `todo_kind`.

Closes PA-1836.

PR checklist:

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -2467,7 +2467,7 @@ and m_pattern a b =
   | G.PatWhen (a1, a2), B.PatWhen (b1, b2) ->
       m_pattern a1 b1 >>= fun () -> m_expr a2 b2
   | G.OtherPat (a1, a2), B.OtherPat (b1, b2) ->
-      m_other_pattern_operator a1 b1 >>= fun () -> (m_list m_any) a2 b2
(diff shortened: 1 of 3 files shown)


brandonspark:
Should we replace all of the `m_other_<stuff>` calls with `m_todo_kind`? I don't see a reason for us to ever do a physical equality check during matching between tokens. @nmote @aryx

aryx:
Yep, for todo_kind we should.
It was done for all those ExprTodo, StmtTodo that used to have a long list of possible constructors, where it's ok to rely on the default polymorphic structural equality (we never used the physical equality; we were using structural equality)

brandonspark:
> Yep, for todo_kind we should.
It was done for all those ExprTodo, StmtTodo that used to have a long list of possible constructors, where it's ok to rely on the default polymorphic structural equality (we never used the physical equality; we were using structural equality)

Oh yes, I meant structural equality. Whoops.

I've now changed the other uses as well.

aryx (approved):

PR #6033, opened 2022-09-02
parse newlines before guards in scala

brandonspark:
**What:**
We currently fail to parse Scala programs which contain newlines before the guard of generator expressions.

**Why:**
This breaks some use cases. For instance, we cannot parse:
```
for {
  x <- 3
  if b 
} yield 3
```

**How:**
I just added a thing which, irregardless of if there is a newline or space, checks beyond it to see if the next token is an `if`. If so, then it enters the `guard` loop.

**Test plan:**
`make test`

PR checklist:

### DIFF
--- a/semgrep-core/src/analyzing/Constant_propagation.ml
+++ b/semgrep-core/src/analyzing/Constant_propagation.ml
@@ -530,7 +530,6 @@ let add_special_constants env lang prog =
                logger#trace "adding special terraform constant for %s" (fst id);
                add_constant_env id (terraform_sid, Lit literal) env
            | _ -> ())
-  else ()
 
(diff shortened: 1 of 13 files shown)


aryx (changes requested):
Need to fix the segfault. Too dangerous to merge without understanding why there is a segfault.

aryx:
Did you manage to fix the segfault issue?

aryx:
or to know where did it come from?

aryx:
need to merge again to resolve the conflict

aryx:
I've fixed the CI error in the pfff PR. You should be able to use the latest pfff now and it should work.

PR #6037, opened 2022-09-02
chore: delete deepexprstmt generic test

brandonspark:
**What:**
The `deep_exprstmt.generic` test is wrong, it says that there should be findings, and there are not. This borks the cheatsheet regenerating script, which expects findings. Also, it doesn't really make sense to do a deep expr-stmt thing on a Generic file.

**Why:**
We should fix this so we don't have wrong tests, and also so we can regenerate the cheatsheet.

**How:**
I deleted the file.

PR checklist:


aryx:
Why this has become an issue? How old is this file? What broke?

aryx (approved):

aryx:
Those tests were maybe not even run?

brandonspark:
> Those tests were maybe not even run?

Yes, these tests were not run. The suite of tests which goes through and finds tests with their associated `.sgrep` files seems to not visit the `POLYGLOT` folder, so these tests are not automatically run (though they are checked by the cheatsheet script).

aryx:
The POLYGLOT directory is used by the testsuite. It's to factorize .sgrep file that are the same with different languages.
For example when we see a tests/java/foo.java test file, we look for tests/java/foo.sgrep for the pattern, but if it's not there we look for tests/POLYGLOT/foo.sgrep.

PR #5981, opened 2022-08-25
refactor: combine two POST requests to upload findings and ignores into one POST request

pangngan:
Previously, the CLI would send two POST requests to the APP: one to upload regular and any experimental findings and another to upload ignored findings. However, as the findings can be sent in one POST request body, we decided that it was unnecessary to have two separate POST requests to send over the information. This PR combines the two POST requests to upload findings and ignores into POST request and updates the test snapshots accordingly.

PR checklist:

- [x] Each source file starts with an up-to-date [summary of why it exists](https://semgrep.dev/docs/contributing/contributing-code/#explaining)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure on any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

Closes APP-1446

PR #6038, opened 2022-09-03
Allow backend sending rules as json instead of yaml string

brendongo:
PR checklist:

### DIFF
--- a/cli/src/semgrep/util.py
+++ b/cli/src/semgrep/util.py
@@ -39,7 +39,7 @@ def is_url(url: str) -> bool:
 
 
 def is_rules(rules: str) -> bool:
-    return rules[:6] == "rules:"
+    return rules[:6] == "rules:" or rules[:8] == '{"rules"'
 
 
 def path_has_permissions(path: Path, permissions: int) -> bool:


brendongo:
@chmccreery https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/config_resolver.py#L634 is a little weird so it was hard to track this down. If we have time might be worth making this less fragile :D

chmccreery (approved):
Very weird, I could swear I already made this change. Maybe there was a bad merge resolution or something. Thanks for tracking down! Agree it's a bit fragile...

underyx (approved):

bmahe:
Hi @brendongo @chmccreery @underyx ,

I am not clear on the purpose of this change.
Would it be possible to:
* Add a description of the purpose of the change?
* Add a test. It looks like the original state was not correct and we had no test to catch that. So can we add a test?

underyx:
@bmahe note that the old line from this PR never made it to a release. So the description would be: this was a follow-up to https://github.com/returntocorp/semgrep/pull/5981 to add support for when the backend sends rules serialized as JSON instead of YAML.

chmccreery:
Agree with the concern though that this wasn't caught by existing CI tests. @brendongo what prompted this change/what is the use-case that was not supported?

underyx:
> what prompted this change/what is the use-case that was not supported?

We changed the backend today to use JSON serialization instead of YAML serialization for this field.

chmccreery:
> We changed the backend today to use JSON serialization instead of YAML serialization for this field.

So shouldn't we update at least some of the tests to use JSON serialized data instead of YAML to prevent an effective revert of this change?

underyx:
yep, that sounds like a good idea

underyx:
I looked into it and there's no tests for any of this config parsing at all, no test case to extend or modify. Unfortunately I don't have the time to write the whole test suite for this so I'm dropping your request @bmahe @chmccreery.

chmccreery:
Ok, I've made a linear ticket for this https://linear.app/r2c/issue/APP-2245/add-tests-for-config-parsing-in-semgrep-cli

PR #6048, opened 2022-09-06
Only import jsonnet when necessary

emjin:
Since jsonnet rules are an internal experiment for now, let's
not require users to have the jsonnet dependency. People who
want to try out jsonnet rules can `pip install jsonnet`.

PR checklist:

### DIFF
--- a/cli/Pipfile
+++ b/cli/Pipfile
@@ -44,6 +44,5 @@ typing_extensions = "*"
 importlib_resources = ">=1.4.0"
 
 [packages]
-semgrep = {editable = true,path = "."}
 jsonschema = "==4.6.1"
-jsonnet = "*"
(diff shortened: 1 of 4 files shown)


brendongo (approved):

spencerdrak (approved):

underyx (approved):

aryx:
What was not working? Pip isn't supposed to install dependenciew automatically when you do pip install semgrep?

spencerdrak:
@aryx It does in a linux environment - however, homebrew has it's own way of doing dependency management. We would've had to update our brew formula to declare the dependency.

I have an action to investigate ways we might smooth out that process.

aryx:
I hope it also does in a macOS environment. pip install semgrep should also work in macOS right?
I guess the problem is more with brew install semgrep.

PR #6049, opened 2022-09-06
feat: Transitivity Detection for Supply Chain

mmcqd:
This PR adds support for detection of whether a dependency is direct or transitive to Semgrep Supply Chain. Currently only support python and JS. It works by reading the manifest file associated with a given lockfile.
PR checklist:


cli/src/semdep/parse_lockfile.py
```diff
@@ -21,7 +23,7 @@
 
 logger = getLogger(__name__)
 
-from semgrep.semgrep_interfaces.semgrep_output_v0 import FoundDependency
+from semgrep.semgrep_interfaces.semgrep_output_v0 import FoundDependency, Transitivity
```
brendongo: You import Transitivity twice

cli/src/semdep/parse_lockfile.py
```diff
@@ -45,38 +51,40 @@ def extract_npm_lockfile_hash(s: str) -> Dict[str, List[str]]:
...
-            parsed_name = "@".join(first_quoted.split("@")[:-1])
-            return parsed_name
+def parse_yarn(
+    lockfile_text: str, manifest_text: Optional[str]
+) -> Generator[FoundDependency, None, None]:
+    def version_sources(line: str) -> Tuple[str, List[Tuple[str, str]]]:
+        constraint_strs = line[:-1].split(",")
+        constraints = []
+        for c in constraint_strs:
+            c = c.strip().strip('"')
```
brendongo: Does .strip("\" ") work or is it possible there are spaces after "

cli/src/semdep/parse_lockfile.py
```diff
@@ -45,38 +51,40 @@ def extract_npm_lockfile_hash(s: str) -> Dict[str, List[str]]:
...
-        if '"' not in line:
-            return line.split("@")[0]
-        else:
-            first_quoted = "".join(line.split('"')[1:2])
-            parsed_name = "@".join(first_quoted.split("@")[:-1])
-            return parsed_name
+def parse_yarn(
+    lockfile_text: str, manifest_text: Optional[str]
+) -> Generator[FoundDependency, None, None]:
+    def version_sources(line: str) -> Tuple[str, List[Tuple[str, str]]]:
```
brendongo, line 57: Would be helpful to have a comment of what you expect line to look like (and also a quick comment on what this does cause it's not immediately clear from function name

cli/src/semdep/parse_lockfile.py
```diff
@@ -45,38 +51,40 @@ def extract_npm_lockfile_hash(s: str) -> Dict[str, List[str]]:
...
+        constraints = []
+        for c in constraint_strs:
+            c = c.strip().strip('"')
+            if c[0] == "@":
+                name, constraint = c[1:].split("@")
+                name = "@" + name
+            else:
+                name, constraint = c.split("@")
+            constraints.append((name, constraint))
+        return constraints[0][0], constraints
```
brendongo, line 72: Would returning Version/SemVer type here work or does it need to be strings?

cli/src/semdep/parse_lockfile.py
```diff
@@ -89,30 +97,43 @@ def remove_quotes(s: str) -> str:
...
     # Newer versions of NPM (>= v7) use 'packages'
     # But 'dependencies' is kept up to date, and 'packages' uses relative, not absolute names
     # https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json
     if "dependencies" in as_json:
         deps = as_json["dependencies"]
-    elif "packages" in as_json:
-        deps = as_json["packages"]
     else:
-        logger.info("Found package-lock with no 'dependencies' or 'packages'")
+        logger.info("Found package-lock with no 'dependencies'")
```
brendongo: Should this be info or debug?

cli/src/semdep/parse_lockfile.py
```diff
@@ -344,45 +407,64 @@ def parse_dep(line: str) -> Optional[FoundDependency]:
     yield from (dep for dep in deps if dep)
 
 
-def parse_Poetry_str(lockfile_text: str) -> Generator[FoundDependency, None, None]:
+def parse_poetry(
+    lockfile_text: str, manifest_text: Optional[str]
+) -> Generator[FoundDependency, None, None]:
+    # poetry.lock file are not quite valid TOML >:(
```
brendongo, line 417: :(

cli/src/semdep/parse_lockfile.py
brendongo: I also cant find these Classes in the atd repo

cli/src/semgrep/target_manager.py
```diff
@@ -725,7 +726,16 @@ def get_lockfile_dependencies(
         lockfiles = self.get_files_for_language(ecosystem).kept
         parsed: Dict[Path, List[FoundDependency]] = {}
         for lockfile in lockfiles:
-            deps = parse_lockfile_str(lockfile.read_text(encoding="utf8"), lockfile)
+            path, lockfile_pattern = lockfile.parents[0], lockfile.parts[-1]
```
brendongo: nit: indexing always makes me nervous but it's probably fine

brendongo (approved):

cli/src/semdep/parse_lockfile.py
mmcqd: probably debug

cli/src/semdep/parse_lockfile.py
mmcqd: I don't think making them Versions would do anything. If "^1.0.0" appears in the manifest then "^1.0.0" will appear in the lockfile. So string equality vs semver equality will be the same.

cli/src/semdep/parse_lockfile.py
mmcqd: I need to make a PR there

cli/src/semdep/parse_lockfile.py
mmcqd: no that should work

PR #6075, opened 2022-09-09
chore: add not-in to eval op

### LINKED ISSUE #6072: metavariable-comparison: $VAR not in [list] construction does not work
**Describe the bug**
The metavariable-comparison rule does not correctly match based on the following construction:
```yaml
  - metavariable-comparison:
      comparison: $ARG not in [443, 80]
      metavariable: $ARG
```

**To Reproduce**
See [this playground example](https://semgrep.dev/playground/s/YZZW), where [the inverse case works properly](https://semgrep.dev/playground/s/644w).

**Expected behavior**
I would expect the example above to produce a single match.

**Workaround**
The correct behavior can be achieved by using the "not $VAR in [list]" construction (thanks for the hint on Slack, @Sjord!).

**Environment**
Playground and local install.

**Use case**
Matching a metavariable against an allowlist of possible values, and alerting if it has a different value. e.g.: Allowed ports, cryptographic algorithms, ...

brandonspark:
**What:**
We do not currently do evaluation of the `not in` operator, which appears in Python. So for instance, using this clause:
```
  - metavariable-comparison:
      comparison: $ARG not in [443, 80]
      metavariable: $ARG
```
will not properly match if `$ARG` is bound to `5`, for instance.

**Why:**
This is just a regular bug we should fix. This was brought to our attention as something that a user was trying to do via the community slack.

**How:**
Added it to the list of supported operators.

**Test plan:**
`make test`

Closes #6072 

PR checklist:

### DIFF
--- a/changelog.d/gh-6072.fixed
+++ b/changelog.d/gh-6072.fixed
@@ -1,0 +1,1 @@
+Allowed metavariable-comparison to make use of the `not in` operator.
--- a/semgrep-core/src/engine/Eval_generic.ml
+++ b/semgrep-core/src/engine/Eval_generic.ml
@@ -281,6 +281,10 @@ and eval_op op values code =
       match v2 with
       | List xs -> Bool (List.mem v1 xs)
       | _ -> Bool false)
+  | G.NotIn, [ v1; v2 ] -> (
+      match v2 with
+      | List xs -> Bool (not (List.mem v1 xs))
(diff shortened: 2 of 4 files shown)


emjin (approved):
lgtm

IagoAbal:
@emjin this PR adds a feature but there is no mention about updating Semgrep-docs.

PR #6078, opened 2022-09-09
Parse non-optional parts of impl statements

### LINKED ISSUE #6076: Semgrep returns complement of pattern result instead of intended target for some Rust patterns
**Describe the bug**
Results for Rust rules are the complement of what the user expects.

**To Reproduce**
https://semgrep.dev/s/2egz

**Expected behavior**
Expected single finding on line 2 in this example. Received a match for everything *except* line 2.

**Screenshots**
image

**Environment**
semgrep.dev, also reproduces with Semgrep CLI 0.111.1

**Use case**

emjin:
Closes #6076

Test plan: make test

PR checklist:

### DIFF
--- a/semgrep-core/src/parsing/tree_sitter/Parse_rust_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_rust_tree_sitter.ml
@@ -3185,7 +3185,7 @@ and map_declaration_statement (env : env) (*_outer_attrs _visibility*) x :
           v1
       in
       let attrs = deoptionalize [ unsafe_attr ] in
-      let impl = token env v2 (* "impl" *) in
+      let impl = ident env v2 (* "impl" *) in
       let tparams =
         match v3 with
         | Some x -> map_type_parameters env x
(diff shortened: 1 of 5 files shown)


semgrep-core/src/parsing/tree_sitter/Parse_rust_tree_sitter.ml
```diff
@@ -3184,9 +3184,9 @@ and map_declaration_statement (env : env) (*_outer_attrs _visibility*) x :
             G.KeywordAttr (G.Unsafe, tok))
           v1
       in
-      let _attrs = deoptionalize [ unsafe_attr ] in
-      let _impl = token env v2 (* "impl" *) in
-      let _type_paramsTODO =
+      let attrs = deoptionalize [ unsafe_attr ] in
+      let impl = token env v2 (* "impl" *) in
```
nmote: Maybe use `str env v2` here so that you can avoid writing `"impl"` below?

nmote (approved):
I'm not very familiar with Rust, or how Semgrep handles it, but I'm slightly worried that the introduction of an `OtherDef` here could lead to matching inconsistencies if different forms of definitions remain unchanged. Is it possible for someone to write a pattern that gets parsed as a `DefStmt` with an ordinary definition kind, that will then fail to match a definition that goes through this code path, or vice versa? If so, is that likely to result in problems in practice?

emjin:
> I'm not very familiar with Rust, or how Semgrep handles it, but I'm slightly worried that the introduction of an OtherDef here could lead to matching inconsistencies if different forms of definitions remain unchanged. Is it possible for someone to write a pattern that gets parsed as a DefStmt with an ordinary definition kind, that will then fail to match a definition that goes through this code path, or vice versa? If so, is that likely to result in problems in practice?

I think it's fine because we're just parsing both the pattern and the target into an OtherDef. Previously, there was no way at all to match an `impl` since we only parsed the body of the `impl` definition, so I don't think we can break anything

semgrep-core/src/parsing/tree_sitter/Parse_rust_tree_sitter.ml
```diff
@@ -3211,9 +3211,13 @@ and map_declaration_statement (env : env) (*_outer_attrs _visibility*) x :
             ty)
           v4
       in
-      let _tyTODO = map_type_ env v5 in
+      let ty = map_type_ env v5 in
       let _where_clauseTODO = Option.map (fun x -> map_where_clause env x) v6 in
-      map_decls_or_semi env v7
+      let body = map_decls_or_semi env v7 in
+      (* TODO not sure what to put for the name *)
```
aryx, line 3217: not "impl" for sure :)
This is similar to type classes in Haskell. It does not really define an entity, instead it extends the definition
of an existing entity/type (in your example Ptr). We probably need a new construct for that in AST_generic.ml

Issue #5996, opened 2022-08-27
Error while running rules: 0 bytes read on a total of 2 expected bytes


psifertex:
**Describe the bug**
Working on getting the Binary Ninja plugin updated and it looks like something has changed since you've published some rules and they're not compatible anymore? 

**To Reproduce**
1. Install semgrep via pip
2. Create a `sample.c` file as shown below.
3. Download [info-leak-on-non-formatted-string](https://github.com/returntocorp/semgrep-rules/blob/develop/c/lang/security/info-leak-on-non-formatted-string.yaml) rule
4. Run: `semgrep scan -c info-leak-on-non-formatted-string.yaml sample.c`
5. Result is the error in the subject of this bug.

```sample.c
int64_t _main()

{
    return 0;
}
```

**Expected behavior**
The scan should run indicating a result found or not.

**Priority**
How important is this to you?
- P0: blocking me from making progress

**Desktop (please complete the following information):**
 - MacOS
 - Apple Silicon
 - Semgrep 
 ```
semgrep --version
0.111.1
```

**Comment by psifertex** [comment]
Upon further investigation I don't think this is related to the rule I'm using I think it's just a broken semgrep install where the provided error just isn't helpful. Even doing:

```
semgrep --config=auto .
```
inside of `semgrep-rules/c/lang/security` produces the same error:

```
Fetching rules from https://semgrep.dev/registry.

Scanning across multiple languages:
    <multilang> | 52 rules × 46 files
              c | 13 rules × 12 files
           yaml | 23 rules × 11 files

    0%|                                                                                                                                                                                                  
|0/69 tasks[ERROR] Error while running rules: 0 bytes read on a total of 2 expected bytes
    0%|                                                                                                                                                                                                  
|0/69 tasks
```

psifertex:
Yeah, this should probably be migrated to the semgrep repo, I can close/re-open if that's easier instead.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1825/error-while-running-rules-0-bytes-read-on-a-total-of-2-expected-bytes. _Note_: this link is for r2c use only and is not accessible publicly.

aryx:
Can you rerun with --debug and copy-paste the output here?

emcd:
[...]

PR #6082, opened 2022-09-12
Write rules and targets to ~/.semgrep

emjin:
Semgrep writes the rules and targets to temporary files, which are sent to semgrep-core. However, the temporary files are essentially just random names, since we still have to remove them ourselves. This was probably done earlier when we didn't use ~/.semgrep as much.

Now, since we already write to ~/.semgrep, there's no reason not to use this folder to write our rules and targets. In addition, it makes our code path simpler and avoids using implementation details of temporary files (e.g. there is nothing preventing Python from deleting the temporary file for us).

Alternatives considered: wrapping `os.remove` calls in a `try-except`. This would achieve the same purpose of avoiding the possibility of Python deciding to remove the temporary file before we remove it, and maintains the random name of the rules and targets, but makes the code more complicated.

Related to https://github.com/returntocorp/semgrep/issues/5996

The one potentially annoying consequence of this is that people using `-d` cannot rely on the generated `semgrep_rules.json` persisting when they run semgrep without `-d`, but that isn't good behavior to rely on anyway.

Test plan: make test. Also, print out the rule file name and target file name and check that they don't exist after the run.

PR checklist:

### DIFF
--- a/cli/src/semgrep/core_runner.py
+++ b/cli/src/semgrep/core_runner.py
@@ -727,16 +727,8 @@ class CoreRunner:
         profiling_data: ProfilingData = ProfilingData()
         parsing_data: ParsingData = ParsingData()
 
-        rule_file_name = (
-            str(state.env.user_data_folder / "semgrep_rules.json")
-            if dump_command_for_core
-            else tempfile.NamedTemporaryFile("w", suffix=".json").name
-        )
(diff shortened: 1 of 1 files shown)


emjin:
This approach doesn't work because now you can't have semgrep running in two places

PR #6083, opened 2022-09-12
Remove metatypes/critical types experiment

emjin:
Metatypes (aka critical types) were added as an experiment after brainstorming at an offsite. Security researchers were finally able to sit down at test it (see https://docs.google.com/document/d/1Q37NILl3zILVrg0eQyf_USDSJ0GRGvvWD23nmNhpzco/edit)

The results found it wasn't easy to write rules which benefited from critical types immediately. Some of these cases would benefit more from the jsonnet experiment! Some need better type propagation for this to be useful. In the meantime, I'm removing critical types to reduce maintenance burden. If we want it again, we can add it back.

Test plan: make test (no critical types tests were added yet :D)

PR checklist:


IagoAbal:
In these cases I think it's great having a reference to the PR we're rolling back: #5249.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -453,22 +453,9 @@ let rec m_name a b =
           (* Try matching against parent classes *)
           try_parents dotted)
   (* extension: metatypes *)
```
IagoAbal: This comment should be removed too?

IagoAbal (approved):

PR #6084, opened 2022-09-12
Guard removal of rules and targets files

emjin:
Semgrep creates temporary rules and targets files to send to semgrep-core. These files may not be removed by the temporary file system. To avoid undefined behavior, we try to manually remove the file but do not crash if we fail.

Related to https://github.com/returntocorp/semgrep/issues/5996

Test plan: make test

PR checklist:

### DIFF
--- a/cli/src/semgrep/core_runner.py
+++ b/cli/src/semgrep/core_runner.py
@@ -908,8 +908,16 @@ class CoreRunner:
                     parsing_data.add_error(err)
             errors.extend(parsed_errors)
 
-        os.remove(rule_file_name)
-        os.remove(target_file_name)
+        try:
+            # We use NamedTemporaryFile for the rules and targets files to
+            # create files with random names, but because of how we open
+            # the files we need to remove them ourselves. However, we are
+            # not guaranteed that a different implementation would not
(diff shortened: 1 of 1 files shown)


underyx (approved):

cli/src/semgrep/core_runner.py
```diff
@@ -908,8 +908,16 @@ def _run_rules_direct_to_semgrep_core(
...
-        os.remove(rule_file_name)
-        os.remove(target_file_name)
+        try:
+            # We use NamedTemporaryFile for the rules and targets files to
+            # create files with random names, but because of how we open
+            # the files we need to remove them ourselves. However, we are
+            # not guaranteed that a different implementation would not
+            # remove the file. Therefore, we wrap this in a try-except
+            os.remove(rule_file_name)
+            os.remove(target_file_name)
```
underyx, line 925: note that this assumes both files would be deleted together

chmccreery: Yeah, might be nice to put these into separate try/except blocks so that the `logger.warning` line could be more specific about which file(s) was not found.

chmccreery (approved):
Thanks @emjin!

PR #6089, opened 2022-09-12
chore(swift): Upgrade semgrep-swift

nmote:
See https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/343 for details on the changes here.

Notably, this fixes the issue where:

```
foo
...
```

was not parsed as two statements, where one was `foo` and the next was the Semgrep ellipsis. This fix should make writing Swift rules more intuitive.

Test plan: Automated tests

PR checklist:


semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -472,16 +473,72 @@ let map_as_operator (env : env) (x : CST.as_operator) =
   | `As_quest tok -> (* as_quest_custom *) token env tok
   | `As_bang tok -> (* as_bang_custom *) token env tok
 
+let map_referenceable_operator (env : env) (x : CST.referenceable_operator) =
```
brandonspark, line 476: was there a reason for this to switch positions?

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
```diff
@@ -3069,7 +3149,15 @@ and map_type_parameters (env : env) ((v1, v2, v3, v4) : CST.type_parameters) =
         v2)
       v3
   in
-  let v4 = (* ">" *) token env v4 in
+  (* TODO handle type constraints? *)
```
brandonspark, line 3152: might these be able to be folded into the `tp_attrs` on the type parameter `v2`? seems they're modifiers on the parameter, i think

brandonspark (approved):
I reviewed the `semgrep-swift` PR but not this one. Oops.

In the words of the wise sages of the East: "looks good to me"

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
nmote: Yeah, it is now referenced in a function that was previously defined above it.

semgrep-core/src/parsing/tree_sitter/Parse_swift_tree_sitter.ml
nmote: I'm not actually sure what these are, but I figured that since they weren't even parsed before, it probably wasn't urgent to handle them. Would you be okay with leaving this for later?

nmote: Normally I would expect a type constraint to exist on each individual type parameter, and not as a separate list after the other type parameters. That's where my confusion is and why I think I'd need to spend a bit of time understanding the construct before deciding how to handle it.

brandonspark: I think before, what I did was I folded type constraints into the entity of the defined thing, or the nearest type.

It's weird because I went and looked at the Swift grammar, and to my understanding our `type_constraints` are the same as `generic-where-clause`, which is not permitted to appear after a singular type parameter, as it does here. My assumption would be that these type constraints (which appear right after a singular type parameter) are modifiers on the type parameter, but given that it doesn't seem to be in the grammar, I'm unsure if that's accurate.

In any case, I think it's fine to just merge this. I'm not even sure why this change was made to the `tree-sitter` grammar.

alex-pinkus: Indeed, the current official grammar does not support this; a `generic-parameter-clause` is just made up of a comma-separated list of `generic-parameter` items, separated by a comma, and surrounded by `<` `>`. This syntax was changed in [`SE-0081`](https://github.com/apple/swift-evolution/blob/main/proposals/0081-move-where-expression.md), prior to which there was permitted to be a single `where` clause within the `generic-parameter-clause`. The motivation in SE-0081 was that where clauses can get very long, with an example given of:

```
func anyCommonElements<T : SequenceType, U : SequenceType where
    T.Generator.Element: Equatable,
    T.Generator.Element == U.Generator.Element>(lhs: T, _ rhs: U) -> Bool
{
    ...
}
```

These days that code fails to compile, printing out:
```
error: 'where' clause next to generic parameters is obsolete, must be written following the declaration's type
```

With the old behavior, the single `where` clause would contain type constraints that apply to any of the generic parameters.

It's reasonable to ask why I added support for it now, when the real compiler dropped support six years ago. The answer, actually, is `semgrep`! I was looking at the repositories [in your parsing-stats directory](https://github.com/returntocorp/semgrep/blob/aadd498a94274c13071031ab41c50c4991bbcdc4/parsing-stats/lang/swift/projects.txt) and I noticed that one of them [uses this obsolete syntax](https://github.com/ReactKit/ReactKit/blob/swift/2.0/ReactKit/Stream.swift#L221). That repository was last updated in 2015, so I suspect it simply doesn't build with the latest compiler. There are a few others for which I made some legacy syntax additions as well.

brandonspark: @alex-pinkus Oh, thank you for the in-depth explanation!

That's super considerate of you actually -- I hadn't realized that you had done this on our behalf. Thanks so much for your help, you've been very accommodating :)

nmote: Nice, thanks for the explanation and for the change!

PR #6090, opened 2022-09-12
fix: semgrep blocking on triaged issues

chmccreery:
DO NOT MERGE: I am considering whether to instead change this on the app side. If semgrep CLI has consistently used these other variable names, then by changing the app instead we could fix older as well as current/future versions of semgrep.

----------

Here's another place where true, typed interfaces would have saved us. 

Since we don't have that yet, I am also trying to figure out a setup in pytest that will mock the /deployments/scans/config response and verify that ScanConfig has the correct properties as a result.

So far, I have tested this manually. I verified that a blocking finding locally returned status code 1. Then I triaged it to ignored and verified that it still had status code 1. Then I made these code changes and verified that it returned status code 0.

PR checklist:

Closes APP-2149


chmccreery:
Closing to fix on app side instead, which will be released sooner and also then fix this issue for folks running older versions of semgrep (this has been broken since before 0.110.0 and maybe never worked properly).

PR #6098, opened 2022-09-13
fix: add backwards compatibility with old endpoints

chmccreery:
We have a customer still running a version of semgrep-app from March 2022 on-prem. These changes give backwards-compatibility with that version of the app for that customer specifically.

Test plan:

Tested against version of semgrep-app from Aug 3, 2022 before the refactor on app. To do this, I downgraded the migrations back through Aug 1, then checked out develop from that date (git checkout `git rev-list -n 1 --before="2022-08-03 13:37" develop`) then ran:

`SEMGREP_APP_URL="http://localhost:5000" SEMGREP_APP_TOKEN=<secret> python -m semgrep ci`

PR checklist:

### DIFF
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -332,8 +332,12 @@ class ScanHandler:
             json=findings_and_ignores,
         )
         # TODO: delete this once all on-prem app instances are gone
-        if response.status_code == 404:
-            response = state.app_session.post(
+        if (
+            response.status_code == 404
+            and state.env.semgrep_url != "https://semgrep.dev"
+        ):
+            # order matters here - findings sends back errors but ignores doesn't
+            ignores_response = state.app_session.post(
                 f"{state.env.semgrep_url}/api/agent/scans/{self.scan_id}/ignores",
                 json={"findings": ignores},
             )
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -116,6 +116,47 @@ class ScanHandler:
                 f"API server at {state.env.semgrep_url} returned type '{type(response.json())}'. Expected a dictionary."
             )
 
+    def fetch_config_and_start_scan_old(self, meta: Dict[str, Any]) -> None:
+        """
+        Get configurations for scan and create scan object
+        Backwards-compatible with versions of semgrep-app from before Aug 2022
+        TODO: Delete once all old, on-prem instances have been deprecated
+        """
+        state = get_state()
+
+        logger.debug("Getting deployment information")
+        get_deployment_url = f"{state.env.semgrep_url}/api/agent/deployments/current"
+        body = self._get_scan_config_from_app(get_deployment_url)
+        self._deployment_id = body["deployment"]["id"]
+        self._deployment_name = body["deployment"]["name"]
+
+        logger.debug("Creating scan")
+        create_scan_url_old = (
+            f"{state.env.semgrep_url}/api/agent/deployments/{self.deployment_id}/scans"
+        )
+        response = state.app_session.post(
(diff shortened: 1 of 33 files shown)


pangngan (approved):

cli/src/semgrep/commands/ci.py
```diff
@@ -297,8 +297,11 @@ def ci(
                 # Note this needs to happen within fix_head_if_github_action
                 # so that metadata of current commit is correct
                 if scan_handler:
-                    scan_handler.fetch_and_init_scan_config(metadata_dict)
-                    scan_handler.start_scan(metadata_dict)
+                    try:
```
brendongo, line 300: Nice

cli/src/semgrep/app/scans.py
```diff
@@ -290,6 +331,16 @@ def report_findings(
...
         )
+        # TODO: delete this once all on-prem app instances are gone
+        if response.status_code == 404:
+            response = state.app_session.post(
+                f"{state.env.semgrep_url}/api/agent/scans/{self.scan_id}/ignores",
+                json={"findings": ignores},
+            )
+            response = state.app_session.post(
+                f"{state.env.semgrep_url}/api/agent/scans/{self.scan_id}/findings",
+                json=findings_and_ignores,
```
brendongo, line 346: This should just be findings right?

cli/src/semgrep/app/scans.py
```diff
@@ -228,21 +269,21 @@ def report_findings(
...
-            ],
+            "findings": findings,
             "searched_paths": [str(t) for t in targets],
-            "rule_ids": all_ids,
-            "ignores": [
-                match.to_app_finding_format(commit_date).to_json()
-                for match in new_ignored
-            ],
+            "rule_ids": rule_ids,
+            "cai_ids": cai_ids,
```
brendongo, line 285: Do we need to send back cai_ids?

cli/src/semgrep/app/scans.py
```diff
@@ -290,6 +331,16 @@ def report_findings(
             f"{state.env.semgrep_url}/api/agent/scans/{self.scan_id}/findings_and_ignores",
             json=findings_and_ignores,
         )
+        # TODO: delete this once all on-prem app instances are gone
+        if response.status_code == 404:
```
brendongo: Can we also add a condition here that state.env.semgrep_url is not https://semgrep.dev so we only fall back to this on VPC

cli/src/semgrep/commands/ci.py
```diff
@@ -297,8 +297,11 @@ def ci(
...
+                        scan_handler.fetch_and_init_scan_config(metadata_dict)
+                        scan_handler.start_scan(metadata_dict)
+                    except Exception:
+                        scan_handler.fetch_config_and_start_scan_old(metadata_dict)
```
brendongo: Probably want the same condition as mentioned in above comment

brendongo (changes requested):
-

cli/src/semgrep/app/scans.py
chmccreery: Yeah, but technically it's the same, just minus the "ignores" key so I figured we could just re-use this (sending extra fields doesn't hurt functionality)

cli/src/semgrep/app/scans.py
chmccreery: I think we might, since their app pre-dates us getting rid of CAI and I don't know if they have CAI enabled. Without this.

cli/src/semgrep/app/scans.py
brendongo: Ah I see

brendongo (approved):

PR #6102, opened 2022-09-15
feat: Support semgrepconfig file for scan metadata

zyannes:
PR checklist:

- [x]  Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)

### DIFF
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -30,6 +30,7 @@ from semgrep.meta import GitMeta
 from semgrep.metrics import MetricsState
 from semgrep.output import OutputHandler
 from semgrep.output import OutputSettings
+from semgrep.project import ProjectConfig
 from semgrep.rule import Rule
 from semgrep.rule_match import RuleMatchMap
 from semgrep.state import get_state
@@ -298,6 +299,8 @@ def ci(
                 # so that metadata of current commit is correct
                 if scan_handler:
                     try:
+                        proj_config = ProjectConfig.load_all()
+                        metadata_dict = {**metadata_dict, **proj_config.to_dict()}
                         scan_handler.fetch_and_init_scan_config(metadata_dict)
                         scan_handler.start_scan(metadata_dict)
                     except Exception as e:
--- a/cli/src/semgrep/project.py
+++ b/cli/src/semgrep/project.py
@@ -1,13 +1,27 @@
 """
 Utility for identifying the URL of the current git project
 """
+import json
+import re
+from copy import deepcopy
+from pathlib import Path
+from typing import Any
+from typing import Dict
+from typing import List
 from typing import Optional
 
+import ruamel.yaml
+from attr import define
+from attr import Factory
+
+from semgrep.git import get_git_root_path
 from semgrep.util import git_check_output
 from semgrep.util import manually_search_file
 from semgrep.verbose_logging import getLogger
 
 logger = getLogger(__name__)
+
+CONFIG_FILE_PATTERN = re.compile(r"^\.semgrepconfig(\.yml|\.yaml)?$")
 
 
 def get_project_url() -> Optional[str]:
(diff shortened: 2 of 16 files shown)


cli/src/semgrep/commands/ci.py
```diff
@@ -298,6 +299,8 @@ def ci(
                 # so that metadata of current commit is correct
                 if scan_handler:
                     try:
+                        proj_config = ProjectConfig.load_all()
+                        metadata_dict = {**metadata_dict, **proj_config.to_dict()}
```
chmccreery, line 303: So it looks like semgrepconfig keys will take precedence over metadata set elsewhere - is that a deliberate choice/something we need to document?

cli/src/semgrep/project.py
```diff
@@ -24,3 +38,71 @@ def get_project_url() -> Optional[str]:
...
+            config: Dict[str, Any] = yaml.load(fp)
+            return config
+
+    @classmethod
+    def load_all(cls) -> "ProjectConfig":
+        src_directory = get_git_root_path()
+        cwd_path = Path.cwd()
+        conf_files = cls._find_all_config_files(src_directory, cwd_path)
+
+        # Sort by depth asc
```
chmccreery: This means deeper files are later in list and thus take precedence, yes?

cli/src/semgrep/project.py
```diff
@@ -24,3 +38,71 @@ def get_project_url() -> Optional[str]:
...
+
+    def to_dict(self) -> Dict[str, Any]:
+        # Convert all tag values to strings
+        data = deepcopy(self.metadata)
+        tags = data.pop("tags", None)
+        if tags:
+            data["tags"] = []
+            for tag in tags:
+                if isinstance(tag, dict):
+                    data["tags"].append(json.dumps(tag))
```
chmccreery: Might be nice to call out explicitly (with examples?) the types of "tags" data you are supporting here and why you are flattening them all to be of type string. It strikes me as a little odd that we support standalone-string tags and also dictionary-type tags. Can we simplify and just support one or has that ship sailed?

cli/src/semgrep/project.py
```diff
@@ -24,3 +38,71 @@ def get_project_url() -> Optional[str]:
...
+                else:
+                    data["tags"].append(str(tag))
+        # Strip out version
```
chmccreery, line 122: why?

cli/tests/unit/test_project_config.py
```diff
@@ -0,0 +1,116 @@
...
+def test_projectconfig__find_all_config_files_basic(git_tmp_path):
+    dir_files = ["test.py", "main.py", "setup.py"]
+    create_mock_dir(git_tmp_path, dir_files)
+    make_semgrepconfig_file(git_tmp_path, CONFIG_TAGS)
+    config_files = ProjectConfig._find_all_config_files(git_tmp_path, git_tmp_path)
+    assert config_files == [git_tmp_path / ".semgrepconfig"]
+
+
+@pytest.mark.quick
+def test_projectconfig__find_all_config_files_monorepo(git_tmp_path):
```
chmccreery, line 32: This is such a beautiful test! It's so readable! 🥇

chmccreery (approved):
Looks good! Super readable, and great tests! 🎉 Only question of substance is around our support of multiple formats for tags, and whether we can unify those somehow (even if not in the .semgrepconfig, maybe by the time they get to the app?)

cli/src/semgrep/project.py
OffBy0x01: From a user perspective, when editing config files, it would be clearer how data will be used/what is expected of me with a dict. For example, if I received a config template with something like this, I could run semgrep without ever adding the correct config data:

```
{
    'name=',
    'animal=',
    'age='
}
```

Just by nature of the parser - I wouldn't be able to do the same with a config template like this (which is a good thing!):

```
{
    'name':,
    'animal':,
    'age':
}
```

cli/src/semgrep/project.py
zyannes: That is correct!

cli/src/semgrep/project.py
zyannes: Great point! I'll update the ProjectConfig to have explicit fields/types and add a validation method.

cli/src/semgrep/project.py
zyannes: Simplified in latest commit

cli/src/semgrep/commands/ci.py
zyannes: Yes this is intentional since if metadata/tags are checked into an SCM, then it should be the source of truth. I viewed this as a similar process to terraform: once a resource is managed by terraform, it will overwrite changes made, for example, in the console.

aryx:
Is there some documentation on this .semgrepconfig file? What is its format? Which fields are allowed?
Or is it still an experimental feature?

zyannes:
> Is there some documentation on this .semgrepconfig file? What is its format? Which fields are allowed? Or is it still an experimental feature?

It's still technically experimental. I'll work with the team to get docs updated for SemgrepConfig once this is released.

chmccreery (approved):
Looks good! Might be nice to add a test case for the issue that @OffBy0x01 mentioned, too!

aryx:
Would be nice to have access to some docs before, so the PA team can also give feedback on it given our work on jsonnet is a bit related to it (I think we can have both, because they serve different purpose, but would be nice to be in the loop).

chmccreery:
> Would be nice to have access to some docs before

I agree with this - and generally worry that we've been a bit remiss on docs updates recently. Maybe you could just make a google doc and link it here so the changes aren't user-facing yet @zyannes, and then when it becomes user-facing, the doc you wrote would help the TW team update our user-facing docs appropriately.

zyannes:
@chmccreery there is information (and examples) in the scope doc which is linked in Linear.

aryx:
Then I think it's ok to give a link to the scope doc or the linear ticket in your PR, so people not familiar with it can just follow the link.
Actually what is this scope doc? I don't see anything related in https://drive.google.com/drive/folders/13ZODLU-6UMWfzLlQdiz-T4Gu64wr-K5B our google shared drive in the scope folder.

chmccreery:
@aryx I believe it is this one: https://docs.google.com/document/d/1Ffp7trrVe9rKzvSLrlb6VaMFBRkW5kW2No7XMMVdxxY/edit#heading=h.1djkkjt93l5y

(it was linked in the Linear project, so if you follow the linear ticket that this closes, and then click to the right of the project label where there is a little arrow, you can get to the linear project, which has a link to the scope doc)

aryx:
thanks @chmccreery . Please add scope docs to the shared R2C google drive in the Scope docs folder, so things are more discoverable by other (right now I've added a link from there to the doc, but it would be better to move the doc there).

chmccreery:
I don't seem to have permission for that on this doc, but maybe @zyannes does?

Issue #5686, opened 2022-07-06
Unable to have a `focus-metavariable` within a `pattern-either` 


enncoded:
**Describe the bug**
I have a `pattern-either` that only has `focus-metavariable` inside of it:

```
- pattern-either:
  - focus-metavariable: $PATH
  - focus-metavariable: $ARGS
```

However, the rule is considered invalid when I try to run/save it.

**To Reproduce**
I'm unable to save the rule on the editor because its invalid syntax, so here's a [Gist](https://gist.github.com/enncoded/442290934503ab36e811d9b655c3be01) (in new playground rule format).

**Expected behavior**
I'd like to be able to match a lot of different function calls that have at least one of two metavariables as arguments and then specify both of those metavariables as potential sinks. In this example, I don't care which sink is matched, as they're both dangerous.

From my understanding, this error is occurring because `focus-metavariable` doesn't match anything on its own, but the things in a `pattern-either` need to match something. However, in this case the matching is already done by the previous `pattern`s and I just want to focus in on stuff that's already matched.

**What is the priority of the bug to you?**

- [x] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
Playground and CLI with Semgrep v0.103.0

**Use case**
This will allow me to write rules where I want to target specific function arguments in multiple different contexts a lot easier, or in other contexts where focusing on multiple already matched things may be useful.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1607/unable-to-have-a-focus-metavariable-within-a-pattern-either. _Note_: this link is for r2c use only and is not accessible publicly.

IagoAbal:
@enncoded could you please provide us with the rule that you wanted to write (and the test code) ?

enncoded:
I posted the relevant section of the rule above under the "To reproduce" section, but I can post the full one on the Linear issue, along with the test code.

enncoded:
We decided to enhance `focus-metavariable` itself instead of changing the behavior of `pattern-either`. In addition to taking in a single metavariable as a YAML string (which has been the behavior so far), it can also to take in a list of metavariables.

PR #6105, opened 2022-09-16
`focus-metavariable` can now take multiple metavariables

enncoded:
Addresses https://github.com/returntocorp/semgrep/issues/5686.

This PR allows you to specify a list of metavariables, in addition to the prior behavior of being able to specify a single one. In the list case, it will act as a union of the ranges of focused code.

For example, if you have a rule such as

```
patterns:
  - pattern: foo($X, ..., $Y)
  - focus-metavariable:
    - $X
    - $Y
```

and run it on the code `foo(1, 2, 3, 4)`, both `1` and `4` would be highlighted, as it is unioning the code matched by `$X` and `$Y`.

However, running a similar rule

```
patterns:
  - pattern: foo($X, ..., $Y)
  - focus-metavariable: $X
  - focus-metavariable: $Y
```

and run it on the same code, nothing will be highlighted, as it is intersecting the code matched by `$X` and `$Y`.

PR checklist:


semgrep-core/src/core/Rule.ml
```diff
@@ -73,6 +73,8 @@ type formula =
    *)
   | Inside of tok * formula
 
+and focus_mv_list = tok * MV.mvar list [@@deriving show]
```
emjin, line 78: Add a comment explaining what this is since it's a bit confusing

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -322,41 +322,99 @@ let run_selector_on_ranges env selector_opt ranges =
...
-let apply_focus_on_ranges env focus_mvars ranges : RM.ranges =
-  (* this will return a new range (or a no match) *)
-  let apply_focus_mvar (range : RM.t) (focus_mvar : MV.mvar) : RM.t option =
-    let* _mvar, mval =
-      List.find_opt
-        (fun (mvar, _mval) -> MV.equal_mvar focus_mvar mvar)
-        range.RM.mvars
+let apply_focus_on_ranges env (focus_mvars_list : R.focus_mv_list list)
+    (ranges : RM.ranges) : RM.ranges =
+  let intersect (r1 : RM.t) (r2 : RM.t) : RM.t option =
```
emjin, line 328: Minor: I personally would define intersect closer to where it's used so that people reading this aren't primed to think about intersecting ranges

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -322,41 +322,99 @@ let run_selector_on_ranges env selector_opt ranges =
...
+    if Range.( $<=$ ) r1.r r2.r then Some r1
+    else if Range.( $<=$ ) r2.r r1.r then Some r2
+    else None
+  in
+  (* this will return a list of new ranges that have been restricted by the variables in focus_mvars *)
+  let apply_focus_mvars (focus_mvars : MV.mvar list) (range : RM.t) : RM.t stack
+      =
+    let fm_mvals =
```
emjin, line 341: I would include comments at each step about what the variable represents. Most importantly, in the last comment clarify that the list represents a union of the fm_mvals ranges.

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -322,41 +322,99 @@ let run_selector_on_ranges env selector_opt ranges =
...
+             focused_ranges)
+    in
+    let intersect_ranges (list1 : RM.t stack) (list2 : RM.t stack) : RM.t stack
+        =
+      match (list1, list2) with
+      | [ range1 ], [ range2 ] -> (
+          match intersect range1 range2 with
+          | Some r -> [ r ]
+          | None -> [])
+      | [ _ ], []
```
emjin, line 408: I'm confused by this; why do we allow empty list in just this case? This is probably obvious but I don't want to think

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -322,41 +322,99 @@ let run_selector_on_ranges env selector_opt ranges =
...
-           final)
-         (Some init_range)
+  let apply_focus_mvars_list init_range =
+    (* focus_mvars_list and focus_ranges_list are lists of focus statements under a single `patterns`,
+     * ie: [focus-metavariable: $A, $B; focus-metavariable: $X, $Y]
+     *)
+    let focused_ranges_list =
+      focus_mvars_list
+      |> Common.map (fun (_tok, focus_mvars) ->
+             (* focus_mvars is [$A, $B] in a single "focus-metavariable: $A, $B" *)
```
emjin: I can't quite parse this

changelog.d/gh-5686.added
```diff
@@ -0,0 +1 @@
+Fixes https://github.com/returntocorp/semgrep/issues/5686. You can now have multiple metavariables under `focus-metavariable`.
```
emjin: I would add a quick explanation of why this is useful

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -322,41 +322,99 @@ let run_selector_on_ranges env selector_opt ranges =
...
+    in
+    let rec intersect_ranges_list (l : RM.t stack stack) : RM.t stack =
+      match l with
+      | [] -> failwith "No focus-metavariable statements found."
+      | [ first ] -> first
+      | first :: second :: tail ->
+          (* Intersecting two separate focus-metavariable statements at a time *)
+          intersect_ranges_list ([ intersect_ranges first second ] @ tail)
+    in
+    intersect_ranges_list focused_ranges_list
```
emjin, line 427: This is an interesting approach---why not use a fold?

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -330,6 +330,17 @@ let parse_str_or_dict env (value : G.expr) : (G.ident, dict) Either.t =
       error_at_expr env value
         "Wrong field for a pattern, expected string or dictionary"
 
+let parse_focus_mvs env (key : key) x =
```
emjin: Add a comment here too about what the different types represent

emjin (approved):
Looks good! Just some small notes. Also, it might be nice to include a test that uses this in the taint context

semgrep-core/src/engine/Match_search_mode.ml
enncoded: Because it can return an empty list on line 393, which is then put back into the list on 411, so it'll appear at some point during a later recursion

semgrep-core/src/engine/Match_search_mode.ml
enncoded: What if I changed it to:

```
             (* focus_mvars is a list of the metavariables under a single
                focus-metavariable statement.

                eg: focus_mvars = [$A, $B] for
                " - focus-metavariable:
                    - $A
                    - $B"*)
```

semgrep-core/src/engine/Match_search_mode.ml
enncoded: I think that's the closest that I can put it, because it's used on both line 371 and 391, which are in different scopes.

semgrep-core/src/engine/Match_search_mode.ml
enncoded: i don't think i was able to figure out how to get the types that the function returns to match up with the init, because iirc fold enforces that those are the same.

enncoded: also figuring out how to do this recursively was fun and felt cool

semgrep-core/src/engine/Match_search_mode.ml
emjin: Yeah, that makes sense!

emjin:
Failing test is unrelated

enncoded:
Failing test is for the test that uses this feature, so it won't pass until this is merged.

enncoded:
Force pushed because I accidentally did `git pull --rebase` in my local branch oops

PR #6106, opened 2022-09-16
fix(python): typed expression translation bug

brandonspark:
**What:**
A little while ago, I made the PR #5949, which introduces a regression which only a DeepSemgrep test catches. Essentially, sole standalone casted expressions are no longer parsed as `DefStmt`s, which has semantic meaning in DeepSemgrep.

**Why:**
This is not the desired behavior, and makes it so that naming fails for class fields, which rely on the `DefStmt`. See the following file:
`self_ex.py`
```
class Ex:
    #DEF: Ex.fld1
    fld1: int
    fld2_dead_ok: int
    fld3: int = 42

    #DEF: Ex.bar
    def bar():
        return 1

    def other_dead_ok():
        return 1

    #DEF: Ex.foo
    def foo():
        #USE: Ex.fld1
        x = self.fld1
        #USE: Ex.bar
        self.bar(x)
        return 0

    
def test_dead_ok():
    #USE: Ex.foo
    Ex().foo();
```
This test makes `fld1` invisible to DeepSemgrep.

**How:**
Made it so that `Cast`s of `Name`s are properly translated to `DefStmt`s, as they were before.

**Test plan:**
`make test` in DeepSemgrep (I've tested that this fixes the regression).

Closes PA-1881.

PR checklist:

### DIFF
--- a/semgrep-core/src/parsing/pfff/Python_to_generic.ml
+++ b/semgrep-core/src/parsing/pfff/Python_to_generic.ml
@@ -735,6 +735,13 @@ and stmt_aux env x =
   | AugAssign (v1, (v2, tok), v3) ->
       let v1 = expr env v1 and v2 = operator v2 and v3 = expr env v3 in
       [ G.exprstmt (G.AssignOp (v1, (v2, tok), v3) |> G.e) ]
+  | Cast (Name (id, _kind, _ref), _tok, ty) ->
+      (* no need to guard with assign_to_vardef here *)
+      let id = name env id in
+      let ty = type_ env ty in
+      let ent = G.basic_entity id in
+      let def = G.VarDef { G.vtype = Some ty; vinit = None } in
+      [ G.DefStmt (ent, def) |> G.s ]
   | Cast (e, tok, ty) ->
       [ G.exprstmt (G.Cast (type_ env ty, info tok, expr env e) |> G.e) ]
   | Raise (t, v1) -> (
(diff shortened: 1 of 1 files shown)


semgrep-core/src/parsing/pfff/Python_to_generic.ml
```diff
@@ -735,6 +735,13 @@ and stmt_aux env x =
   | AugAssign (v1, (v2, tok), v3) ->
       let v1 = expr env v1 and v2 = operator v2 and v3 = expr env v3 in
       [ G.exprstmt (G.AssignOp (v1, (v2, tok), v3) |> G.e) ]
+  | Cast (Name (id, _kind, _ref), _tok, ty) ->
+      (* no need to guard with assign_to_vardef here *)
+      let id = name env id in
+      let ty = type_ env ty in
+      let ent = G.basic_entity id in
+      let def = G.VarDef { G.vtype = Some ty; vinit = None } in
```
IagoAbal, line 750: May be worth adding a comment that this is translated in this very specific way for DeepSemgrep's consumption?

IagoAbal (approved):
Great!

semgrep-core/src/parsing/pfff/Python_to_generic.ml
IagoAbal: Well, after a second thought, I think this translation makes more sense in general anyways. What feels weird is that we parse `fld1 : int` as a cast in the first place. 🤷‍♂️

PR #6113, opened 2022-09-16
generic mode: Allow short text targets without 2D layout

### LINKED ISSUE #6071: Issues when scanning long lines with `generic` mode
**Describe the bug**

* when you run a generic rule, this one specifically https://semgrep.dev/s/7PnQ on a one line snippet of code - it is not working
when you add extra line to a snippet, like here: https://semgrep.dev/s/E6zN - it works
same for the CLI
also if you make first line even longer https://semgrep.dev/playground/s/7PQ2 - it is not working again (with 1 extra line)
but if I add one more extra line https://semgrep.dev/s/Le26 - it works again

* it seems like this behavior also affects the `autofix` feature since it works incorrectly on such lines (but can be not related :v:)
running https://semgrep.dev/s/7PnQ in the CLI will incorrectly fix last finding

**To Reproduce**
https://semgrep.dev/s/7PnQ

mjambon:
Closes https://github.com/returntocorp/semgrep/issues/6071

PR checklist:


aryx (approved):

aryx:
not sure what's going on with the CI failure

changelog.d/gh-6071.changed
```diff
@@ -0,0 +1,6 @@
+generic mode: allow text input without human-readable indentation up to 500
```
IagoAbal, line 1: Can't we handle more than 500 bytes without problems? If we don't raise it then I think we should make it configurable.

mjambon: No, minified or binary code can be extremely slow to handle, so we skip such files. Allowing single-line files up to 500 bytes is a quick fix to prevent a common problem that rule authors run into. The generic mode is for source code, i.e. human-readable text with meaningful indentation.

Removing this and other limits completely can be done with the `--force` option of `spacegrep` to developers who have access to it.

PR #6118, opened 2022-09-17
style(CLI): modify output to separate non-blocking and blocking findings

pangngan:
This PR modifies the CLI output such that now findings are separating into blocking and non-blocking findings. Additionally, if there are blocking findings, we also show a list of blocking rules that were fired. 

image

PR checklist:

Closes APP-2306

### DIFF
--- a/cli/src/semgrep/formatter/text.py
+++ b/cli/src/semgrep/formatter/text.py
@@ -584,7 +584,10 @@ class TextFormatter(BaseFormatter):
         if first_party_blocking_rules:
             formatted_first_party_blocking_rules = [
                 with_color(Colors.foreground, rule_id, bold=True)
-                for rule_id in set(first_party_blocking_rules)
+                for rule_id in sorted(
+                    set(first_party_blocking_rules),
+                    key=first_party_blocking_rules.index,
+                )
             ]
             first_party_blocking_rules_output = (
                 [
--- a/cli/src/semgrep/formatter/text.py
+++ b/cli/src/semgrep/formatter/text.py
@@ -500,7 +500,7 @@ class TextFormatter(BaseFormatter):
                     first_party_blocking.append(match)
                     rule_id = match.match.rule_id.value
                     # When ephemeral rules are run with the -e or --pattern flag in the command-line, the rule_id is set to -.
-                    # The short rule is ran in the command-line and has no associated rule_id
+                    # The rule is ran in the command-line and has no associated rule_id
                     if rule_id != "-":
                         first_party_blocking_rules.append(rule_id)
                 else:
--- a/cli/src/semgrep/formatter/text.py
+++ b/cli/src/semgrep/formatter/text.py
@@ -499,6 +499,7 @@ class TextFormatter(BaseFormatter):
                 if match.is_blocking:
                     first_party_blocking.append(match)
                     rule_id = match.match.rule_id.value
+                    # When ephemeral rules are run with the -e or --pattern flag in the command-line, the rule_id is set to -
                     if rule_id != "-":
                         first_party_blocking_rules.append(rule_id)
                 else:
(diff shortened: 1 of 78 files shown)


cli/src/semgrep/formatter/text.py
```diff
@@ -491,10 +491,18 @@ def format(
...
+        first_party_blocking = []
+        first_party_nonblocking = []
+        first_party_blocking_rules = []
         for match in rule_matches:
             if "sca_info" not in match.extra:
-                first_party.append(match)
+                if match.is_blocking:
+                    first_party_blocking.append(match)
+                    rule_id = match.match.rule_id.value
+                    if rule_id not in first_party_blocking_rules and rule_id != "-":
```
sabrinabrogren: nit: you could make `first_party_blocking_rules` a set so you don't have to check for duplicates

pangngan: Done! I had to sort the set because I wanted to keep the order of the rules listed the same as what you'd see in the list of findings but if listing alphabetically works, I can change `first_party_blocking_rules` to just be initialized as a set, and just add elements to it as well.

cli/src/semgrep/formatter/text.py
```diff
@@ -543,30 +551,60 @@ def format(
...
+                "\nBlocking Findings:\n" + "\n".join(first_party_blocking_output)
+            )
+
+        first_party_blocking_rules_output = []
+        if first_party_blocking_rules:
+            formatted_first_party_blocking_rules = [
+                with_color(Colors.foreground, rule_id, bold=True)
+                for rule_id in sorted(
+                    set(first_party_blocking_rules),
+                    key=first_party_blocking_rules.index,
```
pangngan, line 591: Sorting the set here because I thought it would make sense for us to return the rules in the same order that is returned by the findings listed right above.

cli/src/semgrep/formatter/text.py
```diff
@@ -491,10 +491,18 @@ def format(
...
+        first_party_blocking = []
+        first_party_nonblocking = []
+        first_party_blocking_rules = []
         for match in rule_matches:
             if "sca_info" not in match.extra:
-                first_party.append(match)
+                if match.is_blocking:
+                    first_party_blocking.append(match)
+                    rule_id = match.match.rule_id.value
+                    if rule_id != "-":
```
chmccreery, line 504: When is the rule_id just a dash? Can you add some comment here to explain it? It feels a little fishy to me

cli/src/semgrep/formatter/text.py
```diff
@@ -543,30 +551,60 @@ def format(
...
                 extra.get("color_output", False),
                 extra["per_finding_max_lines_limit"],
                 extra["per_line_max_chars_limit"],
                 extra["dataflow_traces"],
             )
             findings_output.append(
-                "\nFirst-Party Findings:\n" + "\n".join(first_party_output)
+                "\nFirst-Party Non-Blocking Findings:\n"
+                + "\n".join(first_party_nonblocking_output)
+            ) if (reachable or unreachable) else findings_output.append(
```
chmccreery, line 567: Ooh nice touch!

chmccreery (approved):

cli/src/semgrep/formatter/text.py
pangngan: Just added a comment! It's when we run semgrep using the -e or --pattern flag and so there is no rule_id specified there and is just set to "-" instead.

cli/src/semgrep/formatter/text.py
```diff
@@ -491,10 +491,20 @@ def format(
...
+        first_party_nonblocking = []
+        first_party_blocking_rules = []
         for match in rule_matches:
             if "sca_info" not in match.extra:
-                first_party.append(match)
+                if match.is_blocking:
+                    first_party_blocking.append(match)
+                    rule_id = match.match.rule_id.value
+                    # When ephemeral rules are run with the -e or --pattern flag in the command-line, the rule_id is set to -.
+                    # The short rule is ran in the command-line and has no associated rule_id
```
pangngan:
```suggestion
                    # The rule is ran in the command-line and has no associated rule_id
```

cli/src/semgrep/formatter/text.py
chmccreery: Ah, perfect! Thanks :)

PR #6121, opened 2022-09-18
Add: Release process for tagging semgrep schema files

ajbt200128:
PR checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ jobs:
   create_release:
     name: Create the Github Release
     runs-on: ubuntu-latest
-    needs: [upload-wheels, release-ubuntu-16-04]
+    needs: [upload-wheels, release-ubuntu-16-04, release-schema]
     if: ${{ !contains(github.ref, '-test-release') }}
     steps:
       - name: Get the version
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -76,12 +76,54 @@ jobs:
...
+      - uses: actions/download-artifact@v3
+        with:
+          name: rule-schema-yaml
+          path: rule_schema.yaml
+      - name: Upload Schema Files
+        id: upload-semgrep-schema-files
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release --repo returntocorp/semgrep-interfaces upload ${{ steps.get_version.outputs.VERSION }} rule_schema.json
```
spencerdrak, line 122: I don't know if this will work without having created a release on that repository prior

.github/workflows/release.yml
```diff
@@ -76,12 +76,54 @@ jobs:
...
+      - uses: actions/download-artifact@v3
+        with:
+          name: rule-schema-json
+          path: rule_schema.json
+      - uses: actions/download-artifact@v3
+        with:
+          name: rule-schema-yaml
+          path: rule_schema.yaml
+      - name: Upload Schema Files
+        id: upload-semgrep-schema-files
```
spencerdrak, line 118: We'll need to configure auth for this repo in one of a few different ways.

.github/workflows/release.yml
```diff
@@ -76,12 +76,54 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           gh release --repo returntocorp/semgrep edit ${{ steps.get_version.outputs.VERSION }} --draft=false
+      - name: Publish Release Semgrep Interfaces
+        id: publish_release_semgrep_interfaces
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release --repo returntocorp/semgrep-interfaces edit ${{ steps.get_version.outputs.VERSION }} --draft=false
```
spencerdrak, line 129: Auth needed (see above)

spencerdrak (changes requested):

.github/workflows/release.yml
spencerdrak: Typically a release is created off a tag, so we may need to also create a tag in `returntocorp/semgrep-interfaces`.

spencerdrak:
Looks good overall - few things that need to be addressed through. I am happy to discuss further offline if you'd like @ajbt200128

.github/workflows/release.yml
ajbt200128: Yea looks like we'll have to

.github/workflows/release.yml
ajbt200128: Will the token for this repo not work for the other?

spencerdrak (commented):
Approved pending the set up of the authorization bot. Let's work out the details today about setting that up!

spencerdrak (commented):
Approved - only comment here is whether or not this has been tested. I don't know how you and the team plan to test this - my strategy has been to set it up in a different repo and run it to ensure I've not missed things, but some sort of test might be useful as well. 

If the team is comfortable moving ahead without a test and live debugging that could also work, but poses some risks of interrupting the release which could get messy.

ajbt200128:
I tested the individual commands/scripts, my main concern is the auth step, but that's one I think we'd have to figure out in prod? I'm not sure how to set up a mock semgrep + interfaces + all the permissions in a way I feel like would be an accurate test at least

spencerdrak:
I set up the `semgrep-ci` bot on test-gh-actions as well as this repo - feel free to test there if you feel necessary, but testing the scripts is a great start. I'll defer to you there. Also, you may want to let @IagoAbal know if this is going to be merged before tomorrow morning - we had some issues with release last week and might want to see if those are ironed out before you merge. Otherwise, nice work!

spencerdrak (approved):

changelog.d/app-2313.added
```diff
@@ -0,0 +1,3 @@
+returntocorp/semgrep-interfaces will now release a rule schema file in both yaml
```
IagoAbal: I am not sure this should go into the changelog, the changelog is for user-facing changes and this is an internal change.

scripts/prepare-schema.py
```diff
@@ -0,0 +1,34 @@
...
+from pathlib import Path
+
+from ruamel.yaml import YAML
+
+RULE_SCHEMA_PATH = "cli/src/semgrep/semgrep_interfaces/rule_schema.yaml"
+RULE_SCHEMA_OUT = "rule_schema.json"
+RULE_SCHEMA_YAML_OUT = "rule_schema.yaml"
+
+
+def remove_comment(content):
```
IagoAbal, line 13: Why do we need to remove comments?

IagoAbal:
Given that the previous release was not smooth, if this is not urgent, then we may want to wait to see how this release goes. Is that OK? If you think it's urgent, I'm fine merging it now, but I'll revert it if I the release fails and I suspect it may be this.

scripts/prepare-schema.py
ajbt200128: There's a comment:
```
$comment: "!!If you modify this file, you need to update the submodule in returntocorp/semgrep and returntocorp/semgrep-app!!"
```
which doesn't make sense to me to include as part of the release, so this just removes that

changelog.d/app-2313.added
ajbt200128: Removed

.github/workflows/start-release.yml
```diff
@@ -209,15 +209,22 @@ jobs:
...
-      - name: Create release version tag
+      - name: Create semgrep release version tag
         id: create-tag
         run: |
           git config user.name ${{ github.actor }}
           git config user.email ${{ github.actor }}@users.noreply.github.com
           git tag -a -m "Release ${{ needs.release-setup.outputs.version }}" "v${{ needs.release-setup.outputs.version }}"
           git push origin "v${{ needs.release-setup.outputs.version }}"
+      - name: Create semgrep-interfaces release version tag
+        id: create-tag
```
aryx, line 223: @ajbt200128 @spencerdrak this broke the release, because we have 2 steps with the same id.
Why do we need an id in the first place anyway?
cc @mjambon

spencerdrak: It's' used for referencing the output of the steps later - I wonder if we can write a semgrep rule to check for duplicate step IDs. If you don't reference the step later, the id can be omitted, although I am in the habit of always adding them.

PR #6126, opened 2022-09-19
feat!: named argument tainting

brandonspark:
## What:
Currently, we do not currently support arguments that are _named_ within our model for generating taint signatures for functions. This is witnessed by the fact that the `Arg` variant for taint takes in only an integer, and has no information about the name of the argument. This means that, when passing arguments to a function with named arguments, we will not properly be able to match the argument taints to the taints within the body of the function.

## Why:
This is good for taint maturity, as well as necessary for the current plan for inter-module analysis in Terraform, due to the fact that Terraform modules only have named arguments.

## How:
I made the `Arg` variant take in both a string and an integer, meaning the position that it corresponds to, as well as its name.

Some refactoring was necessary to make everything fit within this type.

At the point where we consume this data, what we do is that we first process the positional arguments (which must come first, by Python's named argument rules), and match their taints to the first `n` arguments.. This leaves us with only named arguments, at which point we go and search for them by name in the list of arguments (which may occur out of order).

## Test plan:
No new tests. Because of the fact that this has to do with taint signatures, which are only inferred in a DeepSemgrep run, we cannot test these changes independently of DeepSemgrep.

As such, I've gone and tested these changes myself using DeepSemgrep. Modulo a match duplication bug, this seems to be working.
PR checklist:


semgrep-core/src/tainting/Taint.ml
```diff
@@ -30,7 +30,7 @@ type 'a call_trace =
 
 type source = Rule.taint_source call_trace [@@deriving show]
 type sink = Rule.taint_sink call_trace [@@deriving show]
-type arg_pos = int [@@deriving show]
+type arg_pos = string * int [@@deriving show]
```
aryx, line 33: We should probably rename arg_pos to arg then and use a variant also here?

aryx (commented):
I'll let @IagoAbal review, but this looks good to me after a quick read.

semgrep-core/src/tainting/Taint.ml
brandonspark: You mean, do you want a unary constructor for the `arg` type?

aryx: I mean in other places you've changed the argument type to change from a simple expression to Unamed | Named.
Now in other code we used to have a type to record the position of an argument (an int), and now we don't care just about position so arg_pos seems like a bad name. Or maybe I don't understand what the string and int mean in this arg_pos type in which can a comment would help clarify.

changelog.d/pa-1886.added
```diff
@@ -0,0 +1 @@
+Added support for named arguments in taint tracking.
```
IagoAbal: Mention that this is only relevant for DeepSemgrep users.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -764,15 +777,15 @@ and arguments env xs = xs |> G.unbracket |> Common.map (argument env)
 
 and argument env arg =
   match arg with
-  | G.Arg e -> expr env e
-  | G.ArgKwd (_, e)
-  | G.ArgKwdOptional (_, e) ->
+  | G.Arg e -> Unnamed (expr env e)
+  | G.ArgKwd (id, e)
+  | G.ArgKwdOptional (id, e) ->
       (* TODO: Handle the keyword/label somehow (when relevant). *)
```
IagoAbal: I would remove this `TODO` then.

semgrep-core/src/core/il/IL_lvalue_helpers.ml
```diff
@@ -23,12 +23,17 @@ let compare_name x y =
   let ident_cmp = String.compare (fst x.ident) (fst y.ident) in
   if ident_cmp <> 0 then ident_cmp else Int.compare x.sid y.sid
 
+let exp_of_arg arg =
```
IagoAbal, line 26: This doesn't fit here very well, since this module is about manipulating l-values. We could rename the module to `IL_helpers` otherwise.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -428,9 +478,10 @@ let rec check_tainted_expr env exp : Taints.t * Lval_env.t =
     | Literal _
     | FixmeExp (_, _, None) ->
         (Taints.empty, env.lval_env)
-    | Composite (_, (_, es, _))
+    | Composite (_, (_, es, _)) -> union_map_taints_and_vars env check es
     | Operator (_, es) ->
-        union_map_taints_and_vars env check es
+        union_map_taints_and_vars env check
+          (Common.map IL_lvalue_helpers.exp_of_arg es)
```
IagoAbal: In general I think I prefer to rewrite this as:

```ocaml
| Operator (_, args) ->
  let es = Common.map IL_lvalue_helpers.exp_of_arg es in
  union_map_taints_and_vars env check es
```
Same applies to the other places where we use `exp_of_arg`.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -251,6 +251,56 @@ let union_taints_filtering_labels ~new_ curr =
           if req then Taints.add new_taint taints else taints)
     new_ curr
 
+let find_args_taints args_taints fdef =
+  let pos_args_taints, named_args_taints =
+    Common.partition_either Fun.id args_taints
+  in
+  let argpos_to_taints = Hashtbl.create 11 in
```
IagoAbal: OCaml hashtbl implementation uses powers of two, and `Hashtbl.create` automatically finds the nearest power of two given the _approximate_ size you provide to it. So in this case it would use 16. I tend to just specify the size as a power of 10.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -251,6 +251,56 @@ let union_taints_filtering_labels ~new_ curr =
...
+  (* We first process the positional arguments, and then the named arguments.
+     This is OK because in Python, positional arguments must appear before named arguments.
```
IagoAbal: You mention "because in Python" but this is not restricted to Python in any way.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -251,6 +251,56 @@ let union_taints_filtering_labels ~new_ curr =
...
+      | Some s -> Hashtbl.add argpos_to_taints (Right s) taints)
+    named_args_taints;
+  fun (s, i) ->
+    let taint_opt =
+      match
+        ( Hashtbl.find_opt argpos_to_taints (Right s),
+          Hashtbl.find_opt argpos_to_taints (Left i) )
+      with
+      | Some taints, _ -> Some taints
+      | _, Some taints -> Some taints
```
IagoAbal, line 314: Can it be `Some _, Some _` ?

IagoAbal (changes requested):
`find_args_taints` seems to be tailored to Python but there are more languages with named arguments where the assumptions you are making don't hold. OCaml is the first example that comes to my mind. Why not search for the named argument first, assuming they can be given in any order. Then, take the remaining unnamed arguments in the same order as they are given, and match them with the remaining parameters. Isn't that more general?

semgrep-core/src/tainting/Taint.ml
IagoAbal: I'm not even sure about the `string * int` type, maybe
```ocaml
type arg = Pos of int | Named of string
```

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -542,21 +588,35 @@ let check_tainted_instr env instr : Taints.t * Lval_env.t =
...
+                 match arg with
+                 | Unnamed _ -> (lval_env, Left taints)
+                 | Named (id, _) -> (lval_env, Right (id, taints)))
                env.lval_env
           |> Common2.swap
         in
+        let all_taints =
+          args_taints
+          |> Common.map (function
+               | Left taints -> taints
```
IagoAbal: I would use variant types here instead of `Left` and `Right`, makes the code easier to read. The only inconvenience is that you can't use `partition_either`, but using a regular `partition` isn't very verbose anyways.

semgrep-core/src/tainting/Taint.ml
brandonspark: That is what it started out as. It seems like you would want a sum here, but actually you don't. This isn't going to work because every argument (in Python, at least) can be used both positionally and in a named fashion. If the function is invoked with that positional information, then you need to know which `arg_pos` to look for based on index. If the function is invoked with named information, then you need to know which `arg_pos` to look for based on name.

semgrep-core/src/tainting/Dataflow_tainting.ml
brandonspark: Yes, but we would prefer that if an argument is identified as corresponding to a named parameter, then it should be associated with that parameter. The mapping from integer argument positions is only in the _absence_ of named arguments.

emjin (commented):
Can we add a test or two for this?

brandonspark:
> Can we add a test or two for this?

I can't add a test for this in `semgrep`, because named argument tainting relies on taint signatures, which are only triggered in DeepSemgrep. I have tests I want to add for DeepSemgrep, but they're out of scope for this PR.

IagoAbal (approved):
Nice!

PR #6127, opened 2022-09-19
fix: allow equality checking for taint traces in pattern matches

brandonspark:
## What:
Currently, `Parse_info.t`s ignore the taint traces when comparing for equality.

This means, in particular, that later assumptions on this equality function make deduplication of pattern matches remove taint traces which have different call sites. This means that even if the same function is called with tainted arguments in multiple places, only once will a finding be reported.

## Why:
We should not do this, because we would like multiple findings from different call-sites, even if to the same sink (and from the same rule)

## How:
Made the equality function properly derived.

## Testing plan:
Unfortunately, none. I can't really test this without DeepSemgrep.

PR checklist:


aryx:
It's not Parse_info.t that ignores taint traces, It's Pattern_match.t

semgrep-core/src/core/Pattern_match.ml
```diff
@@ -74,7 +74,7 @@ type t = {
   (* metavars for the pattern match *)
   env : Metavariable.bindings;
   (* Lazy since construction involves forcing lazy token lists. *)
-  taint_trace : taint_trace Lazy.t option; [@equal fun _a _b -> true]
+  taint_trace : taint_trace Lazy.t option;
```
aryx, line 82: Please add a comment like "we used to have [@ equal fun _a _ b -> true] here, but this was not correct because  ...
we now rely also on the deriving eq of taint_trace (which itself relies on Parse_info.t eq)"

aryx (approved):

IagoAbal:
@brandonspark have you checked how this interacts with the CLI without `--dataflow-traces`? By default, the CLI does not show any taint trace, so the user doesn't see the source and what we know are different findings will look like duplicate findings to the user. There is some deduplication being done by CLI but one would need to check that it removes the duplicates as expected.

PR #6130, opened 2022-09-20
Dockerfile: Fix fragmentation of some identifiers

### LINKED ISSUE #2038: The rule dockerfile.best-practice.missing-image-version.missing-image-version does not handle intermediate stages using earlier stages
**Describe the bug**
e.g.

```
FROM python:3.10.1-alpine3.15@sha256:4be65b406f7402b5c4fd5df7173d2fd7ea3fdaa74d9c43b6ebd896197a45c448 as base

# ruleid: missing-image-version
FROM base AS BUILD
```

**To Reproduce**
Use above example of docker file

**Expected behavior**
Intermediate build steps should not be affected by this rule.

**Screenshots**

P2

**Environment**
semgrep docker

**Use case**
I currently have to run individual rules inside docker ruleset to avoid running dockerfile.best-practice.missing-image-version.missing-image-version

### LINKED ISSUE #5229: Dockerfile metavariables for image aliases are not matching
**Describe the bug**
https://semgrep.dev/playground/s/r2c-sec-research-test:missing-image-version

I am trying to update [this rule](https://semgrep.dev/editor/r/dockerfile.best-practice.missing-image-version.missing-image-version) for [this issue](https://github.com/returntocorp/semgrep-rules/issues/2038). I expected the `pattern-not-inside` clause to filter out the last test case because `$ALIAS` should match to `base`, but it does not.

**To Reproduce**
https://semgrep.dev/playground/s/r2c-sec-research-test:missing-image-version

**Expected behavior**
I expected the last test case in the example provided to not match.

**Use case**
This will help me finish resolving this issue: https://github.com/returntocorp/semgrep-rules/issues/2038

mjambon:
Fixes https://github.com/returntocorp/semgrep-rules/issues/2038
Fixes https://github.com/returntocorp/semgrep/issues/5229

PR checklist:


semgrep-core/src/parsing/tree_sitter/Parse_dockerfile_tree_sitter.ml
```diff
@@ -307,6 +309,7 @@ let unquoted_string (env : env) (xs : CST.unquoted_string) : str =
         | `BSLASHSPACE tok -> String_content (str env tok (* "\\ " *))
         | `Imme_expa x -> expansion env x)
       xs
+    |> simplify_fragments
```
aryx, line 312: I wonder if the type for fragments could be changed to enforce the invariant that you should call simplify_fragments.
(a bit like what Brandon did recently with his new Path.ml module).

semgrep-core/tests/dockerfile/from-alias-reuse.dockerfile
```diff
@@ -0,0 +1,7 @@
+# MATCH:
+FROM a as b
+FROM b
+
+# MATCH:
```
aryx, line 5: would be nice to also have negative test cases.

aryx (approved):

semgrep-core/src/parsing/tree_sitter/Parse_dockerfile_tree_sitter.ml
mjambon: We could have a new type with exactly the same structure but which would have to be created with `simplify_fragments`. I think it's a bit overkill but I'll try it as an exercise.

mjambon: I [tried using a private type abbreviation](https://github.com/returntocorp/semgrep/pull/6144/files#r975918900) to ensure that the type `str` has no fragmented literals but it poses problems with pattern matching.

semgrep-core/tests/dockerfile/from-alias-reuse.dockerfile
mjambon: done

mjambon:
The MacOS [job](https://github.com/returntocorp/semgrep/actions/runs/3094439350/jobs/5007814952) fails because https://opam.ocaml.org/ is down and it's needed to run to fetch package info with `opam update`. The Linux jobs don't need it because they run in a container that already has a recent version of the packages and doesn't run `opam update`.

aryx:
Another good reason for having our own ocaml docker image!

aryx (approved):

PR #6134, opened 2022-09-20
Cron to periodically update semgrep-rules

aryx:
test plan:
wait for CI to kick in and check a PR is created
when pusing to the 'crons' branch (or wait for
the cron to trigger)

PR checklist:


cgdolan (approved):

.github/workflows/crons.yml
```diff
@@ -0,0 +1,38 @@
+name: crons
+on:
+  #TODO: if I put pull_request: instead of push: it does not work, weird
```
aryx: @cgdolan can you comment here on the lines where I have a TODO and answer my question or point me to some tricks to use to fix the TODO. There's a few TODOs in this file where I have no idea how to get it working with github action. I can google it each time, but if you already know the answer that would save me some time.

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
+name: Update Semgrep Rules
+on:
+  workflow_dispatch: {}
```
aryx, line 5: what is this workflow dispatch for?

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
...
+  # TODO: Remove this trigger
```
aryx: actually I like to also have it, so it's easy to test improvement to the cron by pushing to this special cron branch (we do that for our parsing-stat and naming-stat crons and it's convenient instead of waiting for the cron to kick in at the scheduled time)

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
...
+      - "cron"
+  #TODO: at some point also set up an actual cron like
+  #  schedule:
+  #  - cron: "43 20 * * *"
+
+jobs:
+  update-semgrep-rules-submodule:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get JWT for semgrep-ci GitHub App
```
aryx, line 41: why do we need that? What was wrong with the github-action default bot?

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
...
+        id: jwt
+        uses: docker://public.ecr.aws/y9k7q4m1/devops/cicd:latest
+        env:
+          EXPIRATION: 600 # seconds
```
aryx, line 48: why do we need this EXPIRATION?

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
...
+          -H "Accept: application/vnd.github.v3+json" \
+          "https://api.github.com/app/installations/${{ secrets.SEMGREP_CI_APP_INSTALLATION_ID }}/access_tokens" | \
+          jq -r .token)"
+          echo "::add-mask::$TOKEN"
+          echo "::set-output name=token::$TOKEN"
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
+          ref: "${{ github.event.repository.default_branch }}"
```
aryx, line 76: can you explain the goal of this magic incantation? Which problem does it solve?

aryx (commented):
Nice, but overall it would be really great if you add comment explaining all the tricks you use. For devops people maybe this is common knowledge, but to PA people this really looks like incomprehensible magic.

.github/workflows/update-semgrep-rules.yml
aryx: why on pull_request was not working?

.github/workflows/update-semgrep-rules.yml
aryx: why can't we simply use git checkout develop?

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
...
+        id: jwt
+        uses: docker://public.ecr.aws/y9k7q4m1/devops/cicd:latest
+        env:
+          EXPIRATION: 600 # seconds
+          ISSUER: ${{ secrets.SEMGREP_CI_APP_ID }} # semgrep-ci GitHub App id
+          PRIVATE_KEY: ${{ secrets.SEMGREP_CI_APP_KEY }}
+      - name: Get token for semgrep-ci GitHub App
+        id: token
+        run: |
+          TOKEN="$(curl -X POST \
```
aryx, line 56: Is there some URL documenting this thing. I have no idea what's going on and why we need that.

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
...
+          "https://api.github.com/app/installations/${{ secrets.SEMGREP_CI_APP_INSTALLATION_ID }}/access_tokens" | \
+          jq -r .token)"
+          echo "::add-mask::$TOKEN"
+          echo "::set-output name=token::$TOKEN"
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
+          ref: "${{ github.event.repository.default_branch }}"
+          token: "${{ steps.token.outputs.token }}"
```
aryx, line 77: what is this token?

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Update Submodule
+        id: update-submodule
+        working-directory: ./semgrep-core/tests/semgrep-rules
+        run: |
+          git checkout develop
+          git pull
+      - run: git status
+      - name: Add and Commit New Files
+        uses: EndBug/add-and-commit@v9
```
aryx: is it better than what I add before? Why use a special library when the code is simpler than what was before?

.github/workflows/update-semgrep-rules.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          add: cli/tests/e2e/snapshots
+          author_name: ${{ github.actor }}
+          author_email: ${{ github.actor }}@users.noreply.github.com
+          message: "chore: Bump Semgrep Rules Submodule"
+          new_branch: update-semgrep-rules-${{ github.run_id }}-${{ github.run_attempt }}
+      - name: Open PR
+        env:
+          GITHUB_TOKEN: "${{ steps.token.outputs.token }}"
```
aryx, line 124: what is this for? this token?

aryx:
Also this does not seem to work anymore. I don't see any PR request created from this.

.github/workflows/update-semgrep-rules.yml
spencerdrak: The default one doesn't allow it to trigger downstream jobs - if the default token is used to open a PR, then checks won't run on that PR. That API might've just changed, though, I am confirming.

.github/workflows/update-semgrep-rules.yml
spencerdrak: We don't want credentials to have a longer life than they need - this is the shortest setting. Ensure that if an attacker got a hold of these credentials after the job runs, they're expired.

.github/workflows/update-semgrep-rules.yml
spencerdrak: There is not currently, but documenting somewhere is not a bad idea.

spencerdrak: will also add comments.

.github/workflows/update-semgrep-rules.yml
spencerdrak: Using git checkout develop would require some form of configuring URLs or auth using the token - this handles that for us.

.github/workflows/update-semgrep-rules.yml
spencerdrak: explained in code comments.

.github/workflows/update-semgrep-rules.yml
spencerdrak: This handles the case where nothing is available to commit - `git commit` fails if there's no changes, and then we'd have to handle catching that error code ourselves.

.github/workflows/update-semgrep-rules.yml
spencerdrak: Explained in code comments.

spencerdrak:
@aryx I added back the `pull_request` trigger so that it would fire and you can test it again - that will need to be removed before you merge, however, since you may not want all PRs to trigger this job.

.github/workflows/update-semgrep-rules.yml
cgdolan: We are using the standard Github-recommended method for short-lived authentication with a GitHub App.

Documentation can be found here: https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps

.github/workflows/update-semgrep-rules.yml
cgdolan: `develop` may not be the name of the default branch in the repository. It could be `master`, `main`, or any arbitrary other name configured by the repository owner. Using indirection here ensures this workflow continue to run even if the default branch name is changed in the future.

.github/workflows/update-semgrep-rules.yml
cgdolan: See https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow for a more detailed explanation of why the default token is not sufficient.

.github/workflows/update-semgrep-rules.yml
cgdolan: The final output of authenticating as a GitHub app is a short-lived server-to-server token that can be used to perform authenticated operations against the GitHub api. The token must be passed as an input to each operation in order for it to be used.

.github/workflows/crons.yml
cgdolan: Responded to as many of the comments as I could. Hope that helps clarify things.

aryx: It did. Thanks a lot!

.github/workflows/update-semgrep-rules.yml
aryx: I understand it's better to use default_branch, but I still don't understand why 'git checkout develop' fails if run after.

.github/workflows/update-semgrep-rules.yml
aryx: ah great!

aryx:
@spencerdrak I've left some TODOs in the file for things I still don't understand or that would require more doc from you,
but I'm merging now because this is already useful.
Let me know if you know the answer to some of those TODOs at some point.

PR #6138, opened 2022-09-20
fix: Use unique home directory for test_api

zyannes:
PR checklist:

### DIFF
--- a/cli/tests/conftest.py
+++ b/cli/tests/conftest.py
@@ -359,6 +359,12 @@ def _run_semgrep(
     return result
 
 
+@pytest.fixture()
+def unique_home_dir(monkeypatch, tmp_path):
+    monkeypatch.setattr(Path, "home", lambda: tmp_path)
+    yield tmp_path
+
+
 @pytest.fixture
 def run_semgrep():
     yield partial(_run_semgrep, strict=False, target_name=None, output_format=None)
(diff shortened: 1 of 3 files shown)


cli/tests/conftest.py
```diff
@@ -359,6 +359,12 @@ def _run_semgrep(
     return result
 
 
+@pytest.fixture()
```
aryx, line 362: Can you explain (in comments in the code) this code? It's rather cryptic to me.
What's monkeypatch?

cli/tests/e2e/test_api.py
```diff
@@ -8,7 +8,7 @@
 
 
 @pytest.mark.slow
-def test_api(capsys, run_semgrep_in_tmp):
+def test_api(unique_home_dir, capsys, run_semgrep_in_tmp):
```
aryx, line 11: Can you explain (in comments in the code) why we need unique_home_dir here? I've read the changelog but I still don't understand
why we got those flaky errors.

zyannes: Added!

cli/tests/conftest.py
zyannes: Added a comment. `monkeypatch` allows you to set/get/del attributes on any imported class. It is often used in tests for usecases like this one, where the return value of a method needs to be overridden. 

In this particular case, monkeypatch is used to override the `Path.home` return value with the temporary directory created by the `tmp_path` fixture. When the `test_api` test runs and the settings file is read from `Path.home() / ".config/.semgrep/settings.yml"`, instead of checking `/home/runner/.config/.semgrep/settings.yml` it will check `/tmp/.../.config/.semgrep/settings.yml`

spencerdrak (approved):

Issue #4964, opened 2022-04-06
[Fix] Metavariable ellipses sometimes match nothing. In a fix it can be necessary to know this.


0xDC0DE:
**Is your feature request related to a problem? Please describe.**
Sometimes metavariable ellipses (`$...ARGS`) match an empty string. When writing a fix, this is necessary to know. For example, when matching (optional) arguments in a function call, when an empty string is matched, a comma less is needed.

Example rule where this is the case:
https://semgrep.dev/s/pieterdc1:autofix-metavariable-ellipses

The rule 
```
rules:
- id: fix-named-ellipses-commas
  pattern: make_transaction($...BEFORE, secure=False, $...AFTER)
  message: Make sure secure is set to True
  languages: [python]
  severity: WARNING
  fix: make_transaction($...BEFORE, secure=True, $...AFTER)
```

Matches both test cases here:
```
# ruleid: transaction
make_transaction(transaction = "transaction", secure=False)

# ruleid: transaction
make_transaction(transaction = "transaction", secure=False, commit=False)
```

However, to write a correct fix, it is necessary to know whether `$...AFTER` matches an empty string or not. If it matches an empty string, the fix should not have a comma after the `secure=True` argument.

**Describe the solution you'd like**
??? I am not sure, the best solution would be to automagically remove the comma if the metavariable matches an empty string. A fix should always result in syntactically correct code.

**Describe alternatives you've considered**
Writing two rules, one where there are no arguments following the argument we want to match, and one where there are arguments following the argument we want to match.

**Use case**
Better fixes for our users.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1112/[fix]-metavariable-ellipses-sometimes-match-nothing-in-a-fix-it-can-be.

IagoAbal:
We could make autofix automagically remove `, $...X` from the fix when `$...X` is empty. Although this would be quite a hack and could lead to surprises, the fact is that autofix doesn't know the syntax of the target language. We would need to implement autofix in semgrep-core to make it syntax-aware. Do you have experience implementing syntax-aware fixing @pieterdc1 ? IIRC @mmcqd was going to look at taking over autofix?

Otherwise maybe we could have a conditional expansion, something like this (disregard the syntax for now, it's more about the concept):

```python
make_transaction($...BEFORE, secure=True$$commaSep($...AFTER))
```

[...]

Issue #4963, opened 2022-04-06
[Fix] Metavariable ellipsis not correctly copied with autofix


0xDC0DE:
**Describe the bug**
A rather simple, straightforward rule results in weird behaviour after applying the autofix:

```
rules:
- id: fix-named-ellipses-commas
  pattern: make_transaction($...BEFORE, secure=False, $...AFTER)
  message: Make sure secure is set to True
  languages: [python]
  severity: WARNING
  fix: make_transaction($...BEFORE, secure=True, $...AFTER)
```

```
make_transaction(transaction = "transaction", secure=False, commit=False)
```
Is corrected to
```
make_transaction(, secure=True, transaction = "transaction", secure=False, commit=False)
```

**To Reproduce**
https://semgrep.dev/s/pieterdc1:autofix-incorrect

**Expected behavior**
This feels like a straightforward autofix.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [X] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1111/[fix]-metavariable-ellipsis-not-correctly-copied-with-autofix.

nmote:
With the work-in-progress on AST-based autofix, this issue is not fully addressed, but it is dramatically improved: https://semgrep.dev/playground/s/2gr5?develop=true

Now, the second example is autofixed to:

```
make_transaction(secure=True, transaction = "transaction", commit=False)
```

It still reorders the arguments, but since they are keyword arguments this shouldn't matter. It at least generates valid code now.

The reordering is due to how Semgrep does matching in the presence of keyword arguments. The same logic that allows Semgrep to match a pattern with a list of keyword arguments against a target with the same keyword arguments in a different order leads to this unexpected autofix behavior. I'd like to address this at some point, but it seems very low priority now that Semgrep does generate valid code here.

PR #6143, opened 2022-09-21
AST-based autofix

nmote:
This implements end-to-end AST-based autofix in semgrep-core. There are
still numerous unhandled cases. The cases that I did handle are enough
to illustrate the design, and are enough to address the specific bug
reported in https://github.com/returntocorp/semgrep/issues/4964. By design, unhandled cases do not lead to crashes or
incorrect results: Semgrep simply falls back on the existing text-based
autofix.

I have left detailed comments inline, so I won't repeat myself here. In
particular, for an outline of the design, see Autofix.ml.

I made changes to Map_AST and Visitor_AST to support visiting more nodes
that are needed to perform autofix. I expect that we will need to add
support for more nodes to these if we don't switch to autogenerated
visitors via ppx deriving visitors.

I also added a `fix` and `languages` field to `rule_id`. I'm not sure if
this is the best way to plumb that information down to JSON_report, but
there is precedent for that decision with the `message` field. I'm not
clear on why this data structure exists, rather than passing the full
rule around.

Incorrect ranges attached to AST nodes will cause issues with this
approach. Hopefully, the validation pass at the end will catch most of
them. We'll address these issues as they come up, and we may also start
using the ranges provided by tree sitter directly rather than
reconstructing them from tokens. Incorrect ranges also cause problems
with the existing text-based autofix.

I'm comfortable with the existing tests for this change, but as we
improve autofix I think we'll want some better testing infrastructure.
The existing end-to-end tests are thorough, but they are a bit clunky.
It would be nice to have semgrep-core tests that verify that AST-based
autofix is working correctly.

The snapshot change is caused by an additional log line that gets
printed when we parse the fix pattern. It uses a temp file, and Semgrep
prints a log line when that temp file is removed.

Test plan:
- Existing automated tests, including e2e tests that test autofix.
- Manually ran against the example in https://github.com/returntocorp/semgrep/issues/4964 (used only the first
  function call in the target file due to an unrelated issue interfering
  with the second) and verified that the extraneous comma is not added.
- Some ad-hoc performance testing against the youtube-dl repository,
  enough to convince myself that all the parsing and matching logic
  dwarfs any additional work we're doing here for autofix. I can put
  together something more rigorous if reviewers would like.

PR checklist:


aryx (commented):
Please add a small comment at the top of each new .mli file, just what's the purpose of the module, its main entry point, whatever can be useful.

nmote:
Sure, will do. This isn't ready for review yet by the way, that's why I have it marked as a draft. I just wanted to get the full CI running.

semgrep-core/src/core/ast/Ugly_print_AST.ml
```diff
@@ -0,0 +1,122 @@
...
+  object (self)
+    method print_any =
+      function
+      | G.E expr -> self#print_expr expr
+      | G.Ar arg -> self#print_argument arg
+      (* TODO Handle other kinds *)
+      | _ -> None
+
+    method print_expr { e; _ } = self#print_expr_kind e
+    method print_expr_kind _ = None
```
emjin, line 99: Just skimming this, not doing a full review, but maybe a Result type would be more clear? It's not immediately obvious to me that "None" means "failure"

nmote: I don't feel strongly, but I do have a preference for `option` here. `option` is frequently used in the standard library and elsewhere in cases where, for whatever reason, a function might not produce a value. `Result.t` is a little bit more annoying to deal with, but it's worth it when you need to attach some data explaining why you don't have a value. I don't think we need such data here. With that said, I'm happy to make the change if you feel strongly.

changelog.d/ast-autofix.fix
```diff
@@ -0,0 +1 @@
+Start to use the AST to render autofixes in some specific cases. As this is extended, autofix will become more correct and more powerful.
```
aryx, line 1: why not use gh-4964.fixed for the changelog given this actually close the issue?

semgrep-core/src/core/Pattern_match.ml
```diff
@@ -101,6 +101,8 @@ and rule_id = {
   id : string;
   (* other parts of a rule (or mini_rule) used in JSON_report.ml *)
   message : string;
+  fix : string option;
```
aryx, line 107: maybe add a TODO comment saying "maybe we should remove those fields and just have a rule: Rule.t;"

semgrep-core/src/core/ast/Ugly_print_AST.ml
```diff
@@ -0,0 +1,122 @@
...
+    (* TODO Add more nodes as needed. *)
+  end
+
+class python_printer : printer_t =
+  object (self)
+    inherit base_printer
+
+    method! print_arguments (_b1, args, _b2) =
+      (* TODO Consider using original token when available? *)
+      let b1 = Immutable_buffer.of_string "(" in
```
aryx: maybe use some alias because Immutable_buffer.of_string is pretty long to read

aryx: Ideally the code should look like '  combine [[x "("] @ combine ~sep:", " [map_all self#print_argument args] @ [x ")"])
so it's more matching the output.

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
+(* Nat Mote
+ *
+ * Copyright (C) 2019-2022 r2c
```
aryx, line 3: I would put just 2022 but as you want.

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
...
+(******************************************************************************)
+
+(* Attempts to render a fix. If unsuccessful, returns None.
+ *
+ * Failure causes include, but are not limited to:
+ * - The fix pattern does not parse.
+ * - A metavariable is bound to an AST node that is not suitable for the context
+ *   in which it is used in the fix.
+ * - Printing of the resulting fix AST fails (probably because there is simply a
+ *   node that is unhandled).
```
aryx, line 56: probably we want to add some logging or debugging somewhere so when this happens we can quickly know which construct is not handled yet.

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
...
+let render_fix lang metavars ~fix_pattern ~target_contents =
+  (* Fixes are not exactly patterns, but they can contain metavariables that
+   * should be substituted with the nodes to which they are bound in the match.
+   * Because they can contain metavariables, we need to parse them as patterns.
+   * *)
+  let* fix_pattern_ast = parse_pattern_opt lang fix_pattern in
+
+  (* Look through the pattern and replace metavariables with the nodes to which
+   * they are bound in the match. *)
+  let* fixed_ast =
```
aryx: fixed_ast makes me thing of the fixed AST of the target file.
Maybe you can use again fix_pattern_ast again?

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
...
+    Autofix_metavar_replacement.replace_metavars metavars fix_pattern_ast
+  in
+
+  (* Try to print the fixed AST. *)
+  let* text =
+    Autofix_printer.print_ast ~lang ~metavars ~target_contents ~fix_pattern_ast
+      ~fix_pattern fixed_ast
+  in
+
+  (* Perform sanity checks for the resulting fix. If they fail, return None *)
```
aryx, line 96: In theory we should also do some sanity check Python side, to make sure we don't generate unparasable code.

aryx: Actually because we sometimes use the incomplete pfff parser to parse the pattern,
we might have matched with metavar some complex code that is not parsable back by the pattern parser.
Another reason though to switch gradually to tree-sitter for parsing the pattern.

semgrep-core/src/fixing/Autofix.mli
```diff
@@ -0,0 +1,6 @@
+val render_fix :
```
aryx, line 3: Please add a small comment in the mli about the goal of the module and the return type of the function.

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
...
+   * should be substituted with the nodes to which they are bound in the match.
+   * Because they can contain metavariables, we need to parse them as patterns.
+   * *)
+  let* fix_pattern_ast = parse_pattern_opt lang fix_pattern in
+
+  (* Look through the pattern and replace metavariables with the nodes to which
+   * they are bound in the match. *)
+  let* fixed_ast =
+    Autofix_metavar_replacement.replace_metavars metavars fix_pattern_ast
+  in
```
aryx, line 88: Why having this intermediate step where you replace in the fix_pattern_ast the metavar with their value?
Why not in Autofix_printer.print_ast() pass down the metavars, and as you encounter a node
that is a metavar, just look it up and get its corresponding text from the target file (without having to pretty/ugly print anything)?

semgrep-core/src/fixing/Autofix_printer.ml
```diff
@@ -0,0 +1,128 @@
...
+ * which are lifted unchanged from either the target file (via metavariable
+ * bindings) or the fix pattern in the rule. This serves two purposes:
+ * - It lets us synthesize autofixes without having to implement printing for
+ *   all of the AST nodes in the fixed AST.
+ * - It lets us make minimal changes to target files by carrying over
+ *   formatting, comments, etc. from the original code.
+ *)
+(******************************************************************************)
+
+module PythonPrinter = Hybrid_print.Make (struct
```
aryx, line 48: functors and objects, wow, we're using advanced features of OCaml this week (brandon started to use private constructors).

semgrep-core/src/fixing/Hybrid_print.ml
```diff
@@ -0,0 +1,60 @@
...
+ * Specifically designed for autofix, where we can print some nodes by lifting
+ * their original text from the source target or pattern.
+ *)
+(******************************************************************************)
+
+module type Printer = sig
+  class printer : Ugly_print_AST.printer_t
+end
+
+(* Classes are not first-class. Functors are the only way to create a class that
```
aryx, line 36: This sounds complicated to me ... Not sure we could not do the same with simple ocaml records (like I do in Visitor_AST.ml), but let's merge this; we can always improve later (or maybe that's the only solution to this problem).

aryx (approved):

semgrep-core/src/fixing/Autofix.mli
```diff
@@ -0,0 +1,6 @@
...
+  Lang.t ->
+  Metavariable.bindings ->
+  fix_pattern:string ->
+  target_contents:string Lazy.t ->
```
aryx, line 16: maybe explain why the lazy content?
The idea that many things can fail before even trying to grab part of the target file (for the metavar), that
it's better to have the content lazy.

semgrep-core/src/fixing/Autofix_printer.mli
```diff
@@ -0,0 +1,10 @@
+(* Print the given AST to a string, using the original text of nodes from the
+ * fix pattern or target file wherever possible *)
+val print_ast :
+  lang:Lang.t ->
+  metavars:Metavariable.bindings ->
```
aryx, line 8: why do we need the metavars again given the fix_pattern_ast has already been gone though
the Metavar_replacement phase?

semgrep-core/src/fixing/Autofix_printer.mli
```diff
@@ -0,0 +1,10 @@
...
+  target_contents:string Lazy.t ->
+  fix_pattern_ast:AST_generic.any ->
+  fix_pattern:string ->
```
aryx, line 20: why do we pass also the fix_pattern?
What's the difference between the AST_generic.any passed below and the fix_pattern_ast argument?

aryx (changes requested):

semgrep-core/src/core/ast/Ugly_print_AST.ml
```diff
@@ -0,0 +1,122 @@
...
+ * nicely (as opposed to a pretty printer such as those based on Wadler's "A
+ * prettier printer" [1]). It would be difficult, if not impossible, to print
+ * code that conforms to everyone's style guides, linters, and formatters, so
+ * the prettifying of code is left to other tools.
+ *
+ * UglyPrinter can print code snippets, not just full programs. Any of the
+ * `print_` methods can be used by consumers.
+ *
+ * If UglyPrinter encounters a construct that it is unable to handle, it returns
+ * `None`. It is crucial that maintainers preserve this property, since
```
aryx, line 34: why not raising an exception instead?

semgrep-core/src/core/ast/Ugly_print_AST.ml
```diff
@@ -0,0 +1,122 @@
...
+    (* TODO Add more nodes as needed. *)
+  end
+
+(* Some if and only if `f` returns `Some` for all elements. Short circuits when
+ * the first `None` is encountered.
+ *
+ * TODO Make tail recursive?
+ * TODO Move to Common?
+ *)
+let rec map_all f = function
```
aryx, line 77: you would not need that if you were using exceptions instead of None

aryx:
Looks great! I've requested changes, but it's mostly for adding more comments in the .mli
so the whole thing is clearer by just looking at the types and the comments in the .mli

changelog.d/ast-autofix.fix
nmote: I considered it, but in the end I don't feel like this fully addresses that issue. Sure, it fixes the one small test case given there, but if you tweak anything about it, it will no longer work. I'd prefer to have more cases filled in so that this is more general before closing that issue. I was really on the fence though, so if you prefer I can make this change.

semgrep-core/src/fixing/Autofix.ml
nmote: I copied the existing license headers that I found and updated 2021 to 2022. I think we should have some consistent license header along with automated enforcement but that's out of scope for this PR.

semgrep-core/src/fixing/Autofix.ml
nmote: I agree that we should do this check against the actual program text that we generate, right before we write it to disk. Unfortunately that's challenging to do in Python, since we'd have to invoke semgrep-core again to do the parsing. I've been thinking about whether we should do more of this in semgrep-core. @mjambon's experiment to move more of the CLI to OCaml could help with this.

I hadn't considered the possibility that the pattern parser might fail where the target parser would succeed, but that's definitely a valid concern. Fortunately, this is designed to be fail-safe, so worst case we would just fall back on the current textual autofix. We should definitely address this later, though.

semgrep-core/src/fixing/Autofix.mli
nmote: Will do -- sorry, I thought I went through all the mli files and added comments, but I missed this one.

semgrep-core/src/fixing/Autofix.ml
nmote: If we did that, we'd be pretty much back to where we are with textual autofix. In the example given in https://github.com/returntocorp/semgrep/issues/4964, we would encounter the `$...AFTER` argument node while printing, and print the empty string. But, we wouldn't know without a special check to see if it was bound to zero arguments, so we would print the extraneous comma just like we currently do. We could start adding checks like that wherever we think they are needed to the printing code, but I think it's much cleaner to first create a valid generic AST and then print it. Plus, I suspect that as we improve autofix, we will want to do more operations over the resulting AST.

I will add a comment explaining this.

semgrep-core/src/fixing/Hybrid_print.ml
nmote: I thought pretty hard about this problem and I didn't choose this design lightly. I wanted to make sure that printer authors didn't have to concern themselves with the hybrid printing logic. All the other designs I considered would have required people to remember to check the external printer whenever they implement printing for some new node, or would have required us to have exactly one monolithic printer for all languages like we currently have in `Pretty_print_AST`, which I have come to believe is somewhat error-prone. The only way that I saw around this was to use dynamic dispatch, but to get the desired behavior, the plain printer had to be the base class. The only way to achieve that with multiple separate language printers and without duplicating code is via a functor.

It's definitely a bit complicated but I do think this is maintainable. Adding a new hybrid printer for a new language is a simple matter of copying and pasting the functor instantiation that I wrote for Python. Adding support for new nodes doesn't really require the maintainer to understand exactly what's going on -- they just have to copy what's already done for the existing nodes. Plus, I'll probably fill in all of these cases pretty soon, and at that point we'll very rarely have to touch this.

And, most important, the authors of printers don't have to interact with this at all.

I'd be happy to discuss specific ideas for alternative solutions, but I believe I've considered all the reasonable options and I like this one the best. I think the runner-up would be to have a monolithic printer for all languages and then we could just inherit from it directly here without using functors, but I'm not a big fan of that.

semgrep-core/src/fixing/Autofix_printer.mli
nmote: We need all of these arguments for use in the external printer. The final argument to this function is the AST that we want to print. However, we also need the metavar bindings so that we can tell which nodes in the final AST came unchanged from a metavar, and the target contents so that we can lift the original text for those nodes from it.

Similarly, we need the AST for the original fix pattern, as well as that fix pattern's text, so that we can tell which nodes came from the pattern and use their original text.

I'll add a comment to this effect.

semgrep-core/src/fixing/Autofix_printer.mli
nmote: See my comment above

semgrep-core/src/core/ast/Ugly_print_AST.ml
nmote: I prefer to use `option` because callers are forced to handle it and cannot forget, especially for something like this where there isn't a clear single entry point where we could catch the exception and keep it internal to the module. If you feel strongly I can change it, though.

nmote:
Thanks for the prompt and thorough review! I've responded inline and I'll go through and make the changes as soon as I can.

aryx:
I don't feel strongly. I'm fine with your design choices, I just wanted to point out alternatives (and maybe ask you to explain more those alternatives in comments and how your design improves over those naive alternatives).
Looks good otherwise.

semgrep-core/src/core/ast/Ugly_print_AST.ml
```diff
@@ -0,0 +1,122 @@
...
+ * implementation-sharing and dynamic dispatch. Dynamic dispatch allows for the
+ * hybrid printer used for autofix, where we use the original text from the
+ * target or pattern for nodes that have been lifted unchanged. This is
+ * implemented by inheriting from the plain printer. With a different
+ * architecture, printer authors would need to deal with the logic for this.
+ * This way, printer authors need only focus on printing AST correctly.
+ *
+ * OCaml's powerful object system will also allow us to share implementation
+ * between printers to whatever extent is appropriate. `Pretty_print_AST`, for
+ * example, gates on language within each function. We could replicate that here
```
emjin, line 48: To be able to do this would we create a function for, say, each expression? I'm wondering if we would be able to do some of the things you can with `Pretty_print_AST`, where you can have most of the languages use the same implementation but three specific ones are different

semgrep-core/src/core/ast/Ugly_print_AST.ml
emjin: +1
This will be really nice, I'm liking the setup otherwise

semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -315,7 +319,8 @@ let run_selector_on_ranges env selector_opt ranges =
       in
       let patterns = [ (pattern, false, pid, fst pstr) ] in
       let res =
-        matches_of_patterns ~range_filter env.xconf env.xtarget patterns
+        matches_of_patterns ~range_filter env.rule env.xconf env.xtarget
```
emjin, line 322: Maybe we could refactor to just pass in the whole env? I feel like what keeps happening is that a function only needs one thing from the env, so we pass it directly. Then, the next person follows the previous convention, until the function has a ton of arguments which are taken out of the env

semgrep-core/src/core/ast/Ugly_print_AST.ml
emjin: I was also thinking that we might want some information about why the autofix failed for our own debugging; also as documentation for later engineers if it's subtle. Pad's suggestion of logging works for that too, though, so if options are a lot cleaner I'm fine with that. The only other thing is that it's easy to forget to log.

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
...
+
+(******************************************************************************)
+(* Main module for AST-based autofix. This module will attempt to synthesize a
+ * fix based a rule's fix pattern and the match's metavariable bindings. *)
+(******************************************************************************)
+
+let parse_pattern_opt lang pattern =
+  try Some (Parse_pattern.parse_pattern lang pattern) with
+  | Timeout _ as e -> Exception.catch_and_reraise e
+  | _ -> None
```
emjin: Parse_pattern will sometimes raise exceptions that include useful messages; it would be nice if we could preserve those too (at least for our own debugging benefit)

emjin: If that's hard as a start let's log that this returned None because `parse_pattern_opt` failed

semgrep-core/src/fixing/Autofix.ml
emjin: Or maybe `fixed_pattern_ast`?

semgrep-core/src/fixing/Autofix.ml
emjin: This made sense to me as the core of "AST-based autofix", for the record. I guess maybe what the module needs is a more explanatory file header

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Failure causes include, but are not limited to:
+ * - The fix pattern does not parse.
+ * - A metavariable is bound to an AST node that is not suitable for the context
+ *   in which it is used in the fix.
+ * - Printing of the resulting fix AST fails (probably because there is simply a
+ *   node that is unhandled).
+ * *)
+let render_fix lang metavars ~fix_pattern ~target_contents =
+  (* Fixes are not exactly patterns, but they can contain metavariables that
```
emjin, line 59: I spent longer than I should have understanding what `fix_pattern` was and I wonder if this comment is part of it. I was additionally confused, I think, because I went in assuming `render_fix` would actually fix the whole target file, rather than just output the fix pattern as text

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
...
+   * *)
+  let* fix_pattern_ast = parse_pattern_opt lang fix_pattern in
+
+  (* Look through the pattern and replace metavariables with the nodes to which
+   * they are bound in the match. *)
+  let* fixed_ast =
+    Autofix_metavar_replacement.replace_metavars metavars fix_pattern_ast
+  in
+
+  (* Try to print the fixed AST. *)
```
emjin: Another place where terms used may have thrown me off. `Fixed AST` makes me think of the whole fixed target AST

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,125 @@
...
+module MV = Metavariable
+
+(******************************************************************************)
+(* Module responsible for traversing a fix pattern AST and replacing
+ * metavariables within it with the AST nodes from the target file to which the
+ * metavariables are bound. *)
+(******************************************************************************)
+
+let make_metavar_tbl bindings =
+  let tbl = Hashtbl.create (List.length bindings) in
```
emjin: This made me wonder if there were any cases where you could have a problem with two matches that have the same range but different metavariables. I don't _think_ this can ever happen when there's a well-formed autofix (because which match would you replace?) but I haven't thought through this carefully.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,125 @@
...
+  let mapper =
+    Map_AST.(
+      mk_visitor
+        {
+          default_visitor with
+          (* TODO handle more kinds of nodes *)
+          kargs =
+            (fun (k, _) args ->
+              (* A metavariable can appear as a single argument, but can be
+               * bound to multiple arguments, so we have to handle this case by
```
emjin: > A metavariable can appear as a single argument, but can be bound to multiple arguments

That isn't self-evident to me. If you have a quick example to show how that happens, would be helpful

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -0,0 +1,70 @@
...
+  let* text =
+    Autofix_printer.print_ast ~lang ~metavars ~target_contents ~fix_pattern_ast
+      ~fix_pattern fixed_ast
+  in
+
+  (* Perform sanity checks for the resulting fix. If they fail, return None *)
+  if validate_fix lang text then Some text else None
```
emjin, line 97: Let's also log here that validating failed and what the fix that failed was

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,125 @@
...
+              let args = List.concat_map map_arg args in
+              k args);
+          kexpr =
+            (fun (k, _) orig ->
+              (* The mapper changes the ID of expressions. This stymies the
+               * print-avoidance implemented later in the autofix process,
+               * because the resulting AST nodes cannot be recognized as the
+               * same as the original ones.
+               *
+               * To work around this, we use the original expression, except for
```
emjin, line 70: Another reason to do this: saving propagated values for later use in the message (via `value($X)`). So the workaround would still be necessary!

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,125 @@
...
+               * changing that behavior in the mapper, in which case this
+               * workaround would no longer be necessary.
+               * *)
+              let expr = k orig in
+              { orig with e = expr.e });
+        })
+  in
+  mapper.Map_AST.vany pattern_ast
+
+(* Check for remaining metavars in the AST. If there are any, that indicates a
```
emjin:
```suggestion
(* Check for remaining metavars in the fix AST. If there are any, that indicates a
```
This throws me off even though it's pretty obvious

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,125 @@
...
+ * failure to properly replace them in the previous step, and the autofix
+ * attempt should be aborted.
+ *
+ * This currently works by comparing all identifiers against the list of bound
+ * metavariables. Some languages allow identifiers that look like metavariables,
+ * so this is less likely to result in a false positive than if we checked for
+ * any identifiers that could be metavariables. However, this wouldn't catch the
+ * case of a malformed fix pattern where the user wrote a metavariable in the
```
emjin, line 92: I support that! My reasoning is that, in that case I think it's better to assume that they meant that not as a metavariable. If we want to be less tolerant I think the better place is as a metacheck, which people have the choice of ignoring

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,125 @@
...
+let make_metavar_tbl bindings =
+  let tbl = Hashtbl.create (List.length bindings) in
+  List.iter (fun (mvar, mvalue) -> Hashtbl.replace tbl mvar mvalue) bindings;
+  tbl
+
+let replace metavar_tbl pattern_ast =
+  (* Use a mapper to traverse the AST. For each metavar, look up what it is
+   * bound to. If the kind of node matches, replace the metavar in the pattern
+   * with the AST node to which it is bound. Note that we can't handle this by
+   * simply visiting the identifier nodes, because metavars are often bound to
```
emjin: I thought I understood this comment and then I didn't. The way I initially read the comment was "because metavars can exist in constructs other than Id", which I then realized (from looking at `has_remaining_metavars`) wasn't true. Now I think it's because by the time you get to `Id` you've lost context that you need earlier. This would probably be obvious to me if I thought about it, but thinking is hard

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,125 @@
...
+ * attempt should be aborted.
+ *
+ * This currently works by comparing all identifiers against the list of bound
+ * metavariables. Some languages allow identifiers that look like metavariables,
+ * so this is less likely to result in a false positive than if we checked for
+ * any identifiers that could be metavariables. However, this wouldn't catch the
+ * case of a malformed fix pattern where the user wrote a metavariable in the
+ * fix that doesn't exist in the rule's pattern.
+ * *)
+let has_remaining_metavars metavar_tbl ast =
```
emjin, line 95: Why not just perform this check when mapper hits the `Id` case? Anything that could have been resolved would be resolved by then. And we wouldn't have to go to the expense of creating another visitor and traversing the AST again

emjin: I guess it's not so bad since fix patterns are small, so feel free to ignore this

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,125 @@
...
+ *   the actual node to which the metavar is bound. For example, a metavar that
+ *   is bound to a statement used in a position that expects an expression.
+ * - A failure in this module to handle some specific case, leading to a
+ *   metavar's continued presence in the tree after replacement has been
+ *   attempted. In this case, this function should detect that and return None.
+ * *)
+let replace_metavars metavars pattern_ast =
+  let metavar_tbl = make_metavar_tbl metavars in
+  let res = replace metavar_tbl pattern_ast in
+  if has_remaining_metavars metavar_tbl res then None else Some res
```
emjin, line 135: Debuggability here too; ideally log what metavariable wasn't successfully replaced

semgrep-core/src/fixing/Autofix_printer.ml
```diff
@@ -0,0 +1,128 @@
...
+  class printer = Ugly_print_AST.python_printer
+end)
+
+let get_printer lang external_printer : Ugly_print_AST.printer_t option =
+  match lang with
+  | Lang.Python -> Some (new PythonPrinter.printer external_printer)
+  | _ -> None
+
+let original_source_of_ast source any =
+  let* start, end_ = Visitor_AST.range_of_any_opt any in
```
emjin, line 61: Note: this is fine now but if we ever start doing autofixes on stmt lists we could run into perf trouble. Just making this note because we've had multiple perf problems with `range_of_any_opt` and it's easy to run into. Getting a range of an expression or a statement is safer because those two are now at least cached.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
emjin: Ohh I forgot about `...$X`

semgrep-core/src/fixing/Autofix_printer.ml
```diff
@@ -0,0 +1,128 @@
...
+  let str = String.sub source starti len in
+  Some str
+
+(* Add each metavariable value to the lookup table so that it can be identified
+ * during printing *)
+let add_metavars tbl metavars =
+  List.iter
+    (fun (_, mval) ->
+      let any = MV.mvalue_to_any mval in
+      Hashtbl.replace tbl any `Metavar;
```
emjin: I'm so confused about what this table looks like 😂. Just going to keep reading and come back

emjin: Ok, I kind of get it now. We're marking each of the nodes as metavar or pattern, which the pretty-printer will use to decide what ast to print as. This could use explanation in the file header. Still a bit confused about "pattern"

semgrep-core/src/fixing/Autofix_printer.ml
```diff
@@ -0,0 +1,128 @@
...
+      match mval with
+      | MV.Args args ->
+          List.iter (fun arg -> Hashtbl.replace tbl (Ar arg) `Metavar) args
+      (* TODO iterate through other metavariable values that are lists *)
+      | _ -> ())
+    metavars
+
+(* Add each AST node to the lookup table so that it can be identified during
+ * printing *)
+let add_ast_nodes tbl ast =
```
emjin: This name surprised me---I didn't expect `add_ast_nodes` to be the function that checked for patterns. This seems to actually be `add_fix_pattern_nodes`?

emjin: Oh I am once again confused because I thought this was going through the whole target AST 🤦. I don't know why I'm so stuck on that.

semgrep-core/src/fixing/Autofix_printer.ml
```diff
@@ -0,0 +1,128 @@
...
+   * failing to look up a node is not catastrophic: it will only mean that in
+   * some cases we may fail to print. *)
+  let tbl = Hashtbl.create 8 in
+  add_metavars tbl metavars;
+  add_ast_nodes tbl fix_pattern_ast;
+  fun any ->
+    let* node = Hashtbl.find_opt tbl any in
+    let* str =
+      match node with
+      | `Metavar -> original_source_of_ast (Lazy.force target_contents) any
```
emjin: It's really counterintuitive to me that "metavar" is the one we can use the original source of the ast for but "pattern" is for the fix pattern? I'm still missing something. I'll leave this file now with my scattered comments. tldr; explain more in the file header with examples

semgrep-core/src/fixing/Autofix_printer.mli
emjin: I was also confused about this :D

semgrep-core/src/fixing/Hybrid_print.ml
emjin: Personally I'm fine with this; the printing stages are separated enough that I don't find the complexity here interferes with my general understanding

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -743,6 +743,7 @@ let minirule_of_pattern lang pattern_string pattern =
     message = "";
     severity = R.Error;
     languages = [ lang ];
+    fix = None;
```
emjin, line 746: It bothers me a little bit that we can't distinguish between "this rule has no fix" and "this rule has a failed fix". I can't think of any place where this could cause problems, it just makes me suspicious.

semgrep-core/src/utils/Unit_immutable_buffer.ml
```diff
@@ -0,0 +1,27 @@
+module B = Immutable_buffer
```
emjin, line 1: yooooo unit test

semgrep-core/src/utils/Immutable_buffer.ml
```diff
@@ -0,0 +1,40 @@
+(* Nat Mote
```
emjin, line 1: It would be so nice if we could have all of our commonly used modules in `utils` like this! But that's a problem for another day

semgrep-core/tests/Test.ml
```diff
@@ -52,6 +52,7 @@ let tests () = List.flatten [
   Unit_memory_limit.tests;
   Unit_SPcre.tests;
   Unit_regexp_engine.tests;
+  Unit_immutable_buffer.tests;
```
emjin, line 55: Are there any integration tests to show this working? Or is that waiting on the autofix test suite? (I saw some discussion but didn't pay attention)

emjin (changes requested):
This is awesome! I really like the way it's laid out, with three printers that build on each other, and I learned a lot from reading the PR. My main concern is that if an autofix isn't synthesized it's impossible to debug why. There are three (four?) possible failure points across multiple files that are all silent. Without reading this PR I wouldn't know what to look for. Otherwise just looking for some more comments, mostly in `Autofix_printer.ml`, which I had a hard time reading, and consistent naming of the fix ast.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
emjin: Oh wait this isn't true since this is only used for synthesizing the fix text. I'll leave this comment as evidence of confusion

semgrep-core/src/core/ast/Ugly_print_AST.ml
nmote: We could do whatever we wanted. A method for each expression could be reasonable. We could have a `monolithic_printer` class that does exactly what `Pretty_print_AST` does and matches on the language in each method, just like the existing printer does. I'm not putting forward any particular proposal on this topic, I just wanted to point out that we aren't constrained by the design choices that I've made here.

semgrep-core/src/engine/Match_search_mode.ml
nmote: I'll look into this. I have a vague sense that I considered it and something stopped me, but maybe it was only laziness :stuck_out_tongue:

semgrep-core/src/core/ast/Ugly_print_AST.ml
nmote: Yeah, that's a good point. Using `Result.t` would let us report with granularity where exactly the printing failed. I agree that that's valuable, but I think doing that well would be a large change to this PR. How about for this PR, I add logging so that we can at least tell which large phase failed, and then experiment with `Result.t` later this week as a followup?

semgrep-core/src/fixing/Autofix.ml
nmote: Do you have suggestions on how I can address this confusion? Would mentioning in the main comment for `render_fix` that it prints the text of only the fix itself help?

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
nmote: Yes, this can happen and it's a problem. That's the root cause of #4963.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
nmote: Yeah, it's not obvious without an example. I'll add that in.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
nmote: Basically, you can't replace an `ident` type with an `argument` type. If you want to replace arguments, you have to start with arguments. I think an example might help drive the point home, so I'll add that to this comment.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
nmote: I think I'm missing some context here, but if there are more reasons than this to change the default behavior of the mapper, we should probably go ahead and do it.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
nmote: I think this is more foolproof because a maintainer of the mapper could forget to call the `k` function to continue the traversal, potentially leading to missed nodes.

semgrep-core/src/fixing/Autofix_printer.ml
nmote: I'm hoping that, when compared to the overall amount of code scanned, the number of actual findings with fixes is low enough that we can get away with some inefficiency here. I read the comments on perf and I guessed that this was only an issue because it was getting called in a very hot code path. I don't think this is going to get called a ton.

I agree we should keep an eye on it, though.

semgrep-core/src/fixing/Autofix_printer.ml
nmote: Sounds good, I'll explain this better. Maybe I should change Pattern to Fix, since that is for nodes that came from the `fix` field in the rule?

semgrep-core/src/fixing/Autofix_printer.ml
nmote: I will add some better comments here.

semgrep-core/src/fixing/Autofix_printer.ml
nmote: This is because the metavar bindings contain the AST nodes that those metavars matched in the original target file. So, when printing those nodes, we need to get them from the target file.

In contrast, when printing nodes from the fix pattern, we need to lift text from the fix pattern in the rule.

I'll add some comments and think about how I can improve the naming here to make it clearer.

semgrep-core/src/runner/Run_semgrep.ml
nmote: This is the minirule field that has the rule's `fix` field as written (if it has one). At this point, we haven't attempted to render a fix for a particular finding.

semgrep-core/tests/Test.ml
nmote: In the summary, I wrote:

> I'm comfortable with the existing tests for this change, but as we
improve autofix I think we'll want some better testing infrastructure.
The existing end-to-end tests are thorough, but they are a bit clunky.
It would be nice to have semgrep-core tests that verify that AST-based
autofix is working correctly.

Does that answer your question?

semgrep-core/src/engine/Match_search_mode.ml
nmote: I looked into this and now I remember what the issue was:

We need the fix in `matches_of_patterns` in order to construct the `mini_rule`: https://github.com/returntocorp/semgrep/blob/7b1ee35f7d4ed5147d81a4f11889ac05379bca16/semgrep-core/src/engine/Match_search_mode.ml#L215

That function is called by `matches_of_xpatterns`: https://github.com/returntocorp/semgrep/blob/7b1ee35f7d4ed5147d81a4f11889ac05379bca16/semgrep-core/src/engine/Match_search_mode.ml#L437

That function is called by `matches_of_formula`: https://github.com/returntocorp/semgrep/blob/7b1ee35f7d4ed5147d81a4f11889ac05379bca16/semgrep-core/src/engine/Match_search_mode.ml#L771

`matches_of_formula` uses the return value of `matches_of_xpatterns` in order to construct the `env`. So, we don't have an `env` to pass down at that point. It seems a bit odd to me that it is structured that way, but I don't want to attempt to restructure it as part of this change.

semgrep-core/src/fixing/Autofix.ml
nmote: I agree that we should do this at some point. As an update to this PR, I'm adding logging which will indicate which of the top level phases of autofix has failed, but I think it will be a little bit more involved to add logging to the printer that will allow us to tell exactly which node failed. Because of that I would prefer to experiment a bit and do that as a followup.

semgrep-core/src/fixing/Autofix.ml
nmote: I'm adding some more explanation here. Given the short length of the `render_fix` function, I'd rather explain inline than add a long file header that people are less likely to read.

nmote:
Thanks, both of you, for your thorough reviews! I believe that I have addressed your concerns by improving readability and explanations. Let me know what you think!

aryx:
probably need to merge develop for fixing the pytest regressions.

nmote:
Just had to handle the new log line for autofix failure in the e2e test

semgrep-core/src/fixing/Autofix.ml
emjin: Mentioning it in the main comment would definitely help

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
emjin: Oh, I see! Is that case addressed by this PR?

nmote: No, it's not

emjin: Might be worth adding a comment about it; this is a confusing scenario

emjin (approved):
Looks great! Thanks for the comments!

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -0,0 +1,140 @@
...
+
+let logger = Logging.get_logger [ __MODULE__ ]
+
+(******************************************************************************)
+(* Module responsible for traversing a fix pattern AST and replacing
+ * metavariables within it with the AST nodes from the target file to which the
+ * metavariables are bound. *)
+(******************************************************************************)
+
+let make_metavar_tbl bindings =
```
aryx: You could use Common.hash_of_list

semgrep-core/src/fixing/Autofix_printer.ml
```diff
@@ -0,0 +1,148 @@
...
+
+let original_source_of_ast source any =
+  let* start, end_ = Visitor_AST.range_of_any_opt any in
+  let starti = start.Parse_info.charpos in
+  let _, _, endi = Parse_info.get_token_end_info end_ in
+  let len = endi - starti in
+  let str = String.sub source starti len in
+  Some str
+
+(* Add each metavariable value to the lookup table so that it can be identified
```
aryx, line 68: An alternative to having those hash and passing around the metavars each time to know whether an AST node comes from the target file would be to mark specially those nodes in the AST! We use a record for
expr and stmts now, so we could a field origin:, that could be set to Target of filename | FixedPattern

semgrep-core/src/fixing/Autofix_printer.ml
```diff
@@ -0,0 +1,148 @@
...
+ * be identified during printing *)
+let add_fix_pattern_ast_nodes (tbl : ast_node_table) ast =
+  let visitor =
+    Visitor_AST.(
+      mk_visitor
+        {
+          default_visitor with
+          kargument =
+            (fun (k, _) arg ->
+              Hashtbl.replace tbl (Ar arg) FixPattern;
```
aryx, line 115: BIG QUESTION:
Shouldn't we check that the argument does not contain any metavariable?
what if the fix: is `foo(1+$X, $...ARGS)`?
Won't you just print `1+$X` as is in the resulting text just because
`1+$X` is an argument coming from the fix_pattern_ast.

aryx (changes requested):
I've just have a BIG QUESTION comment (see in the list of my new comments) where if you add a comment in the code answering my question, I'll approve.
The rest looks good!

semgrep-core/src/fixing/Autofix_printer.ml
aryx: That would require to turn every elements in Metavariable.mvalue deriving from AST_generic as a record (ident, name, expr, stmt. type, pattern, argument, parameter, xml_body). We just have expr and stmt for now.

semgrep-core/src/fixing/Hybrid_print.ml
```diff
@@ -0,0 +1,60 @@
...
+        match primary any with
+        | Some x -> Some x
+        | None -> fallback#print_any any
+
+      method! print_argument arg =
+        match primary (G.Ar arg) with
+        | Some x -> Some x
+        | None -> fallback#print_argument arg
+
+      (* TODO Fill in more cases as needed. *)
```
aryx, line 58: One problem is that each time we will have a bug report for autofix, we will need to update many of those TODOs
(in Hybrid_print, Ugly_print, but also in some visitors that replace the metavar, etc.) and it starts to be many candidates without much debugging help to know where to start.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
nmote: I agree that it's a confusing scenario but it seems unrelated to the code in this PR. It was already a problem before this PR, and it continues to be a problem now. There's already a lot to explain and understand about the central purpose of this code: taking one `Pattern_match.t` and rendering a fix for it to text. I'd rather not also add notes about the circumstances under which this code is currently used by callers.

semgrep-core/src/fixing/Autofix_printer.ml
nmote: I considered this, but I'm reluctant to add a field to basically every AST node to support an operation that occurs on only a very small minority of AST nodes. It would significantly increase memory usage throughout Semgrep, I expect.

We'd also need to then make sure that we set that origin to `None` or `Synthesized` or something when the mapper changes a node, because otherwise we'd incorrectly pull text from the pattern or target when in reality the node has changed and we need to print it ourselves.

semgrep-core/src/fixing/Autofix_printer.ml
nmote: I believe that what would happen right now is that we would fail to replace the `$X` because ordinary expressions are not yet handled in the metavariable replacement step. Then, when checking the resulting AST for lingering metavariables, we would find `$X` and abort at that point.

Once we implement metavariable replacement for this case (which won't be hard, I just wanted to get one scenario working end-to-end first), then we'd replace `$X` with, say, `5`. We'd still add the AST node for `Arg (1 + $X)` here (pardon the pseudocode), but when printing, we'd look up `Arg (1 + 5)`. Since those two nodes are not equal, the hashtbl lookup will fail and we will not pull the original text from the fix pattern.

semgrep-core/src/fixing/Hybrid_print.ml
nmote: I think it will be fairly straightforward to fill in all the TODOs for the language-agnostic parts of the printing machinery for all nodes that are part of `AST_generic.any`, which will give us good coverage. I intend to do this soon. The part that will drag on is actually implementing the printers for each language, but at least that's just one place.

I am also planning to experiment with ways to improve debug logging. I made some changes as part of this PR review so that at least it will now log which main phase of autofix has failed, but both you and @emjin pointed out that more granular information would be better. I'd like to do that as a followup.

nmote:
I've responded to your comments and made some changes. Back to you for another round!

aryx (approved):

aryx:
Merged!

PR #6144, opened 2022-09-21
See what we can do with private type abbreviations

mjambon:
This is a spin-off of https://github.com/returntocorp/semgrep/pull/6130/files/b8cb72ee1712e0ec2c91c96d31734ac3ce0dfd4f#r975040361


semgrep-core/src/parsing/ast/AST_dockerfile.mli
```diff
@@ -0,0 +1,194 @@
...
+
+(* Used for quoted and unquoted strings for now.
+   Must be created with the 'create_str' function to ensure that
+   consecutive literal fragments are collapsed.
+   Converting str -> unchecked_str is done with the type coercion operator :>
+*)
+type unchecked_str = loc * string_fragment list
+type str = private unchecked_str
+
+val create_str : unchecked_str -> str
```
mjambon, line 60: The relevant changes in the module interface are the 3 lines of code above. The requirement of doing `(x :> unchecked_str)` before being able to pattern-match on it is impractical because it can't be done when pattern matching against `Arr_string x`.

PR #6153, opened 2022-09-21
Documenting and cleaning up our github action workflows part 1

aryx:
test plan:
wait for CI to kick in and check if everything is green

PR checklist:

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,21 +1,56 @@
 This directory contains Github actions "workflows" to automate certain
 tasks (e.g., running our testsuite on each PR, automate the release).
 See https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions
-for more information or our Notion page on Github actions.
+for more information on Github actions or our Notion page on Github actions.
+
 Most of those workflows have the 'workflow_dispatch:' directive so you can
 also trigger them manually here: https://github.com/returntocorp/semgrep/actions
 
 Here is a short description of the workflows in this directory:
 
-- lint.yml: running mostly pre-commit checks for our pull requests (PRs)
-- tests.yml:
-- semgrep.yml:
-- start-release.yml:
-- release.yml:
-- e2e-semgrep-ci.yml:
-- find-old-brew-prs.yml:
-- homebrew-core-head.yml:
-- open-bump-pr.yml:
-- revert-semgrep-docker-image.yml:
-- update-semgrep-rules.yml:
-- validate-release.yml:
+- lint.yml: running mostly pre-commit checks for our pull requests (PRs).
+  It runs the same checks that we run with pre-commit locally (configured
+  in semgrep/.pre-commit-config.yaml), but enforced here in CI in case
+  the developer forgot to setup pre-commit.
+
+- tests.yml: building semgrep and running our semgrep testsuite for our PRs.
+  It also runs benchmarks, test our MacOS, Linux, and docker build, and more.
+
+- semgrep.yml: dogfood Semgrep by using our Semgrep Github action
+  and submitting findings to Semgrep App.
+
+- start-release.yml: workflow to manually trigger a new Semgrep release.
+  This internally creates a new branch, call 'make release' on it, and
+  put it on a vXxx branch which then triggers the release workflow below.
+  It also coordinates separate workflows, wait for them to finish to
+  start other workflows. Here are the dependent workflows:
+
+  - release.yml: this workflow is triggered after start-release.yml created
+    a candidate release PR and pushed a vXxx branch. It then performs lots of task such
+    as creating the Linux and MacOS binary artifacts,
+    publishing to PyPy, homebrew, updating our semgrep develop docker image, etc.
+
+  - validate-release.yml: this workflow is activated by start-release.yml
+    once release.yml finished. It performs checks to make sure
+    the artifacts produced by release.yml are valid
+
+  - open-bump-pr.yml: this workflow is activated once we have a new semgrep release
+    (after validate-release.yml) and in turn bumps to the new semgrep version
+    many dependent repositories (semgrep-rpc, semgrep-action, semgrep-app)
+
+- find-old-brew-prs.yml: cron to check whether the homebrew PR created by
+  release.yml has been accepted or is hanging around unmerged
+
+- homebrew-core-head.yml: cron to check that the Homebrew Core Formula
+  created by release.yml works (TODO: why can't this be part of validate-release.yml?
+  because it usually take some time for homebrew developers to accept the PRs
+  so we can't test directly? hence find-old-brew-prs.yml and homebrew-core-head.yml
+  workflows?)
+
+- update-semgrep-rules.yml: cron to update semgrep/semgrep-core/tests/semgrep-rules
+  submodule to its latest version
+
+- revert-semgrep-docker-image.yml: interactive workflow
+  to manually revert the semgrep 'latest' docker image.
+
+- e2e-semgrep-ci.yml: ???
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -14,10 +14,10 @@ Here is a short description of the workflows in this directory:
   the developer forgot to setup pre-commit.
 
 - tests.yml: building semgrep and running our semgrep testsuite for our PRs.
-  It also runs benchmarks, test our MacOS, Linux, and docker build, and more.
+  It also runs benchmarks, test our MacOS, Linux, and Docker build, and more.
 
 - semgrep.yml: dogfood Semgrep by using our Semgrep Github action
-  and submitting findings to Semgrep App.
+  and submitting findings to Semgrep App for bugs in the semgrep repo itself.
 
 - start-release.yml: workflow to manually trigger a new Semgrep release.
   This internally creates a new branch, call 'make release' on it, and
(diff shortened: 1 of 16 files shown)


nmote (approved):

aryx:
To whoever review, the goal is to better document our workflow. I've put some TODOs in the yaml file when I don't know something so if you have the knowledge it would be great if in your review you comment next to it so then I can fill those TODOs with what you've explained.

.github/workflows/e2e-semgrep-ci.yml
```diff
@@ -1,4 +1,7 @@
+#TODO: The goal of this workflow is ???
```
spencerdrak: This workflow is set up to do some very basic E2E checks - makes sure that semgrep will still run both on a pull request and via `semgrep ci`. The context here is that we had some cases where releases went out that were broken in simple ways, and this gives us some confident that nightly semgrep still works.

.github/workflows/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+  - open-bump-pr.yml: this workflow is activated once we have a new semgrep release
+    (after validate-release.yml) and in turn bumps to the new semgrep version
+    many dependent repositories (semgrep-rpc, semgrep-action, semgrep-app)
+
+- find-old-brew-prs.yml: cron to check whether the homebrew PR created by
+  release.yml has been accepted or is hanging around unmerged
+
+- homebrew-core-head.yml: cron to check that the Homebrew Core Formula
+  created by release.yml works (TODO: why can't this be part of validate-release.yml?
```
spencerdrak: These jobs take >60 minutes to run, and require that the latest version of the homebrew formula is up to date, which only occurs after release. The PRs to homebrew also are merged when the homebrew teams gets to them - we shouldn't block on actions we can't control.

.github/workflows/README.md
```diff
@@ -0,0 +1,56 @@
...
+  because it usually take some time for homebrew developers to accept the PRs
+  so we can't test directly? hence find-old-brew-prs.yml and homebrew-core-head.yml
+  workflows?)
```
spencerdrak: Correct assumption.

.github/workflows/lint.yml
```diff
@@ -1,4 +1,7 @@
+# The goal of this workflow is mostly to run pre-commit on every pull requests.
+
 name: lint
+
```
spencerdrak, line 4: I haven't touched much of this workflow, will try and answer questions but may leave gaps based on context.

.github/workflows/lint.yml
```diff
@@ -7,21 +10,31 @@ on:
       - develop
 
 jobs:
+  # Running pre-commit in CI. See semgrep/.pre-commit-config.yaml for
+  # our pre-commit configuration.
   pre-commit:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        #TODO: why do we need this ref:? doesn't checkout@v3 grabs the right branch by default?
```
spencerdrak: It should. I can't speak for this specific case, but in the workflows I've been a part of implementing, we've set this explicitly to remove assumptions and make it clear for follow on maintainers.

.github/workflows/lint.yml
```diff
@@ -72,12 +87,16 @@ jobs:
...
+        #TODO: should we extend that beyond just cli/src?
+        # is it still needed now that we have a ../pull_request_template.md
+        # that also reminds to add a changelog entry
         with:
           GET_FILE_DIFF: true
           PATTERNS: cli/src/semgrep/**
       - id: changelog_diff
         name: Get changelog diff
         uses: technote-space/get-diff-action@v6.1.0
+        #TODO: should we get rid of this whole job now that we use changelog.d/?
```
spencerdrak: Probably, yes.

.github/workflows/tests.yml
```diff
@@ -17,6 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     container: returntocorp/ocaml:alpine-2022-06-09
     steps:
+      #TODO: why do we need this? What does not work if we don't have this?
```
spencerdrak, line 26: This was needed the last time I looked at this part of the workflow (~3 months ago). Maybe @underyx knows.

.github/workflows/e2e-semgrep-ci.yml
spencerdrak: It also allows us to confirm fail-open behavior is still functioning as expected (e.g., we aren't passing on blocking findings)

.github/workflows/homebrew-core-head.yml
```diff
@@ -1,9 +1,12 @@
-# Verify Homebrew Core Formula Works
+# Cron to verify that the Homebrew Core Formula Works.
+# This formula is created by ???
```
spencerdrak: This formula is created by our release process with the PR to `homebrew/homebrew-core`. What this workflow does is uses the latest version of the formula at that repo, but `develop` branch source code from our PR. This serves two purposes:
- Verifies that our changes don't break brew
- Gives us time before release to fix these issues and adjust our homebrew formula if needed.

.github/workflows/start-release.yml
```diff
@@ -45,6 +52,7 @@ jobs:
           submodules: "recursive"
           ref: "${{ github.event.repository.default_branch }}"
           token: "${{ steps.token.outputs.token }}"
+      #TODO: Why do we need this? checkout@v3 does not get the tags by default?
```
spencerdrak: See comment below, however, no. checkout@v3 only gets tags if you do a fully checkout, which is too heavyweight. We don't want all branches and everything that ever existed on the repo, so we just do a lightweight checkout and then get the tags ourselves.

.github/workflows/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+- update-semgrep-rules.yml: cron to update semgrep/semgrep-core/tests/semgrep-rules
+  submodule to its latest version
+
+- revert-semgrep-docker-image.yml: interactive workflow
+  to manually revert the semgrep 'latest' docker image.
+
+- e2e-semgrep-ci.yml: ???
```
spencerdrak: See the comment on the workflow file for a description.

.github/workflows/tests.yml
aryx: I think @mjambon actually wrote those pre-checkout fixes

.github/workflows/lint.yml
spencerdrak: @brendongo might have context.

.github/workflows/lint.yml
```diff
@@ -7,21 +10,31 @@ on:
...
+        #TODO: why do we need this ref:? doesn't checkout@v3 grabs the right branch by default?
         with:
           ref: ${{ github.event.pull_request.head.sha }}
+      # We grab those submodules below because they are the one needed by 'mypy' which
+      # runs as part of pre-commit to check our Python code.
+      # alt: we could also use 'submodules: recursive' instead, but that would be slower
       - name: Fetch semgrep-cli submodules
         run: git submodule update --init --recursive --recommend-shallow cli/src/semgrep/lang cli/src/semgrep/semgrep_interfaces
+      #TODO: why this is needed? pre-commit/action@v3 requires this setup-python to run?
+      # what is this cache-dependency-path?
```
brendongo: This caches the pip installed dependencies that pre-commit uses. Otherwise it will need to reinstall everytime

.github/workflows/lint.yml
```diff
@@ -7,21 +10,31 @@ on:
...
       - uses: actions/setup-python@v4
         with:
           python-version: "3.10"
           cache: pip
           cache-dependency-path: .github/workflows/lint.yml
       - uses: pre-commit/action@v3.0.0
 
+  #TODO: what is the difference with the 'pre-commit' job above? looks only
+  # the extra_args below but what is it for? What will not work without that?
```
brendongo: The extra args runs https://github.com/returntocorp/semgrep/blob/develop/.pre-commit-config.yaml#L150
which dont run on normal pre-commit since they would slow down pre-commit on local development. The intention is a test that runs semgrep-pre-commit. So we can also split this out to a different config?

.github/workflows/lint.yml
brendongo: And yes pre-commit is a python script :D https://github.com/pre-commit/pre-commit

aryx: so each time you use the pre-commit/action, you also need to setup python before? The github action don't have a way to pull automatic dependencies?

.github/workflows/lint.yml
aryx: why not add just the   
```    - uses: pre-commit/action@v3.0.0
        with:
          extra_args: --hook-stage manual
```
in the pre-commit job instead of duplicating all the parts before and make it a separate job?
Because it can run in parallel then?

aryx: I don't see any difference though oin the pre-commit config file between:
```
  # Dogfood! running semgrep in pre-commit!
  - repo: https://github.com/returntocorp/semgrep
    rev: "v0.100.0"
    hooks:
      - id: semgrep
        name: Semgrep Python
        types: [python]
        exclude: "^cli/tests/.+$|^scripts/.+$|^cli/setup.py$"
        args: ["--config", "https://semgrep.dev/p/python", "--error"]
      - id: semgrep
        name: Semgrep Bandit
        types: [python]
        exclude: "^cli/tests/.+$|^scripts/.+$|^cli/setup.py$"
        args: ["--config", "https://semgrep.dev/p/bandit", "--error"]
```
and
```
  # Run develop semgrep. Only used in CI since it's slower for local developmemt.
  # To run locally use `pre-commit run --hook-stage manual semgrep-docker-develop`
  - repo: https://github.com/returntocorp/semgrep
    rev: "v0.100.0"
    hooks:
      - id: semgrep-docker-develop
        name: Semgrep Develop Python
        types: [python]
        exclude: "^cli/tests/.+$|^scripts/.+$|^cli/setup.py$"
        args: ["--config", "p/python", "--error"]
        stages: [manual]
      - id: semgrep-docker-develop
        name: Semgrep Develop Bandit
        types: [python]
        exclude: "^cli/tests/.+$|^scripts/.+$|^cli/setup.py$"
        args: ["--config", "p/bandit", "--error"]
        stages: [manual]
```
in both cases we use rev: "v0.100.0" so we don't really test the develop docker image here.

PR #6157, opened 2022-09-21
feat: add file renames to diffing logic

chmccreery:
The current diffing logic uses the list of files that have findings on the `HEAD` commit, and scans those files in the `BASE` commit. It then goes finding by finding and removes any findings that are found in the `BASE` commit from the `HEAD` commit, so that Semgrep only reports _new_ findings. Semgrep uses a `ci_unique_key` (rule_id, file_path, syntactic_context, and index) to determine that two findings on `BASE`/`HEAD` are actually the same.

The problem with this, is that when users re-name files, the ci_unique_key changes because the file_path changes. To address this, in this PR semgrep starts to use `git diff`'s ability to detect file renames, and substitutes the old file name in for the new file name when scanning the `BASE` commit.

This will prevent us from showing old findings as "new" in a diff scan when a developer is refactoring the code and renaming files. 

Full scan behavior (without baseline_ref) is unchanged by this PR, so semgrep-app will still report these findings as "new" when a full scan is done and findings are uploaded to semgrep-app. This means the findings in renamed files will show up for security engineers, but will never block or spam developers. A follow-up PR in semgrep-app will address that, but it is out of scope for this PR.

PR checklist:

Closes APP-2344


changelog.d/gh-6157.changed
```diff
@@ -0,0 +1,2 @@
+Changed renamed file behavior on diff scans (scans in which a baseline ref is provided).
```
chmccreery:
```suggestion
Changed behavior for renamed files on diff scans (scans in which a baseline ref is provided).
```

DrewDennison (approved):

cli/src/semgrep/git.py
```diff
@@ -37,6 +37,7 @@ class GitStatus(NamedTuple):
     modified: List[Path]
     removed: List[Path]
     unmerged: List[Path]
+    renamed: Dict[str, Path]  # keys are new names, values are old paths
```
brendongo, line 40: nit: Having the key also be Path might be the more expected type?

cli/src/semgrep/rule_match.py
```diff
@@ -205,6 +205,19 @@ def get_ci_unique_key(self) -> Tuple:
...
+        """
+        A unique key that accounts for filepath renames.
+
+        Results in fewer unique findings than cli_unique_key.
+        """
+        try:
+            path = str(self.path.relative_to(Path.cwd()))
+        except (ValueError, FileNotFoundError):
+            path = str(self.path)
+        renamed_path = str(rename_dict[path]) if path in rename_dict else path
```
brendongo, line 218: Wont this only work on the PR that made the rename? Any PR after this would then not be able to do this lookup

brendongo (dismissed):

cli/src/semgrep/rule_match.py
chmccreery: Discussed offline and agreed this should be ok

PR #6160, opened 2022-09-22
fix: don't parse ts patterns to `Pr`

brandonspark:
**What:**
This PR implements a change which parses TS/JS in a way that pattern parsing does not lead to the `Pr` constructor being used.

**Why:**
Someone pulled up in the community slack with this issue:
https://r2c-community.slack.com/archives/C018NJRRCJ0/p1663850040935109

Basically, the pattern 
```
import $K = require('aws-cdk-lib/aws-rds')
...
```
was being parsed to a `Pr`. However, the engine in `Match_patterns` only deals with a few variants of `any`, and namely `Pr` is not one of them. So it just `failwith`s.

**How:**
In the parsing of the pattern, I add a `match` which converts any `Pr` to an `Ss`. This only happens in `parse_pattern`, so there's no risk of it affecting target parsing.

**Comments:**
I'm not in love with the fact that there's this distinction between `Pr` and `Ss`, because I don't really see why there ought to be. We have to insert logic converting between the other (in several places), when really I don't know of a time when we have to dispatch on that information in a meaningful way. Unless there is a strong reason to keep it, I would be inclined towards refactoring to eliminate this variant.

Another solution that was considered was to add a case to `Match_patterns` which handles `Pr`, but we decided not to go for that approach.

PR checklist:

### DIFF
--- a/changelog.d/pa-1910.fixed
+++ b/changelog.d/pa-1910.fixed
@@ -1,0 +1,1 @@
+Added a fix for a bug involving parsing of TS imports.
--- a/changelog.d/pa-1910.fixed
+++ b/changelog.d/pa-1910.fixed
@@ -1,1 +1,1 @@
-Added a fix for a bug involving parsing of TS imports.
+Added a fix for a bug involving parsing of TS imports, where they were not allowed to appear as patterns to a rule.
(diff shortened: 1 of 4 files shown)


aryx (approved):
Yes maybe we should just remove Program or handle it generaically in parse_pattern
to accept not only Ss but also Pr.
Whatever you think he's best.

changelog.d/pa-1910.fixed
```diff
@@ -0,0 +1 @@
+Added a fix for a bug involving parsing of TS imports.
```
IagoAbal: It's very difficult to understand what was the bug about just reading this. The changelog entries should be a bit more detailed. This has been a common complain from @adamkvitek when trying to produce release notes.

PR #6164, opened 2022-09-22
feat(terraform): IL translation of module calls and variable usages

brandonspark:
## What:
This PR has to do with changing all module declarations in Terraform to "fake function calls", which will sync up with work done in previous PRs.

## Why:
See the current scope doc [here](https://docs.google.com/document/d/1yoHq0RNvVhhlH51PyMPzzSvUFIYlmoMavCV2LzRy2d4/edit?usp=sharing). Basically, we are modeling module instantiation as the calling of a fake module function. As such, module definitions have to be translated during IL translation into function calls, with parameters as specified in the declaration. This will hook into the existing taint engine logic.

## How:
We changed module declarations here to be real function calls. The parameters will all have fake `sid`s, though, which will muck with the lvalue matching process. This can be fixed at a later date by changing the lvalue comparison function.

## Test plan:
Since this is having to do with DeepSemgrep, I necessarily cannot test these changes here. It will be testable upon consolidation with other PRs.

PR checklist:

### DIFF
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -566,9 +566,7 @@ and expr_aux env ?(void = false) e_gen =
                                         ( _,
                                           {
                                             id_resolved =
-                                              Some
-                                                ( G.ResolvedName (dotted_id, _),
-                                                  _ );
+                                              { contents = Some _; _ };
                                             _;
                                           } )) as name;
                                   _;
@@ -582,22 +580,24 @@ and expr_aux env ?(void = false) e_gen =
                 {
                   s =
                     DefStmt
-                      ( { name = EN (Id ((arg_name, _), _)); _ },
+                      ( { name = EN (Id (arg_id, _)); _ },
                         VarDef { vinit = Some arg_exp; _ } );
                   _;
                 } ->
-                Right3 (arg_name, arg_exp)
+                Right3 (G.ArgKwd (arg_id, arg_exp))
             | _ ->
                 (* bTODO: Ignore all other cases. They correspond neither to the module's name, or its arguments. *)
                 Middle3 ())
           fields
       in
+      let tok = G.fake "call" in
       match mod_sources with
-      | [ mod_source ] -> call_generic env ~void tok
+      | [ mod_source ] ->
+          call_generic env ~void tok (mod_source |> G.e)
+            (G.fake_bracket mod_args)
       | _ ->
           (* Don't know how to interpret multiple sources. Just let this go to the wildcard case.
          *)
-          let tok = G.fake "call" in
           call_generic env ~void tok e args)
   | G.New (tok, ty, args) ->
       (* TODO: lift up New in IL like we did in AST_generic *)
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -223,10 +223,70 @@ let is_hcl lang =
   | Lang.Hcl -> true
   | _ -> false
 
+let rec hcl_translate_module_call env void (e, args, fields) =
+  (* bTERRAFORM:
+      Terraform modules are translated in the form of `Call`, but we would like to
+      interpret them as actual function calls to fake function definitions.
+      This step in the translation to IL just manufactures that fake function
+      definition.
+      This function is named the same as the "module name" for a given module,
+      which is of the form `dir/module_name` for a module at directory `dir`.
+      The parameters are all of the `variable` blocks.
+
+      We have to translate into an actual IL Call construct with the correct
+      name so that we can let the taint engine do the work.
+  *)
+  let mod_sources, _, mod_args =
+    Common.partition_either3
+      (function
+        | G.F
+            {
+              s =
+                DefStmt
+                  ( {
+                      name =
+                        EN
+                          (Id
+                             ( ("source", _),
+                               {
+                                 id_resolved =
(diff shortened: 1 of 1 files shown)


semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -536,6 +541,72 @@ and expr_aux env ?(void = false) e_gen =
        * interpolated strings, but we do not have an use for it yet during
        * semantic analysis, so in the IL we just unwrap the expression. *)
       expr env e
+  | G.Call
+      ( ({ e = N (Id (("module", _), _)); _ } as e),
+        ((_, [ _; G.Arg { e = Record (_, fields, _); _ } ], _) as args) )
+    when is_hcl env.lang -> (
+      (* bTERRAFORM:
```
IagoAbal: typo: **b**TERRAFORM ?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -536,6 +541,72 @@ and expr_aux env ?(void = false) e_gen =
...
+         OK, so here's the idea.
```
IagoAbal: Nitpicking, but "OK, so here's the idea" doesn't add any useful info.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -536,6 +541,72 @@ and expr_aux env ?(void = false) e_gen =
...
+         This step in the translation to IL just manufactures that fake function
+         definition.
+         This function is named the same as the "module name" for a given module,
+         which is of the form `dir/module_name` for a module at directory `dir`.
+         The parameters are all of the `variable` blocks.
+
+         We have to translate into an actual IL Call construct with the correct
+         name so that we can let the taint engine do the work.
+      *)
+      let mod_sources, _, mod_args =
```
IagoAbal: This is a lot of code so it's better if we put it in a separate function. (Maybe not a lot of code, but a lot of lines of code.)

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -536,6 +541,72 @@ and expr_aux env ?(void = false) e_gen =
...
+                                         contents = Some (G.ResolvedName _, _);
+                                         _;
+                                       };
+                                     _;
+                                   } ) as name);
+                          _;
+                        },
+                        VarDef _ );
+                  _;
+                } ->
```
IagoAbal: What is a module source?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -536,6 +541,72 @@ and expr_aux env ?(void = false) e_gen =
...
+                {
+                  s =
+                    DefStmt
+                      ( { name = EN (Id (arg_id, _)); _ },
+                        VarDef { vinit = Some arg_exp; _ } );
+                  _;
+                } ->
+                Right3 (G.ArgKwd (arg_id, arg_exp))
+            | _ ->
+                (* bTODO: Ignore all other cases. They correspond neither to the module's name, or its arguments. *)
```
IagoAbal: What are these other cases (for example)?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -552,6 +623,13 @@ and expr_aux env ?(void = false) e_gen =
       let tok = G.fake "call" in
       call_generic env ~void tok e args
   | G.L lit -> mk_e (Literal lit) eorig
+  | G.DotAccess ({ e = N (Id (("var", _), _)); _ }, _, FN (Id ((s, t), id_info)))
```
IagoAbal, line 625: Need to add `when is_hcl lang`.

IagoAbal (approved):
I need to see the DeepSemgrep PR to get the full picture but as long as everything is restricted to HCL then it's good.

semgrep-core/src/analyzing/AST_to_IL.ml
brandonspark: According to the Terraform docs:

image

semgrep-core/src/analyzing/AST_to_IL.ml
brandonspark: The module which is being called.

semgrep-core/src/analyzing/AST_to_IL.ml
brandonspark: good catch!

semgrep-core/src/analyzing/AST_to_IL.ml
IagoAbal: I will be more explicit next time. But I expect that you answer that question in a code comment, e.g. adding a link to the Terraform docs where those "other cases" are specified or something.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -218,10 +218,75 @@ let composite_of_container : G.container_operator -> IL.composite_kind =
...
   | Dict -> CDict
 
+let is_hcl lang =
+  match lang with
+  | Lang.Hcl -> true
+  | _ -> false
+
+let rec hcl_translate_module_call env void (e, args, fields) =
+  (* bTERRAFORM:
+      Terraform modules are translated in the form of `Call`, but we would like to
```
aryx, line 228: Could we add a comment explaining why we didn't do this transformation in Parse_hcl_tree_sitter.ml instead?

PR #6165, opened 2022-09-22
feat!: allow lvalue comparison to unknown `sid`s

brandonspark:
## What:
This PR adds the ability to say that any lvalue with matching names are the same, if one of them has an unresolved `sid`.

## Why:
In implementing Terraform changes, I had to construct a fake function definition for each module. In particular, I had to add parameters for the function, after the naming stage, which necessarily have `sid`s as `-1`. When comparing variables for comparison in taint analysis, it will automatically reject them if they have the same name, as long as they don't have the same `sid`s. I would like this functionality so that, even if we don't resolve the name of a variable, we go by identifier matching to say that they are the same.

## How:
Changed the comparison function for names which is used for lvalue comparison.

PR checklist:

### DIFF
--- a/semgrep-core/src/core/il/IL_lvalue_helpers.ml
+++ b/semgrep-core/src/core/il/IL_lvalue_helpers.ml
@@ -22,6 +22,11 @@ open IL
 let compare_name x y =
   let ident_cmp = String.compare (fst x.ident) (fst y.ident) in
   if ident_cmp <> 0 then ident_cmp
+    (* If their names match, and either of the names have an unresolved sid,
+       then let's just say that they are the same.
+       In particular, this will help with Terraform, where we purposefully introduce
+       unresolved names as the parameters to fake functions.
+    *)
   else if x.sid = -1 || y.sid = -1 then 0
   else Int.compare x.sid y.sid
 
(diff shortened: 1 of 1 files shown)


semgrep-core/src/core/il/IL_lvalue_helpers.ml
```diff
@@ -21,7 +21,14 @@ open IL
 
 let compare_name x y =
   let ident_cmp = String.compare (fst x.ident) (fst y.ident) in
-  if ident_cmp <> 0 then ident_cmp else Int.compare x.sid y.sid
+  if ident_cmp <> 0 then ident_cmp
+    (* If their names match, and either of the names have an unresolved sid,
+       then let's just say that they are the same.
+       In particular, this will help with Terraform, where we purposefully introduce
+       unresolved names as the parameters to fake functions.
```
IagoAbal, line 28: For DeepSemgrep, I think we should always try having every name resolved. I am a bit concerned about how this may change the results of taint-mode. The effect may even be positive, but it would be great if we could run some tests on a bunch of repos before applying this. Could you also provide some pointer to know more about those unresolved names that we need to introduce for Terraform?

IagoAbal: As far as I can tell, DeepSemgrep only works with two types of resolved names, corresponding to `AST_generic.LocalVar` and `AST_generic.ResolvedName`. If the name you are interested in is already resolved, and it's local, you can just declare the parameter as the same local name (it seems that DeepSemgrep doesn't distinguish  between `Parameter` and `LocalVar` as Semgrep does). If that name is global, you can transform it into a local by updating the `(AST_generic.id_info_id, IdInfo.t) Hashtbl.t`.

brandonspark:
Made irrelevant by #269.

PR #6173, opened 2022-09-24
sym-prop: Fix propagatation of New expressions

### LINKED ISSUE #6161: Symbolic propagation does not propagate `new`
Reporting on behalf of @backcover7.

We expect this https://semgrep.dev/s/vrBA to work but it doesn't.

IagoAbal:
Fixes: 995163b526d ("fix(js): Parse `New` as a separate construct from `Call` (#5510)")

PR #5510 made `new` expressions different from function calls, thus breaking symbolic propagation of `new`.

Closes #6161
Closes PA-1911

test plan:
make test # added one test

PR checklist:

### DIFF
--- a/semgrep-core/src/analyzing/Dataflow_svalue.ml
+++ b/semgrep-core/src/analyzing/Dataflow_svalue.ml
@@ -373,7 +373,13 @@ and eval_concat env args =
 (* Symbolic evaluation *)
 (*****************************************************************************)
 
-(* Defines the subset of Generic expressions that we can propagate. *)
+(* Defines the subset of Generic expressions that we can propagate.
+ *
+ * In principle we could just propagate any expression, we restricted it mainly
+ * to play it safe. Eventually we could consider lifting up these restrictions
+ * and see what happens. The main problem with the current approach is that
+ * every now and then somebody requests X to be supported by symbolic propagation.
+ *)
 let rec is_symbolic_expr expr =
   match expr.G.e with
   | G.L _ -> true
@@ -383,7 +389,10 @@ let rec is_symbolic_expr expr =
   | G.ArrayAccess (e1, (_, e2, _)) -> is_symbolic_expr e1 && is_symbolic_expr e2
   | G.Call (e, (_, args, _)) ->
       is_symbolic_expr e && List.for_all is_symbolic_arg args
-  | _ -> false
+  | G.New (_, _, args) ->
+      let args = G.unbracket args in
+      List.for_all is_symbolic_arg args
+  | _else -> false
 
 and is_symbolic_arg arg =
   match arg with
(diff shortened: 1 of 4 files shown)


semgrep-core/src/analyzing/Dataflow_svalue.ml
```diff
@@ -383,7 +389,10 @@ let rec is_symbolic_expr expr =
   | G.ArrayAccess (e1, (_, e2, _)) -> is_symbolic_expr e1 && is_symbolic_expr e2
   | G.Call (e, (_, args, _)) ->
       is_symbolic_expr e && List.for_all is_symbolic_arg args
-  | _ -> false
+  | G.New (_, _, args) ->
+      let args = G.unbracket args in
+      List.for_all is_symbolic_arg args
+  | _else -> false
```
emjin, line 395: Nit: Is `_else` a new standard we're trying to adopt? I find `_` perfectly clear; if it's not I can start using `_else` but we should standardize

semgrep-core/src/analyzing/Dataflow_svalue.ml
```diff
@@ -515,9 +524,18 @@ let transfer :
               let ccall = sym_prop instr.iorig in
               update_env_with inp' var ccall
         | CallSpecial
-            (Some { base = Var var; rev_offset = [] }, (Concat, _), args) ->
-            (* var = concat(args) *)
-            let cexp = eval_concat inp' args in
+            (Some { base = Var var; rev_offset = [] }, (special, _), args) ->
+            let cexp =
+              (* We try to evalaute the special function. *)
```
emjin:
```suggestion
              (* We try to evaluate the special function. *)
```

emjin: I think something like `Depending on what the special function is, we evaluate it` might be even more clear? It's ok though. Also, just to be clear, in the case of Concat I assume we will propagate the evaluated value?

emjin (approved):
Very cool!

semgrep-core/src/analyzing/Dataflow_svalue.ml
IagoAbal: I find `| _ -> ...` perfectly clear as I would find ` | $%! -> ...` perfectly clear if `$%!` was the OCaml syntax for the wildcard pattern. I adopted this practice from Haskell land time ago (the GHC project). If people feel confused about it we can vote to forbid anything than `| _ -> ...`. For me `_else` is easier to read at least.

semgrep-core/src/analyzing/Dataflow_svalue.ml
IagoAbal: Thanks for asking these questions, I clarified the comment. I wrote:

> We try to evaluate the special function, if we know how. If we don't know how to evaluate this function, or if the evaluation fails, then we do sym-prop.

So if evaluating `Concat` gives a constant then yes we propagate that, but if we fail to get a constant from `Concat` then we do sym-prop.

PR #6174, opened 2022-09-24
Simplify (and comment more) Dockerfile thx to the new returntocorp/ocaml:alpine

aryx:
Now that the returntocorp/ocaml:alpine use the 'user=root' default,
like most Docker images for languages (e.g., Golang, Rust),
the Dockerfile can be simpler.
See also https://github.com/returntocorp/ocaml-layer/pull/14
and
https://returntocorp.slack.com/archives/C01NXGX2EHZ/p1663937723277069
for a recent discussion on this topic.

test plan:
$ make build-docker
$ docker images # semgrep:latest still at 226MB (before compression)

PR checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,52 @@
 ###############################################################################
 # Overview
 ###############################################################################
-# First, build a *static* 'semgrep-core' binary on Alpine
-# (requires using musl rather than glibc).
+# First, we build a fully *static* 'semgrep-core' binary on Alpine. This
+# binary does not even depend on Glibc because Alpine uses Musl instead
+# which can be statically linked.
 #
-# Then 'semgrep-core' alone is copied to a container which takes care
-# of the 'semgrep-cli' Python wrapping.
+# Then 'semgrep-core' alone is copied to another Alpine-based container
+# which takes care of the 'semgrep-cli' Python wrapping.
+#
+# We use Alpine because it allows to generate the smallest Docker images.
 
 ###############################################################################
 # Step1: build semgrep-core
 ###############################################################################
-# The docker base image below in the FROM currently uses OCaml 4.14.0
+
+# The Docker image below (after the 'FROM') is prepackaged with OCaml, 'opam',
+# and lots of packages that are used by semgrep-core and installed in
+# 'make setup' further below.
 # See https://github.com/returntocorp/ocaml-layer/blob/master/configs/alpine.sh
+# Thanks to this container, 'make setup' finishes very quickly because it's
+# mostly a noop. Alternative container candidates are:
+#
+#  - ocaml/opam:alpine, the official OCaml/opam container, 
+#    but building a Docker image would take longer because
+#    of all the necessary Semgrep dependencies installed in 'make setup'.
+#    Note also that ocaml/opam:alpine default user is 'opam', not 'root', which
+#    is not without problems when used inside GHA or even inside this Dockerfile.
+#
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,111 +1,112 @@
...
 # coupling: if you modify the OCaml version there, you probably also need
 # to modify:
 # - scripts/{osx-release,osx-m1-release,setup-m1-builder}.sh
 # - doc/SEMGREP_CORE_CONTRIBUTING.md
 # - https://github.com/Homebrew/homebrew-core/blob/master/Formula/semgrep.rb
+#
 # Note that many .github/workflows/ use returntocorp/ocaml:alpine, which should
 # be the latest, but may differ from this one.
-FROM returntocorp/ocaml:alpine-2022-06-09@sha256:99b453a838c9d94414991c0fd7be4711aa1bcc120f576e0f0653c7b921ea9718 as semgrep-core
+FROM returntocorp/ocaml:alpine-2022-09-24 as semgrep-core-container
```
spencerdrak, line 57: We may want to keep the pinning around in case these tags ever change - could be cause for confusing errors down the line.

spencerdrak (approved):

Dockerfile
```diff
@@ -114,24 +115,24 @@ RUN apk add --no-cache --virtual=.build-deps build-base && \
...
      SEMGREP_VERSION_CACHE_PATH=/tmp/.cache/semgrep_version \
      SEMGREP_USER_AGENT_APPEND="Docker"
 
-WORKDIR /src
-
 # /semgrep is not needed anymore;
 # semgrep is now available /usr/local/bin thx to the 'pip install' command above
-# (this weirdly does not reduce the size of the docker image though)
+# TODO: this weirdly does not reduce the size of the docker image though
 RUN rm -rf /semgrep
```
mjambon, line 152: It's a good observation. This is how docker images work. The state of the filesystem after each docker instruction is called a layer and remains available in the final image, similarly to diffs in a git history. To save space, we'd have to start another docker build stage like we already [do](https://github.com/returntocorp/semgrep/blob/3b82b71d493d5eaa367e3eec6c8531a08d4bd7f2/Dockerfile#L83) between the ocaml build and the python build. See the [Docker documentation on multistage builds](https://docs.docker.com/build/building/multi-stage/).

mjambon: I don't think we should create another build stage here. I would add a note that the rm is only a cleanup, that it doesn't save space, and link to the doc on multistage builds.

Dockerfile
```diff
@@ -114,24 +115,24 @@ RUN apk add --no-cache --virtual=.build-deps build-base && \
...
 # Get semgrep-core from step1
-COPY --from=semgrep-core /semgrep/semgrep-core/_build/default/src/cli/Main.exe /usr/local/bin/semgrep-core
+COPY --from=semgrep-core-container /src/semgrep/semgrep-core/_build/default/src/cli/Main.exe /usr/local/bin/semgrep-core
 
+# ???
 ENV SEMGREP_IN_DOCKER=1 \
      SEMGREP_VERSION_CACHE_PATH=/tmp/.cache/semgrep_version \
      SEMGREP_USER_AGENT_APPEND="Docker"
 
-WORKDIR /src
```
aryx, line 130: hmm I've removed this because I dunno what it's for (it was not commented), but I think we might rely on this
for commands we tell people to run.
Still does not look like I broke CI so there was no test for it.

PR #6185, opened 2022-09-27
Cleanup GHA workflows and Dockerfile and Makefile

aryx:
This factorizes things that were repeated in many files before

test plan:
wait for CI and check everything is still green

PR checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -144,9 +144,6 @@ jobs:
     env:
       HOME: /root
     steps:
-      #TODO: why do we need this? should be in install-alpine-semgrep-core
-      - run: sudo pip install pipenv==2022.6.7
-      - run: sudo apk add python3-dev
       - name: Make checkout speedy
         run: git config --global fetch.parallel 50
       - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -167,16 +164,13 @@ jobs:
     steps:
       - name: Make checkout speedy
         run: git config --global fetch.parallel 50
-      - name: Checkout
```
spencerdrak, line 170: @aryx my only comment here is to ask why we're removing the names of the steps? I find them to be useful markers especially in longer workflows.

spencerdrak (approved):
Approving from the DevOps side - however someone with more expertise in the actual internals might also want to review.

.github/workflows/release.yml
aryx: It's just repeating the uses:/run: that is just after.
It's like writing code like: 
```
// Calling foo()
foo()
```
For steps where the run: contains a long series of commands, then I keep the name: because it's actually a useful comment that gives an overiew of what this step is doing, but otherwise it's just noise in the file.

PR #6186, opened 2022-09-27
fix(deep-semgrep): Propagate taint through method calls of objects

emjin:
For deep-semgrep

PR checklist:


semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -479,14 +479,15 @@ let check_function_signature env fun_exp args_taints =
...
+                       function. *)
                     id_info =
                       {
-                        G.id_resolved =
-                          {
-                            contents =
-                              Some ((G.ImportedEntity _ | G.ResolvedName _), _);
-                          };
-                        _;
+                        G.id_resolved = { contents = Some _; }; _;
```
IagoAbal: Now I wished somebody had asked me _why_ I was introducing this restriction at the time I did it. :D

IagoAbal (approved):

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: I actually think we should probably not even case-split on `fun_exp` at all. If the function expression is not being resolved (and trying to resolve it's probably cheap?) then we just return `None` and that makes Semgrep take the fall-back assumption (assume it's tainted). If you want you can just do that, and drop the `TODO` note.

emjin: I found a reason this is bad so I'm motivated to make that fix now!

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -471,30 +471,17 @@ let check_function_signature env fun_exp args_taints =
...
-              _;
-            };
+        e = Fetch { base = Var { ident; _ }; rev_offset = []; _ };
         eorig = SameAs eorig;
         _;
+      } )
+  | ( Some hook,
+      (* Case `$X. ... .$F()` *)
+      {
+        e = Fetch { base = _; rev_offset = Dot { ident; _ } :: _; _ };
```
IagoAbal: Why do you need to distinguish those two cases?

emjin: It's for `ident`, so that we can have the token of the function. I can clarify

emjin: Actually, I'll just perform that split lower since we only need it in one case

PR #6187, opened 2022-09-27
feat(terraform): parse block names consistently, avoid mistranslation of records in IL

brandonspark:
## What:
This PR adds in two things. One is the parsing of block names in a consistent way, due to these names being able to be presented in either a string literal, or as a standalone identifier. Two is ignoring unknown fields within a record for HCL programs during IL translation.

## Why:
The first will allow us to avoid being unable to produce findings due to naming issues on blocks, because blocks are expected to have names, instead of string literals.

The second will allow us to avoid being unable to produce findings due to failing at IL translation time. This is because there may be certain constructs within a `resource` block, such as a `lifecycle` block, which does not translate to a `DefStmt`, but an `ExprStmt`. The IL translation engine doesn't know what to do with this though, so it currently just makes the entire module call into a `Fixme`. To avoid that, we will opt to just ignore such fields when translating HCL programs specifically.

## How:
This was achieved via just altering the HCL parser to parse string literals as block names to real names, and to change the mapping of records during IL translation to a `map_filter` which only triggers on parsing an HCL AST.

## Test plan:
`make test` still succeeds. I can't really write a more in-depth test demonstrating the effect on DeepSemgrep, but I have tested this manually, and it matters.

PR checklist:

### DIFF
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -867,7 +867,7 @@ and record env ((_tok, origfields, _) as record_def) =
   let e_gen = G.Record record_def |> G.e in
   let fields =
     origfields
-    |> Common.map (function
+    |> Common.map_filter (function
          | G.F
              {
                s =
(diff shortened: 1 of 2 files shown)


brandonspark:
Irrelevant due to #6226

PR #6189, opened 2022-09-28
Start CLI migration from Python to OCaml

mjambon:
This is experimental work in progress over about two weeks.

PR checklist:


cli/Makefile
```diff
@@ -18,10 +18,30 @@ test:
...
+	$(PYTEST) -n auto -v \
+	  -m 'not no_semgrep_cli and not todo' tests/e2e
+
 # Run the extra tests on public repos and semgrep-rules
 .PHONY: qa
 qa:
-	$(PYTEST) -n auto tests/qa
+	$(PYTEST) -n auto -v tests/qa
+
+# Run the end-to-end tests using mlgrep instead of the Python CLI.
```
aryx: not a big fan of mlgrep. I think osemgrep was better.
or semgrep-ocaml, to differentience from semgrep-python and semgrep-core.

cli/Makefile
```diff
@@ -18,10 +18,30 @@ test:
...
+.PHONY: mlgrep-e2e
+mlgrep-e2e:
+	PYTEST_USE_MLGREP=xxx $(MAKE) e2e
```
aryx: =1 or =true would be more readable than =xxx I think.

cli/tests/semgrep_runner.py
```diff
@@ -0,0 +1,113 @@
+#
+# Run either semgrep in the same process using Click or inovke the
```
aryx: typo, invoke

cli/tests/semgrep_runner.py
```diff
@@ -0,0 +1,113 @@
...
+# It's something like ["semgrep"] or ["python3"; -m; "semgrep"] or
+# ["/path/to/mlgrep"].
+SEMGREP_BASE_COMMAND: List[str] = (
+    [MLGREP_PATH] if USE_MLGREP else [sys.executable, "-m", "semgrep"]
+)
+
+SEMGREP_BASE_COMMAND_STR: str = " ".join(SEMGREP_BASE_COMMAND)
+
+
+class Result:
```
aryx, line 59: No idea what this class is doing.

cli/tests/semgrep_runner.py
```diff
@@ -0,0 +1,113 @@
...
+    proc = Popen(argv, stdout=PIPE, stderr=PIPE, env=full_env)
+    stdout, stderr = proc.communicate()
+    return Result(proc.returncode, stdout.decode("utf-8"), stderr.decode("utf-8"))
+
+
+class SemgrepRunner:
+    """Run either semgrep with CliRunner or with mlgrep.
+
+    It's meant as a drop-in replacement for CliRunner(cli, args).
+    If a property is missing on the runner object, please add it here.
```
aryx, line 123: what do you mean by property of runner object? Do you have an example?

cli/tests/semgrep_runner.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+
+    def __init__(self, env=None, mix_stderr=True):
+        self._use_mlgrep = USE_MLGREP
```
aryx: Shouldn't we define those fields via a @dataclass with types?

semgrep-core/src/mlgrep/cli-lib/Error.ml
```diff
@@ -0,0 +1,37 @@
+(*
```
aryx: Would be nice to have a file somewhere that lists which files have been converted (or half-converted) and which
files remain.

semgrep-core/src/mlgrep/mlgrep-util/Constants.ml
```diff
@@ -0,0 +1,152 @@
+(*
```
aryx: not a big fan of -util suffix for directories.

aryx: and not a big fan of mlgrep- prefix also. maybe just semgrep-core/src/semgrep-ocaml/core/
that contains the common stuff.

semgrep-core/src/mlgrep/mlgrep-util/Exit_code.ml
```diff
@@ -0,0 +1,34 @@
+(*
```
aryx: What's the difference with cli-lib/Error.ml?

semgrep-core/src/runner/Run_semgrep.ml
```diff
@@ -742,6 +758,41 @@ let semgrep_with_raw_results_and_exn_handler config =
       in
       (Some e, res, [])
 
+(* This is ugly, with potentially some filtering operations being done twice.
+   It should get simplified when we get rid of the Python wrapper.
+   For now, we avoid code duplication.
+*)
+let semgrep_with_prepared_rules_and_targets config (x : lang_job) =
```
aryx, line 765: would be nice to put the return type of this.

semgrep-core/src/runner/Runner_config.ml
```diff
@@ -9,6 +9,16 @@ open Common
  * found by extract rules.
  *)
 type output_format = Text | Json of bool (* dots *) [@@deriving show]
+type rule_source = Rule_file of filename | Rules of Rule.t list
```
aryx, line 18: maybe comment on what is usually used when using semgrep-core directly (rule_file), and
one when passed by semgrep-ocaml?

semgrep-core/src/runner/Runner_config.ml
```diff
@@ -24,7 +34,8 @@ type t = {
   (* Main flags *)
   pattern_string : string;
   pattern_file : filename;
-  rules_file : filename;
+  rule_source : rule_source option;
+  lang_job : lang_job option;
```
aryx, line 52: not a big fan of the _job suffix.

semgrep-core/src/targeting/Find_target.ml
```diff
@@ -1,13 +1,52 @@
...
+type baseline_handler = TODO
+type file_ignore = TODO
+type path = string
+
+(*
+   Some rules will use 'include' (required_path_patterns) and 'exclude'
+   (excluded_path_patterns) to select targets that don't have an extension
+   such as 'Dockerfile'. We expect most rules written for a language
+   to use the same combination of include/exclude. This allows caching
+   across the many rules that target the same language.
```
aryx, line 29: nice!

semgrep-core/src/targeting/List_files.mli
```diff
@@ -0,0 +1,38 @@
+(*
+   List files recursively in a safe, efficient, and portable manner.
+
+   Replaces the functions in pfff/commons that use external unix commands.
```
aryx: maybe add "(e.g., find)"

semgrep-core/src/targeting/List_files.mli
```diff
@@ -0,0 +1,38 @@
...
+*)
+
+type path = string
```
aryx, line 8: maybe add a comment about why we don't use FPath.t (yet).

aryx (approved):
Looks like a great start!

aryx:
+2300 LOC. Not bad for a week.

semgrep-core/src/mlgrep/cli-lib/Semgrep_CLI_lib.ml
```diff
@@ -0,0 +1,79 @@
...
+  scan          Run semgrep rules on files
+  shouldafound  Report a false negative in this project.
+|}
+
+let dispatch_subcommand argv =
+  match Array.to_list argv with
+  | [] -> assert false
+  | [ _; ("-h" | "--help") ] ->
+      print_string main_help_msg;
+      exit Error.ok_exit_code
```
aryx: why not just Error.ok_exit_code here, without the exit before?
Anyway this will be done in Main.ml

semgrep-core/src/mlgrep/cli-lib/Semgrep_CLI_lib.ml
```diff
@@ -0,0 +1,79 @@
...
+(* This is used to determine if we should fall back to assuming 'scan'. *)
+let known_subcommands =
+  [ "ci"; "login"; "logout"; "lsp"; "publish"; "scan"; "shouldafound" ]
+
+(* Exit with a code that a proper semgrep implementation would never return.
+   Uncaught OCaml exception result in exit code 2.
+   This is to ensure that the tests that expect error status 2 fail. *)
+let missing_subcommand () =
+  eprintf "This semgrep subcommand is not implemented\n%!";
+  exit Error.not_implemented_in_mlgrep
```
aryx: same, we do not need the exit call here.

cli/Makefile
mjambon: I wanted to avoid a long name. I'm fine with `osemgrep`.

cli/Makefile
mjambon: The point is that the value doesn't matter. Maybe we should parse it and require either `true` or `false`.

cli/Makefile
mjambon: Renamed.

cli/Makefile
mjambon: I added some parsing (overkill but a good exercise)

cli/tests/semgrep_runner.py
mjambon: I explained it.

cli/tests/semgrep_runner.py
mjambon: The runner object is the one created by `CliRunner(cli, args)`, obviously. You need to understand what `CliRunner` does. I explained it briefly in the `Result` class above. It's a hack that pretends to start a new Python program but doesn't, saving the startup time. Not my creation. See https://click.palletsprojects.com/en/8.1.x/testing/

semgrep-core/src/runner/Runner_config.ml
mjambon: It's a unit of work tied to a specific language. I was hoping it would be temporary until we no longer need to split the input by language, but I don't know.

Regarding the name: I don't know what else to call it. `job` is short and not too far off.

semgrep-core/src/runner/Runner_config.ml
mjambon: I added comments.

semgrep-core/src/osemgrep/TODO.md
```diff
@@ -0,0 +1,77 @@
...
+
+The goal of this file is to track progress and share ideas about the
+migration of the Python CLI wrapper to OCaml, known temporarily as
+osemgrep.
+
+## October 10, 2022
+
+by Martin:
+
+- `osemgrep scan` is the only subcommand that was (partially)
```
aryx, line 11: A big TODO is to plug the result returned by semgrep-core and output it in JSON or in text. That's the output/ directory in semgrep-python to port.

semgrep-core/src/targeting/List_files.mli
mjambon: I don't think explaining everything we're not using makes much sense. I introduced 3 different modules that use `type path` independently. I'm all for using `FPath` but it needs to be revised so it's independent of the state of the file system (among other things).

semgrep-core/src/mlgrep/cli-lib/Semgrep_CLI_lib.ml
mjambon: good catch

semgrep-core/src/mlgrep/cli-lib/Semgrep_CLI_lib.ml
mjambon: thx

semgrep-core/src/osemgrep/TODO.md
mjambon: yes, thanks

semgrep-core/Makefile
```diff
@@ -70,6 +70,9 @@ dev:
 	rm -f ../cli/src/semgrep/bin/semgrep-core
 	ln -s ../../../../semgrep-core/bin/semgrep-core \
 	  ../cli/src/semgrep/bin/semgrep-core
+	rm -f ../cli/src/semgrep/bin/semgrep-o
```
aryx: I don't think we need this anymore

aryx (approved):

semgrep-core/Makefile
mjambon: I actually don't know. I made this PR to have the same symlink creation for both semgrep-core and osemgrep: https://github.com/returntocorp/semgrep/pull/6285

PR #5190, opened 2022-05-11
add(test): test functionality that checks autofix behavior

0xDC0DE:
PR checklist:

- [ ] Documentation is up-to-date
- [ ] Changelog is up-to-date
- [ ] Change has no security implications (otherwise, ping security team)

PR #6193, opened 2022-09-28
test: Test autofixes in semgrep-core

nmote:
This extends the existing semgrep-core test runner that we use to test matches so that it can also test autofix. The format is pretty simple: add a `.fix` file with the fix pattern and a `.fixed` file with the expected contents of the target after fixes are applied.

There are existing end-to-end tests which test autofix on the CLI side. Unfortunately, modifying them involves updating a large snapshot file, and maintaining any large quantity of them would be onerous. Additionally, it's not easy to adapt tests from our large existing test suite in semgrep-core to also exercise autofix.

Semgrep's `--test` flag can also test autofix (#5190), but it has the same problems as the existing autofix e2e tests for these purposes.

PR checklist:

### DIFF
--- a/semgrep-core/src/engine/Unit_engine.ml
+++ b/semgrep-core/src/engine/Unit_engine.ml
@@ -218,15 +218,93 @@ let maturity_tests () =
 (* Language-specific tests *)
 (*****************************************************************************)
 
-let sgrep_file_of_target file =
+let related_file_of_target ~ext ~file =
   let d, b, _e = Common2.dbe_of_filename file in
-  let candidate1 = Common2.filename_of_dbe (d, b, "sgrep") in
-  if Sys.file_exists candidate1 then candidate1
+  let candidate1 = Common2.filename_of_dbe (d, b, ext) in
+  if Sys.file_exists candidate1 then Some candidate1
   else
     let d = Filename.concat tests_path "POLYGLOT" in
-    let candidate2 = Common2.filename_of_dbe (d, b, "sgrep") in
-    if Sys.file_exists candidate2 then candidate2
-    else failwith (spf "could not find sgrep file for %s" file)
+    let candidate2 = Common2.filename_of_dbe (d, b, ext) in
+    if Sys.file_exists candidate2 then Some candidate2 else None
+
+let rec fail_on_overlapping_fixes = function
+  | f1 :: f2 :: tl ->
+      let (_, end1), f1_text = f1 in
+      let (start2, _), f2_text = f2 in
+      if end1 > start2 then
+        failwith
+          (spf "found overlapping fixes:\n\n  %s\n\n  %s" f1_text f2_text);
+      fail_on_overlapping_fixes (f2 :: tl)
+  | [ _ ]
+  | [] ->
+      ()
+
+(* Allows the  semgrep-core test runner that we use to test matches to also test
+ * autofix. The format is pretty simple: add a `.fix` file with the fix pattern
+ * and a `.fixed` file with the expected contents of the target after fixes are
+ * applied.
+ *
+ * There are also end-to-end tests which test autofix on the CLI side.
+ * Unfortunately, modifying them involves updating a large snapshot file, and
+ * maintaining any large quantity of them would be onerous. Additionally, it's
+ * not easy to adapt tests from our large existing test suite in semgrep-core to
+ * also exercise autofix.
+ *
+ * Semgrep's `--test` flag can also test autofix
+ * (https://github.com/returntocorp/semgrep/pull/5190), but it has the same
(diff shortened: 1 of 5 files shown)


semgrep-core/src/engine/Unit_engine.ml
```diff
@@ -218,15 +218,79 @@ let maturity_tests () =
...
+      let (start2, _), f2_text = f2 in
+      if end1 > start2 then
+        failwith
+          (spf "found overlapping fixes:\n\n  %s\n\n  %s" f1_text f2_text);
+      fail_on_overlapping_fixes (f2 :: tl)
+  | [ _ ]
+  | [] ->
+      ()
+
+let compare_fixes lang ~file ~fix matches =
```
aryx, line 256: Can you put the comment you have in your PR message about why we do our own autofix regression infra
and how it compares to the other fixed regression infra?

semgrep-core/src/engine/Unit_engine.ml
```diff
@@ -218,15 +218,93 @@ let maturity_tests () =
...
+      let expected_fixed_text =
+        let expected_fixed_file =
+          match related_file_of_target ~ext:"fixed" ~file with
+          | Some file -> file
+          | None -> failwith (spf "could not find fixed file for %s" file)
+        in
+        Common.read_file expected_fixed_file
+      in
+      let file_text = Common.read_file file in
+      let fixes =
```
aryx, line 269: This whole part of applying a set of fixes to a file from a list of matches, checking for overlapping,
and apply in order could be put in Autofix.ml.
If Martin's project is successful, we would do those things OCaml side at some point, and
better in Autofix.ml than in testing code.

semgrep-core/src/engine/Unit_engine.ml
```diff
@@ -218,15 +218,93 @@ let maturity_tests () =
...
+        List.sort
+          (fun ((start1, _), _) ((start2, _), _) -> start1 - start2)
+          fixes
+      in
+      fail_on_overlapping_fixes fixes;
+      (* Switch to bottom to top order so that we don't need to track offsets as
+       * we apply multiple patches *)
+      let fixes = List.rev fixes in
+      let fixed_text =
+        (* Apply the fixes. These string operations are inefficient but should
```
aryx, line 295: yep, exactly :) hence moving to Autofix.ml now might be a good idea.
I'd rather keep the test-specific code in Unit_engine.ml (e.g., reading some .fix, .fixed)
and things that are quite general (applying a set of fixes from matches to a file in the right order with extra checks)
in the regular code (here Autofix.ml or another file)

semgrep-core/tests/python/fix_ellipsis_metavar.py
```diff
@@ -0,0 +1,15 @@
...
+foo(bar=True), foo(bar=True)
+
+foo(bar=False)
+
+foo(baz=True)
+
+# MATCH:
+foo(  bar   =    True  )
+
+# These should work, but end up with overlapping fixes. See
```
aryx, line 11: add a TODO in the comment

aryx: There's also foo(bar=True, x) to test.

A bit sad it's not working ... the whole purpose of $...X is to allow those fixes :)

aryx (approved):

semgrep-core/src/engine/Unit_engine.ml
nmote: I was planning to productionize this and move it over when I needed it, but I can just put it there now instead with a note that it may not be suitable for use in production.

semgrep-core/tests/python/fix_ellipsis_metavar.py
nmote: It does work in production -- the CLI chooses one of the fixes arbitrarily and applies it. It sometimes looks weird since it rearranges the keyword arguments in some cases, but it is correct.

I added the strict check for overlapping fixes here in the test runner because ideally when there are multiple sets of metavariable bindings for a single match, we would pick the one that seems the most like how the code was originally written and apply it. So, I want to test for such cases. But if we decide that's not important we can always relax this.

emjin (approved):
It's definitely not nice that we have two different implementations to apply autofixes, both of which get tested, though only one of which gets used. But, better to have this than no tests in semgrep-core. Hopefully we can move over that code quickly.

semgrep-core/src/engine/Unit_engine.ml
emjin: I'm supportive of this

aryx:
I'm reorganizing semgrep-core/tests/ so I'm merging this, but the PR comment still stand.

PR #6194, opened 2022-09-28
Revert "feat(semgrep-core cli): added implicit rules key (#5822)"

IagoAbal:
This reverts PR #5822 (commit 17e2a57a537ce79b875b9634ba614ae6ba1ae5fe).

We made this change after App removed the `rules:` key from the Editor. But now App put it back, so it's better to not have semgrep-core accept a different rule format than CLI and the Editor.

test plan:
make test

PR checklist:


emjin (approved):
Are these tests duplicated somewhere else now? Maybe we can just delete them if so

IagoAbal:
> Are these tests duplicated somewhere else now? Maybe we can just delete them if so

The tests are not duplicates, they were not meant to test the syntax. @brandonspark just added the syntax for internal convenience, so that when we got a Playground link we could just copy-paste it directly to run with Semgrep-core.

PR #6196, opened 2022-09-28
fix: Use better hash function in AST-based autofix

nmote:
In some work-in-progress to support more kinds of nodes, I ran into an issue where a difference in `e_range` led to a failure to look up a node that should have been printed by lifting the original text. This fixed it.

Test plan: Automated tests, including those added in #6193

PR checklist:


aryx (approved):
A test showing the problem would be nice (that was not working before,
and works thx to the use of AST_general.equal_any
I understand you have another PR for .fix/.fixed, so once this is landed you can
back a test for this one maybe?

nmote:
Yeah, agree that it would be nice. Unfortunately I ran into some unrelated issues with my work in progress (properly inserting parentheses when needed) and became concerned that adding that printing functionality now could lead to regressions over text-based autofix.

I do have a test case that requires this hash change on my branch, and will add it eventually.

PR #6234, opened 2022-10-04
Add support for Jsonnet in the Semgrep Docker image

aryx:
test plan:
```
$ make build-docker
$ docker run --rm -v "${PWD}:/src" semgrep:latest semgrep --config
semgrep.jsonnet
...
Support for Jsonnet rules is experimental and currently meant for internal use only. The syntax may change or be removed at any point.
Semgrep rule registry URL is https://semgrep.dev/registry.

     ocaml | 41 rules × 362 files
    python |  5 rules × 224 files
...
```

PR checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -107,7 +107,9 @@ COPY cli ./
 # Here is why we need the apk packages below:
 #  - build-base: ??
 #  - make, g++: to compile the jsonnet C++ library which is installed
-#    by 'pip install jsonnet'
+#    by 'pip install jsonnet'.
+#    TODO: at some point we should not need the 'pip install jsonnet' because
+#    jsonnet would be mentioned in the setup.py for semgrep as a dependency.
 # TODO? why the mkdir -p /tmp/.cache?
 # hadolint ignore=DL3013
 RUN apk add --no-cache --virtual=.build-deps build-base make g++ &&\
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -102,12 +102,16 @@ COPY cli ./
...
+#  - build-base: ??
+#  - make, g++: to compile the jsonnet C++ library which is installed
+#    by 'pip install jsonnet'
 # TODO? why the mkdir -p /tmp/.cache?
-# TODO: for jsonnet, you need for the build-deps before 'pip install jsonnet'
-#  'apk add g++ make'
 # hadolint ignore=DL3013
-RUN apk add --no-cache --virtual=.build-deps build-base &&\
+RUN apk add --no-cache --virtual=.build-deps build-base make g++ &&\
+     pip install jsonnet &&\
```
brendongo, line 116: Should we add it as a dependency in setup.py or are you trying to keep this internal use only for testing for now?

brendongo: https://github.com/returntocorp/semgrep/blob/develop/cli/setup.py#L108

aryx: yep internal, next step would be setup.py

aryx: I can add a comment about it

PR #6242, opened 2022-10-05
fix(autofix): Abort on failure to replace metavar in string literal

nmote:
At some point in the near future, AST-based autofix should correctly replace metavariables in string literals with their contents. However, until it does, and to protect against future regressions, we should look for metavariables in string literals when doing this validation pass.

Test plan:

Rule:

```
rules:
- id: thing
  pattern: |
    foo("$X")
  message: uh oh
  languages: [python]
  severity: WARNING
  fix: bar("$X")
```

Target:

```
foo("asdf")
```

Run `semgrep-core -l py -json -rules str-metavar.yaml str-metavar.py -json -debug`

Before, observe that it provides a `rendered_fix` of `foo("$X")`. After, observe that it does not provide a `rendered_fix` and that it logs the error.

PR checklist:

### DIFF
--- a/semgrep-core/src/core/ast/Visitor_AST.ml
+++ b/semgrep-core/src/core/ast/Visitor_AST.ml
@@ -53,6 +53,7 @@ type visitor_in = {
   kid_info : (id_info -> unit) * visitor_out -> id_info -> unit;
   ksvalue : (svalue -> unit) * visitor_out -> svalue -> unit;
   kargument : (argument -> unit) * visitor_out -> argument -> unit;
+  klit : (literal -> unit) * visitor_out -> literal -> unit;
 }
 
 and visitor_out = any -> unit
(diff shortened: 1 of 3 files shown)


semgrep-core/src/core/ast/Visitor_AST.ml
```diff
@@ -96,6 +97,7 @@ let default_visitor =
         ());
     ksvalue = (fun (k, _) x -> k x);
     kargument = (fun (k, _) x -> k x);
+    klit = (fun (k, _) x -> k x);
```
emjin, line 100: Technically this does change the default. If the tests pass then it looks like nothing has been relying on the previous behavior, so I think that's fine, just wanted to make sure it was intentional.

emjin (approved):

semgrep-core/src/core/ast/Visitor_AST.ml
nmote: Could you elaborate? I didn't intend to change the visitor behavior.

PR #6251, opened 2022-10-06
refactor(autofix): Split expr printing into different methods

nmote:
Instead of having the pattern match on expr_kind in each language printer, we can have it in the base printer and then dispatch to the methods for each variant. This is what the autogenerated visitors that ppx deriving visitors does, and I think it makes the code more concise and readable overall.

Test plan: Automated tests

PR checklist:


aryx (approved):
Not super convinced, because I like to know what is not handled
(by replacing the | _ ->  with a big disjunction of all the remaining constructors), and
here you would not be able to see what is not handled.

aryx:
But we can try this technique ...

nmote:
That's fair -- let's leave it as-is.

PR #6256, opened 2022-10-06
dogfood: add a rule to detect |_ -> $X, and initial exclude list

aryx:
This is an experiment. Maybe this rule is too annoying. Let's try
to use it and let's see how many (* nosemgrep *) we end up writing.

test plan:
make check

PR checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -258,19 +258,25 @@ release:
 ###############################################################################
 .PHONY: check
 
-# add --verbose for debugging
+#coupling: see also .circleci/config.yml and its 'semgrep' job
 SEMGREP_ARGS=--config semgrep.jsonnet --error --exclude tests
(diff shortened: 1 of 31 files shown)


nmote (approved):
I feel ambivalent about this. In particular:

> If you think adding a new constructor should have no impact on this code, then replace the pattern with `| __else__ ->` instead.

In practice, won't this just mean that people replace the habit of writing `| _ ->` with `| __else__ ->`?

But with that said, I don't really feel strongly one way or the other so I'm fine with this change.

aryx:
yep, there's a danger. But let's experiment, let's see how it goes.

IagoAbal:
Let's have faith in people @nmote! I got poked by this rule and I actually fixed it without `__else__`. 😁

PR #6263, opened 2022-10-06
test(autofix): Unit tests for autofix printer

nmote:
This tests the autofix printer to make sure that it is lifting the original text from the target and fix_pattern where appropriate. Details are in the main comment in `Unit_autofix_printer.ml`.

This also refactors Unit_engine to pull out the code that runs Semgrep matching over some code and rule, so that it can be used here.

Test plan: Picked an arbitrary semgrep-core test and modify it to change the matching behavior. Ensure that `make test` fails. Change one of the new unit tests so that the output doesn't match. Ensure that `make test` fails.

PR checklist:

### DIFF
--- a/semgrep-core/src/engine/Unit_engine.mli
+++ b/semgrep-core/src/engine/Unit_engine.mli
@@ -4,3 +4,12 @@
    'dune utop' from any location.
 *)
 val tests : unit -> Testutil.test list
+
+(* Can be used from other test code to concisely run Semgrep *)
+val match_pattern :
+  lang:Lang.t ->
+  hook:(Pattern_match.t -> unit) ->
+  file:string ->
+  pattern:string ->
+  fix_pattern:string option ->
+  Pattern_match.t list
--- a/semgrep-core/src/fixing/tests/README.txt
+++ b/semgrep-core/src/fixing/tests/README.txt
@@ -1,0 +1,2 @@
+This is a separate directory to avoid a circular dependency with semgrep_engine,
+since we need to run the engine as part of the unit tests
--- a/semgrep-core/src/fixing/tests/Unit_autofix_printer.ml
+++ b/semgrep-core/src/fixing/tests/Unit_autofix_printer.ml
@@ -1,0 +1,124 @@
+open Common
+
+type autofix_printer_test_case = {
+  target : string;
+  pattern : string;
+  fix_pattern : string;
+  expected : string;
+}
+
+(* Test suite for the python autofix printer.
+ *
+ * This runs Semgrep to get a pattern match, then renders the fix. However, it
+ * feeds in fake strings for the target and fix pattern contents, so that we can
+ * test which parts were lifted from the original source and which parts were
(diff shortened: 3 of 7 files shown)


semgrep-core/src/engine/Unit_engine.mli
```diff
@@ -4,3 +4,12 @@
    'dune utop' from any location.
 *)
 val tests : unit -> Testutil.test list
+
+(* Can be used from other test code to concisely run Semgrep *)
+val match_patterns :
```
aryx: You can rename it to just match_pattern as you take just one single pattern.

semgrep-core/src/fixing/tests/README.txt
```diff
@@ -0,0 +1,2 @@
+This is a separate directory to avoid a circular dependency with semgrep_engine,
+since we need to run the engine as part of the unit tests
```
aryx, line 2: An alternative strategy to avoid the circular deps is to make Unit_autofix_printer takes an additional function called semgrep_engine and pass it in Test.ml which has access to all the modules.
I do that in matching/Unit_matcher.ml where the parser is passed as a string, to avoid having matching/ depending on parsing/

semgrep-core/src/fixing/tests/Unit_autofix_printer.ml
```diff
@@ -0,0 +1,105 @@
...
+          let before = Str.first_chars fake_target_contents start in
+          let after = Str.string_after fake_target_contents end_ in
+          before ^ fix_text ^ after
+        in
+        Alcotest.(check string) fix_pattern expected full_fixed_text)
+  in
+  List.iter check
+    [
+      (* TODO foobar should be lifted from the pattern *)
+      ("foobar", "foo", "foo", "foobar");
```
aryx: maybe we could define a record for those tests, so it's easier to remember what each tuple element correspond to.
Maybe short names like {expexted: "foobar"; tgt: "foo"; pat:"foo"; fix"foobar"}

semgrep-core/src/fixing/tests/Unit_autofix_printer.ml
```diff
@@ -0,0 +1,105 @@
...
+ * fixes tested there are correct.
+ *
+ * TODO generalize the test runner once we have multiple languages to test
+ *)
+let test_python_autofix_printer () =
+  let check (expected, target, pattern, fix_pattern) =
+    Common2.with_tmp_file ~str:target ~ext:"py" (fun target_file ->
+        (* Make a fake string to use when printing instead of the target. Fill
+         * it with 't' so that we can identify it in the final output. *)
+        let fake_target_contents = String.make (String.length target) 't' in
```
aryx: those locals could be moved closer to where they are used.

aryx (approved):

semgrep-core/src/fixing/tests/README.txt
nmote: I considered a scheme like this, and will make the change if you would like me to, but I feel like this is more maintainable.

PR #6264, opened 2022-10-06
tainting: Allow propagation of taint through lambdas

### LINKED ISSUE #5971: Taint mode bug when matching the foreach statement
**Describe the bug**
The taint mode might not track data flow through `foreach` statement

**To Reproduce**
https://semgrep.dev/s/94LY

**Expected behavior**
The `foreach` statement should be tracked and the sink function in the provided code should be matched.

P1: important to fix or quite annoying. It might ignore lots of potential findings in the code

**Environment**
https://semgrep.dev/playground/

**Use case**
Tried to do taint mode analysis in the source code to do the security research and audit.

IagoAbal:
... via `pattern-propagators`.

This is useful for propagating taint through higher-order iterators such as `forEach` in Java.

Closes #5971
Closes PA-1813

test plan:
make test # added three tests

PR checklist:

### DIFF
--- a/semgrep-core/tests/rules/taint_propagator_lambda.java
+++ b/semgrep-core/tests/rules/taint_propagator_lambda.java
@@ -1,3 +1,7 @@
+// https://github.com/returntocorp/semgrep/issues/5971
+// only func4 truly tests taint propagators but we include the remaining examples
+// for extra coverage of taint-mode
+
 class Test {
 
     public void func1() {
--- a/semgrep-core/tests/rules/taint_propagator_lambda.java
+++ b/semgrep-core/tests/rules/taint_propagator_lambda.java
@@ -1,0 +1,31 @@
+class Test {
+
+    public void func1() {
+        String nodePred = request.getParameter("h");
+        String nodeSucc = nodePred.toString();
+        //ruleid: test
+        Runtime.getRuntime().exec(nodeSucc);
+    }
+
+    public void func2() {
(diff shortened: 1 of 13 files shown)


semgrep-core/tests/rules/taint_propagator_lambda.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public void func3() {
+        String nodePred = request.getParameter("h");
+        NodeSucc nodeSucc = new NodeSucc(nodePred);
+        //ruleid: test
+        Runtime.getRuntime().exec(nodeSucc);
+    }
+
+    public void func4() {
```
aryx, line 28: This feels like it should be the only tests.
What are testing the other func1/func2/func3?
I don't even see the difference between their code.

aryx (approved):

semgrep-core/tests/rules/taint_propagator_lambda.java
IagoAbal: That was the example reported by the user. For debugging, I always minimize the example. But for testing, I tend to include it as-is, with the hope to increase coverage.

IagoAbal: Clarified this in the test file. I think the other cases are useful to increase the test coverage of taint-mode, but we could perhaps separate non-minimized user-provided examples?

PR #6265, opened 2022-10-06
feat!: dot operators and js opt chaining

### LINKED ISSUE #6201: Semgrep does not support TypeScript  `?` in pattern syntax
**Describe the bug**
In typescript you can write things like this:

`req.user?.accountId` 

Patterns do not accept that as valid and fall over with a parsing error, customers have had pain writing some typescript rules because of this.

**Expected behavior**
support ? as pattern syntax  see the link here https://semgrep.dev/s/1XOk

**Priority**
How important is this to you?
- [x] P2: annoying but not blocking me

brandonspark:
Closes #6201
Closes PA-1943

## What:
This PR adds in a new type to `AST_generic`, called `dot_operator`. This is used to differentiate different kinds of dot accesses.

## Why:
In Javascript, you can do a normal dot access of `x.y`, but also "opt chaining", which looks like `x?.y`. This is essentially just `Option.bind`, but it's a special piece of syntax in the language.

Right now, with just vanilla `DotAccesses`, we have no choice but to match them the same, or to embed the `x` or `y` into weird new constructs. Notably, we can't do this the same way we do it in Rust, where `x?.y` is understood to be the result of `(x?).y`, or the result of doing a dot access on the result of `x?`, because Javascript only has the chaining, and not the standalone postfix `?` operator.

Additionally, adding this new variant instead of offloading this change into the `x` or `y` preserves our current semantic logic as much as possible, because we want to maintain the simple case where the `DotAccess` is taken of something which is a simple identifier.

## How:
Added a new `dot_operator` type, which contains two variants, `Dot` and `QuestDot`. This may be modified in the future to allow more kinds of dot accesses, such as for Kotlin, which has some use cases for this.

PR checklist:


IagoAbal:
@brandonspark the CI failures should go away if you rebase against `develop`.

IagoAbal:
@brandonspark remember to add `Closes #NNNN` to the PR comment so that 1) we keep traceability and 2) the issue is closed.

changelog.d/pa-1943.fixed
```diff
@@ -0,0 +1 @@
+JS/TS: Allowed parsing of patterns using the optional chaining "?." syntax.
```
IagoAbal: This should be `gh-6201.fixed`, we should always use the GitHub issue if it exists. Users don't have access to our Linear board.

IagoAbal:
> Notably, we can't do this the same way we do it in Rust, where x?.y is understood to be the result of (x?).y, or the result of doing a dot access on the result of x?, because Javascript only has the chaining, and not the standalone postfix ? operator.

Does it cause any specific issue to translate `x?.y` as `(x?).y` in JS? JS may not have `x?` but our Generic representation does have it (I'm assuming), and _IF_ the semantic meaning is the same then I don't see the problem.

aryx:
Yes, is it possible to unsugar x?.y  in something that would not require to change our DotAccess?
Or we have to change DotAccess?

brandonspark:
Sorry, I requested review again by accident.

semgrep-core/src/parsing/tree_sitter/Parse_typescript_tree_sitter.ml
```diff
@@ -1097,11 +1098,10 @@ and member_expression (env : env) ((v1, v2, v3) : CST.member_expression) : expr
     =
   let expr = expr_or_prim_expr env v1 in
   (* TODO: distinguish optional chaining "?." from a simple access "." *)
```
nmote: Remove this todo

semgrep-core/src/parsing/pfff/js_to_generic.ml
```diff
@@ -271,7 +271,17 @@ and expr (x : expr) =
...
   | ObjAccess (v1, t, v2) -> (
-      let v1 = expr v1 in
+      let e = expr v1 in
+      let t, is_quest =
+        match t with
+        | Dot, tok -> (info tok, false)
+        | QuestDot, tok -> (info tok, true)
+      in
+      let v1 =
+        if is_quest then
```
aryx: you could move all this code in the match above and avoid having this intermediate is_quest.
let v1 = match t with | Dot _ -> e | QuestDot, tok -> G.Call ....

aryx (approved):
Can you add a comment in AST_generic.ml next to ObjAccess
where you discuss that x?.y is translated in Call Elvis ...
and why we chose this design over adding a new dot_operator type.

PR #6276, opened 2022-10-10
[RFC] Beginning of autofix printing statistics

nmote:
The goal here is to record autofix printing statistics, like we do parsing statistics. We want to look over rules repositories (starting with semgrep-rules), run all the rules that have both a `fix` field and a test case, and measure what fraction of them semgrep-core is able to render a fix for.

I'm requesting feedback on this early because I'm uncertain of my decision to implement this in Bash. I thought it would be a quick and dirty script but it quickly became challenging to write and read. Maybe that's okay since it's not likely to change very often, and isn't part of production, but I'm not sure.

Should I:

1. Finish this Bash script, add comments, and improve names, then add a Python script like `parsing-stats/upload-parsing-rates` to upload the statistics
2. Rewrite this in Python
3. Put as much of this in OCaml as I can, like is done for parsing stats

I am leaning towards (1) or (2) but I would appreciate feedback

PR checklist:

### DIFF
--- a/autofix-printing-stats/.gitignore
+++ b/autofix-printing-stats/.gitignore
@@ -1,0 +1,2 @@
+tmp
+stats.json
--- a/autofix-printing-stats/projects.txt
+++ b/autofix-printing-stats/projects.txt
@@ -1,0 +1,1 @@
+https://github.com/returntocorp/semgrep-rules.git
--- a/autofix-printing-stats/run
+++ b/autofix-printing-stats/run
@@ -1,0 +1,141 @@
+#!/usr/bin/env bash
+#
+# Fetch repositories containing Semgrep rules and test cases, run Semgrep over
+# those with `fix` properties, and gather stats on how many fixes are
+# successfully rendered by semgrep-core.
+
+set -eu -o pipefail
+progname=$(basename "$0")
+
+usage() {
+  cat <<EOF
+Usage: $progname [--upload]
+
+Expects:
+- semgrep-core command must be available
+
+Produces stats.json.
+
+Example: $progname
+EOF
+}
+
+echo using `which semgrep-core`
+
+projects_file=projects.txt
+
+fetch_project_files() {
+  # Note that project names are not unique
+  project=$(basename "${url%.git}")
+  org=$(basename $(dirname "${url%.git}"))
+  name="$org-$project"
+  project_list+=" $name"
+
+  mkdir -p tmp
+  (
+    cd tmp
+    if [[ ! -d "$name" ]]; then
+      echo "Cloning '$name' from '$url'."
+      # Since we do not care about revision history or
+      # Git Large File Storage files, we can shallow clone
+      # and ignore LFS pointers to expedite cloning.
+      GIT_LFS_SKIP_SMUDGE=1 git clone --depth 1 "$url" "$name"
+    else
+      echo "Using local git repo for '$name'."
+      origin_url=$(git -C "$name" remote get-url origin)
+      if [[ "$url" != "$origin_url" ]]; then
+        cat >&2 <<EOF
+Wrong remote URL found in cloned repository '$name':
+  found $origin_url
+  expected $url
+Check that you don't have two project URLs with the same repo name.
+EOF
+        exit 1
+      fi
+    fi
+  )
+}
+
+path_to_lang() {
+  path=$1
+  ext="${path##*.}"
+  lang=""
+  case $ext in
+    py)
+      lang=python
+  esac
+}
+
+record_results() {
+  lang=$1
+  all=$2
+  successes=$3
+  if [[ ! ( -v totals_by_language[$lang] ) ]] ; then
+    totals_by_language[$lang]=0
+  fi
+  if [[ ! ( -v successes_by_language[$lang] ) ]] ; then
+    successes_by_language[$lang]=0
+  fi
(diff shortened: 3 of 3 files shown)


aryx:
maybe @mjambon has an opinion on this given his experience writing the parsing stats in the first place.

aryx:
I would vote for 3, but really I think you should do whatever you think is best.

nmote:
The reason I'm reluctant to do (3) is because all the information needed to compute these stats is already available from semgrep-core's normal output, unlike for parsing where we actually do need to modify semgrep-core. Since we can accomplish this without actually changing the binary that we ship to production, I'd slightly prefer to do this elsewhere.

emjin:
> Maybe that's okay since it's not likely to change very often, and isn't part of production, but I'm not sure.

I actually think it's less okay because it's not likely to change very often. It means that people are likely to forget how the script works and maintaining it becomes really annoying. I would vote for either 2 or 3 with a preference for 3.

Alternatively, is there a 4th option of writing the script separately but in OCaml? I did that for the `examples-annotations` benchmark and the experience was actually really nice; the smoothest coding experience I've ever had. And we'd have types :D

mjambon:
Overall, I concur with @emjin and @pad.

Bash
--

Bash is hard to learn, it's limited in features, and more importantly, It's not reasonable to expect our team members to be highly proficient in Bash. Note that most people think they know how to write shell scripts because they use the shell interactively all the time but they don't. That said, I tend to use Bash a lot for tasks that involve mostly calls to external programs and moving files around. The advantages I find are that Bash is pre-installed everywhere (although if you want the modern features, it gets problematic on MacOS because it comes by default with an antiquated version) and that I can quickly make a script that works, that's safe, and that's reasonably easy to read.

Python
--

I never found that Python was a good companion to OCaml like Bash because there's a lot of overlap in the features they offer and both require heavy tooling. However, if our primary language was Python, I think I would not hesitate to advocate for Python instead of Bash in pretty much all situations. What I don't like in Python as a replacement for shell scripts is:

1. It comes with a massive runtime, possibly in the wrong version (python 2 vs. 3).
2. It doesn't have a convenient syntax to call external commands, redirect IOs, etc. For example, to call `ls` in Bash, you just write `ls`. Python, OCaml, heck even Perl doesn't have that.

(1) is not a problem at all if you already have a whole Python environment set up. (2) is a somewhat minor issue. It makes code more verbose but in exchange, it's not mysterious to readers who don't know shell syntax but know Python. However, we're not a team of expert Python programmers. Unlike many people, I don't think it's reasonable to expect all OCaml programmers to be competent in Python and its ecosystem. There are more interesting languages and platforms out there, in my opinion.

My inclination is to have one language per project since each project is managed by a team and it's important that all team members have some control over all parts of the project. If there were a separate project for running parsing stats or autofix stats and it was justified because it's maintained by a separate team with separate concerns, then they should probably use the most appropriate tool for the job, be it Python or something else.

OCaml
--

OCaml could be decent for writing programs that make a lot of external command calls and deal with files a lot. The benefit of static typing for small programs without advanced data structures is mild. The main benefit of writing all our "scripts" in OCaml would be that everyone on the team that manages the semgrep repo is familiar with OCaml and the tools. Maybe our OCaml scripts would be not the shortest or not the most readable by a non-OCaml programmer, but they would be easy to maintain using the usual toolchain. Type `make` and `make test` and have everything typechecked, built, and tested, without having to think differently about any of that.

What's lacking in OCaml is a standard way of writing the equivalent of shell scripts. Not having a standard way of doing X is a common issue in OCaml and not something that stops the average OCaml programmer. We can adopt third-party libraries to deal with shell-like problems (running external commands, IO redirections) or even make our own. It's not that hard.

The conclusion for OCaml is the same as for Python. I think it makes more sense to have a whole project in the same language. Anything written in secondary languages will tend to get outdated faster due to the lack of investment of the team in these languages.

OCaml tools
--

I personally have made a few attempts to do shell-like programming in OCaml:
* [ocamlscript](https://github.com/ocaml-community/ocamlscript): is for writing single-file OCaml programs that are compiled to native code and don't require a knowledge of a build system. I haven't used it since I stopped doing bioinformatics 15 years ago. I don't think the feature it offers is valuable to us since we don't mind having build and test steps and a whole project setup.
* A hypothetical language that would allow embedding shell code in OCaml and vice-versa, transpiling everything to pure OCaml: https://gist.github.com/mjambon/bb07b24f89fa60c973735307ce9c6cb9

There are of course libraries that people have made to facilitate shell-like programming in OCaml. This would require some research. Something like [Feather](https://github.com/charlesetc/feather) could be useful or maybe just what we need.

Conclusion
--

* I would favor investing in tooling for writing shell-like scripts in OCaml within the semgrep repo rather than Bash or Python, for better, happier maintenance.
* Correct me if I'm wrong but I don't think there's an emergency in moving away from either Bash or Python.
* I volunteer to set up something that allows us to write shell-like scripts in OCaml, which would be developed like the OCaml code in semgrep-core but would result in separate binaries.
* Being able to call semgrep-core as a function that takes a `Runner_config.conf` object would be handy.

brandonspark:
I cannot read Bash so I vote for 2 or 3

nmote:
Thanks everyone for your great input! I am convinced that I should do this in OCaml. I won't make it part of semgrep-core, but instead will just write a standalone script. I would like to take @mjambon up on his offer to create tooling to support this, but for now I will probably just write this as a single file that can be run directly with `ocaml`.

PR #6279, opened 2022-10-11
tainting: Fix field-sensitivity regression

IagoAbal:
Before 0.113.0, each dot-access was an l-value expression on its own,
and we checked l-values bottom-up like any other expression. But in
0.113.0 we began representing a chain of dot-accesses as a single
l-value, and sub-lvalues were no longer checked as expressions of their
own, leading to regressions.

Closes PA-1928
Followed by PR returntocorp/semgrep-rules#2504

test plan:
make test # added one test

PR checklist:

### DIFF
--- a/semgrep-core/src/core/il/IL.ml
+++ b/semgrep-core/src/core/il/IL.ml
@@ -171,7 +171,20 @@ and base =
   (* THINK: Mem of exp -> Deref of name *)
   | Mem of exp
 
-and offset = { o : offset_kind; oorig : orig }
+and offset = {
+  o : offset_kind;
+  oorig : orig;
+      (** `oorig' should be a DotAccess expression and gives us the corresponding
+      * Generic expression for a sub-lvalue. Now that we represent `x.a.b.c` as
+      * { base = "x"; rev_offsets = ["c"; "b"; "a"]}, it makes it very easy to
+      * check whether a sub-lvalue is a source/santizer/sink by just checking
+      * the range of the `oorig'.
+      *
+      * alt: We could compute the range of the sub-lvalue from the ranges of all
+      *      the offsets in the sub-lvalue, but this is probably less efficent
+      *      unless we cache the range here. So it seems better to have `oorig'.
+      *)
+}
 
 and offset_kind =
   (* What about computed field names?
--- a/semgrep-core/src/core/il/IL.ml
+++ b/semgrep-core/src/core/il/IL.ml
@@ -171,7 +171,9 @@ and base =
   (* THINK: Mem of exp -> Deref of name *)
   | Mem of exp
 
-and offset =
+and offset = { o : offset_kind; oorig : orig }
+
+and offset_kind =
   (* What about computed field names?
    * - handle them in Special?
    * - convert in Index with a string exp?
--- a/semgrep-core/src/tainting/Dataflow_tainting.ml
+++ b/semgrep-core/src/tainting/Dataflow_tainting.ml
@@ -33,13 +33,19 @@ let logger = Logging.get_logger [ __MODULE__ ]
 (*****************************************************************************)
 (* Tainting dataflow analysis.
  *
- * This is a very rudimentary tainting analysis.
- * It is a MAY analysis, it finds *potential* bugs (the tainted path could not
- * be feasible in practice).
- * Very coarse grained (taint whole array/object).
- * This is step1 for taint tracking support in semgrep.
- * This was originally in semgrep-core/src/analyze, but it now depends on Pattern_match,
- * so it was moved to semgrep-core/src/engine
+ * - This is a rudimentary taint analysis in some ways, but rather complex in
+ *   other ways... We don't do alias analysis, and inter-procedural support
+ *   (for DeepSemgrep) still doesn't cover some common cases. On the other hand,
+ *   almost _anything_ can be a source/sanitizer/sink, we have taint propagators,
+ *   etc.
+ * - It is a MAY analysis, it finds *potential* bugs (the tainted path could not
+ *   be feasible in practice).
+ * - Field sensitivity is limited to l-values of the form x.a.b.c, see module
+ *   Taint_lval_env and check_tainted_lval for more details. Very coarse grained
+ *   otherwise, e.g. `x[i] = tainted` will taint the whole array,
+ *
+ * old: This was originally in semgrep-core/src/analyze, but it now depends on
+ *      Pattern_match, so it was moved to semgrep-core/src/engine.
  *)
 
 module DataflowX = Dataflow_core.Make (struct
@@ -74,6 +80,11 @@ type config = {
   is_propagator : AST_generic.any -> a_propagator tmatch list;
   is_sink : G.any -> R.taint_sink tmatch list;
   is_sanitizer : G.any -> R.taint_sanitizer tmatch list;
+      (* NOTE [is_sanitizer]:
+       * A sanitizer is more "extreme" than you may expect. When a piece of code is
+       * "sanitized" Semgrep will just not check it. For example, something like
+       * `sanitize(sink(tainted))` will not yield any finding.
+       * *)
   unify_mvars : bool;
   handle_findings : var option -> T.finding list -> Lval_env.t -> unit;
 }
@@ -349,7 +360,7 @@ let find_args_taints args_taints fdef =
 (* Tainted *)
 (*****************************************************************************)
 
-let sanitize_lval lval_env sanitizer_pms lval =
+let sanitize_lval_by_side_effect lval_env sanitizer_pms lval =
   let lval_is_now_safe =
     (* If the l-value is an exact match (overlap > 0.99) for a sanitizer
      * annotation, then we infer that the l-value itself has been updated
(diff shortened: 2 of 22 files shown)


semgrep-core/src/core/il/IL.ml
```diff
@@ -171,7 +171,9 @@ and base =
   (* THINK: Mem of exp -> Deref of name *)
   | Mem of exp
 
-and offset =
+and offset = { o : offset_kind; oorig : orig }
```
emjin: Can you explain here why `oorig` is necessary?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -323,26 +349,29 @@ let find_args_taints args_taints fdef =
...
 (* Check if an expression is sanitized, if so, return a new variable environment. *)
 let exp_is_sanitized env exp =
   match orig_is_sanitized env.config exp.eorig with
   | [] -> None
   | sanitizer_pms -> (
       match exp.e with
-      | Fetch { base = Var var; rev_offset = [] } ->
-          Some (sanitize_var env.lval_env sanitizer_pms var)
-      | _ -> Some env.lval_env)
+      | Fetch lval when LV.lval_is_var_and_dots lval ->
```
emjin, line 382: It's clear from the PR, but if I were writing this code I would not necessarily realize I needed this restriction. Personally I would prefer to separate the condition for this into its own type if possible. It just really changes how I think about this

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -383,128 +412,171 @@ let handle_taint_propagators env x taints =
...
-  let var_env =
+  let lval_env =
     match x with
-    | `Var var ->
-        (* If `x` is a variable, then taint is propagated by side-effect. This
-         * allows us to e.g. propagate taint from `x` to `y` in `f(x,y)`. *)
-        Lval_env.add_var var taints_incoming var_env
+    (* If `x` is a variable, then taint is propagated by side-effect. This
+     * allows us to e.g. propagate taint from `x` to `y` in `f(x,y)`. *)
+    | `Lval lval when LV.lval_is_var_and_dots lval ->
```
emjin, line 435: This one makes it more clear why the restriction is needed, but I'm not a huge fan of having to add the restriction for each case---what if I added a new case which required it but didn't realize? This comment also is still a little confusing because I'm not sure why I would want to propagate taint from `x` to `y` (is it if there's a specific propagator asking for it?). And, if I just saw this, I would wonder, why does it have to be "var and dots" and not just any var?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -383,128 +412,171 @@ let handle_taint_propagators env x taints =
...
+
+let find_lval_taint_sources env ~labels lval =
+  let source_pms = lval_is_source env.config lval in
+  let mut_source_pms, reg_source_pms =
+    (* If the variable is an exact match (overlap > 0.99) for a source
+       * annotation, then we infer that the variable itself is now tainted
+       * (presumably by side-effect) and we will update the `var_env`
+       * accordingly. Otherwise the variable belongs to a piece of code that
+       * is a source of taint, but it is not tainted on its own. *)
+    List.partition is_exact source_pms
```
emjin: Your comment is helpful but I'm still confused. Maybe an example?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -383,128 +412,171 @@ let handle_taint_propagators env x taints =
...
+  and taints_sources_mut =
+    mut_source_pms |> taints_of_matches |> filter_taints_by_labels labels
   in
-  match sanitizer_pms with
+  let lval_env = Lval_env.add lval taints_sources_mut env.lval_env in
+  (Taints.union taints_sources_reg taints_sources_mut, lval_env)
+
+let rec check_tainted_lval env (lval : IL.lval) : Taints.t * Lval_env.t =
+  let new_taints, lval_in_env, lval_env = check_tainted_lval_aux env lval in
+  let taints_from_env = status_to_taints lval_in_env in
```
emjin, line 464: This is interesting; I'm still not clear on why we add this `status_to_taints` and what the implications are

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -383,128 +412,171 @@ let handle_taint_propagators env x taints =
...
+   *  comments below.
+   *)
+  match lval_is_sanitized env.config lval with
   (* TODO: We should check that taint and sanitizer(s) are unifiable. *)
-  | _ :: _ ->
-      let lval_env' = sanitize_var env.lval_env sanitizer_pms var in
-      (Taints.empty, lval_env')
+  | _ :: _ as sanitizer_pms ->
+      (* NOTE [lval/clean]:
+       *  If lval is sanitized, then we will "bubble up" the `Clean status, so any
```
emjin, line 484: Nit: I would think of this as "bubble down".

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -383,128 +412,171 @@ let handle_taint_propagators env x taints =
...
+   *  with the info we have stored in `env.lval_env`. This can be subtle, see
+   *  comments below.
+   *)
+  match lval_is_sanitized env.config lval with
   (* TODO: We should check that taint and sanitizer(s) are unifiable. *)
-  | _ :: _ ->
-      let lval_env' = sanitize_var env.lval_env sanitizer_pms var in
-      (Taints.empty, lval_env')
+  | _ :: _ as sanitizer_pms ->
+      (* NOTE [lval/clean]:
```
emjin, line 483: Why not bubble the sanitization status when the sanitization happens? (Or is that what's happening? I did not look at when this function is called, so maybe it's just a confusion with the name)

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -383,128 +412,171 @@ let handle_taint_propagators env x taints =
...
-         * accordingly. Otherwise the variable belongs to a piece of code that
-         * is a source of taint, but it is not tainted on its own. *)
-        List.partition (fun src -> src.overlap > 0.99) source_pms
+      (* Recursive call, check sub-lvalues first.
+       *
+       * It needs to be done bottom-up because any sub-lvalue can be a source and a
+       * sink by itself, even if an extension of lval is not. For example, given
+       * `x.a.b`, this lvalue may be considered sanitized, but at the same time `x.a`
+       * could be tainted and considered a sink in some context. We cannot just check
+       * `x.a.b` and forget about the sub-lvalues.
```
emjin, line 505: Wait---is `x.a` a sub-lvalue of `x.a.b`? This is just a terminology thing.

emjin (approved):
Left a bunch of questions. Overall this makes sense, but I am still pretty unclear on how `lval_is_sanitized` works. Maybe some examples of scenarios there would be helpful, or a summary of the steps taken. Or a restructuring?

I also would like more tests given how many cases I feel like there are. Looking at `lval_is_sanitized`, there seems to be interactions between sources, propagators, and sanitizers at different parts of the fields. I would like some tests for some of the combinations. One thing I've noticed is our tests usually just have one source + sink + sanitizer, but real rules usually have at least three of each. For this PR, at least from my quick read, it feels like those interactions could be important

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: What do you mean by "separating the condition for this into its own type" ?

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: I take note that this is unclear and I'll add a TODO for a refactoring, maybe should add a task too in Linear. This may involve encapsulating more of this logic inside `Taint_lval_env`. I'm adding some more comments for now. But to clarify this, we are only field sensitive on lvalues of the form `x.a.b.c` which is what `lval_is_var_and_dots` is testing.

In this function we are processing `pattern-propagators`, so anything could be propagated, I was just trying to provide an example of taint-propagation "by side effect". I will have to work on taint doc for a while, bear with me.

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: Have you checked the definition of `status_to_taints`?

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: I think it goes "up" ? You can probably find a better name than me. But essentially if we have `x.a.b.c`, and we're checking `x.a` and it's "clean" (e.g. because it has been sanitized), then we "propagate up" the "clean" to `x.a.b` and `x.a.b.c`.

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: Yep, is it confusing ?

IagoAbal:
Had to resolve conversations to merge but we can continue the discussion here or via Slack, and I can open more PRs to improve docs. I should be updating the taint docs for users next week, so that may also help. Thanks for the review!

PR #6281, opened 2022-10-11
feat(lsif): pfff update, exposed generic functions

brandonspark:
## What:
This change adds some `pfff` changes that are needed for the LSIF experiment, and exposed some functions which are helpful for the LSIF experiment

## Why:
This change is a part of the LSIF experiment, as described by the scope doc here: https://www.notion.so/r2cdev/Semgrep-LSIF-Querying-e9261a3fbee94084a129014d599045bd

We are trying to use the LSIF format to get type information for free, instead of needing to implement a Typescript typechecker ourselves. To that end, DeepSemgrep needs an LSIF command-line flag to enable the experiment.

PR checklist:

### DIFF
--- a/semgrep-core/src/core_cli/Core_CLI.ml
+++ b/semgrep-core/src/core_cli/Core_CLI.ml
@@ -358,7 +358,6 @@ let mk_config () =
     match_format = !match_format;
     mvars = !mvars;
     lsp = !lsp;
-    lsif = false;
     timeout = !timeout;
     timeout_threshold = !timeout_threshold;
     max_memory_mb = !max_memory_mb;
--- a/semgrep-core/src/core_cli/Core_CLI.ml
+++ b/semgrep-core/src/core_cli/Core_CLI.ml
@@ -358,6 +358,7 @@ let mk_config () =
     match_format = !match_format;
     mvars = !mvars;
     lsp = !lsp;
+    lsif = false;
     timeout = !timeout;
     timeout_threshold = !timeout_threshold;
     max_memory_mb = !max_memory_mb;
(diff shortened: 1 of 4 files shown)


aryx:
need to merge develop first and solve the conflicts.

aryx (approved):
We probably want to move the lsif/lsp features to deep-semgrep.
It's a bit ugly to have deep-semgrep specific flags to Runner_config.ml, but
not sure we can do something else.

aryx:
Why do we need the lsif flag in Runner_config?

brandonspark:
> Why do we need the lsif flag in Runner_config?

The `Runner_config.t` is used in `Deep`. The type inherits. So we have to change the type in `semgrep` to make it available in DeepSemgrep's config.

aryx:
You dont need to use runnerconfig. You can pass flag around in deepsemgrep without runnerconfig

brandonspark:
> You dont need to use runnerconfig. You can pass flag around in deepsemgrep without runnerconfig

Would we really prefer to add another argument to `build_idx_and_setup_hooks_for_deep_semgrep`? I can do it, but I figured that it would start getting not nice when we have 5+ arguments to the function. Since we already have so many command-line options in `Runner_config.t` (which are set via cmdline), I figured it fit there.

emjin:
Maybe we can make a `Deep_semgrep_config`? I agree it's not great to have too many arguments to a function

emjin (approved):

aryx:
I would vote for Deep_semgrep_config.
I would try hard to avoid having deep-specific code in semgrep when we can.

PR #6284, opened 2022-10-11
fix(autofix): Add binary operator expression printing

nmote:
This does two things.
* Separates `ordinary_call` and `opcall` printing into two different methods. Binary expressions are desugared in the generic AST to a function application where the target is an operator. Since this is not language-specific, I put the logic to differentiate these in the base_printer.
* Implements parenthesis insertion. As demonstrated by the test cases, parentheses must be added surrounding certain AST nodes. For now, this inserts too many parentheses in the spirit of UglyPrinter, but we can refine this later by tracking the context in which a node appears.
  * As part of this, I removed the `print_any` in `Hybrid_print.ml`. When print avoidance succeeds there, it bypasses the parenthesis insertion in `print_expr`.

These couldn't easily be separated because without implementing binary expression printing, there's no way to test parenthesis insertion. Without implementing parenthesis insertion, binary expression printing would lead to a regression over textual autofix (I can give an example if anyone is interested).

I also filled in metavar replacement and print-avoidance for expressions cases so that the tests can run.

Test plan: Automated tests

PR checklist:


nmote:
The snapshot changes are just formatting differences. The file in question still gets its line count reduced, so it is still testing what it was meant to test. The reason for the change is that now we can lift text from the original target for all expressions, so we can successfully print `requests.post`.

semgrep-core/src/core/ast/Ugly_print_AST.ml
```diff
@@ -107,12 +137,49 @@ class base_printer : printer_t =
...
+    method private print_expr_without_parens { e; _ } = self#print_expr_kind e
     method print_expr_kind _ = print_fail ()
     method print_argument _ = print_fail ()
     method print_arguments _ = print_fail ()
 
+    method print_call e args =
+      match e.G.e with
+      (* Binary operator expressions are desugared to calls, but are different
+       * enough that they merit a separate method here. *)
+      | G.IdSpecial (G.Op _, _) -> self#print_opcall e args
```
aryx, line 173: Why not pass the op as an argument instead of e, that way the signature of print_opcall is also clearer.

semgrep-core/src/core/ast/Ugly_print_AST.ml
```diff
@@ -107,12 +137,49 @@ class base_printer : printer_t =
...
+      | _ -> self#print_ordinary_call e args
+
+    method print_opcall _ _ = print_fail ()
+    method print_ordinary_call _ _ = print_fail ()
+
     (* TODO Add more nodes as needed. *)
+
+    method private needs_parens =
+      function
+      | G.E { e = G.Call ({ e = G.IdSpecial (G.Op _, _); _ }, _); _ } -> true
```
aryx, line 187: maybe add a comment saying that by default we are overly defensive and print extra parents each time.

aryx (approved):

semgrep-core/src/core/ast/Ugly_print_AST.ml
nmote: I did this so that print-avoidance would work for operators. Because we can look up expressions and use their original text, when `print_opcall` calls `print_expr`, this just works. If we passed down the operator, it wouldn't. There is no `any` variant for `operator` (and even if there were, it doesn't have location information so it would require some additional tweaking), or is there one for `special` or `special wrap`. Because the external printer works by storing a table with `AST_generic.any` as the key, making this work would require some changes.

Either we can pass down the expression here, which I don't think makes printing any more onerous since you would just call `print_expr` instead of `print_operator`, or add a variant to `AST_generic.any`, or change how the external printer works to allow it to print non-any nodes, which would mean more boilerplate. I prefer this solution, though I think the last may also be worth exploring. I do not think it is a good idea to add an `any` variant to support this.

I'll add a comment explaining this.

nmote: I'll also add a test that verifies that operators are lifted from the original text (since we can't currently print them, the fact that the existing tests pass are proof of that, but once we implement that, nothing will prevent someone from inadvertently breaking it)

PR #6300, opened 2022-10-12
test: Autofix printing stats

nmote:
This sets up a tool that clones rules repositories, finds the rules that have a `fix:` field, runs them against their test cases, and counts how many semgrep-core can succesfully render using AST-based autofix.

Optionally, it can also upload this data to our metrics dashboard.

This will allow us to track the performance of AST-based autofix on real-world rules.

I leaned heavily on the parsing-stats code when creating this. The part that clones the repositories is pretty much a direct translation from Bash into OCaml.

I wrote this in the style of a shell script, making use of commonly-available tools where appropriate. Down the line, we may want to replace some of these with various OCaml utilities, but for now I think this is reasonable.

I wrote this in OCaml after the discussion on
https://github.com/returntocorp/semgrep/pull/6276. My initial attempt to write it in Bash did not turn out especially well.

I will follow up to create a CI job that runs this daily.

Test plan:

Manually run, both with and without the `--upload` flag. Without, see that the stats for Python are printed. With, see that the stats for Python are printed and
https://dashboard.semgrep.dev/metric/semgrep.core.python.autofix.pct is updated.

Verify that subsequent runs without removing the `tmp` directory are successful as well.

PR checklist:


aryx:
@mjambon might have feedback based on his experience with ocamlscript and other DSL for shell-in-ocaml

mjambon (approved):

mjambon:
I think this is better than if it were in Bash, given the expertise of our team. It's fine if it works right now but I'm curious about how to make this easier in the future. The things I remember would help are:

* no need to do `open Printf` (in that respect, `open Base` would be better)
* primitives for running commands and IO redirections ([Feather](https://github.com/charlesetc/feather))
* support for ocamlfind packages i.e. the `#require` directive. This can be achieved with `#use "topfind";;`.
* a richer standard library (I mean actually standard and loved by the whole community, which doesn't exist)
* alternatively, specialized libraries for dealing with strings and files conveniently. Regexp syntax is pretty bad in all languages. I was very productive with [mikmatch](https://github.com/mjambon/mikmatch), which I used a lot back in the day (but also is a non-standard syntax, which is incompatible with dev tools).
* making scripts compiled as part of CI, if not run, to catch any regression and need for maintenance. I'm not sure how to do that without adding extra files for a build system. Looking for executable files with an ocaml shebang line and running them with `--help` could be sufficient to ensure that everything's in order.

Based on previous experience, I would limit the use of scripts to single-file programs i.e. no includes via `#use`. If we have a need to split a stats script into multiple modules (files), then it's probably best to write a `dune` file (with the caveat that we still need something to ensure CI checks if the code resides outside of the main OCaml codebase (`semgrep-core/`)).

mjambon:
https://github.com/ocaml-community/ocamlscript should still be usable. I think it no longer requires camlp4 but I'm not sure how well ppx extensions are supported. A very easy command-line parsing utility like https://github.com/hammerlab/ppx_deriving_cmdliner would be great (especially with cmdliner completely buried out of user's sight).

PR #6320, opened 2022-10-14
osemgrep: support for --config p/ocaml

aryx:
test plan:
$ time ./semgrep-core/bin/osemgrep --config p/ocaml --json cli/ semgrep-core/src/
but does not work yet!

PR checklist:

### DIFF
--- a/semgrep-core/src/osemgrep/cli_scan/Semgrep_dashdash_config.ml
+++ b/semgrep-core/src/osemgrep/cli_scan/Semgrep_dashdash_config.ml
@@ -72,6 +72,9 @@ let load_rules_from_file file =
 (* Entry point *)
 (*****************************************************************************)
 
+open Lwt.Syntax
+open Cohttp_lwt_unix
+
 let rules_from_dashdash_config config_str =
   (* TOPORT: resolve rule file URLs; for now we assume it's a file. *)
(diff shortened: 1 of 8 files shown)


semgrep-core/src/osemgrep/cli_scan/Semgrep_dashdash_config.ml
```diff
@@ -10,20 +14,82 @@
...
+  logger#debug "loading local config from %s" file;
+  Parse_rule.parse_and_filter_invalid_rules file
 
 (*****************************************************************************)
 (* Entry point *)
 (*****************************************************************************)
 
-let rules_from_dashdash_config config =
+open Lwt.Syntax
+open Cohttp_lwt_unix
```
IagoAbal, line 76: I would use a module alias instead of an `open` here. Not being familiar with Cohttp I would prefer to easily identify what types and functions belong to it. For example, I skimmed the code and I was wondering what that `Client` module was.

aryx: good idea

aryx: for Lwt.syntax we have to open it, to benefit from the let*, but I can add a comment about it.

PR #6327, opened 2022-10-17
tainting: Add new rule options

IagoAbal:
1. `taint_assume_safe_functions` to assume that function calls do not propagate taint from their arguments to their result.
2. `taint_assume_safe_indexes` to assume that a tainted index does not cause an access expression to be tainted.

Both should help reducing FPs. Also, `taint_assume_safe_functions` is meant to replace the `not_conflicting` option in `pattern-sanitizers`, but is not enough by itself. This e.g. `sink(not_tainted(tainted('a')))` will still be flagged due to how sinks work right now (anything within the range of the sink is considered a sink, also `tainted('a')`). We may also need to generalize `pattern-propagators` so you can enumerate the functions that propagate taint.

Closes PA-1541

test plan:
make test # added two tests

PR checklist:


aryx (approved):

ievans:
@IagoAbal this looks great, are these intended to be stable / relied on by external users or should we market them as experimental / internal?

IagoAbal:
> @IagoAbal this looks great, are these intended to be stable / relied on by external users or should we market them as experimental / internal?

`taint_assume_safe_indexes` could be considered stable and can be relied on, the only question is whether to change the default to `true` instead of `false`. The current default value (`false`) is more appropriate for audit rules.

`taint_assume_safe_functions` is meant to be stable, but we could not enable it by default right now due to the problems described in the PR comment, so for now I consider this one experimental.

I'll clarify this in the changelog.

PR #6329, opened 2022-10-17
fix(autofix): Abort on failure to replace metavar anywhere in string literal

nmote:
This is an extension of #6242. What I didn't realize when I wrote that was that textual autofix allows a metavariable to appear anywhere within a string literal. This makes sense, since the implementation can't distinguish between the inside of a string literal and anything else.

Now, after performing metavariable replacement, we search string literals for any instance of metavariables, and we abort if we find one. Previously, we only aborted if the string literal's contents was, exactly, the string literal in question.

Test plan: The test plan of #6242, then repeated with the fix changed to `bar("x$Xy")`

Also repeat the test with the pattern `foo("$X", "$Y")`, the target `foo("asdf", "bar")`, and the fix `bar("$Y", "$X")` to test that the regular expression is joined properly.

PR checklist:

### DIFF
--- a/semgrep-core/src/fixing/Autofix_metavar_replacement.ml
+++ b/semgrep-core/src/fixing/Autofix_metavar_replacement.ml
@@ -107,6 +107,20 @@ let replace metavar_tbl pattern_ast =
  * *)
 let find_remaining_metavars metavar_tbl ast =
   let seen_metavars = ref [] in
+  let str_metavars_regexp =
+    lazy
+      ((* List of metavars that were bound in this match, quoted so that they
+        * can be used safely in a regex *)
+       let quoted_metavars =
+         Hashtbl.to_seq_keys metavar_tbl |> Seq.map Str.quote |> List.of_seq
+       in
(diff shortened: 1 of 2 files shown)


semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -107,6 +107,19 @@ let replace metavar_tbl pattern_ast =
  * *)
 let find_remaining_metavars metavar_tbl ast =
   let seen_metavars = ref [] in
+  let str_metavars_regexp =
+    lazy
+      ((* List of metavars, quoted so that they can be used safely in a regex *)
```
emjin: Maybe clarify that this is a list of the metavars seen in the original code

emjin (approved):
Nice! There might be some other cases we're not covering (maybe import definitions? Regexes?). I don't have an example in mind, though, so I'm happy to handle them when they come up

nmote:
Could be! Let me know if you think of any. We don't have very good test coverage on autofix...

Issue #6247, opened 2022-10-05
Semgrep tries to parse .elm files even though it doesn't support Elm


ievans:
Split out from https://github.com/returntocorp/semgrep/issues/6199

> Having said that, we have a small portion of our codebase in Elm. When we run checks (e.g. semgrep scan --config auto ...) it, as expected, fails to parse the elm files with an error like: [ERROR] Fatal error at line myfile.elm:1: (Failure "not a program").

> The problem is that the exit code from semgrep changes depending on whether there is a subset of scanned files fail to parse but there are no findings (exit code 3) or all files fail to parse (exit code 2) and there are no findings. Some of our tooling rely on the exit code to determine what to do next, and these different behaviours impact our workflow. Even without support for Elm, we'd like to run semgrep to check files for things like secrets, but the error when parsing the files is preventing that.

My understanding is the expected behavior is that if we don't support elm, we don't try to parse elm file.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1961/semgrep-tries-to-parse-elm-files-even-though-it-doesnt-support-elm. _Note_: this link is for r2c use only and is not accessible publicly.

aryx:
seems related to https://github.com/returntocorp/semgrep/issues/6167

aryx:
This should be fixed by #6335

Issue #6167, opened 2022-09-23
semgrep ci fatal error


JannesSmitskamp-sfy:
A fatal error is triggered when running `semgrep ci`. The error condition seems to be triggered only when run with multiple rulesets and a sufficiently large source tree.

Running `semgrep scan` separately on current and baseline-commit versions works fine.

Working setup: semgrep 0.111.1 on python 3.10.6
Error setup: semgrep 0.112.1 on python 3.10.7
Error setup: semgrep 0.114.0 on python 3.10.7 and python 3.10.6

Error log:
```
Running `semgrep ci` without API token but with configs ('p/r2c-security-audit', 'r/kotlin.lang.security', 'r/java.lang.security', 'r/java.spring.security', 'r/contrib.owasp.java', 'r/generic')
Scan environment:
  versions - semgrep 0.114.0 on python 3.10.7
  environment - running in environment git, triggering event is unknown
Fetching configuration from semgrep.dev
Semgrep rule registry URL is https://semgrep.dev/registry.
Scanning across multiple languages:
    <multilang> | 100 rules × 302 files
           java | 185 rules ×  76 files
           json |   2 rules ×  41 files
         kotlin |  19 rules ×  15 files
           yaml |   3 rules ×   7 files
     dockerfile |   4 rules ×   1 file 
  Current version has 175 findings.
Switching repository to baseline commit '6b23b8b37c49b6547689495f0fdc017721f03fff'.
  Will report findings introduced by these commits:
    * [REDACTED COMMITD INFO]
Scanning across multiple languages:
    <multilang> |  3 rules × 13 files
           java | 10 rules × 10 files
[ERROR] Fatal error at line src/serviceTest/resources/test-local.properties:1:
 (Failure "not a program")
====[ BEGIN error trace ]====
(Failure "not a program")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Parse_tree_sitter_helpers.wrap_parser in file "src/parsing/tree_sitter/Parse_tree_sitter_helpers.ml", line 158, characters 13-29
Called from Parse_target.run_parser in file "src/parsing/Parse_target.ml", line 116, characters 18-24
Re-raised at Exception.reraise in file "src/pfff/commons/Exception.ml", line 20, characters 2-41
Called from Parse_target.parse_and_resolve_name in file "src/parsing/Parse_target.ml", line 396, characters 12-42
[...]

Issue #6249, opened 2022-10-05
Error while running on an Android project


twittemb:
**Describe the bug**

I have an issue while running Mobsfscan (which relies on semgrep) on an Android/Kotlin project with the following error.

Environment:

macOS Monterey
proc M1 Ultra
Python 3.8.9 (default, May 17 2022, 12:55:41) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
mobsfscan: v0.1.1

the stack trace seems to incriminate semgrep :-(

```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/xxx/Library/Python/3.8/lib/python/site-packages/libsast/common.py", line 45, in myrunner
    ret[0] = function(*args, **kwargs)
  File "/Users/xxx/Library/Python/3.8/lib/python/site-packages/libsast/core_sgrep/helpers.py", line 50, in invoke_semgrep
    ) = semgrep_main.main(
  File "/Users/xxx/Library/Python/3.8/lib/python/site-packages/semgrep/semgrep_main.py", line 363, in main
    rule_matches_by_rule, semgrep_errors, all_targets, profiling_data = run_rules(
  File "/Users/xxx/Library/Python/3.8/lib/python/site-packages/semgrep/semgrep_main.py", line 154, in run_rules
    ) = core_runner.invoke_semgrep(
  File "/Users/xxx/Library/Python/3.8/lib/python/site-packages/semgrep/core_runner.py", line 758, in invoke_semgrep
    ) = self._run_rules_direct_to_semgrep_core(
  File "/Users/xxx/Library/Python/3.8/lib/python/site-packages/semgrep/core_runner.py", line 703, in _run_rules_direct_to_semgrep_core
    returncode = runner.execute()
  File "/Users/xxx/Library/Python/3.8/lib/python/site-packages/semgrep/core_runner.py", line 283, in execute
    rc = asyncio.run(self._stream_subprocess())
[...]

PR #6335, opened 2022-10-18
change default of --scan-unknown-extensions to false!

aryx:
This is an important UX change.
This might close PA-1932, PA-1949, #6247, #6167

This has bitten us many times. In a pre-commit context,
pre-commit will pass an explicit list of files to semgrep
(the files modified in the PR), and this was causing semgrep
to try for example some Scala rules on some foo.h header files, which does not
make any sense.

This also happened in a semgrep ci context.

test plan:
```
(cli) [pad@thinkstation ~]$ cat targeting.yaml
rules:
- id: test
  patterns:
    - pattern: $FOO
  message: $FOO matched
  languages: [scala]
  severity: WARNING
(cli) [pad@thinkstation ~]$ cat /tmp/foo/foo.h
bla
bla
$XXX
foo ...

(cli) [pad@thinkstation ~]$ semgrep --config targeting.yaml /tmp/foo/foo.h
Nothing to scan.

Ran 1 rule on 0 files: 0 findings.
If Semgrep missed a finding, please send us feedback to let us know!
  $ semgrep shouldafound --help
```

This used to crash because of a parse error on $XXX, but it does not
make any sense to parse foo.h with a Scala parser ...

PR checklist:

### DIFF
--- a/changelog.d/pa-1932.changed
+++ b/changelog.d/pa-1932.changed
@@ -1,4 +1,4 @@
 The `--scan-unknown-extensions` option is now set to false by default.
 This means that from now on `--skip-unknown-extensions` is the default.
 This is an important change that prevents many errors when using
-semgrep in a pre-commit context or in CI.
+Semgrep in a pre-commit context or in CI.
--- a/changelog.d/pa-1932.changed
+++ b/changelog.d/pa-1932.changed
@@ -1,0 +1,4 @@
+The --scan-unknown-extensions option will be now set by default to false
+(which means --skip-unknown-extensions is now the default).
+This is an important change but this will avoid many errors when using
+semgrep in a pre-commit context or in CI.
--- a/changelog.d/pa-1932.changed
+++ b/changelog.d/pa-1932.changed
@@ -1,4 +1,4 @@
 The `--scan-unknown-extensions` option is now set to false by default.
 This means that from now on `--skip-unknown-extensions` is the default.
-This is an important change but this will avoid many errors when using
+This is an important change that prevents many errors when using
 semgrep in a pre-commit context or in CI.
--- a/changelog.d/pa-1932.changed
+++ b/changelog.d/pa-1932.changed
@@ -1,4 +1,4 @@
 The `--scan-unknown-extensions` option is now set to false by default.
-(which means --skip-unknown-extensions is now the default).
+This means that from now on `--skip-unknown-extensions` is the default.
 This is an important change but this will avoid many errors when using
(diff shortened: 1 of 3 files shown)


aryx:
This might also help #6249

changelog.d/pa-1932.changed
```diff
@@ -0,0 +1,4 @@
+The --scan-unknown-extensions option will be now set by default to false
```
adamkvitek:
```suggestion
The `--scan-unknown-extensions` option is now set to false by default.
```

changelog.d/pa-1932.changed
```diff
@@ -0,0 +1,4 @@
...
+(which means --skip-unknown-extensions is now the default).
```
adamkvitek:
```suggestion
This means that from now on `--skip-unknown-extensions` is the default.
```

changelog.d/pa-1932.changed
```diff
@@ -0,0 +1,4 @@
...
+This is an important change but this will avoid many errors when using
```
adamkvitek:
```suggestion
This is an important change that prevents many errors when using
```

changelog.d/pa-1932.changed
```diff
@@ -0,0 +1,4 @@
...
+semgrep in a pre-commit context or in CI.
```
adamkvitek:
```suggestion
Semgrep in a pre-commit context or in CI.
```

adamkvitek (commented):
@aryx I am leaving a few edits to the changelog, only small things. For some reason, I cannot make GitHub suggestions in this repository, so I am leaving them as comments.

aryx:
@adamkvitek I've added you with write access to this repo.

adamkvitek:
Thanks, @aryx. I changed my comments to the suggestions above, commit them at your convenience.

aryx:
Thx! committed.

aryx:
@adamkvitek this is a big change for the user, so we want to be loud about it when we make the next release (which should be today). It can break things.

emjin:
Did we also update `semgrep-docs`? I don't know if we had a note there before but if we didn't we should.

I also think we should include a warning if the user is only trying to scan one file and that file is skipped by `unknown-extension`. Or, we could make sure to include this in the "skipped files" output for a bit (`$NUM explicitly passed files were skipped due to unknown extension`). Even if we're loud, some people will likely miss the announcement.

aryx:
Good ideas!

PR #6336, opened 2022-10-18
tainting: Track taint coming from params' default values

### LINKED ISSUE #6298: Taint mode does not identify Python keyword arguments as a source
**Describe the bug**
In this example, https://semgrep.dev/s/4v4l, I want to match any instance of a string beginning with `http://` that enters the sinks specified. This works in the expected case, but does not work when the string is the default string of a keyword argument.

**To Reproduce**
https://semgrep.dev/s/4v4l

**Expected behavior**
I expect this rule to match on lines 28 and 38 as it's written.

**Screenshots**
image

**Environment**

**Use case**

Switching older "fake taint" rules to taint mode

IagoAbal:
Plus a small refactoring, to make -dfg_tainting use the same code as for running taint rules, otherwise results can differ sometimes.

Closes #6298
Fixes PA-1991

test plan:
make test # added one test

PR checklist:

### DIFF
--- a/semgrep-core/src/engine/Match_tainting_mode.ml
+++ b/semgrep-core/src/engine/Match_tainting_mode.ml
@@ -496,7 +496,6 @@ let check_fundef lang options taint_config opt_ent fdef =
     Some (D.str_of_name name)
   in
   let add_to_env env id ii pdefault =
-    pr2 (spf "add_to_env %s pdefault?=%b" (fst id) (Option.is_some pdefault));
     let var = AST_to_IL.var_of_id_info id ii in
     let source_pms =
       taint_config.D.is_source (G.Tk (snd id))
--- a/semgrep-core/src/engine/Match_tainting_mode.ml
+++ b/semgrep-core/src/engine/Match_tainting_mode.ml
@@ -495,14 +495,22 @@ let check_fundef lang options taint_config opt_ent fdef =
     let* name = AST_to_IL.name_of_entity ent in
     Some (D.str_of_name name)
   in
-  let add_to_env env id ii =
+  let add_to_env env id ii pdefault =
+    pr2 (spf "add_to_env %s pdefault?=%b" (fst id) (Option.is_some pdefault));
     let var = AST_to_IL.var_of_id_info id ii in
-    let taint =
+    let source_pms =
       taint_config.D.is_source (G.Tk (snd id))
+      @
+      match pdefault with
+      | Some e -> taint_config.D.is_source (G.E e)
+      | None -> []
+    in
+    let taints =
+      source_pms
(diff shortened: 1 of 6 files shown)


semgrep-core/src/engine/Match_tainting_mode.ml
```diff
@@ -495,22 +495,31 @@ let check_fundef lang options taint_config opt_ent fdef =
     let* name = AST_to_IL.name_of_entity ent in
     Some (D.str_of_name name)
   in
-  let add_to_env env id ii =
+  let add_to_env env id ii pdefault =
+    pr2 (spf "add_to_env %s pdefault?=%b" (fst id) (Option.is_some pdefault));
```
aryx: remaining pr2 ...
Maybe we need to add a semgrep rule to forbid them.

aryx: use logger#debug if you really need.

semgrep-core/src/engine/Match_tainting_mode.mli
```diff
@@ -20,6 +20,14 @@ val taint_config_of_rule :
   unit) ->
   Dataflow_tainting.config * debug_taint * Matching_explanation.t list
 
+val check_fundef :
```
aryx, line 23: Maybe a small comment explaining the function and its arguments.
We should try to have better documentation in our .mli.

aryx (approved):

semgrep-core/tests/rules/taint_param_default.yaml
```diff
@@ -0,0 +1,29 @@
+rules:
+  - id: request-session-with-http
+    options:
+      symbolic_propagation: true
+    mode: taint
+    pattern-sources:
+      - patterns:
+          - pattern: $URL
```
aryx, line 8: hmm should be "$URL"?
or maybe pattern: |
  "$URL"
(I hate yaml).

aryx (approved):

semgrep-core/src/engine/Match_tainting_mode.ml
IagoAbal: Ooops, sorry, that's meant to go away. Good catch.

semgrep-core/src/engine/Match_tainting_mode.mli
IagoAbal: True, thanks for calling me out here.

semgrep-core/tests/rules/taint_param_default.yaml
IagoAbal: Good point, I guess they may be using `metavariable-pattern` instead of `metavariable-regex` because then they can use `pattern-not-regex` ? cc @minusworld

PR #6337, opened 2022-10-18
feat(autofix): Replace text metavariables in AST-based autofix

### LINKED ISSUE #3648: autofix mishandles embedded quotes in string replacements for Python
Given a rule like this:
```yaml
rules:
- pattern: foo("$S")
  fix: bar("$S")
  languages: [python]
```

autofix will replace:
```python
foo('hello "friend"')
```

with
```python
bar("hello "friend"")
```

which is a syntax error.

**Expected behavior**
Should get a properly quoted string, either:
```python
bar('hello "friend"')
```

or
```python
bar("hello \"friend\"")
```

- P1: important to fix or quite annoying

**Environment**
Latest Semgrep (0.59 at time of writing)

Example in playground: https://semgrep.dev/s/spookylukey:string-quoting-bug

nmote:
This handles autofix for metavars that are bound to text, e.g. `$X` in `foo("$X")`. There are a few interesting implications of how I have chosen to handle this:

- This will lift the original text for the entire string, including quotes, from the target file. I think this is the right behavior. Textual autofix uses the quotes from the fix pattern, and inserts the original string contents inside of them. This leads to issues like #3648. My intuition is that people would expect Semgrep to leave these strings alone, and rarely care about maintaining the quotes that they use in the fix pattern. This behavior is also the simplest to implement.
- This doesn't work with f strings. f strings only match in this case when constant propagation evaluates them to a constant string. The behavior of AST-based autofix with f strings is to replace the f string with part of the contents that it evaluates to, without quotes. In many cases, this will lead to a syntactically-invalid program that will fail the validation step at the end, but in some cases it may slip through and change the meaning of the program. For now, I think this is acceptable because textual autofix works with f strings in almost no cases either, and often results in mangled code.
- Lastly, this will not yet apply autofix in cases where the fix is `bar("old string: $X")`. The main complication with handling that properly is that we will have to escape the string contents since we could be mixing contents from one type of quoted string with that of another.

This also comes with two other necessary changes which would have been difficult to test independently:

- Includes the original token in the `Metavariable.Text` variant. Previously, we stripped the quotes and adjusted the location accordingly. This facilitates metavariable pattern matches down the line. However, it meant that the quotes were excluded when we lifted the original text for that node during printing. To solve this, we just use the original token when creating the `String` AST node during metavariable replacement.
- Adds a `klit` hook into the AST mapper.

Fixes #3648

Test plan: Automated tests

PR checklist:


semgrep-core/src/core/Metavariable.ml
```diff
@@ -75,7 +75,10 @@ type mvalue =
    * where strings are often used to represent entities (e.g., function
    * names).
    *)
-  | Text of string AST_generic.wrap
+  | Text of
```
aryx, line 81: Maybe a comment saying why we keep the original token with enclosing quotes.
I guess the reasoning is that when people write "$X", they don't care whether in the target
code they used 'foo' or "foo", and if we use an autofix with "$X" we want to reput the
exact same string ('foo' or "foo"). The alternative would require a way to match the specific
quote used for the string but this seems heavy.

Not sure though how it will solve autofix like stuff("bar $X").

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
```diff
@@ -122,6 +146,8 @@ let find_remaining_metavars metavar_tbl ast =
             (fun (k, _) lit ->
               (match lit with
               | String (str, _) ->
+                  (* TODO handle the case where the metavar appears within the
+                   * string but is not the entire contents of the string *)
```
aryx: exactly ... not sure how the originfo will help you here.

aryx (approved):

semgrep-core/src/core/Metavariable.ml
nmote: For something like `stuff("bar $X")` we'll have to synthesize a new string, escape the text, and print it out without lifting any original text. So, this doesn't help, but both of these ranges will be wrong and irrelevant since it will be a new synthetic AST node. I will add a comment though.

semgrep-core/src/fixing/Autofix_metavar_replacement.ml
nmote: It won't, see above.

PR #6340, opened 2022-10-18
Update the Dockerfile grammar to reflect changes in generated code.

mjambon:
I was expecting changes due to https://github.com/returntocorp/ocaml-tree-sitter-core/pull/52 but it turns out the only necessary changes were from previous changes in ocaml-tree-sitter-core.

PR checklist:


nmote (approved):

aryx (approved):
does not look clearer than before but whatever.

mjambon:
> does not look clearer than before but whatever.

It's because the tokens are inlined in `Boilerplate.ml` together with a comment that tells you what pattern they are, so the type aliases aren't that useful after all. I forgot about that.

PR #6345, opened 2022-10-19
Add a plugin mechanism for registering proprietary parsers

mjambon:
This adds support for registering an Apex parser when using semgrep-core as a library in semgrep-proprietary.

Here's what we get right now if we try to parse an Apex file with semgrep-core:
```
 semgrep-core -lang apex -e hello <(echo 'hello')
Uncaught exception:
  
  invalid rule no-id, unknown location (FakeTokStr): Invalid pattern for Apex: Plugin.Missing_plugin("Missing Semgrep extension needed for parsing Apex target.")
----- pattern -----
hello
----- end pattern -----


Raised at Run_semgrep.parse_pattern in file "src/runner/Run_semgrep.ml", line 346, characters 6-220
Called from Run_semgrep.pattern_of_config in file "src/runner/Run_semgrep.ml", line 877, characters 26-46
Called from Run_semgrep.semgrep_with_one_pattern in file "src/runner/Run_semgrep.ml", line 892, characters 32-61
Called from Stdlib__Fun.protect in file "fun.ml", line 33, characters 8-15
Re-raised at Stdlib__Fun.protect in file "fun.ml", line 38, characters 6-52
Called from Common.main_boilerplate.(fun) in file "src/pfff/commons/Common.ml", line 1667, characters 12-16
Called from Common.pp_do_in_zero_box in file "src/pfff/commons/Common.ml", line 1633, characters 45-49
Called from Stdlib__Fun.protect in file "fun.ml", line 33, characters 8-15
Re-raised at Stdlib__Fun.protect in file "fun.ml", line 38, characters 6-52
Called from Common.exn_to_real_unixexit in file "src/pfff/commons/Common.ml", line 1630, characters 6-10
Called from Dune__exe__Main in file "src/main/Main.ml", line 59, characters 9-31
```

PR checklist:


semgrep-core/src/parsing/Plugin.mli
```diff
@@ -0,0 +1,28 @@
+(*
```
aryx: maybe rename to Parsing_plugin.ml?

aryx (approved):
Looks good!

aryx (approved):
I like it

nmote (approved):
Nice

semgrep-core/src/parsing/Plugin.ml
```diff
@@ -0,0 +1,65 @@
...
+    | None -> parsers := Some (parse_pattern, parse_target)
+    | Some _existing_parsers ->
+        (* This is a bug *)
+        let msg =
+          spf
+            "Plugin initialization error: a %s parser is being registered \
+             twice."
+            (Lang.to_string lang)
+        in
+        failwith msg
```
emjin: What happens if we hit this failwith? Does it get wrapped as a ParseError or FatalError? This seems like a fairly easy error to cause by accident and not get caught by tests. I'm worried it could cause Semgrep to fail with the `0 bytes out of 2 expected` error, which is really annoying to debug.

(Of course we should also aim to handle fatal errors better but I think it's not bad to think through the error case here too)

semgrep-core/src/parsing/Plugin.mli
mjambon: good call

emjin (commented):
Generally I like the approach! I'm not sure how I feel about the behavior that if we register two parsers, we fail, especially because this function mutates state.

semgrep-core/src/parsing/Plugin.ml
emjin: This sent really late 🤔. I think the comment still stands but on a different line

PR #6349, opened 2022-10-19
ci: Fix autofix-printing-stats job

nmote:
The `returntocorp/semgrep-dev:develop` image doesn't have OCaml, so instead we use an OCaml image and install the production version of Semgrep.

I also added a trigger for this job on branch push, which allowed for testing.

I had to do a bit of a hack (documented inline) to get access to the `semgrep-core` binary, but I think it should be alright.

Test plan: https://app.circleci.com/pipelines/github/returntocorp/semgrep/14680/workflows/75c67aa7-6871-40f4-9d1e-6991a24d8c64/jobs/18758

PR checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,15 +71,29 @@ jobs:
   # Run parsing stats and publish them to the semgrep dashboard.
   autofix-printing-stats:
     docker:
-      - image: returntocorp/semgrep-dev:develop
+      # The returntocorp/semgrep-dev:develop image doesn't have OCaml, so
+      # instead we use an OCaml image and install the production version of
+      # Semgrep.
+      #
+      # TODO build a docker image with both OCaml and semgrep-core available.
+      # Perhaps we can adapt our current Dockerfiles to publish an appropriate
+      # image here?
+      - image: ocaml/opam:ubuntu
     steps:
       - checkout
       - run:
           name: autofix printing stats
           no_output_timeout: 60m
           command: |
-            cd autofix-printing-stats
-            ./run --upload
+            sudo apt-get update
+            sudo apt-get install -y python3-pip jq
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -71,15 +71,22 @@ jobs:
   # Run parsing stats and publish them to the semgrep dashboard.
   autofix-printing-stats:
     docker:
-      - image: returntocorp/semgrep-dev:develop
+      - image: ocaml/opam:ubuntu
```
aryx, line 81: you can add the comment of your PR here (why we don't use returntocorp/ocaml or returntocorp/semgrep-dev
docker images).

aryx (approved):

mjambon (approved):

.circleci/config.yml
```diff
@@ -71,15 +71,22 @@ jobs:
...
-            ./run --upload
+            sudo apt-get update
+            sudo apt-get install -y python3-pip jq
+            pip3 install semgrep
+            # Put semgrep-core on the path. This is a little bit shady but
+            # apparently the `pip show` output is meant to be machine-parseable
+            # so this output should be stable:
+            # https://github.com/pypa/pip/issues/5261
+            PATH=$PATH:"$(pip3 show semgrep | grep '^Location' | sed -e 's/Location: //')/semgrep/bin"
+            opam exec -- ./autofix-printing-stats/run --upload
```
mjambon, line 96: I'd rather publish a docker image corresponding to the end of the first stage for our multistage build, like I suggested in Slack.

nmote: I agree that that would be a better long-term fix. I'll add a TODO here for that, but I'd prefer not to get derailed with that task right now, especially since I am not very familiar with Docker.

PR #6350, opened 2022-10-19
feat(autofix): Handle Id metavars in metavar replacement

nmote:
Some identifiers were already handled since they end up getting bound to expression metavars. But, for those that are bound as identifier metavars, like the ones in the tests, autofix will now correctly replace them.

Test plan: Automated tests

PR checklist:


nmote:
The e2e test changes are because that case is now using AST-based autofix, which conservatively inserts parentheses around binary expressions.

aryx:
Is there a way to implement the minimum parenthesis so we avoid those extra parens to avoid updating all those snapshots?

aryx (approved):

aryx:
or maybe just hardcode something for the ugly python printer that recognize == and avoid the parenthesis around?

nmote:
Yes, there's a way but I explicitly laid it out as out of scope for this month's work. Would you like me to re-evaluate that?

aryx:
I would hack something just for $X == $Y that would be enough for all those snapshots.

aryx:
But up to you.

nmote:
Sure, I'll look into it

nmote:
Actually, I don't think that we can safely leave out the parentheses for `==`. Consider the expression:

`2 + 3 == 4 + 1`
vs
`2 + (3 == 4) + 1`

The latter is kind of nonsense, but Python does happily evaluate it to `3`. If we are printing the latter expression and leave off the parentheses, we change the meaning of the program.

nmote:
I think that to address this, we will need to take into account the context of the expression. This is something that I intend to do, but not as part of this month's project which is focused on getting correct (if ugly) output in as many cases as possible.

PR #6352, opened 2022-10-19
feat!: allow metavariable analysis of terraform propagated values

brandonspark:
## What:
We have already solved the issue of Terraform constant propagation. Those propagated values are stored within the `id_svalue` of the identifier, which is in the form of a `DotAccess`, and are used in matching.

However, metavariable comparison relies on only using propagated values in the form of an `N`. Terraform identifiers don't look like that, so we need to special-case that.

## Why:
We may want to do metavariable comparison on a Terraform identifier which has been propagated to. Consider this example:
```
resource "x" "y" {
  name = "z"
  foo = local.num
}
```
with another file
```
locals {
  num = 364
}
```

If we are interested in whether `foo` is <= 150, then we might write a rule like this:
```

rules:
  - id: retention-not-compliant 
    options:
      symbolic_propagation: true
    languages:
      - hcl
    message: >-
      We want this value <= 150!
    patterns:
      - pattern: |
          resource "x" "..." {... foo = $NUM ...}
      - metavariable-comparison: 
          metavariable: $NUM
          comparison: |
            $NUM <= 150
    severity: WARNING
```

However, this will not work, because `Eval_generic` will not know how to get the propagated value out of the metavariable.

## How:
I added a special case in the `eval` function. Last one, I promise.

## Test plan:
I can't add a test here, but trust that you need this for the DeepSemgrep thing to work.

PR checklist:


semgrep-core/src/engine/Eval_generic.ml
```diff
@@ -219,6 +219,17 @@ let rec eval env code =
...
+  (* TERRAFORM:
+     The propagated value is stored in the `id_svalue` of the DotAccess, for Terraform.
+     We need to add this case here, or else we can't do metavariable analysis on an
+     identifier which contains a propagated value, because identifiers look like
+     DotAccesses in Terraform.
+  *)
+  | G.DotAccess
+      ( { e = G.N (Id ((("local" | "var"), _), _)); _ },
+        _,
+        FN (Id (_, { id_svalue = { contents = Some (Lit lit); _ }; _ })) ) ->
```
IagoAbal: There are already two cases that handle const-prop values, the normal case and the Dockerfile case, and I think you could add this case there too. Also this lacks `when env.constant_propagation ->`, although if you moved with the other two cases that would solve it already.

IagoAbal:
> I can't add a test here, but trust that you need this for the DeepSemgrep thing to work.

Are you sure you can't add a test case here? You can put a constant in the same Terraform file and we have intra-file const-prop for Terraform already I believe.

If I'm correct here, we need a changelog entry too. We may still want the changelog entry if this was just relevant for DeepSemgrep... I'm not sure how we're managing this right now.

brandonspark:
> > I can't add a test here, but trust that you need this for the DeepSemgrep thing to work.
> 
> Are you sure you can't add a test case here? You can put a constant in the same Terraform file and we have intra-file const-prop for Terraform already I believe.
> 
> If I'm correct here, we need a changelog entry too. We may still want the changelog entry if this was just relevant for DeepSemgrep... I'm not sure how we're managing this right now.

Yeah, fair enough, I missed that. Test added!

IagoAbal (approved):
Nice!

PR #6354, opened 2022-10-19
feat(autofix): AST-based autofix for JS

nmote:
All of the printing I have implemented so far for Python is common to many languages, including JS. So, I introduced a `common_printer` class. This is separate from `base_printer` as explained in the comment.

Then, I used that in both the Python printer and JS printer. The rest is just updating a bit of boilerplate to include JS and adding tests.

Test plan: Automated tests. Also, run `autofix-printing-stats/run` and verify that the Python output is unchanged, and JS reports 0.0% (0/1) for the printing stats (there is only one JS fix in semgrep-rules, and it is a statement which is currently out of scope for this project).

PR checklist:


aryx:
wow, interesting to know we have no fix rules for JS.
Once we get the autofix stats in the dashboard DB, we could add some graphs to metabase here: https://metabase.corp.r2c.dev/collection/59-semgrep to track the # rules per lang with fix, and of course the current autofix stats for AST-based autofix coverage.

semgrep-core/src/core/ast/Ugly_print_AST.ml
```diff
@@ -197,41 +197,65 @@ class base_printer : printer_t =
...
+(* Printing for constructs that are identical across many languages. Unlike
+ * base_printer, it's not expected that every printer inherit from this. But
+ * many printers should. This can be split up as needed so that printers can use
+ * pieces a la carte, but care should be taken to avoid making the inheritance
+ * graph overly complicated.
+ *
+ * This is virtual rather than a subclass of base_printer so that language
+ * printers can pull pieces from this and other classes as needed to concisely
+ * build up a complete printer. *)
+class virtual common_printer =
```
aryx, line 209: maybe at some point we can rename common_printer to c_like_printer, or classic_parencalls_printer,
or something more descriptive than common.

aryx (approved):

semgrep-core/src/core/ast/Ugly_print_AST.ml
nmote: Yes! I intend to do this. I tried to come up with a more descriptive name right now, but I wasn't sure what this would turn in to as we added more printing and more languages. I have a feeling that the boundaries will become clearer as we add printing.

nmote:
There is one autofix rule for JS, but it's for a statement so that's out of scope for this month. I was also surprised that there were so few, especially since Python had a handful and JS is more popular.

PR #6356, opened 2022-10-19
fixed: taint labels not being applied to `Dot` offsets

### LINKED ISSUE #6355: Labels not being applied correctly to taint rule
**Describe the bug**
We are producing a finding when we shouldn't be, because a label being used as a "requires" should be preventing the finding, but it is not.

**To Reproduce**
Steps to reproduce the behavior, ideally a link to https://semgrep.dev: https://semgrep.dev/playground/s/JPRQ

**Expected behavior**
There should be no finding, because the `USERINPUT` label is not satisfied by the source.

**Screenshots**

**Environment**

**Use case**
Correctness.

brandonspark:
## What:
Currently, we are producing findings that should be made impossible due to taint labels. These specifically only occur in the case of them resulting from a `Dot` / `DotAccess`.

## Why:
This is a regular bug that should be fixed. In particular, it was causing erroneous findings on these cases:
https://semgrep.dev/playground/s/JPRQ
https://semgrep.dev/playground/s/RW7N

## How:
We weren't filtering for labels when producing taint from a tainted token, which the `Dot` case was degenerating to. I just added in a filtering step.

Fixes #6355 

PR checklist:

### DIFF
--- a/changelog.d/gh-6355.fixed
+++ b/changelog.d/gh-6355.fixed
@@ -1,1 +1,1 @@
-Taint: Fixed a bug where taint labels were not being correctly applied to findings resulting from dot accesses. For instance, a taint pattern with a `requires` involving `x.y` might ignore the `requires` if the `y` produces a match.
+taint-mode: Fixed a bug in the experimental taint-labels feature that caused labels to be incorrectly applied to dot accesses. For instance, if a pattern-source that requires label A and adds label B matches a dot-access expression like x.a, the field a will get the label B even if it does not carry label A as required.
--- a/changelog.d/gh-6355.fixed
+++ b/changelog.d/gh-6355.fixed
@@ -1,0 +1,1 @@
+Taint: Fixed a bug where taint labels were not being correctly applied to findings resulting from dot accesses. For instance, a taint pattern with a `requires` involving `x.y` might ignore the `requires` if the `y` produces a match.
--- a/semgrep-core/src/tainting/Dataflow_tainting.ml
+++ b/semgrep-core/src/tainting/Dataflow_tainting.ml
@@ -409,9 +409,10 @@ let check_tainted_tok env tok =
   | _ :: _ -> (Taints.empty, env.lval_env)
   | [] ->
       let taints = source_pms |> taints_of_matches in
-      (* Empty because we have no base, but we want to filter the taints we got from
-         the `source_pms`.
-         Otherwise, we might let taints from this token escape a label.
+      (* Empty because we want to filter all the previous taints we got from the
+         `source_pms`, but we have nothing to really union it with.
+         Otherwise, we might let taints from this token escape a label. In particular,
+         this may happen in the `Dot` case.
       *)
       let taints = Taints.empty |> union_taints_filtering_labels ~new_:taints in
       let sinks = sink_pms |> Common.map trace_of_match in
--- a/semgrep-core/src/tainting/Dataflow_tainting.ml
+++ b/semgrep-core/src/tainting/Dataflow_tainting.ml
@@ -409,6 +409,11 @@ let check_tainted_tok env tok =
   | _ :: _ -> (Taints.empty, env.lval_env)
   | [] ->
       let taints = source_pms |> taints_of_matches in
+      (* Empty because we have no base, but we want to filter the taints we got from
+         the `source_pms`.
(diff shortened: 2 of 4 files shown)


semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -409,6 +409,11 @@ let check_tainted_tok env tok =
   | _ :: _ -> (Taints.empty, env.lval_env)
   | [] ->
       let taints = source_pms |> taints_of_matches in
+      (* Empty because we have no base, but we want to filter the taints we got from
```
emjin: Totally fair if this comment just needs a lot of context, but from this snippet it's hard to understand. Main sources of confusion are probably "because we have no base"

emjin (commented):
Can you add a test case?

semgrep-core/src/tainting/Dataflow_tainting.ml
emjin: > a tainted token, which the Dot case was degenerating to

This from the PR is helpful, could use it in the code

changelog.d/gh-6355.fixed
```diff
@@ -0,0 +1 @@
+Taint: Fixed a bug where taint labels were not being correctly applied to findings resulting from dot accesses. For instance, a taint pattern with a `requires` involving `x.y` might ignore the `requires` if the `y` produces a match.
```
IagoAbal: Even if just to make the review easier, we should hard wrap these files at 80 chars. It's painful having to scroll horizontally so much. Otherwise I need to open the file "raw" so that I can get soft wrapping by the browser.

IagoAbal:
We need to add at least one regression test to this PR, to make sure we detect the problem the next time.

changelog.d/gh-6355.fixed
IagoAbal: > Taint: Fixed a bug where taint labels were not being correctly applied to findings resulting from dot accesses. For instance, a taint pattern with a `requires` involving `x.y` might ignore the `requires` if the `y` produces a match.

Maybe

> taint-mode: Fixed a bug in the experimental taint-labels feature that caused labels to be incorrectly applied to dot accesses. For instance, if a `pattern-source` that requires label `A` and adds label `B` matches a dot-access expression like `x.a`, the field `a` will get the label `B` even if it does not carry label `A` as required.

? 

cc @adamkvitek may give a more objective opinion.

IagoAbal:
Also, what GitHub issue is this fixing? I saw a bug report filed by yourself that seems to be fixed by this PR.

enncoded:
> Also, what GitHub issue is this fixing? I saw a bug report filed by yourself that seems to be fixed by this PR.

@IagoAbal It's fixing https://github.com/returntocorp/semgrep/issues/6355, which I reported in the PA channel in slack, but Brandon filed a GH issue for it before I finished doing so. I left some details about its importance to me in the Linear issue.

semgrep-core/tests/tainting_rules/go/token_labels.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
+  - id: go-token-labels
+    message: A taint label did not prevent the match coming from a tainted token!
```
emjin, line 3: I got confused so maybe specify that the point is this is an unsatisfiable label and we're testing that the label works. Having the issue helps though

emjin (approved):
I'm not personally a huge fan of the fact that we have to copy this filtering step in four places, but you're just following previous convention so I'm fine with this PR.

semgrep-core/tests/tainting_rules/go/token_labels.go
```diff
@@ -0,0 +1,8 @@
+// See https://github.com/returntocorp/semgrep/issues/6355 if this breaks.
```
IagoAbal, line 1: The `tainting_rules` folder comes from a time when you couldn't pass taint rules to `-rules`, you had to use a different switch `-tainting_rules` or something. Nowadays I just add these tests in `tests/rules` but prefixed with `taint_`. I'll see if I can move all tests soon.

IagoAbal (approved):
Nice! Let's merge it. In #6279 I actually removed `check_tainted_tok` anyways, but in the meantime we fix the bug and it's also useful to get the new test case.

PR #6357, opened 2022-10-19
fix(js): parse specials consistently

brandonspark:
## What:
Special identifiers are parsed inconsistently between the pattern and target in JS.

## Why:
Ty brought up this example in the security research channel: https://semgrep.dev/s/BK7R

Here, `module` is parsed differently in the pattern and AST. We see this via:
```
brandonspark@MacBook-Pro ~/playground> sc -dump_pattern test.js -lang js
E(
  Assign(
    DotAccess(OtherExpr(("Module", ()), []), (),
      FN(
        Id(("exports", ()),
          {id_info_id=1; id_hidden=false; id_resolved=Ref(None);
           id_type=Ref(None); id_svalue=Ref(None); }))), (),
    N(
      Id(("$DATA", ()),
        {id_info_id=2; id_hidden=false; id_resolved=Ref(None);
         id_type=Ref(None); id_svalue=Ref(None); }))))
brandonspark@MacBook-Pro ~/playground> sc -dump_ast test.js -lang js
Pr(
  [ExprStmt(
     Assign(
       DotAccess(
         N(
           Id(("module", ()),
             {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); })), (),
         FN(
           Id(("exports", ()),
             {id_info_id=2; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))), (),
       N(
         Id(("$DATA", ()),
           {id_info_id=3; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }))), ())])
```

## How:
I changed an `idexp` to an `idexp_or_special` in the corresponding place. We use the Treesitter and pfff parsers for both pattern and target, but in different orders. This means that both parsers need to be kept closely in sync, or they will produce differing trees for both.

I'm unfamiliar with Javascript. I see a whole bunch of `idexp`s in `Parse_typescript_tree_sitter.ml`, but I don't know which ones are safe to change to `idexp_or_special`. I will call on a Javascript expert -- @nmote ?

Also, why is it that we even use both parsers in different orders? I don't remember the reason, but would it be feasible to change one of them to be pfff-before-tree-sitter or tree-sitter-before-pfff? That would make bugs like this impossible.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/changelog.d/pa-2030.fixed
+++ b/changelog.d/pa-2030.fixed
@@ -1,0 +1,1 @@
+JS/TS: Fixed a parsing bug where special identifiers were parsed differently in patterns
--- a/semgrep-core/src/parsing/tree_sitter/Parse_typescript_tree_sitter.ml
+++ b/semgrep-core/src/parsing/tree_sitter/Parse_typescript_tree_sitter.ml
@@ -1175,7 +1175,7 @@ and primary_expression (env : env) (x : CST.primary_expression) : expr =
           | `Member_exp x -> member_expression env x
           | `Paren_exp x -> parenthesized_expression env x
           | `Choice_unde x -> identifier_ env x
-          | `Choice_decl x -> reserved_identifier env x |> idexp
+          | `Choice_decl x -> reserved_identifier env x |> idexp_or_special
           | `This tok -> this env tok (* "this" *)
           | `Super tok -> super env tok (* "super" *)
           | `Num tok ->


nmote (approved):
We used to use pfff to parse JS patterns, and both parsers to parse JS targets, but for targets we tried tree sitter first.

I introduced the fallback logic for patterns, but to minimize the size of the change, and because I didn't implement all the pattern features in tree sitter, I put tree sitter second. At some point we should make this more consistent, for sure. maybe even drop pfff altogether. I think it requires more work supporting Semgrep pattern features in the tree sitter parser, though.

In the meantime, we should strive to have the two parsers produce the same result on the same input. Usually we change the tree sitter output to match pfff like you've done here.

I'm also not sure which ones are safe to change to `idexp_or_special`. I think if we were to do this, we'd want to look through systematically at all the places where the pfff parser calls that function, and make sure we are calling it at the equivalent point in the tree sitter conversion. Or, we could just wait until issues crop up.

Or, better yet, we could migrate off of pfff altogether. That might be a bit of an effort, though.

Anyway, this PR looks good. You're just changing the tree sitter parser to mimic the behavior of the pfff parser, which is uncontroversial. A test would be nice, but I won't be a stickler for it here.

PR #6362, opened 2022-10-20
osemgrep: porting env.py

aryx:
test plan:
make test

PR checklist:


semgrep-core/src/osemgrep/TOPORT/env.py
```diff
@@ -1,134 +0,0 @@
-from semgrep.constants import SETTINGS_FILENAME
```
aryx, line 1: - no need for singletons and Factory, and click state and get_state(), and
   @action.default, and methods (see env.py and state.py)
   => just use let which can access environment vars at run-time (no single main in ocaml,
   each module has a "startup")

PR #6363, opened 2022-10-20
fix(autofix): Implement print-avoidance for bare identifier metavariables

nmote:
Previously, we would only lift the original text for identifiers if they were within a `N (Id _)` expression. This is because `Metavariable.mvalue_to_any`  wraps identifier metavariables with expressions.

Now, for the purposes of autofix print avoidance, `Metavariable.ID` values are converted to `AST_generic.I` nodes before they go into the print avoidance table. Additionally, the printer has been modified with a `print_ident` method to facilitate printing at that level of granularity.

I also implemented DotAccess printing for Python so that the test cases that exercise this can pass.

Test plan: Automated tests

PR checklist:


aryx (approved):

nmote:
Re-ran the benchmark and it passed

PR #6364, opened 2022-10-20
Add submodule for semgrep-apex

mjambon:
Note that it's only used in conjunction with semgrep-proprietary. It would be best to keep it there but then we'd have to edit the package name `tree-sitter-lang` in the submodule's dune file which is generated upstream by ocaml-tree-sitter.

Pros:
* simplicity

Cons:
* makes semgrep-core larger than necessary (unless dead code is automatically removed by the compiler/linker, which I doubt)

In the future, we might want to add an option to ocaml-tree-sitter so as to generate a suitable dune file, or some equivalent mechanism such as a dune template that can be instantiated after code generation.

PR checklist:


aryx (approved):

aryx:
I like simplicity.

PR #6381, opened 2022-10-24
fix(autofix): Handle focus-metavariable matches in AST-based autofix

nmote:
At the point where we try to render an autofix, the rule that we have when there is a focus-metavariable is the one that has been created by `Match_env.fake_rule_id` called from
`Match_search_mode.apply_focus_on_ranges`. This fake rule ID is later adjusted in `Range_with_metavars.range_to_pattern_match_adjusted`.

I've updated that later adjustment to copy over the fix and the languages, which are both needed for autofix.

Previously, the fix and languages weren't copied over either in the place where the rule ID was created, or when adjusted later. I'm not sure of the history or the broader architecture here, so I'm not sure why we create a rule ID with dummy data and then update it later, rather than just using the correct info to begin with. But, I think that my changes fit in with the existing design. It might be time to update some names, but I don't think I have the context to suggest improvements.

Test plan:

`./autofix-printing-stats/run` --> 96.4% (27/28) for Python, up from 67.9% (19/28)

Manually run the CLI against the python/distributed/security.yaml rule in semgrep-rules, and verify that it performs the fix correctly.

PR checklist:

### DIFF
--- a/semgrep-core/src/core/Xlang.mli
+++ b/semgrep-core/src/core/Xlang.mli
@@ -21,6 +21,9 @@ val of_lang : Lang.t -> t
 (* raises an exception with error message *)
 val to_lang : t -> Lang.t
 
+(* Does not raise, but returns empty list for all but the L variant *)
+val to_langs : t -> Lang.t list
+
 (* raises an exception with error message *)
 val lang_of_opt_xlang : t option -> Lang.t
 
--- a/semgrep-core/src/engine/Range_with_metavars.ml
+++ b/semgrep-core/src/engine/Range_with_metavars.ml
@@ -37,11 +37,14 @@ let (range_to_pattern_match_adjusted : Rule.t -> t -> Pattern_match.t) =
  fun r range ->
   let m = range.origin in
   let rule_id = m.rule_id in
+  let languages = Xlang.to_langs r.Rule.languages in
   (* adjust the rule id *)
   let rule_id =
     {
       rule_id with
       Pattern_match.id = fst r.Rule.id;
+      fix = r.Rule.fix;
+      languages;
       message =
         r.Rule.message (* keep pattern_str which can be useful to debug *);
     }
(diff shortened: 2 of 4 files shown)


semgrep-core/src/core/Xlang.mli
```diff
@@ -21,6 +21,9 @@ val of_lang : Lang.t -> t
 (* raises an exception with error message *)
 val to_lang : t -> Lang.t
 
+(* Does not raise, but returns empty list for all but the L variant *)
+val to_langs : t -> Lang.t list
```
emjin, line 25: One idea: `to_lang_exn` for the previous?

semgrep-core/src/engine/Range_with_metavars.ml
```diff
@@ -37,11 +37,14 @@ let (range_to_pattern_match_adjusted : Rule.t -> t -> Pattern_match.t) =
  fun r range ->
   let m = range.origin in
   let rule_id = m.rule_id in
+  let languages = Xlang.to_langs r.Rule.languages in
   (* adjust the rule id *)
```
emjin, line 41: Yeah, we might want to change the name `rule_id` at this point. Maybe `rule_info`?

emjin (approved):
Yup, I think this is the right thing to do. Left some suggestions for names but imo you don't have to address them to merge this

semgrep-core/src/core/Xlang.mli
nmote: Good idea. I'll put it in a followup PR since it's mostly unrelated to this one, though.

semgrep-core/src/engine/Range_with_metavars.ml
nmote: Maybe. I'm actually not sure why we don't just pass the whole rule around. When I last looked into this my guess was that it was initially actually just the rule ID, but then people (myself included) added more and more info that would be nice to have.

PR #6382, opened 2022-10-24
chore: Add --verbose flag to autofix printing stats script

nmote:
This will print a list of rule-target pairs that had at least one match where an autofix failed to render.

I also refactored how flags are handled as part of this.

Test plan:

Ran without any flags and it behaved as before. Ran with `--verbose` and got a list of failing test cases. Ran with `--verbose --upload` and got "Uploading statistics" at the beginning (commented out the actual upload so as not to pollute the metrics data).

PR checklist:

### DIFF
--- a/autofix-printing-stats/run
+++ b/autofix-printing-stats/run
@@ -10,6 +10,11 @@
 
 #load "str.cma"
 
+type opts = {
+  upload: bool;
+  verbose: bool;
+}
+
 let spf = Printf.sprintf
 
 (* Sets the current working directory to the location of this executable. May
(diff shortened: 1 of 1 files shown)


autofix-printing-stats/run
```diff
@@ -204,15 +212,18 @@ let upload_results should_upload results =
              ];
            print_newline ()))
 
+let make_opts () =
```
aryx, line 215: Using ppx_deriving_cmdliner like in hello_script.ml would remove the need for this code, but this is fine also like this.

aryx (approved):

autofix-printing-stats/run
nmote: For something this simple, it seemed easier to just roll my own rather than learn another tool and then make sure it's available in CI.

aryx: I agree

PR #6393, opened 2022-10-26
Add support for RUN options for dockerfile

### LINKED ISSUE #6353: I need a way to detect a special "mounted secrets" syntax in Dockerfiles
Dockerfiles have introduced a [special `--secrets` flag](https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information) which permits the use of secrets during the build process that are discarded after the build is complete. This is neat, since you can use secrets to affect the final container image but the image won't contain the secret.

The syntax used for this matches neither Dockerfile nor Bash conventions, and as such Semgrep doesn't quite work like I want it to.

```dockerfile
# syntax=docker/dockerfile:1.2

FROM alpine

# shows secret from default secret location:
RUN --mount=type=secret,id=mysecret cat /run/secrets/mysecret

# shows secret from custom secret location:
RUN --mount=type=secret,id=mysecret,dst=/foobar cat /foobar
```

I have this snippet here (https://semgrep.dev/s/qwwe) which sort of demonstrates what I want to match, but what I'd like to do is capture pieces of this special secrets syntax in metavariables for use later in the pattern. Here's an example snippet for what I'm trying to do: https://semgrep.dev/s/pdd0

```
rules:
  - id: accidentally-stored-secret-copy
    patterns:
      - pattern-inside: |
          RUN ... --mount=type=secret,id=$SECRET ...
      - pattern-either:
          - pattern: ... cat /run/secrets/$SECRET > ...
          - pattern: ... cp /run/secrets/$SECRET ...
          - pattern: ... mv /run/secrets/$SECRET ...
    message: Semgrep found a match $SECRET
    languages:
      - docker
    severity: WARNING
```

I've also tried [matching using regexes](https://semgrep.dev/s/X664) but there are syntax limitations with how metavariables can be reused (I don't think they "concat" to strings like I'm using them), nor does the special `$1` syntax appears to be recyclable in patterns.

Another user came up with [this](https://semgrep.dev/playground/s/RWWg), using only regexes, but that's error-prone and still doesn't account for all cases.

Another option tried is binding the Bash command (e.g., `cp whatever`) to a metavariable and using `metavariable-pattern` with the Bash language.

**Use case**
This will enable an actually-useful Docker security rule where the behavior of the secrets syntax is not well-understood.

**Additional context**
Myself and other users would like this rule! :D

mjambon:
Closes https://github.com/returntocorp/semgrep/issues/6353
Uses https://github.com/camdencheek/tree-sitter-dockerfile/pull/28

PR checklist:


aryx (approved):

fproulx-boostsecurity:
@mjambon looks like that still does not work with a line like (multiple `--mount=type` flags:
```
  RUN --mount=type=secret,id=GITHUB_TOKEN --mount=type=ssh set -ex \
   && poetry install --no-root --without dev
```

mjambon:
@fproulx-boostsecurity what pattern did you try? I added a note here: https://github.com/returntocorp/semgrep/issues/6353#issuecomment-1371579782

PR #6397, opened 2022-10-26
feat(deep)!: Make add/addAll calls from collections objs propagators

emjin:
For https://github.com/returntocorp/semgrep-proprietary/pull/322

Test plan: see tests in linked PR

PR checklist:

### DIFF
--- a/semgrep-core/src/analyzing/AST_to_IL.ml
+++ b/semgrep-core/src/analyzing/AST_to_IL.ml
@@ -225,6 +225,22 @@ let is_hcl lang =
   | Lang.Hcl -> true
   | _ -> false
 
+(* Helper to restrict built-in propagators to certain types. The goal is to
+   eventually have a default pattern-propagators for most propagators, which
+   users can configure, and only use built-in propagators if we perform a
+   special analysis for that class (e.g. key-sensitive analysis for HashMap) *)
+let is_collections_class t =
+  match t with
+  | G.TyN t
+  | TyApply ({ t = TyN t; _ }, _) -> (
+      match t with
+      | Id
+          ( (("List" | "ArrayList" | "Map" | "HashMap" | "Set" | "HashSet"), _),
+            _ ) ->
+          true
+      | _ -> false)
+  | __else__ -> false
+
 (*****************************************************************************)
 (* lvalue *)
 (*****************************************************************************)
@@ -540,6 +556,42 @@ and expr_aux env ?(void = false) e_gen =
       let special = (Eval, tok) in
       let args = arguments env args in
       add_instr env (mk_i (CallSpecial (Some lval, special, args)) eorig);
+      mk_e (Fetch lval) (related_tok tok)
+  | G.Call
+      (* This narrowly adds some calls from objects of certain classes as default
+         propagators. For now, this is a quick fix to make rule-writing easier.
+         TODO In the future, we should only use this for classes where we perform
+         a special analysis. Other propagators should be in a configurable defaults
+         file *)
+      ( {
+          e =
+            G.DotAccess
+              ( ({
+                   e =
+                     G.N
+                       (G.Id
+                         ((id, _), { id_type = { contents = Some { t; _ } }; _ }));
+                   _;
+                 } as obj),
+                _tok,
+                G.FN
+                  (G.Id
+                    ( ((("add" | "addAll") as kind), tok),
+                      { G.id_resolved = { contents = None }; _ } )) );
+          _;
+        },
+        args )
+    when is_collections_class t ->
+      let lval =
+        (* TODO: for cases like adding to a HashMap, we will
+           want to make this key-sensitive. We can do that
+           by adding the key as an offset to the lval *)
+        try lval env obj with
+        | Fixme _ -> fresh_lval ~str:id env tok
+      in
+      let special = (IL_helpers.call_assign_of_string kind, tok) in
+      let args = arguments env args in
+      add_instr env (mk_i (CallAssign (lval, special, args)) eorig);
       mk_e (Fetch lval) (related_tok tok)
   | G.Call
       ({ e = G.IdSpecial (G.InterpolatedElement, _); _ }, (_, [ G.Arg e ], _))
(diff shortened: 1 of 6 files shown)


emjin:
@IagoAbal Let me know what you think of this approach and then, if you like it, I'll add a changelog entry and a test. Alternatively we could gate this behavior so it doesn't affect Semgrep.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -225,6 +225,22 @@ let is_hcl lang =
...
+   users can configure, and only use built-in propagators if we perform a
+   special analysis for that class (e.g. key-sensitive analysis for HashMap) *)
+let is_collections_class t =
+  match t with
+  | G.TyN t
+  | TyApply ({ t = TyN t; _ }, _) -> (
+      match t with
+      | Id
+          ( (("List" | "ArrayList" | "Map" | "HashMap" | "Set" | "HashSet"), _),
+            _ ) ->
```
IagoAbal, line 239: Perhaps better to pass the `env` to this function and restrict this to Java. Just to be on the safe side. 🤔

IagoAbal:
Isn't it better to just reuse `pattern-propagators` and "add" some built-in ones to Java taint rules? That would be something you could easily do in DeepSemgrep so Semgrep wouldn't even be affected. Have you considered that? If so, why do you think that this approach is better?

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -541,6 +557,42 @@ and expr_aux env ?(void = false) e_gen =
...
+                 } as obj),
+                _tok,
+                G.FN
+                  (G.Id
+                    ( ((("add" | "addAll") as kind), tok),
+                      { G.id_resolved = { contents = None }; _ } )) );
+          _;
+        },
+        args )
+    when is_collections_class t ->
```
IagoAbal, line 584: Better to restrict this for Java here too, then we don't have unintended effects for other languages.

semgrep-core/src/core/il/IL.ml
```diff
@@ -250,10 +250,13 @@ and instr_kind =
   | Assign of lval * exp
   | AssignAnon of lval * anonymous_entity
   | Call of lval option * exp (* less: enforce lval? *) * exp argument list
+  | CallAssign of lval * call_assign wrap * exp argument list
```
IagoAbal, line 253: I would prefer not to add another `CallXyz` here, we can just use `CallSpecial` for this and extend `call_special`.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -541,6 +557,42 @@ and expr_aux env ?(void = false) e_gen =
...
+                   e =
+                     G.N
+                       (G.Id
+                         ((id, _), { id_type = { contents = Some { t; _ } }; _ }));
+                   _;
+                 } as obj),
+                _tok,
+                G.FN
+                  (G.Id
+                    ( ((("add" | "addAll") as kind), tok),
```
IagoAbal, line 579: Is the method called `add` or `addAll` in all cases? I think for `HashMap` it is called `put` for example.

semgrep-core/src/core/il/IL_helpers.ml
```diff
@@ -144,6 +146,30 @@ let rlvals_of_node = function
...
+
+let should_clean_lval x =
+  match x.i with
+  | Assign _
+  | AssignAnon _
+  | Call _
+  | CallSpecial _
+  | FixmeInstr _ ->
+      true
+  | CallAssign (_, ((Add | AddAll), _), _) ->
```
IagoAbal, line 164: Do you have bugs in the test suite that need this? If you do `x[i] = clean` we assume that `x` gets cleaned and, although I'm aware that it's not sound, it probably avoids FPs. Rather than making this ad-hoc for Java collections, I would prefer that we have a rule option that let's you choose what kind of assumptions to make. I just recently added `taint_assume_safe_functions` and `taint_assume_safe_indexes` in that spirit. (I know options aren't great but...). Then DeepSemgrep Java could start having different defaults than Semgrep.

semgrep-core/src/analyzing/AST_to_IL.ml
```diff
@@ -541,6 +557,42 @@ and expr_aux env ?(void = false) e_gen =
...
+                      { G.id_resolved = { contents = None }; _ } )) );
+          _;
+        },
+        args )
+    when is_collections_class t ->
+      let lval =
+        (* TODO: for cases like adding to a HashMap, we will
+           want to make this key-sensitive. We can do that
+           by adding the key as an offset to the lval *)
```
IagoAbal, line 588: I think that if we go this way with collections, I would rather translate them already into assignments to lval+offset. Then they would be treated like arrays, and we can configure via options how that should work. At some point, we can make them index-sensitive as you say here.

IagoAbal (changes requested):
I would prefer to either avoid touching Semgrep's code and just add pattern-propagators to DeepSemgrep's Java taint rules, or otherwise I would like to go "further" see collection operations transformed into lval+offset to be handled as if they were arrays (so we avoid CallAssign and all the special handling of these). In the latter case,  we can then tune how operations on lval+index are handled in DeepSemgrep.

emjin:
> Isn't it better to just reuse `pattern-propagators` and "add" some built-in ones to Java taint rules? That would be something you could easily do in DeepSemgrep so Semgrep wouldn't even be affected. Have you considered that? If so, why do you think that this approach is better?

I thought about that direction (see my comment https://github.com/returntocorp/semgrep-proprietary/pull/322), though I didn't think about just encoding it in the DeepSemgrep code. The reason I chose this approach is because I'm confident we'll eventually have to do some special handling of collections. I also would actually rather handle them all as their own thing, rather than translate them to arrays, because there are some things you can do with collections that you can't with arrays (e.g. addAll). Even though this adds some special casing, it makes it cleaner to know what you're working with. But that's just my suspicion, I haven't tried implementing index sensitivity, so I'm very open to changing my mind.

PR #6398, opened 2022-10-26
osemgrep: Port autofix application for `fix`

nmote:
This isn't an exact port of the relevant code in `autofix.py`. I already had some autofix application code in `Autofix.ml` for testing purposes. In #6391 I refactored it to make it usable in this context.

Several differences:
- Instead of tracking details about applied fixes so that we can correctly apply multiple fixes to the same file, we just apply them from bottom to top so that they can't interfere with each other.
- We discard overlapping fixes and apply the ones that we can apply without conflicts. I think the Python CLI just applies them anyway, possibly incorrectly. However, the Python CLI also dedups matches that are the same except for metavariable bindings, which eliminates a major source of overlapping fixes. This logic is not ported yet.
- Here, we apply fixes using the byte offset into the file, instead of reading the file, splitting into lines, and then updating using the line and column like the CLI does. This is a lot simpler.
- When the `--dryrun` flag is passed, the Python CLI reports a list of fixed lines in its JSON output. This hasn't been ported yet, and in fact using the byte offset (see above) will make this a little bit more complicated.

Test plan:

Manual testing with the example given in #4964:

`osemgrep scan -c autofix.yaml autofix.py --autofix`

Note that this is an example where there are multiple matches reported for the same range, due to different possible metavariable bindings. For overlapping ranges, osemgrep applies the first one in the list that semgrep-core returns, just like the Python CLI. So, in this example it applies the same fix as the CLI.

Also, ran the e2e tests:

`PYTEST_USE_OSEMGREP=true pipenv run python -m pytest tests/e2e`

Before: 77 passed

After: 83 passed

PR checklist:

### DIFF
--- a/semgrep-core/src/fixing/Autofix.ml
+++ b/semgrep-core/src/fixing/Autofix.ml
@@ -43,7 +43,7 @@ let validate_fix lang text =
            (Exception.to_string e))
 
 type textedit = {
-  path : string;
+  path : Common.filename;
   (* 0-based byte index, inclusive *)
   start : int;
   (* 0-based byte index, exclusive *)
@@ -93,6 +93,28 @@ let apply_edits_to_text text edits =
   in
   if discarded_edits = [] then Success fixed_text
   else Overlap { partial_result = fixed_text; discarded_edits }
+
+let partition_edits_by_file edits =
+  (* TODO Consider using Common.group_by if we update it to return edits in
+   * order and add a comment specifying that changes to it must maintain that
+   * behavior. *)
+  let edits_by_file = Hashtbl.create 8 in
+  List.iter
+    (fun edit ->
+      let prev =
+        match Hashtbl.find_opt edits_by_file edit.path with
+        | Some lst -> lst
+        | None -> []
+      in
+      Hashtbl.replace edits_by_file edit.path (edit :: prev))
+    edits;
+  (* Restore the original order of the edits as they appeared in the input list.
+   * This is important so that we consistently choose to apply the first edit in
+   * the original input when there are edits for an identical span. *)
+  Hashtbl.filter_map_inplace
+    (fun _file edits -> Some (List.rev edits))
+    edits_by_file;
+  edits_by_file
 
 (******************************************************************************)
 (* Entry Points *)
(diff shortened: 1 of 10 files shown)


ievans:
I wonder if we can just skip the --dryrun flag in the port and maybe add it back later

nmote:
I already ported `--dryrun` in this PR, and it was easy enough to do for the part of it that I ported. It's the `fixed_lines` in particular that is going to be a little annoying. It probably won't be too bad, though. I just mentioned it because I chose to put it off for this specific PR. I don't know what it's used for (if anything).

aryx:
Woohoo! Agree with the approach. This is often what we did with martin; Looking at the original code, taking the good bits but not being afraid to follow a different approach if it makes things significantly simpler.

semgrep-core/src/fixing/Autofix.ml
```diff
@@ -94,6 +94,25 @@ let apply_edits_to_text text edits =
   if discarded_edits = [] then Success fixed_text
   else Overlap { partial_result = fixed_text; discarded_edits }
 
+let partition_edits_by_file edits =
+  let edits_by_file = Hashtbl.create 8 in
+  List.iter
```
aryx, line 102: probably can use one of the Common.group_by function to do this.

semgrep-core/src/fixing/Autofix.mli
```diff
@@ -16,6 +16,22 @@ val render_fix :
   target_contents:string Lazy.t ->
   string option
 
+type textedit = {
+  path : string;
```
aryx, line 20: Common.filename.

semgrep-core/src/osemgrep/cli_scan/Output.ml
```diff
@@ -21,6 +35,25 @@ let output_result (conf : Scan_CLI.conf) (res : Core_runner.result) : unit =
...
+    * At some point we could re-run Semgrep on all files where some fixes
+    * haven't been applied because they overlap with other fixes, and repeat
+    * until all are applied (with some bounds to prevent divergence). This
+    * probably happens rarely enough that it would be very fast, and it would
+    * make the autofix experience better.
+    * *)
+   let modified_files, _failed_fixes = apply_fixes conf cli_output in
+   if not conf.dryrun then
+     if modified_files <> [] then
+       pr2
```
aryx, line 52: You should now be able to use the Logs library. See my recent PR regarding logging.

aryx (approved):
Fantastic!

semgrep-core/src/fixing/Autofix.ml
nmote: Looks like `Common.group_by` reverses the order of the list elements. The order is important here. If I reverse it after calling `group_by`, then I'm relying on `group_by` to always reverse the order of the list elements. If someone changes it later, this will break.

```
let group_by f xs =
  (* use Hashtbl.find_all property *)
  let h = Hashtbl.create 101 in

  (* could use Set *)
  let hkeys = Hashtbl.create 101 in

  xs |> List.iter (fun x ->
    let k = f x in
    Hashtbl.replace hkeys k true;
    Hashtbl.add h k x
  );
  Hashtbl.fold (fun k _ acc -> (k, Hashtbl.find_all h k)::acc) hkeys []
;;

let l = ["1"; "2"; "3"];;

let g = group_by (fun _ -> 5) l;;

(* output: 3, 2, 1 *)
print_endline (g |> List.hd |> snd |> String.concat ", ");;
```

semgrep-core/src/osemgrep/cli_scan/Output.ml
nmote: I tried to faithfully reproduce the Python behavior here. It prints on stderr with no additional text other than what's seen here. I just tried out the Logs library and got:

```
osemgrep: [WARNING] successfully modified 1 file.
```

I had to use `Logs.warn` because otherwise it looks like this wouldn't get printed by default. This output seems weird and it's inconsistent with the existing output.

I guess I'm not sure how faithfully we are trying to reproduce the current behavior. Does it matter whether this is slightly different? If I were writing this from scratch I would probably use `Logs.info` for this, so it appears only when `--verbose` or higher is passed.

Anyway, I'm going to merge this without using Logs here but I'll be happy to make a followup PR depending on your answer here.

nmote:
CI failure is unrelated

PR #6403, opened 2022-10-27
Deduplicate 'create-tag' step ID in start-release workflow.

mjambon:
This is in response to this failure: https://github.com/returntocorp/semgrep/actions/runs/3333367837

> [Invalid workflow file: .github/workflows/start-release.yml#L231](https://github.com/returntocorp/semgrep/actions/runs/3333367837/workflow)
The workflow is not valid. .github/workflows/start-release.yml (Line: 231, Col: 13): The identifier 'create-tag' may not be used more than once within the same scope.

PR checklist:

### DIFF
--- a/.github/workflows/start-release.yml
+++ b/.github/workflows/start-release.yml
@@ -221,14 +221,14 @@ jobs:
           ref: "${{ needs.release-setup.outputs.release-branch }}"
           token: "${{ steps.token.outputs.token }}"
       - name: Create semgrep release version tag
-        id: create-tag
+        id: create-semgrep-tag
         run: |
           git config user.name ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


aryx (approved):
@spencerdrak is there any tool to check the validity of GHA workflow files statically?
CircleCI has such as feature.
(we could also write a semgrep rule for it ... if they dont provide such a tool).

aryx (commented):
This create-tag id is referenced though later.
Should we update also those instances?
Or it looks like it's also a job id, in addition to a step id.
Actually why do we need those step id?

aryx:
I've commented here on the breaking PR: https://github.com/returntocorp/semgrep/pull/6121#pullrequestreview-1157808659

PR #6415, opened 2022-10-29
Make it possible to turn off skipping minimized files

emjin:
See https://github.com/returntocorp/semgrep-proprietary/pull/328

This helps DeepSemgrep. I could instead have added an option to tune the amount of whitespace, but for now I wanted to err on the side of skipping fewer files, since performance is not a priority for DeepSemgrep. We can revisit this later.

Test plan: see the DeepSemgrep PR; nothing should change for Semgrep users

PR checklist:

### DIFF
--- a/semgrep-core/src/targeting/Skip_target.ml
+++ b/semgrep-core/src/targeting/Skip_target.ml
@@ -23,6 +23,13 @@ module Resp = Output_from_core_t
      than 500 bytes at the same time.
    - only some really short source files (2-3 lines) were found to have
      between 5% and 8% whitespace.
+
+  However, this threshold works less well for java files because it is not
+  unusual to have large import blocks at the top, resulting in a very low
+  whitespace fraction. This should not affect very many Java files, so for
+  now it is ok that we leave it. The consequence of reducing the frequency
+  is that Semgrep would likely take longer as it scans more costly minified
+  files. TODO we might want to do some benchmarking and change this default
 *)
 let min_whitespace_frequency = 0.07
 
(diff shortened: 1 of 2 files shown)


semgrep-core/src/targeting/Skip_target.ml
```diff
@@ -23,6 +23,13 @@ module Resp = Output_from_core_t
      than 500 bytes at the same time.
    - only some really short source files (2-3 lines) were found to have
      between 5% and 8% whitespace.
+
+  However, this threshold works less well for java files because it is not
+  unusual to have large import blocks at the top, resulting in a very low
+  whitespace fraction. This should not affect very many Java files, so for
+  now it is ok that we leave it. The consequence of reducing the frequency
+  is that Semgrep would likely take longer as it scans more costly minified
+  files. TODO we might want to do some benchmarking and change this default
```
IagoAbal, line 32: Could we also tweak this default so that it works for your example?

IagoAbal: What was the `stat.ws_freq` for that example?

IagoAbal (approved):

semgrep-core/src/targeting/Skip_target.ml
emjin: Yeah, we could do that. I wanted to just not skip them all for now because they might be helpful and we aren't focused on performance yet.

emjin: From my example I would have lowered it to 0.3

PR #6423, opened 2022-11-01
fix(autofix): Fix statement equality issue

nmote:
So far, AST-based autofix has been tested mostly on simple expressions. However, even some expressions can contain statements. Semgrep represents objects in JS using `DefStmt` nodes. The generated equality functions for the generic AST have a quirk when it comes to comparing statements: they must be wrapped in a function which mutates global state in order to indicate what kind of comparison should be performed. Otherwise, they raise an exception when attempting to check any statements for equality.

We should probably have a Semgrep rule for this, since this is not a typical property of equality functions.

Test plan: Automated tests. The test added here failed previously.

PR checklist:


nmote:
Semgrep rule for this is in #6424

aryx (approved):

PR #6425, opened 2022-11-01
feat(SSC): Support requirements.txt and yarn 2/3

mmcqd:
This PR adds support for reading dependencies from a `requirements.txt` lockfile, assuming a `requirements.in` manifest file, and for reading a `yarn.lock` file produced by the latest versions of yarn.

PR checklist:


cli/tests/e2e/rules/dependency_aware/python-requirements-sca.yaml
```diff
@@ -0,0 +1,10 @@
+rules:
+  - id: python-poetry-sca
```
brendongo: Should we change the id since I think this is testing requirements.txt generated by requirements.in

cli/src/semdep/parse_lockfile.py
```diff
@@ -491,6 +571,33 @@ def parse_dep(s: str) -> FoundDependency:
     yield from (parse_dep(dep) for dep in deps)
 
 
+def parse_requirements(
+    lockfile_text: str, manifest_text: Optional[str]
+) -> Iterator[FoundDependency]:
+    deps = [
+        (i, l.split("==")) for i, l in enumerate(lockfile_text.split("\n")) if "==" in l
```
brendongo: This seems fragile. Are all relevant lines in the requirements.txt guaranteed to have `==`

cli/src/semdep/parse_lockfile.py
```diff
@@ -124,6 +124,86 @@ def remove_trailing_octothorpe(s: str) -> str:
...
+
+    deps = lockfile_text.split("\n\n")[2:]
+    for dep in deps:
+        yield parse_dep(dep)
+
+
+def parse_yarn(
+    lockfile_text: str, manifest_text: Optional[str]
+) -> Iterator[FoundDependency]:
+    if lockfile_text.startswith("# This file is"):
```
brendongo, line 202: Is there something in the structure of the yarn2 we can use? I feel nervous relying on the existence of a comment

brendongo (changes requested):

cli/tests/e2e/rules/dependency_aware/python-requirements-sca.yaml
mmcqd: whoops, yes

cli/src/semdep/parse_lockfile.py
mmcqd: I added a function to remove comments in case people decide to put them on the same line as a package. None of the automated tools do this. Otherwise yes, any line that does not contain an `==` is either a comment, a newline, or was a non-pinned package, which is now filtered at the top.

cli/src/semdep/parse_lockfile.py
mmcqd: There's nothing more robust than that comment. They're all called `yarn.lock` and have similar structure. I don't think relying on a comment is bad. These files are only ever autogenerated and they always have the same comment if they're using yarn2. They always have a different comment if they're using yarn1. The alternative would be like "is there a version name that has no quotes around it, then it must be yarn1", but that's harder to check and no more accurate

mmcqd: I could check for the whole comment instead of the first few words, that might be clearer?

brendongo: I guess this is fine :D

cli/src/semdep/parse_lockfile.py
```diff
@@ -491,6 +572,55 @@ def parse_dep(s: str) -> FoundDependency:
     yield from (parse_dep(dep) for dep in deps)
 
 
+def parse_requirements(
+    lockfile_text: str, manifest_text: Optional[str]
+) -> Iterator[FoundDependency]:
+    for op in ["<", "<=", ">", ">="]:
+        if op in lockfile_text:
+            raise SemgrepError("requirements.txt contains non-pinned versions")
```
brendongo, line 580: Is it fine to raise here or will semgrep now fail on any repo with a requirements.txt file with <=

Probably worth adding a test that we dont crash with the common requirements.txt

brendongo (approved):

cli/src/semdep/parse_lockfile.py
mmcqd: This will get caught by the try/except clauses in `parse_lockfile_str`. So we'll print a warning to the command line and then consider the offending `requirements.txt` to be empty, and it will produce no findings.

PR #6449, opened 2022-11-03
tainting: Track taint as precisely as possible

IagoAbal:
And move all this logic to Taint_lval_env.

Even after adding field-sensitivity, if we had an l-value such as `x.y[i]`, we would track it as `x` rather than as `x.y`. Now we try to be as precise as possible, and we find the longest dotted-offset prefix to track.

We also added support for `thix. ...` l-values, by simply dropping the `this`.

Follows: 9de8d50fc18 ("tainting: Fix field-sensitivity regression (#6279)")
Follows: 756408aafeb ("Field Sensitive Tainting (#5697)")

Closes PA-2086

test plan:
make test # added two tests and updated three more

PR checklist:

### DIFF
--- a/changelog.d/pa-2086.changed
+++ b/changelog.d/pa-2086.changed
@@ -1,0 +1,4 @@
+taint-mode: Semgrep can now track taint through l-values of the form `this.x`.
+It will also be more precise when tracking taint on l-values involving an
+array access, previously if `x.a[i]` was tainted, then `x` itself was tainted;
+now only `x.a` will be considered tainted.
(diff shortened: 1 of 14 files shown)


emjin (approved):
Oh, I like this a lot better! This addresses the concerns I had about your previous PR. Much simpler and easier to think about.

emjin:
I didn't have a lot of time so I reviewed this pretty quickly. Maybe someone else should look at it first. I really like the way this is structured compared to what it was before, though. One note is that likely there is a file header comment that needs to be updated since we are now "field-sensitive-first". If not, there is file header comment that needs to be written.

IagoAbal:
> One note is that likely there is a file header comment that needs to be updated since we are now "field-sensitive-first". If not, there is file header comment that needs to be written.

Could you check the header comment of `Taint_lval_env.ml` ? Is that what you were expecting ?

PR #6453, opened 2022-11-03
feat(scip): install scip files with deepsemgrep

brandonspark:
## What:
This is a part of a multi-step change towards increasing type and naming resolution in DeepSemgrep, by using SCIP information. This is detailed in [this scope doc](https://www.notion.so/r2cdev/LSIF-Integration-408de64b7726488cb3c1c328f548fad1).

## Why:
This will make it so that we can distribute the necessary SCIP files in a relatively easy way, and produce better findings within DeepSemgrep.

## How:
This PR adds a change where, during installation of DeepSemgrep, we also access another endpoint within the App, which will distribute the URL of the S3 bucket which we can download the SCIP artifacts from. These will then be used by DeepSemgrep.

## Test plan:
Can't test this until the App change is added (adding the endpoint).

## Closes:
Closes PA-2108

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -48,9 +48,6 @@ def install_deep_semgrep() -> None:
         logger.info(
             f"Overwriting DeepSemgrep binary already installed in {deep_semgrep_path}"
         )
-    scip_path = Path(core_path).parent / "scip_files.zip"
-    if scip_path.exists():
-        logger.info(f"Overwriting SCIP files already installed in {scip_path}")
 
     if state.app_session.token is None:
         logger.info("run `semgrep login` before using `semgrep install`")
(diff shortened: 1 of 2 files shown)


emjin (changes requested):
I'd rather hold off on installing scip for people until doing this will actually benefit them. `semgrep install-deep-semgrep` is now a command we're giving to customers; at least one person has bought semgrep for deep-semgrep. Let's keep this experiment as internal as possible for now.

(If I've misinterpreted what this does and it is purely internal, feel free to dismiss)

cli/src/semgrep/commands/install.py
```diff
@@ -101,3 +102,63 @@ def install_deep_semgrep() -> None:
         stderr=subprocess.DEVNULL,
     ).rstrip()
     logger.info(f"DeepSemgrep Version Info: ({version})")
+
+
+@click.command(hidden=True)
+@handle_command_errors
+def install_scip() -> None:
+    """
+    Install the DeepSemgrep binary (Experimental)
```
aryx, line 111: you need to update this comment.

aryx (approved):

PR #6464, opened 2022-11-06
#5698 fix(yaml): excessive capture in maps

### LINKED ISSUE #5698: YAML matching sometimes captures too much text
**Describe the bug**
In this example: https://semgrep.dev/s/P3yp, The match on
```
y:
  foo: true
```
extends past `true` and includes the newline and comment following it.

**Expected behavior**
The match range should begin at `y` and end at `true`

MIJOTHY:
## Context

`yaml`-mode captures too much text in mapping contexts; it captures all
text from the final token in the map to the next token in the file. This
includes comments and leading whitespace. This comes from the upstream
library (https://github.com/avsm/ocaml-yaml).

The implemented fix is to use the position of the token in the event
immediately prior to the `Mapping_end` event. If there is no prior
event, fall back to the `Mapping_end` event's position. This should
be safe as a `Mapping_end` must be preceded by a `Mapping_start`
(right????).

See also:
https://r2c-community.slack.com/archives/C01LNF69F9B/p1666872497082059

Fixes https://github.com/returntocorp/semgrep/issues/5698

## Additional Considerations
I originally thought of implementing the fix getting the last element of the accumulator and using that position. It seems neater but seems slightly less performant? I don't mind refactoring to that solution if that's preferred (and correct).

Any more tests I can whack in?

## Other stuff
PR checklist:


semgrep-core/src/parsing/other/yaml_to_generic.ml
```diff
@@ -313,8 +313,17 @@ let parse (env : env) : G.expr list =
         let rest, end_pos = read_sequence acc in
         (seq :: rest, end_pos)
   and read_mappings acc : G.expr list * E.pos =
+    (* Capture the mutable `last_event` field prior to parsing, as the *)
+    (* `Mapping_end` event needs to make use of it. *)
+    let last_ev = env.last_event in
```
MIJOTHY, line 318: I think I would rather grab the last element of the accumulator (if it exists), and use the position of that element. But I get a strange typecheck error when I do so, and my ocaml-fu is inadequate for solving it 
```ocaml
    | E.Mapping_end, pos ->
        (* The `Mapping_end` event position is inaccurate: it specifies the *)
        (* position immediately prior to the next token. This causes comments *)
        (* and the leading whitespace of the next yaml element to be chomped on *)
        (* autofix. Instead, take the position of the previous element, which *)
        (* is either the last element in the mapping, or the start of the map *)
        (* if empty (implicitly). *)
        (match List.rev acc with
         | [] -> (acc, pos)
         | (_, last_pos) :: _ -> (acc, last_pos))
```

```
Line 325, characters 18-21:
325 |          | [] -> (acc, pos)
                        ^^^
Error: This expression has type ('a * 'b) list
       but an expression was expected of type G.expr list
       Type 'a * 'b is not compatible with type G.expr = AST_generic.expr 

```

aryx:
@emjin is the best reviewer for this but she is in vacation and will be back next week.
In the mean time I think I can merge this because this looks ok and emma can give it a second look when she will be back.

Thanks a lot! Awesome contrib.

PR #6498, opened 2022-11-10
Ensure that Rule.formula is serializable

IagoAbal:
The `Regexp_engine.t` type is a wrapper for a C type coming from the C PCRE library, and it is represented by a "Custom" block. Values of this type are not serializable, and will cause an exception if we try calling `Marshal.to_string` (or variant) on them. This can cause problems when serializing any data structure that depends on `Xpattern.t` and `Rule.formula`.

Previously, we got weird crashes after PR #5725 (taint labels). This was due to redefining type `Taint.source` as `Rule.taint_source call_trace`, where `Rule.taint_source` contains a `Rule.formula`. Whenever a taint finding is produced, we compute the taint trace this way:

    lazy (taint_trace_of_src_to_sink source tokens sink)

This `lazy ...` expression creates a thunk that capturies a `source` which contains a `Rule.formula`. Even though the formula is not used to compute the taint trace, it still needs to be serialized in order to serialize the thunk.

So, when using -j 2 (or higher) and -json (in text mode we print findings right away and hence the thunk gets evaluated), this thunk remains unevaluated and reaches the result of `Match_rules.check`. Then Parmap tries to serialize it to communicate the result back to the master process... and if the formula contains any compiled regex, it crashes.

This also affects us if we try to serialize a `Dataflow_tainting.config` for caching purposes. While we could remove the formulas from these types, it is less error prone to just make sure that formulas can be serialized.

Follows: 2e6b3079d51 ("Quick-fix for regression in 0.107.0 after taint labels PR (#5867)")

Regarding perf, this should not cause any noticeable regression, unless we build some synthetic case with lots of rules making an absurd use of regexes. I ran `semgrep -c p/default` on three repos from stress-test-monorepo (elasticsearch, vscode, and mediawiki) and with this change all runs were actually a little bit faster (maybe just coincidence).

test plan:
make test
(In PR #5867 we modified Test_engine.ml to check that the result of Match_rules.check can be serialized, see 'Test_engine.make_tests'.)

PR checklist:

### DIFF
--- a/semgrep-core/src/core/Xpattern.ml
+++ b/semgrep-core/src/core/Xpattern.ml
@@ -21,7 +21,8 @@
 (* Types *)
 (*****************************************************************************)
 
-type regexp = Regexp_engine.t [@@deriving show, eq]
+type compiled_regexp = Regexp_engine.t [@@deriving show, eq]
+type regexp_string = string [@@deriving show, eq]
 
 (* used in the engine for rule->mini_rule and match_result gymnastic *)
 type pattern_id = int [@@deriving show, eq]
@@ -29,7 +30,21 @@ type pattern_id = int [@@deriving show, eq]
 type xpattern_kind =
   | Sem of Pattern.t * Lang.t (* language used for parsing the pattern *)
   | Spacegrep of Spacegrep.Pattern_AST.t
-  | Regexp of regexp
+  | Regexp of regexp_string
+      (** NOTE: We used to keep the compiled regexp of type `Regexp_engine.t', but
+      * that is not a pure OCaml data structure and it cannot be serialized.
+      *
+      * This had previously caused weird Semgrep crashes like
+      *
+      *     Invalid_argument "output_value: abstract value (Custom)"
+      *
+      * after PR #5725 (taint labels), see PA-1724.
+      *
+      * When we use -j 2 (or higher) and -json, then Parmap is used, and the result
+      * of `Match_rules.check' (of type `Report.match_result') will contain some
+      * unevaluated thunks. Then Parmap will try to marshal this value and it will
+      * crash if there is any "Custom" block involved.
+      *)
   | Comby of string
 [@@deriving show, eq]
 
--- a/semgrep-core/src/core/Xpattern.ml
+++ b/semgrep-core/src/core/Xpattern.ml
@@ -23,6 +23,7 @@
 
 type compiled_regexp = Regexp_engine.t [@@deriving show, eq]
 type regexp_string = string [@@deriving show, eq]
+(* see the NOTE "Regexp" below for the need to have this type *)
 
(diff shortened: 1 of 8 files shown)


semgrep-core/src/engine/Match_search_mode.ml
```diff
@@ -124,7 +124,8 @@ let partition_xpatterns xs =
          match pat with
          | XP.Sem (x, _lang) -> Common.push (x, inside, pid, str) semgrep
          | XP.Spacegrep x -> Common.push (x, pid, str) spacegrep
-         | XP.Regexp x -> Common.push (x, pid, str) regexp
+         | XP.Regexp x ->
+             Common.push (Regexp_engine.pcre_compile x, pid, str) regexp
```
aryx, line 128: you can add a [@@profiling] on Regexp_engine.pcre_compile so we can track (thx to -profile) whether
those calls are expensive.

semgrep-core/src/core/Xpattern.ml
```diff
@@ -21,15 +21,30 @@
 (* Types *)
 (*****************************************************************************)
 
-type regexp = Regexp_engine.t [@@deriving show, eq]
+type compiled_regexp = Regexp_engine.t [@@deriving show, eq]
+type regexp_string = string [@@deriving show, eq]
```
aryx, line 25: maybe a comment here like '(* see the NOTE below on Regexp for the need to have this type *)'

aryx (approved):
Would have been great to have a test case showing the issue.
Otherwise looks good.

IagoAbal:
> Would have been great to have a test case showing the issue.
> Otherwise looks good.

Oh we do have that, see `Test_engine.ml`:
```ocaml
             (* Check that the result can be marshalled, as this will be needed
                 when using Parmap! See PA-1724. *)
             (try Marshal.to_string res [ Marshal.Closures ] |> ignore with
             | exn ->
                 failwith
                   (spf "exn on %s (exn = %s)" file (Common.exn_to_s exn)));
```
If you undo this change (and keep the `lazy` in Match_tainting_mode` you get many failing tests).

semgrep-core/src/engine/Match_search_mode.ml
IagoAbal: Good point!

IagoAbal:
I wonder what is going on, tests passed 5 days ago, semgrep-core tests pass locally. I rebased against `develop` and trying again...

aryx:
I did modify some code recently in Parse_rule.ml regarding regexps in parse_xptattern.

aryx:
That's the CI failure. See the error in Parse_rule.ml

IagoAbal:
Yeah, I was running the tests before rebasing so I didn't get that error. I got it when I rebased.

PR #6529, opened 2022-11-15
tainting:  Add experimental option `taint_only_propagate_through_assignments`

IagoAbal:
To disable built-in taint propagators, that is: constructors, operators, dot-access, array indexing, and function calls. By disabling this option, the only thing that propagates taint implicit will be assignment statements.

This is meant to be useful for writing secret detection rules.

Closes PA-2193

test plan:
make test # added one test

PR checklist:

### DIFF
--- a/interfaces/Config_semgrep.atd
+++ b/interfaces/Config_semgrep.atd
@@ -26,6 +26,7 @@ type t = {
   ~taint_unify_mvars <ocaml default="false"> : bool;
   ~taint_assume_safe_functions <ocaml default="false"> : bool;
   ~taint_assume_safe_indexes <ocaml default="false"> : bool;
+  ~taint_builtin_propagators <ocaml default="true"> : bool;
 
   (* 'ac' stands for associative-commutative matching *)
   ~ac_matching <ocaml default="true"> : bool;
--- a/interfaces/Config_semgrep.atd
+++ b/interfaces/Config_semgrep.atd
@@ -26,7 +26,9 @@ type t = {
   ~taint_unify_mvars <ocaml default="false"> : bool;
   ~taint_assume_safe_functions <ocaml default="false"> : bool;
   ~taint_assume_safe_indexes <ocaml default="false"> : bool;
-  ~taint_builtin_propagators <ocaml default="true"> : bool;
+  (* when you are paranoid about minimizing FPs, and probably useful for
+   * writing secret detection rules *)
+  ~taint_only_propagate_through_assignments <ocaml default="false"> : bool;
 
   (* 'ac' stands for associative-commutative matching *)
   ~ac_matching <ocaml default="true"> : bool;
--- a/semgrep-core/src/tainting/Dataflow_tainting.ml
+++ b/semgrep-core/src/tainting/Dataflow_tainting.ml
@@ -121,12 +121,18 @@ let status_to_taints = function
 let is_exact x = x.overlap > 0.99
 
 let union_map_taints_and_vars env f xs =
-  xs
-  |> List.fold_left
-       (fun (taints1, lval_env) x ->
-         let taints2, lval_env = f { env with lval_env } x in
-         (Taints.union taints1 taints2, lval_env))
-       (Taints.empty, env.lval_env)
+  let taints, lval_env =
+    xs
+    |> List.fold_left
+         (fun (taints1, lval_env) x ->
+           let taints2, lval_env = f { env with lval_env } x in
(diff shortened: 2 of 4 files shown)


interfaces/Config_semgrep.atd
```diff
@@ -26,6 +26,7 @@ type t = {
   ~taint_unify_mvars <ocaml default="false"> : bool;
   ~taint_assume_safe_functions <ocaml default="false"> : bool;
   ~taint_assume_safe_indexes <ocaml default="false"> : bool;
+  ~taint_builtin_propagators <ocaml default="true"> : bool;
```
emjin: When I saw this, my first reaction was "why would we want to do this?". We're probably going to lose the knowledge of this. Can we include a note that secrets was the use case that prompted it? Also, maybe we can rename this from "builtin_propagators" to "only_propagate_through_assignments" or even "secrets_mode"

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -581,9 +589,12 @@ and check_tainted_lval_offset env offset =
   | Index e ->
       let taints, lval_env = check_tainted_expr env e in
       let taints =
-        if not env.options.taint_assume_safe_indexes then taints
-        else (* Taint from the index should be ignored. *)
+        if
+          env.options.taint_assume_safe_indexes
+          || not env.options.taint_builtin_propagators
```
emjin: This is getting complicated to read. I could easily imagine myself adding a `assume_safe_fields` but forgetting to add `taint_builtin_propagators`. 

I would prefer that we have one central place where we read the taint options and then set the fields `propagate_through_default_propagators`, `propagate_through_indexes`, and `propagate_through_functions`. Then, when deciding whether or not to return taints, we only have to check one thing.

The central location makes it easy to reason about how to set each field (because you have all the context of all the other fields) and each if statement is a clear and consistent check

semgrep-core/tests/rules/taint_no_builtin_props.py
```diff
@@ -0,0 +1,14 @@
...
+    x = "password"
+    y = x
+    z = y
+    #ruleid: secrets
+    sink(z)
+
+def ok():
+    x = "password"
+    y = x
+    z = y+"ok"
```
emjin, line 12: This test case does not justify to me disabling all propagators for secrets. I feel like I don't have a good understanding of why that is necessary. It seems to me that we just don't want concatenation to propagate taint (though I'd also like to see a more illustrative use case for why---in this case, surely having the password as part of the string). Other methods of propagation still preserve the value of the secret whole. Could we never propagate a secret through an array?

emjin (changes requested):
I might just not have studied the scope doc in enough detail, but I am not convinced that we need to disable all builtin propagators for secrets. It seems like we could miss legitimate cases. I'm also just generally concerned about the options matrix getting too complicated (see comment for a suggestion).

emjin:
Feel free to dismiss my review if this is needed urgently to support SR

IagoAbal:
> I am not convinced that we need to disable all builtin propagators for secrets. It seems like we could miss legitimate cases.

It could be, this is more of an experiment to figure that out. This is just what I inferred from discussions with @LewisArdern. But SR needs to write some rules using this to get a better sense of it.

interfaces/Config_semgrep.atd
IagoAbal: Fair enough, I will add some docs here, and when it has been tested a bit also do semgrep.dev/docs. Regarding the name I'm less sure, but I could take `only_propagate_through_assignments` for now.

semgrep-core/tests/rules/taint_no_builtin_props.py
IagoAbal: We're trying to understand what is needed for secrets. I see Lewis writes rules where he doesn't want "anything" than assignments to propagate taint, so we try that and see how it goes. Probably some of this leads to FNs, but that's always a trade-off you make when you want to minimize FPs. It's still possible to write explicit taint propagators.

emjin (approved):

IagoAbal:
Merged because the changes introduced by this PR require enabling an option which is disabled by default, so it shouldn't have any effect on performance. Plus the benchmarks CI check has been systematically failing on every recent PR.

emjin:
Yeah I tried to click through to see if I could figure out which PR could have caused the latest benchmark failures, but it's so hard to tell. We can change the pinned version

PR #6532, opened 2022-11-16
RFC: WIP on removing desugaring for JS requires

nmote:
I'd like to address the TODOs in the aliasing_require.js test, and also have some consistent logic in Semgrep upon which I can build while adding CommonJS import/export support to DeepSemgrep.

Currently, we desugar `require` calls to `ImportFrom` nodes in one specific case. This only occurs when the `require` is on the RHS of a variable assignment, and the LHS is a destructuring, e.g. `const { f1, f2 } = require("file");`. However, instead of replacing the `require`, the `ImportFrom` appears alongside it. I believe that this was done because there isn't a way to consistently transform require calls into `ImportFrom`. `require` calls are simply expressions, and can appear anywhere. There is also the potential for complicated destructuring where Semgrep bails out from desugaring at all. Thus, to be able to consistently match a pattern such as `require('file')` we need to keep the original node around.

There is no equivalent desugaring to `ImportAs` for `const f = require("file");`, thus the TODO item in the test.

My first attempt was to simply add an equivalent desugaring to `ImportAs`. Unfortunately, this led to an issue when matching the pattern `const $X = require('file');` against the target `const x = require('file')`. Both the target and pattern get desugared. Then, in the target, the name `x` is defined twice, once in the traditional `DefStmt` and once in the additional `ImportAs`. Differences in how these names get resolved mean that the `$X` metavariable is not considered to be bound to different things in the two different places, and so the match fails.

Rather than attempting to figure out how to change that behavior without disrupting other behavior that we rely on, I decided to try to remove the desugaring altogether and instead perform name resolution on `Require`s in `Naming_AST`, as suggested in a comment in the removed code.

That worked fairly well, except that there is also a test case (equivalence_import_require.js) where we want to match `import {$X1, $X2} from "module-name";` with `const { export1, export2 } = require("module-name");`

At first glance, this looks like it just requires a simple adjustment to the matching code to handle this case, now that we don't desugar requires to `ImportFrom` (which is how this matched before). I implemented this matching change here in `Generic_vs_generic.ml`. But, I was stymied by another desugaring. `import {x, y} from 'z'` is desugared into two `ImportFrom` nodes, one for `x` and one for `y`. There appears to be no straightforward way to change matching behavior so that in this one specific case, a single statement in the target can be matched by one or more statements in the pattern. We could change `m_stmts_deep_uncached` to make this work, but this, again, felt like a point where I should take a step back and consider whether the desugaring makese sense.

To solve this, I would like to remove the desugaring of `import {x, y} from 'z';` into multiple `ImportFrom` nodes. To keep current behavior, we would have to make sure that `import {x} from 'z';` also matches the target `import {x, y} from 'z';`. This would be straightforward to implement. We would have to make some changes to the generic AST to represent the underlying JS code with greater fidelity, and that is always a bit of an undertaking since it impacts most of Semgrep. However, I don't know what effects this might have, or whether it might make some analyses more complex. Still, it seems like the right way forward. All of these problems are basically caused by losing fidelity when we desugar. It seems clear to me that the solution is to maintain fidelity and avoid desugaring.

Feedback is welcome.

PR checklist:

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -2211,6 +2211,7 @@ and m_stmt a b =
       m_condition a1 b1 >>= fun () -> m_stmt a2 b2
   | G.DefStmt a1, B.DefStmt b1 -> m_definition a1 b1
   | G.DirectiveStmt a1, B.DirectiveStmt b1 -> m_directive a1 b1
+  | G.DirectiveStmt a1, B.DefStmt b1 -> m_directive_vs_def a1 b1
   | G.DoWhile (a0, a1, a2), B.DoWhile (b0, b1, b2) ->
       m_tok a0 b0 >>= fun () ->
       m_stmt a1 b1 >>= fun () -> m_expr a2 b2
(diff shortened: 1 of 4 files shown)


nmote:
Moving forward with removing desugaring. First PR here: https://github.com/returntocorp/semgrep/pull/6582

PR #6535, opened 2022-11-16
refactoring: introduce Rule.Err that gathers all exns in one type

aryx:
test plan:
make test

PR checklist:

### DIFF
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -320,7 +320,6 @@ type error =
   | InvalidYaml of string * Parse_info.t
   | DuplicateYamlKey of string * Parse_info.t
   | UnparsableYamlException of string
-  | ExceededMemoryLimit of string
 
 (* can't use Error because it's used for severity *)
 exception Err of error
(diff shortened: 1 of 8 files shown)


semgrep-core/src/core/Rule.ml
```diff
@@ -316,6 +314,20 @@ and invalid_rule_error_kind =
...
 
+(* General errors *)
+type error =
+  | InvalidRule of invalid_rule_error
+  | InvalidYaml of string * Parse_info.t
+  | DuplicateYamlKey of string * Parse_info.t
+  | UnparsableYamlException of string
+
+(* can't use Error because it's used for severity *)
+exception Err of error
```
mjambon, line 325: `Err` is fine but `Error` would be fine too, I think. I find that type-driven disambiguation (or whatever it's called) works well. It requires adding type constraints but it's more precise than just the module name. For example, compare:
```ocaml
match (x : R.severity) with
| Error -> ...
```
vs
```ocaml
match x with
| R.Error -> ... (* `R` isn't specific enough to determine the type of x *)
```

PR #6574, opened 2022-11-21
feat(cli): deep dataflow traces for CLI output

brandonspark:
## What:
Currently, our JSON output (and text, and SARIF) for the CLI does not allow long call traces for the taint source, and does not allow taint sink data at all.

## Why:
This is bad, because we want to output that information so people can debug DeepSemgrep, as well as consume that output to do cool stuff. It also just isn't faithful to real DeepSemgrep findings.

## How:
Refactored the interfaces types, and caused it to be output in the proper way.

## Test plan:
I tried it out. It looks like this:
image

PR checklist:


cli/src/semgrep/formatter/text.py
```diff
@@ -181,6 +185,91 @@ def _finding_to_line(
...
+            yield from TextFormatter._match_to_lines(
+                data[0],
+                data[1],
+                color_output,
+                per_finding_max_lines_limit,
+                per_line_max_chars_limit,
+            )
+
+            if intermediate_vars and len(intermediate_vars) > 0:
+                # TODO change this message based on rule kind of we ever use
```
emjin:
```suggestion
                # TODO change this message based on rule kind if we ever use
```

cli/src/semgrep/formatter/text.py
```diff
@@ -181,6 +185,91 @@ def _finding_to_line(
...
+                # dataflow traces for more than just taint
+                yield (8 * " " + "Taint flows through these intermediate variables:")
```
emjin: Better to use a variable for the 8, you have this value in several places

cli/src/semgrep/formatter/text.py
```diff
@@ -181,6 +185,91 @@ def _finding_to_line(
...
+
+    @staticmethod
+    def _call_trace_to_lines(
+        call_trace: out.CliMatchCallTrace,
+        color_output: bool,
+        per_finding_max_lines_limit: Optional[int],
+        per_line_max_chars_limit: Optional[int],
+    ) -> Iterator[str]:
+        trace = call_trace.value
+        if isinstance(trace, out.CliLoc):
```
emjin, line 228: I don't need you to change this because I assume this predates your PR but what are these names

cli/src/semgrep/formatter/text.py
```diff
@@ -140,6 +141,9 @@ def _format_lines(
                     # while stripping a string, the ANSI code for resetting color might also get stripped.
                     line = line + colorama.Style.RESET_ALL
 
+            if show_path:
+                yield f" " * 9 + f"{with_color(Colors.cyan, f'{path}', bold=False)}"
```
emjin: Why is this 9 and not 8 like everything else lol

cli/src/semgrep/join_rule.py
```diff
@@ -369,18 +369,47 @@ def cli_intermediate_vars_to_core_intermediate_vars(
...
+    if isinstance(value, core.CliLoc):
+        return core.CoreMatchCallTrace(core.CoreLoc(value.value[0]))
+    elif isinstance(value, core.CliCall):
+        data, intermediate_vars, call_trace = value.value
+
+        return core.CoreMatchCallTrace(
+            core.CoreCall(
+                (
+                    data[0],
+                    cli_intermediate_vars_to_core_intermediate_vars(intermediate_vars),
```
emjin, line 385: For my own understanding, why do we convert the cli vars back to core vars?

cli/src/semgrep/join_rule.py
```diff
@@ -369,18 +369,47 @@ def cli_intermediate_vars_to_core_intermediate_vars(
...
+                    cli_call_trace_to_core_call_trace(call_trace),
+                )
+            )
+        )
+    else:
+        # unreachable
+        return None
```
emjin, line 393: Let's log a warning when it inevitably is reached

cli/src/semgrep/rule_match.py
```diff
@@ -304,19 +304,45 @@ def is_blocking(self) -> bool:
...
+                return core.CliMatchCallTrace(
+                    core.CliLoc(translate_loc(call_trace.value.value))
+                )
+            elif isinstance(call_trace.value, core.CoreCall):
+                intermediate_vars = []
+                for var in call_trace.value.value[1]:
+                    location, content = translate_loc(var.location)
+                    intermediate_vars.append(
+                        core.CliMatchIntermediateVar(location=location, content=content)
+                    )
```
emjin: This looks to me like it could be a list comprehension, I think we generally prefer those where possible for purity + speed reasons

cli/src/semgrep/rule_match.py
```diff
@@ -304,19 +304,45 @@ def is_blocking(self) -> bool:
 
     @property
     def dataflow_trace(self) -> Optional[core.CliMatchDataflowTrace]:
+        def translate_loc(location: core.Location) -> Tuple[core.Location, str]:
```
emjin, line 309: Would be nice to quickly explain why this is necessary

emjin (approved):
Left some small comments. We should also add a TODO to add e2e tests but we can merge this for now

cli/src/semgrep/join_rule.py
brandonspark: pretty much just because the historic join rule code uses core stuff, whereas it receives cli stuff. no idea why it's necessary, but i just made the types fit

cli/src/semgrep/formatter/text.py
brandonspark: cause the path is slightly indented, to separate it from the intervening messages

cli/src/semgrep/formatter/text.py
brandonspark: "Location" versus "Call". Suppose I could have used "Match"

changelog.d/pa-2116.added
```diff
@@ -0,0 +1 @@
+CLI: Added deep traces to `--dataflow-traces`
```
IagoAbal, line 1: I am not sure this is a official convention, but I would have written it like `DeepSemgrep: ...` to make clear this a DeepSemgrep related change. It can help tech writers decide where to put this in the release notes.

eran-medan:
Small question, is the small inconsistency in the format as designed? What is the reasoning behind it? `location` feels a bit more informative than an array / tuple with `CliLoc` in index 0, and another array as value :) is this following some more standardized format? if so why didn't intermediate_vars get the same treatment. 

Not opening an issue yet as I'm sure it's my lack of understanding and it's more of a question. 

e.g. 

intermediate_vars 

```
"intermediate_vars": [
    {
      "content": "a",
      "location": ... (start/end/path)
    }
]
```

vs the taint sink / source 

```
  "taint_sink/taint_source": [
    "CliLoc",
    [
      {
       ...(start/end/path)
      },
      "sink/source"
    ]
  ]
},
        
``` 
 
Is that as designed? What is the reasoning? :)

PR #6581, opened 2022-11-22
tainting: Keep only the shortest trace we can find for a source

IagoAbal:
And helps avoiding infinite loops when inferring taint signatures for mutually recursive procedures.

Closes PA-2224

test plan:
make test

PR checklist:

### DIFF
--- a/semgrep-core/src/tainting/Taint.ml
+++ b/semgrep-core/src/tainting/Taint.ml
@@ -16,11 +16,19 @@
 module G = AST_generic
 module PM = Pattern_match
 
+let logger = Logging.get_logger [ __MODULE__ ]
+
+(*****************************************************************************)
+(* Call traces *)
+(*****************************************************************************)
+
 type tainted_tokens = G.tok list [@@deriving show]
 (* TODO: Given that the analysis is path-insensitive, the trace should capture
  * all potential paths. So a set of tokens seems more appropriate than a list.
  * TODO: May have to annotate each tainted token with a `call_trace` that explains
- * how it got tainted.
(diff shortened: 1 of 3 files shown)


semgrep-core/src/tainting/Taint.ml
```diff
@@ -16,20 +16,56 @@
...
+ * `x = f(tainted)`, it may be interesting to see how is `f` propagating the
+ * taint on its input.
  *)
 
 type 'a call_trace =
   | PM of PM.t * 'a
   | Call of G.expr * tainted_tokens * 'a call_trace
 [@@deriving show]
 
+let lenght_of_call_trace ct =
```
emjin:
```suggestion
let length_of_call_trace ct =
```

emjin (approved):
Implementation looks fine to me. I wonder how we can make it more visible that we only return one trace for a source. I *think* this is the obvious thing to happen, but I could see it making someone's life hard when trying to debug.

emjin:
In general the taint trace logic was pretty confusing to me. I spent a while looking for it because it happens by side effect when a finding is found. That's why I'm a little worried about this PR's change being hard to find

IagoAbal:
> In general the taint trace logic was pretty confusing to me. I spent a while looking for it because it happens by side effect when a finding is found. That's why I'm a little worried about this PR's change being hard to find

I'll think on it and maybe do some refactoring or improve the docs. For now I added a note in the docs for `Taint.signature` to clarify that we only keep the shortest call-trace for each source. I definitely need to get back to this, but for now I just wanted to fix the infinite loop issue.

PR #6582, opened 2022-11-22
fix(js): Faithfully represent import statements in generic AST

### LINKED ISSUE #5305: TS/JS - Import drops $VALUE before {...}
**Describe the bug**
The addition [for import { } from 'value' ](https://github.com/returntocorp/semgrep/issues/5012) was awesome for TS/JS  but it also would be great to capture the  values before {..} 

```
  - pattern-inside: |
      import {  ...,useRef,... } from "react"
      ...

^ works

      import React,{  ...,useRef,... } from "react"
      ...

^ does not work 
```

Works:
https://semgrep.dev/playground/s/ylo8?develop=true

Does not work:
https://semgrep.dev/s/ylo8

**To Reproduce**
Works:
https://semgrep.dev/playground/s/ylo8?develop=true

Does not work:
https://semgrep.dev/s/ylo8

**Expected behavior**
$REACT should be also captured 

**Priority**
How important is this to you?
- [x] P2: annoying but not blocking me

nmote:
Previously, a statement such as `import {x, y} from 'z'` would have been desugared into two separate `ImportFrom` statements. This caused issues with matching, such as #5305, and it also led to the issue described in #6532.

This PR changes the `ImportFrom` node to have a list of imported values instead of a single imported value. That allows us to represent the example above as a single `ImportFrom` node without desugaring. It also allows `import {} from 'z';` to be represented (though it's unclear to me why anybody would ever write that). This change addresses the issues mentioned above.

Fixes #5305

Test plan: Automated tests

PR checklist:

### DIFF
--- a/semgrep-core/src/core/ast/AST_generic.ml
+++ b/semgrep-core/src/core/ast/AST_generic.ml
@@ -1749,8 +1749,7 @@ and directive_kind =
   | ImportFrom of
       tok (* 'import'/'from' for Python *)
       * module_name
-      * ident
-      * alias option (* as name alias *)
+      * (ident * alias option (* as name alias *)) list
   | ImportAs of tok * module_name * alias option (* as name *)
   (* Bad practice! hard to resolve name locally.
    * We use ImportAll for C/C++ #include and C++ 'using namespace'.
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -3030,16 +3030,17 @@ and m_directive a b =
 
   m_directive_basic a.d b.d >!> fun () ->
   match a.d with
-  (* normalize only if very simple import pattern (no alias) *)
-  | G.ImportFrom (_, _, _, None)
-  | G.ImportAs (_, _, None) -> (
-      (* equivalence: *)
-      let normal_a = Normalize_generic.normalize_import_opt true a.d in
-      let normal_b = Normalize_generic.normalize_import_opt false b.d in
-      match (normal_a, normal_b) with
-      | Some (a0, a1), Some (b0, b1) ->
-          m_tok a0 b0 >>= fun () -> m_module_name_prefix a1 b1
-      | _ -> fail ())
+  (* normalize only if very simple import pattern (no aliases) *)
+  | G.ImportFrom (_, _, imports)
+    when List.for_all
+           (function
+             (* None here means that there is no local alias for the imported
+              * name. *)
+             | _imported_name, None -> true
+             | _imported_name, Some _aliases -> false)
+           imports ->
+      m_normalized_imports a.d b.d
+  | G.ImportAs (_, _, None) -> m_normalized_imports a.d b.d
   (* more complex pattern should not be normalized *)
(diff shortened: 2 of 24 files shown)


nmote:
I just updated this PR with some tests that show some more improved matching behavior. Check out `equivalence_import_variations4.js` and compare to https://semgrep.dev/s/B757

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -3003,16 +3003,15 @@ and m_directive a b =
...
-      let normal_b = Normalize_generic.normalize_import_opt false b.d in
-      match (normal_a, normal_b) with
-      | Some (a0, a1), Some (b0, b1) ->
-          m_tok a0 b0 >>= fun () -> m_module_name_prefix a1 b1
-      | _ -> fail ())
+  (* normalize only if very simple import pattern (no aliases) *)
+  | G.ImportFrom (_, _, imports)
+    when List.for_all
+           (function
+             | _, None -> true
```
emjin: nit: this would be more readable for me if there were underscored names given to the wildcards. In particular, something like `(_name, Some(_aliases))` would let me know what we're filtering for. It's not the norm in this file though so this is fine

semgrep-core/src/parsing/pfff/js_to_generic.ml
```diff
@@ -756,22 +763,22 @@ and require_to_import_in_stmt_opt st =
                          | (s1, _), (s2, _) when s1 =$= s2 -> None
                          | _ -> Some (id2, G.empty_id_info ())
                        in
-                       G.DirectiveStmt
-                         (G.ImportFrom (treq, G.FileName file, id1, alias_opt)
-                         |> G.d)
-                       |> G.s
+                       (id1, alias_opt)
                    | _ -> raise ComplicatedCase)
```
emjin, line 767: lmao this exception

emjin (approved):
I'm not super familiar with Javascript imports, but this seems straightforward and good to me.  I wonder if there are any SR issues this solves, I know they've had friction with our Javascript imports

semgrep-core/src/matching/Generic_vs_generic.ml
nmote: I've made this change and added a comment. Hopefully that's clearer now.

nmote:
Yes, @LewisArdern reported the issue that this closes. Maybe there are others, too.

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -1749,8 +1749,7 @@ and directive_kind =
   | ImportFrom of
       tok (* 'import'/'from' for Python *)
       * module_name
-      * ident
-      * alias option (* as name alias *)
+      * (ident * alias option (* as name alias *)) list
```
aryx, line 1752: Could we add a comment here explaining why we don't "desugar" in multiple ImportFrom like
we do for variable declarations. I already forgot why we do that ...

nmote: https://github.com/returntocorp/semgrep/pull/6634

PR #6584, opened 2022-11-22
tainting: Fix taint tracking for l-values with array accesses

IagoAbal:
Taint going into `x.a[i].c` should not make `x.a[i]` tainted!

Closes PA-2225

test plan:
make test # updated two tests

PR checklist:

### DIFF
--- a/semgrep-core/src/tainting/Taint_lval_env.ml
+++ b/semgrep-core/src/tainting/Taint_lval_env.ml
@@ -69,36 +69,50 @@ let union le1 le2 =
  * not represent the exact same object as the original l-value, but an
  * overapproximation. For example, the normalized l-value of `x[i]` will be `x`,
  * so the taints of any element of an array are tracked via the array itself. *)
-let rec normalize_lval lval =
+let normalize_lval lval =
+  let open Common in
   let { IL.base; rev_offset } = lval in
-  match lval.IL.rev_offset with
-  | [] -> (
-      (* Base case, no offset; we can only track variables. *)
-      match base with
-      | Var _ -> Some lval
-      | VarSpecial _
-      | Mem _ ->
-          None)
-  | _ :: rev_offset' -> (
-      (* Must find the longest prefix of the form x.a_1. ... . a_N. *)
-      let is_dots = LV.is_dots_offset rev_offset in
-      match base with
-      | Var _ when is_dots -> Some lval
-      | VarSpecial _ when is_dots -> (
-          (* this.x.a_1. ... . a_N becomes x.a_1. ... . a_N *)
-          match List.rev lval.rev_offset with
-          | { o = IL.Dot var; _ } :: offset' ->
-              Some { IL.base = Var var; rev_offset = List.rev offset' }
-          | []
-          | { o = IL.Index _; _ } :: _ ->
-              logger#error "normalize_lval: Impossible happened";
-              None)
-      | Var _
-      | VarSpecial _ ->
-          (* Offset-chain is not of the form .a_1. ... . a_N, so drop the last
-           * offset until the condition is met or we reach the base case. *)
-          normalize_lval { base; rev_offset = rev_offset' }
-      | Mem _ -> None)
+  let* base, rev_offset =
+    match base with
+    | Var _ -> Some (base, rev_offset)
+    | VarSpecial _ -> (
+        (* this.x.a_1. ... . a_N becomes x.a_1. ... . a_N *)
+        match List.rev lval.rev_offset with
+        | { o = IL.Dot var; _ } :: offset' -> Some (Var var, List.rev offset')
+        | [] -> None
+        | { o = IL.Index _; _ } :: _ ->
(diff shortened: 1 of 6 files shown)


semgrep-core/src/tainting/Taint_lval_env.ml
```diff
@@ -69,36 +69,50 @@ let union le1 le2 =
...
-      | VarSpecial _ ->
-          (* Offset-chain is not of the form .a_1. ... . a_N, so drop the last
-           * offset until the condition is met or we reach the base case. *)
-          normalize_lval { base; rev_offset = rev_offset' }
-      | Mem _ -> None)
+  let* base, rev_offset =
+    match base with
+    | Var _ -> Some (base, rev_offset)
+    | VarSpecial _ -> (
+        (* this.x.a_1. ... . a_N becomes x.a_1. ... . a_N *)
```
emjin: 🤔 Can this interact with deep-semgrep in any way? Maybe if you have `this.foo().a`, where `this` has type `B`, `B extends A`, and `A.foo()` is tainted? I'm wondering if we would still perform the inheritance check when we need it.

semgrep-core/src/tainting/Taint_lval_env.ml
```diff
@@ -69,36 +69,50 @@ let union le1 le2 =
...
+        match List.rev lval.rev_offset with
+        | { o = IL.Dot var; _ } :: offset' -> Some (Var var, List.rev offset')
+        | [] -> None
```
emjin: So it looks like this case wasn't possible before but now it is. Confused about why, or am I reading it wrong?

semgrep-core/src/tainting/Taint_lval_env.ml
```diff
@@ -69,36 +69,50 @@ let union le1 le2 =
...
-          (* Offset-chain is not of the form .a_1. ... . a_N, so drop the last
-           * offset until the condition is met or we reach the base case. *)
-          normalize_lval { base; rev_offset = rev_offset' }
-      | Mem _ -> None)
+  let* base, rev_offset =
+    match base with
+    | Var _ -> Some (base, rev_offset)
+    | VarSpecial _ -> (
+        (* this.x.a_1. ... . a_N becomes x.a_1. ... . a_N *)
+        match List.rev lval.rev_offset with
```
emjin: Nit: why not just do `rev_offset`, since you decomposed `lval`?

semgrep-core/src/tainting/Taint_lval_env.ml
```diff
@@ -69,36 +69,50 @@ let union le1 le2 =
...
+            logger#error "normalize_lval: Impossible happened";
+            None)
+    | Mem _ -> None
+  in
+  let rev_offset =
+    rev_offset
+    |> List.filter (fun o ->
+           match o.IL.o with
+           | IL.Dot _ -> true
+           | IL.Index _ -> false)
```
emjin, line 94: To confirm my understanding, is it accurate to say here `TODO make tainting through arrays index-sensitive`?

emjin (approved):
Nice! I'm wondering about whether we lose information in the transformation of `this.x` to `x`, but we did that before too.

semgrep-core/src/tainting/Taint_lval_env.ml
IagoAbal: I may not fully understand the potential risk that you are envisioning. But note that l-values in the IL are side-effect free, so this cannot apply to l-values involving method calls. In the case of `this.foo().a`, it has to be represented as `_tmp = this.foo ()` (here the l-value is `this.foo`), and then `_tmp.a`. Also note that l-values are only normalized within `Taint_lval_env` to track taint that is propagated via assignments, propagators, or side-effects. So if `this.foo()` is tainted in the sense that it is a method that returns taint, that is not handled here. We will check whether `this.foo` (normalized as just `foo`) is tainted by some `pattern-source`, but I'm not sure an inheritance check is relevant for that case?

semgrep-core/src/tainting/Taint_lval_env.ml
IagoAbal: Good catch!

semgrep-core/src/tainting/Taint_lval_env.ml
IagoAbal: Good catch, actually both cases are possible now. There used to be a guard checking that the offset was of the form `.a_1 ... .a_N`, that is why they were impossible cases. My bad, copy-pasted that snippet and then forgot to fix it properly.

semgrep-core/src/tainting/Taint_lval_env.ml
IagoAbal: Yes! We would need to check here for constant indexes, but there is also some extra work to do when looking up the lvalues. Since you may taint `x.a` and then look for `x.a[1]` and it should tell you it's tainted.

PR #6585, opened 2022-11-22
feat(js): allow angular JS dependencies to be named

brandonspark:
## What:
In JS, classes may have constructors, which take parameters. In Angular JS, there is a concept of "dependency injection", which allows these dependencies to be specified in the parameters to a class's constructors, when marked with an `@Injectable` or `@Component` decorator. These dependencies may then be referenced outside of the scope of the constructor, but within the class generally. We should change naming so these things work that way, which is different than regular JS.

## Why:
This way, we can produce findings in line with the actual behavior. In particular, Lewis needs it to write some rules.

## How:
I moved the `class_definition` stuff out of `kclass_definition` in `Naming_AST`, because I needed the decorators along with the class definition.

I added extra logic which searches all the top-level definitions in a class, when that class is marked with a decorator `@Injectable` or `@Component`. Then, all parameters to that constructor are added to the parameters that are in scope within the class itself, making the parameters to the constructor visible outside the constructor itself.

## Test plan:
`make test`

Closes PA-2223.

PR checklist:

### DIFF
--- a/semgrep-core/src/naming/Naming_AST.ml
+++ b/semgrep-core/src/naming/Naming_AST.ml
@@ -412,18 +412,16 @@ let params_of_parameters env xs =
    parameters.
    This also works for `@Component`.
 *)
-let js_add_angular_constructor_args env attrs defs =
-  let ( let* ) = Option.bind in
-  match
-    let* () =
-      List.find_map
-        (function
+let js_get_angular_constructor_args env attrs defs =
+  let is_injectable = List.exists 
+    (function
           | NamedAttr (_, Id ((("Injectable" | "Component"), _), _), _) ->
-              Some ()
-          | _ -> None)
+              true 
+          | _ -> false)
         attrs
-    in
-    List.find_map
+  in
+    defs
+  |> List.filter_map
       (function
         | {
             s =
@@ -431,13 +429,10 @@ let js_add_angular_constructor_args env attrs defs =
                 ( { name = EN (Id (("constructor", _), _)); _ },
                   FuncDef { fparams; _ } );
             _;
-          } ->
+          } when is_injectable ->
             Some (params_of_parameters env fparams)
         | _ -> None)
-      defs
-  with
-  | None -> []
-  | Some l -> l
+  |> List.concat
 
 let declare_var env lang id id_info ~explicit vinit vtype =
   let sid = H.gensym () in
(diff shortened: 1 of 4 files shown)


semgrep-core/src/naming/Naming_AST.ml
```diff
@@ -402,6 +403,42 @@ let params_of_parameters env xs =
...
+(* In Angular JS, we have some "Injectable" classes, which are marked with an
+   @Injectable decorator.
+   https://angular.io/guide/dependency-injection-in-action
+   These classes may reference parameters to the constructor of the class, outside
+   of the actual code of the constructor itself.
+   So we must add them to the scope, should we find the decorator and a constructor's
+   parameters.
+   This also works for `@Component`.
+*)
+let js_add_angular_constructor_args env attrs defs =
```
IagoAbal: Maybe `get` instead of `add`, or is this function "adding" something somewhere?

semgrep-core/src/naming/Naming_AST.ml
```diff
@@ -402,6 +403,42 @@ let params_of_parameters env xs =
...
+  let ( let* ) = Option.bind in
+  match
+    let* () =
```
IagoAbal: This feels overcomplicated, why not `let is_injectable = List.exists ...`.  What is the point of using `List.find_map` and return `Some ()` ?

IagoAbal: And then we could also take this outside of the `match ... with`.

semgrep-core/src/naming/Naming_AST.ml
```diff
@@ -402,6 +403,42 @@ let params_of_parameters env xs =
...
+      List.find_map
+        (function
+          | NamedAttr (_, Id ((("Injectable" | "Component"), _), _), _) ->
+              Some ()
+          | _ -> None)
+        attrs
+    in
+    List.find_map
```
IagoAbal: If you used `List.filter_map` instead of `List.find_map` you wouldn't need the `match ... with`.

IagoAbal (approved):

semgrep-core/src/naming/Naming_AST.ml
```diff
@@ -455,18 +492,25 @@ let resolve lang prog =
...
         (fun (k, _v) x ->
           match x with
+          | { attrs; _ }, ClassDef c ->
+              (* todo: we should first process all fields in the class before
+                 * processing the methods, as some languages may allow forward ref.
+              *)
+              let class_params = params_of_parameters env c.cparams in
+              with_new_context InClass env (fun () ->
+                  let special_class_params =
+                    if lang = Lang.Js || lang = Lang.Ts then
```
IagoAbal: I think `Lang.is_js lang` means the same as `lang = Lang.Js || lang = Lang.Ts`.

semgrep-core/src/naming/Naming_AST.ml
brandonspark: good point

PR #5240, opened 2022-05-17
tainting: Allow variables to be taint sources by side-effect

IagoAbal:
Closes PA-1283

test plan:
make test # two tests added

PR checklist:

- [ ] Documentation is up-to-date
- [x] Changelog is up-to-date
- [x] Change has no security implications (otherwise, ping security team)

PR #6599, opened 2022-11-23
tainting: Assume sources and sanitizers are pure

IagoAbal:
Pure in the sense of having no side-effects.

In 0.94.0 we made it so when a pattern-source matched a variable exactly
this was understood as that variable being tainted by side-effect, e.g.
given `tainted(x)` we would taint `x` by side-effect. Same was done for
sanitizers.

This worked well initially, or so we thought, until in 0.113.0 we added
field-sensitivity, and in subsequent versions we applied sources and
sanitizers by side-effect on more general l-values. It was then that we
started to see regressions that were fairly unintuitive for users (see
added tests).

We now correct the situation by adding the `by-side-effect` option to
sources and sanitizers, and requiring this option to be explicitly enabled
in order to apply the source or the sanitizer by side-effect.

Requires https://github.com/returntocorp/semgrep-rules/pull/2577

Fixes: https://github.com/returntocorp/semgrep/commit/653db7dfa787b2e6057ee65dbf1ddb7c16b5bea0 ("tainting: Allow variables to be taint sources by side-effect (https://github.com/returntocorp/semgrep/pull/5240)")
Fixes: https://github.com/returntocorp/semgrep/commit/4ced4711358a6f19da87b59d9e7b3e69070bcdd1 ("tainting: Allow variables to be sanitized by side-effect (https://github.com/returntocorp/semgrep/pull/5266)")

Closes [PA-1629](https://linear.app/r2c/issue/PA-1629)
Closes [PA-2229](https://linear.app/r2c/issue/PA-2229)

Reviewed-by: Emma Jin

test plan:
make test # added three tests

PR checklist:

### DIFF
--- a/semgrep-core/src/parsing/Parse_rule.ml
+++ b/semgrep-core/src/parsing/Parse_rule.ml
@@ -1078,6 +1078,10 @@ let parse_taint_source ~(is_old : bool) env (key : key) (value : G.expr) :
     if is_old then parse_formula_old_from_dict else parse_formula_from_dict
   in
   let source_dict = yaml_to_dict env key value in
+  let source_by_side_effect =
+    take_opt source_dict env parse_bool "by-side-effect"
+    |> Option.value ~default:false
+  in
   let label =
     take_opt source_dict env parse_string "label"
     |> Option.value ~default:R.default_source_label
@@ -1088,7 +1092,7 @@ let parse_taint_source ~(is_old : bool) env (key : key) (value : G.expr) :
     |> Option.value ~default:(R.default_source_requires tok)
   in
   let source_formula = f env source_dict in
-  { source_formula; label; source_requires }
+  { source_formula; source_by_side_effect; label; source_requires }
 
 let parse_taint_propagator ~(is_old : bool) env (key : key) (value : G.expr) :
     Rule.taint_propagator =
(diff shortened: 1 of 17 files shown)


semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -1103,13 +1106,17 @@ let parse_taint_sanitizer ~(is_old : bool) env (key : key) (value : G.expr) =
     if is_old then parse_formula_old_from_dict else parse_formula_from_dict
   in
   let sanitizer_dict = yaml_to_dict env key value in
+  let sanitizer_mutate =
+    take_opt sanitizer_dict env parse_bool "mutate"
```
emjin: I'm assuming we're changing this from `"mutate"` to something else

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -212,6 +212,11 @@ let merge_source_sink_mvars env source_mvars sink_mvars =
     (* The union of both sets, but taking the sink mvars in case of collision. *)
     sink_biased_union_mvars source_mvars sink_mvars
 
+let partition_mutable_sources sources_matches =
```
emjin: Personally I think `mutating_sources` is better? Mutable to me sounds like we're trying to identify if the source is mutable (e.g. a list in Python)

emjin (approved):
In the changelog, it might be nice to give some examples of this unintuitive behavior to help people understand what this change entails. I don't think it's really easy to understand from a summary

semgrep-core/src/parsing/Parse_rule.ml
IagoAbal: Yes, finally it's going to be `taint-by-side-effect`.

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: Agree, I will rename it.

PR #6603, opened 2022-11-24
Check whether semgrep-proprietary still builds

mjambon:
See description in the job config.

Update: this PR is now only about using the script in CI. The script in question is now part of https://github.com/returntocorp/semgrep/pull/6664.

PR checklist:


mjambon:
The alpine image we use doesn't come with `ssh` or with `gh`. We can run `apk add openssh` but what about `gh`? It would need to be installed from a tarball as described [here](https://github.com/cli/cli/issues/658#issuecomment-701375745). It's not a huge deal but I'd like to know if there's a way to post a PR comment without installing `gh` into our docker container.

scripts/build-semgrep-proprietary
```diff
@@ -0,0 +1,46 @@
...
+$(cd "$semgrep_root" && git status)
+EOF
+
+# Use a folder where it's safe to delete any existing semgrep-proprietary.
+workdir=$semgrep_root/../tmp-semgrep-pro
+mkdir -p "$workdir"
+(
+  cd "$workdir"
+  rm -rf semgrep-proprietary
+  git clone "$semgrep_pro_url" --depth 1
```
spencerdrak: I think auth might need to be configured here, since `semgrep-proprietary` is a private repo.

spencerdrak (changes requested):
Overall, this looks good, but there's three comments I have:
- I think it would be best to wait until after we reelase 1.0.0 to merge this in, so our release process freeze has ended
- The auth (as noted) might need some work. LMK if you need some help with that
- I would strongly suggest testing this in our test GH Actions repository. I am happy to explain how that works, or even better, I can find some time to put together a better set of documents around that. I'll touch base when I get those into Notion.

aryx:
Closing; Someone can reopen when it works.
It just clutters our list of pull requests. We must keep it clean.

PR #6624, opened 2022-11-28
semgrep ci runs DeepSemgrep when it is enabled in App

IagoAbal:
Closes PA-2226

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -320,6 +320,10 @@ def ci(
                 logger.info(f"Could not start scan {e}")
                 sys.exit(FATAL_EXIT_CODE)
 
+            # Run DeepSemgrep when available but only for full scans
+            is_full_scan = metadata.merge_base_ref is None
+            deep = scan_handler.deepsemgrep if scan_handler and is_full_scan else False
+
             # Append ignores configured on semgrep.dev
             requested_excludes = scan_handler.ignore_patterns if scan_handler else []
             if requested_excludes:
@@ -346,6 +350,7 @@ def ci(
                 lockfile_scan_info,
             ) = semgrep.semgrep_main.main(
                 core_opts_str=core_opts,
+                deep=deep,
                 output_handler=output_handler,
                 target=[os.curdir],  # semgrep ci only scans cwd
                 pattern=None,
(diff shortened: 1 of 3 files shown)


emjin:
Is it ok to only run deep-semgrep? My concern is that customers might rely on the nightly scans results. I personally think I would expect the toggle to run semgrep and also deep-semgrep

IagoAbal:
I realized I probably need to add some check to avoid running DeepSemgrep in PR scans.

IagoAbal:
> Is it ok to only run deep-semgrep? My concern is that customers might rely on the nightly scans results. I personally think I would expect the toggle to run semgrep and also deep-semgrep

It's as valid concern that we should address for the DeepSemgrep launch. Although there is also an ongoing discussion about it, and whether this should be selected in a per-workflow basis? Right now this will only be run by a selected group of orgs anyways, so I believe it's fine this way. What I did is to restrict it to full scans.

brendongo (approved):
LGTM. do we fail gracefully when deepsemgrep is used but has not been installed correctly yet?

cli/src/semgrep/commands/ci.py
```diff
@@ -346,6 +350,7 @@ def ci(
                 lockfile_scan_info,
             ) = semgrep.semgrep_main.main(
                 core_opts_str=core_opts,
+                deep=deep,
```
emjin, line 353: Could we also add `--time` when we add `--deep` so that we can get metrics? In addition to timings, Brandon is adding the max heap size into the time output.

It's possible this just happens on every app run (I'm not sure what the current state is), so if so we don't need to change anything

IagoAbal:
> LGTM. do we fail gracefully when deepsemgrep is used but has not been installed correctly yet?

This was done already but I believe we do, I get:
```
[ERROR] Could not run deep analysis: DeepSemgrep not installed. Run `semgrep install-deep-semgrep`
```

cli/src/semgrep/commands/ci.py
IagoAbal: It's a good point but let's discuss this as a separate issue.

emjin:
Yay!! Let's start dogfooding this after next release!!!

PR #6626, opened 2022-11-28
fixed(taint): dededuplicate taint findings with same sink, prevent overlapping autofixes

brandonspark:
## What:
Our `no_submatches` pass on taint findings permits removal of matches with exactly the same sink. This should instead be a strict inclusion, otherwise we can delete findings which happen to have the same sink, which is common.

This introduces a bug with our tests where autofixes with the same sink may happen, which will apply twice. When overlapping autofixes occur, they just paste them next to (or on top of) each other. This is a bad thing, since this will almost always wreck the code that is produced. This PR adds in the capability in the CLI to deduplicate matches which overlap, which will allow more consistent and better behavior from Semgrep.

## Why:
This will allow us to report numerous findings when a function is invoked that causes a taint finding multiple times, from the same sink. This will also prevent the extra (likely) duplicate findings that are allowed from wrecking each other in autofix.

## How:
Added strict inclusion for range, and had that be used instead of the existing submatch function.

Also added code to the CLI which arbitrarily picks a match from two matches which overlap, when deciding on matches to autofix. Only one such autofix will be applied.

## Test plan:
`make test`

Test plan is in DeepSemgrep.

Fixes PA-2208

PR checklist:


changelog.d/pa-2208.fixed
```diff
@@ -0,0 +1 @@
+Taint: Fixed an issue where findings with the same sink would only report one finding
```
IagoAbal: It could be explained a bit better using some more words? I understand what it means but I'm not sure @adamkvitek will understand it. Think that we write changelog entries for @bmahe and @adamkvitek.

IagoAbal (approved):

IagoAbal (approved):
We need to check that 1) the CLI deduplicates such findings when _not_ using `--dataflow-traces` and 2) the CLI does _not_ deduplicate such findings when using `--dataflow-traces`. Could you confirm this?

brandonspark:
> We need to check that 1) the CLI deduplicates such findings when _not_ using `--dataflow-traces` and 2) the CLI does _not_ deduplicate such findings when using `--dataflow-traces`. Could you confirm this?

me: Why is that necessary? Surely the CLI doesn't do deduplication
CLI: *does deduplication*

good call...

cli/src/semgrep/rule_match.py
```diff
@@ -190,6 +190,10 @@ def get_cli_unique_key(self) -> Tuple:
             self.start.offset,
             self.end.offset,
             self.message,
+            # This is necessary so we don't deduplicate taint findings which have different sources.
```
IagoAbal, line 193: I'm just a bit concerned that this could introduce duplicate findings for those that are not yet using `--dataflow-traces` like basically everybody using App.

emjin: This is a good point 🤔. Maybe we can talk to Lewis about the expected behavior for CLI.

brandonspark: talked to Lewis, and we believe that the best way to proceed is to duplicate findings in all cases. This will better the user experience (no fixing a finding, and then finding out there's still a finding), as well as being more in line with the JSON output.

PR #6628, opened 2022-11-28
refactor: Add `Require` node to generic AST

nmote:
Currently, in JS, `require('foo')` is represented using `OtherExpr` for the `require` part of that expression. This is fine currently since we don't do any analysis of `require` calls after the AST generic translation. Instead, we desugar them to `ImportFrom` in some cases, and Semgrep's existing naming logic for imported symbols takes it from there.

However, that desugaring is causing some issues. See #6532 for details.

Instead of desugaring, we need to represent `Require` in the AST properly, and analyze it directly. We could simply treat it like a normal function call to a function named `require`, but it's a separate node in the JS AST and doing analysis by matching on specific literal values is more error-prone.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/semgrep-core/src/core/ast/AST_generic.ml
+++ b/semgrep-core/src/core/ast/AST_generic.ml
@@ -748,6 +748,11 @@ and special =
   | Op of operator
   (* less: should be lift up and transformed in Assign at stmt level *)
   | IncrDecr of (incr_decr * prefix_postfix)
+  (* JS: `require('foo')`. Calls to require are different than imports as
+   * represented by e.g. `ImportFrom`. They are expressions rather than top
+   * level statements, and can therefore appear inline in any expression, so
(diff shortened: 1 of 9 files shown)


semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -748,6 +748,7 @@ and special =
   | Op of operator
   (* less: should be lift up and transformed in Assign at stmt level *)
   | IncrDecr of (incr_decr * prefix_postfix)
+  | Require
```
aryx, line 755: Can you add a comment explaining why we need this construct given we already have ImportFrom?
Maybe something like "This is similar to ImportFrom, but at the expression level; in languages such as Javascript or Ruby import/require can appear at any places, not just at the toplevel".

aryx (approved):

PR #6630, opened 2022-11-29
fixed(cli): default `semgrep --deep` jobs to 1

brandonspark:
## What:
`semgrep` defaults the number of jobs to the number of cores on the machine, when not specified. This can cause OOM when using DeepSemgrep, due to the number of jobs taking a lot of memory.

## Why:
We should default it to 1 instead when using `semgrep --deep`, at least for now, while memory usage is unstable.

## How:
I changed it so jobs no longer default, and we set the number of jobs to `1` if `--deep` is specified, and the cpu counts otherwise.

## Test plan:
I tested this manually by printing out the values. The output is as follows:
image

Where `before jobs` denotes what `jobs` was prior to assignment, and `after jobs` for what it was assigned to after.

Closes PA-2231.

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/scan.py
+++ b/cli/src/semgrep/commands/scan.py
@@ -315,10 +315,9 @@ _scan_options: List[Callable] = [
         "-j",
         "--jobs",
         type=int,
-        default=__get_cpu_count(),
         help="""
             Number of subprocesses to use to run checks in parallel. Defaults to the
-            number of cores on the system.
+            number of cores on the system (1 if using --deep).
         """,
     ),
     optgroup.option(
@@ -625,7 +624,7 @@ def scan(
     gitlab_sast: bool,
     gitlab_secrets: bool,
     include: Optional[Tuple[str, ...]],
-    jobs: int,
+    jobs: Optional[int],
     json: bool,
     junit_xml: bool,
     lang: Optional[str],
@@ -711,6 +710,10 @@ def scan(
         abort(
             "Cannot create auto config when metrics are off. Please allow metrics or run with a specific config."
         )
+
+    print("before jobs", jobs)
+    jobs = jobs if jobs else 1 if deep else __get_cpu_count()
+    print("after jobs", jobs)
 
     output_time = time_flag
 
--- a/cli/src/semgrep/commands/scan.py
+++ b/cli/src/semgrep/commands/scan.py
@@ -1,4 +1,3 @@
-import multiprocessing
 import os
 import tempfile
 from itertools import chain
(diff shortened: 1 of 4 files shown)


cli/src/semgrep/commands/scan.py
```diff
@@ -712,6 +711,8 @@ def scan(
             "Cannot create auto config when metrics are off. Please allow metrics or run with a specific config."
         )
 
+    jobs = jobs if jobs else 1 if deep else __get_cpu_count()
```
brendongo: I think this will only set the job to 1 when running `semgrep scan --deep` but not `semgrep ci` and will also avoid setting jobs to 1 https://github.com/returntocorp/semgrep/pull/6624/files

Might be worth setting this further down and closer to deepsemgrep invocation?

cli/src/semgrep/core_runner.py
```diff
@@ -508,14 +516,15 @@ class CoreRunner:
...
+        jobs: Optional[int],
+        deep: bool,
         timeout: int,
         max_memory: int,
         timeout_threshold: int,
         optimizations: str,
         core_opts_str: Optional[str],
     ):
-        self._jobs = jobs
+        self._jobs = jobs if jobs else 1 if deep else __get_cpu_count()
```
brendongo: This is probably fine but would be cleaner to keep get_cpu_count where it was. Separation of concerns. The command line flag parsing at the top level knows what the user wants / the computer default is. Then in this layer we choose to follow or reject that recommendation. (I would also print out something like "number of jobs set to X but setting to 1 because running with --deep"

brendongo: Doing this also has the added benefit of jobs not being undefined all the way down

brandonspark: I don't quite understand. If we pass down something which isn't `Optional`, then if we use the `get_cpu_count` up at `scan` time, at the time we produce the `core_runner` we don't know if this recommendation is something the user wants, or if it's the computer default. Both cases behave differently (in the former, we want to take the user's input, and in the latter, we want to override the computer default if `--deep` is present), but they look the same. That's why I have the `Optional`, to disambiguate the case where there is no user input.

My main confusion is that I don't see how we can make the distinction (follow or reject the recommendation) with only a singular integer of data. Or are you saying that `__get_cpu_count` should be moved back, but still used within the context of `core_runner`? My impression was you were saying that it should both be moved back and used at `scan` time.

brendongo: I was mostly suggesting the `--deep` override happen at point closest to semgrep-core invocation 

Don't we override both user input or get_cpu_count when --deep is used so we dont really need to be able to disambiguate 

dont feel too strongly either way on where __get_cpu_count happens :D approving PR

brendongo (approved):

cli/src/semgrep/core_runner.py
brandonspark: oh I think that the confusion is that the desired behavior, as I understand it from Emma, is that we only override if _both_ `deep` is specified and `-j` was NOT passed in. If people really want to run multicore with DeepSemgrep, if they explicitly say to do so, then we should.

brendongo: AH i see.

PR #6635, opened 2022-11-29
fix(semdep): Stop marking yarn.lock findings as transitive

underyx:
PR checklist:

### DIFF
--- a/cli/src/semdep/parse_lockfile.py
+++ b/cli/src/semdep/parse_lockfile.py
@@ -136,7 +136,7 @@ def parse_yarn1(
         version, line_number = get_version(lines)
         resolved = get_resolved(lines)
         integrity = get_integrity(lines)
-        if manifest_deps is None:
+        if not manifest_deps:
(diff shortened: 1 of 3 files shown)


cli/src/semdep/parse_lockfile.py
```diff
@@ -136,7 +136,7 @@ def remove_trailing_octothorpe(s: str) -> str:
         version, line_number = get_version(lines)
         resolved = get_resolved(lines)
         integrity = get_integrity(lines)
-        if manifest_deps is None:
+        if not manifest_deps:
```
underyx, line 139: We have no package.json parser so `manifest_deps` is always an empty list.

brendongo (dismissed):

cli/src/semdep/parse_lockfile.py
brendongo: dont we parse it in line 122 above?

brendongo (approved):

PR #6636, opened 2022-11-29
Fix(js): Remove desugaring for JS requires

nmote:
## Background

My end goal here is to add CommonJS import/export support to DeepSemgrep. To do this, I need some consistency in how requires are represented.

Currently, we desugar `require` calls to `ImportFrom` nodes in one specific case. This only occurs when the `require` is on the RHS of a variable assignment, and the LHS is a destructuring, e.g. `const { f1, f2 } = require("file");`. However, instead of replacing the `require`, the `ImportFrom` appears alongside it. I believe that this was done because there isn't a way to consistently transform require calls into `ImportFrom`. `require` calls are simply expressions, and can appear anywhere. There is also the potential for complicated destructuring where Semgrep bails out from desugaring at all. Thus, to be able to consistently match a pattern such as `require('file')` we need to keep the original node around.

There is no equivalent desugaring to `ImportAs` for `const f = require("file");`, thus the TODO item in the `aliasing_require` test that is addressed by this change.

Because there is no way to map all `require` calls to the existing import nodes that we have, we need to remove this partial desugaring if we want to build a maintainable analysis.

## Implementation

This removes the code that does the desugaring in `js_to_generic.ml`. With that desugaring gone, we need to make sure that both naming and matching behave as desired, so I've added code to handle `require` calls to both `Naming_AST.ml` and `Generic_vs_generic.ml`.

Some things for reviewers to focus on:
- I added `m_directive_vs_def`. This performs matching on two different kinds of AST nodes. I couldn't find a precedent for this. Is this the right thing to do? Thoughts on naming? For matching partial nodes we generate synthetic nodes in the visitor. Maybe we should use that approach here? I think this is clearer, though.
- Related to the above, I added the `general_matcher` type which takes two type parameters so that `m_list_in_any_order` can operate on a matcher that matches a different type on the left and the right.

## Test plan

Automated tests

PR checklist:

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -2253,6 +2253,7 @@ and m_stmt a b =
       m_condition a1 b1 >>= fun () -> m_stmt a2 b2
   | G.DefStmt a1, B.DefStmt b1 -> m_definition a1 b1
   | G.DirectiveStmt a1, B.DirectiveStmt b1 -> m_directive a1 b1
+  | G.DirectiveStmt a1, B.DefStmt b1 -> m_directive_vs_def a1 b1
   | G.DoWhile (a0, a1, a2), B.DoWhile (b0, b1, b2) ->
       m_tok a0 b0 >>= fun () ->
       m_stmt a1 b1 >>= fun () -> m_expr a2 b2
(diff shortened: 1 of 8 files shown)


nmote:
This grew out of https://github.com/returntocorp/semgrep/pull/6532

emjin:
I was hoping to get to this today, but I still need to think about alternatives a bit more. Current impression: I'm definitely not a fan; I don't like having the DefStmt match the DirectiveStmt. However, it's fair to argue that this is because Javascript's semantics are themselves not nice. Some thoughts:

1. One alternative is to translate `require(...)` into an `ExprStmt(DirectiveStmt(ImportSpecial("require", ...)))` instead of a `Call(IdSpecial("require", ...), ...)`. This would allow us to not have to do the cross-node comparison and make it harder to forget about this special case. My biggest concern with the current implementation is that it would be extremely easy to implement import logic and forget about `require(...)`. On the flip side, I see where this alternative is ugly
2. Do we know if any other languages have this pattern? If so, maybe we could make it more first class with an expression node for `ExpressionImports` or something, which `require` falls under. We still have the special case problem, so this is just something to help if there are multiple languages that have some call that behaves like this

Definitely want thoughts from @aryx before merging

nmote:
I've run into so many issues stemming from desugaring so far that my instinct is to faithfully represent the original source code when in doubt. We'll have to add some new nodes to represent this anyway, so my belief is that it's better to use the Call so that it's closer to the original source.

With that said, I think your ideas are reasonable. My main concern is that we'll lose the ability for `$F(...)` to match `require('y')`, which at least one test (maybe in semgrep-rules? I ran into it, but don't remember now) relies upon: https://semgrep.dev/s/319g

Of course, we could also match a call pattern against an `ImportSpecial` node or whatever it ends up being, but that's no better than what I have now.

Do you think it's worth losing that matching behavior?

I'm not aware of other languages with this behavior. I think you can put a Ruby `require` anywhere, but it works by adding bindings to the global scope rather than introducing local bindings at the point where it is called, so for analysis purposes we don't have to treat it so carefully.

emjin:
Ah, I didn't realize that people would still want to match it as a call. In that case, I guess this does make the most sense 😞.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -3056,6 +3057,85 @@ and m_directive a b =
   | G.OtherDirective _ ->
       fail ()
 
+and m_directive_vs_def a b =
```
emjin, line 3089: I might call this something more specific like `m_directive_vs_import_def` and add another comment explaining some of the tradeoffs considered, so that people are thoughtful about adding more cases to this function

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -3056,6 +3057,85 @@ and m_directive a b =
...
+  let f (filea, importsa) = function
+    (* const x = require('y');
+     *
+     * This is for JS but there is no need to gate on language, since the
```
emjin: imo we might as well gate on JS anyway, it doesn't make the code harder to read and maybe someone will add some language that uses `require` as a CallSpecial but doesn't want this behavior

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -3056,6 +3057,85 @@ and m_directive a b =
   | G.OtherDirective _ ->
       fail ()
 
+and m_directive_vs_def a b =
+  let f (filea, importsa) = function
+    (* const x = require('y');
```
emjin: Does it have to be specifically `const` or is that just an example?

emjin: I personally would just leave both what `a` and `b` should look like. So, for this case, I would say that this matches `const x = require('y')` to `import a`. It just takes me a little longer to notice that we're checking the value of `importsa`, and then I have to scroll to see what it means for `importsa` to be empty

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -3056,6 +3057,85 @@ and m_directive a b =
...
+                };
+            vtype = _;
+          } )
+      when importsa = [] ->
+        (* Match the pattern `import "foo"` against `const x = require("foo")` *)
+        m_wrap m_string filea fileb
+    (* const {x, y} = require('z');
+     *
+     * This is for JS but there is no need to gate on language, since the
+     * `Require` node is unique to JS. *)
```
emjin: Same comments here as the other case

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -3056,6 +3057,85 @@ and m_directive a b =
...
+    | _ -> fail ()
+  in
+  match a.d with
+  (* JS: `import {x, y as z} from 'a';` *)
+  | G.ImportFrom (_, G.FileName filea, importsa) -> f (filea, importsa) b
+  (* JS: `import 'a';` *)
+  | G.ImportAs (_, G.FileName filea, None) -> f (filea, []) b
+  | _ -> fail ()
+
+and m_import_vs_field a b =
```
emjin, line 3159: Maybe gate this under javascript or add a comment specifying that? It's pretty close to the use, so not a big deal, just thinking that if we ever find another language with similar semantics it might not do it the same way

semgrep-core/src/naming/Naming_AST.ml
```diff
@@ -506,6 +506,89 @@ let resolve lang prog =
                   (* TODO? Maybe we need a `with_new_class_scope`. For now, abusing `with_new_function_scope`. *)
                   with_new_function_scope (special_class_params @ class_params)
                     env.names (fun () -> k x))
+          (* const x = require('y');
```
emjin: I would clarify that `require('y')` in Javascript behaves as a local import here

semgrep-core/src/naming/Naming_AST.ml
```diff
@@ -506,6 +506,89 @@ let resolve lang prog =
...
+           *
+           * This is for JS but there is no need to gate on language, since the
```
emjin: Same as before; I see no reason not to gate on language if we can

emjin (approved):
Given that this is the approach we're going with, looks good. Could use some more comments explaining the semantics of require for the poor souls that debug issues with it later

semgrep-core/src/matching/Generic_vs_generic.ml
nmote: The other matching functions use the type of the arguments as part of the name, e.g. `m_directive` takes two directives, `m_definition` takes two definitions. This takes a `directive` and a `definition` so while it's true that it only returns a match on `definition`s that are equivalent to imports, I think the current name is more consistent with the existing naming scheme.

I will definitely add a comment here, though.

semgrep-core/src/matching/Generic_vs_generic.ml
nmote: That's just an example. I will clarify.

nmote: Good point

nmote (commented):
Thanks as always for the thorough review!

PR #6639, opened 2022-11-29
fix(deep): resolve types of function calls made with Id

emjin:
Supports deep-semgrep

Note: as I mention in the other PR, I restricted this to Java to be conservative but I think we should just always
use the type if we have it

Test plan: see PR in that repo

PR checklist:

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -1357,18 +1357,25 @@ and type_of_expr lang e : (G.ident * G.type_ option) option =
   (* deep: same *)
   | B.Call
       ( { e = B.DotAccess (_, _, FN (Id (idb, { B.id_type = tb; _ }))); _ },
-        _args )
-  (* deep: in Java, there can be an implicit `this.`
-     so look this up the type here too *)
-  | B.Call ({ e = N (Id (idb, { B.id_type = tb; _ })); _ }, _args)
-    when lang = Lang.Java -> (
+        _args ) -> (
       match !tb with
       (* less: in OCaml functions can be curried, so we need to match
        * _params and _args to calculate the resulting type.
        *)
       | Some { t = TyFun (_params, tret); _ } -> Some (idb, Some tret)
-      | Some _ -> None
-      | None -> None)
+      | Some _
+      | None ->
+          None)
+  (* deep: in Java, there can be an implicit `this.`
+     so calculate the type in the same way as above
+     THINK: should we do this for all languages? Why not? *)
+  | B.Call ({ e = N (Id (idb, { B.id_type = tb; _ })); _ }, _args)
+    when lang = Lang.Java -> (
+      match !tb with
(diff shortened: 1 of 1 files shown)


semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1357,15 +1357,19 @@ and type_of_expr e : (G.ident * G.type_ option) option =
   (* deep: same *)
   | B.Call
       ( { e = B.DotAccess (_, _, FN (Id (idb, { B.id_type = tb; _ }))); _ },
-        _args ) -> (
+        _args )
+  (* deep: in Java, there can be an implicit `this.`
+     so look this up the type here too *)
+  | B.Call ({ e = N (Id (idb, { B.id_type = tb; _ })); _ }, _args)
+    when lang = Lang.Java -> (
```
nmote: This `when` clause will apply to every part of the pattern match that uses this case. It doesn't sound like that was your intent? If so, could you add a clarification comment?

```
let f z = function
  | 1
  | 2 when z -> print_endline "first"
  | _ -> print_endline "second"
;;

f false 1
```
prints "second"

nmote (changes requested):

nmote (approved):
thanks!

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1363,9 +1363,20 @@ and type_of_expr e : (G.ident * G.type_ option) option =
        * _params and _args to calculate the resulting type.
        *)
       | Some { t = TyFun (_params, tret); _ } -> Some (idb, Some tret)
-      | Some _ -> None
-      | None -> None)
-  | B.ParenExpr (_, e, _) -> type_of_expr e
+      | Some _
+      | None ->
+          None)
+  (* deep: in Java, there can be an implicit `this.`
```
IagoAbal, line 1369: This is not really DeepSemgrep specific, I think.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1363,9 +1363,20 @@ and type_of_expr e : (G.ident * G.type_ option) option =
...
+     so calculate the type in the same way as above
+     THINK: should we do this for all languages? Why not? *)
```
IagoAbal, line 1371: I think we should.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -1363,9 +1363,20 @@ and type_of_expr e : (G.ident * G.type_ option) option =
...
+  | B.Call ({ e = N (Id (idb, { B.id_type = tb; _ })); _ }, _args)
+    when lang = Lang.Java -> (
+      match !tb with
+      | Some { t = TyFun (_params, tret); _ } -> Some (idb, Some tret)
```
IagoAbal, line 1375: But here we need to watch out that the function is fully applied perhaps, it may be a partial application. I guess this is not a problem in Java however.

PR #6640, opened 2022-11-29
tainting: Make identification of sinks more precise

IagoAbal:
This way we will not incorrectly report `sink(ok1 if tainted else ok2)` as a tainted sink. Taint-mode will no longer consider that `tainted` itself is being passed into a sink, but the whole if-expression.

Closes PA-2142

test plan:
make test # tests added

PR checklist:

### DIFF
--- a/changelog.d/pa-2142.fixed
+++ b/changelog.d/pa-2142.fixed
@@ -1,0 +1,3 @@
+taint-mode: Fixed matching of `pattern-sinks` to be more precise, so that e.g.
+it will no longer report `sink(ok1 if tainted else ok2)` as a tainted sink, as
+the expression passed to the `sink` is actually not tainted.
--- a/semgrep-core/src/tainting/Dataflow_tainting.ml
+++ b/semgrep-core/src/tainting/Dataflow_tainting.ml
@@ -65,7 +65,13 @@ module SMap = Map.Make (String)
 
 type var = Var_env.var
 type overlap = float
-type 'spec tmatch = { spec : 'spec; pm : PM.t; overlap : overlap }
+
+type 'spec tmatch = {
+  spec : 'spec;
+  spec_pm : PM.t;
+  range : Range.t;
+  overlap : overlap;
+}
 
 type a_propagator = {
   kind : [ `From | `To ];
@@ -76,7 +82,7 @@ type a_propagator = {
 type config = {
   filepath : Common.filename;
   rule_id : string;
-  is_source : G.any -> Rule.taint_source tmatch list;
+  is_source : G.any -> R.taint_source tmatch list;
   is_propagator : AST_generic.any -> a_propagator tmatch list;
   is_sink : G.any -> R.taint_sink tmatch list;
   is_sanitizer : G.any -> R.taint_sanitizer tmatch list;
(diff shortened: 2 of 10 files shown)


semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -94,12 +100,66 @@ type mapping = Lval_env.t D.mapping
 (* HACK: Tracks tainted functions intrafile. *)
 type fun_env = (var, Taints.t) Hashtbl.t
 
+(* NOTE "Top sinks":
+ * (See note "Taint-tracking via ranges" in 'Match_tainting_mode.ml' for context.)
+ *
+ * Since we use sub-range checks to determine wheter a piece of code is a sink this
```
emjin:
```suggestion
 * Since we use sub-range checks to determine whether a piece of code is a sink this
```

changelog.d/pa-2142.fixed
```diff
@@ -0,0 +1,3 @@
+taint-mode: Fixed matching of `pattern-sinks` to be more precise, so that e.g.
```
emjin, line 1: I notice in the issue you said that you were going to add an `exact: true` option to turn this on, but in the PR it looks like it's always on. Am curious why you changed your mind! (Always on seems correct to me, just wondering)

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -94,12 +100,66 @@ type mapping = Lval_env.t D.mapping
...
+ *
+ * Unfortunately we cannot simply check for exact matches, because sometimes we
+ * cannot match sinks exactly. So, what we do is to check the top-level nodes in the
+ * CFG for sinks, and from all those matches we keep the "best" (larger) fits. We
+ * call these the "top sinks" (matches). Then we use those top sinks as a practical
+ * definition of "exact match", thus if we find  another match which is a strict
+ * sub-range of one of these top sinks, we simply disregard it. In our example above
+ * the top sink match will be `sink(if tainted then ok1 else ok2)`, so we will
+ * disregard `tainted` as a sink because we know there is a better match.
+ *)
```
emjin, line 138: This is a great comment, thank you!

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -94,12 +100,66 @@ type mapping = Lval_env.t D.mapping
...
+  module S = Set.Make (struct
+    type t = R.taint_sink tmatch
+
+    let compare m1 m2 =
+      let sink_id_cmp = String.compare m1.spec.R.sink_id m2.spec.R.sink_id in
+      if sink_id_cmp <> 0 then sink_id_cmp
+      else
+        let r1 = m1.range in
+        let r2 = m2.range in
+        if Range.(r1 $<=$ r2 || r2 $<=$ r1) then 0 else Stdlib.compare r1 r2
```
emjin, line 151: I'm missing something; why do we return 0 if either `r1 <=  r2` or `r2 <= r1`?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -94,12 +100,66 @@ type mapping = Lval_env.t D.mapping
...
+
+  let empty = S.empty
+
+  let add m' sinks =
+    match S.find_opt m' sinks with
+    | None -> S.add m' sinks
+    | Some m ->
+        let r = m.range in
+        let r' = m'.range in
+        if r'.start > r.start || r'.end_ < r.end_ then
```
emjin, line 168: In the comment, you indicate that we disregard a match if it is a "strict sub-range" of one of the top sinks. However, this check just checks if either end is within the other. If this is what we want, the comment should be modified

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -94,12 +100,66 @@ type mapping = Lval_env.t D.mapping
...
+          (* The new match is a worse fit so we keep the current one. *)
+          sinks
+        else (* We found a better (larger) match! *)
```
emjin, line 171: Not strictly true; I believe you can have overlapping matches. Trying to construct an example: let's say the rule defines two sinks, `A.getSession(...)` and `(Session $X).run(...)`. The code has `A.getSession(arg1).run(arg2)` (where `A.getSession(...)` returns a new `Session`). In this case, `r'.start > r.start` but `r'.end_ > r.end_`.

I believe if we encounter an overlap we should always want to keep both sinks. I'm not certain, though, would need to spend more time thinking about not-totally-contrived cases where this could happen.

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -640,7 +726,9 @@ and check_tainted_expr env exp : Taints.t * Lval_env.t =
       (Taints.empty, lval_env)
   | None ->
       let sinks =
-        orig_is_sink env.config exp.eorig |> Common.map trace_of_match
+        orig_is_sink env.config exp.eorig
```
emjin: Are there other uses of `orig_is_sink`? It seems like this is at least the most common use, could we just make it so that `orig_is_sink` defaults to filtering for `best_match` and make the special case not filtering?

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -480,7 +562,11 @@ let rec check_tainted_lval env (lval : IL.lval) : Taints.t * Lval_env.t =
   let new_taints, lval_in_env, lval_env = check_tainted_lval_aux env lval in
   let taints_from_env = status_to_taints lval_in_env in
   let taints = Taints.union new_taints taints_from_env in
-  let sinks = lval_is_sink env.config lval |> Common.map trace_of_match in
+  let sinks =
+    lval_is_sink env.config lval
```
emjin, line 594: I saw the `orig_is_sink` one first but I have a similar comment here too; I'm not familiar enough to know if the default should be `is_best_match`

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -812,31 +918,40 @@ let input_env ~enter_env ~(flow : F.cfg) mapping ni =
...
           match opt_lval with
           | Some lval ->
-              (* We call `check_tainted_var` here because the assigned `var`
+              (* We call `check_tainted_lval` here because the assigned `lval`
                * itself could be annotated as a source of taint. *)
-              check_tainted_lval { env with lval_env = lval_env' } lval |> snd
+              let taints, lval_env' =
+                check_tainted_lval { env with lval_env = lval_env' } lval
+              in
+              (* We check if the instruction is a sink, and if we so the taints
```
emjin:
```suggestion
              (* We check if the instruction is a sink, and if so the taints
```

semgrep-core/src/tainting/Dataflow_tainting.ml
```diff
@@ -903,9 +1018,10 @@ let (fixpoint :
     | None -> Lval_env.empty
     | Some in_env -> in_env
   in
+  let top_sinks = top_level_sinks_in_nodes config flow in
```
emjin: I don't completely understand why we need to initialize this. Would appreciate if you explained! (It might not need a comment, I just want to understand and perhaps I'm a little tired)

emjin (changes requested):
I'm really excited for this! As a solution I think it makes a lot of sense. Requesting changes just to make sure I get clarity on what happens if there are overlapping ranges, and because it looks like the issue this solves is quite old.

I actually think this also solves one of the FPs GovTech [got](https://linear.app/r2c/issue/PA-2267/java-fp-sink-not-respecting-semantics-of-language) though! So I will definitely make sure I get back to this

changelog.d/pa-2142.fixed
IagoAbal: Updated `NOTE "Top sinks"`.

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: Added a comment.

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: Added a comment. But given `S`s definition, `r` is a sub-range of `r'` or vice-versa, so given that I believe that if `r'.start > r.start` then `r'` must be a strict sub-range of `r`.

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: If the rule defines two sinks, and one match is for one sink and another match is for the other sink, then you will never reach this case because the `sink_id`s will be different.

semgrep-core/src/tainting/Dataflow_tainting.ml
IagoAbal: I did some refactoring but we still need the "raw" `orig_is_sink`.

emjin (approved):
Thanks for the clarifications!

PR #6641, opened 2022-11-29
feat!: add memory usage to JSON output

brandonspark:
## What:
This PR adds a field to the JSON output (currently named `max_ocaml_heap_words`, which denotes the maximum number of words used by the OCaml major heap during execution of Semgrep. Notably, this output is only triggered on the  `-json_time` `semgrep-core` setting, and `--time` correspondingly for `semgrep`.

## Why:
This will be helpful, because this information can be passed up and used in our databases (particularly Metabase), so we can do things like plot memory usage against size of repository. This will help us benchmark memory, and determine how well we do when we make improvements in the future.

## How:
Added a field `max_ocaml_heap_words` to the JSON output of both Core and CLI, under the above conditions.

## Test plan:
`make test`

Observed that the `max_ocaml_heap_words` field only appears in `semgrep --time`

Closes PA-2075.

PR checklist:


semgrep-core/src/reporting/Report.ml
```diff
@@ -352,7 +353,12 @@ let make_final_result results rules ~rules_parse_time =
   let extra =
     let mk_profiling () =
       let file_times = results |> Common.map get_profiling in
-      { rules; rules_parse_time; file_times }
+      {
+        rules;
+        rules_parse_time;
+        file_times;
+        max_memory_bytes = (Gc.quick_stat ()).top_heap_words;
```
emjin: This is a good start. Let's leave a comment talking about some of the caveats we discussed. Most importantly, if the most amount of memory is consumed after forking, the real max needs to be multiplied by the number of processes.

emjin: To get bytes multiply by `Sys.word_size`

cli/src/semgrep/core_runner.py
```diff
@@ -871,6 +871,7 @@ def _run_rules_direct_to_semgrep_core(
 
             if ("time" in output_json) and core_output.time:
                 self._add_match_times(profiling_data, core_output.time)
+                profiling_data.set_max_memory_bytes(core_output.time.max_memory_bytes)
```
emjin: For consistency I would rather also have this set by a `self._add_max_memory_bytes()` function

emjin (approved):
Left some comments. Before merging, let's also test this by doing a deep-semgrep run with `--metrics on --time` and check metabase to see that `--max-memory-bytes` shows up, then comparing that number to roughly what we get from a local run. Not going to require a test because we can't verify the number stays correct and frankly that's what I care about

semgrep-core/src/reporting/Report.ml
emjin: (If you don't do that change it from `max_memory_bytes` to `max_memory_words` or at the very least pretend it's also a 64-bit machine and multiply by 8)

brandonspark:
image

it's in the raw blob being sent to metabase, though not as a column

PR #6643, opened 2022-11-30
fix(SSC): Fix a host of yarn2 parse errors

mmcqd:
This PR fixes the following issues:
* Parse failure on versions like `@storybook/react-docgen-typescript-plugin@canary`. This is only present as some kind special version range specifier and never appears as a concrete version. It would only be used to check if the dependency was in the manifest file, so we just parse the version as "canary"
* Parse failure on versions like `@types/ol-ext@npm:@siedlerchr/types-ol-ext@3.0.6`
* Parse failure on versions like `resolve@patch:resolve@^1.1.7#~builtin<compat/resolve>`. These are now just ignored, as they appear to always come with a non-patch version as well.

PR checklist:


brendongo (approved):

brendongo:
Lacking changelog entry

PR #6647, opened 2022-12-01
fix!: deduplicate overlapping autofixes

brandonspark:
## What:
#6626 introduced test cases where autofixes overlap, which is a possibility, especially with taint findings which are not deduplicated purely based on sinks. We should avoid multiple autofixes going and mangling the same space, which more likely than not will create invalid code.

## Why:
If Semgrep creates invalid code, users will not trust our autofixes.

## How:
I just ordered all matches and arbitrarily picked one, whenever two overlap.

PR checklist:

### DIFF
--- a/cli/src/semgrep/autofix.py
+++ b/cli/src/semgrep/autofix.py
@@ -1,11 +1,15 @@
 import re
+from functools import cmp_to_key
 from pathlib import Path
 from typing import Dict
+from typing import Iterable
 from typing import List
 from typing import Set
 from typing import Tuple
 
 from semgrep.error import SemgrepError
+from semgrep.rule import Rule
+from semgrep.rule_match import OrderedRuleMatchList
 from semgrep.rule_match import RuleMatch
 from semgrep.rule_match import RuleMatchMap
 from semgrep.util import unit_str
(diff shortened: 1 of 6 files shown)


cli/src/semgrep/autofix.py
```diff
@@ -130,53 +134,122 @@ def _write_contents(path: Path, contents: str) -> None:
     path.write_text(contents)
 
 
+def matches_compare(x: RuleMatch, y: RuleMatch) -> int:
+    # If paths are not the same, just order based on that.
+    # I just need a total ordering on matches.
+    if x.path < y.path:
```
emjin, line 140: I sometimes will go instead for
```
if x.path != y.path: 
    return x.path - y.path 
else:
    <next case:>
```
but I'm fine with this too. Mostly just mentioning it if you hadn't considered it

cli/src/semgrep/autofix.py
```diff
@@ -130,53 +134,122 @@ def _write_contents(path: Path, contents: str) -> None:
...
+
+    final_matches = []
+
+    ordered_matches = sorted(
+        (match for _, matches in rules_and_matches for match in matches),
+        key=cmp_to_key(matches_compare),
+    )
+    acc = None
+
+    for match in ordered_matches:
```
emjin, line 183: The second answer in https://stackoverflow.com/questions/10366374/what-is-the-pythonic-equivalent-to-the-fold-function-from-functional-program makes it seem like there might be a pythonic way to do this in 3.8 with the assignment operator, but sadly I think we want to support 3.7. I'm just grasping for straws really because this code so clearly wants to be a fold

emjin: @brendongo Do you know a pythonic way to do this?

emjin (approved):
Nice!

PR #6655, opened 2022-12-03
feat(cli, DS): add error message and suggestion to update

brandonspark:
## What:
Sometimes we release changes to Semgrep which break on older versions of DeepSemgrep. Since it's on users to update their versions of DeepSemgrep, we don't want them getting mad over it crashing because of this versioning issue. A bandaid solution is to make Semgrep suggest to update, whenever DeepSemgrep crashes.

## Why:
We don't want customers to get confused and frustrated with our product when DeepSemgrep cryptically crashes, which could conceivably happen to anyone who forgets to update DeepSemgrep (and who does that?)

## How:
Added a `try` which wraps the DeepSemgrep call, so that we can suggest that users update.

image

PR checklist:

### DIFF
--- a/cli/src/semgrep/core_runner.py
+++ b/cli/src/semgrep/core_runner.py
@@ -889,11 +889,12 @@ class CoreRunner:
                 try:
                     returncode = runner.execute()
                 except:
-                    logger.error("DeepSemgrep crashed during execution (unknown reason).\nTry updating to the latest version? (`semgrep --install-deep-semgrep`)")
+                    logger.error(
+                        "DeepSemgrep crashed during execution (unknown reason).\nTry updating to the latest version? (`semgrep --install-deep-semgrep`)"
+                    )
                     sys.exit(0)
             else:
                 returncode = runner.execute()
-
 
             # Process output
             output_json = self._extract_core_output(
--- a/cli/src/semgrep/core_runner.py
+++ b/cli/src/semgrep/core_runner.py
@@ -878,7 +878,22 @@ class CoreRunner:
 
             runner = StreamingSemgrepCore(cmd, plan.num_targets)
             runner.vfs_map = vfs_map
-            returncode = runner.execute()
+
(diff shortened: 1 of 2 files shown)


cli/src/semgrep/core_runner.py
```diff
@@ -878,7 +878,23 @@ def _run_rules_direct_to_semgrep_core(
...
+                # These issues may possibly cause a failure if a user, for instance, updates their
+                # version of Semgrep, but does not update to the latest version of DeepSemgrep.
+
+                # A short bandaid solution for now is to suggest that a user updates to the latest
+                # version, if the DeepSemgrep binary crashes for any reason.
+                try:
+                    returncode = runner.execute()
+                except:
+                    logger.error(
+                        "DeepSemgrep crashed during execution (unknown reason).\nTry updating to the latest version? (`semgrep --install-deep-semgrep`)"
```
emjin:
```suggestion
                        "DeepSemgrep crashed during execution (unknown reason).\n This can sometimes happen because either DeepSemgrep or Semgrep is out of date.\n Try updating your version of DeepSemgrep (`semgrep --install-deep-semgrep`)\n or your version of Semgrep (`pip install semgrep/brew install semgrep`).\n If both are up-to-date and the crash persists, please contact support to report an issue!"
```

cli/src/semgrep/core_runner.py
```diff
@@ -878,7 +878,23 @@ def _run_rules_direct_to_semgrep_core(
...
+                    )
+                    sys.exit(0)
```
emjin:
```suggestion
                    sys.exit(2)
```

emjin: Let's also print out the exception (maybe in gray), with some kind of label

emjin (commented):
We're wrapping this more widely to be safe

emjin (approved):

PR #6656, opened 2022-12-03
fixed!: fix yaml target metavariable parsing

brandonspark:
## What:
Currently, our YAML parser parses metavariables strangely in target files. Because the YAML pattern parser has to be able to use Semgrep metavariables, and the target parser is the same parser, it results in metavariables resulting in the target Generic AST. This causes issues with matching, where metavariables in YAML rules which match ranges that look like metavariables will end up with content that is just the empty string. 

This can be observed at the following Playground link (see the message): https://semgrep.dev/playground/s/brandonspark:yaml-metavariable-matching-metavariable

## Why:
This just doesn't make sense. We should be able to match the metavariable with the metavariable.

## How:
I added a flag to the YAML parser so it is aware of if we are parsing a target. If we are parsing a target, then it should treat metavariables it sees as strings.

## Test plan:
`make test`

PR checklist:


semgrep-core/src/osemgrep/cli_scan/Rule_fetching.ml
```diff
@@ -65,7 +65,7 @@ let parse_yaml_for_jsonnet (file : Common.filename) : AST_jsonnet.program =
   (* TODO? or use Yaml_to_generic.parse_yaml_file which seems
    * to be used to parse semgrep rules?
    *)
-  let gen = Yaml_to_generic.program file in
+  let gen = Yaml_to_generic.program ~is_target:false file in
```
aryx: this should be true

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -1402,11 +1402,11 @@ let parse_file ?error_recovery file =
               if n <> 0 then failwith (spf "error executing %s" cmd);
               let ast = Parse_json.parse_program tmpfile in
               Json_to_generic.program ~unescape_strings:true ast)
-    | FT.Config FT.Yaml -> parse_yaml_rule_file file
+    | FT.Config FT.Yaml -> parse_yaml_rule_file ~is_target:false file
```
aryx, line 1406: this should be true

semgrep-core/src/parsing/other/yaml_to_generic.mli
```diff
@@ -1,10 +1,11 @@
 (* Parsing a YAML file.
  * This may raise Parse_info.Other_error.
  *)
-val program : Common.filename -> AST_generic.program
+val program : is_target:bool -> Common.filename -> AST_generic.program
```
aryx: I don't think you need a label. When we use the program function, it's a target,
when we use any, it's always to parse a pattern.

aryx (changes requested):

semgrep-core/src/parsing/Parse_rule.ml
aryx: Why did you continue to use false?
We're parsing a rule, so the rule will not contain metavariables; only the patterns inside the rule will, but that's using a different function (Parse_pattern).

aryx (approved):

IagoAbal:
Bad idea to use mutable links like https://semgrep.dev/playground/s/brandonspark:yaml-metavariable-matching-metavariable in GitHub issues, better immutable ones: https://semgrep.dev/s/kKJ2.

semgrep-core/src/parsing/other/yaml_to_generic.mli
```diff
@@ -1,3 +1,8 @@
+(* Note!!
+   `program` and `any` will behave slightly differently on metavariables.
```
IagoAbal, line 2: I understand this comment now, but in X months I probably don't. Some of the explanation that went into the PR comment should probably go here too. I will be asking myself what is that "slightly different behavior on metavariables"?

PR #6668, opened 2022-12-06
feat(ojsonnet): desugar array comprehensions and eval specials

brandonspark:
## What:
Adds in desugaring of array comprehensions. Also adds in evaluation of some specials I added.

## Why:
Need it to make `ojsonnet`'s semantics work

## How:
I added two specials, `StdMakeArray` and `StdLength`, which are base functions whose semantics are given in the operational semantics. These are then cased on in `Eval_jsonnet`, which I also implemented the big-step dynamics of.

For the generation of fresh variable names, I just picked an unlikely variable name and appended increasing numbers to it. I realize this is disgusting, but in practice it will never come up. C'mon. Should have used de Bruijn indices if you didn't want me to do that.

Closes PA-2245

PR checklist:

### DIFF
--- a/semgrep-core/src/ojsonnet/interpreting/Core_jsonnet.ml
+++ b/semgrep-core/src/ojsonnet/interpreting/Core_jsonnet.ml
@@ -82,7 +82,7 @@ type expr =
 (* ------------------------------------------------------------------------- *)
 
 (* no Dollar anymore *)
-and special = Self | Super | StdLength | StdMakeArray
+and special = Self | Super
 
 (* the NamedArg are supposed to be the last arguments *)
 and argument = Arg of expr | NamedArg of ident * tok (* = *) * expr
--- a/semgrep-core/src/ojsonnet/interpreting/Core_jsonnet.ml
+++ b/semgrep-core/src/ojsonnet/interpreting/Core_jsonnet.ml
@@ -82,7 +82,7 @@ type expr =
 (* ------------------------------------------------------------------------- *)
 
 (* no Dollar anymore *)
-and special = Self | Super
+and special = Self | Super | StdLength | StdMakeArray
 
 (* the NamedArg are supposed to be the last arguments *)
 and argument = Arg of expr | NamedArg of ident * tok (* = *) * expr
--- a/semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
+++ b/semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
@@ -289,7 +289,7 @@ and desugar_comprehension_helper env expr comps =
       let inner_exp =
         match rest with
         | [] -> mk_array [ expr ]
-        | _ -> desugar_comprehension_helper env expr rest
+        | __else__ -> desugar_comprehension_helper env expr rest
       in
       If (tok, cond, inner_exp, empty_else)
   | CompFor (_, x, _, for_expr) :: rest ->
@@ -307,7 +307,7 @@ and desugar_comprehension_helper env expr comps =
         let inner_exp =
           match rest with
           | [] -> mk_array [ expr ]
-          | _ -> desugar_comprehension_helper env expr rest
+          | __else__ -> desugar_comprehension_helper env expr rest
         in
         let i = freshvar () in
         Lambda
@@ -327,7 +327,7 @@ and desugar_comprehension_helper env expr comps =
           [ B (arr, fk, for_expr) ],
           fk,
           std_join (mk_array [], std_mk_array (std_length (Id arr), f)) )
-  | _ -> failwith "lol"
+  | [] -> failwith "impossible: Empty array comprehension"
 
 and desugar_comprehension env expr comps =
   desugar_expr env (desugar_comprehension_helper env expr comps)
--- a/semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
+++ b/semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
@@ -66,7 +66,9 @@ let error tk s =
   raise (Error (s, tk))
 
 let fk = Parse_info.unsafe_fake_info ""
+let fb x = (fk, x, fk)
 let mk_str_literal (str, tk) = Str (None, DoubleQuote, (fk, [ (str, tk) ], fk))
+let mk_array exprs = A (fk, Array exprs, fk)
 
 let mk_DotAccess_std id : expr =
   let std_id = ("std", fk) in
@@ -76,6 +78,12 @@ and expr_or_null v : expr =
   match v with
   | None -> L (Null fk)
   | Some e -> e
+
+let freshvar =
+  let store = ref 0 in
+  fun () ->
+    incr store;
+    ("reallylongfreshvariablename" ^ string_of_int !store, fk)
 
 let todo _env _v = failwith "TODO"
 
@@ -215,6 +223,8 @@ and desugar_special _env v =
   match v with
   | Self -> C.Self
   | Super -> C.Super
+  | StdLength -> C.StdLength
+  | StdMakeArray -> C.StdMakeArray
   | Dollar -> assert false
 
 and desugar_argument env v =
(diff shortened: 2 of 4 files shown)


aryx:
test plan?

semgrep-core/src/ojsonnet/interpreting/Core_jsonnet.ml
```diff
@@ -82,7 +82,7 @@ type expr =
 (* ------------------------------------------------------------------------- *)
 
 (* no Dollar anymore *)
-and special = Self | Super
+and special = Self | Super | StdLength | StdMakeArray
```
aryx: Why not using the Desugar_jsonnet.mk_DotAccess_std to generate those calls?

semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
```diff
@@ -77,6 +79,12 @@ and expr_or_null v : expr =
   | None -> L (Null fk)
   | Some e -> e
 
+let freshvar =
+  let store = ref 0 in
+  fun () ->
+    incr store;
+    ("reallylongfreshvariablename" ^ string_of_int !store, fk)
```
aryx: you can use something like "!tmp!"
it's guaranteed to not conflict with a variable from the program since !tmp! is not a valid identifier.

semgrep-core/src/ojsonnet/parsing/AST_jsonnet.ml
```diff
@@ -135,7 +135,13 @@ and string_content = string wrap list
  * "core" language where super is at the expression level, so this is
  * probably simpler to have it here instead of in 3 special constructs.
  *)
-and special = Self | Super | Dollar (* ??? *)
+and special =
+  | Self
+  | Super
+  | Dollar
+  (* ??? *)
```
aryx: I'm not convinced. Those things are not in the grammar, and are not in the Core jsonnet definitions in the spec.
I'd rather follow closely the spec and use the mk_DotAccess_std function I've defined.

aryx: In fact, where do you build AST trees containing AST_jsonnet.stdMakeArray? Nowhere ... the Parse_jsonnet_tree_sitter.ml does not build them (and it should not).

semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
```diff
@@ -262,42 +272,65 @@ and desugar_arr_inside env (l, v, r) : C.expr =
...
-    let v2 = desugar_expr env v2 in
-    todo env (v1, v2))
-    env v
+(* Strictly speaking, the semantics say you're supposed to desugar as an expression in tandem with
+   desugaring the comprehension.
+   However, these return two different types in our desugaring process. So we can't do that.
+
+   I hope that applying a single outer-level desugar_expr is equivalent to interleaving it.
+*)
+and desugar_comprehension_helper env expr comps =
```
aryx: can you put the return type for this function like I did for desugar_expr (the original code was autogenerated by otarzan so when you "fill it" with the correct code, it can be useful to add type annotations, which otarzan can't add).

semgrep-core/src/ojsonnet/interpreting/Eval_jsonnet.ml
```diff
@@ -124,6 +125,40 @@ and eval_expr_aux (env : env) (v : expr) : V.value_ =
   | Id (s, tk) -> lookup env tk (LId s)
   | IdSpecial (Self, tk) -> lookup env tk LSelf
   | IdSpecial (Super, tk) -> lookup env tk LSuper
+  (* Should not have to evaluate these on their own. *)
+  | IdSpecial (StdLength, _)
+  | IdSpecial (StdMakeArray, _) ->
+      assert false
+  | Call (IdSpecial (StdLength, _), (_l, args, _r)) ->
```
aryx: I think that if you remove those StdXxx specials and just use mk_DotAccess_Std, you
can intercept instead at Call (ArrayAccess(Id "std", Id "length")) ...

semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
```diff
@@ -262,42 +272,65 @@ and desugar_arr_inside env (l, v, r) : C.expr =
...
+      let empty_else = Some (fk, mk_array []) in
+      let inner_exp =
+        match rest with
+        | [] -> mk_array [ expr ]
+        | __else__ -> desugar_comprehension_helper env expr rest
+      in
+      If (tok, cond, inner_exp, empty_else)
+  | CompFor (_, x, _, for_expr) :: rest ->
+      let std_join (l, r) =
+        Call (DotAccess (Id ("std", fk), fk, ("join", fk)), fb [ Arg l; Arg r ])
```
aryx: use mk_DotAccess_Std

aryx (changes requested):

semgrep-core/src/ojsonnet/parsing/AST_jsonnet.ml
brandonspark: i was relying on `Desugar` to produce them, but you're correct that it shouldn't have been a new variant

semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
```diff
@@ -262,42 +270,65 @@ and desugar_arr_inside env (l, v, r) : C.expr =
...
-and desugar_for_or_if_comp env v =
-  match v with
-  | CompFor v ->
-      let v = desugar_for_comp env v in
-      todo env v
-  | CompIf v ->
-      let v = desugar_if_comp env v in
+  | ArrayComp (expr, for_comp, rest_comp) ->
+      let v = desugar_comprehension env expr (CompFor for_comp :: rest_comp) in
       todo env v
```
aryx, line 275: We probably don't want to return a todo here.
BTW what is your test plan for your PR?

semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
```diff
@@ -262,42 +270,65 @@ and desugar_arr_inside env (l, v, r) : C.expr =
...
-    env v
+(* Strictly speaking, the semantics say you're supposed to desugar as an expression in tandem with
+   desugaring the comprehension.
+   However, these return two different types in our desugaring process. So we can't do that.
+
+   I hope that applying a single outer-level desugar_expr is equivalent to interleaving it.
+*)
+and desugar_comprehension_helper env expr comps : AST_jsonnet.expr =
+  match comps with
+  | CompIf (tok, cond) :: rest ->
```
aryx, line 285: I tried to use the same name than in the spec, so for example here I would use e' instead of cond,
so it really matches the spec, and I would rename the passed expr parameter as 'e', like in the spec.

semgrep-core/src/ojsonnet/interpreting/Desugar_jsonnet.ml
```diff
@@ -262,42 +270,65 @@ and desugar_arr_inside env (l, v, r) : C.expr =
...
+      let empty_else = Some (fk, mk_array []) in
+      let inner_exp =
+        match rest with
+        | [] -> mk_array [ expr ]
+        | __else__ -> desugar_comprehension_helper env expr rest
+      in
+      If (tok, cond, inner_exp, empty_else)
+  | CompFor (_, x, _, for_expr) :: rest ->
```
aryx, line 293: I would use e' instead of for_expr, to follow the spec

semgrep-core/src/ojsonnet/interpreting/Eval_jsonnet.ml
```diff
@@ -124,6 +125,44 @@ and eval_expr_aux (env : env) (v : expr) : V.value_ =
   | Id (s, tk) -> lookup env tk (LId s)
   | IdSpecial (Self, tk) -> lookup env tk LSelf
   | IdSpecial (Super, tk) -> lookup env tk LSuper
+  | Call
+      ( ArrayAccess
+          (Id ("std", _), (_, L (Str (None, DoubleQuote, (_, s, _))), _)),
+        (_l, args, _r) ) -> (
+      match s with
+      | [ ("length", _) ] ->
+          Primitive (Double (float_of_int (List.length args), fk))
```
aryx, line 134: This is wrong. You need to inspect args, which should always have one element, then evaluate it,
then look at its type, and depending on its type (string, array, object) do different things.

semgrep-core/src/ojsonnet/interpreting/Eval_jsonnet.ml
```diff
@@ -124,6 +125,44 @@ and eval_expr_aux (env : env) (v : expr) : V.value_ =
...
+                  error tk
+                    (spf "Improper arguments to std.makeArray: %s, %s" (sv v)
+                       ([%show: expr] e')))
+          | _ ->
+              error tk
+                (spf
+                   "Improper number of arguments to std.makeArray: expected 2, \
+                    got %d"
+                   (List.length args)))
+      | _ -> failwith "impossible: invalid program in dynamics")
```
aryx, line 165: This is wrong too. If the method is not length, or makeArray, then we should default to the normal
evaluation of Call, because those methods should be defined in the imported std.jsonnet.

aryx (approved):
I'll approve because it's good progress, but I think you need to fix the "this is wrong"
or put a TODO and fix it in another PR.

aryx:
hmm, I need the code in this PR so I'll add myself the TODO in another PR and try to improve on it.

PR #6669, opened 2022-12-07
tainting: Fix regression in DeepSemgrep after PR #6640

IagoAbal:
Better check unconditionally whether the function is a sink and, if so, check whether its arguments are tainted. When the function acts like a taint propagator, that will produce some duplicate findings, but they should be filtered out during deduplication anyways.

In addition, the taint of the function expression itself should always be propagated, also in DeepSemgrep. The function/method itself could be a taint source. Plus, for now it helps us identify `x.foo()` as tainted when `x` is tainted, although in the future we want to do this properly.

Fixes: 1eb330d6887 ("tainting: Make identification of sinks more precise (#6640)")

test plan:
make test

PR checklist:

### DIFF
--- a/changelog.d/pa-2304.fixed
+++ b/changelog.d/pa-2304.fixed
@@ -1,0 +1,3 @@
+DeepSemgrep: Fix regression in taint-mode introduced by Semgrep v1.1 that caused
+some findings to be missed. Also, DeepSemgrep will assume, for now, that a method
+call on a tainted object is always tainted.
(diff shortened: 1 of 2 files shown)


brandonspark (approved):

emjin (commented):
Will definitely lead to FPs. I'm comfortable with that right now, because we have a bigger FN problem and because FPs are easier to report than FNs, but we should benchmark this to see the impact.

PR #6694, opened 2022-12-12
fix!: resolvedname aliasing

brandonspark:
## What:
In #6667, I added ability to use `ResolvedName` in matching aliased `IdQualified`s. This, however, prevents some tests from reaching the "boilerplate" case below it, which actually does some stuff we want to do.

## Why:
We should fix this because it causes some Ruby tests to fail.

## How:
I added the logic from the case it no longer reached into the modified case.

## Test plan:
`make test` with these changes from DeepSemgrep.

PR checklist:


nmote (approved):
Thanks!

PR #6698, opened 2022-12-13
fix: Avoid stack overflow when symbol resolves to itself

nmote:
This addresses an issue encountered in DeepSemgrep. I don't know how to trigger this with regular Semgrep.

Test plan: Automated tests, plus new automated tests to be added in DeepSemgrep.

Also ran it manually with `--debug` in the crashing case I have for DeepSemgrep and observed the new log lines.

PR checklist:

### DIFF
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -401,8 +401,24 @@ let m_with_symbolic_propagation ~is_root f b =
       && not is_root)
     ~then_:
       (match b.G.e with
-      | G.N (G.Id (_, { id_svalue = { contents = Some (G.Sym b1) }; _ })) ->
-          f b1
+      | G.N (G.Id ((id, _), { id_svalue = { contents = Some (G.Sym b1) }; _ }))
+        ->
+          (* We shouldn't end up with a symbol that resolves to itself, but if
+           * we do, we shouldn't crash. This simple check will not protect
+           * against complicated paths through which a symbol could resolve to
+           * itself, but if it directly resolves to itself, we can easily catch
+           * it.
+           *
+           * Yes, Semgrep, I want physical equality.
+           *
+           * nosemgrep *)
+          if b1 == b then (
+            logger#warning
+              "Aborting symbolic propagation: Circular reference encountered \
+               (\"%s\")"
+              id;
+            fail ())
+          else f b1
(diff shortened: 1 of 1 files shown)


aryx:
weird, CI seems blocked. Can you make a new commit and push?
You can do `git commit --amend --no-edit && git push --force-with-lease`

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -401,8 +401,24 @@ let m_with_symbolic_propagation ~is_root f b =
       && not is_root)
     ~then_:
       (match b.G.e with
-      | G.N (G.Id (_, { id_svalue = { contents = Some (G.Sym b1) }; _ })) ->
-          f b1
+      | G.N (G.Id ((id, _), { id_svalue = { contents = Some (G.Sym b1) }; _ }))
+        ->
+          (* We shouldn't end up with a symbol that resolves to itself, but if
```
IagoAbal, line 406: This doesn't seem like a proper fix, may be fine to merge for now but we need a Linear ticket to keep track of the issue and make sure it's properly fixed.

IagoAbal: I guess that it's either a bug in name resolution, which is giving both occurrences of `foo` the same `id_info` when it should not, or it is a bug in symbolic-propagation. Since these bugs have pretty bad consequences, perhaps we should add a guardrail here and bound the amount of `id_svalue`s that we can follow. I would prefer that we add that sanity check now in this PR but I understand you may be pressed for time. However please create the Linear ticket and assign it to me. (I would like to know how we found this in the first place, so I can reproduce both the minimized example and the actual example.)

IagoAbal (approved):

semgrep-core/src/matching/Generic_vs_generic.ml
nmote: The test case I have for DeepSemgrep is `const foo = foo` (see my PR there) which is invalid code to begin with. I'm not really worried about analyzing it correctly (whatever that means) -- just worried about not crashing. I'll put together a linear ticket for you, though -- I think the original test case was more reasonable.

I agree this isn't a complete solution, but I think there should be guardrails here anyway. I'd be fine if you'd like to limit depth instead, but as you've noted I am a bit pressed for time at the moment.

Issue #4185, opened 2021-10-30
Autofix is broken because of import statements.


s3rius:
**Describe the bug**
I faced this issue when I was writing my own set of rules in .semgrep.yml.
I've tried to reproduce this bug on semgrep.dev playground and it's reproducible.

**To Reproduce**
Create custom rule with fix.
For example:
```yaml
rules:
  - id: exception-useless-parentheses
    pattern: raise $X()
    fix: raise $X
    message: Found useless parentheses
    languages: [python]
    severity: ERROR
```

This rule must validate that exception with no parameters have no parentheses.
And it works well for this sample:
```python
def test_test():
  raise RandomError()  # Works ok.
```

But if we add import statement for the exception, it breaks.

```python
from randomlib.errors import RandomError


def test_test():
  raise RandomError()
```

Semgrep suggests the following:
```
raise randomlib.errors import RandomError
```

**Expected behavior**

I expect that if I use $X variable it uses only $X variable and no import statement is the suggestion.

**What is the priority of the bug to you?**

- [x] P0: blocking your adoption of Semgrep or workflow
- [x] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
It can be reproduced on https://semgrep.dev/editor and in standalone cli tool. The version of semgrep is 0.70.0.

Kolaer:
```
rules:
- id: exception-useless-parentheses
  patterns:
    - pattern: raise $X()
    - pattern-not-inside: |
        import $X
        ...
  message: Found useless parentheses
  languages: [python]
  severity: ERROR
  fix: raise $X
```
Seems to work fine: https://semgrep.dev/s/Q9Gk

ievans:
@s3rius , did the approach @Kolaer posted unblock you?

s3rius:
Yes, it helped. But it looks really weird. Is it ok?

s3rius:
If it is,  feel free to close the issue.

ievans:
Great :) Let's keep the issue open, because your approach *ought* to work so there is still a bug here. But hopefully it is no longer a P0 blocker for you!

IagoAbal:
Closed by #6900.

aryx:
Did you try it @IagoAbal ? Should we add the test then?

IagoAbal:
I did try it, and took note to add the test. But I should have added the test before closing it. Will do tomorrow. Hopefully I didn't make any mistake.

[...]

PR #6700, opened 2022-12-13
fix(deep): `metavariable-pattern` matching error

brandonspark:
## What:
We are getting crashes in DeepSemgrep. A minimal reproduction is below:
A minimal reproduction is below:
```
rules:
- id: insufficient-postmessage-origin-validation
  message: >-
    blahblah
  languages:
  - javascript
  - typescript
  severity: WARNING
  patterns:
    - pattern: |
        foo('blah', $FUNC)
    - metavariable-pattern:
        pattern: |
          $ANY
        metavariable: $FUNC
```
and
```
var badResolved = 2;

foo("blah", badResolved);
```

The reason for this is that we are permitting metavariables to match in `envf` to `IdQualified`s constructed from `ResolvedName`s, which always have fake tokens. This means that, when looking up the metavariable's range, Semgrep crashes.

## Why:
This comes up in common use cases, and in DeepSemgrep, crashes the program.

## How:
I intercepted the matching logic so that we don't produce a match from a metavariable to an `IdQualified` resulting from a `ResolvedName`.

At first, I thought of a different solution (changing the SAST to have tokens, and making the metavariable range succeed), but what I didn't realize was that there is actually no reason whatsoever to allow the range of that `IdQualified` to succeed. It's a nonsense range, because it will be comprised of things like tokens for identifiers of classes, tokens for identifiers of variables, etc. It doesn't have semantic meaning, and it doesn't even make sense in the `metavariable-regex` case. For instance, for the following code:
```
class A:
  x = 2
```
a metavariable matching the `ResolvedName` of `x` will match this:
```
A:
x
```
because it is using the entire range between the `A` and the `x`.

So no reason to permit this.

## Test plan:
I can't add a test here because `ResolvedName`s occur in DeepSemgrep. But, I did actually try this change out with `DeepSemgrep`, and it causes my tests to pass (and another to no longer xfail!). PR forthcoming.

Closes PA-2263

PR checklist:

### DIFF
--- a/changelog.d/pa-2263.fixed
+++ b/changelog.d/pa-2263.fixed
@@ -1,1 +1,2 @@
-DeepSemgrep: Semgrep no longer crashes in certain cases when `metavariable-pattern/regex` are used on a metavariable referencing another metavariable.
+DeepSemgrep: Semgrep no longer crashes in certain cases when `metavariable-pattern/regex`
+are used on a metavariable referencing a variable which has been imported.
--- a/changelog.d/pa-2263.fixed
+++ b/changelog.d/pa-2263.fixed
@@ -1,2 +1,1 @@
-DeepSemgrep: Semgrep no longer crashes in certain cases when `metavariable-pattern/regex`
-are used on a metavariable referencing a variable which has been imported.
+Semgrep no longer crashes in certain cases when `metavariable-pattern/regex` are used on a metavariable referencing a variable which has been imported.
--- a/changelog.d/pa-2263.fixed
+++ b/changelog.d/pa-2263.fixed
@@ -1,0 +1,1 @@
+DeepSemgrep: Semgrep no longer crashes in certain cases when `metavariable-pattern/regex` are used on a metavariable referencing another metavariable.
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -818,18 +818,19 @@ and m_expr ?(is_root = false) a b =
   | ( _a,
       B.N
         (B.Id
-           ( _idb,
-             {
-               B.id_resolved =
-                 {
-                   contents =
-                     Some
-                       ( ( B.ImportedEntity dotted
-                         | B.ImportedModule (B.DottedName dotted) ),
-                         _sid );
-                 };
-               _;
-             } ) as nb) ) ->
+          ( idb,
+            {
+              B.id_resolved =
+                {
+                  contents =
+                    Some
+                      ( ( B.ImportedEntity dotted
+                        | B.ImportedModule (B.DottedName dotted)
+                        | B.ResolvedName (dotted, _) ),
+                        _sid );
+                };
+              _;
+            } )) ) ->
       (* We used to force to fully qualify entities in the pattern
        * (e.g., with org.foo(...)) but this is confusing for users.
        * We now allow an unqualified pattern like 'foo' to match resolved
@@ -838,20 +839,9 @@ and m_expr ?(is_root = false) a b =
        * bugfix: important to call with empty_id_info() below to avoid
        * infinite recursion.
        *)
-      (match a.e with
-      | G.N na ->
-          m_name na nb
-          >||> m_with_symbolic_propagation ~is_root (fun b1 -> m_expr a b1) b
-      | _ -> fail ())
-      >!>
-      (* try this time a match with the resolved entity *)
-      (* This needs to be gated behind a `>!>`! This is because we want to avoid
-         matching this `dotted` unless we _have_ to, because it is breaking apart
-         an imported name. This means that the tokens of the produced `DotAccess`
-         are potentially all out of order, meaning we should prefer to match the
-         name directly, and then break it apart "safely" in `m_name`.
-      *)
-      fun () -> m_expr a (make_dotted dotted)
+      m_expr a (B.N (B.Id (idb, B.empty_id_info ())) |> G.e)
+      >||> (* try this time a match with the resolved entity *)
+      m_expr a (make_dotted dotted)
   (* equivalence: name resolving on qualified ids (for OCaml) *)
   (* Put this before the next case to prevent overly eager dealiasing *)
   | G.N (G.IdQualified _ as na), B.N ((B.IdQualified _ | B.Id _) as nb) ->
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -436,8 +436,6 @@ let m_regexp_options a_opt b_opt =
    See the case conditioning in `is_resolved` below.
 *)
 let rec m_name ?(is_resolved = false) a b =
-  Common.(
-    pr2 (spf "names are %s and %s" ([%show: G.name] a) ([%show: G.name] b)));
   let try_parents dotted =
     let parents =
       match !hook_find_possible_parents with
@@ -506,17 +504,17 @@ let rec m_name ?(is_resolved = false) a b =
       (* Try the resolved entity and parents *)
       match a with
       (* > If we're matching against a metavariable, don't bother checking
-         * > the resolved entity or parents. It will only cause duplicate matches
-         * > that can't be deduped, since the captured metavariable will be
-         * > different.
-         *
-         * FIXME:
-         * This is actually not the correct way of dealing with the problem,
-         * because there could be `metavariable-xyz` operators filtering the
-         * potential values of the metavariable. See DeepSemgrep commit
-         *
-         *     5b2766ee30e "test: Tests for matching metavariable patterns against resolved names"
-      *)
+       * > the resolved entity or parents. It will only cause duplicate matches
+       * > that can't be deduped, since the captured metavariable will be
+       * > different.
+       *
+       * FIXME:
+       * This is actually not the correct way of dealing with the problem,
+       * because there could be `metavariable-xyz` operators filtering the
+       * potential values of the metavariable. See DeepSemgrep commit
+       *
+       *     5b2766ee30e "test: Tests for matching metavariable patterns against resolved names"
+       *)
       | G.Id ((str, _tok), _info) when MV.is_metavar_name str -> fail ()
       | _ ->
           (* Try matching against parent classes *)
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -749,25 +749,6 @@ and m_expr_root a b = m_expr ~is_root:true a b
  * also add them in m_pattern
  *)
 and m_expr ?(is_root = false) a b =
-  let _match_dotted_if_resolved a b =
-    match b.G.e with
-    | B.N
-        (B.Id
-          ( _,
-            {
-              B.id_resolved =
-                {
-                  contents =
-                    Some
-                      ( ( B.ImportedEntity dotted
-                        | B.ImportedModule (B.DottedName dotted) ),
-                        _sid );
-                };
-              _;
-            } )) ->
-        m_expr a (make_dotted dotted)
-    | _ -> fail ()
-  in
   Trace_matching.(if on then print_expr_pair a b);
   match (a.G.e, b.G.e) with
   (* the order of the matches matters! take care! *)
@@ -878,7 +859,6 @@ and m_expr ?(is_root = false) a b =
       >||> m_with_symbolic_propagation ~is_root (fun b1 -> m_expr a b1) b
   | G.N (G.Id ((str, tok), _id_info)), _b when MV.is_metavar_name str ->
       envf (str, tok) (MV.E b)
-      >||> m_with_symbolic_propagation ~is_root (fun b1 -> m_expr a b1) b
   (* metavar: typed! *)
   | G.TypedMetavar ((str, tok), _, t), _b when MV.is_metavar_name str ->
       with_lang (fun lang -> m_compatible_type lang (str, tok) t b)
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -427,7 +427,8 @@ let m_regexp_options a_opt b_opt =
 (* TODO: factorize with metavariable and aliasing logic in m_expr
  * TODO: remove MV.Id and use always MV.N?
  *)
-let rec m_name a b =
+
+let rec m_name ?(is_resolved = false) a b =
   let try_parents dotted =
     let parents =
       match !hook_find_possible_parents with
@@ -446,7 +447,7 @@ let rec m_name a b =
     | B.ResolvedName (_, alternate_names) ->
         List.fold_left
           (fun acc alternate_name ->
-            acc >||> m_name a (H.name_of_ids alternate_name))
+            acc >||> m_name ~is_resolved:true a (H.name_of_ids alternate_name))
           (fail ()) alternate_names
     | _ -> fail ()
   in
@@ -467,10 +468,15 @@ let rec m_name a b =
                };
              _;
            } as infob) ) ) -> (
+      let is_resolved =
+        match resolved with
+        | B.ResolvedName _ -> true
+        | _ -> false
+      in
       m_name a (B.Id (idb, { infob with B.id_resolved = ref None }))
       >||> try_alternate_names resolved
       (* Try the resolved entity *)
-      >||> m_name a (H.name_of_ids dotted)
+      >||> m_name ~is_resolved a (H.name_of_ids dotted)
       >||>
       (* Try the resolved entity and parents *)
       match a with
@@ -494,7 +500,7 @@ let rec m_name a b =
       (* this will handle metavariables in Id *)
       m_ident_and_id_info (a1, a2) (b1, b2)
   | G.Id ((str, tok), _info), G.IdQualified _ when MV.is_metavar_name str ->
-      envf (str, tok) (MV.N b)
+      if is_resolved then fail () else envf (str, tok) (MV.N b)
   (* equivalence: aliasing (name resolving) part 2 (mostly for OCaml) *)
   | ( G.IdQualified _a1,
       B.IdQualified
--- a/semgrep-core/src/matching/Generic_vs_generic.ml
+++ b/semgrep-core/src/matching/Generic_vs_generic.ml
@@ -480,7 +480,19 @@ let rec m_name ?(is_resolved = false) a b =
               (* m_name a n *)
               return ())
       | _ -> fail ())
-      >||>
+      (* Try to match the resolved name.
+         We want to gate this behind `>!>`, because we prefer not to break it apart
+         if we don't have to! We should only descend underneath the `ResolvedName` if
+         it is not possible to produce a match to the identifier which contains that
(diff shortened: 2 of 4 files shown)


nmote:
I'm not sure about this. Generally, we want to be able to go from:

```
    - pattern: foo.bar(...)
```

to

```
    - pattern: $FOO(...)
    - metavariable-pattern:
        pattern: foo.bar
        metavariable: $FOO
```

with no change in behavior. I believe I've seen bug reports stemming from inconsistencies like this.

mjambon (approved):

mjambon:
looks like GHA is stuck:
image

brandonspark:
> I'm not sure about this. Generally, we want to be able to go from:
> 
> ```
>     - pattern: foo.bar(...)
> ```
> 
> to
> 
> ```
>     - pattern: $FOO(...)
>     - metavariable-pattern:
>         pattern: foo.bar
>         metavariable: $FOO
> ```
> 
> with no change in behavior. I believe I've seen bug reports stemming from inconsistencies like this.

I believe this should still work. In the prior example, we will properly unpack the `ResolvedName` (because this logic only prevents the match to a _metavariable_, not a `DotAccess` like `foo.bar`).

In the second example, it is true we will prevent the match from `$FOO` to `foo.bar`. This is only a good thing though, as it would otherwise crash the `metavariable-pattern`. What should happen is that `$FOO` will match the identifier which _contains_ the `ResolvedName`, and then the `metavariable-pattern` on _that_ should succeed with `foo.bar` matching against it (because it will then unpack the `ResolvedName`). Does that make sense?

I've just tested it. It seems that it works going from:
```
- pattern: foo.bar(...)
```
to
```
- pattern: $FOO
- metavariable-pattern:
    metavariable: $FOO
    pattern: |
      foo.bar(...)
```
specifically.

You have to factor the parens into the `metavariable-pattern`, I think because `foo.bar` is a `DotAccess` and does not match properly against an identifier whose `ResolvedName` is `foo.bar`. Is that acceptable?

changelog.d/pa-2263.fixed
```diff
@@ -0,0 +1 @@
+DeepSemgrep: Semgrep no longer crashes in certain cases when `metavariable-pattern/regex` are used on a metavariable referencing another metavariable.
```
IagoAbal: Perhaps I got it wrong, but I think this changelog entry is not accurate. The error happens when using those operators on a metavariable _that binds to a code variable_. Although I am not sure I understand what did you want to say with "a metavariable referencing another metavariable". 

Also, Semgrep never crashed AFAICT. It's printed like "internal error" but it's an error we are recovering from. It should probably just say "matching warning" or something. The thing is that the operators now should work, and previously they just evaluated to false.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -485,7 +493,25 @@ let rec m_name a b =
...
+         then the pattern metavariable will match the range of the text
+         ```
+         A:
+         def foo
+         ```
+         because it is matching everything between `A` and `foo`.
+
+         So we should not permit this match.
+      *)
+      if is_resolved then fail () else envf (str, tok) (MV.N b)
```
IagoAbal, line 554: Have we considered other solutions?

IagoAbal (changes requested):
I agree with Nat here. And I also think this or a similar problem should be reproducible with Semgrep alone, I remember Nat worked on it, and we tried implementing a similar solution to this, which we later reverted. Have we discussed or considered any other solution?

brandonspark:
> I agree with Nat here. And I also think this or a similar problem should be reproducible with Semgrep alone, I remember Nat worked on it, and we tried implementing a similar solution to this, which we later reverted. Have we discussed or considered any other solution?

I don't believe that the issue that Nat has brought up is at odds with my solution, I think that it's a pre-existing quirk of the matching engine.

All I am doing with this PR is _preventing the match_ from a metavariable to a `ResolvedName` which consists of many different fake tokens. This is nonsensical, because those tokens could have been sourced from _anywhere_ in the file. They could be backwards, upside-down, or even in completely different files. This means that even if they _had_ location data, it would fundamentally disagree with an assumption we make in many places in Semgrep's logic, which is that a metavariable refers to a _contiguous range_ in the original program text.

Even if we did not add location data back, this still represents an issue due to the above assumption. 

In the example Nat gave, we still would not succeed with any attempted remediation of the situation (up to and including just ignoring the warning), because the `metavariable-pattern` will default to false. This prevents a match which, as Nat himself said, should occur in the first example given. I am not proposing a solution to that issue, but I _am_ proposing a solution to other issues.

I agree that `metavariable-pattern` will fail, but the question is simply _whether the metavariable we are `pattern`ing on should be allowed to match the resolved name at all_. As with my answer here:

> What should happen is that $FOO will match the identifier which contains the ResolvedName, and then the metavariable-pattern on that should succeed with foo.bar matching against it (because it will then unpack the ResolvedName). 

I believe that the best solution (that is also consistent with how we treat symbolically propagated values, for instance) is to prevent the match to the resolved name directly.

semgrep-core/src/matching/Generic_vs_generic.ml
brandonspark: We cannot intercept it by `metavariable-pattern` time, or any time once the binding has been produced, because we will not know whether this originated from a `ResolvedName`. One solution is just to remove the error message entirely, which will lead to `metavariable-pattern` failing, which is good, because then no match will be produced (as it should), but it means _without_ a `metavariable-pattern`, we will produce an extraneous match, which seems to "go away" on a `metavariable-pattern`, regardless of the clause. This does not seem to make sense to me.

Another thing we could do is make an effort to also just fail to produce a match if the expression we are binding does not have a range. This is done in `text_of_binding`, at `Eval_generic` time, but this seems to run later than `envf`, and possibly the match which is being `metavariable-pattern`ed on, which is why we still produce an error. We could add some logic to refuse to `envf` if the expression in question does not have location data. I wanted to restrict it more to this case though, since I don't know what that looks like.

brandonspark: It's worth noting that even if we add the latter option (refusing to match nonsensical ranges), this will not work in the grand scheme of my solution for fixing the problems that I see. This is because I intend to make `ResolvedName`s contain location data where possible, and if we produce a match to a `ResolvedName` with true location data, really bad stuff will happen. So I am in favor of recognizing that resolved names behave a little differently, and their location data cannot really be trusted.

IagoAbal (approved):
After a proper chat with Brandon I think this solution should work. I will just re-review the code to finish convincing myself. 😄 Also, as I mentioned to Brandon, this may also be a problem that exists with `ImportedEntity` which is what `Naming_AST` uses when it resolves names.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -485,7 +493,25 @@ let rec m_name a b =
       (* this will handle metavariables in Id *)
       m_ident_and_id_info (a1, a2) (b1, b2)
   | G.Id ((str, tok), _info), G.IdQualified _ when MV.is_metavar_name str ->
-      envf (str, tok) (MV.N b)
+      (* If `is_resolved` is true, then we got the target `IdQualified` from a
```
emjin, line 526: This explanation still left me confused 😅. It's definitely tricky to explain. I think it would be helpful if you zoomed out in the beginning. It's not intuitively clear to me that it's bad for an identifier to have a nonsensical range, so I could have used some context there.

emjin (approved):
I like this solution. It took me a bit to wrap my head around what was happening, but I agree that, given our current behavior, metavariables shouldn't bind resolved names. I do think Nat's concern is important and we should work towards improving behavior there.

nmote (changes requested):
Maybe I'm still misunderstanding, but I still believe that this will disable behavior that we currently allow, and that we want to allow.

> In the second example, it is true we will prevent the match from $FOO to foo.bar. This is only a good thing though, as it would otherwise crash the metavariable-pattern.

It sounds like you believe that the example I gave doesn't work today. It does: https://semgrep.dev/s/K247

I'm reluctant to give up this behavior and I'd like to at least make sure we are on the same page regarding the tradeoffs here before moving forward.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -458,10 +461,15 @@ let rec m_name a b =
                };
              _;
            } as infob) ) ) -> (
+      let is_resolved =
+        match resolved with
+        | B.ResolvedName _ -> true
```
IagoAbal: Should make this more general to cover `ImportedEntity` too. I think this issue is related to #4185, and I remember seeing other issues of this kind. Perhaps `ImportedModule` should be covered too?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -458,10 +461,15 @@ let rec m_name a b =
...
+        | _ -> false
+      in
       m_name a (B.Id (idb, { infob with B.id_resolved = ref None }))
       >||> try_alternate_names resolved
       (* Try the resolved entity *)
-      >||> m_name a (H.name_of_ids dotted)
+      >||> m_name ~is_resolved a (H.name_of_ids dotted)
```
IagoAbal, line 502: It could be more clear to just avoid `>||> m_name ~is_resolved a (H.name_of_ids dotted)` here in the first place. The way it is done works, but it's subtle, and I'm not sure it needs to be.  The way it is done for sym-prop is a lot more clear to me. We simply add a case so that if something could not be matched, we give it a try using the sym-prop'ed info:

```ocaml
  | _, G.N (G.Id _) ->
      m_with_symbolic_propagation ~is_root (fun b1 -> m_expr a b1) b
```

Could we replicate this here?

IagoAbal: Alternatively, do you think that something like this could work?

```ocaml
>||> (if is_metavariable a && is_resolved then
          (* Resolved names do not correspond to actual code, some tokens come from import declarations
           * whereas others come from uses of the imported entities, leading to non-sensical ranges.
           *  Metavariables should only be bound to actual code. *)
          fail ()
      else m_name a (H.name_of_ids dotted))
```
Then we don't need the optional argument passed to `m_name`.

IagoAbal:
I would say that this https://semgrep.dev/s/qzP1 should work. It currently doesn't.

nmote (approved):
Thanks for clarifying. In that case I will let you finish the review since you've already taken the time to understand this thoroughly.

semgrep-core/src/matching/Generic_vs_generic.ml
brandonspark: I'm not entirely sure, because I'm afraid of a case which looks like `$A.x` or `x.$A`. In both cases, we will proceed to match, and in the latter we will succeed because of the match on the `name_last` (which is a good thing). In the former case, I'm afraid of the possibility of `$A` matching a prefix of multiple qualifiers (like z.y.x, matching `z.y`). 

I'd rather keep the logic as is and defer as late as possible, because _then_ we know for sure we _must_ be matching a metavariable to an `IdQualified` (not just a single id, which is fine) that is from a `ResolvedName`.

semgrep-core/src/matching/Generic_vs_generic.mli
```diff
@@ -13,7 +13,11 @@ val m_attribute : AST_generic.attribute Matching_generic.matcher
 val m_partial : AST_generic.partial Matching_generic.matcher
 val m_field : AST_generic.field Matching_generic.matcher
 val m_fields : AST_generic.field list Matching_generic.matcher
-val m_name : AST_generic.name Matching_generic.matcher
+
+(* `is_resolved` is for checking if this name we are matching in the target
+   is resulting from a `ResolvedName`, which has a nonsensical range.
```
IagoAbal: The comment is not accurate since it's specific to `ResolvedName`.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -458,25 +461,34 @@ let rec m_name a b =
                };
              _;
            } as infob) ) ) -> (
+      let is_resolved =
+        match resolved with
+        | B.ResolvedName _
+        | B.ImportedEntity _
```
IagoAbal, line 481: Great! But we need a test here, at least https://semgrep.dev/s/qzP1, and better if we could come up with some other.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -418,7 +418,10 @@ let m_regexp_options a_opt b_opt =
 (* TODO: factorize with metavariable and aliasing logic in m_expr
  * TODO: remove MV.Id and use always MV.N?
  *)
-let rec m_name a b =
+(* `is_resolved` should be true if `b` is the result of unpacking a `ResolvedName`.
```
IagoAbal: Fix comment so that it's not specific to `ResolvedName`. (I wonder if we could merge `ImportedEntity` etc into a single constructor... but maybe that would be for another PR.)

IagoAbal (changes requested):
I tried the branch and it fails this test https://semgrep.dev/s/qzP1.

changelog.d/pa-2263.fixed
```diff
@@ -0,0 +1,2 @@
+DeepSemgrep: Semgrep no longer crashes in certain cases when `metavariable-pattern/regex`
```
IagoAbal: Changelog entry needs update? Not DeepSemgrep specific. Although this is less important, there was never any crash. This actually affected autofix as well. So it's generally about incorrect behavior when metavariables matched imported names.

changelog.d/pa-2263.fixed
```diff
@@ -0,0 +1,2 @@
...
+are used on a metavariable referencing a variable which has been imported.
```
IagoAbal: Also, this should be fixing #4185.

semgrep-core/tests/rules/imported_metavariable_pattern.js
```diff
@@ -0,0 +1,4 @@
+import {bar as b} from 'foo';
+
+//ruleid: imported-metavariable-pattern 
+b();
```
IagoAbal: Do we have a test suite for autofixes? It would be nice to add an autofix test too for #4185.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -427,7 +427,15 @@ let m_regexp_options a_opt b_opt =
 (* TODO: factorize with metavariable and aliasing logic in m_expr
  * TODO: remove MV.Id and use always MV.N?
  *)
-let rec m_name a b =
+(* `is_distributed` should be true if `b` is the result of unpacking a name which
```
IagoAbal: I liked `is_resolved` (or `is_imported`) better. Otherwise I would pick something more clear even if it's longer, in the spirit of `has_tokens_from_many_places`. That one would be a totally fine name and much easier to understand than `is_distributed`.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -446,136 +454,117 @@ let rec m_name a b =
...
-      | IdQualified a1 -> m_name_info a1 b1
-      | Id _ -> fail ())
-  | G.IdQualified a1, B.IdQualified b1 -> m_name_info a1 b1
-  | G.Id _, _
-  | G.IdQualified _, _ ->
-      fail ()
+      (match a with
+      | B.Id ((str, tok), _info) when MV.is_metavar_name str ->
+          (* If `is_distributed` is true, then we got the target `IdQualified` from a
+             `ResolvedName`.
```
IagoAbal: Is this true? Doesn't this also handle the case where the `IdQualified` comes from `Naming_AST`?

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -446,136 +454,117 @@ let rec m_name a b =
...
+                 pass
+             ```
+             then the pattern metavariable will match the range of the text
+             ```
+             A:
+             def foo
+             ```
+             because it is matching everything between `A` and `foo`.
+
+             So we should not permit this match.
```
IagoAbal: You should probably hint here why the `metavariable-pattern` example still works. Nat, Emma, myself had difficulties to understand this, so it's certainly not obvious, and in 6 months it may not even be obvious to you.

changelog.d/pa-2263.fixed
IagoAbal: The title of the PR could also be more general.

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -427,7 +427,10 @@ let m_regexp_options a_opt b_opt =
 (* TODO: factorize with metavariable and aliasing logic in m_expr
  * TODO: remove MV.Id and use always MV.N?
  *)
-let rec m_name a b =
+(* `is_resolved` should be true if `b` is the result of unpacking a `ResolvedName`.
+   See the case conditioning on `is_resolved` below.
+*)
+let rec m_name ?(is_resolved = false) a b =
   let try_parents dotted =
```
IagoAbal, line 439: I wonder if this `hook_find_possible_parents` may also be a source of ids with mixed tokens. These will be the parent classes of `dotted` (assuming `dotted` represents a class).

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -494,7 +506,29 @@ let rec m_name a b =
...
+         then the pattern metavariable will match the range of the text
+         ```
+         A:
+         def foo
+         ```
+         because it is matching everything between `A` and `foo`.
+
+         So we should not permit this match.
+      *)
+      if is_resolved then fail () else envf (str, tok) (MV.N b)
```
IagoAbal, line 554: Does it make sense to just check whether the tokens in `b` are coherent? We could tag `IdQualified`s as "generated" (vs coming from the sources) and that would make it trivial to check. Otherwise I wonder if we're missing cases and how we make sure we don't.

IagoAbal (approved):
Don't know the code very well but the tests pass so I approve too. I hope we don't break anything!

semgrep-core/src/matching/Generic_vs_generic.ml
```diff
@@ -446,136 +454,117 @@ let rec m_name a b =
...
+             A:
+             def foo
+             ```
+             because it is matching everything between `A` and `foo`.
+
+             So we should not permit this match.
+          *)
+          if is_distributed then fail () else envf (str, tok) (MV.N b)
+      | B.Id _ -> fail ()
+      | B.IdQualified a1 -> m_name_info a1 nameinfo)
```
IagoAbal: Previously this was combined with an `>||>` and now it's combined with a `>!>`, could add a comment explaining why?

semgrep-core/tests/rules/imported_metavariable_pattern.js
brandonspark: im getting the same autofix error even with this PR:
image
im not sure this solves that problem

semgrep-core/src/matching/Generic_vs_generic.ml
brandonspark: My next PR (on top of this one) will make _all_ such tokens coherent, so no point to adding a check

aryx (approved):

PR #6711, opened 2022-12-14
fix: Invalidation for file cache in Range.ml

nmote:
This addresses a DeepSemgrep crash when run on the `NodeBB` repo (see test plan), but it's probably possible to trigger this crash in other ways including without DeepSemgrep.

Basically, the issue is that collisions in tmp file names can lead to outdated information available in the cache. If we try to read, a substring starting at, say character 10, but the temp file that is cached is only 5 characters long, we get a crash.

Test plan: See https://github.com/returntocorp/pfff/pull/575

PR checklist:

### DIFF
--- a/changelog.d/cache-invalidation.fixed
+++ b/changelog.d/cache-invalidation.fixed
@@ -1,0 +1,1 @@
+Fixed rare crash that could occur due to stale file caches when temp file names overlap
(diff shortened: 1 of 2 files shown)


nmote:
Don't merge -- I'll update once the pfff PR is merged

aryx (approved):

PR #6724, opened 2022-12-16
fix/feat!: yaml quotes behavior

brandonspark:
## What:
As it stands right now, if you have:
```
key: "hi!"
```
matching pattern
```
key: $FOO
```

`$FOO` will be bound to `hi!`. This stands opposed to other languages, where we would match the verbatim textual contents, and thus match `"hi!"`.

Also, when you use a pattern like:
```
"$FOO"
```
matching a string
```
"foo"
```
the contents of the metavariable will be `"foo"`, and not `foo`, which is also inconsistent with typical Semgrep behavior.

Same with single variables, actually.

## Why:
This makes error crop up in Playground, because the metavariable is reported to be less long than the text it should be matching. Thus, weird display bugs like: https://semgrep.dev/s/ZygA

## How:
I cased on the YAML parser's output, which tells us if it matched something doubly quoted. If it's doubly quoted, we should wrap the content in quotes, for instance.

Also, when parsing a metavariable (to solve the second use case), if the metavariable is also the result of double quotes, then it must be a wrapped metavariable. Those get parsed to strings, so I just let it fall through to the typical case.

## Test plan:
`make test`

Closes PA-2332

PR checklist:

### DIFF
--- a/changelog.d/pa-2332.fixed
+++ b/changelog.d/pa-2332.fixed
@@ -1,0 +1,2 @@
+YAML: Fixed a bug where metavariables matching YAML double-quoted strings would not capture the entire range of the string, and would
+not contain the double-quotes. Also added the ability to properly use patterns like `"$FOO"`, which will unpack the contents of the matched string.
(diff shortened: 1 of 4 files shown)


emjin (approved):
Thank you!! This is a long time coming.

PR #6726, opened 2022-12-16
fix: Fix race condition when creating parse cache dir

nmote:
Since the check and directory creation isn't an atomic operation, it is at risk of race conditions. The directory can get created by another worker process between the check and creation, so then the `mkdir` fails with `EEXIST`.

Test plan: `semgrep --config p/deepsemgrep --deep -j 12` on the Angular repo. Before, there were 11 internal errors on each run (`Unix.Unix_error(Unix.EEXIST, "mkdir", "_deepsemgrepcache")`), now it finishes successfully.

PR checklist:

### DIFF
--- a/semgrep-core/src/runner/Parse_with_caching.ml
+++ b/semgrep-core/src/runner/Parse_with_caching.ml
@@ -98,7 +98,14 @@ let cache_computation use_parsing_cache version_cur file cache_file_of_file f =
 
 let cache_file_of_file parsing_cache_dir suffix filename =
   let dir = parsing_cache_dir in
-  if not (Sys.file_exists dir) then Unix.mkdir dir 0o700;
+  (if not (Sys.file_exists dir) then
+   try
+     Unix.mkdir dir 0o700
+     (* bugfix: Despite the check above, we may race with another thread to
+      * create the directory, leading to a crash if this exception is unhandled.
+      * *)
+   with
+   | Unix.Unix_error (Unix.EEXIST, _, _) -> ());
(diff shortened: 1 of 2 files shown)


semgrep-core/src/runner/Parse_with_caching.ml
```diff
@@ -98,7 +98,14 @@ let cache_computation use_parsing_cache version_cur file cache_file_of_file f =
...
   let dir = parsing_cache_dir in
-  if not (Sys.file_exists dir) then Unix.mkdir dir 0o700;
+  (if not (Sys.file_exists dir) then
+   try
+     Unix.mkdir dir 0o700
+     (* bugfix: Despite the check above, we may race with another thread to
+      * create the directory, leading to a crash if this exception is unhandled.
+      * *)
+   with
+   | Unix.Unix_error (Unix.EEXIST, _, _) -> ());
```
brandonspark, line 108: whats the consequence of not creating this directory? does this just mean we won't cache in this particular instance?

nmote: We need it to exist, we don't need to create it ourselves. That's why we check if it exists before trying to create it. The `EEXIST` error just means that we are trying to make something that already exists.

brandonspark: oh oops, that's smart. i did not think about what the error actually meant.

brandonspark (approved):
Let's Go To Mars!

PR #6749, opened 2022-12-22
feat(matching): do not match resolvednames if possible

brandonspark:
## What:
This PR adds in some changes to the matching logic, to prefer to not unpack the `ResolvedName`, `ImportedEntity`, or `ImportedModule` of an identifier, if at all possible.

This comes up in examples like the following:
```
rules:
- id: insufficient-postmessage-origin-validation
  message: >-
    blahblah
  languages:
  - javascript
  - typescript
  severity: WARNING
  patterns:
    - pattern: |
        foo('blah', $FUNC)
    - metavariable-pattern:
        pattern: |
          $ANY
        metavariable: $FUNC
```
matching
```
var badResolved = 2;

foo("blah", badResolved);
```
The reason for this is that we are permitting metavariables to match in envf to IdQualifieds constructed from ResolvedNames, which always have fake tokens. This means that, when looking up the metavariable's range, Semgrep crashes.

## Why:
We do not want Semgrep to have internal errors.

As a corollary, metavariables should not match to non-contiguous ranges. This causes issues further down, where if we `metavariable-pattern` or something, we will try to extract the source file's text, except because the range doesn't make sense we will cause an error. In addition, this causes weird behavior with `metavariable-regex`, where we can get cases like:
```
import {x as _y} from 'z';

// ok: my_pattern_id
var res = foo(_y);
```
matching
```
patterns:
- pattern: |
    foo($BAR)
- metavariable-regex:
    metavariable: $BAR
    regex: "^[^_].*"
```
In this case, clearly we are looking for the argument to `foo` to not start with an underscore. With Semgrep right now, we will actually produce a match here, on account of the fact that we also match the `ResolvedName`.

An additional motivation is that we want to succeed on matching the `ResolvedName` in _some_ cases. For instance:
```
import {x as y} from 'z';

var res = foo(y)
```
matching
```
patterns:
- pattern: |
    foo($BAR)
- metavariable-pattern:
    metavariable: $BAR
    pattern: |
      z.$X
```
In this case, this happens in two phases. The first is that we want `$BAR` to match to the identifier _containing_ the `ResolvedName` `z.x`, and then we want to be able to decompose on that information in the `metavariable-pattern`. We also want `$X` to be able to match the `x` which was imported, which is on the first line. 

## How:
This PR implements the behavior as described above. It factorizes the `m_name` logic to be slightly easier to understand, and to get rid of the previous way that it was done, which was to produce _two_ matches, one to the identifier which has been cleared of naming data, and one to the fully qualified name.

Now, we prefer to match the original identifier, which contains the naming data. This has the advantage of working analogously to how we do things with symbolic propagation, as well as ensuring that we never allow a metavariable to match a language construct which has a non-contiguous range in the source text. 

Should we fail to produce a match to the original identifier, however, we will then allow the possibility of unpacking the fully qualified name. If we do that, however, we still maintain the property that we should only match a metavariable to a single identifier, at best, and not to any string of tokens which might not be a contiguous range.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/semgrep-core/src/core/ast/AST_generic.ml
+++ b/semgrep-core/src/core/ast/AST_generic.ml
@@ -362,7 +362,7 @@ and id_info = {
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
    *)
-  id_type : type_ option ref;
+  id_type : type_ option ref; [@equal fun _a _b -> true]
   (* type checker (typing) *)
   (* sgrep: this is for sgrep constant propagation hack.
    * todo? associate only with Id?
--- a/semgrep-core/src/core/ast/AST_generic.ml
+++ b/semgrep-core/src/core/ast/AST_generic.ml
@@ -361,6 +361,9 @@ and id_info = {
   (* sgrep: in OCaml we also use that to store the type of
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
+   * We do not use this for equality, because we want to say that two variables
+   * are the same if their resolved names are the same, even if type analysis
+   * did not propagate the same type to them.
    *)
   id_type : type_ option ref; [@equal fun _a _b -> true]
   (* type checker (typing) *)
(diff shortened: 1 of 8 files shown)


aryx:
I think we can close this.

PR #6750, opened 2022-12-23
fix!: yaml parsing for literal and folded styles

### LINKED ISSUE #6731: [YAML] Inconsistent parsing of Semgrep pattern keys when in patterns
**Describe the bug**
`pattern`, `pattern-not`, `pattern-inside`, and `pattern-not-inside` are parsed differently. 

**To Reproduce**
https://semgrep.dev/s/DNby Comment out the metavariable-regex and run. I added a `$PATTERN` at the end of the message, and you can see how in line 27 (the `pattern` case) and 37 (the `pattern-not` case), the metavariable is not populated. This can be further confirmed with the `--json` cli argument.

**Expected behavior**
The `$PATTERN` metavariables should be captured in the `pattern` and `pattern-not` cases, alongside the `pattern-inside` and `pattern-not-inside` cases which currently work as expected.

**Environment**
Semgrep.dev and latest CLI version

**Use case**
We need this to fix one of SR's rule linting workflows.

brandonspark:
## What:
A little bit ago, I did #6724 for fixing the way that our YAML parser interacts with scalar strings. I did not deal with "literal" and "folded" style strings, however, which are those which use `|` and `>` respectively.

## Why:
See https://semgrep.dev/s/RNgK

We get unexpected behavior in the Playground, because the range of the match tends to be shorter than it actually should be. The printed out message (the contents of the metavariable) also tends to be wrong.

## How:
I changed the YAML parser so that we use the event end marker location information to extract the literal textual contents of the file at the relevant spots. The way that the YAML parser works now is that it returns the "ideal contents" of the string within YAML, meaning that it discounts things such as the beginning bar or chevron, and gets rid of some whitespace and newlines which are actually present within the literal text.

In the favor of adhering to the syntax, I opted to have the metavariable's text contain the literal contents of what it should match. This solution ends up being more general than what I did in #6724 

## Test plan:
`make test`

Worth noting that in the metavariables which captured a literal or folded string, they end up terminated with newlines (whereas the double-quoted and single-quoted metavariables do not). I could fix this, but it's unclear to me what the intended behavior should be.

Closes PA-2347
Closes #6731 

PR checklist:

### DIFF
--- a/cli/tests/e2e/snapshots/test_output/test_yaml_metavariables/report.json
+++ b/cli/tests/e2e/snapshots/test_output/test_yaml_metavariables/report.json
@@ -1,0 +1,308 @@
+{
+  "errors": [],
+  "paths": {
+    "_comment": "<add --verbose for a list of skipped paths>",
+    "scanned": [
+      "targets/yaml/target.yaml"
+    ]
+  },
+  "results": [
+    {
+      "check_id": "rules.yaml-key",
+      "end": {
+        "col": 13,
+        "line": 1,
+        "offset": 12
+      },
+      "extra": {
+        "fingerprint": "0x42",
+        "is_ignored": false,
+        "lines": "- key: \"one\"",
+        "message": "\"one\"\n",
+        "metadata": {},
+        "metavars": {
+          "$VALUE": {
+            "abstract_content": "\"one\"",
+            "end": {
+              "col": 13,
+              "line": 1,
+              "offset": 12
+            },
+            "start": {
+              "col": 8,
+              "line": 1,
+              "offset": 7
+            }
+          }
+        },
+        "severity": "ERROR"
+      },
+      "path": "targets/yaml/target.yaml",
+      "start": {
+        "col": 3,
+        "line": 1,
+        "offset": 2
+      }
+    },
+    {
+      "check_id": "rules.yaml-key",
+      "end": {
+        "col": 13,
+        "line": 2,
+        "offset": 25
+      },
+      "extra": {
+        "fingerprint": "0x42",
+        "is_ignored": false,
+        "lines": "- key: 'two'",
+        "message": "'two'\n",
+        "metadata": {},
+        "metavars": {
+          "$VALUE": {
+            "abstract_content": "'two'",
+            "end": {
+              "col": 13,
+              "line": 2,
+              "offset": 25
+            },
+            "start": {
+              "col": 8,
+              "line": 2,
+              "offset": 20
+            }
+          }
+        },
+        "severity": "ERROR"
+      },
+      "path": "targets/yaml/target.yaml",
+      "start": {
+        "col": 3,
+        "line": 2,
+        "offset": 15
+      }
+    },
+    {
+      "check_id": "rules.yaml-key",
+      "end": {
+        "col": 11,
+        "line": 4,
+        "offset": 45
+      },
+      "extra": {
+        "fingerprint": "0x42",
+        "is_ignored": false,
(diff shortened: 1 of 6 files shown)


cli/tests/e2e/snapshots/test_output/test_yaml_metavariables/report.json
```diff
@@ -0,0 +1,308 @@
+{
```
aryx, line 1: those e2e tests are not super readable. Who is gonna look at the value on those expected report.json?

cli/tests/e2e/snapshots/test_output/test_yaml_metavariables/report.json
```diff
@@ -0,0 +1,308 @@
...
+      "end": {
+        "col": 11,
+        "line": 4,
+        "offset": 45
+      },
+      "extra": {
+        "fingerprint": "0x42",
+        "is_ignored": false,
+        "lines": "- key: |\n    three",
+        "message": "|\n    three\n\n",
```
aryx, line 95: In theory we should semantically transform those as just "three"

cli/tests/e2e/test_output.py
```diff
@@ -92,6 +92,34 @@ def test_output_highlighting__force_color_and_no_color(run_semgrep_in_tmp, snaps
     )
 
 
+# This test is just for making sure that our YAML parser interacts properly
+# with metavariables. We don't want to introduce regressions which might
+# mess this up.
+@pytest.mark.kinda_slow
```
aryx: why kinda slow? should be fast, it's a very simple rule and target file.

cli/tests/e2e/test_output.py
```diff
@@ -92,6 +92,34 @@ def test_output_highlighting__force_color_and_no_color(run_semgrep_in_tmp, snaps
...
+        "rules/yaml_key.yaml",
+        target_name="yaml/target.yaml",
+        output_format=OutputFormat.JSON,
+    )
+    parsed_output = json.loads(stdout)
+    assert "results" in parsed_output
+
+    for result in parsed_output["results"]:
+        value = result["extra"]["metavars"]["$VALUE"]
+        content = value["abstract_content"]
```
aryx, line 110: we should get rid of abstract_content, so there is no need to assert anything on its content.

semgrep-core/src/parsing/other/yaml_to_generic.ml
```diff
@@ -51,6 +51,7 @@ module PI = Parse_info
 type env = {
   (* when we parse a pattern, the filename is fake ("<pattern_file>") *)
   file : Common.filename;
+  text : string;
```
aryx, line 59: please add a comment on what this field is for.

aryx (approved):

cli/tests/e2e/snapshots/test_output/test_yaml_metavariables/report.json
brandonspark: Right now, our infrastructure is incompatible with having the contents of the metavariable differ in length from the range of the metavariable itself. As much as I would like to "simplify" the abstract content (and lift it from being the literal text of the range), it seems this would require larger changes in the core and the CLI.

cli/tests/e2e/snapshots/test_output/test_yaml_metavariables/report.json
brandonspark: I've read them all, and verified they look good. As long as they don't change, from now on, then we should be good.

PR #6756, opened 2022-12-23
fix: Use GH API to get merge base instead of fetching

nmote:
As noted inline, fetching enough commits to compute the merge base can be very expensive on large repositories, and reduces or even eliminates the performance benefits that shallow clones provide. We can get around this by calling out to GitHub's rest API to get the merge base between two commits, rather than fetching enough to compute it locally.

Test plan:

```
docker build .
docker tag <hash> nmote/semgrep
docker push nmote/semgrep
```

Then, set up a test repo (https://github.com/nmote/gha-test/) with an action that runs `semgrep ci --config semgrep.yaml --debug --dry-run` using the `nmote/semgrep` image (this repo has to be private so that I can test the authentication portion, but I can add anyone who is interested).

Observe the following log line in the output of the GitHub Action, and observe that Semgrep runs as expected:

```
Got merge base using GitHub API: 73d928285096a2b2bec33e6d7d50cf3def74e8af
```

We could mock out the GH API and add a test for this, but the code here is actually quite simple, and most of what might break is how Semgrep interacts with external systems, which wouldn't be tested.

PR checklist:

### DIFF
--- a/changelog.d/gha-mergebase.fixed
+++ b/changelog.d/gha-mergebase.fixed
@@ -1,0 +1,1 @@
+Use the GitHub REST API when possible to compute the merge base for `semgrep ci`, improving performance on shallow clones of large repositories.
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -379,6 +379,7 @@ def ci(
                 skip_unknown_extensions=(not scan_unknown_extensions),
                 optimizations=optimizations,
                 baseline_commit=metadata.merge_base_ref,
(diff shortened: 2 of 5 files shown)


nmote:
I've filed a ticket to follow up on documentation once this has been proven in production. I have also reached out for a security review because this calls a new API.

aryx (approved):

nmote:
Got security review

LewisArdern:
Linear ticket created to track https://linear.app/r2c/issue/SEC-157/poa-for-github-url-for-gha but :stamp:

Issue #3801, opened 2021-09-01
support for pattern-inside a control flow


gordonwoodhull:
**Describe the bug**
We'd like to use semgrep to do the analysis of control flow that we can (in theory) do with SmPL. I'm not clear if semgrep patterns follow control flow or if it is only syntactic/AST matching.

Say we have some `Foo` provider and we want to match the region from the construction of the `Foo` to the call of `Foo.bar()`:

```yaml
  pattern: |
    $FOO = $PROVIDER.foo();
    ...
    $FOO.bar();
```

This matches the "no blocks" case:
```java
class C {
    public void f() {
        Foo foo = provider.foo();
        something.qux();
        foo.bar();
    }
}
```

and the "ends inside block" case:
```java
class C {
    public void f() {
        Foo foo = provider.foo();
        something.qux();
        if(true) {
            foo.bar();
        }
    }
}
```

but not the "starts in block" case:

```java
class C {
    public void f() {
        Foo foo;
        if(true) {
            foo = provider.foo();
        }
        something.qux();
        foo.bar();
    }
}
```

I see the same behavior for try-blocks, our actual use case. I'm using `if` for simplicity here.

**To Reproduce**
https://semgrep.dev/s/gordonwoodhull:ellipsis-control-flow

**What is the priority of the bug to you?**

P1. We have had great success applying vulnerability remediation fixes to code using semgrep, but we want more control over which patches are selected and where they are applied. I am hoping to use `pattern-inside` to specify a control-flow pattern which should contain the statement to be patched.

ievans:
@gordonwoodhull this looks like a use-case for the new taint mode, which was recently promoted from experimental (but is not yet GA). How does https://semgrep.dev/s/ievans:java-taint-cfg look to you :D 

Docs: https://semgrep.dev/docs/writing-rules/data-flow/

gordonwoodhull:
Thanks @ievans! That's a neat feature, and I guess it confirms my guess that ordinary patterns and ellipsis are syntactic and do not follow control flow.

Unfortunately I am looking for something with deeper integration with autofix. Specifically in our use case we would like to
- modify (or prepend statements) to the source statement (not just the sink)
- use bound metavariables in the `fix`

That's why I was hoping to use `pattern-inside`. I suppose I will fall back to writing external tools that use line numbers reported by semgrep.

[...]

PR #6764, opened 2022-12-30
discussion: pattern-inside-taint and reviving taint patterns

gordonwoodhull:
This is not a contribution yet, just a demo implementation of #3801 pattern inside a control flow, starting from the stale #5993.

Tests successfully in three languages, but it breaks many other tests, so it would need a flag (or a `where`?)

Instead of returning the range of the sink, it returns the range from the start of the source to the end of the sink.

I implemented pattern-inside-taint a year ago, but only found time now to catch up with all the exciting changes and hopefully contribute back.

Is there any interest in reviving taint patterns #5993? As soon as I saw the whiteboard in new pattern operators #5916, I realized that others had implemented this feature in a much better way than I had. 😉

### DIFF
--- a/semgrep-core/src/engine/Match_tainting_mode.ml
+++ b/semgrep-core/src/engine/Match_tainting_mode.ml
@@ -474,8 +474,10 @@ let pm_of_finding finding =
            Lazy.from_val
              (taint_trace_of_src_to_sink source tokens sink))
       in
+      let source_pm, _ = T.pm_of_trace source in
       let sink_pm, _ = T.pm_of_trace sink in
-      Some { sink_pm with env = merged_env; taint_trace }
+      Some { sink_pm with env = merged_env; taint_trace;
+                          range_loc = fst source_pm.range_loc, snd sink_pm.range_loc }
   | T.SrcToReturn _
   (* TODO: We might want to report functions that let input taint
    * go into a sink (?) *)
--- a/semgrep-core/tests/OTHER/rules/pattern-inside-taint.java
+++ b/semgrep-core/tests/OTHER/rules/pattern-inside-taint.java
@@ -1,0 +1,19 @@
+class C {
+    public void f() {
+        Foo foo;
+        // ok: pattern-inside-taint (before taint)
+        foo.qux();
+        if(true) {
+            foo = provider.foo();
+            // ruleid: pattern-inside-taint
+            foo.qux();
+        }
+        // ok: pattern-inside-taint (different object)
+        baz.qux();
+        // ruleid: pattern-inside-taint
+        foo.qux();
+        foo.bar();
+        // ok: pattern-inside-taint (after taint)
+        foo.qux();
+    }
+}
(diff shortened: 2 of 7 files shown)


IagoAbal:
@gordonwoodhull are you able to share more details about the real-world use of this?

IagoAbal:
cc @brandonspark

gordonwoodhull:
Sure.

When fixing vulnerabilities reported by a third-party tool, the flagged line may be before or after the line that needs to be fixed. `pattern-inside-taint` allows us to create a pattern which matches the range of the entire control flow, from the line that needs to be changed through the flagged line, and then we only autofix within that range, not unrelated lines outside the range that happen to match the autofix pattern.

It also allows us to bind metavariables, so that we know both the source-pattern and sink-pattern refer to the same local variables, and our autofix pattern doesn't match lines within the range that refer to other variables. (I understand the taint behavior may have changed here, but it doesn't seem to affect the tests in this PR.)

*Specific example*
For an XXE vulnerability, Veracode may flag `$BUILDER.parse()` but we need to search upward to where `$BUILDER` was instantiated, and change some settings on its factory before instantiation.

This is a feature of equal or greater power to SmPL's context nodes and `...` but as cool as that language was, it never took off outside Coccinelle for C. This is why we are using Semgrep instead.

We also have use cases where we use multiple Semgrep rules to change multiple places in a control flow. This requires extra tooling around Semgrep, but again it uses `pattern-inside-taint` to find those multiple places.

Thanks for your consideration!

brandonspark:
I think this is kinda a neat idea. In general, I'm all for cool new intuitively semantic features in the pattern syntax, and "capture the range of an entire taint flow" is pretty darn intuitive.

Some concerns:
- Is there the possibility of "backwards" ranges? How should that be handled? For instance, suppose we have the following file:
```python
def foo(x):
  sink(x)
  
def bar():
  x = source()
  foo(x)
```
The source appears lexically after the sink. Should the range still go through, and simply match everything from `sink(x)` through `x = source()`?
- What should happen if a taint source and sink are located in different files? (this would be in the case of our inter-file extension, DeepSemgrep) It appears in this case, we should really have no option except for producing no range. The engine simply won't be able to accommodate it. This is an easier choice than the above, but it does indicate maybe the above shouldn't produce a match either.

It seems to me that `pattern-inside-taint`, as conceived, currently uses "[returning] the range from the start of the source to the end of the sink" as a proxy for "returning the range that a taint flowed through". These are not always the same thing, as the above example demonstrates. I wonder how hard it would be to instrument a taint finding with the lines that it went through, and have `pattern-inside-taint` use this notion instead.

Essentially, it would mean that such a finding should be the union of many disparate findings -- each time a source flows to a  relevant call/sink, and every time a function's parameter flows to a relevant sink/other call. @IagoAbal thoughts?

brandonspark:
As an aside on taint patterns, I am very enthusiastic about reviving it (I've been meaning to for some time), but we found that in terms of our own rule-writing wants, it was only a "mild yes". As such, it's been shelved for now in favor of some other, more urgent things. I'd be very happy if we could bring it back, however, and I have no intention of letting it stay dead forever.

gordonwoodhull:
Thanks @brandonspark.

Indeed, I have only considered the intraprocedural case, as that is the scope of our project.

As you hint, there are also inaccuracies within a function, e.g. a taint that starts in an if-block should “skip over” the else block and create two disjoint ranges. These haven’t caused any practical problems for us, but should probably be addressed.

PR #6765, opened 2022-12-31
feat(julia): support the julia language

brandonspark:
image

## Commentary:
It's probably easier to read the file directly than it is to try and look at the diff.

The grammar changes included in `semgrep-julia` were so drastic from the last time I looked at it, that I had to take the previous `Parse_julia_tree_sitter.ml` I wrote and basically scrap it, because so much of it was now different. As such, perhaps it's best to look at this file with a clear eye.

I eliminated pretty much every single TODO, in favor of making sacrifices and tradeoffs that would allow us to parse for now, if at the cost of some semantic information. In particular, I've made liberal use of `Other`, as Julia's implementation means that quite often, things can appear in places that we do not expect. 

In particular, statements and expressions have really no meaning, macro-expressions can appear _everywhere_ an identifier can (making things like macro definitions, imports, etc more difficult), and it's annoying that things which should be patterns are represented in the grammar as just plain expressions. For the latter, I've just injected them into an `Other` for now. Comments have been left, hopefully clearing up the places that I have had to make these tradeoffs, other than just everywhere an `Other` appears.

## Test plan:
`./run-lang julia`

PR checklist:

- [] Documentation is up-to-date

### DIFF
--- a/parsing-stats/lang/julia/projects.txt
+++ b/parsing-stats/lang/julia/projects.txt
@@ -2,7 +2,8 @@
 # Top julia projects from GitHub, sorted by stars.
 # Created by /Users/brandonspark/semgrep/semgrep-core/src/ocaml-tree-sitter-core/scripts/most-starred-for-language.
 #
-https://github.com/JuliaLang/julia
+# This one probably has a lot of edge cases and grammar limitations we don't care about
+# https://github.com/JuliaLang/julia
 https://github.com/fonsp/Pluto.jl
 https://github.com/FluxML/Flux.jl
 https://github.com/JuliaLang/IJulia.jl
--- a/parsing-stats/lang/julia/projects.txt
+++ b/parsing-stats/lang/julia/projects.txt
@@ -1,0 +1,34 @@
+#
+# Top julia projects from GitHub, sorted by stars.
+# Created by /Users/brandonspark/semgrep/semgrep-core/src/ocaml-tree-sitter-core/scripts/most-starred-for-language.
+#
+https://github.com/JuliaLang/julia
+https://github.com/fonsp/Pluto.jl
+https://github.com/FluxML/Flux.jl
+https://github.com/JuliaLang/IJulia.jl
+https://github.com/SciML/DifferentialEquations.jl
+https://github.com/mitmath/18S191
+https://github.com/GenieFramework/Genie.jl
+https://github.com/jump-dev/JuMP.jl
+https://github.com/GiovineItalia/Gadfly.jl
+https://github.com/MakieOrg/Makie.jl
+https://github.com/TuringLang/Turing.jl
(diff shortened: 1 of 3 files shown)


parsing-stats/lang/julia/projects.txt
```diff
@@ -0,0 +1,34 @@
+#
+# Top julia projects from GitHub, sorted by stars.
+# Created by /Users/brandonspark/semgrep/semgrep-core/src/ocaml-tree-sitter-core/scripts/most-starred-for-language.
+#
+https://github.com/JuliaLang/julia
```
aryx: This we could potentially comment because it probably contact lots of test cases exercising edge cases
in the grammar that are not really representative of typical julia programs (we do that for other languages).

semgrep-core/src/parsing/tree_sitter/Parse_julia_tree_sitter.ml
```diff
@@ -1,47 +1,64 @@
 (* Yoann Padioleau
```
aryx: At this point you can put your name here.

semgrep-core/src/parsing/tree_sitter/Parse_julia_tree_sitter.ml
```diff
@@ -1,47 +1,64 @@
...
+   * directly to AST_generic.ml
+   *
+*)
 
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
 
+(* Disable warnings against unused variables *)
+[@@@warning "-26-27"]
```
aryx: This we want to remove ASAP.

semgrep-core/src/parsing/tree_sitter/Parse_julia_tree_sitter.ml
```diff
@@ -1,47 +1,64 @@
...
+
+(* Disable warning against unused 'rec' *)
+[@@@warning "-39"]
+
 type env = unit H.env
 
+let ( let* ) = Option.bind
```
aryx: I think if you open Common you get this let* too.

aryx (approved):
Great job!
Now we probably need to start writing alpha tests for alpha
to check if everything works fine. Probably need to extend the ocaml-tree-sitter-semgrep/lang/..../semgrep-julia/grammar.js and add more semgrep extensions (..., metavars, etc.)

aryx:
Regarding expressions as patterns, this is a recurring problem with other languages (e.g., Python). AST_generic_helpers.ml provide some helpers such as expr_to_pattern(), but it's quite limited.
AST_generic.ml has also a few useful helpers at the end of the file (e.g., one to build opcall).

aryx (approved):

PR #6788, opened 2023-01-06
feat!: unique names

brandonspark:
## What:
This PR adds in a new type, `unique_name`, which is meant to uniquely refer to a given name in a program. It is a `dotted_ident` and `tok`. It also changes `ResolvedName` to instead use this new type.

Motivation and details of this PR are in the accompanying PR, https://github.com/returntocorp/semgrep-proprietary/pull/435

## Test plan:
The actual changes which make use of this will be in `DeepSemgrep`, so I cannot add any substantial tests here. 

PR checklist:

### DIFF
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -308,7 +308,6 @@ and resolved_name_kind =
    * canonical_name, but we also want to match the pattern `foo.bar` so we will
    * store ['foo', 'bar'] as an alternate name.
    * *)
-
   | GlobalName of canonical_name * alternate_name list
 
 (* The enclosed token is not relevant for matching purposes.
@@ -1986,13 +1985,20 @@ let basic_id_info ?(hidden = false) resolved =
 
 (* TODO: move AST_generic_helpers.name_of_id and ids here *)
 
-let canonical_append {unqualified; _} {unqualified = unqualified2; tok} =
-  { unqualified = unqualified @ unqualified2; tok}
+let canonical_append { unqualified; _ } { unqualified = unqualified2; tok } =
+  { unqualified = unqualified @ unqualified2; tok }
 
-let dotted_to_canonical xs = 
-  (* TODO: Change if used on a function name which needs to be disambiguated *)
+let dotted_to_canonical xs =
+  (* WARNING: Do not use this function on a dotted which must be referred to uniquely! *)
   { unqualified = Common.map fst xs; tok = None }
-let canonical_to_dotted tid { unqualified = xs; _ } = xs |> Common.map (fun s -> (s, tid))
+
+(* Here, we ignore the enclosed token.
+   This token is useful for uniqueness purposes, but it is not supposed to be used for
+   matching.
+*)
+let canonical_to_dotted tid { unqualified = xs; _ } =
+  xs |> Common.map (fun s -> (s, tid))
+
 let alternate_name_to_dotted tid xs = xs |> Common.map (fun s -> (s, tid))
 
 (* ------------------------------------------------------------------------- *)
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -1985,6 +1985,18 @@ let basic_id_info ?(hidden = false) resolved =
 
 (* TODO: move AST_generic_helpers.name_of_id and ids here *)
 
+(* `canonical_append` is needed in one spot in the code, but it helps in that we can
+   just look here if we ever need to change something about the way that we deal with
+   canonical names.
+   Right now, we just throw away tokens, but if this is ever needed in the future,
+   we can easily change that.
+
+   I don't believe we should need the tokens right now in the canonical names, however,
+   as our default behavior for the time being will be to just use the `unqualified`
+   parts of the canonical names in the OSS engine.
+
+   The token will be used and passed back by the Pro engine, however.
+*)
 let canonical_append { unqualified; _ } { unqualified = unqualified2; tok } =
   { unqualified = unqualified @ unqualified2; tok }
 
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -382,7 +382,7 @@ and id_info = {
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
    *)
-  id_type : type_ option ref; [@equal fun _a _b -> true]
+  id_type : type_ option ref;
   (* type checker (typing) *)
   (* sgrep: this is for sgrep constant propagation hack.
    * todo? associate only with Id?
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -308,9 +308,17 @@ and resolved_name_kind =
    * canonical_name, but we also want to match the pattern `foo.bar` so we will
    * store ['foo', 'bar'] as an alternate name.
    * *)
+
   | GlobalName of canonical_name * alternate_name list
 
-and canonical_name = string list
+(* The enclosed token is not relevant for matching purposes.
+   `canonical_name` is a record of all the information necessary to
+   disambiguate two names from each other.
+   In particular, two canonical names may indicate two overloaded functions in Java,
+   which are named the same. Then, we need to use the function signature information
+   to disambiguate them from each other.
+*)
+and canonical_name = { unqualified : string list; tok : tok option }
 
 and alternate_name = string list
 [@@deriving show { with_path = false }, eq, hash]
@@ -374,7 +382,7 @@ and id_info = {
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
    *)
-  id_type : type_ option ref;
+  id_type : type_ option ref; [@equal fun _a _b -> true]
   (* type checker (typing) *)
   (* sgrep: this is for sgrep constant propagation hack.
    * todo? associate only with Id?
@@ -1978,10 +1986,14 @@ let basic_id_info ?(hidden = false) resolved =
 
 (* TODO: move AST_generic_helpers.name_of_id and ids here *)
 
-let dotted_to_canonical xs = Common.map fst xs
+let canonical_append {unqualified; _} {unqualified = unqualified2; tok} =
+  { unqualified = unqualified @ unqualified2; tok}
 
-let canonical_to_dotted tid xs =
-  xs |> Common.map (fun s -> (s, Parse_info.fake_info tid s))
+let dotted_to_canonical xs = 
+  (* TODO: Change if used on a function name which needs to be disambiguated *)
+  { unqualified = Common.map fst xs; tok = None }
+let canonical_to_dotted tid { unqualified = xs; _ } = xs |> Common.map (fun s -> (s, tid))
+let alternate_name_to_dotted tid xs = xs |> Common.map (fun s -> (s, tid))
 
 (* ------------------------------------------------------------------------- *)
 (* Entities *)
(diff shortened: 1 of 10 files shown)


semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -352,6 +351,15 @@ and qualifier =
...
   | QExpr of expr * tok
 
+(* This is not relevant for matching purposes.
+   `unique_name` is a record of all the information necessary to
+   disambiguate two `ResolvedName`s from each other.
+   In particular, two `ResolvedName`s may indicate two overloaded functions,
+   which are named the same. Then, we need to use the function signature information
+   to disambiguate them from each other.
+*)
+and unique_name = { dotted : dotted_ident; id : id_info_id option }
```
emjin: You've almost certainly explained this somewhere, but why use `id_info_id` rather than recording the type information with the name directly?

brandonspark: talked in person

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -328,7 +327,7 @@ and resolved_name_kind =
  *
  * sgrep-ext: note that ident can be a metavariable.
  *)
-type name = Id of ident * id_info | IdQualified of qualified_info
```
aryx: can you avoid to make this mutually recursive? I try hard to limit the size of the mutually recursive types in AST_generic.ml, to layer thing, so one can understand things gradually.

semgrep-core/src/core/ast/AST_generic.ml
aryx: We should not use id_info_id. This is not the purpose of id_info_id. See IdInfo.ml in deep-semgrep.
The main goal of id_info_id is to allow to modify back the generic AST while working on a view of the generic AST (SAST).

semgrep-core/src/core/ast/AST_generic.ml
```diff
@@ -362,7 +370,7 @@ and id_info = {
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
    *)
-  id_type : type_ option ref;
+  id_type : type_ option ref; [@equal fun _a _b -> true]
```
aryx: why do you add this?

aryx (changes requested):

aryx:
maybe we want to call unique_name globally_resolved_name, to
differentiate it from the resolved_name used in AST_generic.ml.

semgrep-core/src/core/ast/AST_generic.ml
brandonspark: I'm not sure I see the contradiction. Even if the purpose of `id_info_id` is to modify back the generic AST (because each identifier will contain its own `id_info_id`), I think these uses overlap -- it is necessary for us to produce a unique identifier that we can use to distinguish different instances of identifiers. One already exists, though. It's `id_info_id`. What would you propose instead?

brandonspark: Furthermore , it's highly convenient. At the point of adding an identifier to the local scope in `Naming_SAST`, we add it only once we are already holding its `id_info_id`. I don't see too much of a point of generating a fresh ID, for instance, which will just exactly correspond to that `id_info_id`. I see `id_info_id` as a pointer which identifies certain `DefStmt`s, that we can use for debugging purposes and manual inspection also.

semgrep-core/src/core/ast/AST_generic.ml
brandonspark: This might be a holdover from a previous PR, but basically we are failing to unify metavariables at some point because of the fact that we compare types (probably due to a bug in type inference).

I would be inclined to leave this in, knowing that it will prevent roadblocks down the line.

IagoAbal: If this causes problems we should have a reproducible test case, and probably better to open a separate PR.

semgrep-core/src/core/ast/AST_generic.ml
IagoAbal: An `id_info_id` identifies the variable occurrence not the variable.

libs/ast_generic/AST_generic.ml
```diff
@@ -365,7 +374,7 @@ and id_info = {
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
    *)
-  id_type : type_ option ref;
+  id_type : type_ option ref; [@equal fun _a _b -> true]
```
nmote: I agree that this should be in another PR. If this is needed because of this PR, I think that's a red flag. If it's unrelated, it should not be part of this PR.

libs/ast_generic/AST_generic.ml
```diff
@@ -303,7 +303,16 @@ and resolved_name_kind =
    * resolved_name, but we also want to match the pattern `foo.bar` so we will
    * store ['foo', 'bar'] as an alternate name.
    * *)
-  | ResolvedName of dotted_ident * dotted_ident list
+  | ResolvedName of unique_name * unique_name list
+
+(* This is not relevant for matching purposes.
```
nmote: By "This" you mean the `tok` specifically, right? The way it's phrased, it seems like the entire `unique_name` is not relevant for matching purposes, but I don't think that's true. Could you clarify?

libs/ast_generic/AST_generic.ml
```diff
@@ -303,7 +303,16 @@ and resolved_name_kind =
...
+   `unique_name` is a record of all the information necessary to
+   disambiguate two `ResolvedName`s from each other.
+   In particular, two `ResolvedName`s may indicate two overloaded functions,
+   which are named the same. Then, we need to use the function signature information
+   to disambiguate them from each other.
```
nmote, line 318: If I understand correctly, this disambiguation is currently (and maybe always will be?) only relevant for taint mode. For search mode, since we ignore the `tok`, there is no disambiguation. Is this correct? If so, could you explain? If not, could you add some more information here so that future readers don't come to the same misunderstanding?

nmote (changes requested):
Overall looks good, but I'd like to have another look once you address my comments. Please also update the PR summary now that you are using tokens.

nmote (approved):
Looks good! Please hold off on merging until the corresponding semgrep proprietary PR is ready to go, though.

aryx:
This is in conflict with https://github.com/returntocorp/semgrep/pull/6900 so please do not merge indeed

IagoAbal:
Is this PR currently doing anything apart from changing the formatting of some code? 😕

libs/ast_generic/AST_generic.ml
```diff
@@ -1991,8 +1998,21 @@ let basic_id_info ?(hidden = false) resolved =
 
 (* TODO: move AST_generic_helpers.name_of_id and ids here *)
 
-let dotted_to_canonical xs = Common.map fst xs
-let canonical_to_dotted tid xs = xs |> Common.map (fun s -> (s, tid))
+let canonical_append { unqualified; _ } { unqualified = unqualified2; tok } =
```
IagoAbal, line 2000: I think this needs some comment. Why do we append canonical names, and what's more important, why can we just throw away one of the tokens?

src/naming/Naming_AST.ml
```diff
@@ -840,7 +840,8 @@ let resolve lang prog =
                       let sid = SId.unsafe_default in
                       let rest_of_middle = Common.map fst rest_of_middle in
                       let canonical =
-                        xs @ dotted_to_canonical (rest_of_middle @ [ id ])
+                        canonical_append xs
```
IagoAbal, line 851: So here given `canonical_append`s definition we throw away `xs`'s token, and keep  the one from `dotted_to_canonical ...` which is guaranteed to be `None`. Is that fine?

libs/ast_generic/AST_generic.ml
```diff
@@ -1991,8 +1998,21 @@ let basic_id_info ?(hidden = false) resolved =
...
+  { unqualified = unqualified @ unqualified2; tok }
+
+let dotted_to_canonical xs =
```
IagoAbal, line 2003: Could we avoid this function? I only see it being used together with `canonical_append`, so why not combine them both?

brandonspark: I think I used it a fair few times, such as in `Naming_AST`, even without `canonical_append`. Basically, we need to produce a `canonical_name` either way to fit into the existing framework (since that's what `ImportedEntity` and co take in), so I don't really see a way of avoiding this function.

brandonspark: Oh, I see what happened. There's a few times it was already used (the function being pre-existing prior to this PR), so they won't show up in the diff. But there are like 10 occurrences of it in `Naming_AST`.

src/naming/Naming_AST.ml
brandonspark: It's meant to admit easy refactoring even if it is not. Should it end up being that we need that token around, we should just change `dotted_to_canonical` to embed the final token into the `canonical_name`.

I believe that this is not actually necessary here, since we are erring on the side of not allowing unique characteristics for these canonical names, in the open-source engine. These canonical names may pass back unique characteristics from the deep engine, but here they aren't so important.

libs/ast_generic/AST_generic.ml
brandonspark: There is exactly one point when we need to append canonical names, and that is purely because we are already given an `ImportedEntity` and must combine it with another canonical name. Since we already have the canonical name, we cannot go further upstream to intercept it and avoid appending while they are both canonical.

I think the point of leaving this here, however, is that we have one place we can easily change if we want to change that behavior. Per my other comment, if we want to keep the tokens, we can just change `dotted_to_canonical`. In any event, I don't believe we do need it.

I will add a comment explaining this.

libs/ast_generic/AST_generic.ml
aryx: I still don't understand. sid is supposed to be the thing we use to disambiguate two names (for example
in a function you can have multiple times the same name 'a' for a variable declared in nested blocks, but it does
not matter because we use a different sid for each of those. So why not use sid here for the same purpose, to make sure 2 references to 2 different overloaded function have a different sid ...

aryx: In fact, where do you ever use this tok in the matching code?

aryx (changes requested):

libs/ast_generic/AST_generic.ml
brandonspark: @aryx I believe that we came to the idea to use the `tok` for disambiguation together. Here's what Nat said on the other PR about this:

> Semgrep uses sids but I believe they are created during name resolution. We need to be able to construct Name.ts ahead of name resolution as part of creating the SIGs. I'm sure we could still come up with a scheme to use sids for this, and I think it would be valuable work, but I don't think it would be more difficult to do after merging this PR, and I don't see a reason to increase the complexity of this change beyond what is necessary.

I believe that if we wanted to use `sid`s, it would not be very hard to make a change to do so after the fact, but right now it would add a fair amount of complexity to the PR, which is already quite long. I have no real objections to it, but I would strongly prefer that we hold off until after the initial change, so it can be considered and evaluated separately.

aryx:
I don't think we need those changes at all. Everything can be done with sid.

libs/ast_generic/AST_generic.ml
aryx: Because if we use sids like they're supposed to be used, we don't need https://github.com/returntocorp/semgrep/pull/6788 at all

aryx:
This is closed in favor of https://github.com/returntocorp/semgrep-proprietary/pull/545
which relies on sid and remove the need for adding special code in semgrep for unique_names.

PR #6789, opened 2023-01-06
refactor!: make `sid`, `id_info_id`, and `Node_ID` abstract types

brandonspark:
## What:
This PR allows `sid`, `id_info_id`, and `Node_ID` to be represented by abstract types which are different to the compiler.

## Why:
Currently, `sid`, `id_info_id`, and `Node_ID` are represented by ints.

This is fine, it's an easy and classic way to generate fresh identifiers. But it leads to some measure of unsafety, because those ints can be tampered with, and mistakes can be written in our codebase around them. By making them abstract, we guarantee that we cannot confuse them for other `int`s, or otherwise use them in a way that is not precisely described by the `MkTemp` interface.

We also duplicate some measure of code whenever we allocate our refs and re-write the counter functions. I'm sure that this is not the last we need of temporary counters, so this also helps us in the future, by standardizing our approach.

## How:
`Common` now contains a functor, `MkId`. Because OCaml has generative functors, functors which generate abstract types _always_ generate _different_ types. Thus, the module resulting from `MkId ()` and `MkId ()` have different types to the compiler, which are abstract types that are represented in the back-end by ints.

These modules provide some useful functions and have their own private counters.

## Test plan:
`make test`

PR checklist:


semgrep-core/src/pfff/commons/Common.ml
```diff
@@ -1639,3 +1639,38 @@ let files_of_dir_or_files_no_vcs_nofilter xs =
...
+(*****************************************************************************)
+
+(* OCaml has generative functors.
+   This means that abstract types minted from different applications of the same
+   functor are _always_ different, even if they are constructed in the exact
+   same way.
+   So different instances of the module `MkTemp` will have distinct types.
+   This helps us not conflate types when we need a new kind of temp, and avoid
+   boilerplate code with refs and counters.
+*)
```
mjambon, line 1654: What's a temp?

semgrep-core/src/pfff/commons/Common.mli
```diff
@@ -474,3 +474,12 @@ type 'a smap = 'a SMap.t
...
 val jsoo : bool ref
+
+module MkTemp () : sig
+  type t [@@deriving show, eq, ord, hash]
+
+  val mk : unit -> t
+  val to_int : t -> int
+  val unsafe_default : t
+  val unsafe_reset_counter : unit -> unit
+end
```
mjambon, line 496: You should document the interface (try to do better than your elders...).

mjambon (approved):

semgrep-core/src/pfff/commons/Common.ml
brandonspark: renamed

aryx:
Nice, I like it.

PR #6794, opened 2023-01-10
Add "raw tree" constructs to generic AST

mjambon:
I'm using this with Apex on semgrep-proprietary. It works but I need to add a few tests so I'll make that other PR tomorrow.

This adds a `('tok, 'any) Raw_tree.t` type which is roughly a copy of the module of the same name provided by ocaml-tree-sitter with which the generated `Boilerplate.ml` compiles (see https://github.com/returntocorp/ocaml-tree-sitter-core/pull/58 for some explanations of the plan). The parametrization by the token type and by the type of custom nodes (e.g. `AST_generic.any`) allows us to use this type in language-specific ASTs before the final conversion to the generic AST (e.g. `AST_apex`).

This PR adds new kinds of nodes to the generic AST. This shouldn't create incompatibilities with semgrep-proprietary.

test plan: besides `make test` to check that everything still works in semgrep, the actual demonstration that this system works will come tomorrow with the semgrep-proprietary PR.

PR checklist:


aryx (approved):
I don't fully understand why we can't reuse the Raw_tree.t from ocaml-tree-sitter-core.
Tok should always be a Parse_info.t anyway.

mjambon:
> I don't fully understand why we can't reuse the Raw_tree.t from ocaml-tree-sitter-core.
> Tok should always be a Parse_info.t anyway.

ocaml-tree-sitter needs to test that the generated `Boilerplate.ml` compiles (and ideally works), but it doesn't depend on semgrep or pfff, using its own token type.

aryx:
Got it. Then I was worried about conflicting names because we now have 2 Raw_tree.ml, but it's actually packed under Tree_sitter_run.Raw_tree in ocaml-tree-sitter so all is good!
I think this Raw_tree thing is pretty cool to accelerate our support for new languages. Good job!

mjambon:
I think we don't need the token type to be a type parameter, though. It should become `'any Raw_tree.t` rather than `('tok, 'any) Raw_tree.t`. I'll change it before it's harder to change.

Issue #4477, opened 2022-01-05
Match Jupyter notebooks with python rules


nbrahms:
**Is your feature request related to a problem? Please describe.**
Enforce best practices in Jupyter notebooks.

**Describe the solution you'd like**
Automatically run python rules on notebooks.

**Describe alternatives you've considered**
- Use Semgrep join mode to extract code blocks (suffers from escaping & single-cell scanning)
- Use metavariable-pattern (suffers from escaping in blocks, single-cell scanning)
- Run `jupyter nbconvert --to script` first (requires jupyter installed in local environment)
- Run `jq -r '.cells[] | select(.cell_type == "code") | .source | add'` first (requires jq, may be sensitive to local jq build)

aryx:
This will help https://github.com/returntocorp/semgrep/issues/4477

emjin:
Unclear what we need here to close this. Is there a specific case we need to match or is this closed when we have extract mode rules in p/default and they work for most cases we could want?

aryx:
I guess when we have extract mode rules for jupyter extraction somewhere :) especially p/python and p/default.

PR #6797, opened 2023-01-10
Jupyter Notebook support in "extract mode"

jselvi:
This pull request creates a "json" flag in the "extract mode" experiment feature. This new flag allows to decode the extracted content as a JSON array and, as a result, it can be used to detect python vulnerabilities in Jupyter Notebook files.

This is part of a pull request modifying several repositories:
- https://github.com/returntocorp/semgrep/pull/6797
- https://github.com/returntocorp/semgrep-langs/pull/24
- https://github.com/returntocorp/semgrep-interfaces/pull/46
- https://github.com/returntocorp/semgrep-docs/pull/843

PR checklist:

### DIFF
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -209,7 +209,7 @@ type extract_spec = {
   dst_lang : Xlang.t;
   (* e.g., $...BODY, $CMD *)
   extract : MV.mvar;
-  json : bool;
+  transform : extract_transform;
 }
 
 (* Method to combine extracted ranges within a file:
@@ -217,6 +217,12 @@ type extract_spec = {
     - concatentate them together
 *)
 and extract_reduction = Separate | Concat [@@deriving show]
+
+(* Method to transform extracted content:
+    - either treat them as a raw string; or
+    - transform JSON array into a raw string
+*)
+and extract_transform = Unquote | ConcatJsonArray [@@deriving show]
 
 (*****************************************************************************)
 (* The rule *)
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -209,6 +209,7 @@ type extract_spec = {
   dst_lang : Xlang.t;
   (* e.g., $...BODY, $CMD *)
   extract : MV.mvar;
+  json: bool;
 }
 
 (* Method to combine extracted ranges within a file:
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -222,7 +222,8 @@ and extract_reduction = Separate | Concat [@@deriving show]
     - either treat them as a raw string; or
     - transform JSON array into a raw string
 *)
-and extract_transform = Unquote | ConcatJsonArray [@@deriving show]
+and extract_transform = NoTransform | Unquote | ConcatJsonArray
+[@@deriving show]
 
 (*****************************************************************************)
 (* The rule *)
--- a/semgrep-core/src/core/Rule.ml
+++ b/semgrep-core/src/core/Rule.ml
@@ -209,7 +209,7 @@ type extract_spec = {
   dst_lang : Xlang.t;
   (* e.g., $...BODY, $CMD *)
   extract : MV.mvar;
-  json: bool;
+  json : bool;
 }
 
 (* Method to combine extracted ranges within a file:
--- a/semgrep-core/src/engine/Match_extract_mode.ml
+++ b/semgrep-core/src/engine/Match_extract_mode.ml
@@ -142,12 +142,12 @@ let mk_extract_target dst_lang contents rule_ids =
   }
 
 (* Unescapes JSON array string *)
-let extract_code_from_json json =
+let convert_from_json_array_to_string json =
   let json' = "{ \"semgrep_fake_payload\": " ^ json ^ "}" in
   Yojson.Basic.from_string json'
   |> Yojson.Basic.Util.member "semgrep_fake_payload"
   |> Yojson.Basic.Util.to_list
-  |> Common.map (fun x -> x |> Yojson.Basic.Util.to_string)
+  |> Common.map Yojson.Basic.Util.to_string
   |> String.concat "\n"
 
 (*****************************************************************************)
@@ -310,8 +310,9 @@ let extract_and_concat erule_table xtarget rule_ids matches =
                 in
                 (* Convert from JSON to plaintext, if required *)
                 let contents =
-                  let (`Extract { Rule.json; _ }) = r.Rule.mode in
-                  if json then extract_code_from_json contents_raw
+                  let (`Extract { Rule.transform; _ }) = r.Rule.mode in
+                  if transform = ConcatJsonArray then
+                    convert_from_json_array_to_string contents_raw
                   else contents_raw
                 in
                 logger#trace
@@ -352,11 +353,12 @@ let extract_and_concat erule_table xtarget rule_ids matches =
                   fun ({ Parse_info.charpos; _ } as loc) ->
                     if charpos < consumed_loc.start_pos then map_contents loc
                     else
-                      (* For some reason, with json extraction option, it needs a fix to point the right line *)
+                      (* For some reason, with the concat_json_string_array option, it needs a fix to point the right line *)
                       (* TODO: Find the reason of this behaviour and fix it properly *)
                       let line =
-                        let (`Extract { Rule.json; _ }) = r.Rule.mode in
-                        if json then loc.line - consumed_loc.start_line - 1
+                        let (`Extract { Rule.transform; _ }) = r.Rule.mode in
+                        if transform = ConcatJsonArray then
+                          loc.line - consumed_loc.start_line - 1
                         else loc.line - consumed_loc.start_line
                       in
                       map_snippet
@@ -415,8 +417,9 @@ let extract_as_separate erule_table xtarget rule_ids matches =
              in
              (* Convert from JSON to plaintext, if required *)
              let contents =
-               let (`Extract { Rule.json; _ }) = erule.mode in
-               if json then extract_code_from_json contents_raw
+               let (`Extract { Rule.transform; _ }) = erule.mode in
+               if transform = ConcatJsonArray then
+                 convert_from_json_array_to_string contents_raw
                else contents_raw
              in
              logger#trace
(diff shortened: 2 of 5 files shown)


semgrep-core/src/core/Rule.ml
```diff
@@ -209,6 +209,7 @@ type extract_spec = {
   dst_lang : Xlang.t;
   (* e.g., $...BODY, $CMD *)
   extract : MV.mvar;
+  json : bool;
```
aryx: I think there are a few possible scenario where the extracted code is not in final form and need to be further
transformed. 'json:' is actually not a good name because in this case it's actually a json string array.
Maybe we could have a field like 'transform: concat_json_string_array' where we could have a list of possible
choices for this new transform: field. we could also have 'transform: unquote_string'.
Ideally the extract rules would have done the job, and reduce: concat could do things, but maybe sometimes it's hard and an additional transform: field could help.

semgrep-core/src/engine/Match_extract_mode.ml
```diff
@@ -293,12 +302,18 @@ let extract_and_concat erule_table xtarget rule_ids matches =
...
          (* Read the extracted text from the source file *)
          |> Common.map (fun { start_pos; start_line; start_col; end_pos } ->
-                let contents =
+                let contents_raw =
                   Common.with_open_infile xtarget.Xtarget.file (fun chan ->
                       let extract_size = end_pos - start_pos in
                       seek_in chan start_pos;
                       really_input_string chan extract_size)
                 in
+                (* Convert from JSON to plaintext, if required *)
```
aryx, line 317: add a TODO like:
TODO: note that this extract_code_... mess up the position information so match on the
extract code would not correspond to precise location in the original file.

aryx (changes requested):

aryx:
This looks pretty good.

semgrep-core/src/core/Rule.ml
jselvi: Yeah, I thought something like that, based on a conversation that we had in Slack but, since someone mentioned that this feature may be redesigned by the end of the year, I didn't think it was critical to do something more complex, since it could be completely changed in some months.

semgrep-core/src/engine/Match_extract_mode.ml
jselvi: In theory that is managed by the previous extract mode feature. I just added a fix in that location for the JSON decoding.

semgrep-core/src/core/Rule.ml
aryx: still, 'json:' is currently a bad name, because it should really be concat_json_string_array: true, at which point
doing an enum with transform: concat_json_string_array | unqote_string seems better.

semgrep-core/src/engine/Match_extract_mode.ml
```diff
@@ -141,6 +141,15 @@ let mk_extract_target dst_lang contents rule_ids =
...
   }
 
+(* Unescapes JSON array string *)
+let extract_code_from_json json =
+  let json' = "{ \"semgrep_fake_payload\": " ^ json ^ "}" in
+  Yojson.Basic.from_string json'
+  |> Yojson.Basic.Util.member "semgrep_fake_payload"
+  |> Yojson.Basic.Util.to_list
+  |> Common.map (fun x -> x |> Yojson.Basic.Util.to_string)
+  |> String.concat "\n"
```
kopecs:
```suggestion
(* Unescapes JSON array string *)
let extract_code_from_json json =
  Yojson.Basic.from_string json
  |> Yojson.Basic.Util.to_list
  |> Common.map Yojson.Basic.Util.to_string
  |> String.concat "\n"
```
I'm not sure what the purpose of the object we immediately unpack here would be. I don't have a full dev setup at the moment, but I think it can be simplified with so, from some minor testing with Yojson. Also, I don't recall what the prevailing style in semgrep is, but `(fun x -> x |> Yojson.Basic.Util.to_string)` is just a wordy way to write `Yojson.Basic.Util.to_string`, so I might consider that instead.

kopecs: I also think the name of this function could be improved to be more descriptive of what it actually is doing, given it's not really doing anything to get the code from the json, just processing the array.

semgrep-core/src/engine/Match_extract_mode.ml
```diff
@@ -337,7 +352,13 @@ let extract_and_concat erule_table xtarget rule_ids matches =
                   fun ({ Parse_info.charpos; _ } as loc) ->
                     if charpos < consumed_loc.start_pos then map_contents loc
                     else
-                      let line = loc.line - consumed_loc.start_line in
+                      (* For some reason, with json extraction option, it needs a fix to point the right line *)
+                      (* TODO: Find the reason of this behaviour and fix it properly *)
```
kopecs, line 365: If I had to guess, this is probably an artefact of how the JSON is formatted. Is it something like the following?
```
code: [
  "contents",
  "more contents"
]
```
(looks like yes, from line 27 in the example file below)

Relying on Yojson to do the parsing probably doesn't help here, because we wouldn't have an easy way to distinguish the formatting and corresponding range mapping we'd have to do.

semgrep-core/src/engine/Match_extract_mode.ml
```diff
@@ -386,22 +407,35 @@ let extract_as_separate erule_table xtarget rule_ids matches =
...
                m.rule_id.id m.file start_extract_pos end_extract_pos contents;
              (* Write out the extracted text in a tmpfile *)
              let (`Extract { Rule.dst_lang; _ }) = erule.mode in
              let target = mk_extract_target dst_lang contents rule_ids in
+             (* For some reason, with json extraction option, it needs a fix to point the right line *)
+             (* TODO: Find the reason of this behaviour and fix it properly *)
              let map_loc =
-               map_loc start_extract_pos line_offset col_offset
-                 xtarget.Xtarget.file
+               let (`Extract { Rule.json; _ }) = erule.mode in
```
kopecs: We already have to get some fields out of the rule a couple lines prior, so maybe something like this instead?
```ocaml
let (`Extract { Rule.dst_lang; json; _ }) = erule.mode in (* << just getting both members here *)
```
and then delete new line 432.

semgrep-core/src/engine/Match_extract_mode.ml
kopecs: At a glance, I don't think that adjustment is robust enough to accurately handle what it would need to, but I'm not in a good position to test. Ideally it shouldn't rely on the code array just happening to be formatted in such a way each string is on it's own line, etc..

Are columns also not slightly off in the instances where we needed to delete characters as part of the unescaping?

kopecs (commented):
Had a bit of free time, so I look a quick look. I might've missed some stuff, or have concerns which aren't an issue; I didn't do any testing.

Overall looks pretty good. I think we could figure out some way to handle this sort of thing which doesn't turn into 50 different config knobs, but that's probably a future problem, and more complicated than immediately needed.

For handling the correct source mapping I think there's probably a fair bit more work, but if this is good enough we could punt on it for now. I don't see the mapping working correctly until we know specifically what bytes got deleted from the extracted range, but it seems like things might be workable without doing that here if we're assuming ipynb files are always formatted in a given way (I think columns will still be off, but at that point we might as well fix it all).

semgrep-core/src/core/Rule.ml
jselvi: This has been reimplemented as suggested.

semgrep-core/src/engine/Match_extract_mode.ml
jselvi: Yeah, columns are not very accurate. @aryx mentioned that shouldn't be a problem right now.

semgrep-core/src/engine/Match_extract_mode.ml
jselvi: Updated this as suggested. I kept the "semgrep_fake_payload" trick because it's necessary. If not the extracted content is not a valid JSON and it can't be processed by Yojson.

semgrep-core/src/engine/Match_extract_mode.ml
jselvi: I tried to code a simple json parser from scratch, but I felt like it's reinventing the wheel. Also, these have been my first lines in Ocaml, so I didn't feel I was able to code a good quality JSON parser. Anyway, mapping lines and columns will be something challenging even if you implement your own parser. It could be even more challenging in future transforms.

semgrep-core/src/engine/Match_extract_mode.ml
jselvi: Done.

aryx:
This would help https://github.com/returntocorp/semgrep/issues/4477

semgrep-core/src/core/Rule.ml
```diff
@@ -209,6 +209,7 @@ type extract_spec = {
   dst_lang : Xlang.t;
   (* e.g., $...BODY, $CMD *)
   extract : MV.mvar;
+  transform : extract_transform;
```
aryx, line 212: We need to make this an 'extract_transform option' or add a NoTransfo in extract_transform.

aryx (changes requested):

semgrep-core/src/core/Rule.ml
jselvi: should I also remove the "unquote_string" then? I thought that was the "do nothing" option.

aryx: no it's meant for scenario that I remember where we extract code, but the extracted code is a big string and
to make it valid we need to unquote it.

jselvi: Done.

aryx (approved):

semgrep-core/src/parsing/Parse_rule.ml
```diff
@@ -1220,12 +1237,17 @@ let parse_mode env mode_opt (rule_dict : dict) : R.mode =
       in
       (* TODO: determine fmt---string with interpolated metavars? *)
       let extract = take rule_dict env parse_string "extract" in
+      let transform =
+        take_opt rule_dict env parse_string_wrap "transform"
+        |> Option.map (parse_extract_transform ~id:env.id)
+        |> Option.value ~default:R.Unquote
```
aryx, line 1243: hmm, we probably want NoTransform as a default. I'll fix that.

PR #6804, opened 2023-01-11
Simplify the Raw_tree.t type by removing the 'tok parameter.

mjambon:
1. There's no need to make the token a type parameter. See comments on this earlier PR: https://github.com/returntocorp/semgrep/pull/6794#issuecomment-1377945165
2. The type of Token nodes is now equivalent to `string wrap` as defined in the generic AST, as usual. This makes the matcher work as expected because it's based on comparing the string contents and ignores the tok.

test plan: relying on `make test`. The main subtlety is the custom hash and comparison functions provided by the new `Tok` module.

PR checklist:


aryx (approved):
Simpler, better.

PR #6815, opened 2023-01-11
fix semgrep.jsonnet, problem using new syntax

aryx:
We should find why the regression, but in the mean time
I've commented the problematic rule

test plan:
make check_with_docker

PR checklist:

### DIFF
--- a/semgrep.jsonnet
+++ b/semgrep.jsonnet
@@ -9,21 +9,22 @@ local test_code_globs = ["Unit_*.ml", "Test_*.ml"];
 local less_important_code_globs = ["spacegrep/*", "experiments/*", "scripts/*"];
 
 local semgrep_rules = [
-  #TODO: this rule could be moved in pfff/semgrep.yml at some point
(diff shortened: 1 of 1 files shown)


semgrep.jsonnet
```diff
@@ -9,21 +9,22 @@ local test_code_globs = ["Unit_*.ml", "Test_*.ml"];
...
-    severity: "ERROR",
-    message: |||
-        It is easy to forget to close `open_in` with `close_in`.
-        Use `Common.with_open_infile()` instead.
-    |||,
-  },
+#TODO: does not parse anymore!!
+#  #TODO: this rule could be moved in pfff/semgrep.yml at some point
+#  { id: "pfff-no-open-in",
+#    match: { or: ["open_in_bin ...", "open_in ..."]},
```
aryx, line 15: not sure what changed recently @brandonspark but this rule does not work anymore.
I've tried to replace the or: with any: but same problem.

PR #6817, opened 2023-01-11
feat(SSC): Rewrite all the lockfile parsers

mmcqd:
This is a _very_ big PR, since it involves a major rewrite. Perhaps it should be split into multiple PRs.
What this PR does:
* Introduces a new dependency: parsy, a parser combinator library
* Adds type stubs for parsy, since it has no types
* Rewrite all the existing parsers (except the Cargo.lock parser) using this library
* Enable line numbers for every lockfile format we support
* Enable transitivity detection for Gradle and Gem. This is not enable for Go because we should completely change how we handle go dependencies, but this should be a separate PR
* Replace extremely terrible error messages with things like `Failed to parse file with Exception: expecting "[[package]]" at 15:32`
* Remove support for rust/Cargo.lock. No one is using it, we don't even list it as supported in our docs, and we have no rules for it
* Replace the gradle test with a better one that includes a manifest file
* Add more weird syntax to the requirements.txt test

All the little line number changes in the snapshots are expected. They are now more accurate.

PR checklist:


aryx:
+1,649 −1,288 sad that the line number increases even though you use more elegant parsing.
Hopefully it's also more robust.

mmcqd:
> +1,649 −1,288 sad that the line number increases even though you use more elegant parsing. Hopefully it's also more robust.

A lot of the extra lines are actually imports, since I separated the parsers out into many new files

mmcqd:
Closing in favor of opening a series of smaller PRs. This is unreviewable.

PR #6598, opened 2022-11-23
feat(java): propagate singly assigned private vars

brandonspark:
## What:
We do not allow constant propagation on singly assigned variables in Java. This is only safe if those variables are `private`, so they are not accessed outside of the class itself.

## Why:
We should do this, it will help write Java rules.

## How:
Added it as a condition in constant propagation.

## Test plan:
`make test`

Closes PA-2230

PR checklist:

- [X] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [X] Tests included or PR comment includes a reproducible test plan
- [X] Documentation is up-to-date
- [X] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [X] Change has no security implications (otherwise, ping security team)

If you're unsure on any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #6825, opened 2023-01-12
const-prop: Fix var_stats for l-values such as this.x

### LINKED ISSUE #6793: Miss detection of const value for a Java class variable
**Describe the bug**
Since version [0.123.0](https://github.com/returntocorp/semgrep/releases/tag/v0.123.0) our Java rules about hardcoded conditional are broken. If we set a default value to a private class variable and using a setter method to change it, semgrep did not detect possibile change of it. If we change the function from ```this.test = bool;``` to ```test = bool;``` semgrep works perfectly.

**To Reproduce**
https://semgrep.dev/s/KNb9

**Expected behavior**
Detect that the variable is not const.

**Environment**
Example on semgrep.dev

**Use case**
Using semgrep in ci pipline.

IagoAbal:
In 0.123.0 we allowed const-prop to regard a private class attribute in Java as constant, if it was assigned just once. But when we looked for assigned variables we didn't consider l-values such as `this.x`, so `this.x = 42` was not being counted as an assignment to `x`.

And, refactor code so that it is cleaner plus it will be able to handle e.g. nested tuples in an LHS.

Closes #6793
Follows: 621e5ea3b1e ("propagate isngly assigned private java vars (#6598)")

test plan:
make test # added one test

PR checklist:


.github/CODEOWNERS
```diff
@@ -1,3 +1,5 @@
 
 # Matches all files in the .github/ dir, recursively.
 .github/ @spencerdrak @cgdolan @aryx @bmahe
+semgreo-core/src/analyzing @IagoAbal
```
aryx: semgreo?

semgrep-core/tests/rules/cp_private_class_attr1.yaml
```diff
@@ -0,0 +1,14 @@
+rules:
+  - id: hardcoded-conditional-copy
+    patterns:
+      - pattern-either:
+          - pattern: if (false) { ... }
+    message: This if statement always results in the same action and is unnecessary.
+    languages:
+      - java
+    severity: ERROR
+    metadata:
```
aryx: do we need those metadata in our test files?

aryx (approved):

PR #6836, opened 2023-01-13
fix(perf): stage visitor in `extract_ranges` better

brandonspark:
## What:
This PR stages the computation of the `extract_ranges` visitor such that `mk_visitor` is called only once.

## Why:
Currently, `extract_ranges` sets up a ref, along with some subsidiary helpers that reference it, and then uses those to call the expensive function `mk_visitor`. It then calls that same visitor and returns the contents of the cell.

If we are visiting with essentially the same "functionality", it is wasteful to call `mk_visitor` for every single time that we try to `extract_ranges` from an `any`. This would mean that we call it every single time we match, essentially.

## How:
Moved the computation of the visitor to before the binding of the function value to `extract_ranges`.

## Performance:
I observed, via the `deepsemgrep-sqli-rules.yaml` on the `OpenClinica` repository, the following statistics with `semgrep` normally running:
- `2173M` of memory usage, taking `518.12s` of time to run.

With this change, I observed the following performance:
- `2188M` of memory usage, taking `366.53s` of time to run.

I consider the memory usage variance to be well within normal deviation. The time increase is significant though -- 30% speedup.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/semgrep-core/src/core/ast/Visitor_AST.ml
+++ b/semgrep-core/src/core/ast/Visitor_AST.ml
@@ -1444,7 +1444,8 @@ let first_info_of_any any =
 (*****************************************************************************)
 
 (*s: function [[Lib_AST.extract_info_visitor]] *)
-let extract_ranges recursor =
+let extract_ranges :
+    AST_generic.any -> (PI.token_location * PI.token_location) option =
   let ranges = ref None in
   let smaller t1 t2 =
     if compare t1.PI.charpos t2.PI.charpos < 0 then t1 else t2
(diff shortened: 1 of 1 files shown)


nmote (approved):
Great find!

aryx:
noice!

IagoAbal:
> I observed, via the deepsemgrep-sqli-rules.yaml on the OpenClinica repository, the following statistics with semgrep normally running:

Perhaps I'm nit picking but this change could be tested with Semgrep itself without DeepSemgrep. Just running `p/default` on some large repo should show a perf gain.

PR #6853, opened 2023-01-16
sym-prop: Prevent infinite loops during matching

IagoAbal:
Previously, naming-related bugs have introduced circular references in symbolically-propagated values, leading to infinite loops during matching. There are some sanity checks in place already, but here we introduce a last-resort check to avoid falling into infinite loops. (Q: Should we remove other checks?) We restrict the number of times that the engine can follow symbolically-propagated values during matching.

Closes PA-2324

test plan:
Set `max_NESTED_SYMBOLIC_PROPAGATION` to 1 or 2 and run `make test`, a bunch of tests will fail.

PR checklist:

### DIFF
--- a/changelog.d/pa-2324.fixed
+++ b/changelog.d/pa-2324.fixed
@@ -1,0 +1,4 @@
+In rare situations, mainly in DeepSemgrep and related to naming bugs, the use of
+symbolic propagation could make Semgrep fall into an infinite loop during matching.
+This has been fixed by bounding the number of times that Semgrep can follow
+symbolically-propagated values.
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -389,7 +389,7 @@ let m_deep (deep_fun : G.expr Matching_generic.matcher)
 (* In Match_patterns.match_rules_and_recurse we create a new env for each attempt
  * at matching a "mini-rule" against a sub-AST, so this bound doesn't need to be
  * large. Our test suite only needs it to be >= 3 to pass! *)
-let max_SYMBOLIC_PROPAGATION = 100
+let max_NESTED_SYMBOLIC_PROPAGATION = 50
 
 let m_with_symbolic_propagation ~is_root f b tin =
   if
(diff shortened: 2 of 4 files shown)


changelog.d/pa-2324.fixed
```diff
@@ -0,0 +1,4 @@
+In rare situations, mainly in DeepSemgrep and related to naming bugs, the use of
+symbolic propagation could make Semgrep fall into an infinite loop during matching.
+This has been fixed by bounding the number of times that Semgrep can follow
+symbolically-propagated values.
```
nmote, line 4: Nice clear description

src/matching/Generic_vs_generic.ml
```diff
@@ -406,14 +415,21 @@ let m_with_symbolic_propagation ~is_root f b =
...
               "Aborting symbolic propagation: Circular reference encountered \
                (\"%s\")"
               id;
-            fail ())
-          else f b1
-      | ___else___ -> fail ())
-    ~else_:(fail ())
+            fail () tin)
+          else (
+            incr tin.followed_sym_vals;
```
nmote: I'd be more comfortable if this weren't mutable, and we instead allocated a new `tin` with the updated value. I think it would make it a lot more clear that the limit has to do with the depth of the symbolic propagation. Right now, I would have to understand exactly when we allocate a new `tin` in order to understand at what point we give up on symbolic propagation.

I'd probably have to read quite a bit of code to convince myself that this wouldn't cause a spurious failure when we perform a lot of unrelated symbolic propagation operations as part of one match attempt.

Is there a specific reason that you've done it this way? If not, I'd appreciate it if you switched it to be immutable.

nmote (approved):

changelog.d/pa-2324.fixed
IagoAbal: Thanks!

src/matching/Generic_vs_generic.ml
IagoAbal: Good point! Initially I tried to make it immutable but it looked ugly, I think I just overcomplicated myself. I gave it another try after your comment and I got it done in a simple way. 🤷‍♂️

nmote:
Also, to directly answer your question, I'd be open to the removal of the check that I added for directly circular references, but I would prefer to leave it. I think it's good to catch egregious cases early, especially since I've seen cases where we seem to get some sort of exponential blowup related to symbolic propagation loops, and even 2^50 is quite large.

PR #6878, opened 2023-01-18
fix: Make `$F(x)` match `eval(x)`

### LINKED ISSUE #6877: eval cannot be matched in metavariable
**Describe the bug**
eval call cannot be matched in metavariable because the "eval" string might be a special one in semgrep core.

**To Reproduce**
https://semgrep.dev/s/D0Aq

**Reference**
https://semgrep.slack.com/archives/C0242656NQ2/p1674065402080939?thread_ts=1674063446.010749&cid=C0242656NQ2

nmote:
`eval` is an `IdSpecial` in the generic AST, instead of a regular identifier. However, users expect it to match like an ordinary function call, since it actually is.

I previously added special cases for `this`, `super`, `self`, `parent`, and `require`, which are all `IdSpecial`s to make sure that they all match. I left a TODO for `eval` since it seemed like it should probably match but was unrelated to what I was doing at the time. This fix simply addresses that TODO and adds a test.

Fixes #6877

Test plan: Automated tests

PR checklist:


brandonspark (approved):
the todos always come back

aryx:
they always come back :)

Issue #5487, opened 2022-06-10
Semgrep --test should scan test file regardless of `paths` directive


ben-elttam:
**Is your feature request related to a problem? Please describe.**

When running `semgrep --test` for a rule file `rule.yml` it can find the semgrep unit test file, like `rule.test.yml`, or `rule.java`. If the rule is using `paths.include` directive, the unit test file must be included in the list.

https://semgrep.dev/docs/writing-rules/rule-syntax/#limiting-a-rule-to-paths

e.g. the rule in file `web-xml.yml` must have:

```yaml
paths:
  include:
    - web.xml
    # unit test file
   - web-xml.xml  # <---
```

Without allow-listing the test file, the rule won't run, the rule(s) won't be tested.

**Describe the solution you'd like**

Since semgrep is already finding the file for the unit test to run it, I would like in test mode to magically bypass the `paths.include`, and scan the file.

e.g. I don't want to have to manually add the unit test files to the allow-list, I want them automagically included only when in test mode.

**Describe alternatives you've considered**

**Use case**
This will enable better Quality of Life.

**Additional context**

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1488/semgrep-test-should-scan-test-file-regardless-of-paths-directive. _Note_: this link is for r2c use only and is not accessible publicly.

fopina:
I was almost opening a duplicate from this but I'd like to add an alternative:

Instead of `--test` simply ignoring `paths` constraints (which sounds reasonable and probably works well), there's the option to leverage the fact that everything matching `rule-name*` is initially assessed as test code.

So using a file structure such as below would also work:

```
./my-rule-a/mymodule/views.py
./my-rule-a.yaml
```

This would allow not only to make sure path constraints are not broken (in some rule definition change / regression) but also other more complex scenarios (either for join-mode or deepsemgrep)

Sample rule/test files in https://github.com/fopina/delme/tree/main/semgrep_cannot_test_path_includes

ben-elttam:
> Instead of --test simply ignoring paths constraints (which sounds reasonable and probably works well), there's the option to leverage the fact that everything matching rule-name* is initially assessed as test code.

This is what I wanted. If `rule-name.yml` is going to find `rule-name*` as a test, I want that to be scanned regardless of the rule's `paths.include` directive.

I do like the idea of the tests being a directory tree, which could have many files.

fopina:
> This is what I wanted

[...]

Issue #6880, opened 2023-01-18
rule ID cannot contain dots


fopina:
**Describe the bug**
A rule id cannot contain any dots, as that will break (at least) the `--test` framework

```
rules:
 - id: my.rule.a
```

**To Reproduce**

* Download 2 rules and 2 test code files from https://github.com/fopina/delme/tree/main/semgrep_no_dots_in_rule_name
* Follow the [README](https://github.com/fopina/delme/blob/main/semgrep_no_dots_in_rule_name/README.md):
  * Run `semgrep --validate --config ./` and it does not fail
  * Run `semgrep --test ./` and it does fail with assertions mismatch 

**Expected behavior**

If dots are not allowed in rule IDs, they should error immediately such as in an invalid schema or be ignored entirely. At least `--validate` (or `--test`) should check that.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Environment**
macOS, semgrep 1.3.0

**Use case**
Fixing this bug will document and confirm that dots are not expected in the rule id. Otherwise there are other bugs to be fixed.

**Other**

At least for `--test` I believe the first issue is in https://github.com/returntocorp/semgrep/blob/5bd4f3e0fb6005a4ddacd9cdc1e3a086c8088764/cli/src/semgrep/test.py#L312-L321

It removes suffix after suffix cause the rules in this example (`my.rule.a.yaml` and `my.rule.b.yaml`) to become the same identifier `my` and picking up on both test file for both rules.

I really like the suffix trimming as it allows us to have several test files for the same rule (required in some rules but also nice for readability in some more complex rules):

```
my-rule.yaml
my-rule.py
my-rule.other-test.py
```

But this is not documented in https://semgrep.dev/docs/writing-rules/testing-rules/, only the `.test.yaml` suffix specifically for the YAML language.

So with this issue I'd expect to confirm that:
* dots are indeed not expected
* it is ok to continue to use this `my-rule.whatever.ext` even though it's not documented (could it be documented?)

ben-elttam:
Rule IDs also cannot have spaces. This is allowed in some parts but causes other parts to fail. So, I generally avoid them. It would be good if this behaviour was handled consistently, either tolerated everywhere, or caught early and rejected everywhere.

I'm not sure if this is still the case.
[...]

PR #6883, opened 2023-01-19
Ability to test rules with `paths`

fopina:
PR checklist:

This PR tries to solve https://github.com/returntocorp/semgrep/issues/5487 by allowing _test directories_ to be processed instead of single test files.

### unit test
New unit test for `relatively_eq` included in `tests/unit/test_semgrep_test.py`

### manual e2e
Testing done by using the [same test repository](https://github.com/fopina/delme/tree/main/semgrep_cannot_test_path_includes):

```
$ python -m semgrep --test /Users/fopina/Documents/delme/semgrep_cannot_test_path_includes
0/1: 1 unit tests did not pass:
	✖ my-rule-a
	missed lines: [3], incorrect lines: [4]
	test file path: /Users/fopina/Documents/delme/semgrep_cannot_test_path_includes/my-rule-a/mymodule/views.py


No tests for fixes found.
```

**Note**

`my-rule-a.py` needs to be removed otherwise output is _blocked_ by

```
Found rule id mismatch - file=.../semgrep_cannot_test_path_includes/my-rule-a.py 'ruleid' annotation with no YAML rule={'my-rule-a'}
Failing due to rule id mismatch. There is a test denoted with 'ruleid: <rule name>' where the rule name does not exist or is not expected in the test file.
```


fopina:
Just noticed that `parts[0]` cannot be used without first removing common prefix (such as `language/`

I'll update tomorrow and include an unit test

fopina:
fixed the matching ✅ 
I've also refactored `relatively_eq` to make it simpler for this change and because it was only used in this file and in this context (comparing targets with configs).
I've renamed the its parameters to make its restricted use clearer as well

brendongo:
@fopina I'm not sure I understand the intention of the change could you explain more?

fopina:
@brendongo I avoided duplicating the problem as it is linked to an issue.

Did you miss that or you mean the problem is not clearly stated in the issue?

brendongo:
Oh I somehow missed that 😓

brendongo:
This looks good. Do you mind adding a changelog entry

fopina:
> Oh I somehow missed that

No worries, I just wasn't sure if it was the case. It's easy for a reporter to assume all context is provided in the description when it's not!

Added the changelog entry, I hope that's ok, not that many entries to use as example

emjin (approved):
Thanks for this change and the unit tests! If you have time, we would welcome an update to https://github.com/returntocorp/semgrep-docs

emjin:
Need to look into how best to run pre-commit. @brendongo do you know?

fopina:
> we would welcome an update to https://github.com/returntocorp/semgrep-docs

Would it be ok to do it after this is merged?

Mostly because this PR also depends on the undocumented behavior described in https://github.com/returntocorp/semgrep/issues/6880 so I was thinking of opening a PR (to docs) to add both things.

Just needed confirmation on #6880 that it is ok to make the behaviour official and not meant to be removed/"fixed"

> how best to run pre-commit

I know this wasn't towards me and probably I miss context on what you need but I usually just run pre-commit with the failed id from the pipelines

in this case
```
pre-commit run --all end-of-file-fixer
```

I don't use `pre-commit install` to avoid having all checks running on each commit and sometimes I end up forgetting to run `pre-commit run --all` manually before last commit...

emjin:
> Would it be ok to do it after this is merged?

Yup! Re #6880, I'm happy for that to be official behavior. I'll leave the ticket open until you or we document it!

> I usually just run pre-commit 

Wasn't sure if you were able to run pre-commit on your end. Thanks for running it!

ben-elttam:
> I don't use `pre-commit install` to avoid having all checks running on each commit and sometimes I end up forgetting to run `pre-commit run --all` manually before last commit...

If you do install it and want to avoid the hooks, you can use `git commit -n` (`--no-verify`) to skip the hooks.

ben-elttam:
I've tested this and it works well. The only thing I noticed is you can't have a test file that is all `# ok: <ruleid>` comments. It will trigger the error:

> Found rule id mismatch - file=.../semgrep_cannot_test_path_includes/my-rule-a.py 'ruleid' annotation with no YAML rule={'my-rule-a'}
Failing due to rule id mismatch. There is a test denoted with 'ruleid: <rule name>' where the rule name does not exist or is not expected in the test file.

The file can simply be included with no comments, it will be scanned and should have no matches, so passes. It won't show up in the list of test file path(s), but it is used. If you modify the file so it matches, the test will fail and the file will be listed.

fopina:
The "single-ok" issue already existed though using this feature will make it more visible/common.

A colleague is actually trying to address it in https://github.com/returntocorp/semgrep/pull/6987 ⚡

PR #6888, opened 2023-01-19
cli: Switch to Semgrep PRO binaries (deprecate DeepSemgrep)

IagoAbal:
Closes PA-2389

test plan:
```
% semgrep -c p/default semgrep/parsing-stats/lang/java/tmp/square-retrofit
  # ^ try this with deep-semgrep installed, then remove deep-semgrep and try again,
  # then run `semgrep install-semgrep-pro` and try one more time
```

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -42,7 +42,7 @@ def run_install_semgrep_pro() -> None:
     (semgrep_pro_path, deep_semgrep_path) = determine_semgrep_pro_path()
     if deep_semgrep_path.exists():
         logger.warning(
-            f"""You have an old DeepSemgrep binary installed in {deep_semgrep_path}
+            f"""You have an old DeepSemgrep binary installed in {deep_semgrep_path}.
 DeepSemgrep has been renamed to Semgrep PRO, which we will proceed to install.
 Please delete {deep_semgrep_path} manually to make this warning disappear!
 """
@@ -103,10 +103,10 @@ Please delete {deep_semgrep_path} manually to make this warning disappear!
     logger.info(f"Installed Semgrep PRO to {semgrep_pro_path}")
 
     version = sub_check_output(
-        [str(semgrep_pro_path), "--version"],
+        [str(semgrep_pro_path), "-pro_version"],
         timeout=10,
         encoding="utf-8",
-        stderr=subprocess.DEVNULL,
+        stderr=subprocess.STDOUT,
     ).rstrip()
     logger.info(f"Semgrep PRO Version Info: ({version})")
 
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -40,6 +40,10 @@ def run_install_semgrep_pro() -> None:
     state.terminal.configure(verbose=False, debug=False, quiet=False, force_color=False)
 
     (semgrep_pro_path, deep_semgrep_path) = determine_semgrep_pro_path()
+
+    # TODO This is a temporary solution to help offline users
+    logger.info(f"Semgrep PRO engine will be installed in {semgrep_pro_path}")
+
     if deep_semgrep_path.exists():
         logger.warning(
             f"""You have an old DeepSemgrep binary installed in {deep_semgrep_path}.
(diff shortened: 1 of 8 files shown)


cli/src/semgrep/commands/install.py
```diff
@@ -88,30 +95,30 @@ def run_install_deep_semgrep() -> None:
...
-        os.stat(deep_semgrep_path).st_mode | stat.S_IEXEC | stat.S_IXGRP | stat.S_IXOTH,
+        semgrep_pro_path,
+        os.stat(semgrep_pro_path).st_mode | stat.S_IEXEC | stat.S_IXGRP | stat.S_IXOTH,
     )
-    logger.info(f"Installed deepsemgrep to {deep_semgrep_path}")
+    logger.info(f"Installed Semgrep PRO to {semgrep_pro_path}")
 
     version = sub_check_output(
-        [str(deep_semgrep_path), "--version"],
+        [str(semgrep_pro_path), "--version"],
```
emjin: Pretty strange to have both `--version` checked here are `-pro_version` later 🤔

cli/src/semgrep/commands/install.py
```diff
@@ -30,18 +31,24 @@ def determine_deep_semgrep_path() -> Path:
...
+            f"""You have an old DeepSemgrep binary installed in {deep_semgrep_path}
+DeepSemgrep has been renamed to Semgrep PRO, which we will proceed to install.
+Please delete {deep_semgrep_path} manually to make this warning disappear!
+"""
         )
+    if semgrep_pro_path.exists():
+        logger.info(f"Overwriting Semgrep PRO already installed in {semgrep_pro_path}")
 
     if state.app_session.token is None:
         logger.info("run `semgrep login` before using `semgrep install`")
```
emjin, line 58: I'm sorry, I merged my PR without thinking about this. If it's convenient to in this PR can you update https://github.com/returntocorp/semgrep/pull/6889/files? Otherwise I'll just make a note to fix that before I release

cli/src/semgrep/core_runner.py
```diff
@@ -835,45 +836,82 @@ def _run_rules_direct_to_semgrep_core_helper(
...
+                )  # DEPRECATED: To be removed by Feb 2023 launch
+                pro_path = SemgrepCore.pro_path()
+
+                if pro_path is None:
+                    if deep_path is None:
+                        raise SemgrepError(
+                            "Could not run deep analysis: Semgrep PRO not installed. Run `semgrep install-semgrep-pro`"
+                        )
+                    else:
+                        logger.warning(
```
emjin, line 850: Thank you for all the warnings 😍

cli/src/semgrep/core_runner.py
```diff
@@ -835,45 +836,82 @@ def _run_rules_direct_to_semgrep_core_helper(
...
+                    ).rstrip()
+                    logger.info(f"Semgrep PRO Version Info: ({version})")
+
+                    cmd_bin = pro_path
+                    cmd_args += [
+                        "-json_time"
+                    ]  # TODO: When do we stop adding this by default?
+                    cmd_args += ["-deep_inter_file"]
+                    cmd_args += [root]
+                elif (
```
emjin, line 871: Couldn't we just include this section in the else case on line 849 and then turn the `if` statement on 844 to an `else`? If not, take that to mean that these conditionals are hard to understand :D

cli/src/semgrep/core_runner.py
```diff
@@ -835,45 +836,82 @@ def _run_rules_direct_to_semgrep_core_helper(
...
+                        [pro_path, "-pro_version"],
+                        timeout=10,
+                        encoding="utf-8",
+                        stderr=subprocess.STDOUT,
+                    ).rstrip()
+                    logger.info(f"Semgrep PRO Version Info: ({version})")
+
+                    cmd_bin = pro_path
+                    cmd_args += [
+                        "-json_time"
```
emjin: Isn't this already there?

emjin (approved):
Approving to make things smoother, left some comments. I'll also test it myself before I release

cli/src/semgrep/commands/install.py
IagoAbal: Good catch! Couldn't test this at the time because there was no `semgrep-core-proprietary` binaries in the S3 bucket.

cli/src/semgrep/commands/install.py
IagoAbal: No worries, I rebased.

cli/src/semgrep/core_runner.py
IagoAbal: We could, I think. I'm not sure it would be cleaner (if I'm understanding you correctly). But I didn't put much thought into cleaning this up because the idea is to remove all that `deep-semgrep` related code in the next (or next next) release..

cli/src/semgrep/core_runner.py
IagoAbal: Ooops, indeed!

PR #6891, opened 2023-01-20
fix: Avoid fatal merge-base calls on shallow checkouts

nmote:
#6756 allows us to use the GitHub API to find the merge base, rather than relying on the sometimes slow process of fetching additional commits until we can compute it locally.

However, downstream code assumes that we have enough history to compute the merge base. When it tries to do so, Git fatals and Semgrep aborts. This was reported to us when a customer tried to make use of this optimization.

I missed this while testing the original PR because my test PR on my test repo added the file with the finding. When all the files with findings are added in the PR being scanned, Semgrep doesn't bother doing a baseline scan, which I overlooked.

Test plan:

Same as #6756 except the test PR modifies an existing file with findings rather than creating a new file. When using the current `returntocorp/semgrep` image, Semgrep fails on the merge base call in the same way as described by the customer. When using the locally-built `nmote/semgrep` image with the change in this PR, Semgrep runs correctly.

PR checklist:

### DIFF
--- a/cli/src/semgrep/git.py
+++ b/cli/src/semgrep/git.py
@@ -196,6 +196,16 @@ class BaselineHandler:
 
         return GitStatus(added, modified, removed, unmerged, renamed)
 
+    def _get_git_merge_base(self) -> str:
+        # If we already know that the base commit is the merge base, just return
+        # the base commit. This allows us to operate on shallow checkouts where
+        # we might not have the information locally to compute the merge base.
+        # In this case, calling `git merge-base` may fail.
+        if self._is_mergebase:
+            return self._base_commit
+        else:
+            return git_check_output(["git", "merge-base", self._base_commit, "HEAD"])
+
     def _get_dirty_paths_by_status(self) -> Dict[str, List[Path]]:
         """
         Returns all paths that have a git status, grouped by change type.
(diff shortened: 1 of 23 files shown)


nmote:
I don't think I want to address the Semgrep finding. All I did was modify an existing string that is logged.

cli/src/semgrep/git.py
```diff
@@ -338,10 +341,10 @@ def baseline_context(self) -> Iterator[None]:
 
     def print_git_log(self) -> None:
         base_commit_sha = git_check_output(["git", "rev-parse", self._base_commit])
-        merge_base_sha = git_check_output(
-            ["git", "merge-base", self._base_commit, "HEAD"]
+        merge_base_sha = self._get_git_merge_base()
+        logger.info(
+            "  Will report findings introduced by these commits (may be incomplete for shallow checkouts):"
```
aryx, line 350: is it possible to detect shallow checkout so that we print the additional warning only if we know it's a shallow checkout?

aryx (approved):

cli/src/semgrep/git.py
```diff
@@ -196,6 +196,12 @@ def _get_git_status(self) -> GitStatus:
 
         return GitStatus(added, modified, removed, unmerged, renamed)
 
+    def _get_git_merge_base(self) -> str:
+        if self._is_mergebase:
```
emjin, line 204: Can you explain why we need this check? Wasn't obvious to me so it would be nice to preserve it

emjin (approved):

cli/src/semgrep/git.py
nmote: I'm not sure. At this point I'm reluctant to complicate this further. Do you think it's critical?

nmote: I'm going to merge, but am happy to follow up if you think it's important.

PR #6893, opened 2023-01-20
cli: intuitive order ci findings

dataholly:
This is a small change to provide more intuitive default findings sort order in the app when using the ci command with semgrep.  This is a short-term improvement as other intentional (rather than arbitrary) ordering schemes are in the works.  PR also includes addition of tomli to pre commit yaml to enable `pre-commit run --all` to pass.
 
Current behavior: arbitrary ordering of rules (based on dictionary key ordering) and decreasing alphabetical / line numbering tracking where issues are found.
New behavior: rules with higher severity appear first, file names appear alphabetically and line numbers increment.

Test plan: this change is difficult to fully test locally.  I uploaded findings on a test repo (juicebox) to semgrep.dev and confirm ordering is as expected.

PR checklist:

### DIFF
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -278,7 +278,10 @@ class ScanHandler:
             for matches_of_rule in matches_by_rule.values()
             for match in matches_of_rule
         ]
-        all_matches.reverse()  # rising filename and line numbers in app
+        # we want date stamps assigned by the app to be assigned such that the
+        # current sort by relevant_since results in findings within a given scan
+        # appear in an intuitive order.  this requires reversed ordering here.
+        all_matches.reverse()
         sort_order = {  # used only to order rules by severity
             "EXPERIMENT": 0,
             "INVENTORY": 1,
@@ -286,6 +289,8 @@ class ScanHandler:
             "WARNING": 3,
             "ERROR": 4,
         }
+        # NB: sorted guarantees stable sort, so within a given severity level
+        # issues remain sorted as before
         all_matches = sorted(
             all_matches, key=lambda match: sort_order[match.severity.value]
         )
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -278,6 +278,17 @@ class ScanHandler:
             for matches_of_rule in matches_by_rule.values()
             for match in matches_of_rule
         ]
+        all_matches.reverse()  # rising filename and line numbers in app
+        sort_order = {  # used only to order rules by severity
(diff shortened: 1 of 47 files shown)


cli/src/semgrep/app/scans.py
```diff
@@ -278,6 +278,17 @@ def report_findings(
             for matches_of_rule in matches_by_rule.values()
             for match in matches_of_rule
         ]
+        all_matches.reverse()  # rising filename and line numbers in app
```
aryx: what's the point of reverse if anyway a few lines below you gonna sort the array?

aryx:
Not sure what's going on with the CI failures. I've rerun the failed test ... let's hope they pass now.

cli/src/semgrep/app/scans.py
dataholly: I know it looks unintuitive.  Would a different comment be more useful?  Python 'sorted' maintains the existing sort order of the list, except for the new sort key.  In this code, sorting by severity leaves all "high" severity issues in the original sort order.  The original sort order results in file names appearing in reverse alphabetical order and line numbers decreasing across issues.  I find alphabetical ordering and increasing line numbers more intuitive.

dataholly:
Looks like I hadn't fully understood the end to end testing framework.  I'll dig in on why those are failing, leaving this PR as blocked is appropriate.

cli/src/semgrep/app/scans.py
aryx: Maybe better to be explicit then and fully specify the sorting order rather than relying on obscure internal
of Python functions?

dataholly: "sorted" is a common python function and sort stability is an important and well-documented "guarantee" (see https://docs.python.org/3/howto/sorting.html).

Nevertheless, agree that in general explicit is better than implicit.  

The order in which we sort issues within a given rule is currently controlled by [get_order_key in rule_match.py](https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/rule_match.py#L228) -- we sort by path, start, end, rule_id, and message.  I believe that sort actually occurs at [core_output.py](https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/core_output.py#L213).  That is the more natural place to reverse the order of the within-rule results, but it will break far more tests and it will affect other output types, whereas I was only seeking to change the ordering for CI upload.

brendongo: ~~you can have sorted's key be a tuple for secondary sorting~~

aryx (approved):

PR #6903, opened 2023-01-20
chore(pro): thread through errors for oom and timeout during preprocessing

emjin:
This will help us keep track of which errors are raised by pro engine
users that prevent pro analysis from happening. Also, sets a default
pro timeout for CI users

See https://github.com/returntocorp/semgrep-proprietary/pull/465 for more information

PR checklist:


cli/src/semgrep/commands/scan.py
```diff
@@ -348,6 +348,15 @@ def convert(
             skipped. If set to 0 will not have limit. Defaults to 3.
         """,
     ),
+    optgroup.option(
+        "--pro-timeout",
+        type=int,
+        help=f"""
+            Maximum time to spend on preprocessing for pro features. If set to 0
+            will not have time limit. Defaults to 0 s unless the DeepSemgrep toggle
+            is on, in which case it defaults to 3 hours.
```
nmote: That's a bit confusing

nmote (approved):
Looks good! This will be a breaking change though, right? New versions of the CLI will no longer function with even slightly older versions of semgrep-proprietary.

nmote:
Please don't merge until you're ready to merge the associated changes to semgrep-proprietary

cli/src/semgrep/commands/scan.py
```diff
@@ -348,6 +348,15 @@ def convert(
             skipped. If set to 0 will not have limit. Defaults to 3.
         """,
     ),
+    optgroup.option(
+        "--pro-timeout",
+        type=int,
+        help=f"""
+            Maximum time to spend on preprocessing for pro features. If set to 0
+            will not have time limit. Defaults to 0 s unless the DeepSemgrep toggle
```
IagoAbal: Maybe App should rename the toggle to `Semgrep Pro` (even if it's just in the UI, that should be easy) so that we don't need to refer to "DeepSemgrep" here either. I also find  a bit confusing to refer to the App toggle here. This is the `scan` command command, not the `ci` one.

cli/src/semgrep/commands/scan.py
```diff
@@ -348,6 +348,15 @@ def convert(
             skipped. If set to 0 will not have limit. Defaults to 3.
         """,
     ),
+    optgroup.option(
+        "--pro-timeout",
+        type=int,
+        help=f"""
+            Maximum time to spend on preprocessing for pro features. If set to 0
```
IagoAbal: As I said in the other PR I think this should be "inter-file analysis". Apex is a Pro feature and doesn't need this timeout. And `--interproc` won't need this timeout either I think.

IagoAbal (approved):

cli/src/semgrep/commands/scan.py
emjin: Yeah, I agree that's a bit confusing but I don't know another place to put it

cli/src/semgrep/commands/scan.py
emjin: Made some clarifications

PR #6900, opened 2023-01-20
ResolvedName of dotted_ident -> GlobalName of string list

aryx:
This is the PR for a possible resolution of the
ResolvedName and metavariable mess.
The idea is to not store tokens anymore for resolved names
(now rename to GlobalName), because anyway the tokens
in it can come from different files, or contain tokens whose
range might contain keywords (e.g., in import {bar} from foo).
We want to move away from tokens anyway as a general principle,
and prefer proper pretty printing.

test plan:
make test, and hope for no regression in semgrep.
At some point also test in deep-semgrep and test
with the new testsuite brandon is preparing covering
more edge cases of interactions between resolved names
and metavariable-regex: and metavariable-pattern:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure on any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #6909, opened 2023-01-22
chore: add tests for resolved names

brandonspark:
## What:
This is the test suite that ideally should describe the engine behavior for matching resolved names (as well as imported entities) to metavariables.

The main tests that are contentious are tests 4, 5, 6, and 7, due to things like `focus-metavariable`, `metavariable-regex`, and `metavariable-comparison`.

I did check, and noted that metavariables seem to not unify in situations they should not, as for unresolved metavariables the content of the metavariable "as stated" is what is used for comparison, not the actual content at the location. So the tokens don't change that.

I added it here in `semgrep` because I think it should be able to pass for both OSS and pro.

PR checklist:

### DIFF
--- a/tests/rules/resolved_metavariables.js
+++ b/tests/rules/resolved_metavariables.js
@@ -1,0 +1,20 @@
+// MATCH:
+import {bar as baz} from "foo";
+
+// MATCH:
+var res1 = baz()
+
+// MATCH:
+var res2 = baz()
+
+// MATCH:
+var res3 = baz()
+
+// MATCH:
+var res4 = baz()
+
+var res5 = baz()
+
+var res6 = baz()
+
+var res7 = baz()
--- a/tests/rules/resolved_metavariables.js
+++ b/tests/rules/resolved_metavariables.js
@@ -1,20 +1,26 @@
-// MATCH:
+// ruleid: resolved-metavariables 
 import {bar as baz} from "foo";
 
-// MATCH:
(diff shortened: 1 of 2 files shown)


emjin:
Tests failing

aryx:
yep 
```
this one finding is missing: ../../../tests/rules/resolved_metavariables.js:14
```

aryx:
res4 is failing because the metavariable will contain 'bar as baz} from "foo"' I think, because
we use tokens ranges from 'bar' to 'foo'

aryx:
@brandonspark I don't see how test 6, 7, and 8 should match? bar and foo will never be equal.

aryx:
> "as for unresolved metavariables the content of the metavariable "as stated" is what is used for comparison, not the actual content at the location. So the tokens don't change that."

The actual content at the tokens location is used; this is why res4 is failing. This is why using the tokens
is wrong, because you have bad edge cases like this.

aryx:
Also I don't think we should print a match for the first import line ... it's weird.

aryx:
Oh I understand 6, 7, 8 now, it's to double check that we don't use a single token for the resolved name in which case we get the same content for every part of the name.

aryx:
I've included brandon's testsuite in https://github.com/returntocorp/semgrep/pull/6900
and the tests are now passing there!

brandonspark:
Tests have been added.

PR #6914, opened 2023-01-24
feat!: `unique_name`s

brandonspark:
## What:
This PR adds in `unique_name`, which is a transparent type equivalent to `Name.t`, for use in OSS Semgrep.

## Why:
This will help to uniquely identify names in the PRO engine. This information needs to be passed back to OSS, and since we don't have the `Name.t` type, we mint one that is just equivalent.

## How:
Boilerplate.

PR checklist:

### DIFF
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -303,7 +303,16 @@ and resolved_name_kind =
    * resolved_name, but we also want to match the pattern `foo.bar` so we will
    * store ['foo', 'bar'] as an alternate name.
    * *)
-  | ResolvedName of dotted_ident * dotted_ident list
+  | ResolvedName of unique_name * unique_name list
+
+(* This is not relevant for matching purposes.
+   `unique_name` is a record of all the information necessary to
+   disambiguate two `ResolvedName`s from each other.
+   In particular, two `ResolvedName`s may indicate two overloaded functions,
+   which are named the same. Then, we need to use the function signature information
+   to disambiguate them from each other.
+*)
+and unique_name = { dotted : dotted_ident; tok : tok option }
 [@@deriving show { with_path = false }, eq, hash]
 
 (* Start of big mutually recursive types because of the use of 'any'
@@ -365,7 +374,7 @@ and id_info = {
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
    *)
-  id_type : type_ option ref;
+  id_type : type_ option ref; [@equal fun _a _b -> true]
   (* type checker (typing) *)
   (* sgrep: this is for sgrep constant propagation hack.
    * todo? associate only with Id?
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -374,7 +374,7 @@ and id_info = {
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
    *)
-  id_type : type_ option ref; [@equal fun _a _b -> true]
+  id_type : type_ option ref;
   (* type checker (typing) *)
   (* sgrep: this is for sgrep constant propagation hack.
    * todo? associate only with Id?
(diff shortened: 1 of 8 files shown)


libs/ast_generic/AST_generic.ml
```diff
@@ -365,7 +365,7 @@ and id_info = {
    * a typed entity, which can be interpreted as a TypedMetavar in semgrep.
    * alt: have an explicity type_ field in entity.
    *)
-  id_type : type_ option ref;
+  id_type : type_ option ref; [@equal fun _a _b -> true]
```
nmote: What is this for?

libs/ast_generic/AST_generic_helpers.ml
```diff
@@ -119,6 +119,25 @@ let name_of_ids xs =
           name_info = empty_id_info ();
         }
 
+let name_of_name_t name =
```
nmote: Can you clarify when one would use this vs using `name_t_to_ast_generic_name` below?

libs/ast_generic/AST_generic_helpers.ml
```diff
@@ -170,6 +189,14 @@ let dotted_ident_of_name (n : name) : dotted_ident =
       in
       before @ [ ident ]
 
+let name_t_to_ast_generic_name name =
+  (* this is not right, TODO *)
```
nmote: Elaborate?

libs/ast_generic/Name.ml
```diff
@@ -0,0 +1,162 @@
...
+ * to access when at the head of the list.
+ * It's also faster to expand the name by simply using
+ * '::' instead of 'name @ [id]' (see append_id).
+ * We could even have type t = Qualified of string * string list.
+ * to enforce the name can't be empty.
+ *)
+
+open Ppx_hash_lib.Std.Hash.Builtin
+
+type unique_marker =
```
nmote: Maybe `name_kind`?

src/analyzing/Dataflow_svalue.ml
```diff
@@ -53,10 +53,7 @@ let warning _tok s =
 (*****************************************************************************)
 
 let str_of_name name = spf "%s:%s" (fst name.ident) (G.SId.show name.sid)
-
-let str_of_resolved_name name =
-  let name = Common.map fst name in
-  String.concat "." name
+let str_of_resolved_name name = Name.to_entname name
```
nmote: Sweet

nmote (commented):
I started reviewing, but then I realized there is a test failure. Could you address that? Once you do I'll finish my review.

libs/ast_generic/AST_generic_helpers.ml
brandonspark: I think this was erroneous

libs/ast_generic/AST_generic_helpers.ml
brandonspark: Changed the names to be more specific

libs/ast_generic/AST_generic.ml
brandonspark: I'm certain this is needed in the future, but I guess I'll delete for now

aryx:
I would vote no.

aryx:
Name.ml make sense only for DeepSemgrep so I think it's better to have it in the semgrep-pro repository.
This also avoids to have to modify 2 repositories when you need to modify naming related things.

aryx:
Plus I am in the middle of changing ResolvedName, and plus Name.ml ultimately will contain
information about overloading method which we should not care at all in regular Semgrep.
So this is a big no for me.

brandonspark:
> Name.ml make sense only for DeepSemgrep so I think it's better to have it in the semgrep-pro repository. This also avoids to have to modify 2 repositories when you need to modify naming related things.

We are going to have to pass back a transparent isomorphism of `Name.t` regardless. If it is located here, at least we can avoid having to break the abstraction by exposign a type which is completely equivalent to `Name.t`, as well as avoid general overhead of having around another type to manage.

I disagree that naming is completely localized to PRO, because the naming information from PRO is consumed at this stage. We use the `ResolvedName` by dispatching on it within the matching engine, so it's not unrelated.

brandonspark:
This doesn't need to be a new PR actaully

PR #6915, opened 2023-01-24
fix: cleanup semgrep brew caches before re-running

spencerdrak:
The Nightly homebrew checks were failing due to a not-hermetic build (moved submodules caused us issues again). This cleans up as a stopgap, MTF on a more permanent fix.

There's likely more discussion here on whether this affects other workflows. Will discuss this before marking this PR as ready for review.

Test Plan:

1. Run a nightly test using this branch ([success](https://github.com/returntocorp/semgrep/actions/runs/3992135935))

PR checklist:

### DIFF
--- a/.github/workflows/homebrew-core-head.yml
+++ b/.github/workflows/homebrew-core-head.yml
@@ -25,6 +25,8 @@ jobs:
         # This is sub-optimal - our workflows shouldn't have to conform to their environment.
         # However, on the runner side, we can't hook into the workflow run to clean up after.
         run: brew uninstall semgrep || true
+      - name: Cleanup semgrep
+        run: brew cleanup --prune=all semgrep
       - name: Brew update
         run: brew update
       - name: Brew Install


spencerdrak:
As noted above, the test succeeded with this change: https://github.com/returntocorp/semgrep/actions/runs/3992135935

cgdolan (approved):

.github/workflows/homebrew-core-head.yml
```diff
@@ -25,6 +25,8 @@ jobs:
         # This is sub-optimal - our workflows shouldn't have to conform to their environment.
         # However, on the runner side, we can't hook into the workflow run to clean up after.
         run: brew uninstall semgrep || true
```
aryx, line 27: Why do we need to do that? Why the self-hosted does not clean things by itself for each new job,
like what GHA does.

aryx (approved):

.github/workflows/homebrew-core-head.yml
cgdolan: https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#differences-between-github-hosted-and-self-hosted-runners

spencerdrak: We're using bare-metal instances, whereas the GHA runners are a single-use VM that starts fresh each time. We don't have tooling/processes for spinning up new instances for each job. It's a know limitation and on our radar.

aryx: great link chris, would be nice to add it next to the uninstall step, so people understand why
we need those extra steps.

emjin (approved):

cgdolan (approved):

PR #6916, opened 2023-01-24
chore(ci): run semgrep jsonnet tests in CI only

brendongo:
PR checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -182,6 +182,8 @@ repos:
     hooks:
       - id: semgrep-jsonnet
         name: semgrep jsonnet
+        # Only run this job in CI since it is slow
+        stages: [manual]
         language: docker_image
         # See .pre-commit-hooks.yaml for why we need to set those
         # SEMGREP_XXX variables.  Ideally we would not need this
@@ -218,7 +220,7 @@ repos:
           ]
   # Dogfooding .pre-commit-hooks.yml and setup.py
   - repo: https://github.com/returntocorp/semgrep
-    rev: v0.100.0
+    rev: v1.5.1
     hooks:
       - id: semgrep
         name: Semgrep Python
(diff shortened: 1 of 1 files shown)


aryx:
No! This was added to dogfood the use of jsonnet ... It takes 4s on my machine, which IMHO is not that long. Actually
some of the pre-commit steps take longer I think.

aryx:
What is the test plan?
How do I run manually the semgrep-jsonnet?

brendongo:
@brandonspark how slow is the step on your machine? is it a docker m1 mac issue?

@aryx `pre-commit run --hook-stage manual` to run locally.  This PR doesn't remove it, just makes it run only in CI. https://github.com/returntocorp/semgrep/blob/develop/.github/workflows/lint.yml#L55 is there it runs the hooks marked as manual

aryx:
I'll let @brandonspark and @nmote accept if they are really bothered by the current dogfooding of semgrep-jsonnet.
Personally I think it would have been better if brandon tried to investigage/fix the issue rather than bypassing it everytime
at commit, but if people think it's really too slow (it takes 4secs on my machine and I don't commit a hundred time per day, so it's fine with me), then I can live with running it manually. I personally prefer to wait 4sec in pre-commit than waiting 4min to get the error displayed in CI in github.

nmote:
Adding 4 seconds to commit is pretty frustrating to me. I constantly commit and amend and it's especially annoying because I usually have to run it twice because the formatting step formats and fails instead of formatting and then passing. The longer pre-commit jobs only run on Python changes, which I rarely make (but I am frustrated with the commit delay there too when I do).

@aryx, when we previously discussed this, you said that running this in Docker was temporary and once jsonnet shipped to prod we could go back to running Semgrep natively, which is much faster. What is the current status? I'd prefer to have it run quickly on commit, but if I have to choose between fast commits and Semgrep on commit, I'll choose fast commits.

Plus, I VERY rarely end up with any Semgrep findings from this job. The tradeoff of putting it in CI sounds good to me.

aryx (approved):

aryx:
Ok, let's merge it then.

PR #6917, opened 2023-01-24
feat!: type coercions for Terraform

### LINKED ISSUE #6898: Add Terraform equivalence on automatic type conversions

Terraform supports automatic type conversions between strings, ints, and bools: https://developer.hashicorp.com/terraform/language/expressions/types#type-conversion. E.g. `"42"` and `42`, `"true"` and `true`.

It would be helpful if Semgrep automatically supported these type equivalences so I don't have to manually specify separate cases in rules. Here's a link that exhibits the behavior: https://semgrep.dev/playground/s/D0Kj.

Manually specifying the equivalences

**Use case**
What will this feature enable for you?

**Additional context**

brandonspark:
## What:
This PR adds in coercions from literals of bools, ints, and floats to strings, in the matching engine.

## Why:
This is how Terraform works, and it was requested by a user: #6898 

## How:
I added a wrapper for `m_literal`, which first checks if the language is `hcl`. If so, then it does the coercions on the necessary cases. Otherwise, it will defer to the original `m_literal`, now named `m_literal_inner`.

## Test plan:
`make test`

Closes #6898 

PR checklist:

### DIFF
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -1165,7 +1165,46 @@ and m_for_or_if_comp a b =
   | G.CompIf _, _ ->
       fail ()
 
+(* This just does some special coercions for Terraform.
+   If not a Terraform language, skip straight to `m_literal_inner`.
+   https://developer.hashicorp.com/terraform/language/expressions/type-constraints
+*)
 and m_literal a b =
+  Trace_matching.(if on then print_literal_pair a b);
+  fun tin ->
+    if Lang.equal tin.lang Lang.Hcl then
+      (* Is it ugly? Yes.
+         Is it clear what it does? Also yes, I think. Moreso than factorizing it.
+         So let's just let it be.
+      *)
+      match (a, b) with
+      | G.Int (a1, a2), B.String (b1, b2) ->
+          let i = Option.map Int.to_string a1 in
+          m_wrap_m_string_opt (i, a2) (Some b1, b2) tin
+      | G.String (a1, a2), B.Int (b1, b2) ->
+          let i = Option.map Int.to_string b1 in
+          m_wrap_m_string_opt (Some a1, a2) (i, b2) tin
+      | G.Bool (a1, a2), B.String (b1, b2) ->
+          let b = Bool.to_string a1 in
+          m_wrap m_string (b, a2) (b1, b2) tin
+      | G.String (a1, a2), B.Bool (b1, b2) ->
+          let b = Bool.to_string b1 in
+          m_wrap m_string (a1, a2) (b, b2) tin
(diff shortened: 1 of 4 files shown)


src/matching/Generic_vs_generic.ml
```diff
@@ -1165,7 +1165,46 @@ and m_for_or_if_comp a b =
   | G.CompIf _, _ ->
       fail ()
 
+(* This just does some special coercions for Terraform.
+   If not a Terraform language, skip straight to `m_literal_inner`.
+   https://developer.hashicorp.com/terraform/language/expressions/type-constraints
+*)
 and m_literal a b =
+  Trace_matching.(if on then print_literal_pair a b);
+  fun tin ->
```
aryx: You can use with_lang, so you don't have to pass around tin

src/matching/Generic_vs_generic.ml
```diff
@@ -1165,7 +1165,46 @@ and m_for_or_if_comp a b =
...
+    if Lang.equal tin.lang Lang.Hcl then
+      (* Is it ugly? Yes.
```
aryx: Not very useful comment.

aryx: Or did you mean that you could factorize with
| G.Int (x1, x2), B.String (y1, y2)
| G.String (y1, y2), G.Int (x1, x2) ->
  ???
Why not, since the body is the same? Or is it because
you want to distinguish when the String is from a pattern or from a target?
(possibly important since one can contain metavariable I guess).

src/matching/Generic_vs_generic.ml
```diff
@@ -1218,6 +1257,15 @@ and m_wrap_m_int_opt (a1, a2) (b1, b2) =
       let b1 = Parse_info.str_of_info b2 in
       m_wrap m_string (a1, a2) (b1, b2)
 
+and m_wrap_m_string_opt (a1, a2) (b1, b2) =
+  match (a1, b1) with
+  (* iso: semantic equivalence of value! 0x8 can match 8 *)
```
aryx: You've copied the comment from below but this actually does not apply here ... "0x8" will not match "8" ...

aryx (approved):

src/matching/Generic_vs_generic.ml
brandonspark: fair, i will leave a comment just saying we preserve order

src/matching/Generic_vs_generic.ml
brandonspark: thanks!

src/matching/Generic_vs_generic.ml
brandonspark: oops, that was a mistake

PR #6918, opened 2023-01-24
chore: expose js angular decorators helper

brandonspark:
## What:
To fix a regression in DeepSemgrep, I have to add logic in `Naming_SAST` duplicating something I already did in `Naming_AST`.

To avoid having to duplicate hard-coded values, I want to expose an `is_js_angular_decorator` function which will tell if a decorator is possibly an Angular JS decorator, in which case different behavior should occur.

## Why:
This will allow me to make changes in DeepSemgrep without having to hard-code values

## How:
Exposed aforementioend function

## Test plan:
in Semgrep PRO

No changelog because not user-facing

PR checklist:

### DIFF
--- a/src/naming/Naming_AST.mli
+++ b/src/naming/Naming_AST.mli
@@ -3,4 +3,12 @@
  * specific to a language.
  *)
 val resolve : Lang.t -> AST_generic.program -> unit
+
+(* We expose this language-specific (and framework-specific!)
+   function here because it is needed in Semgrep Pro.
+   This lets us avoid having to duplicate the logic of this function.
+
+   There's not really a better place to put this, so we'll just
+   let it exist here.
+*)
 val is_js_angular_decorator : string -> bool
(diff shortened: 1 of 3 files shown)


src/naming/Naming_AST.mli
```diff
@@ -3,3 +3,4 @@
  * specific to a language.
  *)
 val resolve : Lang.t -> AST_generic.program -> unit
+val is_js_angular_decorator : string -> bool
```
aryx, line 14: This is ugly. Ideally those functions would be in another file, or there will be a comment in the .mli explaining
why the hell we need to do something JS and even Angular specific in what is supposed to be language agnostic.

aryx (approved):

src/naming/Naming_AST.mli
brandonspark: I'd prefer not to open up a whole other file for precisely one helper function, personally. It's naming-related, so I figured this would be a better place. Open to suggestions.

PR #6919, opened 2023-01-24
OCaml support for semgrepignore

mjambon:
This code adds support for most of the gitignore mechanisms, based on the official specification. In addition to in-project `.gitignore` files, we support `.semgrepignore` at the same precedence level, i.e. if a `.semgrepignore` file is found, it behaves as if it were appended to the `.gitignore` present in the same folder. This allows negated patterns (e.g. `!.*.c`) to de-exclude files excluded by gitignore. The `--exclude` feature of the semgrep command line uses the gitignore mechanism, just like `git ls-files --exclude`.

The `--include` option of semgrep is also supported. It is however significantly different from `--exclude` as its patterns determine files to keep rather than files to ignore. If any include pattern matches a target, this target is selected. It is subject to gitignore/semgrepignore filtering after that.

Other things that I added in this PR:
* the new function `Testutil.test` can be used to define and register local tests. It seems to be convenient for modules with only a few simple tests. A major gotcha is that if nothing uses the module containing the tests, the tests won't be registered and won't run.
* the `Testutil_files` module makes it easy to create temporary file hierarchies. This is used for the gitignore tests.
* the `Ftree` module hasn't been used yet. It will be used (and probably heavily modified or deleted) once we implement `Semgrepignore.list_files`.
* the `Git_project` module provides a function to resolve a path into the pair (project root, git path relative to the project root). This will be used to resolve target files from the command line but it's not in use yet.

PR checklist:


aryx:
looks like a good start!

src/osemgrep/targeting/Ftree.mli
```diff
@@ -0,0 +1,81 @@
+(*
+   Lazy file tree used for path filtering.
```
aryx: Maybe explain the name of the file, maybe something like
"Ftree.ml is a tree of FPath.ml"

src/osemgrep/targeting/Gitignore_syntax.mli
```diff
@@ -0,0 +1,30 @@
...
+  (* Line number, starting from 1. *)
+  line_number : int;
+  line_contents : string;
+}
+
+(* Path selector. *)
+type pattern = {
+  loc : loc;
+  (* The matcher tells whether a given path matches the pattern.
+     For example, the pattern /foo matches the path / *)
```
aryx: the pattern /foo matches the path /foo you mean? typo?

src/osemgrep/targeting/Semgrepignore.ml
```diff
@@ -0,0 +1,44 @@
...
+   Migration plan:
+   - print a deprecation notice if an ':include' directive is found in the
+     root .semgrepignore.
+   - stay silent otherwise: this is problematic only in the case of a
+     .semgrepignore that doesn't include the .gitignore explicitly and
+     contains fewer exclusions that the .gitignore. The new behavior will
+     exclude more files than before.
+
+   Questions:
+   - Do some people really run semgrep on purpose on files that are excluded
```
aryx, line 33: I've asked bence and brandon on slack.

src/osemgrep/targeting/dune
```diff
@@ -0,0 +1,12 @@
+; Various operations having to do with selection of files in Semgrep.
+; What they have in common is:
+; - dependency on utilities to manipulate file paths (e.g. globbing);
+; - no dependency on full-fledged language parsing or the resulting ASTs.
+;
+(library
+  (name osemgrep_targeting)
+  (libraries
+    commons
+    fpath
```
aryx, line 10: probably need to add fpath in semgrep.opam

src/osemgrep/targeting/dune
```diff
@@ -0,0 +1,12 @@
+; Various operations having to do with selection of files in Semgrep.
```
aryx, line 1: ideally we can also put this out of osemgrep in libs/gitignore/ and release as a separate OPAM library
at some point.

.pre-commit-config.yaml
```diff
@@ -190,17 +190,17 @@ repos:
         # is not available via setup.py and only (unofficially)
         # available in docker.
         entry: -e SEMGREP_LOG_FILE=/tmp/out.log -e SEMGREP_VERSION_CACHE_PATH=/tmp/cache returntocorp/semgrep:develop semgrep
-        exclude: "^semgrep-core/src/pfff"
-        #note that the --exclude below are useless because pre-commit pass
-        # a list of files explicitely which prevents --exclude to work
-        # hence the separate exclude: above.
+
+        # Both the .semgrepignore file and the --exclude option
```
aryx, line 195: Maybe add a TODO that says that once we redesign file targeting in (o)semgrep
we should not need that anymore.
In the mean time we have to use the pre-commit exclude mechanism because we can't rely
on the semgrep ignore mechanism (bypassed by passing files explicitely by pre-commit).

src/osemgrep/targeting/Glob_lexer.mll
```diff
@@ -0,0 +1,75 @@
...
+
+let syntax_error msg =
+  raise (Syntax_error msg)
+}
+
+rule tokens acc = parse
+| '/'      { tokens (Slash :: acc) lexbuf }
+| "**"     { (* only special if it occupies a whole path component. This
+                is dealt with later. *)
+             tokens (Slashslash :: acc) lexbuf }
```
aryx: You mean StarStar?

aryx (commented):
An important missing piece is a nice set of tests that show all the features
implemented by .semgrepignore and some unit tests showcasing the main features of each
of the modules.

mjambon:
Semgrepignore, --exclude, and --include now work as I intended :tada:
I still need to review the code myself and do some cleanup.

aryx:
Fantastic!

aryx:
399 failed, 129 passed

libs/commons/Testutil_files.mli
```diff
@@ -0,0 +1,63 @@
+(*
+   Utilities for creating, scanning, and deleting a hierarchy
+   of test files.
```
aryx, line 3: maybe an example of use would be nice, like an example of test using this API (or
maybe just say to look at Unit_semgrepignore.ml for a nice example of use).

Note that at first because of the name (Testutil_files.mli), I thought it was creating fake files, in a virtual
filesystem, so maybe it's worth saying this module will create real files, but usually
in /tmp (thanks to with_tempdir()).

Also because it creates real files, what's the advantage of your API vs using directly the Unix or Sys or Fpath
APIs to create/remove files?
Maybe worth a comment too.

tests/Test.ml
```diff
@@ -34,6 +34,10 @@ let any_gen_of_string str =
 (* All tests *)
 (*****************************************************************************)
 
+(* temporary: force initialization of modules that aren't used yet *)
+let () = ignore Osemgrep_targeting.Semgrepignore.select
```
aryx: Is this still true? Why we don't use those modules?

src/osemgrep/targeting/dune
```diff
@@ -0,0 +1,17 @@
...
+; What they have in common is:
+; - dependency on utilities to manipulate file paths (e.g. globbing);
+; - no dependency on full-fledged language parsing or the resulting ASTs.
+;
+(library
+  (name osemgrep_targeting)
```
aryx, line 7: can you add (wrapped false) to be consistent
with the other libraries.

src/osemgrep/targeting/Glob_matcher.mli
```diff
@@ -0,0 +1,90 @@
+(*
+   AST and matching of a glob pattern against a path.
+   This is purely syntaxic: the file system is not accessed.
+*)
+
+(* The location of a pattern, for logging and troubleshooting. *)
+type loc = {
```
aryx, line 16: yet another loc ... would be good to factorize things at some point.

src/osemgrep/targeting/Glob_matcher.mli
```diff
@@ -0,0 +1,90 @@
...
+  (* Line number, starting from 1. *)
+  line_number : int;
+  line_contents : string;
+}
+
+(* Create a location from a pattern string rather than a location in a file. *)
+val string_loc : ?source_name:string -> string -> loc
+val show_loc : loc -> string
+
+type char_class_range = Class_char of char | Range of char * char
```
aryx, line 29: what is the difference between Char_class {complement=false; ranges = [(Class_char 'a')]} and Char 'a' ?

src/osemgrep/targeting/Glob_matcher.mli
```diff
@@ -0,0 +1,90 @@
...
+  | Char of char
+  | Char_class of char_class
+  | Question
+  | Star
+[@@deriving show]
+
+(* A path component is what represents a simple file name in a directory *)
+type component =
+  | Component of component_fragment list
+  | Ellipsis (* '**' = path ellipsis *)
```
aryx: Why not StarStar? Those are actually not ellipsis. Or it's because it's the term used in the spec?

src/osemgrep/targeting/Glob_matcher.mli
```diff
@@ -0,0 +1,90 @@
+(*
+   AST and matching of a glob pattern against a path.
+   This is purely syntaxic: the file system is not accessed.
```
aryx, line 3: Is there a reference to a specification for the syntax of those globs?
Did you follow any particular document?

src/osemgrep/targeting/Glob_matcher.mli
```diff
@@ -0,0 +1,90 @@
...
+   purposes.
+*)
+val compile : source:loc -> pattern -> t
+
+(*
+   Match a path against a pattern:
+   - The path must be slash-separated (not backslash-separated like on
+     Windows).
+   - If the pattern starts with a slash, the path must start with a slash
+     as well. In both cases, the matching starts from the beginning.
```
aryx, line 76: maybe talk whether the pattern is "anchored" by default?

libs/commons/Testutil_files.mli
```diff
@@ -0,0 +1,63 @@
...
+*)
+
+(*
+   We don't deal with other file types than those three.
+   File names must be normal names, i.e. not ".", "..", or "". They may not
+   contain slashes or backslashes either.
+*)
+type t =
+  | Dir of string * t list
+  | File of string * string
```
aryx: To what correspond each string in File and Symlink? Which one is what?
Maybe File of string (* name *) * string (* content *)

and then maybe an example would be nice, like:
Dir "a" [File "a" ""; Dir "b" []; Symlink "alias" "a"].

src/osemgrep/targeting/Glob_matcher.mli
```diff
@@ -0,0 +1,90 @@
...
+   non-matching paths: foo.c/bar
+
+   folder pattern: foo/
+   matching paths: foo/
+   non-matching paths: foo bar/foo/ /foo/ /foo
+*)
+val run : t -> string -> bool
+val source : t -> loc
+val of_path_components : string list -> pattern
+val show : t -> string
```
aryx, line 100: I think you can add the [@deriving show] after the 'type t' instead.

libs/commons/Testutil_files.mli
```diff
@@ -0,0 +1,63 @@
...
+   'include_dirs=true' will return the path to the folders as well.
+*)
+val flatten : ?root:Fpath.t -> ?include_dirs:bool -> t list -> Fpath.t list
+
+(*
+   Read files from a root. Don't follow symlinks.
+   Fail with an exception if we can't read the files or if they're of an
+   exotic kind.
+*)
+val read : Fpath.t -> t list
```
aryx, line 64: This can return also Dir and Symlink? I'm confused by "read files" which sounds it will only return File ...
Maybe rename it to  hierarchy_from_root ?

libs/commons/Testutil_files.mli
```diff
@@ -0,0 +1,63 @@
...
+
+(*
+   Write a file hierarchy into the given root folder,
+   which must already exist.
+*)
+val write : Fpath.t -> t list -> unit
+
+(* Recursive removal (rm -r) *)
+val remove : Fpath.t -> unit
+val is_dir : Fpath.t -> bool
```
aryx, line 74: Fpath does not provide such functions already?

libs/commons/Testutil_files.mli
```diff
@@ -0,0 +1,63 @@
...
+val is_file : Fpath.t -> bool
+val is_symlink : Fpath.t -> bool
+
+(*
+   Create a folder specified as a string. This path is parsed. If relative,
+   it created relative to the root, which must exist and defaults to
+   the process' current directory. If absolute, that path is used
+   and all the parent directories up to the root are created if necessary.
```
aryx, line 82: maybe add in parenthesis (like mkdir -p)

src/osemgrep/targeting/Unit_glob.ml
```diff
@@ -0,0 +1,65 @@
...
+      ("simple", test "abc" "abc" true);
+      ("absolute path", test "/a/b" "/a/b" true);
+      ("star 1", test "*.c" "hello.c" true);
+      ("star 2", test "*.c" "hello.h" false);
+      ("question 1", test "a.?" "a.c" true);
+      ("question 2", test "a.?" "b.c" false);
+      ("relative path 1", test "ab/*.c" "ab/hello.c" true);
+      ("relative path 2", test "ab/*.c" "hello.c" false);
+      ("relative path 3", test "ab/*.c" "ab/" false);
+      ("relative path 4", test "ab/*.c" "x/ab/hello.c" false);
```
aryx, line 38: so this means the pattern is kinda "anchored"? ab/*.c does not match x/ab/hello.c"

src/osemgrep/targeting/Unit_glob.ml
```diff
@@ -0,0 +1,65 @@
...
+      ("ellipsis 1", test "**" "a" true);
+      ("ellipsis 2", test "**" "a/b/c" true);
+      ("ellipsis 3", test "/**" "/a/b" true);
+      ("ellipsis 4", test "/**" "a/b" false);
+      ("ellipsis 5", test "**/" "a/b/c/" true);
+      ("ellipsis 6", test "**/" "a/b/c" false);
+      ("ellipsis 7", test "a/**/z" "a/b/c/d/z" true);
+      ("ellipsis 8", test "a**" "a" true);
+      ("ellipsis 9", test "a**" "abc" true);
+      ("ellipsis 10", test "a**" "a/b" false);
```
aryx, line 61: this one looks weird to me. a** does not match a/b?
I don't see the logic of ** maybe. To mean it means * but also go through /.

src/osemgrep/targeting/Unit_semgrepignore.ml
```diff
@@ -0,0 +1,219 @@
...
+      in
+      let error = ref false in
+      selection
+      |> List.iter (fun (path, should_be_selected) ->
+             let path = Git_path.of_string path in
+             let status, selection_events = Semgrepignore.select filter path in
+             printf "Selection events for path %s:\n" (Git_path.to_string path);
+             print_string
+               (Gitignore_syntax.show_selection_events selection_events);
+             if should_be_selected then (
```
aryx, line 63: you could also match should_be_selected, status with ...
to be more compact.

src/osemgrep/targeting/Unit_semgrepignore.ml
```diff
@@ -0,0 +1,219 @@
...
+            ("/a/", false);
+            ("/a", true);
+            ("/dir/a", true);
+            ("/dir/a/", false);
+            ("/a/b", false);
+          ] );
+      ( "absolute patterns",
+        test_filter
+          [
+            File (".gitignore", "/a\nb/c");
```
aryx, line 150: what's the difference in pattern between /b/c and b/c?
Looks the same to me in behavior.

src/osemgrep/targeting/Unit_semgrepignore.ml
```diff
@@ -0,0 +1,219 @@
...
+            dir "a" [ file "b" ];
+            dir "b" [ file "a"; file "c"; file "d"; dir "b" [ file "c" ] ];
+          ]
+          [
+            ("/a/b", false);
+            ("/b/a", true);
+            ("/b/c", false);
+            ("/b/d", true);
+            ("/b/b/c", true);
```
aryx, line 159: so even with b/c in .gitignore, /b/b/c is not ignored?

src/osemgrep/targeting/Glob_matcher.mli
aryx: maybe just reference glob(3) and glob(7) like you did in Glob_lexer.mli?

src/osemgrep/targeting/Glob_lexer.mli
```diff
@@ -0,0 +1,16 @@
...
+   Parse a glob pattern according to glob(3) and glob(7), which is the
+   POSIX standard for old-fashioned shell globbing patterns for matching
+   file paths. This is what gitignore uses.
+
+   Examples:
+
+   *.c          # all local files with a '.c' extension
+   /tmp/**      # all valid paths under '/tmp/'
+   Thing.ml?    # matches 'Thing.ml' as well as 'Thing.mli', 'Thing.mll', etc.
+   [a-z0-9]     # matches a single character in these ranges
```
aryx, line 12: ** was also part of the old-fashioned shell globbing syntax?

src/osemgrep/targeting/Glob_lexer.mli
```diff
@@ -0,0 +1,16 @@
...
+*)
+
+exception Syntax_error of string
+
+val parse_string : string -> Glob_matcher.pattern
```
aryx, line 18: obvious, but maybe worth adding (* this may raise Syntax_error *) in comment

src/osemgrep/targeting/Glob_parser.mly
```diff
@@ -0,0 +1,37 @@
...
+%}
+%token SLASH QUESTION STAR STARSTAR EOF
+%token <char> CHAR
+%token <Glob_matcher.char_class> CHAR_CLASS
+
+%start <Glob_matcher.component list> components
+%%
+
+components:
+| frags=list(fragment) SLASH comps=components
```
aryx: not a fan of those xxx=yyy. I prefer $1 $2, but whatever.

src/osemgrep/targeting/Glob_matcher.mli
```diff
@@ -0,0 +1,90 @@
...
+(* Append the components, taking care of trailing slashes that prevent
+   us from using a plain list append (@). *)
+val append : pattern -> pattern -> pattern
+
+(*
+   Compile the pattern into something efficient. The source should be
+   the original glob pattern before parsing. It's used only for debugging
+   purposes.
+*)
+val compile : source:loc -> pattern -> t
```
aryx, line 66: maybe worth adding that internally we rely on the ocaml-re library to compile and run globs?

src/osemgrep/targeting/Glob_matcher.ml
```diff
@@ -0,0 +1,131 @@
+(*
+   AST and matching of a glob pattern against a path.
+   This is purely syntaxic: the file system is not accessed.
+
+   We could use Re.Glob from the ocaml-re library but it doesn't expose
+   the AST of the glob pattern, and this prevents us from making
+   transformations required by gitignore such as treating the pattern
+   'foo/bar' as equivalent to '/foo/bar' but not treat 'foo' as '/foo'.
```
aryx: maybe add: "We do use though the Re library internally to compile and run our (transformed) glob patterns.

src/osemgrep/targeting/Glob_matcher.ml
```diff
@@ -0,0 +1,131 @@
...
+let map_root pat =
+  let open Re in
+  seq (bos :: map pat)
+
+(* Compile a pattern into an ocaml-re regexp for fast matching *)
+let compile ~source pat =
+  let re = map_root pat |> Re.compile in
+  { source; re }
+
+let debug = true
```
aryx: do we still want set to true?
Maybe use the Logs or logger library instead? See how I use it in the other osemgrep files.

src/osemgrep/targeting/Glob_matcher.ml
```diff
@@ -0,0 +1,131 @@
...
+  match pat with
+  | [ Component comp ] -> [ map_comp comp; eos ]
+  | [ Ellipsis ] -> []
+  | Component comp :: pat -> map_comp comp :: slash :: map pat
+  | Ellipsis :: pat -> rep (seq [ rep not_slash; slash ]) :: map pat
+  | [] -> [ eos ]
+
+let map_root pat =
+  let open Re in
+  seq (bos :: map pat)
```
aryx, line 111: bos? beginning-of-string? Maybe worth saying "this is what makes the pattern anchored by default"?

src/osemgrep/targeting/Semgrepignore.ml
```diff
@@ -0,0 +1,102 @@
...
+   Honor them with a deprecation warning.
+*)
+
+let create ?include_patterns ?(cli_patterns = []) ~project_root () =
+  if Fpath.is_rel project_root then
+    invalid_arg
+      ("Semgrepignore.create needs an absolute path for the project root: "
+      ^ Fpath.to_string project_root);
+  let include_filter =
+    match include_patterns with
```
aryx: Option.map (Include_filter.create ~project_root)

src/osemgrep/targeting/Realpath.ml
```diff
@@ -0,0 +1,156 @@
...
+  | Ok path -> realpath path |> Fpath.to_string
+  | Error (`Msg msg) -> invalid_arg ("Realpath.realpath_str: " ^ msg)
+
+(**********************************************************)
+(* Tests (Unix only) *)
+(**********************************************************)
+
+let make_test_tree () =
+  let cmd =
+    "rm -rf /tmp/test-realpath\n\
```
aryx: not a fan of rm -rf.
Why not use your Testutils_file library?

src/osemgrep/targeting/Include_filter.ml
```diff
@@ -0,0 +1,89 @@
...
+  match Gitignore_syntax.remove_negator str with
+  | None -> ()
+  | Some _ -> failwith ("--include patterns cannot be negated: " ^ str)
+
+let create ~project_root patterns =
+  List.iter check_nonnegated_pattern patterns;
+  let glob_matchers =
+    Common.map
+      (fun pat ->
+        Gitignore_syntax.parse_pattern
```
aryx, line 23: Why not using the Glob_matcher parser instead? That way you don't even need to check
for negated patterns since Glob_matcher does not allow them anyway.

src/osemgrep/targeting/Gitignore_syntax.mli
```diff
@@ -0,0 +1,62 @@
...
+   Show a list of selection events, one per line, oldest first
+   (unlike the list which has the most recent first).
+*)
+val show_selection_events : selection_event list -> string
+
+(* Path selector. *)
+type path_selector = {
+  loc : Glob_matcher.loc;
+  (* The matcher tells whether a given path matches the pattern.
+     For example, the pattern /foo matches the path /
```
aryx: the pattern /foo matches the path /? How?

src/osemgrep/targeting/Gitignore_syntax.mli
```diff
@@ -0,0 +1,62 @@
...
+
+     The result is the list of selection events, most recent first.
```
aryx: Is this comment still uptodate? matcher does not return a list of selection event, just an option.

src/osemgrep/targeting/Ftree.mli
```diff
@@ -0,0 +1,83 @@
...
+
+   Ftree stands for "file tree" like Fpath for file paths.
```
aryx: Do we still need this module? Seems unused for now.
Maybe add it in another PR when it's actually needed?

src/osemgrep/targeting/Git_path.mli
```diff
@@ -0,0 +1,47 @@
+(*
+   Abstract type for a file path within a git project.
+
+   This avoids issues of Unix-style vs. Windows-style paths. All
+   git paths use '/' as a separator.
+*)
+
+type t = private { string : string; components : string list }
```
aryx: maybe explain what string is? I guess the original argument to of_string?

aryx: maybe rename components to segments, to mimic the terminology in Fpath?
Also why not using Fpath which already provides most of the functions here (append, segments, /, etc.)?

src/osemgrep/targeting/Git_path.ml
```diff
@@ -0,0 +1,143 @@
...
+      Ok (create components)
+
+let of_fpath path = Fpath.segs path |> create
+
+let () =
+  Testutil.test "Git_path" (fun () ->
+      let test_str f input expected_output =
+        Alcotest.(check string) "equal" expected_output (f input)
+      in
+      let rewrite str = to_string (of_string str) in
```
aryx, line 98: I don't really understand what this test. It's the identity function since
anyway you'll store str in the string field and then return it in to_string.

src/osemgrep/targeting/Git_path.ml
```diff
@@ -0,0 +1,143 @@
...
+        | Ok _ -> false
+        | Error _ -> true
+      in
+      test_str norm "a" "a";
+      test_str norm "a/b" "a/b";
+      test_str norm "ab/cd" "ab/cd";
+      test_str norm "/" "/";
+      test_str norm "/a" "/a";
+      test_str norm "/a/b" "/a/b";
+      test_str norm "" ".";
```
aryx, line 120: why do we need yet another module that performs those normalization? We already do that in Realpath.ml,
then Fpath.ml also provides this normalization I think, so why another one?

src/osemgrep/targeting/Gitignore_syntax.mli
```diff
@@ -0,0 +1,62 @@
...
+(*
+   A ternary state indicating whether a path is selected or deselected at
+   the end of a level:
+   - Selected: select the file (= ignore it for git purposes)
+   - Deselected: deselect the file (= don't ignore it for git purposes)
+   - Unmatched: move on to the next level to determine the fate of the path.
+*)
+type selection_event =
+  | Selected of Glob_matcher.loc
+  | Deselected of Glob_matcher.loc
```
aryx, line 19: maybe add a comment like (* this is for negated patterns like !foo.c *)

src/osemgrep/targeting/Gitignore_level.ml
```diff
@@ -0,0 +1,9 @@
+(*
+   A level as defined in the gitignore specification.
+*)
+
+type t = {
```
aryx, line 10: when the .ml mostly contain types, I usually don't define a corresponding .mli since it's mostly a copy-paste.
I would just move all the good comments in Gitignore_level.ml and get rid of Gitignore_level.mli

src/osemgrep/targeting/Gitignore_filter.mli
```diff
@@ -0,0 +1,46 @@
...
+val create :
+  ?gitignore_filenames:string list ->
+  ?higher_priority_levels:Gitignore_level.t list ->
+  ?lower_priority_levels:Gitignore_level.t list ->
+  project_root:Fpath.t ->
+  unit ->
+  t
+
+(*
+   Examine a single absolute* path** and determine whether it is selected
```
aryx: why the extra *?

aryx: Maybe [1] and [2] would be more readable.

aryx (approved):
Nice work!

src/osemgrep/targeting/Git_path.mli
mjambon: I thought "component" was the standard name because that's the terminology Gerd was using in [Ocamlnet](http://projects.camlcity.org/projects/dl/ocamlnet-4.1.9/doc/html-main/Neturl.html). Daniel uses "segment" in Fpath. https://en.wikipedia.org/wiki/Path_(computing) uses "component" and https://en.wikipedia.org/wiki/URL uses "segment". https://www.rfc-editor.org/rfc/rfc3986#section-3.3 uses "segment" while using "path component" to refer to the whole path in a URL. Segment it is, then.

src/osemgrep/targeting/Ftree.mli
mjambon: I think we're not going to use it. Deleted.

src/osemgrep/targeting/Realpath.ml
mjambon: I don't have strong feelings against `rm -rf` _in particular_ but I don't like to rely on shell commands (especially in production). The `Testutils_file` module was created after this... Updating.

src/osemgrep/targeting/Glob_matcher.ml
mjambon: I forgot to turn it off. We want it on for the Semgrepignore unit tests and maybe on occasion in a wider context. The main issue is I can never remember how to use the logging stuff.

src/osemgrep/targeting/Gitignore_syntax.mli
mjambon: This was not intended

src/osemgrep/targeting/Gitignore_level.ml
mjambon: I agree - didn't realize at first it would be only a type definition.

src/osemgrep/targeting/Git_path.ml
mjambon: Git_path guarantees the use of slashes regardless of the platform we're on. Fpath, on the other hand, is platform-dependent, so a path that's valid on a POSIX platform may be invalid on Windows and vice-versa.

The main advantage is that these tests will work on all platforms.

Realpath is for normalizing actual paths in the filesystem. It is expected to be slow and should be used sparingly.

The normalization done by Git_path, like what Fpath does, is purely syntactic and good enough for some uses. It's used to determine whether a git path lies outside the project, in which case it should raise an error. For example, the git path `/src/../../foo` is normalized into `/../foo` which is outside of the git project and always invalid. By the way, Fpath (and POSIX?) path normalization treats `/..` as equivalent to `/` which I find problematic.

src/osemgrep/targeting/dune
mjambon: I'm fine with this practice **only for application code that doesn't get distributed as libraries**. I can't guarantee that the modules being created today have unique names in the OCaml world, and since OCaml hasn't been blessed with namespaces, `(wrapped false)` is likely to create incompatibilities with other libraries.

src/osemgrep/targeting/Glob_matcher.mli
mjambon: I think this new type is justified because this `loc` contains the line of data it's referring to, which we quote in case of an error. Since it's often a pattern that comes from the command line (`--exclude foo`), we need to keep it around. It's acceptable because patterns are short but it would not be acceptable for locations in programs in general.

src/osemgrep/targeting/Glob_matcher.mli
mjambon: It's two syntaxes for the same thing, semantically. It may or may not be compiled into the same thing later (we don't care because who writes `h[e]llo` anyway?).

src/osemgrep/targeting/Glob_matcher.mli
mjambon: I think I wanted an English name that distinguishes it from StarStar because we have a StarStar token that becomes Ellipsis if it's alone between two slashes or two successive stars if it's not alone. I understand that Ellipsis is confusing. Perhaps `Segment_repeat` would work.

mjambon: I'm calling it `Any_subpath` (until I change my mind again, of course).

src/osemgrep/targeting/Glob_matcher.mli
mjambon: It's a hand-written `show` that's more compact than what deriving would produce.

libs/commons/Testutil_files.mli
mjambon: No, because Fpath doesn't access the file system.

src/osemgrep/targeting/Glob_matcher.mli
aryx: still, even if you don't use deriving in the .ml, you can still use deriving in the .mli I think, and it will generate the show signature
(but I agree it's maybe misleading).

src/osemgrep/targeting/Unit_glob.ml
mjambon: yes, the glob patterns are anchored on both ends but not the gitignore patterns as the user sees them. I added that precision to the glob mli after your comment. The patterns found in gitignore/semgrepignore files are rewritten to use this correctly.

Say my project contains:
```
.git/
thing.c
src/
src/.gitignore
src/hello.c
src/tmp/
src/tmp/hello2.c
```
and `src/.gitignore` contains:
```
*.c
```
`*.c` is expanded into the anchored pattern `/src/**/*.c` so as to match `/src/hello.c` and `/src/tmp/hello2.c`.

src/osemgrep/targeting/Unit_glob.ml
mjambon: `**` behaves as `**` only if occupies a syntactic segment of its own. `a**` doesn't make sense and is interpreted as `a*` as per the gitignore spec. Maybe the user meant `a/**` but it's not the same.

src/osemgrep/targeting/Unit_semgrepignore.ml
mjambon: I like the visual separation into 2x2 cases.

src/osemgrep/targeting/Unit_semgrepignore.ml
mjambon: Yes, they should be the same because the presence of a slash in the middle of `b/c` makes it equivalent to `/b/c` as per the gitignore spec. I don't know why this rule exists.

image

src/osemgrep/targeting/Unit_semgrepignore.ml
mjambon: yes, because the pattern `b/c` must be treated as `/b/c`.

src/osemgrep/targeting/Glob_lexer.mli
mjambon: No, I don't think it is. It's documented by gitignore but I can't find a mention of this in the man pages for glob or fnmatch. I fixed the comment above.

src/osemgrep/targeting/Glob_parser.mly
mjambon: We should get you this 😂
image

mjambon: We should get you this 😂
image

src/osemgrep/targeting/Include_filter.ml
mjambon: It's because we need to add some anchoring depending on whether the pattern is considered anchored by git. Remember `a/` is unanchored but `a/b` is anchored. That nonsense is encoded in the `Gitignore_syntax` module.

The `check_nonnegated_pattern` call is to produce an error message if a pattern starts with an exclamation mark because it is almost certainly a mistake by the user. The glob pattern parser would otherwise accept `!` as a normal part of a path segment.

src/osemgrep/targeting/Git_path.ml
mjambon: `of_string` checks a few things.

src/osemgrep/targeting/Glob_matcher.ml
```diff
@@ -0,0 +1,136 @@
...
+  let re = map_root pat |> Re.compile in
+  { source; re }
+
+(* This is used during unit testing. *)
+let debug = ref false
+
+let run matcher path =
+  let res = Re.execp matcher.re path in
+  if !debug then
+    (* expensive string concatenation; may not be suitable for logger#debug *)
```
aryx, line 124: there's a logger#ldebug for lazy debug that takes the string as a lazy argument I think.

src/osemgrep/targeting/Unit_semgrepignore.ml
aryx: Maybe this weird thing about b/c being converted to /b/c but foo not being convertd to /foo needs a special comment somewhere.

mjambon: Yes, I think it's a good idea to explain tricky tests.

PR #6928, opened 2023-01-24
feat(cli): label pro matches

brandonspark:
## What:
This PR labels the matches that occur due to Semgrep and Semgrep PRO separately.

## Why:
Currently, in `pro`, we run once with all the rules labelled `deepsemgrep: true`, and once with all the other rules. These outputs are all collected, however, and it is not evident at reporting time where these matches came from.

This PR just changes it such that we separately report matches from Semgrep and Semgrep PRO.

## How:
Changed the Semgrep output interface, and passed the information up, to be consumed in `text.py`.

I kept the `is_pro_match` field in the CLI output so that App can consume this data, if necessary.

## Test plan:
image

## Future work:
We should display which rules were run with the OSS engine, when `--deep` is specified.

Closes PA-2432

PR checklist:


cli/src/semgrep/formatter/json.py
```diff
@@ -30,6 +30,7 @@ def _rule_match_to_CliMatch(rule_match: RuleMatch) -> out.CliMatch:
             lines="".join(rule_match.lines).rstrip(),
             metavars=rule_match.match.extra.metavars,
             dataflow_trace=rule_match.dataflow_trace,
+            is_pro_match=rule_match.match.extra.is_pro_match,
```
emjin: I suspect we're going to need to differentiate further than this in the future. We'd probably be better off with an enum (I'm expecting we'll want `matcher_engine=SEMGREP|PRO_MULTIFILE|PRO_INTERPROC`)

cli/src/semgrep/formatter/text.py
```diff
@@ -684,19 +684,31 @@ def format(
...
+                )
+            else:
+                oss_matches = [
+                    x for x in first_party_blocking if not x.match.extra.is_pro_match
+                ]
+                pro_matches = [
+                    x for x in first_party_blocking if x.match.extra.is_pro_match
+                ]
+                generate_output(f"{blocking_description}", oss_matches)
+                generate_output(f"Semgrep PRO Findings", pro_matches)
```
emjin, line 715: When we add `semgrep --deep` testing we should have a snapshot test for this output

emjin (commented):
I'll review this again with the semgrep pro PR. Let's also check with @milanwilliams on what the output should look like.

src/core/Pattern_match.ml
```diff
@@ -80,6 +80,11 @@ type t = {
      We now rely on equality of taint traces, which in turn relies on equality of `Parse_info.t`.
   *)
   taint_trace : taint_trace Lazy.t option;
+  (* This is a flag indicating whether this match was produced during a run of Semgrep PRO.
+     This will be overrided later by the PRO engine, on any matches which are produced
+     from a PRO run.
+  *)
+  is_pro_match : bool;
```
emjin: This goes better in `Report.ml` than `Pattern_match.ml`. We don't need to annotate each pattern match with this information, just the final result

emjin (approved):
Approving to not block you but have some comments. Make sure to make a PR on semgrep-interfaces before merging!

src/core/Pattern_match.ml
brandonspark: Resolved in other PR

PR #6930, opened 2023-01-24
feat(SSC): New Lockfile Parsers!

mmcqd:
Made all at once with @brendongo's permission. Should be easier to read because it's just "add new file, slot it into a dictionary" 
* Adds all the new parsers, except for Rust's Cargo.lock, because it's literally never been used. We can discuss removing support for this some other time.
* Adds support for line numbers in every language except Rust (for reasons above), and transitivity detection in every ecosystem except Rust and Go. We actually need to switch to parsing a different Go dependency file all together, but that can be another PR
* Some test additions
* Changes the `upto` and `not_any` parsers to take *args instead of a list, since it's easier to write

PR checklist:


mmcqd:
Thought I could simplify things in the requirements.txt parser by preprocessing some stuff away, but it messes up line numbers :(
Edit: actually it was easy

mmcqd:
All the line number changes are correct. It's mainly changing from indicating the line where the "version" text appears, to indicating the line where the whole dependency description begins

cli/src/semdep/parsers/go_sum.py
```diff
@@ -0,0 +1,62 @@
+"""
+Parser for go.sum files
```
aryx, line 2: Please add a link to a spec when possible for each of those lockfile, possible with a quick example.
For example here you can add a link to https://go.dev/ref/mod#go-sum-files

cli/src/semdep/parsers/requirements.py
```diff
@@ -0,0 +1,141 @@
+"""
+Parsers for requirements.txt and requirements.in files
+This is based on the spec described here: https://pip.pypa.io/en/stable/reference/requirements-file-format/
```
aryx: Nice! thanks.

cli/src/semdep/parsers/gem.py
```diff
@@ -0,0 +1,83 @@
...
+from semdep.parsers.util import mark_line
+from semdep.parsers.util import safe_path_parse
+from semdep.parsers.util import transitivity
+from semdep.parsers.util import upto
+from semgrep.semgrep_interfaces.semgrep_output_v1 import Ecosystem
+from semgrep.semgrep_interfaces.semgrep_output_v1 import FoundDependency
+from semgrep.semgrep_interfaces.semgrep_output_v1 import Gem
+
+
+# Examples:
```
brendongo, line 23: Love the examples. Makes it very easy to understand what is happening

cli/src/semdep/parsers/yarn.py
```diff
@@ -0,0 +1,245 @@
...
+# "@ampproject/remapping@^2.0.0"
+# bad-lib@0.0.8
+def source1(quoted: bool) -> "Parser[Tuple[str,str]]":
+    return (
+        string("@")
+        .optional(default="")
+        .bind(
+            lambda at_prefix: upto("@", consume_other=True).bind(
+                lambda package: upto(
+                    *(['"', ","] + ([":"] if not quoted else []))
```
brendongo: Can you add examples of when there are : and , in the line

cli/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awaremonorepo.yaml-dependency_awaremonorepo/results.txt
```diff
@@ -87,7 +87,7 @@ SEMGREP_USER_AGENT_APPEND="pytest" SEMGREP_SETTINGS_FILE="<MASKED>" SEMGREP_VERS
             "found_dependency": {
               "allowed_hashes": {},
               "ecosystem": "npm",
-              "line_number": 6,
+              "line_number": 5,
```
brendongo, line 90: Did you expect line numbers to change?

cli/src/semdep/parsers/package_lock.py
```diff
@@ -0,0 +1,95 @@
...
+            version = fields.get("version")
+            if not version:
+                logger.info(f"no version for dependency: {package}")
+                continue
+            try:
+                Version(version.as_str())
+            # Version was a github commit
+            except InvalidVersion:
+                logger.info(f"no version for dependency: {package}")
+                continue
```
brendongo: Is this necessary? Does FoundDependency not play well if version is not semver?

brendongo (approved):

cli/src/semdep/parsers/go_sum.py
mmcqd: I think I failed to find specs for some of these a while ago and gave up, but I'm gonna search again. Thanks for pointing out the Go one. Many of them have a page to the effect of "we have lockfiles" but with no grammar. I'll link those though because it's better than nothing.

mmcqd: Ok added everything I could find

cli/src/semdep/parsers/yarn.py
mmcqd: Done

aryx:
Thanks for the links. The examples are also very nice to help understand the grammars.

PR #6931, opened 2023-01-25
feat!: Pro Testing Workflow

brandonspark:
## What:
This PR adds in a GitHub action that, on PR and dispatch, will run the most recent version of Semgrep Pro (that is, the one built by the current `develop` branch) with the latest built `semgrep-core` and `semgrep` code.

## Why:
This is important to make sure that we know when we are releasing breaking changes that are incompatible with the latest version of Semgrep Pro! Once we merge this, then we will know that so long as this test passed (the repo is green), and there hasn't been a possibly-breaking `semgrep-proprietary` PR since then (which there shouldn't be, I think), then it is safe to release.

## How:
Added an extra test which will downloaded Semgrep Pro and run it with the just-built version of Semgrep. This means that on PR, the local changes to the PR will be taken into account.

## Test plan:
CI!

Closes PA-2390

PR checklist:

### DIFF
--- a/.github/workflows/test-semgrep-pro.yaml
+++ b/.github/workflows/test-semgrep-pro.yaml
@@ -27,16 +27,17 @@ jobs:
     name: Set up Docker tag based on if this is a pull request
     runs-on: ubuntu-22.04
     outputs:
-      docker-tag: ${{steps.setup-docker-tag.outputs.docker-tag}}
+      docker-tag: ${{ steps.setup-docker-tag.outputs.docker-tag }}
     steps:
       - name:
         id: setup-docker-tag
         run: |
+          echo "Github event is ${{ github.event_name }}"
           if [ "${{ github.event_name }}" = "pull_request" ]; then
-            echo "docker-tag=${{ github.event.pull_request.number }}" >> $GITHUB_OUTPUT
+            echo ::set-output name=docker-tag::${{ github.event.pull_request.number }}
             echo "Setting docker tag to current pull request number"
           else
-            echo "docker-tag=develop" >> $GITHUB_OUTPUT
+            echo ::set-output name=docker-tag::develop
             echo "Setting dry-run to develop"
           fi
   test-semgrep-pro:
--- a/.github/workflows/test-semgrep-pro.yaml
+++ b/.github/workflows/test-semgrep-pro.yaml
@@ -57,6 +57,9 @@ jobs:
       - name: Load image
         run: |
           docker load --input /tmp/image.tar
+      - name: See images
+        run: |
+          docker images
       - name: Run Semgrep Pro Engine!
         run: |
           docker run --rm -v "$(pwd):/root" -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint=bash "${{ inputs.repository-name }}:${{ needs.setup-docker-tag.outputs.docker-tag }}" /root/test-pro.sh
--- a/.github/workflows/test-semgrep-pro.yaml
+++ b/.github/workflows/test-semgrep-pro.yaml
@@ -68,4 +68,4 @@ jobs:
           pwd
       - name: Run Semgrep Pro Engine!
         run: |
-          docker run --rm -v "$(pwd):/root" -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint=bash "${{ inputs.repository-name }}:pr-${{ needs.setup-docker-tag.outputs.docker-tag }}" /root/test-pro.sh
+          docker run --rm -v "$(pwd):/root" -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint=bash "${{ inputs.repository-name }}:pr-${{ needs.setup-docker-tag.outputs.docker-tag }}" /root/scripts/test-pro.sh
--- a/.github/workflows/test-semgrep-pro.yaml
+++ b/.github/workflows/test-semgrep-pro.yaml
@@ -59,5 +59,4 @@ jobs:
           docker load --input /tmp/image.tar
       - name: Run Semgrep Pro Engine!
         run: |
-          ls
-          docker run --rm -v "$(pwd):/root" -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint=bash "/tmp/image.tar" /root/test-pro.sh
+          docker run --rm -v "$(pwd):/root" -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint=bash "${{ inputs.repository-name }}:${{ needs.setup-docker-tag.outputs.docker-tag }}" /root/test-pro.sh
--- a/.github/workflows/test-semgrep-pro.yaml
+++ b/.github/workflows/test-semgrep-pro.yaml
@@ -71,18 +71,13 @@ jobs:
           role-duration-seconds: 900
           role-session-name: "semgrep-deploy"
           aws-region: us-west-2
-      - name: Download Semgrep Pro binary
+      # This is the `develop` binary, so this is truly the most recent version of
+      # `semgrep-proprietary` from that repository's `develop` branch.
+      # We test with this so that we know whether any changes we make on this PR are breaking with
+      # the `develop` branch of `pro`.
+      - name: Download Semgrep Pro `develop` binary
         run: |
           aws s3 cp s3://web-assets.r2c.dev/assets/semgrep-core-proprietary-manylinux-develop ./semgrep-core-proprietary
-      - name: See images
-        run: |
-          docker images
-      - name: ls
-        run: |
-          ls -a
-      - name: pwd
-        run: |
-          pwd
       - name: Run Semgrep Pro Engine!
         run: |
           docker run --rm -v "$(pwd):/root" -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint=bash "${{ inputs.repository-name }}:pr-${{ needs.setup-docker-tag.outputs.docker-tag }}" /root/scripts/test-pro.sh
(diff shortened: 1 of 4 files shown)


.github/workflows/semgrep-pro.yml
```diff
@@ -0,0 +1,36 @@
...
+    if: (github.actor != 'dependabot[bot]')
+    steps:
+      - uses: actions/checkout@v3
+      - run: |
+          semgrep install-semgrep-pro
+      - run: |
+          echo "public class Foo { public void foo(int x){ return sink(x); } public void main() { foo(tainted); } }" > program.txt
+          echo -e "rules:\n- id: taint-rule\n  message: \"foo\"\n  severity: ERROR\n  languages: [ java ]\n  mode: taint\n  pattern-sources:\n  - pattern: \"tainted\"\n  pattern-sinks:\n  - pattern: \"sink(...)\"" > rule.txt
+      - run: |
+          semgrep --config rule.txt program.text --deep
```
IagoAbal: `--interfile` now

.github/workflows/test-semgrep-pro.yaml
```diff
@@ -0,0 +1,48 @@
...
+        required: true
+        type: string
+        description: Name (key) to use when uploading the docker image tarball as a artifact
+      repository-name:
+        required: true
+        type: string
+        description: The repository/name of the docker image to push, e.g., returntocorp/semgrep
+
+jobs:
+  push-docker:
```
spencerdrak: This job needs a new name

.github/workflows/test-semgrep-pro.yaml
```diff
@@ -0,0 +1,48 @@
...
+        uses: docker/setup-buildx-action@v2
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.artifact-name }}
+          path: /tmp
+      - name: Load image
+        run: |
+          docker load --input /tmp/image.tar
+      - uses: docker/login-action@v2
```
spencerdrak: This step shouldn't be necessary, the auth info is only needed for push/pull. You're doing a load from a local file, so you're fine.

.github/workflows/test-semgrep-pro.yaml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Run Semgrep Pro Engine!
```
spencerdrak, line 81: Does this command handle installing the deep-semgrep binary? if so, is versioning around that required?

.github/workflows/test-semgrep-pro.yaml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          docker run --rm -it "returntocorp/semgrep:pr-{{ github.event.pull_request.number }}" --config "p/deepsemgrep" . --interfile
```
spencerdrak: Does this semgrep command need access to the `SEMGREP_APP_TOKEN` env var? If so, you'll need to something like `docker run --rm -it -e SEMGREP_APP_TOKEN=$SEMGREP_APP_TOKEN "returntocorp/semgrep:pr-...` to ensure that env var is propagated into the container.

.github/workflows/tests.yml
```diff
@@ -352,6 +352,15 @@ jobs:
       artifact-name: image-dev
       repository-name: ${{ github.repository }}-dev
 
+  test-semgrep-pro:
+    needs: [build-test-dev-docker]
+    uses: ./.github/workflows/test-semgrep-pro.yaml
+    if: github.ref == 'refs/heads/develop'
```
spencerdrak: This is your call, but if it's not too expensive, we'd strongly recommend running all checks at PR time in addition to merge time. Might be worth removing this.

.github/workflows/test-semgrep-pro.yaml
spencerdrak: Nit: sub `${{ inputs.repository-name }}` in place of `returntocorp/semgrep`

spencerdrak: It looks like `{{ github.event.pull_request.number }}` will be null in cases where it's a workflow-call, workflow-dispatch, or push event. Consider setting up a step at the top with the following logic:
- if it's a pull request, use the PR number
- if it's not a PR or that field is blank, use the `develop` docker image tag

spencerdrak: Also this should probably be `${{ ... }}`

spencerdrak (changes requested):

.github/workflows/test-semgrep-pro.yaml
brandonspark: good call!

.github/workflows/test-semgrep-pro.yaml
brandonspark: nope, my intention is that we try it with the latest `pro` binary, every time

.github/workflows/tests.yml
brandonspark: i literally don't even know what this does, so let's do it

.github/workflows/test-semgrep-pro.yaml
brandonspark: I didn't really know how to do this, but I approximated

.github/workflows/test-semgrep-pro.yaml
```diff
@@ -0,0 +1,72 @@
...
+          name: ${{ inputs.artifact-name }}
+          path: /tmp
+      - name: Load image
+        run: |
+          docker load --input /tmp/image.tar
+      - name: Run Semgrep Pro Engine!
+        run: |
+          docker run --rm -it -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint bash "${{ inputs.repository_name }}:pr-${{ github.event.pull_request.number }}"
+
+          semgrep install-semgrep-pro
```
spencerdrak: This isn't going to do what you need it to, this will simply run the docker image (which'll then do whatever the default thing is, I think just a regular semgrep scan) and then will try to run semgrep outside the context of the running docker image. You might want to mount a script into the container, and run that script, or just do something of the form: `docker run ... <args> ... returntocorp/semgrep:pr-123 "semgrep install-semgrep-pro; semgrep --config p/deepsemgrep . --interfile"` 

Not sure that syntax works exactly, but happy to discuss if you want more details.

.github/workflows/test-semgrep-pro.yaml
```diff
@@ -0,0 +1,72 @@
...
+          semgrep --config "p/deepsemgrep" . --interfile
+
+  test-semgrep-pro-with-develop:
+    name: Test Semgrep Pro Engine with develop version of Semgrep
+    if: ${{ github.event.pull_request.number == null }}
+    runs-on: ubuntu-22.04
+    env:
+      SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
```
spencerdrak, line 51: duplication, is this still needed here?

.github/workflows/test-semgrep-pro.yaml
```diff
@@ -0,0 +1,72 @@
...
+        description: Name (key) to use when uploading the docker image tarball as a artifact
+      repository-name:
+        required: true
+        type: string
+        description: The repository/name of the docker image to push, e.g., returntocorp/semgrep
+
+jobs:
+  test-semgrep-pro-with-pr:
+    name: Test Semgrep Pro Engine with PR version of Semgrep
+    if: ${{ github.event.pull_request.number != null }}
```
spencerdrak: There's probably better ways to do this, I'll include a short example

spencerdrak: https://github.com/returntocorp/semgrep/blob/46bb38e789dce4ac7f4f45cfadf9acbd261b80e1/.github/workflows/release.yml#L35

Try to do something like this, but have the output be the desired tag. Either something of the form `pr-123` or `develop` if you're not on a PR.

.github/workflows/test-semgrep-pro.yaml
```diff
@@ -0,0 +1,72 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ inputs.artifact-name }}
+          path: /tmp
+      - name: Load image
+        run: |
+          docker load --input /tmp/image.tar
+      - name: Run Semgrep Pro Engine!
+        run: |
+          docker run --rm -it -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint bash "${{ inputs.repository_name }}:pr-${{ github.event.pull_request.number }}"
```
spencerdrak:
```suggestion
          docker run --rm -it -e SEMGREP_APP_TOKEN=${{ secrets.SEMGREP_APP_TOKEN }} --entrypoint bash "${{ inputs.repository_name }}:${{ steps.docker-tag.outputs.docker-tag }}"
```

.github/workflows/test-semgrep-pro.yaml
spencerdrak: You can set an output like `docker-tag` and use it as suggested below.

spencerdrak (changes requested):

brandonspark:
@spencerdrak ok, it's very close to working, but unfortunately it seems it's having trouble finding the script. could that be because the `checkout` is checking out a particular branch (perhaps `develop`) that doesn't have the script?

scripts/test-pro.sh
```diff
@@ -0,0 +1,11 @@
+#! /usr/bin/env bash
```
spencerdrak, line 1: Consider adding `bash -e` here so that failures will short-circuit.

spencerdrak (approved):
one nit, but overall, this LGTM.

brandonspark:
@spencerdrak seems I cannot merge this until it passes all checks -- do you know how to fix the shellcheck errors?

spencerdrak:
@brandonspark [take a read over this](https://github.com/rhysd/actionlint/blob/main/docs/checks.md) - I believe it's complaining about the unquoted `GITHUB_OUTPUT` bash vars.

brandonspark:
@spencerdrak as they say, ten four. that helped a lot, thanks!

PR #6940, opened 2023-01-25
fix(SSC): Fix poetry.lock parser failing on quoted closing bracket

mmcqd:
The parser was failing on a list like `["foo[bar]"]`, now it's not.

PR checklist:


cli/src/semdep/parsers/poetry.py
```diff
@@ -8,29 +8,49 @@
...
+from semdep.parsers.util import upto
 from semgrep.semgrep_interfaces.semgrep_output_v1 import Ecosystem
 from semgrep.semgrep_interfaces.semgrep_output_v1 import FoundDependency
 from semgrep.semgrep_interfaces.semgrep_output_v1 import Pypi
 
+# These use [until] instead of [upto] because [upto] only works on single characters
+# We don't care about the contents of any list or object values right now
 
-# A value in a key-value pair.
+# Using ]\n allows us to avoid issues with closing brackets inside strings
```
brendongo, line 28: Is ]\n in a string not allowed?

brendongo:
```
[
   coverage[
      toml
   ]
]
```

brendongo (approved):

mmcqd:
Ok, the tests broke, this is wrong lol

brendongo:
The updated snapshot looks like something broke

cli/src/semdep/parsers/poetry.py
mmcqd: Well... I don't know for sure honestly, due to the lack of docs. It seems very unlikely to me that poetry would auto-generate an unquoted newline in a string, but I might be wrong. Perhaps better to just write a legit parser for the contents of these things...

mmcqd: It would be
```
[
    "coverage[
        toml
    ]"
]
```
which seems very weird to me. Is there any format that allows unescape newlines in single quote strings?

mmcqd: I'm gonna go with it for now

PR #6943, opened 2023-01-26
fix(SSC): Two bugs connecting reachability and transitivity

mmcqd:
Fixes two bugs:
1. If `foo < 1.0.0` is dangerous, and a lockfile has package `foo 1.0.0` directly, and `foo 0.9.0` transitively, and uses `foo` in their code, we previously would produce a reachable finding at `foo 0.9.0`. This is incorrect, because any use of `foo` in scanned code is actually using `foo 1.0.0`. Now if we find a seemingly reachable transitive dependency, we check to see if there is a direct counterpart, and if so we just skip generating that finding, leaving it to be generated as unreachable later
2. Previously if a rule generated a reachable finding in a lockfile, we would never have the rule generate an unreachable finding in that same lockfile. This was to prevent generating a duplicate unreachable finding for every reachable finding. Now   to avoid this we track the information `(lockfile_path,package,version,transitivity)` of each reachable finding, and only prevent generating unreachable findings when all that information matches a reachable finding we have already produced.
PR checklist:


changelog.d/sc-547.fixed
```diff
@@ -0,0 +1 @@
+Fixed bug where SSC would generate reachable findings for transitive dependencies, even if a direct dependency at a safe version exists
```
underyx: not sure if SSC is a user-facing acronym, maybe write Supply Chain? it's also confusing since Semgrep Code is abbreviated SC

changelog.d/sc-549.fixed
```diff
@@ -0,0 +1 @@
+Fixed bug where SSC would fail to generate unreachable findings for a particular depencency if a reachable finding for that dependency had already been found in the same lockfile
```
underyx:
```suggestion
Fixed bug where SSC would fail to generate unreachable findings for a particular dependency if a reachable finding for that dependency had already been found in the same lockfile
```

underyx:
> 1. This is incorrect, because any use of `foo` in scanned code is actually using `foo 1.0.0`

technically, you can still import any version in javascript by doing something like `import { groupBy } from "node_modules/direct-dep/node_modules/lodash"`. And our semgrep rules would match that, cause I think we check for the library name `lodash` with a string containment operator in the right hand side string.

But that sounds obscure enough to ignore I guess?

changelog.d/sc-549.fixed
underyx: could you add a sentence explaining when you'd want the fixed behavior? I wanted to ask that to understand the point of this PR and figured you might as well explain via the changelog instead of with just a comment :D

mmcqd:
> > 1. This is incorrect, because any use of `foo` in scanned code is actually using `foo 1.0.0`
> 
> technically, you can still import any version in javascript by doing something like `import { groupBy } from "node_modules/direct-dep/node_modules/lodash"`. And our semgrep rules would match that, cause I think we check for the library name `lodash` with a string containment operator in the right hand side string.
> 
> But that sounds obscure enough to ignore I guess?

Yeah, there's actually at least one other edge case like this around having multiple transitive copies of a dependency. I'll open a linear ticket describing them.

changelog.d/sc-547.fixed
mmcqd: just switched to Semgrep

changelog.d/sc-549.fixed
mmcqd: Let me know if the update helps. Basically if `foo < 2.0.0` is vulnerable, and you have `foo 1.0.0` directly and it's reachable, but you also have `foo 1.0.1` transitively, you still want an unreachable finding on that transitive dependency. Before this PR you would not get one.

changelog.d/sc-549.fixed
```diff
@@ -0,0 +1 @@
+Fixed bug where Semgrep would fail to generate unreachable findings for a particular dependency if a reachable finding for that dependency had already been found in the same lockfile, even if there are multiple copies of this dependency at different vulnerable versions.
```
underyx:
```suggestion
Fixed bug where Semgrep would fail to generate unreachable findings for a particular library if a reachable finding for that library had already been found in the same lockfile, even if there are multiple copies of this library at different vulnerable versions.
```
maybe I'd phrase it like this, I always thought of `dependency` as `(library, version)` and that's why I was confused; this is why I think you can have the same library as a direct dependency and as a transitive dependency at the same time

cli/tests/e2e/targets/dependency_aware/transitive_and_direct/direct_reachable_transitive_unreachable/package-lock.json
```diff
@@ -0,0 +1,70 @@
...
+                    "dev": true,
+                    "requires": {
+                        "@types/sizzle": "*"
+                    }
+                }
+            }
+        },
+        "@types/jquery": {
+            "version": "3.3.0",
+            "resolved": "https://registry.npmjs.org/@types/jquery/-/jquery-3.3.22.tgz",
```
underyx, line 53: this says 3.3.22; that's because you manually edited the file, right?

cli/src/semgrep/dependency_aware_rule.py
```diff
@@ -109,9 +117,25 @@ def generate_unreachable_sca_findings(
     return non_reachable_matches, dep_rule_errors, targeted_lockfiles
 
 
+@lru_cache(maxsize=1000)
```
underyx:
```suggestion
@lru_cache(maxsize=100_000)
```

I usually do 100x expected cache size for a typical run, just to prevent crazy memory leaks

cli/src/semgrep/dependency_aware_rule.py
```diff
@@ -109,9 +117,25 @@ def generate_unreachable_sca_findings(
...
+    package: str, deps: Tuple[Tuple[str, Transitivity], ...]
+) -> bool:
+    """
+    Assumes that [dep] is transitive
+    Checks if there is a direct version of the transitive dependency [dep]
+    """
+    for other_package, transitivity in deps:
+        if other_package == package and transitivity == Transitivity(Direct()):
+            return True
+    return False
```
underyx:
```suggestion
    return (package, Transitivity(Direct())) in deps
```

this might be a simpler equivalent

underyx (approved):

cli/tests/e2e/targets/dependency_aware/transitive_and_direct/direct_reachable_transitive_unreachable/package-lock.json
mmcqd: yeah haha

PR #6954, opened 2023-01-26
fix: prevent crash when running tests on config with empty rules list

jtimmons:
PR checklist:

## Description
Fixes a small edge cases where `semgrep --test` fails when running against a yaml file with no rules in it - this is because the `config_contains_fix_key` check assumes that there's at least one rule in the rules list. Found this after our team disabled a subset of our tests.

example rule file to test against:
```yaml
rules: []
```


brendongo:
I don't think this would fix the bug you described. `rules: []` will still pass that condition. You probably want to add a check that the list is non empty before the first element access https://github.com/returntocorp/semgrep/pull/6954/files#diff-26dadf2b538a85df720f7872378ff02d0755ed5f11b3d4e5ea2ac64a0486a605R418

jtimmons:
> I don't think this would fix the bug you described. `rules: []` will still pass that condition. You probably want to add a check that the list is non empty before the first element access https://github.com/returntocorp/semgrep/pull/6954/files#diff-26dadf2b538a85df720f7872378ff02d0755ed5f11b3d4e5ea2ac64a0486a605R418

I confirmed that it fixes our issue locally! The existing check is only for the existence of the `"rules"` key but since `.get()` returns the actual value (if it exists) this will check its truthiness. An empty array is falsey so it hits the `else` condition

brendongo:
Ah yes of course. empty list is non truthy. 😅

brendongo (approved):

brendongo:
Mind adding a changelog entry to the PR and we can get this merged

jtimmons:
> Mind adding a changelog entry to the PR and we can get this merged

done!

changelog.d/gh-6954.fixed
```diff
@@ -0,0 +1 @@
+CLI: No longer fails when running tests against a config file with no rules in it
```
brendongo:
```suggestion
CLI: No longer fails when running tests against a config file with no rules in it

```

brendongo:
Thanks @jtimmons

PR #6957, opened 2023-01-26
feat(devops): add a command to dump just the path to the engine

emjin:
This is a quick PR to output the path to the engine. We'll probably want to also be able to output the path to the pro engine, but for now this will unblock devops. I will follow up this PR with a refactor.

Test plan: `semgrep --engine-path`

PR checklist:


cgdolan (approved):
Thank you!

ievans:
@emjin just musing, would "--show-engine-path" be better? `--engine-path` seems like it takes an argument which is the path to the engine

nmote (approved):
I agree with Isaac, but I don't have a strong opinion. Especially seeing as this will be obsolete with osemgrep.

PR #6964, opened 2023-01-27
GHA: fix regexp for hashFiles for core-cache optimisation

aryx:
Not sure it's worth keeping this optimisation now that
most of the modifications are in OCaml code anyway.

test plan:
We should now see a CI failure, because develop is currently
broken, but it remained unnoticed because we were using
old semgrep-core version to run the tests

PR checklist:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,7 +44,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ocaml-build-artifacts.tgz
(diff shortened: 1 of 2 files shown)


aryx:
Ok now we get the failure in CI: ERROR tests/e2e/test_join_rules.py::test_recursive_join_rules[rules/join_rules/recursive/flask-deep-stored-xss-example/flask-stored-xss.yaml-join_rules/recursive/flask-deep-stored-xss-example]


Note that the regexp was using .github/workflows/tests.yml, which I modifed in this PR, so for sure it was triggering the cache invalidation, but still, I think the regexp should also now catch changes in the new places where we store OCaml code.

PR #6956, opened 2023-01-26
Add a test to ensure all dynamically linked libraries are in /usr/lib

cgdolan:
Run the exact same actions as in the release workflow to shift left issues from release time to commit time. We'll work on de-duplication with reusable workflows in a followup PR. This one focuses on preventing https://github.com/returntocorp/semgrep/issues/6904 from reoccuring.

PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure on any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #6970, opened 2023-01-27
Adopt re-usable workflows

cgdolan:
Follow up to https://github.com/returntocorp/semgrep/pull/6956 that eliminates the duplication between test and release workflows.

PR checklist:


.github/workflows/build-test-osx-m1.yaml
```diff
@@ -0,0 +1,91 @@
+name: build-test-osx-m1
+
+on:
+  workflow_call:
```
spencerdrak, line 5: Can we add a workflow_dispatch here as well just in case it's needed in the future?

.github/workflows/build-test-osx-x86.yaml
```diff
@@ -0,0 +1,87 @@
+name: build-test-osx-x86
+
+on:
+  workflow_call:
```
spencerdrak, line 5: See above, maybe add a workflow_dispatch?

spencerdrak (approved):
LGTM - two nit comments I think we can save for later.

.github/workflows/build-test-osx-m1.yaml
cgdolan: Done

.github/workflows/build-test-osx-x86.yaml
cgdolan: Done

PR #6972, opened 2023-01-27
feat: reorder docker release steps, prevent inconsistencies in event of failures

spencerdrak:
In this way, we avoid releasing docker (a less risky release artifact) before other release artifacts (such as macos/ubuntu builds) in case of issues.

This avoids getting into the state where docker image vX.Y.Z is released, but we can't release on other platforms and need to scrap the release.

Note that this change must be merged after #6970 - the changes there are also included here.

PR checklist:


aryx:
Can we get https://github.com/returntocorp/semgrep/pull/6970 merged first so we can see just the diff related to the reordering?
Looks like the PR content also includes the modifs of chris.

cgdolan (approved):

.github/workflows/release.yml
```diff
@@ -135,6 +135,13 @@ jobs:
     name: Build and push Semgrep Docker image
     runs-on: ubuntu-latest
     if: ${{ !contains(github.ref,'-test-release') }}
+    needs:
```
aryx, line 138: Again a comment would be nice, to explain why we do that here (basically what you had in your PR comment, but
even though we can find back the info with git blame, I think it's better to put it in comment directly).

aryx (approved):

.github/workflows/release.yml
spencerdrak: Thanks for the comment - going to resolve as I want to get this in before @brandonspark releases.

PR #6977, opened 2023-01-27
refactor output, add `OutputExtra` class

brandonspark:
## What:
This collects some common fields that are output by `semgrep-core` in the CLI, namely profiling data, parsing data, matching explanations, and all targets, and puts them into a single extensible `OutputExtra` class.

## Why:
We pass this information up by name through like 7 calls or something, having to make both the code and the types very verbose. It's not always used in every context, so it's better to put them in a Python equivalent of a record, which will also make future work here easier. Before `osemgrep`, anyways.

## How:
Made a `OutputExtra`

PR checklist:


brandonspark:
what, this is so lame, why is it slow

brendongo (approved):

brandonspark:
@brendongo is it fine that it's slow? im not sure why that would be...

aryx (approved):
nice!

PR #7004, opened 2023-01-31
feat(cli): make pretty dataflow traces

brandonspark:
## What:
This PR makes dataflow traces in the CLI more easily readable.

## Why:
Currently, separating lines are not always printed on the findings themselves, leading to big blocks of taint traces which might make it difficult to parse which finding they actually attach to.

## How:
Added a newline where appropriate, and made it such that separating lines between findings are always printed when dataflow traces are specified.

On this file:
```

public class Foo {
   public int foo(int arg) {
     sink(arg);
   }

   public int bar(int arg) {
     sink(arg);
   }

   public int qux(int arg) {
     sink(arg);
   }



   public void main() {
     int x = 150;

     foo(x);
     bar(x);
     qux(x);
   }
}
```
with this rule
```
rules:
- id: test
  mode: taint
  pattern-sources:
    - pattern: |
        150 
  pattern-sinks:
    - pattern: |
        sink(...)
  message: stop
  languages: [java]
  severity: WARNING
  metadata:
    deepsemgrep: true
```
currently we see the following findings, in either the case where we limit it to one call to `foo`, and three calls to all three of `foo`, `bar`, and `qux`:
image
image

After this change, they will now be viewed as:
image
image

Note that this will always add the separating line, even when there is only one finding, when `--dataflow-traces` is specified. I view this as an OK compromise, though.

Closes PA-2471

PR checklist:

### DIFF
--- a/cli/src/semgrep/formatter/text.py
+++ b/cli/src/semgrep/formatter/text.py
@@ -266,7 +266,7 @@ class TextFormatter(BaseFormatter):
                         per_finding_max_lines_limit,
                         per_line_max_chars_limit,
                         False,
-                        True,
+                        False,
                     )
 
             if isinstance(call_trace.value, out.CliCall):
@@ -324,7 +324,7 @@ class TextFormatter(BaseFormatter):
                         per_finding_max_lines_limit,
                         per_line_max_chars_limit,
                         False,
-                        True,
+                        False,
                     )
 
             if sink:
--- a/cli/src/semgrep/formatter/text.py
+++ b/cli/src/semgrep/formatter/text.py
@@ -336,6 +336,7 @@ class TextFormatter(BaseFormatter):
                     per_finding_max_lines_limit,
                     per_line_max_chars_limit,
                 )
+                yield ""
 
     @staticmethod
     def _get_details_shortlink(rule_match: RuleMatch) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


IagoAbal:
I think I would put the `|-----------------------` line _before_ the finding rather than _after_ it. So in between two separating lines you have a finding + its trace. I think that's more consistent with how these separators are used without traces.

cli/src/semgrep/formatter/text.py
```diff
@@ -573,7 +574,10 @@ def _build_text_output(
                 color_output,
                 per_finding_max_lines_limit,
                 per_line_max_chars_limit,
-                is_same_file,
+                # if we have dataflow traces on, then we should print the separator,
```
emjin, line 577: Unclear from just this comment what "the separator" is; if it's clear from the rest of the code just ignore this

emjin (approved):
I agree with Iago that the line would make more sense to be over the finding instead of under, but I don't feel strongly

brandonspark:
New output:

1 finding:
image

2 findings:
image

Findings and dataflow traces:
1 trace:
image

Multiple traces:
image

does that seem all good, @emjin @IagoAbal ?

cli/src/semgrep/formatter/text.py
brandonspark: removed it

IagoAbal:
> does that seem all good, @emjin @IagoAbal ?

Looks great!

IagoAbal:
Don't know how difficult this would be, but ideally we don't repeat `test.java` all the time, we could skip printing the file name when it's the same, just print it when it changes (because of inter-file). I am looking at some taint traces and having the file name there all the time makes them less readable, specially when they are long paths.

If this is difficult, we could at least omit the filename in the `Taint flows through these intermediate variables:` part, since IIRC those are always in the same function as you already are.

PR #7007, opened 2023-01-31
feat!: make nested blocks visible to taint

brandonspark:
## What:
This PR just makes nested blocks visible to the taint engine, notably making them usable as sources or sinks.

## Why:
Sometimes we want to write rules which check if bad values go into nested blocks. This enables that functionality.

For instance, consider the following Terraform code:
```
resource "resource_ty" "resource_name" {
  setting1 = true 

  x { 
    sensitive_setting = var.x
  }
}
```

We would like to be able to write rules which track whether taint goes to the `sensitive_setting`. Unfortunately, it is inside of a `resource`, specifically inside of a `Record`. A `Call` inside of a `Record` is not understood, so this is just killed, and never makes it to the IL, and thus it can't be a source or a sink.

This PR just makes sure it makes it there.

## How:
Just made it so that when in `hcl`, the AST to IL conversion allows nested blocks.

Closes PA-2475

PR checklist:

### DIFF
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -890,7 +890,9 @@ and record env ((_tok, origfields, _) as record_def) =
                 we just want it in there at all so that we can use it as a sink.
              *)
              let field_expr = record env fields in
-             (* it is not clear to me if the prior_expr is necessary here. *)
+             (* We need to use the entire `prior_expr` here, or the range won't be quite
+                right (we'll leave out the identifier)
+             *)
              Some (Field (id, { field_expr with eorig = SameAs prior_expr }))
          | _ when is_hcl env.lang ->
              (* For HCL constructs such as `lifecycle` blocks within a module call, the
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -859,6 +859,39 @@ and record env ((_tok, origfields, _) as record_def) =
                _;
              } ->
              Some (Spread (expr env e))
+         | G.F
+             {
+               s =
+                 G.ExprStmt
+                   ( ({
+                        e =
+                          Call
+                            ( { e = N (Id (id, _)); _ },
+                              (_, [ Arg { e = Record fields; _ } ], _) );
+                        _;
+                      } as prior_expr),
+                     _ );
+               _;
+             }
+           when is_hcl env.lang ->
+             (* This is an inner block of the form
+                someblockhere {
+                  s {
+                    <args>
+                  }
+                }
+
+                We want this to be understood as a record of { <args> } being bound to
+                the name `s`.
+
+                So we just translate it to a field defining `s = <record>`.
+
+                We don't actually really care for it to be specifically defining the name `s`.
+                we just want it in there at all so that we can use it as a sink.
+             *)
+             let field_expr = record env fields in
+             (* it is not clear to me if the prior_expr is necessary here. *)
+             Some (Field (id, { field_expr with eorig = SameAs prior_expr }))
          | _ when is_hcl env.lang ->
              (* For HCL constructs such as `lifecycle` blocks within a module call, the
                 IL translation engine will brick the whole record if it is encountered.
--- a/changelog.d/pa-2475.fixed
+++ b/changelog.d/pa-2475.fixed
@@ -1,4 +1,4 @@
-Terraform: Blocks can now be used as sources and sinks for taint.
+Terraform: Nested blocks can now be used as sources and sinks for taint.
 For instance, the block `x` in
 
 resource $A $B {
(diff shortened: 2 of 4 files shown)


src/analyzing/AST_to_IL.ml
```diff
@@ -859,6 +859,41 @@ and record env ((_tok, origfields, _) as record_def) =
...
+                someblockhere {
+                  s {
+                    <args>
+                  }
+                }
+
+                We want this to be understood as a record of { <args> } being bound to
+                the name `s`.
+
+                So we just translate it to a field defining `s = <record>`.
```
aryx, line 887: Could we do this transformation at parsing time in Parse_hcl_tree_sitter.ml?

brandonspark: i think in principle, yes, because the main reason why we don't already is because some blocks have more than one "parameter":
```
s "parameter1" "parameter2" {
  <stuff>
}
```
this we can't really nicely translate to a `DefStmt`, because we want the matchability. the blocks I'm concerned with are of the form that they don't take in any extra parameters, however:
```
s {
  <stuff>
}
```

so I feel like this could in fact be done at parse time, by translating it to a `DefStmt` defining a record

it would be a digression from the general case of a block, however. you'd see that some blocks would get translated to `DefStmt` and some would become `ExprStmt (Call ...)`. this lack of homogeneity could cause some manner of confusion. what do you think?

aryx: Could we transform the s "parameter1" "parameter2" { } in a DefStmt FuncDef?
If it's really defining a function, then let's encode it as a function?
and then if it has no parameter, it's a variable and we can use DefStmt VarDef, looks simple no?

brandonspark: It looks simple, but unfortunately it's not. I said "parameter", but I didn't mean parameter in the sense of a function, I just meant parameter to a block. An instance of this block may be:

resource "type" "name" {
  arg1 = 1
  arg2 = "hi"
}

where "type" is the type the resource ascribes to, "name" is the name of the resource, and the arguments are the parameters used to configure the resource. It may also look like

settings {
  arg1 = true
}

in the case of a nested block. In neither case is either block truly a "function", and it was my predecessor's choice to keep them as Call's, which is the convention I've kept up with. My question is whether the relative cost of special casing the blocks that look like `settings`, at AST_generic time, is less confusing than special casing it at AST_to_IL time.

brandonspark: It's also worth noting that these are all Terraform specific notions, and that we claim to support "HCL", which is the more general language surrounding it. If we special case, it will go against the general scheme of HCL.

brandonspark: @aryx any thoughts?

aryx: I don't think we care about HCL outside terraform, so if those resources are kinda like functions and variables,
I'd rather transform them in that, so then we can reuse all the IL machinery we have as-is, without having to adapt things later. But I don't understand anything to terraform so ...

brandonspark: @aryx  I am a little hesitant to do this, because I would prefer that we translate things which are of the same syntactic class (blocks in HCL) to the same kind of construct. I feel like I would forget all the time if some blocks translated to `DefStmt`s, and some translated to `Call`s, nevermind the fact that something like
```
lifecycle {
  settings1 = 2
}
```
is not really "defining" a name of "lifecycle", but it is instead just a parameter that happens to be labelled "block", that contains some settings.

And indeed, none of these are really functions in any sense.

Right now, there is a nice invariant in `Terraform.ml` that any `DefStmt`s are strictly introduced via `transform_ast_for_analyzing`, or otherwise our pre-analysis Terraform-specific processing steps. This is nice because they truly are from `variable` blocks which do define names, so I would prefer to keep them as the only sources of truth for things which define names, because otherwise our logic which is intended to interact strictly with those definitions will also coincide with these blocks here, even though they are not definitions. Does that make sense?

IagoAbal: Looking at `Parse_hcl_tree_sitter`, I think it would be better to translate this as a `DefStmt` as Yoann suggests. Although I would always translate it into a `VarDef`/`FieldDefColon` like `<block-id> = (<block-params>, <block-body>)`. So e.g. this below could be represented as `x = (["parameter1"; "parameter2"], <stuff>)`:

```hcl
s "parameter1" "parameter2" {
  <stuff>
}
```

IagoAbal: The thing is that the Terraform translation is arbitrary, we use a `Call` as we could use something else. So it's better to change the `Call` to a `DefStmt` or whatever is needed, rather than keep the `Call` for no particular reason and then add extra code in `AST_to_IL` to cope with that.

aryx (changes requested):
We should put this on hold and try to see if we can generate better generic AST from terraform so that we would not need that.

brandonspark:
@aryx In light of what we discussed at the offsite, might we merge this PR to unblock @LewisArdern from writing Terraform rules?

aryx (approved):

Issue #3328, opened 2021-06-09
Support for Clojure


Gby56:
**Is your feature request related to a problem? Please describe.**
Given the current state of Clojure security tools, having semgrep would be a great addition into the ecosystem.

HN Thread: https://news.ycombinator.com/item?id=24929993
CRUSH, a code review tool that could inspire some rules.
- https://jemurai.com/2020/01/15/oss-crush/
- https://github.com/Jemurai/crush

**Describe the solution you'd like**
Parsing support for Clojure and some initial rules

**Describe alternatives you've considered**
None, there is just this really low-level tool for now https://github.com/clojure/tools.analyzer

DrewDennison:
https://github.com/sogaiu/tree-sitter-clojure and https://github.com/oakmac/tree-sitter-clojure both exist and if they have reasonable coverage. It shouldn't be too hard to add in

aryx:
There's also https://github.com/SergeevPavel/tree-sitter-clojure and https://github.com/Tavistock/tree-sitter-clojure
Which one to choose?

aryx:
https://github.com/sogaiu/tree-sitter-clojure seems to be the most up-to-date and maintained

aryx:
hopefully this will help for other lisp dialects

sogaiu:
Hi @aryx and any other interested folks,

This is just an FYI from sogaiu/tree-sitter-clojure maintainers.

We made some changes recently as described [here](https://github.com/sogaiu/tree-sitter-clojure/issues/33#issuecomment-1362363845).  The short of it is that we fixed one issue and enhanced functionality a bit.

We noticed the [`parser.c` here](https://github.com/returntocorp/semgrep-clojure/blob/main/lib/parser.c) might be a slightly older version than what we have currently (this is a guess based on noticing that it doesn't mention `sym_ns` or `sym_name`).

Our most recent efforts have those bits: https://github.com/sogaiu/tree-sitter-clojure/blob/master/src/parser.c#L109-L110

May be there is some interest in using the more recent version?

We're trying to make an effort to keep users of the grammar / parser informed about potential changes in an effort to mitigate breakage and such.  If there is interest in that sort of thing, please consider subscribing to [this issue](https://github.com/sogaiu/tree-sitter-clojure/issues/33).

Thanks!

---

[...]

PR #7012, opened 2023-01-31
Support metavar and ellipsis on Raw_tree

aryx:
This also adds experiment support for Clojure thanks
to the new code in Generic_vs_generic to provide
better matching on raw_trees (currently generated in
Parse_clojure_tree_sitter.ml coming from the new Boilerplate.ml
of ocaml-tree-sitter-core).

test plan:
test file included
make core-test

PR checklist:


aryx:
This closes #3328

aryx:
@mjambon maybe this could help for Apex too :)

src/matching/Generic_vs_generic.ml
```diff
@@ -1075,10 +1075,43 @@ and m_expr ?(is_root = false) a b =
...
+   *)
+  | List a, List b
+  | Tuple a, Tuple b ->
+      m_list_with_dots m_raw_tree
+        (function
+          | Token ("...", _) -> true
+          | Case (_, Token ("...", _)) -> true
+          | Case (_, Case (_, Token ("...", _))) -> true
+          | _ -> false)
+        ~less_is_ok:false (* empty list can not match non-empty list *) a b
```
mjambon, line 1114: 1. I'm worried that raw-tree lists don't correspond to lists in the original syntax. I think it's gonna be a problem for repeat1(x) constructs that compile to seq(x, repeat(x)) (= Tuple [x; List x]). Maybe that's the only problem, in which case we could do some work on the ocaml-tree-sitter side to convert any seq(x, repeat(x)) into repeat(x). This would lose the static guarantee that the list has at least one element, but it's usually not something we care about anyway.
2. I don't see how tuples matching ellipses is going to work but we'll see. If that works, maybe we don't have to worry about point (1).

PR #7021, opened 2023-01-31
add dry-run to docker push step

spencerdrak:
Test Plan:

Ensure PR checks run as expected (push-dev should do a dry run), then toggle switch and push.

PR checklist:

### DIFF
--- a/.github/workflows/push-docker.yaml
+++ b/.github/workflows/push-docker.yaml
@@ -11,6 +11,11 @@ on:
         required: true
         type: string
         description: The repository/name of the docker image to push, e.g., returntocorp/semgrep
+      dry-run:
+        required: true
+        type: boolean
+        description: Whether a dry-run (e.g., print tags to push) should be peformed. Actually push images if false.
+
   workflow_call:
     inputs:
       artifact-name:
@@ -21,6 +26,10 @@ on:
         required: true
         type: string
         description: The repository/name of the docker image to push, e.g., returntocorp/semgrep
+      dry-run:
+        required: true
+        type: boolean
+        description: Whether a dry-run (e.g., print tags to push) should be peformed. Actually push images if false.
 
 jobs:
   push-docker:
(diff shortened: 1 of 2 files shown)


.github/workflows/push-docker.yaml
```diff
@@ -11,6 +11,11 @@ on:
         required: true
         type: string
         description: The repository/name of the docker image to push, e.g., returntocorp/semgrep
+      dry-run:
+        required: true
+        type: boolean
+        description: Whether a dry-run (e.g., print tags to push) should be peformed. Actually push images if false.
```
cgdolan, line 17:
```suggestion
        description: Whether a dry-run (e.g., print tags to push) should be peformed. Actually push images if false.
        default: true
```

.github/workflows/push-docker.yaml
```diff
@@ -21,6 +26,10 @@ on:
...
```
cgdolan, line 33:
```suggestion
        description: Whether a dry-run (e.g., print tags to push) should be peformed. Actually push images if false.
        default: true
```

cgdolan (approved):

cgdolan (approved):

aryx:
Nice test plan:! Thanks.

.github/workflows/push-docker.yaml
```diff
@@ -21,6 +27,11 @@ on:
         required: true
         type: string
         description: The repository/name of the docker image to push, e.g., returntocorp/semgrep
+      dry-run:
```
aryx, line 30: This is so ugly. Why do we need to repeat the same thing for workflow_dispatch and worflow_call.
This yaml thing is retarded.

aryx (approved):

aryx:
Hmm actually I'm not sure I understand the test plan.
Did you go to https://github.com/returntocorp/semgrep/actions and then run some workflow manually with Dry run set to false?

.github/workflows/push-docker.yaml
cgdolan: I feel this comment falls short of our respect and inclusion values. All technologies have strengths and weaknesses and while I agree that this a weakness, I think it's inappropriate to use `retarded` as a pejorative term here. My understanding from the GitHub documentation on workflow syntax is that the `inputs` context is a child of both `workflow_dispatch` and `workflow_call` and can not be shared between them: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions.

cgdolan (approved):

.github/workflows/push-docker.yaml
aryx: Ok sorry, I'll try to better hide my hate for yaml.

Issue #7033, opened 2023-02-01
pip fails to install correctly the semgrep package from PyPi when installing together with other packages


jdeamicis:
**Describe the bug**
The following command is currently failing:
```
pip install black isort seed-isort-config flake8 flake8-docstrings flake8-bugbear flake8-mypy flake8-colors pep8-naming bandit semgrep
```
It seems like Python starts pulling all the versions of `semgrep` available on PyPi from latest to oldest, which causes the system to end up with the oldest version of `semgrep` available.
I'm not sure whether this is a bug in the `semgrep` package or in PyPi, but I noticed that a new version was recently released (1.8.0, released approx. 12 hours ago), and yesterday the problem was not occurring.

**To Reproduce**
On a clean Python environment, do:
```
$ pip install black isort seed-isort-config flake8 flake8-docstrings flake8-bugbear flake8-mypy flake8-colors pep8-naming bandit semgrep
```

After the installation:
```
$ pip list | grep semgrep
semgrep               0.82.0
```

**Expected behavior**
I would expect the latest version of `semgrep` to be installed.

**What is the priority of the bug to you?**

- [x] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

We can implement a temporary workaround of installing a specific version of `semgrep` (the latest one), but in the future that may cause inconsistencies in case the rules are updated on the semgrep registry.

**Environment**
I'm running on a Python virtualenv and installing `semgrep` from PyPi. The version of Python used does not seem to matter (I could reproduce the issue on 3.9.12 and 3.10.4).

samueljsb:
I believe this is because black 23.1.0 (the latest release from yesterday) requires `packaging>=22.0` and `semgrep` constrains it to `packaging~=21.0`.

Is there a reason `semgrep` needs to pin a specific major version of packaging? The comment in the `install_requires` section says:

> These version are flexible so semgrep can coexist with other tools.

but then advises that the major version should be pinned, which is causing this problem.

jdeamicis:
> I believe this is because black 23.1.0 (the latest release from yesterday) requires `packaging>=22.0` and `semgrep` constrains it to `packaging~=21.0`.
> 
> Is there a reason `semgrep` needs to pin a specific major version of packaging? The comment in the `install_requires` section says:
> 
> > These version are flexible so semgrep can coexist with other tools.
> 
[...]

Issue #7032, opened 2023-02-01
Install via pip causes all version of semgrep to be downloaded


anotherbridge:
**Describe the bug**
When installing semgrep in a freshly created virtual environment via pip all the package versions until 0.82.0 will be downloaded. This causes semgrep to use unnecessary space and also prolongs builds for a significant amount of time. Moreover, it will install version 0.82.0 instead of 1.8.0

Yet, when a previous version of semgrep is installed and an update is performed no issues occur. 

**To Reproduce**
```
python3 -m venv .venv
source .venv/bin/activate
pip install semgrep 
```

**Expected behavior**
Only the latest version, i.e. 1.8.0 should be downloaded and installed. 

**What is the priority of the bug to you?**

- [x] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
This occured on a Debian machine as well as on a Mac having an ARM processor.

samueljsb:
I can't reproduce this myself:

```console
$ python --version --version
Python 3.8.15 (default, Oct 11 2022, 21:52:37)
[Clang 14.0.0 (clang-1400.0.29.102)]
$ python -m venv venv
$ venv/bin/pip install semgrep --no-cache
Collecting semgrep
  Downloading semgrep-1.8.0-cp37.cp38.cp39.py37.py38.py39-none-macosx_10_14_x86_64.whl (23.5 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 23.5/23.5 MB 9.3 MB/s eta 0:00:00
Collecting glom~=22.1
  Downloading glom-22.1.0-py2.py3-none-any.whl (100 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100.7/100.7 KB 11.2 MB/s eta 0:00:00
Collecting jsonschema~=4.6
  Downloading jsonschema-4.17.3-py3-none-any.whl (90 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 90.4/90.4 KB 12.8 MB/s eta 0:00:00
Collecting ruamel.yaml<0.18,>=0.16.0
  Downloading ruamel.yaml-0.17.21-py3-none-any.whl (109 kB)
[...]

PR #7038, opened 2023-02-01
fix(pypi): vendor packaging 21.3

brendongo:
semdep relies on some behavior in packaging 21.3 that does not exist in 23.0. Namely
in 23.0 they removed support for non pypi compliant version strings. Since we operate on
version strings of other ecosystems we need this functionality. We need to switch out out using
packaging library for this but for now just copy the needed code from 21.3 so we can unpin

PR checklist:


cli/tests/unit/test_semver_matching.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+from semdep.package_restrictions import semver_matches
+
+@pytest.mark.quick
+@pytest.mark.parametrize(
+    "expression,candidate,match",
+    [
+        pytest.param(">1.0.0", "1.0.1", True),
```
mmcqd, line 10: these can just be tuples, you don't need the `pytest.param`

cli/tests/unit/test_semver_matching.py
```diff
@@ -0,0 +1,27 @@
...
+        pytest.param(">1.0.0", "2.0.0", True),
+        pytest.param(">=1.0.0", "1.0.0", True),
+        pytest.param(">1.0.0", "0.1.0", False),
+        pytest.param(">1.0.0", "0.1.0-beta", False),
+        pytest.param("<1.0.1", "1.0.1", False),
+        pytest.param("<1.0.1", "1.0.0", True),
+        # pytest.param("<1.0.1", "0.1.0-beta", True),
```
mmcqd: Could xfail this instead of commenting it out? Because in the ideal case it would work. For that you do need `pytest.param`, it's
```
pytest.param("<1.0.1", "0.1.0-beta", True, marks=pytest.mark.xfail)
```
See https://docs.pytest.org/en/7.1.x/how-to/skipping.html#skip-xfail-with-parametrize

mmcqd (approved):

brendongo:
Fixes https://github.com/returntocorp/semgrep/issues/7033 and https://github.com/returntocorp/semgrep/issues/7032

PR #7039, opened 2023-02-01
fix(CLI): Url reformatting for Bitbucket

pangngan:
Problem: 
Bitbucket URLs were incorrectly getting overwritten by `get_url_from_sstp_url`. Specifically, Bitbucket Server (on-prem) repo URLs were not getting parsed correctly. 

Solution:
Remove the call to that function and just use the raw value of the environment variable instead. I checked on a test repo on Bitbucket Cloud and it seems that we could just use the environment variable as is, and hopefully this will fix the issue with Bitbucket Server as well. 

PR checklist:

Closes APP-3279


pangngan:
Also, pre-commit failing is happening for everyone so apparently we can ignore that.

changelog.d/app-3279.fixed
```diff
@@ -0,0 +1 @@
+Fix incorrectly reformatted Bitbucket repository URLs
```
brendongo:
```suggestion
Fix incorrectly reformatted Bitbucket repository URLs

```

brendongo (approved):

changelog.d/app-3279.fixed
brendongo: This was failing pre-commit cause of formatting :D

PR #7045, opened 2023-02-02
fix(pre-commit): add back semgrep in precommit

emjin:
Use local rules for precommit to avoid blocking. Our main flow of rule management is through semgrep ci, so this run is mostly intended to help dogfood.

Test plan: precommit

PR checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -214,42 +214,44 @@ repos:
             # want an OCaml rule to be applied on a script.
             "--skip-unknown-extensions",
           ]
-  #! # Dogfooding .pre-commit-hooks.yml and setup.py
-  #! - repo: https://github.com/returntocorp/semgrep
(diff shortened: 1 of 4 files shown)


aryx (approved):
What's wrong with using p/python and p/bandit?
We can't fix the issues as they come when they change p/python and p/bandit?

emjin:
> What's wrong with using p/python and p/bandit? We can't fix the issues as they come when they change p/python and p/bandit?

It's very annoying to whoever encounters it because pre-commit is a required check and you'll get changes on code you didn't modify. @brendongo tells us that our rule management is mostly done with the `semgrep ci` run, so it doesn't seem worth it to me to have this be blocking. Maybe that's feedback for our pre-commit hook

emjin:
I'll merge this for now as a strict improvement over not having any hook and we can always change the rulepack back if we decide otherwise

PR #7049, opened 2023-02-02
osemgrep: use git ls-files when possible

aryx:
test plan:
osemgrep --config semgrep.jsonnet . --metrics=off --json
now takes only 10s in the semgrep repo (used to take
forever because it was analyzing all the files, including
in submodules).

Note that semgrep itself takes only 5s, and so is 2x faster
probably because we currently don't handle .semgrepignore
which means we analyze more files than needed.

PR checklist:


aryx:
This closes PA-2437

PR #7052, opened 2023-02-02
feat!: persist metavariable pattern bindings

brandonspark:
## What:
This PR makes it so that bindings that occur within a `metavariable-pattern` persist to outside of the `metavariable-pattern` itself.

## Why:
Sometimes, you want to be able to do something with a metavariable that is found within a `metavariable-pattern`.

From within the `metavariable-pattern` itself, the only thing that it can influence is whether or not to kill the range that it is modifying. This means that the original range always remains either completely killed, or untouched. Sometimes it is useful to make use of information that is found within a `metavariable-pattern` however -- in particular, the bindings within might be useful for the purpose of `focus-metavariable`.

For instance, suppose that we are interested in any occurrence of some pattern within another, but only within some particular _context_. For instance, we could write:
```
patterns:
  - pattern: |
      foo($A)
  - metavariable-pattern:
      metavariable: $A
      pattern: |
        bar($B)
```

Here we look for any occurrence of `bar($B)` within `$A`. But what if we are not interested in `bar($B)`'s range itself, but just the _contents_ of the call to `bar`? We want to find all occurrences of some sub-pattern within another, "in the context" of what encloses it, in this case the call to `bar`. In this case, we may instead want to do:
```
patterns:
  - pattern: |
      foo($A)
  - metavariable-pattern:
      metavariable: $A
      pattern: |
        bar($B)
  - focus-metavariable: $B
```

This also lets us use the metavariable of `$B` in the rule message and stuff. Overall, it's just nice to have, and there's no fundamental engine limitation that prevents us from doing so.

This is what this PR does.

## How:
I just made it such that the passed-down `nested_formula_has_match` function, instead of returning a boolean, returns a list of all the possible matches that can occur due to the nested formula.

I also needed to write logic to deal with the fact that, due to the fact that `metavariable-pattern` has an `mvar_context` that inherits the binding environment from the modified match, the matches that occur within the original range have duplicate copies of all the original metavariable bindings. I didn't want to duplicate those, so I just made it such that we only add metavariables from a nested match if they did not occur in the original match. By invariant, this ensures that we only add fresh bindings that were occurring from the `metavariable-pattern`.

Another difficulty was in using the nested bindings. These bindings are all localized to a temp file, which we place the metavariable pattern contents in. This means those bindings, in particular, have the wrong file information (and sometimes the wrong location information). So I had to construct a visitor to change the `mvalue` in each binding to an equivalent `mvalue` from the original file. The types didn't work in my favor here, so I also had to implement an `mvalue_to_any` function, since we only had visitors that worked on `any` and `program` and stuff, but none that work on `mvalue` (or its many variants).

### Why take care to not duplicate metavariables?

I'd prefer that we keep it such that there's only one metavariable of each name in each range. I think it's a clean, nice invariant, which hopefully holds in the first place.

### Why not just deduplicate the metavariables?

I originally had it this way, using almost identical code to that we use for intersecting ranges in the first place. Unfortunately, this is brittle to metavariables which bind sequences of statements, like `Ss`, which is not compatible with `equal_ast_bound_code`. I could have added it in as a case, but there may be more in the future, and by construction what I've done with omitting the existing metavariables should be safe.

### Why write `mvalue_to_any`?

`mvalue` has too many variants, not all of which permit a nice visitor. I could case on just the ones which do, but `Map_AST`'s interface of visitors is limited enough as it is, so I just decided to implement the function.

## Test plan:
`make test`

Also passes `make test` in `pro`.

PR checklist:


tests/rules/.bash_history
```diff
@@ -0,0 +1,10 @@
+ls
+ls
+cd root
+ls
+pwd
+ls
+cd ..
+ls
+cd root
+ls
```
nmote: Delete this

nmote (approved):
Looks pretty reasonable, and since you've run the idea past @IagoAbal and @aryx I think you should be good to go. Just make sure to remove the `.bash_history` file before merging :laughing:

tests/rules/.bash_history
brandonspark: how does this even happen...

changelog.d/pa-2490.added
```diff
@@ -0,0 +1 @@
+Rule syntax: Metavariable bindings bound within `metavariable-pattern` now persist to outside of the `metavariable-pattern`
```
IagoAbal, line 1: This is an important change for the release notes @adamkvitek.

src/engine/Match_search_mode.ml
```diff
@@ -523,17 +523,18 @@ let children_explanations_of_xpat (env : env) (xpat : Xpattern.t) : ME.t list =
...
          match cond with
          | R.CondEval e ->
              let env = Eval_generic.bindings_to_env env.xconf.config bindings in
-             Eval_generic.eval_bool env e
+             Eval_generic.eval_bool env e |> map_bool r
          | R.CondNestedFormula (mvar, opt_lang, formula) ->
              (* TODO: could return expl for nested matching! *)
              Metavariable_pattern.satisfies_metavar_pattern_condition
-               nested_formula_has_matches env r mvar opt_lang formula
+               get_nested_formula_matches env r mvar opt_lang formula
```
IagoAbal, line 537: Wait, this is returning the matches of `metavariable-pattern`!

IagoAbal: This can lead to a significant change in semantics besides just allowing you to refer metavariables bound inside `metavariable-pattern`.

IagoAbal: I imagined this quite differently, that you would add to `r` the metavariables bound by `metavariable-pattern`.

IagoAbal: Oh OK, never mind me, sorry, that is what you did indeed! But the logic is in `Metavariable_pattern`.

src/engine/Match_search_mode.ml
```diff
@@ -523,17 +523,18 @@ let children_explanations_of_xpat (env : env) (xpat : Xpattern.t) : ME.t list =
 
 let rec filter_ranges (env : env) (xs : RM.ranges) (cond : R.metavar_cond) :
```
IagoAbal, line 524: Do we want to allow this?
```yaml
rules:
- id: test
  severity: INFO
  message: Test
  languages: [python]
  patterns:
  - pattern: $A + $B
  - metavariable-pattern:
      metavariable: $B
      patterns:
      - pattern: bar($X)
  - metavariable-pattern:
      metavariable: $X
      pattern: x
```
because this allows that, so the order in which `metavariable-pattern`s are specified will alter the meaning of the rule.
cc @aryx

PR #7053, opened 2023-02-02
feat(js): Allow metavar ellipsis in template strings

nmote:
Previously, a metavar ellipsis in a template string would match only its literal value. This is surprising, since ellipses work in template strings, and metavar ellipses work in regular string literals.

To address this, I updated `m_arguments_concat` to use the `m_list_with_dots_and_metavar_ellipsis` helper function, instead of using a hand-rolled solution to match ellipses.

The change in `Pattern.ml` is so that the optimization step in `Analyze_rule` does not erroneously skip files that don't contain the literal contents of the metavariable ellipsis. Without this change, a manual test of the case added in `tests/rules` fails. However, the automated test runner does not appear to run the optimization step in the same way, and it still passes. I will file an issue to follow up on that.

Test plan: Automated tests

PR checklist:


nmote:
Ticket: https://linear.app/r2c/issue/PA-2491/make-sure-test-runner-is-testing-the-analyze-rule-optimization

emjin (approved):
It's strange to me that we fail on ellipses or metavars but not metavar-ellipses (at least, that's my read of the code---I could be wrong). Since this passes tests though I'm ok with merging this anyway as a fix

emjin (approved):
It's strange to me that we fail on ellipses or metavars but not metavar-ellipses (at least, that's my read of the code---I could be wrong). Since this passes tests though I'm ok with merging this anyway as a fix

nmote:
Regular metavars and regular ellipses already worked, and continue to do so after this change. It's just that regular metavars can only bind to one component of a template string, e.g. a single interpolated expression or a single run of uninterrupted text. See the following tests:

* https://github.com/returntocorp/semgrep/blob/develop/tests/patterns/js/dots_template_literals.sgrep
* https://github.com/returntocorp/semgrep/blob/develop/tests/patterns/js/metavar_template.sgrep
* https://github.com/returntocorp/semgrep/blob/develop/tests/patterns/js/template_string_w_metavar.sgrep

emjin:
I'm not explaining this super well but my read was that the code will do:
```
$X = `...${...}...`

# Matches
matchThis = `foo${bar}baz`;

# Doesn't match
dontMatch = `foobar`
```
but
```
$X = `...${$...INTERPOLATED}...`

# Matches
matchThis = `foo${bar}baz`;

# Matches
dontMatch = `foobar`
```

nmote:
It turns out that the first example matches as expected, and the second example has zero matches (after making some changes so that it's valid JS):

```
// Matches
var matchThis = `foo${bar}baz`;

// Matches
var dontMatch = `foobar`;
```

This behavior is the same before and after this PR. This PR has to do with ellipsis metavariables appearing directly in the template string itself, rather than inside an embedded expression.

emjin:
I'm surprised by that, but that's on me for not reading `m_list_with_dots_and_metavar_ellipsis` more closely. In any case, we should be fine to merge

nmote:
I'm surprised too, but I think it's a separate issue :shrug:

PR #7056, opened 2023-02-02
feat: Allow metavariable-pattern to be more widely used with generic mode

nmote:
As I've explained inline in more detail, we now allow metavariable-pattern to be used on any kind of metavariable binding as long as the target language for the metavariable-pattern is generic. Previously we only allowed metavariable-pattern to be used with a sub language if the metavariable in question was bound to a string literal.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/changelog.d/metavar-pattern-generic.feat
+++ b/changelog.d/metavar-pattern-generic.feat
@@ -1,0 +1,1 @@
+Allow metavariable-pattern clauses that use `language: generic` to run (and potentially match) on any metavariable binding kind, not just strings. For example, with the pattern `foo($...ARGS)`, it is now possible to use a `metavariable-pattern` on `$...ARGS` with `language: generic`, and match using generic mode against whatever text `$...ARGS` is bound to.
--- a/src/engine/Metavariable_pattern.ml
+++ b/src/engine/Metavariable_pattern.ml
@@ -135,8 +135,8 @@ let satisfies_metavar_pattern_condition get_nested_formula_matches env r mvar
               r = { start = 0; end_ = mval_range.end_ - mval_range.start };
             }
           in
-          match (opt_xlang, mval) with
-          | None, __any_mval__ -> (
+          match opt_xlang with
+          | None -> (
               (* We match wrt the same language as the rule.
                * NOTE: A generic pattern nested inside a generic won't work because
                *   generic mode binds metavariables to `MV.Text`, and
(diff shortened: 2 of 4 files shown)


nmote:
Don't merge -- I'll eat the conflicts with https://github.com/returntocorp/semgrep/pull/7052

src/engine/Metavariable_pattern.ml
```diff
@@ -118,53 +118,79 @@ let satisfies_metavar_pattern_condition nested_formula_has_matches env r mvar
...
+                (* Previously we only allowed metavariable-pattern with a
+                 * different language when the metavariable was bound to a
+                 * string. For these cases, we keep the previous behavior, which
+                 * I (nmote) believe makes only the contents of the string
+                 * available for matching in the sub language, and not the
+                 * enclosing quotes.
+                 *
+                 * However, now we also allow non-text metavariable bindings to
+                 * match in generic mode. Because non-text metavariable bindings
+                 * are highly unlikely to be syntactically valid in any
```
IagoAbal, line 216: Do tests break if you allow this for any language and not just `generic` ?

src/engine/Metavariable_pattern.ml
```diff
@@ -118,53 +118,79 @@ let satisfies_metavar_pattern_condition nested_formula_has_matches env r mvar
...
+                    Some content
+                | Xlang.LGeneric, _else_ ->
+                    Some (Range.content_at_range mval_file mval_range)
+                | _else_ -> None
+              in
+              match content with
+              | None ->
+                  (* This is not necessarily an error in the rule, e.g. you may be
+                   * matching `$STRING + ...` and then add a metavariable-pattern on
+                   * `$STRING`. This will only work when `$STRING` binds to some text
```
IagoAbal: ~Need to update this comment?~ Nah it's still valid, it just doesn't apply to `generic` anymore.

IagoAbal (approved):

changelog.d/metavar-pattern-generic.feat
```diff
@@ -0,0 +1 @@
+Allow metavariable-pattern clauses that use `language: generic` to run (and potentially match) on any metavariable binding kind, not just strings.
```
IagoAbal: Isn't this change associated with a customer-related ticket? I would expect the ticker number to be used for the changelog entry instead of `metavar-pattern-generic.feat` and referenced in the PR comment as well.

IagoAbal:
Nice addition!

changelog.d/metavar-pattern-generic.feat
IagoAbal: We could also add more detail in this changelog entry (like an example use case), so that tech-writers can highlight it appropriately in the release notes. cc @adamkvitek

src/engine/Metavariable_pattern.ml
nmote: No. The one test that broke was updated in https://github.com/returntocorp/semgrep-rules/pull/2691. @LewisArdern agreed that the new behavior here makes sense, and updated the rule accordingly.

I wouldn't object strongly to allowing this for any language, but I would want to see a motivating use case.

changelog.d/metavar-pattern-generic.feat
nmote: Well, there are going to be at least 3 PRs associated with that ticket and they can't all use the same changelog file. Doing it this way seems to make the most sense to me.

src/engine/Metavariable_pattern.ml
nmote: I've added a clarification.

changelog.d/metavar-pattern-generic.feat
nmote: I've added an example to the changelog entry.

PR #7062, opened 2023-02-03
Use match based key instead of syntactic content to dedup diffs

emjin:
This is what we expected the cli to do, but apparently it did not. The effect will be that changing `logger.error("log 1")` to `logger.error("log 2")` will not trigger a new finding if the rule is `logger.error(...)`, since the finding was already there.

Test plan: make test

PR checklist:

### DIFF
--- a/cli/tests/e2e/snapshots/test_baseline/test_match_based_key/error.txt
+++ b/cli/tests/e2e/snapshots/test_baseline/test_match_based_key/error.txt
@@ -1,0 +1,6 @@
+Scanning 1 file.
+
+Some files were skipped or only partially analyzed.
+  Scan was limited to files tracked by git.
+
+Ran 1 rule on 1 file: 1 finding.
--- a/cli/tests/e2e/test_baseline.py
+++ b/cli/tests/e2e/test_baseline.py
@@ -3,6 +3,7 @@ import tempfile
 from collections import defaultdict
 from itertools import permutations
 from pathlib import Path
+from typing import List
 from typing import Optional
 
 import pytest
@@ -88,22 +89,16 @@ def _git_merge(ref: str) -> str:
     ).strip()
 
 
-def run_sentinel_scan(check: bool = True, base_commit: Optional[str] = None):
+def run_semgrep_scan(
+    args: List[str], check: bool = True, base_commit: Optional[str] = None
+):
     env = {"LANG": "en_US.UTF-8"}
     env["SEMGREP_USER_AGENT_APPEND"] = "testing"
     unique_settings_file = tempfile.NamedTemporaryFile().name
     Path(unique_settings_file).write_text("has_shown_metrics_notification: true")
     env["SEMGREP_SETTINGS_FILE"] = unique_settings_file
 
-    cmd = SEMGREP_BASE_COMMAND + [
-        "--disable-version-check",
-        "--metrics",
-        "off",
-        "-e",
-        f"$X = {SENTINEL_1}",
-        "-l",
-        "python",
-    ]
+    cmd = SEMGREP_BASE_COMMAND + args
     if base_commit:
         cmd.extend(["--baseline-commit", base_commit])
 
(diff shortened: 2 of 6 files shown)


aryx:
Can we add e2e tests for this? test_match_based_id.py has a few useful tests for those kinds of things I think.

emjin:
> Can we add e2e tests for this? test_match_based_id.py has a few useful tests for those kinds of things I think.

Yeah I'm doing another commit

aryx:
Also with a test we can make sure that osemgrep must also support this new feature.

emjin:
@chmccreery Let me know if there's any documentation to be updated.

cli/tests/e2e/snapshots/test_baseline/test_match_based_key/error.txt
```diff
@@ -0,0 +1,6 @@
+Scanning 1 file.
+
+Some files were skipped or only partially analyzed.
```
chmccreery, line 3: Is this line expected on a non-baseline scan?

cli/tests/e2e/test_baseline.py
```diff
@@ -621,6 +642,40 @@ def test_renamed_file(git_tmp_path, snapshot, new_name):
...
+    # Add head finding
+    foo.write_text(f"logger.log('changed text')\n")  # Change just the content of foo
+    subprocess.run(["git", "add", "."], check=True, capture_output=True)
+    _git_commit(2)
+
+    # Non-baseline scan should report 1 finding
+    output = run_match_based_scan()
+    assert output.stdout != ""
+    snapshot.assert_match(
+        output.stderr.replace(base_commit, "baseline-commit"), "error.txt"
```
chmccreery, line 667: Sort of a side comment, but why do we use stderr? The scan is succeeding, no?

chmccreery (commented):
Thanks @emjin! No documentation that needs updating. However, I've been thinking about this and I have 2 concerns:

1. Please double check me that this doesn't change the match_based_id that we used to calculate at all? Logically, I don't think it should, but if that happened it would be very bad.
2. There is one edge case that I am concerned about, and I'm trying to think through a mitigation strategy:

Assume a rule has 2 findings on the same file. Let's say the rule is logger.$METHOD(...) and the following lines contained the following code in the baseline commit:

1. logger.info("hello, world")
3. logger.info("hello, again")

The user adds logger.info("new message") before the other two findings, so now the following lines contain the following code in the latest commit:

1. logger.info("new message")
2. logger.info("hello, world")
4. logger.info("hello, again")

All of these findings have the same rule_id, path, and formula, so the only differentiator will be the index. In the baseline commit, the index of the "hello, world" finding is 0 and the index of the "hello, again" finding is 1. In the head commit, the index of the NEW finding is 0, the index of the "hello, world" finding is 1 and the index of the "hello, again" finding is 2. This means that the finding on line 3 is going to get reported rather than the finding on line 1 when we subtract them out.

To be clear, this is already a problem with syntactic_id, but I think with the change to using match_based_id we will see this behavior happen more frequently, since the match_based_id contains a lot less user-written-code than syntactic_id and is therefore more likely to have repeats within the same file. I'm not sure there's a way around this, just interested in if you have any ideas. 

For example, I know the CLI has access to the code and the git history. Maybe we can use the information about the changed lines/git diff to also subtract out findings whose line numbers only span unchanged lines of code from both of the scans?

cli/tests/e2e/test_baseline.py
emjin: That's a good question. My understanding is we direct non-result output to stderr so that it makes it easier to filter. I'm not certain, though

cli/tests/e2e/snapshots/test_baseline/test_match_based_key/error.txt
emjin: Yeah, this just means that there were files semgrep couldn't scan completely (partial parse error) or skipped for various reasons

emjin:
> Please double check me that this doesn't change the match_based_id

I also don't think so, but I'm not certain. Let's do a comparison of match_based_id in an example output. I'll also look into how to add a test (or confirm that one already exists), so that someone in the future doesn't do a quick refactor and accidentally break this

> There is one edge case that I am concerned about, and I'm trying to think through a mitigation strategy

That's absolutely a concern with match based key. I believe when people initially settled on this key, there was the agreement that the index, file, rule id, and rule content would be sufficient. (In fact, I have the faint idea that at one point we considered just the first three.) There might be a doc floating around with reasoning provided so that we don't need to redo all that thinking. I'll look into this more on Monday

emjin:
After talking to @chmccreery, we're going to try to use the diffs to mitigate the index problem. To explain the problem, let's say develop has two findings with the same match based id, on lines 20 and 30. Now, let's say that in our diff, we add a function in between lines 20 and 30, moving the finding from line 30 to line 100. In that function, we introduce a new finding on line 40. When we surface that finding, we want to report the one on line 40. The plan is to achieve that by considering which finding could have been introduced by the diff, using a combination of the match location and taint traces.

emjin:
Closing this because we're too concerned about the index problem. Will design another solution

PR #7066, opened 2023-02-03
fix: Correct range for mvars bound in `metavariable-pattern`

nmote:
Followup to #7052. It hasn't been released yet, so I'm not adding anything to the changelog here.

Specifically in the case where the `metavariable-pattern` uses a different language, the metavariable binding location was not correctly calculated.

I just pulled out the `revert_loc` and `fix_loc` functions into an outer scope, so that `revert_loc` is also available in the case where the `metavariable-pattern` is evaluated using a different language.

I added some TODOs inline because I'm suspicious of the way that the location are adjusted. However, since `fix_loc` predates #7502 and `revert_loc` is simply the inverse, I don't want to mess with it right now.

Test plan: Automated tests. Without this change, the added test has no matches, since the `focus-metavariable` range doesn't intersect with the original match's range.

PR checklist:

### DIFF
--- a/src/engine/Metavariable_pattern.ml
+++ b/src/engine/Metavariable_pattern.ml
@@ -135,6 +135,37 @@ let satisfies_metavar_pattern_condition get_nested_formula_matches env r mvar
               r = { start = 0; end_ = mval_range.end_ - mval_range.start };
             }
           in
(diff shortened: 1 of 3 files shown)


brandonspark (approved):
Let's Get This Merged!

PR #7067, opened 2023-02-04
chore: defer persistent bindings from metavariable-pattern until after filter

brandonspark:
## What:
This PR changes it such that persistent metavariable bindings from `metavariable-pattern` are not applied until right before `focus-metavariable`.

## Why:
In #7052, we introduced persistent metavariable bindings that are introduced by the filtering step in `metavariable-pattern`. Unfortunately, they were done in such a way that they could be visible from other `metavariable-pattern`s, making them dependent on the order of the filtering. 

## How:
I just floated up the bindings introduced and collected them through each filtering step, until I combined them right before focusing-time.

## Test plan:
`make test`
and `make test` on `proprietary` (I remembered this time, @nmote!)

I declined a `changelog` entry because the previous change hasn't been released yet, and this should be how it worked in the first place anyways.

PR checklist:

### DIFF
--- a/src/engine/Match_search_mode.ml
+++ b/src/engine/Match_search_mode.ml
@@ -521,20 +521,24 @@ let children_explanations_of_xpat (env : env) (xpat : Xpattern.t) : ME.t list =
 (* Metavariable condition evaluation *)
 (*****************************************************************************)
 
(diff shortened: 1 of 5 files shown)


nmote (approved):
Cool, seems reasonable

PR #7068, opened 2023-02-04
feat!: separate rules by engine

brandonspark:
## What:
This PR makes it so that when an option is specified to run an engine that is not OSS, and there are rules that are run with the OSS engine, we report which rules.

## Why:
Currently, we require that rules be annotated with `deepsemgrep: true`, or else they will not be run with the Pro engine, even when `--interfile` is specified.

This is highly confusing to users, and can result in people getting confused over not having findings, even when `--interfile` is specified. We should be more transparent to users, so they don't get confused (including myself).

I also put this information into the CLI output, because I think it will be useful to consume on the App side.

## How:
I pass around the engine requested, and I have the core output include the engine that was used for each rule, since the decision of what engine to run with which rule is done on the OCaml side.

## Test plan:
With the rule file
```
rules:
- id: test
  mode: taint
  pattern-sources:
    - pattern: |
        150 
  pattern-sinks:
    - pattern: |
        sink(...)
  message: stop
  languages: [java]
  severity: WARNING
  metadata:
    deepsemgrep: true
- id: test2
  pattern: |
    150
  message: stop
  languages: [java]
  severity: WARNING
```
and target file
```

public class Foo {
  public void main () {
    int x = 150;
    
    sink(x);
  }
}
```

without `--interfile` specified, we get
image

but with it, we get:
imageCloses PA-2442

and with verbose:
image

PR checklist:


aryx:
Where is the PR on changes on semgrep-interfaces?

brandonspark:
@aryx sorry, forgot to put it up: https://github.com/returntocorp/semgrep-interfaces/pull/59

cli/src/semgrep/formatter/text.py
```diff
@@ -736,10 +736,28 @@ def generate_output(header: str, matches: List[RuleMatch]) -> None:
...
+        ]
+
+        rules_output = []
+        if cli_output_extra.engine_requested is None or not isinstance(
+            cli_output_extra.engine_requested.value, out.OSS
+        ):
+            if oss_rules:
+                rules_output = [
+                    "\nSome rules were run as OSS rules because `deepsemgrep: true` was not specified.\n"
+                ] + ["These rules were:\n   " + "   \n   ".join(oss_rules)]
```
emjin: I'm concerned because this could be a lot of rules. Maybe we could output the number and only show which rules with `--verbose`?

cli/src/semgrep/formatter/text.py
```diff
@@ -736,10 +736,28 @@ def generate_output(header: str, matches: List[RuleMatch]) -> None:
...
+        rules_by_engine = cli_output_extra.rules if cli_output_extra.rules else []
+
+        oss_rules = [
+            with_color(Colors.foreground, rule.rule_id.value, bold=True)
+            for rule in rules_by_engine
+            if isinstance(rule.engine_kind.value, out.OSS)
+        ]
+
+        rules_output = []
+        if cli_output_extra.engine_requested is None or not isinstance(
```
emjin: Very minor nit: I think I would prefer an `and` here over a nested if but it definitely doesn't matter

cli/src/semgrep/constants.py
```diff
@@ -40,6 +42,12 @@ class EngineType(Enum):
     def is_pro(self) -> bool:
         return self.value >= EngineType.PRO.value
 
+    def to_engine_kind(self) -> out.EngineKind:
```
emjin, line 45: I wonder how your work interacts with stuff like:
```
        if any(match.severity == RuleSeverity.EXPERIMENT for match in new_ignored):
            logger.info("Some experimental rules were run during execution.")
```
(scans.py: 318)

emjin (commented):
The only important comment is the one about the output. I'm wondering if printing out all the rules will be too much in some cases.

cli/src/semgrep/constants.py
brandonspark: this is just the engine kind convertor, surely that shouldn't affect the severity, right?

aryx:
Who asked for this feature?

brandonspark:
> Who asked for this feature?

@aryx I'm not sure anyone explicitly asked for this, but I in particular really want this, and this should exist either way before the launch. Without this, I think customers might be super frustrated if something doesn't work, and it turns out that Semgrep didn't warn them that it was running their rules as OSS.

aryx:
ok, got it.
Maybe add milan in the loop?

cli/src/semgrep/constants.py
emjin: Oh, I more meant on an architectural level, how this affects the readability

brandonspark:
@emjin I added new screenshots in the description

emjin (approved):
Yay let's get this in!

PR #7071, opened 2023-02-04
fix(metrics): send metrics after configs are resolved

emjin:
We determine whether or not to send metrics based on the config type, so we should wait for that determination before trying to send anything.

Test plan: after the release there should be no more places where the engine type is empty

Shouldn't need any privacy policy updates because this is just doing what we claimed to do

PR checklist:


brandonspark (approved):
stamp

PR #7080, opened 2023-02-06
fix(pro): fix the timeout set in CI

emjin:
Test plan: run the docker image on one of the repos that currently times out.

PR checklist:


emjin:
TODO: need to document all of these resource limits

emjin:
Tested on one of the repos that previously times out, it works!

nmote:
Wait, so reducing the timeout makes something that previously timed out no longer time out? Am I reading that right?

emjin:
Previously one of the semgrep pro runs would hit the github actions timeout and not return any results. Now, it hits our timeout first and gives up on interfile analysis, instead running a normal analysis and reporting back results.

nmote (approved):

PR #7081, opened 2023-02-06
chore: handle unparsable URLs

brandonspark:
## What:
This PR makes it so that if we fail to parse an SSTP URL, instead of passing the mangled, incorrect URL to the App, we will instead send the original sstp URL.

## Why:
We have this thread https://returntocorp.slack.com/archives/C01NXGX2EHZ/p1674767808991039 in the slack where we were getting URLs that did not make sense. This was because the URLs that were read in from the environment variables were not parsable by our current implementation of `get_url_from_sstp_url`, and we were outputting a URL that contained `None` in it.

One proposed solution (discussed with @sabrinabrogren) was to port this URL-parsing logic to the App altogether, and just transmit the base URL. Unfortunately, there's several places in the code (such as here: https://github.com/returntocorp/semgrep/blob/22a0bd5b8aaae5a41a367902a9c5a2033db0c7c8/cli/src/semgrep/meta.py#L711) which rely on multiple environment variables, so I find this approach untenable, as the App would need to have access to all of them.

I believe that the next best option is just to avoid mangling the URL, if possible. That's what this PR does.

## How:
If we fail to parse any of the essential fields that are returned by the URL,

PR checklist:


sabrinabrogren:
Ah, this looks like a reasonable approach. Nice! Would you mind adding a test?

cli/src/semgrep/meta.py
```diff
@@ -40,6 +40,10 @@ def get_url_from_sstp_url(sstp_url: Optional[str]) -> Optional[str]:
         # let's just pick https
         protocol = "https"
 
+    # if we know that this URL is going to be unclickable, return the original URL
+    if None in [protocol, result.resource, result.owner, result.name]:
+        return sstp_url
```
aryx, line 50: Can someone add a comment somewhere explaining what is an SSTP URL?
I never heard about SSTP.

brandonspark: i don't really think i can write a comment about it, i've never heard of it either

all i know is that returning a URL which contains `None` in it is bad behavior

brandonspark:
@sabrinabrogren I added a test, if you look at the `meta.json` file, you can see that the `repo_url` is correctly the original URL, and I've confirmed that without this change, it gets mangled.

cli/src/semgrep/meta.py
sabrinabrogren: ah yeah, perhaps we could add [this link](https://stackoverflow.com/questions/31801271/what-are-the-supported-git-url-formats) which talks about all the different git url formats.

sabrinabrogren (approved):

aryx:
Nice link. Still don't know what is an sstp_url though :(

PR #7087, opened 2023-02-07
refactor(cli): select binary path at CoreRunner creation time

brandonspark:
## What:
This PR makes it so that the binary that we use (`semgrep-core` or `semgrep-core-proprietary`) is not selected when we use a `CoreRunner` object, but when we create the `CoreRunner` object in the first place.

## Why:
This is aesthetically pleasing, and also will fix the problem in PA-2484. The logic for selecting this path shouldn't be localized to specifically when we decide to run Semgrep on some rules, the CoreRunner object should just correspond to a particular executable, "statically".

## How:
The `__init__` method on the `CoreRunner` class now contains the logic for dealing with the `pro_path` and whatnot, which used to be in `run_rules_to_semgrep_core_helper`. Some simplifications in logic were able to be done in a few other places, and in particular, `--validate` can just use that pre-selected path.

## Test plan:
For `--dump-command-for-core`:
image

For running Semgrep:
image

For validating an Apex rule:
image

For when using `--pro` with a `deep-semgrep` binary installed:
image

For when using `--pro` without either binary installed:
image

Closes PA-2484

PR checklist:

### DIFF
--- a/cli/src/semgrep/core_runner.py
+++ b/cli/src/semgrep/core_runner.py
@@ -555,6 +555,28 @@ class CoreRunner:
         self._optimizations = optimizations
         self._core_opts = shlex.split(core_opts_str) if core_opts_str else []
 
+        self._binary_path = SemgrepCore.path()
+
+        if engine.is_pro:
+            deep_path = (
+                SemgrepCore.deep_path()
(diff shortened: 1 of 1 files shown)


mjambon (approved):

aryx (approved):
nice

PR #7093, opened 2023-02-07
Minor improvements to CI config flows

jbergler:
- remove support for legacy semgrep-app installations which is no longer required
- add additional debugging for `semgrep ci --debug`

### DIFF
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -125,49 +125,6 @@ class ScanHandler:
                 f"API server at {state.env.semgrep_url} returned type '{type(response.json())}'. Expected a dictionary."
             )
 
-    def fetch_config_and_start_scan_old(self, meta: Dict[str, Any]) -> None:
-        """
-        Get configurations for scan and create scan object
-        Backwards-compatible with versions of semgrep-app from before Aug 2022
-        TODO: Delete once all old, on-prem instances have been deprecated
-        """
-        state = get_state()
-
-        logger.debug("Getting deployment information")
-        get_deployment_url = f"{state.env.semgrep_url}/api/agent/deployments/current"
-        body = self._get_scan_config_from_app(get_deployment_url)
-        self._deployment_id = body["deployment"]["id"]
-        self._deployment_name = body["deployment"]["name"]
-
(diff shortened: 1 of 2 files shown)


cli/src/semgrep/app/scans.py
```diff
@@ -1,7 +1,9 @@
 # Handle communication of findings / errors to semgrep.app
 import json
+import logging
```
brendongo:
```suggestion
from semgrep.verbose_logging import getLogger

logger = getLogger(__name__)
```
is recommended for cli use. It adds a log level for `--verbose`. (we dont use verbose logging in this file but might as well standardize on the logger used in all other files

cli/src/semgrep/app/scans.py
```diff
@@ -197,6 +156,14 @@ def fetch_and_init_scan_config(self, meta: Dict[str, Any]) -> None:
         self._skipped_match_based_ids = body.get("triage_ignored_match_based_ids") or []
         self.ignore_patterns = body.get("ignored_files") or []
 
+        if logger.isEnabledFor(logging.DEBUG):
+            config = deepcopy(body)
+            try:
+                config["rule_config"] = json.loads(config["rule_config"])
```
brendongo, line 162: nit: is `rule_config` always guaranteed to exist in config? Should we use more defensive code?

brendongo (approved):

cli/src/semgrep/app/scans.py
jbergler: I'm importing `logging` here for logging.DEBUG since it's not in `verbose_logging`
`getLogger` is imported and used as you suggested after the imports.

cli/src/semgrep/app/scans.py
jbergler: I don't know when/how it wouldn't be present, but it's a good point that the existing error handling doesn't catch `json.loads(None)` so I can change this to `catch Exception` easily enough to handle all the failures.

cli/src/semgrep/app/scans.py
jbergler: Cleaned up the import to be more specific

PR #7095, opened 2023-02-08
feat(ci): allow users to run an individual repo in CI with a different engine

emjin:
The CLI allows users to choose between running with just pro languages, with intrafile interprocedural analysis, and with interfile analysis. We offer this to CI users too in case they wish to make an exception for a single repo. In general, they should use the toggle to control this.

Test plan: run in CI with docker image

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -196,6 +196,9 @@ def ci(
     optimizations: str,
     dataflow_traces: bool,
     output: Optional[str],
+    pro_languages: bool,
+    pro_intrafile: bool,
+    pro: bool,
     sarif: bool,
     quiet: bool,
     rewrite_rule_ids: bool,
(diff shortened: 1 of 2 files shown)


cli/src/semgrep/commands/ci.py
```diff
@@ -323,9 +326,18 @@ def ci(
             is_full_scan = metadata.merge_base_ref is None
             engine = EngineType.OSS
             if scan_handler and scan_handler.deepsemgrep:
-                engine = (
-                    EngineType.PRO_INTERFILE if is_full_scan else EngineType.PRO_LANG
-                )
+                if is_full_scan:
```
IagoAbal: I imagine some may want to enable `--pro-intrafile` for diff scans too?

cli/src/semgrep/commands/ci.py
```diff
@@ -323,9 +326,18 @@ def ci(
...
+                    if pro_intrafile:
```
IagoAbal: A bit concerned that the flags don't mean the same as for `scan`, e.g. `--pro-intrafile --pro` in `scan` would be the same as `--pro` but here it would be the same as just `--pro-intrafile`.

IagoAbal (approved):

IagoAbal:
Perhaps @nmote's proposal of just adding `--no-pro` would do the job and we would still have time to think more about `--pro` and co later on? Instead of enabling one by one, you'd have to disable it one by one.

cli/src/semgrep/commands/ci.py
emjin: Yeah, I'm concerned about this too. Personally, I would like them to just be equivalent and override the toggle, but I don't have strong feelings.

IagoAbal: I think it's better to override the toggle too.

PR #7096, opened 2023-02-08
fix(pro): allow interprocedural and proprietary language to be pro findings

brandonspark:
## What:
This PR makes it such that running the Pro engine with the intraprocedural engine, or running the Pro engine to analyze a proprietary language, correctly mark produced findings as from "Semgrep Pro Engine".

## Why:
After merging https://github.com/returntocorp/semgrep/pull/6928, we started labelling the matches that resulted from the main code in `Run.ml` in `pro`. However, this is not the only entry point to pro features. We might also run the vanilla engine, instrumented with an extra taint hook, or we may run the vanilla engine, instrumented with proprietary parsers. These were incorrectly being classified as regular "Findings", as opposed to "Semgrep Pro Engine Findings". This PR fixes that.

## How:
I added a `pro` field to the `Runner_config`, and a helper function `is_proprietary` for `Lang`, in the same spirit as `is_js`.
Notably, this means that _all_ findings that result from `semgrep --interproc` will classify as Pro findings, even if they could have been found with the vanilla engine.

## Alternatives considered:

I considered making it such that we only report findings which actually only could have occurred with Pro as Pro findings, but:
1) currently we treat interfile findings the same way, we report them regardless of whether they were truly found as interfile
2) due to interfile constant propagation, this is actually quite difficult to do. even in just the taint finding case, we would have to instrument the taint with the fact that it occurred interfile or something.
So I decided this was not worth it with the time that we have.

I also considered placing the logic for turning matches into pro matches based on proprietary language somewhere else. 
1) I couldn't really place it earlier, because I didn't want to have to duplicate it for each `Match_search_mode`, `Match_tainting_mode`, ... 
2) I couldn't really place it later, because we need access to the language (and thus the xtarget). This is pretty much where xtargets disappear.
3) I couldn't place it in `pro` itself, because `pro` just calls the vanilla engine, and lets that do all the work. I can't intercept the matches from within `pro`.

I considered placing the logic for turning matches from intraprocedural engine somewhere else too, but for similar reasons, I decided to just group it with the logic for proprietary languages. This would mean they wouldn't be distributed. I could have just checked if there was a hook set, but I didn't really like that as much.

I considered not having a flag in the config, but I need some way to signal to the vanilla engine that it was invoked with the `pro` engine. I judged this as relatively clean.

## Test plan:
`make test`, and:

Running on a proprietary language:
image

Running on a non-proprietary language, with and without `--pro`:
image

Running with `--pro-intrafile`:
image

Running with `--interfile`, with a rule that has `interfile: true` and a rule that doesn't (this is the current behavior):
image

PR checklist:


src/core/Xlang.ml
```diff
@@ -76,3 +76,7 @@ let to_string = function
   | L (l, _) -> Lang.to_string l
   | LRegex -> "regex"
   | LGeneric -> "generic"
+
+let is_proprietary = function
+  | L (lang, _) -> Lang.is_proprietary lang
+  | __else__ -> false
```
IagoAbal: Better `| LRegex | LGeneric -> false` here.

IagoAbal:
Does this depend on another PR to `Semgrep-proprietary` that sets this `pro` config to `true`?

IagoAbal:
Also rebase against `develop`, in a PR yesterday I deprecated `--interfile` (now `--pro`) and `--interproc` (now `--pro-intrafile`). 🤷‍♂️ And also I had provisionally changed the text to `Inter-File Analysis Findings:` so you'll have to change it back to `Semgrep Pro Engine Findings:`.

src/runner/Run_semgrep.ml
```diff
@@ -674,9 +675,23 @@ let semgrep_with_rules config ((rules, invalid_rules), rules_parse_time) =
...
+             let matches =
+               Match_rules.check ~match_hook ~timeout:config.timeout
+                 ~timeout_threshold:config.timeout_threshold xconf rules xtarget
+             in
+             (* If our target is a proprietary language, or we've been using the proprietary
+              * engine, then label all the resulting matches with the Pro engine kind.
+              * This can't really be done any later, because we need the language that
+              * we're running on.
+              *)
+             if config.pro || Xlang.is_proprietary xtarget.xlang then
```
IagoAbal: Feels wrong to have an `engine_kind` in `Pattern_match.t` and then use it this way.

IagoAbal: I don't have a strong opinion about having to change this now, but long-term I think we either change how `engine_kind` is decided, or it shouldn't be part of `Pattern_match.t`.

brandonspark:
> Does this depend on another PR to `Semgrep-proprietary` that sets this `pro` config to `true`?

Yes, this one: https://github.com/returntocorp/semgrep-proprietary/pull/520

src/runner/Runner_config.ml
```diff
@@ -48,6 +48,7 @@ type t = {
   (* Main flags *)
   pattern_string : string;
   pattern_file : filename;
+  pro : bool;
```
IagoAbal: What if we avoided adding `pro` here and we just checked that the hooks are setup? Really `pro` here can be thought as a synonym of `Option.is_some !(Match_tainting_mode.hook_setup_hook_function_taint_signature) || Option.is_some !(Dataflow_tainting.hook_function_taint_signature)`.

brandonspark: this is true. i wanted to avoid it because i thought that this line would look more "clean", and also because i was afraid of a future where we have some pro features that don't involve the hooks, but it's probably a safe bet, so I don't mind changing it

src/runner/Run_semgrep.ml
brandonspark: Yeah, I agree, but for data-dependency and type-related reasons, I found this really difficult to do when originally setting it up in https://github.com/returntocorp/semgrep/pull/6928

I think this will make more sense once we implement tracking whether a match resulted specifically from Pro (as in, it couldn't be found with OSS)

IagoAbal (approved):

Issue #6287, opened 2022-10-12
[Go] Backtick string literals not captured correctly


dstrelau:
**Describe the bug**
Go's backtick string literals are not captured properly (or autofix doesn't output them properly?)

**To Reproduce**

https://semgrep.dev/s/1vrv

FWIW, I don't believe this to be specific to JSON content — any content inside backticks fails in this manner.

**Expected behavior**

The autofix is missing the closing backtick and final character. I expect it to be included.

```
assert.Equal(t,`{"foo:"bar", str)
```

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [X] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Use case**

Large-scale automated refactoring of legacy test code that uses backticks a lot. I can't reasonably make these changes by hand, but if autofix worked it'd be ✨ amazing.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1987/[go]-backtick-string-literals-not-captured-correctly. _Note_: this link is for r2c use only and is not accessible publicly.

PR #7108, opened 2023-02-08
fix(go): fix range of raw string literals

brandonspark:
## What:
This PR fixes the tokens for raw string literals in Go, to be faithful to the range of the string in the source text.

## Why:
Currently, raw string literals in Go have tokens which capture just the contents of the string, and not the delimiters. This causes issues with metavariable range tracking, because a metavariable matching a raw string literal will start at the delimiter, but only go for the range of the contents of the delimited string, meaning that it will capture two characters less than the entire source-text string. This messes with Autofix and message interpolation in the Playground.

## How:
I made it such that the token produced from the tree-sitter parser instead now has the entire original source-text string inside of it, which will give it the proper length.

## Test plan:
For a file 
```
"hello"
`hello`
```
before we had:
image

now we have 
image

Note that in the second example of the string, the token is now surrounded by backticks, which accurately reflects the range that it corresponds to.

For a file
```
x = `hello`
```
after this change, we get:
image

and before we had:
image

Note how before, the `message` was cut off, and now it properly has the entire statement.

Closes PA-2206

PR checklist:


nmote:
Does this fix #6287?

brandonspark:
@nmote Yes!
image

nmote:
Cool! Could you use that as a test case?

nmote:
(or something similar)

brandonspark:
@nmote ah yeah, I was trying to think of how to test this, and couldn't come up with a good way. I didn't think of autofix. Will do.

nmote (approved):
Nice! When I suggested testing using autofix, I forgot that since Go doesn't have AST-based autofix, the simpler autofix testing that we have for semgrep-core won't work. You've added an e2e test which is fine, but they can also be a bit annoying to maintain. I'm content with the manual test plan. I'll let you decide whether you want to include the test or not.

brandonspark:
> Nice! When I suggested testing using autofix, I forgot that since Go doesn't have AST-based autofix, the simpler autofix testing that we have for semgrep-core won't work. You've added an e2e test which is fine, but they can also be a bit annoying to maintain. I'm content with the manual test plan. I'll let you decide whether you want to include the test or not.

OK, I think I'll take it out then, since it doesn't really fit precedence with the other tests, and I don't really foresee this changing.

PR #7109, opened 2023-02-08
chore: allow matching inside of `Cast`

brandonspark:
## What:
This PR allows `e1` to try to match to `e2` when matching `Cast(_, _, e2)`.

## Why:
If we are trying to find an expression, it's a fair bet that, in langauges that allow dynamic type casts, we should match the thing that is being type-cast. It's still a valid instance of the expression.

## How:
This is tricky because what you want to do is say that matching a `Cast` allows you to match inside, but also let it match regularly (the other cases). If we just naively write that, we will loop forever, since we will continuously re-enter the `Cast` case. We need to somehow remember that we already unpacked the inner logic, and try the other cases in the next recursive call.

This PR adds in an `is_first_invocation` flag, which is normally always `true`. Any recursive case which should occur exactly once, but might not deconstruct on `a` or `b`, and doesn't want to loop, should set `is_first_invocation` to `false`, and then require `is_first_invocation` on the relevant call. This effectively makes such cases "invisible" after the first pass.

## Test plan:
`make test`, and:
image

## Alternatives considered:

I considered extracting `m_expr` to an inner helper, and using an outer `m_expr` which just does the `Cast` case, and then defers to the inner helper (which will call only the inner helper in the `Cast` case). I thought that would induce more bloat into `Generic_vs_generic` (which `m_expr_root` already does), and this is less invasive of a change, which is still only relevant to the clauses that mention `is_first_invocation`.

Closes PA-1133

PR checklist:

### DIFF
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -760,7 +760,6 @@ and m_expr ?(is_root = false) ?(is_first_invocation = true) a b =
   *)
   | _, G.Cast (_, _, b1) when is_first_invocation ->
       m_expr a b1 >||> m_expr ~is_first_invocation:false a b
-  (* I want to make sure this second call does not re-enter the Cast case... *)
   (* equivalence: name resolving! *)
   (* todo: it would be nice to factorize the aliasing code by just calling
    * m_name, but below we use make_dotted, which is different from what
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -739,7 +739,12 @@ and m_expr_root a b = m_expr ~is_root:true a b
 (* coupling: if you add special sgrep hooks here, you should probably
  * also add them in m_pattern
  *)
-and m_expr ?(is_root = false) a b =
+(* `is_first_invocation` should be false if this is not the first time
+   `m_expr` has recursively tried to match the same `a` and `b`.
+   This allows some cases to be "try-once", and fall-through to the others in
+   all the other cases.
+*)
+and m_expr ?(is_root = false) ?(is_first_invocation = true) a b =
   Trace_matching.(if on then print_expr_pair a b);
   match (a.G.e, b.G.e) with
   (* the order of the matches matters! take care! *)
@@ -748,6 +753,14 @@ and m_expr ?(is_root = false) a b =
   | _, G.Alias (_alias, b1) -> m_expr a b1
   (* equivalence: user-defined equivalence! *)
   | G.DisjExpr (a1, a2), _b -> m_expr a1 b >||> m_expr a2 b
+  (* This case should only run exactly once.
+     This is so we do not endlessly loop trying to match to the same two things.
+     By setting `is_first_invocation` to false, we ensure that we fall-through to
+     the cases that do decompose on `a` or `b`.
+  *)
+  | _, G.Cast (_, _, b1) when is_first_invocation ->
+      m_expr a b1 >||> m_expr ~is_first_invocation:false a b
+  (* I want to make sure this second call does not re-enter the Cast case... *)
   (* equivalence: name resolving! *)
   (* todo: it would be nice to factorize the aliasing code by just calling
    * m_name, but below we use make_dotted, which is different from what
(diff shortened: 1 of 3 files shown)


mjambon (approved):

src/matching/Generic_vs_generic.ml
```diff
@@ -739,7 +739,12 @@ and m_expr_root a b = m_expr ~is_root:true a b
 (* coupling: if you add special sgrep hooks here, you should probably
  * also add them in m_pattern
  *)
-and m_expr ?(is_root = false) a b =
+(* `is_first_invocation` should be false if this is not the first time
+   `m_expr` has recursively tried to match the same `a` and `b`.
```
aryx: put in uppercase 'same'.
Maybe is_first_invocation is not a great label name.

src/matching/Generic_vs_generic.ml
```diff
@@ -748,6 +753,13 @@ and m_expr ?(is_root = false) a b =
   | _, G.Alias (_alias, b1) -> m_expr a b1
   (* equivalence: user-defined equivalence! *)
   | G.DisjExpr (a1, a2), _b -> m_expr a1 b >||> m_expr a2 b
+  (* This case should only run exactly once.
+     This is so we do not endlessly loop trying to match to the same two things.
+     By setting `arguments_have_changed` to false, we ensure that we fall-through to
+     the cases that do decompose on `a` or `b`.
+  *)
+  | _, G.Cast (_, _, b1) when arguments_have_changed ->
+      m_expr a b1 >||> m_expr ~arguments_have_changed:false a b
```
IagoAbal, line 762: Wouldn't something like this work too?
```ocaml
  | a, B.Cast (b1, bt, b2) ->
      (match a with
      | G.Cast (a1, at, a2) ->
          m_type_ a1 b1 >>= fun () ->
          m_tok at bt >>= fun () -> m_expr a2 b2
      | a -> fail ())
      >||> m_expr a b1
```
It would be cleaner than this solution IMO.

aryx: @brandonspark can you simpify the code like @IagoAbal asked? seems far simpler and more maintainable than having
yet another set of label arguments.

brandonspark: this wouldn't work if the choice of `a` and `b` were to fall through to any of the intermediary cases between this one and the `Cast _, Cast _` case below, right?

for instance, we might have the 
```
  | G.N (G.Id ((str, tok), _id_info)), _b when MV.is_metavar_name str ->
      envf (str, tok) (MV.E b)
 ```
 case that we would skip.
 
 concretely, if our input were to be `G.N _, G.Cast (b1, _, b2)`, where the `N` is enclosing an mvar, then the behavior on this proposed change would be to bind the metavariable to `b2`, instead of `Cast`.

Comment (IagoAbal) [reply src/matching/Generic_vs_generic.ml:762]: It has been 5 days but I believe the idea was to replace the case
```ocaml
  | G.Cast (a1, at, a2), B.Cast (b1, bt, b2) -> ...
```
with something like what I proposed above. So you would try the `G.N (G.Id ((str, tok), _id_info)), _b when MV.is_metavar_name str` case before you tried the `a, B.Cast (b1, bt, b2)` case.

brandonspark: im not necessarily sure i understand how you mean, i thought you were proposing to replace the new
```
| _, G.Cast (_, _, b1) when arguments_have_changed ->
      m_expr a b1 >||> m_expr ~arguments_have_changed:false a b
```
case (the third case) with 
```
  | a, B.Cast (b1, bt, b2) ->
      (match a with
      | G.Cast (a1, at, a2) ->
          m_type_ a1 b1 >>= fun () ->
          m_tok at bt >>= fun () -> m_expr a2 b2
      | a -> fail ())
      >||> m_expr a b1
```

if you meant to replace the `Cast _, Cast _` case with the thing above, then i think i misunderstood you. i dont see an issue with that i think, im fine with this.

IagoAbal: > if you meant to replace the Cast _, Cast _ case with the thing above, then i think i misunderstood you. i dont see an issue with that i think, im fine with this.

Yep I believe that's what I meant, I think it should work (but could be wrong) and if so it makes it easier to follow the logic.

PR #7113, opened 2023-02-09
Remove tqdm dependency

underyx:
new progress bars:

https://user-images.githubusercontent.com/1060436/217682587-117e528a-3605-4b46-a601-5b9232e3d481.mp4

PR checklist:


aryx:
Maybe it would be nice to change our testing code so that we don't have to regenerate snapshots each time for those small estethic change. I know sometimes we get rid of semgrep version in the output, maybe we could do something for most of those unit tests and have just a few tests where we really check that the output or error is precisely something regarding the layout.

aryx:
This will help also for osemgrep so that we don't have exactly to have the same layout than the python version.
Of course there will be a few tests where we really look for the exact layout, and those osemgrep would fail, but in many of
those tests we don't really care about the layout, we just care about certain information to be in the output or stderr,
and that's what the snapshot content should focus on.

brendongo (approved):

PR #7114, opened 2023-02-09
[RFC]: allow regex capture group metavariables

brandonspark:
## What:
This PR makes it so metavariables can start with numerals.

## Why:
Currently, regex matches can use capture groups to introduce metavariables with names like `$1`, `$2`, which do not currently pass our `is_metavar` checks. This means they cannot be inspected with `metavariable-regex`, meaning they are produced, and appear in the interpolated message, but are otherwise invisible to the engine.

this causes issues in stuff like https://returntocorp.slack.com/archives/C01NXGX2EHZ/p1675903424282459

from the thread: 
> im sure this solution causes a litany of other issues (which others can enlighten me on), so im putting it up as RFC. open to other ideas (like renaming the produced metavars, perhaps)

PR checklist:

### DIFF
--- a/changelog.d/pa-2514.fixed
+++ b/changelog.d/pa-2514.fixed
@@ -1,0 +1,2 @@
+Fixed a bug where sometimes metavariables from regex capture groups could not be
+`focus-metavariable`d, `metavariable-regex`d, or would otherwise disappear.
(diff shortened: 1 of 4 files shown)


brandonspark:
Now addressed by #7227

PR #7117, opened 2023-02-10
feat!: pro latest testing workflow

brandonspark:
## What:
This PR sets up a workflow, intended to be dispatched manually, to check whether the current `develop` version of Semgrep is incompatible with the latest released version of Semgrep Pro Engine.

## Why:
Ideally, if we make a breaking change to `semgrep` that necessitates a `semgrep-proprietary` change, we will release both at the same time. Sometimes, however, we will have hotfixes which necessitate releasing just one of them individually, without the other. If we are releasing Semgrep, for instance, we want to be sure that it is not incompatible with Semgrep Pro Engine's last released version, or the latest released version for both will be incompatible with each other, and we'll look bad.

This PR just adds in a manual step to verify that they work together.

## How:
It's pretty similar to the `test-semgrep-pro` workflow I set up for testing the `develop` `proprietary` binary, but instead of using the `develop` binary, it uses the latest released one. It also uses the `returntocorp/semgrep:develop` Docker image, which should be updated per PR to Semgrep, and thus correspond to the `develop` branch currently on `semgrep`.

## Test plan:
See the corresponding check in CI.

Closes PA-2522

PR checklist:

### DIFF
--- a/.github/workflows/test-semgrep-pro-latest.yaml
+++ b/.github/workflows/test-semgrep-pro-latest.yaml
@@ -1,9 +1,9 @@
 # The goal of this workflow is to offer a validation step before releasing
-# Semgrep, to ensure that the to-be-released version of Semgrep is not 
+# Semgrep, to ensure that the to-be-released version of Semgrep is not
 # incompatible with the currently released version of Semgrep Pro.
 #
 # If this workflow succeeds, then it should be safe to release Semgrep and
-# not immediately be incompatible with Semgrep Pro. 
+# not immediately be incompatible with Semgrep Pro.
 
 name: test-semgrep-pro-latest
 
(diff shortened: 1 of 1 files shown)


brandonspark:
@spencerdrak i set up this workflow for the purpose of testing whether a prospective Semgrep release will be incompatible with the latest-released Pro Engine version

the workflow is "on pull request", but ignore that. i put it there just so i could see whether it worked. lmk what you think!

aryx:
LGTM but I'll let @spencerdrak review. Maybe there's a way to factorize things.

spencerdrak (approved):
This LGTM but may change based on the results of our discussion on wednesday re: linking semgrep and semgrep pro.

PR #7123, opened 2023-02-10
fix(validate): treat regex metavariables correctly

emjin:
The current validate behavior doesn't work for $1, $2, etc. See: https://semgrep.dev/playground/s/W5qZ

The problem is that $1 appears in the `metavariable-regex` but does not appear in the pattern string by name (it's bound by the capture group).

Test plan: `semgrep --config lewis_rule.yaml lewis_code.txt --validate`

TODO will add tests after launch, should have a while ago

PR checklist:

### DIFF
--- a/changelog.d/validate-regex-mvar.fix
+++ b/changelog.d/validate-regex-mvar.fix
@@ -1,0 +1,2 @@
+Fixed bug with validate for metavariables like $1, $2. Previously,
+it blocked rules that it shouldn't.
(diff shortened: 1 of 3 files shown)


changelog.d/validate-regex-mvar.fix
```diff
@@ -0,0 +1,2 @@
+Fixed bug with validate for metavariables like $1, $2. Previously,
```
brandonspark: i would write something like 
> CLI: Fixed bug with `semgrep --validate` for regex capture group metavariables like $1, $1. Previously, it blocked rules that it shouldn't.

brandonspark (approved):

PR #7128, opened 2023-02-11
fix: Reduce logspew with Pro Engine

nmote:
I was seeing this a lot with `--debug` and it was drowning out other findings and leading to excessive memory use on the CLI side (which stores up all the output and then prints it at the end).

After this change, the CLI's memory usage with `--debug --pro` is greatly reduced.

Test plan: Manually tested with `--debug --pro` and observed greatly reduced memory usage.

PR checklist:

### DIFF
--- a/src/engine/Match_tainting_mode.ml
+++ b/src/engine/Match_tainting_mode.ml
@@ -249,7 +249,7 @@ let range_of_any any =
        * first place? *)
       | G.Anys []
       (* An enormous amount of logspew with Semgrep Pro Engine happens with this
-       * case. TODO Figure out why avoid it? *)
+       * case. TODO Figure out why and avoid it? *)
       | G.Tk _ ->
           ()
       | _else_ ->
--- a/src/engine/Match_tainting_mode.ml
+++ b/src/engine/Match_tainting_mode.ml
@@ -242,14 +242,20 @@ let range_of_any any =
    * the AST at some point. *)
   match Visitor_AST.range_of_any_opt any with
   | None ->
+      (match any with
       (* IL.any_of_orig will return `G.Anys []` for `NoOrig`, and there is
        * no point in issuing this warning in that case.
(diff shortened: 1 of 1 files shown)


src/engine/Match_tainting_mode.ml
```diff
@@ -242,14 +242,20 @@ let range_of_any any =
...
        * no point in issuing this warning in that case.
        * TODO: Perhaps we should avoid the call to `any_in_ranges` in the
        * first place? *)
-      if any <> G.Anys [] then
-        logger#warning
-          "Cannot compute range, there are no real tokens in this AST: %s"
-          (G.show_any any);
+      | G.Anys []
+      (* An enormous amount of logspew with Semgrep Pro Engine happens with this
+       * case. TODO Figure out why avoid it? *)
```
emjin:
```suggestion
      (* An enormous amount of logspew with Semgrep Pro Engine happens with this
       * case. TODO Figure out why and avoid it? *)
```

nmote:
Follow up ticket: https://linear.app/r2c/issue/PA-2533/follow-up-on-logspew-related-to-no-real-tokens

emjin (approved):
I'm pretty curious about why this happens so frequently, but let's disable it for now

src/engine/Match_tainting_mode.ml
```diff
@@ -242,14 +242,20 @@ let range_of_any any =
    * the AST at some point. *)
   match Visitor_AST.range_of_any_opt any with
   | None ->
+      (match any with
       (* IL.any_of_orig will return `G.Anys []` for `NoOrig`, and there is
        * no point in issuing this warning in that case.
        * TODO: Perhaps we should avoid the call to `any_in_ranges` in the
        * first place? *)
-      if any <> G.Anys [] then
-        logger#warning
```
IagoAbal, line 250: I would rather leave it as it was but just change the severity from `warning` to `trace`.

PR #7132, opened 2023-02-13
chore: no longer complain about `interfile: true` when `--interfile` not specified

brandonspark:
## What:
This PR makes it so that the CLI no longer complains that some rules were run as OSS, even if `--interfile` was not specified.

## Why:
This is a disingenuous error because it's just not relevant to the user, if they specify `--pro-intrafile` or `--pro-languages`. It makes them think there's an error when it's not an error.

## How:
I made it such that we only really report this if `--interfile` was requested.

## Notes:
This is the minimal change to get the behavior correct. There is some redundancy in the CLI code (most introduced by myself), that warrants a refactor. Ultimately, I see a kind of tension between parts of the codebase where we only care about whether the Pro Engine is running (in all three of its forms), and parts where we see fit to dispatch on that information. The existence of both of these makes the code kind of ugly right now.

In addition, we will still report rules run with the intrafile engine and pro languages as "OSS" rules. That is something I'm downscoping to fix for this PR. A more disciplined solution is in order, but later.

We also really need e2e testing for `pro`.

## Test plan:
image

PR checklist:

### DIFF
--- a/cli/src/semgrep/output.py
+++ b/cli/src/semgrep/output.py
@@ -486,6 +486,9 @@ class OutputHandler:
         if self.settings.output_format == OutputFormat.SARIF:
             extra["dataflow_traces"] = self.settings.dataflow_traces
 
+        # as opposed to below, we need to distinguish the various kinds of pro engine
+        extra["engine_requested"] = self.engine_requested
+
         # the rules are used only by the SARIF formatter
         return self.formatter.output(
             self.rules,
@@ -496,6 +499,7 @@ class OutputHandler:
                 time=cli_timing,
                 explanations=explanations,
                 rules_by_engine=rules_by_engine,
+                # this flattens the information into just distinguishing "pro" and "not-pro"
                 engine_requested=self.engine_requested.to_engine_kind(),
             ),
             extra,
--- a/changelog.d/pa-2528.fixed
+++ b/changelog.d/pa-2528.fixed
@@ -1,0 +1,2 @@
+CLI: No longer reports rules as being run with a lack of `interfile: true` when interfile
+analysis was not requested.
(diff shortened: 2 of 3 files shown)


emjin:
Can you also leave a screenshot showing it still working as expected with `--pro`?

brandonspark:
image

cli/src/semgrep/output.py
```diff
@@ -496,6 +499,7 @@ def _build_output(
                 time=cli_timing,
                 explanations=explanations,
                 rules_by_engine=rules_by_engine,
+                # this flattens the information into just distinguishing "pro" and "not-pro"
```
emjin, line 502: Is this because the sarif formatter also outputs a check for `interfile: true`? (Comment could include more context)

emjin (approved):

cli/src/semgrep/output.py
brandonspark: no, it was meant to be a reference to the implementation of `to_engine_kind`, which will simply inject into `PRO`, regardless of the particular kind of pro engine variant

emjin: Why do we choose to flatten this though?

brandonspark: the type of `CliOutputExtra` requires an `engine_kind`, which is a sum of only `PRO` and `OSS`. we could make this all four variants, but as I said in the summary, there are places where this makes life hard for us. in particular, this is the same variant that we use to label our matches, which would require a fair sum of work to categorize matches as `PRO_INTRAFILE` findings or `PRO_LANGUAGE` findings, etc. this should be cleaned up later, but I don't think this is a problem for now

emjin: I see. The comment could clarify that the output engine_kind has less information. I would be fine with merging this and improving the comment later though

brandonspark:
merging to not block release

PR #7140, opened 2023-02-13
chore(sca): refactor calculating number of dependencies

brendongo:
PR checklist:


cli/src/semdep/find_lockfiles.py
```diff
@@ -37,23 +33,34 @@
 }
 
 
-def find_single_lockfile(
-    p: Path, ecosystem: Ecosystem
-) -> Optional[Tuple[Path, List[FoundDependency]]]:
+def lockfile_path_to_manfiest_path(lockfile_path: Path) -> Optional[Path]:
```
underyx:
```suggestion
def lockfile_path_to_manifest_path(lockfile_path: Path) -> Optional[Path]:
```

cli/src/semgrep/app/scans.py
```diff
@@ -233,7 +234,7 @@ def report_findings(
         parse_rate: ParsingData,
         total_time: float,
         commit_date: str,
-        lockfile_scan_info: Dict[str, int],
+        lockfile_dependencies: Dict[str, List[FoundDependency]],
```
brendongo, line 240: Now that we have full dependency list here, easy to send data once app side toggle is done.

cli/src/semgrep/semgrep_main.py
```diff
@@ -197,24 +198,27 @@ def run_rules(
...
                 )
                 rule_matches_by_rule[rule] = dep_rule_matches
                 output_handler.handle_semgrep_errors(dep_rule_errors)
-                output_extra.all_targets.union(targeted_lockfiles)
+
+        # Generate stats per lockfile:
+        for ecosystem in ECOSYSTEM_TO_LOCKFILES.keys():
+            for lockfile in target_manager.get_lockfiles(ecosystem):
+                # Add lockfiles as a target that was scanned
+                output_extra.all_targets.add(lockfile)
```
brendongo, line 225: Technically this is an approximation of the truth instead of the exact truth

brendongo: Could be convinced to revert this part to old behavior

cli/src/semgrep/output_extra.py
```diff
@@ -19,5 +20,6 @@ class OutputExtra:
     all_targets: Set[Path]
     profiling_data: ProfilingData
     parsing_data: ParsingData
+    dependencies: Dict[str, List[FoundDependency]]
```
mmcqd: This seems suspect to me. Putting this here directly contradicts what the comment above the class says it's for. And even if we were to change that comment, it's confusing to have the semgrep-core runner return something with a `dependencies` field which is always guaranteed to be empty, since semgrep-core doesn't actually compute that data. I think this should have remained a separate thing that gets passed around, even though it's less pretty. We could consolidate it with `OutputExtra` and make `ScanExtra` or something so that it's easy to add things?

cli/src/semgrep/semgrep_main.py
mmcqd: I can see a couple issues.

1. This field of the scan data is no longer "the lockfiles we scanned while running the rules", now it's "every lockfile in your repo (that isn't semgrepingored {see next point}), no matter what". In practice this is probably ok, because the SSC reachability rules have every language in them (except rust). So this was already happening, but we should probably be explicit about it, and leave a comment indicating this

2. This will mess with the interaction between semgrepignore and lockfile scanning. This is probably an extreme edge case, but currently if `foo/yarn.lock` is ignored, but we scan `foo/bad.js` with a reachability rule, we _will_ scan `foo/yarn.lock`, and with the old code this was reflected in `lockfile_scan_info`: it would include `foo/yarn.lock`. Since the `get_lockfiles` method respects semgrepignore, in the scenario I described `foo/yarn.lock` would _not_ end up in `lockfile_scan_info` (or `lockfile_dependencies`), even though we scanned it.

This is the kind of thing that shouldn't matter 99.9% of the time but could result in a very very unpleasant debugging experience in like a year lol

cli/src/semgrep/target_manager.py
```diff
@@ -735,21 +731,8 @@ def get_files_for_rule(
...
-            if lockfile not in self.lockfile_scan_info:
-                # We haven't seen this file during reachable finding generation
-                self.lockfile_scan_info[str(lockfile)] = len(deps)
-            parsed[lockfile] = deps
-        return parsed
+    def get_lockfiles(self, ecosystem: Ecosystem) -> FrozenSet[Path]:
+        """
+        Return set of paths to lockfiles for a given ecosystem
+        """
+        return self.get_files_for_language(ecosystem).kept
```
mmcqd, line 756: beautiful

cli/src/semdep/find_lockfiles.py
```diff
@@ -37,23 +33,34 @@
 }
 
 
-def find_single_lockfile(
-    p: Path, ecosystem: Ecosystem
-) -> Optional[Tuple[Path, List[FoundDependency]]]:
+def lockfile_path_to_manifest_path(lockfile_path: Path) -> Optional[Path]:
```
mmcqd: This is much nicer!

mmcqd (dismissed):

cli/src/semgrep/app/scans.py
```diff
@@ -301,7 +304,7 @@ def report_findings(
                 "errors": [error.to_dict() for error in errors],
                 "total_time": total_time,
                 "unsupported_exts": dict(ignored_ext_freqs),
-                "lockfile_scan_info": lockfile_scan_info,
+                "lockfile_scan_info": dependency_counts,
```
underyx, line 310: could you add a snapshot test that snapshots the contents of a scan with dependency_counts in it?

brendongo: Tests exist! They just didn't get updated cause this PR doesnt change anything

underyx: oh silly me

cli/src/semgrep/output_extra.py
brendongo: yeah that's fair. will move out of here

cli/src/semgrep/semgrep_main.py
brendongo: yeah (2) is a good reason to keep old way this was calculated

cli/src/semgrep/output_extra.py
brendongo: Done

cli/src/semgrep/semgrep_main.py
brendongo: We switched over to always respecting semgrepignore. I moved the lockfile finding logic used in reachability rules into TargetManager so this should be an accurate depiction of true scanned files now

underyx (approved):

cli/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awareyarn-sass.yaml-dependency_awareyarn/results.txt
```diff
@@ -87,6 +89,6 @@ Nothing to scan.
 └──────────────┘                                                                
                                                                                 
 
-Ran 1 rule on 0 files: 1 finding.
```
underyx, line 90: lol

underyx:
@brendongo I think the searched_paths in test snapshots are now ordered non-deterministically

underyx:
> @brendongo I think the searched_paths in test snapshots are now ordered non-deterministically

Actually they always were, but we only ever had one element in that field in tests previously. I added a sort call in your PR.

PR #7144, opened 2023-02-14
ask git for remote origin url if repo_url not provided

ievans:
Should fix https://linear.app/r2c/issue/PA-2540/missing-hyperlinks-after-local-scan

Testing situation is not ideal. Testing is not using the `git` subprocess command, because if it does, the output will be non-deterministic as pytest is creating a new directory for the repo every time and thus the origin will be different. We could change the origin, but other tests actually rely on it being clonable. 

PR checklist:

### DIFF
--- a/cli/src/semgrep/meta.py
+++ b/cli/src/semgrep/meta.py
@@ -104,17 +104,17 @@ class GitMeta:
         if not repo_url:
             # if the repo URL was not explicitly provided, try getting it from git
             # nosem: use-git-check-output-helper
-            rev_parse = subprocess.run(
+            git_parse = subprocess.run(
                 ["git", "remote", "get-url", "origin"],
                 capture_output=True,
                 encoding="utf-8",
                 timeout=env.git_command_timeout,
             )
-            if rev_parse.returncode != 0:
-                raise Exception(
-                    "Unable to infer repo_url. Set SEMGREP_REPO_URL environment variable or run in a valid git project"
+            if git_parse.returncode != 0:
+                logger.warn(
+                    f"Unable to infer repo_url. Set SEMGREP_REPO_URL environment variable or run in a valid git project with remote origin defined"
                 )
-            repo_url = rev_parse.stdout.strip()
+            repo_url = git_parse.stdout.strip()
 
         return get_url_from_sstp_url(repo_url)
 
--- a/cli/src/semgrep/meta.py
+++ b/cli/src/semgrep/meta.py
@@ -112,6 +112,7 @@ class GitMeta:
                 raise Exception(
                     "Unable to infer repo_url. Set SEMGREP_REPO_URL environment variable or run in a valid git project"
                 )
+            repo_url = rev_parse.stdout.strip()
 
         return get_url_from_sstp_url(repo_url)
 
--- a/cli/src/semgrep/meta.py
+++ b/cli/src/semgrep/meta.py
@@ -99,7 +99,21 @@ class GitMeta:
 
     @property
     def repo_url(self) -> Optional[str]:
-        return get_url_from_sstp_url(os.getenv("SEMGREP_REPO_URL"))
+        repo_url = os.getenv("SEMGREP_REPO_URL")
+        if not repo_url:
+            # if the repo URL was not explicitly provided, try getting it from git
+            rev_parse = subprocess.run(
+                ["git", "config", "--get", "remote.origin.url"],
+                capture_output=True,
+                encoding="utf-8",
+                timeout=env.git_command_timeout,
+            )
+            if rev_parse.returncode != 0:
+                raise Exception(
+                    "Unable to infer repo_url. Set SEMGREP_REPO_URL environment variable or run in a valid git project"
+                )
+
+        return get_url_from_sstp_url(repo_url)
 
     @property
     def commit_sha(self) -> Optional[str]:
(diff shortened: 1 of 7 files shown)


cli/src/semgrep/meta.py
```diff
@@ -99,7 +99,24 @@ def repo_name(self) -> str:
 
     @property
     def repo_url(self) -> Optional[str]:
-        return get_url_from_sstp_url(os.getenv("SEMGREP_REPO_URL"))
+        env = get_state().env
+        repo_url = os.getenv("SEMGREP_REPO_URL")
+        if not repo_url:
+            # if the repo URL was not explicitly provided, try getting it from git
+            # nosem: use-git-check-output-helper
```
underyx, line 106: could you comment why we're not using git_check_output?

cli/src/semgrep/meta.py
```diff
@@ -99,7 +99,24 @@ def repo_name(self) -> str:
...
+            rev_parse = subprocess.run(
+                ["git", "config", "--get", "remote.origin.url"],
+                capture_output=True,
+                encoding="utf-8",
+                timeout=env.git_command_timeout,
+            )
+            if rev_parse.returncode != 0:
+                raise Exception(
+                    "Unable to infer repo_url. Set SEMGREP_REPO_URL environment variable or run in a valid git project"
+                )
```
underyx: doesn't this just crash if the URL couldn't be extracted? people's CI jobs might break

also, if that's intended, I think there's some specific class we should be using, maybe SemgrepError?

cli/src/semgrep/meta.py
```diff
@@ -99,7 +99,24 @@ def repo_name(self) -> str:
...
     @property
     def repo_url(self) -> Optional[str]:
-        return get_url_from_sstp_url(os.getenv("SEMGREP_REPO_URL"))
+        env = get_state().env
+        repo_url = os.getenv("SEMGREP_REPO_URL")
+        if not repo_url:
+            # if the repo URL was not explicitly provided, try getting it from git
+            # nosem: use-git-check-output-helper
+            rev_parse = subprocess.run(
+                ["git", "config", "--get", "remote.origin.url"],
```
underyx:
```suggestion
                ["git", "remote", "get-url", "origin"],
```

I don't actually know, but my hunch is this is the preferred way to get the remote URL

```
$ git config --get remote.origin.url
git@github.com:returntocorp/semgrep-app.git

$ git remote get-url origin
git@github.com:returntocorp/semgrep-app.git
```

cli/src/semgrep/meta.py
ievans: @underyx I have no idea tbh -- I just re-used the same approach in `repo_name`, which explicitly has a nosem as well. Hmm.

cli/src/semgrep/meta.py
ievans: good points -- I'm not sure why we're using Exception, but it's consistent across the entire file. I do think that it would be pretty weird to run a CI scan on a repo without a remote, so I'm ok with it crashing.

underyx: @ievans the issue is you're asserting not just that there is a remote, but also that it's named `origin`. I'm not sure how safe that second assumption is.

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-local/meta.json
```diff
@@ -2,7 +2,7 @@
   "meta": {
     "semgrep_version": "<sanitized version>",
     "repository": "checkout_project_name",
-    "repo_url": null,
+    "repo_url": "git@github.com:example/fake.git",
```
underyx: is this ok? shouldn't this be an https:// url for links on the app to work?

ievans: this is monkeypatched output to get determinism into the test. But I'm working on a better approach now.

cli/src/semgrep/meta.py
ievans: good point -- I'll make it a warning

cli/src/semgrep/meta.py
```diff
@@ -99,7 +99,24 @@ def repo_name(self) -> str:
...
+                ["git", "remote", "get-url", "origin"],
+                capture_output=True,
+                encoding="utf-8",
+                timeout=env.git_command_timeout,
+            )
+            if git_parse.returncode != 0:
+                logger.warn(
+                    f"Unable to infer repo_url. Set SEMGREP_REPO_URL environment variable or run in a valid git project with remote origin defined"
+                )
+            repo_url = git_parse.stdout.strip()
```
underyx, line 117:
```suggestion
            repo_url = None
            if git_parse.returncode == 0:
                repo_url = git_parse.stdout.strip()
            else:
                logger.warn(
                    f"Unable to infer repo_url. Set SEMGREP_REPO_URL environment variable or run in a valid git project with remote origin defined"
                )
```
this might be nice since `get_url_from_sstp_url` should explicitly handle being passed `None`

underyx (approved):

cli/src/semgrep/meta.py
ievans: Going to keep this consistent with other invocations and leave for a later cleanup that will properly use the git check output helper

PR #7145, opened 2023-02-14
fix: Improve parse error messages related to metavariable-pattern

nmote:
Currently, when there is a nested language in a `metavariable-pattern` and we encounter a parse error when parsing the metavariable's contents, we get something like this (appears when providing the `--verbose` flag to the CLI):

```
[WARN] Syntax error at line /tmp/tmp-3427277-52324a.mvar-pattern:1:
 `<` was unexpected
```

This is totally unactionable, and is a bit scary. There is no way to tell what rule or file led to this result. To improve this situation, I propose two changes in this pull request:

First, when we encounter errors when running a nested pattern in a different language for metavariable-pattern, we annotate the resulting error message with information about the nested target language and the rule ID.

Second, when we get parse errors, we now update their locations so that they point to the original source text rather than to the temp file created for the metavariable pattern.

I'm not completely convinced that we should be even issuing these errors at all. I think it is expected that the contents of a metavariable might not always parse cleanly as another language. But, it's probably better than silently declining to provide results because of a parse error that users don't have any way of finding out about. I'd like to see if this messaging change clears up confusion first and iterate if not.

Test plan:

Rule:

```
rules:
- id: find.really.scary.problems
  patterns:
    - pattern: "$X"
    - metavariable-pattern:
        metavariable: $X
        language: js
        pattern: ...
  message: placeholder message
  languages: [python]
  severity: WARNING
```

test.py:
```
y = 'asdf'

x = '<'
```

Run `semgrep scan -c test.yaml . --verbose` and observe the following lines in the output:

```
[WARN] Syntax error at line test.py:3:
 When parsing a snippet as JavaScript for metavariable-pattern in rule 'find.really.scary.problems', `<` was unexpected
```

PR checklist:

### DIFF
--- a/changelog.d/pa-2537.fixed
+++ b/changelog.d/pa-2537.fixed
@@ -1,0 +1,1 @@
+Improve error messages displayed with `--verbose` when the contents of a metavariable fails to parse.
--- a/src/engine/Match_search_mode.ml
+++ b/src/engine/Match_search_mode.ml
@@ -621,7 +621,23 @@ and get_nested_formula_matches env formula range =
       { env.xconf with nested_formula = true }
       env.rule env.xtarget formula (Some range)
   in
-  env.errors := Report.ErrorSet.union res.RP.errors !(env.errors);
+  (* Update the error messages with some more context. Otherwise it will appear
+   * as if, for example, we encountered a parse error while parsing the original
+   * file as the original language. *)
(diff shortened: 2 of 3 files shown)


aryx (approved):

aryx:
Nice

PR #7158, opened 2023-02-14
feat: Use Semgrep to enforce correct changelog extensions

nmote:
Semgrep doesn't really have support for enforcing standards on path names. The best we can do right now is the workaround described in #2438 to include/exclude paths, and then always match any contents in files that the rule applies to.

Unfortunately, the workaround generates a separate finding for each line, but that's really not a big deal.

Test plan:

Add files `asdf.js` and `asdf.fix` to the `changelog.d` directory. Run `semgrep -c semgrep.yml` and observe findings both when the files are empty and when they are populated. Otherwise, observe no findings.

PR checklist:

### DIFF
--- a/semgrep.yml
+++ b/semgrep.yml
@@ -240,3 +240,21 @@ rules:
         - lang_php/analyze/foundation/unit_prolog_php.ml
         - main_db.ml
         - tests/Test.ml
+  - id: bad-changelog-extension
+    pattern-regex: .*
+    message: |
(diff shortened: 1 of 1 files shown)


underyx (approved):

aryx (commented):
Nice! Dogfooding.
cc @mjambon for example of use of include/exclude we need to handle in osemgrep.

PR #7159, opened 2023-02-15
fix: allow metavar ellipsis params in Go

brandonspark:
## What:
This PR adds in support for metavariable ellipses as parameters in Golang.

## Why:
c.f. [this thread](https://returntocorp.slack.com/archives/C01NXGX2EHZ/p1676405651489119), they were not working properly to capture arguments. [This playground example](https://semgrep.dev/s/wb4P), for instance, does not properly produce a match as expected.

This ended up being because single identifier arguments in Golang are interpreted as types, so something like `func foo($...ARGS)` was interpreted as a function with a single argument of type `$....ARGS`. This is kind of silly, and it's a safe bet if anyone writes that, they don't mean a type which is an ellipsis metavariable (what does that even mean?)

## How:
Special-cased it in the pattern parser.

Note I had to add in logic for signatures in `adjust_signatures`, because there's some logic for distributing types across identifiers. I followed the same logic as for `...`, which is to not try and distribute types over it (that wouldn't really make sense anyways).

Due to the special cases for ellipses, I felt it was better to add it as a variant to the Go AST.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/languages/go/ast/ast_go.ml
+++ b/languages/go/ast/ast_go.ml
@@ -92,6 +92,7 @@ and func_type = {
 
 and parameter_binding =
   | ParamClassic of parameter
+  | ParamMetavarEllipsis of ident
   (* sgrep-ext: *)
   | ParamEllipsis of tok
 
--- a/languages/go/ast/ast_go.ml
+++ b/languages/go/ast/ast_go.ml
@@ -92,8 +92,8 @@ and func_type = {
 
 and parameter_binding =
   | ParamClassic of parameter
-  | ParamMetavarEllipsis of ident
   (* sgrep-ext: *)
+  | ParamMetavarEllipsis of ident
   | ParamEllipsis of tok
 
 and parameter = {
--- a/languages/go/ast/ast_go.ml
+++ b/languages/go/ast/ast_go.ml
@@ -93,6 +93,11 @@ and func_type = {
 and parameter_binding =
   | ParamClassic of parameter
   (* sgrep-ext: *)
+  (* For metavariable ellipsis identifier (like $...ARGS), we don't
+     usually separate this out at the AST level. However, single
+     identifiers as parameters are parsed as types in Go, so we can't
+     reuse the ordinary identifier construct. So we use this.
+  *)
   | ParamMetavarEllipsis of ident
   | ParamEllipsis of tok
 
--- a/languages/go/menhir/parser_go.mly
+++ b/languages/go/menhir/parser_go.mly
@@ -1025,6 +1025,17 @@ fnlitdcl: fntype { $1 }
 arg_type:
 |       name_or_type {
     (match $1 with
+    (* This means a param of the form $...<ID>.
+     * Ordinarily, in Golang, a singular identifier is interpreted as an
+     * anonymous argument of the type of that identifier.
+     * So func foo(int) is allowed.
+
+     * When we see a metavariable ellipsis, it's a little silly to consider that
+     * as a type. If someone writes func foo ($...ARGS), they probably just meant to
+     * capture all of the args of the function, and an anonymous argument of type $...ARGS.
+
+     * So we parse it as such here.
+     *)
     | TName [ (s, _) as id ] when AST_generic.is_metavar_ellipsis s ->
         ParamMetavarEllipsis id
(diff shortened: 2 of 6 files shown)


languages/go/ast/ast_go.ml
```diff
@@ -92,6 +92,7 @@ and func_type = {
 
 and parameter_binding =
   | ParamClassic of parameter
+  | ParamMetavarEllipsis of ident
```
nmote: Put this under the `sgrep-ext` comment below

tests/rules/metavar_ellipsis_param.yaml
```diff
@@ -0,0 +1,13 @@
+rules:
+  - id: test
+    languages:
+      - go
+    message: javascript inside html working!
```
nmote: Maybe choose something more generic or something that better describes this case?

nmote (approved):
I might have tried to do this all in go_to_generic.ml, but this change is pretty simple and I accept the explanation that allowing that to be parsed as a type and then reaching in would be odd.

tests/rules/metavar_ellipsis_param.go
```diff
@@ -0,0 +1,32 @@
...
+
+func bar(string, string, bool) { }
+
+func bar(string, int, string, bool) { }
+
+func qux() { }
+
+func qux(int) { }
+
+func qux(x int) { }
```
nmote, line 28: Wait, shouldn't this match? Ellipsis metavariables should match zero or more items.

nmote:
Also I would describe this as either a fix or a feature and I think there should be a changelog entry.

brandonspark:
@nmote oops, i committed but did not push my changelog entry...

languages/go/ast/ast_go.ml
```diff
@@ -93,6 +93,7 @@ and func_type = {
 and parameter_binding =
   | ParamClassic of parameter
   (* sgrep-ext: *)
+  | ParamMetavarEllipsis of ident
```
aryx, line 101: You could add a comment here like 
"for metavar ellipsis identifier (e.g., $...ARGS), we usually don't have a special construct in the AST for it;
Id of ident can represent such a metavar. But for Go parameters, a single identifier is usually interpreted
as a type, so we can't reuse parameter.pname to store $...ARGS, hence this special construct".

languages/go/menhir/parser_go.mly
```diff
@@ -1021,7 +1023,25 @@ fnliteral: fnlitdcl lbrace listsc(stmt) "}"
...
+     * anonymous argument of the type of that identifier.
+     * So func foo(int) is allowed.
+
+     * When we see a metavariable ellipsis, it's a little silly to consider that
+     * as a type. If someone writes func foo ($...ARGS), they probably just meant to
+     * capture all of the args of the function, and not an anonymous argument of type $...ARGS.
+
+     * So we parse it as such here.
+     *)
+    | TName [ (s, _) as id ] when AST_generic.is_metavar_ellipsis s ->
```
aryx, line 1040: You can also guard it with Flag.sgrep_guard, which make sure we are in parsing pattern mode.

tests/rules/metavar_ellipsis_param.yaml
```diff
@@ -0,0 +1,14 @@
+rules:
+  - id: test
+    languages:
+      - go
+    message: |
+      Metavariable ellipses should be able to match arguments in Go! 
+    patterns:
```
aryx: you could use your new syntax here if you want :)

aryx (approved):
Nice

aryx (commented):
Those diffs are relaxing.

brandonspark:
Note that the TODO in the test is going to be due to a bug in the Go target parser for tree-sitter. We effectively do not do the same hack which is done in `parser_go.mly` and stuff like:
```
func qux(string, x int) { }
```
is interpreted as a function of two arguments named `string` and `x`, both of which have type `int`...

additionally, that's not even the full truth, it also thinks the second argument is of name `,` because of a bug in the translator. I'll open up a PR for that shortly.

brandonspark:
The TODO will be fixed (in part) by #7204

PR #7169, opened 2023-02-15
feat!: code path for simultaneous tainting

brandonspark:
## What:
This PR adds in the beginnings of what I am calling "simultaneous tainting", which is when you run multiple taint rules on a single target at the same time. 

This is Step 1 in the plan of my scope doc for running multiple taint rules at the same time: https://www.notion.so/r2cdev/Taint-Mode-for-Multiple-Rules-bd0b2eb571ae4e3aa2b8661bba8b2edc?pvs=4

## Why:
This is as opposed to how we currently do it, where we invoke the taint engine once on an xtarget for every single taint rule. This causes us to run the dataflow engine on a control flow graph many times, when we really don't have to. This is seen in taint labels, which allow us to invoke dataflow analysis once within a single rule, but for "multiple taint rules".

Preliminary results have showed that taking a number of taint rules and reinterpreting them as a single taint rule with multiple taint labels have had good implications for runtime.

## How:
Just added an alternate codepath, and a tentative new file. Current behavior should be unchanged.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/src/core_cli/Core_CLI.ml
+++ b/src/core_cli/Core_CLI.ml
@@ -111,6 +111,7 @@ let use_parsing_cache = ref Runner_config.default.parsing_cache_dir
 (* similar to filter_irrelevant_patterns, but use the whole rule to extract
  * the regexp *)
 let filter_irrelevant_rules = ref Runner_config.default.filter_irrelevant_rules
+let simultaneous_taint = ref Runner_config.default.simultaneous_taint
 
 (* ------------------------------------------------------------------------- *)
 (* flags used by the semgrep-python wrapper *)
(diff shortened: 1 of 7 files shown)


IagoAbal (changes requested):
Better to not include "simultaneous taint" at all in this PR, just change Semgrep so that taint rules are passed all at once to `Match_tainting_mode`. Instead of `check_rule`, we will have a `check_rules`. This will prepare Semgrep/taint-mode for optimizations that look at the entire set of taint rules to run on one file. (This is how it used to be BTW.)

brandonspark:
Deprecated in favor fo #7187

PR #7173, opened 2023-02-16
Break dependency cycle between Logging and Common

emillon:
Hi,

I am part of the dune dev team. While checking for regressions in the upcoming 3.7 release, I found that the `commons` library does not build with it.

After investigating, there is a dependency cycle in `libs/commons`: `Common.ml` depends on `Logging`, and `logging.mli` depends on `Common`.

Dune used to be OK with that, but the new dependency analysis algorithm in 3.7 makes tighter checks and (correctly) detects it. It's kinda OK in that particular case because `Logging` only uses a type alias from `Common`, but dune can not reliably detect this.

So this PR expands the type alias in a way that splits the cycle and makes it compatible with dune 3.7.

PR checklist:

### DIFF
--- a/libs/commons/Logging.mli
+++ b/libs/commons/Logging.mli
@@ -130,4 +130,4 @@ val add_PID_tag : unit -> unit
    Load a json config file that sets log levels for each logger matching
    specific tags.
 *)
-val load_config_file : Common.filename -> unit
+val load_config_file : string -> unit


emillon:
(context for this change in dune: https://github.com/ocaml/dune/issues/7018)

aryx (approved):

aryx:
Isn't ocamlc supposed also to detect those situations and forbid cycles?

ievans:
thanks @emillon, cool to see this work being done proactively!

emillon:
> thanks @emillon, cool to see this work being done proactively!

no problem. we try pretty hard not to break downstream code so when something like that affects the latest release we try to come up with fixes.

> Isn't ocamlc supposed also to detect those situations and forbid cycles?

In that case I think that ocamlc accepts more configurations that what dune supports. Implementations must be passed in topological order (every module after its dependencies), so cycles between ml files are definitely forbidden. As for interfaces, I'm actually not sure what's supported. I think that each cmi is only considered when building the associated cmo/cmx file, that would mean that cycles between interfaces are ok. But it's simpler for dune to just consider that modules as a whole (impl+intf) should form an acyclic graph.

aryx:
I see. Where are you working @emillon ? On recrute des programmeurs OCaml pour semgrep si ca t'interesse :)

PR #7175, opened 2023-02-16
refactor: ppx deriving visitors for generic AST

nmote:
This adds an iter visitor generated by ppx deriving visitors (http://gallium.inria.fr/~fpottier/visitors/manual.pdf) for the generic AST.

To prove it out, I also converted `range_of_any_opt` to use the new visitor. This will be a good proving ground for this approach. Because `range_of_any_opt` is such a hot code path, I was uncertain of the performance implications of using dynamic dispatch here (although to some extent, the current architecture already makes use of explicit dynamic dispatch). Due to this uncertainty, I conducted some performance tests (described below).

I plan to follow up by converting the existing `mk_visitor` function to use the new autogenerated visitors while presenting the same external interface. Then, we can move callsites over one by one and eventually retire that interface.

Test plan:

Existing automated tests

Manual test of performance on juice-shop (rev
c7e1ade8c593875d5fdb6b3b51cef306adf53de9).

Command: `time semgrep-core -l ts -rules default.yml -j 1 .` where `default.yml` is downloaded from https://semgrep.dev/c/p/default.

Methodology: Run command four times both before and after this change, discarding first result for each series of runs. Record "real" time.

Before:

- 56.459s
- 53.678s
- 55.094s

After:

- 56.344s
- 55.342s
- 54.086s

I haven't done a statistical analysis on these results but they look pretty much the same to me.

PR checklist:


mjambon (approved):
awesome!

libs/ast_generic/AST_generic.ml
```diff
@@ -329,6 +329,74 @@ and canonical_name = string list
 and alternate_name = string list
 [@@deriving show { with_path = false }, eq, hash]
 
+(* Used as a parent class for the autogenerated iter visitor *)
```
aryx, line 332: Maybe reference the fact that the visitor is generated at the very very end of this file, because of the big and long set of recursive types.

libs/ast_generic/AST_generic.ml
```diff
@@ -423,7 +491,7 @@ and id_info = {
   id_info_id : id_info_id; [@equal fun _a _b -> true]
 }
 
-and id_info_id = IdInfoId.t
+and id_info_id = (IdInfoId.t[@name "id_info_id_t"])
```
aryx, line 494: maybe explain what's this @name is, at least the first time we encounter one in the file.

libs/ast_generic/Visitor_AST.ml
```diff
@@ -1460,60 +1460,64 @@ let first_info_of_any any =
...
+          let saved_ranges = !ranges in
+          ranges := None;
+          super#visit_stmt ranges stmt;
+          stmt.s_range <- !ranges;
+          match saved_ranges with
+          | None -> ()
+          | Some r -> incorporate_tokens ranges r)
+      | Some range -> incorporate_tokens ranges range
+
+    (* Mimics the old mk_visitor -- otherwise we will end up visiting resolved
```
aryx, line 1509: should we make it the default in the parent visitor then?

nmote: I ended up kind of doing this in #7209

PR #7180, opened 2023-02-17
refactor(taint): factor out `Match_tainting_mode` for multiple rules

brandonspark:
## What:
This PR changes `Match_tainting_mode.check_rule` to instead ingest multiple rules at once, instead of just one rule. It then  uses that logic in `Match_rules` to pass multiple taint rules at once. Note that this PR should not change existing engine behavior.

## Why:
This will allow us to do optimizations that look at all of the taint rules, such as sharing of common patterns, which is a big memory bottleneck right now, due to hardcoded propagators in Pro. We can also do some cool stuff like combining taint rules into a single taint config, meaning that we can run dataflow analysis just once for a given group of taint rules.

## How:
Changed the type for `Match_tainting_mode.check_rule`, and then added logic for grouping rules. We group taint rules by language, because we can't really do multi-rule optimizations between rules meant for different languages.

I'm not in love with the need to make this `per_rule_boilerplate_fn`, but I tried a few other things and I think this is the cleanest way to achieve it.

## Alternatives considered:
I considered leaving the logic for producing a faulty `match_result` from a timed-out rule run where it was, and having `Match_tainting_mode.check_rules` return a list, where each entry might indicate that it timed out. This ended up producing a lot of duplicated code unless I also made it return the relevant rule as well, however, which also meant I had to return the relevant rule in the Search and Extract cases, which was kind of gross. `Unit_engine` also then had to deal with the possibility of returning a bad result, even though it was impossible by construction.

I considered leaving the call to `Match_tainting_mode.check_rules` where it was, and injecting rules into a sum type that indicated if it was a single rule or a group of rules. This then made it so that I couldn't have the timeout function work at the granularity of rules, and I was considered that timing out for a group of rules would change behavior, and that people wouldn't like that.

## Test plan:
`make test`

PR checklist:


src/engine/Match_tainting_mode.mli
```diff
@@ -56,9 +56,16 @@ val check_fundef :
   * This is a low-level function exposed for debugging purposes (-dfg_tainting).
   *)
 
-val check_rule :
-  Rule.taint_rule ->
+val check_rules :
+  Rule.taint_rule list ->
   (string -> Pattern_match.t -> unit) ->
```
IagoAbal: Since you're here, maybe add a label to this parameter.

src/engine/Match_tainting_mode.mli
```diff
@@ -56,9 +56,16 @@ val check_fundef :
...
   Match_env.xconfig ->
   Xtarget.t ->
-  Report.rule_profiling Report.match_result * debug_taint
+  (Rule.rule ->
```
IagoAbal: And a label to this one.

src/engine/Unit_engine.ml
```diff
@@ -591,10 +591,19 @@ let tainting_test lang rules_file file =
...
+               (fun _rule f -> f ())
            in
-           res.matches)
+           match results with
+           | [ res ] -> res.matches
+           (* By construction, `check_rules` should only return the same number of results as rules it
+              was initially given.
+              So this case is impossible.
+           *)
+           | __else__ -> raise Impossible)
```
IagoAbal: I don't mind it here because it's test code, but otherwise I would have used `| [] | _::_::_ -> raise Impossible` instead.

src/engine/Match_rules.ml
```diff
@@ -68,23 +68,14 @@ let skipped_target_of_rule (file_and_more : Xtarget.t) (rule : R.rule) :
...
-    logger#info "forcing eval of ast outside of rules, for better profile";
-    Lazy.force lazy_ast_and_errors |> ignore);
-
-  let cnt_timeout = ref 0 in
-
-  let res_rules, skipped_rules =
+(* This function separates out rules into groups of taint rules by languages,
+   all of the nontaint rules, and the rules which we skip due to prefiltering.
+*)
+let group_rules xconf rules xtarget =
```
IagoAbal, line 74: Nice that you moved this into a separate function.

src/engine/Match_rules.ml
```diff
@@ -101,53 +92,107 @@ let check ~match_hook ~timeout ~timeout_threshold (xconf : Match_env.xconfig)
...
+  (* Given a thunk [f] that computes the results of running the engine on a single
+     rule, this function simply instruments the computation on a single rule with
+     some boilerplate logic, like setting the last matched rule, timing out if
+     it takes too long, and producing a faulty match result in that case.
+
+     In particular, we need this to call [Match_tainting_mode.check_rules], which
+     will iterate over each rule in a different place, and so needs access to this
+     logic.
+  *)
+  let per_rule_boilerplate_fn r f =
```
IagoAbal: It could be nice to lift this one function out of `let check ...` as a top-level function.

IagoAbal (approved):
LGTM,  let's ask @emjin to take a look too.

src/engine/Match_rules.ml
brandonspark: it inherits from a lot of things already in the closure (`cnt_timeout`, `timeout`, `timeout_threshold`, `file`), will it still be nice to make a function with all of those as params? if you think so, im happy to extract, but I was worried that it would be less clean to do so

IagoAbal: I think the `cnt_timeout` could be defined in `per_rule_boilerplate_fn` itself like:
```ocaml
let mk_per_rule_boilerplate_fn ~timeout ~timeout_threshold file =
  let cnt_timeout = ref 0 in
  fun ... -> ...
```
But it's true that it could get somewhat ugly. I don't have a strong opinion here, I leave it to your judgment!

src/engine/Match_rules.ml
```diff
@@ -101,53 +92,109 @@ let check ~match_hook ~timeout ~timeout_threshold (xconf : Match_env.xconfig)
...
+    rule, this function simply instruments the computation on a single rule with
+    some boilerplate logic, like setting the last matched rule, timing out if
+    it takes too long, and producing a faulty match result in that case.
+
+    In particular, we need this to call [Match_tainting_mode.check_rules], which
+    will iterate over each rule in a different place, and so needs access to this
+    logic.
+*)
+let per_rule_boilerplate_fn ~timeout ~timeout_threshold file rule f =
+  let cnt_timeout = ref 0 in
```
IagoAbal, line 124: I guess you need to put `fun f -> ...` after this ?

IagoAbal: Otherwise each invocation of this `per_rule_boilerplate_fn` below will be creating a new `ref` initialized to `0`.
```ocaml
let per_rule_boilerplate_fn =
     per_rule_boilerplate_fn ~timeout ~timeout_threshold file
   in
```

brandonspark: aw man, i used to explicitly teach people not to do this too...

brandonspark: whoops

PR #7182, opened 2023-02-17
chore(scan): Detect token to authenticate with Bitbucket API

pangngan:
PR checklist:

Closes APP-3691


cli/src/semgrep/app/scans.py
```diff
@@ -272,10 +272,18 @@ def report_findings(
...
+            # GitLab.com (cloud)
+            or os.getenv("GITLAB_TOKEN")
+            # Bitbucket Cloud
+            or os.getenv("PIPELINES_JWT_TOKEN")
+        )
+
         findings_and_ignores = {
             # send a backup token in case the app is not available
-            "token": os.getenv("GITHUB_TOKEN"),
-            "gitlab_token": os.getenv("GITLAB_TOKEN"),
```
pangngan, line 278: I will leave the corresponding app code that still detects GitLab tokens like this for now until we make a decision later to stop supporting x versions back of Semgrep.

cli/src/semgrep/app/scans.py
```diff
@@ -272,10 +272,18 @@ def report_findings(
         ignores = [
             match.to_app_finding_format(commit_date).to_json() for match in new_ignored
         ]
+        token = (
+            # GitHub (cloud)
+            os.getenv("GITHUB_TOKEN")
+            # GitLab.com (cloud)
+            or os.getenv("GITLAB_TOKEN")
+            # Bitbucket Cloud
+            or os.getenv("PIPELINES_JWT_TOKEN")
```
pangngan:
```suggestion
            or os.getenv("BITBUCKET_TOKEN")
```

cli/src/semgrep/app/scans.py
```diff
@@ -272,10 +272,18 @@ def report_findings(
...
+            or os.getenv("GITLAB_TOKEN")
+            # Bitbucket Cloud
+            or os.getenv("BITBUCKET_TOKEN")
+        )
+
         findings_and_ignores = {
             # send a backup token in case the app is not available
-            "token": os.getenv("GITHUB_TOKEN"),
-            "gitlab_token": os.getenv("GITLAB_TOKEN"),
+            "token": token,
```
chmccreery, line 286: How do we detect app-side whether this is from github, gitlab, or bitbucket? Is there a corresponding code-change in `semgrep-app`?

chmccreery (commented):
OMG it's h a p p e n i n g g g g!!

cli/src/semgrep/app/scans.py
pangngan: Yup! We look at the repo url to determine which SCM provider it's from and essentially make the assumption that the token is from that SCM provider as well.

chmccreery: Sweet! Would it be any more accurate to send that information as an enum here instead of relying on URL parsing? Or is the URL parsing really as easy as checking the domain for equality with the SCM? Just want to beware of lossy changes and it seems like one piece of information might potentially be lost here.

chmccreery (approved):

cli/src/semgrep/app/scans.py
pangngan: Oh I think it should be really straightforward. For GitHub and GitLab, we're just checking it with the base URL (essentially doing `repo_url == GITHUB_BASE_URL` and `GITHUB_BASE_URL` is defined as a constant equal to "https://github.com" somewhere else. I think it should be okay to do it for Bitbucket Cloud as well since we should just be checking if the url starts with https://bitbucket.org.

chmccreery: Great! As long as this token thing doesn't apply to GHE or GLSM or BitBucket not-cloud, that sounds good 👍

PR #7186, opened 2023-02-18
refactor: Use ppx deriving visitors for Visitor_AST

nmote:
This changes the implementation without changing the external interface. It now uses the autogenerated visitor that I added in #7175. I took pains to keep the behavior identical in this refactor. This has made it much more obvious where the visitor diverges from just doing an ordinary traversal of the tree. I think there are some places where cleanup is warranted.

I also included a change to `Raw_tree` which would otherwise not have had a motivating example. Basically, we want the `kraw` function to be called on every `raw_tree` node. Without the `v_raw_tree` parameter, `Raw_tree.visit` would traverse the entire `raw_tree` without providing any external code an opportunity to examine it, until it reached a `tok` or an `any`. Without this change, I had 14 test failures.

I plan to follow up by cleaning up nonstandard traversal behavior that was most likely unintentional. I also plan to split out the parts that change the traversal from the parts that call the various `k___` functions. Then, I will start moving clients on to the object oriented visitor interface rather than the current one. Finally, I will remove the current interface altogether.

Test plan: Same as #7175:

Before:

- 56.727s
- 56.062s
- 56.167s

After:

- 55.193s
- 59.065s
- 56.455s

There is one outlier result after this change, but overall this does not seem concerning.

PR checklist:

### DIFF
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -366,13 +366,36 @@ class virtual ['self] iter_parent =
 
     method visit_todo_kind env kind = self#visit_wrap self#visit_string env kind
 
-    method visit_raw_tree_t
-        : 'a. ('env -> 'a -> unit) -> 'env -> 'a Raw_tree.t -> unit =
-      fun f env x ->
-        Raw_tree.visit
-          ~v_token:(self#visit_wrap self#visit_string env)
-          ~v_any:(f env) x
+    (* This is a bit of gymnastics. Because Raw_tree.t is polymorphic, deriving
+     * visitors generates a visit_raw_tree method for the monomorphic raw_tree
+     * type, which expects there to be a polymorphic visit_raw_tree_t for `'a
+     * Raw_tree.t`. That's a problem because child classes want to override
+     * visit_raw_tree and intercept *every* raw_tree node, even nested ones. If
+     * we allowed this to be a polymorphic method as expected, there would be no
+     * way to call back into the monomorphic visit_raw_tree when visiting nested
+     * `raw_tree`s.
+     *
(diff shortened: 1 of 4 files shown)


mjambon (approved):

libs/ast_generic/AST_generic.ml
```diff
@@ -366,13 +366,36 @@ class virtual ['self] iter_parent =
...
+     * don't have access to the `raw_tree` type), and call it here.
+     *
+     * This makes this method monomorphic on the `Raw_tree.t` type parameter in
+     * practice.
+     *
+     * If we introduced another use of `Raw_tree.t` in this big recursive type,
+     * with a different type parameter, then we would start getting type errors
+     * because OCaml would expect this method to be polymorphic. However, since
+     * we will likely never use `Raw_tree.t` here with a type parameter other
+     * than `AST_generic.any`, this doesn't matter. *)
```
aryx, line 389: In Visitor_AST and a few other boilerplate code I used to use for all types, 'a. xxxx
to overcome some typing limitations.
Just FYI.

aryx (approved):
nice!

aryx:
removing 900LOC!

libs/ast_generic/AST_generic.ml
nmote: In this PR I have actually changed this from being explicitly polymorphic like you suggested because that works against us in this case. We have to be able to call the monomorphic `self#visit_raw_tree` method from here in order to get the desired visitor behavior, so we have to allow the OCaml type system to specialize the type of this method when it is inherited in the autogenerated visitor.

It's a bit subtle but hopefully we only have to do minimal maintenance on this from here on out.

PR #7187, opened 2023-02-18
feat!: propagator labels

brandonspark:
## What:
This PR adds in the ability to `require` and output a new `label` for `pattern-propagators`.

## Why:
Sometimes it's useful to be able to say a propagator runs only under a certain label. We have this support for sources and sinks, but not propagators!

## How:
Added in parsing for this feature, as well as augmenting the `source` type to be a record containing a `label` field. This `label` field may not agree with the label that the `src` of the enclosed `call_trace` has, because this label may change over the taint's lifetime, in particular by being propagated to a different label.

It couldn't be placed in `taint` because that includes `Arg` taint, which is supposed to be polymorphic in label.

## Alternatives considered:
I considered just editing the existing `source` of the `call_trace` so that it said the original source just had the label it was propagated to, but this seemed confusing to me, since you may now have a source which claims it had a certain label, but which was never written in the original rule.

Closes PA-1633

PR checklist:


changelog.d/pa-1633.added
```diff
@@ -0,0 +1,18 @@
...
+pattern-propagators:
+  - pattern: |
+      $TO.foo($FROM)
+    from: $FROM
+    to: $TO
+    requires: A
+    label: B
+```
+to denote a propagator which only propagates from $FROM to $TO if $FROM has taint label A. In addition, it also taints $TO
+with label B.
```
IagoAbal: Here I understand that the `$TO` will get both label `A` and `B`.

changelog.d/pa-1633.added
```diff
@@ -0,0 +1,18 @@
...
+
+If `label` is not specified, the `to` is tainted with the same label of taint that $FROM has. If `requires` is not specified,
```
IagoAbal: However here I understand that `$TO` will be labeled with `B` but not with `A`.

src/tainting/Taint.ml
```diff
@@ -34,12 +34,21 @@ type tainted_tokens = G.tok list [@@deriving show]
...
   | PM of PM.t * 'a
   | Call of G.expr * tainted_tokens * 'a call_trace
+  (* We keep this as part of a "call trace" in order to denote when
+     an incoming source has been propagated to have a different taint label.
+     We need this so that we can compute the new label of a propagated source,
+     but so we can remember what the original source label was as well.
+     For all intents and purposes, `Propagate` should only be visible in the
+     case where we are calling [label_of_source_trace].
+  *)
+  | Propagate of 'a call_trace * string (* label *)
```
IagoAbal: I think it should be `orig` or `taint` that should be extended rather than `call_trace`.

IagoAbal: We need to re-think taint traces to accommodate labels properly, but in the meantime I think we could just add a `label: string` field to `taint` or similar.

changelog.d/pa-1633.added
brandonspark: hmm, i think i meant this to mean "it additionally (as an action) taints $TO with label B", not "it additionally (as a separate taint) taints $TO with label B". i can rephrase.

src/tainting/Taint.ml
brandonspark: i think a `label: string` field for `taint` probably shouldn't work, because `Arg` as a source is "polymorphic" in its label, is that correct? i can just fall back to changing the `Src source` label

brandonspark: hm. actually.

brandonspark: i think there's a general issue with taint summaries as they relate to taint labels, that we should address first prior to talking about propagators

brandonspark: i know you already identified there were some regressions with pro w.r.t. taint labels but i only just started thinking about it

brandonspark: for context im thinking of an example that looks like this, which produces no findings even though I think it should:
```
public class Foo {
  public void foo(int a) {
    b = convert(a); 

    sink(b);
  }

  public void main() {
    foo("a");
  }
}
```

```
rules:
- id: taint-labels-interproc
  message: |
    we should be able to do interproc taint findings involving labels!
  severity: ERROR
  languages: [java]
  mode: taint
  pattern-sources:
    - pattern: |
        "a"
      label: A
    - pattern: |
        convert(...)
      label: B
      requires: A
  pattern-sinks:
    - pattern: |
        sink(...)
      requires: B
 ```
 
 (you may be aware of this already)

Comment (IagoAbal) [reply src/tainting/Taint.ml]: Yep, I'm aware that taint labels don't work inter-procedurally. I'm fine merging this PR if we replace the `Propagate` constructor with a `label: string` field in `taint` for example, I don't think it will break anything that currently works. There is work to do to update taint signatures and taint traces to work with labels.

File: changelog.d/pa-1633.added [hunk 4]
```diff
@@ -0,0 +1,20 @@
+Taint: `pattern-propagators` now have optional fields `requires` and `label`,
```
IagoAbal, line 1: Perhaps remind here that taint labels is an _experimental_ feature.

src/tainting/Dataflow_tainting.ml
```diff
@@ -544,11 +554,28 @@ let handle_taint_propagators env thing taints =
...
+        *)
+        if eval_label_requires ~labels prop.spec.prop.propagator_requires then
+          (* If we have an output label, change the incoming taints to be
+             of the new label.
+             Otherwise, keep them the same.
+          *)
+          let new_taints =
+            match prop.spec.prop.propagator_label with
+            | None -> taints
+            | Some label -> Taints.map (propagate_taint_to_label label) taints
```
IagoAbal: Now that I look at it again, this actually replaces _all_ labels in `taint` with the propagated `label`. 🤔 Is that what we want to do?
Given:
```yaml
requires: A and B
label: X
```
We could expect `A` and `B` to be replaced with `X`, but if there is also a label `C` in the data... why should it be replaced with `X` ?

IagoAbal: It could be fine to leave it as it is here for now since it's an experimental feature, but I think this would have to change later.

IagoAbal: It may be simpler to just have a `replace-labels:` field that needs to be set explicitly?
```yaml
requires: A and B
replace-labels: [A, B]
label: X
```

brandonspark: im a little wary of having too many fields, but since it's optional i guess its OK. I've added it now.

src/tainting/Taint.mli
```diff
@@ -45,7 +55,7 @@ type signature = finding list
  * "taint polymorphism", making the taint analysis context-sensitive.
  *
  * Also note that, within each function, if there are multiple paths through
- * which a taint source may reach a sink, we do not keep all of them but only
+ * which a taint ssourcee may reach a sink, we do not keep all of them but only
```
IagoAbal: Typo: "ssourcee"

src/tainting/Taint.ml
```diff
@@ -105,18 +120,24 @@ let _show_finding = function
...
-  let pm1, ts1 = pm_of_trace s1 and pm2, ts2 = pm_of_trace s2 in
-  Stdlib.compare
-    (pm1.rule_id, pm1.range_loc, pm1.env, ts1.Rule.label)
-    (pm2.rule_id, pm2.range_loc, pm2.env, ts2.Rule.label)
+  let pm1, ts1 = pm_of_trace s1.call_trace
+  and pm2, ts2 = pm_of_trace s2.call_trace in
+  match String.compare s1.label s2.label with
+  | 0 ->
+      Stdlib.compare
+        (pm1.rule_id, pm1.range_loc, pm1.env, ts1.Rule.label)
```
IagoAbal: Why not just
```ocaml
   Stdlib.compare
     (pm1.rule_id, pm1.range_loc, pm1.env, s1.label, ts1.Rule.label)
     (pm2.rule_id, pm2.range_loc, pm2.env, s2.label, ts2.Rule.label)
```

IagoAbal (approved):
Kudos for making this kind of non-trivial additions to taint-mode! 🎉  I can see that you are able to navigate the code quite well. Feel welcome to make other suggestions to rename functions (like the ones you renamed), add comments, etc.

src/tainting/Taint.ml
brandonspark: ooh, good point. i forgot polymorphic compare was, well, polymorphic

PR #7207, opened 2023-02-22
chore: remove fdef from taint signature hook

brandonspark:
## What:
This PR removes the entire `fdef` from the `hook_function_taint_signature`, as the majority of it is unused. We keep the `parameters`, which denote the parameter list of the function in particular.

## Why:
Function definitions are big, and can take up a lot of memory!

## Test plan:
`make test`.

On `develop`, on `android`, with `deepsemgrep-sqli-rules`, we see the following memory plot:
image

With this change and the same parameters, we see:
image

Closes PA-2563

PR checklist:


IagoAbal:
> I'm not sure if this is user-facing or not.

It is, although I would just describe it as "Reduced peak memory usage during inter-file analysis" or something like that.

IagoAbal (approved):

aryx (approved):
Very nice!
How did you find this could be the source of high memory usage?

IagoAbal:
> Very nice!
> How did you find this could be the source of high memory usage?

https://returntocorp.slack.com/archives/C01NXGX2EHZ/p1677027224039889

PR #7208, opened 2023-02-22
feat(perf): share matches for taint formulas

brandonspark:
## What:
This PR adds the ability to share matches between taint rules running on a different file. In particular, if there is a source, sink, propagator, or sanitizer that is used more than once between multiple rules (or even a single rule), we will cache and share the pre-computed matches.

## Why:
This may have marginal effects on OSS engine. In particular, hardcoded propagators also fall under this category, so we expect to see significant gains for Pro Engine.

## How:
I made a `Formula_tbl` module which allows mapping from rule formulas to things (in this case, matches and match explanations). This module uses structural equality on ASTs to compare formulae, and a newly-derived hash on rule formula. Hashing the Generic AST is possible, but might be kind of hefty, so I've opted to just hash the literal string that the pattern originally came from, in the hopes that it will serve as a "good enough" proxy that still permits equality to behave normally.

We also factor out a lot of `Match_tainting_mode` by a `formula_cache` argument, which must be created prior to running on a bunch of rules. `check_rules` does this for you automatically, behind the scenes, but it must be created manually if you call `taint_config_of_rule` a bunch of times, as Pro Engine does.

## Alternatives considered:

I considered factoring out `taint_config_of_rule` to take in multiple rules, and then hide the taint config creation. `handle_findings` and `xconf` are both different when invoked by Pro Engine, for each taint config, however, so that logic would need to be ported too, and I thought at that point creating an external cache was sufficiently uncumbersome that it would be cleaner to just keep it the way it is (also because I expect `Match_tainting_mode.taint_config_of_rule` to need no more call-sites.

For `Formula_tbl`, I considered caching only things which are literally xpatterns (which would subsume the propagators case), but I didn't want to make it too specialized to the hardcoded propagators. That being said, having the exact same formula is probably pretty rare outside of hardcoded propagators (though it does happen)

## Test plan:
`make test`

I conducted benchmarks via `deepsemgrep-sqli-rules.yaml` on `android`.

On `develop`, we get the following trace:
image

With these sharing changes, we get:
image

Memory usage went up very slightly, which I did not actually anticipate. But time elapsed went down, by 2x in this case.

We actually see elsewhere in the memtrace that memory usage _in total_ goes from 342G to 154G, so we did save on memory (from not recomputing propagators). But for some reason, it seems that the excess propagators are not actually the source of the _maximal_ memory usage.

Closes PA-2553

PR checklist:

### DIFF
--- a/src/engine/Match_tainting_mode.ml
+++ b/src/engine/Match_tainting_mode.ml
@@ -162,7 +162,7 @@ module Formula_tbl = struct
           (* if there's only 1 more use left, there's no point
              in caching it
           *)
-          ranges, expls
+          (ranges, expls)
         else (
           (* otherwise, this is the first time we've seen this
              formula, and we should cache it
--- a/src/engine/Match_tainting_mode.ml
+++ b/src/engine/Match_tainting_mode.ml
@@ -156,34 +156,6 @@ module Formula_tbl = struct
     | Some (ranges, expls) -> (ranges, expls)
 end
 
-(*
-let identify_sharing (rules : Rule.taint_rule list) =
-  let count_tbl = Formula_tbl.create 128 in
-  let flat_formulas =
-    rules |> List.concat_map (fun rule ->
-      let `Taint (spec : R.taint_spec) = rule.R.mode in
-      Common.map (fun source -> source.R.source_formula) (snd spec.sources)
-      @ Common.map (fun sanitizer -> sanitizer.R.sanitizer_formula) spec.sanitizers
-      @ Common.map (fun sink -> sink.R.sink_formula) (snd spec.sinks)
-      @ Common.map (fun propagator -> propagator.R.propagator_formula) spec.propagators
-    )
-  in
-  flat_formulas |> List.iter (fun formula ->
-    match Formula_tbl.find_opt formula_tbl formula with
-    | None -> Formula_tbl.add formula_tbl formula 1
-    | Some x -> Formula_tbl.replace formula_tbl formula (1 + x)
-  );
-  let
-  Formula_tbl.iter (fun k v ->
-    (* if there's only one occurrence, take it out *)
-    if v <= 1 then
-      None
-    else
-      Some v
-  ) formula_tbl;
-  formula_tbl
-  *)
-
 (*****************************************************************************)
 (* Finding matches for taint specs *)
 (*****************************************************************************)
--- a/src/engine/Match_tainting_mode.ml
+++ b/src/engine/Match_tainting_mode.ml
@@ -147,11 +147,11 @@ module Formula_tbl = struct
     let hash = Rule.hash_formula
   end)
 
-  let cached_find_opt formula_cache formula cont =
+  let cached_find_opt formula_cache formula compute_matches_fn =
     match find_opt formula_cache formula with
     | None ->
-        let ranges, expls = cont () in
-        replace formula_cache formula (ranges, expls);
+        let ranges, expls = compute_matches_fn () in
+        add formula_cache formula (ranges, expls);
         (ranges, expls)
     | Some (ranges, expls) -> (ranges, expls)
 end
--- a/src/engine/Match_tainting_mode.ml
+++ b/src/engine/Match_tainting_mode.ml
@@ -132,17 +132,31 @@ type propagator_match = {
 }
 (** Taint will flow from `from` to `to_` through the axiliary variable `id`. *)
 
-module Formula_tbl = Hashtbl.Make (
-    struct
-      type t = Rule.formula
+(* This Formula_tbl structure is used to create a "formula cache", which will
+   permit the sharing of matches resulting from common formulas that occur as
+   sources, sinks, sanitizers, or propagators.
 
-      let equal = AST_utils.with_structural_equal Rule.equal_formula
+   In particular, because of hardcoded propagators, we expect to see lots of
+   sharing from Semgrep Pro Engine.
+*)
+module Formula_tbl = struct
+  include Hashtbl.Make (struct
+    type t = Rule.formula
 
-      let hash = Hashtbl.hash
-    end
-  )
+    let equal = AST_utils.with_structural_equal Rule.equal_formula
+    let hash = Rule.hash_formula
+  end)
 
-  (*
+  let cached_find_opt formula_cache formula cont =
+    match find_opt formula_cache formula with
+    | None ->
+        let ranges, expls = cont () in
+        replace formula_cache formula (ranges, expls);
+        (ranges, expls)
+    | Some (ranges, expls) -> (ranges, expls)
+end
+
+(*
 let identify_sharing (rules : Rule.taint_rule list) =
   let count_tbl = Formula_tbl.create 128 in
   let flat_formulas =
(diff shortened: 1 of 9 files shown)


IagoAbal:
> We actually see elsewhere in the memtrace that memory usage in total goes from 342G to 154G, so we did save on memory (from not recomputing propagators).

Have you checked the GC stats ?

IagoAbal:
> Hashing the Generic AST is possible, but might be kind of hefty

The default `hash` function won't look at the entire AST if that is huge, see https://v2.ocaml.org/api/Hashtbl.html#VALhash_param.

IagoAbal:
> > Hashing the Generic AST is possible, but might be kind of hefty
> 
> The default `hash` function won't look at the entire AST if that is huge, see https://v2.ocaml.org/api/Hashtbl.html#VALhash_param.

Or perhaps you meant that _deriving_ `hash` for Generic AST is what is heavy?

src/engine/Match_tainting_mode.ml
```diff
@@ -132,6 +132,30 @@ type propagator_match = {
...
+
+    let equal = AST_utils.with_structural_equal Rule.equal_formula
+    let hash = Rule.hash_formula
+  end)
+
+  let cached_find_opt formula_cache formula cont =
+    match find_opt formula_cache formula with
+    | None ->
+        let ranges, expls = cont () in
+        replace formula_cache formula (ranges, expls);
```
IagoAbal: You can use `add` here given that you _know_ `formula` is not in the cache. Or is there another reason for `replace` ?

src/engine/Match_tainting_mode.ml
```diff
@@ -132,6 +132,30 @@ type propagator_match = {
...
+*)
+module Formula_tbl = struct
+  include Hashtbl.Make (struct
+    type t = Rule.formula
+
+    let equal = AST_utils.with_structural_equal Rule.equal_formula
+    let hash = Rule.hash_formula
+  end)
+
+  let cached_find_opt formula_cache formula cont =
```
IagoAbal: I'm not sure about passing this `cont` here... but let's check the rest of the code first.

IagoAbal: In any case I think `cont` is a bad name, when I see `cont` I expect to see CPS or something, so you would apply `cont` to the result of looking up the formula in the cache. But this argument is a _delayed_ `range_w_metas_of_formula`.

IagoAbal (changes requested):
I like that it's a rather small change, although it's perhaps _too_ simple. It was reasonable to expect that this would increase the peak memory usage since it's keeping every taint-subformula in memory for the entire analysis of the file.

Potentially this could raise the peak memory usage more than just a little bit. But I'm fine merging this if we put it under a flag.

Otherwise, there should be a step where we count how many times a given formula is needed. Then the cache could just hold the formula's results for as long as it is needed.

IagoAbal:
We should also do some testing like running `p/default` on a bunch of large repos and making sure 1) perf is same or better in all cases, 2) the findings count doesn't change.

brandonspark:
> > We actually see elsewhere in the memtrace that memory usage in total goes from 342G to 154G, so we did save on memory (from not recomputing propagators).
> 
> Have you checked the GC stats ?

I have not. Do you have a recommended way of checking / is there something in particular you're concerned about?

> The default hash function won't look at the entire AST if that is huge, see https://v2.ocaml.org/api/Hashtbl.html#VALhash_param.

For this, I was concerned that `Hashtbl.hash` will hash the tokens as well, leading to ASTs which are structurally the same being hashed differently (and thus invisible to the hash table). The generated hash I was concerned about being hefty.

> Potentially this could raise the peak memory usage more than just a little bit. But I'm fine merging this if we put it under a flag.

Is there something in particular you'd like to see that could avoid merging under a flag? My impression is that merging under a flag is akin to not merging at all, since either an entry point would need to be added to Semgrep CLI to use it (which users would likely not use anyways), or users would just be unaware of it.

I'm happy to work on doing a counting step where we precompute which formulae may be necessary to keep.

> We should also do some testing like running p/default on a bunch of large repos and making sure 1) perf is same or better in all cases, 2) the findings count doesn't change.

I did run `p/default` on `OpenClinica`, and obtained that time went from 53.5 minutes to 40 minutes. I was unable to obtain a memory trace, however, as I suspect it was just too big (`memtrace-viewer` wouldn't let me open it afterwards).

src/engine/Match_tainting_mode.ml
brandonspark: yeah thats fair, i think i was just being overly paranoid and figured better safe than sorry, but I can change it

src/engine/Match_tainting_mode.ml
brandonspark: sure, thats fair. changed!

brandonspark:
Post specialized formula cache sharing:

With sharing, on `android`, with `deepsemgrep-sqli-rules`:
image

Without sharing, with the same parameters:
image

So memory usage did ultimately go down! Time benefits seem similar.

src/engine/Match_tainting_mode.ml
```diff
@@ -648,6 +734,15 @@ let check_rules ~match_hook
        RP.rule_profiling RP.match_result) (rules : R.taint_rule list)
     (xconf : Match_env.xconfig) (xtarget : Xtarget.t) :
     RP.rule_profiling RP.match_result list =
+  (* We create a "formula cache" here, before dealing with individual rules, to
+     permit sharing of matches for sources, sanitizers, propagators, and sinks
+     between rules.
+
+     In particular, this expects to see big gains due to shared propagators,
+     in Semgrep Pro. There may be some benefit in OSS, but it's low-probability.
```
IagoAbal, line 750: I wouldn't say it's low probability, have you tested it?

brandonspark: i haven't, but i find the likelihood of a source, sink, or sanitizer being the exact same between different rules being a relatively chance event (outside of hardcoded propagators). note that they need to be _exactly_ the same, even down to things like metavariable naming

src/engine/Match_tainting_mode.mli
```diff
@@ -30,8 +30,23 @@ val hook_setup_hook_function_taint_signature :
   *   (or we could infer a signature for them too...).
   *)
 
+module Formula_tbl : Hashtbl.S with type key = Rule.formula
```
IagoAbal: Do we need to expose this module?

src/engine/Match_tainting_mode.mli
```diff
@@ -30,8 +30,23 @@ val hook_setup_hook_function_taint_signature :
...
+  Formula_tbl.t
+
 (* It could be a private function, but it is also used by Deep Semgrep. *)
+(* This [formula_cache] argument is exposed here because this function is also
+   a subroutine but the cache itself should be created outside of the any main
+   loop which runs over rules. This cache is only safe to share with if
+   [taint_config_of_rule] is used on the same file!
+*)
 val taint_config_of_rule :
+  formula_cache:
```
IagoAbal: It would be nicer having an abstract type alias `formula_cache`, the rest should be an implementation detail.

src/engine/Match_tainting_mode.ml
```diff
@@ -132,6 +132,84 @@ type propagator_match = {
...
+           @ Common.map
+               (fun propagator -> propagator.R.propagator_formula)
+               spec.propagators)
+  in
+  flat_formulas
+  |> List.iter (fun formula ->
+         match Formula_tbl.find_opt count_tbl formula with
+         | None -> Formula_tbl.add count_tbl formula 1
+         | Some x -> Formula_tbl.replace count_tbl formula (1 + x));
+  (* This new table will map each formula which exhibits sharing
```
IagoAbal: I would go further and keep the actual number of occurrences. Then, whenever the formula is requested via `cached_find_opt` I would decrease this counter. And when the counter reaches `0` then we know it can be freed.

The simpler alternative, is that we require a formula to occur in X% of the rules before we share it. If a formula is just used in two rules, and we have 200 rules to run, and the formula produces a lot of findings (hence consumes a lot of memory), then it may not be worthwhile to cache it. We could experiment with different percentages and find some sweet spot.

IagoAbal:
> > Have you checked the GC stats ?
> 
> I have not. Do you have a recommended way of checking / is there something in particular you're concerned about?

Mainly out of curiosity, I expect these numbers to have improved, but looking at the number of promoted words and major collections can hint if the reduction in runtime is largely due to a reduction in memory allocations.
 
> > The default hash function won't look at the entire AST if that is huge, see https://v2.ocaml.org/api/Hashtbl.html#VALhash_param.
> 
> For this, I was concerned that `Hashtbl.hash` will hash the tokens as well, leading to ASTs which are structurally the same being hashed differently (and thus invisible to the hash table). The generated hash I was concerned about being hefty.

Good point, maybe clarify that with a comment when you `[@hash.ignore]` the Generic AST.

> > We should also do some testing like running p/default on a bunch of large repos and making sure 1) perf is same or better in all cases, 2) the findings count doesn't change.
> 
> I did run `p/default` on `OpenClinica`, and obtained that time went from 53.5 minutes to 40 minutes. I was unable to obtain a memory trace, however, as I suspect it was just too big (`memtrace-viewer` wouldn't let me open it afterwards).

Could you run `p/default` on some more repos using Semgrep OSS? Ideally not just Java but also some JS/TS and Python repos? Don't need to do a memory trace. Simply `/usr/bin/time` to get real time and peak memory, and check that the number of findings is not affected too. I'm a little bit worried that a minor detail can cause a major regression here, despite I don't see how, but it wouldn't be the first time. It's partly my bad not having done it yet, but we need to automate this kind of testing.

IagoAbal:
> > > Have you checked the GC stats ?
> > 
> > 
> > I have not. Do you have a recommended way of checking / is there something in particular you're concerned about?
> 
> Mainly out of curiosity, I expect these numbers to have improved, but looking at the number of promoted words and major collections can hint if the reduction in runtime is largely due to a reduction in memory allocations.

No need to do this, I can check it myself too.

src/engine/Match_tainting_mode.mli
brandonspark: good point, we do not

src/engine/Match_tainting_mode.ml
brandonspark: Here's a summary of the sharing counts in `pro` with `p/default:
```
!!!This is a proprietary extension of semgrep.!!!
!!!You should not call directly this program.!!!
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
2
2
2
2
2
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
here's a table
1
1
1
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
3
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
2
2
4
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
2
2
2
2
2
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
19
here's a table
1
1
1
1
2
3
3
3
3
3
3
3
3
3
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
4
4
4
4
5
7
19
20
62
62
62
62
62
62
62
62
62
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
2
3
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
2
2
2
2
2
2
3
3
3
6
17
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
2
3
5
6
8
8
9
here's a table
1
1
1
1
3
here's a table
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
1
here's a table
1
1
1
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
2
```
where each "here's a table" denotes a new invocation of `mk_specialized_formula_cache`. I used it on a single target, so I'm not actually quite sure why it's being called so many times, but it should be being called for each batch of taint rules, per language. So maybe that's why.

This seems to show some measure of sharing. Notably `p/default` has about 23 `interfile: true` rules, so those should be the ones with hardcoded propagators, I think.

brandonspark:
New stats:

on `develop`:
image

On `develop` with sharing using counting:
image

On `develop` with sharing before counting:
image

where the drastic memory reductions as compared to before are because of #7207. Looks good still.

brandonspark:
With sharing, using `semgrep` on OpenClinica using `p/default`:
image

Without sharing:
image

brandonspark:
On `vscode` using `p/default`:

With sharing:
image


Without sharing:
image

brandonspark:
(it seems that the findings count has stayed about the same)

brandonspark:
(both of the above perf findings are with OSS)

src/engine/Match_tainting_mode.ml
IagoAbal: Asked SR but my intuition is that there should be, because rules for the same framework should use mostly the same sources and sanitizers. And I guess SR just copy-pastes them, although perhaps not always. Almost as soon as SR started using taint mode we started discussing how to share sources across taint rules.

src/engine/Match_tainting_mode.ml
```diff
@@ -132,6 +132,87 @@ type propagator_match = {
...
+
+  let cached_find_opt formula_cache formula compute_matches_fn =
+    match find_opt formula_cache formula with
+    | None ->
+        (* it should not actually be possible for a formula to
+           not be in the formula table
+
+           just don't cache it I guess
+        *)
+        raise Common.Impossible
```
IagoAbal: `logger#error` instead and `comptue_matches_fn` anyways, just make it clear this is defensive programming... but we never know, it's always funny when you get those "impossible" exceptions.

brandonspark: oops, i originally had it as `compute_matches_fn` anyways, but i changed it to impossible while debugging. good catch.

src/engine/Match_tainting_mode.mli
```diff
@@ -31,7 +41,13 @@ val hook_setup_hook_function_taint_signature :
   *)
 
 (* It could be a private function, but it is also used by Deep Semgrep. *)
+(* This [formula_cache] argument is exposed here because this function is also
+   a subroutine but the cache itself should be created outside of the any main
+   loop which runs over rules. This cache is only safe to share with if
+   [taint_config_of_rule] is used on the same file!
+*)
 val taint_config_of_rule :
+  formula_cache:formula_cache ->
```
IagoAbal: Perhaps to make it even more clear we could name this parameter `per_file_formula_cache` ?

IagoAbal (approved):
Nice work! I would like to explore some details more, particularly why this doesn't give us a perf gain with Semgrep OSS... Perhaps there is no sharing indeed, or perhaps we need to do minor rewritings to the rules, or something else.

PR #7217, opened 2023-02-23
refactor: Migrate src/analyzing to generated visitors

nmote:
Test plan: Automated tests

PR checklist:

### DIFF
--- a/src/analyzing/Constant_propagation.ml
+++ b/src/analyzing/Constant_propagation.ml
@@ -345,72 +345,70 @@ let var_stats prog : var_stats =
         stat
   in
 
-  let hooks =
-    {
-      V.default_visitor with
-      V.kdef =
-        (fun (k, _v) x ->
-          match x with
-          | ( {
-                name =
-                  EN
-                    (Id
-                      (id, { id_resolved = { contents = Some (_kind, sid) }; _ }));
-                _;
-              },
-              VarDef { vinit = Some _e; _ } ) ->
-              let var = (H.str_of_ident id, sid) in
-              let stat = get_stat_or_create var h in
-              incr stat.lvalue;
-              k x
-          | _ -> k x);
-      V.kexpr =
-        (fun (k, vout) x ->
-          match x.e with
-          (* TODO: What if there is an asignment inside the `lhs` ? *)
-          | Assign (* v = ... *) (lhs, _, e2)
-          | AssignOp (* v += ... *) (lhs, _, e2) ->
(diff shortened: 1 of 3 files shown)


src/analyzing/Constant_propagation.ml
```diff
@@ -345,72 +345,70 @@ let var_stats prog : var_stats =
...
+            incr stat.lvalue
+        | N (Id (id, { id_resolved = { contents = Some (_kind, sid) }; _ })) ->
+            let var = (H.str_of_ident id, sid) in
+            let stat = get_stat_or_create var h in
+            incr stat.rvalue;
+            super#visit_expr env x
+        | _ -> super#visit_expr env x
+    end
   in
-  let visitor = V.mk_visitor ~vardef_assign:false hooks in
```
aryx, line 412: we don't need to set something equivalent to vardef_assign:false with this new visitor?
By default vardef_assign is set to false?

src/analyzing/Visit_function_defs.ml
```diff
@@ -15,32 +15,33 @@
...
-                v (G.S body)
-            | __else__ -> k def);
-        V.kfunction_definition =
-          (fun (k, _v) def ->
-            f None def;
-            (* go into nested functions *)
-            k def);
-      }
-  in
+  let (lazy v) = visitor_instance in
```
aryx, line 45: why this use of lazy?

aryx (approved):

src/analyzing/Visit_function_defs.ml
nmote: I changed this to be initialized at the top level (see the definition of visitor_instance above) to avoid repeated allocations but I wanted to avoid instantiating the class until it's actually needed. It's a premature optimization for sure but it seems harmless at worst to me.

src/analyzing/Constant_propagation.ml
nmote: Correct, it's set to `false` by default.

nmote: But also, this is using a vanilla visitor rather than the one designed for matching which generates all sorts of extra partial nodes, as well as allowing callers to request other behavior changes using `vardef_assign` and the other flags. I think most callers don't actually need to visit generated partial nodes or do anything else strange, so I'm updating them to use the vanilla visitor instead. Where needed, I'll have callers inherit from the `matching_visitor` that currently backs `mk_visitor`.

PR #7218, opened 2023-02-24
Add hashes to better track findings over time as code changes

chmccreery:
**Problem:**

Our ability to track findings over time has relied for a long time upon syntactic_id and match_based_id, which are both hashes of tuples of information about a match.

The syntactic_id is a hash of the code contents of the finding or "match", the file path, the rule name, and an index. The code contents themselves cannot be passed in plaintext with semgrep's current terms and agreements, but the rest of the items in this tuple can be passed in plaintext (and already are). By including the file path, rule name, and index inside of the hash, we prevent semgrep-app from being able to track a finding across file path renames, rule renames, and index changes. 

The match_based_id, while less sensitive to code changes, has the same problem. In addition, we have discovered that it is overly sensitive to rule updates because it is essentially a hash of the rule pattern itself, and it is in fact so non-specific that it exacerbates problems where we don't know which of two findings got fixed.

**Solution:**

To overcome some of these limitations, we have decided to add in several new hashes, each of which depends _only_ on information that cannot be passed in plaintext:
- **code_hash:** this is an alternative to syntactic_id, and is a hash of the same code contents, without the file path, rule name, and index that are also inside the syntactic_id hash
- **pattern_hash:** this is an alternative to match_based_id and is a hash of the rule pattern with metavariables substituted in, without file path, rule name, or index
- **start_line_hash**: this is a hash of the contents of just the first line of the code match. It may be useful in linking findings together when the code_hash changes due to a line other than the first line changing.
- **end_line_hash:** this is a hash of the contents of just the last line of the code match. It may be useful in linking findings together when the code_hash changes due to a line other than the last line changing.

*Note: We are not doing anything with these hashes yet in the app, but we want to start collecting them, since the app has to support many versions of semgrep. The sooner we start sending these hashes when we upload findings, the faster we can shift our users over for an improved experience.*

PR checklist:


cli/src/semgrep/rule_match.py
```diff
@@ -290,6 +284,74 @@ def get_match_based_id(self) -> str:
...
+        We started collecting this in addition to match_based_id because the match_based_id will
+        change when the file path or rule name or index changes, but all of those can be passed
+        in plaintext, so it does not make sense to include them in the hash.
+
+        By sending the hash of ONLY the pattern contents, we can determine whether a finding
+        has only changed because e.g. the file path changed
+        """
+        # Upon reviewing an old decision,
+        # there's no good reason for us to use MurmurHash3 here,
+        # but we need to keep consistent hashes so we cannot change this easily
```
chmccreery:
```suggestion
```

cli/src/semgrep/rule_match.py
```diff
@@ -290,6 +284,74 @@ def get_match_based_id(self) -> str:
...
+
+        By sending the hash of ONLY the code contents, we can determine whether a finding
+        has moved files (path changed), or moved down a file (index changed) and handle that
+        logic in the app. This also gives us more flexibility for changing the definition of
+        a unique finding in the future, since we can analyze things like code, file, index all
+        independently of one another.
+        """
+        # Upon reviewing an old decision,
+        # there's no good reason for us to use MurmurHash3 here,
+        # but we need to keep consistent hashes so we cannot change this easily
```
chmccreery:
```suggestion
```

cli/src/semgrep/rule_match.py
```diff
@@ -118,6 +121,13 @@ def start(self) -> core.Position:
     def end(self) -> core.Position:
         return self.match.location.end
 
+    def get_individual_line(self, line_number: int) -> str:
```
jbergler, line 124: Since this doesn't do any bounds checking does it need different type signature? 
```suggestion
    def get_individual_line(self, line_number: int) -> str | None:
```

cli/src/semgrep/rule_match.py
```diff
@@ -290,6 +284,68 @@ def get_match_based_id(self) -> str:
...
+        whenever the file path or rule name or index changes, but all of those can be sent in
+        plaintext, so it does not make sense to include them inside a hash.
+
+        By sending the hash of ONLY the code contents, we can determine whether a finding
+        has moved files (path changed), or moved down a file (index changed) and handle that
+        logic in the app. This also gives us more flexibility for changing the definition of
+        a unique finding in the future, since we can analyze things like code, file, index all
+        independently of one another.
+        """
+        hash_int = hash128(str(self.syntactic_context))
```
jbergler: Is murmurhash simply chosen because of historical reasons? If so I'd suggest we switch to something a bit more current and off the shelf like sha256

The murmurhash code looks unmaintained (last commit to the original C library was 8 years ago and we use a pure python re-implementation) and it calls out that the hash is not stable across architectures (32bit vs 64bit produce different values)

cli/tests/e2e/snapshots/test_ci/test_dryrun/results.txt
```diff
@@ -192,6 +198,12 @@ Would have sent findings and ignores blob: {
             },
             "is_blocking": false,
             "match_based_id": "2c4ff12fcdf80ef1c00dd0f566ae102d792c7ba68e560d70f111aae3b3216c0b1b943e74d2ce29c0361f1fbc37bd4e9aafd32c3435a36c61b8bd3963efe0d7a1_0",
+            "hashes": {
```
jbergler, line 201: This is an SCA finding - is it expected that these are all null?

cli/tests/unit/test_match_hashes.py
```diff
@@ -0,0 +1,163 @@
...
+
+@pytest.mark.quick
+def test_line_hashes_hash_correct_line(mocker, double_eqeq_rule, foo_contents):
+    mocker.patch.object(Path, "open", mocker.mock_open(read_data=foo_contents))
+    match_1 = get_rule_match(start_line=4, end_line=5)  # 5 == 5\n6 == 6\n
+    match_2 = get_rule_match(start_line=5, end_line=6)  # 6 == 6\n5 == 5\n
+    matches = RuleMatchSet(double_eqeq_rule)
+    matches.update([match_1, match_2])
+    matches = list(sorted(matches))
+    assert matches[0].start_line_hash != matches[0].end_line_hash
```
jbergler, line 160: Thanks for adding these checks since it doesn't look like snapshots have coverage of these.

cli/src/semgrep/rule_match.py
chmccreery: Good call, thanks for looking into this! Yeah, I just used it because that was what syntactic_id used, but you're right we have an opportunity here to update to something more current, I'll do that.

cli/tests/e2e/snapshots/test_ci/test_dryrun/results.txt
chmccreery: Sort of - it's an artifact of the test set-up being kind of incomplete. I could flesh out all of the tests to have real rules with real finding files, but that's a bit more of an undertaking. Since this test isn't testing specifically for these hashes, I downscoped that out, but happy to revisit if you disagree.

jbergler: No, I'm happy with that - just wanted to make sure we weren't missing something in the unit tests you added.

cli/src/semgrep/rule_match.py
chmccreery: I don't think so? The old behavior, which I've preserved, just returns an empty string if the line was out of range 🤷 Or are you suggesting a more comprehensive change to that behavior?

edit: I guess it really comes down to what we think is more expected if the line doesn't exist. Should we pretend there's an empty line there (semantically equivalent to no line being there) or should we in that case show the lack-of-line by returning None? I don't have strong feelings either way.

jbergler: I think I'm still in ruby land too much of the time, looks like what would return null there raises an error here so this is fine.

```
>>> lines = []
>>> lines[0]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
```

jbergler (approved):

PR #7224, opened 2023-02-25
refactor: Convert AST_stat.ml to use generated visitor

nmote:
This changes the stats somewhat. This is because previously we used the matching visitor which generates various extra nodes for matching purposes. I split this up into two commits so as to demonstrate that this is why the numbers have changed. In the first commit, I change this code to ignore the nodes that are generated just for matching. This causes the numbers to change. In the second commit, I updated this code to use the vanilla autogenerated visitor, and the numbers do not change.

Test plan:

From `stats/parsing-stats`, run `./run-lang julia` (after running `make core && make install` at the repo root).

The untranslated node count is unchanged (149476) but the total node count drops from 9432006 to 9005661

PR checklist:

### DIFF
--- a/src/parsing/tests/AST_stat.ml
+++ b/src/parsing/tests/AST_stat.ml
@@ -9,55 +9,54 @@ type t = Parsing_stat.ast_stat
 let stat (ast : G.program) : t =
   let total_node_count = ref 0 in
   let untranslated_node_count = ref 0 in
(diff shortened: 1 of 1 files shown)


mjambon (approved):
I think it's great. I don't think we're using these stats in any dashboard, and even if did, the variations we care about are a lot more than 5%.

PR #7227, opened 2023-02-25
feat(rule syntax): explicit regex mvar introduction

brandonspark:
## What:
This PR introduces "explicit regex metavariable introduction", which means that capture group metavariables previously introduced via names like `$1`, `$2`, etc, are now required to be explicitly renamed to other metavariables. This looks like:
```
pattern-regex: "one (.*) two (.*)
metavars:
  - $1: $A
  - $2: $B
```
which introduces metavariables `$A` and `$B`, corresponding to teh first and second metavariables captured by the regex.

## Why:
`$1` and `$2`, while intuitive, do not actually match our official regex name schema, which specifies that metavariables cannot begin with a number.

This means that in code such as that for `metavariable-regex` and `metavariable-comparison`, these capture group metavariables pass by unseen, as they are not recognized as metavariables, and thus do not induce lookups. 

## How:
We simply require that rules specify renaming of these capture group metavariables to explicitly given names. This offers both a syntactic indication of the metavariables that have been introduced, as well as preventing invalid metavariables.

This requires a little bit of changing to the rule parsing code, to hold the dict that the given analyzed pattern originated from. This is because, while we may be dispatching on a `pattern-regex`, we may want to look at other entries that are in the dictionary, such as the renamed metavariables.

It's important to note that this change will make it such that previous rules that relied on this numerical-metavariable behavior will no longer work. I considered an alternative that would not alter the existing behavior, but I think the fact that these rules will throw the same error for unscoped metavariables is fairly indicative. Plus, I don't think there are really any solutions that will let us keep old rules unchanged, as this just straight-up didn't work, and is incompatible with our metavariable criteria.

## Alternatives considered:
I considered something like:
```
pattern-regex: "one (.*) two (.*)"
$1: $A
$2: $B
```

However, this causes major pains with the JSON schema. This was the original solution I put up in this PR, but it's very difficult in JSON schema to allow the case where `pattern-regex` is the top-level pattern, and the renames exist among the other top-level keys like `id`, `languages`, and `severity`. This is because metavariable names are not pre-specified, so you need to use a `patternProperties`, which will change the error message if you give an invalid rule. I found it was easier to just put it under an additional key.

## Test plan:
`make test`

I also introduced some rule parsing changes, in particular more `find` calls per pattern, and I wanted to make sure this didn't introduce a significant performance regression. I ran `sc -dump_rule default.yaml`, where `default.yaml` contains the contents of `p/default`, to see what the time benchmarks were. I got, with these changes:

10.82 secs
10.86 secs
10.72 secs

On `develop`, with the same parameters, I got:

11.27 secs
10.80 secs
10.79 secs

So, seems the same.

Error handling for bad name:
image

Using an un-renamed metavariable:
image

Renamed to a non-string:
image

Semgrep `--validate` without renaming:
image

Semgrep `--validate` with renaming:
image

Using Semgrep (with rule schema):
image

Using Semgrep with new syntax (with rule schema):
image

Closes PA-2514

PR checklist:


cli/tests/e2e/rules/regex-nosemgrep.yaml
```diff
@@ -2,6 +2,10 @@ rules:
   - id: detected-aws-account-id
     pattern-regex: |-
       ("|')?(AWS|aws|Aws)?_?(ACCOUNT|account|Account)_?(ID|id|Id)?("|')?\s*(:|=>|=)\s*("|')?[0-9]{12}("|')?
+    metavars:
```
aryx, line 5: not a fan of adding yet another toplevel field. Seems like it should be under pattern-regex:

src/core/Xpattern.ml
```diff
@@ -25,13 +25,26 @@ type compiled_regexp = Regexp_engine.t [@@deriving show, eq]
 type regexp_string = string [@@deriving show, eq]
 (* see the NOTE "Regexp" below for the need to have this type *)
 
+type regexp_xpattern = regexp_string * (string * string) list
```
aryx: why not int * Metavariable.mvar then?

cli/tests/e2e/rules/regex-nosemgrep.yaml
brandonspark: you'd have to add yet another subsidiary key within the `pattern-regex` itself, though. like:

```
pattern-regex:
  regex: ".*"
  renames: ...
```

this seems quite disruptive to me, what do you think?

IagoAbal:
> Why

If this was just a name convention thing I believe we could fix it by just accepting `$\d+` as a valid metavariable. Or using `$A`, `$B`, etc instead. The bigger problem is that they introduce metavariables implicitly, and writing multiple `pattern-regex` with capture groups can strangely produce no findings because the `$1` of one of them does not unify with the `$1` of the other.

cli/tests/e2e/rules/regex-nosemgrep.yaml
IagoAbal: IMO it's fine if we only require this new syntax when you want to assign metavariables to capture groups.

cli/tests/e2e/rules/regex-nosemgrep.yaml
```diff
@@ -2,6 +2,10 @@ rules:
...
+      - $1: $SHOULDNTEXIST
+      - $2: $AWS
```
IagoAbal: I would just write this as `- 2: $AWS`. We're mapping capture group 2 to metavariable `$AWS`. The idea is that there will no be any metavariable `$2`.

src/core/Xpattern.ml
brandonspark: good point!

aryx:
Also I question a bit the whole thing. We could use https://www.regular-expressions.info/named.html

aryx:
Also I question a bit the whole thing. We could use https://www.regular-expressions.info/named.html

mjambon:
> Also I question a bit the whole thing. We could use https://www.regular-expressions.info/named.html

We would have to restrict the name of the capture to an all-caps identifier compatible with metavariables e.g.
```
one (?<ONE>.*) two (?<TWO>.*)
```
If someone writes
```
one (?<one>.*) two (?<two>.*)
```
the regexp will match but we shouldn't make `$one` and `$two` available as metavariables because we don't allow lowercase metavariables elsewhere.

We'd also need to report double definitions as an error:
```
one (?<foo>.*) two (?<foo>.*)
```
^ this is an error in PCRE, at least in the version I use:
```ocaml
# Pcre.extract_all ~pat:{|one (?<foo>.*) two (?<foo>.*)|} {|one a two b|};;
Exception:
Pcre.Error (Pcre.BadPattern ("two named subpatterns have the same name", 25)).
```
Instead, the user perhaps intended to use a back reference:
```ocaml
# Pcre.extract_all ~pat:{|one (?<foo>.*) two \k<foo>|} {|one aa two aa|};;
- : string array array = [|[|"one aa two aa"; "aa"|]|]
```

Note that we have an OCaml parser for the PCRE syntax that we can use to analyze user-specified regexps and produce better error messages than the default runtime error messages from PCRE.

brandonspark:
@aryx It seems the `pcre-ocaml` library only allows you to query for named substrings if you already know
the name of the capture group:
https://mmottl.github.io/pcre-ocaml/api/pcre/Pcre/index.html#val-get_named_substring

Moreover, the implementation looks like it can only do this through an opaque, external function which obtains
the index in the array by the capture group name, so I don't think this information is even stored in the OCaml side:
image

What do you think we should do?

brandonspark:
@aryx poke again about my above comment ^

aryx:
Like martin suggested, we have a regexp parser so we can extract the named group first.

aryx:
I talked with @IagoAbal and I think we have a plan. Rename the $1, $2, to $_1, $_2. Support also $_0 to match the whole thing, so that can also be used in metavariable-comparison. And then later analyze the regexp to extract the named group names, and convert that too to metavariables. That way no need for those metavars: $1 : $XXX extra config. We just reuse existing mechanisms.

aryx (changes requested):
Do not introduce new syntax. Let's reuse the existing mechanism in PCREs to name groups.

aryx:
@brandonspark use languages/regexp/AST.mli if you need to analyze the regexp.

aryx:
Closing in favor of the the discussion we had with Iago.

PR #7229, opened 2023-02-25
AST_to_IL: Fix translation of side-effects in Bool exprs

### LINKED ISSUE #7199: Constant propagation fails to consider branches and conditionals
**Describe the bug**

In what follows, both methods are functionally the same.

```java
class A {
    
    public void method(boolean condition){
      var str = "hello";
      if(condition){
        if("goodbye".equals(str = "goodbye"))
	  X.test(str);
      }
      // ruleid: untitled_rule
      X.test(str);
    }

    public void method2(boolean condition){
      var str = "hello";
      if(condition && "goodbye".equals(str = "goodbye")){
        X.test(str);
      }
      // ruleid: untitled_rule
      X.test(str);
    }
}
```

~~- The pattern `X.test("hello")` fails to match the second call in both methods. It is possible if `condition` is `false`.~~
~~- The pattern `X.test("goodbye")` fails to match the second call in the first method. It is possible if `condition` is `true`.~~
See @IagoAbal's comment (https://github.com/returntocorp/semgrep/issues/7199#issuecomment-1440347945) below. The behavior for `X.test("hello")` is correct but still wrong for `X.test("goodbye")`, as it should not match the second call on `method2`.

**To Reproduce**
[https://semgrep.dev/s/elDE](https://semgrep.dev/s/elDE)

**Expected behavior**
~~`X.test("hello")` should match the second `X.test(str)` call and `X.test("goodbye")` should match all calls.~~
`X.test("goodbye")` should not match the second call on `method2`.

IagoAbal:
We were interpreting `A && x = 42` as `x = 42; A && x`, but the side-effects of `x = 42` only take effect if `A` holds.

We fix this by guarding these side-effects, e.g. interpreting `A && x = 42` as `if (A) x = 42; A && x`.
    
Unfortunately, this will have some side-effects on sym-prop, because we kill sym-prop values at JOIN points, so any sym-prop values are lost after an if-statement. Sym-prop is experimental so that's OK.

Closes #7199

test plan:
make test # tests added

PR checklist:

### DIFF
--- a/cli/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awarelog4shell.yaml-dependency_awarelog4j/results.txt
+++ b/cli/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awarelog4shell.yaml-dependency_awarelog4j/results.txt
@@ -39,19 +39,6 @@ SEMGREP_USER_AGENT_APPEND="pytest" SEMGREP_SETTINGS_FILE="<MASKED>" SEMGREP_VERS
               "line": 33,
               "offset": 1035
             },
-            "propagated_value": {
-              "svalue_abstract_content": "req.getParameter(\"uname\")",
-              "svalue_end": {
-                "col": 52,
-                "line": 19,
-                "offset": 562
-              },
-              "svalue_start": {
-                "col": 27,
-                "line": 19,
-                "offset": 537
-              }
-            },
             "start": {
               "col": 26,
               "line": 33,
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -190,6 +190,14 @@ let pop_stmts env =
   env.stmts := [];
   xs
 
+let with_pre_stmts env f =
+  let saved_stmts = !(env.stmts) in
+  env.stmts := [];
+  let r = f env in
+  let f_stmts = pop_stmts env in
+  env.stmts := saved_stmts;
+  (f_stmts, r)
+
 let bracket_keep f (t1, x, t2) = (t1, f x, t2)
 
 let ident_of_entity_opt ent =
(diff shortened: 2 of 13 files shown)


cli/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awarelog4shell.yaml-dependency_awarelog4j/results.txt
```diff
@@ -39,19 +39,6 @@ SEMGREP_USER_AGENT_APPEND="pytest" SEMGREP_SETTINGS_FILE="<MASKED>" SEMGREP_VERS
               "line": 33,
               "offset": 1035
             },
-            "propagated_value": {
```
aryx, line 42: this seems unrelated. cc @mmcqd

src/analyzing/AST_to_IL.ml
```diff
@@ -445,8 +453,14 @@ and assign env lhs tok rhs_exp e_gen =
 and expr_aux env ?(void = false) e_gen =
```
aryx, line 456: could you explain what this void parameter is?

aryx (approved):

cli/tests/e2e/snapshots/test_dependency_aware_rules/test_dependency_aware_rules/rulesdependency_awarelog4shell.yaml-dependency_awarelog4j/results.txt
IagoAbal: It is related to this change but unrelated to SSC:

> Unfortunately, this will have some side-effects on sym-prop, because we kill sym-prop values at JOIN points, so any sym-prop values are lost after an if-statement. Sym-prop is experimental so that's OK.

IagoAbal: See also this test I added: https://github.com/returntocorp/semgrep/blob/38baaabc8d7d45ba2cf3d5c0e10ef2777c7a7826/tests/rules/sym_prop_no_merge4.java

src/analyzing/AST_to_IL.ml
IagoAbal: Will add a comment.

PR #7188, opened 2023-02-18
fix(rule parsing perf): initial changes to speed up rule parsing

emjin:
Rule parsing is slow because we create and remove a tmp file for each pattern we parse. This is entirely unnecessary; the Lexer class supports lexing strings. While rule parsing is a relatively small amount of time compared to full scans, it is relatively large compared to diff scans or scans of small repos, which affects the feel of semgrep. Also, I am just morally against creating unnecessary tmp files.

TODO: ruby rule parsing is still really slow because of the dyp parser. We would need to switch to tree-sitter. I will do that in a separate PR. Alternatively, we can filter out rules that aren't relevant to the files passed (which I would also do in a separate PR).

Test plan: make test, also time with running p/default on a single file

On p/python this dropped runtime with a small test from 2.4 seconds to 0.5 seconds

PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure on any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #7236, opened 2023-02-27
feat(perf): Only parse rules that will actually be run

emjin:
Pattern parsing--specifically, Ruby pattern parsing--can be slow. We currently have over 1000 rules in p/default, many of which are not run depending on the languages in the repo. We can skip the pattern parsing stage for those rules.

In conjunction with https://github.com/returntocorp/semgrep/pull/7188, reduces rule parsing when run on the semgrep repo  to around a second, down from 13 seconds with just 7188. It is very effective because Ruby pattern parsing takes the vast majority of the time. Any repo without Ruby should see a similar result.

Curently not implemented for osemgrep.

Test plan: make test

PR checklist:

### DIFF
--- a/src/core/Rule.ml
+++ b/src/core/Rule.ml
@@ -275,27 +275,12 @@ type extract_rule = extract_mode rule_info [@@deriving show]
 
 (* the general type *)
 type rule = mode rule_info [@@deriving show]
+type mode_type = Match | Extract [@@deriving show]
 
 (* aliases *)
 type t = rule [@@deriving show]
 type rules = rule list [@@deriving show]
 type hrules = (rule_id, t) Hashtbl.t
-
-(*****************************************************************************)
-(* Helpers *)
-(*****************************************************************************)
-
-let hrules_of_rules (rules : t list) : hrules =
-  rules |> Common.map (fun r -> (fst r.id, r)) |> Common.hash_of_list
-
-let partition_rules (rules : rules) :
-    search_rule list * taint_rule list * extract_rule list =
-  rules
-  |> Common.partition_either3 (fun r ->
-         match r.mode with
-         | `Search _ as s -> Left3 { r with mode = s }
-         | `Taint _ as t -> Middle3 { r with mode = t }
-         | `Extract _ as e -> Right3 { r with mode = e })
 
 (*****************************************************************************)
 (* Error Management *)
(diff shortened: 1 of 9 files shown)


emjin:
```
============================[ summary ]============================
Total time: 19.4735s Config time: 2.8022s Core time: 16.6258s

Semgrep-core time:
Total CPU time: 60.0635s  File parse time: 47.3807s  Rule parse time: 0.9831s  Match time: 18.4821s
Slowest 5/1497 files
libs/spacegrep/examples/my_first_calculator.py.doc (753KB): 5.301s (4.488s to parse)
src/engine/Entropy_data.ml                         (329KB): 2.722s (3.033s to parse)
perf/perf-matching/input/l5000.java                (487KB): 0.620s (2.632s to parse)
perf/perf-matching/input/l10000.js                 (203KB): 11.310s (2.601s to parse)
libs/commons/common2.ml                            (164KB): 0.756s (1.492s to parse)
Slowest 5 rules to match
...skip-tls-verify-service.skip-tls-verify-service:         0.000s
...skip-tls-verify-cluster.skip-tls-verify-cluster:         0.000s
...nfinement-disabled.seccomp-confinement-disabled:         0.000s
...rnetes.security.run-as-non-root.run-as-non-root:         0.000s
...urity.privileged-container.privileged-container:         0.000s

Analyzed: 13 bash files ( 12KB in 0.018 seconds)
          4 c files ( 21KB in 0.020 seconds)
          3 dockerfile files (  9KB in 0.011 seconds)
          420 generic files (  3MB in 10.981 seconds)
          1 go files ( 53KB in 0.012 seconds)
          1 java files (487KB in 0.620 seconds)
          7 js files (225KB in 12.122 seconds)
          9 json files (489KB in 0.050 seconds)
          1 jsonnet files (  4KB in 0.001 seconds)
          766 ocaml files (  6MB in 10.754 seconds)
          235 python files (  1MB in 24.220 seconds)
          37 yaml files (133KB in 0.272 seconds)
Errors:   17 files with errors, see output before the results for details or run with --strict
          PartialParsing (17 files)
                                                                                                 
                                                                                                 
┌──────────────┐                                                                                 
│ Scan Summary │                                                                                 
└──────────────┘                                                                                 
                                                                                                 
  Partially scanned: 17 files only partially analyzed due to a parsing or internal Semgrep error
  Scan skipped: 20 files larger than 1.0 MB, 4709 files matching .semgrepignore patterns
  For a full list of skipped files, run semgrep with the --verbose flag.

Ran 1416 rules on 1497 files: 46 findings.
semgrep --config p/default . --time  130.68s user 9.11s system 414% cpu 33.729 total
```

emjin:
Fun fact running `semgrep --config p/default .` is actually faster than downloading `p/default` as `default.yaml` and running  with the downloaded config, because parsing yaml rules takes much longer than downloading rules (which are downloaded as json). Actually, parsing yaml on the python side takes about as much time as parsing the rules including the patterns on the OCaml side did before my improvements.

src/core/Rule.ml
```diff
@@ -275,28 +275,13 @@ type extract_rule = extract_mode rule_info [@@deriving show]
 
 (* the general type *)
 type rule = mode rule_info [@@deriving show]
+type mode_type = Match | Extract [@@deriving show]
```
aryx, line 278: explain why it's mostly a copy of mode, but without Taint and without arguments.

src/core/Rule.ml
```diff
@@ -336,6 +321,72 @@ type error =
 (* can't use Error because it's used for severity *)
 exception Err of error
 
+(*****************************************************************************)
+(* Lazy parsing *)
+(*****************************************************************************)
+
+type rule_result = Rule of rule | InvalidRuleError of invalid_rule_error
```
aryx, line 328: Did you try instead to just make the parsing of the pattern lazy? so have Xpattern.t defined as lazy?

src/osemgrep/cli_scan/Rule_fetching.ml
```diff
@@ -35,7 +35,7 @@ type rules_and_origin = {
   (* TODO? put a config_id: string option? or config prefix? or
    * compute it later based on the origin?
    *)
-  rules : Rule.rules;
+  rules : Rule.lazy_rule list;
   errors : Rule.invalid_rule_error list;
```
aryx, line 39: but if you don't really parse the rule, errors will always be empty?

src/core/Rule.ml
aryx: (I guess because now we parse rules lazily, but we need to differentiate early whether a rule is
extract or match/taint because for extract rules we need to evaluate them early.)

Maybe change Match to SearchOrTaint

aryx:
Very nice, 13s to 1s! 13x speedup!

src/core/Rule.ml
aryx: That way we don't need define a new rule_result type.

emjin: I did try that first but it seemed like a bigger change. At least, the way I tried it, I was changing the type of rule, but this meant that everywhere rules were processed I had to double check whether the pattern was lazy or not. I liked having an invariant that once the rules were passed down for matching, they were parsed in full.

src/osemgrep/cli_scan/Rule_fetching.ml
emjin: You still can have errors from parsing `id`, `languages`, and `mode`. This is the ugliest thing about my current scheme

src/core/Rule.ml
aryx: I would expect a smaller change. Just change the Xpattern.t in Rule.ml to Xpattern.t lazy.

emjin: But then what about everywhere the xpattern is used? Maybe I'm not understanding something about the type

aryx: I thought there would not be many places where we use Xpattern.t, but indeed at those places
you would need to use Lazy.force or do like Nat does and use (lazy X) in the pattern to force the evaluation
directly from the pattern.

emjin: Ah yeah there are a decent number of places. There's Analyze_rule, Translate_rule, Check_rule, and then at least four places in Match_search_mode. I didn't follow it all the way down the chain to see how many places I would need to change for sure, so there could be more. Given how many places we work with patterns I felt that this was nicer. I don't feel very strongly though, so I can change it if you do

aryx:
superseded by https://github.com/returntocorp/semgrep/pull/7270

PR #7215, opened 2023-02-23
tainting: Make identification of sinks more precise

IagoAbal:
Previously in 1eb330d6887 we did not consider that when adding a better sink range to Top_sinks, there could be not just one but several ranges that were subsumed by the new one.

Closes PA-2509

Fixes: 1eb330d6887 ("tainting: Make identification of sinks more precise (#6640)")

test plan:
make test # added two more tests

PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure on any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #7245, opened 2023-02-28
tainting: Make identification of sinks more precise

IagoAbal:
Folllows: 4020470dd7c ("tainting: Make identification of sinks more precise (#7215)")

Closes PA-2477
Closes PA-2571

test plan:
make test # tests added
Run p/default on django, redash, mypy, angular.js, and bootstrap, and check that perf and number of findings don't change. 

PR checklist:

### DIFF
--- a/changelog.d/pa-2477.changed
+++ b/changelog.d/pa-2477.changed
@@ -1,0 +1,15 @@
+taint-mode: Historically, the matching of taint sinks has been somewhat imprecise.
+For example, `sink(ok if tainted else ok)` was flagged. Recently, we made sink-
+matching more precise for sinks like `sink(...)` declaring that any argument of
+a given function is a sink. Now we make it more precise when specific arguments of
+a function are sinks, like:
+
+```yaml
+pattern-sinks:
+- patterns:
+  - pattern: sink($X, ...)
+  - focus-metavariable: $X
+```
+
+So `sink(ok1 if tainted else ok2)`, `sink(not_a_propagator(tainted))`, and
+`sink(some_array[tainted])`, will not be reported as findings.
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -158,6 +158,11 @@ module Top_sinks = struct
 
   let empty = S.empty
 
+  let _debug sinks =
+    sinks |> S.elements
+    |> Common.map (fun m -> Range.content_at_range m.spec_pm.file m.range)
+    |> String.concat " ; "
+
   let rec add m' sinks =
     (* We check if we have another match for the *same* sink specification
      * (i.e., same 'sink_id'), and if so we must keep the best match and drop
@@ -251,7 +256,7 @@ let orig_is_source config orig = config.is_source (any_of_orig orig)
 let orig_is_sanitized config orig = config.is_sanitizer (any_of_orig orig)
 let orig_is_sink config orig = config.is_sink (any_of_orig orig)
 
-let orig_is_best_sink env orig =
+let orig_is_best_sink env orig : R.taint_sink tmatch list =
   orig_is_sink env.config orig
   |> List.filter (Top_sinks.is_best_match env.top_sinks)
 
@@ -279,9 +284,24 @@ let report_findings env findings =
 let top_level_sinks_in_nodes config flow =
   flow.CFG.reachable |> CFG.NodeiSet.to_seq
   |> Stdcompat.Seq.concat_map (fun ni ->
+         let origs_of_args args =
+           Seq.map (fun a -> (IL_helpers.exp_of_arg a).eorig) (List.to_seq args)
+         in
          let node = flow.CFG.graph#nodes#assoc ni in
          match node.n with
-         | NInstr instr -> orig_is_sink config instr.iorig |> List.to_seq
+         | NInstr instr ->
+             let origs : orig Seq.t =
+               Seq.cons instr.iorig
+                 (match instr.i with
+                 | Call (_, c, args) -> Seq.cons c.eorig (origs_of_args args)
+                 | CallSpecial (_, _, args) -> origs_of_args args
+                 | Assign (_, e) -> List.to_seq [ e.eorig ]
+                 | AssignAnon _
+                 | FixmeInstr _ ->
+                     Seq.empty)
+             in
+             origs
+             |> Seq.concat_map (fun o -> orig_is_sink config o |> List.to_seq)
          | NCond (_, exp)
          | NReturn (_, exp)
          | NThrow (_, exp) ->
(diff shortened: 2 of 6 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -793,8 +842,7 @@ and check_tainted_expr env exp : Taints.t * Lval_env.t =
         handle_taint_propagators { env with lval_env } (`Exp exp) taints
       in
       let taints = Taints.union taints taints_propagated in
-      let findings = findings_of_tainted_sinks env taints sinks in
-      report_findings env findings;
+      check_orig_if_sink env exp.eorig taints;
```
brandonspark, line 845: i like this! it confused me before when both relevant lines were relatively distributed

src/tainting/Dataflow_tainting.ml
```diff
@@ -298,6 +318,27 @@ let top_level_sinks_in_nodes config flow =
...
+(* Checks whether the sink corresponds has the shape
+ *
+ *     patterns:
+ *     - pattern: <func>(<args>)
+ *     - focus-metavariable: $MVAR
+ *
+ * In which case we know that the function call itself is not the sink, but
+ * either the <func> or one (or more) of the <args>.
+ *)
+let is_func_sink_with_focus taint_sink =
```
brandonspark, line 330: i surmise that a pattern like
```
patterns:
- pattern: [$X, ...]
- focus-metavariable: $X
```
would still flag false positives for something like `[ok1 if tainted else ok2]`, due to not being special-cased the same way as here. i am not in love with the fact that we have to do this special casing at all, though i recognize that this would be a valid solution for the problem you're trying to solve.

do you this may present an issue? will there be a great deal of boilerplate in the future needing to add in further special cases?

tests/rules/taint_best_fit_sink4.yaml
```diff
@@ -0,0 +1,18 @@
...
+    mode: taint
+    pattern-sanitizers:
+      - pattern: sanitizer(...)
+      - patterns:
+          - pattern: $S
+          - pattern: $ANY . $S
+    pattern-sinks:
+      - patterns:
+          - focus-metavariable: $SOURCE
+          - pattern: sink($SOURCE, ...)
```
brandonspark, line 15: it would be nice to add some examples of sinks that actually have multiple arguments, for sanity's sake

brandonspark (approved):
otherwise, i see no reason to block this. but i am wary of whether or not this is a bandaid that will need to be reapplied many times in the future.

src/tainting/Dataflow_tainting.ml
```diff
@@ -279,9 +284,24 @@ let report_findings env findings =
 let top_level_sinks_in_nodes config flow =
   flow.CFG.reachable |> CFG.NodeiSet.to_seq
   |> Stdcompat.Seq.concat_map (fun ni ->
+         let origs_of_args args =
+           Seq.map (fun a -> (IL_helpers.exp_of_arg a).eorig) (List.to_seq args)
+         in
          let node = flow.CFG.graph#nodes#assoc ni in
          match node.n with
-         | NInstr instr -> orig_is_sink config instr.iorig |> List.to_seq
+         | NInstr instr ->
```
brandonspark, line 292: maybe also a comment here explaining what these are for

src/tainting/Dataflow_tainting.ml
IagoAbal: Why the `Top_sinks` trick is done is explained (hopefully) in `NOTE "Top sinks"`. Given that,  I'm trying to cover the cases that cause most of the problems right now, while trying not to break anything else. It's true that there are other sinks using `focus-metavariable` that would not benefit from this, but I'm not too worried about it for now. If it's asked for, we can easily add it. But if more cases are requested then I agree we need to generalize this instead.

changelog.d/pa-2477.changed
```diff
@@ -0,0 +1,15 @@
+taint-mode: Historically, the matching of taint sinks has been somewhat imprecise.
+For example, `sink(ok if tainted else ok)` was flagged. Recently, we made sink-
```
aryx, line 2: and it should not be flagged because taint in a condition does not matter, because the result will always
be 'ok' ?

changelog.d/pa-2477.changed
```diff
@@ -0,0 +1,15 @@
...
+matching more precise for sinks like `sink(...)` declaring that any argument of
+a given function is a sink. Now we make it more precise when specific arguments of
```
aryx, line 4: and this is more precise? to declare any argument is a sink?

changelog.d/pa-2477.changed
```diff
@@ -0,0 +1,15 @@
...
+a function are sinks, like:
+
+```yaml
+pattern-sinks:
+- patterns:
+  - pattern: sink($X, ...)
+  - focus-metavariable: $X
+```
+
+So `sink(ok1 if tainted else ok2)`, `sink(not_a_propagator(tainted))`, and
```
aryx, line 14: Tbh I don't understand it. It's too subtle for me.

src/tainting/Dataflow_tainting.ml
IagoAbal: Well, actually the case you wrote should be covered... this should only affect function calls.

aryx (approved):

src/tainting/Dataflow_tainting.ml
brandonspark: oh i know this PR only affects function calls, i just meant someone might write something like this:
https://semgrep.dev/s/B81R

and expect no findings

changelog.d/pa-2477.changed
IagoAbal: Correct.

changelog.d/pa-2477.changed
IagoAbal: That's fair, if you are not used to writing taint rules it can be subtle. I'll try to make it more clear in the docs the next time I give them a pass.

changelog.d/pa-2477.changed
IagoAbal: Hmm, well, we had fixed the aforementioned issue for sink specifications like `sink(...)`. But if you declared a sink via `focus-metavariable` like below then you still had the same problem with `sink(ok if tainted else ok)`.

src/tainting/Dataflow_tainting.ml
IagoAbal: I'll take a look if it's easy to add those cases too, although I haven't got any complain except for the function case.

IagoAbal: https://github.com/returntocorp/semgrep/pull/7466

PR #7247, opened 2023-02-28
fix: brew nightly tests failing

spencerdrak:
Nightly homebrew checks began failing due to an unknown error, but this may be related [to changes to the homebrew API discussed here](https://github.com/orgs/Homebrew/discussions/4150) - we set the correct environment variables and then found that `brew update` was not properly updating our formula unless the same env var was set on the `brew update` step. There's been a lot of discussion about this flag/var recently, since it's newly added.

This should resolve issues moving forward, since it will force all updates/installs/tests to use the source and never the homebrew API.

Test Plan:
- successful run here: https://github.com/returntocorp/semgrep/actions/runs/4297155703/jobs/7489775698

PR checklist:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -52,18 +52,18 @@ jobs:
     with:
       dry-run: true
 
-  # notify-failure:
-  #   needs: [brew-build, release-dry-run]
-  #   name: Notify of Failure
-  #   runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


aryx (approved):

aryx:
would nice nice to add a link to the relevant discussion about this env variable in the workflow file.

cgdolan (approved):

PR #7249, opened 2023-02-28
fix(cli): use percentage for progress bar estimation

brandonspark:
## What:
This PR changes the progress bar logic to be more streamlined, using a percentage measure of progress as opposed to an abstract "tasks", as well as exposing the `update_cli_progress` function so that dot-printing logic is all in one place. Companion to https://github.com/returntocorp/semgrep-proprietary/pull/581

## Why:
The progress bar, as it currently stands, uses "tasks" as its measurement for how far along Semgrep is. Not only is this rather unclear, it's unstandardized between Pro and OSS, since Pro will report 3 * # files as its number of tasks, whereas OSS will report # files as the number of tasks.

## How:
We standardize the denominator of the progress bar at 3 times the # of files, and report a percentage for the CLI's progress. However, we still keep the convention that we expect to see 3x the number of dots from Pro as we do from OSS. We achieve this by simply counting each dot from Pro as 1 of a progress point, whereas OSS dots are three.

## Test plan:
https://asciinema.org/a/3CXH6jEr0uBHfB3FsfC58W2wr

PR checklist:


cli/src/semgrep/core_runner.py
```diff
@@ -269,10 +270,21 @@ async def _core_stdout_processor(self, stream: asyncio.StreamReader) -> None:
...
+                # - once when finishing phase 4, name resolution, on that target
+                # - once when finishing phase 5, taint configs, on that target
+                # - once when finishing analysis on that target as usual
+                #
+                # However, for regular OSS Semgrep, we only print one dot per
+                # target, that being the last bullet point listed above.
+                #
+                # So a dot counts as 0.33 files progress if running Pro, but 1
+                # file's progress if running the OSS engine.
+                advanced_targets = 1 / 3 if self._engine_type.is_interfile else 1
```
underyx: CleanShot 2023-03-01 at 12 06 18@2x

the floating point drift from incrementing by 1/3 is making me uncomfortable; could we just always use integers? set the total to three times the current value, and advance by 3 when not interfile

cli/src/semgrep/core_runner.py
```diff
@@ -435,7 +447,7 @@ def execute(self) -> int:
         with Progress(
             BarColumn(),
             MofNCompleteColumn(),
-            TextColumn("tasks"),
+            TextColumn("files"),
```
underyx: Wanna make sure you're aware why this was called tasks instead of files.

```terminal
$ echo 1 > one.py
$ semgrep --config=auto one.py
    <multilang> |  50 rules × 2 files
         python | 355 rules × 1 file

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0/3 tasks -:--:--
```

The progress bar says there are '3 tasks' total now and I guess it would say '3 files' after your PR. I think it would be very confusing if Semgrep claimed it's scanning 3 files when you explicitly passed only one.

*Note: I see there's a bug now where we say 2 multilang files because it seems we add together the number of files for regex and for generic rules.*

brandonspark: ooh, that makes sense. this work was based off of what @dlukeomalley saw in a dry run, where "tasks" was somewhat confusing. but it looks like there's a reason for it after all.

@dlukeomalley what do you think? is "tasks" still sufficient, or would a more-specific term like "targets" (which hopefully is understood not to correlate to files 1-1) be better?

dlukeomalley: Does it need to say anything and/or show a concrete numerator/denominator? What if instead if was a percentage?

brandonspark: @dlukeomalley Yes, it can be a percentage! See the revised PR description. What do you think?

cli/src/semgrep/core_runner.py
brandonspark: that's a fair point, fixed!

cli/src/semgrep/core_runner.py
```diff
@@ -1087,14 +1098,10 @@ def _run_rules_direct_to_semgrep_core_helper(
...
-                # We expect to see 3 dots for each target, when running interfile analysis:
-                # - once when finishing phase 4, name resolution, on that target
-                # - once when finishing phase 5, taint configs, on that target
-                # - once when finishing analysis on that target as usual
-                cmd,
-                plan.num_targets * 3 if engine.is_interfile else plan.num_targets,
-            )
+            # Multiplied by three, because we have three places in Pro Engine to
+            # report progress, versus one for OSS Engine. This way, we don't get
+            # floating-point drift.
```
underyx:
```suggestion
            # Multiplied by three, because we have three places in Pro Engine to
            # report progress, versus one for OSS Engine.
```

I think this solution is logical enough that it doesn't need to be compared to incrementing by 1/3. In fact, I think bringing up floating-point drift out of nowhere (without introducing the alternative solution) is more confusing than not mentioning it.

cli/src/semgrep/core_runner.py
```diff
@@ -1262,7 +1269,9 @@ def validate_configs(self, configs: Tuple[str, ...]) -> Sequence[SemgrepError]:
                 *configs,
             ]
 
-            runner = StreamingSemgrepCore(cmd, 1)  # only scanning combined rules
+            runner = StreamingSemgrepCore(
+                cmd, 1, self._engine_type
+            )  # only scanning combined rules
```
underyx:
```suggestion
            # only scanning combined rules
            runner = StreamingSemgrepCore(cmd, 1, self._engine_type)
```

changelog.d/pa-2526.fixed
```diff
@@ -0,0 +1,5 @@
+CLI: Made the progress bar reflect Semgrep's actual execution better.
+Instead of abstract "tasks", the progress bar now measures progress in terms of
+how many files Semgrep has processed.
+
+Also, the progress bar should jump less with the Pro Engine.
```
underyx:
```suggestion
CLI: The scan progress bar now shows progress with higher granularity, and has fewer big jumps when using the Pro Engine.

The abstract unit of 'tasks' has been removed, and now only a percentage number will be displayed.
```

underyx (approved):

cli/src/semgrep/core_runner.py
brandonspark: ooh, good point, i wasnt thinking about that

PR #7257, opened 2023-03-01
Add automatic indentation to semgrep CLI output

underyx:
This PR makes output formatting look a bit worse; other parts of #7251 adapt various output lines to rely on auto-indentation.

PR checklist:


underyx:
cc @aryx

cli/src/semgrep/console.py
```diff
@@ -25,18 +27,81 @@ class Title:
...
+
+        return None
+
+    def print(self, *args: Any, deindent: int = 0, **kwargs: Any) -> None:
+        indent = 0
+
+        for arg in args:
+            title = self.extract_title(arg)
+            if title is not None:
+                self.active_title = title
```
brendongo, line 93: what does setting self.active_title do?

brendongo: Looks like this is handling printing titles but not sure how. Would it also be more correct to special case printing strings so console correctly handles any other future objects passed in (aside from Title class)

brendongo: Oh self.active_title is var we created nvm

cli/src/semgrep/console.py
```diff
@@ -25,18 +27,81 @@ class Title:
...
+    @staticmethod
+    def extract_title(current: Optional[RenderableType]) -> Optional[Title]:
+        while isinstance(current, Title) or hasattr(current, "renderable"):
+            if isinstance(current, Title):
+                return current
+            current = getattr(current, "renderable", None)
+
+        return None
+
+    def print(self, *args: Any, deindent: int = 0, **kwargs: Any) -> None:
```
brendongo, line 81: Can you add quick comment explaining how you got this to work. We keep around reference to last title in active_title and figure out indentation level from active_title object

brendongo (approved):

cli/src/semgrep/console.py
```diff
@@ -25,18 +27,81 @@ class Title:
...
+        else:
+            indent = self.auto_indent_size
+
+        indent = max(0, indent - deindent)
+
+        super().print(Padding.indent(Group(*args), indent), **kwargs)
+
+
+console = AutoIndentingConsole(highlighter=None)
+console.width = min(console.width, 120)
```
brendongo: Do you want these global?

PR #7266, opened 2023-03-02
Parametrize engine_type tests

underyx:
This method of defining tests is nice cause:

- Fewer lines of code
- The test cases are easier to read as they're basically a table
- Lower likelihood of an early line of code in the test function accidentally & unexpectedly changing a later line's behavior
- pytest would run the earlier version of the test function only up until the first AssertionError, while this parametrized version will always run once per case, so test reports in CI show a more complete picture of what's failing and what's broken, it's easier to identify the pattern and figure out where you need to look to fix the code


emjin (approved):
Thank you! I figured there was some nice pythonic way to do this but I was already struggling with understanding the mocker

PR #7270, opened 2023-03-02
Perf: parse the semgrep patterns lazily

aryx:
This is an alternative to
https://github.com/returntocorp/semgrep/pull/7236

This is arguably simpler and also work for osemgrep (for free).

test plan:
I tried in regular semgrep and I go from 13s to 5s
(most of those 5s are because of the network I think)

In osemgrep:
```
$ xx scan --config p/default --profile src/osemgrep/targeting/
---------------------
profiling result
---------------------
CLI.dispatch_subcommand                  :      7.061 sec          1 count
Rule_fetching.rules_from_rules_source    :      6.979 sec          1 count
Network.get                              :      6.720 sec          1 count
Parse_rule.parse_and_filter_invalid_rules :      0.258 sec          1 count
Parse_with_caching.cache_computation     :      0.054 sec         48 count
Parse_target.parse_and_resolve_name      :      0.054 sec         48 count
Parsing_helpers.full_charpos_to_pos_large :      0.029 sec         49 count
real_rule:ocaml.lang.portability.slash-tmp.not-portable-tmp-string :      0.014 sec         24 count
Match_search_mode.matches_of_xpatterns   :      0.014 sec         24 count
Match_patterns.check                     :      0.014 sec         24 count
Parse_target.just_parse_with_lang        :      0.013 sec         48 count
Match_patterns.match_e_e                 :      0.011 sec       4318
count
...
```

So parsing the whole p/default takes only 0.258s,
whether it was taking 8s before.

PR checklist:


aryx:
I've learned this 'lazy' in pattern trick from @nmote 
It's pretty nice.

aryx:
What do you think @emjin ? simpler no?
I need to fix the metachecking tests though.

src/core/Xpattern.ml
```diff
@@ -29,7 +29,11 @@ type regexp_string = string [@@deriving show, eq]
 type pattern_id = int [@@deriving show, eq]
 
 type xpattern_kind =
-  | Sem of Pattern.t * Lang.t (* language used for parsing the pattern *)
+  (* opti: parsing Semgrep patterns lazily improves speed significantly.
+   * Parsing 'p/default', which contains more than 1000 rules, including
+   * lots of Ruby rules goes from 13s to just 0.2s (Dyp GLR is pretty slow).
```
emjin: It's not actually that GLR is slow, it's that it takes a lot of time (0.02 s or so) to create the parser. This work is repeated for each pattern (in `p/default` there were ~600 Ruby patterns), so it adds up. Note that the relevant function is `pp ()` so it isn't doing anything special per pattern.

src/osemgrep/cli_scan/Rule_fetching.ml
```diff
@@ -110,6 +110,9 @@ let import_callback base str =
               *)
              Logs.debug (fun m ->
                  m "trying to download from %s" (Uri.to_string url));
+             (* TODO: ask for JSON in headers which improves performance
```
emjin, line 113: I wonder how much the difference is in OCaml. The downside of JSON is that we can't report nice error spans for rules. We can't already, but it would be cool if after the osemgrep port we could

emjin (approved):
Thanks! I didn't realize about `lazy` in pattern, that makes it much nicer than I was expecting.

I guess you need to force it for the metachecking tests

emjin:
Oh I forgot to do this in my PR but we can include a changelog entry that announces that there's a speed up in rule parsing

aryx:
@emjin what do you thing of the snapshot update https://github.com/returntocorp/semgrep/commit/3fb6a3f121733c7d9b27dc3b23e3f1b928a23f2c to make this PR pass?
Sounds legit? We used to fail quickly on parse errors in the ruleset, and now we actually parse the rules,
and report only error when we need the pattern in the rule. Looks like we still print the parse error in the snapshot
but the scanned: list is bigger because we actually parse files for the rules we can handle.

What do you think? Should we accept the new snapshot?

nmote:
> I've learned this 'lazy' in pattern trick from @nmote

It's just a constructor like any other!

emjin:
> What do you think? Should we accept the new snapshot?

I don't really like it. Specifically, the one where there's a pattern parse error but we also have the normal output and say "no results". I think that might cause some people to disregard the pattern parse error.

PR #7273, opened 2023-03-02
fix(extract): Run all rules of dest language on extracted target

emjin:
Semgrep file targeting relies on a relatively fragile system of recording the rules that should b e run for each target. This is a consequence of the current split between Python and OCaml.

Previously, the way that extract rules were separated out caused the numbering to become offset. This in turn meant that the wrong rules were being run on targets. Fixed by this PR

Possibly closes some other issues.

Test plan:
```
➜  extract git:(emma/pa-2591-fix-extract-mode) ✗ pwd
/Users/emma/workspace/semgrep/tests/extract
➜  extract git:(emma/pa-2591-fix-extract-mode) ✗ semgrep --config python_jupyter.yaml python_jupyter.ipynb
Scanning 1 file.

┌─────────┐
│ Results │
└─────────┘
Findings:

  python_jupyter.ipynb
     forbid-pickle
        found pickle

         31┆ "    pickle.dump(data, f)"

┌──────────────┐
│ Scan Summary │
└──────────────┘

Ran 2 rules on 1 file: 1 finding.
```
(This previously gave no findings)

PR checklist:

### DIFF
--- a/cli/tests/e2e/targets/extract/js_html_concat.html
+++ b/cli/tests/e2e/targets/extract/js_html_concat.html
@@ -1,0 +1,31 @@
+<!DOCTYPE html>
+<html>
+    <script>
+        // ruleid: get-elem-id
+        function myFunction() { document.getElementById("demo").innerHTML = "Paragraph changed."; }
+        var x = foo();
+        // ruleid: get-elem-id
+        function myFunction() { document.getElementById("demo").innerHTML = "Paragraph changed."; }
+        // ruleid: cross-tag-taint
+        bar(x);
+    </script>
+
+    <body>
+        <h2>Demo JavaScript in Body</h2>
+
+        <p id="demo">A Paragraph</p>
+
+        <button type="button" onclick="myFunction()">Try it</button>
+
+        <script>
+            // ruleid: get-elem-id
+            function myFunction() { document.getElementById("demo").innerHTML = "Paragraph changed."; }
+            // ruleid: get-elem-id
+            function myFunction() { document.getElementById("demo").innerHTML = "Paragraph changed."; }
+
+            // ruleid: cross-tag-taint
+            bar(x);
+        </script>
+
+    </body>
+</html>
--- a/src/engine/Match_extract_mode.ml
+++ b/src/engine/Match_extract_mode.ml
@@ -126,19 +126,53 @@ let offsets_of_mval extract_mvalue =
            end_pos = end_loc.charpos + end_len;
          })
 
-let mk_extract_target dst_lang contents rule_ids =
-  let dst_lang =
+(* Compute the rules that should be run for the extracted
+   language.
+
+   Note: for normal targets, this decision is made on the Python
+   side. Implementing that for extracted targets would be far more
+   annoying, however. The main implication of selecting the rules
+   here is, if your target is html and your dest lang is javascript,
+   there is no way to ignore an html path for a specific javascript
+   rule. In general extract mode does not allow you to select rules to
+   run on extracted targets so this is in keeping with that design *)
+let rules_for_extracted_lang (all_rules : Rule.t list) =
+  let rules_for_lang_tbl = Hashtbl.create 10 in
+  let memo xlang =
+    match Hashtbl.find_opt rules_for_lang_tbl xlang with
+    | Some rules_for_lang -> rules_for_lang
+    | None ->
+        let rule_ids_for_lang =
+          all_rules
+          |> List.mapi (fun i r -> (i, r))
+          |> List.filter (fun (_i, r) ->
+                 let r_lang = r.Rule.languages in
+                 match (xlang, r_lang) with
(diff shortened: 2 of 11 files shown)


IagoAbal:
~Shouldn't we have a regression test for this?~ We have! Sorry, confused by the test plan.

cli/tests/e2e/targets/extract/js_html_concat.html
```diff
@@ -0,0 +1,31 @@
...
+    <body>
+        <h2>Demo JavaScript in Body</h2>
+
+        <p id="demo">A Paragraph</p>
+
+        <button type="button" onclick="myFunction()">Try it</button>
+
+        <script>
+            // ruleid: get-elem-id
+            function myFunction() { document.getElementById("demo").innerHTML = "Paragraph changed."; }
```
IagoAbal, line 22: What would happen if we had a `bar(x);` here, would it be a finding of rule `cross-tag-taint` ?

src/runner/Run_semgrep.ml
```diff
@@ -559,7 +559,7 @@ let targets_of_config (config : Runner_config.t)
  * the original rules passed via -rules, without the extract-mode rules).
  *)
 let extracted_targets_of_config (config : Runner_config.t)
-    (rule_ids : Rule.rule_id list) (extractors : Rule.extract_rule list) :
+    (extractors : Rule.extract_rule list) (all_rules : Rule.t list) :
```
IagoAbal: I haven't looked much at the code but I don't understand why to pass the `extractors` here when it's a subset of `all_rules` and it's easy to compute right where you need it.

src/engine/Match_extract_mode.ml
```diff
@@ -126,19 +126,53 @@ let offsets_of_mval extract_mvalue =
...
+let rules_for_extracted_lang (all_rules : Rule.t list) =
+  let rules_for_lang_tbl = Hashtbl.create 10 in
+  let memo xlang =
+    match Hashtbl.find_opt rules_for_lang_tbl xlang with
+    | Some rules_for_lang -> rules_for_lang
+    | None ->
+        let rule_ids_for_lang =
+          all_rules
+          |> List.mapi (fun i r -> (i, r))
+          |> List.filter (fun (_i, r) ->
```
IagoAbal, line 148: ~`List.filteri` ?~

src/engine/Match_extract_mode.ml
```diff
@@ -126,19 +126,53 @@ let offsets_of_mval extract_mvalue =
...
+                 let r_lang = r.Rule.languages in
+                 match (xlang, r_lang) with
+                 | Xlang.L (l, _), Xlang.L (rl, rls) ->
+                     List.exists (fun x -> Lang.equal l x) (rl :: rls)
+                 | Xlang.LGeneric, Xlang.LGeneric -> true
+                 | Xlang.LRegex, Xlang.LRegex -> true
+                 | _else -> false)
```
IagoAbal: Perhaps this is better in case we add more members to `Xlang.t`:
```ocaml
| (Xlang.L _ | Xlang.LGeneric | Slang.LRegex), _ -> false
```

IagoAbal (approved):
Not familiar with extract mode implementation, but LGTM.

src/engine/Match_extract_mode.ml
```diff
@@ -126,19 +126,53 @@ let offsets_of_mval extract_mvalue =
...
+          |> Common.map (fun (i, _r) -> i)
```
IagoAbal: Maybe `Common.map fst` is marginally more readable?

src/runner/Run_semgrep.ml
```diff
@@ -660,6 +663,14 @@ let semgrep_with_rules config ((rules, invalid_rules), rules_parse_time) =
                  is in skipped_rules *)
              target.In.rule_nums
              |> List.filter_map (fun r_num -> Hashtbl.find_opt rule_table r_num)
+             (* Don't run the extract rules
+                Note: we can't filter this out earlier because the rule ids need to be stable *)
```
IagoAbal: Seems quite fragile. Shouldn't the rule ids be part of the `Rule.t` type"

IagoAbal: And by "ids" I guess we mean "indexes" since the actual "id" of a rule is a string?

src/engine/Match_extract_mode.ml
emjin: No, I need the pair for later when I add it to the table

src/runner/Run_semgrep.ml
emjin: oops great point

src/runner/Run_semgrep.ml
emjin: Yeah, it's certainly why this broke in the first place 😞. The targets file interface was meant to be temporary before we ported file targeting over. I think that's still the end goal, so I'm not super invested in making it better

PR #7274, opened 2023-03-03
Migrate more code to Fpath (was: Integrate semgrepignore/gitignore into osemgrep)

mjambon:
It started as an integration of the semgrepignore mechanism in osemgrep and ended up into a migration to Fpath (since we shared some code between semgrep-core and osemgrep).

I haven't really finished the semgrepignore integration. The test results are roughly the same as on the main branch but the total differs mysteriously:
```
Output of 'make osemgrep-e2e'

before:
=========== 395 failed, 107 passed, 4116 warnings, 2 rerun in 56.45s ===========

after:
=========== 402 failed, 106 passed, 4164 warnings, 2 rerun in 48.12s ===========
```

I'd like to merge this branch first because of all the Fpath changes, and only then get back to the semgrepignore tests.

Guide to migrating to Fpath
===

[Fpath](https://erratique.ch/software/fpath/doc/Fpath/index.html) is a library that provides a single module for handling native file paths. The main benefit of using it is that having a dedicated type for file paths clarifies function signatures. Fpath does not access the file system in any way. `Fpath.t` is the type of a path. Internally, it is a string, meaning that converting it to a string is cheap. However, converting a path from a string requires some validation. Functions for handling files that used to be in the `Common` module now exist in a dedicated module `File`. For example, `Common.read_file` becomes `File.read_file` when the path type changes from `string` to `Fpath.t`.

Common operations have their own operators:
* `src / "hello.c"`: for appending a single path segment  (see Fpath documentation)
* `abs_path // other_path`: for making `other_path` absolute if it's not absolute already (see Fpath documentation)
* `sprintf "doing something with file %s" !!file`: `!!` is `Fpath.to_string`. We use it quite a lot.

These 3 operators `/`, `//`, and `!!` (and nothing else) are made available with
```ocaml
open File.Operators
```

Migration guidelines:
* Prefer `Fpath.t` over `string` when you have a choice. `Common.filename`, `Common.dirname`, and `Common.path` are aliases for `string`. They will be replaced by `Fpath.t`.
* Use `open File.Operators` and `!!path` to turn a path into a string.
* Use `Fpath.v` to create a path from a string. On Unix, it will fail only if the string is empty or contains a null character. The other function `Fpath.of_string` is more cumbersome to use because it returns a `Result.t`.
* Use `/` and `//` to concatenate paths.
* Use `Fpath` instead of the `Filename` from the standard library. `Fpath` provides equivalent functionality and more.
* Use `File.whatever` instead of `Common.whatever` for handling files.
* Use `File.of_strings` and `File.to_strings` to convert lists of paths.

PR checklist:


.pre-commit-config.yaml
```diff
@@ -189,6 +189,8 @@ repos:
         # 'docker_image' to run semgrep, but for now jsonnet support
         # is not available via setup.py and only (unofficially)
         # available in docker.
+        # If docker complains that --platform isn't supported because
+        # it's experimental, try upgrading your docker server.
```
aryx, line 193: why is this change in?
Also I see below that you've updated the semgrep_interfaces. Why this is part of this PR?
Did you use the latest develop before making this PR?

languages/c/menhir/test_parsing_c.ml
```diff
@@ -7,21 +8,23 @@ module Stat = Parse_info
 let test_parse_c xs =
   Parse_cpp.init_defs !Flag_parsing_cpp.macros_h;
 
+  let xs = File.of_strings xs in
```
aryx, line 11: This does not read well. File.of_strings looks like you gonna create a new file from a set of strings.
I would rather have
File.fpaths_of_strings xs
or Fpath.vs

libs/commons/File.mli
```diff
@@ -0,0 +1,118 @@
...
+   - step 1: here, we expose the same functions as in Common but using the
+             Fpath.t type.
+   - step 2: move the original implementation from Common.ml to here
+             and stop using 'string' for file paths.
+*)
+
+(* TODO: call this type 't'? *)
+type fpath = Fpath.t [@@deriving show, eq]
+(* This type exists only to be used with ppx deriving because Fpath.t
+   doesn't provide 'show' and 'eq' functions.
```
aryx, line 14: Is it something we can ask Daniel Bunzli to add?
Nowaday lots of OCaml people use deriving.

libs/commons/File_type.ml
```diff
@@ -364,7 +365,7 @@ let file_type_of_file file =
   | "r"
   | "R" ->
       PL R
-  | _ when Common2.is_executable file -> Binary e
+  | _ when Common2.is_executable !!file -> Binary e
```
aryx, line 368: I guess we could migrate this Common2.is_executable to File.ml too.

libs/commons/File_type.ml
```diff
@@ -373,7 +374,7 @@ let file_type_of_file file =
   | _ when b = "TARGETS" -> Config Makefile
   | _ when b = ".depend" -> Obj "depend"
   | _ when b = ".emacs" -> PL (Lisp Elisp)
-  | _ when Common2.filesize file > 300_000 -> Obj e
+  | _ when Common2.filesize !!file > 300_000 -> Obj e
```
aryx, line 377: same, could be moved to File.ml

libs/lib_parsing/Skip_code.ml
```diff
@@ -13,6 +13,7 @@
  * license.txt for more details.
  *)
 open Common
+open File.Operators
```
aryx, line 16: once gitignore.ml is made independent of semgrep, we can get rid of Skip_code.ml and use
gitignore.ml instead for this kind of stuff.

libs/lib_parsing/Skip_code.ml
```diff
@@ -86,12 +89,12 @@ let filter_files skip_list ~root relative_paths : string list * string list =
   let relative_paths =
     relative_paths
     |> List.filter (fun rel_path ->
-           let path = Common.readable ~root rel_path in
+           let path = Common.readable ~root:!!root !!rel_path |> Fpath.v in
```
aryx, line 92: Can also be made part of File.ml

src/osemgrep/cli_scan/Rule_fetching.ml
```diff
@@ -40,8 +41,7 @@ type rules_and_origin = {
 }
 
 (* TODO? more complex origin? Remote of Uri.t | Local of filename | Inline? *)
-and origin = Common.filename option (* None for remote files *)
-[@@deriving show]
+and origin = Fpath.t option (* None for remote files *) [@@deriving show]
```
aryx, line 44: So deriving show is working on Fpath.t too?

src/osemgrep/targeting/Git_path.mli
```diff
@@ -43,5 +45,22 @@ val make_absolute : t -> t
...
 val of_fpath : Fpath.t -> t
 
+(* Convert back to a system path. *)
+val to_fpath : root:Fpath.t -> t -> Fpath.t
+
 (* / *)
 val root : t
+
+(*
+   Return an absolute, normalized git path relative to the project root.
```
aryx, line 55: We can get rid of Common2.readable, which provides a similar feature.

aryx (approved):
Nice!

.pre-commit-config.yaml
mjambon: because it was broken on my machine, which I found out at commit time.

languages/c/menhir/test_parsing_c.ml
mjambon: Addressed by https://github.com/returntocorp/semgrep/pull/7406 -> File.Path.of_strings

libs/commons/File_type.ml
mjambon: this will be for another day

src/osemgrep/cli_scan/Rule_fetching.ml
mjambon: I guess only `Fpath.pp` needs to be available, not `Fpath.show`.

src/osemgrep/targeting/Git_path.mli
mjambon: It would be easier if I knew what `Common.readable` does...

PR #7282, opened 2023-03-06
tainting: Java: Basic getX/setX sensitivity

IagoAbal:
Helps PA-2585

test plan:
make test # test added

PR checklist:

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -199,6 +199,7 @@ end
 
 (* THINK: Separate read-only enviroment into a new a "cfg" type? *)
 type env = {
+  lang : Lang.t;
   options : Config_semgrep.t; (* rule options *)
   config : config;
   fun_name : var option;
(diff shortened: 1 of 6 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -968,6 +969,31 @@ let check_tainted_instr env instr : Taints.t * Lval_env.t =
...
+                    ({
+                       base = Var _obj;
+                       rev_offset = [ ({ o = Dot m; _ } as offset) ];
+                     } as lval);
+                _;
+              },
+              [ _ ] )
+            when env.lang =*= Lang.Java
+                 && String.starts_with ~prefix:"set" (fst m.IL.ident) ->
+              let prop_name = "get" ^ Str.string_after (fst m.IL.ident) 3 in
```
emjin, line 987: We're probably going to add more of these hacks for other languages. It might be nice to extract this logic into another function/file

src/tainting/Dataflow_tainting.ml
```diff
@@ -968,6 +969,31 @@ let check_tainted_instr env instr : Taints.t * Lval_env.t =
         let all_args_taints =
           List.fold_left Taints.union Taints.empty all_taints
         in
+        let lval_env =
+          (* HACK: Java: If we encounter `obj.setX(arg)` we interpret this as `obj.getX = arg`. *)
```
emjin, line 973: This seems like we're limiting it more than it needs to be, why not interpret it as `obj.x = arg`?

emjin (approved):
Nice! I like it! It seems to me that `obj.setX` could also be used for `obj.X`, but since most people who have getters and setters have private fields, it doesn't really matter.

src/tainting/Dataflow_tainting.ml
IagoAbal: More PRs coming!

src/tainting/Dataflow_tainting.ml
IagoAbal: In general, `getX`/`setX` may not correspond to an actual attribute in the class. And as you pointed out in another comment you typically have getters/setters with private fields.

PR #7291, opened 2023-03-10
feat(taint): multiple source traces

brandonspark:
## What:
This PR adds in the ability to have dataflow traces which have "multiple sources". More specifically, it consolidates the information of multiple _potential_ sources which all flow into the taint sink.

## Why:
As it currently stands, we generate a single finding per every single source which reaches into a sink, when that sink produces a finding. This is a lot of findings, but they are deduplicated in the CLI side. This means that there is a single finding resulting from something like this:
```
x = "tainted"
y = "tainted"

sink(x, y)
```

and that finding has a dataflow trace which is arbitrary. It could be from x or from y. It depends on how it's deduplicated. This is unclear, because a user might fix the `x` source, upon seeing it in the trace, and then see the finding not go away, because the `y` source is still there.

## How:
This PR changes it so that dataflow traces can originate from several call traces, which might be very different, but still have taint.

This is achieved by killing the previous `finding` representation for taint findings, by changing it to contain the information of all the taints which flow into a sink or return. We do this "lazily", by deferring the actual act of extracting a finding from the "taint finding" as late as possible. This means that the sink `requires` is not evaluated until `Match_tainting_mode`, in `pms_of_finding`, as oppose to `findings_of_tainted_sink`.

Note that I did not strictly need to defer the `requires` evaluation, but it will need to be moved there once we start doing taint labels interprocedurally, so I figured now is a fine time to do it too.

Also, I really wish that we didn't use the term `finding` for the summary information generated for sinks and functions. It's confusing me with Semgrep findings.

This change is strictly internal to Semgrep core right now. Externally, call traces may change slightly if there are multiple sources coming into a sink, but life is hard sometimes.

## Test plan:

Here's what it looks like:
image

`make test`

Not user-facing, so no changelog.

Here's an example of how it looks for default sinks (with no `requires`) with multiple sources:
image

We see the old behavior is maintained.

Here's an example of how it looks for a sink with a `requires`, with multiple sources:
image

We see that we only produce one finding, and the metavariable interpolation does not work (because the metavariables in the sources did not unify). 

PR checklist:


brandonspark:
@IagoAbal this should pass core tests, i'd like to get your opinion on it before i go and do the necessary CLI changes.

one concern i have is that, because `text.py` operates only on `CliMatchDataflowTrace`, we cannot make Semgrep output different taint trace information without also changing the representation of the taint traces, thus opening a one-way door that we can't really close. so i want to be sure that we're ready to do this.

IagoAbal:
> Also, I really wish that we didn't use the term finding for the summary information generated for sinks and functions. It's confusing me with Semgrep findings.

I'm fine renaming "finding" to something else.

src/engine/Match_extract_mode.ml
```diff
@@ -214,8 +214,8 @@ let map_taint_trace map_loc { Pattern_match.source; tokens; sink } =
           }
   in
   {
-    Pattern_match.source = map_taint_call_trace source;
-    tokens = Common.map map_loc tokens;
+    Pattern_match.sources =
+      Common.map (fun (ct, toks) -> (map_taint_call_trace ct, toks)) sources;
```
IagoAbal: You are removing the `map_loc` that was applied here to `tokens`, is it on purpose?

src/tainting/Taint.ml
```diff
@@ -57,6 +57,13 @@ type source = {
 
 type sink = Rule.taint_sink call_trace [@@deriving show]
 
+let rec sink_has_no_requires = function
+  | PM (_pm, x) -> (
+      match x.Rule.sink_requires.G.e with
+      | G.N (G.Id ((s, _), _)) -> s = Rule.default_source_label
```
IagoAbal: If this was really necessary then we could make the `requires` field an `option`.

src/tainting/Taint.ml
```diff
@@ -169,6 +153,35 @@ let _show_taint taint =
...
+
+type taints_to_sink = {
+  (* These taints were incoming to the sink, under a certain
+     REQUIRES expression.
+     When we discharge the taint signature, we will produce
+     a certain number of findings suitable to how the sink was
+     reached.
+  *)
+  taints_with_precondition : taint list * G.expr;
+      (** These [label_origs] are the "requires trace" of the *)
```
IagoAbal: Need to update comment? I don't see any `label_origs`.

src/tainting/Dataflow_tainting.ml
```diff
@@ -421,45 +421,65 @@ and eval_label_args ~labels args =
...
+             | Src source ->
+                 let src_pm, _ = T.pm_of_trace source.call_trace in
+                 src_pm.PM.env
+           in
+           match merge_source_sink_mvars env sink_pm.PM.env bindings with
+           | None -> None
+           | Some _ -> Some (t, bindings))
+  in
+  let* merged_env =
+    taints_and_bindings_with_unifying_mvars |> List.concat_map snd
```
IagoAbal: What if we have two sources A and B, and both can unify with the sink, but we can't unify A with B?

src/engine/Match_tainting_mode.ml
```diff
@@ -447,42 +448,80 @@ let rec convert_taint_call_trace = function
...
-      let sink_pm, _ = T.pm_of_trace sink in
-      Some { sink_pm with env = merged_env; taint_trace }
-  | T.SrcToReturn _
-  (* TODO: We might want to report functions that let input taint
-   * go into a sink (?) *)
-  | T.ArgToSink _
-  | T.ArgToReturn _ ->
-      None
+  | T.ToReturn (_taints, _) ->
+      (* TODO: We might want to report functions that let input taint
```
IagoAbal, line 553: This comment was referring to `ArgToSink`, here it makes no sense.

src/engine/Match_tainting_mode.ml
```diff
@@ -447,42 +448,80 @@ let rec convert_taint_call_trace = function
...
+                 match orig with
+                 | Src src when src.label =*= Rule.default_source_label ->
+                     Left3 (src, tokens)
+                 | Src src -> Middle3 (src, tokens)
+                 | Arg arg -> Right3 arg)
+        in
+        (* The old behavior used to be that, for sinks with a `requires`, we would
+           generate a finding per every single taint source going in. Later deduplication
+           would deal with it.
+           We will continue that behavior for taint sinks with no `requires`, but for multiple
```
IagoAbal: Haven't you done the opposite? For sinks with no `requires:` this generates one finding, and for sinks with `requires:` it generates one finding per sink.

IagoAbal:
> one concern i have is that, because `text.py` operates only on `CliMatchDataflowTrace`, we cannot make Semgrep output different taint trace information without also changing the representation of the taint traces, thus opening a one-way door that we can't really close. so i want to be sure that we're ready to do this.

For now we can just leave it as internal, we need more discussion before changing traces in our interfaces.

src/engine/Match_extract_mode.ml
brandonspark: nope, good catch

brandonspark:
@IagoAbal I decided to just handle the unify mvars case by exhibiting the same behavior. I tried unifying all the sources, but this causes regressions in our test suite, and I wasn't comfortable bricking the behavior. I may have understood what you meant, though.

src/reporting/JSON_report.ml
```diff
@@ -211,10 +211,15 @@ let rec taint_call_trace = function
       let* call_trace = taint_call_trace call_trace in
       Some (Out.CoreCall (location, intermediate_vars, call_trace))
 
-let taint_trace_to_dataflow_trace { source; tokens; sink } :
+let taint_trace_to_dataflow_trace { sources; sink } :
     Out.core_match_dataflow_trace =
+  let call_trace, tokens =
+    match sources with
+    | [] -> raise Common.Impossible
+    | (call_trace, tokens) :: _ -> (call_trace, tokens)
```
IagoAbal, line 228: You could add a comment here to explain why we are ignoring all the other taints. This may be confusing for others without some clarification.

src/tainting/Dataflow_tainting.ml
```diff
@@ -423,45 +423,85 @@ and eval_label_args ~labels args =
...
+             (T.ToSink
+                {
+                  taints_with_precondition = ([ t ], R.get_sink_requires ts);
+                  sink;
+                  merged_env;
+                }))
+  else
+    match
+      taints_and_bindings |> List.concat_map snd
+      |> merge_source_sink_mvars env sink_pm.PM.env
```
IagoAbal, line 547: If source A binds `$X` and source B binds `$X`, and the sink does NOT bind `$X`, then I believe this will return a list of bindings where `$X` will occur twice possibly bound to different values.

IagoAbal (approved):
Nice, looking forward to more PRs. Tomorrow I'll compare this PR against develop on a large number of findings and I'll let you know if I find anything interesting.

IagoAbal:
Could you rebase against develop? When I test this it appears to be significantly slower than develop but I think it may be because of some recent optimizations that were made.

src/tainting/Dataflow_tainting.ml
```diff
@@ -367,6 +367,31 @@ let sink_biased_union_mvars source_mvars sink_mvars =
   in
   Some (source_mvars' @ sink_mvars)
 
+let merge_source_mvars taints_and_bindings =
```
IagoAbal: I think this could end up picking an arbitrary `$X` among the sources. If there are sources A, B, and C, each one binding `$X` to something else. First `merge A B` will remove `$X`, and then we  `merge (merge A B) C` and C will add `$X` again.

src/tainting/Dataflow_tainting.ml
brandonspark: oops, good catch. should be fixed now.

src/tainting/Dataflow_tainting.ml
brandonspark: good catch! i reimplemented + tested, and behavior is the same (as well as working for the 3-source case now)

IagoAbal (approved):

PR #7294, opened 2023-03-11
fix(metrics): Send metrics for app scans

emjin:
We currently do not send metrics for most scans triggered by semgrep ci due to an error in the logic. This makes it harder for us to debug issues for our customers, since there may be useful information in the metrics that we don't have access to. This PR fixes the bug

Test plan: I tested this by running `semgrep ci` and printing out whether metrics were enabled. Also see e2e tests

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -219,7 +219,7 @@ def ci(
         output_format=output_format,
     )
 
-    state.metrics.configure(metrics, metrics)
+    state.metrics.configure(metrics, metrics_legacy)
     state.error_handler.configure(suppress_errors)
     scan_handler = None
 
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -219,15 +219,6 @@ def ci(
         output_format=output_format,
     )
 
-    # Metrics are always on for CI scans. It doesn't make sense for them to
-    # be off, since CI scans have to connect to the Cloud Platform
-    # (For reviewers: is this explanation below clear?)
-    if metrics == MetricsState.OFF or metrics_legacy == MetricsState.OFF:
-        raise click.BadParameter(
-            "--metrics cannot be turned off for ci scans since ci scans require access to the Cloud Platform"
-        )
-    metrics = MetricsState.ON
-
     state.metrics.configure(metrics, metrics)
     state.error_handler.configure(suppress_errors)
     scan_handler = None
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -219,7 +219,16 @@ def ci(
         output_format=output_format,
     )
 
-    state.metrics.configure(metrics, metrics_legacy)
+    # Metrics are always on for CI scans. It doesn't make sense for them to
+    # be off, since CI scans have to connect to the Cloud Platform
+    # (For reviewers: is this explanation below clear?)
+    if metrics == MetricsState.OFF or metrics_legacy == MetricsState.OFF:
+        raise click.BadParameter(
+            "--metrics cannot be turned off for ci scans since ci scans require access to the Cloud Platform"
+        )
+    metrics = MetricsState.ON
+
+    state.metrics.configure(metrics, metrics)
     state.error_handler.configure(suppress_errors)
     scan_handler = None
 
--- a/cli/src/semgrep/metrics.py
+++ b/cli/src/semgrep/metrics.py
@@ -410,6 +410,7 @@ class Metrics:
           - on, sends
           - off, doesn't send
         """
+        # import here to prevent circular import
         from semgrep.state import get_state
 
         state = get_state()
--- a/cli/src/semgrep/metrics.py
+++ b/cli/src/semgrep/metrics.py
@@ -410,8 +410,19 @@ class Metrics:
           - on, sends
           - off, doesn't send
         """
+        from semgrep.state import get_state
+
+        state = get_state()
+
         if self.metrics_state == MetricsState.AUTO:
-            return self.is_using_registry
+            # When running logged in with `semgrep ci`, configs are
+            # resolved before `self.is_using_registry` is set.
+            # However, these scans are still pulling from the registry
+            using_app = (
+                state.command.get_subcommand() == "ci"
+                and state.app_session.is_authenticated()
+            )
+            return self.is_using_registry or using_app
         return self.metrics_state == MetricsState.ON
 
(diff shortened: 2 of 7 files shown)


emjin:
I checked PRIVACY.md and it says:
```
$ semgrep login && semgrep ci   # → metrics enabled (logged in to semgrep.dev)
```
so docs should already be up to date

cli/src/semgrep/commands/ci.py
```diff
@@ -219,7 +219,16 @@ def ci(
         output_format=output_format,
     )
 
-    state.metrics.configure(metrics, metrics_legacy)
+    # Metrics are always on for CI scans. It doesn't make sense for them to
+    # be off, since CI scans have to connect to the Cloud Platform
+    # (For reviewers: is this explanation below clear?)
+    if metrics == MetricsState.OFF or metrics_legacy == MetricsState.OFF:
+        raise click.BadParameter(
+            "--metrics cannot be turned off for ci scans since ci scans require access to the Cloud Platform"
```
ievans:
```suggestion
            "--metrics cannot be turned off for ci scans since they require access to Semgrep Cloud Platform"
```

cli/src/semgrep/commands/ci.py
```diff
@@ -219,7 +219,16 @@ def ci(
...
+        )
+    metrics = MetricsState.ON
```
dataholly: This change looks like it achieves the goal of getting metrics to on, but is there (also?) a more straightforward change we can make such that cloud platform initiated scans are already setting metrics to "ON" in the first place?  I worry we're generating an exception and potentially worrisome warning message on every app generated scan if not, but don't know enough to know where that change to default behavior is needed.

cli/src/semgrep/commands/ci.py
dataholly: or "--metrics cannot be turned off because Semgrep Cloud Platform is needed for ci scans"

emjin:
Edit: it was brought to my attention thatt `semgrep ci` can be run without being logged in, with just local files, so I need to make some changes. Will rerequest review

cli/src/semgrep/commands/ci.py
emjin: If the specific concern is the exception, we don't have to disallow `metrics=off` for ci scans. I added this error because I remember in a previous discussion that we assumed `semgrep ci --metrics=off` was always a confusion (we send info to the app, even if not exactly the metrics). That being said, I'm perfectly happy to not have the warning and follow AUTO behavior.

cli/src/semgrep/commands/ci.py
emjin: @ievans I took this warning out in my new changes because without the locality it makes it a little harder to understand that check. That being said, it's easy to add a check that logged in `semgrep ci` scans cannot be run with `--metrics=off` if that's what we want. I have no strong feelings here.

Argument for disallowing `semgrep ci --metrics=off` when logged in: let's be extra clear to users that we are sending their data and there's no way around that when they run semgrep ci
Argument for respecting `semgrep ci --metrics=off` (but still sending data to the app): people might be ok with what we send to the app but specifically not want us to send extra metrics. It also makes our tests work 😅 because our tests run `semgrep ci --metrics=off`

The latter would be less work for me but from a product/privacy perspective I don't know which one makes more sense, let me know.

cli/src/semgrep/commands/ci.py
dataholly: Nevermind, I missed that you had a new commit with a fix to is_enabled in metrics already, that fix looks good to me.

> In the current case, a scan that uses CI but does not use registry will not use CLI metrics.  We want to change that behavior in general such that by default such a scan would use CLI metrics.  We don't consider CI without registry as an unusual setup, so I found handling it via exception unintuitive.
> 
> For example, could we set this at metrics configure time, e.g. in [metrics.py:configure(...)](https://github.com/returntocorp/semgrep/blob/60b623e4f66265599f0623622a1e1f7e34487972/cli/src/semgrep/metrics.py#L227).  If command is CI we do not check for registry usage before setting metrics to on by default.  But I am definitely not the expert on this code base, so will defer to your expertise.
> 
> I don't have enough context to know how we should handle the other question -- how to handle potentially unexpected / unintuitive setting of command CI with user-specified --metrics=off.  This is a particularly complicated question for users who are logged in to Cloud Platform.

cli/src/semgrep/metrics.py
```diff
@@ -410,8 +410,19 @@ def is_enabled(self) -> bool:
...
         if self.metrics_state == MetricsState.AUTO:
-            return self.is_using_registry
+            # When running logged in with `semgrep ci`, configs are
+            # resolved before `self.is_using_registry` is set.
+            # However, these scans are still pulling from the registry
+            using_app = (
+                state.command.get_subcommand() == "ci"
+                and state.app_session.is_authenticated()
+            )
+            return self.is_using_registry or using_app
```
dataholly, line 426: This looks good to me!

dataholly (commented):
This looks good to me, will let others give the final stamp.

cli/tests/e2e/test_ci.py
```diff
@@ -1262,3 +1263,16 @@ def test_git_failure_error_handler(
         env={"SEMGREP_APP_TOKEN": "fake-key-from-tests"},
     )
     mock_send.assert_called_once_with(mocker.ANY, 2)
+
+
+def test_metrics_enabled(run_semgrep: RunSemgrep, mocker):
```
brendongo, line 1268: Nice test

cli/src/semgrep/commands/ci.py
```diff
@@ -219,7 +219,7 @@ def ci(
         output_format=output_format,
     )
 
-    state.metrics.configure(metrics, metrics_legacy)
+    state.metrics.configure(metrics, metrics)
```
brendongo: Why are we sending the argument twice?

cli/src/semgrep/metrics.py
```diff
@@ -410,8 +410,19 @@ def is_enabled(self) -> bool:
           - on, sends
           - off, doesn't send
         """
+        from semgrep.state import get_state
```
brendongo, line 414: Does this import need to be here?

emjin: Yeah, otherwise I get a circular import. Commented

cli/src/semgrep/commands/ci.py
emjin: oops this was left over from a previous change

aryx (approved):

PR #7310, opened 2023-03-14
tainting: pro: Java: Interprocedural getX-sensitivity

IagoAbal:
test plan:
See returntocorp/semgrep-proprietary#601

PR checklist:

### DIFF
--- a/src/engine/Test_dataflow_tainting.ml
+++ b/src/engine/Test_dataflow_tainting.ml
@@ -34,7 +34,7 @@ let test_tainting lang file options config def =
           let tok1, tok2 = (fst (Taint.pm_of_trace src.call_trace)).range_loc in
           let r = Range.range_of_token_locations tok1 tok2 in
           Range.content_at_range file r
-      | Taint.Arg (s, i) -> spf "arg %s %d" s i
+      | Taint.Arg arg -> Taint._show_arg arg
     in
     taint |> Taint.Taint_set.elements |> Common.map show_taint
     |> String.concat ", "
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -489,7 +489,7 @@ let union_new_taint_sources_filtering_labels ~new_ curr =
   let new_filtered = filter_new_taint_sources_by_labels labels new_ in
   Taints.union new_filtered curr
 
-let find_args_taints args_taints fparams =
+let find_pos_in_actual_args args_taints fparams =
   let pos_args_taints, named_args_taints =
     List.partition_map
       (function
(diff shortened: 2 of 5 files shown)


src/engine/Test_dataflow_tainting.ml
```diff
@@ -34,7 +34,7 @@ let test_tainting lang file options config def =
           let tok1, tok2 = (fst (Taint.pm_of_trace src.call_trace)).range_loc in
           let r = Range.range_of_token_locations tok1 tok2 in
           Range.content_at_range file r
-      | Taint.Arg (s, i) -> spf "arg %s %d" s i
+      | Taint.Arg arg -> Taint._show_arg arg
```
emjin, line 37: Out of curiosity why the underscore?

src/tainting/Dataflow_tainting.ml
```diff
@@ -575,6 +575,30 @@ let propagate_taint_to_label replace_labels label (taint : T.taint) =
   in
   { taint with orig = new_orig }
 
+let add_offset_to_poly_taint_in_st env lval st =
```
emjin: Nit: the name is pretty opaque considering how many words it is. To me it seems like it should be closer to "resolve taints for getters"

src/tainting/Dataflow_tainting.ml
```diff
@@ -772,7 +796,11 @@ and check_tainted_lval_aux env (lval : IL.lval) :
         | (`None | `Tainted _) as st -> (
             match Lval_env.dumb_find lval_env lval with
             | (`Clean | `Tainted _) as st' -> st'
-            | `None -> st)
+            | `None ->
+                (* HACK(field-sensitivity): Java: If we encounter `obj.getX` and `obj` has
+                   * polymorphic taint,  and we know nothing specific about `obj.getX`, then
+                   * we add the same offset `.getX` to the polymorphic taint coming from `obj`. *)
+                add_offset_to_poly_taint_in_st env lval st)
```
emjin: I think I'm missing something. Where do we resolve `setX` vs `setY` to know that `setX` only applies to `x`?

src/tainting/Dataflow_tainting.ml
```diff
@@ -902,8 +949,10 @@ let check_function_signature env fun_exp args_taints =
                  Some
                    (Taints.singleton
                       { orig = Src { src with call_trace }; tokens = [] })
-             | T.ArgToReturn (argpos, tokens, _return_tok) ->
-                 let* arg_taints = taints_of_arg argpos in
+             | T.ArgToReturn (arg, tokens, _return_tok) ->
+                 let* arg_taints =
+                   taints_of_sig_arg env fparams args args_taints arg
```
emjin, line 980: I'm surprised we didn't have to do this before. Does this also improve our field sensitivity in other ways than just the getX case?

emjin (approved):
I think I'm still missing some details in how everything hooks up to make the getX/setX case work, but everything here makes sense and the tests are convincing

src/engine/Test_dataflow_tainting.ml
IagoAbal: It's just a "debugging version" of `show_arg`, in this case it could probably replace `show_arg` because it contains the same information and it's prettier. But I tend to leave the derived `show_arg` as they are.

src/tainting/Dataflow_tainting.ml
IagoAbal: The name is clear to me right now, but perhaps I agree with you in X months time. So I'm renaming it to `resolve_poly_taint_for_java_getters` for now!

src/tainting/Dataflow_tainting.ml
IagoAbal: That was done as part of a previous PR, you reviewed it too I think. There we intercept `obj.setX(foo)` and we propagate the taint of `foo` into `obj.getX`. I made this more clear here I hope:
```ocaml
        let lval_env =
          (* HACK: Java: If we encounter `obj.setX(arg)` we interpret this as `obj.getX = arg`. *)
          propagate_taint_via_java_setter {env with lval_env} e args all_args_taints
        in
```

src/tainting/Dataflow_tainting.ml
IagoAbal: Note that this is an incomplete solution, it doesn't handle all cases, but it's mean to handle the most basic ones. It does work for more than just getters, I have tried it with Python, and I also managed to get something working for JS/TS. But I feel it needs some testing and for now I feel more comfortable restricting it to Java and `getX`.

PR #7312, opened 2023-03-14
feat: allow string literal metavariables in julia

### LINKED ISSUE #7280: Literal metavariables do not work in Julia
**Describe the bug**
Literal metavariables do not work in Julia. See the linked playground. Changing the language from _julia_ to _python_ makes the same pattern match. 

**To Reproduce**
https://semgrep.dev/s/aviks:literal-metavariable

**Expected behavior**
The pattern matches. 

**Screenshots**

**Environment**
Playground

**Use case**
More accurate patterns

brandonspark:
## What:
This PR adds in the ability to properly parse string literal metavariables like `"$A"` in Julia.

## Why:
These were previously being confused with interpolated variables in a string.

## How:
Special-cased it in the parser.

This can probably also happen in the future if you did something like `"foo $A bar"`, but this is kind of a weird pattern and I find it highly unlikely anyone will try that.

## Test plan:
`make test`

Closes #7280 
Closes PA-2630

PR checklist:

### DIFF
--- a/languages/julia/generic/Parse_julia_tree_sitter.ml
+++ b/languages/julia/generic/Parse_julia_tree_sitter.ml
@@ -670,12 +670,16 @@ and map_anon_choice_str_content_838a78d (env : env)
   match x with
   | `Str_content tok -> (* string_content *) Left3 (str env tok)
   | `Str_interp (v1, v2) ->
-      let _v1 = (* "$" *) token env v1 in
+      let dollar_s, dollar_t = (* "$" *) str env v1 in
       let v2 =
         match v2 with
-        | `Id tok ->
-            let id = map_identifier env tok in
-            Middle3 (N (H2.name_of_id id) |> G.e)
+        | `Id tok -> (
+            let s, t = map_identifier env tok in
+            match env.extra with
+            | Pattern ->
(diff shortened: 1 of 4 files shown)


languages/julia/generic/Parse_julia_tree_sitter.ml
```diff
@@ -670,12 +670,16 @@ and map_anon_choice_str_content_838a78d (env : env)
...
       let v2 =
         match v2 with
-        | `Id tok ->
-            let id = map_identifier env tok in
-            Middle3 (N (H2.name_of_id id) |> G.e)
+        | `Id tok -> (
+            let s, t = map_identifier env tok in
+            match env.extra with
+            | Pattern ->
+                let mvar_id = (dollar_s ^ s, PI.combine_infos dollar_t [ t ]) in
```
aryx, line 680: in theory we could check also that is_metavariable_name, but I guess it's fine also like that.

tests/rules/string_mvar_julia.jl
```diff
@@ -0,0 +1,11 @@
+
```
aryx, line 1: I don't think we need a tests/rules/ for that. a tests/patterns/julia/metavar_string.jl should be enough.

aryx (approved):

PR #7313, opened 2023-03-14
cli: start scan before fetching config

chmccreery:
We want to start caching the rules that semgrep runs on a particular scan in our database.

In order to connect a cached config that is generated to the scan that it was generated for (in the even that it was generated for a specific scan), we need to create the scan before generating the rules config.

This PR accomplishes this by separating out the endpoint that is used by scans, which takes a scan_id, from the endpoint that is used by non-scans, which takes a bunch of metadata in order to figure out which rules to run.

After this PR gets merged, for scans we call:
`PUT deployments/scans` to create the scan object in our database
`GET scans/<scan_id>/config` to create the list of rules (and connect them to the scan they were generated for so we can go back later and always have a source of truth for which rules ran on a scan)

For non-scans (dry-run, VS code, etc.) we call:
`GET deployments/scans/config` to create and return the list of rules based on a set of metadata

Fixes APP-3607

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -238,13 +238,13 @@ def ci(
         )
         sys.exit(FATAL_EXIT_CODE)
     elif token:
-        deployment = auth.is_valid_token(token)
-        if not deployment:
+        deployment_name = auth.get_deployment_from_token(token)
+        if not deployment_name:
             logger.info(
                 "API token not valid. Try to run `semgrep logout` and `semgrep login` again.",
             )
             sys.exit(INVALID_API_KEY_EXIT_CODE)
-        scan_handler = ScanHandler(dry_run=dry_run, deployment_name=deployment)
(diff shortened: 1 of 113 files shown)


cli/src/semgrep/commands/ci.py
```diff
@@ -296,18 +297,19 @@ def ci(
...
                     else ""
                 )
-                connection_task = progress_bar.add_task(
-                    f"Fetching configuration from Semgrep Cloud Platform{at_url_maybe}"
-                )
-                scan_handler.fetch_and_init_scan_config(metadata_dict)
-                progress_bar.update(connection_task, completed=100)
 
                 start_scan_task = progress_bar.add_task(
                     f"Reporting start of scan for [bold]{scan_handler.deployment_name}[/bold]"
```
chmccreery, line 302: Most of the code changes in this PR were to enable us to know and report the deployment name back to the user. Previously, we only kept track of the deployment_name _after_ `fetch_and_init_scan_config` had been called, so this involved some refactoring.

jbergler (approved):

PR #7315, opened 2023-03-14
feat(rust): allow let patterns to percolate taint

brandonspark:
## What:
This PR lets Rust let patterns be translated into appropriate structures in the IL, such that it can transfer taint.

## Why:
Currently, very simple programs in Rust that look like they should straightforwardly percolate taint do not, such as:
```
fn main() {
  let file = tainted; 

  sink(file);
}
```

## How:
I avoided putting Rust let-patterns into `EDynamic` of a `OtherExpr`, since `OtherExpr` can't be used for any further semantic purpose other than matching. I instead put it into `EPattern`, and allowed any `EPattern` of an identifier to be translated as normal.

Note that this doesn't handle the tuple case. I leave that to @IagoAbal, as I'm not sure I know the proper IL way to treat this.

## Test plan:
`make test`

Closes PA-2605

PR checklist:

### DIFF
--- a/changelog.d/pa-2605.fixed
+++ b/changelog.d/pa-2605.fixed
@@ -1,0 +1,2 @@
+Rust: Basic let-statement bindings (such as `let x = tainted`) now properly
+carry taint.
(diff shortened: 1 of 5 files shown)


aryx (approved):
Nice.
We can start to add maturity test for taints, just like we have for search, so we can
hope to detect those simple things for new language we add in the futur.

PR #7317, opened 2023-03-15
Add nestingLevel to SARIF

LewisArdern:
This introduces a change to our SARIF spec to support a nicer display when uploading to places like GHAS. Previously it would reference the rule.message in the title, finding shortdescription, and finding fullDescription.

This leverages the help.markdown which will override the fullDescription for any products leveraging it. 

We add references and the helpURI to the references in the show more tab to help the reviewer get up to speed with the finding. We also added the nestedLevels which appeared to help with the show paths.

Example:
image

PR checklist:


aryx:
@LewisArdern do we still want this? Who would be the best reviewer for this?

LewisArdern:
@aryx yes, and @tpetr or @emjin

tpetr:
Let's bring the PR out of draft mode if we're ready for review

cli/src/semgrep/formatter/sarif.py
```diff
@@ -258,11 +261,12 @@ def _rule_to_sarif(rule: Rule) -> Mapping[str, Any]:
         severity = SarifFormatter._rule_to_sarif_severity(rule)
         tags = SarifFormatter._rule_to_sarif_tags(rule)
         security_severity = rule.metadata.get("security-severity")
+
         if security_severity is not None:
             rule_json = {
                 "id": rule.id,
                 "name": rule.id,
-                "shortDescription": {"text": rule.message},
+                "shortDescription": {"text": f"Semgrep Finding - {rule.id}"},
```
tpetr: uber-nit: thoughts on `Semgrep Finding: ` vs. `Semgrep Finding - `? (I personally slightly lean towards the colon, but either are fine. I guess I'm more curious what pattern other tools follow?)

cli/src/semgrep/formatter/sarif.py
```diff
@@ -275,16 +279,33 @@ def _rule_to_sarif(rule: Rule) -> Mapping[str, Any]:
...
                 "properties": {"precision": "very-high", "tags": tags},
             }
 
         rule_url = rule.metadata.get("source")
+        references = []
+
         if rule_url is not None:
             rule_json["helpUri"] = rule_url
-
+            references.extend([f"[Semgrep Rule]({rule_url})"])
```
tpetr: uber-nit: just use `.append()` for single values

cli/src/semgrep/formatter/sarif.py
```diff
@@ -275,16 +279,33 @@ def _rule_to_sarif(rule: Rule) -> Mapping[str, Any]:
...
+            ref = rule.metadata["references"]
+            references.extend(
+                [f"[{r}]({r})" for r in ref]
+                if isinstance(ref, list)
+                else [ref]
+            )
+        if references:
+            r = ""
+            for item in references:
+                r += "\n - {}".format(item)
```
tpetr:
```python
r = "".join([f" - {item}\n" for item in references])
```

tpetr: I might also rename the variable `references_markdown` to make it clearer

cli/src/semgrep/formatter/sarif.py
```diff
@@ -275,16 +279,33 @@ def _rule_to_sarif(rule: Rule) -> Mapping[str, Any]:
...
             rule_json["helpUri"] = rule_url
-
+            references.extend([f"[Semgrep Rule]({rule_url})"])
+
+        if "references" in rule.metadata:
+            ref = rule.metadata["references"]
+            references.extend(
+                [f"[{r}]({r})" for r in ref]
+                if isinstance(ref, list)
+                else [ref]
```
tpetr: why create markdown links when `ref` is a list but not when it's a scalar?

cli/src/semgrep/formatter/sarif.py
tpetr: also lets factor this out into a variable so that we don't have to update it in 2 locations (even better would be to reduce the duplication in the entire `rule_json` dict creation, but that could be done another day)

cli/src/semgrep/formatter/sarif.py
```diff
@@ -275,16 +279,33 @@ def _rule_to_sarif(rule: Rule) -> Mapping[str, Any]:
...
 
         rule_url = rule.metadata.get("source")
+        references = []
+
         if rule_url is not None:
             rule_json["helpUri"] = rule_url
-
+            references.extend([f"[Semgrep Rule]({rule_url})"])
+
+        if "references" in rule.metadata:
```
tpetr: I'd go even further and check for truthiness (e.g. `if rule.metadata.get("references")`)

tpetr (approved):
Looks good at a high level. I have some nits and questions but nothing that would block merging.

cli/src/semgrep/formatter/sarif.py
```diff
@@ -326,6 +346,9 @@ def _rule_to_sarif_tags(rule: Rule) -> Sequence[str]:
                 if isinstance(owasp, list)
                 else [f"OWASP-{owasp}"]
             )
+        if "confidence" in rule.metadata:
```
tpetr:
```suggestion
        if rule.metadata.get("confidence"):
```
^ uber-nit but lets guard against false-y values like ""

tpetr, line 349:
```suggestion
        if rule.metadata.get("confidence"):
```
^ uber-nit but lets guard against false-y values like ""

cli/src/semgrep/formatter/sarif.py
```diff
@@ -275,16 +279,32 @@ def _rule_to_sarif(rule: Rule) -> Mapping[str, Any]:
...
-
+            references.append(f"[Semgrep Rule]({rule_url})")
+
+        if rule.metadata.get("references"):
+            ref = rule.metadata["references"]
+            references.extend(
+                [f"[{r}]({r})" for r in ref]
+                if isinstance(ref, list)
+                else [f"[{ref}]({ref})"]
+            )
```
tpetr, line 302: nit: are we 100% sure these references will always be URLs? if so, I think we can just add them without the markdown links

tpetr (approved):
LMK if you need help updating the snapshots

cli/src/semgrep/formatter/sarif.py
LewisArdern: In semgrep rules at least, they will always be urls, if you don't markdown link it'll just show it without rendering  -- people who write custom rules could technically put anything in there

tpetr: Oh gotcha 👍

LewisArdern:
Still need to wait for the new snapshot to pop up.. but then we are good

cli/src/semgrep/formatter/sarif.py
```diff
@@ -326,6 +349,9 @@ def _rule_to_sarif_tags(rule: Rule) -> Sequence[str]:
                 if isinstance(owasp, list)
                 else [f"OWASP-{owasp}"]
             )
+        if rule.metadata.get("confidence"):
+            confidence = rule.metadata["confidence"]
+            result.extend([f"{confidence} CONFIDENCE"])
```
tpetr:
```suggestion
            result.append(f"{confidence} CONFIDENCE")
```

tpetr (approved):
one last mega-nit

PR #7325, opened 2023-03-15
GHA: update to returntocorp/ocaml:alpine-2023-03-03

aryx:
This should speedup a bit the build in CI

test plan:
check the details of the CI build check and make sure
we don't install OPAM packages (they should all be already
installed in returntocorp/ocaml:alpine-2023-03-03) so
the make install-deps-XXX step should be a noop mostly.

PR checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,8 +3,8 @@
 ###############################################################################
 # The main goals of this pipeline are to:
 # - dogfood Semgrep by running it on each semgrep PRs.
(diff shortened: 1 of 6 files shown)


aryx:
```
opam install -y --deps-only ./libs/ocaml-tree-sitter-core
Warning:  Running as root is not recommended
Nothing to do.
opam install -y --deps-only ./
Warning:  Running as root is not recommended
Nothing to do.
...
```
so seems to work as intended.

PR #7327, opened 2023-03-15
CI: remove use of a dune cache

aryx:
Not sure what was the purpose of this caching code. I think
Bence added the code a long time ago. Maybe when chris
removed some of the semgrep-core caching code bence wrote
he forgot also to remove this bit.

Maybe this is useful when ocaml-layer lags behind,
but this step currently takes 1m, so seems simpler
to just rely on ocaml-layer for caching the needed packages
and just faster too.

test plan:
wait for CI to kick in, look at the time for the Test semgrep-core
job and look if less than 10min51s

PR checklist:

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,6 +1,7 @@
 # This workflow dogfoods Semgrep Code and Supply Chain.
 
 name: semgrep
+
 on:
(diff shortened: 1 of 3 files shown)


aryx:
The 'test semgrep-core' job takes less time without the cache than with it so I guess it's not used.

PR #7350, opened 2023-03-17
Clojure: consider foo/bar as two tokens, as well as $X/bar

### LINKED ISSUE #7311: Clojure: naming resolution broken
**Describe the bug**
I'm currently trying to match an import/require/use statement with a later usage of a function, through an alias.
It looks like this:
```clj
(require '[clojure.xml :as xml]
         '[clojure.zip :as zip])
     


(defn zip-str [s]
  (zip/xml-zip 
      (xml/parse (java.io.ByteArrayInputStream. (.getBytes s)))))

(zip-str "<a href='test.com'/>")
[{:tag :a, :attrs {:href "test.com"}, :content nil} nil]

(println (xml/emit-element (zip/root [{:tag :a, :attrs {:href "test.com"}, :content nil} nil])))
```

And I'm trying to do this, to catch a direct use, maybe later check if XXE protections have not been enabled etc.... (inspiration here https://github.com/clj-holmes/clj-holmes-rules/blob/main/security/xxe-clojure-xml/xxe-clojure-xml.yml)

```
    patterns:
      - pattern: |
          ... clojure.xml :as $X ...
      - pattern: |
          ... 
          ($X/parse ... ) 
          ...
```

**To Reproduce**
https://semgrep.dev/s/rgkj

**Expected behavior**
It should show me a match, because both patterns work independently I think (remove $X for the xml/parse call)

**Use case**
Making more rulez !!! Clojure seriously has zero tooling

aryx:
This closes #7311

test plan:
test file included

PR checklist:


Gby56:
Not sure if I'm doing this wrong, I wanted to try it in docker on the develop or edge images
```
docker run --rm -v "${PWD}:/src" returntocorp/semgrep:develop ls -lia && semgrep scan vuln.clj --config rule.yaml --lang clojure --verbose
```

with vuln.clj containing the test you added, and rule.yaml the simple pattern, I'm not getting a match 🤔

aryx:
I didn't fix the whole rule. I just fixed one of the bug that was causing the rule to not match. There are more bugs ...

Gby56:
Ah ! 😅 my bad, thanks for the explanation, I'll see if I can dive into the contribution guidelines for parsers but I'm a complete newbie in OCaml...

PR #7358, opened 2023-03-20
tainting: First batch of taint maturity tests

IagoAbal:
Found out that C# assignemnts are all encoded as AssignOp(Eq) and that was causing sanitizers not to work as expected, `y = sanitize(x)` was viewed as `y = x = sanitize(x)` and thus `y` was still considered as tainted.

test plan:
make test

PR checklist:

### DIFF
--- a/src/engine/Unit_engine.ml
+++ b/src/engine/Unit_engine.ml
@@ -696,7 +696,7 @@ let lang_tainting_tests () =
 (* TODO: We could have a taint_maturity/POLYGLOT/ directory to put reusable rules
  * that can work for multiple languages (like we have for tests/patterns/POLYGLOT/ *)
 let full_rule_taint_maturity_tests () =
-  let path = Filename.concat tests_path "taint_maturity" in
+  let path = tests_path / "taint_maturity" in
   pack_tests "taint maturity"
     (let tests, _print_summary =
        Test_engine.make_tests ~unit_testing:true [ path ]
--- a/src/engine/Unit_engine.ml
+++ b/src/engine/Unit_engine.ml
@@ -688,6 +688,18 @@ let lang_tainting_tests () =
 (* Full rule tests *)
 (*****************************************************************************)
 
+(* TODO: For now we only have taint maturity tests for Beta, there are no specific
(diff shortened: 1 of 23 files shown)


src/engine/Unit_engine.ml
```diff
@@ -681,6 +681,18 @@ let lang_tainting_tests () =
...
 
+(* TODO: For now we only have taint maturity tests for Beta, there are no specific
+ * tests for GA, but eventually we'll need something more sophisticated. Ideally
+ * we can integrate these tests into the same `maturity_tests` function that packs
+ * search-mode maturity tests. *)
+let full_rule_taint_maturity_tests () =
+  let path = Filename.concat tests_path "taint_maturity" in
+  pack_tests "taint maturity"
+    (let tests, _print_summary =
+       Test_engine.make_tests ~unit_testing:true [ path ]
```
aryx, line 702: What would be nice is to have here an explicit list of test filenames, like "taint_if", that
is then checked for every language, like I do for the search mode maturity tests.

tests/taint_maturity/c/taint_if.yaml
```diff
@@ -0,0 +1,17 @@
+rules:
+  - id: taint-maturity
+    mode: taint
+    languages:
```
aryx, line 4: potentially we could have a taint_maturity/POLYGLOT/ directory to put reusable rules that can work for multiple
languages (like we have for tests/patterns/POLYGLOT/

aryx (approved):

IagoAbal:
I agree with both suggestions, I added a couple of TODOs to take care of both in the next PR.

brandonspark:
do we know where this is going to fall in terms of our "language maturity" criteria?

IagoAbal:
> do we know where this is going to fall in terms of our "language maturity" criteria?

This is just one PR and more will come. But these two tests "taint_seq" and "taint_if" are meant to be Beta-level maturity tests. One potential GA-level test could be related to field sensitivity. Happy to have a chat about this if you have an opinion about how these tests should look like.

PR #7361, opened 2023-03-20
fix: allow semgrep-specific env vars for GitHub and GitLab CI

brandonspark:
## What:
This PR adds in the logic which interprets the Semgrep-specific env vars for metadata, for GitHub and GitLab scans.

## Why:
@r2c-david ran into an issue where setting the `SEMGREP_REPO_NAME` env var was causing no observable difference for GitHub and GitLab scans. It turns out this is because for all scans except those two providers, we have the logic to interpret those `SEMGREP_`-prefixed env vars, but not those two. I just added it in to standardize functionality, so it's not confusing when those env vars have no effect for certain providers.

## How:
I just added in the logic, usually by calls to the superclass `GitMeta`.

Note that I avoided this for `repo_name` and `repo_url`, which `GitMeta` handles the error case for (when the env var is not set). We only want to act if the env var is set, but if it isn't, then we shouldn't use `GitMeta`'s logic, we should use the GitHub and GitLab specific logic.

## Test plan:
Added test, and observed the `meta.json` changing. It seems OK to me, but cc @brendongo 

Closes PA-2644

PR checklist:


cli/src/semgrep/meta.py
```diff
@@ -545,7 +570,10 @@ def branch(self) -> Optional[str]:
         """
         if self.event_name == "pull_request_target":
             return os.getenv("GITHUB_HEAD_REF")
-        return os.getenv("GITHUB_REF")
+        github_ref = os.getenv("GITHUB_REF")
+        if github_ref:
+            return github_ref
+        return super().branch
```
brendongo, line 574: if we want to keep behavior consistent shouldn't we return SEMGREP_BRANCH before GITHUB_REF?

cli/src/semgrep/meta.py
```diff
@@ -593,15 +621,39 @@ def _fetch_branch_get_merge_base(branch_name: str, head_sha: str) -> str:
 
     @property
     def repo_name(self) -> str:
-        return os.getenv("CI_PROJECT_PATH", "[unknown]")
+        project_path = os.getenv("CI_PROJECT_PATH")
```
brendongo: Note that os.getenv second arg is a default value if things are not set. So this can probably be rewritten as:
```
return os.getenv("SEMGREP_REPO_NAME", os.getenv("CI_PROJECT_PATH"))
```

cli/src/semgrep/meta.py
```diff
@@ -593,15 +621,39 @@ def _fetch_branch_get_merge_base(branch_name: str, head_sha: str) -> str:
...
+        semgrep_repo_name = os.getenv("SEMGREP_REPO_NAME")
+
+        # see NOTE: no-super
+        if semgrep_repo_name:
+            return semgrep_repo_name
+        if project_path:
+            return project_path
+        return super().repo_name
```
brendongo, line 627: Why see NOTE: no-super when you call super?

underyx: super, no, I also see a NOTE: no-super note and want no NOTE: no-super note

cli/src/semgrep/meta.py
brandonspark: yep, that's fair, fixed

cli/src/semgrep/meta.py
brandonspark: I suppose this note is meant to be moreso describing why I don't call `super()` _first_. I isolate the part of the `super()` call that I want, which is the `SEMGREP_REPO_NAME`, but because the `super()` call is otherwise side effecting, I avoid calling that first, prior to looking at the actual `CI_PROJECT_PATH` env var.

I can be more explicit about that.

brandonspark:
@brendongo sorry for forgetting about this for a while, i was away for a bit and preoccupied with some other things

chmccreery (approved):

PR #7365, opened 2023-03-20
feat: support webhooks for Zapier rather than relying on GH integrations

spencerdrak:
This will simplify follow-on actions we take in Zapier.

I'll be manually testing this step once the PR has approval - Zapier needs to be prepared to accept this request. 

Confirmed that because URL is kept as secret, GH will mask any occurrences in the log output. 

PR checklist:

### DIFF
--- a/.github/workflows/start-release.yml
+++ b/.github/workflows/start-release.yml
@@ -430,6 +430,9 @@ jobs:
       - run: echo "${{needs.get-version.outputs.version}}"
       - name: Notify Success
         run: |
+          curl "${{ secrets.ZAPIER_WEBHOOK_URL }}" \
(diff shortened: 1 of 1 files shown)


pabloest (approved):
Very cool, @spencerdrak!

.github/workflows/start-release.yml
```diff
@@ -430,6 +430,9 @@ jobs:
       - run: echo "${{needs.get-version.outputs.version}}"
       - name: Notify Success
         run: |
+          curl "${{ secrets.ZAPIER_WEBHOOK_URL }}" \
```
aryx, line 434: What does this do? A comment would be nice, it's not obvious from the code what is the effect of the command. Does it post something on Zapier? What is the URL where we can see the pushed information?

spencerdrak: Addressed - thanks!

cgdolan (approved):

PR #7366, opened 2023-03-21
Don't install libraries or executables when running 'make build'.

mjambon:
`make` was calling `dune install` for no good reason. This was installing any internal library or executable declared as public.

Fixed:
- `make` = `make build`: no longer installs OCaml libraries or executables.
- `make install`: installs `semgrep-core` and `semgrep` correctly. This now installs from the local Python code rather than installing the public semgrep package. This doesn't install OCaml libraries or executables other than `semgrep-core`.

New:
- `make uninstall`: uninstall what may have been installed by `make install`.

These makefile targets are intended for semgrep developers only so I don't think we should document them as part of the changelog.

Test plan: let's see if CI breaks.

PR checklist:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,7 +6,7 @@ PR checklist:
 - [ ] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
 - [ ] Change has no security implications (otherwise, ping security team)
 
-If you're unsure on any of this, please see:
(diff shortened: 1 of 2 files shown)


mjambon:
@cgdolan @spencerdrak I believe you got added to the PR because I fixed a typo in the PR template. Same for @bmahe ?

spencerdrak:
@mjambon that's correct, happy to drop a review if you'd like but I think other folks will know the internals of this better.

aryx (approved):
Nice

PR #7380, opened 2023-03-22
fix(typescript): restore GA parse rate

brandonspark:
`angular` was causing a few errors, which didn't seem to be on our OCaml side in the log, since they just reported "FAILED TO FULLY PARSE" (meaning a `tree-sitter` partial parse error). I decided to just delete it. They seemed to have `test` files too:
image

image

PR checklist:

### DIFF
--- a/stats/parsing-stats/lang/typescript/projects.txt
+++ b/stats/parsing-stats/lang/typescript/projects.txt
@@ -9,7 +9,7 @@
 #
 
 https://github.com/ant-design/ant-design
-https://github.com/angular/angular
+# https://github.com/angular/angular
 https://github.com/ionic-team/ionic
 https://github.com/angular/angular-cli
 https://github.com/akveo/ngx-admin


nmote:
I'm okay with removing benchmark repos when it makes sense, but I don't think that the fact that they have parse errors, when we are trying to measure our parse rate, is a sufficient justification. Could you convince me that we shouldn't care about these particular parse errors?

stats/parsing-stats/lang/typescript/projects.txt
```diff
@@ -9,7 +9,7 @@
 #
 
 https://github.com/ant-design/ant-design
-https://github.com/angular/angular
+# https://github.com/angular/angular
```
aryx, line 12: I agree with Nat. At least you need to explain why this was commented.
Like for skipping the OCaml repo in ../ocaml/projects.txt I've added a comment explaining why it was skipped.

nmote:
I'm removing myself as a reviewer so that this doesn't show up in my queue. Feel free to re-add me if you make an update.

aryx:
Closing. Cleaning up PRs.

PR #7382, opened 2023-03-22
fix(jsonnet): support relative paths for imports of local files

emjin:
Locally imported files are supposed to be supported as valid paths for rules, but currently we load them incorrectly if `rule.yaml` is called from a different directory (e.g. `rules/rule.yaml`). Fixed by this PR.

Test plan: existing jsonnet hook in precommit should catch most cases. For an example of the `libjsonnet` case working, talk to Emma.

PR checklist:


aryx (approved):
Would be nice to have a regression test for it
(useful to check this feature is also implemented in osemgrep when the port is finished).

PR #7390, opened 2023-03-23
fix: remove old checkouts on non-hermetic runners

spencerdrak:
Leftover moved submodules were causing issues on non-hermetic runners.

PR checklist:


emjin (approved):
Nuclear but I see no reason why we shouldn't do it. Do we also need it for linux?

spencerdrak:
@emjin agreed - this is what we're doing in every other case, including the non-hermetic runners in Semgrep Pro as well as any runner that is GH hosted, since we get a new env each time. 

No, linux uses a GH hosted runner, with a fresh env each run.

Once I get a codeowner review I'll merge.

bmahe (approved):

PR #7396, opened 2023-03-25
Fix some aspects of target filtering in osemgrep

mjambon:
~~Current status: I'm not completely done with this branch. There are still some revisions to make after @aryx's review from one week ago, and I need to review the whole thing, in particular the `--project-root` option which I added.~~

Notably:
* No longer convert scanning roots and project roots to a physical path. These input files may be symbolic links and we're fine with it like in the original semgrep CLI.
* Return targets skipped by gitignore/semgrepignore

We still don't pass the e2e test for semgrepignore because the JSON output is slightly different. The reference output produced by legacy semgrep lacks the following:
* skipped targets `.gitignore` and `.semgrepignore`
* the field `core.engine_requested`

I'm sure there will be a lot more cases where osemgrep and semgrep don't return exactly the same JSON. Sometimes it's because the order of some fields doesn't matter. In other cases, it's because there's an improvement in osemgrep. What we care about is catching regressions but we don't want to fix old bugs in the legacy implementation. I need to think about a solution. Perhaps it will involve comparing JSON data in a smarter way than exact matches.

PR checklist:


src/osemgrep/cli_scan/Scan_CLI.ml
```diff
@@ -154,6 +164,21 @@ let o_version_check : bool Term.t =
  *  rules' languages. These options alter which files Semgrep scans."
  *)
 
+let o_project_root : string option Term.t =
```
aryx: please add a big comment like (* NEW: This is a new option! it's not in legacy semgrep *).
I actually put the new flags in another section below.

src/osemgrep/cli_scan/Scan_CLI.ml
```diff
@@ -587,9 +612,9 @@ let cmdline_term : conf Term.t =
   (* !The parameters must be in alphabetic orders to match the order
    * of the corresponding '$ o_xx $' further below! *)
   let combine autofix baseline_commit config debug dryrun dump_ast dump_config
-      emacs error exclude exclude_rule_ids force_color include_ json lang
-      max_memory_mb max_target_bytes metrics num_jobs optimizations pattern
-      profile quiet replacement respect_git_ignore rewrite_rule_ids
+      emacs error project_root exclude exclude_rule_ids force_color include_
```
aryx: NO! see the comment above, please keep the order alphabetic.

aryx (approved):
This is a lot of unrelated changes put together ... I understand you had to do this
in a hurry, and there's the pressure of Robur, but in the future please try to make
separate PR for separate fixes.

src/osemgrep/cli_scan/Scan_CLI.ml
mjambon: I appreciate your newfound enthusiasm for keeping lists ordered alphabetically 😄

mjambon:
The goal of the PR was to make the test_semgrepignore test succeed, which it still doesn't due to irrelevant differences in JSON output. I'm going to merge this because it fixes basic file targeting and I'll solve the JSON issue in a separate PR.

PR #7410, opened 2023-03-28
bugfix: fix bug with -raja on List.hd exn

aryx:
test plan:
in juice-shop directory:
semgrep --config p/default --core-opts="-raja"
does not thrown an error anymore.

PR checklist:

### DIFF
--- a/src/experiments/misc/Raja_experiment.ml
+++ b/src/experiments/misc/Raja_experiment.ml
@@ -35,11 +35,17 @@ let ranges_of_path (path : Fpath.t) : Function_range.ranges =
   match Hashtbl.find_opt cache path with
   | Some x -> x
   | None ->
-      (* alt: could use just_parse_with_lang and get the lang
-       * from the languages: field in the rule corresponding to
-       * the rule_id of the match
-       *)
-      let ast = Parse_target.parse_program (Fpath.to_string path) in
+      let ast =
+        (* parse_program may raise the "hd" exn when we can't infer
+         * the language from the filename.
+         * alt: we could use just_parse_with_lang and get the language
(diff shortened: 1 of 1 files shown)


src/experiments/misc/Raja_experiment.ml
```diff
@@ -35,11 +35,17 @@ let ranges_of_path (path : Fpath.t) : Function_range.ranges =
...
   | None ->
-      (* alt: could use just_parse_with_lang and get the lang
-       * from the languages: field in the rule corresponding to
-       * the rule_id of the match
-       *)
-      let ast = Parse_target.parse_program (Fpath.to_string path) in
+      let ast =
+        (* parse_program may raise the "hd" exn when we can't infer
+         * the language from the filename.
+         * alt: we could use just_parse_with_lang and get the language
```
emjin: I definitely like the alt suggestion better. This is an experiment so ok with this for now, but maybe change the "alt" to "TODO"

emjin (approved):

mjambon:
Time to get rid of all uses of `List.hd`? I don't see a justification for using it.

* if receiving an empty list is a bug, use match-with + `assert false`
* if an empty list is ok, use match-with + any coping mechanism other than `Failure "hd"` (`option`, `result`, `failwith msg`, ...)
* if brevity is essential, define and use `hd_opt : 'a list -> 'a option` or `hd_exn : msg:string -> 'a list -> 'a`.

aryx:
Go for it. Write a rule and autofix :)

mjambon:
The problem is I don't know when it's a bug and when it's not. I guess I'll assume it's always a bug.

PR #7415, opened 2023-03-29
Implementation of nosemgrep tag in OCaml

dinosaure:
PR checklist:

This PR wants to implement the `nosemgrep` tag into the `osemgrep` OCaml code as the Python implemention (`semgrep`). Currently, from my handmade tests, the `nosemgrep` tag works. However, the end-to-end test fails due to some other things unrelated to this PR (like absolute path of files into the `scanned` JSON field).

Some questions persists however:
- [ ] do we have a function to _trim_ ID collected?
- [ ] my usage of `cli_error` is correct?
- [ ] `nosem_line_comment_re` seems unused by the current Python implemention, should I delete it?

I make this PR as a draft because I'm still unsure that it's ready to merge.

### DIFF
--- a/libs/commons/File.mli
+++ b/libs/commons/File.mli
@@ -126,3 +126,12 @@ val erase_this_temp_file : Fpath.t -> unit
 (*****************************************************************************)
 val is_executable : Fpath.t -> bool
 val filesize : Fpath.t -> int
+
+(* [lines_of_file (start_line, end_line) file] returns
+ * the list of lines from start_line to end_line included.
+ *
+ * Note that the returned lines do not contain \n.
+ *
+ * This function is slow, you should not use it!
+ *)
+val lines_of_file : int * int -> Common.filename -> string list
--- a/src/osemgrep/reporting/Nosem.ml
+++ b/src/osemgrep/reporting/Nosem.ml
@@ -35,7 +35,7 @@ let rule_id_re_str = {|(?:[:=][\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?|}
    * nosem and nosemgrep should be interchangeable
 *)
 let nosem_inline_re_str = {| nosem(?:grep)?|} ^ rule_id_re_str
-let _nosem_inline_re = SPcre.regexp nosem_inline_re_str ~flags:[ `CASELESS ]
+let nosem_inline_re = SPcre.regexp nosem_inline_re_str ~flags:[ `CASELESS ]
 
 (*
    As a hack adapted from semgrep-agent,
@@ -43,6 +43,7 @@ let _nosem_inline_re = SPcre.regexp nosem_inline_re_str ~flags:[ `CASELESS ]
 *)
 let _nosem_inline_comment_re =
   SPcre.regexp (spf {|[:#/]+%s$|} nosem_inline_re_str) ~flags:[ `CASELESS ]
+(* XXX(dinosaure): seems unused by the last version of `cli/` *)
 
 (*
    A nosemgrep comment alone on its line.
@@ -56,7 +57,7 @@ let _nosem_inline_comment_re =
      # nosemgrep
      print('nosemgrep');
 *)
-let _nosem_previous_line_re =
+let nosem_previous_line_re =
   SPcre.regexp
     ({|^[^a-zA-Z0-9]* nosem(?:grep)?|} ^ rule_id_re_str)
     ~flags:[ `CASELESS ]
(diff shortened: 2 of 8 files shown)


src/osemgrep/reporting/Nosem.ml
```diff
@@ -35,14 +35,15 @@ let rule_id_re_str = {|(?:[:=][\s]?(?P<ids>([^,\s](?:[,\s]+)?)+))?|}
...
-let _nosem_inline_re = SPcre.regexp nosem_inline_re_str ~flags:[ `CASELESS ]
+let nosem_inline_re = SPcre.regexp nosem_inline_re_str ~flags:[ `CASELESS ]
 
 (*
    As a hack adapted from semgrep-agent,
    we assume comment markers are one of these special characters
 *)
 let _nosem_inline_comment_re =
   SPcre.regexp (spf {|[:#/]+%s$|} nosem_inline_re_str) ~flags:[ `CASELESS ]
+(* XXX(dinosaure): seems unused by the last version of `cli/` *)
```
aryx: yes, feel free to remove the dead code here and also in cli/.

src/osemgrep/reporting/Nosem.ml
```diff
@@ -69,11 +70,96 @@ let _nosem_previous_line_re =
...
 
-(* TODO: probably need to pass some flags from the CLI *)
-let process_ignores (out : Out.cli_output) : Out.cli_output =
-  (* TODO: port nosemgrep.py, filter in out.matches the
-     matches where the match has a corresponding nosem: annotation in
-     the corresponding file, and add new errors if nosem: has the wrong
-     format in the file.
-  *)
-  out
+let rule_match_nosem ~strict (rule_match : Out.cli_match) :
```
aryx, line 78: you can move this in the Helpers section (it's not an entry point of the module).

src/osemgrep/reporting/Nosem.ml
```diff
@@ -69,11 +70,96 @@ let _nosem_previous_line_re =
...
+
+  let previous_line, line =
+    match lines with
+    | line0 :: line1 :: _ when rule_match.Out.start.line > 0 ->
+        (Some line0, Some line1)
+    | line :: _ -> (None, Some line)
+    | [] (* XXX(dinosaure): is it possible? *) -> (None, None)
+  in
+
+  let recognise_and_collect ~rex line =
```
aryx: avoid nested functions. Move out in Helper section, and give type and small explanation if needed.

src/osemgrep/reporting/Nosem.ml
```diff
@@ -69,11 +70,96 @@ let _nosem_previous_line_re =
...
+    Result.map
+      (Array.map (fun subst ->
```
aryx: We tend to favor using |> when we can, to be able to 
see first the data structure we operate on, and then the actuall code, so here
like (SPcre.exec_all ~rex line |> Array.map (fun subst ...)

But if you feel strongly about not using |>, that's fine, some of us don't.

src/osemgrep/reporting/Nosem.ml
```diff
@@ -69,11 +70,96 @@ let _nosem_previous_line_re =
...
+           try Some (Pcre.get_named_substring rex "ids" subst) with
+           | _ -> None))
+      (SPcre.exec_all ~rex line)
+    |> Result.to_option
+  in
+
+  let no_ids = Array.for_all Option.is_none in
+
+  let ids_line, ids_previous_line =
+    ( Option.bind line (recognise_and_collect ~rex:nosem_inline_re),
```
aryx: I never used Option.bind, but as a general rule with option we sometimes prefer to be explicit
and use match x | Some -> .. None -> ...
especially if the resulting code is actually not that longer than using Option.map, Option.bind.

aryx (approved):

aryx (changes requested):
Some small requests, but overall looks good.

aryx (approved):

libs/commons/File.mli
```diff
@@ -126,3 +126,12 @@ val erase_this_temp_file : Fpath.t -> unit
...
 val filesize : Fpath.t -> int
+
+(* [lines_of_file (start_line, end_line) file] returns
+ * the list of lines from start_line to end_line included.
+ *
+ * Note that the returned lines do not contain \n.
+ *
+ * This function is slow, you should not use it!
+ *)
+val lines_of_file : int * int -> Common.filename -> string list
```
mjambon, line 137: This module (which was created recently) should use only `Fpath.t` for file paths. We're moving away from `Common.filename` which is an alias for `string`. We haven't finished migrating all our code to `Fpath`, so if it's too inconvenient to use the fpath version of this function, move the non-fpath version to the `Common` module and have a wrapper using fpath in the `File` module.

PR #7416, opened 2023-03-29
tainting: Add new option to avoid taint propagation via comparisons

IagoAbal:
Helps PA-2645

test plan:
make test # new test added

PR checklist:

### DIFF
--- a/changelog.d/pa-2645.changed
+++ b/changelog.d/pa-2645.changed
@@ -1,0 +1,4 @@
+taint-mode: Added option `taint_assume_safe_comparisons`, disabled by default, that
+prevents comparison operators to propagate taint, so e.g. `tainted != "something"`
+will not be considered tainted. Note that this a syntactic check, if the operator
+is overloaded to perform a different operation this will not be detected.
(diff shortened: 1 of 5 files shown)


IagoAbal:
Opened ticket to update docs: PA-2670

changelog.d/pa-2645.changed
```diff
@@ -0,0 +1,4 @@
+taint-mode: Added option `taint_assume_safe_comparisons`, disabled by default, that
+prevents comparison operators to propagate taint, so e.g. `tainted != "something"`
+will not be considered tainted. Note that this a syntactic check, if the operator
+is overloaded to perform a different operation this will not be detected.
```
mjambon, line 4: I don't know where you are in the doc writing, but I think it would be great to explain in what scenario this option is useful. Is it because comparing a variable against any other value is assumed to always be part of some sort of sanitization? When is it a safe assumption and when is it not?

mjambon (approved):

aryx (approved):
should we set the default to true instead?

changelog.d/pa-2645.changed
IagoAbal: I'll clarify this better in the changelog, but this option simply makes the comparison expression `tainted != "something"` (or `tainted >= something`, etc) not tainted, however `tainted` will continue to be tainted. Previously, the result of comparing `tainted` against some other value, that is the resulting Boolean `true`/`false` was considered tainted.

IagoAbal:
> should we set the default to true instead?

Ideally, but I'm worried about doing that without using type info to make sure the operators are truly comparison operators. If the operator is overloaded and it actually means something else, then we may be doing it wrong. Don't know if this is common or uncommon in the languages we support. For now it's mainly something for @LewisArdern to experiment with.

PR #7419, opened 2023-03-29
feat(scala): parsing for quoted exprs, using params, soft modifiers

brandonspark:
## What:
This PR adds in parsing for a few Scala constructs, namely quoted expressions/patterns, `using` parameters, and soft modifiers like "inline" and "open".

## Why:
This will increase our Scala parse rate.

## How:
The first two were relatively straightforward. I decided to make `bindings` have a `tok option`, rather than attach anything in the parameter itself, because parameters can only have attributes that are annotations or modifiers. `using` is not truly a modifier in the sense of the Scala grammar, but something which is explicitly specified as existing as a separate kind of parameter, so I felt it was better to separate it out.

For soft modifiers, I just inserted the logic as stated in the Scala 3 specification, which is that a soft modifier is only a modifier if it is followed by a modifier, or the start of a declaration.

I could have separated this into several PRs but I thought that would get really annoying.

## Test plan:
Parse rate went up. Before we had `0.918339362478076` (without `finagle` because it causes a segfault on my machine), but after we get `0.9189344037430826`

Also, `make test`.

PR checklist:


aryx (approved):
Great work. A bit sad that it does not improve that much the parse rate. Still, this is
progress.

PR #7422, opened 2023-03-30
feat(scala): parse match types

brandonspark:
## What:
This PR adds the ability to parse match types, which are type expressions that case on the particular type of something.

## Why:
Trying to parse more programs.

## How:
I used the existing `case_clause` code and parameterized it, as it seems in the Scala grammar there's quite a few things of that form, except with different nonterminals for the LHS and RHS of the cases.

I put it into a `Switch` and then embedded that into `type_`, as it seems we have no actual analogue in `type_` as of right now. I wanted to preserve usage of the ellipses, and I thought this was the most explicit construct that denoted what it was doing. I don't think we'll need a special other construct for this for a very long time, if ever.

## Test plan:
Automated tests, and parse rate went from `0.9189344037430826` to `0.9189534728003734`

PR checklist:

### DIFF
--- a/changelog.d/pa-2673.added
+++ b/changelog.d/pa-2673.added
@@ -1,0 +1,4 @@
+Scala: Can now parse programs containing matches on types, such as:
+type t = K match {
+  case Int => String
+}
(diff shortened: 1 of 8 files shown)


changelog.d/pa-2672.added
```diff
@@ -0,0 +1,2 @@
+Scala: Semgrep can now parse programs that contain quoted expressions, context
```
aryx: This should not be in this PR.

aryx (commented):
This mix 2 PRs in one

aryx (approved):
Nice work.

PR #7424, opened 2023-03-30
feat(SSC): Use go.mod files instead of go.sum files

mmcqd:
[`go.sum` files are not lockfiles](https://stackoverflow.com/questions/69825533/why-does-go-sum-include-so-many-older-packages), and they contain multiple copies of versions of packages that are not actually being used by a project. This PR switches to using `go.mod` files as a source of dependency information for go projects. `go.mod` files actually _do_ function as lockfiles, as the go package manager automatically edits them to include only exact versions and all transitive dependencies.

PR checklist:


aryx:
Note that for semgrep-pro @nmote also needed to parse go.mod files to resolve correctly names. We made a very
shallow parser for those go.mod files, but having all our parsers in OCaml would improve things a little bit here and avoid
code duplication.

nmote:
Yeah, it definitely would be nice to have things consolidated. But, in this case the "parser" I wrote is dead simple and only a few lines long. I don't think there's a huge benefit to be had in this specific case by avoiding duplication.

cli/tests/performance/test_semdep.py
```diff
@@ -14,7 +14,6 @@
         for file_size, max_time in [("10k", 3), ("50k", 15), ("100k", 30)]
         for target in [
             "Gemfile.lock",
-            "go.sum",
```
brendongo, line 17: Do we want to add go.mod to the perf tests?

brendongo (approved):

cli/tests/performance/test_semdep.py
mmcqd: done

PR #7427, opened 2023-03-30
fix(autofix): Fix rare Python argument reordering

nmote:
Due to unordered keyword matching, and how it leads ellipsis metavariables to be bound to arguments, in some cases Semgrep can generate autofixes that reorder keyword arguments. In Python, this is a problem when it moves keyword arguments in front of positional arguments, since Python disallows that.

This addresses this case by adding a transformation step after the fix is generated that moves positional arguments in front of keyword arguments.

I've also included a change so that the `Textedit` module only reports conflicting edits if they overlap and are not identical. This allows the test cases I added to pass. Semgrep generates multiple matches, but due to the changes here they end up with the same autofix.

Partially addresses #4963

Test plan: Automated tests

PR checklist:

### DIFF
--- a/changelog.d/keywordarg-autofix.fixed
+++ b/changelog.d/keywordarg-autofix.fixed
@@ -1,0 +1,1 @@
+In rare cases, Semgrep could generate invalid autofixes where Python keyword arguments were placed before positional arguments. When using AST-based autofix, it no longer makes that error.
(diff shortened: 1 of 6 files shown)


emjin (approved):
lgtm

aryx (commented):
AST-based autofix ftw!

PR #7428, opened 2023-03-30
feat(scala): indentation tokens

brandonspark:
## What:
This PR adds in `INDENT` and `DEDENT` tokens to the token stream of Scala programs, to be consumed at a later date.

## Why:
In the Scala 3 specification, there is a notion of "optional braces" (https://docs.scala-lang.org/scala3/reference/syntax.html#optional-braces), where explicit curly braces can be omitted in favor of clear indentation and dedentation.

As it happens, right now, we only insert tokens for some number of spaces and newlines. Things like whether there is an indent or dedent are lost on us, so we don't have enough information in the token stream to distinguish cases like
```
object Foo:
  val x = 2
val y = 3
```
and
```
object Foo:
  val x = 2
  val y = 3
```
(actually, we might, because there would be no space before `val y = 3` in the first, but imagine this was further nested one level, and you get the point)

## How:
I wrote a fold which just goes and inserts `INDENT` and `DEDENT` tokens any time that it finds we are on a new line, and the column has changed with respect to the start column on the prior line.

Every other function which skips newlines and spaces and stuff also skips these tokens, so they are functionally "invisible" unless you explicitly go looking for them, which future functions can be written to do (such as for detection of these optional braces).

## Test plan:
Parse rate has not changed.
image

I also manually inspected the token stream for an example:
```
{
<
  >
    {

    }
  {

}
}
```
produces
```
[(LBRACE ()); (Nl ()); (OP ("<", ())); (Nl ()); (Space ()); INDENT;
  (OP (">", ())); (Space ()); (Nl ()); (Space ()); INDENT; (LBRACE ());
  (Space ()); (Nl ()); (Nl ()); (Space ()); (RBRACE ()); (Nl ()); (Space ());
  DEDENT; (LBRACE ()); (Nl ()); (Nl ()); DEDENT; (RBRACE ()); (Nl ());
  (RBRACE ()); (Nl ()); (EOF ())]
```
As we can see, we get two indents and two dedents, in the right places.

PR checklist:

### DIFF
--- a/languages/scala/recursive_descent/Parsing_hacks_scala.ml
+++ b/languages/scala/recursive_descent/Parsing_hacks_scala.ml
@@ -1,6 +1,6 @@
 (* Brandon Wu
  *
- * Copyright (C) 2019 r2c
+ * Copyright (C) 2023 r2c
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public License
--- a/languages/scala/recursive_descent/Parsing_hacks_scala.ml
+++ b/languages/scala/recursive_descent/Parsing_hacks_scala.ml
@@ -1,0 +1,55 @@
+(* Brandon Wu
+ *
+ * Copyright (C) 2019 r2c
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * version 2.1 as published by the Free Software Foundation.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


aryx:
There is some code related to indent/dedent in languages/python/menhir/Lexer_python.mll
Dunno if there's some idea you could steal from this file.

languages/scala/recursive_descent/Parsing_hacks_scala.ml
```diff
@@ -0,0 +1,55 @@
+(* Brandon Wu
+ *
+ * Copyright (C) 2019 r2c
```
aryx: 2023

aryx (approved):

brandonspark:
@aryx i looked at that, its pretty similar i think, and i think ultimately implements what i did in this PR

PR #7429, opened 2023-03-30
Eliminate List.hd and List.tl

mjambon:
The automatic replacement by `Common.hd_exn` and `Common.tl_exn` is not a big improvement but it would take a while to address all these issues at the source. I did a manual replacement in a few cases before giving up and applying the default replacement everywhere else.

Note that I have never used `List.hd` or `List.tl` in my own code, usually preferring something like:
```ocaml
match xs with
| [] -> assert false
| x :: xs -> ...
```

This is from the days when OCaml didn't show stack traces*, making it really hard to find where an exception was raised. `assert false` includes the source location in the exception payload, so it's been my go-to solution since. With stack traces, both `Assert_failure` and `Failure "hd"` look bad to a user anyway, and a stack trace solves the location issue.

\* Does OCaml always show stack traces now?

PR checklist:


src/experiments/misc/Raja_experiment.ml
```diff
@@ -43,7 +44,9 @@ let ranges_of_path (path : Fpath.t) : Function_range.ranges =
          * the rule_id of the match. That would require to pass
          * more info to ranges_of_path() though.
          *)
-        try Parse_target.parse_program (Fpath.to_string path) with
+        try Parse_target.parse_program !!path with
+        (* TODO: does this still catch what it's supposed to catch
+           since we no longer use List.hd? *)
         | Failure "hd" -> []
```
mjambon, line 50: @aryx I need your help here. I don't know what raises the `Failure "hd"` we're expecting here.

aryx: yes, I'll fix your branch today hopefully with the right fix. We should throw an exn like
CantInferLangFromFilename in Parse_target.parse_program instead of abusing Failure "hd".

IagoAbal:
I don't use `List.hd` or `List.tl` either, perhaps very rarely. I think `ocamlopt` only generates stack traces with `-g`. I try to avoid `assert false` in Semgrep unless I'm pretty sure something will never happen, otherwise I log an error message and I try to let Semgrep keep running.

PR #7433, opened 2023-03-30
feat(scala): template optional braces

brandonspark:
## What:
This PR makes it such that, when parsing a `Template`, we can have optional braces, signalled by a colon, which is terminated upon dedentation of the current indentation block.

This is a refinement of the first iteration of this PR, which made some choices that were out of haste, and because they worked, but without understanding why or whether it was a good idea. In particular, I made it so stat separators could accept any token, because NEWLINEs weren't being emitted as I wanted. This is a bad idea, and in this iteration we go and do it how the Scala 3 compiler does it.

## Why:
This is part of the Scala 3 grammar.

## How:
There's a few things that were done here.

The lexing hack to insert INDENTs and DEDENTs was made more sophisticated. Previously, we inserted them willy-nilly based on column relations between lines, but this actually ends up producing a whole lot of INDENTs and DEDENTs, not all of which are relevant. We followed the specification of the Scala 3 compiler here, and only inserted INDENTs after certain keywords. In addition, we only inserted DEDENTs when a previous "indentation region" was ended, i.e. there was a token to the left of a previously emitted INDENT.

In parsing, we also got more particular. Instead of just accepting any stat separator, we changed it so we properly emit NEWLINEs while inside of indented regions. We do this by treating indents like LBRACE and RBRACE, in that they may appear in the `sepRegions`. This is tricky, though, and doesn't work exactly the same, for a few reasons:
- `INDENT` is not a real token, so it must be skipped by `nextToken`. Otherwise, you would need to insert logic that looks for indents in many unrelated places, and this would mess with our ability to predict the next nonterminal by looking at a few "real" tokens of lookahead.
- Since `DEDENT` is the same way, this means we may need to play catch-up with many implicit tokens that were skipped, particularly because a DEDNT might end many indentation regions. This is different than `LBRACE` and `RBRACE`, where we can always trust the call stack to go and pop them in the right order, and you can never be on multiple RBRACEs "at once". Dealing with playing catch-up produces extremely messy code.
- Not all `INDENT` and `DEDENT`s will be used for indentation regions. Only some are.

So what we do is as the Scala compiler does. We end up stopping on DEDENTs when they end the current indentation region. This means that it could possibly mess with our lookahead, but it doesn't matter, because this should only happen in cases where we deliberately choose to enter an indentation region. This means we can still use the call stack to pop the `DEDENT`s.

## Test plan:
Parse rate goes up.
`0.9382998981972374`

Note that this is very slightly smaller than the previous iteration of this PR, which was
```
lang/scala/stats.json:3
   "global": {
     "name": "*",
     "parsing_rate": 0.938544762228358,
     "line_count": 2307403,
     "error_line_count": 141802,
     "file_count": 14684,
     "error_file_count": 1031,
     "total_node_count": 31164817,
     "untranslated_node_count": 103526
   }
```
but I think it's fine, because that PR probably allowed some false parsing.

(previous rate: `0.9189534728003734`)

I also manually verified that the two added test cases parse to
```
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("CodeGenerater", ()),
                {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[]; tparams=[]; },
      ClassDef(
        {ckind=(Class, ()); cextends=[]; cimplements=[]; cmixins=[]; cparams=[
         ];
         cbody=[F(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("foo", ()),
                               {id_info_id=2; id_hidden=false; id_resolved=Ref(
                                None); }));
                      attrs=[]; tparams=[]; },
                     FuncDef({fkind=(Method, ()); fparams=[]; frettype=None; fbody=FBExpr(L(Int((Some(3), ())))); }))));
                F(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("bar", ()),
                               {id_info_id=3; id_hidden=false; id_resolved=Ref(
                                None); }));
                      attrs=[]; tparams=[]; },
                     FuncDef({fkind=(Method, ()); fparams=[]; frettype=None; fbody=FBExpr(L(Int((Some(4), ())))); }))))];
         })));
   DefStmt(
     ({
       name=EN(
              Id(("x", ()),
                {id_info_id=4; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[KeywordAttr((Const, ()))]; tparams=[]; },
      VarDef({vinit=Some(L(Int((Some(2), ())))); vtype=None; })))]
```
and
```
Pr(
  [DefStmt(
     ({
       name=EN(
              Id(("CodeGenerater", ()),
                {id_info_id=1; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[]; tparams=[]; },
      ClassDef(
        {ckind=(Object, ()); cextends=[]; cimplements=[]; cmixins=[]; cparams=[
         ];
         cbody=[F(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("foo", ()),
                               {id_info_id=2; id_hidden=false; id_resolved=Ref(
                                None); }));
                      attrs=[]; tparams=[]; },
                     FuncDef({fkind=(Method, ()); fparams=[]; frettype=None; fbody=FBExpr(L(Int((Some(3), ())))); }))));
                F(
                  DefStmt(
                    ({
                      name=EN(
                             Id(("bar", ()),
                               {id_info_id=3; id_hidden=false; id_resolved=Ref(
                                None); }));
                      attrs=[]; tparams=[]; },
                     FuncDef({fkind=(Method, ()); fparams=[]; frettype=None; fbody=FBExpr(L(Int((Some(4), ())))); }))))];
         })));
   DefStmt(
     ({
       name=EN(
              Id(("x", ()),
                {id_info_id=4; id_hidden=false; id_resolved=Ref(None); id_type=Ref(None); id_svalue=Ref(None); }));
       attrs=[KeywordAttr((Const, ()))]; tparams=[]; },
      VarDef({vinit=Some(L(Int((Some(2), ())))); vtype=None; })))])
```
As we can see, the `val x` is properly outside of the `class` or `object`.

PR checklist:


languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
```diff
@@ -456,6 +456,38 @@ let lookingAhead2 body1 body2 in_ =
...
+
+     def bar() = 3
+
+   We ideally want to keep scanning the contents of Foo until we find a DEDENT,
+   which presumably signals the end of the class block. Unfortunately, we might
+   run into a DEDENT which is not the end of the class, such as the DEDENT
+   published right before the second "def", due to the indentation difference
+   between it and the 3.
+
+   So instead, we can just check the columns. We follow a rule that the first token
```
aryx: Is it an official Scala rule?

languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
```diff
@@ -3387,23 +3419,34 @@ let templateStatSeq ~isPre in_ : self_type option * block =
...
 *)
 
 let templateBody ~isPre in_ : template_body =
-  (* ast: self, EmptyTree.asList *)
-  inBraces (templateStatSeq ~isPre) in_
+  match in_.token with
+  | LBRACE _ ->
+      (* ast: self, EmptyTree.asList *)
+      inBraces (templateStatSeq ~isPre) in_
+  | _ ->
```
aryx, line 3554: Should we check that it's actually a Colon here?

aryx (approved):

languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
brandonspark: the `accept` will throw an error anyways, so it's the same as throwing an explicit new error in the third wildcard case. by convention, in other parts of this parser where we narrow down to a single case, we generally wildcard it and just enforce constraints via exceptions

languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
brandonspark: It's actually more like how I do it now, where we check the columns _as they relate to previous indentation regions_.

brandonspark:
@aryx i changed this PR since, would appreciate another look

aryx (approved):

PR #7437, opened 2023-03-31
fix(autofix/python): Fix range of expressions wrapped in parens

### LINKED ISSUE #2902: Invalid pattern capture with parenthesis for Python
**Describe the bug**
When capturing patterns for Python, semgrep ignores parenthesis matching in some cases.

**To Reproduce**
For the code below:

```python
    if cond(1, 2, z) or not (z(x) is k(y)) or cond333(3 ** x):
        return
```

Pattern: `not $X is $Y`

**Expected behavior**
Do not capture the `not (z(x) is k(y))`, because it is `not ($X is $Y)` pattern.

Since it captures the last closing parenthesis `)` wrongly, it is impossible to write a safe autofix rule to fix this pattern to `$X is not $Y`.

**Screenshots**
image
As you can see the screenshot, it wrongly captures `not (z(x) is k(y)` **ignoring** the last closing parenthesis `)`.

If this behaviour is pervasive in Python, it should be fixed quickly.

nmote:
Previously, the `$X` metavariable binding in the new fix_binop.py example was given the range associated with `1 + (5 + 3`. This is because the parentheses get dropped by the parser, so the location of the second plus operation is associated with only `5 + 3`. In isolation that's fine, but in cases like this where it is part of a larger expression, this can mean that one parenthesis is included in the range while the other is not.

This issue affects both text-based autofix and AST-based autofix. Both need to know which snippet to replace with the fix, and both need to know the correct location of the original text for bound metavariables.

My first attempt to address this was to create `ParenExpr` nodes in the generic AST, which are already used on a limited basis for Go. Even after adding cases to unwrap them in `m_expr`, there was still a fairly large difference in matching behavior. I assume that there are a lot of places outside of `m_expr` that match on expression nodes, and we'd have to add unwrapping logic everywhere that happens. This seemed error prone and hard to maintain. This is another case that illustrates the tension between an AST, most useful for matching and analysis, and a CST, most useful for manipulating and printing code.

Instead, I decided to write the range, including the parentheses, to `e_range` when the expression node was created. This is a bit of a departure from the way e_range has historically been used. Until now, as far as I know, it has only been a cache. Now, it serves as the canonical source of range information in this case. I think that this is a good direction to move in -- many of our location issues stem from some tokens not making it into the resulting AST, so I think that explicitly setting the range based on locations from the parser, when the node is constructed, will help. That way, we would be free to drop semantically meaningless tokens from the AST without worrying that they are needed to compute locations.

Fixes #2902

Test plan:

- Automated tests
- Manual test with the example from #2902 (https://semgrep.dev/playground/s/20WB)
  - Also manual test after putting the tree sitter parser first in Parse_target2.ml

PR checklist:

### DIFF
--- a/libs/ast_generic/AST_generic_helpers.ml
+++ b/libs/ast_generic/AST_generic_helpers.ml
@@ -401,3 +401,16 @@ let undo_ac_matching_nf tok op : expr list -> expr option = function
         |> G.e
       in
       Some (List.fold_left mk_op (mk_op a1 a2) args)
+
+let set_e_range l r e =
+  match
+    (Parse_info.token_location_of_info l, Parse_info.token_location_of_info r)
+  with
+  | Ok l, Ok r -> e.e_range <- Some (l, r)
+  | Error _, _
+  | _, Error _ ->
+      (* Probably not super useful to dump the whole expression, or to log the
+       * fake tokens themselves. Perhaps this will be useful for debugging
+       * isolated examples, though. *)
+      logger#debug "set_e_range failed: missing token location";
+      ()
--- a/src/fixing/Autofix.ml
+++ b/src/fixing/Autofix.ml
@@ -63,6 +63,12 @@ let transform_fix lang ast =
           mk_visitor
             {
               default_visitor with
+              (* The default visitor behavior is to create a new expr node every
+               * time, with a new e_id and e_range. Avoid this by overriding
+               * such that we take the new expr_kind and put it in the original
+               * expression. *)
+              kexpr =
+                (fun (k, _) expr -> AST_generic.{ expr with e = (k expr).e });
               kargs =
                 (fun (k, _) args ->
                   let args =
--- a/changelog.d/gh-2902.fixed
+++ b/changelog.d/gh-2902.fixed
@@ -1,0 +1,1 @@
+Fixed an issue where incorrect ranges for expressions containing parentheses could lead Semgrep to generate invalid autofixes in Python.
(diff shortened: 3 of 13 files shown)


libs/ast_generic/AST_generic_helpers.ml
```diff
@@ -401,3 +401,12 @@ let undo_ac_matching_nf tok op : expr list -> expr option = function
...
       Some (List.fold_left mk_op (mk_op a1 a2) args)
+
+let set_e_range l r e =
+  match
+    (Parse_info.token_location_of_info l, Parse_info.token_location_of_info r)
+  with
+  | Ok l, Ok r -> e.e_range <- Some (l, r)
+  | Error _, _
+  | _, Error _ ->
+      ()
```
aryx, line 416: we might want to log the error (not that anybody will ever look at those logs ...)

src/fixing/Autofix.ml
```diff
@@ -63,6 +63,8 @@ let transform_fix lang ast =
           mk_visitor
             {
               default_visitor with
+              kexpr =
```
aryx, line 70: What does this do?
need a comment. I don't get it.

aryx (approved):
I like it. Yes we should gradually switch to use e_range.

src/fixing/Autofix.ml
nmote: Oh sorry, meant to comment here but forgot... will do.

PR #7445, opened 2023-04-03
Osemgrep: lang rules matrix

hannesm:
This adds prelimary "Scan Status" output, similar to what is in the python version.

Some questions are remaning:
- (a) the Python code prints an origin table, based on "rule.metadata, ("semgrep.dev", "rule", "origin"), default="unknown" -- we should probably do the same
 - (b) stdout vs stderr -- esp. with json output, I suspect we should output the "Scan status" to stderr?
 - (c) supply chain analysis -- would you mind to educate me with an example passed to semgrep so I can see the output?
 - (d) pro rules (see (a)), would be great to have a running example for this as well

Now, there are some differences, as seen below (indentation, "<multilang>" vs "generic" and "regex"). I tried to get indentation with OCaml's formatter but wasn't lucky.

Python output (on semgrep --config r/ocaml --config r/python src/osemgrep/):
    
```
┌─────────────┐
│ Scan Status │
└─────────────┘
Scanning 160 files tracked by git with 429 Code rules:
    
<multilang>       5     159          Community     429
ocaml            32      86
python          385      57
    
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:02
```
    
Osemgrep output:
    
```
---------------
| Scan status |
---------------
Scanning 160 files tracked by git with 429 Code rule:
    
Language Rules Files
--------------------
generic      8  160
Python     385   57
OCaml       32   86
regex        4  160
```

PR checklist:

### DIFF
--- a/src/osemgrep/cli_scan/Core_runner.ml
+++ b/src/osemgrep/cli_scan/Core_runner.ml
@@ -201,15 +201,13 @@ let pp_status rules targets lang_jobs respect_git_ignore ppf () =
   Fmt_helpers.pp_table
     ("Language", [ "Rules"; "Files" ])
     ppf
-    (List.combine
-       (Common.map
-          (fun (job : Runner_config.lang_job) ->
-            Xlang.to_string job.Runner_config.lang)
-          lang_jobs)
-       (Common.map
-          (fun Runner_config.{ targets; rules; _ } ->
-            [ List.length rules; List.length targets ])
-          lang_jobs))
+    (lang_jobs
+    |> List.fold_left
+         (fun acc Runner_config.{ lang; targets; rules } ->
+           (Xlang.to_string lang, [ List.length rules; List.length targets ])
+           :: acc)
+         []
+    |> List.rev)
 
 (*************************************************************************)
 (* Entry point *)
--- a/src/osemgrep/cli_scan/Core_runner.ml
+++ b/src/osemgrep/cli_scan/Core_runner.ml
@@ -181,55 +181,8 @@ let runner_config_of_conf (conf : conf) : Runner_config.t =
         version = Version.version;
       }
 
-let pp_table (h1, heading) ppf entries =
-  let int_size i =
-    let rec dec acc = function
-      | 0 -> acc
-      | n -> dec (succ acc) (n / 10)
-    in
-    dec 1 i
-  in
-  let len1, lengths =
-    let acc = Common.map String.length heading in
-    List.fold_left
-      (fun (n1, needed) (c1, curr) ->
-        ( max (String.length c1) n1,
-          Common.map2 max needed (Common.map int_size curr) ))
-      (String.length h1, acc)
-      entries
-  in
-  let llen = List.fold_left (fun acc w -> acc + w + 1) len1 lengths in
-  let line = String.make llen '-' in
-  let pad str_size len =
-    let to_pad = succ (len - str_size) in
-    String.make to_pad ' '
-  in
-  Fmt.pf ppf "%s%s" h1 (pad (String.length h1) len1);
-  List.iter2
-    (fun h l -> Fmt.pf ppf "%s%s" h (pad (String.length h) l))
-    heading lengths;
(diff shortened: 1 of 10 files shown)


src/osemgrep/cli_scan/Core_runner.ml
```diff
@@ -181,16 +181,96 @@ let runner_config_of_conf (conf : conf) : Runner_config.t =
...
+        (String.capitalize_ascii e1)
+        (pad (String.length e1) len1);
+      List.iter2
+        (fun e l -> Fmt.pf ppf "%s%u" (pad (int_size e) l) e)
+        entries lengths;
+      Fmt.pf ppf "@.")
+    entries;
+  Fmt.pf ppf "@."
+
+let pp_heading ppf txt =
```
aryx: We probably want to move those helper functions to a separate file. maybe src/osemgrep/reporting/Fmt_helpers.ml

aryx: With also a Fmt_helpers.mli with some basic documentation about each helper function (ex of output for example).

aryx (approved):
Looks good.
Weird that we already have different numbers between osemgrep and pysemgrep.

aryx:
a) yes, feel free to define some helpers functions to parse and categorize the metadata field of a rule
b) probably yes; just check what pysemgrep does and do the same. It might not fit well with how Logs is working though.
c) we can skip this part for now, as it's not part of the scope doc to handle SCA rules
d) not important for now, can keep as a TODO in the code. I think pro rules have a special field (interfile: true) in the metadata, and special origin.

hannesm:
> Looks good. Weird that we already have different numbers between osemgrep and pysemgrep.

Indeed, also pysemgrep uses "159" as number of files (for e.g. `<multilang>`), while osemgrep uses 160.

Also, the number of rules in pysemgrep from the table do not sum up to 429.

aryx:
Good. We're reaching the point where we find bugs in pysemgrep and osemgrep is correct.

aryx (approved):

hannesm:
Usage of stdout and stderr:

With --json, pysemgrep only outputs the json on stdout (which is nice for redirecting output and using shell pipes).

Osemgrep: the Logs library by default uses stdout for `Logs.app` (i.e. "application level") and stderr for all other log levls.

I recommend the following: we use the normal "Format.printf" (or Printf.printf) for application output (i.e. json -- already done in Common.pr), and Logs only for Scan Status / Scan Summary / ... -- which we configure to print on stderr as well.

hannesm:
> configure Logs to print app level on stderr as well.

done in [f4cbdf0](https://github.com/returntocorp/semgrep/pull/7445/commits/f4cbdf06c97fa0540e50cc99f4c2f2ef341ec139)

src/osemgrep/cli_scan/Core_runner.ml
```diff
@@ -181,16 +181,93 @@ let runner_config_of_conf (conf : conf) : Runner_config.t =
...
+    ppf
+    (List.combine
+       (Common.map
+          (fun (job : Runner_config.lang_job) ->
+            Xlang.to_string job.Runner_config.lang)
+          lang_jobs)
+       (Common.map
+          (fun Runner_config.{ targets; rules; _ } ->
+            [ List.length rules; List.length targets ])
+          lang_jobs))
```
hannesm: this is nothing I'm fund of... and should write it differently

hannesm: done in 6f2af64c6

src/osemgrep/cli_scan/Core_runner.ml
hannesm: done in [8532ca4](https://github.com/returntocorp/semgrep/pull/7445/commits/8532ca47b40bcf67365ecfc2be078e8db31311d9)

aryx (approved):

aryx (approved):

PR #7460, opened 2023-04-05
fix: Avoid crashing when outputting a string containing [/#]

nmote:
The `rich` library uses in-band signaling, and crashes when it detects what it interprets as a closing tag without a matching open tag. If Semgrep tries to print a finding that contains such a snippet, it crashes.

We can avoid this by disabling markup in the `rich` library. We can still do formatting, we just need to use its out-of-band signaling options.

See https://github.com/Textualize/rich/issues/2636

Test plan:

From `cli/src/semgrep`, run `python console.py`. The little test there does not crash. Without the `markup` argument, the newly-added line does crash.

Also ran `semgrep scan --pro --config p/default .` from the Juice Shop repo, while logged in. Before, this crashed. Now, it does not. The output looks reasonable -- it doesn't appear that we were relying on the `rich` library's markup, or at least not in any major way with the `scan` command.

PR checklist:

### DIFF
--- a/changelog.d/rich-markup-crash.fixed
+++ b/changelog.d/rich-markup-crash.fixed
@@ -1,0 +1,1 @@
+Fix an issue that could lead to a crash when printing findings that contain snippets that look like markup to the Rich Python library
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -274,7 +274,7 @@ def ci(
     )
 
     console.print(Title("Scan Environment", order=2))
-    console.print(debugging_table)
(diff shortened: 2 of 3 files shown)


emjin (approved):
Thank you! Hopefully we can add a regression test later

underyx:
CleanShot 2023-04-13 at 16 47 50@2x

@nmote I see quite a few usages of `[bold]` markup. Surprised this didn't break.

nmote:
I may have only searched `cli/src/semgrep`. Maybe it didn't break because it uses a different console instance? Anyway, I'd rather risk outputting markup syntax as text than risk the crashes we had before.

PR #7463, opened 2023-04-05
tainting: Fix bug in field sensitivity

IagoAbal:
Closes PA-2486

We have to distinguish between "clean in the environment" versus "matches a sanitizer". If an l-value such as `x.a.b` is tainted, but that same occurrence of `x.a` matches a sanitizer, we will not taint `x.a.b` (the sanitizer "blocks" the taint). However, if `x.a` is just clean in the environment from a previous sanitization, then `x.a.b` must get tainted.

test plan:
make test # test added

PR checklist:


emjin (approved):
Makes sense. I was initially a bit confused, because I expected `Sanitized` to mean "permanently sanitized". The comments made it clear for me, though, and on second thought I can't think of why you would need that.

PR #7467, opened 2023-04-05
fix(matching): simplify cast matching and fix double matching bug

### LINKED ISSUE #7403: Duplicate search-mode matches when target has an explicit cast
**Describe the bug**
Matches for patterns containing casts produce duplicates, one including the cast and one without.

See https://github.com/returntocorp/semgrep/pull/7109

**To Reproduce**
https://semgrep.dev/s/0wlB

**Expected behavior**
This should return one finding, not two.

**Screenshots**
N/A

**Environment**
semgrep.dev

**Use case**
Exposed via internal rule-writing efforts

brandonspark:
## What:
This PR fixes a double-matching bug we had.

## Why:
We were getting a double match from matching a `Cast` to a `Cast`, because we would match once for the constituents and once for the whole expressions.

## How:
We apply the same logic as symbolic propagation, to only descend into the constituents if we are not already at the root.

## Test plan:
image

## Alternatives considered:

I tried applying @IagoAbal 's suggestion for simplifying the matching logic, by changing it to 
```
  | _, B.Cast (b1, bt, b2) ->
      (match a.G.e with
      | G.Cast (a1, at, a2) ->
          m_type_ a1 b1 >>= fun () ->
          m_tok at bt >>= fun () -> m_expr a2 b2
      | _ -> fail ())
      >||> if not is_root then m_expr a b2 else fail ()
```

However, on either side you place it of the `N _, _ when is_metavariable` case, this causes problems. If it's before, you get issues with `N <mvar>, Cast` case, because it's possible to not produce a match (when it's not at the root), even though you still want that match to occur.

Closes #7403 

PR checklist:

### DIFF
--- a/changelog.d/gh-7403.fixed
+++ b/changelog.d/gh-7403.fixed
@@ -1,1 +1,4 @@
-Matching: Fixed a bug where explicit
+Matching: Fixed a bug where explicit casts of expressions would produce two matches to
+other explicit casts.
+
+So for instance, a pattern `(int $X)` in Java would match twice to `(int) 5`.
--- a/changelog.d/gh-7403.fixed
+++ b/changelog.d/gh-7403.fixed
@@ -1,0 +1,1 @@
+Matching: Fixed a bug where explicit
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -725,7 +725,12 @@ and m_expr_root a b = m_expr ~is_root:true a b
 (* coupling: if you add special sgrep hooks here, you should probably
  * also add them in m_pattern
  *)
-and m_expr ?(is_root = false) a b =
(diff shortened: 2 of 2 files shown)


changelog.d/gh-7403.fixed
```diff
@@ -0,0 +1,4 @@
+Matching: Fixed a bug where explicit casts of expressions would produce two matches to
```
IagoAbal, line 1: I think it's simpler than this, any expression pattern `E` matches `E` as well as `(T) E`. And e.g. `x` matches `x` and `(int) x`.

IagoAbal (approved):
I need to convince myself better than we really need `~arguments_have_changed`, but the fix is good anyways.

However, please add a test case, it's easy to make one that works for semgrep-core, simply add a new line as here https://semgrep.dev/s/6YNq.

aryx (approved):

PR #7472, opened 2023-04-06
fix: Ignore tokens from YAML aliases when computing range

nmote:
This way, the location of the match does not include the original location of the aliased section. See the difference in output for the `anchor.yaml` test below.

#4517 excluded these tokens from the visitor that gets the tokens from an AST node, but not from the visitor that computes ranges. The semgrep core tests use the tokens from the pattern match instead of the range associated with the pattern match, which is why, according to the tests, that has worked. Unfortunately, I believe the range as reported to end users has been wrong this whole time.

I will follow up with a change that updates the test runner.

Test plan:

From `tests/patterns/yaml`:

`semgrep scan -l yaml -e "$(cat anchor.sgrep)" anchor.yaml`

Before: https://gist.github.com/nmote/e6008ecb38e20749a6163916542b0160
After: https://gist.github.com/nmote/fd3d2800484213d2a75fe818172ae227

PR checklist:

### DIFF
--- a/changelog.d/yaml-alias.fixed
+++ b/changelog.d/yaml-alias.fixed
@@ -1,0 +1,1 @@
+Fixed the range reported by findings for YAML files that include an anchor, so that the match does not include the original location of the snippet bound to the anchor.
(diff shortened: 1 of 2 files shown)


emjin (approved):
😬 Good catch, thank you!

PR #7473, opened 2023-04-06
Tests: mask taint traces in the benchmarks

emjin:
The taint traces can change pretty easily when we add intermediate variables, so let's mask them.

Also in this PR: add the options passed to semgrep in the benchmarks name so that they are reported when uploaded.

Test plan: see the outputted findings and check dashboard.semgrep.dev

PR checklist:


perf/run-benchmarks
```diff
@@ -480,10 +482,13 @@ def run_benchmarks(
                 continue
 
             std_findings = {}
+            options = [re.sub(r"\W+", "", option) for option in corpus.semgrep_options]
```
nmote: Should this be sorted to avoid perturbations due only to order?

nmote (approved):

PR #7477, opened 2023-04-06
feat(scala): parse rhs indented block expr

brandonspark:
## What:
This PR allows us to parse indented `blockExpr`s that occur as the RHS to def statements.

## Why:
Parse rate.

In reality, _any_ `blockExpr` anywhere could be indented, but this causes a bunch of weirdness with cases like
```
def foo() =
  3
```
Are we supposed to read that as `3`, or an indented block expression which contains `3`, in other words:
```
def foo() = {
  3
}
```

I don't think it's very clear. So to be safe, to not mess up a bunch of parse trees every time we have an indented expression, I'm only allowing it in this case where we have it on the RHS of a `def`, and we can see if more cases fail later.

## How:
I made a parameter, `is_block_expr`, which is set by parsing the RHS to a `def` and noticing that it is both indented, and not the start to an expression.

We want this to make the `expr` call immediately descend down the recursive call chain to the `simpleExpr` `blockExpr` case, but unfortunately since there's a bunch of intermediary calls, we can't do that. We'll solve it by keeping the `is_block_expr` optional param up until it reaches that brace, to kind of guide the call to the `blockExpr` case.

## Test plan:
`make test`

Parse rate goes from `0.9460688054925819` to `0.95648527803769`.

PR checklist:

### DIFF
--- a/languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
+++ b/languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
@@ -545,6 +545,26 @@ let closeIndentRegion in_ =
       skipToken in_
   | _ -> failwith "precondition failure"
 
+(* If this next token is indented, it's quite likely we're inside of an implicit
+   indented blockExpr, which can start off an `expr`.
+
+   For instance, something like
+   def foo() =
+     val x = 3
+
+   It's not guaranteed, however, such as if we saw something like
+
+   def foo() =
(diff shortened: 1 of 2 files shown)


mjambon (approved):

mjambon:
Great. I'll let you add a changelog entry if you think that's needed.

aryx (approved):

PR #7480, opened 2023-04-07
feat(scala): parse unparenthesized conditions in `if`, `while`

brandonspark:
## What:
This PR lets us parse conditions without parens in `if` and `while`, which is a thing in Scala 3.

## Why:
Parse rate.

## How:
Added two cases.

I tried this a bit ago and ran into an issue, because in Scala, you can have unlimited postfix identifiers after an expression.

So when we saw something like `<expr> then`, it was interpreted as a postfix call from the identifier `then` to `expr`. We can't change this by having `then` lexed as a keyword, because `then` is only a _soft_ keyword.

I just made it such that this postfix logic activates specifically only if it's not `then`. I anticipate this should be fairly
unlikely, anyways.

## Test plan:
`make test`
Parse rate goes to `0.9589261173709144` from `0.9570153111528416`

PR checklist:

### DIFF
--- a/languages/scala/ast/AST_scala.ml
+++ b/languages/scala/ast/AST_scala.ml
@@ -392,6 +392,7 @@ and modifier_kind =
   (* just for variables/fields/class params *)
   | Val (* immutable *)
   | Var (* mutable *)
+  | EnumClass (* for enum classes *)
 
 and annotation = tok (* @ *) * type_ (* usually just a TyName*) * arguments list
 and attribute = A of annotation | M of modifier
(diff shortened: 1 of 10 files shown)


languages/scala/ast/AST_scala.ml
```diff
@@ -392,6 +392,7 @@ and modifier_kind =
   (* just for variables/fields/class params *)
   | Val (* immutable *)
   | Var (* mutable *)
+  | EnumClass (* for enum classes *)
```
aryx, line 395: this should be not be part of this PR. It has nothing to do with the title

aryx (changes requested):

aryx (approved):

languages/scala/ast/AST_scala.ml
brandonspark: same as before

PR #7481, opened 2023-04-07
test: Update core test engine to use match range

nmote:
This is consistent with what is reported via JSON and therefore what the CLI and all end users see.

Previously, the test runner used the tokens associated with the match rather than the range associated with the match, which led to different results in two test cases. The first is the one addressed by #7472, and was due to an inconsistency between the visitor used for computing ranges and the visitor used for extracting tokens from AST. The second is the one caused by #7437 and updated here. That difference is caused by the fact that we now update `e_range` to include the range of the parentheses, even though the tokens associated with the parentheses do not appear in the AST.

The `cp_label.py` test was added in #3042. I had to do some shenanigans to make it continue to test what it was originally intended to test. This is explained inline.

Test plan:

Automated tests

PR checklist:

### DIFF
--- a/src/engine/Unit_engine.ml
+++ b/src/engine/Unit_engine.ml
@@ -347,15 +347,9 @@ let regression_tests_for_lang ~polyglot_pattern_path ~with_caching files lang =
                (fun () ->
                  let matches =
                    match_pattern ~lang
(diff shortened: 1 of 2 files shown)


emjin (approved):
Thanks!

PR #7489, opened 2023-04-10
feat(scala): `enum`s parsing

brandonspark:
## What:
This PR makes it so we can parse the Scala `enum` construct, from Scala 3.

## Why:
I noticed this was unparseable.

## How:
We inject it into much the same construct as Java and Swift, which is a class marked as an `EnumClass`, where the
enums themselves are individual `EnumEntryDef` definitions.

This is made tricky somewhat by the fact that Scala allows the mixing and matching of `enum` variants that are both
simple instantiated class instances, like:
```
enum Color(val rgb: Int):
  case Red   extends Color(0xFF0000)
  case Green extends Color(0x00FF00)
  case Blue  extends Color(0x0000FF)
```

but also algebraic datatypes, such as
```
enum Option[+T]:
  case Some(x: T)
  case None
```

Now, this would not be a problem, normally. Swift does this too. But Swift only permits it to be one of these at a time. Scala actually allows mixing and matching of either.

So while in the Swift case, I would determine which case we were in, and thus then either inject it into a `ClassDef`, or a `TypeDef`, we actually will have to be able to replicate behavior of both. This is tricky because the `EnumEntryDef` construct currently does not support parameters, which the arguments to the enum variants actually are. As such, we'd need to be able to do one of two things:
1) either embed `parameter` into `argument` somehow
2) allow `EnumEntryDef` to support parameters

I chose the first option, by abusing `OtherArg` to get it into the syntax tree at all. I think this won't end up really mattering, anyways, because we're unlikely to need this semantic information later on.

For the `extends`, I chose to ignore this in the translation, because I don't think we have the appropriate constructs yet. It's confusing, because you are not guaranteed to have to extend an instance of the same class. Technically this desugars to a kind of binding.

## Test plan:
Parse rate goes down slightly, from `0.9590214626573684` to `0.9580818781981301`. I tested it out and this seems to be due to the fact that making "enum" a token messes up some Scala 2 programs where this wasn't a keyword.

It seems that programs which use `enum` also heavily use other Scala 3 features I haven't gotten around to parsing yet, so we wouldn't expect to see a parsing rate increase here. More to come.

Closes PA-2691

PR checklist:

### DIFF
--- a/languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
+++ b/languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
@@ -262,30 +262,6 @@ let adjustSepRegions lastToken in_ =
     | LPAREN info, xs -> RPAREN info :: xs
     | LBRACKET info, xs -> RBRACKET info :: xs
     | LBRACE info, xs -> RBRACE info :: xs
-    | Kcase info, xs -> (
-        (* pad: the original code generate different tokens for
-         * 'case object' and 'case class' in postProcessToken, I guess
-         * to simplify code here. Instead I lookahead here and have
-         * a simpler postProcessToken.
-         *)
-
-        (* ugly: if 'case' is the first token of the file, it will actually
-         * still be in in_.rest because of the way mk_env currently works.
-         * so we double check this initial condition here with =*=
-         * (on purpose not =~=)
-         *)
-        let rest =
-          match in_.rest with
-          | x :: xs when x =*= lastToken -> xs
-          | xs -> xs
-        in
-        match in_next_token rest with
-        | Some (Kobject _ | Kclass _) -> xs
-        | Some x ->
-            if !debug_newline then
-              logger#info "found case for arrow, next = %s" (T.show x);
-            ARROW info :: xs
-        | None -> xs)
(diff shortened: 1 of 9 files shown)


languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
```diff
@@ -298,15 +274,18 @@ let adjustSepRegions lastToken in_ =
...
-    (* pad: not that an arrow can also be used outside of a case, so
-     * here we dont raise an error if we don't find a match.
-     *)
-    | (ARROW _ as x), y :: ys when x =~= y -> ys
     | _, xs -> xs
   in
   in_.sepRegions <- newRegions;
   ()
 
+let inSepRegion tok f in_ =
```
aryx, line 282: this code was in a different PR too ...

aryx (approved):

languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
brandonspark: same as before, i think it's because of the PRs being merged out of order (i was planning on merging them myself after approval)

PR #7492, opened 2023-04-11
osemgrep: pass "output" (Core_runner.result -> unit) to invoke_semgrep_core

hannesm:
Previously, the output (scan summary, skipped files) was before the actual findings. Now it is possible to output them at the end, as done in the python implementation.

Another option (to avoid passing a higher-order function) would be to expose a separate binding in Core_runner and call that from Scan_subcommand after the `output` was done. Any preferences?

PR checklist:

### DIFF
--- a/src/osemgrep/cli_scan/Core_runner.ml
+++ b/src/osemgrep/cli_scan/Core_runner.ml
@@ -402,7 +402,7 @@ let pp_skipped ppf
    Take in rules and targets and return object with findings.
 *)
 let invoke_semgrep_core ?(respect_git_ignore = true) ?(ignored_targets = [])
-    (conf : conf) (all_rules : Rule.t list)
+    (output : result -> unit) (conf : conf) (all_rules : Rule.t list)
     (rule_errors : Rule.invalid_rule_error list) (all_targets : Fpath.t list) :
     result =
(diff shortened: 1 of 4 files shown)


aryx (changes requested):
I don't undertstand the goal here. But I'd rather separate things more cleanly and have
invoke_semgrep_core just do the computation and return a complex result,
and then later in Scan_subcommand actually doing the output.

hannesm:
Thanks for your comment. The goal is as follows:

python semgrep outputs:
```
┌─────────────┐
│ Scan Status │
└─────────────┘
...

┌──────────────────┐
│ 48 Code Findings │
└──────────────────┘
...

┌──────────────┐
│ Scan Summary │
└──────────────┘
...
```

While before this PR, osemgrep output is:
```
┌─────────────┐
│ Scan Status │
└─────────────┘
...

┌──────────────┐
│ Scan Summary │
└──────────────┘
...

┌──────────────────┐
│ 48 Code Findings │
└──────────────────┘
...

```

So, this PR moves the "Scan Summary" behind the "Code Findings".

aryx:
In any case, I'd rather not pass the higher-order func in invoke_semgrep_core.

hannesm:
superseeded by #7499

PR #7493, opened 2023-04-11
feat(scala): `given` definitions

brandonspark:
## What:
This PR lets us parse Scala `given` definitions, as specified in the Scala 3 grammar.

## Why:
Parse rate.

## How:
Did the thing.
There's a few places where we have to do some obnoxious hard-coded prediction because it's not super clear what case we're in. In one place, we do it as the Scala 3 compiler does it, and in another I hand-rolled the logic based on what I thought we should do.

It's worth noting that `given` definitions are for "contextual parameters", or essentially fulfilling the role of ad-hoc type-based polymorphism (a la Haskell typeclasses, or Rust traits, or something), so it's not immediately pertinent to us. Since I didn't really have a good construct to put them in, I fully embedded them into `Other`. We can change this later if we feel like.

I also added `with` as a token which can start an indentation region, c.f. the Scala 3 compiler: https://github.com/lampepfl/dotty/blob/865aa639c98e0a8771366b3ebc9580cc8b61bfeb/compiler/src/dotty/tools/dotc/parsing/Tokens.scala#L276

## Test plan:
`make test`, and parse rate went from `0.9581009472554209` to `0.966201829502692`

Closes PA-2692

PR checklist:


aryx (changes requested):
please rebase, so we don't see the code related to enums.

stats/parsing-stats/lang/scala/projects.txt
```diff
@@ -8,7 +8,7 @@ https://github.com/akka/akka
 https://github.com/yahoo/CMAK
 https://github.com/ornicar/lila
 https://github.com/gitbucket/gitbucket
-https://github.com/twitter/finagle
+# https://github.com/twitter/finagle
```
aryx, line 11: would be good to put that back.
Otherwise the 1% improvement from 95.5 to 96.5 could just come from that?

aryx (approved):
great work

PR #7494, opened 2023-04-11
feat(scala): parse exports, `for`, and `opaque`

brandonspark:
## What:
This PR adds in `export` statements.

## Why:
Parse rate!

## How:
Exports are almost identical to imports. I just decided to inject them into an `OtherDirective`.

It didn't feel right to me to put exports under the `I` and `import` type, so I just decided to make a new variant.

## Test plan:
Parse rate actually goes _down_, from `0.966201829502692` to `0.963349705274718`.

This isn't because of the parser doing bad things, I observed this change was due to lexing the `export` keyword, which could mess with Scala 2 programs that use it as an identifier.

The parse rate didn't otherwise go up, I found, because it seems usage of `export` is highly correlated with other language features we cannot parse yet, like `opaque` types. That's up next.

Closes PA-2693

PR checklist:


changelog.d/pa-2694.added
```diff
@@ -0,0 +1 @@
+Scala: Can now parse top-level definitions (as added in Scala 3)
```
aryx, line 1: How many PRs this PR contains?
I still don't understand how all your PR contain stuff from other PRs. How do you use git to make your PR? Or it is because each PR depends on one another and you get confused in using gits with so many branches?

aryx (approved):

PR #7496, opened 2023-04-11
feat(scala): parse indented `for` exprs

brandonspark:
## What:
This PR makes it so we can parse `for` expressions whose enumerators are indented, as opposed to wrapped in braces/parens.

## Why:
Parse rate.

## How:
Cased on if it was in a braces or parens. If not, it's assumed to be indented, so we just call `enumerators` directly.

I deliberately did not call `inBracesOrIndented`, because entering the indentation region of the `for` will mess us up at the end, when determining when to end the `enumerators`. This is because we will see a `DEDENT`, but because of the way the code is structured, this will make it difficult to properly have the tokens synchronized properly when we break out of the enumerators sequence.

This is similar to what the Scala compiler does: https://github.com/lampepfl/dotty/blob/865aa639c98e0a8771366b3ebc9580cc8b61bfeb/compiler/src/dotty/tools/dotc/parsing/Parsers.scala#L2730

## Test plan:
`make test`, and parse rate goes from `0.9672796646272888` to `0.9707749361511622`

Closes PA-2695

PR checklist:

### DIFF
--- a/changelog.d/pa-2694.added
+++ b/changelog.d/pa-2694.added
@@ -1,0 +1,1 @@
+Scala: Can now parse top-level definitions (as added in Scala 3)
--- a/languages/scala/ast/AST_scala.ml
+++ b/languages/scala/ast/AST_scala.ml
@@ -387,6 +387,7 @@ and modifier_kind =
   | Override
   | Inline
   | Open
+  | Opaque
   (* pad: not in original spec *)
   | CaseClassOrObject
   (* less: rewrite as Packaging and object def like in original code? *)
--- a/languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
+++ b/languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
@@ -1687,7 +1687,7 @@ let pattern in_ : pattern = noSeq pattern in_
  *               | while `(` Expr `)` {nl} Expr
  *               | `while` Expr `do` Expr
  *               | do Expr [semi] while `(` Expr `)`
- *               | for (`(` Enumerators `)` | `{` Enumerators `}`) {nl} [yield] Expr
+ *               | for (`(` Enumerators `)` | `{` Enumerators `}` | Enumerators) {nl} [yield] Expr
  *               | throw Expr
  *               | return [Expr]
  *               | [SimpleExpr `.`] Id `=` Expr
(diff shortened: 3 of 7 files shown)


languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
```diff
@@ -2378,8 +2378,28 @@ and parseFor in_ : stmt =
...
+           up when determining when to end, because if we see the future DEDENT and break,
+           then we will still be on the NEWLINE.
+
+           It's easiest to just not emit the DEDENT tokens at all, which is what happens
+           if we do not enter the indentation region.
+
+           See
+           https://github.com/lampepfl/dotty/blob/865aa639c98e0a8771366b3ebc9580cc8b61bfeb/compiler/src/dotty/tools/dotc/parsing/Parsers.scala#L2730
+        *)
+        fb (PI.unsafe_fake_info "") (enumerators in_)
```
IagoAbal, line 2402: Can we not use unsafe fake tokens here? The `unsafe_` is meant to flag that it shouldn't be abused. In this case we have some token around readily available to create a "safe" fake token.

IagoAbal (approved):
Looks good to me, but I'm not familiar with this parser.

changelog.d/pa-2694.added
```diff
@@ -0,0 +1 @@
+Scala: Can now parse top-level definitions (as added in Scala 3)
```
aryx, line 1: this should not be part of this PR

languages/scala/ast/AST_scala.ml
```diff
@@ -387,6 +387,7 @@ and modifier_kind =
   | Override
   | Inline
   | Open
+  | Opaque
```
aryx, line 390: This should not be in this PR.

aryx (changes requested):

aryx (approved):

changelog.d/pa-2694.added
brandonspark: i suspect it was not, but the merge base probably changed due to all the stacked PRs

languages/scala/ast/AST_scala.ml
brandonspark: same as above

PR #7506, opened 2023-04-12
fix(SSC): Include line numbers in match for findings in lockfiles

mmcqd:
Previously lockfile-only findings would always set the location of their finding to `(0,0,0),(0,0,0)`, but our current parsers can identify line numbers of findings for every language except Rust. So we should include this data. 

This also solves the problem of some lockfile-only supply chains findings appearing to be incorrectly deduplicated. This was because we were printing both of them at the position `(0,0,0),(0,0,0)` of the file.

Since we don't track column numbers in lockfiles, I just set the location as spanning from the start of the relevant line to the start of the next line. Technically this is wrong, and the end position should be the final column of the actually relevant line. This causes the next line to get printed unhighlighted, but I don't think this is a big deal.

Here's what this looks like in practice:

# Before
Screen Shot 2023-04-12 at 11 33 55 AM

# After
Screen Shot 2023-04-12 at 11 32 58 AM

PR checklist:


mmcqd:
hm wait, this test is not doing the right thing

cli/src/semgrep/dependency_aware_rule.py
```diff
@@ -98,8 +98,16 @@ def generate_unreachable_sca_findings(
                         rule_id=core.RuleId(rule.id),
                         location=core.Location(
                             path=str(lockfile_path),
-                            start=core.Position(0, 0, 0),
-                            end=core.Position(0, 0, 0),
+                            start=core.Position(found_dep.line_number or 0, 0, 0),
```
mmcqd, line 101: this is the only bit that actually needs review

adamberman (approved):

PR #7507, opened 2023-04-12
php: metavariable-pattern: Only add <?php when needed

IagoAbal:
PR #5501 adds `<?php` to the content to be parsed as PHP without checking if the `<?php` opening is already present. The PHP parser does not like that.

Closes PA-2696
Follows #5501

test plan:
make test # test added

PR checklist:

### DIFF
--- a/src/engine/Metavariable_pattern.ml
+++ b/src/engine/Metavariable_pattern.ml
@@ -30,7 +30,9 @@ let logger = Logging.get_logger [ __MODULE__ ]
 
 let adjust_content_for_language (xlang : Xlang.t) (content : string) : string =
   match xlang with
-  | Xlang.L (Lang.Php, _) -> "<?php " ^ content
+  | Xlang.L (Lang.Php, _)
+    when not (content =~ {|[ \t\n]*<\?\(php\|=\)?[ \t\n]+|}) ->
+      "<?php " ^ content
   | __else__ -> content
 
 (* This function adds mvars to a range, but only the mvars which are not already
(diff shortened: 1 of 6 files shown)


src/engine/Metavariable_pattern.ml
```diff
@@ -30,7 +30,9 @@ let logger = Logging.get_logger [ __MODULE__ ]
 
 let adjust_content_for_language (xlang : Xlang.t) (content : string) : string =
   match xlang with
-  | Xlang.L (Lang.Php, _) -> "<?php " ^ content
+  | Xlang.L (Lang.Php, _)
+    when not (content =~ {|[ \t\n]*<\?\(php\|=\)?[ \t\n]+|}) ->
+      "<?php " ^ content
```
mjambon, line 39: It's better than before but shouldn't this be built into the parser? Doesn't the pattern parser do something like this already?

I'm also wondering how come a closing `?>` isn't required. I guess it's valid PHP but it would be great to document it.

How about moving the logic to where PHP parsing happens. Something like this maybe:
```ocaml
let adjust_content_for_language (xlang : Xlang.t) (content : string) : string =
  match xlang with
  | Xlang.L (Lang.Php, prog_str) -> Lib_parsing_php.wrap_extracted_code prog_str
  | ...
```

mjambon:
You forgot to add a changelog entry.

mjambon (approved):

src/engine/Metavariable_pattern.ml
IagoAbal: Good points, ~applied!~ added some comments. Unfortunately `Lib_parsing_xyz` modules aren't available in the engine package and I prefer not to touch dune files for this.

IagoAbal:
Merged despite one failed test because:
1. Tests passed before and I just added a comment.
2. The test seems to fail because it cannot `git clone` this repo https://github.com/dropbox/groupy.
3. I also cannot clone the repo locally.
4. The repo page says: "This repository is currently being migrated. It's locked while the migration is in progress."

PR #7535, opened 2023-04-14
cli: Sort output to so that it's easier to compare between runs

IagoAbal:
test plan:
Run `semgrep -c p/default-v2 --verbose` twice on the same repo and diff the output

test plan:
make test # tests added

PR checklist:

### DIFF
--- a/changelog.d/pa-2700.changed
+++ b/changelog.d/pa-2700.changed
@@ -1,0 +1,2 @@
+The different lists of skipped files output by Semgrep when given --verbose will
+now be sorted, to make it easier to `diff` the outputs of two runs.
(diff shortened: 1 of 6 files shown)


nmote (approved):

aryx:
I'm surprised this does not require to regenerate some snapshots.

IagoAbal:
> I'm surprised this does not require to regenerate some snapshots.

I guess it's because all our snapshots store the JSON output and this only affects the text output.

Now that you mention it, I realize it's actually bad that this is not reflected in any test, it may be broken in OSemgrep.

IagoAbal:
Added tests!

PR #7539, opened 2023-04-14
convert to typescript + dont use global state for libyaml module

tpetr:
1. Convert engine package to typescript, which makes it infinitely easier to import into other projects (companion PR for languages coming in the near future)
2. Tweak how the engine pulls in the libyaml WASM module so that we don't have to rely on global state


js/engine/src/index.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  return {
+    lookupLang,
+    addParser: (parser: Parser) => {
+      parser.setMountPoints(getMountPoints()); // inherit engine's mount points
+      languages.set(parser.getLang(), parser);
+    },
+    hasParser: (lang: Lang) => {
+      return languages.has(lang);
+    },
```
zyannes: super-nit: Can make this a bit simpler with 
```suggestion
    hasParser: (lang: Lang) => languages.has(lang),
```

zyannes (approved):

PR #7563, opened 2023-04-18
Wire up js langs

tpetr:
This PR is quite large, but should be digestible if you break it down by the first 3 commits:

1. Some final massaging of the ocaml <> JS bridge + implement a couple languages
2. Implement the rest of the languages
3. Update the workflow to build & test the JS engine + all languages


aryx:
CI seems blocked ...

tpetr:
Yep, investigating

aryx:
I foresee we will never get a full list of green checks anymore

tpetr:
Seemed to be a transient issue with cache restore. If we see this happen often we can disable testing the js langs outside of the nightly run and releases

aryx:
I want to release 1.19.0 today before merging this

.github/workflows/build-test-javascript.yaml
```diff
@@ -19,42 +18,79 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: true
+      - name: Set up tree-sitter
+        run: |
+          (cd libs/ocaml-tree-sitter-core && ./configure && ./scripts/install-tree-sitter-lib)
+      - name: Cache git checkout
```
aryx, line 24: Do we need this? Does it improve significantly the build time?

aryx (approved):

.github/workflows/build-test-javascript.yaml
tpetr: 40+ seconds to do a full clone vs. 4s to restore repository from cache

PR #7564, opened 2023-04-18
fix(taint): update keys with `Taints.map`

brandonspark:
## What:
This PR fixes a bug which made `Taints.map` cause taints to become conflated with each other, and erase each other's existence even though they shouldn't.

## Why:
This was causing cases to fail, because taints would become deleted when they shouldn't be. See https://semgrep.dev/s/Onon

## How:
This was happening because the internal representation of `Taint_set` is as a `Taint_map`. Since we're using the regular `map` function on maps (which doesn't touch the keys), this would make keys associate to different values, but then cause clashes between different values which originated from the same key.

I fixed it by simply making the `map` function also work on the keys. Because the type of keys was different than the type of maps, I had to change the key type for `Taint_map` as well, because otherwise I'd be shoehorned into a function which worked only on `orig` or `taint`, when I want the freedom to choose both.

## Test plan:
`make test`

Closes PA-2682

PR checklist:

### DIFF
--- a/src/tainting/Taint.ml
+++ b/src/tainting/Taint.ml
@@ -210,10 +210,10 @@ let pick_taint taint1 taint2 =
 
 module Taint_set = struct
   module Taint_map = Map.Make (struct
-    type t = taint
+    type t = orig
 
     let compare k1 k2 =
-      match (k1.orig, k2.orig) with
+      match (k1, k2) with
       | Arg _, Src _ -> -1
       | Src _, Arg _ -> 1
       | Arg a1, Arg a2 -> Stdlib.compare a1 a2
@@ -255,7 +255,7 @@ module Taint_set = struct
      * coupling: If this changes, make sure to update docs for the `Taint.signature` type.
      *)
     set
-    |> Taint_map.update taint (function
+    |> Taint_map.update taint.orig (function
          | None -> Some taint
          | Some taint' -> Some (pick_taint taint taint'))
 
@@ -274,12 +274,19 @@ module Taint_set = struct
      want to map the keys too.
      Unfortunately, the keys and values are different types, so it's not as
      straightforward.
-     Fortunately, we can exploit a property of
+     Fortunately, we can exploit a property of the map, which is that the
+     `orig` of the domain and codomain should be the same. So it should be fine
+     to simply map the codomain taint, and then take its `orig` as the key.
   *)
   let map f set =
     let bindings = Taint_map.bindings set in
     bindings
-    |> Common.map (fun (t1, t2) -> (f t1, f t2))
+    (* Here, we assume the invariant that the orig must be
+       the same in the domain and codomain.
+    *)
+    |> Common.map (fun (_, t2) ->
+           let new_taint = f t2 in
+           (new_taint.orig, new_taint))
     |> List.to_seq |> Taint_map.of_seq
 
   let iter f set = Taint_map.iter (fun _k -> f) set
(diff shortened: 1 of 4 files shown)


src/tainting/Taint.ml
```diff
@@ -210,10 +210,10 @@ let pick_taint taint1 taint2 =
 
 module Taint_set = struct
   module Taint_map = Map.Make (struct
-    type t = orig
+    type t = taint
```
IagoAbal: A map `orig -> taint` kind of made sense, it mapped an `orig` to a "trace" (a list of tainted tokens). Having `taint -> taint` where `f(x) = x` I don't think makes a lot of sense.

src/tainting/Taint.ml
```diff
@@ -255,7 +255,7 @@ module Taint_set = struct
      * coupling: If this changes, make sure to update docs for the `Taint.signature` type.
      *)
     set
-    |> Taint_map.update taint.orig (function
+    |> Taint_map.update taint (function
```
IagoAbal: And I don't think this is correct any more? This is not guaranteed to maintain `f(x) = x`.

IagoAbal: It works since the tests pass, but "breaks" what one would guess it's an invariant of the data structure.

src/tainting/Taint.ml
```diff
@@ -268,7 +268,20 @@ module Taint_set = struct
...
 
   let singleton taint = add taint empty
-  let map f set = Taint_map.map f set
+
+  (* Because `Taint_set` is internally represented with a map, we cannot just
+     map the codomain taint, using the internal provided `map` function. We
+     want to map the keys too.
+     Unfortunately, the keys and values are different types, so it's not as
+     straightforward.
+     Fortunately, we can exploit a property of
```
IagoAbal: This comment seems outdated and incomplete.

src/tainting/Taint.ml
brandonspark: I suppose I'm still not clear on the original design. Is it not the case that `orig -> taint` maps an `orig` to not just a trace, but an `orig` itself? Why not just have a set of `taint` itself?

brandonspark: Or, why not have an `orig -> trace` map? The `f(x) = x` nature of the map is built-in to the current functionality of the set, I think, and it's not something I'm trying to add, explicitly.

IagoAbal: The idea is that the `orig` is the "unique" thing we are tracking, the source of taint. But `taint` contains more than the `orig`, contains the "trace" aka `tainted_tokens`. And we want to map each `orig` to the "best" (whatever we decide that means) `tainted_tokens`. But yeah, it could have been `orig -> tainted_tokens`, I imagine I just picked `orig -> taint` for convenience.

IagoAbal: Anyways, my main problem with the PR is the comment below. https://github.com/returntocorp/semgrep/pull/7564#discussion_r1170526232. The bug was caused by `map` not respecting the invariant that the `orig` key must be the same as the `orig` of the value. And this non-written invariant seems to be broken again just in another place.

IagoAbal (approved):
Provided that you fix these issues and the tests pass, feel free to merge.

brandonspark:
@IagoAbal i couldn't do quite what we discussed, because actually it's not the `tainted_tokens` that matters in the codomain, it's the `call_trace`. regardless, the fact that we have this invariant that the `orig` should be the same in the domain and codomain makes me think that the map can be simply done by just mapping the codomain `taint`. let me know what you think.

PR #7574, opened 2023-04-19
refactor: rename ast_generic/Tok.ml to ast_generic/Tok_.ml

aryx:
This will allow to define a propose lib_parsing/Tok.ml
extracted from Parse_info.ml

test plan:
make core-test

PR checklist:


nmote:
> This will allow to define a propose lib_parsing/Tok.ml

I'm not very happy about the idea of `Tok` and `Tok_` both floating around. I think that if they are two different things, the names should briefly describe the differences. If they are basically the same thing, we should figure out how to unify them. I already get confused about the differences between a lot of these different data structures and the different aliases for them.

IagoAbal:
These refactorings need to be applied to Semgrep Pro too. Important to take into consideration if you plan to   do more refactoring after this PR. Otherwise when you want to work with Pro you suddenly have to deal with lots of small changes you are unfamiliar with.

aryx:
Ultimately I want to get rid of Tok_.ml

Yes I'll do the PR for semgrep pro too. I'm OnCall this week and so will do the release.

PR #7581, opened 2023-04-19
refactor(taint): allow multiple sink traces

brandonspark:
## What:
This PR refactors our internal taint representation so that a taint finding can have multiple traces, when it comes to sinks.

## Why:
Before, we made a simplifying assumption that when discharging a function taint signature, we could produce a bunch of small taint "findings", one per taint going into the function. This was fine when interprocedural taint was not using taint labels, but it doesn't suffice for a world that has them. This is because a taint finding may occur due to multiple input taints.
 
However, this comes with its own difficulties. Our type representation for `taints_to_sinks` assumed that there was only a single taint going to the sink, meaning a single trace for the sink. This is not true, however, as the input taints to a function call might go to different arguments, resulting in many different call traces, all for different taints.

## How:
I altered some type representations so that we could have multiple traces for multiple taints going through a function call to a given sink.

This was done by, instead of storing just a `taint` for every single taint going into the sink, I paired each `taint` with a `unit call_trace`.

This was using a `unit call_trace` because I didn't want the `taint` to be at the very end of the `call_trace`. If it was, I would have to do a bunch of work to traverse it to figure out if we had an `Arg` or `Src` variant, which would be gross, and also take linear time. This information was necessary because in `check_function_signature` we need to dispatch on the variant in order to determine if we need to substitute it with taints going in to the function. The actual information at the end of the call trace is irrelevant, we just need the trace.

## Test plan:
See https://github.com/returntocorp/semgrep-proprietary/pull/672

## Future work:
This is only step 1 of interprocedural taint labels. This deals with the problem of instantiating multiple taints within the body of a function that is called, but doesn't deal with more nuanced conditional behavior for taints within the flow of the function, such as `requires` attached to sources, or labeled propagators. More work as detailed in my scope doc at https://www.notion.so/r2cdev/Interprocedural-Taint-Labels-138ca7ca4d21463a8f481ccc9995903b?pvs=4 is needed to accomplish that.

Closes PA-2708

PR checklist:


changelog.d/pa-2708.added
```diff
@@ -0,0 +1 @@
+Taint: Added support for simple cases of interprocedural taint labels
```
IagoAbal: Maybe add `Pro: ...` so that it's clear this is a Pro feature.

src/core/Pattern_match.ml
```diff
@@ -54,10 +54,9 @@ type taint_call_trace =
     }
 [@@deriving show, eq]
 
-type taint_trace = {
-  sources : (taint_call_trace * tainted_tokens) list;
-  sink : taint_call_trace;
-}
+type taint_trace =
+  (* source *)
+  (taint_call_trace * tainted_tokens * (* sink *) taint_call_trace) list
```
IagoAbal: Could we define a new record type, `taint_trace_item` or something?

IagoAbal: Why do we now have multiple sinks for a `taint_trace` ?

brandonspark: Yes, we could.

brandonspark: These aren't truly multiple sinks, they are multiple paths to the _same_ sink. This is because the taint going to a single sink might take multiple routes to get there, for instance if multiple different arguments are tainted.

IagoAbal: Thanks, got it! So we have the trace-from-source-to-current function, then the internal-trace-within-the-function (the tainted tokens), then the trace from-current-function-to-sink. Maybe this is good to explain in a comment somewhere.

IagoAbal: Or having a record, using some long descriptive names for the fields.

src/tainting/Taint.mli
```diff
@@ -36,7 +36,7 @@ type orig =
 type taint = { orig : orig; tokens : tainted_tokens } [@@deriving show]
 
 type taints_to_sink = {
-  taints_with_precondition : taint list * AST_generic.expr;
+  taints_with_precondition : (taint * unit call_trace) list * AST_generic.expr;
```
IagoAbal: So this `unit call_trace` is the trace from-current-function-to-sink. Needs a comment / record type / or new type to make it more clear.

src/tainting/Taint.mli
```diff
@@ -22,7 +22,7 @@ type source = {
 }
 [@@deriving show]
 
-type sink = Rule.taint_sink call_trace [@@deriving show]
+type sink = Pattern_match.t * Rule.taint_sink [@@deriving show]
```
IagoAbal: Better record type IMO.

src/tainting/Dataflow_tainting.ml
```diff
@@ -1205,34 +1222,32 @@ let check_function_signature env fun_exp args args_taints =
...
+              taints
+              |> List.concat_map (fun (t, ct) ->
+                     match t.T.orig with
+                     | T.Src _ ->
+                         (* Here, we do not modify the call trace or the taint.
+                            This is because this means that, without our intervention, a
+                            source of taint reaches the sink upon invocation of this function.
+                            As such, we don't need to touch its call trace.
+                         *)
+                         [ (t, ct) ]
```
IagoAbal: Do we need to do anything with this? Previously we did nothing. The idea is that that finding will be reported when that function is analyzed.

IagoAbal:
> ... such as requires attached to sources, or labeled propagators. More work as detailed in my scope doc at ...

Is the precondition attached to sources ignored?

src/tainting/Dataflow_tainting.ml
brandonspark: Previously, we did nothing because we had an assumption that only one taint was necessary in order to produce a taint finding. This worked fine in the case where we didn't have interprocedural taint labels, but as we know, that is no longer the case.

To get it to work interprocedurally with taint labels, we will actually have to keep around these source taints. This is because, whereas previously we could assume that any `Src` taint we saw was "done", in the sense that we had already reported everything we needed from it (finding or not), it's no longer necessarily the case that it will be irrelevant to a future taint finding. Consider the following code:
```
def bar(y):
  foo(y)
  
def foo(x):
  a = source_a
  sink_of_a_and_b(a, x)
```
In this case, `foo` will get a `ToSink` with the `Arg` taint from `x`, and the `Src` taint from `source_a`. This is fine, but when we use the taint signature in the body of `bar`, we have to decide what taint signature to produce for `bar`.

In this case, it's not true that we can safely discard the `Src` taint from `source_a`. If that was true, we would never be able to produce a taint finding from calling `bar`, even though there should be one if `bar` were called with a `B` taint. This is because the `Src` taint is still flowing in to the sink of `sink_of_a_and_b`, albeit gated behind the call to `foo`. So we should keep it around.

brandonspark:
> Is the precondition attached to sources ignored?

In the interprocedural case, yes. `Arg` taint doesn't know how to interface with source `requires`, yet.

IagoAbal (approved):

IagoAbal:
> > Is the precondition attached to sources ignored?
> 
> In the interprocedural case, yes. `Arg` taint doesn't know how to interface with source `requires`, yet.

Wait when we report a finding it is being check that the taints satisfy the requires no?

src/tainting/Dataflow_tainting.ml
IagoAbal: I think you are right, can you write a comment using that same or similar example attached to the code? Your comment here makes it a lot more clear, perhaps just shorten it a bit.

brandonspark:
> Wait when we report a finding it is being check that the taints satisfy the requires no?
Yes, for sinks this is fine. But at the time that we generate the taint signature for a function which contains a source with a `requires` (which, by the way, I think we really need a term for), the taint engine does not know how to deal with a "hypothetical" taint. This means that it might not assume that a taint exists, because it does not know about the `Arg` taint being instantiated in a way to satisfy the source's `requires`. Example:
```
def foo(x):
  y = source_that_requires_a_and_produces_b(x)
  sink_that_requires_b(y)
```
In this example, we will infer that the `x` taint reaches `y`, but if we call it with `x : A`, we will not get any finding. This is because when the `Arg` taint of `x` reaches `y`, the `Arg` taint does not satisfy the `requires: A` and thus we produce no `B` taint.

src/tainting/Dataflow_tainting.ml
brandonspark: yep, good idea. done!

PR #7598, opened 2023-04-20
feat(lsp-ocaml): Port LS to OCaml

ajbt200128:
Ported the LS to OCaml!
Corresponding VSCode extension branch [here](https://github.com/returntocorp/semgrep-vscode/tree/austin/lsp-ocaml)

This should be backwards compatible with the current VSCode extension, but some of its settings are no longer relevant and don't do anything. If you're reviewing/testing this, I'd recommend checking out the branch above, and using that.

This is my first real experience with OCaml so feel free to leave any nitpicking/style guide comments!

PR checklist:


aryx:
+1,861 −1,848
pretty stable change in LOC :) I guess OCaml/Python have similar expressivity

aryx:
I think ls is not a super well know abbreviation. LSP is, so maybe rename to lsp/ or
the more verbose language_server/.
same for the package name, so we would use semgrep.language_server

Note: the LSP_experiment.ml I wrote a long time ago was to leverage type information in the OCaml language server for better typed metavariable in Semgrep for OCaml.

ajbt200128:
> I think ls is not a super well know abbreviation. LSP is, so maybe rename to lsp/ or the more verbose language_server/. same for the package name, so we would use semgrep.language_server

Yea fair, I tried looking around at one point and some people say "LSP implementation" "LSP" or just "Language Server". The idea of calling something the "Language Server Protocol Server" makes me sad, but I've accepted it. I'll probably go for "language_server"

aryx:
you did a mistake with git, because now we have all the changes of Brandon and Tom in your PR.
Can you merge develop correctly so that we just see your changes?

changelog.d/pa-lsp.added
```diff
@@ -0,0 +1 @@
+Language Server has been moved to OCaml core, with major speed improvements
```
aryx, line 1: ... has been reimplemented in OCaml ...

cli/src/semgrep/lsp/schema.json
```diff
@@ -63,40 +43,10 @@
...
-      "type": "boolean",
-      "default": true,
-      "description": "Skip files ignored by git"
-    },
-    "semgrep.lsp.watchOpenFiles": {
-      "type": "boolean",
-      "default": true,
-      "description": "Scan all opened files automatically on open and save."
-    },
-    "semgrep.lsp.watchWorkspace": {
```
aryx, line 81: what happened to all those schema elements? We don't support those features anymore?

cli/src/semgrep/metrics.py
```diff
@@ -191,7 +185,6 @@ class Metrics:
             errors=ErrorsSchema(),
             performance=PerformanceSchema(),
             value=ValueSchema(features=set()),
-            fix_rate=FixRateSchema(),
```
aryx, line 194: Why we don't care about fix_rate anymore?

src/ls/CodeActions.mli
```diff
@@ -0,0 +1,4 @@
+val code_actions_of_results :
```
aryx, line 1: In general we try very hard to document things in the .mli. In fact, most of the time for PRs I just look at the .mli.
So here it would be probably better to rename the function to code_actions_of_core_matches
and explain what is the string list second argument (since it's not super clear from the type what it is).

src/ls/CodeActions.ml
```diff
@@ -0,0 +1,40 @@
...
+  let matches =
+    Common2.filter
+      (fun (m : Semgrep_output_v1_t.core_match) ->
+        m.location.path = file && m.extra.rendered_fix <> None)
+      matches
+  in
+  Common.map code_action_of_match matches
+
+let code_actions_of_results matches files =
+  Common2.map_flatten (code_actions_of_file matches) files
```
aryx, line 40: Use of Common2 is deprecated. Here you can probably use List.concat_map
We could add a semgrep rule for it  ...

src/ls/CodeActions.mli
aryx: I guess the string list is a list of filenames. In theory you could use Common.filename, which is an alias to a string,
but recently we started to switch to more strongly typed file names with Fpath.t.
That requires to use a few extra functions. See also libs/commons/File.mli

src/ls/CodeActions.ml
```diff
@@ -0,0 +1,40 @@
+open Lsp
+open Types
+open Lsp_util
+
```
aryx, line 4: A Toplevel comment in the file rougly explaining the goal of the module would be helpful.
See other files in semgrep as example (e.g., AST_generic.ml, but does not have have to be that long :) )

src/ls/Diagnostics.mli
```diff
@@ -0,0 +1,2 @@
+val diagnostics_of_results :
```
aryx, line 1: Again, a small comment in the .mli will help.
The .mli are the most important files.
That's usually what we read when we want to use other people's code.

src/ls/Git_helper.ml
```diff
@@ -0,0 +1,81 @@
+(*diff unified format regex https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html#Detailed-Unified
```
aryx, line 1: This is mostly independent of the language server, so it should be moved out, e.g., in libs/
and actually there are a few places where we do git related stuff that we should merge together.
maybe we can start a libs/lib_git/ to gather such code.

src/ls/Git_helper.ml
```diff
@@ -0,0 +1,81 @@
...
+ * @@ -10,3 +20,3 @@
+ * where the 3 is the number of lines that were changed
+ * We use a named capture group for the lines, and then split on the comma if it's a multiline diff *)
+let _git_diff_lines_re = {|@@ -\d*,?\d* \+(?P<lines>\d*,?\d*) @@|}
+let git_diff_lines_re = SPcre.regexp _git_diff_lines_re
+
+let is_git_repo () =
+  let cmd = ("git", [| "git"; "rev-parse"; "--is-inside-work-tree" |]) in
+  (* We use Lwt_process.exec here instead of the standard library because we can do timeouts *)
+  let%lwt status =
```
aryx, line 19: @dinosaure started to use the Bos library for such things. We should probably unify things.

src/ls/LS.mli
```diff
@@ -0,0 +1 @@
+val start : Runner_config.t -> unit
```
aryx, line 1: comment

aryx: Can be "Entry point of the language server. This will start the server following the LSP protocol and listen on port XXX for requests from VScode or other LSP clients.

src/ls/Lsp_util.mli
```diff
@@ -0,0 +1,2 @@
+val range_of_location : Semgrep_output_v1_t.location -> Lsp.Types.Range.t
```
aryx, line 1: Those are actually not utilities related to LSP but utilities related to Semgrep.
Maybe Convert_utils.mli

src/ls/Lsp_util.mli
```diff
@@ -0,0 +1,2 @@
...
+val string_of_severity : Rule.severity -> Lsp.Types.DiagnosticSeverity.t
```
aryx, line 2: This does not return a strings maybe lsp_severity_of_rule_severity
(the final type may be a string, but this is an implementation detail we should hide. You use DiagnosticSeverity.Error
in the code, so this looks like a real type)

aryx (approved):

src/ls/Processed_run.ml
```diff
@@ -0,0 +1,146 @@
+open Git_helper
+
+type t = Semgrep_output_v1_t.core_match * Rule.rule
+
+(* This is copied from osemgrep/Nosem.ml, when porting is done we should use
+   stuff from there *)
+(* Trying to include that part of osemgrep causes a dependency cycle :( *)
```
aryx, line 7: Ok, I'll try to fix the dependency cycle.

src/ls/Processed_run.mli
```diff
@@ -0,0 +1,8 @@
+type t = Semgrep_output_v1_t.core_match * Rule.rule
+
+val of_matches :
```
aryx, line 3: small comment, again. Anything that can help the consumer of this module to understand how to use it.

src/ls/Unit_LS.ml
```diff
@@ -0,0 +1,250 @@
+open Lsp
+open Types
+open Testutil
+module Out = Output_from_core_t
+
+(** Try to test all of the more complex parts of the LS, but save the e2e stuff
```
aryx, line 6: great comment

src/ls/Unit_LS.ml
```diff
@@ -0,0 +1,250 @@
...
+  let test_git () =
+    let workspace = mock_workspace ~git:true () in
+    let file1 = add_file ~git:true workspace () in
+    let file2 = add_file ~git:false workspace () in
+    let files = [ file1; file2 ] in
+    let expected = files in
+    test_session files expected workspace false
+  in
+  let test_git_dirty () =
+    let workspace = mock_workspace ~git:true () in
```
aryx, line 143: @mjambon started to write some test utilities in help creating test files and dirs.
See libs/Testutil_files.mli

aryx: Maybe we need to extend it to support those ~git:xxx flags.

src/ls/dune
```diff
@@ -0,0 +1,23 @@
+(library
+ (public_name semgrep.ls)
```
aryx, line 2: I think ls -> language_server everywhere will be better.

aryx (approved):
Great work!

aryx:
Usually when we approve but don't merge the PR that means the PR is mostly good, but we let the author
address the comments and then merge it. There is no need for a second review, just try your best to address the comment
(if you disagree then don't change the code).

aryx:
The CI failure seems unrelated. I'll try to fix it. All the PRs on develop are currently broken because of this dev-docker CI step failing.

src/ls/Git_helper.ml
dinosaure: I think it's indeed the right choice & allow us to be outside the `lwt` monad.

PR #7465, opened 2023-04-05
feat(kotlin): update to latest tree-sitter

brandonspark:
As title.

## Test plan:
```
lang/kotlin/stats.json:3
   "global": {
     "name": "*",
     "parsing_rate": 0.9867509086298707,
     "line_count": 662536,
     "error_line_count": 8778,
     "file_count": 6960,
     "error_file_count": 213,
     "total_node_count": 7272535,
     "untranslated_node_count": 23381
   },
```

PR checklist:

- [X] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [X] Tests included or PR comment includes a reproducible test plan
- [X] Documentation is up-to-date
- [X] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [X] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #7599, opened 2023-04-20
fix(kotlin): allow interpolated idents to match explicit interpolation

brandonspark:
## What:
This PR allows interpolated idents in Kotlin, such as `"$foo"`, to be translated the same as explicitly interpolated expressions.

## Why:
This means that `"${...}"` will match `"$foo"`.

## How:
Just changed the Kotlin translation.

## Alternatives considered:
I tried making this change more generically, but a bunch of things broke.

More fundamentally, I tried to make explicitly interpolated arguments match regular arguments, so that something like `Call (InterpolatedElement, [a])`, `b` could match `a` to `b` (sort of "unpacking" the `InterpolatedElement`). This caused issues, however, with things like when the RHS was a constant string, and we would then match something interpolated to something non-interpolated, for instance ``#{$FOO}`` to ``a``, in Ruby.

I then tried to make it such that we translated `Middle3` interpolation arguments, such as `"$foo"`, to explicitly interpolated things. This ends up causing issues with other languages though, such as PHP, where we just shove everything into a `Middle3`, so this breaks things like `\`...\``.

I decided it was better to just keep this simple and local to Kotlin. This is only guaranteed to help the case where we're dealing with an implicitly interpolated identifier, which should be the same as the explicitly interpolated one anyways.

Closes PA-2711

PR checklist:


brandonspark:
The tests will be failing, but that's because of #7465.

changelog.d/pa-2711.added
```diff
@@ -0,0 +1,2 @@
+Kotlin: Interpolated identifiers in strings, such as "$foo", are now properly
+able to match explicitly interpolated expressions, like "${...}".
```
emjin, line 2: Your test has `...{...}...` (no $), is that just a language equivalence?

languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
```diff
@@ -1362,9 +1362,9 @@ and interpolation (env : env) (x : CST.interpolation) =
       let v3 = token env v3 (* "}" *) in
       Right3 (v1, Some v2, v3)
   | `DOLLAR_simple_id (v1, v2) ->
-      let _v1 = token env v1 (* "$" *) in
+      let v1 = token env v1 (* "$" *) in
       let v2 = simple_identifier env v2 in
-      Middle3 (N (Id (v2, empty_id_info ())) |> G.e)
+      Right3 (v1, Some (N (Id (v2, empty_id_info ())) |> G.e), v1)
```
emjin, line 1362: Wow I hate this use of `Middle3` which gets ultimately resolved in AST_generic.interpolated for readability but that's a larger complaint

emjin (approved):
Seems reasonable to me

aryx:
update?

changelog.d/pa-2711.added
brandonspark: oops, yeah that's a mistake on my part, good catch

PR #7609, opened 2023-04-24
bring back libpcre + move yaml stuff into libyaml

tpetr:
We learned the hard way that Javascript's regex engine doesn't support all the PCRE features that semgrep uses. This PR adds back the libpcre WASM module.

Note that the pcre bindings I wrote work for our purposes, but are gross and half-complete (see `js/libpcre/runtime.js`). Worst-case scenario I'll improve in a subsequent PR.


js/Makefile
```diff
@@ -3,12 +3,14 @@ default: libyaml engine languages
 
 .PHONY: build
 build:
+	cd libpcre; $(MAKE) build
```
aryx, line 6: At this point you can use a for loop ...
```
rec:
	set -e; for i in $(MAKESUBDIRS); do $(MAKE) -C $$i all || exit 1; done
```

aryx (approved):

PR #5131, opened 2022-05-02
GitHub codespaces

ethompsy:
PR checklist:

- [ ] Documentation is up-to-date
- [ ] Changelog is up-to-date
- [ ] Change has no security implications (otherwise, ping security team)

PR #7610, opened 2023-04-24
fix semgrep-dev docker build

tpetr:
At some point the `pre-commit` alpine package took a dependency on python 3.11, which broke docker builds because alpine 3.17's `python3` package installs 3.10. Installing `python3` from edge fixes the build for now.

### DIFF
--- a/dockerfiles/semgrep-dev.Dockerfile
+++ b/dockerfiles/semgrep-dev.Dockerfile
@@ -26,7 +26,8 @@ RUN apk add --no-cache \
 #
 # Install python >= 3.11 needed by pre-commit (for what operation?)
 RUN apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/main python3 && \
-  [[ "$(python3 --version)" =~ 3\.11\.[0-9]+ ]]
+  python3 --version > /tmp/python-version && \
+  grep -F 3.11 /tmp/python-version
 # Install version XXX of pre-commit needed for YYY (requires python >= 3.11):
 RUN apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing pre-commit
 
--- a/dockerfiles/semgrep-dev.Dockerfile
+++ b/dockerfiles/semgrep-dev.Dockerfile
@@ -17,10 +17,13 @@ FROM returntocorp/semgrep:develop
 # Various utilities. We can always install them during the CI job but it's
 # it's nice to do it here while we're root.
 #
+# TODO: remove the explicit python3 install once our version of alpine has python3.11
+#
 RUN apk add --no-cache \
   bash \
   curl \
   jq && \
+  apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/main python3 && \
   apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing pre-commit
 
 # Let the user know how their container was built
--- a/dockerfiles/semgrep-dev.Dockerfile
+++ b/dockerfiles/semgrep-dev.Dockerfile
@@ -17,14 +17,18 @@ FROM returntocorp/semgrep:develop
 # Various utilities. We can always install them during the CI job but it's
 # it's nice to do it here while we're root.
 #
-# TODO: remove the explicit python3 install once our version of alpine has python3.11
-#
 RUN apk add --no-cache \
   bash \
   curl \
-  jq && \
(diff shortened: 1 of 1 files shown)


dockerfiles/semgrep-dev.Dockerfile
```diff
@@ -17,10 +17,13 @@ FROM returntocorp/semgrep:develop
...
 # it's nice to do it here while we're root.
 #
+# TODO: remove the explicit python3 install once our version of alpine has python3.11
+#
 RUN apk add --no-cache \
   bash \
   curl \
   jq && \
+  apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/main python3 && \
   apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing pre-commit
```
mjambon: I have questions:
* why are these commands run in the same `RUN` instruction? If you don't know better than I do, let's break it into 3 RUNs.
* what requires python 3.11? -> comment
* adding a test for python 3.11:
```dockerfile
# install python 3.11 needed for XXX
RUN apk add ... && [[ "$(python3 --version)" =~ 3\.11\.[0-9]+ ]]
```
* while we're at it, let's add: `# TODO: explain why we need a specific version of pre-commit and what version it is`

mjambon (approved):

mjambon:
ugh, dockerfile uses sh, not bash. I need to fix my contribution.

mjambon:
There are errors on "tests / build-test-javascript / build-test-npm-packages" but I'm going to ignore them since we only touched a dockerfile.

dockerfiles/semgrep-dev.Dockerfile
tpetr: 1. Chained `RUN` is a Dockerfile best practice. It reduces the number of layers and image size, which results in faster pulls. Let's bring this back in a future PR.
2. Thanks for adding the clarifying comment re: python 3.11 and pre-commit. I wonder if we even need pre-commit in this image? (Probably, I haven't taken the time to find where it's being used)
3. Two things:
- The `python3 --version` check isn't doing what you expect. The docker image (`alpine-python`) manually installs python (and other python-related things) outside of apk, so your check is not actually confirming that a satisfactory version of the `python3` apk package is installed. You'd need to check the output of something like `apk list --installed python3`.
- However, I'm not convinced that this explicit check is useful as-is, since it simply fails the build without relaying helpful information to the user. I feel that the original `apk` error output did a decent job telegraphing the problem (that `pre-commit` wants 3.11 but alpine only has a 3.10 package) and we were simply confused because the image provides an out-of-band version of 3.11. This will hopefully all become moot when alpine 3.18 is release and/or 3.11 is added to the `main` repository.

mjambon:
> Chained RUN is a Dockerfile best practice. It reduces the number of layers and image size, which results in faster pulls. Let's bring this back in a future PR.

I prefer multiple RUNs because if one step fails, I don't have to re-run all the previous steps to get to that point. This matters a lot during development.

If we want to speed things up when building or loading the image (by how much really?), we should add a build step that squashes all the RUNs into one automatically.

mjambon:
> The python3 --version check isn't doing what you expect. The docker image (alpine-python) manually installs python (and other python-related things) outside of apk, so your check is not actually confirming that a satisfactory version of the python3 apk package is installed. You'd need to check the output of something like apk list --installed python3.

Sure. This is one way of saying this is all broken.

Wouldn't these dependency problems be solved by installing all the apk packages from the same repository?

mjambon:
The line
```
apk add --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing pre-commit
```
was added in https://github.com/returntocorp/semgrep/pull/5131 which was reviewed by @underyx . Maybe he knows what's going on.

tpetr:
> Wouldn't these dependency problems be solved by installing all the apk packages from the same repository?

Yep, one solution could be to update `/etc/apk/repositories` to point to the `edge` branch instead of `3.17` (though the Alpine folks [discourage using edge](https://wiki.alpinelinux.org/wiki/Edge)).

After thinking about this for a bit, though, I think an even easier approach would be to install `pre-commit` via pip or [pipx](https://pypa.github.io/pipx/) instead of apk.

aryx:
There is no documentation in https://github.com/returntocorp/semgrep/pull/5131 on how to use this github codespace thing. None of is relying on pre-commit in the semgrep dev container. I would vote to just remove those things ... unless someone comment on how to use this codespace thing. @underyx do you use this github codespace thing with the semgrep dev container?

PR #7611, opened 2023-04-24
fix(cli): git URL parsing for subgroups

brandonspark:
y'all would have thought this was Lagrange's theorem, because today we're talking about subgroups

## What:
This PR fixes URL parsing for URLs which have a "subgroup" structure, which is to say that it is in a "nested grouping structure". Previously, we would butcher these URLs, and produce things like `https://https//gitlab.com/r2cexamples2/group2/group3/chess-game` from `https://gitlab.com/r2cexamples2/group2/group3/chess-game.git`.

## Why:
We want clickable URLs.

## How:
Made it such that when parsing the "owner" of a project, it can include slashes (so it might be multiple such entries in the URL, instead of just one).

## Test plan:
Added a test suite for git URL parsing, which we can use for the future.

I don't actually know what the end target URL is supposed to look like, but I presume it should look similar to what the working example is (which produces `https://gitlab.com/r2cexamples2/webgoat` from `https://gitlab.com/r2cexamples2/webgoat.git`)

Closes PA-2669

PR checklist:

### DIFF
--- a/changelog.d/pa-2669.fixed
+++ b/changelog.d/pa-2669.fixed
@@ -1,0 +1,3 @@
+CLI: Fixed a bug where Git projects with URLs with subgroups would not parse correctly,
+and produce non-clickable links in Semgrep App. These are such as:
+https://gitlab.com/example/group2/group3/test-case.git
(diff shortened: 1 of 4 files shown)


chmccreery (approved):

brandonspark:
ignore the snapshot changes, they were from a previous iteration of the PR

PR #7616, opened 2023-04-25
Implement the Meta module

dinosaure:
PR checklist:

This PR introduces the new module `Meta.ml` which allows `osemgrep` to aggregate metadata from environment. Several questions persists and I reach the point where the code seems correct. However, the PR is not finished where I implemented only Git and GitHub modules.

These are the questions that arise and may have an impact on the continuity of this work:
1) Dependencies, currently I added 3 dependencies:
  - `bos` to invoke Git and get some informations about the Git repository
  - [`digestif`](https://github.com/mirage/digestif) to have a clean representation of the SHA1
  - [`emile`](https://github.com/dinosaure/emile) to have something which represents an email address (correctly)

  Only `bos` is highly required even if we can replace it by something else. About SHA1 and email address, we can keep a simple `string` to represent them.

2) the `Glom` module. It's a poor lense module to get metadata from a JSON file available into the CI environment. This module is only used to implement the `GitHub` module. I'm not sure that something apparent already exists so I implemented something quick and efficient for the purpose. The main point is that it uses a GADT to "coerce" JSON types into caml types (which can increase the complexity and understanding of this part of the code).

3) The design of `Meta.ml` in general. `meta.py` is built around a base class which the other classes must inherit. This kind of mechanism exists in OCaml, but the use of objects and classes is not common with this language. I decided to start with a fairly simple model of a `module type S` implementing the properties expected of classes in Python and which the various implementations should respect.

4) What seems to be most interesting is the "compilation" of the context into a Python dictionary (the `to_dict` method). I have reproduced this method in "compiling" to a "record" `type 'extension t` which can be extended (I have indeed noticed that the dictionary can have more metadata than initially expected).

5) The last point concerns the collection of information. I've tried to detach the calculation of the various metadata from what should be done when we run the `Cmdliner.Term.eval`. Thus, I expose "cmdliner" terms so that they can be reused (and especially composed with what will happen to `cli_ci/Ci_CLI.ml`). I could have just used `Sys.getenv` but I prefer not to base my work on any global state.

For these different points, this is the design that should follow for the other contexts (CircleCI, Bitbucket, etc.). In order to avoid wasting time, I would like to clarify these points before continuing further in order to start moving in the right direction :+1:.


src/osemgrep/cli_ci/Meta.mli
```diff
@@ -0,0 +1,66 @@
+(** Collect information about the project from the environment. *)
```
aryx: Can you rename to Project_metadata.ml?

And in the .ml toplevel comment says it was translated from meta.py

src/osemgrep/cli_ci/Meta.mli
```diff
@@ -0,0 +1,66 @@
...
+
+type 'e t = {
```
aryx: I have no idea what meta.py is doing. Can you add a toplevel comment in the .mli explaining
the goal of this module. Is it use to collect project metadata to send metrics data to semgrep's server?
What is a project? Should we define a Project.ml? It is git related?

aryx: Is it a project in the sense of the Semgrep app? Again maybe a Project.ml would be useful?

src/osemgrep/cli_ci/Meta.ml
```diff
@@ -0,0 +1,742 @@
...
+
+Try running the command yourself to debug the issue.|}
+      in
+      Logs.warn (fun m -> m fmt Bos.Cmd.pp cmd);
+      failwith "Error when we run a git command"
+
+(* XXX(dinosaure): [meta.py] uses a library [glom] which is able to get values
+   from a JSON value. This module implements what we need to get some values
+   from a JSON file. It's like a poor lense for JSON... *)
+module Glom = struct
```
aryx: You can put that in a separate file then, and give it a proper name (and in its toplevel comment
just say it imiates the Python Glom library).

aryx (changes requested):

src/osemgrep/cli_ci/Meta.mli
```diff
@@ -0,0 +1,66 @@
...
+  val get_merge_base_ref : env -> Digestif.SHA1.t option Lwt.t
+
+  (** Cmdliner helpers. *)
+
+  val env : env Cmdliner.Term.t
+  val make : env -> extension t
+  val term : extension t Cmdliner.Term.t
+end
+
+module Git : S with type extension = unit
```
aryx: I'm not a huge fan of nested modules. Would it be possible to move this module in a separate file?
Git_metadata.ml and Github_metadata.ml maybe?

aryx (changes requested):

aryx:
I'm fine with Bos, digestif, and emile.

aryx:
But is meta.py needed for the metrics work? I see you put the code in cli_ci/ which we actually don't want to port for now.

src/osemgrep/cli_ci/Meta.mli
dinosaure: > Is it a project in the sense of the Semgrep app? Again maybe a Project.ml would be useful?

As far as I understand, the main goal of `meta.py` is to aggregate metadata when `semgrep` is used by a CI (and, by this way, `semgrep` can lint a project for every PRs). We can surely rename it `Project` if you prefer. I don't have strong opinion about names.

dinosaure:
> But is meta.py needed for the metrics work? I see you put the code in cli_ci/ which we actually don't want to port for now.

Currently, it's more about the `ci` than `metrics`. `meta.py` is more related to when we use `semgrep` into a CI to lint a project.

aryx:
You can undraft when you think it's ready to be merged

aryx:
the circleci: semgrep failures are legit!
(you should configure pre-commit locally to find them before CI tells you about it)

semgrep.opam
```diff
@@ -70,6 +70,8 @@ depends: [
   "js_of_ocaml-ppx" {= "5.1.1"}
   "ctypes_stubs_js"
   "integers_stubs_js"
+  "emile"
```
aryx, line 56: you also need to add digestif.c also here.
semgrep.opam is the only thing we use right now I think in CI to install packages.

aryx (changes requested):
you need to fix the CI (legit) errors

commons.opam
```diff
@@ -28,6 +28,8 @@ depends: [
   "ppx_hash" {>= "v0.14.0"}
   "parmap" {>= "1.2.4"}
   "odoc" {with-doc}
+  "digestif" {>= "1.0.0"}
```
aryx: Sorry about the inconvenience, but can you merge the latest develop?
Those things are now specified in the dune-project file.

src/osemgrep/cli_ci/Git_metadata.ml
```diff
@@ -0,0 +1,171 @@
+(*****************************************************************************)
+(* Types and constants *)
+(*****************************************************************************)
+
+type extension = unit
+
+type env = {
+  _SEMGREP_REPO_NAME : string option;
```
aryx, line 8: smart, nice way to match the env variable by prefixing with a _.
I'll do that too maybe in Semgrep_envvars.mli then!

dinosaure: Yes, it's a question of style that one may not like 🙂

aryx (approved):

PR #7619, opened 2023-04-25
only run semgrep pro tests if repo is owned by returntocorp

tpetr:
will test by forking this branch

### DIFF
--- a/.github/workflows/test-semgrep-pro-latest.yaml
+++ b/.github/workflows/test-semgrep-pro-latest.yaml
@@ -13,6 +13,7 @@ on:
 jobs:
   test-semgrep-pro-latest:
     name: Test Semgrep Pro Engine (latest release)
+    if: github.repository_owner == 'returntocorp' # only returntocorp has the necessary credentials to access semgrep pro
     runs-on: ubuntu-22.04
     permissions:
       id-token: write
--- a/.github/workflows/test-semgrep-pro.yaml
+++ b/.github/workflows/test-semgrep-pro.yaml
@@ -42,6 +42,7 @@ jobs:
           fi
   test-semgrep-pro:
     name: Test Semgrep Pro Engine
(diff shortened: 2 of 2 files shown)


aryx (approved):

spencerdrak:
@tpetr thanks for opening this up - my only suggestion here would be to add another trigger to the job so that it also runs on develop. Not sure the performance implications of doing this, but, it might be a nice safety net so that we know which changes caused issues and aren't getting unrelated failures in follow-on PRs. 

Alternatively we can check this nightly, but I don't know what the team would prefer.

.github/workflows/test-semgrep-pro-latest.yaml
```diff
@@ -13,6 +13,7 @@ on:
 jobs:
```
spencerdrak, line 13:
```suggestion
  push:
    branches:
      - "develop"
jobs:
```

spencerdrak (commented):
I'll defer to @aryx or @brandonspark on what they'd prefer to see here, but this should be able to run just fine on develop.

.github/workflows/test-semgrep-pro-latest.yaml
spencerdrak: Also haven't tested this explicitly, so @tpetr may need to run another test from his fork

spencerdrak:
@tpetr thanks for opening this up - my only suggestion here would be to add another trigger to the job so that it also runs on develop. Not sure the performance implications of doing this, but, it might be a nice safety net so that we know which changes caused issues and aren't getting unrelated failures in follow-on PRs. 

Alternatively we can check this nightly, but I don't know what the team would prefer.

.github/workflows/test-semgrep-pro-latest.yaml
brandonspark: i see no reason not to do this, so i'm down

aryx (approved):

PR #7622, opened 2023-04-25
osemgrep: color complete log lines, avoid the logging level

hannesm:
test plan:
 make core
 osemgrep --config r/python --include foo --exclude foo .

-> now the line "Paths that match both --include and --exclude will be skipped by Semgrep." is yellow

Though this does no longer compose -- if styles in log messages are used, the color will be dropped (the Fmt library does composition (reading the current style, and adding the new style), but does not expose the required functionailty basically for "style start" fmt "style end", only if fmt is known upfront.

I don't think this is an issue since all bold/color styling is only done in normal application output (not in warning/error) -- but I wanted to mention it.

PR checklist:

### DIFF
--- a/libs/commons/Logs_helpers.ml
+++ b/libs/commons/Logs_helpers.ml
@@ -1,23 +1,50 @@
+(* ansii escape sequences for colored output, depending on log level *)
+let color level =
+  match level with
+  | Logs.Warning -> Some "33" (*yellow*)
+  | Logs.Error -> Some "31" (*red*)
+  | _else -> None
+
+(* print an ansi escape sequence *)
+let pp_sgr ppf style =
+  Format.pp_print_as ppf 0 "\027[";
+  Format.pp_print_as ppf 0 style;
+  Format.pp_print_as ppf 0 "m"
+
+(* log reporter *)
+let reporter ?(dst = Format.err_formatter) () =
+  let report _src level ~over k msgf =
+    let pp_style, style, style_off =
+      match color level with
+      | None -> ((fun _ppf _style -> ()), "", "")
+      | Some x -> (pp_sgr, x, "0")
+    in
+    let k _ =
+      over ();
+      k ()
+    in
+    let r =
+      msgf @@ fun ?header:_ ?tags:_ fmt ->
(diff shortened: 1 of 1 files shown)


hannesm:
This also drops the `<app-name> [log level]` prefix from log output.

libs/commons/Logs_helpers.ml
```diff
@@ -1,23 +1,50 @@
+(* ansii escape sequences for colored output, depending on log level *)
```
aryx, line 1: ANSI

libs/commons/Logs_helpers.ml
```diff
@@ -1,23 +1,50 @@
...
+let color level =
+  match level with
+  | Logs.Warning -> Some "33" (*yellow*)
+  | Logs.Error -> Some "31" (*red*)
+  | _else -> None
+
+(* print an ansi escape sequence *)
```
aryx, line 8: ANSI

libs/commons/Logs_helpers.ml
```diff
@@ -1,23 +1,50 @@
...
+let pp_sgr ppf style =
+  Format.pp_print_as ppf 0 "\027[";
```
aryx, line 10: there's an OCaml library ANSITerminal that we could use maybe here (we already use it in libs/commons/Console.ml,
so no need to add it to the dependencies). Or maybe just add a comment explaining why we don't use
ANSITerminal lib here (not worth it).

aryx (approved):
I didn't understand your comment in the PR about the composition, but maybe add
an explaination in Logs_helpers.mli about limitations of the reporter.

hannesm:
> I didn't understand your comment in the PR about the composition

It is not relevant here and now for our purposes.

If you would `` Logs.err (fun m -> m "foo %a bar" Fmt.(styled `Bold string) "hello") ``, you will see:
foo (* in red *) hello (* in red & bold *) bar (* plain default foreground *) -- while the expectation would be to see the "bar" in red (and not bold) again

The underlying reason is that Fmt preserves and merges styles. Unfortunately I haven't found a way to apply the color using Fmt.styled to the entire (user-supplied) formatter (so you see the manual emission of ANSI escape sequences in this PR).

PR #7623, opened 2023-04-25
fix(swift): let lambdas translate to IL

brandonspark:
## What:
This PR makes it so that lambdas in Swift now correctly survive to the IL.

## Why:
This prevents taint analysis on the interior of lambdas.

## How:
Previously, this was happening because of injecting into a new variant, `OSWS_Closure`, however this ends up putting the body of the lambda in an unhandled construct, creating a `Fixme` node. We just injected it into the list of statements denoting the lambda, because it could potentially be useful still for matching purposes. This shouldn't affect any semantic analysis, though.

Closes PA-2718

PR checklist:

### DIFF
--- a/languages/swift/generic/Parse_swift_tree_sitter.ml
+++ b/languages/swift/generic/Parse_swift_tree_sitter.ml
@@ -1849,13 +1849,18 @@ and map_lambda_literal (env : env) ((v1, v2, v3, v4) : CST.lambda_literal) :
     (* We inject into an `OtherStmtWithStmt` here because we needed somewhere for the
        capture list to go. This is not permitted with just a regular statement, so there is
        a new variant `OSWS_Closure` to signify this.
+       We put it into the top-level statements so that we do not prevent the inner statements
+       from surviving IL translation. It shouldn't mess anything up, because it shouldn't be
+       translated.
     *)
+    let capture_group_stmt =
+      G.exprstmt
+        (G.OtherExpr
+           (("CaptureGroup", v1), Common.map (fun x -> G.Pa x) captures)
+        |> G.e)
+    in
     G.FBStmt
-      (G.OtherStmtWithStmt
-         ( G.OSWS_Closure,
-           Common.map (fun x -> G.Pa x) captures,
-           G.Block (Tok.unsafe_fake_bracket stmts) |> G.s )
-      |> G.s)
+      (G.Block (Tok.unsafe_fake_bracket (capture_group_stmt :: stmts)) |> G.s)
   in
   let _rb = (* "}" *) token env v4 in
   let def =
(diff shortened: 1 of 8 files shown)


aryx:
LGTM. @IagoAbal ?

languages/swift/generic/Parse_swift_tree_sitter.ml
```diff
@@ -1849,13 +1849,18 @@ and map_lambda_literal (env : env) ((v1, v2, v3, v4) : CST.lambda_literal) :
     (* We inject into an `OtherStmtWithStmt` here because we needed somewhere for the
```
IagoAbal: Doesn't this comment now need to be fixed?

languages/swift/generic/Parse_swift_tree_sitter.ml
```diff
@@ -1849,13 +1849,18 @@ and map_lambda_literal (env : env) ((v1, v2, v3, v4) : CST.lambda_literal) :
...
     *)
+    let capture_group_stmt =
+      G.exprstmt
+        (G.OtherExpr
+           (("CaptureGroup", v1), Common.map (fun x -> G.Pa x) captures)
+        |> G.e)
+    in
     G.FBStmt
-      (G.OtherStmtWithStmt
-         ( G.OSWS_Closure,
```
IagoAbal, line 1855: Do we still need `OSWS_Closure` then?

languages/swift/generic/Parse_swift_tree_sitter.ml
```diff
@@ -1849,13 +1849,18 @@ and map_lambda_literal (env : env) ((v1, v2, v3, v4) : CST.lambda_literal) :
...
+       translated.
     *)
+    let capture_group_stmt =
+      G.exprstmt
+        (G.OtherExpr
+           (("CaptureGroup", v1), Common.map (fun x -> G.Pa x) captures)
+        |> G.e)
+    in
     G.FBStmt
-      (G.OtherStmtWithStmt
```
IagoAbal, line 1854: I'm not arguing for one solution or the other, but is this better than adding a case for `OSWS_Closure` in `AST_to_IL` ? Why was `OSWS_Closure` added in the first place?

IagoAbal (approved):
Looks OK to me, but I would check if we can just drop using `OSWS_Closure` (and if so, we should clean up the Generic AST since this seems to have been added for Swift only). Actually you (Brandon) added it in: ac25c765a57 "feat(swift): most of swift parsing (#5870)".

languages/swift/generic/Parse_swift_tree_sitter.ml
brandonspark: it was added because at that time, i didn't know enough to use `OtherStmt` as much as ive learned to now. i really wanted somewhere for the capture list to go, and i wanted it to be "nice", but now i understand that this suffices

i also now reviewed what capture lists in Swift do and i am highly suspicious we would ever use it for semantic analysis. so now i am fine throwing it in the garbage. before, i think i had hopes we might use it.

languages/swift/generic/Parse_swift_tree_sitter.ml
brandonspark: nope! fixed

languages/swift/generic/Parse_swift_tree_sitter.ml
brandonspark: yes, thank you

PR #7624, opened 2023-04-25
chore(core): flag to not do name resolution/const prop

brandonspark:
## What:
This PR adds in a `-no_resolving` flag to `semgrep-core`, which will cause ASTs to be parsed without constant propagation or naming.

## Why:
This would allow speedier searches for the VS Code extension, possibly.

## Test plan:
I ran it and it was noticeably faster. 5s to 2.5s on `django`

PR checklist:

### DIFF
--- a/src/configuring/Flag_semgrep.ml
+++ b/src/configuring/Flag_semgrep.ml
@@ -71,3 +71,9 @@ let equivalence_mode = ref false
 
 (* One-off experiment for Raja (See Raja_experiment.ml) *)
 let raja = ref false
+
+(* No constant propagation or naming
+   This is useful for quick searches which don't need full
+   semantic information. We could use this for the VS Code extension.
+*)
+let no_resolving = ref false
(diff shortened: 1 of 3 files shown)


nmote:
I think people understand that searching a whole codebase might take a little bit. I'm concerned that this will lead to a strange user experience where a pattern gives one set of results in one interface, and a different set of results in another. I'd prefer to keep things consistent across the board

brandonspark:
@nmote i personally would never use a searching tool which took more than a few seconds. as it stands currently, `django` is not even a large repository. we take 40s to parse the `vscode` repository alone, and im certain more than half of that is just resolution. personally, i think its easier to put a disclaimer on the extension that inline searches might be slightly different than that of semgrep CLI

at any rate, this flag isn't in use yet, but i'm using it in my own terminal for benchmarking purposes, so we can discuss more on the PR that interfaces with the extension, i think

aryx (commented):
No. Naming and constant propagation is part of Semgrep.

PR #7625, opened 2023-04-25
Clean up, comments, and move login nudge to VSCode

ajbt200128:
Added some comments, renamed some stuff. Moved login nudge logic to VSCode. Also fixed bug where if the extension opened on the "main" vscode page, it'd crash as it would have no working folders
PR checklist:


aryx:
What is the reproducible test plan (one of the item in the check list above).
We generally try to add a `test plan:` section in the PR that explains how one can test the contribution of the PR.
Somtimes it can be as simple as 'make test', but sometimes it requires more explanations.
It's useful for the reader to understand the motivation of the work, and it's also useful in 6 months if someone use git blame.

aryx (approved):
Great. We usually use the _helpers and utils_ suffix, and it's also fine to keep
the comments in the .ml next to the function.
One might thing that it's a bit tedious to have to duplicate things in the .ml and .mli, but IMHO this is one of the best thing of OCaml; to have separate signature files (the .mli) so one can read and focus just on the interface of a file. It requires a little bit of duplication, but it's totaly worth it.

PR #7627, opened 2023-04-26
add liquidjs as an explicit dev dependency

tpetr:
We've seen intermittent failures in the JS build because I forgot to add `liquidjs` as an explicit dev dependency, which results in races when multiple instances of npx try to globally install liquidjs at the same time.

### DIFF
--- a/js/languages/bash/package-lock.json
+++ b/js/languages/bash/package-lock.json
@@ -13,7 +13,8 @@
       },
       "devDependencies": {
         "esbuild": "^0.17.17",
-        "jest": "^29.5.0"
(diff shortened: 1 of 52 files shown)


aryx (approved):
All this duplication show the need maybe to factorize things? Maybe we can autogenerate those package.json as one of the Make step?

tpetr:
We already auto-generate package.json via make, they just get checked into GitHub once they’ve been created. I agree that it’s getting a bit ridiculous, though. I can move package.json generation into the dist folder (i.e. autogenerate at build time) in a separate PR 👍

PR #7631, opened 2023-04-26
osemgrep, match report: fix highlighting (otherwise out of bounds)

hannesm:
test plan:
make core
osemgrep --config r/python src/osemgrep/TOPORT/util.py

previously, there was an out of bounds error, since the assumption was that start line and end line was the same (and thus the cut point was negative).

now, the issue was solved, and for safety there's a "max start ..", so idx2 will always be >= idx1 in cut

PR checklist:

### DIFF
--- a/src/osemgrep/reporting/Matches_report.ml
+++ b/src/osemgrep/reporting/Matches_report.ml
@@ -165,11 +165,14 @@ let pp_finding ~max_chars_per_line ~max_lines_per_finding ~color_output
           else col m.start.col + ellipsis_len (line_off > 0)
         in
         let end_color =
-          if line_number >= m.end_.line then
-            min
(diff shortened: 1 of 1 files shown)


hannesm:
The computation is intricate (due to ellipsis and dedenting), the python code first highlights (makes text bold) of the line, and then cuts the part to be displayed.

In OCaml, we instead compute the offsets where to start and stop making text bold.

aryx (approved):
Would be good to add test cases for those.

aryx:
The test case could be a new tests/e2e/... that contains an example that was triggering an array out of bound in osemgrep.

hannesm:
yes, I'll add test cases for the rendering later this week.

PR #7633, opened 2023-04-26
fix js build

tpetr:
The JS build wasn't installing opam deps, which was fine until `ocaml-layer` fell out of sync.

~This is just a quick fix to unblock the release; I'll submit a better fix (i.e. figure out why `make install-deps-for-semgrep-core` wasnt working (see [comment](https://github.com/returntocorp/semgrep/blob/develop/.github/workflows/build-test-javascript.yaml#L34))) later today.~ <-- got this working

### DIFF
--- a/.github/workflows/build-test-javascript.yaml
+++ b/.github/workflows/build-test-javascript.yaml
@@ -39,7 +39,7 @@ jobs:
           # TODO: figure out why we need to do this
           ln -s $TREESITTER_INCDIR/tree_sitter /usr/include/tree_sitter
 
-          opam install -y --deps-only ./libs/ocaml-tree-sitter-core
+          opam install -vv -y --deps-only ./libs/ocaml-tree-sitter-core
           opam install -y --deps-only ./
 
           dune build js --profile=release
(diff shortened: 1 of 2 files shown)


aryx (approved):

aryx:
still failing

tpetr:
Yep, it occurred to me right after I pushed that this was the exact reason why the makefile recipe wasn't working 🤦

tpetr:
JS build is now failing ~due to jsoo being incompatible with the recent language server changes (I believe):~ (see https://github.com/returntocorp/semgrep/pull/7633#issuecomment-1524027821)

```
FAIL src/index.test.js (12.699 s)
  ✕ it loads the engine (313 ms)

  ● it loads the engine

    thrown: Array [
      0,
      Array [
        248,
        "Failure",
        -3,
      ],
      "lwt_unix_iov_max not implemented",
    ]

      1 | const { EngineFactory } = require("../dist/index");
      2 |
    > 3 | test("it loads the engine", async () => {
        | ^
      4 |   const engine = await EngineFactory("./dist/semgrep-engine.wasm");
      5 |   expect(engine.lookupLang("python")).toBe(24);
      6 | });
```

This is different from the original issue this PR was opened to fix, so I'm going to merge this and figure out the lwt issue in a separate PR.

tpetr:
`git bisect` tells me that https://github.com/returntocorp/semgrep/pull/7632 introduced the engine's dependency on lwt. We might have to wrap some of this in `Common.jsoo` checks -- digging into it (cc @aryx )

PR #7639, opened 2023-04-26
feat!: interprocedural taint labels

brandonspark:
## What:
This PR adds in support for interprocedural taint labels, with the exception of labeled propagators, which have been downscoped.

## Why:
SR needs interprocedural taint labels!!!

## How:
This executes the last few steps of my scope doc here: https://www.notion.so/r2cdev/Interprocedural-Taint-Labels-138ca7ca4d21463a8f481ccc9995903b?pvs=4

Essentially, the last remaining step in this was in implementing the notion of "preconditions" and "hypothetical taint". A hypothetical taint is when a taint may or may not exist, depending on the input of polymorphic taint variables. We carry these along during our taint analysis, and then decide whether or not to emit a finding once we substitute for the argument taints at the function call-site.

## Future work:
I purposefully left out the implementation for labeled propagators. Since they're new, I surmise they probably won't be very used yet, and I want to get this change out soon. It's also a more complex problem than I imagined, you would need to be able to conjoin two different preconditions together in order to handle substituting for arguments, as well as change the type representation so that `Arg` taint can have preconditions too. Ultimately, I wanted to keep the edit distance small, and this is already a substantial change.

## Test plan:
See https://github.com/returntocorp/semgrep-proprietary/pull/700 for tests, and performance.

Closes PA-2507

PR checklist:


src/tainting/Taint.ml
```diff
@@ -94,30 +75,129 @@ let _show_arg { pos = s, i; offset = os } =
     Printf.sprintf "arg(%s)#%d.%s" s i os_str
   else Printf.sprintf "arg(%s)#%d" s i
 
+(*****************************************************************************)
+(* Preconditions *)
+(*****************************************************************************)
+
+(* We roll our own boolean formula type here because we need to be
+   able to use polymorphic compare. Comparison on the Generic AST
+   isn't possible, and polymorphic compare would otherwise take into account
```
IagoAbal: There are comparison functions for Generic AST that take care of this. But at some point having our own type may be cleaner so I'm not against it.

src/tainting/Taint.ml
```diff
@@ -94,30 +75,129 @@ let _show_arg { pos = s, i; offset = os } =
...
+*)
+type precondition =
+  | Label of string
+  | Bool of bool
+  | And of precondition list
+  | Or of precondition list
+  | Not of precondition
+[@@deriving show]
+
+let rec expr_to_precondition e =
```
IagoAbal: I think it's better that we replace the `check` in `parse_taint_requires` with a function  that already checks it _and_ convert it into `Taint.precondition` ? Maybe that requires moving the `precondition` type to `Rule`. But otherwise we would need to keep both in sync.

src/tainting/Taint.ml
```diff
@@ -94,30 +75,129 @@ let _show_arg { pos = s, i; offset = os } =
...
+        [ { taint with orig = Src src } ]
+    | Src ({ precondition = Some (incoming, expr); _ } as src) ->
+        let new_incoming = List.concat_map aux incoming in
+        let new_precondition = Some (new_incoming, expr) in
+        [
+          { taint with orig = Src { src with precondition = new_precondition } };
+        ]
+  in
+  (* We only want to replace preconditions! This function is called after
+     we already have replaced the top-level Arg taints, meaning that if we
```
IagoAbal: Sounds like we are doing this in two steps, couldn't we just have a single function that does everything so we avoid this kind of explanation?

src/tainting/Taint.ml
```diff
@@ -94,30 +75,129 @@ let _show_arg { pos = s, i; offset = os } =
...
-  Src { call_trace = PM (pm, x); label = x.label }
+and orig = Src of source | Arg of arg [@@deriving show]
+and taint = { orig : orig; tokens : tainted_tokens } [@@deriving show]
+
+let substitute_precondition_arg_taint ~arg_fn taint =
+  let rec aux taint =
+    match taint.orig with
+    | Arg arg -> arg_fn arg
+    | Src ({ precondition = None; _ } as src) ->
+        [ { taint with orig = Src src } ]
```
IagoAbal: Isn't this the same as just `[ taint ]` ?

src/tainting/Taint.ml
```diff
@@ -94,30 +75,129 @@ let _show_arg { pos = s, i; offset = os } =
...
    * For our purposes, this is OK.
    *)
   let pm1, ts1 = pm_of_trace s1.call_trace
   and pm2, ts2 = pm_of_trace s2.call_trace in
-  Stdlib.compare
-    (pm1.rule_id, pm1.range_loc, pm1.env, s1.label, ts1.Rule.label)
-    (pm2.rule_id, pm2.range_loc, pm2.env, s2.label, ts2.Rule.label)
+  match
+    Stdlib.compare
+      (pm1.rule_id, pm1.range_loc, pm1.env, s1.label, ts1.Rule.label)
```
IagoAbal: Couldn't just add the precondition as yet another element in this tuple?

changelog.d/pa-2507.added
```diff
@@ -0,0 +1,2 @@
+Pro: Taint: Taint labels now work interprocedurally, for everything except
+labeled propagators.
```
IagoAbal: What happens if you add labeled propagators then?

src/tainting/Taint.mli
```diff
@@ -19,23 +36,26 @@ type source = {
         this label may have changed, for instance by being propagated to
         a different label.
       *)
+  precondition : (taint list * precondition) option;
```
IagoAbal: It would better if we could just keep everything as a `Taint_set.t` instead of `taint list`s. Are duplicates desirable in the precondition?

src/tainting/Dataflow_tainting.ml
```diff
@@ -435,48 +435,224 @@ let partition_mutating_sources sources_matches =
   |> List.partition (fun (m : R.taint_source tmatch) ->
          m.spec.source_by_side_effect && is_exact m)
 
+let propagate_taint_via_java_setter env e args all_args_taints =
```
IagoAbal: Has this function also changed? It would be great if this kind of changes weren't mixed with moving code around because it makes it a lot harder to review.

src/tainting/Dataflow_tainting.ml
```diff
@@ -824,7 +838,18 @@ let handle_taint_propagators env thing taints =
         (* Only propagate if the current set of taint labels can satisfy the
            propagator's requires precondition.
         *)
-        if eval_label_requires ~labels prop.spec.prop.propagator_requires then
+        (* TODO(brandon): Interprocedural propagator labels
+           This is trickier than I thought. You have to augment the Arg taints
+           with preconditions as well, and allow conjunction, because when you
+           replace an Arg taint with a precondition, all the produced taints
```
IagoAbal, line 813: Not sure I understand this part. Do you have some concrete example to discuss around it?

src/tainting/Dataflow_tainting.ml
```diff
@@ -435,48 +435,224 @@ let partition_mutating_sources sources_matches =
...
+*)
+let rec taint_can_exist taint =
+  match taint.T.orig with
+  | Arg _ -> false
+  | Src { precondition = None; _ } -> true
+  | Src { precondition = Some (incoming, expr); _ } ->
+      solve_precondition (incoming, expr)
+
+and solve_precondition (incoming, expr) =
+  let labels = List.filter taint_can_exist incoming |> labels_of_taints in
```
IagoAbal: So we just ignore the `Arg` taints when evaluating the requires?

src/tainting/Taint.ml
brandonspark: you're so right. parse, don't validate

src/tainting/Taint.mli
brandonspark: no, but this will require that we make `taint` mutually recursive with `Taint_set`, and i don't actually even know how you would do that, easily. recursive modules? it seems like a big hassle, what do you think?

src/tainting/Dataflow_tainting.ml
```diff
@@ -680,24 +715,6 @@ let find_pos_in_actual_args args_taints fparams =
         "cannot match taint variable with function arguments (%i: %s)" i s;
     taint_opt
 
-(* This function is used to convert some taint thing we're holding
-   to one which has been propagated to a new label.
-   See [handle_taint_propagators] for more.
-*)
-let propagate_taint_to_label replace_labels label (taint : T.taint) =
```
brandonspark, line 704: I just moved this function.

src/tainting/Dataflow_tainting.ml
brandonspark: no, sorry, i was just re-organizing and thought it didn't fit with the headers. i can add it back.

src/tainting/Dataflow_tainting.ml
brandonspark: Yep, under hypothesis that any such `Arg` taints we encounter should have been substituted for, already. This isn't always the case, because we might run `pm_of_finding` on a finding within a function call, but then ignoring them is
the right thing to do, because we don't yet have enough information on the `Arg` taint (that will be instantiated in a different finding, if the function is called at all)

changelog.d/pa-2507.added
brandonspark: They should work as before, but just not interprocedurally. You should not expect a labelled propagator to propagate if it is within a function call, because it should almost never propagate.

src/tainting/Taint.ml
brandonspark: The precondition doesn't just use polymorphic compare, it also uses `compare_taint`, which eventually calls `compare_sources` again. Since `compare_sources` itself isn't just polymorphic compare, I assume that we don't want to allow unrestricted polymorphic comparison.

src/tainting/Taint.ml
brandonspark: yep, you're correct. dunno how that happened.

src/tainting/Taint.mli
IagoAbal: Ooops yeah I didn't realize they would become recursive. OK let's deal with this later on during a clean up. But yeah the most obvious solution would be using recursive modules.

IagoAbal (approved):
Nice work!

src/tainting/Dataflow_tainting.ml
```diff
@@ -1253,24 +1276,14 @@ let check_function_signature env fun_exp args args_taints =
                        in
                        Some
                          (`Return
-                           (arg_taints
+                           (arg_taints |> Taints.of_list
```
IagoAbal: If there are no taints (or we fail to compute them) do we really want to return an empty set of taints here?

src/tainting/Dataflow_tainting.ml
```diff
@@ -1208,14 +1208,38 @@ let check_function_signature env fun_exp args args_taints =
...
+      (* Our first pass will be to substitute the args for taints.
+         We can't do this indiscriminately at the beginning, because
+         we might need to use some of the information of the pre-substitution
+         taints and the post-substitution taints, for instance the tokens.
+
+         So we will isolate this as a specific step to be applied as necessary.
+      *)
+      let first_pass_substitute_arg_with_taints arg =
+        match taints_of_sig_arg env fparams fun_exp args args_taints arg with
+        | None -> []
```
IagoAbal: I think it's better to keep `Taint_set.t` as long as it is possible. (Probably need to rethink that data structure too but that's another issue.)

src/tainting/Dataflow_tainting.ml
```diff
@@ -1208,14 +1208,38 @@ let check_function_signature env fun_exp args args_taints =
...
+          given argument, within the body of the function.
+      *)
+      (* Our first pass will be to substitute the args for taints.
+         We can't do this indiscriminately at the beginning, because
+         we might need to use some of the information of the pre-substitution
+         taints and the post-substitution taints, for instance the tokens.
+
+         So we will isolate this as a specific step to be applied as necessary.
+      *)
+      let first_pass_substitute_arg_with_taints arg =
```
IagoAbal: Kind of prefer the `arg_to_taints` name it used to have when it was defined inside the `ToSink` case.

src/tainting/Dataflow_tainting.ml
```diff
@@ -1348,36 +1341,70 @@ let check_function_signature env fun_exp args args_taints =
...
+                   | _ -> [ `UpdateEnv (dst_lval, Taints.of_list dst_taints) ])
+      in
+      (* This is our second pass for taint substitution, which will run
+          on all of the taints equally. This is because we need to do this
+          on Src taints, even if they weren't the results of substitutions,
+          because they may have dependencies on Arg taints. The alternative
+          is to specifically do the substitution in each case, which is easy
+          to overlook.
+
+          This step could be done on the args taints before we substitute,
```
IagoAbal: Can't this be done in `process_sig` already even if we don't do it directly on `args_taints` ?

IagoAbal: We're basically just calling `taints_of_sig_arg` on the `Arg` taints of the preconditions.

brandonspark: Yes, but I think it needs to be done uniformly for every case, meaning we would have to call `T.map_preconditions substitute_preconditions` in every branch, which is easy to forget, and also doesn't have a type-level distinction. I personally believe it's better to isolate this step and do it for every single case, so we can't miss it. What do you think?

src/tainting/Dataflow_tainting.ml
brandonspark: I think it doesn't matter, because the `Return []` case below would behave the same as if the item hadn't existed in the folded list in the first place, but I've changed it to early return `None` on the empty list anyways.

src/tainting/Dataflow_tainting.ml
IagoAbal: It does get complicated because of the various ways that one needs to tweak the call traces.

PR #7641, opened 2023-04-26
refactor: Make custom map visitor publicly accessible

nmote:
This moves the nonstandard parts of of the mapper into AST_generic. Now, the mapper in `Map_AST`'s only responsibility is to override methods and call the appropriate client-supplied function. This makes it possible for callsites to start using the custom visitor directly, at which point we can delete `Map_AST.mk_visitor`.

Test plan: Automated tests

PR checklist:


libs/ast_generic/AST_generic.ml
```diff
@@ -2405,3 +2395,48 @@ let is_metavar_name s = Common.( =~ ) s "^\\(\\$[A-Z_][A-Z_0-9]*\\)$"
...
+
+    method! visit_tok _env v =
+      (* Make a copy, since Tok.t is mutable *)
+      let v = { v with Tok.token = v.token } in
+      v
+
+    method! visit_expr env x =
+      let ekind = self#visit_expr_kind env x.e in
+      (* TODO? reuse the e_id or create a new one? *)
+      e ekind
```
nmote, line 2432: This calls `e`, so I had to put it below the definition. I moved `iter_no_id_info` too, for consistency.

libs/ast_generic/AST_generic.ml
```diff
@@ -2405,3 +2395,48 @@ let is_metavar_name s = Common.( =~ ) s "^\\(\\$[A-Z_][A-Z_0-9]*\\)$"
...
+  object (_self : 'self)
+    inherit ['self] iter
+    method! visit_id_info _env _info = ()
+  end
+
+(* This is based on the old behavior of Map_AST.mk_visitor. If you are adding a
+ * new callsite, you should consider whether the nonstandard behavior here makes
+ * sense for your use case. Think about using the autogenerated AST_generic.map
+ * directly. *)
+class virtual ['self] map_custom =
```
emjin: I'd call this `Map_AST_default` instead or something similar

libs/ast_generic/AST_generic.ml
```diff
@@ -2405,3 +2395,48 @@ let is_metavar_name s = Common.( =~ ) s "^\\(\\$[A-Z_][A-Z_0-9]*\\)$"
...
+ * it's not desirable to recurse into id_info since it contains resolved names
+ * and svalues which often contain nodes that are already present elsewhere in
+ * the AST. This matches the default behavior of the old mk_visitor. *)
+class virtual ['self] iter_no_id_info =
+  object (_self : 'self)
+    inherit ['self] iter
+    method! visit_id_info _env _info = ()
+  end
+
+(* This is based on the old behavior of Map_AST.mk_visitor. If you are adding a
```
emjin: To someone new "the old behavior of Map_AST.mk_visitor" wouldn't mean much. I would find it more helpful to say "This is tailored to Map_AST, see comment in Map_AST about why it makes certain choices" (if such a comment exists).

In contrast the comment above is great because it explains what's special about iter_no_id_info

emjin (approved):
lgtm

libs/ast_generic/AST_generic.ml
nmote: I will elaborate a bit, but I'm going to delete most, if not all of `Map_AST` shortly so referring to a comment there won't help much.

libs/ast_generic/AST_generic.ml
nmote: Hmm, maybe I'll call it `map_legacy`? I'm not really sure exactly what to call it. Its defining characteristic is that it exists to mimic the previous behavior.

PR #7649, opened 2023-04-26
fix(pro): fix looping dataflow from get methods

brandonspark:
## What:
This PR makes it so that dataflow does not occasionally loop when analyzing `getX` methods in a loop.

## Why:
We want analysis to terminate.

## How:
We were appending infinitely to the `Arg` taint's `offset`, making the dataflow analysis think the in-and-out-sets were constantly changing.

This PR enforces that we eventually always reach a cap, by ensuring we never append the same offset more than once, to the same arg.

## Test plan:
No longer loops on `examples-annotations/repos/java/SQLi/fp-scanner-repos/aws-doc-sdk-examples/java/example_code/kms/src/main/java/aws/example/kms/ListKeyPolicies.java` using `deepsemgrep-sqli-rules.yaml`.

See https://github.com/returntocorp/semgrep-proprietary/pull/692 for tests.

PR checklist:

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -738,10 +738,16 @@ let resolve_poly_taint_for_java_getters env lval st =
               taints
               |> Taints.map (fun taint ->
                      match taint.orig with
-                     | Arg arg ->
+                     | Arg ({ offset; _ } as arg) when not (List.mem n offset)
+                       ->
+                         (* If the offset we are trying to take is already in the
+                            list of offsets, don't append it! This is so we don't
+                            never-endingly loop the dataflow and make it think the
+                            Arg taint is never-endingly changing.
+                         *)
                          let arg' = { arg with offset = arg.offset @ [ n ] } in
                          { taint with orig = Arg arg' }
-                     | Src _ -> taint)
+                     | _ -> taint)
             in
             `Tainted taints')
     | _ :: _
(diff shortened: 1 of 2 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -738,10 +738,16 @@ let resolve_poly_taint_for_java_getters env lval st =
               taints
               |> Taints.map (fun taint ->
                      match taint.orig with
-                     | Arg arg ->
+                     | Arg ({ offset; _ } as arg) when not (List.mem n offset)
+                       ->
+                         (* If the offset we are trying to take is already in the
```
IagoAbal, line 743: Maybe add the example `while (true) { x = x.getX(); }` to the comment, it's obvious now it may no longer be in X months.

src/tainting/Dataflow_tainting.ml
```diff
@@ -738,10 +738,16 @@ let resolve_poly_taint_for_java_getters env lval st =
...
+                            list of offsets, don't append it! This is so we don't
+                            never-endingly loop the dataflow and make it think the
+                            Arg taint is never-endingly changing.
+                         *)
                          let arg' = { arg with offset = arg.offset @ [ n ] } in
                          { taint with orig = Arg arg' }
-                     | Src _ -> taint)
+                     | _ -> taint)
```
IagoAbal: No `_` please, `| Arg _ | Src _ ->` is better.

IagoAbal (approved):

PR #7651, opened 2023-04-27
Move the contents of the *.opam files to dune-project.

mjambon:
This implements the solution provided by @hhugo and @emillon and others that I missed at the time: https://discuss.ocaml.org/t/how-to-avoid-having-a-bunch-of-opam-files-at-the-project-root/11675/2

The opam files are now generated by `dune build` but we use them to install the project dependencies before running `dune build`. That's why for now we keep the nonempty ones under git control until we find a way to generate them before the build.

test plan:
```
make setup  # should not produce ugly warnings or errors from 'opam lint'
make
```

PR checklist:


Makefile
```diff
@@ -156,6 +156,9 @@ clean:
 core-clean:
 	dune clean
 	rm -f bin
+	# we still need to keep the nonempty opam files in git for
```
IagoAbal, line 159: We could do this with `git clean -fx *.opam`.

ast_generic.opam
```diff
@@ -25,23 +25,33 @@ Right now this generic AST is mostly the factorized union of the ASTs of:
...
+authors: ["Yoann Padioleau <pad@semgrep.com>"]
 license: "LGPL-2.1-only"
 homepage: "https://semgrep.dev"
-dev-repo: "git+https://github.com/returntocorp/semgrep"
 bug-reports: "https://github.com/returntocorp/semgrep/issues"
-
 depends: [
   "ocaml" {>= "4.12.0"}
-  "dune" {>= "3.2.0" }
+  "dune" {>= "2.7" & >= "3.2.0"}
```
IagoAbal, line 36: Isn't the same as `>= 3.2.0` ?

ast_generic.opam
```diff
@@ -1,5 +1,5 @@
+# This file is generated by dune, edit dune-project instead
```
IagoAbal, line 1: Is this file is generated by Dune, do we need to keep it?

aryx (approved):
The comments of iago are valid, but let's merge this. We can address those later.
This seems like a good step forward.

dune-project
```diff
@@ -1,3 +1,432 @@
 (lang dune 2.7)
```
IagoAbal, line 1: I guess this in combination with a `depends` below that requires `dune (>= 3.2.0)` is what producing those weird `"dune" {>= "2.7" & >= "3.2.0"}` constraints?

mjambon: yes, I think so.

ast_generic.opam
mjambon: I kept it because it lists dependencies. If the convention is that all the project dependencies are explicitly listed in `semgrep.opam`, then we don't need other opam files.

Makefile
mjambon: will be fixed in upcoming PR

PR #7673, opened 2023-04-28
osemgrep: Filter rules (esp. generic & regex)

hannesm:
test plan: make core && osemgrep --config r/python src/osemgrep

before: 
```
  Language      Rules   Files
 ─────────────────────────────
  Generic           8     151
  Python          385      55
  Regex             4     151
```

now:
```
  Language         Rules   Files
 ────────────────────────────────
  <multilang>          4     161
  Python             385      55
```

pysemgrep:
```
  Language      Rules   Files
 ─────────────────────────────
  <multilang>       5     150
  python          385      55                                                                
```

Not clear to me why the difference is there (4 vs 5)

PR checklist:

### DIFF
--- a/src/osemgrep/cli_scan/Core_runner.ml
+++ b/src/osemgrep/cli_scan/Core_runner.ml
@@ -137,23 +137,50 @@ let split_jobs_by_language all_rules all_targets : Lang_job.t list =
   let grouped_rules = group_rules_by_target_language all_rules in
   let cache = Find_target.create_cache () in
   grouped_rules
-  |> Common.map (fun (lang, rules) ->
-         let targets =
-           all_targets
-           |> List.filter (fun target ->
-                  rules
-                  |> List.exists (fun (rule : Rule.t) ->
-                         let required_path_patterns, excluded_path_patterns =
-                           match rule.paths with
-                           | Some { include_; exclude } -> (include_, exclude)
-                           | None -> ([], [])
-                         in
-                         Find_target.filter_target_for_lang ~cache ~lang
-                           ~required_path_patterns ~excluded_path_patterns
-                           target))
+  |> Common.map_filter (fun (lang, rules) ->
+         let rules, targets =
+           match lang with
+           | Xlang.LGeneric
+           | Xlang.LRegex ->
+               let rules, targets =
+                 List.fold_left
+                   (fun (rules, targets) rule ->
+                     let required_path_patterns, excluded_path_patterns =
+                       match rule.Rule.paths with
+                       | Some { include_; exclude } -> (include_, exclude)
+                       | None -> ([], [])
+                     in
+                     let targets' =
+                       all_targets
+                       |> List.filter (fun target ->
+                              Find_target.filter_target_for_lang ~cache ~lang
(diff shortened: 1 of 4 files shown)


aryx:
pre-commit failing

aryx (approved):
Nice

src/osemgrep/reporting/Status_report.ml
```diff
@@ -27,13 +27,24 @@ let pp_status ~num_rules ~num_targets ~respect_git_ignore lang_jobs ppf =
...
     |> List.fold_left
-         (fun acc Lang_job.{ lang; targets; rules } ->
-           (Xlang.to_string lang, [ List.length rules; List.length targets ])
-           :: acc)
+         (fun acc (lang, rules, targets) ->
+           match List.partition (fun (l, _) -> l = lang) acc with
+           | [], others -> (lang, [ rules; targets ]) :: others
+           | [ (_, [ r1; t1 ]) ], others ->
+               (lang, [ rules + r1; targets + t1 ]) :: others
+           | _ -> assert false)
```
hannesm, line 48: this ..map.. fold_left looks ugly, and likely the Common.group_by has a better complexity

src/osemgrep/cli_scan/Core_runner.ml
```diff
@@ -137,23 +137,50 @@ let split_jobs_by_language all_rules all_targets : Lang_job.t list =
...
-                           | None -> ([], [])
-                         in
-                         Find_target.filter_target_for_lang ~cache ~lang
-                           ~required_path_patterns ~excluded_path_patterns
-                           target))
+  |> Common.map_filter (fun (lang, rules) ->
+         let rules, targets =
+           match lang with
+           | Xlang.LGeneric
+           | Xlang.LRegex ->
```
hannesm, line 144: here this is special-cased for regex & generic rules, where not only targets but also rules are filtered.

the drawback is quite some complexity, in addition to the List.sort_uniq of the targets below (maybe a set would be better)

another drawback is that still generic and regex are applied to all the targets that any generic/regex rule matched:

if rule X says *ml and rule Y *html, and there are some ml and html files around, both X and Y will be applied to all ml and html files (as far as I can tell) -- does this make sense?

aryx: Y should be applied only to *html

src/osemgrep/reporting/Status_report.ml
aryx: yes ... when I fold and fold and fold and I need to thread things, sometimes I rewrite the code imperatively
and use a let res = ref [] in. and use List.iter and iter and iter, and update the res.
It's sometimes cleaner.

aryx (approved):

aryx:
maybe incomplete, but this looks like a good step. Let's merge it.

src/osemgrep/cli_scan/Core_runner.ml
hannesm: yes, and here we need to adjust osemgrep // look into the logic of TargetManager how the rules * targets are transported to Semgrep_core...

Issue #1917, opened 2020-10-29
Auto-fix has... gone haywire


msorens:
**Describe the bug**
Auto-fix for the rule given below sometimes works, but sometimes... well, you have to see it to believe it. Read on...

Input
```
class ReportDataService {

   isEmpty(data) {
    if (data === undefined || data.length === 0 || (data.length === 1 && data[0] === 'passed')) {
      return true;
    }
    return false;
  }

  isZero(count: number): boolean {
    if (count === 0) {
      return true;
    }
    return false;
  }

  isAllZeros(data) {
    const array = Object.keys(data).map(itm => data[itm]);
    const result = remove(array, function (value) {
      return value !== 0;
    });
    return this.isEmpty(result);
  }
}
```

Rule
```
rules:
- id: simple-booleans
  patterns:
    - pattern-either: 
      - pattern: |
          if ($EXPR) { return true; } else { return false; }
      - pattern: |
          if (!$EXPR) { return false; } else { return true; }
      - pattern: |
          if ($EXPR) { return true; } return false;
      - pattern: |
          if (!$EXPR) { return false; } return true;
  message: |
    simple booleans
  severity: WARNING
  fix: |
    return $EXPR;
```
Output
```
class ReportDataService {

isEmpty(data) {
    return data === undefined || data.length === 0 || (data.length === 1 && data[0] === 'passed';
  }

  isZero(count: number): boolean {
    if (count === 0) {
      return true;
    }
    return count === 0;) {
    const array = Object.keys(data).map(itm => data[itm]);
    const result = remove(array, function (value) {
      return value !== 0;
    });
    return this.isEmpty(result);
  }
}
```
As is often the case, a picture shows the problem more dramatically.
 - Two occurrences of the rule flagged: ✅ 
 - First occurrence fixed.  ✅ 
 - Second occurrence... mangled. ❌ 

<img width="1001" alt="image" src="https://user-images.githubusercontent.com/6817500/97516906-6441a080-1951-11eb-94a8-5fdee3280c79.png">

**To Reproduce**
The code mangling does not occur on semgrep.dev, but there a different problem manifests: duplicating the semicolon: https://semgrep.dev/s/5WGD/
To reproduce the main problem outlined above you can use the input fragment above or, because the repo is open source, you can use my real file: 

[...]

PR #7686, opened 2023-04-28
fix(js): Avoid dropping parentheses when autofixing

### LINKED ISSUE #6233: Autofix mangles TypeScript expression containing parentheses
**Describe the bug**
I found a case where the autofix functionality incorrectly transforms a TypeScript expression containing parentheses by eating an opening parenthesis.

**To Reproduce**
Steps to reproduce the behavior, ideally a link to https://semgrep.dev: https://semgrep.dev/s/Aq24

**Expected behavior**

The transformed expression should be

```typescript
function test(obj, first, second, val) {
  obj[first][second] = val
}

test(m, t, (e as SomeType).some_key, e as SomeType);
```

But instead it is

```typescript
function test(obj, first, second, val) {
  obj[first][second] = val
}

test(m, t, e as SomeType).some_key, e as SomeType);
        //^ Note the missing opening parenthesis.
```

nmote:
This is equivalent to #7437, but for JS. Now we include parentheses in the range for expressions, so we don't end up in weird situations where the location of an expression spans one parenthesis but not its match.

Fixes #6233

Probably fixes a couple other issues too (will check after).

Test plan:

* Automated tests
* Manual test of the case given in #6233
* Put the Menhir parser first in `Parse_target2.ml` then manual test of https://semgrep.dev/s/EAYw.

PR checklist:


languages/javascript/generic/js_to_generic.ml
```diff
@@ -223,60 +223,65 @@ and literal x : G.literal =
...
+      G.Cast (v3, v2, v1) |> G.e
   | ExprTodo (v1, v2) ->
       let v2 = list expr v2 in
-      G.OtherExpr (v1, v2 |> Common.map (fun e -> G.E e))
-  | L x -> G.L (literal x)
+      G.OtherExpr (v1, v2 |> Common.map (fun e -> G.E e)) |> G.e
+  | ParenExpr (l, e, r) ->
+      let e = expr e in
+      H.set_e_range l r e;
+      e
```
emjin, line 246: Out of curiosity, why did you choose this rather than translating to an AST_generic.ParenExpr?

emjin (approved):

languages/javascript/generic/js_to_generic.ml
nmote: See #7437 for a more detailed explanation, but basically when I tried that with Python it broke a whole bunch of stuff

nmote: We actually should probably delete ParenExpr from the AST and write to `e_range` instead for the one spot it's actually used

nmote:
Also fixes #1917

nmote:
Also fixes the autofix problems with #2051 (after a tweak to the metavariable-regex to allow the parens that are now part of the expression).

PR #7688, opened 2023-04-28
refactor: Remove ParenExpr

nmote:
I believe that this was only added to make sure the ranges were correct for expressions surrounded by parentheses, especially for autofix. However, I ran into correctness issues when I tried to use it for Python in #7437. I assume that this is because there are places that should match and recurse on `ParenExpr` that were not doing so. It would be difficult to make sure that we are handling it correctly throughout Semgrep.

Instead, I set `e_range` to include the locations of the parentheses. This worked well, and I followed up by doing the same for JS in #7686. I think it's time to take this approach for other languages, and simplify the generic AST and analyses that run on it by removing this node.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/cli/tests/e2e/snapshots/test_autofix/test_autofix/rulesautofixocaml_paren_expr.yaml-autofixocaml_paren_expr.ml-dryrun/results.json
+++ b/cli/tests/e2e/snapshots/test_autofix/test_autofix/rulesautofixocaml_paren_expr.yaml-autofixocaml_paren_expr.ml-dryrun/results.json
@@ -59,9 +59,9 @@
       "extra": {
         "engine_kind": "OSS",
         "fingerprint": "0x42",
-        "fix": "a_function_call (wrap ((the_argument)))",
+        "fix": "a_function_call (wrap (the_argument))",
         "fixed_lines": [
-          "let two = a_function_call (wrap ((the_argument)))"
+          "let two = a_function_call (wrap (the_argument))"
         ],
         "is_ignored": false,
         "lines": "let two = a_function_call (the_argument)",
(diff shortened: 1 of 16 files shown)


nmote:
I believe the snapshot changes are because when a metavariable is bound to an expression consisting solely of an identifier, it is given the `Id` variant defined in `Metavariable.ml`, and the rest of the expression, including the `e_range`, is dropped. This is innocuous because the problematic cases where we had to make sure the parentheses were included were for more complex expressions where one parenthesis could end up included and the other could be omitted, e.g. in `1 + (2)`.

aryx (approved):
Very nice!

cli/tests/e2e/snapshots/test_autofix/test_autofix/rulesautofixocaml_paren_expr.yaml-autofixocaml_paren_expr.ml-dryrun/results.json
```diff
@@ -59,26 +59,26 @@
       "extra": {
         "engine_kind": "OSS",
         "fingerprint": "0x42",
-        "fix": "a_function_call (wrap ((the_argument)))",
+        "fix": "a_function_call (wrap (the_argument))",
```
IagoAbal, line 62: I don't expect a refactoring to change behavior.

nmote: Fair -- it's a very minor change in behavior though and I'm not sure how else to categorize this PR. It seems more like a refactor than any of the other options here: https://www.conventionalcommits.org/en/v1.0.0/

PR #7696, opened 2023-05-01
feat(scala): parse if expr `then` better

brandonspark:
## What:
This PR makes it so that we parse the `then` expression in a Scala if-then-else better.

## Why:
Previously, we were unable to parse things such as 
```
if (true)
  x.foo();
  val x = 2
else
  4
```
because our `blockStatSeq` parser is limited to things which do not start with expressions. This was so we had a distinction between blockStatSeq's and just regular `expr`s, which would otherwise conflate, and cause us to incorrectly parse things as a `blockExpr` when it's too general.

In addition, there are difficulties with parsing it as such, because we have things such as
```
if (true)
  3 else 4
```
where the expression is indented, which would ordinarily start an indentation region, but there is no corresponding 
DEDENT at the end of the expression. This messes with things like `inBracesOrIndented` and friends, which expect to parse a single nonterminal, and then parse the DEDENT. Here, however, there's not just a single nonterminal before the DEDENT, because we have the `else 4` in addition to the `3`.

## How:
We added the `indentexExprOrBlackStatSeqUntil` function, which means to parse something which is an indented expression, or a block stat seq. In addition, there is the possibility of a terminating token, which is cnotextual. In the case of a `then` expression, this is `else`. In the case of the terminating token, we don't require a DEDENT, so we can't use a normal parsing combinator.

## Test plan:
Included tests, and parse rate `0.9785559783011464` -> `0.9800450983204928`

PR checklist:

### DIFF
--- a/languages/scala/ast/AST_scala.ml
+++ b/languages/scala/ast/AST_scala.ml
@@ -355,6 +355,7 @@ and block_stat =
   | D of definition
   | I of import
   | Ex of import
+  | Ext of extension
   | E of expr
   (* just at the beginning of top_stat *)
   | Package of package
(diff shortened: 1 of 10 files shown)


aryx (approved):
The extension stuff seems unrelated. Could have been a separate PR I think.

PR #7705, opened 2023-05-02
feat(scala): indent expr case clause blocks

brandonspark:
## What:
This PR makes it so that blocks inside an `exprCaseClause` (so, for instance, under a `catch`) properly emit indentation
regions for blocks.

## Why:
This would otherwise mess up programs like
```
val x = 
  try 
    3
  catch
    case _ =>
      var x = 2

private val y = 3
```
because failing to emit the `INDENT` on the `var` would cause us to be unable to emit the `DEDENT` on the `private`, meaning we wouldn't know where to end the block.

## How:
Changed it so we went through `expr` instead of straight through `block`, which will take care of the necessary indentation-region logic.

To reuse code, I factorized out `caseClause` by the RHS in question. This allows us to use it for both the `block` and `expr` case, for the RHS.

## Test plan:
Included test, and parse rate `0.9811493700926973` -> `0.9814874124719436`

PR checklist:

### DIFF
--- a/languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
+++ b/languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
@@ -2278,13 +2278,13 @@ and guard in_ : guard option =
              Some (ii, stripParens x)
          | _ -> None)
 
-and caseBlock in_ : tok * block =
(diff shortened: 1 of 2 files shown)


aryx (approved):
This is getting seriously complicated.

PR #7721, opened 2023-05-02
feat(scala): soft `export`, fix template intro

brandonspark:
## What:
This PR makes it so that `export` is now properly a soft keyword again, as well as fixing `isDefIntro` to not incorrectly
flag things which start `TmplStat`, as opposed to `TmplDef` (which is correct)

## Why:
Before, when implementing `export`, I took the path of least resistance and just made it a keyword, with the logic that it would be easier to see whether or not this had a significant effect on parse rate later. It turns out, it does, so we need to not do that.

## How:
In addition, our `isDefIntro` function relies on `isTemplateIntro`. Unfortunately, this is only correct if `isTemplateIntro` corresponds to tokens that start template _defs_, but not template _stats_, which includes other things like end markers, extensions, and exports (which have only been added recently). This means that we enter the `Def` case even in cases where we shouldn't, so I fixed it (and gave it a more descriptive name) so we don't conflate the two.

I also just deleted the `export` keyword, and added proper handling for exports elsewhere. The machinery for making sure we don't use `export` as an identifier in places we shouldn't is fairly streamlined, so I just added it to the (newly named) function `isTemplateIntroSoftKeyword`.

Oh yeah, making `export` soft also makes it so we might spuriously enter the `exportClause` nonterminal when `export` starts a `templateStat` or similar, so I also added a token of lookahead, so we only enter `exportClause` when it's followed by an `id`.

## Test plan:
Included test, and parse rate `0.9841003067084511` -> `0.9883969120262044`

Closes PA-2748

PR checklist:

### DIFF
--- a/languages/scala/recursive_descent/Lexer_scala.mll
+++ b/languages/scala/recursive_descent/Lexer_scala.mll
@@ -395,6 +395,16 @@ rule token = parse
         | "package"     -> Kpackage t
         | "import"   -> Kimport t
 
+        (* Scala 3: There are more keywords, but they are "soft" keywords,
+           meaning that we cannot lex them here as keywords!
+
+           Soft keywords behave like keywords only in certain contextual
+           situations, and otherwise behave as normal identifiers.
+
+           We just case on them as `ID_LOWER` in `Parser_scala_recursive_descent`,
+           such as `accept (ID_LOWER ("export", ab)) in_`.
+         *)
+
         | "abstract"       -> Kabstract t
         | "final"       -> Kfinal t
         | "private"       -> Kprivate t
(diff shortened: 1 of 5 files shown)


languages/scala/recursive_descent/Lexer_scala.mll
```diff
@@ -394,7 +394,6 @@ rule token = parse
 
         | "package"     -> Kpackage t
         | "import"   -> Kimport t
```
aryx, line 396: maybe add a comment around here saying there are more keywords in Scala but they
are not lexed here because they are "soft keywords" which are valid only in Scala 3 ...

aryx: Those soft keywords are recognized in Parser_scala_recursive_descent.ml, for example with
accept (ID_LOWER ("export", ab)) in_;

aryx (approved):

PR #7728, opened 2023-05-03
feat - [COD-16] - make cli hit the new endpoint for the reliable fixed status on semgrep app

vivekkhimani:
## Problem
We need to start using a new endpoint to fetch scan configurations from the Semgrep app. This new endpoint provides us with a more reliable fixed status by sending the rules from the previous scan. Previously, if findings were closed due to a rule change, they were marked as fixed, which wasn't ideal. However, we couldn't precisely identify findings closed by a rule change at that time.

## Solution
We've created a new endpoint on the backend that sends us the rules from the previous scan in the configuration. This PR ensures the following:

- [x] Prevent findings opened by the previous scan from blocking the current scan. This is automatically handled since all previous scan rules have ```dev.semgrep.actions = []```, making them non-blocking.
- [x] Filter out findings and rules from the previous scan using annotated metadata.
- [x] Display only the current scan rules and findings in the CLI output.
- [x] Send the findings from the previous scan back to the app through the ```findings_and_ignores``` endpoint. The request body for this endpoint will include a new field called ```prev_scan_findings``` to avoid interference with the current scan findings.
- [x] Update the CLI to use the new endpoint. We can begin doing this even before the app is ready to parse previous scan findings, as it's just another field in the request body that can be handled later when the CLI change is implemented.

## Test Plan
- [x] Extensive manual testing was conducted by running the app and CLI together, ensuring that rule configurations and findings are annotated correctly.
- [x] The automated snapshot test diffs serve as excellent tests on their own. While this was a work in progress and had bugs, the snapshot tests showed random changes and incorrect totals for regular and ignore findings. However, the latest auto-generated diff (https://github.com/returntocorp/semgrep/commit/f4ca34c9da53500802cd0e4271b73519d4ef546c) is in a clean state, except for the ```prev_scan_findings``` field in the findings_and_ignores request body.
- [x] Modify the snapshot tests so that we also have some data for ```prev_scan_findings``` and it's not always empty. Although we have extensively tested this logic at the backend, it'll be nice to have another layer of tests on CLI.
- [x] Add unit tests for the match based id computation when we have rules from previous scan with different rule ids but everything else being the same.

## PR checklist:


cli/src/semgrep/rule_match.py
```diff
@@ -513,6 +554,28 @@ def to_app_finding_format(self, commit_date: str) -> out.Finding:
...
+            return RuleScanSource.previous_scan
+        else:
+            return RuleScanSource.unannotated
+
+    @property
+    def is_prev_scan(self) -> bool:
+        return self.scan_source == RuleScanSource.previous_scan
+
+    @property
+    def annotated_rule_name(self) -> str:
```
chmccreery, line 576: Nice!

cli/src/semgrep/rule.py
```diff
@@ -249,6 +250,24 @@ def product(self) -> RuleProduct:
             else RuleProduct.sast
         )
 
+    @property
+    def scan_source(self) -> RuleScanSource:
```
chmccreery, line 266: This looks identical to the code in `rule_match.py`. Do we need it in both places? Is there a way to use e.g. `rule_match.rule.scan_source` instead of having this property added to both types?

cli/src/semgrep/commands/ci.py
```diff
@@ -415,10 +415,13 @@ def ci(
...
     nonblocking_rules: List[Rule] = []
+    prev_scan_rules: List[Rule] = []
     cai_rules: List[Rule] = []
     for rule in filtered_rules:
         if "r2c-internal-cai" in rule.id:
             cai_rules.append(rule)
+        elif rule.is_prev_scan:
+            prev_scan_rules.append(rule)
         else:
             if rule.is_blocking:
```
chmccreery: nit: I think you can safely flatten this into just one if/elif/elif/else instead of nesting two if/else blocks

cli/src/semgrep/commands/ci.py
```diff
@@ -465,7 +476,7 @@ def ci(
         all_targets=output_extra.all_targets,
         ignore_log=ignore_log,
         profiler=profiler,
-        filtered_rules=filtered_rules,
+        filtered_rules=[*blocking_rules, *nonblocking_rules, *cai_rules],
```
chmccreery: This may be a bug from before this PR, but this function handles which findings are actually displayed to the user, right? If so, I am surprised to see cai_rules here, since those findings should be hidden. If it is right that cai_rules are included here, I find it odd that prev_scan_rules are not, so just want to check on what this is used for.

cli/src/semgrep/app/scans.py
```diff
@@ -300,6 +310,7 @@ def report_findings(
             # send a backup token in case the app is not available
             "token": token,
             "findings": findings,
+            "prev_scan_findings": prev_scan_findings,
             "searched_paths": [str(t) for t in sorted(targets)],
             "renamed_paths": [str(rt) for rt in sorted(renamed_targets)],
             "rule_ids": rule_ids,
```
chmccreery: the pattern here makes me wonder if we should be sending `prev_scan_ids` to the app as well, but I know you've been e2e testing just fine, so we probably don't need to?

chmccreery (commented):
Code looks good! Are there any tests for the new behavior? e.g. that previous-scan findings are not shown in the output but are in the JSON that is sent to the backend? I am only seeing snapshot updates

chmccreery (commented):
Oh, a test that the match_based_id and syntactic_id don't change between two scans where a rule-id gets updated would be another important test, I think!

cli/src/semgrep/app/scans.py
vivekkhimani: yeah i think the the findings include the rule metadata where we have access to the rule_ids (original and annotated) for the previous scan findings. So I figured it's not worth sending more data.

What do you think?

cli/src/semgrep/commands/ci.py
vivekkhimani: done!

cli/src/semgrep/commands/ci.py
vivekkhimani: Good catch there! I traced it down fully and looks like we only directly use ```filtered_rules``` for the scan timing metrics. 

However, I found this one snippet where ```filtered_rules (self.filtered_rules)``` is used to get ```num_rules```, which is part of the conditionals used to generate some user facing output lines (```suggestion_line``` and ```stats_line```). 

Really makes me think that the way to go is to exclude ```cai_rules``` as well and just have blocking and non-blocking rules. But again, I do think we also include (or at least don't intentionally exclude them) ```cai_rules``` in the scan summary part so maybe we should leave it as it is?

I am not sure. 😅 

Screen Shot 2023-05-12 at 6 26 25 PM

cli/src/semgrep/rule.py
vivekkhimani: I think we do because we don't really have a reference to ```Rule``` object within the ```RuleMatch``` and only store the metadata. Also, there are places where we separately need to handle a set of ```Rule``` objects or ```RuleMatch``` objects and that's where this is super handy.

vivekkhimani:
> Code looks good! Are there any tests for the new behavior? e.g. that previous-scan findings are not shown in the output but are in the JSON that is sent to the backend? I am only seeing snapshot updates

Yes, actually I ended up making a [bunch of changes to test_ci.py](https://github.com/returntocorp/semgrep/blob/428e067446ad25ff06ea631f20f3185c4dc8c354/cli/tests/e2e/test_ci.py#L188) which resulted in the snapshot updates, which ideally serve as the test for the new behavior.

In summary, I updated the config we used for running the e2e tests such that it includes the new rule metadata for all the rules and also includes a rule from ```prev-scan```. The idea was that the snapshot updates generated because of the config change should not change anything about the existing regular ```findings``` and ```ignored_findings```, in ```findings_and_ignores.json```, and should add a new finding under ```prev_scan_findings``` field.

Yes, a ton of snapshot changes were generated as ```test_ci.py``` run e2e test for different SCMs and configurations, but essentially, they are all repetitive changes. Following is the summary of snapshot changes:

1. updated metadata in the existing regular and ignores finding (which is expected).
2. new ```prev_scan_findings``` field was created which included a finding from the previous scan which had the same syntactic and match_based_id as the regular and ignored findings, because it ran on the same piece of code.


>Oh, a test that the match_based_id and syntactic_id don't change between two scans where a rule-id gets updated would be another important test, I think!

And yes, this is again being tested as the part of config changes in test_ci.py. If you notice, the ```prev_scan``` rule has a completely different ```rule_id``` (check_id) than a current scan rule but it's essentially running the pattern of ```eqeq-four``` which is a current scan rule. Because of this we got exact same syntactic_id and match_based_id for current and previous scan findings.

I know this is a lot so happy to also quickly sync over a call and discuss this but also very happy to include more rules in the config or just add more tests as and when relevant.

chmccreery (changes requested):
We talked async. @vivekkhimani is going to write some unit tests to supplement the snapshot tests so protect against future breakages!

vivekkhimani:
@mjambon and @aryx these are few of the changes we are making on CLI-side for the reliable fixed status work. I have added details in the PR description and also attaching a [scope doc for this project](https://www.notion.so/r2cdev/Reliable-Fixed-Status-a610030bbe894f6da1869a1e01defef5?pvs=4).

Tl;dr-
As of today, we marked any existing findings disappearing due to a "rule change" or "code change" as "fixed". In theory, we should only be annotating the findings closed due to "code change" as fixed. As fix rate is one of our core metrics, we believe this work has to be done. It includes including the rules from the previous scan alongside the current (or up-to-date) version of the rules, so we can precisely filter out the findings that our closed by the code change.

Please let me know if you have questions with the CLI rewrite work in progress.

aryx:
> @mjambon and @aryx these are few of the changes we are making on CLI-side for the reliable fixed status work. I have added details in the PR description and also attaching a [scope doc for this project](https://www.notion.so/r2cdev/Reliable-Fixed-Status-a610030bbe894f6da1869a1e01defef5?pvs=4).
> 
> Tl;dr- As of today, we marked any existing findings disappearing due to a "rule change" or "code change" as "fixed". In theory, we should only be annotating the findings closed due to "code change" as fixed. As fix rate is one of our core metrics, we believe this work has to be done. It includes including the rules from the previous scan alongside the current (or up-to-date) version of the rules, so we can precisely filter out the findings that our closed by the code change.
> 
> Please let me know if you have questions with the CLI rewrite work in progress.

We didn't start yet porting the semgrep ci part (maybe we'll start in july), so we should be fine.

aryx:
Closing to cleanup the list of opened PRs before release. Feel free to reopen.

cli/src/semgrep/commands/ci.py
```diff
@@ -415,25 +415,35 @@ def ci(
...
 
     # Split up matches into respective categories
     blocking_matches_by_rule: RuleMatchMap = defaultdict(list)
     nonblocking_matches_by_rule: RuleMatchMap = defaultdict(list)
     cai_matches_by_rule: RuleMatchMap = defaultdict(list)
 
+    # Remove the prev scan matches by the rules that are in the current scan
+    # Done before the next loop to avoid interfering with ignore logic
+    removed_prev_scan_matches = {
+        rule: [match for match in matches if (not match.is_prev_scan)]
```
chmccreery: Do we need this? Doesn't the `if rule.is_prev_scan` two lines below this do this already?

cli/src/semgrep/commands/ci.py
chmccreery: I'm pretty sure cai_rules are totally deprecated anyway, so should be safe to remove!

chmccreery (approved):
Looks good! Let's do this! 🎉 Let's make sure someone on OSS team signs off, too, since technically they own this code.

cli/src/semgrep/commands/ci.py
vivekkhimani: You're right haha, we dont't! It was just me being paranoid.

cli/src/semgrep/app/scans.py
```diff
@@ -154,10 +154,9 @@ def fetch_and_init_scan_config(self, meta: Dict[str, Any]) -> None:
         if self.dry_run:
             app_get_config_url = f"{state.env.semgrep_url}/{DEFAULT_SEMGREP_APP_CONFIG_URL}?{self._scan_params}"
         else:
-            app_get_config_url = f"{state.env.semgrep_url}/{DEFAULT_SEMGREP_APP_CONFIG_URL}?{self._scan_params}"
```
brendongo, line 160: It concerns me that there are no params in the new endpoint. How does it determine if it's a forced sca only scan?

chmccreery: This should be sent now in the `start_scan` function, which is now called before this. However, looking in the GitMeta to_dict method, I don't actually see that specific argument anymore, so we should definitely verify that all the pieces the app expects are in fact being sent by the CLI, good call!

chmccreery: Oh, actually, the same metadata dict that is passed into this function is sent in its entirety to the start_scan endpoint, so it should be good 👍 But wow, hardening the typing at this interface would really decrease my blood pressure.

cli/src/semgrep/app/scans.py
```diff
@@ -325,7 +324,9 @@ def report_findings(
             else 0,
             "dependency_parser_errors": [e.to_json() for e in dependency_parser_errors],
             "stats": {
-                "findings": len(new_matches),
+                "findings": len(
+                    [match for match in new_matches if not match.is_prev_scan]
```
brendongo: This seems error prone and requires a higher mental load. Do the transient matches need to be reported in the same place as matches? aka can we partition the transient matches and actual matches earlier in the code and deal with them in a combined data structure as little as possible?

cli/src/semgrep/rule.py
```diff
@@ -261,6 +262,24 @@ def product(self) -> RuleProduct:
...
+            return RuleScanSource.new_version
+        elif src == "new-rule":
+            return RuleScanSource.new_rule
+        elif src == "previous-scan":
+            return RuleScanSource.previous_scan
+        else:
+            return RuleScanSource.unannotated
+
+    @property
+    def is_prev_scan(self) -> bool:
```
brendongo: it isn't immediately clear what is_prev_scan means.

Does this mean it existed in a previous scan and still exists now?

Maybe it's better to call this something like `for_transient_scan` or something more specific than previous scan.

aryx:
cc @hannesm who recently start to port semgrep ci to osemgrep.

cli/src/semgrep/app/scans.py
aryx: Yes, we should start using ATD to specify those metadata. @hannesm recently start to port semgrep ci to OCaml, and those JSON blobs are scary. Where is the code in the app that reads back those JSON blobs?
Can we start defining a Project_metadata.atd in semgrep-interfaces?

cli/src/semgrep/constants.py
```diff
@@ -61,6 +61,14 @@ def _missing_(cls: Type[Enum], value: object) -> Enum:
         raise ValueError(f"invalid rule severity value: {value}")
 
 
+class RuleScanSource(Enum):
```
aryx, line 64: If this is used also on semgrep-app side, we need to start specify this in ATD in semgrep-interfaces.

cli/src/semgrep/rule_match.py
```diff
@@ -513,6 +554,28 @@ def to_app_finding_format(self, commit_date: str) -> out.Finding:
             ret.sca_info = self.extra["sca_info"]
         return ret
 
+    @property
+    def scan_source(self) -> RuleScanSource:
```
aryx, line 558: This should be autogenerated from ATD definitions. This boilerplate is annoying to maintain manually.

aryx (commented):
I would be far more comfortable if we were specifiying formally those JSON before doing those changes. That way it would also be easier to port to osemgrep and it would be also safer.

cli/src/semgrep/app/scans.py
vivekkhimani: @brendongo we did consider completely separating them out from the get-go and reporting ```findings```, ```transient_scan_findings```, ```ignored_findings```, and ```transient_scan_ignored_findings``` separately.

However, that required us to make more changes to the API spec on the app-side. Currently as all our findings are annotated with the relevant scan_source, we handle all the filtering on the app-side so users who upgrade to a specific version of the CLI (and their orgs have a feature flag enabled) can start getting the reliable fixed status.

Also, if for any reason if we end up nuking this project or deciding to not moving forward, the current way of doing things leaves us with a 2-way door to stop sending out metadata from the app and things will function normally without any changes to the API spec.

However, if we add a new field in the API spec we will be left with the ```transient_scan_*``` fields forever even if we don't end up using it anywhere. 

Therefore, we just decided to send everything together so it leaves us with a 2-way door in the future as the app-side changes we are making to this project are kinda sensitive and we easily wanna be able to revert if anything goes wrong as we can't fix CLI once it's rolled out to the users.

cli/src/semgrep/rule.py
vivekkhimani: did this!

cli/src/semgrep/constants.py
vivekkhimani: As discussed, I am adding a \#todo here so we can tackle it as a part of strongly typed interfaces work that we plan to do soon!

cli/src/semgrep/rule_match.py
vivekkhimani: As discussed, I am adding a #todo here so we can tackle it as a part of strongly typed interfaces work that we plan to do soon!

cli/src/semgrep/app/scans.py
brendongo: Because of the way scan_source is annotated, you can still turn off this functionality even with cli side filtering by just not sending any extra annotated rules so I don't find the two way door argument compelling. And the downside of a field we expect but is never used seems minimal.

If we really want to avoid making app side changes I would still make changes on the cli side to have two separate lists for new_matches and transient_new_matches and if you really want to just use a single list in the API, combine them at the end. The error states I'm hoping to avoid are somehow printing, counting, blocking etc. on any new_match that is a is_prev_scan which is highly likely if we keep them all in the same object since now any time that list is used, the programmer needs to be aware that it needs to be filtered.

brendongo: Take for example line 323 above. If app sends the transient policy with associated actions, that might be wrong because we didn't filter out the transient rules

vivekkhimani: Oooh okay yes, I do find the argument for programmer needing to have context around the prev scan stuff while making any changes to the rules or matches list compelling. I do see that being a pitfall for anyone not aware about these changes.

Okay great, I can separate out these findings upstream and keep it that way until we finally end up making an API call where we can merge it.

Thanks, @brendongo!

vivekkhimani: @brendongo for line 323, it shouldn't still break anything as we annotated the "transient scan rules" in the backend to never be blocking. Which means those matches will never be included in line 323. However, yes, I do see that being confusing without devs having enough context.

vivekkhimani: Discussed offline w @brendongo that it's okay to merge this in for now as we are going to port to OCaml soon.

PR #7729, opened 2023-05-03
add(sca): changes to report lockfile parse errors in UI

brendongo:
Before this PR, lockfile parse errors were just being printed to stderr but now they are reported and can be displayed in the webUI. The parse errors are only reported on logged in `semgrep ci` runs.

Also makes the following changes:
- Update what the manifest file of gradle is
- Added better typing to parsing logic

PR checklist:

### DIFF
--- a/cli/src/semdep/parsers/gem.py
+++ b/cli/src/semdep/parsers/gem.py
@@ -71,7 +71,7 @@ gemfile = (
 def parse_gemfile(
     lockfile_path: Path, manifest_path: Optional[Path]
 ) -> List[FoundDependency]:
-    deps_opt = safe_path_parse(lockfile_path, gemfile, ParserName("gemfile.lock"))
+    deps_opt = safe_path_parse(lockfile_path, gemfile, ParserName.gemfile_lock)
     if not deps_opt:
         return []
     deps, manifest_deps = deps_opt
--- a/cli/src/semdep/parsers/gem.py
+++ b/cli/src/semdep/parsers/gem.py
@@ -11,6 +11,7 @@ from semdep.external.parsy import string
 from semdep.external.parsy import success
 from semdep.parsers.util import consume_line
 from semdep.parsers.util import mark_line
+from semdep.parsers.util import ParserName
 from semdep.parsers.util import safe_path_parse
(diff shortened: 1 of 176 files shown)


cli/src/semdep/parsers/gem.py
```diff
@@ -70,7 +71,7 @@
 def parse_gemfile(
     lockfile_path: Path, manifest_path: Optional[Path]
 ) -> List[FoundDependency]:
-    deps_opt = safe_path_parse(lockfile_path, gemfile)
+    deps_opt = safe_path_parse(lockfile_path, gemfile, ParserName("gemfile.lock"))
```
brendongo: For each call to safe_path_parse, added an identifier so that we know what parser we tried to use on the given path

cli/src/semdep/parsers/yarn.py
```diff
@@ -235,7 +236,8 @@ def parse_yarn(
     manifest_deps = get_manifest_deps(manifest_path)
     yarn_version = 1 if lockfile_text.startswith(YARN1_PREFIX) else 2
     parser = yarn1 if yarn_version == 1 else yarn2
-    deps = safe_path_parse(lockfile_path, parser)
+    parser_name = ParserName("yarn1") if yarn_version == 1 else ParserName("yarn2")
```
brendongo: Backend needs to know what parser was used especially for cases like this where same lockfile name could be parsed differently

cli/src/semgrep/app/scans.py
```diff
@@ -319,6 +321,7 @@ def report_findings(
             "exit_code": 1
             if any(match.is_blocking and not match.is_ignored for match in all_matches)
             else 0,
+            "dependency_parser_errors": [e.to_json() for e in dependency_parser_errors],
```
brendongo, line 324: Send as part of `complete` API call

brendongo: This means any parse errors will not be captured if there are Fatal errors running semgrep

brendongo:
I'm gonna isolate the test

mmcqd (approved):

PR #7745, opened 2023-05-04
EXPERIMENT disabling matching cache

IagoAbal:
PR checklist:

### DIFF
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -319,7 +319,7 @@ let m_module_name a b =
  * we could probably refactor this code to not need so many arguments.
  *)
 let m_deep (deep_fun : G.expr Matching_generic.matcher)
-    (first_fun : G.expr -> 't -> tin -> tout) (sub_fun : 't -> G.expr Seq.t)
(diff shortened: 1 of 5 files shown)


aryx:
The code in PR does not look like what the title says, but I was telling to martin yesterday
it could be a good idea to remove the stmts matching cache entirely as it turned out it didn't
really improved things a lot and it adds quite a lot of complexity to Generic_vs_generic stmt matching.

IagoAbal:
> The code in PR does not look like what the title says, but I was telling to martin yesterday
> it could be a good idea to remove the stmts matching cache entirely as it turned out it didn't
> really improved things a lot and it adds quite a lot of complexity to Generic_vs_generic stmt matching.

Sorry I created this PR on top of another PR and that made the diff pretty confusing, I have fixed it now. But it is an ugly experiment to see what happens if we get rid of that stmts maching cache indeed. I wanted to run the benchmarks and so far my conclusion is that indeed, in practice, it doesn't really do much. I started to look into this due to another issue that would require modify this code, and I found the existing code very complex.

aryx:
Go for it. If you don't I'll do it soon.

PR #7751, opened 2023-05-05
refer to langs by string instead of number

tpetr:
jsoo internally represents `Lang.t` by its ordinal number, which means that adding a language requires rolling out a new version of the engine and *all* parsers. This PR updates the code to use the language's name instead, which should be more stable.

Also, added support for parser name overrides (e.g. `python3` is parsed by the `python` parser) since we don't have a perfect 1:1 mapping between langs and parsers.

There's a lot of copypasta in the tests -- I need to factor out some stuff in a future PR.

test plan:
- [x] full test suite
- [x] local playground testing

### DIFF
--- a/js/engine/Main.ml
+++ b/js/engine/Main.ml
@@ -56,12 +56,7 @@ let _ =
               |> Js.string)
                (Js.string pattern)
 
-       method isValidLang lang =
-         match Lang.of_string_opt (Js.to_string lang) with
-         | Some _ -> true
-         | None -> false
-
-       method lookupParserName lang_js_str =
+       method getParserForLang lang_js_str =
          let lang_str = Js.to_string lang_js_str in
          match Hashtbl.find_opt parser_lang_overrides lang_str with
          | Some parser -> Js.some (Js.string parser)
@@ -72,8 +67,6 @@ let _ =
                    (Js.string
                       (String.lowercase_ascii (Lang.to_lowercase_alnum lang)))
              | None -> Js.null)
-
-       method lookupLang lang = Lang.of_string (Js.to_string lang)
 
        method execute (language : Js.js_string Js.t)
            (rule_file : Js.js_string Js.t) (source_file : Js.js_string Js.t)
--- a/js/engine/Main.ml
+++ b/js/engine/Main.ml
@@ -9,7 +9,14 @@ external get_jsoo_mount_point : unit -> 'any list = "get_jsoo_mount_point"
    For example, the Python parser handles all of the Python Langs *)
 let parser_lang_overrides =
   Common.hash_of_list
-    [ ("python2", "python"); ("python3", "python"); ("py", "python") ]
+    [
+      ("python2", "python");
+      ("python3", "python");
+      ("py", "python");
+      ("xml", "html");
+      ("scheme", "lisp");
+      ("clojure", "lisp");
+    ]
 
 let _ =
   Common.jsoo := true;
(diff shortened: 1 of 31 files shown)


js/engine/Main.ml
```diff
@@ -5,6 +5,19 @@ open Js_of_ocaml
    (see companion setter in Semgrep_js_shared.ml) *)
 external get_jsoo_mount_point : unit -> 'any list = "get_jsoo_mount_point"
 
+(* Overrides are required because we don't have a 1:1 correspondence between Lang.t and parsers
+   For example, the Python parser handles all of the Python Langs *)
+let parser_lang_overrides =
```
aryx: Ideally we should use functions in Lang.ml (see _build/default/src/core/Lang.ml for the code after jinja generation).
or add functions there.

js/engine/Main.ml
```diff
@@ -24,20 +37,36 @@ let _ =
 
        method deleteFile filename = Sys.remove (Js.to_string filename)
 
-       method setParsePattern (func : bool -> Lang.t -> string -> 'a) =
-         Parse_pattern.parse_pattern_ref := func
+       method setParsePattern
+           (func : bool -> Js.js_string Js.t -> Js.js_string Js.t -> 'a) =
```
aryx: should define a type jstring = Js.js_string Js.t.
This is too long to read each time.

js/engine/Main.ml
```diff
@@ -24,20 +37,36 @@ let _ =
 
        method deleteFile filename = Sys.remove (Js.to_string filename)
 
-       method setParsePattern (func : bool -> Lang.t -> string -> 'a) =
-         Parse_pattern.parse_pattern_ref := func
+       method setParsePattern
```
aryx: Ideally we would define separately the interface of this object that you export in a
Semgrep_js.mli
so we could explain the type of functions like setParsePattern. For example what is the first
bool parameter of func below? What does the string correspond to?

aryx: It might be a bit tricky because OCaml objects and Jsoo are a bit tricky. Maybe I'll try to refactor a bit
the code next week.

js/engine/Main.ml
```diff
@@ -24,20 +37,36 @@ let _ =
...
+           (func : bool -> Js.js_string Js.t -> Js.js_string Js.t -> 'a) =
+         Parse_pattern.parse_pattern_ref :=
+           fun print_error lang target ->
+             func print_error
+               (lang |> Lang.to_lowercase_alnum |> String.lowercase_ascii
```
aryx: why we need lowercase_ascii in adition to to_lowercase_alnum? Should already be in lowercase accordoing
to the name of the to_lowercase_alnum function.

js/engine/Main.ml
```diff
@@ -24,20 +37,36 @@ let _ =
...
+              |> Js.string)
+               (Js.string pattern)
 
-       method lookupLang (lang : Js.js_string Js.t) : Lang.t =
-         Lang.of_string (Js.to_string lang)
+       method getParserForLang lang_js_str =
+         let lang_str = Js.to_string lang_js_str in
+         match Hashtbl.find_opt parser_lang_overrides lang_str with
+         | Some parser -> Js.some (Js.string parser)
```
aryx: Rename Some parser to Some final_lang_str

js/engine/Main.ml
aryx: Or maybe add a comment saying the interface is defined in index.d.ts in the Engine interface?

aryx: Maybe we could use atdgen and atdts to define this interface, so at least we can typecheck
they agree with each other.

aryx (approved):

js/engine/Main.ml
aryx: Hmm actually we can't use atdgen for this. I wonder if there is a way to make this a bit safer, to ensure what is exported is indeed what is defined in engine.d.ts

aryx: Maybe https://github.com/ocsigen/ts2ocaml/blob/main/docs/js_of_ocaml.md

js/engine/Main.ml
tpetr: Good call — I’ll add what I need there

js/engine/Main.ml
tpetr: I can start with a manually created .mli for now, but ts2ocaml looks interesting

js/engine/Main.ml
tpetr: Forgot to add a callout for this: I noticed that the first character of to_lowercase_alnum is uppercase (caught this in the js tests), and I think it just hasn’t been a problem yet because Lang.of_string is case insensitive. We should probably fix though

tpetr: For now, I'm going to remove the `String.lowercase_ascii`s and normalize the parser name when the FE fetches the js/wasm

js/engine/Main.ml
tpetr: Looking at this with fresh eyes, `setParsePattern` and `setJustParseWithLang` aren't intended to be public -- `js/engine/src/index.js` uses them internally. I can add a note explaining this, though.

js/engine/Main.ml
tpetr: On second thought, the lang -> parser mapping is somewhat arbitrary, so I'm just going to move this logic into the TurboMode class in semgrep-app

js/engine/Main.ml
tpetr: It doesn't look like anyone is really using `to_lowercase_alnum`, so I'll fix the jinja template

aryx: There are a few callers in Test_parsing.ml in semgrep and in Deep_actions.ml in semgrep-pro, but maybe the code can be rewritten.

tpetr: factored out the fix here: https://github.com/returntocorp/semgrep/pull/7761

tpetr:
Merging to get our build green again -- happy to address any additional feedback in another PR

PR #7753, opened 2023-05-05
yaml parsing is baked into the engine

tpetr:
Not worth creating a standalone parser for yaml -- it's already baked into the engine via semgrep.parsing

### DIFF
--- a/js/engine/Main.ml
+++ b/js/engine/Main.ml
@@ -28,17 +28,32 @@ let _ =
        method setParsePattern (func : jbool -> jstring -> jstring -> 'a) =
          Parse_pattern.parse_pattern_ref :=
            fun print_error lang pattern ->
-             func (Js.bool print_error)
-               (Js.string (Lang.to_lowercase_alnum lang))
-               (Js.string pattern)
+             match lang with
+             (* The Yaml and JSON parsers are embedded in the engine because it's a
+                core component needed to parse rules *)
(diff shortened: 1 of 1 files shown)


js/engine/Main.ml
```diff
@@ -25,11 +25,24 @@ let _ =
        method deleteFile filename = Sys.remove (Js.to_string filename)
 
        method setParsePattern (func : bool -> Lang.t -> string -> 'a) =
-         Parse_pattern.parse_pattern_ref := func
+         Parse_pattern.parse_pattern_ref :=
+           fun print_errors lang str ->
+             match lang with
+             | Lang.Yaml -> Yaml_to_generic.any str
```
aryx: maybe add a comment like
(* The Yaml parser is already in engine.js because it is a core component needed to parse the rules.
 * No need to call the external parser.js *)

aryx: BTW the JS/JSON/TS parsers are currently also in engine.js. See src/parsing/Parse_target.ml
for similar reasons to the Yaml parser; because rules can be written in JSON, and
to parse JSON we abuse the JS/TS parser.
Maybe add also a comment like (* coupling: See src/Parsing/Parse_target.ml list of languages builtin engine.js *)

aryx (approved):

js/engine/Main.ml
tpetr: Will do 👍

aryx (approved):

PR #6197, opened 2022-09-29
Disable Bloom filter optimization by default

IagoAbal:
It conflicts with const-prop and sym-prop and appears not to provide a major net benefit nowadays.

To close #4670 we still need to tackle filter-irrelevant-rules interaction with const-prop.

Helps #4670
Helps PR #6179
Closes PA-1927

test plan:
make test # added one test

PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure on any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

Issue #4670, opened 2022-02-14
[python] multiline strings should be semantically equivalent


nbrahms:
**Describe the bug**
In python, multiline strings are semantically equivalent to single-line strings. Semgrep should treat them as such.

**To Reproduce**
https://semgrep.dev/s/q9Yb

**Expected behavior**
Match either single- or multiline-strings with a single pattern

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Use case**
cf. #4669

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-868/[python]-multiline-strings-should-be-semantically-equivalent.

IagoAbal:
Oh, it looked like a const-prop bug but actually it seems that const-prop is working well here. It seems like Semgrep never tries matching `"password"` against `"pass" "word"` and that's why there is no finding.

IagoAbal:
This is caused by the Boom filter optimization, it works if you disable it. cc @aryx @emjin 

I don't know if this would affect perf, but we should probably not be using string literals for the Bloom annotation. At least not with `constant_propagation: true`.

Or, maybe this is a good reason to add `e_svalue` to Generic `expr`. 😄 (This may also affect perf, but it's quite useful for const-prop!)

emjin:
We're probably fine perf-wise without the bloom filter optimization so it's ok if we disable it...but personally I'd rather just add e_svalue to expr if that also helps constant propagation

IagoAbal:
There is also a problem with filter-irrelevant-rules using string literals for filtering, so we would also need to tackle that for this example to work.

IagoAbal:
Oops, this is not yet solved.

PR #7760, opened 2023-05-08
cleanup: Remove Bloom filter optimization

IagoAbal:
Follows: 2428d17ee7c ("Disable Bloom filter optimization by default (#6197)")

It was probably more impactful when it was first introduced, but the engine has evolved and nowadays it does not seem to have a noticeable impact on performance. We disabled it by default on Sep 2022 and this went completely unnoticed.

What triggered this was that the Bloom filter had tricky interactions with const-prop and sym-prop, see #4670, and PA-1920 (or PR #6179).

Ran p/default on 30 repos from stress-test-monorepo with some interesting results, removing -bloom_filter caused an average speedup of 1.15x, and 9 repos in particular particularly benefited with an average speedup of 1.35x (PowerShell, cpython, flask, hadoop, kubernetes, maven, metabase, mypy, react)! Perhaps this could be due to removing `s_strings` from `stmt` in Generic AST?

test plan:
make test

PR checklist:


aryx:
Need to update some script because there's one legitimate CI failure (you can ignore the one for JS).

aryx (approved):
I wonder if we could store somewhere in a comment a trace of our use of bloom filter; just so that in 1 year if someone come with the idea of using a Bloom filter, he can see in the code
it might not be such a good idea.

IagoAbal:
> I wonder if we could store somewhere in a comment a trace of our use of bloom filter; just so that in 1 year if someone come with the idea of using a Bloom filter, he can see in the code
> it might not be such a good idea.

Yeah I agree, but where to put it?

IagoAbal:
> > I wonder if we could store somewhere in a comment a trace of our use of bloom filter; just so that in 1 year if someone come with the idea of using a Bloom filter, he can see in the code
> > it might not be such a good idea.
> 
> Yeah I agree, but where to put it?

Added a note to the `stmt` type in AST_generic.

PR #7761, opened 2023-05-08
fix to_lowercase_alnum, update current users to to_capitalized_alnum

tpetr:
`Lang.to_lowercase_alnum` does not do what it claims 😞 

This PR:
- updates `to_lowercase_alnum` to do the correct thing
- adds `to_capitalized_alnum` with the original functionality
- migrates current usage from `to_lowercase_alnum` to `to_capitalized_alnum` to ensure nothing changes for folks

I'll make a companion PR for Semgprep Pro once this is merged.

test plan:
- [ ] `make test`

### DIFF
--- a/src/core/Lang.ml.j2
+++ b/src/core/Lang.ml.j2
@@ -122,7 +122,7 @@ let to_lowercase_alnum = function
 {% for item in langs %}  | {{ item.id }} -> "{{ item.id | lower }}"
 {% endfor %}
 
-(* must match [a-z][a-z0-9]* *)
+(* must match [A-Z][a-z0-9]* *)
 let to_capitalized_alnum = function
 {% for item in langs %}  | {{ item.id }} -> "{{ item.id }}"
 {% endfor %}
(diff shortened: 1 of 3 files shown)


src/core/Lang.ml.j2
```diff
@@ -119,6 +119,11 @@ let to_string = function
 
 (* must match [a-z][a-z0-9]* *)
 let to_lowercase_alnum = function
+{% for item in langs %}  | {{ item.id }} -> "{{ item.id | lower }}"
+{% endfor %}
+
+(* must match [a-z][a-z0-9]* *)
```
aryx: you mean [A-Z][...]?

aryx (approved):

src/core/Lang.ml.j2
tpetr: Yep! Copypasta fail

tpetr:
only failure is the pre-existing js build issue -- merging

Issue #6318, opened 2022-10-14
Can't use autofix to remove some code


aryx:
with this rule:
```
 - id: python-typing-osemgrep
   pattern: from typing import $X
   fix: ""
   languages: [ python ]
   message: found one
   severity: ERROR
```
I can't use it on some codebase to remove all the typing imports.
`semgrep --config ~/refactoring.yaml . --autofix` does not apply any fixes.

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-2000/cant-autofix-to-remove-some-code. _Note_: this link is for r2c use only and is not accessible publicly.

alanisaac:
I'm also looking for the same feature.  I think the thread in this issue is related: https://github.com/returntocorp/semgrep/issues/5674

joshlory:
Without this it's hard to use Semgrep for code cleanup. @nmote I think the line you're looking for is here?

https://github.com/returntocorp/semgrep/blob/33d49d5d351601598c882eb3fa9f758c441ea101/cli/src/semgrep/core_output.py#L168

> I don't think AST-based autofix would affect it. It doesn't affect fix-regex at all, since fix-regex is by definition text-based. I bet this is due to a Python truthiness test somewhere like `if replacement:`, since the empty string is falsy. I bet that finding that spot and replacing `if replacement:` with `if replacement is not None:` would address this issue.

https://github.com/returntocorp/semgrep/issues/5674#issuecomment-1307473104

nmote:
@joshlory that looks pretty likely. Are you interested in putting up a PR?

nmote:
Fixed by #7772

PR #7757, opened 2023-05-08
[Cairo 1.0] Add transform Tree-Sitter -> Generic AST

Frostweeds:
PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [ ] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #7772, opened 2023-05-09
cli: Allow empty string in the fix field #6318

ihji:
PR checklist:


aryx (commented):
Nice work!

libs/commons/Textedit.ml
```diff
@@ -53,9 +53,25 @@ let remove_overlapping_edits edits =
...
   let after = Str.string_after text end_ in
-  before ^ replacement_text ^ after
+  let starts_with_newline =
+    String.length before == 0
+    || String.get before (String.length before - 1) == '\n'
+  in
+  let ends_with_newline =
+    String.length after == 0 || String.get after 0 == '\n'
+  in
+  let is_empty_replacement = String.length replacement_text == 0 in
```
aryx: This code looks a bit complicated. I wonder if we could simplify it.

nmote: I agree. I also think that any special business logic should happen outside of the `Textedit` module. A `Textedit.t` is a very simple data structure, and applying it to a file should also be a simple, straightforward operation. If we need to modify the `Textedit.t` to remove newlines, let's do it at the callsite.

nmote:
Other than the change in `Textedit.ml`, this looks great! Please make sure to go over the checkboxes in the PR description, too.

libs/commons/Textedit.ml
ihji: @nmote It seems difficult to apply the same logic outside of the `apply_edit_to_text` function since we would lose the context of the replacement. Although not impossible, it would require more complex logic to keep track of every change introduced by `TextEdit` module. Instead, we could to create a separate function to remove newlines from empty replacements, which made the code easier to read.

ihji:
@nmote Do you think this change requires a change log entry? It's somewhat user facing but probably too minor.

libs/commons/Textedit.ml
nmote: I've taken a closer look at how things work today. I still would like to avoid complicating the `apply_edit_to_text` function. Instead, I think that there should be a pass that looks for newlines and modifies the list of `Textedit.t`s to apply, so that they will also replace a newline character as appropriate.

There is already some business logic in `apply_edits`, so I think it would be fine to put that there. And we already have the file contents there, so we wouldn't need to do anything extra to find the additional context.

Also, please make sure to handle CRLF line endings as well.

nmote:
Yes, this fixes a bug that has been reported by multiple users. I think it merits a changelog entry.

changelog.d/gh-7772.fixed
```diff
@@ -0,0 +1 @@
+Enable automatic removal of matched codes by allowing an empty string in the fix field.
```
nmote: Usually we name this after the GitHub issue that we're fixing, not the pull request that fixed it

libs/commons/Textedit.ml
```diff
@@ -53,7 +53,7 @@ let remove_overlapping_edits edits =
   f [] [] edits
 
 let apply_edit_to_text text { start; end_; replacement_text; _ } =
-  let before = Str.first_chars text start in
+  let before = Str.string_before text start in
```
nmote, line 56: Could you either explain this change or remove it?

libs/commons/Textedit.ml
```diff
@@ -95,12 +95,28 @@ let partition_edits_by_file edits =
...
+  else edit
+
 let apply_edits ~dryrun edits =
   let edits_by_file = partition_edits_by_file edits in
   let all_conflicting_edits = ref [] in
   Hashtbl.iter
     (fun file file_edits ->
       let file_text = Common.read_file file in
+      let file_edits =
+        List.map (remove_newline_for_empty_replacement file_text) file_edits
```
nmote: Not sure why our Semgrep rule didn't catch this, but we use `Common.map` because it's tail recursive on long lists so as to avoid stack overflows.

nmote (changes requested):
Nice, this is looking good. I left a few minor comments inline. I'd like to take another quick look after you address those.

libs/commons/Textedit.ml
ihji: Fixed. Thanks for the comment. It's interesting to note that `List.map` isn't tail-recursive 😮

libs/commons/Textedit.ml
ihji: Both of these functions achieve the same result. However, I rewrote the function since it is more consistent with the usage of `string_after` on the line below.

changelog.d/gh-7772.fixed
ihji: Fixed. I believed that the ID was a pull request number because I searched for an existing file `gh-7757` as an example, and there is a PR 7757. Thank you for the clarification.

nmote (approved):
Nice, looks good!

PR #7774, opened 2023-05-10
osemgrep: uniquify rules before applying them (fixes test_deduplication)

hannesm:
To uniquify the rules, derive equality for the `type rule`.

test plan:
  make osempass

Now, the test_check::test_deduplication passes

PR checklist:

### DIFF
--- a/src/osemgrep/cli_scan/Scan_subcommand.ml
+++ b/src/osemgrep/cli_scan/Scan_subcommand.ml
@@ -199,6 +199,11 @@ let run (conf : Scan_CLI.conf) : Exit_code.t =
       let rules, errors =
         Rule_fetching.partition_rules_and_errors rules_and_origins
       in
+      let rules =
+        Common.uniq_by
+          (fun r1 r2 -> String.equal (fst r1.Rule.id) (fst r2.Rule.id))
+          rules
+      in
       let filtered_rules =
         Rule_filtering.filter_rules conf.rule_filtering_conf rules
       in
--- a/src/osemgrep/cli_scan/Scan_subcommand.ml
+++ b/src/osemgrep/cli_scan/Scan_subcommand.ml
@@ -199,6 +199,7 @@ let run (conf : Scan_CLI.conf) : Exit_code.t =
       let rules, errors =
         Rule_fetching.partition_rules_and_errors rules_and_origins
       in
+      (* TODO: we should probably warn the user about rules using the same id *)
       let rules =
         Common.uniq_by
           (fun r1 r2 -> String.equal (fst r1.Rule.id) (fst r2.Rule.id))
--- a/src/parsing/yaml_to_generic.ml
+++ b/src/parsing/yaml_to_generic.ml
@@ -221,49 +221,51 @@ let scalar (_tag, pos, value, style) env : G.expr * E.pos =
   else
     let token = mk_tok ~style pos value env in
     let expr =
-      (match value with
-      | "__sgrep_ellipses__" -> G.Ellipsis (Tok.fake_tok token "...")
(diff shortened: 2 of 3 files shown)


aryx (commented):
I'm not sure you need to use deriving eq. In fact I don't even know how it works because the patterns in a rule now can be closure and generally equality fails on such closures.
I think you only need to do uniq_by the rule id. The rule id should be enough to dedup
2 rules.

hannesm:
ok, I was following what the python code does (in rule.py, it defines rule_without_metadata and uses that for uniquifying. I'll revise to revert the deriving eq and use the rule_id instead.

hannesm:
@aryx applied your suggestion, leading to a much smaller patch, and force-pushed

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -199,6 +199,11 @@ let run (conf : Scan_CLI.conf) : Exit_code.t =
       let rules, errors =
         Rule_fetching.partition_rules_and_errors rules_and_origins
       in
+      let rules =
```
aryx, line 203: Maybe add a comment like (* TODO: we should probably warn the user about rules using the same id *)

aryx, line 203: Maybe add a comment like (* TODO: we should probably warn the user about rules using the same id *)

aryx (approved):

src/osemgrep/cli_scan/Scan_subcommand.ml
hannesm: done in [e5962c4](https://github.com/returntocorp/semgrep/pull/7774/commits/e5962c4a46ef2195ca60a2fdbebfb7461875323b)

aryx:
Weird errors about "4" vs 4.
Did you get this error when you used the deriving eq approach?
Maybe the skipped duplicate is not the same because the order differ ... still the 2 duplicates have this version: "4"
so I don't see why the test is failing now.

hannesm:
Indeed, that is an issue. The underlying source is:
```OCaml
      | _ -> (
          try G.L (G.Float (Some (float_of_string value), token)) with
          | _ -> G.L (G.String (fb (value, token)))))
```

in `src/parsing/yaml_to_generic.ml` (`let scalar`) -- so the YAML in the rule input contains `version: "4"`, which can be nicely parsed by `float_of_string`.

In [f4a442e](https://github.com/returntocorp/semgrep/pull/7774/commits/f4a442e4207f46bee84347b6c3af76564119963e) I propose to use `String` when `quoted` is true. Please have a careful check that this looks fine to you (I have only glanced the YAML spec, and am not sure whether this treatment is good). I hope the CI / tests will uncover potential issues.

src/parsing/yaml_to_generic.ml
```diff
@@ -221,49 +221,51 @@ let scalar (_tag, pos, value, style) env : G.expr * E.pos =
...
-      | "-.inf" -> G.L (G.Float (Some neg_infinity, token))
-      | ".inf" -> G.L (G.Float (Some neg_infinity, token))
-      | ".nan"
-      | ".NaN"
-      | ".NAN" ->
-          G.L (G.Float (Some nan, token))
-      | _ -> (
-          try G.L (G.Float (Some (float_of_string value), token)) with
-          | _ -> G.L (G.String (fb (value, token)))))
+      (if quoted then G.L (G.String (fb (value, token)))
```
aryx, line 224: @emjin any comment on this code? Looks good to you?

aryx (approved):

aryx (approved):

PR #7777, opened 2023-05-10
semgrep.js: more missing language methods + tests + update types

tpetr:
- need a way to clear the `missingLanguages` set so that we're not constantly attempting to load unloadable parsers in the frontend
- forgot to update types file in previous PR (will automate generation of this file in the near future)
- update tests

test plan:
- test suite
- tested against this branch with a local playground instance

### DIFF
--- a/js/engine/src/index.d.ts
+++ b/js/engine/src/index.d.ts
@@ -16,5 +16,8 @@ export interface Engine {
   ) => string;
   writeFile: (filename: string, content: string) => void;
   deleteFile: (filename: string) => void;
+  isMissingLanguages: () => boolean;
+  getMissingLanguages: () => string[];
+  clearMissingLanguages: () => void;
 }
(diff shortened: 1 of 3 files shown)


zyannes (approved):

js/engine/src/index.d.ts
```diff
@@ -16,5 +16,8 @@ export interface Engine {
   ) => string;
   writeFile: (filename: string, content: string) => void;
   deleteFile: (filename: string) => void;
+  isMissingLanguages: () => boolean;
```
aryx, line 19: Would be good to document a bit those methods. One of the thing I love the most about OCaml is the presence of those .mli signature files that let you document clearly the interface of a module. Then user of the module don't eve need to open the implementation file; just reading the signature and the comments in it (and types) should be clear enough.

PR #7778, opened 2023-05-10
Converge semgrep metrics

dinosaure:
PR checklist:

This PR wants to converge `semgrep_metrics.atd` between the Python code and the OCaml code. Currently, this PR requires returntocorp/semgrep-interfaces#71. I tried to find the best way to fit into the OCaml and the Python code without big changes. However (/cc @mjambon), it seems we can not express with `atdpy` that some values have defaults on the Python side but must be present on the JSON side.

Indeed, one test fails where some unset values are not output by the Python JSON encoder. It will be nice to:
1) be able to express that via `atdpy`
2) or update the test (and what we expects about the JSON output)

I marked this PR as a draft due to the pending PR about `semgrep-interfaces`.


.gitmodules
```diff
@@ -100,3 +100,5 @@
 [submodule "languages/cairo/tree-sitter/semgrep-cairo"]
 	path = languages/cairo/tree-sitter/semgrep-cairo
 	url = https://github.com/returntocorp/semgrep-cairo
+[submodule "cli/src/semgrep/semgrep_interfaces"]
```
aryx: This will need to be removed

aryx (commented):
Very nice! I like it.

dinosaure:
Currently, tests still fails because one test `test_metrics_payload[pro_flag0]` expects to **not** have the `maxMemoryBytes` field and `test_metrics_payload[pro_flag1]` expects to have this field (even if it's not set - with the `null` value). So we must do a choice on that and transform (as I did on this PR) the `maxMemoryBytes` field as an optional field (and upgrade the `test_metrics_payload[pro_flag1]` accordingly) or define it as a required field (and upgrade the `test_metrics_payload[pro_flag0]` accordingly).

aryx:
I wonder how the original python code was working to allow both situations ... weird.

src/osemgrep/core/Metrics_.ml
```diff
@@ -71,6 +71,14 @@ let default_payload =
         profilingTimes = [];
         maxMemoryBytes = None;
       };
+    extension =
```
aryx: BTW could you try to rename Metrics_.ml to just Metrics.ml
or maybe it goes in conflict with some library we use that already have a Metrics.ml, but in any case
it's worth adding  a comment in the .ml to say why it's not called more simply just Metrics.ml

aryx (changes requested):

src/osemgrep/core/Metrics_.ml
dinosaure: I think @hannesm renamed it due to the usage of the `metrics` library - a clash occurs between the library and this module. I can add a note indeed.

aryx:
remove the Draft thing when you think this is ready for review @dinosaure

aryx (approved):
This is great

aryx:
I think you need to remove the old Semgrep_metrics.atd now that we use the one in semgrep-interfaces/ no?
I don't see the file deleted in the PR.

aryx (approved):

PR #7782, opened 2023-05-11
osemgrep: rule validation

hannesm:
Now that #7774 is merged (and quoted strings aren't treat as Float anymore), we can leverage the `read_string_wrap` to disallow Float.

The other check in place is that `paths` may only contain `include` and `exclude` keys.

Test plan:
 - make osempass

The test test_rule_parser.py::test_rule_parser__failure now passes (and has been marked with osempass).

PR checklist:

### DIFF
--- a/src/parsing/Parse_rule.ml
+++ b/src/parsing/Parse_rule.ml
@@ -480,8 +480,10 @@ let parse_paths env key value =
     ( take_opt paths_dict env parse_string_list "include",
       take_opt paths_dict env parse_string_list "exclude" )
   in
+  (* alt: we could use report_unparsed_fields(), but better to raise an error for now
+     to be compatible with pysemgrep *)
   if Hashtbl.length paths_dict.h > 0 then
-    yaml_error paths_dict.first_tok
+    error_at_key env key
       "Additional properties are not allowed (only 'include' and 'exclude' are \
        supported)";
   { R.include_ = optlist_to_list inc_opt; exclude = optlist_to_list exc_opt }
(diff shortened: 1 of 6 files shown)


src/parsing/Parse_rule.ml
```diff
@@ -483,6 +480,10 @@ let parse_paths env key value =
     ( take_opt paths_dict env parse_string_list "include",
       take_opt paths_dict env parse_string_list "exclude" )
   in
+  if Hashtbl.length paths_dict.h > 0 then
```
aryx, line 485: maybe add a comment like (* alt: we could use report_unparsed_fields(), but better to raise an error for now
to be compatible with pysemgrep *)

src/parsing/Parse_rule.ml
```diff
@@ -483,6 +480,10 @@ let parse_paths env key value =
...
+    yaml_error paths_dict.first_tok
```
aryx: It's better to use error_at_key because this generates an exn that can be captured in parse_generic_ast to just skip the rule with the wrong syntax (instead of aborting the whole file).

aryx (changes requested):
Please use error_at_key

aryx (approved):

aryx:
CI failure on                                                                      91   ../../../tests/rules/pattern_int.yaml....

aryx:
in your test plan it's usually good to run 'make osempass' from cli/
and also make core-test from the toplevel

aryx (approved):

aryx (approved):

PR #7784, opened 2023-05-11
osemgrep: do not print the rule id if interactive

hannesm:
test plan:
 - osemgrep -e ': Common.filename' -l ocaml src/

compare output to pysemgrep

PR checklist:

### DIFF
--- a/src/osemgrep/reporting/Matches_report.ml
+++ b/src/osemgrep/reporting/Matches_report.ml
@@ -213,6 +213,8 @@ let pp_text_outputs ~max_chars_per_line ~max_lines_per_finding ~color_output ppf
       msg
     in
     let print =
+      cur.check_id <> "-"
+      &&
       match last_message with
       | None -> true
       | Some m -> m <> cur.extra.message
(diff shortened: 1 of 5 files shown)


src/osemgrep/reporting/Matches_report.ml
```diff
@@ -213,6 +213,8 @@ let pp_text_outputs ~max_chars_per_line ~max_lines_per_finding ~color_output ppf
       msg
     in
     let print =
+      cur.check_id <> "-"
```
aryx: maybe you can use Constants.cli_rule_id like we do in Rule_fetching.rules_from_rules_source() for the Pattern case.

aryx (changes requested):

src/osemgrep/reporting/Matches_report.ml
aryx: and maybe we should rename this constant ... it was taken from the pysemgrep code I think.
maybe rule_id_for_dash_e

hannesm: sure, done in 6ea96868b

aryx (approved):

PR #7786, opened 2023-05-12
osemgrep: Find_target: support include/exclude patterns better

hannesm:
Now, rule `paths` are matched the same as in pysemgrep.

There are various tests that pass now

test plan:
- make osempass

PR checklist:

### DIFF
--- a/src/targeting/Find_target.ml
+++ b/src/targeting/Find_target.ml
@@ -428,10 +428,16 @@ let get_targets conf scanning_roots =
 let create_cache () = Hashtbl.create 1000
 
 let match_glob_pattern ~pat path =
+  (* if pattern uses a * and no /, use filename  *)
   let regex = Glob.Parse.parse_string pat in
+  let path =
+    if String.contains pat '*' && not (String.contains pat '/') then
+      Fpath.filename path
+    else Fpath.to_string path
+  in
   Glob.Match.run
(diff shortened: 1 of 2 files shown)


src/targeting/Find_target.ml
```diff
@@ -428,10 +428,16 @@ let get_targets conf scanning_roots =
 let create_cache () = Hashtbl.create 1000
 
 let match_glob_pattern ~pat path =
+  (* if pattern uses a * and no /, use filename  *)
   let regex = Glob.Parse.parse_string pat in
+  let path =
```
aryx, line 433: I don't think this is enough. For example in our semgrep.jsonnet file we have for the exclude:
```
        "yaml_to_generic.ml",
	#skipping more stuff
        "parsing/tree_sitter/*",
        "metachecking/*",
```
and this should exclude any parsing/tree_sitter/ code, wherever they are; they don't have to be at the toplevel of the project.
I think we probably need to use something like Parse_gitignore.parse_pattern and also find a way to transform the path as a ppath so we can reuse some of the tricky filtering logic in Gitignore_filter

aryx: But @mjambon would know better. He wrote the Glob and Gitignore libraries.

aryx:
I'll take over. There are many things that are currently wrong in osemgrep regarding file targeting. We need to refactor
first.

PR #7789, opened 2023-05-12
tainting: Do not use `Metavariable.equal_mvalue` directly

### LINKED ISSUE #7694: Spurious error appeared suddenly in 1.20.0
**Describe the bug**

Having just upgrade to v1.20.0 and having changed no files in the repo, these errors just started appearing.
These errors were NOT present with v1.19.0.
Note that line 1 of each of the 2 noted files was a standard JavaScript import statement (`import PropTypes from 'prop-types'`)

```
% semgrep -f p/cwe-top-25                     
      
. . . standard info left out here...
                                                                                                                        
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╸ 100% 0:00:01                                                                                                                        
[ERROR] Fatal error at line src/components/system-view.jsx:1:
 (Failure "Call AST_utils.with_xxx_equal to avoid this error.")
====[ BEGIN error trace ]====
(Failure "Call AST_utils.with_xxx_equal to avoid this error.")
=====[ END error trace ]=====

[ERROR] Fatal error at line src/components/policy-authoring/aux-container.jsx:1:
 (Failure "Call AST_utils.with_xxx_equal to avoid this error.")
====[ BEGIN error trace ]====
(Failure "Call AST_utils.with_xxx_equal to avoid this error.")
=====[ END error trace ]=====
```

The same issue appears when running it just on either single file in question.
Note the mention of the internal semgrep errors was present with previous semgrep versions; it is just the fatal error that is new with 1.20.0.
```
% semgrep -f p/cwe-top-25 src/components/system-view.jsx
               
               
┌─────────────┐
│ Scan Status │
└─────────────┘
  Scanning 1 file tracked by git with 207 Code rules:
                                                                                                                        
  <multilang>       3       2          Community     207                                                                
  js               41       1                                                                                           
                                                                                                                        
  ━━━━━━━━━━━━━━━━━━━━━━━━━━╸━━━━━━━━━━━━━  67% 0:00:00                                                                                                                        
[ERROR] Fatal error at line src/components/system-view.jsx:1:
 (Failure "Call AST_utils.with_xxx_equal to avoid this error.")
====[ BEGIN error trace ]====
(Failure "Call AST_utils.with_xxx_equal to avoid this error.")
=====[ END error trace ]=====
              
┌──────────────┐
│ Scan Summary │
└──────────────┘
  Partially scanned: 1 files only partially analyzed due to parsing or internal Semgrep errors
```

**To Reproduce**
Apparently just need to supply a file that causes Semgrep errors.

**Expected behavior**
No error.

**Environment**
Both local binary and docker versions do the same thing.

IagoAbal:
If the `mvalue`s being compared happen to contain `stmt`s, then this will raise a fatal error:

    Failure "Call AST_utils.with_xxx_equal to avoid this error."

We already had a use of `Metavariable.equal_mvalue` previously but this did not cause much trouble because it was only used if option `taint_unify_mvars` was set. But then we started using `equal_mvalue` for essentially all taint-labels rules at the same time as SR was adding more taint-labels rules... making this fatal error more likely.

Replaced `equal_mvalue` with `Matching_generic.equal_ast_bound_code` given that it's actually what we use when unifying metavariables.

Also added a Semgrep rule to prevent this in the future.

Fixes: 8588460 ("feat(taint): multiple source traces (#7291)")

Closes #7694

test plan:
make test # test added

PR checklist:

### DIFF
--- a/changelog.d/gh-7694.fixed
+++ b/changelog.d/gh-7694.fixed
@@ -1,0 +1,2 @@
+taint-mode: Fixed bug in taint labels that was causing some fatal errors:
+> Failure "Call AST_utils.with_xxx_equal to avoid this error."
--- a/semgrep.yml
+++ b/semgrep.yml
@@ -365,3 +365,19 @@ rules:
     fix: Common.tl_exn "unexpected empty list"
     languages: [ocaml]
     severity: WARNING
+
+  - id: no-metavariable-equal-xyz
+    patterns:
+      - pattern: Metavariable.$F
+      - pattern-not: Metavariable.equal_mvar
+      - metavariable-regex:
+          metavariable: $F
+          regex: equal_.*
+    message: >-
+      Don't use 'Metavariable.$F' directly, see 'AST_utils'. Instead use
+      either 'Metavariable.Structural.$F' or 'Metavariable.Referential.$F'.
+      You can also consider replacing `equal_mvalue` with
+      'Matching_generic.equal_ast_bound_code'.
+    fix: Metavariable.Structural.$F
+    languages: [ocaml]
+    severity: WARNING
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -367,7 +367,7 @@ let is_func_sink_with_focus taint_sink =
       true
   | __else__ -> false
 
-let unify_mvars_sets mvars1 mvars2 =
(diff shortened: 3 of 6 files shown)


nmote (approved):
Good find

emjin (approved):
Looks good, but I also wonder if we could override `Metavariable.equal_mvar` to choose one so that the failure can't happen

PR #7792, opened 2023-05-14
fix: sym-prop: Make sure we perform cycle checks

IagoAbal:
Fixes stack overflows that started to happen in 1.19.0.

In eeaecf7a9b4 we extended flow-insensitive sym-prop to work for global variables that are assigned just once in the file, but we forgot about cycle checks... and we allowed assignments like `p = p` to be propagated introducing a cycle in the `id_svalue` relation.

Part of the problem is also that sometimes we wrongly infer that something is assigned just once, for example,  we were not considering that in `for p in FOO` the `p` is being assigned, so `p` is assigned more than once to start with. So now we forbid propagating any assignment where the defined variable occurs in the defining expression, because that is a hint that the variable may be assigned more than once.

Also added some safeguard to (hopefully) prevent this from happening again.

Fixes: eeaecf7a9b4 ("const-prop: Private static variables set once in a static block (#7527)")
Closes PA-2740

test plan:
make test # test added

PR checklist:

### DIFF
--- a/src/analyzing/Constant_propagation.ml
+++ b/src/analyzing/Constant_propagation.ml
@@ -117,6 +117,34 @@ let rec lvars_in_lhs expr =
   | Container ((Tuple | Array), (_, es, _)) -> List.concat_map lvars_in_lhs es
   | __else__ -> []
 
+let no_cycles_in_sym_prop sid exp =
+  let for_all_sid : (G.sid -> bool) -> G.any -> bool =
+    (* Check that all sid's satisfy a given condition. We use refs so that
+     * we can have a single visitor for all calls, given that the old
+     * `mk_visitor` was kind of expensive, and constructing a visitor object
+     * may be as well. *)
+    let ff = ref (fun _ -> true) in
+    let ok = ref true in
+    let vout =
+      object
+        inherit [_] G.iter
+
+        method! visit_resolved_name _env (_, sid) =
+          ok := !ok && !ff sid;
+          if not !ok then raise Exit
+      end
+    in
+    fun f ast ->
+      ff := f;
+      ok := true;
+      try
+        vout#visit_any () ast;
+        !ok
+      with
+      | Exit -> false
+  in
+  for_all_sid (fun sid' -> sid' <> sid) (E exp)
+
 (*****************************************************************************)
 (* Environment Helpers *)
 (*****************************************************************************)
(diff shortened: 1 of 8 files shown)


src/analyzing/Constant_propagation.ml
```diff
@@ -117,6 +117,34 @@ let rec lvars_in_lhs expr =
   | Container ((Tuple | Array), (_, es, _)) -> List.concat_map lvars_in_lhs es
   | __else__ -> []
 
+let no_cycles_in_sym_prop sid exp =
+  let for_all_sid : (G.sid -> bool) -> G.any -> bool =
+    (* Check that all sid's satisfy a given condition. We use refs so that
+     * we can have a single visitor for all calls, given that the old
+     * `mk_visitor` was kind of expensive, and constructing a visitor object
+     * may be as well. *)
+    let ff = ref (fun _ -> true) in
```
emjin: Nit: maybe better to make this `(fun _ -> false_)`? So that things break if someone doesn't override `f`

src/analyzing/Dataflow_svalue.ml
```diff
@@ -400,6 +395,59 @@ let eval_or_sym_prop env exp =
   | G.NotCst -> sym_prop exp.eorig
   | c -> c
 
+let no_cycles_in_svalue (id_info : G.id_info) svalue =
+  let for_all_id_info : (G.id_info -> bool) -> G.any -> bool =
+    (* Check that all id_info's satisfy a given condition. We use refs so that
+     * we can have a single visitor for all calls, given that the old
+     * `mk_visitor` was pretty expensive, and constructing a visitor object may
+     * be as well. *)
+    let ff = ref (fun _ -> true) in
```
emjin: Same comment as the other one

emjin (approved):
This all looks reasonable to me. My biggest qualm is that I don't leave the PR feeling convinced that we don't have other potential cycles or that we'll avoid introducing cycles in the future, since it feels like it would be easy to make the same mistake somewhere else.

IagoAbal:
> This all looks reasonable to me. My biggest qualm is that I don't leave the PR feeling convinced that we don't have other potential cycles or that we'll avoid introducing cycles in the future, since it feels like it would be easy to make the same mistake somewhere else.

It's not a silver bullet but in principle forbidding direct assignments to `id_svalue` would have prevented this one. Basically every time we encounter a cycle in sym-prop we add a new safeguard to prevent problems in the future. Eventually that should make it pretty safe.

src/analyzing/Constant_propagation.ml
IagoAbal: Good point, let's put `assert false` then to be extra sure.

PR #7800, opened 2023-05-15
feat(kotlin): allow string literal mvars in patterns

brandonspark:
## What:
This PR adds [literal metavariables](https://semgrep.dev/docs/writing-rules/pattern-syntax/#literal-metavariables) to Kotlin patterns.

## Why:
We sometimes want to be able to analyze the contents of a string in Kotlin. In particular, SR would like this functionality.

## How:
I specifically made any string which looks like `"$FOO"` into a literal metavariable, by permitting the Generic translation into `String`.

This will not take away our ability to match Kotlin strings that contain a single interpolated identifier, as that can be done via the pattern `"${FOO}`, which is understood to be semantically the same as the Kotlin string `$FOO`.

Closes PA-2755

PR checklist:


tests/patterns/kotlin/single_interpolated_ident.sgrep
```diff
@@ -0,0 +1 @@
+"${FOO}
```
emjin: I'm surprised this pattern works without a close quote

brandonspark: fck

brandonspark: i guess it was a tree-sitter partial parse. good catch though!

tests/patterns/kotlin/string_metavar.kt
```diff
@@ -0,0 +1,12 @@
+
+// not this because it has no part of it thats not the interpolation
+val x = "$X"
```
emjin, line 3: Being unfamiliar with Kotlin, I was confused by these tests. Didn't realize that `$var` was the syntax for interpolation. It would help me if the test was instead `var` and `$var` or similar instead of reusing `$X`

tests/patterns/kotlin/single_interpolated_ident.kt
```diff
@@ -0,0 +1,8 @@
+
+val x = "whatever"
+
+// ERROR: match
+val x = "$FOO"
```
emjin: Same comment as the `string_metavar` case. Also, I assume `$FOO` here isn't valid Kotlin because there's no variable `FOO` in scope? Not that our tests all need to be compilable but that's another place where it would help me understand what was going on

emjin: And you can just apply this comment to `not_string_metavar` as well

languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
```diff
@@ -1889,19 +1896,32 @@ and statements (env : env) ((v1, v2, v3) : CST.statements) =
...
-  in
-  let v3 = token env v3 in
-  G.interpolated (v1, v2, v3)
+  let l = token env v1 in
+  let r = token env v3 in
+  match v2 with
+  | [ `Interp (`DOLLAR_simple_id (v1, v2)) ] when in_pattern env ->
+      (* This is something of the form "$X". This is interpreted as an interpolated
+         string of a single identifier, but if this is the pattern,
+         it's probably not what the person writing the rule meant.
```
emjin, line 1905: Would appreciate something along the lines of "so instead we interpret this as ..."

languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
```diff
@@ -1889,19 +1896,32 @@ and statements (env : env) ((v1, v2, v3) : CST.statements) =
...
+
+         This is without loss of expression, because a string containing an interpolated
```
emjin: I would phrase this as "This does not affect Semgrep's expressive power" or something like that just to be clear

languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
```diff
@@ -31,13 +31,20 @@ module H2 = AST_generic_helpers
...
+
+type context = Program | Pattern
+type env = context H.env
 
 let token = H.token
 let str = H.str
 let fb = Tok.unsafe_fake_bracket
 let sc tok = Tok.sc tok
 
+let in_pattern env =
```
emjin, line 43: I think the usual way we do this is by setting `Flag_semgrep` (unless I misunderstood what's happening here)

emjin (approved):
Looks pretty good

tests/patterns/kotlin/single_interpolated_ident.kt
brandonspark: `var` is a keyword in kotlin, so i think in this case i'm just going to use an assigned-to variable

languages/kotlin/generic/Parse_kotlin_tree_sitter.ml
brandonspark: this is done in the Swift parser: https://github.com/returntocorp/semgrep/blob/78547922c75e15f288f0258a2d4f04e0d60018b8/languages/swift/generic/Parse_swift_tree_sitter.ml#L55
and Julia:
https://github.com/returntocorp/semgrep/blob/78547922c75e15f288f0258a2d4f04e0d60018b8/languages/julia/generic/Parse_julia_tree_sitter.ml#L42

i think in other places we use `sgrep_guard` and the `sgrep_mode` ref, but honestly i think this is cleaner than using a ref. two variants, and two entry points (`pattern` or `program`) are fairly idiomatic for our tree_sitter parsers, i think. what do you think?

emjin: I don't have a strong opinion either way, just wanted us to be consistent. If we've already started doing this then that's ok with me

tests/patterns/kotlin/string_metavar.kt
brandonspark: actually, this one is intended, because the pattern has to look like a metavariable, so its intentionally supposed to be confusing (because its solving a confusing problem)

PR #7812, opened 2023-05-16
Trim ids collected and remove remaining bytes at the end of them

dinosaure:
PR checklist:

This PR trims `ids` collected by `Nosem.ml` and remove remaining bytes that are not a part of the `id`.

### DIFF
--- a/src/reporting/Nosemgrep.ml
+++ b/src/reporting/Nosemgrep.ml
@@ -124,6 +124,9 @@ let rule_match_nosem ~strict (rule_match : Out.cli_match) :
           (Option.value ~default:[||] ids_previous_line)
       in
       let ids = Common.map_filter Fun.id (Array.to_list ids) in
+      let ids = Common.map (String.split_on_char ' ') ids in
+      let ids = Common.map List.hd (* nosemgrep: list-hd *) ids in
+      (* [String.split_on_char] can **not** return an empty list. *)
       (* check if the id specified by the user is the [rule_match]'s [rule_id]. *)
       List.fold_left
         (fun (result, errors) id ->


aryx (approved):

src/reporting/Nosemgrep.ml
```diff
@@ -124,6 +135,13 @@ let rule_match_nosem ~strict (rule_match : Out.cli_match) :
           (Option.value ~default:[||] ids_previous_line)
       in
       let ids = Common.map_filter Fun.id (Array.to_list ids) in
+      let ids = Common.map String_utils.to_unicode_string ids in
+      let ids = Common.map Result.to_option ids in
+      let ids = Common.map_filter Fun.id ids in
+      let ids = Common.map (String_utils.unicode_span ~sat:no_space) ids in
+      let ids = Common.map fst ids in
+      let ids = Common.map String_utils.unicode_string_to_utf_8_string ids in
+      Fmt.epr "%a\n%!" Fmt.(Dump.list string) ids;
```
aryx: Also probably better to remove those debugging instructions.

aryx (changes requested):
Comments, deforest, and probably update the dune and dune-project files are required.

hannesm:
maybe a stupid question, but the `span` and unicode stuff seems to be like a big hammer... isn't it sufficient to get the rule_id as input, and use `String.index` with the space, tab, newline character -- and use the one that has the lowest index thereof for String.sub (or Str.first_chars)? when I understand this correctly, a rule id is terminated by any whitespace character (certainly, a question would be what to do with unicode whitespace characters, or non-printable characters (such as BEL))...

aryx:
yep, I agree with hannes, I wonder if there's a simpler way to solve it.

dinosaure:
I updated the PR according to your review and simplify a bit the process :+1:.

PR #7606, opened 2023-04-21
CLI: Allow --pro to take a single target that is not a dir

IagoAbal:
test plan:
% semgrep --pro -c p/default-v2 some_file.js # now works!

PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #7814, opened 2023-05-16
fix(test): update snapshot test which should work now

brandonspark:
## What:
This PR updates a snapshot test which is now working, which is supposed to work after #7606.

## Why:
#7606 made it such that `--pro` can now take in a single file. We had a metrics payload test which used `--pro`, with a single file, that used to crash. After that PR, however, it no longer crashes, so we should update the test.

It's not clear to me why it seems it's only failing locally. CI passed on the relevant PR, and on PRs past it, so for some reason CI wasn't complaining earlier. Let's see if it does now.

## How:
I just updated the snapshot. We see that the file is correctly parsed, and a result returned now.

## Test plan:
I ascertained that this was the case by bisecting, and seeing how far back I had to go until the snapshot stopped changing. I then got that it was introduced by #7606.

PR checklist:

### DIFF
--- a/cli/tests/e2e/snapshots/test_metrics/test_metrics_payload/pro_flag0/metrics-payload.json
+++ b/cli/tests/e2e/snapshots/test_metrics/test_metrics_payload/pro_flag0/metrics-payload.json
@@ -5,15 +5,66 @@
     "configNamesHash": "d3e03c9938537c0a0668cec7204f37f1e9da87d03e18e8cc444dd7454b36c4d5",
     "isAuthenticated": false,
     "projectHash": null,
+    "rulesHash": "e9641bba77ccfc7908dd80437de848365e2508233b7b42ca197d17fcc0eb0402",
     "version": "x.x.x"
   },
   "errors": {
-    "returnCode": 2
+    "errors": [],
+    "returnCode": 0
   },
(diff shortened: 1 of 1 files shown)


aryx:
nope, CI prefers the snapshot without those new fields. See https://github.com/returntocorp/semgrep/commit/e76d7e4300896fb205a3d79e0abf667a1677baa4

brandonspark:
@aryx yeah, i see that CI is giving a different answer, but it seems that local runs of `make test` are returning different results... i'm not sure why

aryx:
Who wrote this test?

brandonspark:
@aryx it looks like it was added in this commit: https://github.com/returntocorp/semgrep/commit/0623929094116905978f4d37002208623fb0908e by Bence

It makes total sense to me why Iago's PR should change the result of this test. I just don't know why it's different in CI but not locally...

aryx:
@underyx any insight on why this test: cli/tests/e2e/snapshots/test_metrics/test_metrics_payload/pro_flag0/metrics-payload.json works in CI but when run locally it gives different metrics (which actually appear to be the correct one)?

aryx:
Took over by https://github.com/returntocorp/semgrep/pull/7831

PR #7815, opened 2023-05-16
feat(scala): catch pattern ellipsis

### LINKED ISSUE #7807: scala: Cannot write a pattern to match every try-catch
The pattern I tried to write at first doesn't work: https://semgrep.dev/s/E24L

This works https://semgrep.dev/s/7Rve but it's not very flexible, it doesn't allow you to match _every_ try-catch.

brandonspark:
## What:
This PR makes it so you can put an ellipsis inside of a `catch { }`, as a valid pattern.

## Why:
This makes it so you can now write a pattern which matches every kind of `try-catch`, which SR would like to do.

## How:
Just translated a `catch { ... }` into a catch clause with an ellipsis pattern and ellipsis body, which matches everything.

## Test plan:
`make test`

Closes #7807 

PR checklist:


nmote (approved):
Seems legit

languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
```diff
@@ -2730,6 +2743,14 @@ and parseTry in_ : stmt =
     | Kcatch ii -> (
         nextToken in_;
         match in_.token with
+        | LBRACE _ when isEllipsisCase in_ ->
```
aryx, line 2746: please add a comment like (* semgrep: this is only needed to parse semgrep patterns! *)
actually you might want to guard it with Flag_parsing.sgrep_guard

aryx (approved):

languages/scala/recursive_descent/Parser_scala_recursive_descent.ml
brandonspark: good catch!

PR #7826, opened 2023-05-17
fix: Introduce type inference using Type.ml

nmote:
Here, I bring `Type.ml` over from the Pro Engine, and use it to it to implement a couple cases of type inference, just enough to showcase the idea. I don't think that `AST_generic.type_` is suitable for use as a type representation for any sophisticated type inference.

This way, we can build out local type inference in OSS, improving matching for typed metavariables and possibly other things when local type inference is able to determine a type. We can also make use of the Pro Engine's resolved types for symbols defined outside of the local scope to further improve typed metavariable matching in the Pro Engine.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/src/core/Type.ml
+++ b/src/core/Type.ml
@@ -1,0 +1,211 @@
+(* Yoann Padioleau
+ *
+ * Copyright (C) 2022, 2023 r2c
+ *
+ *)
+open Common
+module G = AST_generic
+
+let logger = Logging.get_logger [ __MODULE__ ]
+
+(*****************************************************************************)
+(* Prelude *)
+(*****************************************************************************)
+(* Types for type inference.
+ *
+ * Why not simply reuse AST_generic.type_ ? Because here we want to make
+ * sure type names are fully resolved! AST_generic is also meant to represent
+ * types that people could actually write as part of programs. This is a more
+ * abstract representation of types, divorced from the notion of a syntax tree.
+ * For example, we don't include tokens here. We may introduce types here that
+ * wouldn't appear as written types in a program. We might even introduce type
+ * variables here, which would not be appropriate to include in AST_generic.
+ *
+ * In the future, we may also want to apply more complex normalizations to
+ * simplify typing/naming.
+ *
+ * TODO: polymorphic types? type parameters are resolved names?
+ *
+ * These types are polymorphic to support the Pro Engine, which represents
+ * resolved names differently.
+ *)
+
+(*****************************************************************************)
+(* Types *)
+(*****************************************************************************)
+type todo_kind = string option [@@deriving show]
+
+(* Fully qualified name *)
+type 'resolved name = 'resolved * 'resolved type_argument list
+and 'resolved type_argument = TA of 'resolved t | OtherTypeArg of todo_kind
+
+and 'resolved t =
+  | N of 'resolved name * (* alt names *) G.alternate_name list
+  (* e.g. for unresolved types in core libraries
+   * TODO? generalize and have just NotYetHandled of G.type_?
+   * so at least we are as good as regular Semgrep for typing?
+   *)
+  | UnresolvedName of string * 'resolved type_argument list
+  | Builtin of builtin_type
+  | Null (* for null analysis, alt: TOption of t *)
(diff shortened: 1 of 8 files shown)


src/core/Type.ml
```diff
@@ -0,0 +1,211 @@
...
+ * These types are polymorphic to support the Pro Engine, which represents
+ * resolved names differently.
+ *)
+
+(*****************************************************************************)
+(* Types *)
+(*****************************************************************************)
+type todo_kind = string option [@@deriving show]
+
+(* Fully qualified name *)
```
aryx, line 38: maybe explain why we use a polymorphic type because this module is also used in semgrep-pro
where the resolved name is difference.

aryx: Also maybe time to define a Type.mli given the non-type-definition parts is getting big and .mli are better
documentation in that case.

src/matching/Generic_vs_generic.ml
```diff
@@ -1436,8 +1436,11 @@ and m_compatible_type lang typed_mvar t e =
 (* returns possibly the inferred type of the expression,
  * as well as an ident option that can then be used to query LSP to get the
  * type of the ident.
+ *
+ * Old type inference over `AST_generic.type_`. Let's migrate all of this to the
```
aryx: add a TODO:

src/matching/Generic_vs_generic.ml
```diff
@@ -1499,6 +1502,47 @@ and type_of_expr lang e : G.type_ option * G.ident option =
...
   | _else_ -> (None, None)
 
+(* returns possibly the inferred type of the expression,
+ * as well as an ident option that can then be used to query LSP to get the
+ * type of the ident.
+ *
+ * New Type inference over `Type.t`. Prefer this to the old type inference over
+ * `AST_generic.type_`. Eventually we'll do all the type inference here and
+ * delete the old. *)
+and type_of_expr_new lang e : G.name Type.t * G.ident option =
```
aryx: Could we move this code out of Generic_vs_generic.ml? in a typing/Typing.ml ?

src/matching/Generic_vs_generic.ml
```diff
@@ -1499,6 +1502,47 @@ and type_of_expr lang e : G.type_ option * G.ident option =
...
+            (B.Plus | B.Minus (* TODO more *)),
+            Type.Builtin Type.Int ) ->
+            Type.Builtin Type.Int
+        | _else_ -> Type.NoType
+      in
+      (t, None)
+  | _else_ -> (Type.NoType, None)
+
+and type_of_expr lang e : G.type_ option * G.ident option =
+  let t, id = type_of_expr_new lang e in
```
aryx: but this time 't' contains only fake tokens right? Is it an issue? You remember
the mess there was with the Name.t refactoring in semgrep-pro, where resolved names
suddently were "pure" without tokens but then we had to pass the token of an existing
id when converting back to the generic AST.

aryx (approved):

src/matching/Generic_vs_generic.ml
nmote: Correct. I think this is going to happen if we do type inference anyway, even if we use `AST_generic.type_`. We'll end up pulling types out of thin air in that case too, so we'll need to deal with having fake tokens.

I don't think it matters in this case though. We use this type for typed metavariable matching, but the metavariable still binds to the original expression, not the inferred type. The issues that I recall with fake tokens in synthetic AST nodes were related to metavariable bindings. I think this type should just be used to determine whether or not to match.

src/core/Type.ml
nmote: I'm not a fan of repeating potentially-large variant types in an `mli`. Especially because in this case I anticipate generating visitors for it at some point, which aren't compatible with an `mli`.

src/matching/Generic_vs_generic.ml
nmote: Sure -- I was contemplating something like this anyway. Thanks for the nudge.

src/core/Type.ml
nmote: Oh wait, I already wrote a comment that explains the polymorphism.

PR #7827, opened 2023-05-17
js: make yaml parsing work

tpetr:
(initial feedback welcome but not 100% ready to merge yet)

linter errors are because I didnt get around to installing pre-commit to my other computer


js/engine/dune
```diff
@@ -25,3 +25,12 @@
  (modes js)
  (preprocess
   (pps js_of_ocaml-ppx)))
+
+; we have to massage the libyaml memory offsets
+; because the architecture that ocaml-yaml and ocaml-ctypes
+; was built with isn't compatible with wasm32 :(
```
aryx: and there is no way to cross-compile it for 32bit?
I need a diagram explaining the compilation workflow here. I'm a bit list.

js/engine/fix-libyaml-offsets.py
```diff
@@ -0,0 +1,75 @@
+import sys
+import re
+
+RE_LINE = re.compile(r"( +)var f(\$\d+)? = \[0, ftype, (\d+), fname\];$")
+
+MAPPING = {
```
aryx: what the hell is that??

aryx (commented):
This seems wrong. There has to be a simpler way.

js/engine/dune
tpetr: Sure, it basically looks like this:
- `semgrep` uses `ocaml-yaml` for YAML parsing
- `ocaml-yaml` uses `ocaml-ctypes` to interface with the `libyaml` C library (which does the actual YAML parsing)
- `ocaml-ctypes` uses `ocamlc` at build time to extract the architecture-specific information (e.g. primitive size, pointer size, pointer alignment, etc...) needed to translate the libyaml data layout ([example](https://github.com/avsm/ocaml-yaml/blob/master/types/bindings/yaml_bindings_types.ml#L344-L352)) into the actual memory offsets.

Note: you might be wondering why we didn't hit this for `libpcre`? That's because `pcre-ocaml` uses [external functions](https://v2.ocaml.org/manual/intfc.html) directly ([example](https://github.com/mmottl/pcre-ocaml/blob/master/src/pcre.ml#L247)), so all memory arithmetic is handled in the stub code (which, in the jsoo case, I had to write myself)

Re: cross-compilation, I considered creating an additional opam switch for a 32-bit ocaml compiler (which would fix the pointer size issue, but I'm not sure if all of the primitive sizes would still add up (e.g. `long` is 8 bytes on my machine, but 4 in wasm)), but a) I wasn't wild about the additional complexity[1], b) it looks like the newest opam switch i can create is `ocaml-variants.4.10.2+32bit` (which is quite old), and c) this might force us to build everything inside a special docker container which makes local development more complicated. If you have any extra insight or advice I'm all ears 🙂 

Another option I'm considering is moving the fixup code into an initialization method inside the engine -- basically iterating through the values in `Yaml_types.M` and fixing up the offsets as needed.

[1]: I acknowledge that this PR looks complex in its own way -- it is a rough draft -- take a closer look once I've cleaned up and documented the python script and moved this out of Draft

tpetr: Newer approach in https://github.com/returntocorp/semgrep/pull/7827/commits/68ec65812a8547a25dabadb713b2d06851ac483d is a little easier to digest. tl;dr we exploit Javascript's inherent mutability to update the memory offsets at engine initialization time.

tpetr:
This branch got way out of date and I'm struggling to rebase properly, so I reopened here: https://github.com/returntocorp/semgrep/pull/7918

PR #7828, opened 2023-05-17
Use pre-generated Language.{mli,ml}

mjambon:
This gets rid of the jinja build step. See https://github.com/returntocorp/semgrep-langs/pull/32.

CI checks should pass as usual.

PR checklist:


aryx:
nice!
I still think we should move Language.ml to semgrep-interfaces (And get rid of this semgrep-langs module)
and use more ATD there, but this PR looks fine otherwise.

aryx (approved):
Love it

IagoAbal:
When it comes to the maturity type
```ocaml
type maturity =
| Alpha
| Beta
| Develop
| Ga
```
it could be better to have them declared in the right order, so that built-in compare works as expected. Also, officially we only have 3 maturity levels, is `Alpha` what we advertise as `Experimental` and `Develop` something we use internally that is below `Experimental` ?

mjambon:
@IagoAbal makes sense. I'll update semgrep-langs to sort maturity by increasing maturity.

mjambon:
```
    @pytest.mark.quick
    def test_no_duplicate_reverse_exts() -> None:
        """
        Ensures one-to-one assumption of mapping from reverse file extensions to language in lang.json
        """
        exts = set()
        for d in LANGUAGE.definition_by_id.values():
            for e in d.reverse_exts:
                if e in exts:
>                   raise Exception(f"Duplicate language key {e}")
E                   Exception: Duplicate language key .py
```
I don't know why we're getting this error only now. This is not a new test and the Python{2,3} extensions haven't changed.

Update: it's subtle stuff that wasn't explained because JSON doesn't support comments.

PR #7835, opened 2023-05-18
refactor: Delete Type_generic.ml

nmote:
Now that we have `Type.ml`, it makes more sense to use the `builtin_type` defined there than to have a separate utility. It also made sense to extract a utility for converting back to `AST_generic.type_` since we need that too.

I made some minor behavior tweaks here so this isn't a pure refactor, but the changes probably aren't noticeable.

Test plan: Automated tests

PR checklist:


nmote:
I think it might make sense to split out a lot of the helpers/converters to `Type_helpers.ml` or something along those lines, but for the moment I'd prefer to keep iterating and see what seems appropriate once this is a little bit more settled.

aryx (approved):

PR #7837, opened 2023-05-18
add(ci): check return value of /complete

brendongo:
On scan complete during logged in `semgrep ci` scans, check returned exit code to
see if should block scans. This is to support incoming features that requires
information from semgrep.dev

PR checklist:

### DIFF
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -251,7 +251,7 @@ class ScanHandler:
         lockfile_dependencies: Dict[str, List[FoundDependency]],
         dependency_parser_errors: List[DependencyParserError],
         engine_requested: "EngineType",
-    ) -> Tuple[int, str]:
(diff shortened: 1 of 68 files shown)


cli/src/semgrep/app/scans.py
```diff
@@ -391,3 +392,6 @@ def report_findings(
             raise Exception(
                 f"API server at {state.env.semgrep_url} returned this error: {response.text}"
             )
+
+        ret = response.json()
+        return (ret.get("exit_code", 0), ret.get("reason", ""))
```
mmcqd: does the app ever not send `exit_code`?

mmcqd (approved):

cli/src/semgrep/app/scans.py
brendongo: it currently does not 😓 and better to be backwards compatible if we decide to stop

PR #7840, opened 2023-05-19
feat(ls): add LS support for showAst requests

brandonspark:
## What:
This PR adds in LS support for `semgrep/showAst` requests, which are meant to be sent from the VS Code extension, to see the dumped AST for the currently open file.

## Why:
I imagine this will be helpful for developers and people who write rules.

## How:
I basically copied `ocaml-lsp` and pattern matched on what I needed to do.

## Test plan:
`make test`

PR checklist:


aryx:
The test plan is quite mysterious. How did you manage to write this code if you can't even test if it works?

brandonspark:
@aryx yeah, i was debating putting it up as "draft". i wrote it by looking through copious examples (`ocaml-lsp` and `rust-analyzer`, mostly). im just waiting on austin so he can tell me how i would actually go about verifying that it's correct

ajbt200128 (changes requested):
Overall looks good

About testing from a high level my testing plans have been:
* test whatever can be tested w/out mocking LSP messages in Unit_LS.ml (i.e. the logic to check if a match is in a changed line of a file is tested there)
* whatever requires mocking LSP messages/lifecycle stuff is tested on the python side for now `cli/tests/e2e/test_ls.py` (this is where you'd want to right tests for commands)
* As far as for VSCode, I don't really have any tests there, as 90% of the code is just hooking up commands or UI, and I want to try to keep any complicated logic out of it as best as possible (so porting to other IDEs is as simple as possible)

src/language_server/custom_requests/ShowAst.ml
```diff
@@ -0,0 +1,72 @@
...
+  let lang = Lang.lang_of_filename_exn (Fpath.v file) in
+  (* copied from -dump_ast *)
+  let { Parsing_result2.ast; _ } =
+    if named then Parse_target.parse_and_resolve_name lang file
+    else Parse_target.just_parse_with_lang lang file
+  in
+  let v = Meta_AST.vof_any (AST_generic.Pr ast) in
+  (* 80 columns is too little *)
+  Format.set_margin 120;
+  let s = OCaml.string_of_v v in
```
ajbt200128, line 74: Mentioned this in the VSCode PR, but depending on how you want to display the AST, you may want to turn this into some sort of JSON, instead of formatted text, so we can display it better in VSCode

src/language_server/custom_requests/ShowAst.ml
```diff
@@ -0,0 +1,72 @@
...
+open Lsp
+
+(*****************************************************************************)
+(* Prelude *)
+(*****************************************************************************)
+(* Custom LSP request for a command which will show the Semgrep
+   Generic AST of the given file.
+
+   This file largely follows the structure of ocaml-lsp, such as:
+   https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/src/custom_requests/req_wrapping_ast_node.ml
```
ajbt200128, line 25: I like this pattern, we should probably do the same for custom notifications

src/language_server/custom_requests/ShowAst.mli
```diff
@@ -0,0 +1,2 @@
+val meth : string
+val on_request : Jsonrpc.Structured.t option -> Yojson.Safe.t option
```
ajbt200128, line 2: This is more of a question for my benefit than for the review, but is this somewhere we'd want to use signatures (i.e. any custom request file would implement some custom request signature)?

brandonspark: I'm actually not sure of the best way to do this, because I'm not sure that we could make each file conform to a signature, for example. We could put a module in each file that ascribes to the same signature, but that would encapsulate the data twice, and also I figure each `on_request` method might have different parameters. For now, I'm OK with different signatures for each file.

src/language_server/custom_requests/ShowAst.ml
brandonspark: I'm not actually sure if there's a way to generate a `Yojson` value from our `AST_generic`. For now, can we move with the formatted string, and then fix it later if we see a need to make it nicer?

brandonspark:
@ajbt200128 i took a break from this for a very long time... sorry about bringing it back. but hopefully, it has better coverage now

src/language_server/custom_requests/ShowAst.mli
ajbt200128: Makes sense, that's what I was thinking to but was unsure

src/language_server/custom_requests/ShowAst.ml
ajbt200128: I think you should be able to do
`ast |> AST_generic_to_v1.program |> Ast_generic_v1_j.string_of_program`
or another similar Ast_generic_v1_j method. There's an atd spec of the ast so I think there should be a way if that's not it.

I'm ok with the text view tho, mostly leaving this comment for if we want to come back to it in the future

ajbt200128 (approved):
Tests and everything look good! We might want to wait to merge until we've tested it e2e w/vscode, but also the tests look like they're passing so it should be pretty safe, up to you

ajbt200128:
Looks good, merge whenever @brandonspark

aryx:
Ok CI is passing so I'll merge it for brandon.

PR #7844, opened 2023-05-19
osemgrep: some more output tweaks

hannesm:
for one, it is indentation of the scan status section

on the other note, this respects the environment variables for negetable flags

test plan:
- make osempass

PR checklist:

### DIFF
--- a/libs/commons/Cmdliner_helpers.ml
+++ b/libs/commons/Cmdliner_helpers.ml
@@ -28,14 +28,34 @@ let add_option_dashes option_names =
    backward compatibility with the Python CLI.
    See https://github.com/dbuenzli/cmdliner/issues/164
 *)
-let negatable_flag ?(default = false) ?env ~neg_options ~doc options =
+let negatable_flag ?(default = false) ~neg_options ~doc options =
   let neg_doc =
     let options_str = add_option_dashes options |> String.concat "/" in
     Printf.sprintf "negates %s" options_str
   in
-  let enable = (true, Arg.info options ~doc ?env) in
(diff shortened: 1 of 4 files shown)


libs/commons/Cmdliner_helpers.ml
```diff
@@ -28,15 +28,35 @@ let add_option_dashes option_names =
...
     Printf.sprintf "negates %s" options_str
   in
-  let enable = (true, Arg.info options ~doc ?env) in
+  let enable = (true, Arg.info options ~doc) in
   let disable = (false, Arg.info neg_options ~doc:neg_doc) in
   Arg.value (Arg.vflag default [ enable; disable ])
 
+(* Cmdliner.Arg.vflag does not support environment variables, thus we use
+   Arg.value manually if we need supporting environment variables as well *)
+let negatable_flag_with_env ?(default = false) ?env ~neg_options ~doc options =
```
aryx, line 42: Should we just replace the buggy negatable_flag by this new implementation?

aryx (approved):

libs/commons/Cmdliner_helpers.ml
hannesm: We can of course replace the `negatable_flags` by this code. I removed the `env` argument from `negatable_flags` (to ensure future uses won't end up in the same situation).

I somehow prefer the `Arg.vflag` approach better (more stuff is handled in Cmdliner) than the hand-crafted `negatable_flag_with_env`, but I also understand only using one approach leads to easier to understand code and less code paths to maintain.

PR #7845, opened 2023-05-19
fix(perf): set better limits on analyze_rule

### LINKED ISSUE #7839: filter-irrelevant-rule causes memory/perf blowup on some rules
**Describe the bug**
https://semgrep.dev/playground/s/E2B5 takes forever (> 13 minutes).

So does https://semgrep.dev/playground/s/LbgX (rule = `kotlin_faster_but_longer.yaml`, target = `kotlin_slow_import.kt`). However, though this one is longer, it takes less time (2 minutes)

The two rules are pretty similar; the shorter rule has deleted some patterns compared to the longer rule.

**To Reproduce**

Create files from the playground example.

```
➜  misc semgrep --config kotlin_faster_but_longer.yaml kotlin_slow_import.kt -d
➜  misc sc -json -rules /Users/emma/.semgrep/semgrep_rules.json -j 12 -targets /Users/emma/.semgrep/semgrep_targets.txt -timeout 30 -timeout_threshold 3 -max_memory 0 -json_time -fast -debug
[0.043  Info       Main.Core_CLI        ] Executed as: /Users/emma/workspace/semgrep/_build/install/default/bin/semgrep-core -json -rules /Users/emma/.semgrep/semgrep_rules.json -j 12 -targets /Users/emma/.semgrep/semgrep_targets.txt -timeout 30 -timeout_threshold 3 -max_memory 0 -json_time -fast -debug
[0.043  Info       Main.Core_CLI        ] Version: semgrep-core version: 1.22.0
[0.043  Info       Main.Run_semgrep     ] Parsing /Users/emma/.semgrep/semgrep_rules.json:
[0.049  Info       Main.Run_semgrep     ] extracting nested content from 1 files
[0.049  Info       Main.Run_semgrep     ] processing 1 files, skipping 0 files
[0.000  Info       Main.Run_semgrep     ] [51223] Analyzing kotlin_slow_import.kt
[0.897  Warning    Main.Memory_limit    ] [51223] [kotlin_slow_import.kt] large heap size: 673 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[1.496  Warning    Main.Memory_limit    ] [51223] [kotlin_slow_import.kt] large heap size: 1024 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[2.484  Warning    Main.Memory_limit    ] [51223] [kotlin_slow_import.kt] large heap size: 1791 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[8.372  Warning    Main.Memory_limit    ] [51223] [kotlin_slow_import.kt] large heap size: 4765 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[15.810 Warning    Main.Memory_limit    ] [51223] [kotlin_slow_import.kt] large heap size: 8335 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[28.459 Warning    Main.Memory_limit    ] [51223] [kotlin_slow_import.kt] large heap size: 14578 MiB (memory limit is 0 MiB). If a crash follows, you could suspect OOM.
[60.922 Error      Main.Analyze_rule    ] [51223] CNF size exploded on rule id ktor_request_xss
```

Takes about 2 minutes.

Then, try running without filter-irrelevant-rules.

```
➜  misc time sc -json -rules /Users/emma/.semgrep/semgrep_rules.json -j 12 -targets /Users/emma/.semgrep/semgrep_targets.txt -timeout 30 -timeout_threshold 3 -max_memory 0 -no_filter_irrelevant_rules 
.
{"matches":[{"rule_id":"ktor_request_xss","location":{"path":"kotlin_slow_import.kt","start":{"line":56,"col":23,"offset":2983},"end":{"line":56,"col":36,"offset":2996}},"extra":{"message":"","metavars":{"$F":{"start":{"line":20,"col":36,"offset":592},"end":{"line":20,"col":42,"offset":598},"abstract_content":"accept"},"$RESPFUNC":{"start":{"line":55,"col":18,"offset":2947},"end":{"line":55,"col":30,"offset":2959},"abstract_content":"respondBytes"},"$INPUT":{"start":{"line":56,"col":23,"offset":2983},"end":{"line":56,"col":36,"offset":2996},"abstract_content":"\"\"hi! ${${resp}\""}},"dataflow_trace":{"taint_source":["CoreLoc",{"path":"kotlin_slow_import.kt","start":{"line":20,"col":32,"offset":588},"end":{"line":20,"col":44,"offset":600}}],"intermediate_vars":[{"location":{"path":"kotlin_slow_import.kt","start":{"line":19,"col":17,"offset":516},"end":{"line":19,"col":21,"offset":520}}}],"taint_sink":["CoreLoc",{"path":"kotlin_slow_import.kt","start":{"line":56,"col":23,"offset":2983},"end":{"line":56,"col":36,"offset":2996}}]},"engine_kind":"OSS"}}],"errors":[],"skipped_rules":[],"explanations":[],"stats":{"okfiles":1,"errorfiles":0},"rules_by_engine":[["ktor_request_xss","OSS"]],"engine_requested":"OSS"}
/Users/emma/workspace/semgrep/_build/install/default/bin/semgrep-core -json    0.05s user 0.03s system 29% cpu 0.260 total
```

**Expected behavior**
Expect the optimized version to not be significantly slower.

**Environment**

**Use case**

emjin:
Closes https://github.com/returntocorp/semgrep/issues/7839

In Analyze_rule, we distribute out formulas like `(x0 ^ x1) v (y0 ^ y1)` into cnf. Let's call `x0 ^ x1` p and `y0 ^ y1` q. Currently, we check that `p` and `q` are individually under 1,000,000 elements to prevent memory blowup. However, `p` and `q` produce a formula that is `p * q` elements long. So, even under these constraints, if `p` and `q` are both large, this step can still take too long.

Instead, what we need to do is guard the result. This PR currently requires that that result be under 10,000,000 elements. It leaves the `p` and `q` restrictions intact just in case. This allows us to take advantage of `List.compare_length` in case we get an incredibly long `p` or `q`.

To consider: do we still need the gates for `p` and `q`? Is it possible to get a list for `p` or `q` so large that traversing it is a problem?

Test plan: will add tests

PR checklist:

### DIFF
--- a/changelog.d/gh-7839.fixed
+++ b/changelog.d/gh-7839.fixed
@@ -1,0 +1,2 @@
+Fix for very long runtimes that could happen due to one of our optimizations. We now detect when that might
+happen and skip the optimization.
--- a/src/optimizing/Analyze_rule.ml
+++ b/src/optimizing/Analyze_rule.ml
@@ -205,13 +205,20 @@ let rec (cnf : Rule.formula -> cnf_step0) =
       let zs = Common.map (fun (_t, cond) -> And [ Or [ LCond cond ] ]) conds in
       And (ys @ zs |> List.concat_map (function And ors -> ors))
   | R.Or (_, xs) ->
-      let is_dangerously_large l = List.compare_length_with l 1_000_000 > 0 in
+      let is_dangerously_large p q =
+        List.compare_length_with p 1_000_000 > 0
+        || List.compare_length_with q 1_000_000 > 0
+        ||
+        let p_len = List.length p in
+        let q_len = List.length q in
+        (* Divide rather than multiply to avoid integer overflow *)
+        p_len > Int.div 10_000_000 q_len
+      in
       let ys = Common.map cnf xs in
       List.fold_left
         (fun (And ps) (And qs) ->
           (* Abort before this starts consuming insane amounts of memory. *)
-          if is_dangerously_large ps || is_dangerously_large qs then
-            raise CNF_exploded;
+          if is_dangerously_large ps qs then raise CNF_exploded;
           (* Distributive law *)
           And
             (ps
--- a/src/optimizing/Analyze_rule.ml
+++ b/src/optimizing/Analyze_rule.ml
@@ -603,7 +603,8 @@ let regexp_prefilter_of_rule (r : R.rule) =
             regexp_prefilter_of_formula f
         | `Taint spec -> regexp_prefilter_of_taint_rule r.R.id spec
       with
-      (* TODO: see tests/rules/tainted-filename.yaml *)
+      (* TODO: see tests/rules/tainted-filename.yaml,
+                   tests/rules/kotlin_slow_import.yaml *)
       | CNF_exploded ->
           logger#error "CNF size exploded on rule id %s" rule_id;
           None
--- a/tests/rules/kotlin_slow_import.kt
+++ b/tests/rules/kotlin_slow_import.kt
@@ -51,8 +51,8 @@ fun Route.customerRouting() {
                     // "userAgent (raw): ${req.userAgent()} \n" +
                     // "userAgent (decoded): ${URLDecoder.decode(req.userAgent())} \n"
 
-            // ruleid: ktor_request_xss_2
             call.respondBytes(
+                // ruleid: ktor_request_xss_2
                 bytes="hi! ${resp}",
                 contentType = ContentType.Text.Html,
             )
--- a/tests/rules/kotlin_slow_import.kt
+++ b/tests/rules/kotlin_slow_import.kt
@@ -1,0 +1,61 @@
+package com.example.routes
+
+import com.example.models.Customer
+import com.example.models.customerStorage
+import io.ktor.http.*
+import io.ktor.server.application.*
+import io.ktor.server.logging.*
+import io.ktor.server.request.*
+import io.ktor.server.response.*
+import io.ktor.server.routing.*
+import java.io.StringWriter
+import java.net.URLDecoder
+
+fun Route.customerRouting() {
+    route("/customer") {
+        get("hi/{asdf?}") {
+            val req = call.request
+            val local = req.local
+            val resp = "Customer stored correctly. \n" +
+                    "accept: ${req.accept()} \n"
+                    // "acceptCharset: ${req.acceptCharset()} \n" +
+                    // "acceptCharsetItems: ${req.acceptCharsetItems()} \n" +
+                    // "acceptEncoding: ${req.acceptEncoding()} \n" +
+                    // "acceptEncodingItems: ${req.acceptEncodingItems()} \n" +
+                    // "acceptItems: ${req.acceptItems()} \n" +
+                    // "acceptLanguage: ${req.acceptLanguage()} \n" +
+                    // "acceptLanguageItems: ${req.acceptLanguageItems()} \n" +
+                    // "cacheControl: ${req.cacheControl()} \n" +
+                    // "cookies: ${req.cookies["asdf"]} \n" +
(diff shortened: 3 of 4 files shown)


enncoded:
woohoo thank you so much!!

aryx (approved):

src/optimizing/Analyze_rule.ml
```diff
@@ -205,13 +205,20 @@ let rec (cnf : Rule.formula -> cnf_step0) =
       let zs = Common.map (fun (_t, cond) -> And [ Or [ LCond cond ] ]) conds in
       And (ys @ zs |> List.concat_map (function And ors -> ors))
   | R.Or (_, xs) ->
-      let is_dangerously_large l = List.compare_length_with l 1_000_000 > 0 in
+      let is_dangerously_large p q =
+        List.compare_length_with p 1_000_000 > 0
+        || List.compare_length_with q 1_000_000 > 0
+        ||
+        let p_len = List.length p in
```
IagoAbal, line 212: Once we use `List.length` here then it would be better to refactor the code above to just use `p_len` and `q_len`, it becomes pointless to use `List.compare_length_with`.

IagoAbal: Actually I'm not sure we even need `List.compare_length_with p 1_000_000 > 0  || List.compare_length_with q 1_000_000 > 0`, probably the new condition suffices.

changelog.d/gh-7839.fixed
```diff
@@ -0,0 +1,2 @@
+Fix for very long runtimes that could happen due to one of our optimizations. We now detect when that might
```
IagoAbal, line 1: I'm nitpicking, but as a user I would find this changelog entry kind of confusing, and as a developer I would like to know what optimization was that. I think users are less interested in knowing that the long runtimes could happen because of an optimization. They are more interested in knowing _when_ that happens, i.e. what kind of files/rules would trigger this perf issue.

src/optimizing/Analyze_rule.ml
IagoAbal: OK I saw your PR comment, my bad not having read it before, although I think that comment belonged here in the code.

IagoAbal: I think using `List.compare_length_with` here was an acceptable "premature optimization" because it didn't really add any complexity and it could potentially speed up things. But computing the length of a large list should be pretty fast. And the problem with how it looks now is that the _normal_ path will be to evaluate the three Boolean conditions (because the CNF will rarely explode), so we're just making this check (a bit) slower by keeping the previous conditions. And the new condition is _almost_ stricter than the previous ones anyways. You could have `ps` having 1_000_001 elements, and `qs` having up to 9 elements without your new condition becoming true but... does it matter? If it matters then better to just change your condition to `p_len > Int.div 1_000_000 q_len`.

emjin: That's a fair objection. I figured List.length was fast enough it didn't matter, and there is the slight chance that a list could be truly pathologically long. But I wasn't sure if that could actually happen without the rule being somehow invalid. That's why I made the comment a PR comment, I expected that I would change the code.

I don't really have strong feelings, the numbers here are largely based on vibes. (I know that 50 million and 100 million are definitely too long but not what the lower bound is)

changelog.d/gh-7839.fixed
emjin: Thanks for the feedback! I'm too lazy to make a PR to improve the changelog since I don't have a much better explanation 😅 but I'll keep that in mind for next time

PR #7846, opened 2023-05-19
fix: Handle variables in type inference

nmote:
Out of the two new test cases, previously only the first one matched. We already handled the special case where a typed metavariable was matching a variable with a resolved type. However, the second case did not match.

In many cases we can assume that the type of a variable has already been resolved. This is handled by `Naming_AST.ml` (and the equivalent in Pro Engine). In some languages, type annotations do not exist on all variables, so we might need to recurse on the `id_svalue` if it has one. For now, we assume that naming has been able to resolve the type of a variable.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/src/typing/Typing.ml
+++ b/src/typing/Typing.ml
@@ -101,6 +101,9 @@ and type_of_expr_new lang e : G.name Type.t * G.ident option =
         | _else_ -> Type.NoType
       in
       (t, None)
+  | G.N (Id (ident, id_info)) ->
+      let t = resolved_type_of_id_info lang id_info in
+      (t, Some ident)
   | G.Call ({ e = IdSpecial (Op op, _); _ }, (_l, [ Arg e1; Arg e2 ], _r)) ->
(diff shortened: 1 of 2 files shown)


src/typing/Typing.ml
```diff
@@ -125,3 +128,17 @@ and type_of_expr lang e : G.type_ option * G.ident option =
   with
   | None -> type_of_expr_old lang e
   | Some t -> (Some t, id)
+
+and resolved_type_of_id_info lang info : G.name Type.t =
+  match !(info.G.id_type) with
```
aryx, line 133: Do we want to change id_type to be of Type.t instead of AST_generic.type_?

aryx (approved):

src/typing/Typing.ml
nmote: I do, yes. I haven't yet looked into exactly what that will require.

PR #7850, opened 2023-05-21
Fix plurality of word files in output messages

allandao:
PR checklist:

- [ ] Tests included or PR comment includes a reproducible test plan. Intentionally left blank, please see concern in comments

### DIFF
--- a/src/osemgrep/reporting/Summary_report.ml
+++ b/src/osemgrep/reporting/Summary_report.ml
@@ -61,22 +61,26 @@ let pp_summary ppf
   *)
   let opt_msg msg = function
     | [] -> None
-    | xs -> Some (string_of_int (List.length xs) ^ " " ^ msg)
+    | xs ->
+      let file_count = List.length xs in
+      let file_str = if file_count = 1 then "file" else "files" in
+      Some (string_of_int file_count ^ " " ^ file_str ^ " " ^ msg)
   in
   let out_skipped =
     let mb = string_of_int Stdlib.(max_target_bytes / 1000 / 1000) in
     Common.map_filter Fun.id
       [
-        opt_msg "files not matching --include patterns" include_ignored;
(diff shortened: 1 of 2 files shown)


allandao:
Tests included or PR comment includes a reproducible test plan:
Create a sample file, such as test.js:
```
if (a == b) {.  // Purposeful syntax error to trigger partial scan output
	console.log("hello world");
}
```

Run to trigger Partially Scanned: 1 file and Skipped Scan: 1 file respectively
```
semgrep --config="p/auto" --include="*.txt"
semgrep --config="p/auto"
```

allandao:
I have some concerns about this PR creating tech debt:

1. Automatic scanning/parsing may break in some instances

Current csers or particular features that scan for the phrase "files ... " in output messages may now fail in cases of an example output message such as "1 file ignored not matching --include patterns" as we now present both "file" and "files" as things to search for. However, given that the rest of the summary already had correct plurality, scanning code should have already been robust enough the handle the edge case of 1 file instances.

2. Snapshots become out of date
Ex: https://github.com/search?q=repo%3Areturntocorp%2Fsemgrep+Partially+scanned%3A&type=code

This change may necessitate that these snapshots change given that the plurality of the word "files" would be incorrect. Thank you!

src/osemgrep/reporting/Summary_report.ml
```diff
@@ -61,22 +61,26 @@ let pp_summary ppf
   *)
   let opt_msg msg = function
     | [] -> None
-    | xs -> Some (string_of_int (List.length xs) ^ " " ^ msg)
+    | xs ->
+      let file_count = List.length xs in
+      let file_str = if file_count = 1 then "file" else "files" in
```
aryx, line 66: We have a function for that in libs/commons/String_utils.ml call unit_str that you could reuse

PR #7852, opened 2023-05-22
sym-prop: Allow some Sym-svalues to survive after JOINs

IagoAbal:
Previously any Sym-value would be "killed" at a JOIN point, as a precautionary measure to avoid `id_svalue` cycles. Here we relax this condition to allow trivial cases to work. Basically, if a Sym-value for a variable 'x' was recorded before a fork in the CFG, and then we reach the JOIN without 'x' having been modified, so that the Sym-value assigned to 'x' in all branches is exactly (physically) the same, then we allow that Sym-value to survive.

Helps PA-2739

test plan:
make test # new tests

PR checklist:


tests/rules/sym_prop_merge.py
```diff
@@ -0,0 +1,20 @@
...
+    if cond:
+        y = 42
+    #ruleid: test
+    foo(x)
+
+def test2():
+    x = bar()
+    while cond:
+        y = 42
+    #todoruleid: test
```
aryx, line 12: why this one is not handled?

aryx (approved):

tests/rules/sym_prop_merge.py
IagoAbal: I haven't debug it to be 100% sure, but I believe it's related to Dataflow_svalue not making use of `_|_`  and instead considering that any variable for which we don't know anything it's a non-constant. That introduces some limitations with loops, there are also examples with const-prop. But left the `todoruleid` there to investigate/fix in the future.

PR #7853, opened 2023-05-22
Makefile: in minimal-build target, only build semgrep-core and osemgrep

hannesm:
in core target, build everything

This cuts a rebuild (the second "make minimal-build" in make minimal-build ; make minimal-build from 10s down to 2.5s)

PR checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,8 +120,8 @@ symlink-core-for-cli:
 
 # Minimal build of the semgrep-core executable. Intended for the docker build.
 # Requires the environment variables set by the included file above.
-# Builds only the two main binaries mainly needed for development. If you
-# need other binaries, look at the rules below.
+# Builds only the two main binaries needed for development.
+# If you need other binaries, look at the rules below.
 .PHONY: minimal-build
 minimal-build:
 	dune build _build/install/default/bin/semgrep-core _build/install/default/bin/osemgrep
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,6 @@ build:
 core:
 	rm -f bin
 	$(MAKE) minimal-build
-	dune build ./_build/default/tests/test.exe
 	# make executables easily accessible for manual testing:
 	test -e bin || ln -s _build/install/default/bin .
 
@@ -250,9 +249,8 @@ test:
 core-test: core
 	# The test executable has a few options that can be useful
 	# in some contexts.
-	# The following command ensures that we can call 'test.exe --help'
-	# without having to chdir into the test data folder.
-	./_build/default/tests/test.exe --show-errors --help 2>&1 >/dev/null
+	dune build ./_build/default/src/tests/test.exe
+	cd ./_build/default/src/tests/ && ./test.exe --show-errors --help 2>&1 >/dev/null
 	$(MAKE) -C libs/spacegrep test
 	dune runtest -f --no-buffer
 
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,6 @@ build:
 core:
 	rm -f bin
 	$(MAKE) minimal-build
-	dune build
 	dune build ./_build/default/tests/test.exe
 	# make executables easily accessible for manual testing:
 	test -e bin || ln -s _build/install/default/bin .
@@ -138,6 +137,34 @@ build-docker:
 build-otarzan:
 	rm -f bin
 	dune build _build/install/default/bin/otarzan
+	test -e bin || ln -s _build/install/default/bin .
+
+# Build just this executable
+.PHONY: build-pfff
+build-pfff:
+	rm -f bin
+	dune build _build/install/default/bin/pfff
+	test -e bin || ln -s _build/install/default/bin .
+
+# Build just this executable
+.PHONY: build-parse-cairo
+build-parse-cairo:
+	rm -f bin
+	dune build _build/install/default/bin/parse-cairo
+	test -e bin || ln -s _build/install/default/bin .
+
+# Build just this executable
+.PHONY: build-spacegrep
+build-spacegrep:
+	rm -f bin
+	dune build _build/install/default/bin/spacegrep
+	test -e bin || ln -s _build/install/default/bin .
+
+# Build just this executable
+.PHONY: build-oncall
+build-oncall:
+	rm -f bin
+	dune build _build/install/default/bin/oncall
 	test -e bin || ln -s _build/install/default/bin .
 
 # Build the js_of_ocaml portion of the semgrep javascript packages
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,7 @@ build:
 core:
 	rm -f bin
 	$(MAKE) minimal-build
+	dune build
 	dune build ./_build/default/tests/test.exe
 	# make executables easily accessible for manual testing:
 	test -e bin || ln -s _build/install/default/bin .
@@ -123,7 +124,7 @@ symlink-core-for-cli:
 # Requires the environment variables set by the included file above.
 .PHONY: minimal-build
 minimal-build:
-	dune build
+	dune build _build/install/default/bin/semgrep-core _build/install/default/bin/osemgrep
 
 # It is better to run this from a fresh repo or after a 'make clean',
 # to not send too much data to the Docker daemon.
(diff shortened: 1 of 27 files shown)


src/tests/dune
```diff
@@ -29,8 +29,5 @@
    (pps
       ppx_deriving.show
    ))
+ (deps (source_tree ../tests) (source_tree ../cli/tests/e2e/snapshots))
```
aryx, line 36: nice! This was the thing that I didn't know about.
Maybe add a small comment before it, explaining that without this line, a modification to tests/patterns/xxx/foo.yml
would not be "seen" by dune runtest -f, which may use the cached _build/tests/... files and run the test suite with
old version of the file.

aryx (approved):

hannesm:
The error is:
```
File "languages/lua/tree-sitter/semgrep-lua/lib/dune", line 39, characters 11-18:
39 |     (names scanner)
                ^^^^^^^
(cd _build/default/languages/lua/tree-sitter/semgrep-lua/lib && /usr/bin/cc -fPIC -I . -g -I /Users/ssm-user/.opam/4.14.0/lib/ocaml -I /Users/ssm-user/.opam/4.14.0/lib/ANSITerminal -I /Users/ssm-user/.opam/4.14.0/lib/atdgen-runtime -I /Users/ssm-user/.opam/4.14.0/lib/base/caml -I /Users/ssm-user/.opam/4.14.0/lib/biniou -I /Users/ssm-user/.opam/4.14.0/lib/cmdliner -I /Users/ssm-user/.opam/4.14.0/lib/easy-format -I /Users/ssm-user/.opam/4.14.0/lib/parsexp -I /Users/ssm-user/.opam/4.14.0/lib/ppx_deriving/runtime -I /Users/ssm-user/.opam/4.14.0/lib/ppx_sexp_conv/runtime-lib -I /Users/ssm-user/.opam/4.14.0/lib/re -I /Users/ssm-user/.opam/4.14.0/lib/result -I /Users/ssm-user/.opam/4.14.0/lib/seq -I /Users/ssm-user/.opam/4.14.0/lib/sexplib -I /Users/ssm-user/.opam/4.14.0/lib/sexplib0 -I /Users/ssm-user/.opam/4.14.0/lib/tsort -I /Users/ssm-user/.opam/4.14.0/lib/yojson -I ../../../../../libs/ocaml-tree-sitter-core/src/bindings/lib -I ../../../../../libs/ocaml-tree-sitter-core/src/gen/lib -I ../../../../../libs/ocaml-tree-sitter-core/src/run/lib -o scanner.o -c scanner.cc)
scanner.cc:1:10: fatal error: 'tree_sitter/parser.h' file not found
#include <tree_sitter/parser.h>
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

in a `make minimal-build`. I don't quite understand (yet) where the `-I ../../../../../libs/ocaml-tree-sitter-core/src/gen/lib -I ../../../../../libs/ocaml-tree-sitter-core/src/run/lib` is coming from, and why
```
  (foreign_stubs
    (language c)
    (names parser bindings)
    (flags -std=c99
           -fPIC
           -I %{env:TREESITTER_INCDIR=/usr/local/include}
           -I .)
  )
```

seem to not have an effect. I'll work on this further.

aryx:
regarding tree-sitter, there is this important line in the Makefile:
-include libs/ocaml-tree-sitter-core/tree-sitter-config.mk

to setup some environment variables for dune to be able to find the tree-sitter libs. This is why we can't run 'dune build' directly from the terminal but have to go through make, which set the environment variables.
Dunno if it has a link with your error though.

hannesm:
My remark above lead to https://github.com/returntocorp/ocaml-tree-sitter-core/pull/63 which will lead to updates of tree-sitter submodules in here, and once we have that, the compilation error is gone.

Makefile
```diff
@@ -121,9 +120,11 @@ symlink-core-for-cli:
 
 # Minimal build of the semgrep-core executable. Intended for the docker build.
 # Requires the environment variables set by the included file above.
+# Builds only the two main binaries mainly needed for development. If you
```
aryx: Can remove one of the main

Makefile
```diff
@@ -220,9 +249,8 @@ test:
 core-test: core
 	# The test executable has a few options that can be useful
 	# in some contexts.
-	# The following command ensures that we can call 'test.exe --help'
-	# without having to chdir into the test data folder.
-	./_build/default/tests/test.exe --show-errors --help 2>&1 >/dev/null
+	dune build ./_build/default/src/tests/test.exe
+	cd ./_build/default/src/tests/ && ./test.exe --show-errors --help 2>&1 >/dev/null
```
aryx: could we instead run ./_build/default/src/tests/test.exe from the root, so that we would not need all those ../../../../ ?

src/tests/dune
aryx, line 36: This would require a big comment like 
; If you add test code that depends on some data, you need to update the deps below otherwise
; dune buld will just copy the data under _build/ and will not use the latest version of this data, even if you modify
; this data.

Makefile
hannesm: the issue is that `dune runtest` two lines below executes `test.exe` within the `_build/default/src/tests` directory. but I'm looking into whether we can change that.

Makefile
hannesm: done in 4165f8ea650cbe358af250c44b763c4af45e9086

Makefile
hannesm: done in fe26ede0bb57f7f7269bf3705eda2f573c53bc36

src/tests/dune
hannesm: done in 6f6946960d23f7a255e32d0fcb0018696d4b5925 -- is this understandable?

With this PR, I can "dune runtest" (or make core-test), then modify any rule, and a subsequent make core-test will rerun the test suite and report failure (if I messed up the rule).

So, the "will not use the latest version of this data, even if you modify this data." is not true anymore, since now dune knows that if ../tests (or ../cli/tests/e2e/snapshots) is modified, there's need to re-run the test suite.

Makefile
```diff
@@ -220,9 +249,8 @@ test:
 core-test: core
 	# The test executable has a few options that can be useful
 	# in some contexts.
-	# The following command ensures that we can call 'test.exe --help'
-	# without having to chdir into the test data folder.
-	./_build/default/tests/test.exe --show-errors --help 2>&1 >/dev/null
+	dune build ./_build/default/src/tests/test.exe
```
aryx, line 252: I remember the meaning of that comment. There was a time where we could not even run
test.exe outside of the _build/... directory because we were doing early initialization of 
OCaml global constants by accessing files from ../../../tests.
So that's why we added those 2 cryptic commands.
Maybe you can keep back the old command (and make it less cryptic).

src/tests/Test.ml
```diff
@@ -93,13 +88,24 @@ let tests () = List.flatten [
...
+  try tests () with
+  | e ->
+      let exn = Exception.catch e in
+      [
+        ( "cannot load test data - not a real test",
+          fun () -> Exception.reraise exn );
+      ]
 
 let main () =
+  let rec parent () =
```
aryx: maybe add a comment like
(* find the root of the semgrep repo as many of our tests rely on 'let test_path = "tests/"' to
find their test files *)

aryx: Also maybe print a message like 'chdir to X', to let the developer knows.

aryx (approved):

Makefile
hannesm: done in a350defb8eaa36e96a1d887613eef4cf46cfc184

src/tests/Test.ml
hannesm: done in a350defb8eaa36e96a1d887613eef4cf46cfc184

hannesm:
rebased on develop branch, CI is hopefully fine now.

hannesm:
CI run shows that ruby and vue are still missing the `-I %{env:TREESITTER_INCDIR=/usr/local/include}` in the dune file.

EDIT: fixed in #[7873](https://github.com/returntocorp/semgrep/pull/7873)

aryx (approved):

aryx (approved):

PR #7856, opened 2023-05-22
osemgrep: skeleton for the new 'osemgrep interactive' subcommand

aryx:
test plan:
osemgrep interactive --help

PR checklist:


brandonspark (approved):
looks good to me! would it be possible to also enter this case if `osemgrep -i` or something is specified? `interactive` is kind of long, and id rather not everyone just be forced to alias it

aryx:
The problem with overloading osemgrep with all those flags is that the configuration becomes a mess.
Actually when you type semgrep <blah>, it's really semgrep scan <blah> that you type. In the past we made the mistate of
providing semgrep --test, semgrep --validate, and it bites us, so it's better I think to separate clearly things when they operate very differently. Those are separate subcommands of semgrep.

aryx:
cc @brendongo who also pushed a long time ago for separate subcommands.

PR #7865, opened 2023-05-23
Fix bug with yarn v1 parsing

adamberman:
Yarn v1 dependency blocks look something like this:

```
my-package@1.2.3:
  version "1.2.3"
  resolved "https://some.place.on/the/internet"
  integrity sha1-some-hash
```

The first line defining the package can have a variety of formats:
* `my-package@1.2.3:`
* `"@company/my-package@^2.3.4":`
* `my-package@^2.0.0, my-package@^2.0.2:`

We recently learned about a new valid syntax for defining a package where a version constraint is not included at all in the header: `my-package-without-version-constraint:`. We were failing to parse this.

To solve this, we refactor the header parser into two regex parsers.

PR checklist:


mmcqd (approved):
I have to admit I find this regex pretty confusing:
```
@?([^:,"]*(:?(?!\n)[^:,"]*)*)"?
```
but the yarn v1 format is pretty confusing, this works, and what we had before was pretty confusing too.

Good to go once you add a changelog entry

PR #7867, opened 2023-05-23
tainting: pro: JS/TS: Interprocedural field-sensitivity

IagoAbal:
Helps PA-2570

test plan:
See returntocorp/semgrep-proprietary#743

PR checklist:

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -738,7 +738,7 @@ let propagate_taint_via_java_setter env e args all_args_taints =
   | __else__ -> env.lval_env
 
 let fix_poly_taint_with_field env lval st =
-  if env.lang =*= Java then
+  if env.lang =*= Lang.Java || Lang.is_js env.lang then
     match lval.rev_offset with
     | { o = Dot n; _ } :: _ -> (
         match st with
@@ -903,11 +903,8 @@ let find_lval_taint_sources env ~labels lval =
   let lval_env = Lval_env.add env.lval_env lval taints_sources_mut in
   (Taints.union taints_sources_reg taints_sources_mut, lval_env)
 
-let rec check_tainted_lval env ?(fix_poly_field = false) (lval : IL.lval) :
-    Taints.t * Lval_env.t =
-  let new_taints, lval_in_env, lval_env =
-    check_tainted_lval_aux env ~fix_poly_field lval
-  in
+let rec check_tainted_lval env (lval : IL.lval) : Taints.t * Lval_env.t =
+  let new_taints, lval_in_env, lval_env = check_tainted_lval_aux env lval in
   let taints_from_env = status_to_taints lval_in_env in
   let taints = Taints.union new_taints taints_from_env in
   let sinks =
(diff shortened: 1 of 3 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -986,13 +977,11 @@ and check_tainted_lval_aux env ~fix_poly_field (lval : IL.lval) :
...
-                if fix_poly_field then
-                  (* HACK(field-sensitivity): If we encounter `obj.x` and `obj` has
-                     * polymorphic taint,  and we know nothing specific about `obj.x`, then
-                     * we add the same offset `.x` to the polymorphic taint coming from `obj`.
-                     * See also 'propagate_taint_via_java_setter'. *)
-                  fix_poly_taint_with_field env lval st
-                else st)
+                (* HACK(field-sensitivity): If we encounter `obj.x` and `obj` has
+                   * polymorphic taint,  and we know nothing specific about `obj.x`, then
+                   * we add the same offset `.x` to the polymorphic taint coming from `obj`.
```
emjin: Just to clarify, is this so that when `obj` gets tainted, the taint is split into `obj.x` and `obj.y` being tainted, so that they can later be individually cleaned?

src/tainting/Dataflow_tainting.ml
```diff
@@ -1174,28 +1163,46 @@ let check_tainted_var env (var : IL.name) : Taints.t * Lval_env.t =
...
+        match (arg_exp.e, sig_arg.offset) with
+        | Fetch ({ base = Var obj; _ } as arg_lval), _ ->
+            let lval =
+              {
+                arg_lval with
+                rev_offset = List.rev_append os arg_lval.rev_offset;
+              }
+            in
+            Some (lval, obj)
+        | Record fields, [ o ] -> (
```
emjin, line 1210: Example of this case would be helpful, I'm still a bit confused about what the check `fst fld = fst o.ident` is for. (I have an sense but it's not concrete)

emjin (approved):
Seems reasonable. Checking my understanding: did the `lval_of_sig_arg` change happen because JS lets you do `{a, b, c} = x` whereas Java doesn't?

src/tainting/Dataflow_tainting.ml
IagoAbal: Added this comment which I hope it's helpful:
```ocaml
* For example, given `function foo(o) { sink(o.x); }`, and being '0 the
* polymorphic taint of `o`, this allows us to record that what goes into
* the sink is '0.x (and not just '0). So if later we encounter `foo(obj)`
* where `obj.y` is tainted but `obj.x` is not tainted, we will not
* produce a finding.
```

IagoAbal:
Thanks for the questions!

> Seems reasonable. Checking my understanding: did the `lval_of_sig_arg` change happen because JS lets you do `{a, b, c} = x` whereas Java doesn't?

The `{a, b, c} = x` should be handled by AST_to_IL, I'd need to write some test to make sure it works. 🤔 But in principle it should be translated as:
```js
a = x.a
b = x.b
c = x.c
```

The change to `lval_of_sig_arg` is to handle cases like `foo({x="tainted", y="safe"})`, so when looking at the signature of `foo`, if it says "'0.y goes into a sink", we can resolve that '0.y is referring to `"safe"` in this call and therefore there is no finding. Added a comment.

src/tainting/Dataflow_tainting.ml
emjin: ohhh got it

emjin:
Understanding check failed 😅. The new comments are super helpful, thanks!

PR #7882, opened 2023-05-25
fix: validate rule id

jbergler:
Prior to this change any characters were accepted for a rule `id` which is probably a bit ambitious.
This change limits the rule id to `/^[a-zA-Z0-9._-]*$/` to keep things clean and tidy.

If there is a need in the future we can look at expanding the character set.


brandonspark (approved):

jbergler:
@minusworld in #2574 you added a unicode rule id, but I can't glean from that PR if or why that might be important. Any thoughts on me changing [this](https://github.com/returntocorp/semgrep/pull/2574/files#diff-1b70a4a051810c0f1dc5f33c2dc5200889c66421a0202c9d29cf3907df47c3d8_) test case to not have a unicode rule id.

```
rules:
- id: 寿司
```

```
% rg '寿司'
cli/tests/e2e/rules/nosem-unicode.yaml
2:  - id: 寿司

cli/tests/e2e/targets/advanced_nosem/nosem-unicode.py
1:test_nosem_func()  # nosem: rules.寿司
2:test_nosem_func()  # NOSEM: rules.寿司
3:test_nosem_func()  # nosemgrep: rules.寿司
5:test_nosem_func()  # nOseMgREP: rules.寿司
```

aryx:
cc @dinosaure . rule ids in nosemgrep are now easier to parse :)
sorry for the extra work ...

Issue #7883, opened 2023-05-25
aliengrep: add an option for caseless matching


mjambon:
**Is your feature request related to a problem? Please describe.**

Caseless matching allows matching case-insensitive languages like HTML and derivatives, HTTP headers, English text, etc. It is possible and easy to implement in aliengrep thanks to PCRE having options for case-insensitive matching.

**Describe the solution you'd like**

Support a new option in the `options` section of Semgrep rules:
```
options:
  generic_caseless: true
```

Issue #7881, opened 2023-05-25
aliengrep: leading and trailing ellipses should be anchored


mjambon:
**Describe the bug**

The pattern
```
foo: ...
```

Should match everything in
```
foo: bar
```
rather than just matching `foo:` i.e. it should behave like spacegrep.

**To Reproduce**

```
$ cd semgrep
$ semgrep -c cli/tests/e2e/rules/aliengrep/httpresponse.yaml cli/tests/e2e/targets/aliengrep/httpresponse.txt
...
```
should find a match.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

PR #7885, opened 2023-05-25
Aliengrep integration + rule ID type + 'languages' type

mjambon:
I should have done these other things in different PRs but it was too late by the time I realized how big those changes were. So, here we go. The main changes brought by this branch are:

* `options.generic_engine: aliengrep` in rules allows semgrep to use aliengrep when `generic` is specified in the `languages` list.
* I added some end-to-end tests for aliengrep. I created two issues for a [matching bug](https://github.com/returntocorp/semgrep/issues/7881) and a [missing feature](https://github.com/returntocorp/semgrep/issues/7883) but it's nothing big. Overall, it works.
* The `languages` field is now translated (in OCaml only, not Python) into a `languages` type that distinguishes target selector (langs) from target analyzer (xlang). "generic" now means "generic target selection" and the parsing/matching engine is specified separately. This was done to avoid making things weird for the user but we still have to support "regex" and "none" which both mean "generic target selection" + "regex engine".
* Along the way, I got annoyed that rule IDs were reported as `Common.filename` (= `string`) so I created a type for them.
* I modified `Input_to_core.atd` so that it uses the `Xlang.t` type directly rather than a string that needs to be converted later. This is an example to show that we could use the same mechanism for other types (generally strings or ints that serve as some kind of ID with its own type).

I left comments to explain those things. The Python side gave me a bit of a headache. I tried various things that I reverted, concluding that the least we touch, the better.

I will have to document the features of aliengrep. It shouldn't be too hard because it has a lot in common with spacegrep. For now, there's a readme in `libs/aliengrep`.

test plan: `make test` should work

Uses: https://github.com/returntocorp/semgrep-langs/pull/36

PR checklist:


mjambon:
I just force-added a target file with the `.log` extension (which otherwise is gitignored). Maybe I shouldn't have done this?

semgrep.jsonnet
```diff
@@ -25,62 +25,7 @@ local semgrep_rules = [
...
-        "parsing/tree_sitter/",
-        "metachecking/",
-	# outside src/, not really semgrep code
-	"libs/",
-	"languages/",
-	"tools/",
-        ] + test_code_globs + less_important_code_globs,
-    }
-  }
+#  }
```
mjambon, line 28: Oh, and I removed this because it was counterproductive. I think we should have paused at the time of adding this long list of path exclusions.

src/core/Rule.ml
```diff
@@ -28,7 +28,46 @@ open Ppx_hash_lib.Std.Hash.Builtin
...
+  let of_string x = x
+  let to_string_list x = x
+  let of_string_list x = x
+  let compare = String.compare
+
+  let ends_with r ~suffix:inc_or_exc_rule =
+    r = inc_or_exc_rule || String.ends_with ~suffix:("." ^ inc_or_exc_rule) r
+end
+
+type rule_id = ID.t [@@deriving show, eq]
```
mjambon, line 71: We could move this ID module to its own file if people prefer it that way. I like either way.

mjambon:
I think it's wrong to have language options in the global `options` object, since various steps of the rule can specify languages, we should be able to support different language options for each step. This calls for a `language-options` section next to `languages`. We should still merge this PR ASAP (because it keeps breaking due to merge conflicts and this stuff is unused and experimental so it's fine to change it later) but I think I'll move the `generic_`* options to `language-options` later.

interfaces/Input_to_core.atd
```diff
@@ -18,13 +18,15 @@
...
   path: string (* source file *);
-  (* examples: "ocaml", "python", but also "generic" or "regexp".
-   * less: would be great to use Lang.t here, but atd does not
-   * allow (by design) to use types defined in other files.
-   *)
-  language: string;
+  (* Must be a valid "language" as defined in Xlang.mli.
+     examples: "ocaml", "python", but also "spacegrep" or "regexp".
+     less: would be great to use Lang.t here, but atd does not
+     allow (by design) to use types defined in other files. *)
```
aryx, line 28: well actually how far is your module system for ATD? Did you finish it?

cli/tests/e2e/rules/aliengrep/dockerfile.yaml
```diff
@@ -0,0 +1,17 @@
+rules:
+  - id: double-root
+    languages:
+      - generic
+    options:
+      generic_engine: aliengrep
+    patterns:
+      - pattern-either:
```
aryx: why use a pattern-either when you have just one choice?

cli/tests/e2e/rules/aliengrep/html.yaml
```diff
@@ -0,0 +1,12 @@
+rules:
```
aryx, line 1: why put those tests in cli/tests/e2e/rules? Why not simply in tests/rules/?

cli/tests/e2e/test_aliengrep.py
```diff
@@ -0,0 +1,54 @@
+# Test aliengrep integration in Semgrep rules
+import pytest
+from tests.fixtures import RunSemgrep
+
+
+@pytest.mark.osempass
```
aryx, line 6: yeah! osempass!

semgrep.jsonnet
aryx: Ok

src/core/Rule.ml
```diff
@@ -28,7 +28,46 @@ open Ppx_hash_lib.Std.Hash.Builtin
...
+  (*
+     Rule ids are prepended with the `path.to.the.rules.file.`, so
+     when comparing a rule (r) with the rule to be included or excluded,
+     allow for a preceding path
+
+       "path.to.foo.bar" ~suffix:"foo.bar" -> true
+       "foo.bar" ~suffix:"foo.bar" -> true
+       "xfoo.bar" ~suffix:"foo.bar" -> false
+
+     TODO we may want to allow globs for rules
```
aryx: we do already, see the recent extension @emjin did for extract mode where people want to specify the set of rules one can with the extract rule.

src/core/Rule.ml
aryx: It's fine

src/engine/Unit_engine.ml
```diff
@@ -349,7 +349,9 @@ let regression_tests_for_lang ~polyglot_pattern_path ~with_caching files lang =
                    match_pattern ~lang
                      ~hook:(fun { Pattern_match.range_loc; _ } ->
                        let start_loc, _end_loc = range_loc in
-                       E.error "test pattern" start_loc "" Out.SemgrepMatchFound)
+                       E.error
+                         (Rule.ID.of_string "test pattern")
```
aryx, line 353: actually space are not allowed in rule ids. Unicode chars have also been recently banned by Jonas.

aryx (approved):
Nice. Lots of cool stuff in addition to aliengrep support.

interfaces/Input_to_core.atd
mjambon: I stopped because it got unreasonably complicated to add to atdgen, while it was rather straightforward to add to atdpy. This is when I decided we should implement a new atdml command using the same model as atdpy or atdts:
* this is the modules PR, which works fine for atdpy and atdts but contains massive unfinished refactoring for atdgen that was an attempt to understand the code better: https://github.com/ahrefs/atd/pull/297. I might merge the atdpy/atdts implementation first, after which we'll have imports for Python and TypeScript but not for OCaml...
* issue pushing for an atdml: https://github.com/ahrefs/atd/issues/313

cli/tests/e2e/rules/aliengrep/html.yaml
mjambon: I replicated what we had for spacegrep. We need to test pysemgrep especially because of the generic/generic_engine translation into spacegrep or aliengrep.

What's the status of tests/rules? Are they used for testing both pysemgrep and semgrep-core?

cli/tests/e2e/rules/aliengrep/dockerfile.yaml
mjambon: I copied these tests from spacegrep. I'll simplify this.

src/core/Rule.ml
mjambon: This is an old comment I copied from somewhere else. Removing.

src/engine/Unit_engine.ml
mjambon: I asked @jbergler about the spec since I'm not sure where to look. I might leave a TODO for now and merge.

PR #7896, opened 2023-05-26
feat(interactive): turbo mode

brandonspark:
## What:
This PR adds in turbo mode for Interactive Mode.

## Why:
It's really friggin cool and nice to have it be instantly responsive

## How:
We use threads to compute the matches concurrently, while taking input from the user. This allows (relatively)
smooth display of the query bar changing, while not requiring us to block while computing the matches from the query, which can be done relatively fast, but not as fast as people can type.

## Problems:
There are a few problems which currently exist
1) The first query is always the longest, due to AST caching. This only takes about a second or two, but `Lazy.force` is not thread-safe. This means that if you instantly start typing more than 1 character, threads will start complaining about being killed because they are trying to concurrently force the ASTs which are currently being parsed.
2) If you type too fast, Interactive Mode will stall out and stop updating the display, until your next keypress. This is probably due to the way that we enter the `Term.event` call, I think that call is blocking when we should have waited a bit longer for more input, or something.
3) Originally, I wanted to use the `Term.pending` function, which is supposed to tell you when `Term.event` is blocking or not. However, I did a lot of experimentation, and found that it was incredibly unreliable, and very often would just always tell you that the call would not block, even when it would. So that's why I resorted to manually polling `stdin` using `select`.
4) The query bar can lag sometimes

Otherwise, works pretty well.

## Test plan:
https://asciinema.org/a/uyIpst16czMDgVdQT1F2KQNwm

PR checklist:


src/osemgrep/cli_interactive/Interactive_subcommand.ml
```diff
@@ -56,11 +57,19 @@ type matches_by_file = {
 type state = {
   xlang : Xlang.t;
   xtargets : Xtarget.t list;
-  file_zipper : matches_by_file Pointed_zipper.t;
-  cur_line_rev : char list;
+  file_zipper : matches_by_file Pointed_zipper.t ref;
+      (** A ref, because our matches might change at any time when the
+          thread which is doing matches completes.
+          The thread needs to be able to communicate the new file zipper
+          back to the main process.
```
aryx, line 64: the main thread

aryx (approved):

PR #7905, opened 2023-05-26
fix: Match metavariables against builtins in typed metavariables

nmote:
I broke this in #7868. I added a test that illustrates the issue. Previously we'd match this case using `m_name` since both the pattern and target were `TyN`s. Now, `String` ends up being a `Type.Builtin Type.String` (`Integer` is not currently considered a builtin, but probably should be).

Test plan: Automated tests

PR checklist:


nmote:
Actually, I am rethinking this

nmote:
I've rethought it a bit, and it turns out to not be quite true that metavariable bindings don't escape from the type portion of the typed metavariable. I've updated my comment accordingly. Regardless, this fixes a real problem, and while it doesn't fix all the problems with typed metavariables it's still an improvement, so I'm marking it ready for review.

I'll sort out some of the other issues I've discovered while looking into this in future PRs.

aryx (approved):

Issue #7787, opened 2023-05-12
Ruby: taint not connecting source to sink when method call is done with omitted parenthesis


joaxcar:
**Describe the bug**
In Ruby you can omit the parenthesis on a method call, and in particular if the method does not take any arguments you can call it by just its name, like this
```ruby
def my_func
   "hello"
end

my_func
```
The `my_func` in the end is an expression and a function call and the program will return "hello". When writing a taint query in Semgrep for Ruby I can not get Semgrep to follow this data path from source to sink

<img width="359" alt="ruby" src="https://github.com/returntocorp/semgrep/assets/37298485/076c08b9-72b4-4d38-9116-62e7c3c4d408">

this is somewhat related to https://github.com/returntocorp/semgrep/issues/5222 but more serious. Also see https://ruby-for-beginners.rubymonstas.org/bonus/parentheses.html

**To Reproduce**
Test this in the playground (https://semgrep.dev/s/pJld)
Query
```yaml
rules:
  - id: untitled_rule
    languages:
      - ruby
    mode: taint
    message: $DATA
    pattern-sinks:
      - patterns:
          - pattern: $T
          - pattern-inside: print($T)
    pattern-sources:
      - patterns:
          - pattern: |
              def source
              $DATA
              end
```
and the ruby file
```ruby
# The source
def source
  data
end

# This one will fail to match, but is a valid call
def sink
    value = source
    print(value)
end

# The rest here will hit
def sink
    value = source()
    print(value)
end

def sink
    value = source ""
    print(value)
end

def sink
    value = source 1
    print(value)
end

def sink
    value = source ()
    print(value)
end

```

**Expected behavior**
All the sinks above should hit the query, allowing taint to follow though method calls without parenthesis. 

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [x] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Use case**
What will fixing this bug enable for you?

Find bugs in Ruby code that are now being missed as this is a regular pattern

I am new to this so maybe I am doing something wrong, but this is a blocker for my workflow

kopecs:
Seems like this was fixed in https://github.com/semgrep/semgrep/pull/7906.

PR #7906, opened 2023-05-27
taint analysis #7787: Check function calls without parameters or parenthesis in Ruby

ihji:
This pull request aims to address the issue raised in #7787 by incorporating a specific handling for Ruby function calls that lack parentheses or parameters during the IL transformation.

test plan:
See returntocorp/semgrep-proprietary#748

PR checklist:

### DIFF
--- a/changelog.d/gh-7787.fixed
+++ b/changelog.d/gh-7787.fixed
@@ -1,1 +1,1 @@
-taint analysis: Check function calls without parameters or parenthesis in Ruby
+Pro (taint analysis): Check function calls without parameters or parenthesis in Ruby
--- a/changelog.d/gh-7787.fixed
+++ b/changelog.d/gh-7787.fixed
@@ -1,0 +1,1 @@
+taint analysis: Check function calls without parameters or parenthesis in Ruby
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -33,6 +33,10 @@ let logger = Logging.get_logger [ __MODULE__ ]
 (*****************************************************************************)
 (* Types *)
 (*****************************************************************************)
+module IdentSet = Set.Make (String)
+
+type ctx = { entity_names : IdentSet.t }
+
 type env = {
   lang : Lang.t;
   (* stmts hidden inside expressions that we want to move out of 'exp',
@@ -45,10 +49,19 @@ type env = {
   *)
   break_labels : label list;
   cont_label : label option;
+  ctx : ctx;
 }
 
+let empty_ctx = { entity_names = IdentSet.empty }
+
 let empty_env (lang : Lang.t) : env =
-  { stmts = ref []; break_labels = []; cont_label = None; lang }
+  {
+    stmts = ref [];
+    break_labels = [];
+    cont_label = None;
+    ctx = empty_ctx;
+    lang;
+  }
 
 (*****************************************************************************)
 (* Error management *)
@@ -241,6 +254,9 @@ let mk_class_constructor_name (ty : G.type_) cons_id_info =
     when Option.is_some !(cons_id_info.G.id_resolved) ->
       Some (G.Id (id, cons_id_info))
   | __else__ -> None
+
+let add_entity_name ctx ident =
+  { entity_names = IdentSet.add (H.str_of_ident ident) ctx.entity_names }
 
 (*****************************************************************************)
 (* lvalue *)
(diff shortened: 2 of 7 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -1435,8 +1435,17 @@ let check_function_call_arguments env args =
  * report the finding too (by side effect). *)
 let check_tainted_instr env instr : Taints.t * Lval_env.t =
   let check_expr env = check_tainted_expr env in
-  let check_instr = function
-    | Assign (_, e) -> check_expr env e
+  let rec check_instr = function
```
IagoAbal: ~We would need a test case and a changelog too.~
Need a changelog entry.

src/tainting/Dataflow_tainting.ml
```diff
@@ -1435,8 +1435,17 @@ let check_function_call_arguments env args =
...
+    | Assign (_, e) -> (
+        match e.e with
+        (* Considering that Ruby permits function calls without parameters or
+         * parentheses, we are checking whether the identifier has a function
+         * signature specifically for Ruby. *)
+        | Fetch _
```
IagoAbal: Not very fond of doing it here. Could you try doing this hack in AST_to_IL? It would be almost the same thing, but then you generate a `Call` instead of an `Assign` when these conditions are met.  Ideally we check the type of the lval (for Pro) or the existence of a funcdef for this lval, but you can leave that as a TODO for now. Ideally this is fixed already in the Generic AST indeed, but it's OK having a temporary hack in AST_to_IL.

IagoAbal:
Remember to add a test plan, see the checklist:
>  Tests included or PR comment includes a reproducible test plan

I added one for you.

changelog.d/gh-7787.fixed
```diff
@@ -0,0 +1 @@
+taint analysis: Check function calls without parameters or parenthesis in Ruby
```
IagoAbal: I guess this a Pro feature (inteproc), maybe add `Pro:` in the changelog to make that clear.

src/analyzing/AST_to_IL.ml
```diff
@@ -33,6 +33,10 @@ let logger = Logging.get_logger [ __MODULE__ ]
 (*****************************************************************************)
 (* Types *)
 (*****************************************************************************)
+module IdentSet = Set.Make (String)
+
+type ctx = { entity_names : IdentSet.t }
```
IagoAbal, line 38: Why adding a `ctx` type ? Can't we just add `entity_names : IdentSet.t` to the `env`?

src/analyzing/AST_to_IL.ml
```diff
@@ -615,7 +631,21 @@ and expr_aux env ?(void = false) e_gen =
...
+      let exp = mk_e (Fetch lval) eorig in
+      let ident_function_call_hack exp =
+        (* Taking into account Ruby's ability to allow function calls without
+         * parameters or parentheses, we are conducting a check to determine
+         * if a function with the same name as the identifier exists, specifically
+         * for Ruby. *)
+        match lval with
+        | { base = Var { ident; _ }; _ }
+          when env.lang =*= Lang.Ruby
+               && IdentSet.mem (H.str_of_ident ident) env.ctx.entity_names ->
```
IagoAbal, line 643: Sorry, I thought you would leave this as a TODO so I didn't think much about it. But since this is for Pro, I would expect Pro to have set `ident.id_info.id_type` to `TyFun _`, so we wouldn't need to keep any `ctx` here. Also, that way it would work for functions that are defined outside the current file.

IagoAbal: Hmm, never mind, let's use an `IdentSet` for now. Otherwise this wouldn't work with `--pro-intrafile`.

IagoAbal (approved):
Remember to update the PR to Pro and make sure the tests run.

src/analyzing/AST_to_IL.ml
ihji: My intention was to provide clarity by consolidating the information from the inter-procedural context within a distinct record. This approach not only ensures that the `ctx` field contains relevant data but also allows for easier addition of additional information in the future, if needed.

src/tainting/Dataflow_tainting.ml
ihji: Fixed.

src/tainting/Dataflow_tainting.ml
ihji: Fixed.

changelog.d/gh-7787.fixed
ihji: Fixed.

PR #7918, opened 2023-05-30
js: fix libyaml bindings

tpetr:
This PR fixes memory alignment issues that was preventing YAML parsing from working properly. TL;DR: `ocaml-yaml` uses `ocaml-ctypes` to generate bindings for the `libyaml` library. Since this is done at compile time, the memory offsets are calculated based off of the build machine (64-bit address space --> 8 byte pointer size). WebAssembly, however, is 32-bit (4 bytes), and this mismatch prevented us from interfacing with the library properly.

To fix, we override the memory offsets in JavaScript (exploiting its inherent mutability) at startup. This is a little weird, is much faster than figuring out how to cross-compile `ocaml-yaml` against a 32-bit architecture.

There is a weird off-by-one issue with the index of the end token in Yaml_to_generic, but I'll fix in a separate PR if it proves to be an issue.

test plan:
- make test
- confirm `js/examples/yaml.html` works properly

### DIFF
--- a/js/engine/Main.ml
+++ b/js/engine/Main.ml
@@ -5,6 +5,52 @@ open Js_of_ocaml
    (see companion setter in Semgrep_js_shared.ml) *)
 external get_jsoo_mount_point : unit -> 'any list = "get_jsoo_mount_point"
 
+(* semgrep uses ocaml-yaml, which uses ocaml-ctypes to use the libyaml C library
+   this is problematic for webassembly because it doesn't match the architecture
+   of the build machine (e.g. most computers these days have a 64-bit address space,
+   but WASM is currently 32-bit). cross-compilation ended up being too much work, so
+   instead we exploit javascript's mutability to correct some memory offsets for WASM *)
+external override_yaml_ctypes_field_offset :
+  ('a, 'b) Yaml_bindings.T.field -> int -> unit
+  = "override_yaml_ctypes_field_offset"
+
+let fix_libyaml_field_offsets_for_wasm () =
+  (* mark *)
+  override_yaml_ctypes_field_offset Yaml_types.M.Mark.line 4;
+  override_yaml_ctypes_field_offset Yaml_types.M.Mark.column 8;
+
+  (* event *)
+  override_yaml_ctypes_field_offset Yaml_types.M.Event.data 4;
+  override_yaml_ctypes_field_offset Yaml_types.M.Event.start_mark 32;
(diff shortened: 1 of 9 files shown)


aryx:
Could be worth asking on discuss.ocaml.org if there's an easier way to fix the issue.

js/engine/Main.ml
```diff
@@ -5,6 +5,52 @@ open Js_of_ocaml
...
+   of the build machine (e.g. most computers these days have a 64-bit address space,
+   but WASM is currently 32-bit). cross-compilation ended up being too much work, so
+   instead we exploit javascript's mutability to correct some memory offsets for WASM *)
+external override_yaml_ctypes_field_offset :
+  ('a, 'b) Yaml_bindings.T.field -> int -> unit
+  = "override_yaml_ctypes_field_offset"
+
+let fix_libyaml_field_offsets_for_wasm () =
+  (* mark *)
+  override_yaml_ctypes_field_offset Yaml_types.M.Mark.line 4;
```
aryx, line 19: How did you end up with all those magic constants below?
you disassembled the resulting WASM?
If tomorrow the bus hit you, how someone else can update this code if say the libyaml C library changes a little bit?

aryx (approved):
Lots of magic constants and hardcoded array indexes ...
Would be better to document those at least.

tpetr:
> Could be worth asking on discuss.ocaml.org if there's an easier way to fix the issue.

Good idea, if folks have a cleaner solution I'll apply it in another PR 👍 

edit: posted question here: https://discuss.ocaml.org/t/using-ocaml-ctypes-with-webassembly/12279

js/engine/Main.ml
tpetr: Yep, when I was investigating this I created a macro that I used to print the offset of every struct field. I'll restructure and add more comments to make it clearer what's going on here.

js/engine/Main.ml
```diff
@@ -5,6 +5,52 @@ open Js_of_ocaml
...
+  override_yaml_ctypes_field_offset Yaml_types.M.Mark.column 8;
+
+  (* event *)
+  override_yaml_ctypes_field_offset Yaml_types.M.Event.data 4;
+  override_yaml_ctypes_field_offset Yaml_types.M.Event.start_mark 32;
+  override_yaml_ctypes_field_offset Yaml_types.M.Event.start_mark 44;
```
tpetr, line 25: argh, typo

PR #7923, opened 2023-05-31
feat(interactive): ascii art resizing

brandonspark:
## What:
This PR makes it so the Interactive Mode UI resizes based on the size of the terminal.

## Why:
Previously, the ASCII art always displayed, and was at a fixed width and height, meaning it could stretch off the edges if your terminal was too small. Now, we try to resize to either a smaller ASCII image, or if the terminal is really really small, we just don't display the ASCII images.

## How:
Added functions which get the ASCII art, which depend on the width and height. Since the Semgrep logo is wider than it is tall, width is used as the primary metric for selection. The ghost is taller than it is wide, so height is used as the primary metric.

## Test plan:
This is what I got by making the terminal as small as it could go without losing any of the ASCII images:
image

image

image

image

As you can see, there's generous enough whitespace that this should be good.

PR checklist:


aryx (approved):
At some point we should move this ascii art data out of Interactive_subcommand.ml (I like small files) which is getting big.

PR #7929, opened 2023-05-31
fix: Remove special-case for typed-metavar matching

nmote:
Previously, if a typed metavariable was matched against an identifier, it would go through the old `m_type` matching codepath rather than the new type inference codepath. For consistency and simplicity, I have removed that special case and consolidated on the new type inference and matching.

This required me to make some improvements to the new type inference and new matching in order to pass tests. Notably, metavariables inside of the types in typed metavariables are now bound. Since the AST to which they are bound is the result of type inference, it is synthetic. I adopted the pattern pioneered in https://github.com/returntocorp/semgrep/pull/6900 of using a local token from the AST to associate a synthetic AST with a real program location.

The change in test behavior is because we can now match on unresolved types. The type we infer for `MessageDigest` is just `UnresolvedName "MessageDigest"` (in a simplified representation). As noted in `Typing.type_of_name`, this should most likely be `Typeof "MessageDigest"`, but we conflate class instances with the type of the class itself throughout our type inference infrastructure, with this result. I also reformatted that test since I found the mix of tabs and spaces confusing.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/src/core/Type.ml
+++ b/src/core/Type.ml
@@ -107,8 +107,13 @@ let rec to_name_opt lang ty =
   | Pointer ty when lang =*= Lang.Go -> to_name_opt lang ty
   | _else_ -> None
 
-let mkt str =
-  G.TyN (G.Id ((str, Tok.unsafe_fake_tok str), G.empty_id_info ())) |> G.t
+let mkt ?tok str =
+  let tok =
+    match tok with
+    | Some tok -> tok
+    | None -> Tok.unsafe_fake_tok str
+  in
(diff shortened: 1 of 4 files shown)


nmote:
I'm labeling this a fix because it does improve behavior, but I'm not including a changelog entry because the previous deficiencies in typed metavariable matching were largely masked by the special case that I removed here, and I think it's unlikely that people will notice. This is mostly a refactor.

aryx (approved):

PR #7931, opened 2023-05-31
clean up yaml memory offset overrides

tpetr:
- Isolates and hopefully clarifies my yaml memory offset hack
- Re-adds `print_event()` to libyaml so that others can inspect the struct offsets in the future

[This discussion thread](https://discuss.ocaml.org/t/using-ocaml-ctypes-with-webassembly/12279/7) has convinced me that the right solution here is to cross-compile, but my opam-fu is weak, so let's continue rolling with my gross hack in the short-term and we can revisit once the dust settles (thanks @yallop for the tips)

test plan:
- dune build && make test
- confirm js/examples/yaml.html works


js/libyaml/extra.c
```diff
@@ -11,3 +11,82 @@ YAML_DECLARE(int)
 get_sizeof_yaml_event_t() {
 	return sizeof(yaml_event_t);
 }
+
```
aryx, line 14: would be nice to add a comment about extra.c in the README about your modification
and what is the purpose of this mysterious print_event.

aryx (approved):

PR #7932, opened 2023-05-31
feat(tests): pointed zipper expect tests, new testing script

brandonspark:
## What:
This PR does some renaming for Interactive Mode, and adds new "expect tests" for Pointed_zipper. To make sure that they display nicely, there is a new script `run-core-test.sh`, because otherwise the expect tests tend to run before the core tests, and get drowned in output. Also, we wanted to add a message warning to use `dune promote` to apply expect tests.

## Why:
See this post: https://blog.janestreet.com/the-joy-of-expect-tests/

Writing expect tests is nice because you simply print what you expect to see, and then can programmatically apply the diff if it looks OK. Otherwise, you might need to do very annoying Boolean conditioning to get `true` or `false` on what you expect, and overall it would be not a nice workflow.

What's nice is you can set up a great deal of expect tests, run the diff once, and then look over it, without needing to copy paste anything from your terminal.

## How:
Added a new testing script so we could test the output of `dune runtest src/osemgrep`, as well as run the two separately. Added a bunch of expect tests for `Pointed_zipper`.

## Test plan:
`make test`

Here's what it looks like if an expect test fails on `make test`:
image

PR checklist:


brandonspark:
@aryx @mjambon any thoughts?

Makefile
```diff
@@ -254,7 +254,7 @@ core-test: core build-spacegrep
 	# from the directory of the checkou
 	./_build/default/src/tests/test.exe --show-errors --help 2>&1 >/dev/null
 	$(MAKE) -C libs/spacegrep test
-	dune runtest -f --no-buffer
+	./scripts/run-core-test.sh
```
mjambon: Please don't use file extensions for executables because they're command names and a `.sh` or `.py` extension is an abstraction leak: the user should not have to care about how the command is implemented (see https://google.github.io/styleguide/shellguide.html#s2-shell-files-and-interpreter-invocation). Note that Python programmers have difficulties with these concepts (also, the language doesn't have separate interfaces and many of their tools start with `py` even if the application has nothing to do with Python).

src/osemgrep/cli_interactive/Unit_pointed_zipper.ml
```diff
@@ -0,0 +1,150 @@
...
+  let zipper = PZ.move_right zipper in
+  emit zipper;
+  Common.(
+    pr2
+      (spf "absolute: %d, relative: %d"
+         (PZ.absolute_position zipper)
+         (PZ.relative_position zipper)));
+  [%expect {|
+    [3 | 3, 4, (5)]
+    absolute: 4, relative: 2 |}]
```
mjambon: looks good!

mjambon (approved):

PR #7934, opened 2023-06-01
Implement anchored ellipsis where it makes sense

### LINKED ISSUE #7881: aliengrep: leading and trailing ellipses should be anchored

mjambon:
This matches the spacegrep behavior.

For example, in multiline mode, `...` alone matches the whole input rather than matching the empty sequence many times. In single-line mode, `...` alone matches a whole line for each line in the input. In other contexts where `...` or `....` is sandwiched between other elements, the behavior is still to prefer the shortest match possible.

Fixes https://github.com/returntocorp/semgrep/issues/7881

test plan: `make test`

PR checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,12 +249,20 @@ test:
 core-test: core build-spacegrep
 	# The test executable has a few options that can be useful
 	# in some contexts.
+	$(MAKE) build-core-test
 	dune build ./_build/default/src/tests/test.exe
 	# The following command ensures that we can call 'test.exe --help'
-	# from the directory of the checkou
+	# from the directory of the checkout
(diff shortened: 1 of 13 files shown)


mjambon:
I still haven't written the documentation for any of aliengrep.

PR #7936, opened 2023-06-01
tainting: Add options to ignore taint based on types

IagoAbal:
In many cases Boolean or integer values do not pose a security risk even if they come from a tainted source. For example, a user input of type integer does not pose a risk of SQLi.

Also checked types to prevent attaching an offset to poly taint when the offset corresponds to a class method.

Also fixed type inference to take some corner cases into consideration.

test plan:
make test # new tests

PR checklist:

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -743,7 +743,7 @@ let fix_poly_taint_with_field env lval st =
                    || not
                         (Stdcompat.String.starts_with ~prefix:"get"
                            (fst n.ident)) ->
-                (* We have an l-value like `o.f` where `f` is has a function type,
+                (* We have an l-value like `o.f` where `f` has a function type,
                  * so it's a method call. We will only fix poly-taint in such case
                  * when it's an unresolved Java `getX` method. In any other case,
                  * we do nothing here. *)
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -458,6 +458,33 @@ let partition_mutating_sources sources_matches =
          m.spec.source_by_side_effect && is_exact m)
 
 (*****************************************************************************)
+(* Types *)
+(*****************************************************************************)
+
+let type_of_expr env e =
+  match e.eorig with
+  | SameAs eorig -> Typing.type_of_expr env.lang eorig |> fst
+  | _else -> Type.NoType
+
+(* We only check this at a few key places to avoid calling `type_of_expr` too
+ * many times which could be bad for perf (but haven't properly benchmarked):
+ * - assignments
+ * - return's
+ * - function calls
+ * TODO: Ideally we add an `e_type` field and have a type-inference pass to
+ *  fill it in, so that every expression has its known type available without
+ *  extra cost.
+ *)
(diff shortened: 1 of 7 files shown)


src/typing/Typing.ml
```diff
@@ -53,15 +53,27 @@ let rec type_of_expr lang e : G.name Type.t * G.ident option =
       let t2, _id = type_of_expr lang e2 in
       let t =
         match (t1, op, t2) with
+        | Type.Builtin Type.Int, (G.Plus | G.Minus (* TODO more *)), _
+        | _, (G.Plus | G.Minus (* TODO more *)), Type.Builtin Type.Int
+          when not (Lang.is_js lang) ->
+            (* THINK: Is this correct in every other language that we support? *)
```
nmote: No, it isn't. At the very least, Java also does implicit coercions to string when an int and a string are operands to `+`. I bet other JVM languages do this too.

src/typing/Typing.ml
```diff
@@ -53,15 +53,27 @@ let rec type_of_expr lang e : G.name Type.t * G.ident option =
...
-            | G.GtE | G.And | G.Or ),
+            | G.GtE | G.In | G.NotIn | G.Is | G.NotIs | G.And ),
             _ ) ->
             Type.Builtin Type.Bool
+        | Type.Builtin Type.Bool, G.Or, _
+        | _, G.Or, Type.Builtin Type.Bool ->
+            Type.Builtin Type.Bool
+        | _, G.Or, _ when lang =*= Lang.Java ->
+            (* E.g. in Python you can write `x or ""` to mean `""` in case `x` is `None`.
+             * THINK: Is there a similar idiom involving `and`/`&&` ? *)
```
nmote, line 77: Yeah there is actually. I think most languages with truthiness and falsiness actually can return an operand to `||` and `&&` rather than a fresh boolean value. This change is a step towards better correctness but I think we need to do more. Most likely, for `&&` and `||`, we should return the least upper bound of the two operands for such languages.

Anyway, your change is an improvement so I'm not asking you to do this here. Just musing.

src/tainting/Dataflow_tainting.ml
```diff
@@ -457,6 +457,33 @@ let partition_mutating_sources sources_matches =
...
+  | _else -> Type.NoType
+
+(* We only check this at a few key places to avoid calling `type_of_expr` too
+ * many times which could be bad for perf (but haven't properly benchmarked):
+ * - assignments
+ * - return's
+ * - function calls
+ * TODO: Ideally we add an `e_type` field and have a type-inference pass to
+ *  fill it in, so that every expression has its known type available without
+ *  extra cost.
```
nmote, line 476: I'm a bit concerned about the memory cost of this but I agree that it would be really nice to have if we can afford it.

nmote (approved):
Nice, I really like this.

src/tainting/Dataflow_tainting.ml
IagoAbal: Yeah large files may be a problem.

src/typing/Typing.ml
IagoAbal: True, I should have known about Java.

IagoAbal: Thanks for flagging this!

PR #7942, opened 2023-06-02
feat(Java): Guess types when we can't resolve them definitively

nmote:
We currently use type inference to handle typed metavariables. For example, a Java pattern might include `(boolean $X)` which will match any expression that is inferred to have the type `boolean`.

This PR extends type inference to include heuristics which we use to make a best guess for types of expressions whose types we cannot infer using traditional means. This typically happens when there is an unresolved name involved in the type.

While the heuristics are based on naming patterns in the Java standard library, they are not limited to the standard library. Nor do we intend to replicate the entire Java standard library as heuristics. Instead, we should include heuristics when we believe that they correctly describe the type of an expression with high confidence, whether they are based on a convention in the standard library or other common conventions.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/tests/patterns/java/metavar_typed_bool.java
+++ b/tests/patterns/java/metavar_typed_bool.java
@@ -1,3 +1,5 @@
+import java.util.*;
+
 public class metavar_typed_bool {
   public static void main(String[] args) {
     boolean x = true;
@@ -34,6 +36,7 @@ public class metavar_typed_bool {
     System.out.println(!(1 == 2));
 
     metavar_typed_bool.overloaded();
+    metavar_typed_bool.stdlib();
   }
 
   public static void overloaded() {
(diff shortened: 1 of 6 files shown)


tests/patterns/java/metavar_typed_bool.java
```diff
@@ -68,4 +71,40 @@ public static void overloaded() {
...
+
+    Map<String, String> m1 = new HashMap<>();
+    m1.put("x", "y");
+    // OK:
+    System.out.println(m1.get("x"));
+
+    Map<String, Boolean> m2 = new HashMap<>();
+    m1.put("x", false);
+    // MATCH:
+    System.out.println(m2.get("x"));
```
IagoAbal, line 108: Does thing handle Java for-each loops or will that go into another PR?

IagoAbal (approved):

tests/patterns/java/metavar_typed_bool.java
IagoAbal: Also need to handle `getX` methods. Although maybe this probably needs to be done in Pro.

nmote: No, it doesn't handle for-each loops. And yeah, I think the `getX` heuristics need to go into Pro since we need detailed information about the resolved type, not just its name.

PR #7946, opened 2023-06-04
fix incorrect yaml positions

tpetr:
YAML parsing was fixed in https://github.com/returntocorp/semgrep/pull/7918 and https://github.com/returntocorp/semgrep/pull/7931, but I didn't catch that node start/end positions were incorrect. This was because in `ctypes_read()` we were returning a JS int when we should have been returning `UInt32` instead.

test plan:
- confirmed locally that the playground works properly w/ a yaml pattern
- added unit test to confirm yaml check works and returns correct positions
- `dune build && make test` succeeded


brandonspark (approved):

src/parsing/yaml_to_generic.ml
```diff
@@ -132,8 +132,7 @@ let mk_bracket
         _;
       } ) v env =
   (* The end index needs to be adjusted by one because the token is off *)
-  (* TODO: figure out why we get an off-by-one with jsoo *)
-  let e_index' = if !Common.jsoo then e_index else e_index - 1 in
+  let e_index' = e_index - 1 in
```
aryx, line 135: That's how you realized there was a int32 vs int64 issue with libyaml?

tpetr: Not quite, the libyaml 32-bit vs. 64-bit realization was me staring at the memory offsets for long enough, and then realizing that `int` and `size_t` are the machine word size.

In this case, the ctypes glue code was incorrectly creating `size_t` primitives as a JS number when it should have been jsoo's `UInt32` type. I didn't immediately catch this issue because the type mismatch resulted in the value being equivalent to zero (in a perfect world an exception would have been thrown)

PR #7950, opened 2023-06-05
add workflow for uploading JS artifacts to S3

tpetr:
- removes NPM packaging from build-test-javascript workflow (not currently using & easy to re-add later)
- uploads semgrep JS assets to an S3 bucket so that we can integrate into the app for turbo more
- misc GHA artifact fixes (wasn't tagging w/ git sha in a couple places)

test plan:
- GHA workflows are passing


apt-itude (approved):
LGTM

PR #7956, opened 2023-06-06
feat: Additional heuristics for Java type inference

nmote:
We still don't intend to embed the entire Java standard library, but I've added a few more common methods that will be useful heuristics for type inference. I've also added a few more tests to spot check them.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/src/core/Type.ml
+++ b/src/core/Type.ml
@@ -157,20 +157,23 @@ let builtin_type_of_string _langTODO str =
   | "number" -> Some Number
   | __else__ -> None
 
-(* coupling: Inverse of builtin_type_of_string
- *
- * TODO: Check lang to get proper builtin type for more languages *)
-let ast_generic_type_of_builtin_type ?tok lang t =
+(* TODO: Check lang to get proper builtin type for more languages *)
+let name_of_builtin_type lang t =
   match (lang, t) with
-  | _, Int -> mkt ?tok "int"
-  | _, Float -> mkt ?tok "float"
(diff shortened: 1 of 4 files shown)


nmote:
Looks like #7942 still hasn't gone out, and the changelog entry for that basically covers what I've done here, too.

src/core/Type.ml
```diff
@@ -157,20 +157,23 @@ let builtin_type_of_string _langTODO str =
...
-  | _, Float -> mkt ?tok "float"
-  | Lang.Java, String -> mkt ?tok "String"
-  | _, String -> mkt ?tok "string"
-  | Lang.Java, Bool -> mkt ?tok "boolean"
-  | _, Bool -> mkt ?tok "bool"
+  | _, Int -> "int"
+  | _, Float -> "float"
+  | Lang.Java, String -> "String"
+  | _, String -> "string"
+  | Lang.Java, Bool -> "boolean"
```
IagoAbal, line 167: Why `"boolean"` here and not `"bool"` or `"Boolean"` (capitalized)? Why no special case for Java+`Int` ? The special cases probably deserve some comment.

IagoAbal (approved):
Nice!

IagoAbal:
I'll be merging this since I have to refactor some of this code I think.

src/core/Type.ml
nmote: The primitive boolean type in Java is called `boolean`. The primitive integer type in Java is called `int`. So that's why.

PR #7967, opened 2023-06-06
tainting: Filter parameters taint by type

IagoAbal:
If a parameter has Boolean or number type, and the correspoding option is enabled, we don't assign any taint to that parameter (polymorphic or due to a default value).

Also, tweaked `fix_poly_taint_with_field` to check the type of fields, and except for unresolved Java getter methods, we attach an offset to polymorphic taint if the offset as a function type (i.e. it is a method).

Note that I added a typing traversal to infer whether a field is a method and thus assign it a function type. This conflicted with the typing heuristics which only triggered when a method had no type assigned. Decided to move the heuristics to be part of this typing traversal.

test plan:
make test # new test

PR checklist:


nmote:
It looks like `type_of_expr` no longer includes the typing heuristics? Do the heuristics still work with typed metavariables in search mode or only taint now?

IagoAbal:
> It looks like `type_of_expr` no longer includes the typing heuristics? Do the heuristics still work with typed metavariables in search mode or only taint now?

They work in all modes. We have a typing visitor that runs through the entire program and fills in the `id_type` of names used as method calls. If they match one of the heuristics we give them whatever type has been "guessed", otherwise we give them a "TyFun [todo] todo"-kind of type (simply to know that they are functions). This is to avoid tracking polymorphic taint on object methods.

nmote (approved):
Oh I see, yeah it's right there in `Parse_target.ml`. Looks good!

PR #7973, opened 2023-06-06
fix(cli): Allow diff scans to happen in dirty repos

ajbt200128:
Users can now run `semgrep <command> --baseline-commit COMMIT` in an unclean working tree. Semgrep will now scan staged files in baseline mode, in addition to its normal functionality.

We probably want to add to the documentation somewhere that users should be able to put
`ci --baseline-commit HEAD` as args for pre-commit, to scan + block based on CI rules and if they are in the blocking column. This will also report findings to the app, so `--dry-run` can be added to not do this.

This PR also adds a new pre-commit hook `semgrep-ci` which runs semgrep w/ the above args

PR checklist:


emjin:
There was a thread earlier where someone posted about a tool they made to do something similar. Would be great to respond in that thread once this is out to update them! https://semgrep.slack.com/archives/C01LNF69F9B/p1682841222100279

.pre-commit-hooks.yaml
```diff
@@ -14,6 +14,19 @@
...
   args: ["--disable-version-check", "--quiet", "--skip-unknown-extensions"]
 
+- id: semgrep-ci
+  name: semgrep
+  entry: semgrep
+  # If you use 'language: python', then pre-commit expects a toplevel 'setup.py'.
+  # This ./setup.py was defined just for pre-commit. Essentially, setup.py
+  # runs 'pip install semgrep==0.xxx.yyy'. It means that this hook does not even
+  # rely on the code in this repository; it just relies on pip and the availability
+  # of semgrep in pypy.
```
emjin: Thanks for this comment! Coming in without context, it took me a few reads. My interpretation of it now is that we (the semgrep maintainers) need to create a `setup.py` when we have a `language: python` pre-commit hook. Pre-commit will run that `setup.py` before running the commands in the hook, which means that we'll always use the latest available semgrep in pip.

If this is wrong, definitely clarify in the comment! If it's true, I think what's going on is, because I came in without knowing what the normal behavior was, the comment was a bit hard to follow. I could have used some framing

I'm also just generally curious:
- Why does `language: python` trigger this
- If we consider relying on pip's semgrep a good or bad thing

cli/tests/e2e/snapshots/test_baseline/test_staged_changes/error.txt
```diff
@@ -0,0 +1,29 @@
...
+               
+┌─────────────┐
+│ Scan Status │
+└─────────────┘
+  Scanning 1 file tracked by git with 1 Code rule:
+  Scanning 1 file.
+  Current version has 2 findings.
+
+Switching repository to baseline commit '4b63ca493df05e223404a80ac5f30d2aa830a704'.
+  Will report findings introduced by these commits (may be incomplete for shallow checkouts):
```
emjin: This text makes me think that we're going to report findings introduced by the baseline commit rather than subtract it out. It also makes me think that we're going to leave their repo in the baseline commit stage, which would annoy me

emjin: Maybe just link to docs and include warnings like "(may be incomplete for shallow checkouts)" in the docs? Alternatively, if this is a pretty common scenario, let's improve the warning to make it clear what condition they're looking for and what they can do. Something like:

`Note: You may see findings that weren't introduced by you. This can happen when we perform a shallow checkout. We do this because ___. You can fix that by ___`

cli/src/semgrep/git.py
```diff
@@ -68,20 +69,15 @@ def __init__(self, base_commit: str, is_mergebase: bool = False) -> None:
...
-        self._dirty_paths_by_status: Optional[Dict[str, List[Path]]] = None
 
         try:
             # Check commit hash exists
             git_check_output(["git", "cat-file", "-e", base_commit])
 
             self.status = self._get_git_status()
-            self._abort_on_pending_changes()
-            self._abort_on_conflicting_untracked_paths(self.status)
         except subprocess.CalledProcessError as e:
```
emjin, line 81: The way git UI renders the indentation on line 84 confused me so much as to what's happening

cli/src/semgrep/git.py
```diff
@@ -296,52 +219,26 @@ def baseline_context(self) -> Iterator[None]:
...
-                    raise Exception(
-                        f"Fatal error restoring Git state; please restore your repository state manually:\n{output}"
-                    )
+        cwd = Path.cwd()
+        git_root = get_git_root_path()
+        relative_path = cwd.relative_to(git_root)
+        with tempfile.TemporaryDirectory() as tmpdir:
+            try:
+                merge_base_sha = self._get_git_merge_base()
+                logger.debug("Running git checkout for baseline context")
```
emjin, line 244: Can you clarify that what we're actually doing is checking out the baseline commit in a new directory?

emjin (approved):
Thanks for the tests! I'm approving this from my knowledge, but it might be worth checking with @underyx and @brendongo if there are any potentially bad states that we should be testing. For instance, I don't know if there are any places where making a temporary directory could be problematic.

I would also appreciate some profiling of baseline scanning on large repos. You can clone https://github.com/returntocorp/examples-annotations or https://github.com/returntocorp/stress-test-monorepo for the purpose. If you do the former, `examples-annotations/repos/java/SQLi/cve-repos/OpenClinica` or the same path with `dotCMS` are both good examples. I'm concerned that this will slow down developers who work on monorepos. Depending on the results of that profiling, we might want to think about maintaining a faster alternative

.pre-commit-hooks.yaml
aryx: actually I added the comment. Austin copy pasted it from the previous id. It would actually be better to remove it here (and improve the one above).

aryx: I also don't understand your question, but trying to clarify, please read the toplevel comment in this file. This file is here to make 'semgrep' a possible pre-commit "plugin". When people use pre-commit, they can run many linters, some code formatters, etc. For example if you look at our own .pre-commit-config.yaml, we have: 
```  
- repo: https://github.com/myint/autoflake
    rev: v1.4
    hooks:
      - id: autoflake
```

aryx: But for the magic to work, autoflake authors had to add a special file to tell precommit how to install autoflake. See
 https://github.com/PyCQA/autoflake/blob/main/.pre-commit-hooks.yaml and the format of this "hooks" file
is defined here: https://pre-commit.com/#new-hooks (as mentioned in the toplevel comment).

aryx: Let me know if you still don't understand the comment.

cli/src/semgrep/git.py
aryx: Yes, looks like this git worktree is a very nice command.
Can you explain why it's useful here and why it allowed you to remove so much code in this PR.

cli/tests/e2e/snapshots/test_baseline/test_staged_changes/error.txt
ajbt200128: Good point, I never updated these printouts, I'll do that so it accurately reflects what's happening, which is creating a separate work tree, not checking out or switching anything

ajbt200128: It looks like this normally would print out a list of commits, and so this comment makes sense when doing a baseline say for a PR, but not for this test which is doing a baseline check comparing a dirty git state against HEAD. I'll see if I can update the logic so it only prints out when it makes sense

cli/src/semgrep/git.py
ajbt200128: Yea will do

cli/src/semgrep/git.py
ajbt200128: These two lines would check for if the git state was clean (no unstaged/staged files, or untracked files that conflict with the baseline commit), as before we were performing a git checkout in the same working tree. Now we just create a new working tree at the baseline, so we don't need to check these.

.pre-commit-hooks.yaml
ajbt200128: Will delete the 2nd occurence of this comment and update the above one

ajbt200128: Re: why we need it, I think pad's explanation makes sense, but the way I thought of it is it's similar to Docker's base image feature. pre-commit creates a new "environment" for each tool, with dependencies locally installed, as not to conflict with global state. This means pre-commit needs to know what deps are needed for each tool, and choosing python means python binaries (and pip) will be installed in this local environment, before `pip install semgrep` is ran. So for example if you have v1.25.0 installed globally, and the newest on pypi is v1.26.0, pre-commit will install the newest version locally in their environment thing, but not globally.

As far as relying on pip's semgrep, I don't see why not, it seems the least complex solution

ajbt200128:
> I would also appreciate some profiling of baseline scanning on large repos. 

Tried examples-annotations
- `examples-annotations/repos/java/SQLi/cve-repos/OpenClinica`
  - v1.25.0  `1.30s user 0.21s system 43% cpu 3.495 total`
  - this branch `1.18s user 0.46s system 40% cpu 3.986 total`
- `.../dotCMS`
  - v1.25.0 `1.28s user 0.22s system 48% cpu 3.083 total`
  - this branch `1.18s user 0.45s system 52% cpu 3.121 total`

This makes sense to what I expected. We're making a new directory + copying over everything, so the extra overhead is coming from the system calls for those. My guess is that the warning about the benchmarks above is probably safe to ignore, as for small repos (the tests above) the time spent scanning <<< time spent in system calls, but for large repos it's the opposite

emjin:
The benchmark message is happening in a bunch of repos :( We must have introduced a slight slowdown earlier

emjin:
Hm, ok, that's not awful, though definitely not great. I'm not a fan of waiting 3 or 4 seconds for pre-commit to complete. And it's quite common for our customers to have monorepos. I don't see an obvious solution, but it's definitely something I expect we'll have to revisit. Maybe with osemgrep we can copy only the files that are affected, for example.

cli/src/semgrep/git.py
emjin: Oh sorry this was just a gripe about github :D

cli/src/semgrep/git.py
emjin: @aryx Just to clarify, we've removed a lot of code because we changed the approach. Previously, we refused to run baseline scans when running `git checkout` would cause the user to lose changes. This meant we had to check for dirty paths and similar. Now, we don't have to do any of those checks, because instead we're just making a git copy of the repo into a separate directory.

The comments on how git worktree works are really helpful though! Just wanted to clarify that it's doing something different from the removed code

.pre-commit-hooks.yaml
emjin: The modified comment is helpful, thanks!

ajbt200128:
Yea, I think with osemgrep there'll be a lot more options for speeding this up, but for now this feels fine to me

aryx:
yes I think git worktree is a pretty expensive operation on a big monorepo that we want to avoid.
What we could do instead is just iterate over all the files modified between the current state and the baseline, get their old content (there is a command in git to show the content of a file at a certain SHA), and run semgrep on that file.

aryx:
@ajbt200128 @milanwilliams you might want to comment more on
https://github.com/navhits/semgrep-precommit/issues/1

andrew-lim-highspot:
@ajbt200128 I'm encountering an issue with using `--baseline-commit HEAD` during git commit. I'm getting the below git worktree error when running git commit with the semgrep scan pre-commit hook:

```
Creating git worktree from 'HEAD' to scan baseline.
[ERROR] Command failed with exit code: 128
        -----
        Command failed with output:
        Preparing worktree (detached HEAD dfdbee78)
fatal: .git/index: index file open failed: Not a directory
        Failed to run 'git worktree add /var/folders/_4/_bx2xt892l98t48d0zvycw4m0000gp/T/tmp244vnd3h dfdbee78ea52ea9a565d6c360d0e2a216088e385'. Possible reasons:
        - the git binary is not available
        - the current working directory is not a git repository
        - the baseline commit is not a parent of the current commit
          (if you are running through semgrep-app, check if you are setting `SEMGREP_BRANCH` or `SEMGREP_BASELINE_COMMIT` properly)
        - the current working directory is not marked as safe
          (fix with `git config --global --add safe.directory $(pwd)`)
        Try running the command yourself to debug the issue.
```

My .pre-commit-config.yaml looks like this:

```
repos:
  - repo: https://github.com/semgrep/pre-commit
    rev: 'v1.140.0'
    hooks:
      - id: semgrep
        args: ['--config', 'p/default', '--error', '--skip-unknown-extensions', '--metrics', 'off', '--baseline-commit', 'HEAD']
        verbose: true
```

What's strange is that running semgrep manually with `--baseline-commit HEAD` and also running `pre-commit run` works perfectly. It's only failing when I run `git commit`. Any ideas about this?

PR #7975, opened 2023-06-07
taint analysis: Improve handling of dataflow for tainted value propagation in class field definitions

### LINKED ISSUE #7742: Kotlin does not detect flow of string inside Objects
**Describe the bug**
Semgrep appears to ignore dataflow inside objects https://semgrep.dev/s/JKYj 

```
object Hi {
    private const val MYSTRING = "123"

    // should match
    private val magicString = Base64.decode(MYSTRING, Base64.DEFAULT).toString(Charsets.UTF_8)

}
```

Is ignored from the pattern in the rule above

**To Reproduce**
https://semgrep.dev/s/JKYj

**Expected behavior**
The rule should match inside objects too

**Environment**

**Use case**
Match more findings in Kotlin

ihji:
This commit resolves an issue where dataflow was not correctly accounted for when tainted values flowed through field definitions in class definitions. In JVM-based programming languages such as Java, Kotlin, and Scala, the order of field definitions is significant, and they function like statements during object initialization. To address this, during the transformation from AST to IL, we now treat a list of field definitions as a sequence of statements, allowing for proper tracking of tainted values during object initialization.

Closes #7742

PR checklist:


IagoAbal:
Is there a ticket associated with this? Shouldn't we add a changelog entry too?

tests/tainting_rules/kotlin/tainted_args.yaml
```diff
@@ -0,0 +1,16 @@
+rules:
```
IagoAbal: Sorry for the confusion, `tests/tainting_rules` should not be used anymore... It exists for historical reasons. Need to clean that up. Taint tests now go under `tests/rules/taint_*`.

src/analyzing/AST_to_IL.ml
```diff
@@ -1241,6 +1241,9 @@ and stmt_aux env st =
       let ss, e' = expr_with_pre_stmts env e in
       let lv = lval_of_ent env ent in
       ss @ [ mk_s (Instr (mk_i (Assign (lv, e')) (Related (G.S st)))) ]
+  | G.DefStmt (_, G.ClassDef { cbody = fields; _ }) ->
```
IagoAbal: In general I don't think this is correct, the fields could be modified by calls to methods no? It's correct if the only thing you're analyzing are the fields themselves. It would be useful to get more context here.

ihji: Certainly, it is possible to modify the fields after initialization. However, the primary objective of this PR is to perform data flow analysis specifically during object initialization. The purpose is to address issue #7742 by resolving a limitation that cannot be resolved using AST alone. In the AST, each field in the object definition is treated independently, but in taint mode, there is a partial feasibility in handling this scenario.

However, since I lack sufficient context on the design decisions made during IL transformation, I am also open to discarding these changes if the negative side effects outweigh the positives.

IagoAbal: Then IMO we should have a special visiting step that just analyzes fields in `Match_tainting_mode.ml`. ~Although I think this could be solved already with sym-prop. The #7742's case should be even simpler because it only needs const-prop.~ However when I open #7742 and I follow the Playground link, what I see is a search-mode rule.

ihji: The author of #7742 attempted to perform semi-taint analysis in search mode. However, this approach would not work with class fields without significant modifications to the class_definition AST. The PR enables the same analysis in taint mode, which appears to be more suitable for this particular case.

tests/rules/tainted_anon_type.yaml
```diff
@@ -0,0 +1,11 @@
+rules:
```
IagoAbal: `tests/rules/taint_...` instead of `tainted_....`, there are a few tests that break the convention already, no big deal but FYI.

changelog.d/gh-7742.fixed
```diff
@@ -0,0 +1 @@
+taint analysis: Improve handling of dataflow for tainted value propagation in class field definitions
```
IagoAbal, line 1: We should do our best so that changelog entries are easy to understand by e.g. technical writers. Not always easy, but to keep in mind. If I didn't know what is this about, and I just read that changelog entry, I wouldn't really know what had changed. From the user perspective, what works now that didn't work before?

IagoAbal: cc @adamkvitek

tests/rules/tainted_anon_type.cs
```diff
@@ -0,0 +1,15 @@
+class Asdf {
+    class ReturnObject1 {
+        public string Secret;
+    }
+
+    public void Test1() {
+        // ruleid: tainted-anon-type
+        Log(new ReturnObject1() { Secret = GetSecret() });
```
IagoAbal: No problem to have it here, but this test seems unrelated to the change, am I correct?

IagoAbal (approved):

src/analyzing/AST_to_IL.ml
IagoAbal: I'm not so sure about that, have you actually talked to Lewis about it?

IagoAbal: I am happy to merge this, I am not so sure you are actually fixing #7742.

tests/rules/tainted_anon_type.yaml
ihji: renamed.

changelog.d/gh-7742.fixed
ihji: amended the change description.

tests/rules/tainted_anon_type.cs
ihji: I included the test case because it appeared that this PR might also address #7995, but it turned out that it only works for scenarios where the source and target are in the same expression.

Removed the anonymous type test case.

IagoAbal:
If you add
```
Closes #7742
```
to your PR comment, we keep traceability and GitHub automatically closes the ticket when this is merged.

changelog.d/gh-7742.fixed
```diff
@@ -0,0 +1,10 @@
...
+
+This change resolves an issue where dataflow was not correctly accounted for
```
IagoAbal, line 3: This is quite similar to the PR comment which is oriented towards other PA members. This should be directed to non-PA people, so IMO e.g. implementation details aren't needed here. The last sentence is what I would personally put first, "Track tainted values during object initialization" or similar, followed by a description of what stuff now works "Semgrep will now track the flow of taint sources into sinks during object initialization" or similar. It doesn't need to be very long, in fact I would try to keep it short.  

Sorry for nit picking, I am not trying to be pedantic. Of course, I am not necessarily right, but we could check with Bruno and Adam for a second opinion. Both of them have complained in the past about the quality of the changelog entries. I'm just trying to push a bit so that over time we write better changelogs!

PR #7976, opened 2023-06-07
fix(taint): Improve the Taint comparison for offsets

emjin:
We use Stdlib.compare for offsets for taints in the Taint set, but offsets include variables of type `IL.ident`. These include tokens, id_info, etc. Make compare more precise.

Test plan: make test, especially see benchmark tests

PR checklist:

### DIFF
--- a/src/tainting/Taint.ml
+++ b/src/tainting/Taint.ml
@@ -134,10 +134,7 @@ and compare_args a1 a2 =
   let pos1 = a1.pos in
   let pos2 = a2.pos in
   match Stdlib.compare pos1 pos2 with
-  | 0 ->
-      List.compare
-        (fun n1 n2 -> IL_helpers.compare_name n1 n2)
-        a1.offset a2.offset
+  | 0 -> List.compare IL_helpers.compare_name a1.offset a2.offset
   | other -> other
 
 and compare_orig orig1 orig2 =
--- a/src/tainting/Taint.ml
+++ b/src/tainting/Taint.ml
@@ -130,6 +130,21 @@ and compare_sources s1 s2 =
       Option.compare compare_precondition s1.precondition s2.precondition
   | other -> other
 
+and compare_name n1 n2 =
+  match
+    AST_utils.with_structural_equal G.equal_ident n1.IL.ident n2.IL.ident
+  with
+  | true -> Stdlib.compare n1.sid n2.sid
+  | false -> Stdlib.compare n1.ident n2.ident
+
+(* type arg = { pos : arg_pos; offset : IL.name list } *)
+and compare_args a1 a2 =
+  let pos1 = a1.pos in
+  let pos2 = a2.pos in
+  match Stdlib.compare pos1 pos2 with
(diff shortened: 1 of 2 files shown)


emjin:
This is a draft until I see the test results

src/tainting/Taint.ml
```diff
@@ -130,6 +130,21 @@ and compare_sources s1 s2 =
       Option.compare compare_precondition s1.precondition s2.precondition
   | other -> other
 
+and compare_name n1 n2 =
```
IagoAbal: There is a `compare_name` in `IL_helpers`.

src/tainting/Taint.ml
```diff
@@ -130,6 +130,21 @@ and compare_sources s1 s2 =
...
+  with
+  | true -> Stdlib.compare n1.sid n2.sid
+  | false -> Stdlib.compare n1.ident n2.ident
+
+(* type arg = { pos : arg_pos; offset : IL.name list } *)
+and compare_args a1 a2 =
+  let pos1 = a1.pos in
+  let pos2 = a2.pos in
+  match Stdlib.compare pos1 pos2 with
+  | 0 -> List.compare (fun n1 n2 -> compare_name n1 n2) a1.offset a2.offset
```
IagoAbal: Yep this needs to be fixed, we added the offsets afterwards and forgot to update the comparison function. This explains the duplicates, good catch!

IagoAbal:
It may improve "perf" but this is a bug fix.

src/tainting/Taint.ml
```diff
@@ -130,6 +130,16 @@ and compare_sources s1 s2 =
       Option.compare compare_precondition s1.precondition s2.precondition
   | other -> other
 
+and compare_args a1 a2 =
+  let pos1 = a1.pos in
+  let pos2 = a2.pos in
+  match Stdlib.compare pos1 pos2 with
+  | 0 ->
+      List.compare
+        (fun n1 n2 -> IL_helpers.compare_name n1 n2)
```
IagoAbal: In this particular case I think `List.compare IL_helpers.compare_name ...` is preferable. Not need to change, just nit picking.

src/tainting/Taint.ml
```diff
@@ -130,6 +130,16 @@ and compare_sources s1 s2 =
       Option.compare compare_precondition s1.precondition s2.precondition
   | other -> other
 
+and compare_args a1 a2 =
+  let pos1 = a1.pos in
+  let pos2 = a2.pos in
+  match Stdlib.compare pos1 pos2 with
```
IagoAbal, line 136: I wonder if we should avoid `Stdlib.compare` here too!

IagoAbal (approved):
Nice catch! BTW, no changelog?

src/tainting/Taint.ml
emjin: Thanks for pointing it out, I wrote this when I had multiple conditions and forgot to change it back

src/tainting/Taint.ml
emjin: I'll open a separate PR to experiment with options

PR #7977, opened 2023-06-07
feat(taint): Relate Java getters and setters to properties

nmote:
Previously in Java, Semgrep would consider `o.getX()` tainted if a tainted argument was passed to `o.setX(...)`. However, it would not relate that taint back to the underlying property itself. This means that taint would only be tracked if getters and setters were used in concert. If a regular assignment to the property was made, followed by a getter, or if a setter was used, followed by a direct access to the property, Semgrep would fail to track taint.

Here, I've changed the handling of `o.setX(...)` to make Semgrep consider the property `o.x` tainted, and added functionality to consider `o.getX()` tainted if `o.x` is tainted. The modified tests illustrate the change in behavior.

Test plan: Automated tests

PR checklist:

### DIFF
--- a/changelog.d/getters.added
+++ b/changelog.d/getters.added
@@ -1,0 +1,1 @@
+In Java, Semgrep can now track taint through more getters and setters. It could already relate setters to getters (e.g. `o.setX(taint); o.getX()` but now it can relate setters and getters to properties (e.g. `o.setX(taint); o.x`).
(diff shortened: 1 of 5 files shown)


IagoAbal:
Nice, will do some refactoring here before merging.

IagoAbal:
I don't expect decent Java code mixing getters/setters and direct reads/writes of attributes. It could be useful in Pro for autogenerated getters/setters if we didn't resolve them, but then in Pro the `sid` of the attribute won't be `-1`, since it will be resolved.

IagoAbal (approved):

PR #7993, opened 2023-06-08
tainting: Fix perf regression caused by #7936

IagoAbal:
Calls to `type_of_expr` seem not to be cheap and even though we tried to limit them this cause a slowdown of about 25% in about a dozen repos in our benchmark. We should just not call `type_of_expr` unless at least one of the `taint_assume_safe_{booleans,numbers}` has been set, so rules that do not use these options remain unaffected. Long term we should make `type_of_expr` less costly.

Fixes: 64cb387f7a0 ("tainting: Add options to ignore taint based on types (#7936)")

test plan:
Run semgrep-compare full workflow based on stress-test-monorepo

PR checklist:

### DIFF
--- a/changelog.d/pa-2777-1.fixed
+++ b/changelog.d/pa-2777-1.fixed
@@ -1,0 +1,1 @@
+taint-mode: Fixed performance regression in 1.24.0 that affected taint rules.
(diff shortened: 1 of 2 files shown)


aryx (approved):
Please add a note in comment explaining why we have this intermediate function and also the performance note that we can get 25% slower on certain repo if we're not careful
becaus type_of_expr is slow

PR #7994, opened 2023-06-08
Add a timestamp when we use osemgrep with --debug

dinosaure:
PR checklist:

This small PR add a timestamp on the `DEBUG` output (it concerns only the `DEBUG` level). The metrics can be interesting to see and debug few things. Actually, we do the computation when we print but we probably should use the tags mechanism. /cc @aryx

### DIFF
--- a/libs/commons/Logs_helpers.ml
+++ b/libs/commons/Logs_helpers.ml
@@ -12,8 +12,22 @@ let pp_sgr ppf style =
   Format.pp_print_as ppf 0 style;
   Format.pp_print_as ppf 0 "m"
 
+(** This global is used by our reporter to do the diff between the last
+    log we outputed and the current one. Actually, the implementation is a
+    bit dumb and probably show weird metrics when we use [lwt]. For such
+    case, it's better to add a _counter_ and use the tag mechanism to really
+    show right metrics.
+
+    This variable is set when we configure loggers. *)
+
+let time_program_start = ref 0.
+
+let now () =
+  Mtime.to_uint64_ns (Mtime_clock.now ())
+  |> Duration.to_sec_64 |> Int64.to_float
+
 (* log reporter *)
-let reporter ?(dst = Format.err_formatter) () =
+let reporter ?(with_timestamp = false) ?(dst = Fmt.stderr) () =
   let report _src level ~over k msgf =
     let pp_style, style, style_off =
       match color level with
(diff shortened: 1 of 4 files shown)


aryx (approved):

dinosaure:
I updated the pull request and this is the new output:
image

To clarify, timestamp appears only when we use the flag `--debug` . Colors are added on the `Logs.reporter` now.

dinosaure:
I did some changes requested by @aryx and this is the final output:
image

The counter start from the beginning of the program and show seconds instead of `{micro,nano}`seconds instead of reset it at each `Logs.debug` calls.

libs/commons/Logs_helpers.ml
```diff
@@ -1,34 +1,35 @@
...
+    case, it's better to add a _counter_ and use the tag mechanism to really
+    show right metrics. *)
 
-(* print an ANSI escape sequence - not worth to use an extra library
-   (such as ANSIterminal) for this *)
-let pp_sgr ppf style =
-  Format.pp_print_as ppf 0 "\027[";
-  Format.pp_print_as ppf 0 style;
-  Format.pp_print_as ppf 0 "m"
+let value = ref 0.
```
aryx: value is not a great name. Maybe time_program_start would be better

libs/commons/Logs_helpers.ml
```diff
@@ -1,34 +1,35 @@
-(* ANSI escape sequences for colored output, depending on log level *)
```
aryx: you'll need to also keep this code, otherwise we get regressions in make osempass

libs/commons/Logs_helpers.ml
```diff
@@ -1,34 +1,35 @@
...
+      match with_timestamp with
+      | true ->
+          let current = now () in
+          Fmt.kpf k dst
+            ("@[[+%010.2fms]%a: " ^^ fmt ^^ "@]@.")
+            (current -. !value) Logs_fmt.pp_header (level, header)
+      | _ ->
+          Fmt.kpf k dst
+            ("@[%a" ^^ fmt ^^ "@]@.")
+            Logs_fmt.pp_header (level, header)
```
aryx, line 49: when we're not in --debug, we need to keep the exact same log format as before, otherwise
we get failures in 'make osempass' like the one you get right now.
you need to replace Logs_fmt.pp_header with the previous pp_style that was there before

aryx (changes requested):

PR #7996, opened 2023-06-08
chore: replace polymorphic compare with derived compare

emjin:
We were bitten previously by using Stdlib.compare and then changing the type. We could use ppx_deriving instead. This PR explores that option

Test plan: make test

PR checklist:


emjin:
@IagoAbal Let me know what you think, if you like this I can replace all of our Stdlib.compare instances in `Taint.ml`

src/tainting/Taint.ml
```diff
@@ -65,7 +66,7 @@ let rec _show_call_trace show_thing = function
 (* Signatures *)
 (*****************************************************************************)
 
-type arg_pos = string * int [@@deriving show]
+type arg_pos = string * int [@@deriving show, compare]
```
IagoAbal, line 69: Is that any better than polymorphic compare? It's probably faster, and maybe automatically ignores functions instead of raising an exception. But if we had used `ppx_compare` previously and then changed the data type as we did, would have that given us any warning that we needed to write a custom `compare` for the `offset`s?

src/tainting/Taint.ml
```diff
@@ -17,6 +17,7 @@ module G = AST_generic
 module PM = Pattern_match
 module R = Rule
 module LabelSet = Set.Make (String)
+open Ppx_compare_lib.Builtin
```
IagoAbal, line 20: It's a bit ugly to have to open this module, can't be done without it?

src/tainting/Taint.ml
```diff
@@ -65,7 +66,7 @@ let rec _show_call_trace show_thing = function
 (* Signatures *)
 (*****************************************************************************)
 
-type arg_pos = string * int [@@deriving show]
+type arg_pos = string * int [@@deriving show, compare]
 type arg = { pos : arg_pos; offset : IL.name list } [@@deriving show]
```
IagoAbal, line 70: Don't you have to also change this
```ocaml
Stdlib.compare pos1 pos2
```
in `compare_args` ?

emjin: Yup, just wanted to check the change with you before I went and changed all the Stdlib.compares

src/tainting/Taint.ml
emjin: I couldn't figure out how 😞

src/tainting/Taint.ml
emjin: Oh wait, did I not change any of the compares? Sorry, I must have lost a change at some point. I'll update the PR

src/tainting/Taint.ml
emjin: I expect it would have complained that it didn't have a `compare` function for `offset`, right?

IagoAbal: True, in this particular case. But if we add this `deriving compare` everywhere then I'm not sure it helps. For example, if the types were mutually recursive this would happen rather easily that you change something and don't realize you need a custom `compare`.

IagoAbal (approved):
I'm OK merging this for now as an experiment, although I don't promise it won't go away in some future clean up. TBH I'm not convinced that `ppx_compare` is much better than polymorphic compare, and for `arg_pos` writing a custom `compare` is trivial. Perhaps things look different if we see it used in more places. Remember to update the PR to actually use that derived `compare` function!

src/tainting/Taint.ml
```diff
@@ -114,6 +115,8 @@ and compare_sources s1 s2 =
   let pm1, ts1 = pm_of_trace s1.call_trace
   and pm2, ts2 = pm_of_trace s2.call_trace in
   match
+    (* TODO: I'm pretty suspicious of Stdlib.compare here,
+       the metavariable environments include tokens *)
```
IagoAbal, line 119: Totally agree.

IagoAbal: I'm really considering hand-writting almost every compare.

emjin: I'm going to handwrite this compare today and see what happens

PR #7997, opened 2023-06-08
feat(taint): make the taint limits configurable

emjin:
Allow semgrep-core/-proprietary users to configure the taint limits via command-line argument.

As noted in the help descriptions, this is meant for internal testing.

Test plan: ran locally, printed out when the limits were hit. Confirmed that the limits were hit without `-max_tainted_lvals 0 -max_taint_set_size 0` 

PR checklist:

### DIFF
--- a/src/configuring/Flag_semgrep.ml
+++ b/src/configuring/Flag_semgrep.ml
@@ -33,6 +33,12 @@ let max_cache = ref false
 (* Maximum size of a single target file, in bytes (exceptions apply). *)
 let max_target_bytes = ref 5_000_000
 
+(* Maximum number of tainted lvals to save. *)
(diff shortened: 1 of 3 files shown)


IagoAbal:
Do we really need this? I'm not very fond of exposing these as options. I don't expect we'll be adjusting these limits very often, nor we need to get these numbers super fine tuned.

IagoAbal (approved):
If you think this will help you a lot right now let's merge.

emjin:
Thanks! I mostly need to just be able to turn them on or off, but I made it so that I could set the value because why not. Currently I've been deleting them and running make but it's easy to forget that I've deleted them or not

PR #7998, opened 2023-06-08
tainting: Filter actual arguments taint by type

IagoAbal:

test plan:
make test # new tests

PR checklist:

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -1512,9 +1512,11 @@ let check_function_call_arguments env args =
     args
     |> List.fold_left_map
          (fun (lval_env, all_taints) arg ->
-           let taints, lval_env =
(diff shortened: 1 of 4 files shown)


nmote (approved):
Nice!

If I understand correctly, this just checks the type of the top-level expression, right? Could you add a test case like `sink("asdf" + (x != "safe))` to illustrate its behavior? Ideally that would be filtered out, too.

IagoAbal:
> Nice!
> 
> If I understand correctly, this just checks the type of the top-level expression, right? Could you add a test case like `sink("asdf" + (x != "safe))` to illustrate its behavior? Ideally that would be filtered out, too.

Good point, actually it may not work, need to use `check_function_call_arguments` also for `CallSpecial`.

IagoAbal:
At this point we may well just check the type of every expression... in a future clean up.

PR #8007, opened 2023-06-10
refactor(interactive): kill `Unix.select`

brandonspark:
## What:
This PR kills our previous workflow of polling `stdin` every `0.0005` seconds with `Unix.select`, and instead opts for using threads to determine when events should happen.

## Why:
Ideally, looping but taking actions on events should be able to be done with the `Term.pending` call, which supposedly tells us when there is a pending event to the terminal, to be acted upon. Past experimentation shows that this does not work, so we must find another way.

This way is cleaner, because we don't need to rely on `Unix.select`.

## How:
Instead of `Unix.select`, we spin up another thread (call it the _event thread_) which simply tightly loops and blocks via `Term.event`. This is fine, because this just means that the event thread blocks -- the main thread is free to take input and refresh on an asynchronous flag (such as the turbo thread completing matches).

## Test plan:
https://asciinema.org/a/VMyjaMUZsTCehnV9783qQHqPS

PR checklist:

### DIFF
--- a/src/osemgrep/cli_interactive/Interactive_subcommand.ml
+++ b/src/osemgrep/cli_interactive/Interactive_subcommand.ml
@@ -94,6 +94,13 @@ type state = {
   should_continue_iterating_targets : bool ref;
 }
 
+type event =
+  | End
+  | Key of Unescape.key
+  | Mouse of Unescape.mouse
+  | Paste of Unescape.paste
+  | Resize of { cols : int; rows : int }
+
 (*****************************************************************************)
 (* Constants *)
 (*****************************************************************************)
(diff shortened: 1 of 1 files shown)


src/osemgrep/cli_interactive/Interactive_subcommand.ml
```diff
@@ -114,6 +118,15 @@ let fg_line_num = A.(fg neutral_yellow)
 *)
 let should_refresh = ref false
 
+let event_queue :
```
aryx: please define the type separately, and if possible avoid Polymorphic variants; just use regular constructors.

src/osemgrep/cli_interactive/Interactive_subcommand.ml
```diff
@@ -804,10 +817,78 @@ let stop_thread_if_turbo state =
...
+   signal to the interactive loop that an event has occurred, and cause a
+   rerender.
+*)
+let spawn_event_thread term =
+  Thread.create
+    (fun _ ->
+      let rec loop () =
+        let e = Term.event term in
+        Queue.add e event_queue;
+        (* THINK: yield here? *)
```
aryx: reading the terminal with Term.event will yield the thread if there's nothing to read on the terminal.

aryx (approved):

PR #8012, opened 2023-06-12
feat(interactive): buffer matches

brandonspark:
## What:
This PR makes it so that, instead of only showing matches only once they are computed for every single file in the workspace, the matches are buffered in per file as they are computed.

## Why:
This results in an even faster and more tactile experience for the Interactive Mode user.

## How:
Now, the file zipper for a given state instance is updated in-place asynchronously by its associated turbo thread. This is precarious, however, because this means we need to make sure that we don't race when we write to the zipper for other reasons (for instance, when moving the file pointer or match pointer around). As such, we have `atomic_map_file_zipper`, and the file zipper is now protected by a lock, which is regenerated for each state instance.

I observed that naively making these changes makes the REPL noticeably slower, such as when you type really fast or backspace a lot. This seemed to be because we were spawning so many turbo threads even when there were events left to do, so I made it so we only spawn turbo threads if there are no pending events that would spawn turbo threads (which would mean that the turbo thread we spawned was redundant anyways).

## Test plan:
It's mega fast!!!
https://asciinema.org/a/qyLUJksrqrcDgeGLw8w2NPa6r

PR checklist:


src/osemgrep/cli_interactive/Framed_zipper.mli
```diff
@@ -43,6 +43,7 @@ type 'a t
 val move_left : 'a t -> 'a t
```
aryx: > A zipper is, of course, the classic data structure for a
>   list's "one hole context", of traversing a list via being located
 >  at a single point.

Remove the 'of course', and add a link to a web page explaining what a zipper is.

aryx: maybe rename move_up, move_down? so it's close to the scrolling metaphor? (even though there are horizontal scrolling it's true; still in the use case of semgrep interactive it's up/down scrolling).

src/osemgrep/cli_interactive/Framed_zipper.mli
```diff
@@ -43,6 +43,7 @@ type 'a t
...
 val move_right : 'a t -> 'a t
 val of_list : int -> 'a list -> 'a t
+val append : 'a -> 'a t -> 'a t
```
aryx, line 51: maybe add a few comments in the .mli about those different functions (Does not have to be very long).
For example for append maybe explain what's the semantic of it if the frame is at the end? does the frame move also if you add a new element?

src/osemgrep/cli_interactive/Interactive_subcommand.ml
```diff
@@ -398,45 +431,64 @@ let matches_of_new_iformula (new_iform : iformula) (state : state) :
...
+                        Match_rules.is_relevant_rule_for_xtarget fake_rule xconf
+                          xtarget
+                      then
+                        (* Calling the engine! *)
+                        Some
+                          (Match_search_mode.check_rule fake_rule hook xconf
+                             xtarget)
+                      else None
+                    in
+                    matches
```
aryx: this function is getting pretty long. Maybe you can introduce an helper function with a useful name.

src/osemgrep/cli_interactive/Interactive_subcommand.ml
```diff
@@ -796,24 +855,26 @@ let execute_command (state : state) =
...
-              Framed_zipper.empty_with_max_len (height_of_files state.term)
-        | _, None -> ()
-        | _, Some pat ->
-            let new_iformula = IPat (pat, true) in
-            let file_zipper = matches_of_new_iformula new_iformula state in
-            state.file_zipper := file_zipper;
-            should_refresh := true)
-      ()
+  (* Let's only spawn a new turbo thread if we don't have more events in
+     our queue.
```
aryx: But what if the next event is a noop? Then no thread will ever be created for it?
Maybe add a comment why this situation can't happen.

aryx (approved):

src/osemgrep/cli_interactive/Interactive_subcommand.ml
brandonspark: oh, good point. actually, this would mess us up in that instance. i changed it so we only don't spawn if the next is guaranteed to create another turbo thread

PR #8017, opened 2023-06-12
Add few functions which fill metrics

dinosaure:
PR checklist:

This PR adds various functions that I try to translate from the Python code in order to update the subsequent metrics. It adds everything that seems to be useful while trying to keep a certain correspondence between the values manipulated and those available. It requires returntocorp/semgrep-interfaces#81.

### DIFF
--- a/libs/commons/File.ml
+++ b/libs/commons/File.ml
@@ -44,6 +44,21 @@ let erase_this_temp_file path = Common.erase_this_temp_file !!path
 let is_executable path = Common2.is_executable !!path
 let filesize path = Common2.filesize !!path
 
+let find_first_match_with_whole_line path ?split:(chr = '\n') =
+  Bos.OS.File.with_ic path @@ fun ic term ->
+  let len = in_channel_length ic in
+  let res = Bytes.create len in
+  really_input ic res 0 len;
+  let lines = Bytes.split_on_char chr res in
+  let lines = Common.map Bytes.unsafe_to_string lines in
+  List.find_opt
+    (fun str -> Option.is_some (String_utils.contains term str))
+    lines
+
+let find_first_match_with_whole_line path ?split term =
+  find_first_match_with_whole_line path ?split term
+  |> Result.to_option |> Option.join
+
 let filemtime file =
   if !Common.jsoo then failwith "JSOO: File.filemtime"
   else (Unix.stat !!file).Unix.st_mtime
--- a/libs/commons/File.mli
+++ b/libs/commons/File.mli
@@ -128,6 +128,21 @@ val new_temp_file : string (* prefix *) -> string (* suffix *) -> Fpath.t
 val erase_temp_files : unit -> unit
 val erase_this_temp_file : Fpath.t -> unit
 
+val find_first_match_with_whole_line :
+  Fpath.t -> ?split:char -> string -> string option
+(** [find_first_match_with_whole_line path ~split term] opens [path], split it
+    by the given [split] character (defaults to ['\n']) and tries to return the
+    {b first} full element which contains the given [term].
+
+    For instance, you can search the first line which contains ["semgrep"]:
+
+    {[
+      find_first_match_with_whole_line my_file "semgrep"
+    ]}
+
+    It will returns the first {b full} line which contains the ["semgrep"]
+    occurrence. *)
+
 (*****************************************************************************)
 (* File properties *)
 (*****************************************************************************)
--- a/libs/commons/String_utils.mli
+++ b/libs/commons/String_utils.mli
@@ -1,2 +1,7 @@
 (* ??? *)
 val unit_str : ?pad:bool -> int -> string -> string
+
+val contains : string -> string -> int option
+(** [contains term str] is a {b naive} implementation which tries to find [term]
+    into the given [str]. It returns the position where it find the {b first}
+    occurrence of [term]. Otherwise, it returns [None]. *)
--- a/libs/git_wrapper/Git_wrapper.ml
+++ b/libs/git_wrapper/Git_wrapper.ml
@@ -179,3 +179,13 @@ let commit cwd msg =
   match Bos.OS.Cmd.run_status cmd with
   | Ok (`Exited 0) -> ()
   | _ -> raise (Error "Error running git commit")
+
+let get_project_url () =
+  let cmd = Bos.Cmd.(v "git" % "ls-remote" % "--get-url") in
+  let out = Bos.OS.Cmd.run_out cmd in
+  match Bos.OS.Cmd.out_string ~trim:true out with
+  | Ok (url, _) -> Some url
+  | Error _ ->
+      File.find_first_match_with_whole_line (Fpath.v ".git/config") ".com"
+(* TODO(dinosaure): this function is pretty weak. We probably should handle that
+   by a new environment variable. *)
--- a/libs/git_wrapper/Git_wrapper.ml
+++ b/libs/git_wrapper/Git_wrapper.ml
@@ -187,5 +187,8 @@ let get_project_url () =
   | Ok (url, _) -> Some url
   | Error _ ->
       File.find_first_match_with_whole_line (Fpath.v ".git/config") ".com"
-(* TODO(dinosaure): this function is pretty weak. We probably should handle that
-   by a new environment variable. *)
+(* TODO(dinosaure): this line is pretty weak due to the [".com"] (what happens
+   when the domain is [".io"]?). We probably should handle that by a new
+   environment variable. I just copied what [pysemgrep] does.
+   [git ls-remote --get-url] is also enough and if we can not get such
+   information, that's fine - the metadata is used only [Metrics_] actually. *)
--- a/src/core/Rule.ml
+++ b/src/core/Rule.ml
@@ -706,3 +706,16 @@ let rule_of_xpattern (xlang : Xlang.t) (xpat : Xpattern.t) : rule =
     paths = None;
     metadata = None;
   }
+
+(* TODO(dinosaure): Currently, on the Python side, we remove the metadatas and
+   serialise the rule in JSON format, then produce the hash from this
+   serialisation. However, there is no way (yet?) to serialise OCaml rules into
+   JSON format. We decided to use [Marshal] and then produce the hash
+   accordingly (ignoring the metadatas, as in Python).
+
+   However, this hash is used for metrics. So we don't produce the same hash as
+   the one produced by the Python implementation. *)
+let sha256_of_rule rule =
+  let rule = { rule with metadata = None } in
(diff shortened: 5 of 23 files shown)


libs/commons/File.mli
```diff
@@ -128,6 +128,20 @@ val new_temp_file : string (* prefix *) -> string (* suffix *) -> Fpath.t
 val erase_temp_files : unit -> unit
 val erase_this_temp_file : Fpath.t -> unit
 
+val manually_search_file : Fpath.t -> ?split:char -> string -> string option
```
aryx: why "manually"? It's done by a machine so there's nothing manual here.
Maybe find_first_match_with_whole_line?

libs/commons/File.ml
```diff
@@ -44,6 +44,25 @@ let erase_this_temp_file path = Common.erase_this_temp_file !!path
...
+let manually_search_file path ?split:(chr = '\n') =
+  Bos.OS.File.with_ic path @@ fun ic term ->
+  let len = in_channel_length ic in
+  let res = Bytes.create len in
+  really_input ic res 0 len;
+  let lines = Bytes.split_on_char chr res in
+  let lines = Common.map Bytes.unsafe_to_string lines in
+  let matches =
+    List.filter
+      (fun str -> Option.is_some (String_utils.contains term str))
```
aryx: you're returning all the matches but then return just the first one,
so maybe better to stop ASAP and raise a Found x, if you care only about the first match.

libs/commons/String_utils.mli
```diff
@@ -1,2 +1,7 @@
 (* ??? *)
 val unit_str : ?pad:bool -> int -> string -> string
+
+val contains : string -> string -> int option
+(** [contains term str] is a {b naive} implementation which tries to find [term]
```
aryx, line 5: What's the difference with Str.search_forward (Str.regexp_string term) 0 ?

src/core/Rule.ml
```diff
@@ -676,3 +676,16 @@ let rule_of_xpattern (xlang : Xlang.t) (xpat : Xpattern.t) : rule =
...
     metadata = None;
   }
+
+(* TODO(dinosaure): Currently, on the Python side, we remove the metadatas and
+   serialise the rule in JSON format, then produce the hash from this
+   serialisation. However, there is no way (yet?) to serialise OCaml rules into
+   JSON format. We decided to use [Marshal] and then produce the hash
+   accordingly (ignoring the metadatas, as in Python).
+
+   However, this hash is used for metrics. So we don't produce the same hash as
```
aryx: Ok. Yes there are a few places where we have this problem and where it's gonna be hard to return
the exact same thing than pysemgrep. Fine for now.

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -152,6 +152,15 @@ let analyze_skipped (skipped : Out.skipped_target list) =
     try List.assoc `Other groups with
     | Not_found -> [] )
 
+let get_project_url () =
```
aryx: You can move this in libs/git_wrapper/Git_wrapper.ml

semgrep.opam
```diff
@@ -36,6 +36,7 @@ depends: [
   "ppx_hash"
   "ppx_inline_test"
   "ppx_sexp_conv" {!= "v0.15.0" & != "v0.15.1"}
+  "ppx_accessor"
```
aryx: Where this is used?

src/osemgrep/core/Metrics_.ml
```diff
@@ -112,3 +119,157 @@ let g = ref default
 let configure config =
   g := { !g with config };
   ()
+
+open Semgrep_metrics_t
+
+let add_engine_type ~name =
+  g := Lun.(set (t_payload >> payload_value >> value_engineRequested)) name !g
+
+let is_using_registry () = Lun.(get t_is_using_registry) !g
```
aryx: I'd rather use (!g).is_using_registry than using Lun here.

aryx (approved):
Not a huge fan of Lun, but at the same time I don't really see a simpler alternative; I wish
the Semgrep_metrics had mutable fields we could just go deep and modify
them deepy.

aryx:
I've merged your 'mutable' modif to semgrep_metrics.atd, so feel free to use the latest main semgrep_interfaces (already merged in develop).

libs/commons/String_utils.mli
dinosaure: Nothing, if you prefer to use `Str` instead of, I can replace it :+1: - I'm not an user of `Str` in general.

aryx: Yes I think it's better to reuse standard library functions.

libs/git_wrapper/Git_wrapper.ml
```diff
@@ -179,3 +179,13 @@ let commit cwd msg =
...
   | _ -> raise (Error "Error running git commit")
+
+let get_project_url () =
+  let cmd = Bos.Cmd.(v "git" % "ls-remote" % "--get-url") in
+  let out = Bos.OS.Cmd.run_out cmd in
+  match Bos.OS.Cmd.out_string ~trim:true out with
+  | Ok (url, _) -> Some url
+  | Error _ ->
+      File.find_first_match_with_whole_line (Fpath.v ".git/config") ".com"
+(* TODO(dinosaure): this function is pretty weak. We probably should handle that
```
aryx: What do you mean by weak? What is the issue?

src/osemgrep/cli/CLI.ml
```diff
@@ -276,5 +276,6 @@ let main argv : Exit_code.t =
   *)
   (*TOADAPT? adapt more of Common.boilerplate? *)
   let exit_code = safe_run ~debug (fun () -> dispatch_subcommand argv) in
+  Metrics_.add_exit_code exit_code;
   before_exit ~profile ();
```
aryx, line 284: I guess we need to send the payload at some point before exiting.

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -267,12 +277,38 @@ let run (conf : Scan_CLI.conf) : Exit_code.t =
                 Some (file_match_results_hook conf filtered_rules) )
           | { output_format; _ } -> (output_format, None)
         in
+        Profiler.save profiler ~name:"core_time";
```
aryx: I think it would be better to group those 2 calls in one if it's possible, for example
Profile.execute_and_profile ~name:"core_time" (fun () -> ...)

aryx: when it's possible hwen the 2 calls are in the same scope

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -267,12 +277,38 @@ let run (conf : Scan_CLI.conf) : Exit_code.t =
...
             targets
         in
+        Profiler.save profiler ~name:"core_time";
+
+        Metrics_.add_engine_type
+          ~name:
+            (Format.asprintf "%a" Out.pp_engine_kind
+               res.Core_runner.core.engine_requested);
+
+        let filtered_matches =
```
aryx: maybe you can put this in an helper function, like stat_rule_id_count or something, and give a nice comment
and signature.

src/osemgrep/core/Metrics_.mli
```diff
@@ -1,3 +1,5 @@
+module Out = Semgrep_output_v1_t
```
aryx: we avoid module aliases in .mli.
Just replace Out with the full name in the type signature.

src/osemgrep/core/Profiler.mli
```diff
@@ -0,0 +1,29 @@
+(** A really basic and linear profiler.
```
aryx: what does linear mean in this context?

aryx: Maybe you can add a comment like

related: libs/profiling/Profiling.ml also provide a poor's man profiler, but here we want to restrict profiling
to broad categories that are compatible with pysemgrep, so better to define another module for it.

aryx (changes requested):
Ok now we need to code that actually send the metrics to our backend.
Otherwise looks good!

libs/git_wrapper/Git_wrapper.ml
dinosaure: I mean, for my point of view, depending on what we can have into the `.git/config`, the result can be wrong. For instance, in Robur, we use `https://git.robur.coop/` which will not be recognized by this function. I'm not sure why we need to `find_first_match_with_whole_line` here when `git ls-remote --get-url` seems right and enough.

aryx: ah it's because of the .com? Yes, seems bad, but maybe you can add a comment with your example (and saying you wrote this because that's how it was done in pysemgrep)

src/osemgrep/cli/CLI.ml
aryx: maybe add a giant TODO comment, explaining what remains to be done here.

src/osemgrep/cli_scan/Scan_subcommand.ml
aryx: or at least a Profiler.start and Profiler.end

aryx: (or better, both)

src/osemgrep/cli/CLI.ml
dinosaure: Actually, if only the `scan` sub-command matters, we probably can add this line here:
https://github.com/returntocorp/semgrep/blob/b031ff9a55cdc380fee335e0424384f0d1e4beff/src/osemgrep/cli_scan/Scan_subcommand.ml#L338-L340

And the the `Metrics_.send` after that.

aryx: long term we want to support all command. Actually @hannesm recently improved significantly 'osemgrep ci', so we should just imitate what pysemgrep is doing.

dinosaure: Ok, I will write a big `TODO`

aryx:
@dinosaure regressions for osempass

aryx (approved):

aryx:
```      "caml_digestif_md5_ctx_size not implemented",```

aryx:
I guess we need to give a dummy implementation for jsoo.

aryx:
@dinosaure can you fix the conflict?
I think it's because @hannesm did a modification to this file recently.

dinosaure:
I rebased the PR, let's see the CI.

aryx (approved):

PR #8025, opened 2023-06-12
fix(CI): add `ppx-expect` to opam

brandonspark:
## What:
This PR adds `ppx-expect` to our `semgrep.opam` file, so CI runs fine again.

PR checklist:

### DIFF
--- a/semgrep.opam
+++ b/semgrep.opam
@@ -35,6 +35,7 @@ depends: [
   "ppx_deriving"
   "ppx_hash"
   "ppx_inline_test"
+  "ppx_expect"
   "ppx_sexp_conv" {!= "v0.15.0" & != "v0.15.1"}
   "re"
   "pcre"


tpetr (approved):
looks like GHA is unhappy, but LGTM once you rerun the checks

emjin (approved):
I don't know why you have to do this manually but let's try it

aryx:
I think that with the recent optimization Hannes did to our build, we don't call 'dune build' anymore, but instead
'dune build _build/some/precise/binary', so maybe when you modify the dune-project and type make, it does not regenerate the semgrep.opam file anymore, so you have to modify it manually.
The reason we have semgrep.opam committed, and not removed like the other autogenerated .opam files, is that we need semgrep.opam in 'make setup', which runs even before dune is installed. I'll try to add comments in the Makefile and dune-project to clarify those things.

PR #8028, opened 2023-06-13
sophia/pa-2807 limit core autodetect to 16 

SophiaSR:
Semgrep's memory usage grows linearly with the number of jobs. For now, limiting --jobs to 16 when autodetected. Otherwise, we can get into a situation where we detect a large number of underlying CPU cores, but the customer hasn't allocated a corresponding amount of memory.

Test Plan: For now tested by setting the max cores to 4, on a 10 core machine. Checked to see that this would run with 4 when no value given for jobs, but would run as 10 when set explicitly to 10 by user. 

PR checklist:

### DIFF
--- a/cli/src/semgrep/engine.py
+++ b/cli/src/semgrep/engine.py
@@ -63,7 +63,8 @@ class EngineType(Enum):
     def default_jobs(self) -> int:
         if self == EngineType.PRO_INTERFILE:
             return 1
-        return self.get_cpu_count()
+
+        return min(16, self.get_cpu_count())
 
     @property
     def default_max_memory(self) -> int:
--- a/cli/src/semgrep/engine.py
+++ b/cli/src/semgrep/engine.py
@@ -63,7 +63,7 @@ class EngineType(Enum):
     def default_jobs(self) -> int:
         if self == EngineType.PRO_INTERFILE:
             return 1
-        """Maxing out number of cores used to 16 if more not requested to not overload on large machines"""
+        # Maxing out number of cores used to 16 if more not requested to not overload on large machines
         return min(16, self.get_cpu_count())
 
     @property
--- a/cli/src/semgrep/engine.py
+++ b/cli/src/semgrep/engine.py
@@ -63,7 +63,7 @@ class EngineType(Enum):
     def default_jobs(self) -> int:
         if self == EngineType.PRO_INTERFILE:
             return 1
-
+        """Maxing out number of cores used to 16 if more not requested to not overload on large machines"""
         return min(16, self.get_cpu_count())
 
     @property
--- a/changelog.d/pa-2807.changed
+++ b/changelog.d/pa-2807.changed
@@ -1,0 +1,1 @@
+Updated the maximum number of cores autodetected to 16 to prevent overloading on large machines when users do not specify number of jobs themselves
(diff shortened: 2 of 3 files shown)


brandonspark:
So *you* got the perf work.

cli/src/semgrep/engine.py
```diff
@@ -63,7 +63,8 @@ def get_cpu_count() -> int:
     def default_jobs(self) -> int:
         if self == EngineType.PRO_INTERFILE:
             return 1
-        return self.get_cpu_count()
+
```
aryx: Please add a comment in the code why we use a max of 16 (mostly copy what you put in your PR explanation in the code instead).

aryx (changes requested):
First PR!

Please can you also update osemgrep code, because this python code will at some point be removed. The relevant code is in src/osemgrep/cli_scan/Scan_CLI.ml, look for the call to get_cpu_count().

SophiaSR:
Added comments, and updated the osemgrep code as well. Not sure why the lint/pre-commit test is failing though, could someone take a look at that or point me in the right direction?

aryx:
hmm, I get the same error in another PR. Something is broken on develop probably.

cli/src/semgrep/engine.py
```diff
@@ -63,7 +63,8 @@ def get_cpu_count() -> int:
     def default_jobs(self) -> int:
         if self == EngineType.PRO_INTERFILE:
             return 1
-        return self.get_cpu_count()
+        """Maxing out number of cores used to 16 if more not requested to not overload on large machines"""
```
aryx: I think in Python those docstring comments are usually written to explain the whole function and put
just after the ':'.
I think here you should use the '#' syntax for this kind of comment

aryx:
This is a user-facing change, so you should add an entry in changelog.d/pa-2807.changed

aryx (changes requested):
Mostly good

aryx:
Hmm let me rerun the failing lint / pre-commit workflow.

aryx:
seems to work now.

aryx:
(Github action is quite flaky, just FYI, so it's not unusual we need to relaunch manually some failing jobs).

aryx:
This is a user-facing change, so you should add an entry in changelog.d/[pa-2807](https://linear.app/r2c/issue/PA-2807).changed

SophiaSR:
Added entry in changelog and updated to correct syntax

aryx (approved):

emjin (approved):
🚀

cli/src/semgrep/engine.py
```diff
@@ -63,7 +63,8 @@ def get_cpu_count() -> int:
     def default_jobs(self) -> int:
         if self == EngineType.PRO_INTERFILE:
             return 1
-        return self.get_cpu_count()
+        # Maxing out number of cores used to 16 if more not requested to not overload on large machines
```
IagoAbal, line 66: If I didn't know what happened here, and I just found this comment, I would not know what "overload" means here.

IagoAbal:
I left a comment in Linear too, but this solution seems too ad-hoc. And I'm sure those using larger machines with the proper amount of memory will not realize that we have this limit, it's a bit like `interfile: true`, you know it's there but it's so easy to forget about it.

cli/src/semgrep/engine.py
```diff
@@ -63,7 +63,8 @@ def get_cpu_count() -> int:
...
+        return min(16, self.get_cpu_count())
```
IagoAbal, line 67: And we shouldn't be doing this silently, there should be warning being logged here IMO.

PR #8047, opened 2023-06-14
accept patterns with standalone decorators (Python)

### LINKED ISSUE #4722: Python decorators cannot be matched in isolation
### Describe the bug
The pattern syntax does not allow matching a Python pattern in isolation, and requires it to be specified along with a function or method.

### To Reproduce
Demo of error: https://semgrep.dev/s/JgRj

Or more simply, any pattern like `@patch(...)` is not permitted.

### Expected behavior
I would expect it to be possible to match decorators as isolated pattern expressions.

### What is the priority of the bug to you?

### Environment
`pip install semgrep==0.82.0` on Ubuntu Linux 20.04.

### Use case
Same as https://github.com/returntocorp/semgrep/issues/4721 -- this was an alternative approach that would have avoided the inadvertent removal of any additional decorators, along with the whitespace issues of <https://github.com/returntocorp/semgrep/issues/3070>.

yosefAlsuhaibani:
closes https://github.com/returntocorp/semgrep/issues/4722

Modified python parsing & matching to accept patterns of the form `@deco(...)` to match code of the form 
```python
#ERROR:
@deco("rate")
def fun():
  return None
```

### previously: 

Python patterns involving decorators couldn't be expressed via `@$NAME(...)`; the parser required the decorator to be paired up with a function or class declaration such as:

```python
@$NAME(...)
def func():
  ...

# or 

@$NAME(...)
class room:
  ...
```

### Test plan:

I've written tests that test this capability in `tests/patterns/python/standalone_decorator.{py,sgrep}`, which I've used to test if this works as well as using all of the other tests in `test/*` via `make test`
PR checklist:


languages/python/ast/AST_python.ml
```diff
@@ -387,6 +387,10 @@ type stmt =
   (* should be allowed just at the toplevel *)
   | FunctionDef of function_definition
   | ClassDef of class_definition
+  | Decorators of decorator list
+(* Allowed only in Semgrep; as we want to allow patterns like @$NAME()
+ * in a Semgrep pattern. Not valid Python
+ *)
```
emjin: We generally put the comment before the code (see Exec above)

languages/python/generic/Python_to_generic.ml
```diff
@@ -590,6 +590,10 @@ and stmt_aux env x =
         }
       in
       [ G.DefStmt (ent, G.ClassDef def) |> G.s ]
+  | Decorators ld ->
+      let env = { env with context = InPattern } in
+      let ld = list (fun dec -> G.At (decorator env dec)) ld in
+      [ G.OtherStmt (G.OS_LoneAttrs, ld) |> G.s ]
```
emjin: Subtlety with patterns: we'd rather parse this to a `G.Any (G.Attribute (...))` if possible, or something similar. It doesn't need to be in a statement list, the only time we'll ever want to parse this is if it's a standalone pattern*. 

See `Ruby_to_generic.any` for an example of how we've done similar modifications in the past

*If I'm wrong it won't be hard to change back

libs/ast_generic/AST_generic.ml
```diff
@@ -1332,7 +1332,8 @@ and other_stmt_operator =
   | OS_ThrowFrom
   | OS_ThrowNothing
   | OS_ThrowArgsLocation
-    (* Python2: `raise expr, expr` and `raise expr, expr, exr` *)
+    (* Python-semgrep: allow decorators without function decls *)
+  | OS_LoneAttrs (* Python2: `raise expr, expr` and `raise expr, expr, exr` *)
```
emjin: Let's avoid having to add this if possible

tests/patterns/python/standalone_decorator.py
```diff
@@ -0,0 +1,16 @@
...
+  return 42
+
+#ERROR:
+@decorator(1,2,3)
+def grumble():
+  return "wahoo"
+
+#ERROR:
+@decorator("oh")
+@decorator("my")
```
emjin, line 14: Hm, I think I'd expect line 13 to match as well as a user. I believe my proposed change to `G.Any (G.At (...))` will do precisely that because it'll try to match both `G.At (...)` nodes, instead of just capturing the list.

Because we ran into this, can you include another test case with `@$NAME($X)`? Let's also include for each set of tests at least one example that shouldn't match.

emjin (changes requested):
Great work! With patterns that match partial constructs like this, we sometimes take a different approach. Let me know if you have any questions after reading the comments

languages/python/generic/Python_to_generic.ml
yosefAlsuhaibani: would a sequence of decorators not be allowed? so something like:
```
$NAME($X)
$deco("hello")
```

tests/patterns/python/standalone_decorator.py
yosefAlsuhaibani: I've added additional tests to catch what you've described, and indeed it does miss those findings, which is troublesome! I've extended the old tests in `tests/patterns/python/standalone_decorator.{py,sgrep}`, and added a new set of tests for `@$NAME($X)` in `tests/patterns/python/standalone_decorator-named.{py,sgrep}`

yosefAlsuhaibani: Also: now we do parse lone decorators in `semgrep` patterns as `any`!

libs/ast_generic/AST_generic.ml
```diff
@@ -1332,7 +1332,7 @@ and other_stmt_operator =
   | OS_ThrowFrom
   | OS_ThrowNothing
   | OS_ThrowArgsLocation
-    (* Python2: `raise expr, expr` and `raise expr, expr, exr` *)
+  (* Python2: `raise expr, expr` and `raise expr, expr, exr` *)
```
emjin: I'm guessing this comment isn't meant to be there?

emjin (approved):
Awesome work! Be sure to add a changelog entry before merging

tests/patterns/python/standalone_decorator_named.py
emjin, line 1: nit: I wouldn't call this "named", maybe "single_arg"

libs/ast_generic/AST_generic.ml
yosefAlsuhaibani: I think its miss aligned; should be above `OS_ThrowFrom`?

yosefAlsuhaibani: Wait nvm 😭 I think I mistakingly added it

changelog.d/gh-4722.added
emjin, line 1: It's helpful to give a quick example of a pattern you could write

yosefAlsuhaibani: Added!

changelog.d/gh-4722.added
```diff
@@ -0,0 +1,8 @@
+Added lone decorators as a valid Python semgrep pattern, so for example `$NAME($X)` will
```
IagoAbal, line 1: Typo here, it should have been `@$NAME($X)` I guess.

IagoAbal: Nice changelog entry BTW! To the point, and pretty easy to understand for anybody as it should be.

PR #8052, opened 2023-06-14
feat(matching): match fields with stmts patterns

brandonspark:
## What:
This PR makes it so that `Ss` patterns can now match onto `fields` AST nodes.

## Why:
SR has been asking for this for a while, but sometimes they want to write a rule which takes advantage of some sequencing of statements. Unfortunately, because class fields are different than statements, they cannot write a rule which looks like a sequence of statements, because it won't match onto the fields!

This distinction between fields and statements is, to my understanding, so you can freely permute the fields when doing things like matching classes against classes. I'm all for that, in the `fields` to `fields` case, but I don't really see a reason for why we shouldn't be able to match a pattern which _looks_ like a series of fields, to a series of fields, even if the fields are supposedly unordered. This might make it so we don't get findings that we might expect, if the findings are truly unordered, but that's better than not getting them at all.

See: https://github.com/returntocorp/semgrep/issues/7742

## How:
I just copied the logic from `v_stmts` to `v_fields`.

## Test plan:
`make test`

Closes PA-2754

PR checklist:


brandonspark:
@aryx

src/matching/Match_patterns.ml
```diff
@@ -477,6 +477,64 @@ let check2 ~hook mvar_context range_filter (config, equivs) rules
...
+                 Profiling.profile_code "Semgrep_generic.kfields" (fun () ->
+                     let x = Common.map (fun (F x) -> x) x in
+                     let env = MG.empty_environment cache lang config in
+                     let matches_with_env = match_sts_sts rule pattern x env in
+                     if matches_with_env <> [] then
+                       (* Found a match *)
+                       matches_with_env
+                       |> List.iter (fun (env : MG.tin) ->
+                              let span = env.stmts_match_span in
+                              match Stmts_match_span.location span with
```
aryx, line 515: @IagoAbal where are you with the deprecation of martin's stmt optimizations?
This is the kind of thing we do that would not be needed anymore.

aryx (approved):
Very very nice.
There is a comment in ast_generic.ml about fields:
```
 * note: It is tempting to want to 'field' be just an alias for 'stmt',
 * but fields can be matched in any order, so it is probably better
 * to keep them separate.
```
maybe you can add an extra note after like:
note: we do some special things in xxx.ml so that stmts patterns can also be matched against fields.

src/matching/Match_patterns.ml
brandonspark: @IagoAbal

brandonspark: should I wait on this, or merge?

aryx: no you can merge, it was just an FYI to @IagoAbal that if he still plans to remove martin's stmt optimization, that would have simplified this code.
I'll merge.

IagoAbal: I'll got for it, I was too focused on requests from one customer.

PR #8053, opened 2023-06-14
Adds date comparison, allowing features such as checking date expiration Closes: #7992

### LINKED ISSUE #7992: Expiration date check in the metavariable-comparison condition
Hi guys!
I would like to check if a date has expired, as in this simple example: https://semgrep.dev/playground/s/DrwY

In the case where you are asking yourself for practical examples. I need this feature to verify inside some XML files that some TLS certificates are not expired.
(Example: https://developer.android.com/training/articles/security-config look at the `<pin-set>` expiration attribute)

As always, thank you guys! :heart:

SophiaSR:
Attempt 2, this time hopefully branched it correctly. 

This adds the functionality requested in issue 7992. Currently supports strptime("yyyy-mm-dd") only as a way to create a date. Also has today() to get the current date, again currently supporting only year, month, and date.

PR checklist:


brandonspark:
CI failure looks unrelated, I was getting this earlier. You can rerun the tests by going to details (at the bottom next to the pre-commit job) -> re-run jobs (at the top right) -> re-run failed jobs.

src/engine/Eval_generic.ml
```diff
@@ -202,6 +204,31 @@ let rec eval env code =
...
+      match v with
+      | String s -> (
+          let yyyy_mm_dd = String.split_on_char '-' s in
+          match yyyy_mm_dd with
+          | [ y; m; d ] -> (
+              match
+                (int_of_string_opt y, int_of_string_opt m, int_of_string_opt d)
+              with
+              | Some yv, Some mv, Some dv ->
+                  if CalendarLib.Date.is_valid_date yv mv dv then
```
brandonspark: You can fold these together, if you use a `when` clause!
```
| Some yv, Some mv, Some dv when CalendarLib.Date.is_valid_date yv mv dv ->
```

brandonspark: Saves a case :)

src/engine/Eval_generic.ml
```diff
@@ -246,18 +273,32 @@ and eval_op op values code =
...
   | G.Lt, [ Float i1; Int i2 ] -> Bool (i1 < float_of_int i2)
+  | G.Lt, [ Date (y1, m1, d1); Date (y2, m2, d2) ] ->
+      Bool
+        (if y1 =*= y2 then if m1 =*= m2 then d1 < d2 else m1 < m2 else y1 < y2)
   | G.LtE, [ Int i1; Int i2 ] -> Bool (i1 <= i2)
   | G.LtE, [ Float i1; Float i2 ] -> Bool (i1 <= i2)
   | G.LtE, [ Int i1; Float i2 ] -> Bool (float_of_int i1 <= i2)
   | G.LtE, [ Float i1; Int i2 ] -> Bool (i1 <= float_of_int i2)
+  | G.LtE, [ Date (y1, m1, d1); Date (y2, m2, d2) ] -> (
+      match eval_op G.Gt [ Date (y1, m1, d1); Date (y2, m2, d2) ] code with
```
brandonspark: nice!

src/engine/Eval_generic.ml
```diff
@@ -318,6 +359,8 @@ and eval_str _env ~code v =
     | Int i -> string_of_int i
     | Float f -> string_of_float f
     | String s -> s
+    | Date (y, m, d) ->
+        string_of_int y ^ "-" ^ string_of_int m ^ "-" ^ string_of_int d
```
brandonspark: Could you use a format string instead?

Try `Common.spf "%s-%s-%s" (string_of_int y) (string_of_int m) (string_of_int d)`

src/engine/Eval_generic.mli
```diff
@@ -5,6 +5,7 @@ type value =
   (* the string does not contain the enclosing '"' *)
   | String of string
   | List of value list
+  | Date of int * int * int
```
brandonspark: Could you add a comment here in the `.mli` too, about the meaning of each of the `int`s?

src/engine/Eval_generic.ml
brandonspark: although, actually, perhaps it would be best to extract a helper function, `date_compare`, which produces an 
`int` for the relationship between two dates, `-1`, `0`, and `1` (I know in SML we have `datatype order = LESS | EQUAL | GREATER` but unfortunately OCaml doesn't, so let's just go with convention, I know it's gross)

brandonspark (changes requested):
Just some small changes I'd recommend.

Could you also add a test demonstrating this date comparison logic? You can add it in `tests/rules`, you'll want to add one file `<filename>.yaml`, with the YAML rule which demonstrates this new eval behavior, and a target file `<filename>.<ext>`, in whatever language you please, with an extension fitting the language of the rule.

src/engine/Eval_generic.mli
SophiaSR: Done!

src/engine/Eval_generic.ml
SophiaSR: Done!

src/engine/Eval_generic.ml
SophiaSR: Done!

src/engine/Eval_generic.ml
SophiaSR: Oo nice, didn't know this syntax, done!

src/engine/Eval_generic.ml
```diff
@@ -139,6 +141,18 @@ let value_of_lit ~code x =
   | G.Float (Some f, _t) -> Float f
   | _ -> raise (NotHandled code)
 
+(* helper function for comparing dates *)
+let date_compare date1 date2 =
```
brandonspark: Could you pattern match in the args instead, so we avoid the unary `match`?

src/engine/Eval_generic.ml
```diff
@@ -139,6 +141,18 @@ let value_of_lit ~code x =
...
+  match (date1, date2) with
+  | (y1, m1, d1), (y2, m2, d2) ->
+      if y1 < y2 then -1
```
brandonspark: I believe there's `Int.compare : int -> int -> int`, so you could do something like
```
let y_cmp, m_cmp, d_cmp = Int.compare (y1, y2), Int.compare (m1, m2), Int.compare (d1, d2) in
if y_cmp = 0 then
  ...
else ...
```
(less efficient, but negligibly so)

It would be nice if we had a first-class way of doing this logic, like some kind of comparison chaining operator, in our common library, but AFAIK we don't. Oh well.

src/engine/Eval_generic.ml
```diff
@@ -246,18 +284,30 @@ and eval_op op values code =
   | G.Gt, [ Float i1; Float i2 ] -> Bool (i1 > i2)
   | G.Gt, [ Int i1; Float i2 ] -> Bool (float_of_int i1 > i2)
   | G.Gt, [ Float i1; Int i2 ] -> Bool (i1 > float_of_int i2)
+  | G.Gt, [ Date (y1, m1, d1); Date (y2, m2, d2) ] ->
+      Bool (date_compare (y1, m1, d1) (y2, m2, d2) =*= 1)
```
brandonspark: We could avoid explicitly deconstructing the args here, since we just pass it into `date_compare`. 

Extra:
If you bound it to say, `d1`, and `d2`, we could also get freaky fancy and write:
```
eval_op G.Gt [ Int (date_compare d1 d2); Int 0 ] code
```
and so on, in the other cases.

I'll leave the `eval_op` thing to your judgement, though. I could go either way on that.

tests/rules/date_comparison.py
```diff
@@ -0,0 +1,4 @@
+#OK
+day1 = "3000-01-15"
+#ERROR: match
+day2 = "1906-12-09"
```
brandonspark, line 4: Could we add a test demonstrating when the string does not resemble a date? (just in this file is fine)

brandonspark (changes requested):
Last touches. Looking good!

src/engine/Eval_generic.ml
brandonspark: Wait! I'm wrong! There's `Stdlib.compare`, so we could `Stdlib.compare` on the dates directly.

Now, I hate polymorphic compare as much as the next person, but this code isn't prone to refactoring, so we should be OK.

IagoAbal:
Could you fix the PR title? You can check the Git history for inspiration.
> Sophia/pa 7992 adding date comparison

You could also add
```
Closes #7992
```
to the PR description.

src/engine/Eval_generic.ml
```diff
@@ -45,6 +45,8 @@ type value =
   | Float of float
   | String of string (* string without the enclosing '"' *)
   | List of value list
+  | Date of int * int * int
```
aryx: I think it would be simpler to not introduce an extra value and just use Float.
Unix.gettimeofday returns simply a float representing the number of seconds since a special date in 1970.
Then you can just use the regular < to compare dates.
so today() could internally return a float, and strptime() would still parse the string, but convert the date to a float.

src/engine/Eval_generic.ml
```diff
@@ -246,18 +284,30 @@ and eval_op op values code =
   | G.Gt, [ Float i1; Float i2 ] -> Bool (i1 > i2)
   | G.Gt, [ Int i1; Float i2 ] -> Bool (float_of_int i1 > i2)
   | G.Gt, [ Float i1; Int i2 ] -> Bool (i1 > float_of_int i2)
+  | G.Gt, [ Date (y1, m1, d1); Date (y2, m2, d2) ] ->
+      Bool (date_compare (y1, m1, d1) (y2, m2, d2) =*= 1)
   | G.GtE, [ Int i1; Int i2 ] -> Bool (i1 >= i2)
   | G.GtE, [ Float i1; Float i2 ] -> Bool (i1 >= i2)
   | G.GtE, [ Int i1; Float i2 ] -> Bool (float_of_int i1 >= i2)
   | G.GtE, [ Float i1; Int i2 ] -> Bool (i1 >= float_of_int i2)
+  | G.GtE, [ Date (y1, m1, d1); Date (y2, m2, d2) ] ->
```
aryx: if the date was a float, you would not need all those extra cases. The case for Float would just work.

aryx (changes requested):

aryx:
I like types, but a date is actually more complex, it also has hours, min, seconds, so instead of defining
our own type, simpler to use what OCaml did and simply use a float.

src/engine/Eval_generic.ml
SophiaSR: I think the problem with this is that the user requested being able to do something like date($X) (which I represent strptime, the python syntax for that functionality) where $X is a string of (for example) format "2023-04-15" not just a float. If treating dates as just floats, then would still need to do conversions from these strings into floats. Could represent with just floats, but it will still require a good amount of conversion to do what the user wanted either way. I'm also worried that if we just use a float, and then convert these strings to a float representation then we lose timezones.

SophiaSR: Or not lose timezones, but need to adjust more to timezones, happy to change to floats though if that is the consensus

src/engine/Eval_generic.ml
emjin: I have an experiment to derive compare (https://github.com/returntocorp/semgrep/pull/7996) which might be useful here

tests/rules/date_comparison.py
SophiaSR: Added!

src/engine/Eval_generic.ml
SophiaSR: Switched to using float representation as @aryx suggested, so no longer have separate date representation

aryx:
`this one finding was not expected: tests/rules/date_comparison.py:4 (tests/rules/date_comparison.py:4:0: "Semgrep match found": Semgrep found a match)`

tests/rules/date_comparison.py
```diff
@@ -0,0 +1,6 @@
+#OK
+day1 = "3000-01-15"
+#ERROR: match
```
aryx: it's a bit ugly, but tests under tests/patterns/ use the #ERROR: syntax to specify where something should match,
but in tests/rules/ we use the #ruleid: xxx syntax.
That's why CI is failing because it finds a match at this line, but there's no #ruleid:  before.
Look at the other test files in this directory for example.

aryx (approved):

src/engine/Eval_generic.ml
```diff
@@ -45,6 +45,8 @@ type value =
   | Float of float
   | String of string (* string without the enclosing '"' *)
   | List of value list
+  (*year, month, date for now, can maybe add support for other formats later?*)
```
aryx: you can remove this comment now

aryx (commented):
Ok next step is to make a PR on semgrep-docs to add those 2 functions.

aryx (approved):

PR #8038, opened 2023-06-13
Allow @ in rule id validation

pachuka:
PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [ ] Tests included or PR comment includes a reproducible test plan
- [ ] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

Related to https://github.com/returntocorp/semgrep/pull/8026

For folks that publish their `semgrep` rules as namespaced `npm` packages the current validation would be a breaking change.

For example for rules at the path:

`node_modules/@myorg/semgrep-rules/dist/rules.yml`

This gets run as:
`"node_modules.@myorg.semgrep-rules.dist.rule-name"`

which fails with:

```
[ERROR] Fatal error at line :1:
 Rule.ID.Malformed_rule_ID("node_modules.@myorg.semgrep-rules.dist.rule-name")
====[ BEGIN error trace ]====
Rule.ID.Malformed_rule_ID("node_modules.@myorg.semgrep-rules.dist.rule-name")
Raised at Rule.ID.of_string in file "src/core/Rule.ml", line 73, characters 45-72
Called from Parse_rule.parse_one_rule in file "src/parsing/Parse_rule.ml", line 1613, characters 5-34
Called from Parse_rule.parse_generic_ast.(fun) in file "src/parsing/Parse_rule.ml", line 1709, characters 22-47
```

PR #8057, opened 2023-06-15
Sanitize rule IDs that are rewritten by prepending a file path.

mjambon:
Follow-up to https://github.com/returntocorp/semgrep/pull/8038 and https://github.com/returntocorp/semgrep/pull/8041

Any special character in the file path is removed silently. However, the original rule ID as specified in the rule file remains untouched and must be a valid rule ID.

PR checklist:


jbergler (approved):

cli/tests/e2e/test_rule_id.py
```diff
@@ -0,0 +1,20 @@
+# Test rule ID syntax
+import pytest
+from tests.fixtures import RunSemgrep
+
+
+@pytest.mark.osempass
```
aryx, line 6: yeah :) osempass :)

src/core/Rule.ml
```diff
@@ -66,17 +69,20 @@ end = struct
...
-        introducing major breakage.
-        We may need to support other characters such as '@'
-        as used in Node.js file paths.
-        See https://github.com/returntocorp/semgrep/pull/8038
+  let validate =
+    let rex = SPcre.regexp "^[a-zA-Z0-9._-]*$" in
+    fun str -> SPcre.pmatch_noerr ~rex str
+
+  let sanitize_string str =
+    let buf = Buffer.create (String.length str) in
```
aryx, line 77: you could use a regexp replace, like you did in Python

aryx (approved):

src/core/Rule.ml
mjambon: sure, but are there benefits? I'd have to look up the function calls and make sure the regexp is compiled only once (which I didn't do it for Python). A decent `String` module would provide `String.filter` and `String.replace` in the standard library and I'd use one of these.

PR #7972, opened 2023-06-06
Add docker non-root user

cgdolan:
PR checklist:

- [ ] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [ ] Tests included or PR comment includes a reproducible test plan
- [ ] Documentation is up-to-date
- [ ] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [ ] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #8058, opened 2023-06-15
Remove the semgrep-dev docker image

aryx:
This extra docker image was just complicating things unnecessarily given the size
of the extra packages (bash, curl, jq), used in a few of our
CI jobs are not adding much to the size of the semgrep docker image.
KISS.

Worse, certain updates to the docker image such as https://github.com/returntocorp/semgrep/pull/7972
was causing errors in the semgrep dev docker image that was not detected until the next PR.
As chris said: "Dev docker container should probably build off the container that's built in the current CI run, not semgrep:develop, so that test failures are caught in the PR that introduces them not the next PR."
As martin suggested, simpler to just get rid of it.

test plan:
wait for green CI checks

PR checklist:


aryx:
At least those errors are not happening anymore:
https://github.com/returntocorp/semgrep/actions/runs/5273263956/jobs/9536471217?pr=8056
(well we're cheating a bit because we just remove the job ...)

aryx:
A possible fix suggested by bruno was simply to modify semgrep-dev with:
```
USER root

RUN apk add --no-cache \
  bash \
  curl \
  jq

USER semgrep
```
but even simpler to just remove the semgrep-dev extra image.

aryx:
Another fix suggested by Chris:
```
Getting rid of the dev container sgtm, although I think it should be fairly straightforward to keep it using multi-stage docker builds (we do this for the app). tl;dr: put all the changes from the dev docker file into the regular docker file, then build two different images by invoking docker with the --target flag: https://stackoverflow.com/questions/49028097/use-multi-stage-docker-files-for-outputting-multiple-images
```

PR #8073, opened 2023-06-16
fix: [ci] revert back to root user in docker image

spencerdrak:
This reverts commit ea884475618c8b63c4794e44dc09447a902cb140.

Dogfooding has encountered some issues with this change, I'll propose a revert and we can sort this out without needed to fix forward.

PR checklist:

### DIFF
--- a/.github/workflows/test-semgrep-pro-latest.yaml
+++ b/.github/workflows/test-semgrep-pro-latest.yaml
@@ -5,18 +5,19 @@
 # If this workflow succeeds, then it should be safe to release Semgrep and
 # not immediately be incompatible with Semgrep Pro.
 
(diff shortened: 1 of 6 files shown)


cgdolan (approved):

spencerdrak:
@brandonspark should the `test-semgrep-pro-latest` workflow be running on PRs? It seems like it's a duplicate of `test-semgrep-pro`, but it pulls develop image which is breaking this workflow since that image is having issues rn. 

Is it safe to not run that check on PRs?

spencerdrak:
Chatted with brandon. It appears the workflow was duplicated and causing issues in any case, so we set it to only run on pushes to develop.

PR #8075, opened 2023-06-16
feat(multistep): supporting changes for multistep rules

emjin:
See [PR in semgrep-proprietary](https://github.com/returntocorp/semgrep-proprietary/pull/792) for details

PR checklist:


libs/ast_generic/AST_generic.ml
```diff
@@ -273,7 +273,7 @@ module IdInfoId = Gensym.MkId ()
  *)
 (* a single unique gensym'ed number. *)
 type sid = SId.t
-and resolved_name = resolved_name_kind * sid
+and resolved_name = resolved_name_kind * sid [@@deriving show, eq]
```
aryx: you should not add @@deriving in the middle of a mutually recursive type.
There is already a deriving show and eq further below. There is no need for this one.

libs/ast_generic/AST_generic.ml
```diff
@@ -548,12 +548,15 @@ and qualifier =
 (*****************************************************************************)
 and id_info = {
   id_resolved : resolved_name option ref;
+      [@equal
```
aryx, line 547: why do we need to do anything special here? Why the default autogenerated eq does not work here?

libs/ast_generic/AST_utils.ml
```diff
@@ -40,26 +40,57 @@ module Node_ID = Gensym.MkId ()
    so as to select structural or referential equality.
 *)
 
-type busy_with_equal = Not_busy | Structural_equal | Referential_equal
+type busy_with_equal =
+  | Not_busy
+  | Structural_equal
+  | Referential_equal
+  | Syntactic_equal
```
aryx, line 52: you need to add a comment 4 lines above explaining what is a syntactic equal and why we need yet another form of equality.

libs/ast_generic/AST_utils.ml
```diff
@@ -40,26 +40,57 @@ module Node_ID = Gensym.MkId ()
...
 
 (* global state! managed by the with_equal_* functions *)
 let busy_with_equal = ref Not_busy
 
+let equal_id_info equal a b =
+  match !busy_with_equal with
+  | Not_busy -> failwith "Call AST_utils.with_xxx_equal to avoid this error."
+  | Syntactic_equal -> true
```
aryx, line 60: need a comment explaining why for syntatic_equal we don't care about id_info, and in which context we need that (I guess for join-mode, but could be nice to repeat it here).

aryx (approved):

PR #8078, opened 2023-06-18
dataflow: Set a strict timeout on fixpoint computations

IagoAbal:
We shouldn't allow a bug, a bad rule, or a "bad" file ruining a Semgrep run, better to abort the analysis and move on.

Note that `Time_limit.set_timeout` cannot be nested.

test plan:
Run `p/default-v2` on a subset of 31 repos from stress-test-monorepo, the average speedup was 1.2x.
(No difference in number of findings.)

PR checklist:

### DIFF
--- a/src/configuring/Limits_semgrep.ml
+++ b/src/configuring/Limits_semgrep.ml
@@ -1,4 +1,11 @@
 (*****************************************************************************)
+(* Const/sym ("svalue") propagation *)
+(*****************************************************************************)
+
+(* Timeout in seconds. *)
+let svalue_prop_FIXPOINT_TIMEOUT = 0.1
+
+(*****************************************************************************)
 (* Taint analysis *)
 (*****************************************************************************)
 
@@ -14,6 +21,9 @@
  * not propagate taint for data with Boolean and integer type. Improving some
  * of the data structures involved may help too.
  *)
+
+(* Timeout in seconds. *)
+let taint_FIXPOINT_TIMEOUT = 0.1
 
 (** Bounds the number of l-values we can track. *)
 let taint_MAX_TAINTED_LVALS = 100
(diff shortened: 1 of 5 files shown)


aryx (approved):

aryx:
But shouldn't the general Process_limits timeout process handle those cases too? Do we need an extra timeout mechanism?
(maybe would be good to add a comment why yet another timeout mechanism).

IagoAbal:
> But shouldn't the general Process_limits timeout process handle those cases too? Do we need an extra timeout mechanism?
> (maybe would be good to add a comment why yet another timeout mechanism).

I'll add a comment. But in short, `Time_limit` cannot be nested (right now) and if a timeout triggers seems to be pretty expensive. So expensive that setting a lower timeout than the default (30s) can actually make your runs _slower_! So unless we fix that, I think it's good to have as a last-resort hard-timeout, but not as a general timeout mechanism.

I'll run a quick test to get some actual data, it has been a while since I experimented with this.

IagoAbal:
I'll experiment with #8082 too, but TBH not being able to nest `Time_limit` already makes this necessary in Pro right now. There is already a global `Time_limit` for the entire "build_idx_and_setup_hooks_for_deep_semgrep", but I also want to make sure we don't spend too much time inferring taint signatures. I'll add a note to revisit this later on.

src/configuring/Limits_semgrep.ml
```diff
@@ -1,3 +1,12 @@
+(*****************************************************************************)
+(* Const/sym ("svalue") propagation *)
+(*****************************************************************************)
+
+(* TODO: Report these timeouts as errors in 'Report.match_result' *)
+(* Timeout in seconds.
+ * So e.g. the perf of svalue-prop does not prevent rules from running on a file. *)
```
aryx: I would add also here the comment on why we can't use just Time_limit and that we need extra timeout mechanism.

aryx (approved):

PR #8082, opened 2023-06-20
cli: Set default timeout to just 3s

IagoAbal:
PR checklist:


aryx (changes requested):
you need to change also in osemgrep, in src/osemgrep/cli_scan/Scan_CLI.ml I think, look at the default global constant.

IagoAbal:
Just opened this for comparison with PR #8078, but here the average speedup is just ~1.07x (vs ~1.20x), and it was significantly faster for only 5 (out of 31) repos (vs 15).

PR #8085, opened 2023-06-20
C: Const prop not working in stack array decls

### LINKED ISSUE #8037: Constant propagation not happening in basic C function case
**Describe the bug**
The function is simply:
```
void fun() {
  int fixed = 42;
  // ruleid: match_fixed_size_buffer
  char a[42];
  // ruleid: match_fixed_size_buffer
  char b[fixed];
  // ok: match_fixed_size_buffer
  char c[unknown];
}
```

But the rule
```
rules:
  - id: match_fixed_size_buffer-copy
    patterns:
      - pattern: |
          $TYPE $BUF[42];
    message: var $BUF - size $SIZE
    languages:
      - c
    severity: WARNING
```
fails to match

**To Reproduce**
https://semgrep.dev/playground/s/b36Q

yosefAlsuhaibani:
closes #8037 

It used to be the case that we weren't able to match the following with the pattern `$TYPE $NAME[404];`:
```c
int main() {
  int bad = 404;
  #ERROR:
  int arr[bad]; //old semgrep wouldn't match this line!
  return 0;
}
```
but now we do; the fix is by translating array decls as follows in IL:
```
_tmp := bad 
DEF arr
```

test plan: `make test` + wrote a couple of test sets in `semgrep/tests/patterns/c`; one that is straight line code and the other has nested control flow.

PR checklist:


changelog.d/gh-8037.fixed
```diff
@@ -0,0 +1,13 @@
+Constant propagation is now applied to declerations inside stack array declerations in C; so
```
aryx: typo declerations (2 times)

aryx (approved):

changelog.d/gh-8037.fixed
yosefAlsuhaibani:
```
Error: Unbound value declerations 
Hint: did you mean declarations?
```

emjin (approved):
Nice!

changelog.d/gh-8037.fixed
```diff
@@ -0,0 +1,13 @@
+Constant propagation is now applied to declarations inside stack array declarations in C; so
```
IagoAbal: I believe you wanted to write "... is now applied inside stack array ..." and the "to declarations" that is there in the middle may be the result of you having rewritten this sentence?

yosefAlsuhaibani: Ah Indeed; removed the `declarations inside`.

src/analyzing/AST_to_IL.ml
```diff
@@ -1241,6 +1241,14 @@ and stmt_aux env st =
       let ss, e' = expr_with_pre_stmts env e in
       let lv = lval_of_ent env ent in
       ss @ [ mk_s (Instr (mk_i (Assign (lv, e')) (Related (G.S st)))) ]
+  | G.DefStmt (ent, G.VarDef { G.vinit = None; vtype = Some ty }) -> (
```
IagoAbal: I think it's better to write this pattern instead:
```
... vtype = Some { t = G.TyArray ((_, Some e, _), _); _} ...
```
so that if it's not an array type as you want, then we don't end up in your default `[]` case, which is "masking" whatever the default case used to be.

emjin:
uhh yikes the benchmarks are probably not you

src/analyzing/AST_to_IL.ml
```diff
@@ -1241,6 +1241,14 @@ and stmt_aux env st =
...
+      match ty.t with
+      | G.TyArray ((_, Some e, _), _) ->
+          let ss, e' = expr_with_pre_stmts env e in
```
IagoAbal: Also it's worth having a comment here explaining what is the purpose of this.

IagoAbal (approved):

PR #8096, opened 2023-06-21
tainting: Improvements to taint_assume_safe_{booleans,numbers}

IagoAbal:
Mainly:

- Use type info provided by explicit casts.
- Use const-prop info to infer types.
- Guess the type of `length` when used as an attribute.
- Add `long` and `Long` as "synonyms" for `int` and `Integer`, and `double` for `float`.
- Sometimes we cannot infer the type of the RHS but the LHS is a variable with e.g. a number type, so we can avoid tracking it.

test plan:
make test # new tests

PR checklist:

### DIFF
--- a/src/typing/Typing.ml
+++ b/src/typing/Typing.ml
@@ -22,22 +22,19 @@ module G = AST_generic
 let rec type_of_expr lang e : G.name Type.t * G.ident option =
   match e.G.e with
   | G.L lit ->
-      let t =
-        match lit with
-        (* NB: We could infer Type.Number for JS int/float literals, but we can
-         * handle that relationship in matching and we can be more precise for
-         * now. One actual rule uses `float` for a typed metavariable in JS so
-         * let's avoid breaking that for now at least. *)
-        | G.Int _ -> Type.Builtin Type.Int
-        | G.Float _ -> Type.Builtin Type.Float
-        | G.Bool _ -> Type.Builtin Type.Bool
-        | G.String _ -> Type.Builtin Type.String
-        | _else_ -> Type.NoType
-      in
+      let t = type_of_lit lit in
       (t, None)
+  | G.DotAccess
+      (_obj, _, FN (Id (("length", _), { id_type = { contents = None }; _ })))
+    when lang =*= Lang.Java ->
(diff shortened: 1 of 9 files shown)


src/typing/Typing.ml
```diff
@@ -22,22 +22,19 @@ module G = AST_generic
...
-      in
+      let t = type_of_lit lit in
       (t, None)
+  | G.DotAccess
+      (_obj, _, FN (Id (("length", _), { id_type = { contents = None }; _ })))
+    when lang =*= Lang.Java ->
+      (* TODO: Fix guess_type_of_dotaccess to take an "expected type" (as in bidirectional
+       * type checking) so that we can distinguish `length` as a method (e.g. `String`) and
+       * `length` as an attribute (arrays). *)
+      (Type.Builtin Type.Int, None)
```
nmote: I'd really like to keep the guessing parts of type inference separate. This is a heuristic, not a part of proper type inference. Plus, won't this make it so that we never guess `String.length` to be a method?

Is there a reason that this can't go in `guess_type_of_dotaccess` and instead of using bidirectional type inference, we check the type of the receiver and use that to inform the guess?

nmote (changes requested):
Nice set of improvements! I have one concern inline that I'd like to discuss though.

src/typing/Typing.ml
IagoAbal: To be honest `type_of_expr` is also somewhat heuristic, so I think the whole "Typing" is about guessing types, regardless. I'll try something. But the type of the "receiver" isn't always available, since we're working with partial information.

nmote (changes requested):
I'm a bit confused by the whole wildcard thing. What set of values does a wildcard type model? Could you add some comments at the type definition?

IagoAbal:
> I'm a bit confused by the whole wildcard thing. What set of values does a wildcard type model? Could you add some comments at the type definition?

If you're still confused then please approve and I will just remove that whole code. And I will open a new PR with a different reviewer. Precisely I didn't want to do this kind of work in this PR, because it was out of scope. And I want to keep it somewhat within scope.

PR #8099, opened 2023-06-21
tainting: Improvements to taint_assume_safe_{booleans,numbers}

IagoAbal:
Mainly:

- Use type info provided by explicit casts.
- Use const-prop info to infer types.
- Guess the type of `length` when used as an attribute.
- Add `long` and `Long` as "synonyms" for `int` and `Integer`, and `double` for `float`.
- Guess types for `decode`, `valueOf` and `parseXyz` methods in classes `Boolean`, `Integer`, `Long`, and `Float`.
- Sometimes we cannot infer the type of the RHS but the LHS is a variable with e.g. a number type, so we can avoid tracking it.

test plan:
make test # new tests

PR checklist:

### DIFF
--- a/src/typing/Typing.ml
+++ b/src/typing/Typing.ml
@@ -27,9 +27,15 @@ let rec type_of_expr lang e : G.name Type.t * G.ident option =
   | G.DotAccess
       (_obj, _, FN (Id (("length", _), { id_type = { contents = None }; _ })))
     when lang =*= Lang.Java ->
-      (* TODO: Fix guess_type_of_dotaccess to take an "expected type" (as in bidirectional
-       * type checking) so that we can distinguish `length` as a method (e.g. `String`) and
-       * `length` as an attribute (arrays). *)
+      (* TODO: Move this to 'guess_type_of_dotaccess', but somehow we need to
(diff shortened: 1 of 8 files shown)


src/typing/Typing.ml
```diff
@@ -22,22 +22,19 @@ module G = AST_generic
...
-        | G.Float _ -> Type.Builtin Type.Float
-        | G.Bool _ -> Type.Builtin Type.Bool
-        | G.String _ -> Type.Builtin Type.String
-        | _else_ -> Type.NoType
-      in
+      let t = type_of_lit lit in
       (t, None)
+  | G.DotAccess
+      (_obj, _, FN (Id (("length", _), { id_type = { contents = None }; _ })))
+    when lang =*= Lang.Java ->
```
nmote, line 29: As discussed I'd appreciate a comment here explaining how this does not conflict with the existing guessing for e.g. `string.length()`

nmote (approved):
As we discussed I think this is the best option for the moment. Thanks for talking through the alternatives with me.

nmote (approved):

PR #8100, opened 2023-06-21
Update parser to accept post-pip614 decorators

### LINKED ISSUE #4946: Python's Relaxing Grammar Restrictions On Decorators (PEP 614) is not supported
**Describe the bug**

https://peps.python.org/pep-0614/ (in Python 3.9, released 2020-10-05) generalises Python decorator syntax from `@a.b.c(any, arguments[here])` to accept most expressions, like `@foo[0].bar`. It seems semgrep currently rejects any file that contains this syntax:

```
[WARN] Semgrep Core — Syntax error
An error occurred while invoking the Semgrep engine. Please help us fix this by creating an issue at https://github.com/returntocorp/semgrep

At line target.py:5: `[` was unexpected
```

**To Reproduce**

Have any Python file (not rule) that contains a decorator using the relaxed syntax. No rules can be evaluated on this file. Example (this code runs and prints `foo`):

```python
x = "bad string"

decorators = [lambda f: f]

@decorators[0]
def foo():
    print("foo")

foo()
```

https://semgrep.dev/s/eJPZ

```json
{"code":3,"level":"warn","message":"[WARN] Semgrep Core — Syntax error\nAn error occurred while invoking the Semgrep engine. Please help us fix this by creating an issue at https://github.com/returntocorp/semgrep\n\nAt line target.py:5: `[` was unexpected\n","path":"target.py","type":"Syntax error"}
```

**Expected behavior**

The file should be parseable, and most rules should be evaluated. 

(To be clear, this isn't about improved support for rules about decorators, that's covered by issues like #4722.)

**Screenshots**
N/A

**Environment**
semgrep.dev

**Use case**

We use the generalised decorator syntax in some cases, and so we're having several files not checked with semgrep.

yosefAlsuhaibani:
closes #4946

Semgrep used to not be able to parse the following:
```python
#Doesn't parse this post-pip614 decorator
@why := i.s.this.al["lowed"].oh("my")
def think():
  return None
```
In other words: our Python grammar defined a decorator to be `@ dotted_name arg_list_opt NEWLINE` when it should be `@ named_expr NEWLINE` instead as per [pip0614](https://peps.python.org/pep-0614/).

Now we are able to parse and match on pip614 decorators!

Test plan: `make test` + wrote a couple of test sets in `tests/patterns/python/`.

PR checklist:


aryx:
Love the example in the PR subject :)

tests/patterns/python/pip614-extended-decorator-grammer1.py
```diff
@@ -0,0 +1,15 @@
+#ERROR:
```
aryx, line 1: if the fix address just a parsing issue with a language, there is no need to add a .sgrep example for it.
Instead you can just add the new test file in tests/parsing/python/

If you also want to make sure that we accept the new syntax in the pattern, then yes you want a .sgrep file.

aryx (approved):

tests/patterns/python/pip614-extended-decorator-grammer1.py
yosefAlsuhaibani: Understood: I was worried if the choice of translation between Python AST ==> Generic AST would ruin attribute matching, and was why I included these patterns test.

it did catch a matching bug for when I chose to translate all decorators as `G.OtherAttribute`, instead of attempting to trans some as `G.NamedAttr`.

aryx:
I get warnings when compiling though about arglist_paren2_opt is unreachable as well as decorator_name.
Could you remove those dead grammar rules?

PR #8102, opened 2023-06-22
feat(rule syntax): Metavariable Type Extension for Semgrep Rule Syntax

ihji:
This commit introduces a dedicated field for annotating the type information of metavariables. By adopting this approach, instead of relying solely on language-specific casting syntax, we provide an additional way to enhance the overall usability by eliminating the need to write redundant type cast expressions for a single metavariable.

Moreover, the new syntax brings other benefits, including improved support for target languages that lack built-in casting syntax. It also promotes a unified approach to expressing type, pattern, and regex constraints for metavariables, resulting in improved consistency across rule definitions.

Current syntax:
```
rules:
  - id: no-string-eqeq
    severity: WARNING
    message: find errors
    languages:
      - java
    patterns:
      - pattern-not: null == (String $Y)
      - pattern: $X == (String $Y)
```

Proposed syntax:
```
rules:
  - id: no-string-eqeq
    severity: WARNING
    message: find errors
    languages:
      - java
    patterns:
      - pattern-not: null == $Y
      - pattern: $X == $Y
      - metavariable-type:
          metavariable: $Y
          type: String
```

PR checklist:

### DIFF
--- a/src/core/Rule.ml
+++ b/src/core/Rule.ml
@@ -173,6 +173,7 @@ and metavar_cond =
    *)
   | CondRegexp of
       MV.mvar * Xpattern.regexp_string * bool (* constant-propagation *)
+  | CondType of MV.mvar * Xlang.t option * string * AST_generic.type_
   | CondAnalysis of MV.mvar * metavar_analysis_kind
   | CondNestedFormula of MV.mvar * Xlang.t option * formula
 
--- a/src/core/Rule.ml
+++ b/src/core/Rule.ml
@@ -173,7 +173,11 @@ and metavar_cond =
    *)
   | CondRegexp of
       MV.mvar * Xpattern.regexp_string * bool (* constant-propagation *)
-  | CondType of MV.mvar * Xlang.t option * string * AST_generic.type_
+  | CondType of
+      MV.mvar
+      * Xlang.t option (* when the type expression is in different lang *)
+      * string (* raw input string saved for regenerating rule yaml *)
+      * AST_generic.type_ (* LATER: could parse lazily, like the patterns *)
   | CondAnalysis of MV.mvar * metavar_analysis_kind
   | CondNestedFormula of MV.mvar * Xlang.t option * formula
 
--- a/src/engine/Match_search_mode.ml
+++ b/src/engine/Match_search_mode.ml
@@ -24,6 +24,7 @@ module RP = Report
 module RM = Range_with_metavars
 module E = Semgrep_error_code
 module ME = Matching_explanation
+module GG = Generic_vs_generic
 open Match_env
 
 let logger = Logging.get_logger [ __MODULE__ ]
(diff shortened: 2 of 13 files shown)


src/core/Rule.ml
```diff
@@ -173,6 +173,7 @@ and metavar_cond =
    *)
   | CondRegexp of
       MV.mvar * Xpattern.regexp_string * bool (* constant-propagation *)
+  | CondType of MV.mvar * Xlang.t option * string * AST_generic.type_
```
aryx: can you add a comment about what is the string and why we need an Xlang.t here?

src/matching/Generic_vs_generic.mli
```diff
@@ -22,3 +22,11 @@ val m_any : AST_generic.any Matching_generic.matcher
 
 val hook_find_possible_parents :
   (AST_generic.dotted_ident -> AST_generic.name list) option ref
+
+val m_compatible_type :
```
aryx, line 27: you can add a comment like (* used now for metavariable-type: in Match_search_mode.ml *)

src/optimizing/Analyze_rule.ml
```diff
@@ -317,6 +317,7 @@ and metavarcond_step1 x =
   | R.CondNestedFormula _ -> None
   | R.CondRegexp (mvar, re, const_prop) ->
       Some (MvarRegexp (mvar, re, const_prop))
+  | R.CondType _ -> None
```
aryx, line 321: you can add a comment like
(* TODO? maybe we should extract the strings from the type constraint *)

src/engine/Match_search_mode.ml
```diff
@@ -547,6 +548,38 @@ let rec filter_ranges (env : env) (xs : (RM.t * MV.bindings list) list)
...
+                 in
+                 let matches =
+                   GG.m_compatible_type lang
+                     (mvar, Tok.unsafe_fake_tok "")
+                     t e env
+                 in
+                 logger#info
+                   "range %d-%d filtered from metavar %s type mismatch."
+                   r.r.start r.r.end_ mvar;
+                 matches <> [] |> map_bool r
```
aryx, line 582: in theory the type can also contain metavariables, but we probably don't want to use that in other parts of the rules,
so it's probably fine to just check whether the match is empty or not and disregard any new metavariable bindings
discovered during the process.
Maybe worth a comment though.

src/core/Rule.ml
aryx: Maybe also add a comment like (* LATER: we could also parse the type lazily, like we for the patterns *)

src/parsing/Parse_rule.ml
```diff
@@ -453,6 +453,31 @@ let parse_python_expression env key s =
...
+      |> unwrap_typed_metavar env key
+  | Xlang.LRegex
+  | Xlang.LSpacegrep
+  | Xlang.LAliengrep ->
+      error_at_key env.id key
+        "`type` is not supported with regex, spacegrep or aliengrep."
+
+and wrap_type_expr env key lang str =
+  match lang with
+  | Lang.Java -> spf "(%s x)" str
```
aryx, line 472: what is this 'x'?
Maybe add a comment like  (* coupling: we use 'x' here because it's a hack we use in XXX.ml to parse YYY *)

aryx: Or just explain that we don't care about this 'x' in unwrap_typed_metavar.

aryx (approved):
Great work!

right now you use a hack to parse Java types by just abusing the Cast syntax, but I think in the futur we may need a different approach for other languages.

aryx:
you need a changelog entry with a good example in it, and later improve semgrep-docs

ihji:
#8119

PR #8107, opened 2023-06-22
Fixing php pattern parsing for attributes

### LINKED ISSUE #7398: Add support for multiple PHP attributes
**Describe the bug**

PHP 8 introduced [attributes](https://www.php.net/manual/en/language.attributes.overview.php), which Semgrep currently supports. However, it doesn't support specifying multiple attributes at the same time in a single pattern (like with Python decorators). E.g.

```php
#[Attr1]
#[Attr2]
function $FUNC (...) { ... }
```

**To Reproduce**

- Two attributes don't work: https://semgrep.dev/s/1x1D
- One attribute does: https://semgrep.dev/s/9wx4, https://semgrep.dev/s/ygpr

**Expected behavior**

Ability to specify multiple attributes at the same time.

**Screenshots**

**Environment**

**Use case**

SophiaSR:
Previously, php pattern parsing could only parse multiple attributes in the format `#[Attr1,Attr2,..]` This fix adds support for parsing patterns such as  
`#[Attr1]
#[Attr2]
...`

PR checklist:

### DIFF
--- a/languages/php/menhir/parser_php.mly
+++ b/languages/php/menhir/parser_php.mly
@@ -904,7 +904,9 @@ return_type: ":" type_php                 { $1, $2 }
 (* Attributes *)
 (*************************************************************************)
 (* PHP 8 extension (was using << >> in HPHP) *)
-attributes: "#[" listc(attribute) "]" { ($1, $2, $3) }
+attributes:
+| "#[" listc(attribute) "]" { ($1, $2, $3) }
+| attributes attributes {match ($1, $2) with ((op,xs1,cl),(_,xs2,_)) -> (op,xs1@xs2,cl)}
 
 attribute:
  | ident                                  { Attribute $1 }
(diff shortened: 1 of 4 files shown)


languages/php/menhir/parser_php.mly
```diff
@@ -904,7 +904,9 @@ return_type: ":" type_php                 { $1, $2 }
 (* Attributes *)
 (*************************************************************************)
 (* PHP 8 extension (was using << >> in HPHP) *)
-attributes: "#[" listc(attribute) "]" { ($1, $2, $3) }
+attributes:
+| "#[" listc(attribute) "]" { ($1, $2, $3) }
+| attributes attributes {match ($1, $2) with ((op,xs1,cl),(_,xs2,_)) -> (op,xs1@xs2,cl)}
```
aryx: we usually use lp/rp (left paren, rparen), and in that case
we probably want the left paren from $1 and right paren of $2, so that we get
the right range.

aryx: We need the right set of tokens because we use them to get the range of constructs, and for autofix to know
all the parts we need to remove.

aryx (approved):

PR #8112, opened 2023-06-22
feat(matching): match record field of ellipsis to any definition

brandonspark:
## What:
This PR establishes an equivalence between a `FieldDefColon` of an ellipsis and any other definition kind.

## Why:
`FieldDefColon` of an ellipsis arises from a record field, something like `{x : ...}` in Javascript.

However, a record might contain a definition which is not a `FieldDefColon`! A good example of this is 
```
const x = foo({
    func: () => {
        return 1;
    },
})
```
which is actually a record of a `FuncDef`. However, based on the pattern's translation, we should want to match the lambda, even if it's in a `FuncDef`. So we establish this equivalence.

## How:
Just wrote the case in `Generic_vs_generic`.

## Alternatives considered:
We could alternatively consider establishing a `definition_kind` variant which itself represents an ellipsis, and translate into that instead of the original `FieldDefColon`.

I took a poll in Slack, and this seemed to be the better option (https://semgrepinc.slack.com/archives/C01NXGX2EHZ/p1687384401974879). I also just like it better. We can reconsider if this comes up more often.

## Test plan:
`make test`

Closes PA-2878

PR checklist:


tests/patterns/ts/record_functions.ts
```diff
@@ -0,0 +1,20 @@
...
+})
+
+// ERROR: match
+const y = foo({
+    func: function (opts) {
+        return 2;
+    },
+})
+
+// ERROR: match
```
emjin, line 15: It's nice to also have an example of something that shouldn't match

src/matching/Generic_vs_generic.ml
```diff
@@ -2800,6 +2800,20 @@ and m_list__m_type_parameter a b =
...
 and m_definition_kind a b =
   match (a, b) with
+  (* We maintain an equivalence between a FieldDefColon of an ellipsis
+     with any other definition. This is because FieldDefColon appears
+     when describing the RHS of a record, so this describes a record
+     like
+     { x = ... }
+     The FieldDefColon is only if it's not some other kind of
+     `definition_kind` structure, however, like a `FuncDef`, so we
+     would like to be able to match those with the ellipsis anyways.
```
emjin: This one sentence is a bit confusing in wording, could use some cleaning up.

Great comment overall!

src/matching/Generic_vs_generic.ml
```diff
@@ -2800,6 +2800,20 @@ and m_list__m_type_parameter a b =
...
+
+     A concrete example is TS, when you have
+     { func: function (opts) { return "whatever"; } }
+     which is a record containing an entry which is a `FuncDef`.
+  *)
+  | G.FieldDefColon { vinit = Some { e = Ellipsis _; _ }; _ }, _ -> return ()
```
emjin: I would feel more comfortable enumerating the options for B here, in case we add some new construct that shouldn't match. If you're confident that could never happen, making it instead `_any_other_kind_of_def` would be helpful to show that you've thought about it.

emjin (approved):

PR #8115, opened 2023-06-22
feat!: kill numeric capture group mvars

brandonspark:
## What:
This PR kills regex capture group mvars like `$1`, `$2`, which are implicitly introduced by capture groups, and instead only allows metavariables from explicitly named capture groups.

## Why:
The primary motivation is that if capture groups introduce mvars, two regular expression patterns can both introduce the same metavariable, such as `$1`, and cause both not to unify. This is extremely untransparent to the user, though, and this kind of implicit declaration is a big no-no. Metavariables should be introduced obviously.

The secondary motivation is that, according to our metavariable naming schema (https://github.com/returntocorp/semgrep/blob/c27d999964c26a8ff3f2e5597ab6fe1422d571cf/src/core/Metavariable.ml#L217), metavariables such as `$1` and `$2` are not even metavariables! This was causing issues where stuff like `metavariable-regex` could not use the value of those metavariables, because it would fail a metavariable name check (see https://linear.app/semgrep/issue/PA-2514/cannot-metavariable-regex-on-regex-capture-group-mvars)

## How:
We analyze the regular expression using the `Pcre` library to instead get all of the entries by named capture groups (https://www.regular-expressions.info/named.html), and introduce _those_ (prefixed with a `$`) as metavariables.

I also added a check in `Parse_rule` to see if the named capture groups of a regular expression are valid metavariables or not. If they are not, I emitted a warning, because I wasn't sure if there was legitimate use for such capture group unless it was for binding a metavariable. I would also be OK with rejecting the rule. Open to suggestions.

## Alternatives considered:
I considered introducing the capture group metavariables again, but changing their names to have an underscore first (such as `$_1`, `$_2`, etc), which would solve the secondary concern. This wouldn't solve the implicit unification, though, which I think is a big problem, so I chose to only allow the explicitly named capture groups to bind metavariables.

## Downscoping:
For this PR, I downscoped `metavariable-regex`, which could theoretically be treated the same way, and be allowed to introduce capture group metavariables. Since we didn't do it before, though, even for numeric metavariables, I decided that was out of scope for this PR.

## Test plan:
`make test`

Closes PA-2514
Closes PA-2765

PR checklist:


spencerdrak:
@brandonspark can you rebase/merge develop into this PR? It's showing a lot of files changed that I don't think you've actually messed with. Thanks!

cli/tests/e2e/snapshots/test_rule_validation/test_validation_of_invalid_rules/rulesinvalid-rulesinvalid-regex-capture-group.yaml-basicstupid.py/results.json
```diff
@@ -0,0 +1,22 @@
+{
+  "errors": [
+    {
+      "code": 2,
+      "level": "error",
+      "message": "WARNING: unable to find a config; path `rules/invalid-rules/invalid-regex-capture-group.yaml` does not exist",
```
aryx: I don't think that's the error we want.

aryx (approved):
Great work!

cli/tests/e2e/snapshots/test_rule_validation/test_validation_of_invalid_rules/rulesinvalid-rulesinvalid-regex-capture-group.yaml-basicstupid.py/results.json
brandonspark: indeed it is not, unfortunately snapshot updates do not remove extraneous snapshots. this was a relic from an earlier testing phase. deleted!

PR #8118, opened 2023-06-22
feat(perf): remove more instances of Stdlib.compare

emjin:
Taint.compare_sources previously used Stdlib.compare on a few values, including Metavariable.env. Since these are metavariable bindings, which incldue tokens, the comparison is likely to consider many equal bindings non-equal. This is fine for correctness purposes but impacts perf necessary.

Instead, in this PR I write a proper compare function. I derive compare in a few cases, because writing compare by hand is not super fun. The ppx_compare extension requires you to open a module, however, so I instead copied the functions into Common (see the comment there).

My understanding is that I have to do this because this is a Jane Street library and they expect you to open Base or Core, which probably also have these functions copied, so I think my decision is reasonable. Please let me know if there's a better way.

Test plan: automated tests

Perf: Ran OpenClinica with p/default-v2. Finished in 6:07 with develop and 6:11 with my changes

PR checklist:


libs/commons/Common.ml
```diff
@@ -66,6 +66,23 @@ let ( =*= ) = ( = )
  *)
 let ( = ) = String.equal
 
+(*****************************************************************************)
+(* Comparison *)
+(*****************************************************************************)
+
+(* To use `ppx_compare`, it's necessary to have `Ppx_compare_lib.Builtin
```
aryx: Why using ppx_compare? Why not ppx_deriving and using the 'ord' which seems equivalent.
See https://github.com/ocaml-ppx/ppx_deriving

libs/commons/dune
```diff
@@ -25,6 +25,7 @@
      ppx_deriving.show
      ppx_deriving.eq
      ppx_deriving.ord
```
aryx: see we already include ppx_deriving.ord.
Did you try deriving ord instead? What was wrong with it?

aryx (approved):
Each time we define an explicit compare_xxx function it would be nice
to explain why we could not rely on ppx_deriving.ord (or ppx_compare) to autogenerate it.

libs/commons/Common.ml
emjin: Oh, thanks! I was looking for an alternative but did not find it

src/tainting/Taint.ml
```diff
@@ -104,33 +104,45 @@ let rec compare_precondition (ts1, f1) (ts2, f2) =
...
+    String.compare
+      (Rule.ID.to_string pm1.PM.rule_id.id)
+      (Rule.ID.to_string pm2.PM.rule_id.id)
+  in
+  if compare_rule_id <> 0 then compare_rule_id
+  else
+    let compare_range_loc = compare pm1.range_loc pm2.range_loc in
+    if compare_range_loc <> 0 then compare_range_loc
+    else if Metavariable.Structural.equal_bindings pm1.env pm2.env then 0
+    else Stdlib.compare pm1.env pm2.env
```
IagoAbal: Why should we use `Stdlib.compare` here?

IagoAbal:
> Edit: I don't know what happened in that previous check. I ran it again and it took 2 hours 39 minutes. Based on the output I think this run is most likely correct. The last run I did took over 4 hours without completing, but it's unclear to me if this is the result of my change or another one that happened in between, so I'll do another run tonight.

Please update when possible!

emjin:
Will retime tomorrow

Edit: failed to get to this

emjin:
It was annoying to time this accurately because latest semgrep-proprietary develop does not currently work with latest semgrep develop. I ran using Pad's branch that fixes build errors, using semgrep's develop and then my branch after merging develop.

Good news: we no longer take hours! Finished in 6:07 with develop and 6:11 with my changes. I don't know if it's significant that it took longer with my changes

src/tainting/Taint.ml
```diff
@@ -17,24 +17,25 @@ module G = AST_generic
...
+ * I just thought that if you carefully considered whether that automagic comparison
+ * is OK for each data type then you are fine... So we used `Stdlib.compare` in several
+ * places, until one day `Taint.arg` evolved, and we added an `offset` to it, and we
+ * forgot to revisit whether `Stdlib.compare` was still a good option (it wasn't)...
+ * and this led to duplicates which led to an explosion in the size of taint sets and
+ * to big perf problems. I think the only way to get warned about this is to write these
  * comparisons manually, no matter how painful it is, so the typechecker will force
- * you to revisit the comparison functions if the types change.
+ * you to revisit the comparison functions if the types change. It's safe to use
+ * ppx_deriving when all the types are primitive, but if any is not then it can be
```
IagoAbal: "can be cause"

src/tainting/Taint.ml
```diff
@@ -84,11 +85,32 @@ let length_of_call_trace ct =
...
+  (* It's important that we only return 0 if the two bindings are
+     structurally equal. Otherwise, there will be many duplicates.
+     We use Stdlib.compare in the other case because deriving
+     compare is rather difficult and the specific ordering doesn't
+     matter. *)
+  if Metavariable.Structural.equal_bindings env1 env2 then 0
+  else Stdlib.compare env1 env2
+
+let compare_matches pm1 pm2 =
+  let compare_rule_id =
```
IagoAbal: Would be good to follow the same style as in the other `compare` functions:
```ocaml
match ... with
| 0 -> ...
| other -> other
```
Unless you have a point to prefer this other style? I used to write them the same way, but I very recently changed to the `match ... with | 0 -> ...` as it feels more readable. (I believe I copied it from Brandon.)

src/tainting/Taint.ml
```diff
@@ -84,11 +85,32 @@ let length_of_call_trace ct =
 
 type sink = { pm : Pattern_match.t; rule_sink : R.taint_sink } [@@deriving show]
 
+let compare_metavar_env env1 env2 =
+  (* It's important that we only return 0 if the two bindings are
+     structurally equal. Otherwise, there will be many duplicates.
+     We use Stdlib.compare in the other case because deriving
+     compare is rather difficult and the specific ordering doesn't
+     matter. *)
+  if Metavariable.Structural.equal_bindings env1 env2 then 0
```
IagoAbal, line 94: What if we had a `Set` module for `Metavariable.bindings` and used it here ? `Set` has a `compare` already implemented and it would not be sensitive to ordering.

IagoAbal (approved):
Nice! Could you perhaps update the PR comment? Rather than having an "edit:" note I think it's better to just have the latest info. It would be great to have a better picture of its perf impact too, although it feels like a good change regardless.

src/tainting/Taint.ml
emjin: Wouldn't we need to also write the compare function for these tests?

PR #8124, opened 2023-06-23
tainting: Avoid passing around duplicate findings

IagoAbal:
test plan:
make test

PR checklist:

### DIFF
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -1550,7 +1550,7 @@ let check_function_signature env fun_exp args args_taints =
                    else [ `UpdateEnv (dst_lval, dst_taints) ])
       in
       Some
-        (fun_sig
+        (fun_sig |> T.Findings.elements
         |> List.concat_map process_sig
         |> List.fold_left
              (fun (taints_acc, lval_env) fsig ->
--- a/src/tainting/Dataflow_tainting.mli
+++ b/src/tainting/Dataflow_tainting.mli
@@ -100,7 +100,7 @@ val str_of_name : IL.name -> var
 val hook_function_taint_signature :
   (config ->
   AST_generic.expr ->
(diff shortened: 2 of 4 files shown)


aryx (approved):
Can't we just derive those compare with ppx_deriving?

IagoAbal:
> Can't we just derive those compare with ppx_deriving?

I will add a comment. I want to get rid of `Stdlib.compare` in taint code and I see `ppx_deriving` being essentially the same, "automagic" comparisons. These are very convenient but sometimes the comparison derived is not what you want. Initially I just thought that if you carefully considered whether that automagic comparison is OK for each data type then you're fine. But then one data type (in this case `Taint.arg`) evolved, and we added an `offset` to it, and we forgot to revisit the use of automagic comparisons... and this led to duplicates which led to an explosion in the size of taint sets and to big perf problems. I think the only way to get warned about this is to write them manually and the typechecker will force you to revisit the comparison functions if the types change. That's why I'm also using 
```ocaml
let compare_sink { pm = pm1; rule_sink = sink1 } { pm = pm2; rule_sink = sink2 }
     =
```
instead of
```ocaml
let compare_sink sink1 sink2
     =
```
and then accessing `.pm` and `.rule_sink`. If we pattern-match, and later we add new fields to the record, the typechecker will let us know that we need to revisit `compare_sink`.

IagoAbal:
I'll consider `ppx_deriving` during the "upcoming taint clean up" which I am planning for this summer. I guess that can work so far as you don't blindly add `@deriving compare` everywhere, some functions you may need to write manually.

emjin:
Related: I made a PR which starts to use deriving compare https://github.com/returntocorp/semgrep-proprietary/pull/803. I also handwrote some, and am not against handwriting in some places, but I consider it too annoying to have to write your own compare functions for every type.

aryx:
We need to setup some policy on when it makes sense to derive automatically, and when it's better to handwrite the compare function.

PR #8125, opened 2023-06-23
fix: Go typing metavariables with `new` and `&` (reference)

### LINKED ISSUE #6733: Go: Typed metavariables don't work with := &MY_STRUCT{}
**Describe the bug**
When using a short variable declaration, semgrep doesn't detect the type of a metavariable if the right hand side of the assignment is a reference to a struct (or new)

Example:
```go
client := &http.Client{}
// OR
client2 := new(http.Client)
```

**To Reproduce**
[semgrep playground link](https://semgrep.dev/s/obwan02:golang-taint-sources)

**Expected behavior**
I would expect similar behaviour to javascript/typescript's var keyword , C++'s auto keyword, Java's var etc. where assignment propagates the type of the RHS to the LHS.

**Screenshots**
N/A

**Environment**
Running in semgrep.dev

**Use case**
We need this to make accurate rules for static analysis

yosefAlsuhaibani:
closes #6733 

Fixed a translation & typing bug; it used to be the case that `make` & `new` in go_ast were also just `call make` & `call new`; now:
1) `call make(tau)` ==gen-trans==> `new(tau)`
2) `call new(tau)` ==gen-trans==> `&new(tau)` //new returns a ref

When resolving types while naming, semgrep now resolves the type of `Ref exp` in go.

Test plan: `make test` + a test set in `patterns`!

PR checklist:


languages/go/generic/go_to_generic.ml
```diff
@@ -243,6 +243,15 @@ let top_func () =
...
         G.ArrayAccess (v1, v2)
+    (* TODO: describe why this translation.
+     *)
+    | Call (Id ("new", t), None, (l, [ ArgType ty ], r)) ->
+        G.Ref
+          ( fake t "new",
+            G.New (fake t "new", type_ ty, G.empty_id_info (), (l, [], r))
+            |> G.e )
+    | Call (Id ("make", t), None, (l, [ ArgType ty ], r)) ->
+        G.New (fake t "make", type_ ty, G.empty_id_info (), (l, [], r))
```
nmote, line 257: I'd like to see a test case exercising this

languages/go/generic/go_to_generic.ml
```diff
@@ -243,6 +243,15 @@ let top_func () =
     | Index (v1, v2) ->
         let v1 = expr v1 and v2 = bracket index v2 in
         G.ArrayAccess (v1, v2)
+    (* TODO: describe why this translation.
+     *)
```
nmote: This seems like an easy enough TODO to do now

src/naming/Naming_AST.ml
```diff
@@ -350,6 +350,10 @@ let get_resolved_type lang (vinit, vtype) =
       (* alt: lookup id in env to get its type, which would be cleaner *)
       | Some { e = N (Id (_, { id_type; _ })); _ } -> !id_type
       | Some { e = New (_, tp, _, (_, _, _)); _ } -> Some tp
+      | Some { e = Ref (tok, exp); _ } when lang =*= Go ->
```
nmote: Why just limit to Go? At the very least could you add a comment explaining that choice?

nmote (approved):
Looks good overall but I do have a few comments inline that I'd like you to address. Thanks!

src/naming/Naming_AST.ml
yosefAlsuhaibani: Ah! It was during testing and wanted to limit this to only go: I've tested it with other languages and it works.

PR #8139, opened 2023-06-26
feat: migrate workloads off self-hosted x86 runners

spencerdrak:
We recently set up checks to test migrating back to github-hosted x86 runners for our nightly test and semgrep build workflows. These tests have been successful in the background:
- [Nightly brew install](https://github.com/returntocorp/semgrep/actions/workflows/nightly-macos-12.yml)
- [Built-test-osx-x86](https://github.com/returntocorp/semgrep/actions/workflows/build-test-osx-x86-macos-12.yaml)

We've also bumped into some issues with slowdowns, disk usage, and other issues on our self-hosted runners. Best practice is to get off of these runners and back onto ephemeral nodes, where we can. 

We may be able to cleanup some of the installation scripts in `scripts/` dir that these workflows reference - however, I'm leaving this in the same configuration as our background testing, to ensure we're successful and I can open a follow-on PR for cleanup as needed. This PR is to unblock us.

This PR now implements a cache for the OSX x86 builds for the `~/.opam` directory. The key will be (`runner-os` + `runner-arch` + `opam-switch-name`) which should give us enough specificity. We'll have more work coming down the line to restore some symmetry between envs. 

PR checklist:


cgdolan (approved):

aryx:
What is the difference in speed? I think we migrated to self-hosted runner because the macos GHA runner were super slow by default? Is the situation better now?

aryx:
Also why we have no issue with our m1 macos runner but lots of issues with the x86 macos runners?

aryx:
The m1 are also self-hosted ...

aryx (changes requested):
Far slower.
See https://github.com/returntocorp/semgrep/actions/runs/5380013827/jobs/9762077002?pr=8140 which finishes in 4min

aryx:
What can do to make the osx-x86 runner as reliable as the osx-m1 runner? I don't see why we have problems
with only one.

.github/workflows/nightly.yml
```diff
@@ -19,21 +19,11 @@ on:
 jobs:
   brew-build:
     name: Build Semgrep via Brew from `returntocorp/semgrep:develop`
-    runs-on: ["self-hosted", "macOS", "X64"]
+    runs-on: macos-12
```
aryx, line 22: those one I'm fine running with macos-12 instead of self-hosted. They can be slow. it's ok, it's run during the night.

spencerdrak:
> What is the difference in speed? I think we migrated to self-hosted runner because the macos GHA runner were super slow by default? Is the situation better now?

@aryx We migrated off of `macos-10.15` because those runners were being removed, but we couldn't migrate to `macos-11` or `macos-12` because we were seeing some intermittent failures on those runners with the `ocamlc` binary. I can dig up more info on this if needed. We set up self-hosted runners in their place to keep the lights on.

> Also why we have no issue with our m1 macos runner but lots of issues with the x86 macos runners?

I believe it's because we're running a lot more workloads on the x86 runner (the brew stuff + our regular test and build), but I haven't dug into all the performance differences between the two.

> Far slower.

Can we somehow cache the `opam switch` creation? IIRC, that was the bit that actually takes a long time. I don't know when I'll be able to handle that work, but that might help with some of the speed issues. It seems that actually compiling semgrep runs pretty quick.

aryx:
| I believe it's because we're running a lot more workloads on the x86 runner
I see the same job in our workflow for the x86 and m1. There's the nightly job (that we could switch to macos-12, this job
can be slow, it does not matter), but it's nightly, so run once per day, so I don't see how this can impact our self-hosted
runners ...

How many self-hosted runners we have for osx-x86 and osx-m1? Where are they configured? How do I add more runner?

spencerdrak:
> but it's nightly, so run once per day, so I don't see how this can impact our self-hosted runners ...

It could be other factors as well - I haven't dug in but that was just my best guess. It's the only major difference between what runs on two sets of runners. Could also be that the M1 runners are nearing their limits, I'll need to go check those out as well. 

> How many self-hosted runners we have for osx-x86 and osx-m1? Where are they configured? How do I add more runner?

We have two of each - we'd need to spin up the new AWS VMs, configure them, and then add them to the node pools. I would also have to check where we are at with our usage, we don't get an unlimited number of mac-based VMs and the limit is actually rather low.

On a separate note, this PR also adds back the hermeticity of these jobs for the x86 runners, which I think is a property that can correct a lot of the issues we've been seeing with our self-hosted runners. It's also why we saw such a slowdown, because we're no longer keeping the opam switch and other pre-built items around between runs. Maybe a solution here is to just look for better ways to reuse those components?

aryx:
yes, maybe we should try to cache the ~/.opam. It will be annoying if we update to a new opam switch (e.g., OCaml 5.0),
so would be good to document things very well so if we need to update to a new OCaml updating the cache is easy.

spencerdrak:
Note to the reviewer, this is the first run with this cache key (I updated to make it more specific) - this'll be a longer run, but I can re-run after for a faster runtime. 

See here for a previously cached run, the step in question runs in about ~11 mins. https://github.com/returntocorp/semgrep/actions/runs/5382192185/jobs/9767265395

mjambon (approved):

.github/workflows/build-test-osx-x86.yaml
```diff
@@ -1,29 +1,53 @@
...
+        if: ${{ inputs.use-cache }}
+        with:
+          path: ~/.opam
+          key: "${{ runner.os }}-${{ runner.arch }}-${{ env.OPAM_SWITCH_NAME }}-opam-deps"
       - run: |
-          ./scripts/osx-setup-for-release.sh
+          ./scripts/osx-setup-for-release.sh "${{ env.OPAM_SWITCH_NAME }}"
           opam exec -- make core
           mkdir -p artifacts
           cp ./bin/semgrep-core artifacts
```
mjambon, line 56: Please give a name to this step... It took me a while to figure out why semgrep was being built by the script `./scripts/osx-setup-for-release.sh`. This script doesn't build semgrep but `./scripts/osx-setup-for-release.sh` is the title that GHA assigns automatically to the step.

mjambon: I just did that.

mjambon: I think I'm going to split the step into installing dependencies and compiling semgrep since both are big.

mjambon: I did the split. I hope it works.

mjambon: It took 7 min, this time. Could be 5 min if we skipped `opam update`.

.github/workflows/build-test-osx-x86.yaml
```diff
@@ -1,29 +1,56 @@
...
-        run: |
-          rm -rf ./*
-          rm -f .gitmodules
-          rm -rf .git
       - uses: actions/checkout@v3
         with:
           submodules: true
-      - run: |
-          ./scripts/osx-setup-for-release.sh
+      # Note that this action handles both cache read and cache write.
```
aryx, line 41: Ok I still don't understand how this caching works.
When do we save this ~/.opam directory? At the very end of the jobs? or just after the next step?
Because we may want to cache only the content of ~/.opam after scripts/osx-setup-for-release.sh
has finished (where it will run the opam switch create 4.14.0 the very first time), but not the content
of ~/.opam after 'opam exec -- make core', which does extra thing.
Is it possible to give a scope to the cache? To tell GHA to save the content of a directory only after a certain step?

spencerdrak: @aryx yes, this step runs at the end of the job. It is possible to break this into two steps - the cache read and cache write individually - I can do that if we'd prefer and it would make things more clear. 

I have some follow up work here, I can include this as well.

PR #8140, opened 2023-06-26
precommit: switch back to fast precommit for semgrep jsonnet

aryx:
Now that 1.29.0 ship with osemgrep, we can use it for
our precommit again

test plan:
modify Main.ml and add a call to Main.ml,
run git commit and see error

The full story is that we originally were using the python hook, which was fast. Then we added jsonnet
support in pysemgrep, and wanted to use it in precommit to dogfood, but we could not use anymore
the python hook because our pip package was not packaged with jsonnet (which had some extra dependency
we didn't want our users to suffer from (libstdc++, jsonnet python package), so it was packaged in our docker
image but not in pip, so we had to use the docker image for precommit if we wanted jsonnet.
Now a few months later and we have osemgrep with ojsonnet builtin (no need external dependencies),
so we can use again the python hook, which dispatch now to osemgrep when we use the --experimental flag
since 1.29.0.

PR checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -177,33 +177,15 @@ repos:
   # ----------------------------------------------------------
   # Dogfood! running semgrep in pre-commit!
   # ----------------------------------------------------------
-  - repo: local
+  - repo: https://github.com/returntocorp/semgrep
+    rev: v1.29.0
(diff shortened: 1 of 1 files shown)


nmote (approved):
Nice!

PR #8151, opened 2023-06-27
feat/fix(SSC): Switch to using our existing yaml parser for pnpm, add v6+ support

mmcqd:
We already have a yaml parser that tracks line information in pysemgrep, we should not have re-invented the wheel (my bad lol). This makes it trivial to add support for later versions of pnpm, so we do that as well. This still needs testing and some error guarding to avoid KeyErrors. We should also use this to replace the `yarn.lock` v2 parser, since unlike yarn v1, it's actually valid yaml.

To make using the existing yaml parser easier, I moved the check that some parsed yaml is a valid semgrep config to a separate function, and I expanded the typing/exception handling of `safe_path_parse`.

This also adds support for pnpm files with weird package keys like tarball names or git urls.

test plan: existing tests still pass, and I've added a bunch of parsing tests from OSVScanner

PR checklist:


cli/src/semdep/parsers/util.py
```diff
@@ -257,12 +261,32 @@ def to_json(self) -> Dict[str, Union[Optional[str], Optional[int]]]:
...
+) -> Optional[A]:
+    ...
+
+
+def safe_path_parse(
+    path: Any,
+    parser: Any,
+    parser_name: Any,
+    preprocess: Any = lambda ξ: ξ,  # ξ kinda looks like a string hehe
+) -> Any:
```
brendongo, line 289: What is happening here?

brendongo (approved):

cli/src/semdep/parsers/util.py
mmcqd: `@overload` lets you give two (non-overlapping) types to a function. I'm making it so `safe_path_parse` can be called either with a `Parser` object, or with just a `str -> A` function, so that I can pass in the `parse_yaml_preserve_spans` function. The actual definition just has `Any` because our pre-commit settings require type annotations on functions, using `@overload` means only the annotations with that decorator actually mean anything

PR #8160, opened 2023-06-27
fix(go): `make` construct is not properly matched to 

yosefAlsuhaibani:
## issue
The pattern `x := make(int,...)` has the following behavior:
```go
//matches the following line:
x := make(int,4,5)
//but not this line:
x := make(int)
```
this is due to a change introduced in #8125, where we try to translate `make` into a `new` in our generic AST; it makes a faulty assumption that the only argument is the type, ignoring slice args: so the following would happen:
```
make(int)     ==gen-trans==> New(int)
make(int,...) ==gen-trans==> Call("make",[int, ellipses])
```
This PR fixes that.

test plan: `make test` + written a test in `tests/patterns`

PR checklist:


nmote (approved):
Nice!

PR #8161, opened 2023-06-27
add support for asynchronous processing of scan results

jbergler:
To ensure that we `semgrep ci` performance continues to scale, we will be moving some of the processing from /upload_findings_and_ignores into a background task.

Since the CLI is dependent on some of those processed results, we need to ensure that the CLI will retry periodically until the scan has been processed.

PR checklist:

### DIFF
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -377,10 +377,14 @@ class ScanHandler:
         try:
             response.raise_for_status()
 
-            resp_errors = response.json()["errors"]
+            res = response.json()
+            resp_errors = res["errors"]
             for error in resp_errors:
                 message = error["message"]
                 click.echo(f"Server returned following warning: {message}", err=True)
+
+            if "task_id" in res:
+                complete["task_id"] = res["task_id"]
 
         except requests.RequestException:
             raise Exception(f"API server returned this error: {response.text}")
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -369,7 +369,7 @@ class ScanHandler:
             logger.debug(f"Sending complete blob: {json.dumps(complete, indent=4)}")
 
         response = state.app_session.post(
-            f"{state.env.semgrep_url}/api/agent/scans/{self.scan_id}/findings_and_ignores",
+            f"{state.env.semgrep_url}/api/agent/scans/{self.scan_id}/results",
             timeout=state.env.upload_findings_timeout,
             json=findings_and_ignores,
         )
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -389,7 +389,7 @@ class ScanHandler:
         except requests.RequestException:
             raise Exception(f"API server returned this error: {response.text}")
 
-        try_until = datetime.now() + timedelta(minutes=2)
+        try_until = datetime.now() + timedelta(minutes=10)
         while datetime.now() < try_until:
             logger.debug("Sending /complete")
 
--- a/cli/src/semgrep/app/scans.py
+++ b/cli/src/semgrep/app/scans.py
@@ -3,7 +3,10 @@ import json
 import os
 from collections import Counter
 from copy import deepcopy
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


cli/src/semgrep/app/scans.py
```diff
@@ -366,35 +369,56 @@ def report_findings(
...
+        while datetime.now() < try_until:
+            logger.debug("Sending /complete")
 
-        try:
-            response.raise_for_status()
-        except requests.RequestException:
-            raise Exception(
-                f"API server at {state.env.semgrep_url} returned this error: {response.text}"
+            # mark as complete
+            response = state.app_session.post(
```
vivekkhimani, line 397: One thing that I am concerned about here is that the session (requests.Session) object used to make the POST request here has it's own retry logic setup. It's defined here - https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/app/session.py#L112.

I am wondering how this might interfere with the retry logic that we define for the ```/complete``` endpoint below. 😅

vivekkhimani (commented):
The logic itself looks great but I was wondering if it's a good idea to make the retry logic cleaner using something like ```tenacity``` or ```retrying```?

I kinda wrote this super quick, so haven't tested it properly, but maybe something on the lines of:

```python
from tenacity import retry, stop_after_delay, wait_fixed
```

```python
    @retry(
        stop=stop_after_delay(120),
        wait=wait_fixed(5),
        retry_if_exception_type=LookupError,
        reraise=True,
    )
    def complete_scan(self, complete: dict[str, Any]) -> Any:
        state = get_state()
        response = state.app_session.post(
            f"{state.env.semgrep_url}/api/agent/scans/{self.scan_id}/complete",
            json=complete,
        )
        response.raise_for_status()
        return response.json()
```

And then the ```report_findings``` can have something like:

```python
        try:
            response = self.complete_scan(complete)
            if response.get("success", False):
                return (
                    response.get("app_block_override", False),
                    response.get("app_block_reason", ""),
                )
            else:
                click.echo("Waiting for semgrep.dev to process scan results...")
                raise LookupError(
                    f"API server at {state.env.semgrep_url} never completed processing the scan results."
                )
        except requests.RequestException as ex:
            raise Exception(
                f"API server at {state.env.semgrep_url} returned this error: {ex.response.text}"
            )
```

Again, I still have concerns about the app session retry logic interfering with this.

aryx:
We started to port semgrep ci to osemgrep (see https://github.com/returntocorp/semgrep/blob/develop/src/osemgrep/cli_ci/Ci_subcommand.ml ) so ideally we will need to port this to OCaml soon.

cli/src/semgrep/app/scans.py
```diff
@@ -366,35 +369,56 @@ def report_findings(
...
         except requests.RequestException:
             raise Exception(f"API server returned this error: {response.text}")
 
-        # mark as complete
-        response = state.app_session.post(
-            f"{state.env.semgrep_url}/api/agent/scans/{self.scan_id}/complete",
-            timeout=state.env.upload_findings_timeout,
-            json=complete,
-        )
+        try_until = datetime.now() + timedelta(minutes=2)
```
brendongo: is 2 minutes enough?

cli/src/semgrep/app/scans.py
jbergler: Yeah, I didn't bother because it's going to get ported very soon anyway.

jbergler:
> We started to port semgrep ci to osemgrep (see https://github.com/returntocorp/semgrep/blob/develop/src/osemgrep/cli_ci/Ci_subcommand.ml ) so ideally we will need to port this to OCaml soon.

@aryx Yeah, I wanted to get something basic in so we can do our testing with returntocorp and fine tune all the things before it gets ported (since the ocaml learning curve is high for me right now, although I should work through that at some point)

cli/src/semgrep/app/scans.py
jbergler: Good point - I bumped it to 10min which should be plenty, we can tune from there.

aryx:
make sense! Sounds good to me.

vivekkhimani (approved):
🚀

PR #8168, opened 2023-06-28
osemgrep ci: report findings

hannesm:
this is at the moment reporting null to the complete endpoint, but does report to findings (some partial mocked data -- see TODO items).

test plan: make core

PR checklist:

### DIFF
--- a/src/osemgrep/networking/Scan_helper.ml
+++ b/src/osemgrep/networking/Scan_helper.ml
@@ -115,3 +115,106 @@ let report_failure ~dry_run ~token ~scan_id exit_code =
     | Error (code, msg) ->
         Error
           ("API server returned " ^ string_of_int code ^ ", this error: " ^ msg)
+
+(* TODO the server reply would be specified in ATD *)
+let extract_errors data =
+  try
+    match JSON.json_of_string data with
+    | JSON.Object xs -> (
+        match List.assoc_opt "errors" xs with
+        | Some (JSON.Array errs) ->
+            List.iter
+              (fun err ->
+                match err with
+                | JSON.Object xs -> (
+                    match List.assoc_opt "message" xs with
+                    | Some (String s) ->
+                        Logs.warn (fun m ->
+                            m "Server returned following warning: %s" s)
+                    | _else ->
+                        Logs.err (fun m ->
+                            m "Couldn't find message in %s"
+                              (JSON.string_of_json err)))
+                | _else ->
+                    Logs.err (fun m ->
+                        m "Couldn't find message in %s"
+                          (JSON.string_of_json err)))
+              errs
+        | _else ->
+            Logs.err (fun m ->
+                m "Couldn't find errors in %s"
+                  (JSON.string_of_json (JSON.Object xs))))
+    | json ->
+        Logs.err (fun m -> m "Not a json object %s" (JSON.string_of_json json))
+  with
+  | e ->
+      Logs.err (fun m ->
+          m "Failed to decode server reply as json %s: %s"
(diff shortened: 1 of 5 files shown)


src/osemgrep/networking/Scan_helper.ml
```diff
@@ -115,3 +115,106 @@ let report_failure ~dry_run ~token ~scan_id exit_code =
     | Error (code, msg) ->
         Error
           ("API server returned " ^ string_of_int code ^ ", this error: " ^ msg)
+
+(* TODO the server reply would be specified in ATD *)
```
aryx: do you know to which endpoints this refers too? Might be nice to add also in the comment the endpoint.

src/osemgrep/networking/Scan_helper.ml
```diff
@@ -115,3 +115,106 @@ let report_failure ~dry_run ~token ~scan_id exit_code =
...
+                  (JSON.string_of_json (JSON.Object xs))))
+    | json ->
+        Logs.err (fun m -> m "Not a json object %s" (JSON.string_of_json json))
+  with
+  | e ->
+      Logs.err (fun m ->
+          m "Failed to decode server reply as json %s: %s"
+            (Printexc.to_string e) data)
+
+(* TODO the server reply should be specified in ATD *)
```
aryx: same here

aryx:
pre-commit failure (semgrep jsonnet) in     src/osemgrep/cli_ci/Ci_subcommand.ml

aryx:
Not sure why we don't get the actual error message ...
We're now dogfooding osemgrep in precommit, so this might be a bug in osemgrep :(
In any case you can run 'make check' locally to get the same error.

aryx (changes requested):
Looks good, just need to fix precommit

hannesm:
addressed the review comments, and pre-check

src/osemgrep/networking/Scan_helper.ml
hannesm: done in [d8aaaf4](https://github.com/returntocorp/semgrep/pull/8168/commits/d8aaaf41a50b5dc72a71fb776e75a8422f4b87d5)

src/osemgrep/networking/Scan_helper.ml
hannesm: done in [d8aaaf4](https://github.com/returntocorp/semgrep/pull/8168/commits/d8aaaf41a50b5dc72a71fb776e75a8422f4b87d5)

hannesm:
> Not sure why we don't get the actual error message ... We're now dogfooding osemgrep in precommit, so this might be a bug in osemgrep :( In any case you can run 'make check' locally to get the same error.

Strange. The command line is morally `osemgrep --config semgrep.jsonnet --error --skip-unknown-extensions .` -- which locally reports the findings as pysemgrep (on stdout)...

But what is even stranger is that there's no report of "N findings", only ever 0 findings are reported. :/

And additionally, the exit code is 2 for osemgrep, not 1 which would be expected.

EDIT: there's as well duplicate headings and tables... and on a Linux machine running `pre-commit run semgrep -a` results in the same behaviour: only the filename is printed, but not the finding... but I can't reproduce by invoking osemgrep on the shell, only if executed via pre-commit...

aryx:
pre-commit is a bit weird because it does run many semgrep (or osemgrep) commands with different batch of files, which is why we have lots of similar boilerplate output repeated (we should really have a --precommit flag to optimize output for precommit).

aryx:
I think there's a way to run precommit in verbose mode so you can see the different commands that are executed.

PR #8170, opened 2023-06-28
feat: [ci] set cache timeout

spencerdrak:
Default Cache retrieval timeout is 10m, but in the average case (at least from the small sample size we have), cache read should take only ~30s. Set the timeout to 2m in case of a read issue, we shave 8m off the total runtime. 

Test Plan: 
- CI on this PR. 

PR checklist:

### DIFF
--- a/.github/workflows/build-test-osx-x86.yaml
+++ b/.github/workflows/build-test-osx-x86.yaml
@@ -88,6 +88,8 @@ jobs:
         id: cache-opam
         uses: actions/cache@v3
         if: ${{ inputs.use-cache }}
+        env:
+          SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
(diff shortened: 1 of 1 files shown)


cgdolan (approved):

brendongo (approved):

.github/workflows/build-test-osx-x86.yaml
```diff
@@ -88,6 +88,8 @@ jobs:
         id: cache-opam
         uses: actions/cache@v3
         if: ${{ inputs.use-cache }}
+        env:
```
aryx, line 91: would be better to put what was in your PR comment in the code itself, so next time someone look at this they can understand the motivations for those values (yes one can use git blame, but better to avoid one indirection).

PR #8175, opened 2023-06-28
adding tests to show loops can now be parsed in csharp closes #6568

### LINKED ISSUE #6568: C#: loops seemingly cannot be matched through patterns
**Describe the bug**
It seems impossible to find `while`, `for` and `foreach` loops in C# through patterns. Strangely enough, the exact same code in Java can be found. Also substituting `while` with `if` in both the code and the pattern (so the code and pattern are the same otherwise) does result in a match.

**To Reproduce**
[Link to https://semgrep.dev `while` in C#](https://semgrep.dev/s/O5lb) <= cannot be found
[Link to https://semgrepdev `while` in Java](https://semgrep.dev/s/yv0Z) <= can be found
[Link to https://semgrep.dev `if` in C#](https://semgrep.dev/s/pvPd) <= can be found

**Expected behavior**
I would expect a match on the `while` loop with the current C# code and pattern. Also in general, I would expect `while` and other types of loops (`for`, `foreach`) to be detectable in C#, like they are in Java.

**Screenshots**
image

image

image

**Environment**
-

**Use case**
For educational purposes, we provide automatically generated feedback on the code of students. Additionally we wish to automatically detect if the student has written the code as intended. Fixing this bug means we can do both these things when it comes to loops.

SophiaSR:
Just adding test cases to verify that with the new C sharp parser, loops in patterns are actually parsed correctly
PR checklist:


aryx:
You can mention the issues it closes, this this closes #xxx or PA-xxx (if the issue was on linear).

aryx (approved):

PR #8177, opened 2023-06-28
fix(SSC): Handle comments, empty tables, and allow quoted keys in poetry lockfiles

mmcqd:
Now we don't fail to parse poetry.lock files with arbitrary comments, and we can handle things like this:
```
[foo]
"quoted key" = "value"
```
and this:
```
[foo]
[bar]
key = value
```
test plan: see added tests

PR checklist:

### DIFF
--- a/cli/tests/e2e/targets/dependency_aware/poetry_quoted_key/pyproject.toml
+++ b/cli/tests/e2e/targets/dependency_aware/poetry_quoted_key/pyproject.toml
@@ -1,0 +1,14 @@
+[tool.poetry]
+name = "test"
+version = "0.0.1"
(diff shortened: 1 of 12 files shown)


cli/tests/e2e/targets/dependency_aware/poetry_quoted_key/pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.poetry]
```
adamberman, line 1: do you want to test the arbitrary comment removal part of the manifest file too?

adamberman (approved):

PR #8178, opened 2023-06-28
feat(be): Added ability to process new sca config

SayyantR:
[SC-866](https://linear.app/semgrep/issue/SC-866/make-cli-block-supply-chain-rules-properly)

This PR is part of a bigger SSC [project](https://linear.app/semgrep/project/blocking-findings-in-pr-c74e0ac32c83) that will allow users to toggle PR blocking for SSC findings that are reachable and of either high or critical vulnerability. This PR specifically allows us the CLI to block SCA findings that the App has identified as blocking worthy and the CLI has deemed reachable/ always reachable.

Screenshot 2023-06-28 at 4 26 32 PM

PR checklist:


cli/src/semgrep/rule_match.py
```diff
@@ -356,6 +356,18 @@ def get_end_line_hash(self) -> str:
         """
         last_line = self.get_individual_line(self.end.line)
         return hashlib.sha256(last_line.encode()).hexdigest()
+    
+    def is_direct(self) -> bool:
```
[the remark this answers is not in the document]
mmcqd: I think this rule is only useful if you don't have a typechecker, and we do. Probably fine to ignore.

mmcqd: Two comments:
1. `is_direct` it isn't really clear what `is_direct` means from the name. Probably we should have some `ScaMatch` subclass, but for now I'd maybe call it `is_sca_match_in_direct_dependency`. Verbose, but clear.
2. `"sca_info"` will not be present in `self.extra` if this wasn't an sca finding, so this should be checked before you access it

cli/src/semgrep/rule_match.py
```diff
@@ -356,6 +356,18 @@ def get_end_line_hash(self) -> str:
...
+        return (
+            self.extra["sca_info"].dependency_match.found_dependency.transitivity
+            == "direct"
+        )
+    
+    def is_reachable_or_always_reachable(self):
```
mmcqd: Would also rename this to `is_reachable_or_always_reachable_sca_rule` and add the check that `"sca_info"` is present.

mmcqd (changes requested):
Just need some tests and this should be looking good

cli/src/semgrep/rule_match.py
SayyantR: Regarding the "sca_info" check, I believe it is performed in the `is_blocking()` method before I call the methods that use that field?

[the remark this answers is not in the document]
mmcqd: hm, actually, we use `@property` a lot, so for the sake of stylistic consistency, I'd add it

mmcqd: I'm thinking about other people using this. If you have a method on a `RuleMatch` called `is_direct`, it's not at all clear that it might raise an exception, it should just return false if the match wasn't an sca finding

mmcqd (approved):

mmcqd:
Some of the confusing looking test changes are due to the order of findings changing because of my edits to CI output to account for blocking rules with non-blocking findings

cli/src/semgrep/rule_match.py
```diff
@@ -412,7 +437,13 @@ def is_blocking(self) -> bool:
         """
         blocking = "block" in self.metadata.get("dev.semgrep.actions", ["block"])
         if "sca_info" in self.extra:
-            return blocking and self.extra["sca_info"].reachable
+            if (
+                self.is_always_reachable_sca_match
```
brendongo, line 441: why are reachable transitives non-blocking? what about unknown?

mmcqd: transitive reachables are blocking. transitive "always-reachable"s are non-blocking, because we think it'll be too noisy.

PR #8179, opened 2023-06-28
fix(go): fixed make translation issue

### LINKED ISSUE #8171: Mishandling of make in Go for slices vs. channels
**Describe the bug**
Semgrep does not properly capture the differences between:

```go
ch := make(chan string)
ch := make(chan string, 1)  // Missed by $CHANNEL := make(..., $FOO) and $CHANNEL := make($FOO, $VAR)
x := make(int)
```

**To Reproduce**
Run [this rule](https://semgrep.dev/playground/r/trailofbits.go.hanging-goroutine.hanging-goroutine) and tests with 1.30. Notice that it results in 0 findings. 

Next, remove the below. from the same rule

```
    - pattern-inside: |
        $CHANNEL := make(...)
        ...
```

Notice the are 4 results from the tests, including 2 FPs not previously reported in 1.28. Next, change the `pattern-not-inside` to:

```
$CHANNEL := make(..., $T)
...
``` 

to

```
        $CHANNEL := make($TYPE, $INT)
        ...
```
Same incorrect results. Finally, change the above to 

```
    - pattern-not-inside: |
        $CHANNEL := make(chan $TYPE, $INT)
        ...
```

Now the rule works as expected and the tests pass, indicating that Semgrep is getting confused by the fact that `make` includes they keyword chan AND a space + a type as part of the first argument for Go channels. However, note that if we re-add the below, the rule goes back to returning 0 results.

```
    - pattern-inside: |
        $CHANNEL := make(...)
        ...
```

**Expected behavior**
All existing tests should continue to pass. Run [this rule](https://semgrep.dev/playground/r/trailofbits.go.hanging-goroutine.hanging-goroutine) and tests with 1.28 to see the expected results.

**Environment**
Local installation of Semgrep 1.30, web playground for 1.28

**Use case**

Find misuse of channel bugs in Go effectively and continue running rules that were previously working well but now are missing bugs.

yosefAlsuhaibani:
closes #8171 

This handles the regression introduced by translating `make` differently, and should handle patterns like `make($A,$B)`; `make(...,$X)`; `make(...)`.

test plan: `make test` + new tests

PR checklist:


languages/go/generic/go_to_generic.ml
```diff
@@ -226,6 +226,26 @@ let top_func () =
...
+    (* the first arg is indeed a type *)
+    | ArgType ty -> return (type_ ty, args)
+    (* metavar *)
+    | Arg (Id v1) -> return (G.TyN (G.Id (v1, G.empty_id_info ())) |> G.t, args)
+    (* ... *)
+    | Arg (Ellipsis tok) ->
+        return (G.TyEllipsis tok |> G.t, G.Arg (G.Ellipsis tok |> G.e) :: args)
+    (* exp *)
+    | Arg exp -> return (G.TyExpr (expr exp) |> G.t, args)
+    | _ -> raise Impossible
```
aryx: better to avoid those _ when we can and instead explicitely list the remaining cases, that way
if one day we add a new constructor in Arg, the compiler can tell us maybe we need to update this code.

aryx: Actually we used to have a semgrep check for that, not sure why it's not triggering here.
Or maybe someone removed this check.

aryx (approved):

mjambon:
This didn't automatically close issue #8171 <confused parrot>. Closing by hand.

PR #8190, opened 2023-06-29
Report tree-sitter errors flagged as MISSING nodes.

mjambon:
Update: pausing for now since it's taking longer than anticipated. I'll get back to it when I'm done with my urgent tasks.

This is a change in the ocaml-tree-sitter-core runtime library. It will surface up parsing errors that were previously unreported across all tree-sitter-based parsers used by semgrep.

This uses https://github.com/returntocorp/semgrep-rules/pull/2993. See discussion there.

test plan: usual CI tests

I'm adding a changelog entry since this affects users.

PR checklist:


aryx:
This might have an impact on semgrep users, with new parsing errors coming up for people, so we want to be careful before merging it and warn people.

mjambon:
I fixed the problems in semgrep-rules. In the semgrep repo, I moved away some broken tests into a `todo/` folder when I didn't know what to do.

What we need now is:

1. make sure broken patterns are reported as warnings and not as fatal errors (which is the case right now, it seems).
2. report or track the tests that were moved to `todo/` folders.

aryx:
Do I still need to review this?

mjambon:
@aryx I removed you because you were on vacation but then I paused the effort for 2 weeks. I'll ping the team when I need a review. I still have to check the situation of error reporting for broken patterns. I fixed a bunch broken patterns in semgrep-rules, so I'm a bit worried about the consequences on people's custom rules where they forgot a semicolon in a pattern (which is now an error).

mjambon:
I also have to look at CI failures.

nmote:
Feel free to re-add me when this is ready for review

mjambon:
There's a serious performance regression that I wasn't expecting at all. I can reproduce it in a striking manner: the same test (one pattern on one big js file) went from 3 s to 90 s.

mjambon:
For a language like JavaScript where we see the perf regression, the issue that we try two parsers one after the other until one succeeds. Since we introduced a new error condition, there will be more retrying on average. In this case, it's catastrophic because the tree-sitter parser happens to be about 30 times faster than our fallback parser and the target file is big. I think the solution is to not treat inserted (aka "missing") tokens as errors causing us to retry parsing.

cli/tests/qa/test_public_repos.py
```diff
@@ -67,6 +67,7 @@ def assert_sentinel_results(repo_path, sentinel_path, language):
     ]
 
     # nosemgrep: python.lang.security.audit.dangerous-subprocess-use.dangerous-subprocess-use
+    print(f"semgrep command: {cmd}")
```
aryx, line 70: was this added on purpose?

cli/tests/qa/test_public_repos.py
```diff
@@ -157,7 +158,7 @@ def test_semgrep_on_repo(monkeypatch, tmp_path, repo):
         sentinel_path = repo_path / sentinel_info["filename"]
         with sentinel_path.open("w") as sentinel_file:
             sentinel_file.write(sentinel_info["file_contents"])
-
+        print(f"assert_sentinel_results({repo_path}, {sentinel_path}, {language})")
```
aryx, line 161: same? Looks like debugging statements

js/languages/cpp/Parser.ml
```diff
@@ -7,7 +7,10 @@ let parse_pattern print_errors _ str =
   let any =
     str
     |> Pfff_or_tree_sitter.run_pattern ~print_errors
-         [ TreeSitterPat Parse_cpp_tree_sitter.parse_pattern ]
+         [
+           PfffPat (Parse_cpp.any_of_string Flag_parsing_cpp.Cplusplus);
```
aryx, line 11: Is this related to missing nodes?

src/parsing/Pfff_or_tree_sitter.ml
```diff
@@ -36,10 +36,12 @@ type 'ast parser =
   | Pfff of (Common.filename -> 'ast * Parsing_stat.t)
   | TreeSitter of (Common.filename -> 'ast Tree_sitter_run.Parsing_result.t)
 
+(* This type is parametrized by the AST type because ??? *)
```
aryx, line 39: because we don't always generate directly generic ASTs; we sometimes generate intermediate ASTs hence the 
need for polymorphic type (so you can have Ast_cpp.program internal_result, or Ast_php.program internal_result.

src/parsing/Pfff_or_tree_sitter.ml
```diff
@@ -96,14 +89,28 @@ let extract_pattern_from_tree_sitter_result
...
 (*****************************************************************************)
 
+(* Serious error = any parsing error that causes us to resort to an alternate
+   parser. *)
+let is_serious_error (err : Tree_sitter_run.Tree_sitter_error.t) =
+  match err.kind with
+  | Internal
+  | Error_node ->
+      true
+  | Missing_node -> false
```
aryx, line 107: maybe you could add a comment that if we consider missing_node a serious_error, this will trigger
the fallback to pfff parsers for many more cases resulting in big perf and parsing regressions, hence
this setting.

aryx (commented):
Nice work.
Why the many new parsing/todo/??? You were not familiar with the language so
just move it to todo/ because you didn't know the parsing fix?

cli/tests/qa/test_public_repos.py
mjambon: I forgot to remove them, but also I think all tests should print important debugging information by default.

js/languages/cpp/Parser.ml
mjambon: Weird. I'll look into this.

mjambon: This is the JavaScript implementation. It now does the same as the OCaml implementation. I left a comment about this coupling in the other file (`src/parsing_languages/Parse_pattern2.ml`). I'll do the same here.

mjambon: Oh yeah, I didn't leave a "coupling:" comment here because it would make sense to leave one for each language (`js/languages/*/Parser.ml`).

mjambon: ok, it's not that bad because only a few parsers use a non-tree-sitter parser. I added a bunch of comments.

cli/tests/qa/test_public_repos.py
mjambon: I removed this one and kept the other one + a comment.

cli/tests/qa/test_public_repos.py
mjambon: see other comment thread

mjambon:
> Why the many new parsing/todo/??? You were not familiar with the language so
> just move it to todo/ because you didn't know the parsing fix?

These should be cases where there was a parsing bug rather than a syntax error in the code. These are mostly patterns. Let me reassess and at least file bug reports.

PR #8191, opened 2023-06-29
move create_rule earlier

mmcqd:
`create_rule` calls `Path.open` on the rule schema path, but this test mocks the open method to return some python, which causes the yaml parser called on the result of opening the rule schema to run on this python and fail. No idea why this only just started happening, but I moved the call to before the mock.

PR checklist:


emjin (approved):
Love it

PR #8193, opened 2023-06-30
Adopt ephemeral Cirrus Runners for M1 workflows

cgdolan:
M1 workflows currently execute on self-hosted Mac instances that run the GitHub Actions Runner software. A known limitation of this approach is the lack of isolation between workflow runs. This means state introduced in one workflow run (for example, installing packages) can carry over into the next workflow run leading to confusing and non-deterministic behavior.

Introduce isolation between M1 workflow runs by adopting ephemeral self-hosted Mac instances from Cirrus Runners.  These instances are special-purpose VMs designed to drop into existing GitHub Actions that get created on-demand and torn down after a single use. The isolation allows us to share a single pool of M1 instances across all our organization's repos for better latency and throughput as compared to the existing approach where we dedicate 1-2 instances per repo.

PR checklist:


cgdolan:
Merged in https://github.com/returntocorp/semgrep/pull/8278

PR #8194, opened 2023-06-30
feat!: OCaml 5.0

brandonspark:
## What:
This PR updates our version of OCaml to 5.0.

## Why:
OCaml 5.0 is the shiny new major release of OCaml, and comes with a slew of new, fun features. 
- A brand new runtime, which we anticipate should help fix a segfault issue currently being experienced by Interactive Mode
- Multicore support, which will allow true parallelism within Semgrep, which should provide major performance capabilities
- Effect handlers, better compiler error messages, and other fun things which will help us develop Semgrep

## How:
This was a rather involved change, involving the creation of new 5.0-specific `returntocorp/ocaml` Docker images. You can see those changes here: https://github.com/returntocorp/ocaml-layer/pull/31.

In addition, we had to touch places in our build systems, CI environments, and of course within the codebase itself. Those changes are littered here and there within this PR.

## Test plan:
Automated tests succeed.

`make dev-setup` succeeds locally, and compilation and pre-commit also succeed.


cli/tests/e2e/test_check.py
```diff
@@ -599,7 +599,13 @@ def test_max_memory(run_semgrep_in_tmp: RunSemgrep, snapshot):
     snapshot.assert_match(stderr, "error.txt")
 
 
+# It seems that OCaml 5.0 does not use the system stack https://github.com/ocaml-multicore/docs/blob/main/ocaml_5_design.md#stack-layout
+# note that it says it manages its own runtime stack, which are managed in fibers
+# fiber stacks are allocated w/ mallloc + free aka its in the heap.
+# So I thionk this test is no longer valid as of OCaml 5.0. The exception is C calls
+# but ocaml handles the runtime stack switching to system stack for them so we should be good
 @pytest.mark.slow
+@pytest.mark.skip(reason="OCaml 5.0 does not use system stack")
```
ajbt200128, line 604: @brendongo @aryx This makes sense right?

js/languages/ruby/Parser.ml
```diff
@@ -1,5 +1,9 @@
-let parse_pattern _ _ str =
-  let any = Parse_ruby.any_of_string str in
+let parse_pattern print_errors _ str =
+  let any =
+    str
+    |> Pfff_or_tree_sitter.run_pattern ~print_errors
+         [ TreeSitterPat Parse_ruby_tree_sitter.parse_pattern ]
+  in
```
ajbt200128: It seems that we only use dyp for ruby patterns. I'm not why sure it breaks with  5.0.0 but it seems the tree sitter parse pattern works fine. Is there a reason we're not using it in Parse_pattern/not to use it here?

ajbt200128: If we're ok with it, this commit fixes the failing JS test.

cli/tests/e2e/test_check.py
ajbt200128: also possibly @mjambon since he's usually up to date on his ocaml lore

aryx: @brandonspark (not brendongo)

aryx: Adding also @ihji who investigated a bit also switching to ocaml 5.0

Makefile
```diff
@@ -301,6 +301,7 @@ VIRTENV='virtualenv==20.21.0'
 # https://stackoverflow.com/questions/63515454/why-does-pip3-install-pipenv-give-error-error-cannot-uninstall-distlib
 install-deps-ALPINE-for-pysemgrep:
 	apk add --no-cache $(ALPINE_APK_DEPS_PYSEMGREP)
+	pip uninstall -y virtualenv
```
tpetr, line 315: Why uninstall?

languages/scala/recursive_descent/Parser_scala.mly
```diff
@@ -60,7 +60,7 @@
 (*************************************************************************)
 
 (* unused rec flag *)
-[@@@warning "-39"]
+[@@@warning "-39-8"]
```
tpetr, line 63: Just curious, what is 8?

dune-project
```diff
@@ -389,10 +389,10 @@ For more information see https://semgrep.dev
 ;TODO: restore  "bisect_ppx" {>= "2.5.0"} once can use ppxlib 0.22.0
   (depends
     ; core deps
-    (ocaml (>= 4.13.0))
+    (ocaml (>= 5.0.0))
     (dune (>= 2.7.0 ))
     ; parser generators
-    (menhir (= 20211128)) ; Newer versions cause massive build slowdowns
+    (menhir (= 20220210)) ; Newer versions cause massive build slowdowns
```
aryx: is the comment still true? Is the new version you pinned is slow?

aryx:
ALL CHECKS ARE PASSING!!!!! IS IT TRUE!!! WE CAN SWITCH TO 5.0??

.github/workflows/build-test-osx-arm64.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
         "ghcr.io/cirruslabs/macos-monterey-xcode:latest",
       ]
     env:
-      OPAM_SWITCH_NAME: "4.14.0"
+      OPAM_SWITCH_NAME: "5.0.0"
```
aryx, line 35: Should we switch to 5.1.0 directly (which was released a few weeks ago)?

aryx:
@brandonspark can we mark this read for review? Can we merge this?

dune-project
brandonspark: I would need to investigate this, I missed it.

brandonspark:
@aryx I marked it as draft still because I hadn't done the pre-review lookover, yet -- I literally got the green checkmark as I was boarding my flight.

Makefile
brandonspark: I will leave a comment, I didn't get to elaborate on this because I abruptly had to stop working due to my flight.

TL;DR there is a host of nasty, messy `virtualenv` errors we get from the following line (which simply installs `virtualenv`). The linked Stackoverflow post seems to indicate this is due to a conflict between a previously-installed version of `virtualenv`.

I investigated the Docker images and both the working image and the newer image do have `virtualenv` installed (which originates as a dependency when installing `pre-commit`), but the disparity is in their versioning. I tried a great deal of options (including force installing, installing `virtualenv-clone`, not installing the deps, etc), but the only option that seemed to work was uninstalling `virtualenv` entirely prior to reinstalling it.

I viewed this as the lowest friction solution. I plan to document this better and proceed like this, but if you have a better suggestion, I would love to do something different.

languages/scala/recursive_descent/Parser_scala.mly
brandonspark: "Nonexhaustive match" in a `match ... with`. This file starts getting compilation warnings on 5.0. Since it's an unused file, I decided to just disable the warning and proceed.

.github/workflows/build-test-osx-arm64.yaml
brandonspark: I might prefer that we hold that one off for later -- this would require me to basically revisit every change I've made, and another PR to `ocaml-layer` to boot (which might induce more dependency hell).

Let's get this milestone in quickly, while it's easy to do, and worry about 5.1 soon after.

aryx: ok

dune-project
brandonspark: It looks like compilation actually is faster...
(on 5.0)
image

(on `develop`)
image

Makefile
tpetr: Gotcha, I agree it's not worth spending the calories right now. A brief clarifying comment sounds good to me 👍

libs/process_limits/Unit_memory_limit.ml
```diff
@@ -23,22 +23,27 @@ let with_debug_alarm f =
 (*
    Grow the stack until some limit expressed in bytes.
```
ajbt200128, line 24: I think we can just delete this whole test now? Since the stack is no longer real

ajbt200128 (approved):
Do we want to also remove all of the stack size limit related flags/code?

ihji:
Great work 👍 
Do you have any performance benchmarks that compare the scanning speed between OCaml 4.14.0 and OCaml 5.0.0? I would like to see if the transition to 5.0.0 has made the scanning speed better, worse, or on par.

aryx (approved):
Let's do it!

PR #8196, opened 2023-06-30
refactor: kill `ppx_sexp_conv`

brandonspark:
## What:
This PR _almost_ completely kills `ppx_sexp_conv`

## Why:
`ppx_sexp_conv` is currently being used at version `0.14.3`, because later versions of `ppx_sexp_conv` run into this issue, which prevents us from compiling: https://github.com/janestreet/ppx_sexp_conv/issues/40

Unfortunately, `ppx_sexp_conv` version `0.14.3` is in conflict with OCaml 5.0, which we are considering migrating to. 

We used to use `ppx_sexp_conv` to dump tree-sitter CSTs, but now after the invention of `Raw_tree`, we no longer need it. So we can kill it by updating all of our `semgrep-*` repositories, now with updated dune files that no longer require `ppx_sexp_conv`, and boilerplate code which does not use it.

## How:
We already updated `ocaml-tree-sitter-core` so future updates to `semgrep-*` repos distribute `dune` files which do not have this dependency. This PR just goes and painstakingly updates each of those repositories.

Quite a few test cases will fail, if you also update the `ocaml-tree-sitter-core` runtime, because of the MISSING node changes.

## Test plan:
`make test`

PR checklist:


languages/typescript/tree-sitter/Boilerplate_tree_sitter_typescript.ml
```diff
@@ -0,0 +1,3187 @@
...
+
+   3. Edit the new file Boilerplate_tree_sitter_typescript.ml:
+      a. Copy this header.
+      b. Add the "type assertion" case. Look for 'Type_asse' in
+         the new generated file semgrep-typescript/lib/Boilerplate.ml
+*)
+(**
+   Boilerplate to be used as a template when mapping the tsx + typescript CST
+   to another type of tree.
+*)
```
mjambon, line 40: I would add something like this:

> We need this hand-modified file because it provides the `dump_tree` function that works on the merged TSX/TypeScript CST type.

akuhlens (commented):
I think there is a bit of a race condition here with #8234 which updates the rust submodule.

aryx:
Looks like a good change. Can we rebase and merge this @brandonspark ?

brandonspark:
@aryx finally got around to rebasing it. I also re-updated all of the submodules, to verify that I didn't accidentally regress any. Seems they all should be the most recent iterations of each of those submodules.

aryx:
adding @nmote who talked about using ppx_sexp_conv recently for semgrep-pro ...

dune-project
```diff
@@ -417,7 +417,9 @@ For more information see https://semsgrep.dev
     ; affecting CST type definitions generated by old versions of
     ; ocaml-tree-sitter. New generated code no longer uses ppx_sexp_conv.
     ; Handwritten code can easily work around this bug.
-    (ppx_sexp_conv ( = v0.14.3))
+    ; We used to use this library, but it conflicts with OCaml 5.0. So
```
aryx, line 420: great comment

tests/parsing/swift/statements.swift
```diff
@@ -48,22 +48,22 @@ var y { }
 var x {
   get { return 2 }
   _modify { }
-  set { return () }
+  set { return 2 }
```
aryx: why did you need to change this test file?

aryx (approved):

nmote:
Wait yeah I am seriously considering using this. Was going to use `ppx_yojson_conv` but we already had this dependency. Is the concern just about the version or is even the most recent version in conflict with OCaml 5.0?

nmote:
I guess the issue was never addressed upstream so we have no choice unless we want to go fix it upstream. I will find another solution.

tests/parsing/swift/statements.swift
brandonspark: I just fixed this while I was here, basically, because this was breaking some of Martin's tests and I think he xfailed it. This was hand-in-hand with the missing nodes, because this is an invalid Swift program, and this causes us to suddenly fail.

aryx (approved):

PR #8198, opened 2023-06-30
fix(js): Include location of `async` keyword in arrow func range

### LINKED ISSUE #7353: typescript/javascript `async` arrow function not fixed/replaced properly
**Describe the bug**

An async argument doesn't get replaced properly. Below I have a demo of a pattern and a fix that are equivalent, so I would expect no changes in the "fix". However, they output is invalid code. 

**To Reproduce**
https://semgrep.dev/s/PZKq

```
rules:
  - id: demo
    pattern: fnTakingAsyncArg($X)
    fix: fnTakingAsyncArg($X)
    message: |
      Verifying that an async param is not matched.
    languages:
      - ts
    severity: WARNING
```

**Expected behavior**

I'd expect that nothing would actually change, and `fnTakingAsyncArg(async () => {})` would be replaced with `fnTakingAsyncArg(async () => {})`. However, the result is `fnTakingAsyncArg(=>{})`

**Screenshots**
CleanShot 2023-03-17 at 09 33 06@2x

**Environment**

Binary

nmote:
We don't include the `async` attribute (or any others, if there are others possible) as part of the generic AST. We might want to change that, but we might not want to complicate the generic AST for this one case.

However, since that token does not get included, we get the range wrong. This is especially a problem for metavariable bindings when they are used in autofix. See the test case.

We now set `e_range` when constructing the AST. I've used this technique in a few places in the past to address location issues. To simplify the logic here, I've also refactored a bit to add a utility which will set the `e_range` of an expression based on a list of `AST_generic.any` nodes. Note that this makes it so that calls to `extract_ranges` now entail one additional allocation for the singleton list. While this is a hot code path I very much doubt that one more allocation will make a difference considering all the other work that goes into computing the ranges.

Fixes #7353

Test plan: Automated tests

PR checklist:

### DIFF
--- a/changelog.d/gh-7353.fixed
+++ b/changelog.d/gh-7353.fixed
@@ -1,0 +1,1 @@
+Fixed an issue leading to incorrect autofix results involving JS/TS async arrow functions (e.g. `async () => {}`, etc.).
--- a/languages/javascript/generic/js_to_generic.ml
+++ b/languages/javascript/generic/js_to_generic.ml
@@ -298,9 +298,15 @@ and expr (x : expr) =
       | Right e -> G.DotAccess (v1, t, G.FDynamic e))
       |> G.e
   | Fun (v1, _v2TODO) ->
-      let def, _more_attrs = fun_ v1 in
-      (* todo? assert more_attrs = []? *)
-      G.Lambda def |> G.e
+      let def, more_attrs = fun_ v1 in
+      (* TODO: Include attrs in generic AST? Where? *)
+      let e = G.Lambda def |> G.e in
+      (* Since the attrs aren't included in the AST, at least update the range
+       * to include them. See
+       * https://github.com/returntocorp/semgrep/issues/7353 *)
+      let attrs_any = Common.map (fun attr -> G.At attr) more_attrs in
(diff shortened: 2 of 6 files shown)


emjin (approved):
If we have other places where we fix the range like this, I'm ok with it. I think I'd prefer to add the tokens to the generic AST, so that we have a consistent approach to how we indicate ranges, rather than sometimes using `e_range` and sometimes using the tokens.  I'm worried that it'll make it hard for people to fix range related problems because it'll be hard for them to know how it's being computed.

However, all that said, if I were to choose between tokens and storing ranges for each node, I would choose ranges.

nmote:
> If we have other places where we fix the range like this, I'm ok with it.

I first introduced this idea here, and I provided a bit more motivation in that PR description: https://github.com/returntocorp/semgrep/pull/7437

In general, I think it's a good idea to use `e_range` as the canonical source of the range for an expression. Relying on tokens for ranges has caused us quite a few problems. In fact, I'd like to see us remove tokens altogether at some point. We shouldn't need them to perform any analysis and they are just taking up space. Plus it's a pain to create fake tokens when building synthetic ASTs.

emjin:
> In fact, I'd like to see us remove tokens altogether at some point.

That would be great, totally supportive of this as a step towards that world

PR #8201, opened 2023-07-01
feat(typed metavar): Typed metavariable support for Rust

### LINKED ISSUE #8200: feat(typed metavar): Typed metavariable support for Rust
Typed metavariable support for Rust.

Creating `TypedMetavar` using Rust's casting syntax `as`

ihji:
fix #8200 

PR checklist:

### DIFF
--- a/languages/rust/generic/Parse_rust_tree_sitter.ml
+++ b/languages/rust/generic/Parse_rust_tree_sitter.ml
@@ -49,6 +49,11 @@ let deoptionalize l =
   in
   deopt [] l
 
+let in_pattern env =
+  match env.H.extra with
+  | Target -> false
+  | Pattern -> true
+
 (*****************************************************************************)
 (* Intermediate AST-like types *)
 (*****************************************************************************)
@@ -1410,12 +1415,18 @@ and map_expression (env : env) (x : CST.expression) =
         }
       in
       G.Lambda func_def
-  | `Paren_exp (v1, v2, v3) ->
-      let _lparen = token env v1 (* "(" *) in
-      let expr = map_expression env v2 in
-      let _rparen = token env v3 (* ")" *) in
-      let x = expr in
-      x.G.e
+  | `Paren_exp (_v1, v2, _v3) -> (
+      match v2 with
+      | `Exp e ->
+          let x = map_expression env e in
+          x.G.e
+      | `Semg_typed_meta (v, c, t) ->
+          let metavar = ident env v in
+          if AST_generic.is_metavar_name (fst metavar) && in_pattern env then
+            let colon = token env c in
+            let type_ = map_type_ env t in
+            G.TypedMetavar (metavar, colon, type_)
+          else raise Impossible)
   | `Struct_exp (v1, v2) ->
       let name : G.name =
         match v1 with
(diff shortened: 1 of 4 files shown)


languages/rust/generic/Parse_rust_tree_sitter.ml
```diff
@@ -1194,11 +1194,14 @@ and map_expression (env : env) (x : CST.expression) =
       in
       let rhs = map_expression env v3 in
       G.AssignOp (lhs, (op, tok), rhs)
-  | `Type_cast_exp (v1, v2, v3) ->
+  | `Type_cast_exp (v1, v2, v3) -> (
```
brandonspark: One thing I'm slightly afraid of is the fact that this takes away the possibility of matching an explicit cast within Rust, at least by using a metavariable. For instance, suppose I wanted to find cases of casting a variable to `int32`, I would probably want to write the pattern:
```
$X as int32
```
but this will now match as a typed metavariable, and match many things which are not just an explicit cast. I'd like it if we could preserve the choice, to the rule-writer, of which they want to match.

Additionally, this is a bit of a tough case, because I think the language of `as` lends itself to more thinking of code where you have explicitly said, "transform this expression into one of this type", versus the typing in parameters (like `x : t`), which is more passive (it reflects a property, that `x` has type `t`). In an ideal world, I'd prefer to write a pattern like `$X : ty` in Rust, to avoid that. We could do this by modifying the `tree-sitter` Semgrep grammar for Rust, but that might be a lot of work, so we don't have to do that.

At the very least, I notice that usually we surround Semgrep typed metavariables with parentheses. Could we move this case under `Paren_exp`, down below, so that we continue this behavior (and we allow some measure of using patterns as casts instead of typed metavariables)?

I will consult about the `$X : ty` thing in the meantime.

languages/rust/generic/Parse_rust_tree_sitter.ml
```diff
@@ -1194,11 +1194,14 @@ and map_expression (env : env) (x : CST.expression) =
...
       let expr = map_expression env v1 in
       let as_ = token env v2 (* "as" *) in
       let type_ = map_type_ env v3 in
-      G.Cast (type_, as_, expr)
+      match expr with
+      | { e = N (Id ((n, t), _)); _ } when AST_generic.is_metavar_name n ->
```
brandonspark: Could we case on the `env` (which may be `Pattern` or `Target`, defined above) to ensure that this only happens during pattern parsing? You can consult `Parse_julia_tree_sitter` for an example of where we do this.

brandonspark (changes requested):

languages/rust/generic/Parse_rust_tree_sitter.ml
brandonspark: It seems that we have consensus on preferring `$X : ty` as the syntax for typed metavariables, as per this thread: https://semgrepinc.slack.com/archives/C01NXGX2EHZ/p1688587347895529

Could you make that change? You should have to touch the `grammar.js` from `semgrep-rust` within `ocaml-tree-sitter-semgrep`, and it will entail making a PR to `ocaml-tree-sitter-semgrep` with the updated grammar, and then a PR to `semgrep` updating our version of `semgrep-rust`.

brandonspark (approved):
Benchmarks seems unrelated (same benchmarks issues as for everyone recently)

brandonspark:
(I deleted the comment because benchmarks CI seemed to be complaining about it... not sure why)

Issue #2252, opened 2020-12-14
Support ARM architecture


ajinabraham:
Prioriry: Low

**Is your feature request related to a problem? Please describe.**
Was trying to use semgrep on my android tablet. 
install is successful, but cannot invoke semgrep.

```
File "/data/data/com.termux/files/usr/lib/python3.9/site-packages/semgrep/util.py", line 135, in sub_run
    result = subprocess.run(cmd, **kwargs)  # nosem: python.lang.security.audit.dangerous-subprocess-use.dangerous-subprocess-use
  File "/data/data/com.termux/files/usr/lib/python3.9/subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/data/data/com.termux/files/usr/lib/python3.9/subprocess.py", line 947, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/data/data/com.termux/files/usr/lib/python3.9/subprocess.py", line 1817, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 8] Exec format error: 'semgrep-core'
- Semantic Grep  13
Traceback (most recent call last):
  File "/data/data/com.termux/files/usr/bin/njsscan", line 8, in <module>
    sys.exit(main())
  File "/data/data/com.termux/files/usr/lib/python3.9/site-packages/njsscan/__main__.py", line 66, in main
    scan_results = NJSScan(
  File "/data/data/com.termux/files/usr/lib/python3.9/site-packages/njsscan/njsscan.py", line 37, in scan
    result = scanner.scan()
  File "/data/data/com.termux/files/usr/lib/python3.9/site-packages/libsast/scanner.py", line 64, in scan
    results['semantic_grep'] = SemanticGrep(
  File "/data/data/com.termux/files/usr/lib/python3.9/site-packages/libsast/core_sgrep/semantic_sgrep.py", line 37, in scan
    self.format_output(sgrep_out)
  File "/data/data/com.termux/files/usr/lib/python3.9/site-packages/libsast/core_sgrep/semantic_sgrep.py", line 42, in format_output
    self.findings['errors'] = results['errors']
TypeError: 'NoneType' object is not subscriptable
```
I guess this is because the binary was not targeted for ARM.

**Describe the solution you'd like**

Would be nice to have a arm compatible semgrep-core

**Describe alternatives you've considered**

**Additional context**
Not top priority.

mschwager:
[...]

PR #8205, opened 2023-07-03
Build multi-arch docker images

tpetr:
This PR introduces support for multi-arch semgrep docker builds (amd64 + arm64). A few things to note:
- GHA only provides amd64 runners, so ARM builds are abhorrently slow ([45 mins](https://github.com/returntocorp/semgrep/actions/runs/5456229166/jobs/9928763358)). I cut us over to depot.dev, which provides dedicated ARM hardware and a faster docker cache. `depot/build-push-action` will fall back to the standard docker if depot.dev is offline.
- We run some pre-flight checks prior to pushing the image, which warranted some tricky maneuvers because neither docker nor depot can export (and docker can't import) a multi-arch image archive, so instead:
  - We build each arch individually, run the checks, and then upload the archive (name suffixed with the arch)
  - In `push-docker-multi-arch`, we:
    1. Extract the contents of the two images
    2. Generate a manifest list so that the docker image will show up as multi-arch in Docker Hub
    3. Push both images' blobs and manifests to Docker Hub
    4. Push the manifest list for the desired tags to Docker Hub

(this is a bit of a brain-melter so I'm happy to explain further. I can also add comments to the workflow)

All that being said, we should re-assess if `validate-docker-build.sh` is still useful (or if we can push to a staging repository), so that we can rely on docker's/depot's baked-in multi-arch builds

test plan:
- [x] manually triggered `build-test-docker` to push this branch as `returntocorp/test`. Confirmed that the [tom-arm64](https://hub.docker.com/r/returntocorp/test/tags) image works on both amd64 and arm64.

cc: https://github.com/returntocorp/semgrep/issues/2252

### DIFF
--- a/.github/workflows/build-test-docker.yaml
+++ b/.github/workflows/build-test-docker.yaml
@@ -165,3 +165,8 @@ jobs:
         with:
           name: ${{ inputs.artifact-name }}-arch-${{ matrix.architecture }}
           path: /tmp/image.tar
+      - uses: actions/upload-artifact@v3
+        if: ${{ matrix.architecture == 'amd64' }} # appease the test-semgrep-pro workflow
+        with:
+          name: ${{ inputs.artifact-name }}
+          path: /tmp/image.tar
--- a/.github/workflows/build-test-docker.yaml
+++ b/.github/workflows/build-test-docker.yaml
@@ -27,10 +27,15 @@ on:
         required: true
         type: boolean
         description: Whether or not to run validation on the built image
+      use-depot:
+        required: true
+        type: boolean
+        description: Whether or not to use depot.dev for docker builds
+        default: false
       push-immediately:
         required: true
         type: boolean
-        description: Whether or not to push immediately
+        description: Whether or not to immediately push the docker image after building
         default: false
 
   workflow_call:
@@ -55,6 +60,11 @@ on:
         required: true
         type: boolean
         description: Whether or not to run validation on the built image
+      use-depot:
+        required: true
+        type: boolean
+        description: Whether or not to use depot.dev for docker builds
+        default: false
 
 jobs:
   build-test-docker:
@@ -81,11 +91,24 @@ jobs:
           # :latest is also added automatically
           tags: ${{ inputs.docker-tags }}
       - uses: depot/setup-action@v1
-      - name: Build image
-        id: build-image
+        if: ${{ inputs.use-depot == true }}
+      - name: Build image (depot)
+        id: build-image-depot
+        if: ${{ inputs.use-depot == true }}
         uses: depot/build-push-action@v1.7.1
         with:
           project: fhmxj6w9z8
+          platforms: linux/${{ matrix.architecture }}
+          outputs: type=docker,dest=/tmp/image.tar
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          file: ${{ inputs.file }}
+          buildx-fallback: true
+      - name: Build image (docker)
+        id: build-image-docker
+        if: ${{ inputs.use-depot == false }}
+        uses: docker/build-push-action@v1.7.1
+        with:
           platforms: linux/${{ matrix.architecture }}
           outputs: type=docker,dest=/tmp/image.tar
           cache-from: type=gha,src=/tmp/.buildx-cache
@@ -93,7 +116,6 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           file: ${{ inputs.file }}
-          buildx-fallback: true
       - name: Load image
         run: |
           docker load --input /tmp/image.tar
--- a/.github/workflows/build-test-docker.yaml
+++ b/.github/workflows/build-test-docker.yaml
@@ -31,6 +31,10 @@ on:
         type: boolean
         description: Whether or not to use depot.dev instead of docker buildx to build the docker image
         default: false
+      multi-arch:
+        type: boolean
+        description: Whether or not to build a multi-arch docker image
+        default: false
 
   workflow_call:
     inputs:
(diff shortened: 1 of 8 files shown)


.github/workflows/build-test-docker.yaml
```diff
@@ -17,14 +17,21 @@ on:
...
         description: Dockerfile to build
+        default: Dockerfile
       enable-tests:
         required: true
         type: boolean
         description: Whether or not to run validation on the built image
+      push-immediately:
+        required: true
+        type: boolean
+        description: Whether or not to push immediately
```
jbergler: nit: this description doesn't really explain what this means, can you add an example or clarify in some other way?

cli/src/semgrep/cli.py
```diff
@@ -71,7 +71,7 @@ def cli(ctx: click.Context) -> None:
     state = get_state()
     state.terminal.init_for_cli()
 
-    abort_if_linux_arm64()
+    # abort_if_linux_arm64()
```
jbergler: Can probably just remove this yea?

.github/workflows/build-test-docker.yaml
```diff
@@ -17,14 +17,26 @@ on:
...
+        default: Dockerfile
       enable-tests:
         required: true
         type: boolean
         description: Whether or not to run validation on the built image
+      use-depot:
+        required: true
+        type: boolean
+        description: Whether or not to use depot.dev for docker builds
+        default: false
```
apt-itude: Bit of a nitpick here, but do these really need to be marked as `required` if you've provided a default?

.github/workflows/build-test-docker.yaml
```diff
@@ -48,38 +60,81 @@ on:
...
+    strategy:
+      matrix:
+        architecture:
+          - amd64
+          - arm64
     steps:
       - uses: docker/setup-buildx-action@v2
+      - if: ${{ matrix.architecture != 'amd64' }}
+        uses: docker/setup-qemu-action@v2
+      - uses: docker/setup-buildx-action@v2
```
apt-itude, line 120: Is it intentional that this step happens both before and after the `setup-qemu-action` step?

.github/workflows/push-docker.yaml
```diff
@@ -61,3 +69,57 @@ jobs:
...
+      - uses: actions/download-artifact@v3
+        with:
+          path: /tmp/artifacts
+      - run: |
+          artifact_filenames=(/tmp/artifacts/${{ inputs.artifact-name }}-arch-*/image.tar)
+          if [[ ${#artifact_filenames[@]} == 0 ]]; then
+            echo "No artifacts found matching ${{ inputs.artifact-name }}-arch-*"
+            exit 1
+          fi
+          mkdir -p /tmp/merged/blobs/sha256
```
apt-itude: A couple comments to explain the necessity and behavior of these more complex steps would probably be helpful for someone who has to touch this in the future so they don't have to dig up this PR.

apt-itude (commented):
This is very cool, but also:

> All that being said, we should re-assess if validate-docker-build.sh is still useful (or if we can push to a staging repository), so that we can rely on docker's/depot's baked-in multi-arch builds

Agreed

.github/workflows/build-test-docker.yaml
tpetr: Nope! Good catch.

.github/workflows/build-test-docker.yaml
tpetr: Probably not

.github/workflows/push-docker.yaml
tpetr: Added some comments and refactored things a bit so that we can flip back to non-depot single-arch builds if necessary

.github/workflows/build-test-docker.yaml
```diff
@@ -83,3 +102,66 @@ jobs:
...
+          # :latest is also added automatically
+          tags: ${{ inputs.docker-tags }}
+      - uses: depot/setup-action@v1
+        if: ${{ inputs.use-depot-for-docker-build }}
+      - name: Build image (via depot.dev)
+        id: build-image-depot
+        if: ${{ inputs.use-depot-for-docker-build }}
+        uses: depot/build-push-action@v1.7.1
+        with:
+          project: fhmxj6w9z8
```
cgdolan, line 135: How does authentication with depot work? Can we limit execution to only workflows running in this repository (i.e. not forks or other repos where this project id is specified in a workflow)?

cgdolan (approved):

.github/workflows/build-test-docker.yaml
cgdolan: tl;dr: from offline conversation, yes - this is handled by OIDC.

apt-itude (approved):
LGTM

tpetr:
Note to self (and anyone curious): the defaults in this PR enables depot for all docker builds, but only enables multi-arch for to the main branch (i.e. PR merges). I'll enable multi-arch for releases (and delete some cruft) once these changes have had time to bake.

.github/workflows/push-docker.yaml
```diff
@@ -61,3 +69,84 @@ jobs:
...
+          # Find docker image artifacts
+          image_filenames=(/tmp/artifacts/${{ inputs.artifact-name }}-arch-*/image.tar)
+          if [[ ${#image_filenames[@]} == 0 ]]; then
+            echo "No artifacts found matching ${{ inputs.artifact-name }}-arch-*"
+            exit 1
+          else
+            echo "Found images: ${image_filenames[@]}"
+          fi
+
+          # Create an empty manifest list. This will ultimately be the union of all the arch-specific image manifests we intend to publish.
```
aryx, line 96: A link to some doc about multi-arch docker image would be useful. I have no idea what are those manifest file
for example, and whoever will have to modify this file in 6 months could use some help to accelerate the learning.

PR #8206, opened 2023-07-03
fix(rust): fix metavariable name patterns

brandonspark:
## What:
This PR makes it so patterns that register as `Name` variants, that are metavariables, work properly as such.

## Why:
A `Name` pattern with a metavariable should be strictly more general than an `E (N (Id ...))` that denotes the same thing, by virtue of being able to match to `name` constructs that are not expressions, for instance the inside of a Constructor.

Right now, they are not such, however, as `Name` patterns can only match to other `Name`s. This means that things like metavariables in Rust (which parse to `Name` patterns) straight up do not work.

## How:
Made it so that `Name` patterns which are strictly metavariables are added as an `E` pattern. This will let them operate as they should.

In the process, I also updated the Rust parser with ellipsis metavariables, and added tests demonstrating its behavior.

## Test plan:
`make test`

Closes PA-2915
Closes PA-2912

PR checklist:


src/matching/Match_patterns.ml
```diff
@@ -258,6 +258,16 @@ let check2 ~hook mvar_context range_filter (config, equivs) rules
            | Fld pattern -> Common.push (pattern, rule) fld_rules
            | Flds pattern -> Common.push (pattern, rule) flds_rules
            | Partial pattern -> Common.push (pattern, rule) partial_rules
+           (* If we receive an incoming Name pattern, we want to use it
+              as a name only if it's not a metavariable because then it
+              should be matched against more nodes.
+              So we sort based on if its a metavariable.
+           *)
+           (* THINK: should we do this for named ellipses also? *)
+           | Name (Id ((s, _), _) as name) when Metavariable.is_metavar_name s
```
IagoAbal: Shouldn't we perhaps convert this earlier to an `E (Id ...)` like in `Parse_rust_tree_sitter`?

IagoAbal (approved):

src/matching/Match_patterns.ml
IagoAbal: My understanding is that in the other languages a standalone `$X` is just parsed as `E (Id ...)`, so that's the "contract" and `Parse_rust_tree_sitter` breaks it.

brandonspark: Hmm, I was thinking about it being useful for names to remain name patterns. But I realize you're saying just to convert the metavariable case, which I think is reasonable. I think we can actually do this in `Parse_pattern`, instead, you're right.

Issue #5221, opened 2022-05-16
[Rust] identifiers and deep ellipsis in macro calls are not parsed correctly


Matthewbialek:
**Describe the bug**
Slack thread: https://r2c-community.slack.com/archives/C018NJRRCJ0/p1652628754749719

I have noticed in a few of the rules that I have written for **RUST** that I can not match a variable that is a result of invocation of a method that is later used in a method call. 
Here is a  boiled down example of this : 
          ...
          $NPI = $OBJ.password(...);
          ...
          println!(<... $NPI ...>);
The NPI in println will never match. If I take out the NPI in the println and replace it with "..." it matches
Test Code : 
fn main() {
    // Leak sensitive info via print statement
    let pwd = password.password();
    println!(pwd);
}    
**To Reproduce**
Steps to reproduce the behavior, ideally a link to https://semgrep.dev:
https://semgrep.dev/s/gjzx

**Expected behavior**
I expect:
    println!(pwd);
to match
**Screenshots**
https://semgrep.dev/s/gjzx

**What is the priority of the bug to you?**

- [ X] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [ ] P2: regular bug that should get fixed

**Environment**
If not using semgrep.dev: are you running off docker, an official binary, a local build?
semgrep.dev
https://semgrep.dev/s/gjzx
**Use case**
What will fixing this bug enable for you?
It will empower me to use Semgrep as a SAST tool for Rust

r2c-demo:
This issue is synced in Linear at https://linear.app/r2c/issue/PA-1293/[rust]-match-parsing-with-variables-in-invocation. _Note_: this link is for r2c use only and is not accessible publicly.

aryx:
The problem is not related to the invocation, but to arguments to macro calls like println!
In theory you can pass anything to a macro call.

aryx:
and the pattern println!(<... $NPI ...>); is not parsed correctly.

PR #8209, opened 2023-07-04
fix(rust): best effort macro args 

brandonspark:
## What:
This PR makes it so we do best-effort to try and parse macro arguments as real arguments, if we can.

## Why:
In a macro such as `println!`, a bunch of random stuff can come after it. More specifically, macro items can come after it, which really just tokenize all the stuff and then break them down into various `any` constructs. It's really very unstructured.

Usually, we will translate them into `OtherArg` if we don't get precisely one argument. But, a macro call might receive many arguments. The problem is that the comma is not ignored, it's part of the macro items. So we need to process it ourselves.

## How:
We make a "best effort" to try and recognize sequences of macro items which look like parameters to a function call.

If we see anything which looks like a sequence of things which can be interpreted as expressions, delimited by commas, then we parse it as a true function call's arguments. Otherwise, we do what we did before. This might not suffice for every case, but it should be true in every case it lands in.

Notably, this permits the singleton and empty case for arguments as well.

## Test plan:
`make test`

Closes PA-2902

PR checklist:


changelog.d/pa-2902.fixed
```diff
@@ -0,0 +1,2 @@
+Rust: Fixed a bug where taint could not propagate through
```
IagoAbal: You have more context than me for sure, but I'm not sure I'd qualify this as a bug, may it be a feature?

languages/rust/generic/Parse_rust_tree_sitter.ml
```diff
@@ -82,11 +82,46 @@ type rust_macro_item =
...
- * they are not separators between rust_macro_items.
- *)
 let rec macro_items_to_anys (xs : rust_macro_item list) : G.any list =
-  xs |> Common.map macro_item_to_any
+  (* Note that the commas are considered like any other tokens in a Rust macro;
+     * they are not separators between rust_macro_items.
+  *)
+  (* Thus, we need this machinery in order to parse a list of macro items
+     to see if the macro invocation looks like a normal expression. For instance,
+     if a sequence of macro items looks like
```
IagoAbal, line 91: Is there a risk to get this wrong?

changelog.d/pa-2902.fixed
brandonspark: honestly, the line is hazy to me too. i'm not really sure where one ends and the other begins... "fixed", because we should have been able to do this in the first place. but also, it's reasonable that we didn't, and we just downscoped it, so feature?? not sure.

brandonspark: i can rename it to feature, i don't mind

languages/rust/generic/Parse_rust_tree_sitter.ml
brandonspark: the default behavior is we fail to parse as a normal sequence of arguments, and the translation proceeds as before. if we get it wrong, then we might turn it into `Args` when we shouldn't, and therefore pass normal arguments into the macro. However, otherwise it would just be a sequence of `any`s, like `G.E, G.Tk, G.E, G.Tk` which I doubt would have any particular semantic behavior. so i don't think there's any risk.

languages/rust/generic/Parse_rust_tree_sitter.ml
```diff
@@ -82,11 +82,46 @@ type rust_macro_item =
...
+    | MacTree _ ->
+        None
+  in
+  let rec try_as_normal_args = function
+    | [] -> Some []
+    | [ mac ] ->
+        let* arg = macro_item_to_arg mac in
+        Some [ arg ]
+    (* For now, we just directly case on the token to see if its a comma.
+       This is a fragile approach, but we're close to the origin of the
```
IagoAbal: In which sense it is "fragile", and what does "relatively fine" mean?

languages/rust/generic/Parse_rust_tree_sitter.ml
```diff
@@ -82,11 +82,46 @@ type rust_macro_item =
...
+       token, so this should be relatively fine.
+       It's also a lot more work to bring this information over from
+       when we first parse the token.
+    *)
+    | mac :: MacAny (G.Tk (Tok.OriginTok { str = ","; _ })) :: rest ->
+        let* arg = macro_item_to_arg mac in
+        Option.map (fun res -> arg :: res) (try_as_normal_args rest)
```
IagoAbal: Why not just use `let*` here too ?

IagoAbal (approved):

changelog.d/pa-2902.fixed
IagoAbal: No strong opinions on my side, just wanted to make you think about it.

IagoAbal: Features are something more positive than bugs, so that if both could apply, perhaps that could be a reason to prefer presenting something as a feature. 🤔 🤷‍♂️

languages/rust/generic/Parse_rust_tree_sitter.ml
brandonspark: I suppose I mean "fragile", as in there could be weird cases where the string of the token is not exactly `","`, because I don't fully trust pattern matching on the string inside the token. I don't know that for sure, but this is just less statically safe to my mind. "Relatively fine", as in, there's few opportunities for the string inside the token to change, and there's not much else we could have done prior to this to make this logic easier.

brandonspark: I can leave a comment

languages/rust/generic/Parse_rust_tree_sitter.ml
brandonspark: Good call!

changelog.d/pa-2902.fixed
brandonspark: oh that's a good point, I didn't think of that one. will change!

aryx:
Very nice!
Maybe this will fix https://github.com/returntocorp/semgrep/issues/5221

PR #8216, opened 2023-07-05
feat(tainting): allow pattern params as sources

brandonspark:
## What:
This PR allows the tainting of arguments in languages where you can pattern match in function args, like Rust and OCaml.

## Why:
A common pattern is to taint the argument of a function, in its argument position, and then expect that identifier to be tainted in the actual body of the function.

We weren't able to do this before in languages where those arguments are actually patterns.

## How:
Because patterns are recursive, and the identifier you are tainting might be arbitrarily deep, I ended up implementing a visitor (which can take in an `any`, for maximum generality) which collects the identifiers in the patterns that are given as input, so that we could add them all to the environment as tainted, if they are sources.

## Test plan:
`make test`

Closes PA-2917
Closes PA-2919

PR checklist:

### DIFF
--- a/libs/ast_generic/AST_generic_helpers.ml
+++ b/libs/ast_generic/AST_generic_helpers.ml
@@ -528,52 +528,6 @@ let range_of_any_opt any =
   [@@profiling]
 
 (*****************************************************************************)
-(* Identifiers from any *)
-(*****************************************************************************)
-
-class ['self] id_visitor =
-  object (_self : 'self)
-    inherit ['self] AST_generic.iter_no_id_info as super
-
-    method! visit_name store name =
-      match name with
-      | Id (id, id_info) ->
-          super#visit_name store name;
-          push (id, id_info) store
-      | IdQualified _ -> super#visit_name store name
-
-    method! visit_pattern store pat =
-      match pat with
-      | PatAs (_, (id, id_info))
-      | PatId (id, id_info) ->
-          super#visit_pattern store pat;
-          push (id, id_info) store
-      | PatLiteral _
-      | PatConstructor _
-      | PatRecord _
-      | PatTuple _
-      | PatList _
-      | PatKeyVal _
-      | PatUnderscore _
-      | PatDisj _
-      | PatTyped _
-      | PatWhen _
-      | PatType _
-      | PatEllipsis _
-      | DisjPat _
-      | OtherPat _ ->
-          super#visit_pattern store pat
-  end
-
-let ids_of_any : AST_generic.any -> (G.ident * G.id_info) list =
-  let v = new id_visitor in
-  let ids = ref [] in
-  fun any ->
-    v#visit_any ids any;
-    !ids
-  [@@profiling]
-
-(*****************************************************************************)
 (* Fix token locations *)
 (*****************************************************************************)
 
--- a/libs/ast_generic/AST_generic_helpers.ml
+++ b/libs/ast_generic/AST_generic_helpers.ml
@@ -528,6 +528,52 @@ let range_of_any_opt any =
   [@@profiling]
 
 (*****************************************************************************)
+(* Identifiers from any *)
+(*****************************************************************************)
+
+class ['self] id_visitor =
+  object (_self : 'self)
+    inherit ['self] AST_generic.iter_no_id_info as super
+
+    method! visit_name store name =
+      match name with
+      | Id (id, id_info) ->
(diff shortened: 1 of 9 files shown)


src/engine/Match_tainting_mode.ml
```diff
@@ -681,8 +681,18 @@ let check_fundef lang options taint_config opt_ent ctx java_props_cache fdef =
                     add_to_env env id ii None
                 | G.F _ -> env)
               env fields
+        | G.ParamPattern pat ->
```
IagoAbal, line 684: Need to refactor to avoid this duplication, but in Pro there is also a place in `Deep_tainting` where you would need to add support for `ParamPattern` (where we assign polymorphic taint).

libs/ast_generic/AST_generic_helpers.ml
```diff
@@ -527,6 +527,52 @@ let range_of_any_opt any =
...
+      | PatWhen _
+      | PatType _
+      | PatEllipsis _
+      | DisjPat _
+      | OtherPat _ ->
+          super#visit_pattern store pat
+  end
+
+let ids_of_any : AST_generic.any -> (G.ident * G.id_info) list =
+  let v = new id_visitor in
```
IagoAbal: ~Do we need a new `id_visitor` every time? I think for other visitors we use a `lazy` or some other way to cache the visitor.~

IagoAbal: Ooops, this was already taken care of.

src/engine/Match_tainting_mode.ml
```diff
@@ -681,8 +681,18 @@ let check_fundef lang options taint_config opt_ent ctx java_props_cache fdef =
...
+            (* Here, we just get all the identifiers in the pattern, which may
+               themselves be sources.
```
IagoAbal, line 686: A shorter comment but with a concrete example could be better.

IagoAbal (approved):
Nice!

src/analyzing/Visit_pattern_ids.ml
```diff
@@ -0,0 +1,51 @@
...
+
+    method! visit_pattern store pat =
+      match pat with
+      | PatAs (_, (id, id_info))
+      | PatId (id, id_info) ->
+          super#visit_pattern store pat;
+          Common.push (id, id_info) store
+      | PatLiteral _
+      | PatConstructor _
+      | PatRecord _
```
IagoAbal: Weren't you going to treat records specially to avoid adding variables to the context that should not be in it?

src/analyzing/Visit_pattern_ids.ml
```diff
@@ -0,0 +1,51 @@
...
+      | PatList _
+      | PatKeyVal _
+      | PatUnderscore _
+      | PatDisj _
+      | PatTyped _
+      | PatWhen _
+      | PatType _
+      | PatEllipsis _
+      | DisjPat _
+      | OtherPat _ ->
```
IagoAbal, line 65: In Deep_tainting we had to handle this `OtherPat` case specially https://github.com/returntocorp/semgrep-proprietary/pull/833/files#diff-c1d42c3d0b66c474284f82bb2b7fad3ab670965fee3a6cade993129f63320877L324 (not sure why it wasn't a `PatRecord` 🤷‍♂️), and then traverse the fields of the record https://github.com/returntocorp/semgrep-proprietary/pull/833/files#diff-c1d42c3d0b66c474284f82bb2b7fad3ab670965fee3a6cade993129f63320877L332. Have you tried this in Pro?

src/analyzing/Visit_pattern_ids.ml
brandonspark: This shouldn't add the labels of the variables though, right? We only visit the inner patterns recursively, we don't add the labels (which are `dotted_ident`s) to the env. Correct me if I'm wrong.

brandonspark: Actually, it's the other way around here I guess, where we aren't adding the record patterns (not the labels) properly.

src/analyzing/Visit_pattern_ids.ml
brandonspark: Oh, I thought that was subsumed under the `PatRecord` case, but I only just noticed it's an expression record and not a PatRecord. I just restored that case here, in the most recent commit.

PR #8217, opened 2023-07-05
feat: [nightly] install semgrep via brew non-interactively

spencerdrak:
In our nightly tests, we install semgrep via brew to check for compatibility. In some cases when this fails, brew will stop and wait for user input, which will not be given, and GHA times out the job after 6 hours. 

Run the brew install in non-interactive mode. 

PR checklist:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -35,6 +35,7 @@ jobs:
       - name: Brew Install
         env:
           HOMEBREW_NO_INSTALL_FROM_API: 1
+          NONINTERACTIVE: 1
         run: brew install semgrep --HEAD --debug
       - name: Check installed correctly
         env:


spencerdrak:
Reference: https://docs.brew.sh/Installation

cgdolan (approved):

.github/workflows/nightly.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       - name: Brew Install
         env:
           HOMEBREW_NO_INSTALL_FROM_API: 1
+          NONINTERACTIVE: 1
```
aryx, line 38: Instead of your comment in the PR subject, it would be better to add your comment in the yaml file instead, so in 6 months someone understand why we need this NONINTERACTIVE flag.

PR #8218, opened 2023-07-05
use variable for last docker validation test as well

tpetr:
We're seeing random failures in the arm64 test image step -- I have a feeling that the `docker run` process is exiting too soon and closing the pipe before `grep` can finish.

I thought the stdin test was the culprit, so I saved the output to a variable and then echo-ed it to grep, but it looks like the other test is flaky too. This PR applies this change to the other flakey test.

test plan:
- checks pass


tpetr:
heads up @brandonspark @spencerdrak this should fix the test failure you're seeing in your PRs

cgdolan (approved):

spencerdrak (approved):

brandonspark:
@tpetr ah i was wondering about that, thanks!

brandonspark (approved):

PR #8220, opened 2023-07-05
Fix broken links to example rules in readme

jbergler:
Just a bit of house keeping to un-break these links!


pangngan (approved):
🎉

ievans:
ty @jbergler !

PR #8222, opened 2023-07-06
feat(SSC): Scan un-edited lockfiles in PR scans

mmcqd:
If a PR introduces a new vulnerable usage of a dependency that was already in their lockfile, but does not edit their lockfile, we currently will not catch it, because on PR scans we only target files that were edited in the PR. 

This fixes that by adding an `ignore_baseline_handler` option to the file targeting code that just ignores whether or not we're doing a PR scan. Then `find_single_lockfile`, which is used when generating reachable findings to locate the lockfile associated to a given code file, sets this to `True`. So reachable finding generation can bypass the `baseline_handler`, but we won't just scan every lockfile on every PR scan.

I also update `generate_reachable_sca_findings` to return a `FrozenSet[Path]` of the lockfiles scanned during its execution, which are added to `all_targets` and the scan's dependencies.

PR checklist:

### DIFF
--- a/cli/src/semgrep/dependency_aware_rule.py
+++ b/cli/src/semgrep/dependency_aware_rule.py
@@ -2,6 +2,7 @@ from functools import lru_cache
 from pathlib import Path
 from typing import Callable
 from typing import Dict
+from typing import FrozenSet
 from typing import Iterator
 from typing import List
 from typing import Tuple
@@ -140,7 +141,10 @@ def transivite_dep_is_also_direct(
 def generate_reachable_sca_findings(
     matches: List[RuleMatch], rule: Rule, target_manager: TargetManager
 ) -> Tuple[
-    List[RuleMatch], List[SemgrepError], Callable[[Path, FoundDependency], bool]
+    List[RuleMatch],
+    List[SemgrepError],
+    Callable[[Path, FoundDependency], bool],
+    FrozenSet[Path],
 ]:
     depends_on_keys = rule.project_depends_on
     dep_rule_errors: List[SemgrepError] = []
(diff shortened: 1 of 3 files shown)


cli/src/semgrep/dependency_aware_rule.py
```diff
@@ -163,6 +168,7 @@ def generate_reachable_sca_findings(
                 )
                 if lockfile_path is None:
                     continue
+                scanned_lockfiles.append(lockfile_path)
```
brendongo, line 167: wont this be all lockfiles with a finding not necessarily scanned lockfiles? aka it will miss lockfiles without findings

cli/src/semgrep/target_manager.py
```diff
@@ -471,15 +477,15 @@ def files_from_filesystem(self) -> FrozenSet[Path]:
         )
 
     @lru_cache(maxsize=None)
-    def files(self) -> FrozenSet[Path]:
+    def files(self, ignore_baseline_handler: bool = False) -> FrozenSet[Path]:
```
brendongo, line 487: probably want to edit function level comment on what ignore_baseline_handler does when set to true (maybe also why we need it)

aryx:
Can we close this PR @mmcqd ?

aryx:
Been inactive for almost 2 months, closing.

cli/src/semgrep/dependency_aware_rule.py
brendongo: note: that's ok since if the PR did not introduce a new reachable finding, the lockfile doesnt need to be scanned so it's ok to miss it

PR #8224, opened 2023-07-06
improve cache hits by omitting unnecessary folders from semgrep-core docker build

tpetr:
One of the first steps in our `Dockerfile` is to copy the entire semgrep repo into the container building `semgrep-core`, which means that we have to pay the cost of a full semgrep-core build any time any file is changed (because the resulting docker cache miss invalidates the cache of all subsequent steps). This PR filters out some folders that we know will never affect the `semgrep-core` build, which should improve our cache hit rates, and in turn, our docker build times.

test plan:
- [x] checks pass
- [x] pushed docker image works properly


apt-itude (approved):

Dockerfile
```diff
@@ -15,6 +15,28 @@
...
+# semgrep-core. This is done in a multi-stage build so that the final COPY
+# happens in a single layer.
+
+FROM busybox:stable as semgrep-core-files
+WORKDIR /src/semgrep
+
+# copy over the entire semgrep repository
+COPY . .
+
+# remove folders that aren't necessary for the semgrep-core build
```
aryx, line 34: Not a fan. Is it worth it? Now if in the futur we add more directories in semgrep we need to modify this list ... You avoid the docker build time for PRs that modify the cli/ and js/ which are rare, so is it really worth the added complexity?

aryx: why not adding stats/ to this list? what about perf/? Seems like an arbitrary list that is bound to be obsolete in a few weeks and not really optimizing anything.

aryx: Or is it more useful because we started to switch to depot.dev and the optimization is more visible?

tpetr: Yep, these changes have been a huge help for me as I've iterated on changes to GHA, python wheels, semgrep.js, etc... (basically anything that isn't semgrep-core/osemgrep itself). The docker build has done down from minutes to ~10 sec if everything is cached.

You're right that there are likely a few more folders that I could exclude, but I wanted to stick with the folders that I knew for certain weren't necessary for the docker build. It is unfortunate that this list will become out-of-date over time, but the way I see it:
- if we start needing a folder that is deleted, the build will fail and we'll fix it
- if a new folder is added that is causing unacceptable slowness, someone will get annoyed and we'll fix it

PR #8225, opened 2023-07-06
fix(rust): match attributes attached to functions

akuhlens:
What:
This PR fixes the rust conversion to generic so that it isn't discarding attributes. Attribute parsing isn't complete yet as attributes patterns don't allow nested ellipsis, metavariables, or discrimination based on nested parts of the pattern. This will be added in a future PR.

Why:
This allows functions and other declarations to be pattern matched against based on the attributes on them.

How:
Modified the conversion to generic AST for rust lowering attributes from how they are currently parsed. Converts meta_item in the current parse to attribute in `Ast_generic` and collects the leading outer attribute of a declaration and attaches them to the following declaration.

Test plan:
I added two tests demonstrating matching and its capabilities and limitations.
`make test`

This addresses but isn't a complete fix for [PA-2903](https://linear.app/semgrep/issue/PA-2903)

PR checklist:


languages/rust/generic/Parse_rust_tree_sitter.ml
```diff
@@ -674,10 +657,23 @@ and map_range_pattern_bound (env : env) (x : CST.anon_choice_lit_pat_0884ef0) :
...
+    G.argument =
+  let name = map_path_name env v1 in
+  match v2 with
+  | None -> G.(Arg (N name |> e))
+  | Some (`Meta_args x) ->
+      G.(Arg (Call (N name |> e, map_meta_arguments env x) |> e))
+  | Some (`EQ_lit (v1, v2)) ->
+      let _equals = token env v1 (* "=" *) in
+      let lit = map_literal env v2 in
+      G.(OtherArg (fake_id "MetaArgAssign", [ E (L lit |> e) ]))
```
brandonspark: seems you lost the `name` here, probably just need it as an additional entry in the `any list`

languages/rust/generic/Parse_rust_tree_sitter.ml
```diff
@@ -2063,23 +2057,17 @@ and map_meta_arguments (env : env) ((v1, v2, v3, v4) : CST.meta_arguments) :
...
-        | `EQ_lit (v1, v2) ->
-            let _equals = token env v1 (* "=" *) in
-            let lit = map_literal env v2 in
-            MetaItemLiteral lit
-        | `Meta_args x -> MetaItemMetaArgs (map_meta_arguments env x))
-      v2
-  in
-  (path, value)
+  match v2 with
+  | None -> NamedAttr (tok, name, (G.sc, [], G.sc))
```
brandonspark: instead of wrapping in `(G.sc, _, G.sc)`, you can use `fb`, defined above! (which stands for "fake bracket")

languages/rust/generic/Parse_rust_tree_sitter.ml
```diff
@@ -2063,23 +2057,17 @@ and map_meta_arguments (env : env) ((v1, v2, v3, v4) : CST.meta_arguments) :
...
+  | Some (`Meta_args x) -> NamedAttr (tok, name, map_meta_arguments env x)
+  | Some (`EQ_lit (v1, v2)) ->
+      let _equals = token env v1 (* "=" *) in
+      let lit = map_literal env v2 in
+      OtherAttribute (("AttrAssign", tok), [ G.E (G.L lit |> G.e) ])
```
brandonspark: same as above

tests/patterns/rust/attribute_matching.rs
```diff
@@ -0,0 +1,30 @@
...
+#[more(complicated, attribute)]
+fn foo1() {
+}
+
+#[test]
+fn fooToo() {
+}
+
+// MATCH:
+#[test]
```
brandonspark: If the order is reversed, should a match be produced? Can we add a test case exhibiting this behavior?

changelog.d/pa-2903.fixed
```diff
@@ -0,0 +1 @@
+Fixed rust attribute patterns to allow matching on simple attribute syntax.
```
brandonspark, line 1: Sometimes, providing an example of what can now be matched is nice.

brandonspark (approved):
Looks good! Awesome stuff! Just a few inline comments to address.

mjambon (approved):

tests/patterns/rust/attribute_matching.rs
akuhlens: Added a test case to demonstrate.

PR #8231, opened 2023-07-07
fallback default to workaround inputs.multi-arch being null

jbergler:
Fixes this release workflow issue:
```
release: .github#L1
The template is not valid. .github/workflows/release.yml (Line: 152, Col: 19): Unexpected value ''
```


tpetr (approved):
Whacky, thanks

PR #8232, opened 2023-07-07
fix: [ci] update handling of multi-arch input for release

spencerdrak:
Our last release failed due to the `multi-arch` input being null, which was not acceptable as a boolean input. [A fix was applied here](https://github.com/returntocorp/semgrep/commit/8157c0cfaaa039f444f10dca5e042acdad1f7d2b), but that also passes false as a string, so I'm not sure that it'll be acceptable either, and small test shows that this might not work as expected. 

To remedy this, we can handle the multi-arch input the same way handle dry-run input, and explicitly check for the values we're looking for. 

I'm also open to updating how we handle/trigger the `release.yml` job to standardize things, but would prefer to do that large refactor in a separate step. This PR just uses a pattern that we're already relying on to get us back, but we can update once release is complete.

Test Plan:
- [Dry run here](https://github.com/returntocorp/semgrep/actions/runs/5487629034) - note that this dry-run still isn't a 100% valid test, since we aren't running it based on a tag push. We won't know for sure until we attempt to actually push the image if this has been updated successfully. 

PR checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,11 +41,12 @@ on:
       - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
 
 jobs:
-  dry-run:
-    name: Evaluate Dry Run
+  inputs:
+    name: Evaluate Inputs
     runs-on: ubuntu-22.04
     outputs:
       dry-run: ${{steps.dry-run.outputs.dry-run}}
+      multi-arch: ${{steps.multi-arch.outputs.multi-arch}}
     steps:
       - name: Evaluate Dry Run
         id: dry-run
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -57,12 +58,22 @@ jobs:
             echo "dry-run=false" >> $GITHUB_OUTPUT
             echo "Setting dry-run to FALSE"
           fi
+      - name: Evaluate Multi-Arch
+        id: multi-arch
+        run: |
+          if [ "${{ inputs.multi-arch }}" = "true" ]; then
```
tpetr: uberest nit but can we consolidate on `[[ foo == bar ]]` instead of `[ foo = bar ]`? the latter works but it's confusing to folks

spencerdrak: Yes, and then I'll run another dry-run. Thanks!

.github/workflows/release.yml
```diff
@@ -41,28 +41,39 @@ on:
...
             echo "dry-run=true" >> $GITHUB_OUTPUT
             echo "Setting dry-run to TRUE"
           else
             echo "dry-run=false" >> $GITHUB_OUTPUT
             echo "Setting dry-run to FALSE"
           fi
+      - name: Evaluate Multi-Arch
+        id: multi-arch
+        run: |
+          if [[ "${{ inputs.multi-arch }}" == "true" ]]; then
```
tpetr, line 64: isnt the best practice to pass these through via env vars?

tpetr (approved):
LGTM, just a nit about the multi-arch thing

cgdolan (approved):
This looks fine to me but I still can't shake the feeling that this is more complicated than it needs to be.

spencerdrak:
> This looks fine to me but I still can't shake the feeling that this is more complicated than it needs to be.

@cgdolan agreed - I think we ought to take a look at refactoring this in a separate PR to remove the complexity. I'll take that as an AI, if we're alright with keeping things status-quo for now. I don't want to introduce any bugs by trying to refactor this now.

apt-itude (approved):

.github/workflows/release.yml
spencerdrak: I think so - however, I'll open a PR to remove this entirely and simplify the handling of the inputs here.

PR #8235, opened 2023-07-07
enable multi-arch builds for releases

tpetr:
We've had enough multi-arch branch builds where I feel comfortable with this.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,7 +61,7 @@ jobs:
       - name: Evaluate Multi-Arch
         id: multi-arch
         run: |
-          if [[ "${{ inputs.multi-arch }}" == "true" ]]; then
+          if [[ "${{ inputs.multi-arch }}" == "true" || "${{ inputs.multi-arch }}" == "" ]]; then
             echo "multi-arch=true" >> $GITHUB_OUTPUT
             echo "Setting multi-arch to TRUE"
           else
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -120,6 +120,7 @@ jobs:
       artifact-name: image-release
       file: Dockerfile
       enable-tests: true
+      multi-arch: true
```
spencerdrak: Does this not also need the conditional, or should we always be building multi-arch images and only sometimes be pushing them?

.github/workflows/start-release.yml
```diff
@@ -26,6 +26,11 @@ on:
         type: boolean
         description: Check the box for a dry-run - A dry-run will not push any external state (branches, tags, images, or PyPI packages).
         default: false
+      multi-arch:
```
spencerdrak, line 29: This is currently unused, but will be used if we start triggering the `release.yml` directly from this job. Do we want to leave that around until that time?

tpetr: Gotcha, I figure it's fine to leave around. Ultimately I'm going to remove the option to not build multi-arch (just want a little more bake time) so all the `multi-arch`es are temporary

.github/workflows/release.yml
tpetr: Right, I'll switch to conditional. We'll need build and push to match.

spencerdrak (approved):

PR #8237, opened 2023-07-07
update multi-arch pusher to push :latest

tpetr:
Our multi-arch docker image pusher isn't updating the `latest` tag, lets fix that

test plan:
- checks pass
- `latest` tag is indeed updated


spencerdrak (approved):

tpetr:
turns out we don't need to do this

PR #8243, opened 2023-07-10
build aarch64 python wheels

tpetr:
This PR adds support for building `aarch64` python wheels. This isn't simply a copypasta of the `x86_64` workflow because the cost of ARM emulation in GHA would make this job take forever. Instead, we:
- Add a new intermediate stage to the Dockerfile (`semgrep-wheel`) which:
   1. installs some prereqs (`build-base` because `ruamel.yaml` has native code  that needs to be compiled, `zip` because we gotta zip)
   2. copies the arch-specific `semgrep-core` binary from the `semgrep-core-container` stage
   3. builds the python wheel
   4. runs a script to validate that the wheel installs correctly and semgrep works
  (We don't have to worry about folks needlessly building this image because `docker build` only targets the last image in the Dockerfile by default)
- Add a new workflow that:
   - Kicks off a docker build targeting the `semgrep-wheel` image. This will be very fast because the docker layer cache will be populated already
   - If the image builds successfully, we extract `dist.zip` from the image and upload it as an artifact
- Remove the ARM64 check in the CLI (woo!)

If this ends up working smoothly, I'd like to consolidate on using this workflow for the `x86_64` wheel as well. (future PR)

NOTE: the CLI depends on `ruamel.yaml` which doesn't currently have an aarch64 python wheel published to PyPI, which means that gcc is required in order to install this dependency. Not ideal, but better than what we have currently (which is nothing). Medium-term we can look into switching from ruamel.yaml to pyyaml, and long-term osemgrep will make this go away.

test plan:
- checks pass
- grabbed the `manylinux-wheel-aarch64` artifact and confirmed it installed correctly and semgrep scan worked in an arm64 docker container

related: https://github.com/returntocorp/semgrep/issues/2252

### DIFF
--- a/.github/workflows/build-test-manylinux-aarch64.yaml
+++ b/.github/workflows/build-test-manylinux-aarch64.yaml
@@ -28,5 +28,5 @@ jobs:
           docker run --platform linux/arm64 --rm ${{ steps.build-semgrep-wheel.outputs.imageid }} zip -r - dist > /tmp/dist.zip
       - uses: actions/upload-artifact@v3
         with:
-          name: wheel-manylinux-aarch64
+          name: manylinux-wheel-aarch64
           path: /tmp/dist.zip
--- a/.github/workflows/build-test-manylinux-aarch64.yaml
+++ b/.github/workflows/build-test-manylinux-aarch64.yaml
@@ -1,0 +1,32 @@
+name: build-test-manylinux-aarch64
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  build-wheel:
+    name: Build and test Semgrep python wheel
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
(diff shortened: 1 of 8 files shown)


.github/workflows/build-test-manylinux-aarch64.yaml
```diff
@@ -0,0 +1,32 @@
...
+        uses: depot/build-push-action@v1.7.1
+        with:
+          project: fhmxj6w9z8
+          platforms: linux/arm64
+          outputs: type=docker,dest=/tmp/image.tar
+          target: semgrep-wheel
+      - name: Extract wheel from docker image
+        run: |
+          docker load --input /tmp/image.tar
+          docker run --platform linux/arm64 --rm ${{ steps.build-semgrep-wheel.outputs.imageid }} zip -r - dist > /tmp/dist.zip
```
apt-itude, line 28: Sick

apt-itude (approved):

PR #8199, opened 2023-07-01
Create a semgrep user for our docker container so that people can run…

bmahe:
… it as a non-root user

PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #8251, opened 2023-07-11
tainting: Fix Taint.pick_taint

IagoAbal:
Given two taints with the same source, Semgrep is supposed to pick the taint with the shortest call trace, but due to a typo this was not guaranteed.

Found this by chance while working on a new version of `pick_taint`, I don't have a test case for it.

Also fixed non-tail-rec use of `List.filter_map` (not sure why we had not found this before).

test plan:
make test

PR checklist:

### DIFF
--- a/semgrep.yml
+++ b/semgrep.yml
@@ -408,3 +408,11 @@ rules:
     fix: ""
     languages: [ocaml]
     severity: WARNING
+
+  - id: no-match-x-x-with
+    pattern: match $X, $X with ... -> ...
+    message: >-
+      Pattern matching on ($X, $X) is pointless, you are comparing $X with itself.
+      You probably meant ($X, something-else) and this is due to a typo.
+    languages: [ocaml]
+    severity: WARNING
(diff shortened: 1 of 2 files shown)


emjin (approved):
You could perhaps add a test case using snapshot testing?

brandonspark:
that's incredible that I never noticed that, even though I have explicit memories of reading these lines

emjin:
Honestly we could write a semgrep rule for `match ($X, $X)`

IagoAbal:
> Honestly we could write a semgrep rule for match ($X, $X)

I can do that instead of adding a regression test which I don't have. I guess that if I would try, I would eventually find one, basically Semgrep outputting a longer trace than the "optimal" one. But going through that work (could be fast or maybe not...) just to fix this seems a bit too much!

IagoAbal:
> > Honestly we could write a semgrep rule for match ($X, $X)
> 
> I can do that instead of adding a regression test which I don't have. I guess that if I would try, I would eventually find one, basically Semgrep outputting a longer trace than the "optimal" one. But going through that work (could be fast or maybe not...) just to fix this seems a bit too much!

Actually it's not even possible to reproduce this in Semgrep alone... because what was flawed was the comparison of the call trace, and that requires Pro.

IagoAbal:
Merging because ci/circleci was passing before, and it stop working apparently due to PR #8199. PR #8267 is meant to fix ci/circleci but pre-commit is failing.... which this PR should fix.

PR #8252, opened 2023-07-11
fix: [ci] nightly brew tests failing

spencerdrak:
Brew nightly verification has been failing - this error was originally obscured by a required update to the homebrew formula, which was applied to upstream `homebrew/homebrew-core` on 07JUL. This PR updates the way that `semgrep` is installed via brew on the Github-Hosted x86 MacOS runners. Because python3 already exists on the system, the brew link step fails, so we must also tell brew to link semgrep with the `--overwrite` flag.

See https://github.com/Homebrew/brew/issues/1742 for context on the brew link step.

It's unclear why this wasn't failing during our test runs - I noticed that new `macos-12` runner images were published around the same time, but they make no mention of how python was installed. In any case, this appears to be the way that the brew maintainers recommend proceeding. 

Test Plan:
- waiting for a test run here to pass: https://github.com/returntocorp/semgrep/actions/runs/5521938423/jobs/10070781425

PR checklist:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -23,15 +23,16 @@ jobs:
     runs-on: macos-12
     steps:
       - name: Brew update
+        env:
+          HOMEBREW_NO_INSTALL_FROM_API: 1
         run: brew update --debug --verbose
       - name: Brew Install
         env:
           NONINTERACTIVE: 1
+          HOMEBREW_NO_INSTALL_FROM_API: 1
         # See https://github.com/Homebrew/brew/issues/1742 for context on the brew link step.
         run: brew install semgrep --HEAD --debug || brew link --overwrite semgrep
       - name: Check installed correctly
+        env:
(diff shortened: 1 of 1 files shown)


spencerdrak:
Looks like pre-commit is failing here due to osemgrep, but was passing locally. I think it should be safe to merge this, but checking with the team first.

apt-itude (approved):

cgdolan (approved):

.github/workflows/nightly.yml
```diff
@@ -36,7 +36,8 @@ jobs:
         env:
           HOMEBREW_NO_INSTALL_FROM_API: 1
           NONINTERACTIVE: 1
-        run: brew install semgrep --HEAD --debug
+        # See https://github.com/Homebrew/brew/issues/1742 for context on the brew link step.
```
aryx, line 39: great comment!

PR #8258, opened 2023-07-12
fix(supply-chain): baseline scans reporting on existing findings

brendongo:
lockfile parsing was caching based only on file name which was resulting in the lockfile being cached even between baseline and head scans. When changes were added above an existing vulnerability this would cause match_based_id to access the wrong lines (everything will be off by new additions or deletions) which caused deduplication logic to be incorrect

This PR adds a test to catch this case and also caches the parsing on both filename and last update time of the file.

PR checklist:

### DIFF
--- a/cli/src/semdep/parse_lockfile.py
+++ b/cli/src/semdep/parse_lockfile.py
@@ -1,3 +1,4 @@
+import os
 from functools import lru_cache
 from pathlib import Path
 from typing import Callable
@@ -114,7 +115,6 @@ def lockfile_path_to_manifest_path(lockfile_path: Path) -> Optional[Path]:
     return manifest_path
 
 
-@lru_cache(maxsize=1000)
 def parse_lockfile_path(
     lockfile_path: Path,
 ) -> Tuple[List[FoundDependency], Optional[DependencyParserError]]:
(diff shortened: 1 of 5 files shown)


cli/src/semdep/parse_lockfile.py
```diff
@@ -123,6 +123,22 @@ def parse_lockfile_path(
...
+
+    Raises SemgrepError if the lockfile is not supported
+    """
+    file_changed_timestamp = os.stat(lockfile_path).st_mtime
+    return _parse_lockfile_path_helper(lockfile_path, file_changed_timestamp)
+
+
+@lru_cache(maxsize=1000)
+def _parse_lockfile_path_helper(
+    lockfile_path: Path, file_changed_timestamp: float
```
adamberman, line 135: maybe a dumb question: does file_changed_timestamp get used anywhere? or is it just that `@lru_cache` sees that the param changes and doesn't return the cached value?

brendongo: not used anywhere. the cache key are all arguments to a function + function name

brendongo:
Slowdown and linting error seem unrelated to PR and have appeared in other merged PRs

adamberman (approved):

PR #8260, opened 2023-07-12
implement pcre_names_stub

tpetr:
semgrep.js was failing to evaluate rules that contained regexes because we started doing stuff with capture group names at some point. This PR wires up that missing piece

test plan:
- add test for regex
- checks and tests pass

/cc @LewisArdern

### DIFF
--- a/js/libpcre/runtime.js
+++ b/js/libpcre/runtime.js
@@ -2,35 +2,35 @@
 var libpcre = globalThis.LibPcreModule;
 
 //Provides: NULL const
-var NULL = 0;
+const NULL = 0;
 
 //Provides: PCRE_CONFIG_UTF8 const
-var PCRE_CONFIG_UTF8 = 0;
+const PCRE_CONFIG_UTF8 = 0;
 //Provides: PCRE_CONFIG_NEWLINE const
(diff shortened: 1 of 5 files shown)


js/libpcre/runtime.js
```diff
@@ -372,6 +381,57 @@ function pcre_exec_stub_bc(
...
+  for (var i = 0; i < name_count; i++) {
+    result[i] = libpcre.UTF8ToString(tbl_ptr + 2);
+    tbl_ptr += entry_size;
+  }
+
+  libpcre._free(name_count_ptr);
+  libpcre._free(entry_size_ptr);
+  libpcre._free(tbl_ptr_ptr);
+
+  return caml_js_to_array(result);
```
ajbt200128, line 442: Unrelated to review: These 5 lines cause me so much pain. Perl common regular expressions. Pointers. Memory freeing. OCaml to JS. All in a javascript file. Compiled to WASM. Pain

ajbt200128 (approved):
LGTM.

js/libpcre/runtime.js
tpetr: This is why they pay us the big bucks 💸

PR #8266, opened 2023-07-13
feat(osemgrep): Add support for pro engine in osemgrep via hook

emjin:
Osemgrep currently only allows users to run with the OSS engine. This PR adds in the `pro`, `pro-intrafile`, `pro-languages`, and `oss` flags and adjusts behavior depending on the flag. The first three flags only work when run with the binary built with semgrep-proprietary.

Test plan: see PR in semgrep-proprietary

PR checklist:

### DIFF
--- a/src/osemgrep/cli_scan/Core_runner.ml
+++ b/src/osemgrep/cli_scan/Core_runner.ml
@@ -53,7 +53,8 @@ type result = {
   *)
 }
 
-(* function *)
+(* Type for the core runner function, which can either be invoked by
+   invoke_semgrep_core or invoke_semgrep_core_proprietary *)
 
 type semgrep_core_runner =
   ?respect_git_ignore:bool ->
--- a/src/osemgrep/cli_scan/Core_runner.ml
+++ b/src/osemgrep/cli_scan/Core_runner.ml
@@ -53,6 +53,19 @@ type result = {
   *)
 }
 
+(* function *)
+
+type semgrep_core_runner =
+  ?respect_git_ignore:bool ->
+  ?file_match_results_hook:
+    (Fpath.t -> Report.partial_profiling Report.match_result -> unit) option ->
+  conf ->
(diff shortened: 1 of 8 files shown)


src/osemgrep/cli_scan/Core_runner.ml
```diff
@@ -53,6 +53,19 @@ type result = {
   *)
 }
 
+(* function *)
+
+type semgrep_core_runner =
```
aryx, line 59: maybe a comment explaining we introduced a type because it can be either the OSS semgrep-core (Run_semgrep.semgrep_with_rules_xxx)
or the semgrep-pro function (Run.deep_with_rules)

src/osemgrep/configuring/Engine_type.ml
```diff
@@ -0,0 +1,5 @@
+module Out = Semgrep_output_v1_t
+
+(* TODO it would be better for this also to be in the atd *)
```
aryx, line 3: would be better also to have
type engine_kind =
 | Oss
 | Pro of pro_flavor
instead of this ugly pair.

src/runner/Runner_config.ml
```diff
@@ -74,6 +74,9 @@ type t = {
 }
 [@@deriving show]
 
+type semgrep_engine =
```
aryx, line 79: maybe a comment explaining here this can be OSS or Pro and
maybe a short explanation of the return type (like what is the last Fpath.t list. I know it's the list
of processed target, but because the type is now far away from Run_semgrep.semgrep_with_rules_and_exn... it's better to repeat what it is here.

aryx (approved):
Let's do it!

src/osemgrep/configuring/Engine_type.ml
emjin: Good catch; I did this for a reason that no longer applies

PR #8278, opened 2023-07-15
Adopt ephemeral Cirrus Runners for M1 workflows

cgdolan:
M1 workflows currently execute on self-hosted Mac instances that run the GitHub Actions Runner software. A known limitation of this approach is the lack of isolation between workflow runs. This means state introduced in one workflow run (for example, installing packages) can carry over into the next workflow run leading to confusing and non-deterministic behavior.

Introduce isolation between M1 workflow runs by adopting ephemeral self-hosted Mac instances from Cirrus Runners. These instances are special-purpose VMs designed to drop into existing GitHub Actions that get created on-demand and torn down after a single use. The isolation allows us to share a single pool of M1 instances across all our organization's repos for better latency and throughput as compared to the existing approach where we dedicate 1-2 instances per repo.

PR checklist:


spencerdrak (approved):

.github/workflows/build-test-osx-m1.yaml
```diff
@@ -7,41 +7,60 @@ name: build-test-osx-m1
...
-    # we might as well move out of a self-hosted runner for M1 too.
-    # This would remove the need for those ugly hard cleanup below because
-    # our self-hosted runner builds are not "hermetic".
-    runs-on: [self-hosted, macOS, ARM64, aws]
+    runs-on:
+      [
+        "self-hosted",
+        "macOS",
+        "ARM64",
+        "ghcr.io/cirruslabs/macos-monterey-xcode:latest",
```
aryx, line 32: would be nice to add a link explaining what those cirrus runner are. I have no idea what this link is about.

aryx: What is Cirrus CI? What's the difference with GHA? Please add comments.

.github/workflows/build-test-osx-m1.yaml
```diff
@@ -7,41 +7,60 @@ name: build-test-osx-m1
...
     env:
       OPAM_SWITCH_NAME: "4.14.0"
     steps:
-      # Force hard cleanup of any old checkouts. This is needed because we run on
-      # self-hosted CI runners here and when submodules move around they are no longer
-      # removed by default.
-      # This hack misses some files (like hidden files) but should take care of the
-      # issues we've been seeing.
-      - name: Clean Checkout
+      - name: Setup runner directory
```
aryx, line 37: Can you add a comment explaining why we now need this extra setup? Is it because we use this Cirrus thing?
Is there a link to a doc explaining the need for this setup?

.github/workflows/build-test-osx-m1.yaml
```diff
@@ -7,41 +7,60 @@ name: build-test-osx-m1
...
         run: |
-          rm -rf ./*
-          rm -f .gitmodules
-          rm -rf .git
-      - name: Make checkout speedy
-        run: git config --global fetch.parallel 50
+          sudo mkdir /Users/runner
+          sudo chown admin:staff /Users/runner
+          sudo chmod 750 /Users/runner
+      - uses: actions/setup-python@v4
```
aryx, line 42: why do we need python now here?

.github/workflows/build-test-osx-m1.yaml
```diff
@@ -52,35 +71,60 @@ jobs:
...
       - uses: actions/download-artifact@v3
         with:
           name: semgrep-m1-${{ github.sha }}
-      - run: |
+      - name: Build Wheels
+        run: |
           unzip artifacts.zip
           cp artifacts/semgrep-core cli/src/semgrep/bin
-          ./scripts/build-wheels.sh
+          ./scripts/build-wheels.sh --plat-name macosx_11_0_arm64
```
aryx, line 101: again a link explaining what is this magic macosx_ string would be useful.
Is it something related to Cirrus? To Setuptools called in build-wheels.sh?

PR #8280, opened 2023-07-15
tainting: Deduplicate label preconditions in taint sets

IagoAbal:
If we have two taint labels that are basically the same, but that have different "incoming" taints in their precondition, we can just merge their taints. This removes a lot of redundancy in taint sets. In addition, we also remove duplicates from the "incoming" taints.

test plan:
Run the following with and without this change, while inspecting the findings inferred for each function.
```
semgrep --pro \
  -c r/java.lang.security.audit.formatted-sql-string.formatted-sql-string \
  examples-annotations/repos/java/SQLi/cve-repos/dotCMS
  # Check for example the findings of `getUniqueFieldErrorMessage` that
  # returns a taint set that is noticeably smaller after this change.
```

PR checklist:

### DIFF
--- a/src/tainting/Taint.ml
+++ b/src/tainting/Taint.ml
@@ -139,31 +139,33 @@ type source = {
 and orig = Src of source | Arg of arg [@@deriving show]
 and taint = { orig : orig; tokens : tainted_tokens } [@@deriving show]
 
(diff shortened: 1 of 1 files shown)


emjin (approved):
lgtm; this reminds me I need to go back to my PR to remove Stdlib.compare

PR #8281, opened 2023-07-15
feat(promql): add initial translations

MichaHoffmann:
PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [ ] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

Depends on: https://github.com/returntocorp/semgrep-interfaces/pull/93 after which the submodule needs to be updated

PR #8291, opened 2023-07-17
add buildx-fallback to build-test-manylinux-aarch64

tpetr:
stopgap to make fork PRs succeed until we come up with a better plan

test plan:
- checks pass

/cc https://github.com/returntocorp/semgrep/pull/8281

### DIFF
--- a/.github/workflows/build-test-manylinux-aarch64.yaml
+++ b/.github/workflows/build-test-manylinux-aarch64.yaml
@@ -13,6 +13,7 @@ jobs:
       id-token: write
     steps:
       - uses: docker/setup-qemu-action@v2
+      - uses: docker/setup-buildx-action@v2
       - uses: depot/setup-action@v1
       - name: Build and test python wheel
         id: build-semgrep-wheel
@@ -22,6 +23,7 @@ jobs:
           platforms: linux/arm64
           outputs: type=docker,dest=/tmp/image.tar
           target: semgrep-wheel
+          buildx-fallback: true
       - name: Extract wheel from docker image
         run: |
           docker load --input /tmp/image.tar


cgdolan (approved):

.github/workflows/build-test-manylinux-aarch64.yaml
```diff
@@ -13,6 +13,7 @@ jobs:
       id-token: write
     steps:
       - uses: docker/setup-qemu-action@v2
+      - uses: docker/setup-buildx-action@v2
```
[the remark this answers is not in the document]
tpetr: Thanks Semgrep

.github/workflows/build-test-manylinux-aarch64.yaml
```diff
@@ -22,6 +23,7 @@ jobs:
           platforms: linux/arm64
           outputs: type=docker,dest=/tmp/image.tar
           target: semgrep-wheel
+          buildx-fallback: true
```
aryx, line 26: please add a comment, saying we need this for forked PRs.
Otherwise there's no way we gonna remember it in 6 months.

cgdolan: The context about why this change was introduced is still captured in the commit message for future archaeologists to discover with `git blame`.

aryx: No! We don't want to force people to use git blame to understand why the code is written the way it is. The code should be readable and understandable without having to do some archeology!

aryx: We try hard to follow this rule in the semgrep code, so we should do the same for the GHA workflows.

tpetr: I didn't articulate this well in the PR body, but the core issue here is that I intended to (but forgot) to set `buildx-fallback: true` when I created this workflow. This wasn't a new learning to fix a specific problem (depot could fail for any number of reasons), so I'm more inclined to direct folks to the docs instead of adding a one-off comment: https://github.com/depot/build-push-action#inputs

aryx: FYI I've added some notes about comments vs git blame in https://www.notion.so/semgrep/How-we-code-together-ea3b47c3f3f04e57bd9f6167caff1335#1a28f23f4dd74c2488b1517bbc640e11

PR #8293, opened 2023-07-17
Dockerfile: add comments

aryx:
test plan:
make build-docker

PR checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@
 # which can be statically linked.
 #
 # Then 'semgrep-core' alone is copied to another Alpine-based container
-# which takes care of the 'semgrep-cli' Python wrapping.
+# which takes care of the 'semgrep-cli' (a.k.a. pysemgrep) Python wrapping.
 #
 # We use Alpine because it allows to generate the smallest Docker images.
 # We use this two-steps process because *building* semgrep-core itself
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -32,6 +32,9 @@ WORKDIR /src/semgrep
 COPY . .
 
 # remove files and folders that aren't necessary for the semgrep-core build
+# coupling: see the (dirs ...) directive in the toplevel dune file for the list
+# of directories containing OCaml code and which should not be added below
+# (except js/ which contains OCaml code but is not used to build semgrep-core)
```
tpetr, line 37: This idea is probably too complex, but I wonder if we could extract those dirs from the dune file to ensure that we're including the necessary folders?

Dockerfile
```diff
@@ -60,6 +63,10 @@ COPY cli/src/semgrep/semgrep_interfaces cli/src/semgrep/semgrep_interfaces
 #    is not without problems when used inside Github actions (GHA) or even inside
 #    this Dockerfile.
 #
+#    update: we recently started to cache the ~/.opam/ directory in CI so
+#    in theory we could get rid of ocaml-layer and instead use the official
+#    opam docker image combined with this ~/.opam/ caching to speedup things.
```
tpetr, line 68: I don't think we can depend on the CI cache for arm64 builds

Dockerfile
```diff
@@ -68,6 +75,8 @@ COPY cli/src/semgrep/semgrep_interfaces cli/src/semgrep/semgrep_interfaces
 #    tools like gcc, make, which are not provided by default on Alpine.
 #
 # An alternative to ocaml-layer would be to use https://depot.dev/
+# update: we actually started to use depot.dev to speedup multi-arch (arm)
+# docker image, so maybe we could use it to get rid of ocaml-layer
```
tpetr, line 79: we definitely could but would need to think a little more about opam lockfiles (or something equivalent) so that we're not over- or under-caching

tpetr (approved):
LGTM overall, thanks

PR #8294, opened 2023-07-17
fix: [ci] release workflow non-functional

spencerdrak:
Release workflow was misconfigured and had been failing since Friday - we did not get any notifications about this because the workflow file was invalid and could not run (therefore, it couldn't send us the slack message about it's failure). In addition, we've changed the name of the OSX artifacts which would've broken, had this workflow been able to run successfully.

This PR gets release back on the rails by:
- allowing jobs that build wheels to pass an artifact name as output, to standardize usage
- Adding the required missing job for aarch64 linux wheels. 

Test Plan:
- release dry-run here: https://github.com/returntocorp/semgrep/actions/runs/5582933738

I'll re-enable release workflow once this goes in.

@cgdolan it looks like a while back, you set up linting on PRs for GHA that I would've expected to catch this - I'm going to dig into that separately.

PR checklist:

### DIFF
--- a/.github/workflows/build-test-manylinux-aarch64.yaml
+++ b/.github/workflows/build-test-manylinux-aarch64.yaml
@@ -3,6 +3,13 @@ name: build-test-manylinux-aarch64
 on:
   workflow_dispatch:
   workflow_call:
+    outputs:
+      wheel-artifact-name:
+        description: "The name under which the wheel artifact was stored"
+        value: ${{ jobs.build-wheel.outputs.wheel-artifact-name }}
+
+env:
+  WHEEL_ARTIFACT_NAME: manylinux-aarch64-wheel
 
 jobs:
   build-wheel:
@@ -11,6 +18,8 @@ jobs:
     permissions:
       contents: read
       id-token: write
+    outputs:
+      wheel-artifact-name: ${{ steps.wheel-artifact-name.outputs.wheel-artifact-name }}
     steps:
       - uses: docker/setup-qemu-action@v2
       - uses: docker/setup-buildx-action@v2
@@ -30,5 +39,9 @@ jobs:
           docker run --platform linux/arm64 --rm ${{ steps.build-semgrep-wheel.outputs.imageid }} zip -r - dist > /tmp/dist.zip
       - uses: actions/upload-artifact@v3
         with:
-          name: manylinux-wheel-aarch64
+          name: "${{ env.WHEEL_ARTIFACT_NAME }}"
           path: /tmp/dist.zip
+      - name: Set Wheel Name Output
+        id: wheel-artifact-name
+        run: |
+          echo "wheel-artifact-name=${WHEEL_ARTIFACT_NAME}" >> $GITHUB_OUTPUT
--- a/.github/workflows/build-test-manylinux-aarch64.yaml
+++ b/.github/workflows/build-test-manylinux-aarch64.yaml
@@ -3,13 +3,6 @@ name: build-test-manylinux-aarch64
 on:
   workflow_dispatch:
   workflow_call:
-    outputs:
-      wheel-artifact-name:
-        description: "The name under which the wheel artifact was stored"
-        value: ${{ jobs.build-wheel.outputs.wheel-artifact-name }}
-
-env:
-  WHEEL_ARTIFACT_NAME: manylinux-aarch64-wheel
 
 jobs:
   build-wheel:
@@ -18,8 +11,6 @@ jobs:
     permissions:
       contents: read
       id-token: write
-    outputs:
-      wheel-artifact-name: ${{ steps.wheel-artifact-name.outputs.wheel-artifact-name }}
     steps:
       - uses: docker/setup-qemu-action@v2
       - uses: docker/setup-buildx-action@v2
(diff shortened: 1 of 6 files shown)


cgdolan (approved):

tpetr (approved):
Thank you! Sorry for the headache here

.github/workflows/build-test-osx-arm64.yaml
```diff
@@ -13,13 +13,24 @@ on:
         required: true
         type: boolean
         default: true
+      wheel-artifact-name:
+        description: "The name under which to store the wheel artifact"
+        type: string
+        default: osx-arm64-wheel
```
aryx: Is there anywhere where we set the wheel-artifact-name for this workflow to something other than osx-arm64-wheel?
If not, then what's the point of introducing this extra intermediate constant that looks like a variable but is not?

spencerdrak: In release.yml, there is one other place this is referenced in the `upload-wheels` job. Unfortunately, we can't set an env var in release.yml to hold this value, but this will likely be helpful with some follow up work I have on deck to factorize the OSX workflows.

aryx: if it's referenced but never set to a different value, it means it's a constant and we should just inline its content instead of introducing those intermediate name that complicate things IMHO.

spencerdrak: Ok, I can remove that bit then. thanks!

spencerdrak: @aryx swapped that around to make it such that the wheel artifact names are outputs which can be defined once and referenced, not inputs - lmk if that's acceptable. thanks!

.github/workflows/build-test-manylinux-aarch64.yaml
```diff
@@ -3,6 +3,13 @@ name: build-test-manylinux-aarch64
 on:
   workflow_dispatch:
   workflow_call:
+    outputs:
```
aryx: Honestly the way we do things in Yaml looks completely foreign to me. If we were in a sane programming language,
we could just return values, takes parameters, define constant in a shared file, and we would not need all those weird ways to do very simple things.

.github/workflows/release.yml
```diff
@@ -138,9 +134,21 @@ jobs:
         build-test-osx-x86,
         build-test-osx-arm64,
       ]
+    outputs:
+      manylinux-x86-wheel-artifact-name: ${{ steps.wait-for-build-test.outputs.manylinux-x86-wheel-artifact-name }}
```
aryx: This solution looks even worse to me. But again why we introduce those intermediate step variables in the first place?

.github/workflows/release.yml
```diff
@@ -159,33 +167,33 @@ jobs:
       - name: Download Artifact
         uses: actions/download-artifact@v3
         with:
-          name: manylinux-wheel
-          path: manylinux-wheel
+          name: ${{ needs.wait-for-build-test.outputs.manylinux-x86-wheel-artifact-name }}
```
aryx: Why not simply replace this name: ${{ ...} with the actual value we know it contains? It's a constant!
What's the gain of using this very long variable name, than even contain this ...manylinux-x86-wheel.... in it
with simply manylinux-x86-wheel!
It's a shorter, and it removes the need for those ENV:, for those needs.wait.xxx everywhere, for those echo ... >> $GITHUB_OUTPUT.

aryx (dismissed):
Looks too complicated. Just use directly 'manylinux-x86-wheel' where it's needed.

cgdolan:
> Looks too complicated. Just use directly 'manylinux-x86-wheel' where it's needed.

Directly using these string values in multiple places is the problem we're trying to solve in this PR. We realized that the next release would likely fail after the changes in https://github.com/returntocorp/semgrep/pull/8286 because the wheel producer step outputs an artifact under a different name than the wheel consumer step expects. The fix for this is to declare the output name once and reference it in such a way that if the value changes all callers get the updated value, which is what we've implemented here.

cgdolan (approved):

aryx:
And how using steps.xxx.whateverlong.manywheel-x86-arm variable names will be better than using the manywheel-x86-arm string?
What do you win? If you mispelled something, the string or the variable name you will get a runtime error in both cases, so I'd rather use the simpler code.

cgdolan:
With variables, invalid references are surfaced much earlier before/at PR time by IDEs and workflow linting whereas with raw strings invalid references are surfaced much later at workflow runtime.

Here's what I see when I introduce a typo for example:

Screenshot 2023-07-18 at 8 30 07 AM

aryx:
Does the linting work across files? If you reference an output from a workflow in another yaml file is it detected?
If you make a typo in echo "manylinux-x86-wheel-artifact-name
is it detected?

spencerdrak:
Kicked off a release dry-run with these new changes here: https://github.com/returntocorp/semgrep/actions/runs/5589925058

cgdolan:
In the interest of getting this PR submitted and unblocking the release process, I've downscoped the changes to a minimal subset needed to fix the blockers. @aryx PTAL.

tpetr (approved):
LGTM assuming the dry run succeeds

cgdolan:
Dry run has succeeded and I've implemented the initial feedback as asked. I'm going to dismiss the review requesting changes so we can get this in and unblock releases.

aryx (commented):
Thx chris. Appreciate.

PR #8297, opened 2023-07-18
restructuring jsonnet to carry around an environment rather than using explicit lazy

SophiaSR:
This is a restructuring of the types in jsonnet. 
The key points are that now instead of an explicit "lazy" wrapping of objects and arrays, we now carry around the environment as well. this is needed since we need to be able to make modifications to the environment in which object fields are evaluated based on which object they are part of. Another part of this restructuring includes delaying the setting of self/super until the latest possible point, thus being able to access the correct fields when objects are merged. 
PR checklist:


libs/ojsonnet/Core_jsonnet.ml
```diff
@@ -161,11 +161,65 @@ and for_comp = tok (* 'for' *) * ident * tok (* 'in' *) * expr
 [@@deriving show { with_path = false }]
 
 (*****************************************************************************)
-(* Program *)
+(* Env *)
 (*****************************************************************************)
+type env = {
```
aryx: Why move those types to Core_jsonnet.ml? We could keep Value_jsonnet.ml and move the environment that was in Eval_jsonnet.ml in Value_jsonnet.ml.

I try hard to "separate concerns" and so define data-structures in different files if they are quite independent.
For instance we have some phases like Desugar_jsonnet.ml that can produce Core_jsonnet.program and this phase does not have to know about environment and values.

libs/ojsonnet/Core_jsonnet.ml
```diff
@@ -161,11 +161,65 @@ and for_comp = tok (* 'for' *) * ident * tok (* 'in' *) * expr
...
+   * evaluation so we model this by allowing unevaluated expressions in environment below.
+   *)
+  locals : (local_id, local) Map_.t;
+  (* for call tracing *)
+  depth : int;
+}
 
-type program = expr [@@deriving show]
+and local_id = LSelf | LSuper | LId of string
+and local = Val of (value_ * env) | Expr of (expr * env)
```
aryx: you could rename this type to lazy_expr or even better lazy_value and add a comment about the environment, that it's used to be simply value_ Lazy.t but that we switch to explicit env because of + and the need to access the environment etc. etc. (and that with lazy/closure there was no way we could do that).

libs/ojsonnet/Core_jsonnet.ml
```diff
@@ -161,11 +161,65 @@ and for_comp = tok (* 'for' *) * ident * tok (* 'in' *) * expr
...
+  | Str of string AST_jsonnet.wrap
+
+and object_ = asserts list * value_field list
+
+(* opti? make it a hashtbl of string -> field for faster lookup? *)
+and value_field = {
+  (* like Str, strictly evaluated! *)
+  vfld_name : string AST_jsonnet.wrap;
+  vfld_hidden : AST_jsonnet.hidden AST_jsonnet.wrap;
+  vfld_value : expr * env;
```
aryx: Here it should be lazy_value I think, like for locals, so accessing a field is evaluated only once.

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -31,18 +30,19 @@ let logger = Logging.get_logger [ __MODULE__ ]
...
-   * environment where the locals are defined. Jsonnet uses lazy
-   * evaluation so we model this by using Lazy below.
-   *)
-  locals : (local_id, V.value_ Lazy.t) Map_.t;
-  (* for call tracing *)
-  depth : int;
-}
 
-and local_id = LSelf | LSuper | LId of string
+(*type env = {
```
aryx: This comment should be removed (and the environment type moved to Value_jsonnet.ml instead).

libs/ojsonnet/Core_jsonnet.ml
aryx: It's nice to document design decisions and document design changes in comments, so one can remember why we did the things the way we did it and how we arrived to the final design. Otherwise in 6 months one can look at the code and ask "but why the hell they didn't simply use a value_ lazy.t instead of passing around this environment"

libs/ojsonnet/Core_jsonnet.ml
```diff
@@ -161,11 +161,65 @@ and for_comp = tok (* 'for' *) * ident * tok (* 'in' *) * expr
...
+and local = Val of (value_ * env) | Expr of (expr * env)
+
+(*****************************************************************************)
+(* Values *)
+(*****************************************************************************)
+and value_ =
+  | Primitive of primitive
+  | ObjectVal of object_ AST_jsonnet.bracket
+  | Function of function_definition
+  | Array of (expr * env) array AST_jsonnet.bracket
```
aryx: same here, it should be lazy_value array AST_jsonnet.bracket, otherwise
we would reevaluate each time array elements.

libs/ojsonnet/Core_jsonnet.ml
aryx: Maybe it could be  Val of ... | Unevaluated of expr * env
and also because env needs to be present in both case, we could use a record instead and
document why we also need env even for the Val case (which is subtle, but useful because of nested
array elts or fields, and again this is the kind of things that is nice to document in comments).

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -173,70 +165,80 @@ let rec eval_expr (env : env) (v : expr) : V.value_ =
...
           match
             fields
-            |> List.find_opt (fun (field : V.field) -> fst field.fld_name = fld)
+            |> List.find_opt (fun (field : value_field) ->
+                   fst field.vfld_name = fld)
           with
           | None -> error tk (spf "field '%s' not present in %s" fld (sv e))
-          | Some fld -> Lazy.force fld.fld_value.v)
+          | Some fld ->
+              let new_self = ObjectVal (_l, (_assertsTODO, fields), _r) in
```
aryx: This could be a in a separate PR. This is unrelated to the refactoring.

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -654,9 +627,60 @@ and eval_for_comp env v =
...
+(*Same as eval_expr but with profiling *)
+and eval_program (e : Core_jsonnet.program) (env : env) : value_ =
+  let v = eval_expr env e in
   v
   [@@profiling]
+
+(*****************************************************************************)
+(* Manfestation *)
+(*****************************************************************************)
+and manifest_value (v : value_) : JSON.t =
```
aryx: Again if possible if would be better to have this in a separate Manifest_jsonnet.ml file to separate concerns. Sometimes we can't because things are mutually recursive, but if it's not it's better to move in another file (mutual recursivity can also be broken by passing the mutual recursive function as a parameter also).

aryx (changes requested):

libs/ojsonnet/Eval_jsonnet.ml
SophiaSR: manifest and eval are unfortunately mutually recursive. The to_string coercion for adding strings to things that are not strings calls manifest which in turn now calls eval_expr

libs/ojsonnet/Core_jsonnet.ml
SophiaSR: I think the reason I moved these originally, is I tried to keep environment in a separate file, which then lead to some mutual recursion issues, so I ended up moving everything into 1. But, yes! Moving env into Value_jsonnet works, will fix  that.

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -15,8 +15,8 @@
 open Common
 open Core_jsonnet
 module A = AST_jsonnet
-module V = Value_jsonnet
 module J = JSON
+open Value_jsonnet
```
aryx: Please put back the module alias instead of open. It's not documented anywhere I think but we usually try to avoid to use open Xxx because it makes it harder when reading the code to see where functions or types are coming from. It's usually
ok to open Common (because we're supposed to know the 50 basic functions that are in there), and to also open the file defining the main data structure used in a file (e.g., in Naming_AST.ml we do 'open AST_generic' because we know that's the only data structure used).
But here, we manipulate 2 data structures, which actually have sometimes the same name, so it can be confusing to know what is what, so better to qualify with V.xx the produced value.

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -102,63 +76,73 @@ let log_call env str tk =
...
-  | V.Object _ -> "object"
-  | V.Array _ -> "array"
-  | V.Function _ -> "function"
-
-let std_primivite_equals _env (v : V.value_) (v' : V.value_) : bool =
+  | Primitive (Null _) -> "null"
+  | Primitive (Bool _) -> "boolean"
+  | Primitive (Double _) -> "number"
+  | Primitive (Str _) -> "string"
+  | ObjectVal _ -> "object"
```
aryx: See, because of your open you had to rename some constructors, but if instead you qualify, there's no need to rename

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -102,63 +76,73 @@ let log_call env str tk =
...
-               let v = lazy (eval_expr env e) in
-               { V.v; e })
+        xs
+        |> Common.map (fun x -> { value = Unevaluated x; env })
+        |> Array.of_list
       in
-      V.Array (l, elts, r)
-  | Lambda v -> V.Function v
+      Array (l, elts, r)
+  | Lambda v -> Function v
```
aryx: Again, better keep Lambda, so it mimics more the parallel between Core_jsonnet and Value_jsonnet.

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -551,27 +539,33 @@ and eval_binary_op env el (op, tk) er =
...
-    | V.Array (_, _, _), V.Array (_, [||], _) -> Sup
-    | V.Array (al, ax, ar), V.Array (bl, bx, br) -> (
-        let a0 = Lazy.force ax.(0).v in
-        let b0 = Lazy.force bx.(0).v in
+    | Array (_, [||], _), Array (_, [||], _) -> Eq
+    | Array (_, [||], _), Array (_, _, _) -> Inf
+    | Array (_, _, _), Array (_, [||], _) -> Sup
+    | Array (al, ax, ar), Array (bl, bx, br) -> (
+        let a0 =
+          match ax.(0).value with
```
aryx: You can probably factorize all those match Val | Unevaluated in a separate function like
eval_lazy_value so you don't have to repeat yourself all the time.

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -586,53 +580,41 @@ and eval_std_cmp env tk (el : expr) (er : expr) : cmp =
 (* eval_obj_inside *)
 (*****************************************************************************)
 
-and eval_obj_inside env (l, x, r) : V.value_ =
+and eval_obj_inside env (l, x, r) : value_ =
+  let hdupes = Hashtbl.create 16 in
```
aryx: you should move back hdupes under the match Object since we don't use it in the other case?

libs/ojsonnet/Manifest_jsonnet.mli
```diff
@@ -1,7 +0,0 @@
-exception Error of string * Tok.t
-
-(* TODO: at some point we probably would prefer an AST_generic.program
- * which could track the origin of tokens through import, eval, and
- * manifestation and that we could pass to osemgrep to evaluate the rules.
- *)
-val manifest_value : Value_jsonnet.value_ -> JSON.t
```
aryx, line 7: We could keep Manifest_jsonnet.ml, but just call the one in Eval_jsonnet.ml, so at least we mimic more the jsonnet spec in the way the files are organized.

libs/ojsonnet/Value_jsonnet.ml
```diff
@@ -1,74 +1,68 @@
-(* Yoann Padioleau
```
aryx, line 1: Why removing this comment?
Please put it back, and you can add your name after mine at the top
(and change r2c in Semgrep Inc. instead)

libs/ojsonnet/Value_jsonnet.ml
```diff
@@ -1,74 +1,68 @@
...
-module A = AST_jsonnet
-module C = Core_jsonnet
-
 (*****************************************************************************)
-(* Prelude *)
+(* Env *)
 (*****************************************************************************)
-(* Jsonnet "values".
- *
- * See https://jsonnet.org/ref/spec.html#jsonnet_values
```
aryx, line 23: please restore this toplevel comment too.
Each of our file has this Copyright then Prelude comments, so good to be consistent.

libs/ojsonnet/Value_jsonnet.ml
```diff
@@ -1,74 +1,68 @@
...
- * modify it under the terms of the GNU Lesser General Public License
- * version 2.1 as published by the Free Software Foundation, with the
- * special exception on linking described in file LICENSE.
- *
- * This library is distributed in the hope that it will be useful, but
- * WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the file
- * LICENSE for more details.
- *)
-module A = AST_jsonnet
```
aryx, line 15: You can also restore those module aliases to keep the code shorter.

src/osemgrep/networking/Rule_fetching.ml
```diff
@@ -266,7 +266,7 @@ let parse_rule ~registry_caching (file : Fpath.t) :
           ~import_callback:(import_callback ~registry_caching)
           file ast
       in
-      let value_ = Eval_jsonnet.eval_program core in
+      let value_ = Eval_jsonnet.eval_program core Value_jsonnet.empty_env in
```
aryx: Maybe we should define an Eval_jsonnet.eval_program_with_env and an Eval_jsonnet.eval_program
that calls the other one with an empty env, so we don't have to repeat ourselves everywhere by passing
this empty env. This also hide implementation details (the fact that we refactored the code to use an env instead of lazy value) and don't push this detail to the user of the API (who should not care about those details).

src/parsing/Manifest_jsonnet_to_AST_generic.ml
```diff
@@ -12,9 +12,11 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the file
  * LICENSE for more details.
  *)
-module V = Value_jsonnet
+
 module G = AST_generic
 module A = AST_jsonnet
+module E = Eval_jsonnet
+open Value_jsonnet
```
aryx: again, please avoid using open and instead qualify. 
Here we manipulate 3 different complex data structures, so better qualify.

aryx (approved):
I approve but that does not mean you should not address the comments. It's just mean there's no need for another round of review; what remains is simple to fix and the rest looks fine so you should try your best to address the comments and then merge to move forward.

src/osemgrep/networking/Rule_fetching.ml
SophiaSR: Can do! We will still have to expose the one with environment in the mli though since Manifest_jsonnet_to_AST_generic has to call the one with a specified environment

PR #8300, opened 2023-07-18
cleanup: remove AST_generic.stmt s_use_cache not used anymore

aryx:
This was part of the stmts caching opti that Martin did a long
time ago that ended up not being worth the code complexity.
Iago recently remove the stmts caching code but we forgot
to remove certain fields and files.

test plan:
make core

PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #8302, opened 2023-07-18
cleanup: remove AST_generic.s_backrefs

aryx:
Followup to #8300 to remove cruft related to
the stmts caching

test plan:
make core

PR checklist:

### DIFF
--- a/libs/ast_generic/AST_generic.ml
+++ b/libs/ast_generic/AST_generic.ml
@@ -418,7 +418,6 @@ class virtual ['self] iter_parent =
     method visit_resolved_name _env _ = ()
     method visit_tok _env _ = ()
     method visit_node_id_t _env _ = ()
-    method visit_string_set_t _env _ = ()
   end
 
 (* Basically a copy paste of iter_parent above, but with different return types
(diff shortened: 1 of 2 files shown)


libs/ast_generic/AST_generic.ml
```diff
@@ -484,7 +483,6 @@ class virtual ['self] map_parent =
     method visit_resolved_name _env x = x
     method visit_tok _env x = x
     method visit_node_id_t _env x = x
-    method visit_string_set_t _env x = x
```
aryx, line 487: @nmote FYI I had a very cryptic error message by ocamlc with an error when instantiating a visitor.
Once I removed s_backrefs below, this string set was not used, so the visit_string_set_t suddently contains
some free variables and got an error like
"lazy_t contains free variables, impossible to instantiate".
I got lucky I found it was because of those visit_string_set_t methods ...

nmote: Maybe all of these should have type annotations?

aryx: yes, so that when we remove the type we get a clear typing error instead of those 'free type variable ...' for a different type.

PR #8312, opened 2023-07-18
refactor(tainting): Add a proper constructor for `this'

IagoAbal:
We were representing `this` as a special argument with name "this" at position -1...

test plan:
make test

PR checklist:

### DIFF
--- a/src/tainting/Taint.ml
+++ b/src/tainting/Taint.ml
@@ -108,16 +108,26 @@ let rec _show_call_trace show_thing = function
 (* Signatures *)
 (*****************************************************************************)
 
-type arg_pos = string * int [@@deriving show, compare]
-type arg = { pos : arg_pos; offset : IL.name list } [@@deriving show]
+type arg_pos = { name : string; index : int } [@@deriving show, compare]
+type arg_base = Bthis | Barg of arg_pos [@@deriving show, compare]
+type arg = { base : arg_base; offset : IL.name list } [@@deriving show]
 
-let _show_arg { pos = s, i; offset = os } =
+let _show_pos { name = s; index = i } = Printf.sprintf "arg(%s@%d)" s i
+
+let _show_base base =
+  match base with
+  | Bthis -> "this"
+  | Barg pos -> _show_pos pos
+
+let _show_arg { base; offset = os } =
+  _show_base base
+  ^
   if os <> [] then
     let os_str =
       os |> Common.map (fun n -> fst n.IL.ident) |> String.concat "."
     in
-    Printf.sprintf "arg(%s)#%d.%s" s i os_str
-  else Printf.sprintf "arg(%s)#%d" s i
+    "." ^ os_str
+  else ""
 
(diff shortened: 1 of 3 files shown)


src/tainting/Taint.mli
```diff
@@ -16,8 +16,9 @@ type 'a call_trace =
 type sink = { pm : Pattern_match.t; rule_sink : Rule.taint_sink }
 [@@deriving show]
 
-type arg_pos = string * int [@@deriving show]
-type arg = { pos : arg_pos; offset : IL.name list } [@@deriving show]
+type arg_pos = { name : string; index : int } [@@deriving show]
+type arg_base = Bthis | Barg of arg_pos [@@deriving show]
+type arg = { base : arg_base; offset : IL.name list } [@@deriving show]
```
aryx, line 37: This would benefit from some comments and concrete example.
The whole module could also have a toplevel comment explaining the whole approach (and for example how this approach makes our tainting context-sensitive in Semgrep-pro because of our use of signatures).

aryx: For example I don't really understand what offset is here. An example would really help.

src/tainting/Taint.ml
```diff
@@ -108,16 +108,26 @@ let rec _show_call_trace show_thing = function
...
-type arg_pos = string * int [@@deriving show, compare]
-type arg = { pos : arg_pos; offset : IL.name list } [@@deriving show]
+type arg_pos = { name : string; index : int } [@@deriving show, compare]
+type arg_base = Bthis | Barg of arg_pos [@@deriving show, compare]
+type arg = { base : arg_base; offset : IL.name list } [@@deriving show]
 
-let _show_arg { pos = s, i; offset = os } =
+let _show_pos { name = s; index = i } = Printf.sprintf "arg(%s@%d)" s i
+
+let _show_base base =
```
aryx, line 117: why the _prefix? And why not use deriving show instead?

aryx (approved):

src/tainting/Taint.ml
IagoAbal: There is a `deriving show` that is useful to display a small amount of info, but usually it's too verbose, so the `_show` functions are better to everyday's debugging. Happy to change the prefix to something else if there is a good suggestion.

aryx: maybe debug_ then? You can also annotate the types and adjust what deriving show is doing by default if the default is too verbose for you.

src/tainting/Taint.mli
IagoAbal: Sure thing, I'll revisit the existing comments and add some more.

emjin (approved):
Why the prefix with `B`? Otherwise, nice, much better 😂

IagoAbal:
> Why the prefix with `B`? Otherwise, nice, much better 😂

There is another constructor named `Arg`, if I end up renaming it then I could remove the `B` prefix. It just stands for "base".

PR #8319, opened 2023-07-19
delaying super/self assignment in jsonnet

SophiaSR:
To have self/super be correctly assigned during the plus operator, it is necessary to delay this assignment until the point where the fields are accessed (as per the spec). This PR does that, as well as moves the tests that now pass due to this change into the pass folder.

PR checklist:


libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -190,7 +190,16 @@ and eval_expr (env : V.env) (v : expr) : V.value_ =
                    fst field.fld_name = fld)
           with
           | None -> error tk (spf "field '%s' not present in %s" fld (sv e))
-          | Some fld -> evaluate_lazy_value_ fld.fld_value)
+          | Some fld -> (
+              let new_self = V.Object (_l, (_assertsTODO, fields), _r) in
```
aryx, line 194: maybe add the comment in the PR here also.

aryx: Also for people not familiar with the jsonnet spec, it might be useful to give a quick summary of the semantic of self/super here (ok maybe not super, but at least self).

aryx: I have a hard time for example understand why this code works better than what was before where we were also assigning LSelf to new_self.
Do you have an example showing why we need the code the way it's written and why before where we bind LSelf in
eval_obj_inside it could not work?

aryx: Then we could put this example in comment and show the trace of its evaluation and expose the subtelities.

aryx (approved):

libs/ojsonnet/Eval_jsonnet.ml
SophiaSR: Yep, can add. Here's a small specific example is the following (will also add it to code). 
`({ name : self.y } + {y : 3})["name"]`
If we do the binding to self in eval_obj_inside, we end up with self bound to 
`o = {name:self.y}`
Then evaluating the plus would give (semantically)
`({name : o.y, y : 3})` 
and when doing the access, we try to evaluate
` {name:self.y}.y` which doesn't exist. 
However, if we delay self/super binding til access (like in the spec) we end up with
`({name : self.y, y : 3})["name"]`
 then on access, self gets bound to
` o = {name : self.y, y:3}`
so we get` o.y` when we access name, and now are left with the correct final value of `3`.

aryx: great example!

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -670,6 +668,22 @@ and manifest_value (v : V.value_) : JSON.t =
                    let new_self = V.Object (_l, (_assertsTODO, fields), _r) in
                    let locals =
                      fld_value.env.locals
+                     (* We need to do these assignment on field access rather than on
```
aryx, line 671: I think this comment would be better in eval_expr in the field access code rather than here in manifest_value.

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -670,6 +668,22 @@ and manifest_value (v : V.value_) : JSON.t =
...
+                        {name : self.y}.y
+                        Error no such field.
+                        However, if we only assign self on access, we get the following (correct) evaluation
+                        ({ name : self.y } + {y : 42})["name"]
+                        { name : self.y, y : 42 }["name"]
+                        {name: self.y, y : 42}[y]
+                        42 *)
                      |> Map_.add V.LSelf
                           { V.value = Val new_self; env = fld_value.env }
                      |> Map_.add V.LSuper
```
aryx, line 689: Also is this correct? Seems like Super is always bound to an empty_obj. I know you have more PRs fixing + later, but maybe would be worth to add a TODO comment saying this handling of LSuper is incorrect and need to be fixed.

libs/ojsonnet/Eval_jsonnet.ml
aryx, line 194: Why we don't simply reuse 'e' instead of building this new_self? or just do
let new_self = e (or use 'as o' in the pattern to bind the value in the V.Object ...)

libs/ojsonnet/Eval_jsonnet.ml
```diff
@@ -190,7 +190,16 @@ and eval_expr (env : V.env) (v : expr) : V.value_ =
                    fst field.fld_name = fld)
           with
           | None -> error tk (spf "field '%s' not present in %s" fld (sv e))
-          | Some fld -> evaluate_lazy_value_ fld.fld_value)
+          | Some fld -> (
+              let new_self = V.Object (_l, (_assertsTODO, fields), _r) in
+              let locals =
```
aryx, line 195: you could move this code under the case for Unevaluated since it's relevant only there

libs/ojsonnet/Eval_jsonnet.ml
aryx: Actually I think I understand. This is correct code probably. By default super is the empty obj, 
it's just in + that we do something special and bind it to the left object.

PR #8321, opened 2023-07-19
fix: move Semgrep LS to OSemgrep

ajbt200128:
Warning: this looks like a big PR but git is lame and did not detect that a bunch of files were moved and slightly tweaked.

This PR moves the language server from semgrep-core to osemgrep, netting it a bunch of speed improvements, and overall cleaner code.

### Language Server

Primarily, on the python side, we've removed all code relating to pulling/parsing rules, finding targets, and scan configuration (i.e. timeout, max target size etc.) this has all been moved to either Session.ml, or UserSettings.ml.

Because osemgrep provides functionality for getting rules from remote sources (CI + registry), and targeting, we no longer have to compute this on the python side -> write a file -> parse it on the core side. This improves error messaging and handling, speed when scanning new files/folders, and general stability. Obviously, it decreases complexity by a ton.

OSemgrep also provides a much better interface for running scans, meaning we could simplify a lot of code in Diagnostics, post processing, and in the actual semgrep runner, as we are now using `cli_match`, which has the rule message w/ metavars rendered, the fix rendered, and some other things that before we had to compute ourselves. Also, it 
handles nosem/nosemgrep tags, which we had to do manually in the LS before.

Finally, the language server runs inside a LWT runtime (process?), i.e. inside `Lwt_main.run`, but so do the http helpers for CI/registry related things. Sadly, sanely, LWT does not allow nested calls to their runtime, which means that for some config/scan helpers, I had to create an async version, which is essentially the same as the original version, but without a final call to the LWT runtime. These methods have a suffix of _async, and the original methods just call these and pass them to the runtime to make them synchronous.  

### Cohttp

Outside of the Language Server changes listed above, we also switched our http client to cohttp. This is because the previous http client would randomly fail on DNS request timeouts. Also, cohttp provides jsoo bindings which will be useful in the future.

## Reorganization

For the most part, the organization is the same, except that we finally split out the JSONRPC logic of the language server from the semgrep/message handling. The RPCServer.ml file is existing code, just split out. This did not have to be a part of this PR, but I figured I'd do it while I'm already moving things around. The goal was to separate any generic JSONRPC/LSP logic, from Semgrep LS logic, resulting in cleaner code, that should be less confusing for future contributors.

## Testing

Most of the testing remains the same, and all passes, except testing related to CI/logging in. These tests previously relied on the fact that all CI rules were pulled on the python side, so we could easily mock them out. Now the CI logic is in the language server, and there's not much testing infra there, especially to mock this out. I decided to just remove these tests and test them manually. I'm also ok with this as we are now using logic from osemgrep to get rules, and those (should) be already tested, so I'm ok leaning on that too.

PR checklist:


cli/src/semgrep/app/session.py
```diff
@@ -185,6 +185,7 @@ def request(self, *args: Any, **kwargs: Any) -> requests.Response:
         method, url = args
         error_handler.push_request(method, url, **kwargs)
         try:
+            print(f"Sending request: {args} {kwargs}")
```
aryx: Do we still want that or was it used just for debugging?

cli/src/semgrep/lsp/server.py
```diff
@@ -29,66 +25,23 @@
 # core/osemgrep does't have the target manager yet, we need to do that stuff
```
aryx: you can remove this comment now

cli/src/semgrep/lsp/server.py
```diff
@@ -29,66 +25,23 @@
...
-            str(self.config.timeout),
-            "-timeout_threshold",
-            str(self.config.timeout_threshold),
-            "-fast",
-            "-ls",
-        ]
-        if self.config.debug:
-            cmd.append("-debug")
-
+        path = compute_executable_path("osemgrep")
```
aryx: osemgrep is actually not part of the distribution anymore; it has been merged with the semgrep-core binary.
We used to have a symlink from osemgrep to semgrep-core, and in that case when running the binary
will look at its argv0 and realize it was called as osemgrep and switch to osemgrep behavior, but we removed that links in the pip package for example (just to reduce the number of binaries in the PATH for the user).
So the way to call osemgrep is actually to just call 'semgrep' (should be in the PATH), but to force osemgrep you can use the --experimental flag

aryx: See ```
    path = find_semgrep_core_path()
    # If you call semgrep-core as osemgrep, then we get osemgrep behavior,
    # see src/main/Main.ml
    sys.argv[0] = "osemgrep"
    os.execvp(str(path), sys.argv)
``` 
in cli/bin/semgrep, the semgrep entry point.

File: src/osemgrep/language_server/CodeActions.mli [hunk 4]
```diff
@@ -0,0 +1,6 @@
+val code_actions_of_cli_matches :
+  Semgrep_output_v1_t.cli_match list ->
+  Fpath.t list ->
+  [> `CodeAction of Lsp.Types.CodeAction.t ] list
+(** [code_actions_of_cli_matches matches files] returns a list of code actions
```
aryx, line 5: This comment states the obvious given the name of the functions, so I wonder if there's a way to bring useful information instead in it, like a link to what is a code action, an example of match and the kind of structured returned by the function.

src/osemgrep/language_server/Diagnostics.mli
```diff
@@ -0,0 +1,6 @@
+val diagnostics_of_results :
+  Semgrep_output_v1_t.cli_match list ->
+  Fpath.t list ->
+  Lsp.Server_notification.t list
+(** [diagnostics_of_results results files] returns a list of LSP diagnostics
```
aryx, line 5: Again a concrete example would maybe help.

src/osemgrep/language_server/RPCServer.mli
```diff
@@ -0,0 +1,24 @@
+module State : sig
```
aryx, line 10: Maybe a toplevel comment would help understand the need for this module, or some history behind it, design decisions, whatever can help understand it.

src/osemgrep/language_server/UserSettings.mli
```diff
@@ -0,0 +1,17 @@
+type t = {
+  configuration : string list;
+  exclude : string list;
+  include_ : string list;
+  jobs : int;
```
aryx, line 5: Ideally we would like to reuse Core_runner.conf here.
Maybe there's some mutual directory dependency issue, but this can be fixed by moving Core_runner.conf in a separete file under src/osemgrep/core/Core_runner_conf.ml for example

aryx (approved):
Great work!

cli/src/semgrep/app/session.py
ajbt200128: For debugging, will remove it

cli/src/semgrep/lsp/server.py
ajbt200128: Ahhhh ok, I was a bit confused about how this worked, I'll clean this up then

src/osemgrep/language_server/CodeActions.mli
ajbt200128: I'll add examples and a link to the LSP docs!

src/osemgrep/language_server/Diagnostics.mli
ajbt200128: WIll add example + link to docs

src/osemgrep/language_server/UserSettings.mli
ajbt200128: Originally I was going to do this, but when we send the settings from the extension to the language server, the json is structured the same as in this file. It seemed easier to not change that structure at all (and risk breaking backwards compatibility), and this way we can use derive yojson and it's cleaner. That's also why I added the helper function to convert from a user settings struct to a core_runner.conf

ajbt200128:
Also note to self: write an architecture/contributing notion page after this PR

aryx:
What remains to merge this?

ajbt200128:
> What remains to merge this?
> 

Nothing, I thought I enabled auto merge but I did not :(

ajbt200128:
Oh NVM a test is still failing I have to fix

cli/src/semgrep/lsp/server.py
```diff
@@ -17,80 +13,39 @@
...
 from semgrep.lsp.config import LSPConfig
+from semgrep.semgrep_core import compute_executable_path
 from semgrep.state import get_state
 from semgrep.types import JsonObject
 
 log = logging.getLogger(__name__)
 
 # This is essentially a "middleware" between the client and the core LS. Since
-# core/osemgrep does't have the target manager yet, we need to do that stuff
+# core/osemgrep does't have login functionality, we need to do that stuff
```
aryx, line 24: 'osemgrep login' is supported now (but might not be fully tested true).

src/osemgrep/language_server/UserSettings.mli
aryx: I see. Then please add your design decision rational in a comment in the code, so one can understand why we don't reuse core_runner.conf

cli/src/semgrep/lsp/server.py
aryx: I think this is the only thing remaining to fix, to call semgrep --experimental instead of osemgrep (which does not exist in the latest distributions).
Otherwise I love the added comments in the .mli Very useful.

src/osemgrep/networking/Http_helpers.ml
```diff
@@ -1,52 +1,63 @@
-open Http_lwt_client
-
-(* The http-lwt-client package offers:
```
aryx, line 3: Would have been good in this file to add a toplevel comment about the design decision
to choose cohttp over other libraries, especially to explain that we got
lots of DNS timeout and even though this was partially fixed by Robur
(see the blog post about it), we still got many DNS issues and so decided to switch
back to cohttp unfortunately.

Also might be worth adding some comments about what happened with yosef and the TLS and
ocaml-tls vs openssh.

PR #8323, opened 2023-07-19
fix(SSC): Stop raising exceptions during supply chain parsing

mmcqd:
The original point of `safe_path_parse` was to prevent any exceptions from being raised during parsing, so that, for example, if we failed to parse a manifest file we could still parse the lockfile. The introduction of error tracking undid this behavior, so this PR re-introduces it, by having parsing functions return a list a errors that occurred during parsing.

I also added a `safe_parse_lockfile_and_manifest` function to abstract some of the repeated logic present across all the parsing functions.

PR checklist:


cli/src/semdep/parsers/util.py
```diff
@@ -331,13 +321,36 @@ def safe_path_parse(
...
-            raise DependencyParserError(path, parser_name, reason, line + 1, col + 1)
+            return DependencyParserError(
+                file_to_parse.path, file_to_parse.parser_name, reason, line + 1, col + 1
+            )
+
+
+def safe_parse_lockfile_and_manifest(
+    lockfile_to_parse: DependencyFileToParse[A],
+    manifest_to_parse: Optional[DependencyFileToParse[B]],
+) -> Tuple[Optional[A], Optional[B], List[DependencyParserError]]:
```
brendongo, line 343: needs function level comment

brendongo (approved):

PR #8334, opened 2023-07-20
tainting: Add experimental control sources

IagoAbal:
These sources taint the "control" (the program counter, sort to speak) rather than the data. They allow writing reachability queries, do we reach B from A ?

test plan:
make test # new test

PR checklist:

### DIFF
--- a/changelog.d/pa-2958.added
+++ b/changelog.d/pa-2958.added
@@ -1,0 +1,22 @@
+taint-mode: Added **experimental** `control: true` option to `pattern-sources`,
+e.g.:
+
+```yaml
+    pattern-sources:
+      - control: true
+        pattern: source(...)
+```
+
+Such sources taint the "control flow" (or the program counter) so that it is
+possible to implement reachability queries that do not require the flow of any
+data. Thus, Semgrep reports a finding in the code below, because after `source()`
+the flow of control will reach `sink()`, even if no data is flowing between both:
+
+```python
+def test():
+  source()
+  foo()
+  bar()
+  #ruleid: test
+  sink()
+```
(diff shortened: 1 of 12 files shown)


changelog.d/pa-2958.added
```diff
@@ -0,0 +1,22 @@
+taint-mode: Added **experimental** `control: true` option to `pattern-sources`,
+e.g.:
+
+```yaml
+    pattern-sources:
+      - control: true
+        pattern: source(...)
+```
+
+Such sources taint the "control flow" (or the program counter) so that it is
```
aryx, line 10: Nice, but who asked for this? SR?

IagoAbal: Three or four customers!

IagoAbal: And I heard that SSC could find it useful too, but I haven't checked with them.

src/parsing/Parse_rule.ml
```diff
@@ -1322,6 +1322,9 @@ let parse_taint_source ~(is_old : bool) env (key : key) (value : G.expr) :
       take_opt dict env parse_bool "by-side-effect"
       |> Option.value ~default:false
     in
+    let source_control =
+      take_opt dict env parse_bool "control" |> Option.value ~default:false
```
aryx, line 1326: should this be an options: of the rule instead?

aryx (approved):

src/parsing/Parse_rule.ml
IagoAbal: It could be that way too, but having it as an option in the source gives us more flexibility. A sink can now depend on both data and control sources (using taint labels).

PR #8335, opened 2023-07-20
fix: [ci] improve caching for opam

spencerdrak:
Our CI for OSX has slowed down, due to the fact that the cache was being read successfully, but was never being updated, and therefore was out of date. This means that the `opam update` step had considerable amounts of work to do on each run. 

This PR improves caching for the `~/.opam` dir. Testing shows that the prefix-based `restore-keys` is working as expected. 

We may see cache miss on this PR until this is merged into `develop` - caches from feature branches will use the default branches cache, but until that gets populated, we won't see the benefit. 

Test Plan:
- CI on this PR
- Ensure develop checks run
- Monitor follow-on PRs for speedup

PR checklist:

### DIFF
--- a/.github/workflows/build-test-osx-arm64.yaml
+++ b/.github/workflows/build-test-osx-arm64.yaml
@@ -54,9 +54,9 @@ jobs:
         with:
           path: ~/.opam
           #TODO: we should add the md5sum of opam.lock as part of the key
-          key: "${{ runner.os }}-${{ runner.arch }}-${{ env.OPAM_SWITCH_NAME }}-opam-deps-${{ github.run_id }}"
+          key: ${{ runner.os }}-${{ runner.arch }}-${{ env.OPAM_SWITCH_NAME }}-opam-deps-${{ github.run_id }}
           restore-keys: |
-            "${{ runner.os }}-${{ runner.arch }}-${{ env.OPAM_SWITCH_NAME }}-opam-deps"
+            ${{ runner.os }}-${{ runner.arch }}-${{ env.OPAM_SWITCH_NAME }}-opam-deps
       - name: Install dependencies
         run: |
           ./scripts/osx-setup-for-release.sh "${{ env.OPAM_SWITCH_NAME }}"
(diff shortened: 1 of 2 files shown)


cgdolan (approved):

.github/workflows/build-test-osx-arm64.yaml
```diff
@@ -45,16 +45,18 @@ jobs:
...
+      - name: Cache Opam
+        uses: actions/cache@v3
         if: ${{ inputs.use-cache }}
         env:
           SEGMENT_DOWNLOAD_TIMEOUT_MINS: 2
         with:
           path: ~/.opam
           #TODO: we should add the md5sum of opam.lock as part of the key
-          key: "${{ runner.os }}-${{ runner.arch }}-${{ env.OPAM_SWITCH_NAME }}-opam-deps"
+          key: ${{ runner.os }}-${{ runner.arch }}-${{ env.OPAM_SWITCH_NAME }}-opam-deps-${{ github.run_id }}
```
aryx, line 57: Again, please add comments in the code rather than the commit message.
I have no idea how this key: restore-keys work, and it does not seem intuitive to me at all, so a small
explanation would be useful here.

aryx: Also you went back and forth on this using actions/cache vs actions/cache/restore but in the process you removed one of my comment that was explaining how actions/cache works (and the fast that there was a post-cache phase at the end that was saving things).
Please restore my old comment, so in 6 months someone from PA can understand what's going on.

.github/workflows/build-test-osx-x86.yaml
```diff
@@ -78,20 +80,17 @@ jobs:
...
-        id: restore-opam-cache
-        uses: actions/cache/restore@v3
+      - name: Cache Opam
+        uses: actions/cache@v3
         if: ${{ inputs.use-cache }}
-        # The size of the tgz of ~/.opam is around 500MB, which usually takes less than
-        # 2min to download from GHA. If it takes more than that, that means something
-        # wrong is going on in which case it's better to timeout early than spend
-        # 10min (the default timeout) and then still timeout.
         env:
```
aryx, line 86: Why removing that comment?
I wrote that comment because PA people don't have a deep understanding of GHA but according to our devops philosophy, PA people are supposed to maintain/understand/modify those files (with some help from infra when we need to do new/tricky things), so we need to keep those comments.

PR #8336, opened 2023-07-20
set proper platform tag for python wheels

tpetr:
Backstory: Prior to supporting linux/arm64, we published a semgrep Python wheel with the [platform compatibility tag](https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/) set to `any`, which tells pip to fall back to this if there are no other platform-specific wheels available.

When we started publishing wheels for both architectures, I naively assumed that things would Just Work. Unfortunately, the platform compatibility tags that setup.py was setting by default (`linux_x86_64` and `linux_aarch64`) are not accepted by PyPI (the reason why is kind of a long story and has to do with incompatibilities between glibc and musl).

It turns out the platform compatibility tag is literally just the suffix on the wheel filename (no changes to the actual content of the wheel), so this PR basically just updates the generated wheel to have the proper filenames.

The key word here is filename**s**. In order to support both glibc (e.g. Debian) and musl (e.g. Alpine) platforms, we need to publish wheels targeting `manylinux2014` and `musllinux_1_0`, which is why we have the two-copies-and-a-delete in build-wheels.sh

NOTE: After typing all this out I realized that we should really have _something_ as the fallback `any` wheel, so I just updated the PR have `linux_x86_64` wheels remapped to `any`. Doing this ensures that this change only affects ARM users -- Intel users will keep using the `any` package as usual

test plan:
- [x] checks pass
- [x] grabbed artifacts from GHA and published to test.pypi.org

### DIFF
--- a/.github/workflows/build-test-manylinux-aarch64.yaml
+++ b/.github/workflows/build-test-manylinux-aarch64.yaml
@@ -27,7 +27,7 @@ jobs:
       - name: Extract wheel from docker image
         run: |
           docker load --input /tmp/image.tar
-          docker run --platform linux/arm64 --rm ${{ steps.build-semgrep-wheel.outputs.imageid }} zip -r - dist > /tmp/dist.zip
+          docker run --platform linux/arm64 --rm ${{ steps.build-semgrep-wheel.outputs.imageid }} cat cli/dist.zip > /tmp/dist.zip
       - uses: actions/upload-artifact@v3
         with:
           name: manylinux-aarch64-wheel
--- a/.github/workflows/build-test-manylinux-x86.yaml
+++ b/.github/workflows/build-test-manylinux-x86.yaml
@@ -35,7 +35,7 @@ jobs:
 
   test-wheels-manylinux:
     runs-on: ubuntu-latest
-    container: quay.io/pypa/manylinux1_x86_64
+    container: quay.io/pypa/manylinux2014_x86_64
     needs: [build-wheels-manylinux]
     steps:
       - uses: actions/download-artifact@v1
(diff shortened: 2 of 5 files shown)


spencerdrak:
Kicked off a dry-run here: https://github.com/returntocorp/semgrep/actions/runs/5623167875

I thought I opened a PR to make it so that we run dry-runs of release when our core workflow files change, I'll have to go find that work so we can do this automatically.

tpetr:
@spencerdrak sorry, I should have marked this as draft. I need to fix a regex (and add comments) and then this should be good to go.

spencerdrak:
Dry-run here: https://github.com/returntocorp/semgrep/actions/runs/5623394101

I'll dig up that work to trigger this automatically upon changes to our "core" (release.yml, build-test-*.yml) workflows

Sorry, jumped the gun again. Didn't realize this was still in draft. I'll restart that job when this is marked as ready!

tpetr:
@spencerdrak this should be good for legit testing. Writing up a PR description now.

spencerdrak:
https://github.com/returntocorp/semgrep/actions/runs/5624710368 - will automate this soon

tpetr:
This should be good for a review and merge once tests pass.

spencerdrak:
@tpetr looks like aarch64 was producing the following wheels during the release dry-run:
```
inflating: dist/semgrep-1.32.0-cp37.cp38.cp39.cp310.cp311.py37.py38.py39.py310.py311-none-manylinux2014_aarch64.whl  
inflating: dist/semgrep-1.32.0-cp37.cp38.cp39.cp310.cp311.py37.py38.py39.py310.py311-none-musllinux_1_0_aarch64.whl
```

That seems correct to me, yeah?

tpetr:
@spencerdrak yep, that's what we want!

tpetr:
confirmed that the matrix of (manylinux2014, musllinux_1_0) x (x86_64, aarch64) all pip install correctly and a semgrep scan works

spencerdrak (approved):
LGTM [- final release dry run is here,](https://github.com/returntocorp/semgrep/actions/runs/5625130368) it appears that wheels are tagged appropriately for both linux x86 and arm64.

PR #8337, opened 2023-07-20
fix: reorg for semgrep pro

ajbt200128:
This PR reorganizes Semgrep JS files some more so that way we can also support the pro engine (https://github.com/returntocorp/semgrep-proprietary/pull/835). As a bonus, this file makes it so the normal OSS Semgrep JS can now support multiple files :D.

NOTE: this changes the `engine` method, which will require a change on the app side

PR checklist:


js/Makefile
```diff
@@ -40,9 +40,21 @@ languages:
 engine:
 	cd engine; $(MAKE)
 
+.PHONY: pro-setup
+pro-setup:
```
aryx: Why can't this be done in a semgrep-pro/js/Makefile instead?
Would be good to document somewhere why we do it that way, what were the pro/cons (example of cons was having to duplicate lots of code such as XX and YY)

js/Makefile
```diff
@@ -40,9 +40,21 @@ languages:
...
+ifdef SEMGREP_PRO
```
aryx: Why do we do that? What's the effect? What happens if we remove this for semgrep-pro? We get some linking error?

js/engine/Main.ml
```diff
@@ -71,22 +71,44 @@ let _ =
          | Some lang -> Js.some (Js.string (Lang.to_lowercase_alnum lang))
          | None -> Js.null
 
-       method execute language rule_file source_file : string =
+       method execute language rule_file root source_files : string =
```
aryx: Maybe a comment explaining why we need root here? Just explain this is for semgrep-pro or something

aryx: Also maybe add a comment note such as
(* note that here we take a list of source files! The playground generally parses just one file at a time, but
 * we accept a list of files because XXX
 *)

js/engine/Main.ml
```diff
@@ -71,22 +71,44 @@ let _ =
...
+         let xlang = Xlang.of_string (Js.to_string language) in
+         let rules_and_errors =
```
aryx, line 86: If this is mostly a copy paste of some code in Run_semgrep.ml or something, please add a (* coupling:  *) comment
explaining the copy-paste and how it differs from the original.

js/engine/src/index.d.ts
```diff
@@ -12,7 +12,8 @@ export interface Engine {
   execute: (
     language: string,
     rulesFilename: string,
-    targetFilename: string
+    root: string,
```
aryx, line 15: Would be nice to document a bit better those interface files (they are like .mli, and we usually try hard
to have good documentation in it, to better understand the APIs).

aryx: For example it's not obvious why we take a list of files and a root now instead of a single file.

aryx (approved):

js/engine/Main.ml
aryx: Or maybe just say turbo mode used to be just for semgrep OSS in the playground with a single target file, but we plan to add support in the playground for semgrep-pro, in which case we might have to execute semgrep on multiple files.

js/Makefile
ajbt200128: We could move it there, this just seemed simplest to me to reuse things (instructions stay exactly the same, just add an environment variable). I'll add a comment explaining this

js/Makefile
ajbt200128: Yes, a bunch of build files for the parsers and engine depend on a specific build directory, and this is easier than making changes to all of those. I'll leave a comment about this

js/engine/Main.ml
ajbt200128: > plan to add support in the playground for semgrep-pro, in which case we might have to execute semgrep on multiple files.

Yea this is why, I'll add a comment

js/engine/Main.ml
ajbt200128: Will do

js/engine/src/index.d.ts
ajbt200128: Yea I'll add some more comments

js/Makefile
tpetr: I'd really prefer it if we kept pro-specific stuff in the pro repo. It feels wrong to have proprietary stuff in the OSS makefile. if it's essential that keep these changes in the OSS repo, can you help me better understand why?

js/Makefile
```diff
@@ -40,9 +40,21 @@ languages:
...
+	ln -s -f ../_build ../_build
```
tpetr: what does this line do? it seems like a noop

js/Makefile
tpetr: what would those changes look like?

js/engine/Main.ml
tpetr: - what exactly is `root`? is it the path we use for resolution in the case that a file is a relative path? I agree w/ Pad that we should comment this, but we're not beholden to copying ocaml's variable names and it can't hurt to make the variable more self-explanitory
- what's your plan to update semgrep-app?

js/engine/src/index.d.ts
```diff
@@ -12,7 +12,8 @@ export interface Engine {
...
+    targetFilenames: string[]
```
tpetr, line 16: re: my question about semgrep-app, can we make this `string[] | string` and coerce `string` to `string[]` inside index.js?

js/Makefile
ajbt200128: It's not essential, just a choice I made, I can move it to Pro

js/Makefile
ajbt200128: anywhere `_build` is referenced would have to by grabbed from an environment variable, as currently they are all hardcoded to find _build in `semgrep/_build`. We'd have to change some of the references in the engine files, and in the shared lang files. Imo symlinking feels less hacky than passing this as an environment variable, but I don't feel strongly. I also checked if dune will tell you the build dir location, and it will not

js/Makefile
ajbt200128: it's not, symlinks are weird and relative. So we're creating a symlink called `_build` in the parent directory, that will point to `_build` in the symlink's parent directory. So we're linking semgrep-pro's _build to semgrep/_build

js/engine/Main.ml
ajbt200128: root is the root of scanning, and it's required for semgrep pro intrafile. OSS wise it doesn't do anything, it's just here to keep the interface the same

ajbt200128: As far as my plan to update semgrep-app, I was just going to make a PR that bumps the default version and updates the interface to use this new one. I think this should be fine since the version is hardcoded, so nothing should break in the next release

js/engine/src/index.d.ts
ajbt200128: execute right now is a direct ref to the ocaml jsoo output. We can make execute an anonymous function that calls it, and coerce, but I'm not sure I see a reason to. With this PR oss semgrep js can scan multiple files, and I don't see a reason to remove that, even if we're not using it right now. Also we're the only ones using this interface, so when we update semgrep-app (see response above), I'll just fix it and things should be ok. I'm going to have to do this to support the root field anyways, which we need if we want semgrep pro js, and want to keep the interfaces/build process simple.

aryx:
Can we merge this?

ajbt200128:
Yea, was waiting until after hack week/when I got back from defcon to

PR #8339, opened 2023-07-20
feat!(...): update parser + schema

yosefAlsuhaibani:
PR checklist:


libs/networking/Http.ml
```diff
@@ -1,23 +1,12 @@
...
-
-(* The HTTP lwt client library uses happy_eyeballs as the underlying
-   layer for establishing connections. It uses DNS, and comes with a small DNS
-   cache. We use a single happy_eyeballs instance to reuse the cache present.
-*)
-let happy_eyeballs = Happy_eyeballs_lwt.create ()
-
-(* TODO: extend to allow to curl with JSON as answer *)
-let get ?headers url =
+open Piaf
```
aryx: This will be in conflict with @ajbt200128 who is also modifying this file to use another Http library ...
We need to figure things out here and use just one if possible

aryx: But yes we can move to libs/networking/

aryx: We can also have Http_helpers1.ml and Http_helpers2.ml for now, where one use cohttp and ther other piaf
(or maybe http_helpers_piaf.ml and http_helpers_cohttp.ml)

dune-project
```diff
@@ -129,6 +129,10 @@ a few other projects developed at r2c.
   (synopsis "XXX")
   (description "XXX")
 )
+(package (name http)
```
aryx: maybe networking if it's in libs/networking

src/core/Rule.ml
```diff
@@ -358,6 +358,23 @@ and extract_reduction = Separate | Concat [@@deriving show]
 and extract_transform = NoTransform | Unquote | ConcatJsonArray
 [@@deriving show]
 
+(*****************************************************************************)
+(* Secrets mode *)
+(*****************************************************************************)
+
+(*TODO: actualy figure this out*)
```
aryx: What do you need to figure out?
Maybe be clearer in the comment on what is the TODO

src/core/Rule.ml
```diff
@@ -358,6 +358,23 @@ and extract_reduction = Separate | Concat [@@deriving show]
...
+type secrets_spec = {
+  (* secrets: *)
+  secrets : formula list;
+  (* request: *)
```
aryx: better define intermediate structures for those, like request_spec, response_spec (or just request and response)

src/core/Rule.ml
```diff
@@ -358,6 +358,23 @@ and extract_reduction = Separate | Concat [@@deriving show]
...
+  url : string;
```
aryx: Can we use Uri.t here?

src/core/Rule.ml
```diff
@@ -358,6 +358,23 @@ and extract_reduction = Separate | Concat [@@deriving show]
...
+  method_ : string;
```
aryx: Can we use a more precise type?
The http libraries usually define this like Get | Post | ...

src/core/Rule.ml
```diff
@@ -358,6 +358,23 @@ and extract_reduction = Separate | Concat [@@deriving show]
...
+  headers : (string * string) list;
+  (* response: *)
```
aryx: same, better to have intermediate record for response since I guess it will grow

src/engine/Match_rules.ml
```diff
@@ -100,7 +100,7 @@ let group_rules xconf rules xtarget =
            | _ when not relevant_rule -> Right3 r
            | `Taint _ as mode -> Left3 { r with mode }
            | (`Extract _ | `Search _) as mode -> Middle3 { r with mode }
-           | `Step _ ->
+           | (`Step _ | `Secrets _)->
```
aryx: so I guess this will be handled at some point in semgrep-pro?

src/osemgrep/cli_ci/Github_metadata.ml
```diff
@@ -202,15 +202,27 @@ let _find_branchoff_point_from_github_api env =
   | Some gh_token, Some api_url, Some head_branch_hash -> (
       let headers = [ ("Authorization", Fmt.str "Bearer %s" gh_token) ] in
       let open Lwt.Infix in
-      Http_lwt_client.request ~meth:`GET ~headers
+        (* This is the worst thing I've ever done, I am so sorry
```
aryx: No!
Fix it.
The code was working before, just define your own http_helpers.ml functions to not bother
existing code.

libs/networking/Http.ml
aryx: Also please keep as Http_helpers.ml, so it's clearer it's just some wrappers over another
official http library, otherwise it looks like we use an official Http robust library.

src/osemgrep/cli_login/Login_subcommand.ml
```diff
@@ -96,7 +96,7 @@ let run (conf : Login_CLI.conf) : Exit_code.t =
                     {|{"token_request_key": "|} ^ Uuidm.to_string session_id
                     ^ {|"}|}
                   in
-                  match Http_helpers.post ~body url with
+                  match Http.post ~body url with
```
aryx: Might be better to restore Http_helpers_piaf.ml or something.

src/parsing/Parse_rule.ml
```diff
@@ -1669,6 +1705,8 @@ let parse_mode env mode_opt (rule_dict : dict) : R.mode =
       in
       `Extract
         { formula; dst_lang; extract_rule_ids; extract; reduce; transform }
+  | Some ("semgrep_internal_postprocessor", _), _ ->
```
aryx, line 1714: please use a better name, maybe semgrep_internal_secrets.
Otherwise I have no idea wiaht postprocessor mean/

aryx (changes requested):

src/parsing/Parse_rule.ml
emjin: This was a compromise with Lewis to make him ok with merging the semgrep changes into the public facing repo; it's intentionally vague.

src/engine/Match_rules.ml
yosefAlsuhaibani: Yes; evaluation happens there.

src/core/Rule.ml
yosefAlsuhaibani: I think thats an old TODO; when I was drafting syntaxes; will delete that line

src/osemgrep/cli_ci/Github_metadata.ml
yosefAlsuhaibani: Oh my bad 😭 I left this as a placeholder when I was moving between libraries, will fix...

src/core/Rule.ml
yosefAlsuhaibani: Added `request_spec` && `response_spec`!

src/core/Rule.ml
yosefAlsuhaibani: Changed it to `meth : [`GET | `POST]`.

src/osemgrep/cli_login/Login_subcommand.ml
yosefAlsuhaibani: I'll just restore it to `Http_helpers`

src/parsing/Parse_rule.ml
emjin: @yosefAlsuhaibani We should add a TODO to make this more specific

src/core/Rule.ml
```diff
@@ -275,6 +275,31 @@ and extract_transform = NoTransform | Unquote | ConcatJsonArray
...
 and extract_reduction = Separate | Concat [@@deriving show]
 
+(*****************************************************************************)
+(* secrets mode *)
+(*****************************************************************************)
+
+type request_spec = {
+  url : Uri.t;
+  meth : [`GET | `POST];
+  headers : (string * string) list
```
emjin: What are each of these strings? It would be nice to have comments or make new types

src/core/Rule.ml
```diff
@@ -275,6 +275,31 @@ and extract_transform = NoTransform | Unquote | ConcatJsonArray
...
+}
+[@@deriving show]
+
+type response_sepc = {
+  return_code : int;
+}
+[@@deriving show]
+
+type secrets_spec = {
+  (* postprocessor-patterns: *)
```
emjin: Needs more explanation. What does postprocessor-patterns refer to? Why is it associated with `secrets:`?

src/core/Rule.ml
```diff
@@ -421,9 +447,23 @@ type 'mode rule_info = {
...
 [@@deriving show]
 
+(* If you know your function accepts only a certain kind of rule,
+ * you can use those precise types below.
+ *)
+type search_rule = search_mode rule_info [@@deriving show]
+type taint_rule = taint_mode rule_info [@@deriving show]
+type extract_rule = extract_mode rule_info [@@deriving show]
+type secrets_rule = secrets_mode rule_info [@@deriving show]
+type step_rule = step_mode rule_info [@@deriving show]
```
emjin: When did this get renamed to `step_mode` and `step_rule` (no `s`)?

src/core/Rule.ml
```diff
@@ -275,6 +275,31 @@ and extract_transform = NoTransform | Unquote | ConcatJsonArray
...
+(*****************************************************************************)
+
+type request_spec = {
+  url : Uri.t;
+  meth : [`GET | `POST];
+  headers : (string * string) list
+}
+[@@deriving show]
+
+type response_sepc = {
```
aryx: typo, spec no sepc.
Also I would remove those _spec prefix (I have a PR where I removed it from extract; I kept the
taint_spec because there's code depending on this name type, and I was lazy, but I think it would
be simpler to remove those _spec suffix from the start here.

libs/networking/dune
```diff
@@ -0,0 +1,22 @@
+(library
+ (public_name http)
```
aryx: better networking for the name, to match the directory
(and so that later we can add helpers for other networking stuff, not just http)

src/parsing/Parse_rule.ml
aryx: But we have lots of use of `Secrets in the code ... in the public repo ... even this discussion is in the public repo
so I don't really get obfuscating it in the rule.

aryx (dismissed):

src/parsing/Parse_rule.ml
```diff
@@ -304,6 +305,17 @@ let parse_string_wrap env (key : key) x =
 (* TODO: delete at some point, should use parse_string_wrap instead *)
 let parse_string env (key : key) x = parse_string_wrap env key x |> fst
 
+let parse_method env (key : key) x =
+  let meth = parse_string env key x in
```
emjin, line 308: Minor: I would just call this "method"

src/parsing/Parse_rule.ml
emjin: The rules are user facing, so I believe the goal is to keep users from getting used to `mode: secrets` (since this is experimental and we think it will change) and to keep it a little quieter, since people are more likely to see rules than look at our code base. I don't think we're trying that hard to keep it secret though, all our paying customers know it's coming soon and it's not like they're under NDA.

Lewis would prefer we also don't use secrets in the code but he's ok with it if it would be very annoying for us as maintainers.

libs/networking/Http.ml
yosefAlsuhaibani: We're switching to only Cohttp + moving networking to a libs + changing `Http.*` to `Http_helpers.*`

src/core/Rule.ml
yosefAlsuhaibani: I think this was due to a kerfuffle a merge between my branch & develop; will change them back

yosefAlsuhaibani: I think this is due to a kerfuffle when I was merging into develop; reverted back

src/core/Rule.ml
yosefAlsuhaibani: Added bits of documentation; should I expand a bit more?

src/core/Rule.ml
yosefAlsuhaibani: Removed _spec

src/core/Rule.ml
yosefAlsuhaibani: I think I've updated the types since then to add more context to what this type encodes; `headers` field might be a bit ambigous since its still just `string * string`, but I could update it to a `headers : { name : string; value : string } list`; Or `Cohttp.Headers.t`.

src/core/Rule.ml
yosefAlsuhaibani: Add intermediate records for `response` & `request`

src/core/Rule.ml
emjin: A record would be nice!

src/core/Rule.ml
```diff
@@ -275,6 +275,52 @@ and extract_transform = NoTransform | Unquote | ConcatJsonArray
 *)
 and extract_reduction = Separate | Concat [@@deriving show]
 
+(*****************************************************************************)
+(* secrets mode *)
+(*****************************************************************************)
+
+(* This type encodes a basic HTTP request; mainly used for in the secrets
```
aryx, line 282: Great comment!
Also might be useful to explain why we don't reuse the request type of http libraries like cohttp, which have their own request type.

src/core/Rule.ml
```diff
@@ -275,6 +275,52 @@ and extract_transform = NoTransform | Unquote | ConcatJsonArray
...
+ *)
+type response = { return_code : int } [@@deriving show]
+
+type secrets = {
+  (* postprocessor-patterns:
+   * Each pattern in this list represents a piece of a "secret"; with any
+   * bindings made available in the request post matching.
+   *)
+  secrets : formula list;
+  (* request: *)
```
aryx: you can remove the comment, it's not useful anymore

src/core/Rule.ml
```diff
@@ -275,6 +275,52 @@ and extract_transform = NoTransform | Unquote | ConcatJsonArray
...
+  request : request;
+  (* response: *)
```
aryx: same, remove old comment.

aryx (commented):
LGTM.
I'll let @emjin approve.

aryx:
Lots of CI failures to fix though.

emjin (approved):
Let's merge it!

PR #8348, opened 2023-07-23
fix(yaml,extract): remove style and chomp indicators from raw string tokens

MichaHoffmann:
Previously the tokens for raw or folded yaml strings would include style ( `|` or `>` ) and chomp ( `+` or `-` ) markers which breaks extractions since they do not technically belong to the matched string. 

PR checklist:

### DIFF
--- a/changelog.d/gh-8348.fixed
+++ b/changelog.d/gh-8348.fixed
@@ -1,0 +1,1 @@
+yaml: exclude style markers from matched token in block scalars
--- a/cli/tests/e2e/snapshots/test_metavariable_pattern/test1/results.json
+++ b/cli/tests/e2e/snapshots/test_metavariable_pattern/test1/results.json
@@ -67,16 +67,16 @@
         },
         "metavars": {
           "$SHELL": {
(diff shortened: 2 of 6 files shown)


MichaHoffmann:
Test failures are related to new tokens; the previous assertions assert on matching the "|-" too. Waiting with the fix for a quick review if the method is the correct one.

emjin (commented):
The approach looks good to me! It would be nice to have tests to cover all the potential markers, but I'm happy to merge this without that as soon as tests pass.

One thing is there are a lot of extra whitespaces added (we use ocamlformat 0.21.0). I don't think I can modify your PR to fix them, but I can make a PR after this is merged.

MichaHoffmann:
> The approach looks good to me! It would be nice to have tests to cover all the potential markers, but I'm happy to merge this without that as soon as tests pass.
> 
> One thing is there are a lot of extra whitespaces added (we use ocamlformat 0.21.0). I don't think I can modify your PR to fix them, but I can make a PR after this is merged.

weird i also though i used ocamlformat 0.21.0! added it via `ocamlPackages.ocamlformat` in my nix shell. maybe its a newer one; i revert the whitespace changes

emjin (approved):
Thank you!

emjin:
Benchmarks are failing because they need to post a comment, so we can safely ignore that

PR #8350, opened 2023-07-24
sym-prop: Inspect 'id_svalue's during cycle checks

IagoAbal:
We were overwriting `visit_id_info` in a way that it was not visiting 'id_svalue's, therefore the cycle checks did not detect transitive dependencies.

Closes PA-2933

test plan:
make test # new test
See PA-2933 reproduction

PR checklist:

### DIFF
--- a/src/analyzing/Dataflow_svalue.ml
+++ b/src/analyzing/Dataflow_svalue.ml
@@ -406,18 +406,32 @@ let no_cycles_in_svalue (id_info : G.id_info) svalue =
      * we can have a single visitor for all calls, given that the old
      * `mk_visitor` was pretty expensive, and constructing a visitor object may
      * be as well. *)
+    let i = ref 0 in
     let ff = ref (fun _ -> assert false) in
     let ok = ref true in
     let vout =
-      object
-        inherit [_] G.iter
+      object (self : 'self)
+        inherit [_] G.iter_no_id_info
 
-        method! visit_id_info _env ii =
+        method! visit_id_info env ii =
           ok := !ok && !ff ii;
+          (match !(ii.id_svalue) with
+          | Some (Sym e) when !ok ->
+              (* Following `id_svalue`s can explode in pathological cases,
+               * see 'tests/rules/sym_prop_explosion.js', so we need to
+               * set a bound. *)
+              if !i < 1000 then (
+                incr i;
+                self#visit_expr env e)
+              else ok := false
+          | None
+          | Some _ ->
+              ());
           if not !ok then raise Exit
       end
     in
     fun f ast ->
+      i := 0;
       ff := f;
       ok := true;
       try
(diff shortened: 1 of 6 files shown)


src/analyzing/Dataflow_svalue.ml
```diff
@@ -406,18 +406,32 @@ let no_cycles_in_svalue (id_info : G.id_info) svalue =
...
      * `mk_visitor` was pretty expensive, and constructing a visitor object may
      * be as well. *)
+    let i = ref 0 in
     let ff = ref (fun _ -> assert false) in
     let ok = ref true in
     let vout =
-      object
-        inherit [_] G.iter
+      object (self : 'self)
+        inherit [_] G.iter_no_id_info
```
emjin, line 414: Why do we `iter_no_id_info` when we're visiting `id_info`? Curious for my understanding

src/analyzing/Dataflow_svalue.ml
```diff
@@ -436,7 +450,7 @@ let no_cycles_in_svalue (id_info : G.id_info) svalue =
       for_all_id_info
         (fun ii ->
           (* Note the use of physical equality, we are looking for the *same*
-           * id_svalue ref, that tells us it's the same variable occurrence. *)
+                * id_svalue ref, that tells us it's the same variable occurrence. *)
```
emjin: minor typo here

src/analyzing/Dataflow_svalue.ml
```diff
@@ -406,18 +406,32 @@ let no_cycles_in_svalue (id_info : G.id_info) svalue =
...
 
-        method! visit_id_info _env ii =
+        method! visit_id_info env ii =
           ok := !ok && !ff ii;
+          (match !(ii.id_svalue) with
+          | Some (Sym e) when !ok ->
+              (* Following `id_svalue`s can explode in pathological cases,
+               * see 'tests/rules/sym_prop_explosion.js', so we need to
+               * set a bound. *)
+              if !i < 1000 then (
```
emjin: Should we put 1000 in `Limits_semgrep`?

emjin (approved):
LGTM. If symbolic propagation is the most common cause of stack overflows, maybe we can try to include a note to that effect in the error message when we see one. We should maybe also think about how we could do more stress testing to find these ourselves earlier

emjin:
Note that I am not seeing these benchmark slowdowns on a lot of other PRs. If the PR causes a slowdown we can still merge, just noting that I think they might be related

Edit: nvm, I see them on another PR. I'm concerned that we've had so many benchmark repins now

src/analyzing/Dataflow_svalue.ml
IagoAbal: We're overwriting the `visit_id_info` and not using the one from `super` so we can use any of them really. But I thought it was slightly more clear to inherit from `iter_no_id_info` because it does not visit `id_info` so you immediately know that `visit_id_info` does not depend on inherited behavior.

src/analyzing/Dataflow_svalue.ml
IagoAbal: Sure why not!

IagoAbal:
> LGTM. If symbolic propagation is the most common cause of stack overflows, maybe we can try to include a note to that effect in the error message when we see one. We should maybe also think about how we could do more stress testing to find these ourselves earlier

If it wasn't because of Parmap, this should not be crashing Semgrep. We may need to improve Parmap ourselves in this respect.

Also, although the cause was this one and it's fixed by this PR, I should try to investigate where exactly where the stack overflow actually happened, to add an extra check there.

PR #8352, opened 2023-07-24
cleanup: move Rule.ID in separate file Rule_ID.ml

aryx:
I try to reduce the size of Rule.ml which is getting big.
Separate concerns should be in separate files.

test plan:
make core

PR checklist:

### DIFF
--- a/src/core/Mini_rule.ml
+++ b/src/core/Mini_rule.ml
@@ -32,7 +32,7 @@
 (*****************************************************************************)
 
 type rule = {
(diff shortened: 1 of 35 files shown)


emjin:
We're starting to see this benchmark slowdown message on a lot of PRs starting today. This seems to be the first one where it happened. I'm suspicious it might have caused a slowdown somehow

aryx:
It's a pure refactoring; there's no change of semantic, just a module moved out of Rule.ml in his own file.

PR #8356, opened 2023-07-24
feat(php): Support case insensitive function calls and classes

### LINKED ISSUE #7231: PHP function call matching not case insensitive
**Describe the bug**
In PHP function names are not case sensitive. This allows a function to be defined as `lol` and called as `Lol` (or any other variation). Function matching within Semgrep does not account for this and will only match in a case sensitive situation. Documentation detailing this can be found [here](https://www.php.net/manual/en/functions.user-defined.php)

**To Reproduce**
https://semgrep.dev/playground/s/jrozner:php-function-case-sensitivity

**Expected behavior**
Semgrep should be able to match all variations of a function call regardless of case of the letters

**Screenshots**

**Environment**
Reproducible in all environments

**Use case**
More accurate detection in PHP. This can currently result in false negatives.

akuhlens:
## What
Adds matching support for languages that have case insensitive identifiers and demonstrates their usage for Php.

closes #7231

## How
Adds a boolean field to `id_info` fields and updates `Generic_vs_generic.ml` and `Matching_generic.ml` to respect these fields. I originally thought it would be easier to add a special case for Php in matching, but extending `Matching_generic.ml` to be language specific becomes troublesome because `equal_ast_bound_code` is called from outside this submodules in contexts that could possibly be addressing variables for multiple languages (or at least that was my take on the situation and types at play).

## Remaining Work To Do
This was shared as a draft to communicate my work and get feedback about how to add the bitfield to id_info and update the submodule.
- [x] Modify `id_info` to contain a `id_flags` field that contains a bitfield instead of having two separate boolean fields (id_case_insensitive, and id_hidden).
- [x] Clean up code and document purpose.
- [x] Add change log entry.
- [x] Submit pull request for `semgrep-rules` submodule and update submodule to point to main branch again. (Not actually sure the correct order to do this without breaking things). [(ongoing here)](https://github.com/returntocorp/semgrep-rules/pull/3013)

## Testing
Adds a few test cases matching against identifiers and metavariables in a case insensitive fashion and updates `tests/semgrep-rules` that were disabled due lack of support for this. Note, `tests/semgrep-rules` is currently pointing to a branch that I need to open a pull request for.

PR checklist:


tests/rules/metavar_call.php
```diff
@@ -0,0 +1,18 @@
+<?php
+function foo() {
+
+    # ruleid:test
```
aryx, line 4: Ok but I don't think it will work for a pattern like

$X == $X

and then the target PHP code is
foo() == FOO()

Can you add a test for that?

akuhlens: Added as `metavar_call_case_insensitve.php`.

src/matching/Generic_vs_generic.ml
```diff
@@ -187,7 +193,13 @@ let m_ident a b =
       let re_match = Matching_generic.regexp_matcher_of_regexp_string stra in
       if re_match strb then return () else fail ()
   (* general case *)
-  | a, b -> (m_wrap m_string) a b
+  (* TODO: Should this respect case_sensitivity of ids? *)
+  | a, b ->
+      with_lang (fun lang ->
+          if lang =*= Lang.Php then
+            Printf.printf "%s %s\n%s\n\n" (fst a) (fst b)
+              Printexc.(get_callstack 100 |> raw_backtrace_to_string);
```
akuhlens: Whoops cleaning this up right now.

src/matching/Generic_vs_generic.ml
```diff
@@ -583,41 +598,44 @@ and m_ident_and_empty_id_info a1 b1 =
...
+(*      * import, without this tag, which would prevent *)
+(*      * matching. We need to correctly resolve names and always compare with *)
+(*      * the resolved_name instead of the name used in the code *)
+(*      * (which can be an alias) *)
+(*      * *)
+(*      * old: (m_ref (m_option m_type_)) a2 b2 *)
+(*      * the same is true for types! Now we sometimes propagate type annotations *)
+(*      * in Naming_AST.ml, but we do that in the source file, not the pattern, *)
+(*      * which would prevent a match. *)
+(*      * More generally, id_info is something populated and used on the *)
```
akuhlens: Willing to put this back to being uncommented, but I got tired of changing the pattern match that was just being used for documentation.

akuhlens:
> 📸 The pytest shapshots changed in your PR. Please carefully review these changes and make sure they are intended:
> 
> 1. Review the changes at [f67618b](https://github.com/returntocorp/semgrep/commit/f67618b9d91dd165e560a20394d132c8dace57a1)
> 2. Accept the new snapshots with
>    ```
>    git fetch origin && git cherry-pick f67618b && git push
>    ```

I think these changes to the python output are fine. I am pretty sure they are caused by the update to `libs/ast_generic/Meta_ast.ml`, but I would appreciate someone confirming this for me.

akuhlens:
The tests that are failing are the ones that I have an approved pull request in semgrep-rules for. I just can't land that PR until those tests pass on develop, which requires I land this diff first.

ihji:
@akuhlens Thanks for the awesome work! Just out of curiosity, what were the major benefits of introducing bit encoding instead of using two boolean record fields? I might be missing some context, but it appears a bit over-engineered to me.

akuhlens:
> @akuhlens Thanks for the awesome work! Just out of curiosity, what were the major benefits of introducing bit encoding instead of using two boolean record fields? I might be missing some context, but it appears a bit over-engineered to me.

I probably wouldn't have done it that way at first either, but there was some concern on the team that adding more boolean flags to such a prevalent node in the AST might have performance implications. I went ahead and provided the optimized representation since there were concerns.
https://semgrepinc.slack.com/archives/C01NXGX2EHZ/p1689791184708529

As a side note, the language specific approach in this discussion ended up not working which is why there is no naive diff.

aryx:
You can update the semgrep-rules submodule in this repo to point to your experimental semgrep-rules, so that we can be sure
it works.

cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_file_parser__failure__error_messages/settings1/error.txt
```diff
@@ -45,7 +45,7 @@ Files skipped:
 
   [1m[24mPartially analyzed due to parsing or internal Semgrep errors[0m
 
-   • [36m[22m[24mtargets/bad/invalid_python.py (2 lines skipped)[0m
+   • [36m[22m[24mtargets/bad/invalid_python.py (3 lines skipped)[0m
```
aryx: Do you know why your PR has an impact on this snapshot?

cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_file_parser__failure__error_messages/settings1/out.json
```diff
@@ -3,13 +3,13 @@
     {
       "code": 3,
       "level": "warn",
-      "message": "Syntax error at line targets/bad/invalid_python.py:1:\n `def foo(a)\n    tree = ElementTree.parse('country_data.xml')` was unexpected",
+      "message": "Syntax error at line targets/bad/invalid_python.py:1:\n `def foo(a)\n    tree = ElementTree.parse('country_data.xml')\n    root = tree.getroot()` was unexpected",
```
aryx: same, seems weird to have this snapshot updated.

languages/php/generic/php_to_generic.ml
```diff
@@ -56,9 +56,9 @@ let ident v = wrap string v
 let var v = wrap string v
 let qualified_ident v = list ident v
 
-let name_of_qualified_ident xs =
+let name_of_qualified_ident ?(case_insensitive = false) xs =
```
aryx: since we're in a PHP file I would set the default value to true here instead (I know you explicitely
call this function with ~case_insensitive:true below, but again because I read a conversion for PHP, I'm surprised
to see false here because I know PHP is case insensitive, so better set the default to true.

libs/ast_generic/Id_info_flags.mli
```diff
@@ -0,0 +1,29 @@
...
+ * special exception on linking described in file LICENSE.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the file
+ * LICENSE for more details.
+ *)
+
+type packed
+type unpacked = { hidden : bool; case_insensitive : bool }
```
aryx: maybe better to move the documentation about hidden from the .ml to the .mli (it used to be in ast_generic.ml,
so in a .ml, but there's no ast_generic.mli because the file was mostly type definitions that is annoying to duplicate). You can also add a quick comment about case_insensitive here in the .mli such as
(* this is useful for languages such as PHP or Apex which use case_insensitive names *)

libs/commons/Immediate_bitfield.ml
```diff
@@ -0,0 +1,52 @@
...
+ * version 2.1 as published by the Free Software Foundation, with the
+ * special exception on linking described in file LICENSE.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the file
+ * LICENSE for more details.
+ *)
+
+(* This library implements a bitfield that can be stored as an
```
aryx: better to move some useful info for the user of the API in the .mli rather than the .ml
In the .ml we can add comments directed to the implementor or maintainer of the API.

src/matching/Generic_vs_generic.ml
```diff
@@ -167,6 +167,12 @@ let should_match_call = function
   | G.IncrDecr _ ->
       false
 
+let m_id_string case_insensitive =
+  let open String in
```
aryx: Not a huge fan of open or of local open.
Qualify is usually clearer, unless it really adds lots of text, but in that case we usually
prefer module aliasing with shorter names for the alias (e.g., module G = AST_generic).

aryx (changes requested):
I don't think this works if the metavariable binds to an expression rather than directly
to an idenifier.
For example if you have the pattern $X == $X,

and you have 1 + bar() == 1 + Bar()

I don't think your PR will return a match.
It's because for that we call MV.Structural.equal_mvalue a b, 
and I don't think we do anything with case sensitivity there.

languages/php/generic/php_to_generic.ml
akuhlens: PHP is only case insensitive for functions and classes. Of note variables are strangely not case sensitive, but fair I will switch it and make a note above the function.

cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_file_parser__failure__error_messages/settings1/error.txt
akuhlens: I don't, my assumption was that because I updated how the AST is reported when printed out (by adding an extra field) it added an extra line somewhere in an error message that was printing out some of the AST. I will manually run these tests to confirm that is the case.

cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_file_parser__failure__error_messages/settings1/out.json
akuhlens: Same

tests/rules/metavar_call_case_insensitive.yaml
```diff
@@ -0,0 +1,7 @@
+rules:
+- id: test
+  message: Php function references are case insensitive.
+  languages: [php]
+  severity: WARNING
+  patterns:
+  - pattern: $FUN(...) == $FUN(...)
```
aryx, line 12: Replace $FUN(...) by $X here, and see if it still works.

aryx:
To be fair, this is an important improvement about what we had before; maybe my $X == $X match of complex expression should be done in a separate PR.

aryx:
@akuhlens Maybe fix the failing tests, and add a giant TODO somewhere (maybe in equal_bind_ast case) that equality between metavar expression is still done in a case-sensitive way (because we use the derived eq). In theory we should use Generic_vs_generic too when we check for equality of metavars ... In any case this looks like progress to what we had before
so maybe we can merge this as a good step forward.

akuhlens:
@aryx Sorry, I am just seeing these last two comments. Maybe I need to  adjust my notification settings to give github notification more visibility, or perhaps they were here when I replied to your comments and I just got caught up with the bug you noticed. Either way I will take steps to make sure it doesn't happen again. I did see your comment in slack on Monday but have been out sick with covid until today. I will fix the other things you mentioned and leave a TODO and commented out test case for the derived equality case.

tests/rules/metavar_call_case_insensitive.yaml
akuhlens: Yeah, this doesn't work. Left a todo here and in Matching_generic.ml . We will get it in the next PR.

cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_file_parser__failure__error_messages/settings1/error.txt
akuhlens: It doesn't as it turns out. This is broken on my machine for some reason, even on develop.

cli/tests/e2e/snapshots/test_semgrep_core_parse_error/test_file_parser__failure__error_messages/settings1/out.json
akuhlens: Fixed.

aryx (approved):

Issue #8254, opened 2023-07-11
Autocomplete on zsh errors on no config given


davidholsgrove:
**Describe the bug**
Following documentation (https://semgrep.dev/docs/cli-reference/#autocomplete) on adding 

```
eval "$(_SEMGREP_COMPLETE=zsh_source semgrep)"
```

to `~/.zshrc`, the cli is printing error message on shell start that the config is invalid;

```
[ERROR] invalid configuration file found (1 configs were invalid)
```

The reason being `One of these properties is missing: 'rules'`. 

Adding in an empty `rules: []` to `~/.semgrep/settings.yml`, the config is considered "valid" but the error on `_SEMGREP_COMPLETE=zsh_source semgrep` is now (exit code 7):

```
[ERROR] No config given. Run with `--config auto` or see https://semgrep.dev/docs/running-rules/ for instructions on running with a specific config
```

**To Reproduce**
Move `~/.semgrep/settings.yml` aside, login with the cli, then attempt to add zsh autocomplete.

**Expected behavior**
The documented process worked previously, until a recent semgrep cli update.

**What is the priority of the bug to you?**

- [ ] P0: blocking your adoption of Semgrep or workflow
- [ ] P1: important to fix or quite annoying
- [x] P2: regular bug that should get fixed

**Environment**
Homebrew installed semgrep cli `1.31.2`

**Use case**
Use of autocomplete, and clean shell starting

emjin:
@ajbt200128 maybe you know something about this?

ajbt200128:
Hmm this was happening to me too, but I just assumed it was because I mess with my zshrc file and break things all the time. I'll take a look, I think something just broke it

armchairlinguist:
Thanks y'all - FYI the same symptoms were affecting me with Bash, which might make it more likely to be a general breakage.

ajbt200128:
Closed by https://github.com/returntocorp/semgrep/pull/8360

PR #8359, opened 2023-07-24
Fix autocomplete

ajbt200128:
Calling pysemgrep through os.exec breaks autocomplete see [here](https://click.palletsprojects.com/en/8.1.x/shell-completion/#enabling-completion). This fixes it

## Test plan

```bash
cd semgrep/cli
pipenv install
pipenv shell
eval "$(_SEMGREP_COMPLETE=zsh_source semgrep)"
semgrep --co<tab>
```

PR checklist:

- [X] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [X] Tests included or PR comment includes a reproducible test plan
- [X] Documentation is up-to-date
- [X] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [X] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #8360, opened 2023-07-24
Fix: testing error

ajbt200128:
TL;DR; we were calling semgrep without a main guard, and as a process, causing some weird exec error see [this SO issue](https://stackoverflow.com/questions/24374288/where-to-put-freeze-support-in-a-python-script)

This PR also fixes the same bug that https://github.com/returntocorp/semgrep/pull/8359 fixes, in the same way, so if we merge this we can close that PR and accompanying issue.

## Test plan

```bash
cd semgrep/cli
pipenv install
pipenv shell
semgrep --test --pro --config=. # in archive.zip provided on linear
```

PR checklist:

### DIFF
--- a/cli/bin/semgrep
+++ b/cli/bin/semgrep
@@ -68,25 +68,27 @@ def find_semgrep_core_path():
     sys.exit(2)
        
 
-if "--experimental" in sys.argv:
-    #alt: we could  extend osemgrep (and pysemgrep) to accept a new --experimental
-    # flag instead of removing the flag below. However, in osemgrep we do a few
-    # pattern matching of Sys.argv and if we have 'osemgrep --experimental --help'
-    # this would be interpreted as 'osemgrep scan --experimental --help' (because
-    # scan is the default command), and we would not get the simple semgrep --help
-    # message showing all the subcommands. So simpler to just remove --experimental.
-    sys.argv.remove("--experimental")
-    # We could have moved the code below in a separate 'osemgrep', like for 'pysemgrep',
-    # but we don't want users to be exposed to another command, so better to hide it.
-    # We expose 'pysemgrep' because osemgrep itself might need to fallback to pysemgrep
-    # and it's better to avoid the possibility of an infinite loop by simply using
-    # a different program name. Morever, in case of big problems, we can always
-    # tell users to run pysemgrep instead of semgrep and be sure they'll get the
-    # old behavior.
-    path = find_semgrep_core_path()
-    # If you call semgrep-core as osemgrep, then we get osemgrep behavior,
-    # see src/main/Main.ml
-    sys.argv[0] = "osemgrep"
-    os.execvp(str(path), sys.argv)
-else:
-    os.execvp("pysemgrep", sys.argv)
+if __name__ == "__main__":
+    if "--experimental" in sys.argv:
+        #alt: we could  extend osemgrep (and pysemgrep) to accept a new --experimental
+        # flag instead of removing the flag below. However, in osemgrep we do a few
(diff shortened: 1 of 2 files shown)


ajbt200128:
Also closes #8254

cli/bin/semgrep
```diff
@@ -68,25 +68,27 @@ def find_semgrep_core_path():
...
-    # tell users to run pysemgrep instead of semgrep and be sure they'll get the
-    # old behavior.
-    path = find_semgrep_core_path()
-    # If you call semgrep-core as osemgrep, then we get osemgrep behavior,
-    # see src/main/Main.ml
-    sys.argv[0] = "osemgrep"
-    os.execvp(str(path), sys.argv)
-else:
-    os.execvp("pysemgrep", sys.argv)
+if __name__ == "__main__":
```
aryx, line 80: Can you add a comment explaining this guard for non Python programmers?
And where do we call cli/bin/semgrep in our codebase? and why we need this guard?

aryx (approved):

cli/bin/semgrep
ajbt200128: I added a comment I think explains it, but idk for sure

ajbt200128: Python is weird, python multiprocessing is even weirder

cli/bin/semgrep
```diff
@@ -68,25 +68,36 @@ def find_semgrep_core_path():
...
+        # a different program name. Morever, in case of big problems, we can always
+        # tell users to run pysemgrep instead of semgrep and be sure they'll get the
+        # old behavior.
+        path = find_semgrep_core_path()
+        # If you call semgrep-core as osemgrep, then we get osemgrep behavior,
+        # see src/main/Main.ml
+        sys.argv[0] = "osemgrep"
+        os.execvp(str(path), sys.argv)
+    else:
+        import semgrep.__main__
```
aryx, line 102: Oh I didn't see this change in the original PR ... could we restore the execvp("pysemgrep", ...)?
Why does semgrep --test execute the code in this file anyway?

aryx: And why this was causing an error (I still don't understand the scenario in which we crash and the different code paths).

ajbt200128: No we can't remove thjs change, it's part of the fix for this issue, and for fixing autocomplete. There's some explanation for why it fixes autocomplete in the closed PR linked above. I'm also somewhat confused, but I think the best explanation I can give is that for testing, we were calling the invoke semgrep function from multiprocess. So Semgrep is called -> enters into the above -> forks and runs semgrep cli -> test command called -> we run `invoke_semgrep` a few times in parallel -> Since python has the GIL, it can't run a function in a separate thread, it has to start a whole new interpreter (or something, I'm not 100% sure) -> new process means we run all top level code -> since semgrep is imported, this means the above semgrep top level code is called (this phase is called bootstrapping) -> os.exec is called during bootstrapping, before the function that is this process's main is run, and this replaces the mid bootstrapped process before it's had a chance to run -> python does not like this.

The __name__ check stops this from happening, but I'm not sure why the import semgrep.__main__ is needed too. Just that import part fixes autocomplete too (I believe it's something to do with passing environment variables). 

The above is really just my best guess, and I think it's mostly right, but this bug was weird and fix was weird. I think the main reason is that python just really isn't good at multiprocessing, and there's a lot of foot guns

aryx: Then we need to add your big comment in the script.
Also what is autocomplete? I don't think it's supported by osemgrep anyway, so we could get rid of it.

aryx: I'll add some of your comment in my PR

PR #8362, opened 2023-07-24
fix(rust): enhanced macro args

brandonspark:
## What:
This PR fixes a bug where macro calls that had `&` or `*` inside would not transmit taint properly, as well as macro args with `.` due to dot accesses.

## Why:
Rust support.

## How:
Added logic to deal with an arg has a `*` or `&` in front of it, as well as when there is `.` after an arg.

This ends up being a "bidirectional" thing where we try to carry our arguments forward in anticipation of seeing a `.`, but we carry them backwards and apply the prefix operators after, if we see them.

Note that we only deal with the case where there is one such prefix token. This is because if there are multiple, they end up
getting folded into a single token, so the logic becomes gross. For instance, `&*x` is parsed as a single token `&*` before `x`, so this would need to be treated more carefully.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/changelog.d/pa-2951.fixed
+++ b/changelog.d/pa-2951.fixed
@@ -1,0 +1,2 @@
+Rust: Macro calls which involve dereferencing and reference operators
+(such as `foo!(&x)` and `foo!(*x)`) now properly transmit taint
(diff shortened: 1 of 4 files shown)


aryx (approved):
Argh, this is ugly.

Maybe we could alter the tree-sitter grammar and for macro arguments try
first to parse things as a regular argument (which would generate a cleaner AST),
and in case of issue just go back to fully general macro argument.

PR #8363, opened 2023-07-24
feat(matching): allow metavar params to match special params

brandonspark:
## What:
This PR allows lone metavar parameters, with no other characteristics, to match certain special parameter constructs.

This PR also fixes how Rust translates parameters that look just like `$X`, which is interpreted as a singular type name of a metavariable, as opposed to a metavariable parameter.

## Why:
This will allow matching special parameters, like `ParamPattern`, `ParamReceiver`, and `OtherParam`. Actually, it allows matching only those, because the other ones like `ParamSplat` and `ParamRest`, do not behave like singular parameters.

## How:
Just added a case for it in `Generic_vs_generic`.

Notably, we only match if the metavar param in question has no `ptype`, `pdefault`, etc, because otherwise we might miss a constraining factor.

## Test plan:
`make test`

PR checklist:


aryx (approved):
We could guard this with if lang =*= Rust, but let's go for generality.

Issue #8365, opened 2023-07-25
Adding support for naming propagation when the lhs of a var def is an ident pattern


ihji:
Adding support for naming propagation when the left-hand side (lhs) of a variable definition is an identifier pattern.

ihji:
Fixed by #8366

PR #8364, opened 2023-07-25
feat(types): infer types better for annotated pattern ids

brandonspark:
## What:
This PR makes it so that pattern IDs which have type annotations now transmit that information.

## Why:
We want more types, so typed metavariables work better.

## How:
Just added `EPattern (PatId ...)` to the cases for `Naming_AST`, as it's pretty much equivalent to `EN (N (Id ...))`.

## Test plan:
`make test`

PR checklist:

- [X] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [X] Tests included or PR comment includes a reproducible test plan
- [X] Documentation is up-to-date
- [X] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [X] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #8366, opened 2023-07-25
feat(naming): Adding support for naming propagation when the lhs of a var def is an ident pattern

ihji:
PR checklist:


changelog.d/gh-8365.added
```diff
@@ -0,0 +1,9 @@
+feat(naming): Adding support for naming propagation when the left-hand side (lhs) of a variable definition is an identifier pattern
```
nmote:
```suggestion
Added support for naming propagation when the left-hand side (lhs) of a variable definition is an identifier pattern
```

We settled on not using the conventional commit style for changelog entries, I believe. We had a recent slack discussion on this. We also typically use the past tense in changelog entries.

tests/rules/typed_metavar_not_rust.rs
```diff
@@ -18,3 +18,9 @@ async fn index(form: web::Form<MyForm>,
         response.body("default string".to_string()),
     }
 }
+
+async fn index2() -> impl Responder {
+    let response: HttpResponseBuilder = get_resonse();
```
nmote:
```suggestion
    let response: HttpResponseBuilder = get_response();
```

Inconsequential, but there's a typo here.

src/naming/Naming_AST.ml
```diff
@@ -590,6 +590,7 @@ let resolve lang prog =
                     (* TODO handle nested destructuring? *)
                 | _ -> ())
               fields
+        | { name = EPattern (PatId (id, id_info)); _ }, VarDef { vinit; vtype }
```
nmote, line 596: Could you add a comment here with a brief example so that future readers understand why we're handling this case? Also, do you think there should be a TODO to handle more pattern kinds?

nmote (approved):
Could you go through the checkboxes in the PR and check them if applicable?

brandonspark:
funnily enough, it seems we have implemented the same PR: https://github.com/returntocorp/semgrep/pull/8364/files

src/naming/Naming_AST.ml
ihji: Done. Thanks for the comments!

ihji:
> funnily enough, it seems we have implemented the same PR: https://github.com/returntocorp/semgrep/pull/8364/files

😅 

I see that both pull requests have essentially the same change, but it still seems beneficial to merge the test cases. Would you like me to bring in your test cases, or would you prefer to take the tests from this pull request?

nmote:
@ihji, since your PR is already approved and you have a more thorough changelog entry, could you take @brandonspark's test cases and then merge your PR?

PR #8369, opened 2023-07-25
simplify build-test-docker workflow

tpetr:
`depot/build-push-action` has been working well and will fall back to `docker/build-push-action` in case of failure, so we're good to remove all the `docker/build-push-action` steps and related conditionals

test plan:
- checks pass


aryx (approved):
a red diff is a good diff

PR #8370, opened 2023-07-25
dispatch to osemgrep by default

aryx:
... but quickly dispatch back to pysemgrep for most commands,
unless you used --experimental.

But 'semgrep interactive' does not require anymore the --experimental!
Next target is 'semgrep lsp' e2e in OCaml, without --experimental.

test plan:
cd cli
pipenv install --dev
pipenv shell
time semgrep --help
=> 0.090s
time pysemgrep --help
=> 0.368s
semgrep --experimental scan --help
=> osemgrep man page
semgrep scan --help
=> pysemgrep man page as we default back to pysemgrep
semgrep interactive -l ocaml .
=> works! without --experimental

PR checklist:


aryx:
@brandonspark you can now simply do 'semgrep interactive -l ...', no need --experimental anymore.

brandonspark:
wooh!

cli/bin/semgrep
```diff
@@ -66,38 +72,42 @@ def find_semgrep_core_path():
...
+# pysemgrep and it's better to avoid the possibility of an infinite loop
+# by simply using a different program name. Morever, in case of big problems,
+# we can always tell users to run pysemgrep instead of semgrep and be sure
+# they'll get the old behavior.
+def call_osemgrep():
+    path = find_semgrep_core_path()
+    # If you call semgrep-core as osemgrep, then we get osemgrep behavior,
+    # see src/main/Main.ml
+    sys.argv[0] = "osemgrep"
+    os.execvp(str(path), sys.argv)
```
ajbt200128, line 90: I think that a lot of the errors in the tests are because we're not passing environment variables here, which is important to not break semgrep in CI.

cli/bin/semgrep
```diff
@@ -66,38 +72,42 @@ def find_semgrep_core_path():
...
-        # tell users to run pysemgrep instead of semgrep and be sure they'll get the
-        # old behavior.
-        path = find_semgrep_core_path()
-        # If you call semgrep-core as osemgrep, then we get osemgrep behavior,
-        # see src/main/Main.ml
-        sys.argv[0] = "osemgrep"
-        os.execvp(str(path), sys.argv)
+    if "--legacy" in sys.argv:
+       sys.argv.remove("--legacy")
+       #TODO? or execvp("pysemgrep", sys.argv)?
```
ajbt200128: No, we should not call execvp here, it will break the fix we just made for tests + autocomplete. I tried adding a longer explanation in that PR but the short answer is calling execvp from python to run more python breaks things

cli/bin/semgrep
```diff
@@ -66,38 +72,42 @@ def find_semgrep_core_path():
...
-        # a different program name. Morever, in case of big problems, we can always
-        # tell users to run pysemgrep instead of semgrep and be sure they'll get the
-        # old behavior.
-        path = find_semgrep_core_path()
-        # If you call semgrep-core as osemgrep, then we get osemgrep behavior,
-        # see src/main/Main.ml
-        sys.argv[0] = "osemgrep"
-        os.execvp(str(path), sys.argv)
+    if "--legacy" in sys.argv:
+       sys.argv.remove("--legacy")
```
ajbt200128, line 110: We should make sure we add this to the documentation right away, so users know how to get pysemgrep if things start regressing

cli/bin/semgrep
ajbt200128: Actually checking the docs, it seems execvp will inherit environment variables by default so I think this is fine

src/osemgrep/cli/CLI.ml
```diff
@@ -137,22 +142,32 @@ let dispatch_subcommand argv =
...
+      | "shouldafound" when experimental -> missing_subcommand ()
+      (* TODO: next target for not requiring the 'when experimental' guard! *)
+      | "lsp" when experimental -> Lsp_subcommand.main subcmd_argv
+      (* osemgrep-only: and by default! no need experimental! *)
       | "interactive" -> Interactive_subcommand.main subcmd_argv
       (* LATER: "dump", "test", "validate" *)
-      | _else_ -> (* should have defaulted to 'scan' above *) assert false)
+      | _else_ ->
+          if experimental (* should have defaulted to 'scan' above *) then
+            assert false
```
ajbt200128: Can we make this a better error message than assert false since this is now exposed to end users?

ajbt200128 (commented):
I know I said I would try and switch the python tests to calling semgrep directly, instead of through the pylib, but I feel that change might make sense in this PR (or before). I worry that there is some edge case (something in CI that requires environment variables or something like that) that we're not going to catch without testing by calling it this way.

I also wonder if maybe it's worth making the CLI.ml changes (i.e. osemgrep falls back to pysemgrep), and trying to dog food first by switching everywhere we use semgrep to `semgrep --experimental`.

brandonspark:
seems fine to me, i'll wait on another's approval though

cli/bin/semgrep
aryx: yes, children inherits the environment of the parent (UNIX got many things right!)

aryx:
Good comments @ajbt200128 . I agree getting e2e tests to use semgrep is high-priority!
Also here we very quickly dispatch back to pysemgrep for everything (except interactive, which anyway is not supported in pysemgrep), so I think we should be fine.

aryx:
@ajbt200128 any idea why the test_ls.py tests are now failing?

cli/bin/semgrep
aryx: I've added a bigger comment

src/osemgrep/cli/CLI.ml
aryx: no it should never happen, but I can make an error message (even though it should never happen).

cli/bin/semgrep
aryx: I've added a changelog entry.

cli/bin/semgrep
```diff
@@ -4,7 +4,9 @@
 # the Semgrep HomeBrew package, and the Semgrep Docker container.
 #
 # In the futur we may have different entry points when packaging Semgrep
```
emjin:
```suggestion
# In the future we may have different entry points when packaging Semgrep
```

cli/bin/semgrep
```diff
@@ -30,22 +33,25 @@ import shutil
...
 # but setuptools and pip adjust this line when installing semgrep so we need
-# this instead.
+# to do this instead.
 import warnings
 warnings.filterwarnings("ignore", category=DeprecationWarning)
 
 # Add the directory containing this script in the PATH, so the pysemgrep
 # script will also be in the PATH.
 # Some people don't have semgrep in their PATH and call it instead
 # explicitely as in /path/to/somewhere/bin/semgrep, but this means
```
emjin:
```suggestion
# explicitly as in /path/to/somewhere/bin/semgrep, but this means
```

src/osemgrep/cli/CLI.ml
```diff
@@ -137,22 +144,35 @@ let dispatch_subcommand argv =
...
+       *)
+      | "ci" when experimental -> Ci_subcommand.main subcmd_argv
+      | "install-semgrep-pro" when experimental -> missing_subcommand ()
+      | "login" when experimental -> Login_subcommand.main subcmd_argv
+      | "logout" when experimental -> Logout_subcommand.main subcmd_argv
+      | "publish" when experimental -> missing_subcommand ()
+      | "scan" when experimental -> Scan_subcommand.main subcmd_argv
+      | "shouldafound" when experimental -> missing_subcommand ()
+      (* TODO: next target for not requiring the 'when experimental' guard! *)
+      | "lsp" when experimental -> Lsp_subcommand.main subcmd_argv
```
ajbt200128, line 165: Ahh ok, tests are failing for the language server because before we would run pysemgrep -> call `osemgrep lsp` and use it as a subprocess. now this never happens, I'll add a fix

IagoAbal:
How much does this add to the start-up time of `semgrep` ?

cli/bin/semgrep
```diff
@@ -62,42 +68,57 @@ def find_semgrep_core_path():
...
+       # but this causes some regressions with --test (see PA-2963)
+       # and autocomplete (see #8359)
+       #TODO: we should get rid of autocomplete anyway (it's a Python Click
+       # thing not supported by osemgrep anyway),
+       #TODO: we should fix --test instead.
+       # The past investigation of Austin is available in #8360 PR comments
+       import semgrep.__main__
+       sys.exit(semgrep.__main__.main())
+    elif "--experimental" in sys.argv:
+       call_osemgrep()
```
emjin: For my understanding: why does this not need `sys.exit`?

cli/src/semgrep/lsp/server.py
```diff
@@ -33,7 +33,7 @@ def __init__(self) -> None:
         self.polling: bool = True
 
     def start_ls(self) -> None:
-        cmd = ["osemgrep", "lsp"]
+        cmd = ["osemgrep", "lsp", "--experimental"]
```
emjin: Why does this need `--experimental`?

aryx:
I would say: 0.090 
 - 0.15 for the Python startup time of cli/bin/semgrep (does not use much import so it's fast)
 - 0.75 for the OCaml startup time

cli/src/semgrep/lsp/server.py
aryx: This need to be "semgrep", not "osemgrep", because osemgrep is not part of the pip package nor Docker image.

src/osemgrep/cli_scan/Scan_CLI.ml
```diff
@@ -759,11 +761,12 @@ let cmdline_term ~allow_empty_config : conf Term.t =
           Rules_source.Pattern (pat, Some xlang, fix)
       | _, (Some pat, None, fix) -> (
           match common.maturity with
-          | Some CLI_common.Legacy ->
+          (* osemgrep-only: better: can use -e without -l! *)
+          | Some CLI_common.MDevelop -> Rules_source.Pattern (pat, None, fix)
+          | None
+          | Some (CLI_common.MExperimental | CLI_common.MLegacy) ->
```
emjin, line 767: Why does this need to error for MExperimental?

cli/src/semgrep/lsp/server.py
aryx: In fact I don't even know how this can work ..., what is the path of this osemgrep and how our test infrastructure can find this osemgrep.

src/osemgrep/cli_scan/Scan_CLI.ml
emjin: Edit: I see. Maybe we could have a better name, e.g. MLegacyCompatible, or we could do something like MExperimental, MLegacy, MPySemgrep

cli/src/semgrep/lsp/server.py
ajbt200128: we're not calling osemgrep, we're calling semgrep-core and passing the above as arguments. I'll rename `cmd` to `args`. This is functioning the same as `call_osemgrep`, but looks different because we're using Popen and not os.exec here so we can keep it as a child process

cli/src/semgrep/lsp/server.py
```diff
@@ -33,10 +33,10 @@ def __init__(self) -> None:
         self.polling: bool = True
 
     def start_ls(self) -> None:
-        cmd = ["osemgrep", "lsp"]
+        args = ["osemgrep", "lsp", "--experimental"]
```
emjin, line 36: Why do we need to add `--experimental`?

emjin: I now understand why; would prefer better names for these flags

emjin (approved):
It looks good to me; I don't know much about the exec-ing but the tests pass. I would really prefer better names for the flags `--develop`, `--experimental`, `--legacy`, however. `--experimental` does not do what I intuitively expect

cli/bin/semgrep
ajbt200128: os.exec is a thin wrapper to the OS's exec, which starts the process in place of the caller.

ajbt200128: i.e. not as a child process, it just replaces it

emjin:
> Also here we very quickly dispatch back to pysemgrep for everything (except interactive, which anyway is not supported in pysemgrep), so I think we should be fine.

I think the question here is about whether the tests are indeed calling this path, or if they're directly invoking pysemgrep and skipping this exec flow. If it's the latter, I'm not as comfortable merging without adding some tests specifically for this flow

cli/src/semgrep/lsp/server.py
ajbt200128: before, calling `osemgrep subcommand` would just run osemgrep's implementation of that subcommand. Now, osemgrep calls back to pysemgrep by default, and only runs it's own implementation of a subcommand if --experimental is passed. Once I port login stuff, we will be able to delete this whole file and just use osemgrep directly

ajbt200128:
@emjin yea right now the tests call the invoke semgrep function on the python side. None of the tests handle anything with the execution flow and I have the same worries hence the comment. Note that messing with the entrypoint stuff previously is what broke the test subcommand and autocomplete, as it wasn't caught by tests.

emjin (changes requested):
Did not realize this didn't go through automated tests; we should have at least one test for this entry point

aryx:
Which test do you propose?

emjin:
Can we set one up that calls semgrep and doesn't dispatch directly to the Python? From there I don't care what it does, a basic rule on basic file is fine

aryx:
Making our e2e test switch from calling semgrep via internal python function to forking semgrep is the right long-term solution, but will require a different PR like I said to Austin above. This will also slow down quite a lot 'make test' (which I'm fine with; better to be correct than fast but incorrect).

aryx:
And why this new e2e test would give you more confidence?
I can start an e2e test infra in ocaml (instead of pytest), but since most of our e2e tests are using pytest (and calling invoke_semgrep() direcly), not sure one ocaml e2e test will improve things a lot.

aryx:
BTW a few of our CI tests were failing when I made this PR, meaning there are few places where we exercise
this cli/bin/semgrep binary (including test_ls.py, our Dockerfile, etc.).

emjin (approved):
Ok, I did not have clarity on how the semgrep tests work, so I went and checked.

In semgrep_runner.py:
```
# The semgrep command suitable to run semgrep as a separate process.
# It's something like ["semgrep"] or ["python3"; -m; "semgrep"] or
# ["/path/to/osemgrep"].
SEMGREP_BASE_COMMAND: List[str] = (
    [OSEMGREP_PATH] if USE_OSEMGREP else [sys.executable, "-m", "semgrep"]
)

SEMGREP_BASE_COMMAND_STR: str = " ".join(SEMGREP_BASE_COMMAND)
```

This looks to me like it's executing `semgrep` (or `osemgrep`), and it's used in `test_check`, including at least one matching test. That satisfies what I was asking for.

ajbt200128:
Oh good catch, I didn't know test_check calls semgrep directly in some of the tests

ajbt200128:
Actually it does not, `python -m semgrep` is different it imports the python module, and does not call the entrypoint :(

ajbt200128:
Just confirmed, breaking osemgrep does not break any python tests :( (as suggested by emma, I changed the pysemgrep callback in CLI to something random)

IagoAbal:
I am not sure why but running benchmarks to compare develop vs this PR, as well as 1.33.2 vs this PR, I get one more finding with this PR when analyzing `gitlabhq`:

```diff
 ┌───────────────────┐
-│ 113 Code Findings │
+│ 114 Code Findings │
 └───────────────────┘
+                                                                                
+    repos/gitlabhq/app/assets/javascripts/issues/create_merge_request_dropdown.j
+  s                                                                             
+       javascript.browser.security.open-redirect-from-function.js-open-redirect-
+  from-function                                                                 
+          The application accepts potentially user-controlled input `{ data }`  
+  which can control the                                                         
+          location of the current window context. This can lead two types of    
+  vulnerabilities open-                                                         
+          redirection and Cross-Site-Scripting (XSS) with JavaScript URIs. It is
+  recommended to                                                                
+          validate user-controllable input before allowing it to control the    
+  redirection.                                                                  
+          Details: https://sg.run/P39e                                          
+                                                                                
+          161┆ window.location.href = data.url;
```

Could it be because this PR needs to be rebased ?

ajbt200128:
Ok, I updated `semgrep_runner.py` to actually call the correct entrypoint. So it should work how Emma expects now, and I confirmed breaking osemgrep breaks some tests in `test_check.py`

ajbt200128:
Ok I think the failing tests are now due to the actual changes in this PR

ajbt200128:
It looks like those tests were failing because path would be empty, and so python couldn't find the semgrep module when running them. I think it's fair to assume that python's module directory will be on the path, as is the case with literally every unix computer that has python installed

aryx:
good investigation @emjin. TIL we were already doing real e2e testing in test_check.py

IagoAbal:
False alarm, same number of findings now. 👍

PR #8373, opened 2023-07-25
feat(ssc): Surface Child Dependencies for Maven

SayyantR:
This PR is part of the [Path to Transitivity](https://www.notion.so/semgrep/Path-to-Transitivity-a1539071c8074acf986de73f4cc6778d) project. In this PR, we introduced some logic to parse each dependency's child and modify the `FoundDependency` class to now include the list of packages that the dependency depends on. This data is then sent to the app via the `/complete` endpoint where we will construct the adjacency list (see scope doc linked above).

PR checklist:


cli/src/semdep/parsers/pom_tree.py
```diff
@@ -66,6 +75,51 @@
     << string("\n").optional()
 )
 
+# Dictionaries are not hashable, so we need to convert them to objects which are
+# This will speed up the next iteration that takes place where we remove duplicates
+def dict_to_obj(d: Dict[str, Any]) -> ParsedDependency:
```
mmcqd: Why not just create these in the first place? You don't need dictionaries at all

cli/src/semdep/parsers/pom_tree.py
```diff
@@ -66,6 +75,51 @@
...
+    return ParsedDependency(
```
mmcqd: Considering this has some maven-specific data, I'd probably define this within this file

cli/src/semdep/parsers/util.py
```diff
@@ -318,6 +320,38 @@ def safe_parse_lockfile_and_manifest(
     return parsed_lockfile, parsed_manifest, errors
 
 
+class ParsedDependency:
```
mmcqd, line 323: This should be a `@dataclass` so that you don't have to write an `__init__` method

cli/tests/e2e/test_dependency_aware_rules.py
```diff
@@ -293,3 +294,37 @@ def test_no_lockfiles(run_semgrep: RunSemgrep, monkeypatch, tmp_path, snapshot):
...
+
+# These are file targets which are ecosystems that we can calculate the path to
+# transitivity for
+@pytest.mark.parametrize(
+    "target_supports_path_to_transitivity",
+    [
+        "targets/dependency_aware/log4j/maven_dep_tree.txt",
+    ],
+)
+def test_child_construction(
```
mmcqd: I think I'd just move this into the `test_paths_to_transitivity` file, and maybe move that to the unit test folder.

mmcqd (changes requested):

mmcqd (dismissed):

mmcqd (approved):

PR #8376, opened 2023-07-25
fix(test): fix php parsing errors in test

ihji:
PR checklist:

### DIFF
--- a/tests/rules/metavar_regex_include.php
+++ b/tests/rules/metavar_regex_include.php
@@ -1,5 +1,5 @@
 <?php
 
 //ruleid: xxx
-include("foobar")
+include("foobar");
 
--- a/tests/rules/taint_wo_side_effects2.php
+++ b/tests/rules/taint_wo_side_effects2.php
@@ -4,4 +4,4 @@
 echo $_SERVER['REQUEST_URI'];
 
 // ok:regression_0121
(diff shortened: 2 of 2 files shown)


akuhlens (approved):
This looks fine to me. One of the parse errors seems to marked to succeed and the other to fail. I would be interested to see if that needs changing too. 

Also there is the question of what has changed to surface this error, and why has it just now changed? It seems like it didn't use to cause a parse failure to forget a semi colon. I wonder why it does now.

ihji:
> This looks fine to me. One of the parse errors seems to marked to succeed and the other to fail. I would be interested to see if that needs changing too.
> 
> Also there is the question of what has changed to surface this error, and why has it just now changed? It seems like it didn't use to cause a parse failure to forget a semi colon. I wonder why it does now.

The parsing error appears to have been present for months, but it has recently become noticeable. While I'm not sure of the root cause, my priority is to fix the error since it is causing issues with local testing for developers. The PHP code without the semicolon is considered invalid by the PHP tree sitter grammar, so it definitely needs to be fixed.

ihji:
Merged with the `benchmarks-full` testing error, as it is unrelated to the current change.

aryx:
yes would be great to know why the error got triggered just now ... What changed recently that caused that to suddently happen????

PR #8371, opened 2023-07-25
simplify and consolidate linux wheels

tpetr:
PR checklist:

- [ ] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [ ] Tests included or PR comment includes a reproducible test plan
- [ ] Documentation is up-to-date
- [ ] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [ ] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #8378, opened 2023-07-26
short-term fix for amd64 docker build failing

tpetr:
Skip python wheel validation if no .whl files match the glob passed to validate-wheels.sh. I'm working on a better fix in https://github.com/returntocorp/semgrep/pull/8371

@aryx


aryx (approved):
I don't understand anything, but if it unbreaks my build ...
Would be good to do a longer-term fix though.

PR #8381, opened 2023-07-27
feat(eval): add parse_promql_duration comparison function

MichaHoffmann:
The approach to parse a duration is a bit gross but seemed to work well enough in tests. The function name is debatable since this is only interesting in duration metavariables captured from promql i think.

PR checklist:


aryx:
Until now we've tried in Eval_generic to use the name of Python functions, as most people are familiar with Python so it's a good starting point. Dunno what's the corresponding function in Python for this "milliseconds" but we can always refine that later in a separate PR.

aryx:
At some point you also need to make a PR for our semgrep-docs so we update also:
https://semgrep.dev/docs/writing-rules/rule-syntax/#metavariable-comparison

aryx:
Please add a changelog.d/ entry for this. Every user-facing improvement should be added in the changelog so users can learn about it.

MichaHoffmann:
> Until now we've tried in Eval_generic to use the name of Python functions, as most people are familiar with Python so it's a good starting point. Dunno what's the corresponding function in Python for this "milliseconds" but we can always refine that later in a separate PR.

there is no corresponding python function really in this case! its also only useful for promql durations.

aryx:
See https://github.com/returntocorp/semgrep/pull/8053 for example of changelog.d entry

src/engine/Eval_generic.ml
```diff
@@ -236,6 +278,13 @@ let rec eval env code =
       | __else__ -> raise (NotHandled code))
   | G.Call ({ e = G.N (G.Id (("today", _), _)); _ }, (_, _, _)) ->
       Float (Unix.time ())
+  (* Convert duration string to integer milliseconds*)
```
aryx: maybe add a comment then here saying there's no equivalent python function so we use the name used
in prometheus (with a link to the prometheus doc).

MichaHoffmann: i dont think there is a corresponding go function too ( or maybe `time.ParseDuration(...).Milliseconds()` which is a bit of a mouthfull)

aryx (approved):

PR #8384, opened 2023-07-27
Sophia/lambda calc substitution for jsonnet

SophiaSR:
Substitution model implementation of jsonnet interpreter.
PR checklist:


libs/ojsonnet/Core_jsonnet_subst.ml
```diff
@@ -0,0 +1,163 @@
+(* Yoann Padioleau
```
aryx: What is the difference with Core_jsonnet.ml? seems like exactly the same content, can you get rid of it?

libs/ojsonnet/Desugar_jsonnet_subst.ml
```diff
@@ -0,0 +1,505 @@
+(* Yoann Padioleau
```
aryx: Again what's the difference with the original?
I was thinking we just need to have a Value_jsonnet_subst.ml and Eval_jsonnet_subst.ml.
We should try to keep Desugar and Core as-is, or if you need subst specific stuff to either
make the 2 original types polymorph or something so we can reuse them or clearly mark in the code
where the subst variant is different from the original code.

libs/ojsonnet/Eval_jsonnet_subst.ml
```diff
@@ -0,0 +1,913 @@
...
+  | BinaryOp (el, (op, tk), er) ->
+      BinaryOp (substitute id sub el, (op, tk), substitute id sub er)
+  | If (tif, e1, e2, e3) ->
+      If (tif, substitute id sub e1, substitute id sub e2, substitute id sub e3)
+  | Error (tk, e) -> Error (tk, substitute id sub e)
+  | ExprTodo ((_s, _tk), _ast_expr) -> ExprTodo ((_s, _tk), _ast_expr)
+
+(*TODO*)
+
+let rec substitute_kw kw sub expr =
```
aryx, line 220: maybe add in comment reference to the spec, which part of the spec this function refers to or implement?

aryx: So then one can look at the code and find the corresponding part in the spec and check if they match.

aryx:
Nice work!
I think there's lots of small things we should fix, but let's merge this in; we can always iterate from there.

aryx (approved):

PR #8406, opened 2023-08-01
feat(comparison): add in and not in for strings

brandonspark:
## What:
This PR makes it so that you can use `in` and `not in` in the Python sense, for strings.

## Why:
Feature request from community, see https://semgrep.slack.com/archives/C018NJRRCJ0/p1690862677954849

## How:
Just added the operators, and tests. Cleaned up some code while I was at it.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/src/engine/Eval_generic.ml
+++ b/src/engine/Eval_generic.ml
@@ -334,12 +334,17 @@ and eval_op op values code =
   (* TODO? dangerous use of polymorphic <> ? *)
   | G.NotEq, [ v1; v2 ] -> Bool (v1 <> v2)
   | G.In, [ v1; v2 ] -> (
-      match v2 with
-      | List xs -> Bool (List.mem v1 xs)
+      match (v1, v2) with
+      | _, List xs -> Bool (List.mem v1 xs)
+      | String v1, String v2 ->
+          Bool (Common2.string_match_substring (Str.regexp v1) v2)
       | _ -> Bool false)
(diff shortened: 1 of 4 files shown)


src/engine/Eval_generic.ml
```diff
@@ -334,12 +334,17 @@ and eval_op op values code =
   (* TODO? dangerous use of polymorphic <> ? *)
   | G.NotEq, [ v1; v2 ] -> Bool (v1 <> v2)
   | G.In, [ v1; v2 ] -> (
-      match v2 with
-      | List xs -> Bool (List.mem v1 xs)
+      match (v1, v2) with
+      | _, List xs -> Bool (List.mem v1 xs)
+      | String v1, String v2 ->
+          Bool (Common2.string_match_substring (Str.regexp v1) v2)
```
aryx: So v1 is a regular expression? In python you can do ".*" in "abc" and it returns true? I just tried and it's not the case
so you probably want to use Str.regexp_string instead of Str.regexp.

aryx (changes requested):

src/engine/Eval_generic.ml
brandonspark: good catch! didn't realize the difference. fixed now.

aryx (approved):

ievans:
@brandonspark is there a corresponding update to the docs?

brandonspark:
@ievans There is not. I'll add it, thanks!

PR #8414, opened 2023-08-02
tainting: Add `by-side-effect: only` option to sources

IagoAbal:
This specifies a taint source that _only_ propagates taint by side-effect, which is useful to write "double-op" kind of rules (e.g., double-free).

Closes PA-2980
Closes PA-3124

test plan:
make test # new tests

### DIFF
--- a/changelog.d/pa-2980.added
+++ b/changelog.d/pa-2980.added
@@ -1,0 +1,33 @@
+taint-mode: The `by-side-effect` option for taint sources (only) now accepts a
+third value `only` (besides `true` and `false`). Setting `by-side-effect: only`
+will define a taint source that *only* propagates by side effect. This option
+should allow (ab)using taint-mode for writing some typestate rules.
+
+For example, this taint rule:
+
+```yaml
+pattern-sources:
+  - by-side-effect: only
+    patterns:
+    - pattern: lock($L)
+    - focus-metavariable: $L
+pattern-sanitizers:
+  - by-side-effect: true
+    patterns:
+    - pattern: unlock($L)
+    - focus-metavariable: $L
+pattern-sinks:
+  - pattern: lock($L)
+```
+
+will match the second `lock(x)` in this code:
+
+```python
+lock(x) # no finding
+lock(x) # finding
+```
+
+The first `lock(x)` will not result in any finding, because the occurrence of `x` in
+itself will not be tainted. Only after the function call we will record that `x` is
+tainted (as a side-effect of `lock`). The second `lock(x)` will result in a finding
+because the `x` has been tainted by the previous `lock(x)`.
--- a/src/tainting/Dataflow_tainting.ml
+++ b/src/tainting/Dataflow_tainting.ml
@@ -340,10 +340,17 @@ let merge_source_sink_mvars env source_mvars sink_mvars =
     (* The union of both sets, but taking the sink mvars in case of collision. *)
     sink_biased_union_mvars source_mvars sink_mvars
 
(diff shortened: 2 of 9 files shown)


src/tainting/Dataflow_tainting.ml
```diff
@@ -340,10 +340,17 @@ let merge_source_sink_mvars env source_mvars sink_mvars =
     (* The union of both sets, but taking the sink mvars in case of collision. *)
     sink_biased_union_mvars source_mvars sink_mvars
 
-let partition_mutating_sources sources_matches =
+let partition_sources_by_side_effect_oyn sources_matches =
```
amchiclet, line 343: `partition_sources_by_side_effect_only`? (s/oyn/only)

changelog.d/pa-2980.added
```diff
@@ -0,0 +1,33 @@
...
+    - focus-metavariable: $L
+pattern-sinks:
+  - pattern: lock($L)
+```
+
+will match the second `lock(x)` in this code:
+
+```python
+lock(x) # no finding
+lock(x) # finding
```
amchiclet, line 27: I'm wondering if this should be the actual expected behavior when `by-side-effect: true` or not. That way, we wouldn't need to introduce another enum at all. (But if it breaks backward-compatibility, maybe it's not a good thing.)

Personally, I would expect `x` to be tainted, only after the first `lock(x)`.

Is there an example, where the nuanced difference between `true` vs `only` is needed?

Also here's another naive comment.

If I understand correctly, the term `only` here means "propagate only, don't sink". But at first, I read `only` as "only side-effects propagate the taint, and things like assignments don't".

I'm wondering if there's some less ambiguous word or not, but can't think of a good one.

Some ideas
* `by-side-effect: after` or
* `by-side-effect: after-statement` or
* something else that says that the side-effect doesn't happen just yet

amchiclet (approved):
I have some high-level comments/questions, but not sure if they makes sense or not.

So I'm going ahead approving the PR in case you'd like to move forward without any changes.

ihji (approved):

src/tainting/Dataflow_tainting.ml
ihji: seems like `oyn` is short for `only/yes/no`.

amchiclet: You're right. I'll resolve this.

changelog.d/pa-2980.added
IagoAbal: > I'm wondering if this should be the actual expected behavior when by-side-effect: true or not.

Makes sense, but that indeed would break backwards compatibility. 🤔 I can give it a try and see if we break anything in our tests. I would still feel a bit uneasy about breaking other people's rules, that's why I went for adding `only`. I also wonder if there are rules that do need what `by-side-effect: true` provides right now. It's quite easy to make wrong assumptions based on limited info.

> I read only as "only side-effects propagate the taint, and things like assignments don't".

I read it as "this source only produces taint by side-effect", indeed, not by assignment. Although once the taint is produced, it can indeed propagate by assignment. But naming is difficult and I'm not the best at picking names. I can ask SR for feedback.

IagoAbal: I gave it a try to just make `true` the same as `only` but it does break some tests/rules, when taint labels are involved. In some cases I can rewrite the rule easily, but in one case a taint label is used as a "cleanser" (sanitizer) and I think we then need the combination of "returning taint on-site + updating the l-value by side-effect" too.

PR #8417, opened 2023-08-02
nit: Use semgrep for program name

zzeleznick:
## Description

Running different semgrep commands was showing different program names which was a little confusing (to me).

```
./cli/bin/semgrep publish --help                                                                                                                        
Usage: pysemgrep publish [OPTIONS] TARGET
```

vs.

```
./cli/bin/semgrep --help                                                                                                                        
Usage: semgrep [OPTIONS] COMMAND [ARGS]...
```

See https://semgrepinc.slack.com/archives/C01NXGX2EHZ/p1691013559954119

With this change

```
./cli/bin/semgrep publish --help                                                                                                                        
Usage: semgrep publish [OPTIONS] TARGET
```

### DIFF
--- a/cli/src/semgrep/__main__.py
+++ b/cli/src/semgrep/__main__.py
@@ -3,7 +3,10 @@ import sys
 
 from semgrep.cli import cli
 
-
+# NOTE: This is the entrypoint for the `pysemgrep` command.
+# To match the program usage help between pysemgrep (legacy)
+# and osemgrep (new) – and to hide complexity for our users -
+# here we specify `semgrep` as the program name for pysemgrep.
 def main() -> int:
     cli(prog_name="semgrep")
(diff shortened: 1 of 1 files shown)


cli/src/semgrep/__main__.py
```diff
@@ -5,7 +5,7 @@
 
 
 def main() -> int:
-    cli()
+    cli(prog_name='semgrep')
```
aryx: maybe add a comment like # this code is not run anymore from cli/bin/semgrep but from cli/bin/pysemgrep (exec'ed from cli/bin/semgrep) because of osemgrep, hence the prog_name adjustment to avoid listing 'pysemgrep for arv0'

aryx (approved):

PR #8423, opened 2023-08-03
Use env model when handling std library functions to speed up evaluation in jsonnet

SophiaSR:
In the substitution model of eval, we just always pass an empty environment now, so that we can use the same value file for both. With that, we can then call the environment model evaluation for std library functions so that they don't take as long. We want to use substitution model for other things since it implements more features correctly, but since std library does not use feature that aren't implemented correctly in the environment model, it is safe to revert to it just for std library.

Test plan: added a timer for evaluation then manifestation in Test_ojsonnet.ml.
When running this on tests/jsonnet/eval_pass/array_comprehension2.jsonnet with the purely lambda calculus version, we get an estimate of ~11.0 seconds to run. 
When running on this version, we get ~0.1 seconds.
PR checklist:


libs/ojsonnet/Eval_jsonnet_subst.ml
```diff
@@ -1,13 +1,24 @@
 open Core_jsonnet
 open Common
-module V = Value_jsonnet_subst
+module V = Value_jsonnet
 module A = AST_jsonnet
 module J = JSON
 
 exception Error of string * Tok.t
 
 type cmp = Inf | Eq | Sup
```
aryx, line 9: would be good to  factorize code as much as possible with Eval_jsonnet.ml, for example here reuse the cmp type
from the other file.

libs/ojsonnet/Eval_jsonnet_subst.ml
```diff
@@ -1,13 +1,24 @@
...
 
+(*Creates std so that we can add it to the environment when we switch back to
+  * environment model for handling standard library functions
+*)
+let pre_std = Std_jsonnet.get_std_jsonnet ()
+
+let path =
+  match Fpath.of_string "../" with
```
aryx, line 18: I don't understand this code. What is path? Seems like it's aways .. so maybe the variable
should be renamed to let parent ?

libs/ojsonnet/Eval_jsonnet_subst.ml
```diff
@@ -581,7 +593,19 @@ and eval_std_cmp tk (el : expr) (er : expr) : cmp =
...
-  match eval_expr e0 with
+  (* Check if this is a standard library call. If it is, call the environment
+     *   model evaluation to get more efficiency *)
+  let eval_func =
+    match e0 with
+    | ArrayAccess
+        (Id ("std", _), (_, L (Str (None, DoubleQuote, (_, [ _ ], _))), _)) ->
+        (* set locals so that "std" shows up in the environment when evaluating *)
+        let local_wrap = Local (fk, [ B (("std", fk), fk, std) ], fk, e0) in
+        print_string "going into eval_jsonnet";
```
aryx: Need to remove this print_string. You can use Logs if you want to log things.

libs/ojsonnet/Value_jsonnet.ml
```diff
@@ -28,7 +28,15 @@ module A = AST_jsonnet
...
 type env = {
-  (* The spec uses a lambda-calculus inspired substitution model, but
+  (* There are currently two implementations of evaluation, one in
+   * Eval_jsonnet, the other in Eval_jsonnet_subst. The former
+   * uses an environment mode, while the later uses lambda calc
+   * substitutions. Currently, the lambda calc version passes more
+   * tests but is inneficient. Keeping both implementations to
+   * possibly mix the two, to get the efficiency benifit of environments
+   * while keeping the simpler super/self implementation given
+   * by substitution model. In the substitution model, we always
```
aryx, line 38: I wonder if we could use Eval_jsonnet.ml even more; like handling self/super with lambda calclus but everything else with the old environment-based evaluator.

aryx (approved):
Would be good in a test plan to have a simple bench that shows
the speedup, like on XXX.jsonnet, we go from 3s to 0.1 s thx to this specialization for std.

libs/ojsonnet/Eval_jsonnet_subst.ml
SophiaSR: The filepath is a placeholder (we don't actually care about what it is, since it's not referenced, just there to make the types work out). I can add a comment explaining that!

libs/ojsonnet/Value_jsonnet.ml
SophiaSR: Yeah, I think I'll try to do that way next. Wanted to get a version of mixing the two models working that sped up the current bottleneck in tests first.

aryx (approved):

PR #8427, opened 2023-08-08
Python import weirdness

ajbt200128:
See #8360 for the initial issue. I'm 99% sure #8370 broke #8360 since the first PR was merged, then the 2nd, and they were in the same release. I spent ~30 minutes trying to understand this issue after I fixed it, and left a comment explaining why I think this fixes it, but I couldn't find real documentation on this and I suspect only a deep understanding of the import system, GIL, and how multiprocessing works could explain this. I think we should be fine until osemgrep, but we should be extra careful about touching these files until then.

One note is that when we support/switch to the scan command from osemgrep, we'll need to be careful that we handle all the other places semgrep is invoked on the python side (such as test).

## Test plan

```bash
cd cli && pipenv install
pipenv shell
cd EXTRACT_PATH #path of archive from linear issue
semgrep --legacy --test --pro --config=. # all 4 tests pass
semgrep --test --pro --config=. # all 4 tests pass
```

PR checklist:

### DIFF
--- a/cli/tests/e2e/test_cli_test.py
+++ b/cli/tests/e2e/test_cli_test.py
@@ -1,7 +1,11 @@
+import os
 import re
+import subprocess
+from pathlib import Path
 
 import pytest
 from tests.fixtures import RunSemgrep
+from tests.semgrep_runner import SEMGREP_BASE_COMMAND
 
 from semgrep.constants import OutputFormat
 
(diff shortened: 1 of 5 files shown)


emjin (approved):
Are there tests we could add for this?

ajbt200128:
I totally forgot to write a test, oops, I'll see if I can write one that works

cli/tests/e2e/test_cli_test.py
```diff
@@ -147,3 +151,26 @@ def test_parse_errors(run_semgrep_in_tmp: RunSemgrep, snapshot):
...
+@pytest.mark.slow
+def test_cli_test_from_entrypoint(snapshot):
+    env = {}
+    env["PATH"] = os.environ.get("PATH", "")
+
+    cmd = SEMGREP_BASE_COMMAND + [
+        "--test",
+        "--config",
+        "rules/cli_test/multiple_annotations/multiple-annotations.yaml",
+        "targets/cli_test/multiple_annotations/multiple-annotations-bad.py",
```
aryx, line 165: why are you using multiple-annotations-bad.py here?
Looks like this file was not even used in a test before ...

aryx: Why did you write this test? why using --test? This PR does not seem to have anything to do with --test ...

aryx: Fixed here I think https://github.com/semgrep/semgrep/pull/10189

PR #852, opened 2020-05-28
Optimize deep statement matching

aryx:
This should close #827 and #664

The code to handle foo(); ...; bar(); was very naive
and was doing lots of useless work. This fixes that.

Test plan:
time pipenv run semgrep -f ~/semgrep/tests/PERF/ajin.yaml ~/semgrep/tests/PERF/three.js
3.2s

(was 3min before)

+ /home/pad/github/semgrep/semgrep-core/_build/default/bin/Main.exe -profile -lang py -f tests/PERF/ellipsis-python.sgrep tests/PERF/my_first_calculator.py
---------------------
profiling result
---------------------
Main total                               :      1.670 sec          1 count
Parse_python.parse                       :      0.918 sec          1 count
Parse_python.tokens                      :      0.525 sec          2 count
Semgrep.check                            :      0.458 sec          1 count
Parser_python.main                       :      0.277 sec          1 count
Semgrep.match_sts_sts                    :      0.186 sec      41627 count

(was 85sec before)

+ /home/pad/github/semgrep/semgrep-core/_build/default/bin/Main.exe -profile -lang js -f tests/PERF/ellipsis-js.sgrep tests/PERF/three.js
---------------------
profiling result
---------------------
Main total                               :      2.151 sec          1 count
Parse_js.parse                           :      1.236 sec          1 count
Parse_js.tokens                          :      0.398 sec          2 count
Semgrep.check                            :      0.389 sec          1 count
Semgrep.match_sts_sts                    :      0.239 sec      16824 count
Parser_js.module_item                    :      0.192 sec        609 count

(was a lot more before)

PR #8440, opened 2023-08-10
matching: ellipsis: Always try going deep

### LINKED ISSUE #660: python.lang.maintainability.useless-inner-function false positives
https://semgrep.live/7KvR?registry=python.lang.maintainability.useless-innerfunction&sourceurl=https%3A%2F%2Fraw.githubusercontent.com%2Fdropbox%2Fdropbox-api-v2-repl%2Fa057a71995e12fb4cd2fff6fa4d917969e665fda%2Frepl.py

IagoAbal:
Pattern `... bar()` would behave strangely because it would only try to match deeply if it did not match anything non-deeply. So depending on how the target code looked, `... bar()` would match or not match inside e.g. an `if` statement. This was confusing and not a great property for Semgrep to have.

The regression was introduced in PR #852 as part of a set of optimizations done back in 0.9.0 (!), but at present reverting this one thing does not seem to have any negative perf impact.

Follows: c1ca429c183 ("Optimize deep statement matching (#852)")

Closes returntocorp/semgrep-rules#660
Closes PA-2992

test plan:
make test # new tests
Also, compared against develop running p/default on 32 repos from stress-test-monorepo, and no meaningful slowdown or increase in memory usage was observed.

PR checklist:

### DIFF
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -2249,15 +2249,14 @@ and m_stmts_deep ~inside ~less_is_ok (xsa : G.stmt list) (xsb : G.stmt list) =
       m_stmts_deep ~inside ~less_is_ok xsa bbs
   | ( ({ s = G.ExprStmt ({ e = G.Ellipsis _i; _ }, _); _ } :: _ as xsa),
       (_ :: _ as xsb) ) ->
-      (* let's first try without going deep *)
-      m_list__m_stmt xsa xsb >!> fun () ->
-      if_config
-        (fun x -> x.go_deeper_stmt)
-        ~then_:
-          (match SubAST_generic.flatten_substmts_of_stmts xsb with
-          | None -> fail () (* was already flat *)
-          | Some (xsb, _UNUSED_last_stmt) -> m_list__m_stmt xsa xsb)
-        ~else_:(fail ())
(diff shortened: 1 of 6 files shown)


IagoAbal:
The CI benchmarks failing is unrelated, they fail on every PR.

I'll open a PR to semgrep-rules to fix some test expectations, but first I want to get a review on this PR to be sure it will be merged.

src/matching/Generic_vs_generic.ml
```diff
@@ -2237,15 +2237,14 @@ and m_stmts_deep ~inside ~less_is_ok (xsa : G.stmt list) (xsb : G.stmt list) =
       m_stmts_deep ~inside ~less_is_ok xsa bbs
   | ( ({ s = G.ExprStmt ({ e = G.Ellipsis _i; _ }, _); _ } :: _ as xsa),
       (_ :: _ as xsb) ) ->
-      (* let's first try without going deep *)
-      m_list__m_stmt xsa xsb >!> fun () ->
```
aryx, line 2253: each time we use >!>, there is a bug because we cut some matches.
Ideally we should remove this combinator.
Ideally we should use CTL instead of those deep stmt hacks.

aryx (approved):

IagoAbal:
Rebased and don't know what happened... will look into it.

PR #8443, opened 2023-08-11
Simplify exception printer registration

mjambon:
* Add a version constraint on the Base library to avoid problems due to https://github.com/janestreet/base/issues/146
* Register exception printers in the module where the exception is declared when possible.
* Add good exception printers for osemgrep fatal errors (probably only useful in case of a bug when the exception isn't caught properly)
* Add notice on how to register exception printers in the file `Exception.mli` (distinguishing locally-defined exceptions from external exceptions with bad printers)

test plan: I don't think we have tests for checking the formatting of exceptions since exceptions shouldn't be thrown into the user's face when everything works as expected. I think it's ok how it is. If some exception isn't perfectly formatted, we can fix it later.

PR checklist:

### DIFF
--- a/libs/commons/Exception.mli
+++ b/libs/commons/Exception.mli
@@ -22,6 +22,50 @@
 
    * Do not pass exceptions around without an accompanying trace.
    * Do not re-raise exceptions using 'raise'.
+
+   When to register exception printers?
+   ====================================
+
+   Custom exception printers are desirable for printing structured exceptions
+   because the default exception printers don't go very deep.
+   This is done with 'Printexc.register_printer'. Two cases should be
+   distinguished:
+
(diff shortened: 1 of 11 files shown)


nmote (commented):
Seems reasonable, thanks for tracking this down!

libs/commons/Exception.mli
```diff
@@ -22,6 +22,50 @@
 
    * Do not pass exceptions around without an accompanying trace.
    * Do not re-raise exceptions using 'raise'.
+
+   When to register exception printers?
```
aryx, line 26: great comment.

aryx (commented):
Nice!

PR #8444, opened 2023-08-11
fix(julia): avoid singleton seq for julia parenexpr

brandonspark:
## What:
This PR allows Julia code containing singleton parentheses expressions to be parsed correctly.

## Why:
We were parsing singleton paren exprs as singleton `Seq`s, which messed with some semantic stuff, like `metavariable-comparison`.

## How:
Cased on it so we don't produce singleton `Seq`s.

## Test plan:
`make test`

Closes PA-2991

PR checklist:

### DIFF
--- a/languages/julia/generic/Parse_julia_tree_sitter.ml
+++ b/languages/julia/generic/Parse_julia_tree_sitter.ml
@@ -1853,7 +1853,14 @@ and map_quotable (env : env) (x : CST.quotable) : expr =
             v2)
           v3
       in
-      let base = Seq (v2 :: v3) |> G.e in
+      let base =
+        match v3 with
+        (* This means we would produce a singleton Seq. Let's not do that, and just
+           take the expression itself.
+        *)
+        | [] -> v2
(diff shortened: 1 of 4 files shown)


MichaHoffmann (approved):
lgtm

languages/julia/generic/Parse_julia_tree_sitter.ml
```diff
@@ -1853,7 +1853,14 @@ and map_quotable (env : env) (x : CST.quotable) : expr =
             v2)
           v3
       in
-      let base = Seq (v2 :: v3) |> G.e in
+      let base =
+        match v3 with
+        (* This means we would produce a singleton Seq. Let's not do that, and just
+           take the expression itself.
+        *)
+        | [] -> v2
```
aryx: you could return also a ParenExpr. You could also just return v2 but set its range
to also cover the opening and closing parenthesis for autofixing to work better on such code.

aryx (approved):

PR #8462, opened 2023-08-15
Adding location tracking for jsonnet

SophiaSR:
I want to clean up the code a bit more and make the error messages prettier, and also add trace for operators before opening this as a real pull request, but opening a draft PR right now for feedback on the current state!

PR checklist:


aryx:
What is the test plan: ?

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
aryx, line 1: What is this file??

languages/cpp/ast/ast_cpp.ml
```diff
@@ -1026,7 +1026,11 @@ let get_original_token_location = function
 let make_expanded ii =
   (* TODO? use Pos.fake_pos? *)
   let no_virt_loc =
-    ( { Tok.str = ""; pos = { charpos = 0; line = 0; column = 0; file = "" } },
+    ( {
```
aryx, line 1029: maybe make an helper function in Tok.ml to create those empty location,
so next time we add a field we don't have to update this too.

aryx: You can use the strategy we use for AST_generic.empty_id_info that use default parameter.

languages/cpp/menhir/parsing_hacks_define.ml
```diff
@@ -74,6 +74,7 @@ let mark_end_define ii =
     Tok.OriginTok
       {
         str = "";
+        trace = [];
```
aryx, line 77: same here, we probably want to use an helper function to create an empty location
amd then use { empty_loc with pos = { ... } }

libs/ast_generic/AST_generic_to_v1.ml
```diff
@@ -58,7 +58,8 @@ let error any =
 (* Mapper *)
 (*****************************************************************************)
 
-let map_token_location { Tok.str; pos = { charpos; line; column; file } } =
+let map_token_location
+    { Tok.str; trace = _; pos = { charpos; line; column; file } } =
```
aryx, line 62: here you can use ; _ }
to not care about the other fields, so you don't even have to match trace.

libs/ast_generic/Meta_AST.ml
```diff
@@ -20,6 +20,7 @@ let vof_filename v = OCaml.vof_string v
 let vof_token_location
     {
       Tok.str = v_str;
+      trace = _;
```
aryx, line 23: same, better to use _;  and not even name the trace field.

libs/lib_parsing/Tok.ml
```diff
@@ -196,11 +197,22 @@ let unsafe_fake_bracket x = (unsafe_fake_tok "(", x, unsafe_fake_tok ")")
 (* Accessors *)
 (*****************************************************************************)
 
-let stringpos_of_tok (x : t) : string =
+let string_pos_of_tok_no_trace (x : t) : string =
   match loc_of_tok x with
   | Ok loc -> Pos.string_of_pos loc.pos
   | Error msg -> spf "unknown location (%s)" msg
 
+let stringpos_of_tok (x : t) : string =
```
aryx, line 205: maybe better to invert and keep the original stringpos_of_tok and
add a new string_pos_of_tok_with_trace

libs/lib_parsing/Tok.mli
```diff
@@ -19,11 +19,12 @@
 
 type location = {
   str : string; (* the content of the token starting at pos *)
+  trace : t list;
```
aryx, line 22: this requires a big comment.

aryx: I wonder if instead we could use the ExpandedTok case in Tok.t to represent those tokens.

libs/ojsonnet/Core_jsonnet.ml
```diff
@@ -57,26 +57,28 @@ type ident = string wrap [@@deriving show]
  * no Import (expanded during desugaring).
  * TODO? add Import that resolves lazily during Eval?
  *)
-type expr =
+type expr_with_trace = expr * Tok.t list
```
aryx, line 60: I don't understand why we need this in addition to the extension you added to Tok.t.
Can you add a comment?

libs/ojsonnet/Desugar_jsonnet.ml
```diff
@@ -116,107 +116,111 @@ let desugar_bracket ofa env (v1, v2, v3) =
 
 let desugar_ident env v : C.ident = (desugar_wrap desugar_string) env v
 
-let rec desugar_expr env v : C.expr =
-  try desugar_expr_aux env v with
+let rec desugar_expr trace env v : C.program =
```
aryx, line 119: it's better to add the trace in the environment; that's what environments are for, to store information to pass down.

libs/ojsonnet/Core_jsonnet.ml
aryx: It's a bit heavy because this require to change lots of code, and pattern matching an expression is a bit tedious now
with this pair.
If the problem you had was to find a way to adjust the Tok.t embedded inside AST_jsonnet or Core_jsonnet as you desugar, you could instead use a visitor; We can actually automatically derive those visitors (see what we do in AST_generic.ml) so then you just need to visit the Core_jsonnet AST and adjust all the tokens in one pass.

libs/ojsonnet/Eval_jsonnet_subst.ml
```diff
@@ -4,7 +4,7 @@ module V = Value_jsonnet
 module A = AST_jsonnet
 module J = JSON
 
-exception Error of string * Tok.t
+exception Error of string * Tok.t * Tok.t list
```
aryx, line 7: Again better to instead use the trace field (or ExpandedTok) in Tok.t to avoid having to change all our APIs.

src/engine/Xpattern_match_aliengrep.ml
```diff
@@ -6,7 +6,7 @@ let convert_pos ~file (loc : Aliengrep.Match.loc) =
   (* single "token" spanning the whole match *)
   let charpos = loc.start in
   let line, column = Xpattern_matcher.line_col_of_charpos file charpos in
-  { Tok.str = loc.substring; pos = { charpos; file; line; column } }
+  { Tok.str = loc.substring; trace = []; pos = { charpos; file; line; column } }
```
aryx, line 9: this can be replace with
let pos = { charpos; file; line; column} (or  let pos = mk_pos ~charpos ~file ~line ~column) in
Tok.mk_loc ~str ~pos (and use a default value for ?trace=[])

aryx (changes requested):
For this kind of change that entail lots of collateral evolutions (lots of code to update because pattern match need to be updated), it's better to make a first PR with just the core changes, to avoid doing all those collateral evolutions, especially if they have to be undo if the design decision was maybe not the best).

aryx:
at the same time your demo is arriving very soon, so maybe it's better to continue that way to get a working demo.

PR #8472, opened 2023-08-16
refactor(matching): clean up matching env

brandonspark:
## What:
This PR just removes all the excessive calls to `empty_environment`, which creates a functional environment value that doesn't change, and simply calls it once and passes it down.

## Why:
This has the advantage that if we add program-specific information into the environment, we will not need to compute it multiple times / pass it down excessively. This will be the case for matching wildcard imports, which I would like to be able to do.

## How:
Killed all the `empty_environment` calls and just made it one call, at the beginning of `check2`. I also renamed it to `m_env` so as to distinguish it from the `env` passed down by the visitor.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/src/matching/Match_patterns.ml
+++ b/src/matching/Match_patterns.ml
@@ -141,12 +141,10 @@ let (rule_id_of_mini_rule : Mini_rule.t -> Pattern_match.rule_id) =
     languages = mr.languages;
   }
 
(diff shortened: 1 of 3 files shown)


aryx (approved):

aryx:
Love those refactoring PR before the PR adding the actual feature.

PR #8496, opened 2023-08-19
Per-rule version constraints min-version/max-version

mjambon:
Key points:
* Each rule now support Semgrep version constraints `min-version` and `max-version`. If the Semgrep version doesn't fit within these constraints, the rule is skipped. It is a requirement that skipping such a rule be not fatal.
* We're introducing a new kind of rule error which is a new kind of error within the existing type (I tried the alternative of having a dedicated field for version incompatibilities but it got too complicated).
* Invalid rules used to cause semgrep to abort after parsing the rules and to discard all errors except the first one (note to us all: never ever discard errors silently). I changed this behavior in semgrep-core and in osemgrep (`--experimental`) so that we complete the scan even if some rules are invalid. I'm not sure about the consequences. The comments I removed mentioned that we were matching pysemgrep's behavior which aborts as soon as it finds an invalid rule (still doesn't tell us why it would abort early, though) and I haven't scrutinized pysemgrep long enough to see if it still performs some rule validation and early abort.
* A new `skipped_rules` field is now returned by semgrep (JSON only). It has the same type as the one that was already exposed by semgrep-core. It is not exposed by pysemgrep (yet?), only osemgrep for now. Both disabled in osemgrep for now so as to match pysemgrep's output.
* The `errors` field is now populated with errors coming from invalid rules in both pysemgrep and osemgrep.
* A new error level `Info` was introduced, associated with exit code 0.

Uses https://github.com/returntocorp/semgrep-interfaces/pull/113

PR checklist:


aryx:
fail CI

semgrep.opam
```diff
@@ -6,7 +6,7 @@ synopsis:
 description: """
 Semgrep is like grep but for searching patterns at the AST level.
 
-For more information see https://semsgrep.dev
+For more information see https://semgrep.com
```
aryx: this actually redirect to semgrep.dev, so maybe simpler to keep semgrep.dev?

cli/src/semgrep/core_output.py
```diff
@@ -33,8 +33,8 @@
 def _core_location_to_error_span(location: core.Location) -> out.ErrorSpan:
     return out.ErrorSpan(
         file=location.path,
-        start=out.PositionBis(line=location.start.line, col=location.start.col),
-        end=out.PositionBis(line=location.end.line, col=location.end.col),
+        start=location.start,
```
aryx, line 37: would have been better to have those changes in a separate PR, but ok.

src/core/Rule.ml
```diff
@@ -587,18 +592,37 @@ and invalid_rule_error_kind =
...
-type error =
+type error_kind =
   | InvalidRule of invalid_rule_error
   | InvalidYaml of string * Tok.t
   | DuplicateYamlKey of string * Tok.t
   | UnparsableYamlException of string
 
-(* can't use Error because it's used for severity *)
-exception Err of error
+type error = {
```
aryx, line 609: This could also be a separate PR

src/osemgrep/cli_scan/Cli_json_output.ml
```diff
@@ -215,15 +224,18 @@ let error_spans ~(error_type : Out.core_error_kind) ~(location : Out.location) =
          spans = [dataclasses.replace(..., config_path=yaml_path)]
       *)
       let span =
+        (* This code matches the Python code.
+           Not sure what it does, frankly. *)
```
aryx, line 221: This was useful for the playground I think. @emjin wrote it I think.

src/reporting/Semgrep_error_code.ml
```diff
@@ -108,56 +154,28 @@ let known_exn_to_error ?(rule_id = None) file (e : Exception.t) : error option =
         | Tok.OriginTok { str; _ } -> spf "`%s` was unexpected" str
         | __else__ -> "unknown reason"
       in
-      Some (mk_error_tok tok msg Out.ParseError)
+      Some (mk_error_tok rule_id tok msg Out.ParseError)
   | Parsing_error.Other_error (s, tok) ->
-      Some (mk_error_tok ~rule_id tok s Out.SpecifiedParseError)
-  | R.Err err -> (
-      match err with
```
aryx, line 115: why all of this was deleted?

aryx (approved):

aryx:
Where is the PR on semgrep-interfaces?

semgrep.opam
mjambon: I don't think it matters. There was a typo.

src/reporting/Semgrep_error_code.ml
mjambon: It moved to the function `opt_error_of_rule_error` above.

mjambon: The exception was renamed `Rule.Err` -> `Rule.Error` because the other `Error` constructor occurring in another type is elegantly disambiguated with a type annotation.

In general, use this:
```
match (x : Foo.bar) with
| Error _ -> ...
```
rather than
```
match x with
| Foo.Error _ -> ...
```
because the latter can be ambiguous and misses an opportunity to tell the reader what type we're dealing with.

mjambon:
I still have to write a changelog, write the docs, and review the whole thing myself, and in particular the changes in the output format.

src/core/Rule.ml
mjambon: yes. Eventually, I'll learn.

mjambon:
phew. The tests pass and I'm done with fixes in the ATD interface.

aryx:
And we forgot to test with pysemgrep ...

PR #8504, opened 2023-08-21
osemgrep: rename __main__.py to simple main.py and remove some shebang

aryx:
test plan:
cd cli
make e2e
wait for green CI check

PR checklist:


ajbt200128 (changes requested):
Is there a reason to remove __main__.py? It's pretty standard to have one (see [docs](https://docs.python.org/3/library/__main__.html#main-py-in-python-packages)), and removing it disables loading pysemgrep as a module. I'm not sure if this will break anything for us, but considering how common this is, it could possibly break things for people outside of the company using semgrep.

aryx:
That's precisely why I did this. People can still load semgrep code, but yes they will break if they use python -m semgrep, which is intended because soon this will not work anymore.

aryx:
As I was doing that I've found a few occurrence of old code in our CI using python -m semgrep and I'm glad it broke so I was able to find it and fix it.

aryx:
If people want to call semgrep, they must use cli/bin/semgrep entry point, not python -m semgrep or whatever other way.

aryx:
I can add comment in main.py to explain the rational behind it maybe.

ajbt200128 (approved):

PR #8505, opened 2023-08-21
osemgrep: cleanup cli.py and move git safe dir to Dockerfile

aryx:
Not sure why this was done in cli.py but it seems a better
separation of concern and least suprising to do this
git config in the Dockerfile instead.

test plan:
make e2e

PR checklist:

### DIFF
--- a/cli/src/semgrep/cli.py
+++ b/cli/src/semgrep/cli.py
@@ -12,48 +12,15 @@ from semgrep.commands.publish import publish
 from semgrep.commands.scan import scan
 from semgrep.default_group import DefaultGroup
 from semgrep.state import get_state
-from semgrep.util import git_check_output
 from semgrep.verbose_logging import getLogger
 
 logger = getLogger(__name__)
-
-
-def maybe_set_git_safe_directories() -> None:
-    """
-    Configure Git to be willing to run in any directory when we're in Docker.
-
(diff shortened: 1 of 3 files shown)


cli/src/semgrep/cli.py
```diff
@@ -12,48 +12,15 @@
...
 @click.help_option("--help", "-h")
 @click.pass_context
 def cli(ctx: click.Context) -> None:
-    """
-    To get started quickly, run `semgrep scan --config auto`
-
-    Run `semgrep SUBCOMMAND --help` for more information on each subcommand
-
-    If no subcommand is passed, will run `scan` subcommand by default
-    """
```
ajbt200128, line 56: Why delete this documentation? It's user facing (prints when running `semgrep --help`), so if it's out of date I think we should instead update it

ajbt200128 (approved):
Looks good, but I don't think we want to delete that documentation

cli/src/semgrep/cli.py
aryx: Because this is now provided by osemgrep. cli/bin/semgrep now calls by default osemgrep which does handle the 'semgrep --help'.

ajbt200128: Ahh ok I see

PR #8506, opened 2023-08-21
osemgrep: get rid of command.py

aryx:
This class seems an overcomplicated solution for just checking whether
we ran from semgrep ci. Maybe it was not mean to be extended
and further used, but it's just used for one thing right now,
so let's just simplify the code and get rid of it.

test plan:
make e2e

PR checklist:


cli/src/semgrep/command.py
```diff
@@ -1,26 +0,0 @@
-"""
```
aryx, line 1: @emjin you wrote this file according to git blame. It's ok if I remove it?

PR #8514, opened 2023-08-21
feat: match in the presence of wildcard imports

brandonspark:
## What:
This PR makes it such that patterns that look like qualified names can take into account "prefixes" due to wildcard imports like `from A import *` in Python.

## Why:
Previously, matching a qualified name such as `A.B.C.x` was tricky, because the following programs were all possible:
```
foo(A.B.C.x)
```
```
from A import *
foo(B.C.x)
```
```
from A.B import *
foo(C.x)
```
```
from A.B.C import *
foo(x)
```

However, all of them had different matching behavior, because matching did not know how to properly take into account the wildcard import, which makes identifiers within its scope have an implicit prefix to its qualified name, potentially.

This means that a pattern which matched all four would have to be a `pattern-either` of all four options. This led to slowdowns when writing rules, and just general lack of convenience for rule-writers. We want to make it such that Semgrep understands that the naive pattern `A.B.C.x` can demonstrate differing behavior, depending on whether there is a wildcard import around with some prefix of the pattern's qualified name.

## How:
This PR makes it such that the `Matching_generic.env` has a `wildcard_imports` field in it, which is obtained by visiting the AST of a target prior to matching, and compiling that list of wildcard imports.

Then, when `m_name` is invoked, a code path has been introduced which considers the possibility of "unpacking" the wildcard import's prefix. For instance, if there is a `from A.B import *` somewhere within a program, we consider unpacking a prefix of `A.B` when matching any name.

This is as described at this scope doc: https://www.notion.so/Matching-in-the-Presence-of-Wildcard-Imports-fb004e0580874944ac23c084c69be667?d=4a91132287364654934dea529cb2dbab#69cae00fdfc246f2b0a1edcecf86d10a

We chose to put it into the matching environment, because it's quite a bit of work to consolidate this information at the naming level. To make this a naming problem, we would have to store the wildcard import information within our naming context, and update it accordingly to the AST's scoping rules. Then, we would have to store this information within the AST itself, which would lead to both probable significant memory usage, but also an inconvenience in the representation of identifiers, because this potential prefix might be used at many identifiers.

Since we expect that scoping of wildcard imports is likely a non-issue in most practical code, the simplest and cleanest solution is to put it into the matching environment. It is worth noting that this makes `Generic_vs_generic` less dependent upon specifically the two nodes being compared, because there might be "hidden information" floating around in the environment which alters the match, but I believe this to be a generally desirable change.

## Test plan:
`make test`

Closes PA-1006

PR checklist:


src/analyzing/Visit_wildcard_imports.ml
```diff
@@ -0,0 +1,36 @@
...
+
+    method! visit_directive_kind store dk =
+      match dk with
+      | ImportAll (_t, DottedName name, _t') -> Common.push name store
+      | ImportAll (_, FileName _, _)
+      | _ ->
+          ()
+  end
+
+let visit : AST_generic.any -> G.ident list list =
```
aryx: I'm a bit worried about the performance of this; maybe it would be simpler to just  look at the toplevel directives,
so you don't need to recurse and visit every nodes.

src/engine/Match_search_mode.ml
```diff
@@ -564,8 +564,11 @@ let rec filter_ranges (env : env) (xs : (RM.t * MV.bindings list) list)
                    | Xlang.LAliengrep ->
                        raise Impossible
                  in
+                 let ast, _ = Lazy.force env.xtarget.lazy_ast_and_errors in
+                 (* This will incur some cost from the visitor, but it shouldn't be much. *)
```
aryx: this comment is hard to read because it's not obvious from environment_of_any that internally this runs a visitor.

src/matching/Generic_vs_generic.ml
```diff
@@ -517,6 +522,44 @@ let rec m_name a b =
...
+         This can mess with lots of things like `metavariable-regex` or even join mode,
+         which expect to see metavariables matching sensible ranges. To maintain that we
+         don't mess those things up, we will try to give these inserted prefixes the same
+         location data as the name that it is modifying.
+      *)
+      match AST_generic_helpers.dotted_ident_of_name b with
+      | [] ->
+          logger#warning
+            "Somehow got empty dotted ident from name %s during matching"
+            ([%show: G.name] b);
```
aryx: nah, don't use those %show: thing. just use show_name.

src/matching/Matching_generic.ml
```diff
@@ -389,8 +390,16 @@ let (envf : MV.mvar G.wrap -> MV.mvalue -> tin -> tout) =
         (lazy (spf "envf: success, %s (%s)" mvar (MV.str_of_mval any)));
       return new_binding
 
-let empty_environment lang config =
-  { mv = []; stmts_matched = []; lang; config; deref_sym_vals = 0 }
+let environment_of_any lang config any =
```
aryx: maybe have an environment_for_program that we should use at most place,
and then a separate environment_for_any that I'm not even sure we need.

tests/patterns/python/any_qualified_with_wildcard.py
```diff
@@ -0,0 +1,16 @@
+# will match because wildcard imports in matching are acontextual
```
aryx: maybe add an "unfortunately"

tests/patterns/python/scoped_wildcard.py
```diff
@@ -0,0 +1,17 @@
+
+# match because wildcard imports do not respect scoping in matching
```
aryx: also add unfortunately.
Maybe we should mark certain ERROR: really as TODO:

aryx (approved):
LGTM.
Dunno if @IagoAbal has more feedback but this is mostly how I imagine to implement the feature for a v0.

brandonspark:
wait lmao why did I not merge this

brandonspark:
oh nvm I'm waiting on @IagoAbal

changelog.d/pa-1006.added
```diff
@@ -0,0 +1,8 @@
+Matching: Qualified names written as patterns can now match valid instances of
```
IagoAbal, line 1: Nice changelog entry!

src/analyzing/Visit_wildcard_imports.ml
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the file
+ * license.txt for more details.
+ *)
+
+module G = AST_generic
+
+(* alt:
```
IagoAbal, line 18: If we put here an alternative, I would expect a few lines discussing pros/cons between both.

src/analyzing/Visit_wildcard_imports.ml
```diff
@@ -0,0 +1,53 @@
...
+  (* we only collect top-level wildcard imports *)
+  List.fold_right
+    (fun x acc ->
+      match x with
+      | {
+       G.s = DirectiveStmt { d = ImportAll (_t, DottedName name, _t'); _ };
+       _;
+      } ->
+          name :: acc
+      | { s = DirectiveStmt { d = ImportAll (_, FileName _, _); _ }; _ }
```
IagoAbal: I'm not sure what is the purpose of this case if we just "or" it with the wildcard `_` case?

src/analyzing/Visit_wildcard_imports.ml
```diff
@@ -0,0 +1,53 @@
...
+     let v = new import_visitor in
+     fun any ->
+       let ids = ref [] in
+       v#visit_any ids any;
+       !ids
+     [@@profiling]
+*)
+
+let visit_toplevel prog : G.ident list list =
+  (* we only collect top-level wildcard imports *)
```
IagoAbal: OK so probably we should discuss here why we only take the top-level imports, or point to somewhere else where this may be discussed.

src/analyzing/Visit_wildcard_imports.ml
```diff
@@ -0,0 +1,53 @@
...
+   let visit : AST_generic.any -> G.ident list list =
+     let v = new import_visitor in
+     fun any ->
+       let ids = ref [] in
+       v#visit_any ids any;
+       !ids
+     [@@profiling]
+*)
+
+let visit_toplevel prog : G.ident list list =
```
IagoAbal: Better `: G.dotted_ident list`.

src/matching/Matching_generic.ml
```diff
@@ -114,6 +114,7 @@ type tin = {
   lang : Lang.t;
   config : Rule_options.t;
   deref_sym_vals : int;
+  wildcard_imports : AST_generic.ident list list;
```
IagoAbal: `AST_generic.dotted_ident list`

IagoAbal: (We should probably have a type alias for `dotted_ident` with an unboxing annotation...)

src/matching/Matching_generic.ml
```diff
@@ -389,8 +390,26 @@ let (envf : MV.mvar G.wrap -> MV.mvalue -> tin -> tout) =
...
+    wildcard_imports = [];
+  }
+
+let environment_of_program lang config prog =
+  let wildcard_imports = Visit_wildcard_imports.visit_toplevel prog in
+  { (default_environment lang config) with wildcard_imports }
+
+let environment_of_any lang config any =
+  match any with
+  | G.Ss prog
```
IagoAbal: Could this mean that in some cases we end up looking at non-top-level wildcard imports? If we understand the implications it would be nice to document them here, otherwise maybe we should just cover the `Pr` case?

src/matching/Matching_generic.mli
```diff
@@ -10,6 +10,7 @@ type tin = {
   deref_sym_vals : int;
       (** Counts the number of times that we "follow" symbollically propagated
     * values. This is bound to prevent potential infinite loops. *)
+  wildcard_imports : AST_generic.ident list list;
```
IagoAbal: `AST_generic.dotted_ident list` and some doc string explaining the purpose of this list.

tests/patterns/python/any_qualified_with_wildcard.py
```diff
@@ -0,0 +1,16 @@
+# will match because wildcard imports in matching are acontextual, unfortunately
+# ERROR:
+bar(
```
IagoAbal, line 4: Should we perhaps just collect wildcard imports before any definition is made?

tests/patterns/python/scoped_wildcard.py
```diff
@@ -0,0 +1,17 @@
+
```
IagoAbal, line 1: Very nice that you took the time to write tests to cover different cases, also the TODO ones.

changelog.d/pa-1006.added
```diff
@@ -0,0 +1,8 @@
...
+identifiers which lie underneath a wildcard import. For instance, in Python,
+we could write the pattern `A.B.C.x`, and match the usage in the program
+```
+from A.B import *
+
+foo(C.x)
```
IagoAbal, line 7: We need a ticket in https://linear.app/semgrep/project/documentation-93e79895033e to remember to document this feature.

tests/patterns/python/any_qualified_with_wildcard.sgrep
```diff
@@ -0,0 +1 @@
+$X.$Y
```
IagoAbal, line 1: Makes me wonder if we should support this equivalence in the present of metavariables. It could lead to many FPs. Personally I would only apply the equivalence when the dotted ident has no metavariables. If you're looking for `A.foo` and I find `foo` within a `from A import *` I think it's reasonable to match. But I don't think it's reasonable to match e.g. `A.$X` against every unknown identifier just because there is a `from A import *`. I think this is better done in Pro with inter-file naming. I also don't think this is something that SR has really requested?

tests/patterns/python/wildcard_qualified_mvar.sgrep
```diff
@@ -0,0 +1 @@
+$A.x
```
IagoAbal, line 1: Same here, I don't remember SR asking for this? And what happens here if we have multiple wildcard imports? (We could think of some corner cases involving multiple wildcard imports.)

src/matching/Generic_vs_generic.ml
```diff
@@ -357,7 +357,12 @@ let m_regexp_options a_opt b_opt =
 (* TODO: factorize with metavariable and aliasing logic in m_expr
  * TODO: remove MV.Id and use always MV.N?
  *)
-let rec m_name a b =
+let rec m_name_inner a b =
+  (* Inside of this function, any recursive call to m_name means to not
+     start another round of entry-specific logic. We only recurse with
+     the inner m_name function.
+  *)
+  let m_name = m_name_inner in
```
IagoAbal, line 365: Why not just change calls to `m_name` with `m_name_inner` ? Seems confusing to have a `let-rec` where `m_name` is defined and within a subset of this `let-rec`, redefine `m_name` to something else.

src/matching/Generic_vs_generic.ml
```diff
@@ -517,6 +522,44 @@ let rec m_name a b =
...
 
+(* This is just an entry point for m_name_inner, which just ensures that we only
+   ever unpack wildcard imports once, before entering the main recursive loop of
+   m_name_inner.
+   Since we might reach m_name through many entry points (IdQualified,
+   DotAccess, Id, etc), it's easier to intercept all of these "naming-related"
+   matching procedures at a common point. This is where they all meet.
+*)
+and m_name a b =
+  m_name_inner a b >!> fun () tin ->
```
IagoAbal, line 543: Why `>!>` here instead of `>||>` ? Good to explain it.

src/matching/Generic_vs_generic.ml
```diff
@@ -517,6 +522,44 @@ let rec m_name a b =
...
+  List.fold_left
+    (fun acc import ->
+      acc
+      >||>
+      (* These imports face the same issue as the original problem with resolved names,
+         which is that these idents could contain confusing location data. In particular,
```
IagoAbal: Another reason to not allow this equivalence in the presence of metavariables.

src/matching/Generic_vs_generic.ml
```diff
@@ -517,6 +522,44 @@ let rec m_name a b =
...
+         these wildcard imports could be pulled from anywhere.
+         This can mess with lots of things like `metavariable-regex` or even join mode,
+         which expect to see metavariables matching sensible ranges. To maintain that we
+         don't mess those things up, we will try to give these inserted prefixes the same
+         location data as the name that it is modifying.
+      *)
+      match AST_generic_helpers.dotted_ident_of_name b with
```
IagoAbal: Correct me if I'm wrong but this seems to apply to `b` even if `b` happens to be an already *resolved* name ?

IagoAbal: Could you add a test with something like:

```python
import A.foo
from B import *

foo()
```

and try to match `B.foo()` ? What will that do ?

IagoAbal: I think you already have `module H = AST_generic_helpers`.

IagoAbal:
> oh nvm I'm waiting on @IagoAbal

Thanks for waiting for my review! I am positive about the PR but I have two main requests:

1. Make sure that we do not apply the equivalence to names that are already resolved. (Perhaps this is already the case? I couldn't be sure of that looking at the code.) If there is `import A.foo` and `from B import *`, presumably `foo()` will be resolved to `A.foo()` during naming. So we should not match `B.foo()` against `foo()`. This would be to avoid FPs / undesirable matches. And for Pro in particular, this equivalence should not apply when inter-file naming has already resolved names correctly (hopefully).

2. By default, I would prefer that we do not apply the equivalence when the pattern contains metavariables, like `$X.$Y`, `$X.foo`, or `A.$Y`. Again, because this is much more likely to produce incorrect/undesirable matches, and I am not aware of SR requesting this feature (could be wrong). When we match `A.foo()` against `from A import *` + `foo()` we're making reasonable assumptions. But I don't think matching `$X.$Y()` against `from A import *` + `foo()` is that reasonable... what if there is also `from B import *`? Then we would have to generate two matches for `foo()`, as `A.foo()` and as `B.foo()`. @aryx proposes that this could be put under a rule option, I agree with that if we have it disabled by default.

src/analyzing/Visit_wildcard_imports.ml
brandonspark: that was supposed to be guarding against a potential refactoring of the `module_name` type, for instance the addition of another variant constructor

brandonspark: although now i realize because of the wildcard it actually does no such thing

tests/patterns/python/any_qualified_with_wildcard.py
brandonspark: I believe this to be an unrealistic scenario, and that it would be better to be generous with our matching, than miss a case. I think it would be easy, for instance, to whitelist certain kinds of definitions which it can appear before, and then accidentally forget about one.

changelog.d/pa-1006.added
brandonspark: https://linear.app/semgrep/project/documentation-93e79895033e

src/matching/Generic_vs_generic.ml
brandonspark: i was afraid of someone who later touched `m_name_inner` forgetting, and accidentally calling the external `m_name`

src/matching/Generic_vs_generic.ml
brandonspark: I kind of viewed this as a similar situation to symbolic propagation, where there is some extra information "stored within" the node in question that we are matching. I didn't want to incur this unpacking cost (however much that may be) for a name we were going to match anyway, so I decided to make it a `>!>`. Thoughts?

brandonspark:
@IagoAbal good points! those changes should be made now

src/matching/Generic_vs_generic.ml
```diff
@@ -503,20 +512,91 @@ let rec m_name a b =
...
+  let dotted_contains_mvars dotted =
+    List.exists (fun (s, _) -> Metavariable.is_metavar_name s) dotted
+  in
+  let dotted_b = H.dotted_ident_of_name b in
+  m_name_inner a b >!> fun () tin ->
+  match (b, dotted_b) with
+  (* If we are matching with a pattern that has a metavariable in it,
+     let's not unpack the wildcard import.
+
+     Usually, someone wants a wildcard import to activate when they know the full path
```
aryx, line 548: great comment

src/matching/Matching_generic.ml
```diff
@@ -389,8 +390,26 @@ let (envf : MV.mvar G.wrap -> MV.mvalue -> tin -> tout) =
...
+let environment_of_program lang config prog =
+  let wildcard_imports = Visit_wildcard_imports.visit_toplevel prog in
+  { (default_environment lang config) with wildcard_imports }
+
+let environment_of_any lang config any =
+  match any with
+  | G.Pr prog -> environment_of_program lang config prog
+  | _ -> default_environment lang config
+
+let wipe_wildcard_imports f tin = f { tin with wildcard_imports = [] }
```
aryx, line 412: tbh I don't fully understand when and why we need to use wipe_wildcard_imports in the Generic_vs_generic.ml, but let's go forward  and merge this.

aryx (approved):

src/matching/Generic_vs_generic.ml
IagoAbal: AFAICT we use `>||>` with symbolic propagation.

PR #8515, opened 2023-08-21
chore: kill numeric capture metavars for real

brandonspark:
## What:
This PR finally removes numeric capture group metavariables, as it's been about two months since we said we would, in Release 1.30: https://github.com/returntocorp/semgrep/releases/tag/v1.30.0

## Why:
We agreed to wait for a bit to sunset the feature, and it's about time.

## How:
Removed the code and then updated the tests.

## Test plan:
`make test`

PR checklist:

### DIFF
--- a/changelog.d/pa-3012.changed
+++ b/changelog.d/pa-3012.changed
@@ -1,0 +1,5 @@
+Rule syntax: Numeric capture group metavariables in regex patterns
+have been removed. Now, in order to introduce metavariables in
+regular expressions, you must use named capture groups.
+
+This was something deprecated in 1.30, that has finally been removed.
--- a/cli/tests/e2e/snapshots/test_nosemgrep/test_regex_rule__nosemgrep/results.json
+++ b/cli/tests/e2e/snapshots/test_nosemgrep/test_regex_rule__nosemgrep/results.json
@@ -25,58 +25,6 @@
           "source-rule-url": "https://github.com/grab/secret-scanner/blob/master/scanner/signatures/pattern.go"
         },
         "metavars": {
-          "$2": {
-            "abstract_content": "aws",
-            "end": {
-              "col": 4,
-              "line": 3,
-              "offset": 47
-            },
-            "start": {
-              "col": 1,
-              "line": 3,
-              "offset": 44
-            }
-          },
-          "$3": {
-            "abstract_content": "account",
-            "end": {
-              "col": 12,
-              "line": 3,
-              "offset": 55
-            },
-            "start": {
-              "col": 5,
-              "line": 3,
-              "offset": 48
-            }
-          },
-          "$4": {
-            "abstract_content": "id",
-            "end": {
-              "col": 15,
-              "line": 3,
-              "offset": 58
-            },
-            "start": {
-              "col": 13,
-              "line": 3,
-              "offset": 56
-            }
-          },
-          "$6": {
-            "abstract_content": ":",
-            "end": {
-              "col": 16,
-              "line": 3,
-              "offset": 59
-            },
-            "start": {
-              "col": 15,
-              "line": 3,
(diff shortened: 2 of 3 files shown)


aryx:
Honestly I'm not super convinced we should do this. I like the $1 $2 ... it's easy and it works that way in a few languages (e.g., Perl) so it's also familiar.

brandonspark:
@aryx i agree its easy, but in a dangerous way. i think that the implicit introduction of metavariables (and their associated unification) can cause a rule writer extreme pain without knowing why. i myself just ran into this yesterday, because of a rule `aws-network-acl-allows-public-ingress` (https://github.com/returntocorp/semgrep-rules/blob/develop/terraform/aws/security/aws-ec2-security-group-allows-public-ingress.yaml) because of these capture groups -- it fails to properly `not` out when matching the string `127.0.0.0/8` with the pattern
```
     - pattern-not-regex: ^127\.\d{1,3}\.\d{1,3}\.\d{1,3}/(8|9|[1-3][0-9])$
 ```
but it succeeds with
```
     - pattern-not-regex: ^127\.\d{1,3}\.\d{1,3}\.\d{1,3}/8$
 ```
because this no longer has the capture group around.

I would greatly like to remove this, so that rule-writers don't face traps when they are writing rules.

brandonspark:
@aryx further thoughts on this?

aryx:
Yes, not a big fan. If the $1 is creating some conflicts in some rules because of unwanted unify_mvars, then maybe we should just disable this unification for those numeric capture metavars, just like we do in tainting where you have to use unify_mvars: true, if you want this behavior.

aryx:
cc also @ievans who is the one who asked for those $1, $2, ... support a long time ago.
I think we should keep them (even if named capture groups are also now supported).

ievans:
Inclined to agree with Pad, in the sense that this makes simple rules very simple indeed, and more complex rules are still possible.

ievans:
Also I imagine this breaks backwards compat -- what % of rules do we observe using these groups?

ievans:
Also I imagine this breaks backwards compat -- what % of rules do we observe using these groups?

brandonspark:
@aryx @ievans I'm all for simplifying rules, but I think it's also important to acknowledge how we got here in the first place, namely the fact that because the introduced metavariables `$1, $2, ...` are not real metavariable names, this means they do not interface well with things that metavariables are supposed to be able to do.

For a concrete example, it is not possible currently to `metavariable-regex` on a numeric capture group metavariable, and it does not seem feasible to change our schema for what metavariable names should be (which is a deeply ingrained assumption in our codebase).

So I see three things we could do:
1) Go to the effort to support numeric metavariables in all possible circumstances, which might extend into the future with code we write having to do with metavariables
2) Leave numeric metavariables in this kind of intermediate state where they work for some things and not for others. To be clear, this is the current state of the world.
3) Kill numeric capture group metavariables

I would be fine with either 1 or 3, but without changing the naming schema for metavariables, I do not foresee a sustainable way to do 1). Thoughts?

aryx:
I would vote for 2, but disable the unification of numeric metavars between different patterns, so one does not get bad surprise.

brandonspark:
@aryx Why is it that it's preferred to have such "second-class" metavariables which do not interact predictably for the rest of our machinery? I see that this solution would be the quickest way forward, but I am simply concerned that having such unpredictable rule-writing behavior is simply going to introduce more friction in the future.

aryx:
1) would be ideal, but it's possibly a good amount of work, and I'm not sure it's worth it (we have lots of higher priority work). This is why I voted for 2
but fix the annoying situation where numeric metavar interact in bad ways, so we get the most of it with the least amount of effort.
Happy to revisit in a few months when we have users really complaining about it.

PR #8521, opened 2023-08-22
osemgrep: switch to real e2e testing, part 2

aryx:
This is a followup to #8492.
Our end-to-end (e2e) tests ran by 'make e2e' were not
fully end-to-end because they often didn't call semgrep
but instead use an optimization hack called the ClickRunner
to avoid a fork.
With this PR, we switch back to a real fork, which is slower,
but correct.
There's only a few remaining e2e tests that still use the
Click_runner (mostly test_ci.py because of heavy use of Python
mocking).

test plan:
make e2e!

PR checklist:

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -40,9 +40,9 @@ clean:
 
 # From fastest to slowest (on pad's machine):
 # - 'make quick-tests' (=~ 30s)
-# - 'make e2e' (=~ 1m14s)
-# - 'make kinda-quick-tests' (=~ 2min30s )
-# - 'make test' (=~ 10min)
+# - 'make e2e' (=~ 2m20s)
+# - 'make kinda-quick-tests' (=~ 3min40s )
+# - 'make test' (=~ 11min)
 
 PYTEST ?= pipenv run pytest -v --tb=short --durations=10
 
--- a/cli/tests/e2e/test_ci.py
+++ b/cli/tests/e2e/test_ci.py
@@ -1,3 +1,13 @@
+##############################################################################
+# Prelude
+##############################################################################
+# Testing 'semgrep ci' "end-to-end".
+#
+# TODO: actually most of the tests in this file rely on use_click_runner=True
+# because of some mocking and monkeypatching. Thus, this is this not
+# a real e2e test because cli/bin/semgrep is not invoked.
+# Try to use environment variables instead of Python monkey patching
+# so that those tests can also pass with osemgrep.
 import json
 import re
 import shutil
@@ -24,6 +34,9 @@ from semgrep.meta import GitlabMeta
 from semgrep.meta import GitMeta
 from semgrep.metrics import Metrics
 
+##############################################################################
+# Constants
+##############################################################################
 
 pytestmark = pytest.mark.kinda_slow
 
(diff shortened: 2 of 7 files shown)


cli/Makefile
```diff
@@ -40,9 +40,9 @@ clean:
 
 # From fastest to slowest (on pad's machine):
 # - 'make quick-tests' (=~ 30s)
-# - 'make e2e' (=~ 1m14s)
-# - 'make kinda-quick-tests' (=~ 2min30s )
-# - 'make test' (=~ 10min)
+# - 'make e2e' (=~ 2m20s)
+# - 'make kinda-quick-tests' (=~ 3min40s )
+# - 'make test' (=~ 11min)
```
akuhlens, line 45: Does make `make test` includes `make e2e`? If so this isn't too bad at all.

cli/tests/e2e/test_ci.py
```diff
@@ -1,3 +1,13 @@
+##############################################################################
+# Prelude
+##############################################################################
+# Testing 'semgrep ci' "end-to-end".
+#
+# TODO: actually most of the tests in this file rely on use_click_runner=True
+# because of some mocking and monkeypatching. Thus, this is this not
+# a real e2e test because cli/bin/semgrep is not invoked.
+# Try to use environment variables instead of Python monkey patching
+# so that those tests can also pass with osemgrep.
```
akuhlens, line 10: This is a little sad to hear. What percentage of tests would you estimate are truly running end to end? It will be interesting to see how much work there is to merely convert the tests to being truly end to end. As I understand it, this would be a prerequisite before porting the code that would need to be tested by these tests anyway. Is that correct?

cli/tests/e2e/test_ci.py
```diff
@@ -1718,6 +1765,7 @@ def test_existing_supply_chain_finding(
         strict=False,
         assert_exit_code=None,
         env={"SEMGREP_APP_TOKEN": "fake_key"},
+        use_click_runner=True,  # TODO: probably because rely on some mocking
     )
     snapshot.assert_match(
```
akuhlens, line 1770: I take it for all the `use_click_runner=True` tests in this file, that you tried them without the click runner and they failed?

akuhlens (approved):
Great, and not nearly as expensive as I was afraid it was going to be!

cli/Makefile
aryx: yes it does include (well the logic is a bit complicated in the Makefile but 'make test' run e2e tests).

cli/tests/e2e/test_ci.py
aryx: there's 40 tests_xxx.py in cli/tests/e2e/, so it's not too bad that currently just one file (well test_fixtest.py and test_ignores.py have a few use_click_runner=True but we should be able to fix them easily) requires this flag.
Anyway, most of our effort in osemgrep is on porting 'semgrep scan' and we are delaying porting 'semgrep ci' which is tested by test_ci.py, so we're fine I think.

cli/tests/e2e/test_ci.py
aryx: yep

ajbt200128:
Yea the CI testing is tough... even with LSP moving those parts to ocaml made them hard to test. I may try tackling better ocaml side CI testing when I have timne

aryx:
In some sense this reverts https://github.com/returntocorp/semgrep/pull/5444

PR #8527, opened 2023-08-22
feat(git wrapper): Run with git worktree

ihji:
OCaml port of BaselineHandler.baseline_context from pysemgrep

Test plan: manually tested by running `scan_files` function with different worktree snapshots of semgrep repo

PR checklist:

### DIFF
--- a/libs/git_wrapper/Git_wrapper.ml
+++ b/libs/git_wrapper/Git_wrapper.ml
@@ -121,6 +121,52 @@ let files_from_git_ls ~cwd =
   in
   files |> File.Path.of_strings
 
+let get_git_root_path () =
+  let cmd = Bos.Cmd.(v "git" % "rev-parse" % "--show-toplevel") in
+  let path_r = Bos.OS.Cmd.run_out cmd in
+  let result = Bos.OS.Cmd.out_string ~trim:true path_r in
+  match result with
+  | Ok (path, (_, `Exited 0)) -> path
+  | _ -> raise (Error "Could not get git root from git rev-parse")
+
+let run_with_worktree ~commit f =
+  let cwd = Sys.getcwd () |> Fpath.v |> Fpath.to_dir_path in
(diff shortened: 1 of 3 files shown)


libs/git_wrapper/Git_wrapper.mli
```diff
@@ -16,6 +16,24 @@ type status = {
 *)
 val files_from_git_ls : cwd:Fpath.t -> Fpath.t list
 
+(* Executing a function inside a directory created from git-worktree.
+
+   `git worktree` is doing 90% of the heavy lifting here. Docs:
```
aryx, line 21: maybe this comment needs to be updated. 90% of which lifting?
I think this comment should be put next to the code that use git-worktree for the baseline problem.

libs/git_wrapper/Git_wrapper.ml
```diff
@@ -121,6 +121,52 @@ let files_from_git_ls ~cwd =
...
+  let cwd = Sys.getcwd () |> Fpath.v |> Fpath.to_dir_path in
+  let git_root = get_git_root_path () |> Fpath.v |> Fpath.to_dir_path in
+  let relative_path =
+    match Fpath.relativize ~root:git_root cwd with
+    | Some p -> p
+    | None -> raise (Error "")
+  in
+  let rand_dir () =
+    let uuid = Uuidm.v `V4 in
+    let dir_name = "semgrep_git_worktree_" ^ Uuidm.to_string uuid in
```
aryx, line 142: why do we need this uuidm thing? Filename.get_temp_dir_name is not enough to get a random directory?

aryx (approved):

aryx:
It's usually better to have in a single PR the new library code and the code that uses this library, so the test plan
can usually be automated and clear. So here would be better to have a single PR implementing
--baseline in osemgrep.
We can merge this for now still.

PR #8534, opened 2023-08-23
feat(secrets): unvalidated secrets

akuhlens:
## WHAT
Adds a validation state to the pattern_match.t and updates semgrep_interfaces to match with optional fields.

[Opened a PR in semgrep-proprietay](https://github.com/returntocorp/semgrep-proprietary/pull/922)
[Opened a PR in semgrep-interfaces](https://github.com/returntocorp/semgrep-interfaces/pull/111)
 

## Testing
Ran testing as stated in this [PRO-PR](https://github.com/returntocorp/semgrep-proprietary/pull/922) and relying on CI to make sure nothing else breaks.

PR checklist:

  - Should I do a changelog entry for this since it is before launch?


akuhlens:
Since this adds to the JSON I will probably have to go back and regenerated some snapshots.

src/core/Pattern_match.ml
```diff
@@ -78,6 +78,13 @@ type taint_trace_item = {
 type taint_trace = taint_trace_item list [@@deriving show, eq]
 type engine_kind = OSS | Pro [@@deriving show, eq]
 
+type validation_state =
```
brandonspark, line 86: could you add the comment in the record about what this is here as well?

brandonspark: and also, could we try to find a more descriptive name? "validation state" definitely confused me when i first saw it. would be nice to link it back to its use case

brandonspark (approved):
otherwise, lgtm!

src/core/Pattern_match.ml
akuhlens: I am not going to rename this for now, because it is the same name across semgrep-core, the cli, and the app, and I didn't choose that name, but I did add a comment that should clarify.

src/core/Pattern_match.ml
```diff
@@ -78,6 +78,18 @@ type taint_trace_item = {
 type taint_trace = taint_trace_item list [@@deriving show, eq]
 type engine_kind = OSS | Pro [@@deriving show, eq]
 
+(* This type is used by postprocessors to report back the validatity
```
nmote:
```suggestion
(* This type is used by postprocessors to report back the validity
```

src/core/Pattern_match.ml
```diff
@@ -78,6 +78,18 @@ type taint_trace_item = {
...
+   of a finding. No_validator is currently also used when no
+   validation has yet occurred, which if that becomes confusing we
+   could adjust that, by adding another state. This corresponds to
+   a identically named type in semgrep_interfaces output types.*)
+type validation_state =
```
nmote, line 86: Could you refer to the type generated by ATD rather than duplicating it here? Not saying you should do this necessarily, definitely follow any existing convention, but I am curious what you think.

akuhlens: I could, but others don't seems to. For example I was following engine_kind which doesn't.

src/core/Pattern_match.ml
```diff
@@ -78,6 +78,18 @@ type taint_trace_item = {
 type taint_trace = taint_trace_item list [@@deriving show, eq]
 type engine_kind = OSS | Pro [@@deriving show, eq]
 
+(* This type is used by postprocessors to report back the validity
+   of a finding. No_validator is currently also used when no
+   validation has yet occurred, which if that becomes confusing we
+   could adjust that, by adding another state. This corresponds to
```
aryx, line 84: you can add instead a coupling comment like
coupling: with similar type defined in semgrep-interfaces/semgrep_output_v1.atd

aryx (approved):

PR #8539, opened 2023-08-24
feat(ruby): KILL DYP

brandonspark:
## What:
This PR kills the `Dyp` parser for Ruby in favor of the `tree-sitter` parser.

## Why:
The `Dyp` parser is old, crusty, and slow. As long as I have been working at Semgrep, I have noticed that watching tests run from `make test` hits a noticeable slowdown when running the Ruby tests. It turns out, due to @emjin 's experiments, that this is because the parser for Ruby has an immense startup time.

Our current `Dyp` parser for Ruby is non-standard, and a legacy from a bygone time. It works, but it is both slow, a special-case among our parsers, and recently we have found that it has started causing errors when it comes to upgrading to OCaml 5.0. Since OCaml 5.0 has recently begun increasing in priority for us, it is now time for us to sunset this parser, in favor of the shinier `tree-sitter` technology.

## How:
Unfortunately, since we barely use `tree-sitter`, this means that our generic translation is almost nonexistent. It is extremely outdated, meaning that the translation (and grammar) must be brought up to speed, and augmented with Semgrep constructs, which had not been done previously. See https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/444 and https://github.com/returntocorp/ocaml-tree-sitter-semgrep/pull/446.

A significant amount of work had to be done to make parsing up to the same level of power as the old Menhir parser. In particular, there are problems having to do with:
1) ambiguity of sequences of statements separated by ellipses with range expressions
2) ambiguity of range expressions `foo ...` with command calls with Semgrep ellipses `foo ..., x, y, z`
3) appearance of metavariables in places where users would expect it (`$.constant` and `$.identifier`)
4) ambiguity of metavariable ellipses with range expressions
5) ambiguity of metavariables and `$.global_variable`
6) ... to denote the contents of a hash or hash pattern

Most of these problems were fixed in the above PRs, at the point of `ocaml-tree-sitter-semgrep`.

There were also problems with updating the Generic translation. In particular, bringing the Ruby grammar up to speed seemed to introduce Ruby 2.7, which introduced a brand new pattern syntax for pattern matching, which required a large addition to the AST to emulate a pattern.

We also had to ensure that the translation roughly matched the same as the old Ruby translation from Menhir.

## Test plan:
`make test`, and
```
lang/ruby/stats.json:3
   "global": {
     "name": "*",
     "parsing_rate": 0.9995633741049506,
     "line_count": 5201249,
     "error_line_count": 2271,
     "file_count": 54143,
     "error_file_count": 139,
     "total_node_count": 74939847,
     "untranslated_node_count": 16271
   },
```

Closes PA-3017

PR checklist:


aryx:
probably need a changelog at least to warn ruby users; it's an internal change but this may have an effect on users.

brandonspark:
@aryx this PR should be ready for review now!

languages/ruby/generic/ruby_to_generic.ml
```diff
@@ -819,17 +946,37 @@ and list_stmt1 xs =
...
 and list_stmts xs = list expr_as_stmt xs
 
-let program xs = list_stmts xs
+let program xs =
+  Common.save_excursion Flag_parsing.sgrep_mode false (fun () -> list_stmts xs)
 
 let any x =
-  match x with
-  | E x -> (
+  (* We need this sgrep_mode flag because we want some branching behavior
```
aryx, line 953: great comment

aryx (approved):
Great work!
We can now probably remove dyp in dune-project (and regenerate semgrep.opam from it)

PR #8542, opened 2023-08-24
ftr(scan): include commit_timestamp in ci

ThatDevShparki:
## Context
In Semgrep Cloud Platform, scans are used to capture contributor information for `semgrep ci` initiated scans. A component of this as been assuming that the commit associated with the scan occurred at relatively the same time as the scan began. This is not always the case when there are scheduled scans on older repos with ancient commits.

## Solution
Include the `commit_timestamp` along with other commit information sent in `scan.meta`

Test Plan:
- Update snapshots to show that we are now including `commit_timestamp` in the `scan.meta` object

PR checklist:


ievans:
👍 @ThatDevShparki does this require an update to PRIVACY.md as well?

ThatDevShparki:
> 👍 @ThatDevShparki does this require an update to PRIVACY.md as well?

Great callout, just updated it! It seems like we already mention we capture timestamps, but I added to the commit metadata section that we are now explicitly capturing the timestamp along with the commit hash.

PRIVACY.md
```diff
@@ -58,7 +58,7 @@ The following never leave your environment and are not sent or shared with anyon
 The following data will never leave your environment as part of metrics.
```
chmccreery, line 58: This language is confusing - as we then go on to say that it will leave your environment under certain conditions

cli/src/semgrep/meta.py
```diff
@@ -187,6 +187,7 @@ def to_dict(self) -> Dict[str, Any]:
             "commit_author_username": None,
             "commit_author_image_url": None,
             "commit_title": commit_title,
+            "commit_timestamp": self.commit_datetime,
```
chmccreery: So this was already collected and defined and sent to the app as part of each Finding - do you agree? Looking at `RuleMatch.to_app_finding_format()`

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines-overwrite-autodetected-variables/meta.json
```diff
@@ -11,6 +11,7 @@
     "commit_author_username": null,
     "commit_author_image_url": null,
     "commit_title": "Some other commit/ message",
+    "commit_timestamp": "1672552800",
```
chmccreery: I always prefer human-readable timestamps, but that's personal preference. Either way, it would be helpful to include explicit timezone information here (if it's always UTC, call that out, if not, label with timezone)

chmccreery (commented):
Code looks good to me - might need another review/approval from the privacy side

cli/src/semgrep/meta.py
ThatDevShparki: Yes, this was sent with each Finding, we just want to include it in the scan meta as well moving forward

PRIVACY.md
ThatDevShparki: It was confusing to me as well. I opted to not update the language outside the scope of these changes but that should maybe a rewording here would be nice. I can ping technical writers to check in on this

chmccreery (approved):
Approving to unblock, but I do feel that the time-zone comment is important to resolve. I think it probably makes sense to convert it to UTC for consistency with other timestamps that are saved in the app.

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines-overwrite-autodetected-variables/meta.json
ThatDevShparki: I've updated the code the match the same format that we use with findings, something similar to `2023-07-24T23:38:18`

werrett:
LGTM. No concerns around collecting timestamps like this.

werrett (approved):

PR #8543, opened 2023-08-24
Update Python Help Section and Begin Standardization of Error Display

zzeleznick:
# Description
This PR intends to improve the developer experience for new users by providing an updated help section.

As an implementation note, `semgrep --help` invokes the help section from `osemgrep` which is the static output (without color) of running the pysemgrep help (e.g. `./cli/bin/pysemgrep --help` / `semgrep --legacy --help`). When running `semgrep` without any arguments, we default to `semgrep scan` which hooks into the python implementation and will show the colored help message.

### `semgrep` without any arguments
| Before | After |
| -- | -- |
| Screenshot 2023-08-24 at 5 33 35 PM | Screenshot 2023-08-24 at 4 02 41 PM |
| `$ semgrep` | `$ semgrep` |

### `semgrep` help

| Before | After |
| -- | -- |
| Screenshot 2023-08-24 at 4 15 41 PM | Screenshot 2023-08-25 at 11 46 13 AM |
| `$ semgrep --help` | `$ semgrep --help` |

## PR checklist:


cli/src/semgrep/cli.py
```diff
@@ -10,17 +10,25 @@
...
 
 
-@click.group(cls=DefaultGroup, default_command="scan", name="semgrep")
+@click.group(
+    cls=DefaultGroup, default_command="scan", name="semgrep", epilog=DEFAULT_EPILOGUE
+)
 @click.help_option("--help", "-h")
 @click.pass_context
 def cli(ctx: click.Context) -> None:
+    """
```
aryx, line 52: This became useless to change because this is not used anymore. Semgrep --help is now handled by osemgrep so this part of pysemgrep is dead.

cli/src/semgrep/commands/lsp.py
```diff
@@ -23,7 +23,7 @@ def init_log(name: str, level: int, logfile: Optional[str] = None) -> None:
     log.addHandler(handler)
 
 
-@click.command()
+@click.command(hidden=True)
```
aryx, line 26: there is no option to hide the visibility of flags in osemgrep, so this modification is useless.
The help text for command is automatically generated by the Cmdliner OCaml library which generates a man page
.

cli/src/semgrep/commands/wrapper.py
```diff
@@ -4,12 +4,26 @@
...
 
 
+class CommandSection(click.Command):
+    section: str = (
+        "Commands"  # Name of the section to group the command under in the help menu
+    )
+    priority: int = 0  # Order of the command group (lowest is first)
+
+
+class AdvancedCommand(CommandSection):
```
aryx, line 22: again, not useful as all of that is now handled by osemgrep.
If you want to improve the CLI UX, you need to do it on the OCaml side first.

aryx (changes requested):

aryx:
`we will need to call the python implementation for semgrep --help`
We will not. The change need to be in osemgrep for this, not pysemgrep. We're already not relying on Click and python for the help message and this week I'll remove all the help message for each flag as it becomes redundant with the help of the flag in osemgrep.

cli/src/semgrep/cli.py
zzeleznick: @aryx – Given that we're writing the help message manually for `osemgrep`, we could use the click to write the help message output/
  https://github.com/returntocorp/semgrep/blob/a0033e754c59573b6b9a38b5ac7c2373d8854eb6/src/osemgrep/cli/CLI.ml#L25C48-L26C39

cli/src/semgrep/commands/lsp.py
zzeleznick: This flag controls the visibility of the subcommand when running `pysemgrep`'s help not the command's help text.

cli/src/semgrep/cli.py
aryx: the toplevel help message is handled manually, but the semgrep xxx --help for each subcommand is generated automatically by Cmdliner (the equivalent of Click but for OCaml), and I want the  source of truth to be osemgrep, not pysemgrep. In fact this week I'll remove all the help message for each CLI args in pysemgrep because it's redundant with the CLI parsing in osemgrep.

cli/src/semgrep/commands/lsp.py
aryx: We don't want to hide lsp as a subcommand. cc @ajbt200128

zzeleznick: this was the direction from design

Screenshot 2023-08-25 at 2 06 51 PM

cc @adamboros

cli/src/semgrep/cli.py
zzeleznick: for clarity, this line here is only used for the toplevel help message

cli/src/semgrep/commands/wrapper.py
zzeleznick: This code change is only used for generating the help message (ordering the sections).

cli/src/semgrep/commands/lsp.py
ajbt200128: Yea, I'm not sure about the publish flag but I don't see why we'd want to hide the lsp command. If anything we should display it AND link to the editor integrations page

bmahe: @aryx can you detail why we would not want to hide the lsp subcommand?

aryx: I don't know when this design decision was discussed, but I think it would be better next time to involve me or austin about this kind of discussion as we have more context about how semgrep CLI is used.

aryx: To answer your question bruno, it's just what austin said, we actually want to advertise semgrep lsp more, so people know about it.

zzeleznick:
Closing in favor of returning only a static string on the OCaml side. Will use OCaml to display the help message instead of python on scan fallback without arguments (i.e. `semgrep`).

PR #8551, opened 2023-08-27
feat: implicit return support (for ruby and julia)

### LINKED ISSUE #8408: [Julia] Equivalence of operators and functions and of explicit or no `return` keyword
Most Julia operators are functions as well. For example, `1 + 2` is equivalent to `+(1, 2)`:
```julia
julia> Meta.dump(:(1 + 2))
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol +
    2: Int64 1
    3: Int64 2

julia> Meta.dump(:(+(1, 2)))
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol +
    2: Int64 1
    3: Int64 2
```

Semgrep doesn't catch this: https://semgrep.dev/playground/s/e06d

Also, in Julia the value of the last evaluated expression in a function is the return value - `f(x) = x` is the same as `f(x) = return x`. Semgrep doesn't catch this either: https://semgrep.dev/playground/s/vNgl

Having available a rule option, as @aryx suggested ("`options: treat_operator_as_funcalls: true`") for the first issue, would be nice.

I wouldn't want this to be the default for neither of the issues I mentioned. Most times when I want to match a function call I don't want to match operators as well. And what if I want to check if a function uses the `return` keyword explicitly? I wouldn't be able to do that by default anymore (unless maybe using regex).

**Use case**
I think it's useful to be able to match all usages of an operator in the form `$OP($ARG1, $ARG2, ...)` (for analysing/comparing the operands, for example), or to analyse the return value of a function (right now this can only be done for functions that explicitly return a value using the `return` keyword).

amchiclet:
The implementation revolves around 3 new fields added to `tin`
* inside_function
* parent_is_last_stmt_trans
* self_is_last_stmt

`inside_function` is initialized to false and becomes true whenever a function is visited.

`parent_is_last_stmt_trans` is initialized to true and when visiting each substmt, it tells the substmt whether itself upto root is true. This means that as soon as it hits the first false, this value will always be false when visiting the substmts.

`self_is_last_stmt` is initialized to true except for blocks. When we enter a block, it is set to false, except for only the last statement in that block which is set to true.

All of this is to figure out whether a statement is (transitively) the last statement inside a function or not.

Started with 2 languages which is Ruby and Julia. This feature is now enabled by default, but I also added a rule option to allow users to turn this off.

This is one of the sub-tasks for #8408.

Added new tests and tested for regressions with `make core-test`.

### DIFF
--- a/interfaces/Rule_options.atd
+++ b/interfaces/Rule_options.atd
@@ -153,6 +153,12 @@ type t = {
   (* Whether a rule should be considered for interfile analysis. *)
   ~interfile <ocaml default="false"> : bool;
 
+  (* Whether last expression of a block is treated the same as a return statement.
+     This option is specific only to languages that treat the last statement of
+     a block inside functions as a return statement, such as Ruby and Julia.
+  *)
+  ~implicit_return <ocaml default="true"> : bool;
+
   (* TODO: equivalences:
    *   - require_to_import (but need pass config to Js_to_generic)
    *)
--- a/src/matching/Generic_vs_generic.ml
+++ b/src/matching/Generic_vs_generic.ml
@@ -74,6 +74,39 @@ let hook_find_possible_parents = ref None
 (* Extra Helpers *)
 (*****************************************************************************)
 
+(* Helper functions below that are prefixed with imp_ret modify the environment to support implicit return. *)
+let imp_ret_env_enter_function_body (tin : tin) =
+  [
+    {
+      tin with
+      inside_function = true;
+      parent_is_last_stmt_trans = true;
+      self_is_last_stmt = true;
+    };
+  ]
+
+let imp_ret_env_enter_substmt (tin : tin) =
+  let parent_is_last_stmt_trans =
+    tin.parent_is_last_stmt_trans && tin.self_is_last_stmt
+  in
+  [ { tin with parent_is_last_stmt_trans; self_is_last_stmt = true } ]
+
+let imp_ret_env_enter_try_catch ~has_finally_clause tin =
+  let parent_is_last_stmt_trans =
+    if has_finally_clause then false
+    else tin.parent_is_last_stmt_trans && tin.self_is_last_stmt
+  in
+  [ { tin with parent_is_last_stmt_trans; self_is_last_stmt = true } ]
+
+let imp_ret_env_enter_finally ~parent_tin tin =
+  let parent_is_last_stmt_trans =
+    parent_tin.parent_is_last_stmt_trans && parent_tin.self_is_last_stmt
+  in
+  [ { tin with parent_is_last_stmt_trans; self_is_last_stmt = true } ]
+
+let imp_ret_env_enter_stmt_in_block ~is_last tin =
+  [ { tin with self_is_last_stmt = is_last } ]
+
 let env_add_matched_stmt rightmost_stmt (tin : tin) =
   [ extend_stmts_matched rightmost_stmt tin ]
 
@@ -994,6 +1027,8 @@ and m_expr ?(is_root = false) ?(arguments_have_changed = true) a b =
   | G.Ref (a0, a1), B.Ref (b0, b1) -> m_tok a0 b0 >>= fun () -> m_expr a1 b1
   | G.DeRef (a0, a1), B.DeRef (b0, b1) -> m_tok a0 b0 >>= fun () -> m_expr a1 b1
   | G.StmtExpr a1, B.StmtExpr b1 -> m_stmt a1 b1
+  (* implicit return *)
+  | G.StmtExpr { G.s = G.Return (_, Some a, _); _ }, _ -> m_implicit_return a b
   | G.OtherExpr (a1, a2), B.OtherExpr (b1, b2) ->
       m_todo_kind a1 b1 >>= fun () -> (m_list m_any) a2 b2
   | G.RawExpr a, B.RawExpr b -> m_raw_tree a b
@@ -2264,8 +2299,13 @@ and m_stmts_deep ~inside ~less_is_ok (xsa : G.stmt list) (xsb : G.stmt list) =
     when MV.is_metavar_ellipsis s ->
       (* less: for metavariable ellipsis, does it make sense to go deep? *)
       m_list__m_stmt xsa xsb
+  (* last element *)
+  | [ xa ], [ xb ] ->
+      imp_ret_env_enter_stmt_in_block ~is_last:true >>= fun () ->
+      m_stmt xa xb >>= fun () -> env_add_matched_stmt xb
   (* the general case *)
   | xa :: aas, xb :: bbs ->
+      imp_ret_env_enter_stmt_in_block ~is_last:false >>= fun () ->
       m_stmt xa xb >>= fun () ->
       env_add_matched_stmt xb >>= fun () ->
       m_stmts_deep ~inside ~less_is_ok aas bbs
@@ -2400,6 +2440,7 @@ and m_stmt a b =
   (* the general case *)
   (* ... will now allow a subset of stmts (less_is_ok = false here) *)
   | G.Block a1, B.Block b1 ->
+      imp_ret_env_enter_substmt >>= fun () ->
       m_bracket (m_stmts_deep ~inside:false ~less_is_ok:false) a1 b1
   (* equivalence: vardef ==> assign, and go deep.
    * coupling: with Visitor_AST.v_vardef_as_assign_expr which also deals with
@@ -2428,8 +2469,16 @@ and m_stmt a b =
         ~else_:(fail ())
   (* equivalence: *)
   | G.ExprStmt (a1, _), B.Return (_, Some b1, _) -> m_expr_deep a1 b1
+  (* implicit return *)
+  | ( G.Return (_, Some a1, _),
+      B.ExprStmt ({ e = B.Lambda { fbody = FBExpr b1; _ }; _ }, _) )
+  | G.Return (_, Some a1, _), B.DefStmt (_, FuncDef { fbody = FBExpr b1; _ }) ->
+      let* () = imp_ret_env_enter_function_body in
+      m_implicit_return a1 b1
+  | G.Return (_, Some a1, _), B.ExprStmt (b1, _) -> m_implicit_return a1 b1
   (* boilerplate *)
   | G.If (a0, a1, a2, a3), B.If (b0, b1, b2, b3) ->
+      imp_ret_env_enter_substmt >>= fun () ->
       m_tok a0 b0 >>= fun () ->
       (* too many regressions doing m_expr_deep by default; Use DeepEllipsis *)
       m_condition a1 b1 >>= fun () ->
@@ -2470,10 +2519,7 @@ and m_stmt a b =
       let* () = m_expr a1 b1 in
       m_tok asc bsc
   | G.Try (a0, a1, a2, a3), B.Try (b0, b1, b2, b3) ->
-      let* () = m_tok a0 b0 in
-      let* () = m_stmt a1 b1 in
-      let* () = (m_list m_catch) a2 b2 in
-      (m_option m_finally) a3 b3
+      m_try_catch_finally (a0, a1, a2, a3) (b0, b1, b2, b3)
   | G.Assert (a0, aargs, asc), B.Assert (b0, bargs, bsc) ->
       let* () = m_tok a0 b0 in
       let* () = m_arguments aargs bargs in
@@ -2594,6 +2640,20 @@ and m_label a b =
   match (a, b) with
   | a, b -> m_ident a b
 
+and m_try_catch_finally (a0, a1, a2, a3) (b0, b1, b2, b3) tin =
+  (let has_finally_clause =
+     match (a3, b3) with
+     | Some _, Some _ -> true
+     | _, _ -> false
+   in
+   let* () = imp_ret_env_enter_try_catch ~has_finally_clause in
+   let* () = m_tok a0 b0 in
+   let* () = m_stmt a1 b1 in
+   let* () = (m_list m_catch) a2 b2 in
+   let* () = imp_ret_env_enter_finally ~parent_tin:tin in
+   (m_option m_finally) a3 b3)
+    tin
+
 and m_catch a b =
   match (a, b) with
   | (at, a1, a2), (bt, b1, b2) ->
@@ -2605,7 +2665,7 @@ and m_catch_exn a b =
   (* dots: *)
   | G.CatchPattern (G.PatEllipsis _), _ -> return ()
   (* boilerplate *)
-  | G.CatchPattern a, CatchPattern b -> m_pattern a b
+  | G.CatchPattern a, B.CatchPattern b -> m_pattern a b
   | G.CatchParam a, B.CatchParam b -> m_parameter_classic a b
   | G.OtherCatch (a0, a1), B.OtherCatch (b0, b1) ->
       let* () = m_todo_kind a0 b0 in
@@ -2660,6 +2720,20 @@ and m_other_stmt_with_stmt_operator a b =
   | G.OSWS_Iterator, _
   | G.OSWS_Todo, _ ->
       fail ()
+
+and m_implicit_return a b tin =
+  let has_lang_support =
+    match tin.lang with
+    | Lang.Ruby
+    | Lang.Julia ->
+        true
+    | __else__ -> false
+  in
+  if
+    tin.config.implicit_return && has_lang_support && tin.inside_function
+    && tin.parent_is_last_stmt_trans && tin.self_is_last_stmt
+  then m_expr_root a b tin
+  else fail () tin
 
 (*****************************************************************************)
 (* Pattern *)
@@ -2884,6 +2958,7 @@ and m_function_definition a b =
       { B.fparams = b1; frettype = b2; fbody = b3; fkind = b4 } ) ->
       m_parameters a1 b1 >>= fun () ->
       (m_option_none_can_match_some m_type_) a2 b2 >>= fun () ->
+      imp_ret_env_enter_function_body >>= fun () ->
       m_function_body a3 b3 >>= fun () -> m_wrap m_function_kind a4 b4
 
 and m_function_body a b =
--- a/src/matching/Match_patterns.ml
+++ b/src/matching/Match_patterns.ml
@@ -37,6 +37,15 @@ let profile_mini_rules = ref false
 (* Types *)
 (*****************************************************************************)
 
+(* Extra context to pass to the environment in the visitor to support implicit return.
+ * More details in Matching_generic.tin.
+ *)
+type imp_ret_ctx = {
+  inside_function : bool;
+  parent_is_last_stmt_trans : bool;
+  self_is_last_stmt : bool;
+}
+
 (*****************************************************************************)
 (* Debugging *)
 (*****************************************************************************)
@@ -67,11 +76,22 @@ let match_e_e rule a b env =
   set_last_matched_rule rule (fun () -> GG.m_expr_root a b env)
   [@@profiling]
 
-let match_st_st rule a b env =
-  set_last_matched_rule rule (fun () -> GG.m_stmt a b env)
+let match_st_st ?(inside_function = false) ?(parent_is_last_stmt_trans = false)
+    ?(self_is_last_stmt = false) rule a b (env : Matching_generic.tin) =
+  set_last_matched_rule rule (fun () ->
+      let env =
+        {
+          env with
+          inside_function;
+          parent_is_last_stmt_trans;
+          self_is_last_stmt;
+        }
+      in
+      GG.m_stmt a b env)
   [@@profiling]
 
-let match_sts_sts rule a b env =
+let match_sts_sts ?(inside_function = false)
+    ?(parent_is_last_stmt_trans = false) rule a b env =
   set_last_matched_rule rule (fun () ->
       (* When matching statements, we need not only to report whether
        * there is match, but also the actual statements that were matched.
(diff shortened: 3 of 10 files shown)


amchiclet:
PR checklist:

amchiclet:
Sorry this PR is larger than I expected. If you'd like to discuss the changes offline, or would like me to break it down, please let me know.

amchiclet:
I see slow benchmarks, but also see them on other recent PRs, so I'm not entirely sure if this PR is related or not.

src/matching/Match_patterns.ml
```diff
@@ -37,6 +37,15 @@ let profile_mini_rules = ref false
 (* Types *)
 (*****************************************************************************)
 
+(* Extra context to pass to the environment in the visitor to support implicit return.
+ * More details in Matching_generic.tin.
+ *)
+type imp_ret_ctx = {
```
aryx, line 43: I think it would be better to add this information in AST_generic.stmt instead of in tin.
We could have an extra pass while visiting AST_Generic that populate those fields, and then in Generic_vs_generic
we could just consult those fields; we would not need those imp_ret_enter_xxx because the job would have been done before. Maybe in Naming_AST.ml or in another visitor.

aryx: @IagoAbal what do you think? @brandonspark ?

aryx:
Otherwise the PR looks good. I would just change where we compute the impl_return stmt info.

src/matching/Match_patterns.ml
amchiclet: AST_generic actually does feel like a proper place to put these fields in. It seems like the new fields are info about the AST, and not about matching. I'll wait for more feedback.

src/matching/Generic_vs_generic.ml
```diff
@@ -2428,8 +2469,16 @@ and m_stmt a b =
         ~else_:(fail ())
   (* equivalence: *)
   | G.ExprStmt (a1, _), B.Return (_, Some b1, _) -> m_expr_deep a1 b1
+  (* implicit return *)
+  | ( G.Return (_, Some a1, _),
+      B.ExprStmt ({ e = B.Lambda { fbody = FBExpr b1; _ }; _ }, _) )
```
brandonspark, line 2474: I'm not super clear on why this case is here, could there be a comment?

src/matching/Match_patterns.ml
```diff
@@ -402,11 +444,29 @@ let check2 ~hook mvar_context range_filter (config, equivs) rules
...
+          (* Cases like ExprStmt(StmtExpr s) may cause duplicate matches because
+           * it's possible that
+           *   ExprStmt(StmtExpr s)
+           * will match once, and
+           *   s
+           * will also match again. Ideally, the generic ASTs would not include
+           * these cases, but it seems to be the case for at least Julia.
+           *)
+          match x with
+          | { s = ExprStmt ({ e = StmtExpr _; _ }, _); _ } ->
```
brandonspark, line 463: I think wherever possible, we should allow `Generic_vs_generic` to operate under the simplest assumptions, because those upstream issues are simply the fault of an incorrect translation, and should be changed. This one is on me, because this translation should never have happened in the first place, but I think we should avoid introducing this unrelated code where possible.

brandonspark (commented):
One thing which concerns me is how much `imp_ret` specific logic that this introduces to `Generic_vs_generic`. For instance, I can, with no other context, be reading `m_try_catch` or `check2`, and just see a blow up of logic about last stmts and whatnot. This is fine, but it's a lot of code for this feature, and in particular in a way that we haven't had to do before (this would be one of the first instances of such semantics-specific code in those places, I think).

Without having a whole lot of context, my question is primarily whether there is another way to do this. You probably have considered this already, but what would be the primary challenge in doing this preprocessing (or postprocessing) step on the AST of these languages with implicit returns, before we end up at matching? Can we catch this kind of thing more upstream? It's nice, I think, if we make simplifying steps to make the semantics clear in the Generic AST, as opposed to catching up on the Generic AST and doing a bunch of work to make evident what those properties are, so if we could insert these `Return`s explicitly in the AST before we arrive at matching, I think that would be quite nice.

When I added some fields to `tin` for matching-specific contextual things (in my case, wildcard imports, which is not yet merged), it was with the understanding that any other method would bloat the Generic AST massively, and cause type-level issues with the AST. I just want to be sure that we've exhausted all options before going here. I think that Pad is saying something similar to what I am, but the question of the hour seems to be _can we move this step any earlier?_ Whether by introducing a `Return` or storing it in the `stmt` itself.

src/matching/Match_patterns.ml
```diff
@@ -365,9 +399,17 @@ let check2 ~hook mvar_context range_filter (config, equivs) rules
            * TODO: bloom filter was removed, undo this inlining?
            *)
           let visit_stmt () =
+            let inside_function = env.extra.inside_function in
```
brandonspark, line 402: here is an example of where there is a bunch of introduced code 
which is not necessarily clearly related to the outer context (at the very least, I'd like a comment)

src/matching/Match_patterns.ml
brandonspark: I would be fine with this, but I am firstly wondering about why we could not insert a `Return` into the AST in the prescribed spots, re my review below.

amchiclet:
The code with these new fields passed as function arguments looked even messier. :) I like the idea of moving this earlier.

> Whether by introducing a Return or storing it in the stmt itself.

We would probably need to introduce an additional constructor `ImplicitReturn` to differentiate between the explicit ones in case the user doesn't want to match implicit returns. At this point, I'm leaning a bit towards storing it in `stmt`.

src/matching/Match_patterns.ml
amchiclet: Originally, I think I had a warning log in there, which somehow got removed. I agree with fixing it earlier. I'm just thinking about catching it for all languages.

Makes me think, maybe it might be a good idea to have a unit test to catch these patterns for any parser introduced/updated.

aryx:
I like ImplicitReturn because adding fields in stmt also increase memory
usage as you add it for every stmts.

brandonspark:
Could we also just wrap `Return` with a boolean, rather than introduce a new constructor?

brandonspark:
(these are all coding details, I trust you to pick the right option at this point)

amchiclet:
> Could we also just wrap `Return` with a boolean, rather than introduce a new constructor?

Do you mean something like this

```
and stmt_kind =
| ...
| Return of return
| ...

and return = {
  ast: tok * expr option * sc;
  is_implicit: bool;
}
```

or did you have something else in mind?

brandonspark:
> > Could we also just wrap `Return` with a boolean, rather than introduce a new constructor?
> 
> Do you mean something like this
> 
> ```
> and stmt_kind =
> | ...
> | Return of return
> | ...
> 
> and return = {
>   ast: tok * expr option * sc;
>   is_implicit: bool;
> }
> ```
> 
> or did you have something else in mind?

yes, I like that!

amchiclet:
Thanks for clarifying! I'll think about it a bit more.

interfaces/Rule_options.atd
```diff
@@ -153,6 +153,12 @@ type t = {
   (* Whether a rule should be considered for interfile analysis. *)
   ~interfile <ocaml default="false"> : bool;
 
+  (* Whether last expression of a block is treated the same as a return statement.
+     This option is specific only to languages that treat the last statement of
```
IagoAbal, line 157: Is this only for the last statement? What the "last statement" means here exactly?

src/matching/Match_patterns.ml
```diff
@@ -37,6 +37,15 @@ let profile_mini_rules = ref false
 (* Types *)
 (*****************************************************************************)
 
+(* Extra context to pass to the environment in the visitor to support implicit return.
+ * More details in Matching_generic.tin.
+ *)
+type imp_ret_ctx = {
+  inside_function : bool;
+  parent_is_last_stmt_trans : bool;
```
IagoAbal, line 45: Would need more docs for these two last fields.

IagoAbal: Never mind I missed the comment pointing to `tin`, although maybe `tin` should just incorporate this type instead of "inlining" it.

src/matching/Matching_generic.mli
```diff
@@ -9,7 +9,10 @@ type tin = {
   config : Rule_options.t;
   deref_sym_vals : int;
       (** Counts the number of times that we "follow" symbollically propagated
-    * values. This is bound to prevent potential infinite loops. *)
+       * values. This is bound to prevent potential infinite loops. *)
+  inside_function : bool;
```
IagoAbal, line 13: Shouldn't the docstrings be in the `.mli` ?

interfaces/Rule_options.atd
amchiclet: I'll add better documentation. Basically I'm trying to say that a "last statement" is any expression/statement that is the last to be evaluated before returning from a function no matter what path it took. So there can be many "last statements" in a function.

src/matching/Matching_generic.mli
amchiclet: You're right! Thanks for pointing this out. From @brandonspark and @aryx 's input, I think I'm actually moving away from putting these fields in `tin`, but your feedback still applies anyway, so I'll keep that in mind.

src/matching/Match_patterns.ml
```diff
@@ -365,9 +399,17 @@ let check2 ~hook mvar_context range_filter (config, equivs) rules
...
+            let inside_function = env.extra.inside_function in
+            let parent_is_last_stmt_trans =
+              env.extra.parent_is_last_stmt_trans
+            in
+            let self_is_last_stmt = env.extra.self_is_last_stmt in
             !stmt_rules
             |> List.iter (fun (pattern, rule) ->
-                   let matches_with_env = match_st_st rule pattern x m_env in
+                   let matches_with_env =
+                     match_st_st ~inside_function ~parent_is_last_stmt_trans
```
IagoAbal, line 410: Maybe it's better to just pass that `env.extra` ?

src/matching/Match_patterns.ml
IagoAbal: I agree with Yoann here, I would rather keep much of this logic separated. Also, wouldn't be better if we just used the CFG to find out what statements are the "return"ing ones? AST_to_IL probably needs some fixes.

amchiclet: > why we could not insert a Return into the AST in the prescribed spots, re my review below.

In case somebody comes across this thread and has the same question, here's [said discussion](https://github.com/returntocorp/semgrep/pull/8551#issuecomment-1696028357).

> Also, wouldn't be better if we just used the CFG to find out what statements are the "return"ing ones? AST_to_IL probably needs some fixes

@IagoAbal I want to make sure I understand you correctly.

You are saying that AST_to_IL has a lot of existing functionality with CFG related stuff. But to find returning statements, AST_to_IL may need a bit more changes.

Is my understanding correct, or did you mean something else?

aryx: yes but we don't build the IL until we need it in tainting, so it's probably simpler to just extend Naming_AST.ml 
or add another pass in Parse_target.parse_and_resolve_name that do this special tagging of stmts, or
that change some node to add a new ImplReturn AST_generic stmt node.

IagoAbal:
> I like ImplicitReturn because adding fields in stmt also increase memory usage as you add it for every stmts.
> […](#)

If we use bit-vectors like for `id_flags` it may not be a big problem, and we can reuse for other things in the future. Rewriting Generic to introduce a new `ImplicitReturn` will also increase memory allocation.  Would need to benchmark.

aryx:
True. Ok let's go with adding extra field to AST_generic.stmt then, marking whether a stmt is the last one or whatever is needed.

src/matching/Match_patterns.ml
IagoAbal: We build the IL already for constant propagation which we do right away after parsing a file, even before any matching is done.

> You are saying that AST_to_IL has a lot of existing functionality with CFG related stuff. But to find returning statements, AST_to_IL may need a bit more changes.

Determining what are the implicit returns should be easy once you have a CFG, I think. The problem is that the construction of the IL may need to be fixed for Ruby/Julia so that that CFG is the correct one. That said, doing that will benefit taint analysis for those languages too.

Since this is something that taint also needs to do (and we also have a limited inter-proc const-prop), I think it's better we do this once in AST_to_IL.

amchiclet:
Discussed this PR with @IagoAbal offline. Here's the summary (adding my own comments too):
* Determining returning statements from the CFG seems more robust than an adhoc visitor.
* An adhoc visitor may not be that bad though. We cannot think of any corner cases at the moment. Maybe things related to GOTO?
* IL.stmt doesn't seem to have a link back to the generic AST at the moment because it's [commented out](https://github.com/returntocorp/semgrep/blob/2f2741327928af17de36c6fb442feacdd628006c/src/il/IL.ml#L322), but we may not really need it, since taint uses expressions.
* Whether we annotate `Return` or add `ImplicitReturn`, we will need to modify the generic AST to add the nodes. Matching only requires adding new cases. For taint, the IL will need to point to the expression which is returned.
* If we annotate `StmtExpr`, then we may not need to modify the generic AST. Matching only requires adding new cases. For taint, the IL will already be pointing to the expression node.
* No matter what approach, IL will benefit from this change. I think currently there's a [hack](https://github.com/returntocorp/semgrep/blob/2f2741327928af17de36c6fb442feacdd628006c/src/analyzing/AST_to_IL.ml#L1599) that can be modified after this change.
* Annotating with bit vectors is a potential optimization if using bools increate memory footprint. My gut feeling is that it doesn't matter for now because we're only tracking one thing. In fact, using bit vectors to track one bit may increase code complexity. Bit vectors may come in handy in the future, but not sure if this is over-engineering or not. If benchmarking shows that bools are bad, then yes, maybe we can look at bit vectors.

To me, the tradeoffs seem hard to fully grasp in advance, especially because I'm not familiar with IL, so I will just go ahead and try out what seems to be the robust and easy-to-understand thing, which is
1. using the CFG to check for returning statements
2. adding `ImplicitReturn` nodes.

I'll see what I learn from that and may update my approach. If anyone has further thoughts, let me know.

IagoAbal:
Nice summary @amchiclet !

> adding ImplicitReturn nodes.

I am not very fond of this because it extends Generic with yet another node (which we usually try to avoid), and it requires rewriting the Generic tree so potentially re-allocating a lot of stuff (and I don't think we do this for anything else). Naming, constant-propagation etc all rely on `ref`s in the Generic tree that we update accordingly. Also just recently we added two flags to ids and we did that via `id_flags` (yet another `ref`). So to me that is the "standard" we have set, and adding new nodes for recording such properties would deviate from that. It's always OK to deviate from the standard if needed, but then I would expect to have a reason for why it is needed. In the IL it's different, if we had to identify implicit returns already during AST-to-IL (which may be the case), then it is better to just add the `Return` nodes in the IL.

amchiclet:
Thanks for some more context!

> it [ImplicitReturn] requires rewriting the Generic tree so potentially re-allocating a lot of stuff

This comment applies to adding extra `Return` nodes to the Generic tree as well right?

So it seems, for Generic, annotating `StmtExpr` with additional info may be a good next direction to try first.

To paraphrase your words, modifying the Generic AST after analysis (as opposed to when building the tree) requires more book-keeping and isn't consistent with the way extra info is typically added to the AST. For IL, adding the `Return` nodes during AST-to-IL is fine because we're just about to build the tree.

amchiclet:
Some preliminary results while experimenting on ideas.

Here's a ruby function

```
# test.rb
def f()                                                                                                                                                                                                                                                                       
  if true                                                                                                                                                                                                                                                                     
    0                                                                                                                                                                                                                                                                         
  else                                                                                                                                                                                                                                                                        
    1                                                                                                                                                                                                                                                                         
  end                                                                                                                                                                                                                                                                         
end
```

and its CFG obtained by running `sc -cfg_il test.rb`

![cfg](https://github.com/returntocorp/semgrep/assets/728595/d49b19ac-affa-48d2-ae6d-20a9cf10ef48)

So it does seem like working backwards from the exit node to find the first non-join node along every path would be a strategy that works. (I probably need to figure out if there are other nodes than join that I should be aware of too.)

The flow would probably look like
1. Create CFG without [the hack to insert implicit returns](https://github.com/returntocorp/semgrep/blob/fe2c9deaa7f1d0d0d565ca07bc751f34e21ec530/src/analyzing/AST_to_IL.ml#L1599-L1620)
1. Use the CFG to find which expressions are executed before exiting
    * Currently the CFG only supports [`predecessors`](https://github.com/returntocorp/semgrep/blob/fe2c9deaa7f1d0d0d565ca07bc751f34e21ec530/src/il/CFG.ml#L50), so may need to add functionality to support this.
1. Augment [`StmtExpr`](https://github.com/returntocorp/semgrep/blob/fe2c9deaa7f1d0d0d565ca07bc751f34e21ec530/libs/ast_generic/AST_generic.ml#L765) or maybe [`expr`](https://github.com/returntocorp/semgrep/blob/fe2c9deaa7f1d0d0d565ca07bc751f34e21ec530/libs/ast_generic/AST_generic.ml#L610) with something like `is_implicit_return: ref bool`
    * Augmenting `StmtExpr` seems to break a lot of things because rather than `StmtExpr of stmt`, it would need to be something like `StmtExpr of expr_stmt` instead. Not a big deal if it's the right thing to do, but tedious.
    * Augmenting `expr` is just adding a field and seems less intrusive because I would just need to modify how the record is [initialized by default](https://github.com/returntocorp/semgrep/blob/fe2c9deaa7f1d0d0d565ca07bc751f34e21ec530/libs/ast_generic/AST_generic.ml#L2118). Not sure about performance hit, so need to benchmark.
1. Find all expressions that return from a function and trace it back to the generic AST. I think this can be done through [`eorig`](https://github.com/returntocorp/semgrep/blob/fe2c9deaa7f1d0d0d565ca07bc751f34e21ec530/src/il/IL.ml#L217).
1. Annotate that Generic AST by setting the flag `is_implicit_return` to true. (When I say "setting the flag", I say it in a loose sense. It can be done through bit vectors.)

At this point, our Generic AST is properly annotated with implicit return info, the next time we build the CFG, such as in constant propagation, we can insert [`IL.Return`](https://github.com/returntocorp/semgrep/blob/fe2c9deaa7f1d0d0d565ca07bc751f34e21ec530/src/il/IL.ml#L334)  in the right place.

This whole process involves creating the CFG once to figure out which expressions are returning, and after that re-create or modify the CFG to insert `IL.Return` statements at proper places. Or it could be done in one go. Not sure, but I'll probably know more when I try to implement it.

I think I'm going to go ahead and try this out. Feel free to comment if anyone can foresee issues besides a potential performance hit, which would need benchmarking to say for sure.

IagoAbal:
> This comment applies to adding extra Return nodes to the Generic tree as well right?

Yep.

> is_implicit_return: ref bool

Better  a `mutable` field, it's what we are using in `expr` and `stmt` at the moment. There is a mix of `ref` and `mutable` at the moment, unfortunately.

> Augmenting StmtExpr seems to break a lot of things because rather than StmtExpr of stmt, it would need to be something like StmtExpr of expr_stmt instead. Not a big deal if it's the right thing to do, but tedious.

Can you annotate `StmtExpr` ? Ideally we annotate `stmt` instead of `expr`, but if we do this from the CFG we don't have access to the `stmt`s. 🤔 Unless we do something special for `ExprStmt` (I'm assuming all implicit returns involve `ExprStmt`).

amchiclet:
Some updates:
* I haven't tried annotating `StmtExpr`, but annotating `expr` definitely works. Annotating `expr` allows me to match `Return (_, Some e1, _)` with `ExprStmt (e2, _)` when `e2.is_implicit_return` is true.
* With this change, I can remove the [implicit return hack](https://github.com/returntocorp/semgrep/blob/9977d62927e4e597cb7422cda619177369a335b0/src/analyzing/AST_to_IL.ml#L1610-L1619), and still get implicit return support.

A lot of the new tests are passing, but some are failing. This is mainly because how the CFG is created, so to fix it, I would probably need to update CFG_build too. (It can be done separately, while leaving the failing tests in this PR as TODO items.

Basically here are the issues:
* The CFG of try-catch-finally statement doesn't look correct. (This is how the CFG is created, but it affects the correctness of implicit return.) In particular, the last statement in `try` flows to `catch`. Ideally, any statement that can cause an exception in `try` flows to `catch`. Additionally the last statement in `try` should flow to `finally` too.
* Try-catch-else-finally support in the CFG isn't implemented. I get a TODO block.
* Switch cases in Ruby are not fallthrough, but the CFG always falls through. I think we probably need a flag to differentiate languages that fallthrough by default, like C, and ones that don't, like Ruby.

Another thing that definitely needs to be fixed if I'm proceeding with this plan is that taint analysis in Rust has a new failing test after removing the implicit return hack.

The problem with Rust is that the code
```
fn foo(s:String) -> String {
  s
}
```

is giving me an empty CFG (only an enter and an exit node). The code is [here](https://github.com/returntocorp/semgrep/blob/9977d62927e4e597cb7422cda619177369a335b0/src/analyzing/AST_to_IL.ml#L1196-L1201).

What's happening is that there are no instructions pushed to the IL statement stack if we're using the existing variable `s`. So there are no IL statements created in this case.

I'll look more into this tomorrow. Maybe we can just add a new aux variable whenever we see an `ExprStmt` with a `Fetch` inside.

aryx:
Thx for the update. I think to move forward it would better to fix the 3 issues you mention separately, independently of the implicit_return. Once the 3 issues are fixed we can resume work on this implicit_return.

aryx:
4 issues actually because there's also the bug with the Rust code you mentioned.

amchiclet:
I agree with the plan. I'll start working on the CFG issues, and return (😄) here later.

amchiclet:
An updated implementation of implicit return support is at https://github.com/returntocorp/semgrep/pull/8826.

Closing this PR and continuing the discussion there.

PR #8556, opened 2023-08-28
fix: extract semgrep python wheel in a more robust way

tpetr:
We've been seeing nightly verification failures due to the zip file containing the aarch64 python wheel being corrupted: https://github.com/returntocorp/semgrep/actions/runs/6002039013/job/16278300160

It turns out we're hitting an edge case with how we're extracting the zip file out of the semgrep image: In `docker run --platform linux/arm64 --rm ${{ steps.build-semgrep-wheel.outputs.imageid }} cat cli/dist.zip > /tmp/dist.zip`, not all of stdout is being written to `/tmp/dist.zip`.

This PR takes a different approach which should be more robust (and also doesn't require us running anything inside the semgrep container):
1. Create a new container based on the semgrep image
2. Use `docker export` to export the container's filesystem as a tarball
3. Only keep `dist.zip`

/cc @brandonspark 

test plan:
- confirmed commands worked locally and that the resulting zipfile was valid
- checks pass

PR checklist:

### DIFF
--- a/.github/workflows/build-test-manylinux-aarch64.yaml
+++ b/.github/workflows/build-test-manylinux-aarch64.yaml
@@ -38,6 +38,13 @@ jobs:
 
           # clean up after ourselves
           docker rm $CONTAINER_ID
+
+          # note: this was originally accomplished by running `cat` from within the container and redirecting stdout to a file:
+          #
+          # docker run --platform linux/arm64 --rm ${{ steps.build-semgrep-wheel.outputs.imageid }} cat cli/dist.zip > /tmp/dist.zip
+          #
+          # we ended up hitting an edge case where the final ~100kb doesn't always get written to disk, which results in a corrupted zip file.
+          # our theory is that the pipe between the container and the host is being closed prematurely.
       - uses: actions/upload-artifact@v3
         with:
           name: manylinux-aarch64-wheel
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-manylinux-aarch64.yaml
```diff
@@ -26,9 +26,19 @@ jobs:
...
+          # load the docker image containing the semgrep python wheel
           docker load --input /tmp/image.tar
-          docker run --platform linux/arm64 --rm ${{ steps.build-semgrep-wheel.outputs.imageid }} cat cli/dist.zip > /tmp/dist.zip
+
+          # create a new docker container using the image we just loaded
+          # note: `docker create` simply prepares the container filesystem -- nothing is executed!
+          CONTAINER_ID=$(docker create ${{ steps.build-semgrep-wheel.outputs.imageid }})
+
+          # use `docker export` to extract the python wheel zipfile out of the container
+          docker export $CONTAINER_ID | tar xv semgrep/cli/dist.zip
```
brandonspark, line 37: could you add a comment explaining what the alternative used to be and why it was wrong? think it would help to read this code in the future (and also its not clear to me why the previous method wasn't writing all of `stdout`)

brandonspark (approved):
thanks so much!

aryx (approved):

bmahe (approved):

PR #8558, opened 2023-08-28
feat: Temporary CLI UI to display secrets validation.

akuhlens:
# What
Adds textual output to the CLI when a validation_state is returned besides the default `NO_VALIDATOR`. This is a temporary change and is slated to be replaced in a week or two. This is just to give some feedback on the CLI for alpha and beta testers.

## PR checklist:

  - Not sure if this is considered user facing since it is only for beta testing purposes.

# Testing
Manual test in [this PR for semgrep-proprietary](https://github.com/returntocorp/semgrep-proprietary/pull/934).

### DIFF
--- a/cli/src/semgrep/formatter/text.py
+++ b/cli/src/semgrep/formatter/text.py
@@ -656,6 +656,7 @@ def print_text_output(
         ):
             console.print(line)
 
+        # Temporary CLI UI until a more thorough implementation.
         if "validation_state" in rule_match.extra:
             validation_state = rule_match.extra["validation_state"]
             if validation_state == "NO_VALIDATOR":
--- a/cli/src/semgrep/formatter/text.py
+++ b/cli/src/semgrep/formatter/text.py
@@ -549,7 +549,6 @@ def print_text_output(
     dataflow_traces: bool,
 ) -> None:
     last_file = None
-    last_rule_id = None
     last_message = None
     sorted_rule_matches = sorted(rule_matches, key=lambda r: (r.path, r.rule_id))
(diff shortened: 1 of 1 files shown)


cli/src/semgrep/formatter/text.py
```diff
@@ -656,6 +656,24 @@ def print_text_output(
...
+            if validation_state != "NO_VALIDATOR":
+                msg = ""
+                if validation_state == "CONFIRMED_VALID":
+                    msg = "Semgrep confirmed this secret is still valid."
+                elif validation_state == "CONFIRMED_INVALID":
+                    msg = "Semgrep confirmed this secret is invalid."
+                elif validation_state == "CONFIRMED_ERROR":
+                    msg = "Semgrep encountered a network error while trying to validate this secret."
+                console.print(
+                    f"{8 * ' '}{with_color(Colors.foreground, msg, bold=True)}"
```
kopecs: Should we print a newline at the end of this as well? It seemed a bit dense to me / hard to tell when the finding ended and the next one started.

cli/src/semgrep/formatter/text.py
```diff
@@ -656,6 +656,24 @@ def print_text_output(
...
+                )
+        else:
+            console.print("\n")
```
nmote: Won't this add an extra newline to the output if `validation_state` is not available? Shouldn't we aim to keep things the same as before in that case?

nmote (approved):

cli/src/semgrep/formatter/text.py
akuhlens: Right you are.

PR #8566, opened 2023-08-29
matching: Make `metavariable-type` default to false

IagoAbal:
Like every other metavariable-xyz operator.

In case we could not evaluate `metavariable-type`, we were defaulting to "true", that is, to let the range pass. But every other metavariable-xyz operator defaults to "false", if it cannot be evaluated then it filters out the range.

Also, this removes a `logger#info` reporting ranges filtered out by `metavariable-type` that was likely to be noisy, and no other `metavariable-xyz` operator logs such situations anyways.

Closes PA-3027

test plan:
```yaml
    patterns:
      - pattern: private int $X;
      - metavariable-type:
          metavariable: $Y # $Y is not bound
          type: int
```
no longer produces a match when run on
```java
class Test {
  private int x;
}
```

PR checklist:


tests/rules/metavar_type_invalid.yaml
```diff
@@ -0,0 +1,11 @@
+rules:
+  - id: test
+    message: Test
+    languages:
+      - java
+    severity: WARNING
+    patterns:
+      - pattern: private $T x;
+      - metavariable-type:
+          metavariable: $T
```
aryx: why this is not matching??? Or did you make a typo and you wanted to write 'metavariable: $X' (which is not bounded in the pattern, like you had in the changelog).

aryx (changes requested):

tests/rules/metavar_type_invalid.yaml
ihji: > why this is not matching???

`$T` represents a type identifier so it doesn't match with a type itself. `private int $T` would match.

And I think Iago also wanted to write `metavariable: $X`.

ihji (approved):

changelog.d/pa-3027.fixed
```diff
@@ -0,0 +1,12 @@
+When `metavariable-type` cannot be evaluated then it defauls to "false", that is,
+it filters out the range. Therefore e.g. this rule:
+
+```yaml
+    patterns:
+      - pattern: private $T x
```
aryx: This example is confusing to me because I forgot metavariable-type is about checking the identifier
type, not the type itself.
So maybe `private int $X` and the metavariable-type: metavariable: $Y
would be better?
Even though we should really syntactically reject such rules in the metachecker.

tests/rules/metavar_type_invalid.yaml
aryx: oh right, I got confused. the example is a bit confusing though.

aryx (approved):

tests/rules/metavar_type_invalid.yaml
IagoAbal: Ooops, sorry, I wrote the wrong test indeed.

changelog.d/pa-3027.fixed
IagoAbal: Ignoring I wrote the wrong tests, I agree this should be caught by the meta checker. I was also thinking whether to write a test at all for this.

IagoAbal:
Once I fixed the test I actually cannot commit it because it makes the test suite fail. No "matching errors" allowed (despite they are more like warnings than fatal errors). So I just inlined the test example in the test plan.

IagoAbal:
> Exception raised: `missing field 'check_id' in JSON object of type 'CoreMatch'`

This error seems unrelated to this PR, probably Pro needs to be updated to use the latest Semgrep interfaces.

aryx:
I did update semgrep-pro, but we still get those errors.

PR #8567, opened 2023-08-29
fix(julia): distinguish `import` from `using`

brandonspark:
## What:
This PR makes it so that, when matching in Julia, we distinguish `using` imports from `import` imports, which are two different variants on import resolution.

## Why:
These are semantically different, and would be useful to match separately.

## How:
Just added a `d_attr` attribute which separates the two. While I was there, I did some spring cleaning and fixed the annoying `ExprStmt (StmtExpr ...)` thing.

## Test plan:
`make test`

Closes PA-3028

PR checklist:


libs/ast_generic/AST_generic_helpers.mli
```diff
@@ -5,6 +5,7 @@ val str_of_ident : AST_generic.ident -> string
 exception NotAnExpr
 
 val expr_to_pattern : AST_generic.expr -> AST_generic.pattern
+val expr_to_stmt : AST_generic.expr -> AST_generic.stmt
```
aryx, line 12: maybe a small comment that if the expr was an StmtExpr, we actually return the embedded stmt (and not generate an ExprStmt (StmtExpr ...)

aryx (approved):

PR #8570, opened 2023-08-30
Add http testing framework

ajbt200128:
PR checklist:


libs/networking/Testing_client.mli
```diff
@@ -0,0 +1,14 @@
+type test_response = { response : Cohttp_lwt.Response.t; body_path : string }
```
aryx, line 31: maybe you can add some explanations in the .mli on how to use this library. Maybe the code of a full test using
those mocks.

libs/networking/http_helpers.ml
```diff
@@ -1,17 +1,26 @@
...
 (* This way we can use the async methods in the language server,
    and other places too
 *)
 
+(*****************************************************************************)
+(* Client *)
+(*****************************************************************************)
+
+(* Create a client reference so we can swap it out with a testing version *)
+let client_ref = ref (module Client : Cohttp_lwt.S.Client)
```
aryx: first class modules!

aryx (commented):
Nice! Just need an example now.

libs/networking/http_helpers.ml
akuhlens: I didn't know you could do that!

aryx: https://dev.realworldocaml.org/first-class-modules.html#:~:text=First%2Dclass%20modules%20are%20ordinary,language%20to%20use%20them%20effectively.
But we should avoid using such advanced features. We should use them only as a last resort (and this look like a good use case given the way the Cohttp library is written).

libs/networking/Testing_client.mli
ajbt200128: Yea, was just a POC to show andre, should be updated now

libs/networking/http_helpers.ml
ajbt200128: Yea, it was either this or functors and that seemed too heavy handed.

ajbt200128:
Also added a test bytecode target, and an action to run tests w/ earlybird. And added a basic LS test

src/osemgrep/language_server/Unit_LS.ml
```diff
@@ -224,5 +224,25 @@ let processed_run () =
   in
   pack_tests "Processed Run" tests
 
+let session_rules () =
+  let with_ci_client =
+    let make_fn req body =
+      ignore body;
+      Testing_client.check_method req "GET";
+      Lwt.return Testing_client.(basic_response "./tests/ls/ci/response.json")
```
akuhlens, line 232: Could you ensure that there is some easy way to check the message against a file here? My concern is that Request objects from cohttp are pretty opaque, just want to ensure that there isn't more code needed for this, or if we need to introduce a non-opaque request similar to `Rule.request`

src/tests/dune
```diff
@@ -25,6 +25,7 @@
     osemgrep_targeting
     osemgrep_language_server
   )
+ (modes native byte)
```
aryx: maybe add a comment about byte, which is it allows to produce a test.bc and use
earlybird on it

Makefile
```diff
@@ -188,6 +188,12 @@ core-test:
 	./_build/default/src/tests/test.exe --show-errors --help 2>&1 >/dev/null
 	./scripts/run-core-test
 
+.PHONY: core-test-bc
+core-test-bc:
```
aryx: you can remove the core- prefix here

.vscode/tasks.json
```diff
@@ -26,6 +26,14 @@
         "kind": "build"
       }
     },
+    {
+      "label": "Build Test Bytecode",
+      "type": "shell",
+      "command": "make core-test-bc",
```
aryx: make test-bc should be enough

akuhlens (commented):
This looks great! Thank you. Follow up on the comment that I made. I am okay with the majority of the work for that actually being in a seperate PR if need be, as long as you are confident the interface doesn't need to change.

libs/networking/Testing_client.ml
```diff
@@ -0,0 +1,148 @@
...
+ * special exception on linking described in file LICENSE.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the file
+ * LICENSE for more details.
+ *)
+
+(* Commentary *)
+(* This is a testing client to mock out http_helpers.ml so we
```
aryx: better to put that in the .mli.
In theory people should just read the interface file (.mli) to understand how to use your library.
They can look at the .ml to understand the implementation, get possibly comments about design decisions.

aryx (approved):

src/osemgrep/language_server/Unit_LS.ml
ajbt200128: @akuhlens there's a `check_body` helper function. If you check the header of Testing_client, there should be an example there

libs/networking/Testing_client.ml
```diff
@@ -0,0 +1,118 @@
...
+           (Cohttp.Header.to_string (Cohttp.Request.headers req)))
+  | Some actual_header ->
+      Alcotest.(check string) "header" header_val actual_header
+
+(*****************************************************************************)
+(* Entrypoint *)
+(*****************************************************************************)
+
+let with_testing_client make_fn test_fn () =
+  let prev_client = !Http_helpers.client_ref in
```
aryx, line 110: You should use Common.save_excursion (it's a function I wrote inspired by the same function in emacs lisp).

aryx: It's more tolerant to exn that could be thrown in test_fn ()

src/tests/dune
```diff
@@ -25,6 +25,7 @@
     osemgrep_targeting
     osemgrep_language_server
   )
+ (modes native) ; bytecode here for debugging
```
aryx, line 28: so why did you remove the bytecode from the modes?

libs/networking/Testing_client.ml
ajbt200128: I didn't know about that, very cool

src/tests/dune
ajbt200128: Forgot to remove the comment. Going to revisit this today in a separate PR (wanted to get this merged for andre)

PR #8572, opened 2023-08-30
Build and push rootless docker container

cgdolan:
PR checklist:

### DIFF
--- a/.github/workflows/build-test-docker.yaml
+++ b/.github/workflows/build-test-docker.yaml
@@ -5,6 +5,10 @@ name: build-test-docker
 on:
   workflow_dispatch:
     inputs:
+      docker-flavor:
+        required: true
+        type: string
+        description: A multi-line string in the format accepted by docker metadata tag action for the flavor of the image
       docker-tags:
         required: true
         type: string
@@ -34,6 +38,10 @@ on:
 
   workflow_call:
     inputs:
+      docker-flavor:
+        required: true
+        type: string
+        description: A multi-line string in the format accepted by docker metadata tag action for the flavor of the image
       docker-tags:
         required: true
         type: string
@@ -79,8 +87,7 @@ jobs:
         uses: docker/metadata-action@v4
         with:
           images: ${{ inputs.repository-name }}
-          flavor: |
-            latest=false # pass latest tag in inputs
+          flavor: ${{ inputs.docker-flavor }}
           tags: ${{ inputs.docker-tags }}
       - uses: depot/setup-action@v1
       - name: Build image
--- a/.github/workflows/build-test-docker.yaml
+++ b/.github/workflows/build-test-docker.yaml
@@ -23,6 +23,10 @@ on:
         type: string
         description: Dockerfile to build
         default: Dockerfile
+      target:
+        required: true
+        type: string
+        description: Dockerfile target to build
       enable-tests:
         required: true
         type: boolean
@@ -46,6 +50,10 @@ on:
         required: true
         type: string
         description: Dockerfile to build
+      target:
+        required: true
+        type: string
+        description: Dockerfile target to build
       enable-tests:
         required: true
         type: boolean
@@ -71,6 +79,8 @@ jobs:
         uses: docker/metadata-action@v4
         with:
           images: ${{ inputs.repository-name }}
+          flavor: |
+            latest=false # pass latest tag in inputs
           tags: ${{ inputs.docker-tags }}
       - uses: depot/setup-action@v1
       - name: Build image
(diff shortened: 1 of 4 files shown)


.github/workflows/build-test-docker.yaml
```diff
@@ -46,6 +50,10 @@ on:
         required: true
         type: string
         description: Dockerfile to build
+      target:
```
aryx, line 61: There is no way to factorize things with the workflow_dispatch above to avoid repeating ourselves?

.github/workflows/build-test-docker.yaml
```diff
@@ -71,6 +79,8 @@ jobs:
         uses: docker/metadata-action@v4
         with:
           images: ${{ inputs.repository-name }}
+          flavor: |
```
aryx: What is this flavor for? not sure I understand the use of this 'latest' tag

.github/workflows/tests.yml
```diff
@@ -327,6 +328,33 @@ jobs:
       repository-name: ${{ github.repository }}
       dry-run: false
 
+  build-docker-rootless:
+    uses: ./.github/workflows/build-test-docker.yaml
+    secrets: inherit
+    with:
+      docker-tags: |
+        type=semver,pattern={{version}},suffix=-rootless
```
aryx: what are all those docker tags for? I don't really understand the syntax but why do we need all
those things?

Dockerfile
```diff
@@ -232,3 +232,6 @@ RUN printf "[safe]\n	directory = /src"  > ~semgrep/.gitconfig && \
 # to interactively explore the docker image.
 CMD ["semgrep", "--help"]
 LABEL maintainer="support@semgrep.com"
+
```
aryx, line 236: maybe update the comment 10 lines above about the commented #USER semgrep. Maybe something like "see the rootless docker target below"

aryx (approved):
Thanks.
Why do we need this though? Was it impossible to make it rootless by default, so we don't need to build 2 images each time? Could you add a comment maybe in the Dockerfile next to the rootless target about why we provide 2 images (I guess it's easier for some customers, but would be nice to have some scenario in which the root one is needed and using the rootless would make too many customers suddently fail or something.

cgdolan:
> 

Sure, added a comment in the new nonroot docker stage explaining why we can't just make this the default.

Dockerfile
cgdolan: Done

.github/workflows/build-test-docker.yaml
cgdolan: The `flavor` parameter of the docker metadata action sets options that used for items in the `tags` section below. This makes it possible to add a common suffix to all tags in a single place rather than adding the suffix to each tag individually. The docs for this action are pretty good and explain the different configuration options in more detail if you're interested: https://github.com/docker/metadata-action#about

.github/workflows/tests.yml
cgdolan: This is the configuration that set different tags on images that get pushed to Dockerhub depending on the context: https://hub.docker.com/r/returntocorp/semgrep/tags. For example, `latest` for every commit to develop, `pr-nnnn` for every pull request, and `sha-nnnnnnnn` for every commit in the repo.

This block uses the same tagging strategy as the root container but adds a suffix `-nonroot` to each tag. In this PR, it results in two images being pushed - one with:
```Pushing image tags...
-> returntocorp/semgrep:pr-8572
-> returntocorp/semgrep:sha-9ed0477
```
and another with:
```
Pushing image tags...
-> returntocorp/semgrep:pr-8572-nonroot
-> returntocorp/semgrep:sha-9ed0477-nonroot
```

Once merged to develop, it'll cause two images to be pushed `returntocorp/semgrep:latest` and `returntocorp/semgrep:latest-nonroot`.

.github/workflows/build-test-docker.yaml
cgdolan: I looked around and unfortunately it doesn't seem like there's any way to template this :(

gedigi:
Please note [just above your change](https://github.com/returntocorp/semgrep/blob/develop/Dockerfile#L219-L222) that `USER semgrep` was disabled because of issues with CircleCI (and GitHub Actions.)

See:
- Open issue: https://github.com/actions/checkout/issues/956
- Failed GitHub Actions run: https://github.com/aptos-labs/aptos-core/actions/runs/6054718321/job/16432581065

aryx:
reverting this diff @gedigi. Should be fixed in 1.38.3 later today.

PR #8581, opened 2023-08-30
refactor http requests in secrets

akuhlens:
# What
Factors out the http requests from secrets to clean up code and make it more easily mocked for testing purposes.

# Testing
In semgrep-proprietary in [this PR](https://github.com/returntocorp/semgrep-proprietary/pull/933).
```bash
$ make test
# Manually check that the result is equivalent to tests/secrets/unverified/result.json
$ semgrep --pro --json --config tests/secrets/unverified/rules.yaml tests/secrets/unverified/src
```

PR checklist:

### DIFF
--- a/libs/networking/http_request.ml
+++ b/libs/networking/http_request.ml
@@ -7,7 +7,7 @@ open Common
 open Cohttp
 open Cohttp_lwt_unix
 
-let logger = Logging.get_logger[__MODULE__]
+let logger = Logging.get_logger [ __MODULE__ ]
 
 (* This type encodes a basic HTTP request; mainly used for in the secrets
  * post-processor; such that a basic http request like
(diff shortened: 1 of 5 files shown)


src/core/dune
```diff
@@ -16,6 +16,7 @@
    glob
    lib_parsing ; Parse_info.ml
    ast_generic
+   networking
```
aryx, line 19: This is a bit unfortunate ... to depend on networking code in "the core".

aryx: maybe add a comment next to the dependency saying it's just used for the Http_Request.t type in it which is used in Rule.ml for secrets.

libs/networking/http_request.ml
```diff
@@ -0,0 +1,79 @@
+(* Yosef Alsuhaibani
```
aryx, line 1: Please add also a Http_request.mli (you can also use capitalized module name, so they match how they are used; we sometimes forget to do it, but we should for new files at leasT).

aryx (changes requested):

aryx:
I hope this will work well with the work @ajbt200128 is doing also on mocking http stuff.

aryx:
Should we close this @akuhlens ?

PR #8587, opened 2023-08-30
Sinat101 max memory help update

sinat101:
Updated `--max-memory` help description to make it more clear/concise. To say "Defaults to 0 for all CLI scans." implies a different default for non-CLI scans, where in practicality the default is 0 for all scans except when using Pro Engine, where the default is 5000.

PR checklist:

### DIFF
--- a/changelog.d/max_memory_help.fixed
+++ b/changelog.d/max_memory_help.fixed
@@ -1,0 +1,1 @@
+Updated `--max-memory` help description to make it more clear/concise. To say "Defaults to 0 for all CLI scans." implies a different default for non-CLI scans, where in practicality the default is 0 for all scans except when using Pro Engine, where the default is 5000.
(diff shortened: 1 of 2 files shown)


aryx (approved):

aryx:
hmm, btw you also need to modify src/osemgrep/cli_scan/Scan_CLI.ml  you can do that in another PR since I've already merged this one.

sinat101:
@aryx are you sure? i looked in that file and searched for terms in the help text but did not find anything. from what i could tell `scan.py` was the only file containing the help text.

aryx:
It's here: https://github.com/returntocorp/semgrep/blob/4ca58c6548cddde90d375748ec7a49d19fc9ee18/src/osemgrep/cli_scan/Scan_CLI.ml#L294
but because there is a bug in the flag name; it should be "max-memory" not "max-memory-mb", so you can fix that and fix the related help message a few lines below!

sinat101:
> It's here:
> 
> https://github.com/returntocorp/semgrep/blob/4ca58c6548cddde90d375748ec7a49d19fc9ee18/src/osemgrep/cli_scan/Scan_CLI.ml#L294
> 
> 
> but because there is a bug in the flag name; it should be "max-memory" not "max-memory-mb", so you can fix that and fix the related help message a few lines below!

ahh, i was searching for "--max" and "5000", explains why i didn't find it. thanks!

PR #8588, opened 2023-08-30
refactor: Unkill `ppx_sexp_conv`

nmote:
In https://github.com/returntocorp/semgrep/pull/8196, we removed `ppx_sexp_conv`. It had an issue
(https://github.com/janestreet/ppx_sexp_conv/issues/40) which made versions newer than 0.14.3 unusable for us. However, 0.14.3 is incompatible with OCaml 5.0, so we had to remove it entirely in order to move towards that.

But, 0.16.0 fixed the issue.

I am working on shipping names/types from the Java standard library with the Pro Engine, to help with name and type resolution. I need a shelf-stable serialization format for this -- one where minor changes to the datatype will either be inconsequential or quickly detected. OCaml's built-in marshaling is not safe in the face of changes to the datatypes being marshaled, and can lead to memory corruption issues in these cases.

In contrast, we will get an error on deserialization if an sexp cannot be converted back to a valid OCaml value, so I'd like to use `ppx_sexp_conv` for this.

In this PR, I add `ppx_sexp_conv` version 0.16.0 as a dependency, and use it on a handful of data structures. The autogenerated code is not currently used, but I plan to use it in an upcoming change to the Pro Engine.

Test plan: Automated tests

PR checklist:


semgrep.opam
```diff
@@ -40,6 +40,7 @@ depends: [
   "ppx_hash"
   "ppx_inline_test"
   "ppx_expect"
+  "ppx_sexp_conv" {= "v0.16.0"}
```
aryx, line 43: this needs to be set in the dune-project file (and semgrep.opam is regenerated from it if you type 'make setup')
We keep the generated file semgrep.opam in git because there's a chicken and egg problem so we keep it.

aryx (approved):
Not a huge fan of sexplib (don't like those needed open Sexplib.Std for example), and also because I think we could just marshall with a version number attached
before the marshalled data (or just store the signatures as regular .java files; I'm surprised reading sexps is significantly faster than parsing java files with just signatures in it), but we already had this discussion so let's merge this.

semgrep.opam
nmote: I did set it in the `dune-project` file... did I miss something?

nmote:
> I think we could just marshall with a version number attached before the marshalled data

Given the extent to which these data structures are scattered across several different files, I don't want to rely on a manual process by which developers update version numbers in order to guarantee memory safety. That seems doomed to fail. If you feel strongly, I could see if there is a reasonable way to use ppx to generate a hash for a datatype (not a value, but the type definition itself), which we could safely rely on to reject marshaled data of the wrong version. Would you like me to do that?

> (or just store the signatures as regular .java files; I'm surprised reading sexps is significantly faster than parsing java files with just signatures in it)

Yes, it's substantially faster, and also doesn't require us to commit the java files, use a submodule, or require that every developer checks out their own copy separately. We could probably afford the build time hit, though it wouldn't be pleasant, but I really want to avoid making the java files available during day-to-day development.

semgrep.opam
aryx: Well I don't see any modification to dune-project it in this PR.

aryx:
nah, go merge it. It's fine like that. Let's try your approach.

PR #8589, opened 2023-08-30
[feat:CLI] ℹ️ Add Nudge for Supply Chain Awareness 

zzeleznick:
## Description
This PR adds a nudge to users who have not logged in and for those who haven't tried our **free** [supply chain dependency analysis feature](https://semgrep.dev/products/semgrep-supply-chain/).

### Demo 
| Before | After |
| -- | -- |
| <video src="https://github.com/returntocorp/semgrep/assets/5779832/eb00f90b-b1e5-4d45-a46a-7e1421f4e6c2"/> | <video src="https://github.com/returntocorp/semgrep/assets/5779832/e6bbcae1-0573-4fba-8238-b2ba10aa5ebb"/> | 

### Adds
- messaging to improve awareness of our supply chain product
- section title for the progress bar
- emojis for attention callout to version update and no findings bottom banner

### Modifies
- spacing between section headers and tables 
- behavior of when the tables should display
- behavior of when the rule counts should display

<hr>

### DIFF
--- a/cli/bin/pysemgrep
+++ b/cli/bin/pysemgrep
@@ -11,6 +11,7 @@
 # with semgrep.main.main
 # (e.g., in /opt/local/lib/site-packages/python3.11/semgrep/)
 
+import os
 import sys
 # This file is not part of the Python 'semgrep' package; it's a script.
 # It's an escape hatch for users to pysemgrep in case of problems.
@@ -44,5 +45,6 @@ import sys
 # https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods
 # It doesn't explain in enough detail to give a real answer, hence the guess above.
 if __name__ == "__main__":
+    os.environ["SEMGREP_NEW_CLI_UX"] = "1"  # Only enable new CLI UX when pysemgrep is invoked directly
     import semgrep.main
     sys.exit(semgrep.main.main())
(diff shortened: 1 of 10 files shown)


DrewDennison (approved):

bmahe (approved):

aryx (changes requested):
There are 172 test regressions, so let's take our time here before approving.

aryx:
To avoid the regressions, and to help for porting the code later to osemgrep, we could use a global feature flag like new_cli_UX,
that would be set to true by default, but could be changed to false when running our test infra, so this also gives
room for experimentation. Of course we don't want the code in it to be too big, because our e2e tests would actually not
be faithful to what is happening, but that would allow to experiment in a new UX in python without having to update all our tests and also having to update the OCaml code.

cli/bin/pysemgrep
```diff
@@ -44,5 +45,7 @@ import sys
 # https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods
 # It doesn't explain in enough detail to give a real answer, hence the guess above.
 if __name__ == "__main__":
+    # Only enable new CLI UX when pysemgrep is invoked directly and with interactive terminal
+    os.environ["SEMGREP_NEW_CLI_UX"] =  f"{int(os.isatty(sys.stdout.fileno()))}"
```
aryx, line 49: so this does not return true when we invoke semgrep from pytest in rum_semgrep_in_tmp?

aryx (approved):

PR #8592, opened 2023-08-31
feat: implement --baseline-commit in osemgrep

ihji:
This commit brings in the ported version of the --baseline-commit option to osemgrep, which was initially implemented in pysemgrep.

PR checklist:


cli/tests/e2e/test_baseline.py
```diff
@@ -125,6 +126,27 @@ def run_sentinel_scan(check: bool = True, base_commit: Optional[str] = None):
         raise e
 
 
+def assert_out_match(snapshot, output, snapshot_name):
+    return snapshot.assert_match(textwrap.dedent(output.stdout), snapshot_name)
+
+
+def assert_err_match(snapshot, output, snapshot_name, replace_base_commit=None):
+    if "PYTEST_USE_OSEMGREP" in os.environ:
```
aryx, line 134: ok, but it would be good to at least check that we still return an error code in osemgrep.
We should add a separate assert that make sure the exit code for pysemgrep (and osemgrep) is
an error for those baseline tests.

libs/paths/Ppath.ml
```diff
@@ -205,7 +205,7 @@ let remove_prefix root path =
   let path = Fpath.to_dir_path path in
   (* now we can call this function to remove the root prefix from path *)
   match Fpath.rem_prefix root path with
-  | None -> None
+  | None -> if Fpath.equal root path then Some (Fpath.v ".") else None
```
aryx, line 208: would be good to add this example in the doc in Ppath.mli and maybe add a test showing this behavior in Unit_ppath.ml

src/osemgrep/cli_scan/Scan_subcommand.mli
```diff
@@ -11,11 +11,20 @@ val main : string array -> Exit_code.t
 val run_conf : Scan_CLI.conf -> Exit_code.t
 val run_scan_conf : Scan_CLI.conf -> Exit_code.t
 
+type diff_config = { diff_targets : Fpath.t list; diff_depth : int option }
```
aryx, line 29: This require more comments. What is diff_depth for example?
And why we need an extra type for those kinds of targets? We already have
functions that accept a list of target, so why we must care whether the targets
come from a diff or from another way?

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -27,13 +27,22 @@ open Common
 
 module Out = Semgrep_output_v1_t
 module RP = Report
+module SS = Set.Make (String)
 
 (*****************************************************************************)
 (* To run the pro engine, including multistep rules *)
 (*****************************************************************************)
 
+type diff_config = { diff_targets : Fpath.t list; diff_depth : int option }
```
aryx, line 36: Maybe add a comment like
(* This is used to filter matches from files that are not in the commit (see the --baseline-commit option). 
  * diff_targets contain the list of all files involved in a commit; diff_depth is ???
 *)

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -200,6 +209,58 @@ let rules_and_counted_matches (res : Core_runner.result) : (Rule.t * int) list =
       | None -> acc)
     map []
 
+let remove_matches_in_baseline commit baseline head renamed =
```
aryx: Please add a short comment about what this function does, and maybe add the type for the argument
with let foo (commit : ??) (baseline : ??) (head : ??) (renamed: ??) : ?? =
OCaml has type inference but it's nice for the reader to have the types in the source code.

src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -267,14 +328,106 @@ let run_scan_files (conf : Scan_CLI.conf) (profiler : Profiler.t)
...
-        targets
+        ~file_match_results_hook conf.core_runner_conf rules errors targets
+    in
+    let exn_and_matches =
+      match conf.targeting_conf.baseline_commit with
+      | None ->
+          Profiler.record profiler ~name:"core_time"
+            (core targets filtered_rules)
+      | Some baseline_commit ->
+          (* diff scan mode *)
```
aryx, line 432: This can be extracted in a separate function as the code of run_scan_files() is already very long.
You can add a section (*************) with Differential scanning as a title and put the 2 functions you wrote in it, with
some short explanations before.

aryx (approved):
Very nice!

src/osemgrep/cli_scan/Scan_subcommand.ml
ihji: Done.

src/osemgrep/cli_scan/Scan_subcommand.ml
ihji: Done.

src/osemgrep/cli_scan/Scan_subcommand.mli
ihji: Done.

src/osemgrep/cli_scan/Scan_subcommand.ml
ihji: Done. Added the comment to .mli file.

cli/tests/e2e/test_baseline.py
ihji: Done. The exit code is already checked when the engine has completed with errors. For example, https://github.com/returntocorp/semgrep/blob/86250ee8a05de978087061ce60bc2de7937da71e/cli/tests/e2e/test_baseline.py#L712

libs/paths/Ppath.ml
ihji: Tests and examples are added.

PR #8627, opened 2023-09-05
feat(secrets): add option to have auth headers added

kopecs:
Currently just adds a key for aws sigv4, but we could add more in the future. Used by pro.

PR checklist:


kopecs:
Still requires minor update to interfaces (adding `auth` key under `request`).

libs/commons/dune
```diff
@@ -2,6 +2,7 @@
  (public_name commons)
  (wrapped false)
  (libraries
+   digestif.c
```
akuhlens: Did you ever get a reponse about why we were using the ocaml version? I we should double check that all the tests go through on the pro-engine before landing this because there are some CI tests that use an ocaml to js backend that could choke it they are using this.

akuhlens (approved):
Make sure all the tests pass for pro-engine before committing, but otherwise looks as expected.

libs/commons/dune
kopecs: Hmm, I had changed this back---may just have not pushed that commit. Not needed since we aren't using `aws` directly, but that would be another blocker even if the upstream function worked (they rely on the C version and since this isn't wrapped in dune (I think) we get a link error)

kopecs:
@akuhlens Do you know if I need to do anything to get CLA to report status?

src/core/Rule.ml
```diff
@@ -318,6 +318,17 @@ and extract_reduction = Separate | Concat [@@deriving show]
 type header = { name : string; value : string } [@@deriving show]
 type meth = [ `DELETE | `GET | `POST | `HEAD | `PUT ] [@@deriving show]
 
+(* Used to request additional auth headers are computed and added automatically,
```
aryx, line 321: I don't understand this comment.
Maybe an example would help.

kopecs: I'll add something. The gist is that you might want an `Authorization` header (or other additions/modification) with an HMAC or something over the whole HTTP frame, but that might depend on interpolated metavariable values, current UTC time, etc..; so you can't add it yourself.

PR #8628, opened 2023-09-05
js: add distclean recipe

tpetr:
- adds `distclean` makefile recipe, which clears out additional folders like `node_modules` and the libpcre downloads folder

test plan:
- local `make distclean` succeeds
- local `make test` succeeds after distclean
- checks pass

### DIFF
--- a/js/Makefile
+++ b/js/Makefile
@@ -46,3 +46,10 @@ clean:
 	cd libyaml; $(MAKE) clean
 	cd engine; $(MAKE) clean
 	cd languages; $(MAKE) clean
+
+.PHONY: distclean
+distclean:
+	cd libpcre; $(MAKE) distclean
+	cd libyaml; $(MAKE) distclean
+	cd engine; $(MAKE) distclean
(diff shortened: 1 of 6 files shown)


js/engine/tests/__snapshots__/index.test.js.snap
```diff
@@ -22,24 +22,26 @@ exports[`yaml parser parses a pattern with pattern-regex 1`] = `
...
+              "offset": 3,
+            },
+            "path": "test.yaml",
+            "start": {
+              "col": 1,
+              "line": 1,
+              "offset": 0,
+            },
           },
+          "rule_id": "test",
```
aryx: this is wrong. you're not using the latest semgrep-interfaces.

aryx (approved):

aryx:
you should not need to update the snapshot (I actually regenerated this file monday I think, because of changes I did in semgrep-interfaces).

tpetr:
> you should not need to update the snapshot

right, my branch must have been in a weird state or something -- fixing

PR #8633, opened 2023-09-05
fix: persist full scan diagnostics until change

ajbt200128:
Previously, we would rescan a file whenever it was opened. This is not performant, and also would clear diagnostics from a full scan. Now we do not scan a file when opened unless it has never been scanned before in the session. 
 

PR checklist:

### DIFF
--- a/src/osemgrep/language_server/LS.ml
+++ b/src/osemgrep/language_server/LS.ml
@@ -117,10 +117,21 @@ module MessageHandler = struct
           let content = Some first.text in
           scan_file ~content server uri;
           server
-      | CN.DidSaveTextDocument { textDocument = { uri }; _ }
-      | CN.TextDocumentDidOpen { textDocument = { uri; _ } } ->
-          Logs.app (fun m -> m "Scanning file %s" (Uri.to_string uri));
+      | CN.DidSaveTextDocument { textDocument = { uri }; _ } ->
+          Logs.app (fun m -> m "Scanning file %s on save" (Uri.to_string uri));
           scan_file server uri;
+          server
+      | CN.TextDocumentDidOpen { textDocument = { uri; _ } } ->
+          let path = uri |> Uri.to_path |> Fpath.v in
+          let prev_scan = Session.previous_scan_of_file server.session path in
+          (* We usually scan every file on startup, so let's only rescan an opened
+             file if there weren't previous results *)
+          if Option.is_some prev_scan then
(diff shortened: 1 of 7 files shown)


src/osemgrep/language_server/LS.ml
```diff
@@ -117,11 +117,22 @@ module MessageHandler = struct
...
+      | CN.DidSaveTextDocument { textDocument = { uri }; _ } ->
+          Logs.app (fun m -> m "Scanning file %s on save" (Uri.to_string uri));
           scan_file server uri;
           server
+      | CN.TextDocumentDidOpen { textDocument = { uri; _ } } ->
+          let path = uri |> Uri.to_path |> Fpath.v in
+          let prev_scan = Session.previous_scan_of_file server.session path in
+          (* We usually scan every file on startup, so let's only rescan an opened
+             file if there weren't previous results *)
+          if Option.is_some prev_scan then
```
aryx, line 129: usually more idiomatic to do a match prev_scan with Some _ -> ... | None -> ...

src/osemgrep/language_server/Session.mli
```diff
@@ -10,7 +10,7 @@ type t = {
   incoming : Lwt_io.input_channel;
   outgoing : Lwt_io.output_channel;
   workspace_folders : Fpath.t list;
-  documents : (Fpath.t, Semgrep_output_v1_t.cli_match list) Hashtbl.t;
+  cached_scans : (Fpath.t, Semgrep_output_v1_t.cli_match list) Hashtbl.t;
```
aryx, line 13: You do not store also the time at which the scan was done on this file? So if a file is modified, how do you know
the cache is invalid and we should rerun semgrep for this file?

aryx (approved):

src/osemgrep/language_server/LS.ml
ajbt200128: Fair

src/osemgrep/language_server/Session.mli
ajbt200128: We scan everytime the file is changed, or when its saved, and also initially when a project is opened. This will stay up to date unless a user performs a git checkout (or modifies a file with a different editor, but that's not something easy to handle). This was already an issue that creates phantom diagnostics. What we should do is setup lsp "file watchers" to detect when git HEAD changes, and then rescan, but that's something that should be in a separate PR

PR #8642, opened 2023-09-06
js: only publish JS artifacts to S3 for develop branch, release branch, or if PR has a special label

tpetr:
We're currently uploading semgrep.js artifacts on push for _every_ PR, which is a bit excessive. This PR updates the `upload-javascript-artifacts` workflow to only run:
- on push to `develop` (so that we can test the latest-and-greatest)
- on push to a release branch (starts with `release-`)
- if the PR is labelled with `publish-js` (because there are still occasionally times where it's helpful for troubleshooting)

test plan:
- checks pass

### DIFF
--- a/.github/workflows/build-test-javascript.yaml
+++ b/.github/workflows/build-test-javascript.yaml
@@ -97,7 +97,8 @@ jobs:
 
   upload-javascript-artifacts:
     needs: [build-test-js-artifacts]
-    if: github.ref == 'refs/heads/develop' || github.event.pull_request.head.repo.full_name == github.repository
+    # in english: (this pull request is not from a fork) AND (branch is "develop" OR branch starts with "release-" OR pull request has a "publish-js" label applied to it)
+    if: github.event.pull_request.head.repo.full_name == github.repository && (github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/heads/release-') || contains(github.event.pull_request.labels.*.name, 'publish-js'))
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-javascript.yaml
```diff
@@ -97,7 +97,8 @@ jobs:
 
   upload-javascript-artifacts:
     needs: [build-test-js-artifacts]
-    if: github.ref == 'refs/heads/develop' || github.event.pull_request.head.repo.full_name == github.repository
+    # in english: (this pull request is not from a fork) AND (branch is "develop" OR branch starts with "release-" OR pull request has a "publish-js" label applied to it)
```
aryx, line 100: love the english translation

aryx: Also what is this coding language? we should really go back to shell script and a Makefile.

aryx (approved):

.github/workflows/build-test-javascript.yaml
tpetr: It's GitHub Actions' crazy expression language. It's not bad for simple stuff but it can get hairy quickly. See:
- https://docs.github.com/en/actions/learn-github-actions/expressions
- https://docs.github.com/en/actions/learn-github-actions/contexts
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

PR #8644, opened 2023-09-06
fix: disable capture group unification

brandonspark:
## What:
This PR prevents numeric capture group metavariables from invisibly unifying.

## Why:
Numeric capture group metavariables (like `$1`, `$2`, etc) are introduced implicitly by regular expressions using capture groups, which are essentially parentheses around certain regular expressions.

This is done "implicitly" because these metavariables are never named, and simply happen whenever parentheses do. This can lead to unpredictable results, however, because a rule-writer might accidentally introduce these metavariables, and then fail to get matches because the first such metavariable (`$1`) fails to match.

## How:
Just prevented it when ranges are being combined. We don't need to touch `Matching_generic` because `Generic_vs_generic` guards all `envf` calls with `is_metavar_name`.

## Test plan:
`make test`

Closes PA-3050

PR checklist:

### DIFF
--- a/src/engine/Range_with_metavars.ml
+++ b/src/engine/Range_with_metavars.ml
@@ -65,6 +65,7 @@ let included_in config rv1 rv2 =
      |> List.for_all (fun (mvar, mval1) ->
             match List.assoc_opt mvar rv2.mvars with
             | None -> true
+            | _ when Metavariable.is_metavar_for_capture_group mvar -> true
             | Some mval2 ->
                 Matching_generic.equal_ast_bound_code config mval1 mval2)
 
--- a/src/engine/Range_with_metavars.ml
+++ b/src/engine/Range_with_metavars.ml
@@ -65,6 +65,14 @@ let included_in config rv1 rv2 =
      |> List.for_all (fun (mvar, mval1) ->
             match List.assoc_opt mvar rv2.mvars with
             | None -> true
+            (* Capture group metavariables (of the form $1, $2, etc) may
+               be introduced implicitly via regular expressions that have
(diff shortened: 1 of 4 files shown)


src/engine/Range_with_metavars.ml
```diff
@@ -65,6 +65,15 @@ let included_in config rv1 rv2 =
      |> List.for_all (fun (mvar, mval1) ->
             match List.assoc_opt mvar rv2.mvars with
             | None -> true
+            (* Capture group metavariables (of the form $1, $2, etc) may
```
aryx: maybe rename to Numerric capture group metavariables

src/engine/Range_with_metavars.ml
```diff
@@ -65,6 +65,15 @@ let included_in config rv1 rv2 =
...
+               be introduced implicitly via regular expressions that have
+               capture groups in them.
+               The unification of these metavariables can be dangerous, as
+               it will prevent matches, when users may not even know these
+               metavariables exist.
+               To be safe, let's assume they always unify.
```
aryx: You can maybe add a note like
* note that this does not concern named-capture-group-metavars introduced via the <?xxx> regexp syntax. 
* Those are still unified (as they are explicitely named)

aryx (approved):
Thx!

aryx:
In theory we could add an option: in Rule_options.atd like numeric_capture_unify_mvars which defaults to false
(like we have a taint_unify_mvars), but I guess we can introduce it if people start to complain about this new behavior
(or maybe we can tell them they shoukd used named-capture-group instead).

PR #8646, opened 2023-09-06
cleanup: add IL node for fetch expressions used as statements

amchiclet:
For expression-oriented languages, such as Rust, everything is an expression. Moreover, the last expression of a function is implicitly returned.

Currently, the Rust function

```
  fn f(x:i32) {
    x;
    x;
    x;
  }
```

disregards the first two xs, and gives us a CFG with only one return node for the last x. The return node for the last x is also a hack that explicitly inserts the node to the last expression statement in a function.

This change treats expressions as statements in expression-oriented languages. It generates an IL assigning to a fresh var for these expression statements.

The tradeoff is creating fresh vars that aren't used, but this gives us a more complete CFG. This CFG will help when implementing the next version of implicit return support.

Closes [PA-3056](https://linear.app/semgrep/issue/PA-3056).

Test plan:

Ran `make core-test` and it passes all tests.

Also ran `sc -cfg_il` on the following rust code

```
  fn f(x:i32) {
    x;
    x;
    x;
  }
```

and the CFG shows 2 assignments and one return node. Before this PR, it only showed the single return node.


amchiclet:
PR checklist:

src/analyzing/AST_to_IL.ml
```diff
@@ -178,7 +187,27 @@ let add_instr env instr = Common.push (mk_s (Instr instr)) env.stmts
  * itself is already a variable! *)
 let mk_aux_var ?str env tok exp =
   match exp.e with
-  | Fetch ({ base = Var var; rev_offset = []; _ } as lval) -> (var, lval)
+  | Fetch ({ base = Var var; rev_offset = []; _ } as lval) ->
```
aryx: This seems like a dangerous place to put this code. mk_aux_var is used in many context.
Wouldn't it be possible instead to add some special code in stmt_aux() for the ExprStmt case
and detect when no statement was generated and create a fake assign there? 
What do you think @IagoAbal ?

aryx (changes requested):

src/analyzing/AST_to_IL.ml
```diff
@@ -178,7 +187,27 @@ let add_instr env instr = Common.push (mk_s (Instr instr)) env.stmts
...
+       * expressions as statements. This is needed in case the
+       * expression is returning from a function.
+       *
+       * If we don't generate this assignment, functions like,
+       *   function f(s) {
+       *     s
+       *   }
+       * will end up having an empty CFG, losing information about s.
+       *)
+      (if env.lang |> treats_expr_as_stmt then
```
aryx: also you can just write treats_expr_as_stmt env.lang
we reserve |> for complex pipeline (e.g., big_expr |> some_func |> some_other_func |> ...) or when the argument is really complex and we'd rather have the reader
understand what is the object first.

src/analyzing/AST_to_IL.ml
aryx: Then we would no need treat_expr_as_stmt guard; each time we have an ExprStmt that does not generate
any instruction, that probably means something is wrong and we should still generate a node
(in theory we could issue a warning because in your Rust example the first two x are wrong; people should not write such code).

IagoAbal:
> Addresses: https://linear.app/semgrep/issue/PA-3056

You can just write "Closes PA-3056" and the bot will pick that up!

src/analyzing/AST_to_IL.ml
```diff
@@ -112,6 +112,15 @@ let fixme_stmt kind gany =
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
+
+let treats_expr_as_stmt (lang : Lang.t) =
+  match lang with
+  | Rust
+  | Julia
+  | Ruby ->
```
IagoAbal: Maybe add OCaml here already?

src/analyzing/AST_to_IL.ml
IagoAbal: +1 to what Yoann said.

src/analyzing/AST_to_IL.ml
IagoAbal: Also agree with Yoann here, we could just unconditionally add this assignment in `stmt_aux` for the `ExprStmt` case. We could still make use of `treat_expr_as_stmt` to determine whether the value of `~void` (right now we always set it to `true` in an `ExprStmt`).

src/analyzing/AST_to_IL.ml
```diff
@@ -112,6 +112,15 @@ let fixme_stmt kind gany =
 (*****************************************************************************)
 (* Helpers *)
 (*****************************************************************************)
+
```
IagoAbal, line 115: You need a changelog entry, `vim changelog.d/pa-3056.fixed` and write a user-oriented description of the change. The example that you used in the PR comment it's perfect IMO.

src/analyzing/AST_to_IL.ml
amchiclet: Thanks for the tip! I need to re-wire my head a bit.

What I had in mind was that it reads to me more naturally as: "The language treats expressions as statements".

src/analyzing/AST_to_IL.ml
amchiclet: Sounds good, thanks! I'll come back with an update.

src/analyzing/AST_to_IL.ml
amchiclet: Got it. This one was a bit fuzzy for me because it changes the CFG, but I wasn't sure if the change affects the behavior that the user sees or not. I suppose the user can always use the `cfg_il` flag and this is considered user-facing.

By the way,  I like it that you specify `vim`. 😃

src/analyzing/AST_to_IL.ml
amchiclet: Some updates:

At first, I thought it wouldn't hurt to unconditionally create a fake assignments for the `ExprStmt` case in `stmt_aux` and not call `mk_aux_var`, but there are some issues with this approach.

The code

```
x = 0
```

results in the IL
```
x = 0
tmp = x
```

where both `x`s point to the same generic ast.

During constant propagation, because the second `x` is [on the RHS, its svalue is updated](https://github.com/returntocorp/semgrep/blob/2691ed6fe4feb6499438358532b6929c5b37f306/src/analyzing/Dataflow_svalue.ml#L654-L655) to 0.

This means 0 matches both with x and 0 in `x=0`, but we only want to match with the rhs for constant propagation so this turned out to be incorrect.

Checking for empty results like @aryx suggested seems to be a good approach. I implemented it and it works. I'll send the cleaned up code next week. I'm reluctant to add warnings because while the first two `x`s in
```
  fn f(x:i32) {
    x;
    x;
    x;
  }
```
doesn't make sense, the final `x` is a legitimate one.

I couldn't find good use of setting `~void` here. The `~void` parameter doesn't seem to have any affect on the AST_Generic.N case that I'm interested here. I may be missing what you're trying to suggest.

IagoAbal: Whoops, perhaps we should do this `update_svalue` on the Generic AST instead. Anyways, just checking if there is no instructions generated as Yoann suggested is probably better.

Re `~void`, see `add_call`, it is used to determine whether we store the result of a function call in an auxiliary variable. It's part of a hack that we should refine in the future. If we encounter `x.foo();`, we assume that `x.foo()` either returns `void` or its value is non-important here, and then const-prop may look at that to assume that `x.foo()` may have side-effects (so if `x` was a constant it may stop being it). I was thinking that maybe we want to tweak this for languages like Rust, but on a second thought, better not to touch this yet. So never mind, sorry for confusing you.

amchiclet: Updated the code.

src/analyzing/AST_to_IL.ml
amchiclet: Added changelog.

src/analyzing/AST_to_IL.ml
amchiclet: This code was removed entirely, so I'll just mark this resolved.

amchiclet:
Benchmark slowness is happening in other PRs too. I believe this PR isn't related.

src/analyzing/AST_to_IL.ml
amchiclet: `treats_expr_as_stmt` was removed in the last update. Marking this as resolved.

changelog.d/pa-3056.fixed
```diff
@@ -0,0 +1,18 @@
...
+  x;
+  x;
+  x;
+}
+
+
+in expression-oriented languages, such as Rust, will treat the last x
+as if it was "return x".
+
+Currently we use a hack to support this. To support it in a more
```
aryx: so you need to understand that the changelog is mostly directed to users of Semgrep; At each release they can have a look at the changelog and know what's new. They don't want to learn about how things were before, that something was a hack, etc. so keep it to the point. In some cases because it's really an internal change and the actual visible change for the user will be in another PR you don't have to add a changelog.d/ entry.

For your PR, we do not support yet this implicit return thing (it will be in another PR), so maybbe just say:
```
Fixed the CFG for expression-oriented languages such as Rust so that on the example below, we generate 3 nodes instead of just one.
```
You can have a longer changelog for the actual PR implementing the implicit return.
```

**Review by aryx — APPROVED** [review]

File: changelog.d/pa-3056.fixed [hunk 6]
```diff
@@ -0,0 +1,18 @@
+Programs like
+
+fn f(x:i32) {
+  x;
```
IagoAbal: I would just write `x = g()` or something.

changelog.d/pa-3056.fixed
```diff
@@ -0,0 +1,18 @@
...
+  x;
+  x;
+}
+
+
+in expression-oriented languages, such as Rust, will treat the last x
```
IagoAbal: I would say something like:

> Semgrep now recognizes "implicit returns" in expression-oriented languages such as Ruby or Rust. For example: ... HERE YOUR EXAMPLE ...The last statement `x` will be considered equivalent to (and in fact match) `return x`.

IagoAbal: Sorry, I got confused, this PR actually doesn't fix that yet.

changelog.d/pa-3056.fixed
IagoAbal: +1 just write what the user will see and the details about how we did it belongs to the code and/or PR description.

IagoAbal: BTW I forgot or I'm missing why it is useful to keep the first two `x;` in the IL?

src/analyzing/AST_to_IL.ml
```diff
@@ -1193,12 +1194,53 @@ and mk_switch_break_label env tok =
...
       (* optimize? pass context to expr when no need for return value? *)
       let ss, e = expr_with_pre_stmts ~void:true env eorig in
       mk_aux_var env tok e |> ignore;
       let ss' = pop_stmts env in
-      ss @ ss'
+      match ss @ ss' with
+      | [] ->
+          (* This case may happen when we have a function like
+           *
+           *   function some_function(some_var) {
```
IagoAbal, line 1208: Wasn't this handled by the "implicit_return_hack"?

IagoAbal: OK it's explained below.

changelog.d/pa-3056.fixed
IagoAbal: TBH I would personally not add any changelog for this. Right now it won't have any user-visible effect, it's just an internal thing we're doing in preparation for supporting implicit returns.

amchiclet: @IagoAbal I actually agree with not having a changelog. 😄 

In the first round [you said](https://github.com/returntocorp/semgrep/pull/8646#discussion_r1318355249) it needs a changelog.

I assume now that it's clearer that the PR is not user-facing, it's also clearer it doesn't need a changelog, right?

IagoAbal: Sorry about that then. But yeah if this will not be noticed by any user then it doesn't need a changelog entry.

amchiclet: Alright, I'll remove this file. Thanks!

changelog.d/pa-3056.fixed
amchiclet: Removed the file, so just resolving this.

changelog.d/pa-3056.fixed
amchiclet: Removed!

amchiclet:
Resolved all comments, so going to rebase from develop and force push now.

amchiclet:
Not sure why lint is [failing](https://github.com/returntocorp/semgrep/actions/runs/6175305712/job/16761859935?pr=8646#step:5:209). I'll see what can be done here.

amchiclet:
My guess is that updated changelog.d I pulled from the develop branch doesn't have the proper suffix and is not passing lint.

There's a file [`gh-8679`](https://github.com/returntocorp/semgrep/blob/amarin/cfg-handle-expr-stmt-in-expression-oriented-languages/changelog.d/gh-8679) with no `.added` or `.fixed` etc.

I'll wait to see if newer PRs that include this file are affected the same way or not.

amchiclet:
Lint no longer fails after rebasing today. This PR is ready to be merged unless there are more changes required. Slow benchmarks are happening on other PRs, and I believe this PR is unrelated.

IagoAbal (approved):

PR #8653, opened 2023-09-07
Add Minimal Metrics Support for osemgrep

zzeleznick:
## Description
This PR adds minimal (functional) support for metrics within osemgrep. We should expect to see a 200 response from our metrics endpoint upon the conclusion of a subcommand (with metrics enabled).

## Demo

https://github.com/returntocorp/semgrep/assets/5779832/762e0436-7da5-40a0-a4f3-4632976280c5

### Adds
- required fields for our minimal metrics payload
- `send_metrics` implementation to send metrics payload
- `scan` and `install-ci` experimental subcommands will log metrics when enabled

### Modifies
- sets `chunked` to `false` by default to match (presumed) expectations of the metrics payload endpoint
- enables metrics URL custom for testing purposes (e.g. terminal 1: `while true; do { cat response.txt; } | nc -l 8092; done` and terminal 2: `SEMGREP_METRICS_URL="http://127.0.0.1:8092" ./bin/osemgrep install-ci --experimental --dry-run --debug .`)

## Test Plan
- [x] We see a 200 response from our metrics endpoint upon the conclusion of a subcommand (with metrics enabled).
- [x] CI tests pass
- [x] We see our metrics in snowflake

PR checklist:


aryx:
Please expand a bit the test plan to describe the full testing procedure, that is the precise comamnd you run, the precise URL you went too to observe the metrics data.
Could be useful also in Metrics_.ml to fully document the life of a metrics query, from osemgrep code to a payload, to the actual serialization in osemgrep, to an endpoint, to the code handling this endpoint (where is this code), to a lambda (defined where), to a backend, .... to metabase where it can be obsevered at this URL: X.

Basically everything you learned in the last few days about our metrics pipeline.

src/osemgrep/cli/CLI.ml
```diff
@@ -251,17 +309,15 @@ let main argv : Exit_code.t =
...
-      state.app_session.user_agent.tags.add(f"command/{subcommand}")
-      state.metrics.add_feature("subcommand", subcommand)
-      maybe_set_git_safe_directories()
-  *)
   (*TOADAPT? adapt more of Common.boilerplate? *)
   let exit_code = safe_run ~debug (fun () -> dispatch_subcommand argv) in
   Metrics_.add_exit_code exit_code;
+  send_metrics ();
+
   (* TODO(dinosaure): currently, even if we record the [exit_code], we will
```
aryx: maybe you can remove this comment now?

src/osemgrep/cli_install/Install_CLI.ml
```diff
@@ -72,6 +73,15 @@ let o_update : bool Term.t =
   let doc = "Update any existing workflow and secrets (absent=false)." in
   H.negatable_flag [ "update" ] ~neg_options:[ "no-update" ] ~default:false ~doc
 
+(* NOTE: Should only be required when previous attempts only partially succeeded *)
+let o_dry_run : bool Term.t =
+  let doc =
+    "Simulate outputs without actually running any system commands \
+     (absent=false)."
+  in
+  H.negatable_flag [ "dry-run" ] ~neg_options:[ "no-dry-run" ] ~default:false
```
aryx: I don't think you need the negatable option here.
just "dry-run" should be enough, nobody is gonna ever used --no-dry-run

src/osemgrep/cli_install/Install_subcommand.ml
```diff
@@ -397,7 +404,8 @@ let add_semgrep_workflow ~repo ~token ~update : unit =
 (*****************************************************************************)
 
 let run (conf : Install_CLI.conf) : Exit_code.t =
-  CLI_common.setup_logging ~force_color:true ~level:conf.commons.logging_level;
+  CLI_common.setup_logging ~force_color:true ~level:conf.common.logging_level;
+  Metrics_.configure Metrics_.Auto;
```
aryx, line 413: maybe we should accept the --metrics option here?
or maybe just add a comment saying it does not make sense to let the user do --metrics=off
for install-ci so we set it to Auto by default.

src/osemgrep/configuring/Semgrep_envvars.ml
```diff
@@ -54,15 +54,19 @@ let in_env var = env_opt var <> None
 type t = {
   semgrep_url : Uri.t;
   fail_open_url : Uri.t;
+  metrics_url : Uri.t;
   app_token : string option;
+  integration_name : string option;
```
aryx, line 59: maybe a comment explaining what is this integration_name?

dune-project
```diff
@@ -443,7 +443,7 @@ For more information see https://semgrep.dev
     ; web stuff
     uri
     cohttp-lwt-unix
-    tls-lwt ; needed for TLS support in the cohttp HTTP client
+    (lwt_ssl (>= 1.2.0)) ; needed for HTTPS support in the cohttp HTTP client
```
aryx: and what about TLS? What about the code that needed TLS in cohttp?

libs/networking/http_helpers.mli
```diff
@@ -12,6 +12,7 @@ val get_async :
 val post_async :
   body:string ->
   ?headers:(string * string) list ->
+  ?chunked:bool ->
```
aryx, line 15: maybe a small comment about chunked is?
(many of the PA programmers, including me, are not super familiar with the http world,
so I can understand the body and headers of a rerquest, but chunked?

src/osemgrep/cli/CLI.ml
```diff
@@ -31,12 +31,74 @@ open Common
...
+     1. GITHUB_ACTIONS specific output requirements
+     2. Any NO_COLOR / SEMGREP_FORCE_NO_COLOR behavior
+*)
+(* let init_for_cli () : unit =
+   ()
+*)
+
+let metrics_init () : unit =
+  let () =
+    match Env.v.user_agent_append with
```
aryx: You can use instead `Env.v.user_agent_append |> Option.iter (fun str -> ...);`

src/osemgrep/cli/CLI.ml
```diff
@@ -31,12 +31,74 @@ open Common
...
+    | Some str -> Metrics_.add_user_agent_tag ~str
+    | _ -> ()
+  in
+  let settings = Semgrep_settings.load () in
+  let api_token = settings.Semgrep_settings.api_token in
+  let anonymous_user_id = settings.Semgrep_settings.anonymous_user_id in
+  Metrics_.init ~anonymous_user_id ~ci:Env.v.is_ci;
```
aryx: seems weird to have a call to Metrics_.add_user_agent_tag before a call to Metrics_.init

src/osemgrep/cli/CLI.ml
```diff
@@ -31,12 +31,74 @@ open Common
...
+  Metrics_.add_token (Some api_token);
+  ()
+
+let log_cli_feature flag : unit =
```
aryx, line 81: please add the type of the parameter when you can (OCaml has type inference, but it's usually nice for the reader to add type annotations for the parameters (no need for the locals though).

src/osemgrep/cli/CLI.ml
```diff
@@ -31,12 +31,74 @@ open Common
...
+  Metrics_.add_feature "cli-flag"
+    (flag
+    |> Base.String.chop_prefix_if_exists ~prefix:"-"
+    |> Base.String.chop_prefix_if_exists ~prefix:"-")
+
+(* NOTE: We could implement send_metrics in Metrics_.ml,
```
aryx, line 87: nice comment

src/osemgrep/configuring/Semgrep_envvars.ml
```diff
@@ -54,15 +54,19 @@ let in_env var = env_opt var <> None
...
   version_check_url : Uri.t;
   version_check_timeout : int;
   version_check_cache_path : Fpath.t;
   git_command_timeout : int;
   src_directory : Fpath.t;
+  user_agent_append : string option;
```
aryx, line 65: maybe a comment explaining in which context this environment variable is useful.
Who is using it?

src/osemgrep/core/Metrics_.ml
```diff
@@ -24,6 +26,8 @@ module Out = Semgrep_output_v1_t
 (*****************************************************************************)
 
 let _metrics_endpoint = "https://metrics.semgrep.dev"
```
aryx, line 83: In OCaml we prefix a variable with an underscore when it's actually not used. 
It's because by default OCaml will complain if a variable is unused, and so you add a _ prefix when it's unused on purpose. I think now you're actually using metrics_endpoint and so on so you can remove the _ prefix.

src/osemgrep/core/Metrics_.ml
```diff
@@ -123,6 +137,49 @@ open Semgrep_metrics_t
...
 let is_using_registry () = g.is_using_registry
+let is_enabled () = g.config <> Off
+
+let set_is_using_registry ~is_using_registry =
+  g.is_using_registry <- is_using_registry
+
+let set_anonymous_user_id ~anonymous_user_id =
+  g.payload.anonymous_user_id <- anonymous_user_id
+
+let set_started_at ~started_at = g.payload.started_at <- started_at
```
aryx, line 203: This is a bit annoying to have all those wrapper functions for each field ...
we should just have a payload() function that the caller just assign the field directly.

aryx (changes requested):
Overall looks good.

src/osemgrep/core/Metrics_.ml
zzeleznick: I'm open to other implementations. 

In python we might do something like
```
>>> payload = {}
>>> payload.update({"x": 2, "y": 3})
>>> payload
{'x': 2, 'y': 3}
```

If you can point me to code reference or sample you have in mind I can take another pass.

src/osemgrep/core/Metrics_.ml
zzeleznick: Right now I'm using `Env.v.metrics_url` (so I could test the metrics payloads locally by passing in a custom endpoint) and copied the url over as the default (and did not use the variable).

I kept this var here as a reference but it's currently unused.

src/osemgrep/configuring/Semgrep_envvars.ml
zzeleznick: Sure. It's currently used in our Dockerfile image.

src/osemgrep/configuring/Semgrep_envvars.ml
zzeleznick: I'm actually not very familiar with this one (think it's some labeling that may not be widely used). tried to collect stray env vars and put them in their new home here.

src/osemgrep/cli/CLI.ml
zzeleznick: yeah – I'll add a comment.

src/osemgrep/core/Metrics_.ml
aryx: no we don't want that because we want proper typechecking (like using an undefined field).
But you can keep it like that; I'll refactor things in a separate PR.

src/osemgrep/configuring/Semgrep_envvars.ml
zzeleznick: added docstrings in the `.mli`

aryx (approved):

PR #8655, opened 2023-09-08
o: Login abstraction + testing

ajbt200128:
This PR refactors the login osemgrep subcommand so the logic is more easily testable, and so the language server can utilize it. It also includes some tests for it!

This PR also changes the environment variables into a ref so it can be modified at runtime for testing purposes. I'm open to alternatives to this, but for ease of testing I do think we need some way to modify this at runtime. The alternative would be changing envvars to a function, but I worry that would risk the determinism of osemgrep.

PR checklist:


src/osemgrep/networking/Semgrep_login.mli
```diff
@@ -0,0 +1,36 @@
...
+val fetch_token :
+  ?min_wait:int ->
+  ?next_wait:int ->
+  ?max_retries:int ->
+  ?wait_hook:(unit -> unit) ->
+  login_session ->
+  (string * string, string) result
+(** [fetch_token ?min_wait ?next_wait ?max_retries wait_hook login_session] will
+  * fetch the token using the request token and url the login session. It will retry up to [max_retries]
+  * times, waiting [min_wait] seconds between each retry, and increasing the
```
zzeleznick: errata: measured in `ms` not `seconds`

src/osemgrep/networking/Semgrep_login.mli
```diff
@@ -0,0 +1,36 @@
...
+  *)
+
+val save_token : string -> (unit, string) result
+(** [save_token token] will save the token to the user's home directory.
+  * If it fails, it will return an error message.
+  *)
+
+val is_logged_in : unit -> bool
+(** [is_logged_in ()] will check if the user is logged in by checking if a
+  * valid token in the settings file exists.
```
zzeleznick: aside: somewhat misleading docstring as the implementation just checks if the token exists

zzeleznick: for clarity to my comment above, we only save valid tokens (using `get_deployment_from_token`) but a token could be a revoked. 

mention this in part bc I've been looking at how to implement some messaging changes based on which features are enabled for a deployment

zzeleznick (commented):
I'll let Pad have the official honor of approving this PR, but this design seems reasonable to me based on the implementation tradeoffs. I left a few minor comments on the docstrings. 

Moving forward, if we'd like to start having dynamic configuration, e.g. is supply chain available for the current deployment, I wonder if this mutable settings singleton would be the appropriate container to pass this value around.

src/osemgrep/configuring/Semgrep_envvars.ml
```diff
@@ -71,7 +71,7 @@ type t = {
 }
 
 (* less: make it Lazy? so at least not run in ocaml init time before main() *)
-let v : t =
+let v : t ref =
```
aryx, line 74: An alternative would be to keep it like this but use mutable fields in the structure, which
allows to avoid using ! when reading fields. But I guess the ref is fine and maybe better.

src/osemgrep/configuring/Semgrep_settings.ml
```diff
@@ -21,7 +21,7 @@ let default_settings =
     anonymous_user_id = Uuidm.v `V4;
   }
 
-let settings = Semgrep_envvars.v.user_settings_file
+let settings () = !Semgrep_envvars.v.user_settings_file
```
aryx: maybe simpler to remove this intermediate function and inline its use.

aryx: or just define it locally inside the load function, as it seems used only there, so you can do
let settings = !Semgrep_envvars.v.user_setting_file in there.

src/osemgrep/networking/Semgrep_login.mli
aryx: maybe put back the comment zz wrote next to the min_xxx max_xx in Login_subcommand.ml,
and add the suffix in_ms so the dimension of the type is clearer.

src/osemgrep/networking/Semgrep_login.ml
```diff
@@ -0,0 +1,115 @@
...
+      add_query_params'
+        (with_path !Semgrep_envvars.v.semgrep_url "login")
+        [
+          ("cli-token", Uuidm.to_string session_id);
+          ("docker", if !Semgrep_envvars.v.in_docker then "True" else "False");
+          ("gha", if !Semgrep_envvars.v.in_gh_action then "True" else "False");
+        ]) )
+
+let save_token token =
+  let settings = Semgrep_settings.load () in
```
aryx, line 41: I'm not completely convinced we want to load each time the settings; maybe it would be better to pass
it as a parameter. But let's go with this, maybe it's simpler.

src/osemgrep/networking/Unit_Login.ml
```diff
@@ -0,0 +1,151 @@
...
+let ok_token = "ok_token"
+let bad_token = "bad_token"
+
+let login_session =
+  ( Uuidm.of_string "00000000-0000-0000-0000-000000000000" |> Option.get,
+    snd (Semgrep_login.make_login_url ()) )
+
+let fake_settings =
+  "has_shown_metrics_notification: true\n\
+   anonymous_user_id: 5a06eb03-071d-4c89-b297-993c4912fa8f"
```
aryx: what is this magic string? you picked that randomly? or it's your anonymous_user_id you found in your own ~/.semgrep/settings.yml?

src/osemgrep/networking/Unit_Login.ml
```diff
@@ -0,0 +1,151 @@
...
+        let status, body =
+          match Testing_client.get_header req "Authorization" with
+          | Some "Bearer ok_token" -> (200, "./tests/login/ok_response.json")
+          | Some "Bearer bad_token" -> (401, "./tests/login/bad_response.json")
+          | _ -> failwith "Unexpected token"
+        in
+        Lwt.return Testing_client.(basic_response ~status body)
+    | "/api/agent/tokens/requests" ->
+        let%lwt () =
+          Testing_client.check_body body "./tests/login/fetch_body.json"
```
aryx: not a huge fan of this Testing_client name. Maybe Http_mock ?

src/osemgrep/networking/Unit_Login.ml
```diff
@@ -0,0 +1,151 @@
...
+
+let login_session =
+  ( Uuidm.of_string "00000000-0000-0000-0000-000000000000" |> Option.get,
+    snd (Semgrep_login.make_login_url ()) )
+
+let fake_settings =
+  "has_shown_metrics_notification: true\n\
+   anonymous_user_id: 5a06eb03-071d-4c89-b297-993c4912fa8f"
+
+let with_login_client =
```
aryx: maybe with_fake_login_responses?
I don't find with_login_client very readable when it's used.

src/osemgrep/networking/Unit_Login.ml
```diff
@@ -0,0 +1,151 @@
...
+  let make_fn req body =
+    ignore body;
+    ignore req;
+    Lwt.return Testing_client.(basic_response ~status:404 "/dev/null")
+  in
+  Testing_client.with_testing_client make_fn
+
+let with_tmp_settings_file f () =
+  Common2.with_tmp_file ~str:fake_settings ~ext:"yml" (fun tmp_settings_file ->
+      let old_settings = !Semgrep_envvars.v in
```
aryx, line 69: maybe don't call it settings? It's confusing as it's actually the content of the whole envvars.
Maybe old_env?

src/osemgrep/networking/Unit_Login.ml
```diff
@@ -0,0 +1,151 @@
...
+   anonymous_user_id: 5a06eb03-071d-4c89-b297-993c4912fa8f"
+
+let with_login_client =
+  let make_fn req body =
+    let uri = Cohttp.Request.uri req in
+    match Uri.path uri with
+    | "/api/agent/deployments/current" ->
+        let status, body =
+          match Testing_client.get_header req "Authorization" with
+          | Some "Bearer ok_token" -> (200, "./tests/login/ok_response.json")
```
aryx, line 43: Maybe it would be better to return directly the content of the file instead of the file?
So we can avoid creating those tests/ files.
Or maybe change Testing_client body to be a variant such as File of Fpath.t | Content of string?

aryx (approved):
It is my honor to approve this PR.

src/osemgrep/networking/Semgrep_login.mli
ajbt200128: No that's fair, I misremembered how this function and save_token worked

src/osemgrep/configuring/Semgrep_envvars.ml
ajbt200128: My thought process was that by using a ref here, we make it harder to actually mutate these values, so people are discouraged from doing it unless they really need to. Also a ref makes it a lot easier to save these settings and then restore them in testing.

src/osemgrep/networking/Semgrep_login.ml
ajbt200128: it cleaned up the code a lot, and this felt cheap enough. I'm ok changing it back if it presents any issues

src/osemgrep/networking/Unit_Login.ml
ajbt200128: Oops this is my anonymous ID, I'll change it to something that's obviously not real. This is just a fake user settings file for testing

src/osemgrep/networking/Unit_Login.ml
ajbt200128: Thought the same, I'll change it to Http_mock_client

src/osemgrep/networking/Unit_Login.ml
ajbt200128: sure

src/osemgrep/networking/Unit_Login.ml
ajbt200128: how about with_mock_envvars?

src/osemgrep/networking/Unit_Login.ml
ajbt200128: Yea @kopecs and I discussed this. I believe he's going to change that soon, I didn't want to step on his toes

kopecs: Yeah see #8667

PR #8656, opened 2023-09-08
fix(cfg): case statement in ruby does not fallthrough

amchiclet:
There's already support for switch statements that don't fallthrough, but it's only enabled in Go.

This PR enables this for Ruby and includes a tainting test.

Test plan: added a new test and tested with `make core-test`.

Closes [PA-3055](https://linear.app/semgrep/issue/PA-3055)

### DIFF
--- a/changelog.d/pa-3055.fixed
+++ b/changelog.d/pa-3055.fixed
@@ -1,0 +1,1 @@
+The CFG now supports case statements in Ruby, which does not fall through.
--- a/src/analyzing/AST_to_IL.ml
+++ b/src/analyzing/AST_to_IL.ml
@@ -1167,7 +1167,9 @@ and type_ env (ty : G.type_) : type_ =
 
 (* TODO: What other languages have no fallthrough? *)
 and no_switch_fallthrough : Lang.t -> bool = function
-  | Go -> true
+  | Go
+  | Ruby ->
+      true
   | _ -> false
 
 and mk_break_continue_labels env tok =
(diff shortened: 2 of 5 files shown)


amchiclet:
PR checklist:

amchiclet:
I believe the slow benchmarks are unrelated.

src/analyzing/AST_to_IL.ml
```diff
@@ -1167,7 +1167,9 @@ and type_ env (ty : G.type_) : type_ =
 
 (* TODO: What other languages have no fallthrough? *)
```
aryx, line 1168: maybe study a bit more the other languages and remove this comment? or just remove the comment.

aryx (approved):

changelog.d/pa-3055.fixed
```diff
@@ -0,0 +1 @@
+The CFG now supports case statements in Ruby, which does not fall through.
```
IagoAbal, line 1: This entry is not very good TBH, I do not think any user will understand this. For users it's confusing already because you talk about CFGs. But I would be confused myself if I didn't look at the rest of the PR. Why not something like "Semgrep now handles Ruby case statements correctly and does not allow control to fall-through", and then you put a code snippet e.g. with `source` and `sink` that shows what now works correctly.

amchiclet: Sounds good. Can I still do this after it's merged, but before the release?

IagoAbal: For sure, the actual changelog is only generated at release time. If you want to fix this I really appreciate it, but my intention was that you just considered this for the next time. 🙏

PR #8658, opened 2023-09-08
Tweak `semgrep ci` timeouts and logging

jbergler:
As we work to make `semgrep ci` more reliable we're finding that a few things show up in our debugging:
- We're missing debug info for the start scan step, so lets add the scan id to the output and add the meta payload to the debug logs.
- Under high load, jobs can occasionally take 15 minutes to process, for now lets raise the limit to 30min while we work on optimisations there. 
- Sometimes the client times out and never gets the response from the backend, which is awkward because the timeouts are the same at both ends. Ideally the client's timeouts are slightly higher than the backends so that the errors propagate appropriately.

Tested locally.

PR checklist:

### DIFF
--- a/changelog.d/gh-8656.changed
+++ b/changelog.d/gh-8656.changed
@@ -1,0 +1,1 @@
+Further improvements to timeouts and logging for `semgrep ci`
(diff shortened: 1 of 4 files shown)


zyannes (approved):

brendongo (approved):

aryx:
why this was merged? The CI tests were failing.

PR #8665, opened 2023-09-08
Tweak semgrep ci timeouts and logging

jbergler:
As we work to make `semgrep ci` more reliable we're finding that a few things show up in our debugging:
- We're missing debug info for the start scan step, so lets add the scan id to the output and add the meta payload to the debug logs.
- Under high load, jobs can occasionally take 15 minutes to process, for now lets raise the limit to 20min while we work on optimisations there. Let's also slow down the polling after the first two minutes to reduce load on the backend.
- Sometimes the client times out and never gets the response from the backend, which is awkward because the timeouts are the same at both ends. Ideally the client's timeouts are slightly higher than the backends so that the errors propagate appropriately.

Tested locally.

PR checklist:


cli/src/semgrep/app/scans.py
```diff
@@ -203,7 +203,7 @@ def start_scan(self, meta: Dict[str, Any]) -> None:
             logger.info(f"Would have sent POST request to create scan")
             return
 
-        logger.debug("Starting scan")
+        logger.debug(f"Starting scan: {json.dumps({'meta': meta}, indent=4)}")
```
zyannes, line 206: We'll likely want to mask this in the tests so the magicmock doesn't cause any regressions to the snapshots

jbergler: We don't snapshot the --debug version of the logs, so I think this is fine.

zyannes (approved):

PR #8667, opened 2023-09-08
feat: update HTTP mock for secrets testing

kopecs:
Test plan: `make test` in pro. See also https://github.com/returntocorp/semgrep-proprietary/pull/951.

PR checklist:


kopecs:
cc: @ajbt200128 not sure what the mypy issue is. Do you know?

ajbt200128:
> cc: @ajbt200128 not sure what the mypy issue is. Do you know?

This issue is unrelated to this PR. It looks like you are trying to merge a diff version of semgrep_interfaces than is in develop. This is def where the mypy issue is coming from

libs/networking/Http_mock_client.mli
```diff
@@ -66,6 +73,13 @@ val check_header : Cohttp_lwt.Request.t -> string -> string -> unit
   * Example: [check_header request "Authorization" "Bearer <token>"]
   *)
 
+val check_headers : Cohttp.Header.t -> Cohttp.Header.t -> unit
+(** [check_header request header header_value] will use Alcotest to assert
+  * two lists of headers are equal. Note that order and header capitlisation
```
ajbt200128, line 78: Nit: Should we ignore capitalization? Imo no, I assume most servers don't care about header capitalization, but also that'd really suck to debug if they do

libs/networking/Http_mock_client.ml
```diff
@@ -126,3 +136,146 @@ let with_testing_client make_fn test_fn () =
...
+            take_while (fun s -> String.length s <> 0) lines
+          in
+          let headers = parse_headers headers in
+          let body = String.concat "\n" body |> Body.of_string in
+          ( {
+              Cohttp.Response.version;
+              headers;
+              status;
+              (* Not sure what this field is for. Not documented *)
+              flush = false;
```
ajbt200128: Probably that IO should be flushed when writing this body vs. not flushing and allowing chunking/streaming. Most likely want this true, but idk if it'll actually change anything since this isn't ever going over the network

libs/networking/Http_mock_client.mli
kopecs: If a server isn't following the RfC I'm inclined to say it isn't our problem. I'd be more worried about creating work for if we change how headers are generated in fixing tests which should pass.

libs/networking/Http_mock_client.ml
kopecs: I don't think its buffered at all based on my testing, but I can make it true. Cohttp doesn't document the actual use.

libs/networking/Http_mock_client.mli
```diff
@@ -79,3 +93,40 @@ val with_testing_client : make_response_fn -> (unit -> unit) -> unit -> unit
...
+  * will return a new http client which may be used in lieu of the
+  * [http_helpers] client, which rather than performing network requests will
+  * instead look for a matching request in the given file and yield the
+  * given paired response. If no request matches, a test failure will be
+  * raised with Alcotest.
+  *
+  * The file format is
+  * ```
+  * > REQUEST (possibly multiple lines)
+  * < RESPONSE (possibly multiple lines)
```
ajbt200128, line 115: I'm ok with a custom file format here, but would appreciate more of an explanation of the format. It looks like this is just the actual bytes of the http request, but not sure, and I'd prefer not having to read parser code to figure it out.

libs/networking/Http_mock_client.mli
ajbt200128: Didn't know it was part of the RFC, totally agree then

libs/networking/Http_mock_client.mli
kopecs: Yes, it is essentially the actual bytes modulo line endings (\r\n not being required). Do you not feel the later example clarifies sufficiently? My priority is actually being able to use this more so than documenting every function precisely.

kopecs: resolve so ci runs

libs/networking/Http_mock_client.mli
kopecs: Current ref for posterity: https://www.rfc-editor.org/rfc/rfc9110.html#section-5.1-3

libs/networking/Http_mock_client.mli
ajbt200128: Can you add a comment about that? Just so people know how the test works in case they need to debug it

ajbt200128 (approved):

PR #8671, opened 2023-09-11
Make the `semgrep ci` tests for github envs more accurate

jbergler:
The mocked env didn't accurately reflect the behavior of github actions which meant that the snapshots weren't accurately reflecting what we send to semgrep.dev in a scan. Before I make changes to the behavior I thought it appropriate to update these tests so that the diffs in future PRs are easier to review/understand.

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -110,7 +110,7 @@ def fix_head_if_github_action(metadata: GitMeta) -> None:
     logger.info(f"Not on head ref: {metadata.head_branch_hash}; checking that out now.")
     git_check_output(["git", "checkout", metadata.head_branch_hash])
 
-    atexit.register(git_check_output, ["git", "checkout", stashed_rev])
+    atexit.register(git_check_output, ["git", "checkout", stashed_rev], os.getcwd())
 
 
 @click.command(context_settings=CONTEXT_SETTINGS)
--- a/cli/tests/e2e/test_ci.py
+++ b/cli/tests/e2e/test_ci.py
@@ -404,6 +404,7 @@ def mock_autofix(request, mocker):
 # The tests
 ##############################################################################
 
+
 @pytest.mark.parametrize(
     "env",
     [
--- a/cli/tests/e2e/test_ci.py
+++ b/cli/tests/e2e/test_ci.py
@@ -63,6 +63,34 @@ BAD_CONFIG = dedent(
 FROZEN_ISOTIMESTAMP = "1970-01-01T00:00:00"
 
 
+# To ensure our tests are as accurate as possible, lets try to autodetect what GITHUB_ vars
+# the app code uses, so the tests can enforce the env is mocked appropriately.
+_cli_src = (Path(__file__).parent.parent.parent / "src").resolve()
+USED_GITHUB_VARS = set(
+    subprocess.run(
+        f"git grep -hPo 'GITHUB_[\\w_]*' {_cli_src}", shell=True, capture_output=True
+    )
+    .stdout.decode()
+    .strip()
+    .split("\n")
+) - {
+    "GITHUB_TOKEN",  # not used in the cli, just passed to the backend
+    "GITHUB_EVENT_PATH",  # TODO: mock this for more than just PR events
+}
+
+assert "GITHUB_ACTIONS" in USED_GITHUB_VARS  # ensure the parsing did something
+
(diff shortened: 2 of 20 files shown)


cli/src/semgrep/commands/ci.py
```diff
@@ -110,7 +110,7 @@ def fix_head_if_github_action(metadata: GitMeta) -> None:
     logger.info(f"Not on head ref: {metadata.head_branch_hash}; checking that out now.")
     git_check_output(["git", "checkout", metadata.head_branch_hash])
 
-    atexit.register(git_check_output, ["git", "checkout", stashed_rev])
+    atexit.register(git_check_output, ["git", "checkout", stashed_rev], os.getcwd())
```
jbergler, line 113: I think this is an edge case that only shows up in testing because of how we invoke semgrep during the tests but by the time the atexit runs the working directory has changed, we can pass the current value to git_check_output to ensure this behaves consistently and reduces some noise in the test env.

cli/tests/e2e/test_ci.py
```diff
@@ -195,7 +223,7 @@ def git_tmp_path_with_commit(monkeypatch, tmp_path, mocker):
         check=True,
         capture_output=True,
     )
-    subprocess.run(["git", "pull", "origin"])
+    subprocess.run(["git", "fetch", "origin"])
```
jbergler, line 226: `pull` throws warnings here because there is no branch checked out in this new repo we created, `fetch` is less distracting for people trying to debug because it removes the meaningless error.

aryx:
Not super familiar with this code, but LGTM!

cli/tests/e2e/test_ci.py
```diff
@@ -376,6 +404,11 @@ def mock_autofix(request, mocker):
 # The tests
 ##############################################################################
 
+# "GITHUB_EVENT_NAME": "",
+#     "GITHUB_EVENT_PATH": "",
+#     "GITHUB_HEAD_REF": "", #
+#     "GITHUB_REF": "",
```
brendongo: mean to keep this?

brendongo (approved):

cli/tests/e2e/test_ci.py
jbergler: Nope, cleaned up.

cli/src/semgrep/commands/ci.py
aryx: would have been good to add this comment in the code itself. People in 3 months will wonder why we have this os.getcwd() argument at the end.

aryx:
thx for splitting PR in 2! I love when we first refactor to clean and then make the actual feature PR in another PR.

cli/tests/e2e/test_ci.py
```diff
@@ -63,6 +63,34 @@
 FROZEN_ISOTIMESTAMP = "1970-01-01T00:00:00"
 
 
+# To ensure our tests are as accurate as possible, lets try to autodetect what GITHUB_ vars
+# the app code uses, so the tests can enforce the env is mocked appropriately.
```
aryx, line 67: What is "the app code" in this context? Semgrep-app? the code in semgrep that interacts with the app?

cli/tests/e2e/test_ci.py
```diff
@@ -63,6 +63,34 @@
...
+_cli_src = (Path(__file__).parent.parent.parent / "src").resolve()
+USED_GITHUB_VARS = set(
+    subprocess.run(
+        f"git grep -hPo 'GITHUB_[\\w_]*' {_cli_src}", shell=True, capture_output=True
```
aryx, line 71: This currently captures the GITHUB_xxx string in cli/src/semgrep/interfaces/semgrep_output_v1.atd
which make 'make e2e' fail locally.
It's because I have
```
[submodule]
  recurse = true
  fetchJobs = 20
```
in my .gitconfig.

This git grep command seems a bit fragile to differences in environment ...
or maybe we should restrict it to the cli/src/semgrep/app/ directory?

cli/tests/e2e/test_ci.py
jbergler: I'm using "app" here as the generic application, ie semgrep cli

cli/tests/e2e/test_ci.py
jbergler: Restricting the grep might the safest, if you're ok with it - then I'd try this first though: https://github.com/returntocorp/semgrep/pull/8818

PR #8672, opened 2023-09-11
perf: hack: Reduce size of Input_to_core_t.targets

IagoAbal:
Helps PA-2660
(but I don't think it fixes it, we probably have other memory issues)

test plan:
% semgrep -c p/default -j64 stress-test-monorepo/repos/intellij-community
or use `Obj.reachable_words` to compare the size of `targets`

PR checklist:

### DIFF
--- a/src/runner/Run_semgrep.ml
+++ b/src/runner/Run_semgrep.ml
@@ -568,6 +568,34 @@ let xtarget_of_file (config : Runner_config.t) (xlang : Xlang.t)
     lazy_ast_and_errors;
   }
 
+(* [HACK "targets hash consing"]
+ * Since most rules do not define any 'includes'/'excludes', the 'rule_nums'
+ * for most targets of the same language tend to be the same. This results
+ * in a lot of wasted space. This data structure is then copied to each
+ * Parmap worker and, despite copy-on-write, due to GC, we may end up with
+ * one copy per thread. If we e.g. run p/default on intellij's repo, using
+ * 48 threads, this data structure alone may take over 4G of physical memory.
+ * With hash consing, it would take 480M instead (about 9x less memory).
+ * This should not be needed any more once we do file targeting in OCaml. *)
+let hack_targets_hash_consing (targets : Input_to_core_t.targets) =
+  let module T = Hashtbl.Make (struct
+    type t = int list
+
+    let equal xs ys = List.compare Int.compare xs ys =|= 0
+    let hash = Hashtbl.hash
+  end) in
+  let target_mappings =
+    let shared = T.create 10 in
+    targets.target_mappings
+    |> Common.map (fun (m : Input_to_core_t.target) ->
+           match T.find_opt shared m.rule_nums with
+           | None ->
+               T.add shared m.rule_nums m.rule_nums;
+               m
+           | Some rule_nums -> { m with rule_nums })
+  in
+  { targets with target_mappings }
+
 (* Compute the set of targets, either by reading what was passed
  * in -target, or by using our poor's man file targeting with
  * Find_target.files_of_dirs_or_files.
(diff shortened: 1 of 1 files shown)


src/runner/Run_semgrep.ml
```diff
@@ -624,6 +652,7 @@ let targets_of_config (config : Runner_config.t)
         | Target_file target_file ->
             File.read_file target_file |> In.targets_of_string
       in
+      let targets = hack_targets_hash_consing targets in
```
emjin, line 655: My only concern is that this could cause problems if we transform the targets later. There are two places we mess with the targets list: extract mode, and now secrets. I believe for both of these we're just adding new targets, but it doesn't seem implausible that one day we'd want to remove a rule from a target list for a specific target

emjin (approved):
Nice! I am a little bit concerned about the possibility of someone later wanting to modify the rules for a specific target. Overall, I don't think it prevents us from using the hack, but maybe we could include a warning where the targets are created in `semgrep_with_rules`

src/runner/Run_semgrep.ml
```diff
@@ -568,6 +568,34 @@ let xtarget_of_file (config : Runner_config.t) (xlang : Xlang.t)
     lazy_ast_and_errors;
   }
 
+(* [HACK "targets hash consing"]
+ * Since most rules do not define any 'includes'/'excludes', the 'rule_nums'
```
aryx, line 572: the include/exclude are actually handled now in semgrep-core ... so we definitely don't need anymore this complex Input_to_core.atd data structure

aryx: See the call to Filter_target.filter_paths in Run_semgrep.ml

aryx (commented):
What is the priority of this? Because it might be better to do the right thing and change Input_to_core.atd to not do all this complicated thing instead now that we do filter paths in semgrep-core (I added it for osemgrep)

src/runner/Run_semgrep.ml
IagoAbal: It was just an experiment but since it's just one function and easy to remove without breaking anything, I thought it could be useful to merge if that may help users with large monorepos. But assuming that our per-rule include/exclude is equivalent to that of pysemgrep, then we could just send Semgrep-core the list of targets per language, so we would just have:

```ocaml
type target <ocaml attr="deriving show"> = {
  (* Must be a valid "language" as defined in Xlang.mli.
     examples: "ocaml", "python", but also "spacegrep" or "regexp".
     less: would be great to use Lang.t here, but atd does not
     allow (by design) to use types defined in other files. *)
  language: xlang;
}
```

And semgrep-core will decide what rules run on each target. Is that what you are thinking of as the right thing?

aryx: Ideally yes; it's worth trying ... but maybe it's more complicated than I think to remove this rul_num field
and we get some regressions.

aryx: @IagoAbal if you want to help a little bit in our effort to get file targeting fully ocaml-side by Q3, this is a good way to help :)

emjin: Wait it's handled now regardless of whether you call it with `osemgrep` or not?

aryx: correct.
there's even a comment about it.

emjin: Let's get rid of Input_to_core.targets!!!!!!!!!!

emjin: (How much do we trust it? Do we need `--legacy` to still pass all the mappings?)

aryx:
Closing in favor on removing the problem at the source in Input_to_core.atd rule_nums field.

PR #8679, opened 2023-09-12
Create --code flag for users who want to run code-only scans

chmccreery:
PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/ci.py
+++ b/cli/src/semgrep/commands/ci.py
@@ -155,6 +155,7 @@ def fix_head_if_github_action(metadata: GitMeta) -> None:
     is_flag=True,
     hidden=True,
 )
+@click.option("--code", is_flag=True, hidden=True)
 @click.option("--beta-testing-secrets", is_flag=True, hidden=True)
 @click.option(
     "--suppress-errors/--no-suppress-errors",
@@ -175,6 +176,7 @@ def ci(
     autofix: bool,
     baseline_commit: Optional[str],
     beta_testing_secrets: bool,
+    code: bool,
     core_opts: Optional[str],
     config: Optional[Tuple[str, ...]],
     debug: bool,
@@ -309,6 +311,7 @@ def ci(
             console.print(Title("Connection", order=2))
             metadata_dict = metadata.to_dict()
             metadata_dict["is_sca_scan"] = supply_chain
+            metadata_dict["is_code_scan"] = code
             metadata_dict["is_secrets_scan"] = beta_testing_secrets
             proj_config = ProjectConfig.load_all()
             metadata_dict = {**metadata_dict, **proj_config.to_dict()}
--- a/cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines-overwrite-autodetected-variables/meta.json
+++ b/cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines-overwrite-autodetected-variables/meta.json
@@ -20,6 +20,7 @@
     "scan_environment": "azure-pipelines",
     "is_full_scan": true,
     "is_sca_scan": false,
+    "is_code_scan": false,
     "is_secrets_scan": false
   }
 }
--- a/changelog.d/gh-8679
+++ b/changelog.d/gh-8679
@@ -1,1 +1,1 @@
-Allow Semgrep CI users to specify Code product using `--code` command-line option. This works the same as `--supply-chain` now and fleshes out the product suite. 
+Allow Semgrep CI users to specify Code product using `--code` command-line option. This works the same as `--supply-chain` now and fleshes out the product suite.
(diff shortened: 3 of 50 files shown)


cli/src/semgrep/commands/ci.py
```diff
@@ -155,6 +155,7 @@ def fix_head_if_github_action(metadata: GitMeta) -> None:
     is_flag=True,
     hidden=True,
 )
+@click.option("--code", is_flag=True, hidden=True)
```
aryx, line 158: should we start to think of a more general approach like --products=code,sca instead of a flag for each product?

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines-overwrite-autodetected-variables/meta.json
```diff
@@ -20,6 +20,7 @@
     "scan_environment": "azure-pipelines",
     "is_full_scan": true,
     "is_sca_scan": false,
```
aryx, line 22: would be better to define a proper product type in semgrep_output_v1.atd
with Sca | Code | Secrets | ...
and then keep here instead a list of those products instead of 10 booleans.

aryx (commented):
can we specify the schema for this meta field in semgrep_output_v1.atd instead
of those untyped fields that are growing (and also use better types
than a long list of booleans)

cli/src/semgrep/commands/ci.py
akuhlens: I definitely think we should, doesn't seem to be what @milanwilliams is thinking is needed right now. You two should talk.

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines-overwrite-autodetected-variables/meta.json
akuhlens: I agree that this interface would be better if it was typed. I don't think this PR is the right place for that. Maybe we can add types (or ask someone else to) in preparation for the switch to osemgrep?

akuhlens: One nice thing about @aryx's proposed type representation is that it would be more stable to changes in these snapshots.

akuhlens (approved):
This is precisely what I would have done. We really should type this interface. Not sure who has the spare cycles to do it.

cli/src/semgrep/commands/ci.py
chmccreery: there has been a lot of discussion on this but people seem divided, so I think better to leave as-is in the absence of agreement to bother changing it

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines-overwrite-autodetected-variables/meta.json
chmccreery: agree with typing the interface, I can try to put up a PR for that later, but would rather not block this work on that change. Changing the interface is a whole can of worms, because the app always has to stay backwards-compatible with old versions of the CLI, so we will have to maintain backwards compatibility with the list of booleans approach anyway. I think we should hold off on that change until the interface is strongly typed (I will try to get to that this week).

cli/src/semgrep/commands/ci.py
aryx: If we were adding --secrets, I would be almost ok, but --code is pretty vague, looks like we want to code something ... so at least --products=code  is giving the necessary context to understand what code means

cli/tests/e2e/snapshots/test_ci/test_full_run/autofix-azure-pipelines-overwrite-autodetected-variables/meta.json
aryx: That's exactly why I'd rather make it better now, because otherwise we will have to support this ugly is_code_scan boolean forever.

aryx (changes requested):
I'd rather make it right now, that live with those bad flags and bad json fields that we have to support forever ...

aryx:
I can make the PR to change semgrep-interfaces if you want @chmccreery

chmccreery:
> I can make the PR to change semgrep-interfaces if you want @chmccreery

That would be great, @aryx! I will hold off on merging the app-side PR to make it backwards compatible and forwards-compatible with whatever you decide to do here

aryx:
Ok started to define types in https://github.com/returntocorp/semgrep-interfaces/pull/131
but looking at the code in meta.py and ci.py it might be quite complicated to migrate to it, so probably better to merge this PR

aryx (approved):

aryx:
circle CI seems unrelated, let's merge this.

PR #8685, opened 2023-09-12
Update Dockerfile to place semgrep-core in /home/semgrep/bin/ and update PATH

sinat101:
This addresses the permissions issues faced when attempting to use Pro Engine while running semgrep in the non-root Docker image.

PR checklist:


sinat101:
Tested locally by building the docker image and running the following:
```bash
[Sep 12 2023 01:23:55 AM] : sina @ ~/Code/mygithubpage : docker run --rm -v "${PWD}:/src" \
-e SEMGREP_APP_TOKEN="***" \
semgrep \
/bin/bash -c "semgrep install-semgrep-pro; ls -la /home/semgrep; semgrep ci --pro"
Semgrep Pro Engine will be installed in /home/semgrep/semgrep-core-proprietary
Downloading... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 160.5/160.5 MB 13.9 MB/s 0:00:00Successfully installed Semgrep Pro Engine (version b351eb2)!
total 299644
drwxr-sr-x    1 semgrep  semgrep       4096 Sep 12 05:27 .
drwxr-xr-x    1 root     root          4096 Sep 12 05:10 ..
-rw-r--r--    1 semgrep  semgrep         40 Sep 12 05:26 .gitconfig
drwxr-sr-x    2 semgrep  semgrep       4096 Sep 12 05:26 .semgrep
lrwxrwxrwx    1 root     root            12 Sep 12 05:10 osemgrep -> semgrep-core
-r-xr-xr-x    1 root     root     146289680 Sep 12 04:59 semgrep-core
-rwxr-xr-x    1 semgrep  semgrep  160521000 Sep 12 05:27 semgrep-core-proprietary
                  
                  
┌────────────────┐
│ Debugging Info │
└────────────────┘
                  
  SCAN ENVIRONMENT
  versions    - semgrep 1.39.0 on python 3.11.4                        
            
  CONNECTION
  Reporting start of scan for sinat101                  
  Fetching configuration from Semgrep Cloud Platform    
  Enabled products: Semgrep Code, Semgrep Supply Chain          
        
  ENGINE
  Using Semgrep Pro Version: b351eb2
  Installed at /home/semgrep/semgrep-core-proprietary
               
               
┌─────────────┐
│ Scan Status │
└─────────────┘
  Scanning 3 files tracked by git with 1677 Code rules, 581 Pro rules:
                                                                                
  <multilang>      61       9          Community    1096                        
  python          349       2          Pro rules     581                        
                                                                                
Semgrep Pro Engine may be slower and show different results than Semgrep OSS.
  2233 rules ran in a within-a-file fashion because `interfile: true` was not   
  specified.                                                                    
  (Use --verbose to see which ones.)

                
                
┌──────────────┐
│ Scan Summary │
└──────────────┘

  Found 0 findings (0 blocking) from 1677 rules.
  Uploading scan results  
  Finalizing scan           View results in Semgrep Cloud Platform:
    https://semgrep.dev/orgs/sinat101/findings
    https://semgrep.dev/orgs/sinat101/supply-chain
  ```

**Comment by sinat101** [comment]
closing in favor of https://github.com/returntocorp/semgrep/pull/8705

**Comment by sinat101** [comment]
closing in favor of https://github.com/returntocorp/semgrep/pull/8705

File: changelog.d/pa-3026.fixed [hunk 1]
```diff
@@ -0,0 +1,4 @@
+Updated location of `semgrep-core` binary to `/home/semgrep` and
```
aryx: this is actually /home/semgrep/bin

aryx (commented):
I think I prefer this to https://github.com/returntocorp/semgrep/pull/8705

aryx:
The other PR requires 64 lines to be added, 14 to be removed, this one 15 to add, 7 to remove; simpler wins.

aryx (approved):

Dockerfile
```diff
@@ -214,7 +217,8 @@ WORKDIR /src
 # Better to avoid running semgrep as root
 # See https://stackoverflow.com/questions/49193283/why-it-is-unsafe-to-run-applications-as-root-in-docker-container
 RUN adduser -D -u 1000 -h /home/semgrep semgrep \
-    && chown semgrep /src
+    && chown semgrep /src \
+    && chown semgrep:semgrep /home/semgrep/bin
```
aryx: why semgrep:semgrep whereas the other chown are using simply semgrep?
Seems better to be consistent no?

aryx (approved):

sinat101:
> The other PR requires 64 lines to be added, 14 to be removed, this one 15 to add, 7 to remove; simpler wins.

@aryx totally understand, no worries. Please give me some time to add a few tweaks/improvements before we merge, thank you!

Dockerfile
sinat101: @aryx i tried only setting the user to `semgrep` but then the group owner remains as `root`, which i thought might potentially be problematic. i don't know why we did `&& chown semgrep /src` to be honest, because the directory is still owned by `root`, just the files that are mounted at runtime are owned by `semgrep`:
```bash
/src $ ls -la
total 20
drwxr-xr-x    6 root     root           192 Sep 15 12:49 .
drwxr-xr-x    1 root     root          4096 Sep 17 19:51 ..
drwxr-xr-x   12 semgrep  semgrep        384 Sep 15 12:49 .git
-rw-r--r--    1 semgrep  semgrep         58 Sep 15 12:49 README.md
-rw-r--r--    1 semgrep  semgrep       3112 Sep 15 12:49 Sina_Taghizadeh_PortScanner.py
-rw-r--r--    1 semgrep  semgrep       7131 Sep 15 12:49 Sina_Taghizadeh_RoutingSim.py
```

Dockerfile
```diff
@@ -202,7 +202,7 @@ COPY --from=semgrep-core-container /src/semgrep/_build/default/src/main/Main.exe
 
 RUN ln -s semgrep-core /usr/local/bin/osemgrep
 
-# ???
+# Set some env variables
```
aryx: you can remove your comment. I now those set some env variables, it's just it would be good to document where those env variable are used and for which purpose. For example it's inclear what  this SEMGREP_USER_AGENT_APPEND is doing. It seems redundant with SEMGREP_IN_DOCKER. If we know we are in docker, why we need yet another env variable that set to Docker?

aryx: This is why I put those ??? so the person who know can add a comment explaining the purpose of those environment variables.

Dockerfile
```diff
@@ -239,4 +239,18 @@ LABEL maintainer="support@semgrep.com"
 # on the mounted volume when using instructions for running semgrep with docker:
 # `docker run -v "${PWD}:/src" -i returntocorp/semgrep semgrep`
 FROM semgrep-cli AS nonroot
+
+# Remove previous `osemgrep` symlink
```
aryx: you can remove those comments, they just paraphrasing what is below and so add extra things to read without any benefit.

A comment should be useful, such as explaining a design decision, explaining why we do it this way.
In this case maybe something like
"Without the adjustment belows, a user would not be able to run semgrep --install-pro in a nonroot docker image
because semgrep-core would be in /usr/local/bin which is protected. This is why we need to move it to a non-protected place.
alt: we could also do this work directly it in the root docker image
"

aryx (approved):

changelog.d/pa-3026.fixed
```diff
@@ -0,0 +1,4 @@
+For the `nonroot` Docker build stage, moved `semgrep-core` to
+`/home/semgrep/bin` and updated `$PATH` env variable with the
```
aryx, line 2: Those changelog entries are directed to the users of semgrep; they don't really care about how things were achieved; What matters is more "You can now run semgrep --install-semgrep-pro in nonroot docker images".

aryx (approved):

PR #8687, opened 2023-09-12
const-prop: Handle implicit number-to-string conversions

IagoAbal:
In languages like Java or JS/TS numbers are implicitly converted into strings when used in a string-concatenation expression.

Closes PRO-169

test plan:
make test # new test

PR checklist:

### DIFF
--- a/src/analyzing/Constant_propagation.ml
+++ b/src/analyzing/Constant_propagation.ml
@@ -286,10 +286,18 @@ let binop_bool_cst op b1 b2 =
       Some (Cst Cbool)
   | _b1, _b2 -> None
 
-let concat_string_cst s1 s2 =
+let concat_string_cst env s1 s2 =
   match (s1, s2) with
   | Some (Lit (String (l, (s1, t1), r))), Some (Lit (String (_, (s2, _), _))) ->
       Some (Lit (String (l, (s1 ^ s2, t1), r)))
+  | Some (Lit (String (l, (s1, t1), r))), Some (Lit (Int (Some m, _)))
+    when is_lang env Lang.Java || is_js env ->
+      (* implicit int-to-string conversion *)
+      Some (Lit (String (l, (s1 ^ string_of_int m, t1), r)))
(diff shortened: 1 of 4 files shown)


src/analyzing/Constant_propagation.ml
```diff
@@ -286,10 +286,18 @@ let binop_bool_cst op b1 b2 =
...
   | Some (Lit (String (l, (s1, t1), r))), Some (Lit (String (_, (s2, _), _))) ->
       Some (Lit (String (l, (s1 ^ s2, t1), r)))
+  | Some (Lit (String (l, (s1, t1), r))), Some (Lit (Int (Some m, _)))
+    when is_lang env Lang.Java || is_js env ->
+      (* implicit int-to-string conversion *)
+      Some (Lit (String (l, (s1 ^ string_of_int m, t1), r)))
+  | Some (Lit (String (l, (s1, t1), r))), Some (Lit (Float (Some m, _)))
+    when Float.is_integer m && is_js env ->
+      (* implicit float-to-string conversion *)
+      Some (Lit (String (l, (s1 ^ string_of_int (int_of_float m), t1), r)))
```
ihji: does this only cover the case like `"abc"+1.0`? What happens when we do `"abc"+1.5`?

aryx: I agree, seems wrong; Why not using string_of_float?

aryx (approved):

src/analyzing/Constant_propagation.ml
IagoAbal: Fair enough, I'll add the float case too! But in JS at least, the 1.0 seems to be printed as "1" so OCaml's `string_of_float` doesn't work.

IagoAbal: So I was just handling the integer case... but in JS we parse all numbers as floats, so I was detecting when the float corresponds to an integer. I'll leave a comment too.

PR #8688, opened 2023-09-12
feat: JSON.member function in commons

kopecs:
PR checklist:

### DIFF
--- a/libs/commons/JSON.ml
+++ b/libs/commons/JSON.ml
@@ -15,6 +15,14 @@ type t =
   | Null
 [@@deriving show]
 
+let member m j =
+  match j with
+  | Object members ->
+      List.find_map
+        (fun (m', x) -> if String.equal m m' then Some x else None)
+        members
+  | _ -> None
+
 let rec (to_yojson : t -> Y.t) = function
   | Object xs -> `Assoc (xs |> List.map (fun (s, t) -> (s, to_yojson t)))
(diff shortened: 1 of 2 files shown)


brandonspark (approved):

libs/commons/JSON.ml
```diff
@@ -15,6 +15,14 @@ type t =
   | Null
 [@@deriving show]
 
+let member m j =
```
aryx, line 18: Should probably be called lookup at this point, 

Also maybe time to define a JSON.mli to better document the API of this module.

aryx (changes requested):

libs/commons/JSON.ml
kopecs: The equivalent function in Yojson is named [`member`](https://ocaml-community.github.io/yojson/yojson/Yojson/Basic/Util/index.html#val-member) (albeit that one raises an exception, rather than returning an option), so I was going for consistency here. Also, I would find member more descriptive than lookup in this context.

Happy to make a .mli file.

kopecs: @aryx in that case, do you understand the "(* compatibility mode with json-wheel *)" comment? Not sure where that should end up/relevancy thereof (can't find adding commit---looks like git history got lost since you moved this all from pfff).

aryx (approved):

PR #8690, opened 2023-09-12
feat(cfg): add support for try-catch-else-finally

amchiclet:
In Ruby, the try-catch-else (as well as try-catch-else-finally) statement is captured as an [`OtherStmtWithStmt`](https://github.com/returntocorp/semgrep/blob/965c454931a1c2f3939b16761b801ab3fdd4ed86/libs/ast_generic/AST_generic.ml#L1204) in the generic ast and looks something like

```
OtherStmtWithStmt (
    OSWS_Else_in_try,
    a_list_of_any,
    Block(..., [ try_catch_finally_stmt; else_stmt], ...)
)
```

Python and Julia also have this construct.

Previously this statement isn't handled in AST_to_IL.ml. This PR adds support for it.

Test plan:
* `make core-test` and added more tests.

Closes
* [PA-3053](https://linear.app/semgrep/issue/PA-3053).
* [PA-3054](https://linear.app/semgrep/issue/PA-3054).


amchiclet:
@aryx @IagoAbal 

Can I get some early thoughts on the direction of this PR?

In particular,
* I switched up where the `else` token belongs. The original `ruby_to_generic.ml` looks wrong to me, but I might have misunderstood it.
* I used the `any list` in [`OtherStmtWithStmt`](https://github.com/returntocorp/semgrep/blob/965c454931a1c2f3939b16761b801ab3fdd4ed86/libs/ast_generic/AST_generic.ml#L1204) to pass along the token info for `else`. I'm not sure if `any list` was intended for this or not.
* I used [`OtherCatch`](https://github.com/returntocorp/semgrep/blob/965c454931a1c2f3939b16761b801ab3fdd4ed86/libs/ast_generic/AST_generic.ml#L1258) as the catch pattern. I'm not sure if I should be adding a new constructor `ElseCatch` instead.

src/analyzing/AST_to_IL.ml
```diff
@@ -1452,6 +1470,21 @@ and stmt_aux env st =
   | G.OtherStmtWithStmt (G.OSWS_Block _, [ G.E objorig ], stmt1) ->
       let ss, _TODO_obj = expr_with_pre_stmts env objorig in
       ss @ stmt env stmt1
+  (* Ruby: begin ... rescue ... else ... ensure ... *)
+  | G.OtherStmtWithStmt
```
aryx: Usually it's better to keep the AST_to_IL.ml as simple as possible and as general as possible. So here it would probably be better to adapt ruby_to_generic.ml to generate a better AST so that the existing code in AST_to_IL.ml would handle it for free.
I'm not super familiar with the semantic of begin/try/else/rescue in Ruby, so that's why I hacked the current translation, because I didn't have the time to make it better, but if you now fully understand the semantic of it, then just improve ruby_to_generic.ml to generate the right Try/Catch/Finally construct directly,  possibly extending the generic AST to make the current try/catch construct more general.

Those OtherStmtWithStmt are used as a last resort when we don't have the time to fully understand the semantic and we still want some CFG for it and some support for it, but in the long term we should never use those last-resort cases.

src/analyzing/AST_to_IL.ml
```diff
@@ -363,6 +363,40 @@ and pattern_assign_statements env ?(eorig = NoOrig) exp pat =
...
+        (name, todo_pattern @ catch_stmt) :: acc)
+      [] catches
+  in
+  let all_handlers_rev =
+    match opt_else with
+    | None -> catches_stmt_rev
+    | Some (tok, else_st) ->
+        let name = fresh_var env tok in
+        let todo_pattern =
+          fixme_stmt ToDo (G.Ce (G.OtherCatch (("else_catch", tok), [])))
```
aryx: This OtherCatch, just like OtherStmtWithStmt, OtherExpr, etc. are really last resort constructs when we're lazy and we don't want to properly handle the thing, especially if it's for translating a construct that seem used for only one language. Maybe here there's the opportunity to improve the AST_generic.catch_exn type and add indeed a CatchElse case, with a nice comment before explaining maybe its semantic, the difference between else and finally, etc.
Otherwise by using OtherCatch with those string like "else_catch" is not super type-safe because it's strongly coupled with the string we generate in ruby_to_generic.ml, which if we change one day the typechecker will not tell us to also update the code in AST_to_IL.ml

aryx (changes requested):
I would vote for improving the AST_generic with a CatchElse, and generating a better AST_generic in ruby_to_generic.ml, without the use of an OtherStmtWithStmt
(again maybe you will need to change a little bit the Try stmt type to accomodate Ruby, or maybe you can find a way with just CatchElse to not need to alter the Try stmt).

Also it's good to add test case showing how this PR improve things, so here maybe a taint rule or something showing how we handle more case because the CFG is handled in a better way internally.

src/analyzing/AST_to_IL.ml
IagoAbal: I agree overall with Yoann, BUT if we add a proper test in this PR then we'll know if this breaks. And we could leave improving `AST_generic` for a separate PR perhaps.

IagoAbal: Wait, this is just a `ToDo` IL node that you are generating, so this won't break anything anyways.

src/analyzing/AST_to_IL.ml
```diff
@@ -1452,6 +1470,21 @@ and stmt_aux env st =
...
+        [ G.Tk else_tok ],
+        {
+          s =
+            G.Block
+              ( _,
+                [ { s = G.Try (_, try_st, catches, opt_finally); _ }; else_st ],
+                _ );
+          _;
+        } ) ->
+      try_catch_else_finally env ~try_st ~catches
```
IagoAbal: Definitely need some taint test to exercise this translation.

src/analyzing/AST_to_IL.ml
IagoAbal: Could we then try adding

```ocaml
type try_else = tok * stmt
...
| Try of tok * stmt * catch list * try_else option * finally option 
```

?

Or do we need something else?

src/analyzing/AST_to_IL.ml
IagoAbal: I am just no sure why we use `(G.OtherCatch (("else_catch", tok), []))` for the TODO node. That is not how the `else` is encoded by the Ruby parser anyways. You could just put the token there, and use `fresh_var ~str:"_try_else"` so that if someone examines the IL easily knows what that TODO corresponds to.

src/analyzing/AST_to_IL.ml
amchiclet: I'd like to try adding this and will update on findings.

amchiclet:
Updates:

I tried the approach to support `Try of tok * stmt * catch list * try_else option * finally option` (with the new `try_else` clause), and pushed the new changes to this PR.

It required some grunt work to make different languages compile, and some extra work for Python, but it helped simplify the matching logic. I feel like this is the right approach, but I may be missing something.

Don't pay attention to CFG_builder.ml yet. It's still WIP, and I need a lot more tests.

But here are some things I'd like to call out. Mostly related to Python.

First, I had to modify [`Visitor_python.ml`](languages/python/menhir/Visitor_python.ml) and that file has a comment saying 
```
  (* start of auto generation *)                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                              
  (* generated by ocamltarzan with: camlp4o -o /tmp/yyy.ml -I pa/ pa_type_conv.cmo pa_visitor.cmo  pr_o.cmo /tmp/xxx.ml  *)
```
so I'm not sure if I should have touched this file or generate it from with `ocamltarzan` like the comment says.

Also I'm not sure what to do for [`AST_generic_to_v1.ml`](libs/ast_generic/AST_generic_to_v1.ml) yet. It seems like version 1 is a serializable format defined as a `.atd` file, and to change it, I need to define version 2 so it doesn't break things. I just added a TODO comment, because changing the interface seems like a big deal.

Otherwise, do you think like extending `Try` to be more generic is the right approach after roughly looking at the potential changes?

changelog.d/invalidfile.added
aryx, line 1: What is this invalidfile.added ?

languages/julia/generic/Parse_julia_tree_sitter.ml
```diff
@@ -2071,33 +2077,25 @@ and map_statement (env : env) (x : CST.statement) : stmt list =
...
-                | Some _x ->
-                    (* map_else_clause env x *)
-                    (* TODO: We can't actually accommodate this within the Generic AST.
-                       "else" is something kind of weird where it's something you enter
-                       if you do not enter the "catch", but it's not the same as a
-                        "finally", which always runs in either case.
-                       https://docs.julialang.org/en/v1/manual/control-flow/#else-Clauses
-                    *)
-                    None
+                | Some x -> Some (map_try_else_clause env x)
```
aryx: great! that means this is useful not only for Ruby but also Julia! cc @brandonspark

languages/python/generic/Python_to_generic.ml
```diff
@@ -768,28 +768,12 @@ and stmt_aux env x =
...
-                     (G.OS_TryOrElse, orelse |> Common.map (fun x -> G.S x))
-                   |> G.s;
-                 ])
-            |> G.s;
-          ])
-  | TryFinally (t, v1, t2, v2) ->
-      let v1 = list_stmt1 env v1 and v2 = list_stmt1 env v2 in
-      (* could lift down the Try in v1 *)
-      [ G.Try (t, v1, [], Some (t2, v2)) |> G.s ]
+      and orelse = option (try_else env) v3
```
aryx, line 774: so useful also for Python! Woohoo!

libs/ast_generic/AST_generic_to_v1.ml
```diff
@@ -766,12 +766,16 @@ and map_stmt x : B.stmt =
...
-    | Try (t, v1, v2, v3) ->
+    | Try (t, v1, v2, v3, v4) ->
         let t = map_tok t in
         let v1 = map_stmt v1
         and v2 = map_of_list map_catch v2
-        and v3 = map_of_option map_finally v3 in
-        `Try (t, v1, v2, v3)
+        and _v3 = map_of_option map_try_else v3
+        and v4 = map_of_option map_finally v4 in
+        (* TODO: some languages such as ruby have try-catch-else-finally
```
aryx, line 775: yes this TODO is fine. You can keep it as a TODO.

src/analyzing/CFG_build.ml
```diff
@@ -47,15 +47,16 @@ type label_key = string * G.sid
...
-  (* When there is a 'return' we need to know the exit node to link to *)
-  exiti : F.nodei;
   (* Attaches labels to nodes. *)
   labels : (label_key, F.nodei) Hashtbl.t;
   (* Gotos pending to be resolved, a list of Goto nodes and the label
    * to which they are jumping. *)
   gotos : (nodei * label_key) list ref;
-  (* If we are inside a Try, this is the start of the handlers. *)
-  try_catches_opt : F.nodei option;
+  (* TODO: add comments. *)
```
aryx: yes please add comments

src/analyzing/CFG_build.ml
```diff
@@ -231,7 +233,9 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
       let finalthen = cfg_stmt_list state (Some newfakethen) st in
       state.g |> add_arc_from_opt (finalthen, newi);
       CfgFirstLast (newi, Some newfakeelse)
-  | Label label -> CfgLabel label
+  | Label label ->
+      Printf.printf "where the heck is this label?\n";
```
aryx: no printf!

src/matching/Trace_matching.ml
```diff
@@ -13,7 +13,7 @@ open Printf
 let on = false
 
 (* max_depth: Print the nodes just deep enough to see something useful. *)
-let max_depth = 3
+let max_depth = 30
```
aryx: probably need to restore this to 3 after you debugged things

aryx (commented):
Looks good! This is great when we extend the generic AST and it helps for the translation of more than just one language (here it helps 3 so this is great).

changelog.d/invalidfile.added
amchiclet: Oh I was testing some lint errors on a different branch and somehow it got included in this branch. I'll remove it.

src/analyzing/CFG_build.ml
amchiclet: Oops. :)

src/analyzing/CFG_build.ml
```diff
@@ -47,15 +47,27 @@ type label_key = string * G.sid
...
-  try_catches_opt : F.nodei option;
+  (* Destination node that a return node should go to. *)
+  return_destination : F.nodei;
+  (* True when a return may occur in any of the nodes being visited.
+   * Initialized as false, and once set to true, remains true forever.
+   * This field helps us determine whether we should propagate the return
+   * or not after executing a finally clause in exception handling.
+   *)
+  may_return : bool ref;
+  (* Destination node that a return node should go to. *)
```
ihji: typo here.

src/analyzing/CFG_build.ml
```diff
@@ -240,41 +247,91 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
   | Return (tok, e) ->
       let newi = state.g#add_node { F.n = F.NReturn (tok, e) } in
       state.g |> add_arc_from_opt (previ, newi);
-      state.g |> add_arc (newi, state.exiti);
+      state.g |> add_arc (newi, state.return_destination);
+      state.may_return := true;
       CfgFirstLast (newi, None)
-  | Try (try_st, catches, finally_st) ->
+  | Try (try_st, catches, else_st, finally_st) ->
```
ihji, line 255: This change doesn't seem to be straightforward. It looks like we'll need thorough testing to ensure the new logic works correctly. Do we currently have a unit test for the CFG building code?

ihji (commented):
Great job! I'm really looking forward to seeing some test cases in action for this new change.

src/analyzing/CFG_build.ml
aryx: We don't have unit tests for that I think. The test plan is usually "run -cfg_il and looks like it's ok".
Maybe we should start some test infra there (in a separate PR). Not super easy though to specify some tests and asserts that are readable.

amchiclet: Currently, some of the tainting and constant propagation unit tests rely on the correctness of the CFG.

I'm currently writing some more unit tests that exercise these new arcs. Once I have those, I'll update this PR again, and I think it will be ready for review.

If a separate infra is useful, then that would be in a separate PR.

Thanks for all of your early feedback!

src/analyzing/CFG_build.ml
amchiclet: Thanks!

amchiclet:
I'm confused by the failing tests because they work fine in my local copy. Let me try switching it to "ready for review" and see if the tests pass or not.

amchiclet:
Still failing. I'm creating a new PR. If that passes, I'll close this one and point to this PR from that one to save this discussion history.

amchiclet:
OK, so new PR (rebased from develop) is also failing on the same test `tests/semgrep-rules/ruby/rails/security/audit/avoid-tainted-file-access.yaml`.

I'll try to figure out what's going on and continue the discussion in this PR.

amchiclet:
Tried to reproduce by fetching the same commit hash (from [this line](https://github.com/returntocorp/semgrep/actions/runs/6203530534/job/16844691910?pr=8690#step:3:46)), but still can't reproduce the failing test.

```
git clone git@github.com:returntocorp/semgrep.git semgrep-temp-delete-me-please
cd semgrep-temp-delete-me-please
git fetch origin +5cb126741dc3968193df8015845558402c76d790:refs/remotes/pull/8690/merge
git checkout refs/remotes/pull/8690/merge
make dev-setup
make core-test

# all tests pass locally
```

Next, I'll probably try the different `make` steps and possibly see if the running in alpine (what CI does) makes a difference or not.

amchiclet:
Updates:

I tried running the tests in an alpine environment, but on mac os, not ubuntu like in CI.

```
# Get the same docker image as CI
docker pull returntocorp/ocaml:alpine-2023-06-16

# "Log in" to container
docker run --name debug -it returntocorp/ocaml:alpine-2023-06-16

# Get the code from this PR and build
git clone https://github.com/returntocorp/semgrep.git &&
  cd semgrep &&
  git fetch origin +5cb126741dc3968193df8015845558402c76d790:refs/remotes/pull/8690/merge &&
  git checkout refs/remotes/pull/8690/merge &&
  git submodule sync &&
  git submodule update --init

opam $(opam env)

make install-deps-ALPINE-for-semgrep-core &&
  make install-deps-for-semgrep-core &&
  make core

opam exec -- make core-test
```

The tests were still passing.

I have a hypothesis that the CFG for `avoid-tainted-file-access.rb` is too large, especially with the changes in this PR, and the CI environment is not as powerful as my workstation. I hypothesize there's some cap, either memory or time, that affects correctness of the control flow to avoid using too much time and memory. Just a guess.

This PR assumes that function calls may throw exceptions, so there's an extra arc to the exit node when there's no surrounding try-catch. Previously there was no arc if there's no surrounding try-catch.

I think I will remove this extra arc for function calls to keep the CFG roughly the same size like before. I tried unconditionally removing this extra arc, and `avoid-tainted-file-access.rb` passes in CI again. (It always passes locally.)

I'll update the code tomorrow, and hopefully it will finally be ready for review.

amchiclet:
Alternatively, breaking down the single big (200+ lines) function in `avoid-tainted-file-access.rb` into multiple smaller functions seems to solve this specific issue too, but I feel like 200+ line functions are reasonable in real life, so breaking this test down only masks the issue.

src/analyzing/AST_to_IL.ml
amchiclet: Went with the unified `Try` statement direction. Marking this as resolved.

src/analyzing/AST_to_IL.ml
amchiclet: This is direction I've been taking. Seems like things are looking good. Resolving this comment.

amchiclet:
Ok, I think this PR is ready for review. Phewwww.

aryx (approved):
I didn't have a deep look at the CFG_build, but overall looks good to me!
Great work!

ihji (approved):
Looks great! Your effort here is truly remarkable and highly appreciated 👍

changelog.d/pa-3054.added
```diff
@@ -0,0 +1,32 @@
...
+    d = tainted_input
+  except RuntimeError:
+    a = tainted_input
+    c = sanitize(c)
+  else:
+    b = tainted_input
+  finally:
+    d = sanitize(d)
+
+  # a is not tainted because exception wasn't thrown
```
IagoAbal: For any arbitrary code, how do you know that no exception was thrown?

src/matching/SubAST_generic.ml
```diff
@@ -302,13 +302,18 @@ let substmts_of_stmt st =
...
       @ (xs |> Common.map Common2.thd3)
       @
-      match opt with
+      match opt1 with
       | None -> []
-      | Some (_, st) -> [ st ])
+      | Some (_, st) -> (
+          [ st ]
+          @
+          match opt2 with
```
IagoAbal: This doesn't look right, here `opt2` is only matched if `opt1` is `Some _`. If `opt1` is `None` and `opt2` is `Some _`, I think we still want to visit `opt2`, no? Shouldn't they be matched independently of each other?

tests/rules/taint_exception.py
```diff
@@ -0,0 +1,242 @@
+# This test aims to exercise various exception paths.
```
IagoAbal, line 1: It's also interesting to write a test using constant propagation, because it will also be affected by this change, and it behaves differently than taint (must vs may).

src/analyzing/CFG_build.ml
```diff
@@ -47,15 +47,33 @@ type label_key = string * G.sid
...
   (* Gotos pending to be resolved, a list of Goto nodes and the label
    * to which they are jumping. *)
   gotos : (nodei * label_key) list ref;
-  (* If we are inside a Try, this is the start of the handlers. *)
-  try_catches_opt : F.nodei option;
+  (* Destination node that a return node should go to. *)
+  return_destination : F.nodei;
+  (* True when a return may occur in any of the nodes being visited.
+   * Initialized as false, and once set to true, remains true forever.
+   * This field helps us determine whether we should propagate the return
```
IagoAbal: Could you clarify which situation is this describing?
> This field helps us determine whether we should propagate the return or not after executing a finally clause in exception handling.

What is the "the return" here? Could you provide an example?

IagoAbal: Hmm, OK, is this to make sure that a `return` inside a `try` will execute the `finally` clause before returning? IMO this deserved a separate PR (but let it be here for now).

src/analyzing/CFG_build.ml
```diff
@@ -240,41 +259,122 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
+      (* The following will help determine whether we should exit
+       * or continue the flow after the whole statement
+       *)
+      (* When this is true, the flow stops after the finally clause. *)
+      let try_must_return =
+        match finaltry with
+        | None -> !(state.may_return) && not !(state.may_throw)
+        | _ -> false
+      in
+      (* When this is true, control must go to the else clause. *)
```
IagoAbal: And if it is `false`, does it mean it will not go through the `else` clause?

src/analyzing/CFG_build.ml
```diff
@@ -240,41 +259,122 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
+      let try_must_not_exit =
+        match finaltry with
+        | None -> false
+        | _ -> (not !(state.may_throw)) && not !(state.may_return)
```
IagoAbal: Better `Some _`. In general we consider `_` wrong unless enumerating all possibilities is unreasonable. (We have a Semgrep rule to detect these cases but for some reason not always triggers.)

src/analyzing/CFG_build.ml
```diff
@@ -240,41 +259,122 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
+      in
+
+      (* From end of try to else. *)
+      state.g |> add_arc_from_opt (finaltry, elsei);
```
IagoAbal, line 282: I'm confused because previously we have:

```ocaml
      (* When this is true, control must go to the else clause. *)
      let try_must_not_exit =
      ...
```
and here I see `finaltry -> elsei` being set unconditionally.

src/analyzing/CFG_build.ml
```diff
@@ -240,41 +259,122 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
+      in
+      let else_state = { state with throw_destination; return_destination } in
+      let finalelse = cfg_stmt_list else_state (Some elsei) else_st in
+
+      (* If the else clause is reached, and this is true, then the flow stops
+       * after the finally clause.
+       *)
+      let else_must_exit =
+        match finalelse with
+        | None -> true
```
IagoAbal: Don't we need to check `may_return` here?

src/analyzing/CFG_build.ml
```diff
@@ -240,41 +259,122 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
+      state.g |> add_arc_from_opt (finalelse, newfakefinally);
+
+      (* From catches to finally in case of uncaught exceptions. *)
       state.g |> add_arc (catchesi, newfakefinally);
+
       catches
       |> List.iter (fun (_, catch_st) ->
-             let finalcatch = cfg_stmt_list state (Some catchesi) catch_st in
+             let catch_state =
+               { state with throw_destination; return_destination }
```
IagoAbal: So this is the same `throw_destination` as we used for `else` ?

src/analyzing/CFG_build.ml
```diff
@@ -240,41 +259,122 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
   | Return (tok, e) ->
       let newi = state.g#add_node { F.n = F.NReturn (tok, e) } in
       state.g |> add_arc_from_opt (previ, newi);
-      state.g |> add_arc (newi, state.exiti);
+      state.g |> add_arc (newi, state.return_destination);
+      state.may_return := true;
```
IagoAbal: What happens if you have this code:

```python
if something:
  return;
try:
  ...
```
The `return;` will set `may_return` to true even though it's not inside the `try`, right? How would that affect the CFG of the `try`?

src/analyzing/CFG_build.ml
```diff
@@ -169,7 +187,8 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
                * this should reduce false positives (since it's a must-analysis).
                * Ideally we should have a preceeding analysis that infers which calls
                * may (or may not) raise exceptions. *)
-            state.g |> add_arc_opt_to_opt (Some newi, state.try_catches_opt);
+            state.g |> add_arc_opt_to_opt (Some newi, state.throw_destination);
+            state.may_throw := true;
```
IagoAbal: What if you have this code:

```python
try:
  foo();
  ...
finally:
 ...
bar();
try:
  ... 
```
I suspect the `foo()` in the first `try` will set `may_throw` and that will affect the CFG of the second `try`.

IagoAbal (changes requested):
The changes for adding `else` are uncontroversial, but you also made changes to the CFG generation where I have a lot of questions. (The changelog is a bit misleading in this respect.)

changelog.d/pa-3054.added
IagoAbal: If there is a potential path for `a` to get tainted (and that seems to be the case) then it should be tainted, since it's a may-analysis. But I also don't clearly see why this `a` is not tainted here based on the code, need to check the tests you wrote too.

src/analyzing/CFG_build.ml
```diff
@@ -240,41 +259,122 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
+       *)
+      if !(state.may_throw) then
+        state.g |> add_arc_opt_to_opt (finalfinally, state.throw_destination);
+      if !(state.may_return) then
+        state.g |> add_arc_from_opt (finalfinally, state.return_destination);
+
+      (* If we absolutely know that we need to propagate the exit, then
+       * we don't continue the dataflow after this statement.
+       *)
+      if try_must_return || (try_must_not_exit && else_must_exit) then
```
IagoAbal: All these conditions (`try_must_return` etc) are basically only used here, so I would rather defined them right here too.

IagoAbal: Also, what if `try_must_return` is `false` but the `try` ends in `goto`. Does this work correctly?

tests/rules/taint_exception.py
```diff
@@ -0,0 +1,242 @@
...
+
+def throw_exits(input):
+  clean = None
+  raise RuntimeError()
+
+  # ok:
+  sink = clean
```
IagoAbal: `sink = clean` ? Or `sink(clean)` ?

tests/rules/taint_exception.py
```diff
@@ -0,0 +1,242 @@
...
+  #ok: unreachable, because throw inside the catch clause exits
+  sink(clean)
+
+def throw_in_else_exits(input):
+  clean = None
+  try:
+    pass
+  except Exception as e:
+    clean = input
+    #ok: unreachable, because exception was not thrown
```
IagoAbal, line 26: I wonder if we're trying to be too smart here. Who writes this code? And if they write it, is it really that bad that we report a finding here?

changelog.d/pa-3054.added
IagoAbal: Could you point me where in the code we ensure that `a` he isn't tainted?

tests/rules/taint_exception.py
IagoAbal: Could you point me where in the code we make this unreachable?

IagoAbal: Ah OK I see, now the arcs to the `catch` are only added via `throw_destination`. Fine with that, I think it deserves a comment where we used to connect `finaltry` with `catchesi`, like "we do not directly connect finaltry with catchesi but whenever we encounter a Call or a Throw inside a Try, we add an arc ...".

tests/rules/taint_exception.py
```diff
@@ -0,0 +1,242 @@
...
+      clean2 = input
+      #ok: unreachable because the try clause returns
+      sink(clean2)
+    finally:
+      # Finally clauses are always reachable when present
+      dirty1 = input
+
+    clean3 = input
+    #ok: unreachable because the inner try is returning
+    sink = clean3
```
IagoAbal: `sink(clean3)` ???

src/matching/SubAST_generic.ml
amchiclet: I think you're right. Thanks.

tests/rules/taint_exception.py
amchiclet: I thought I updated all this. Thanks. I'll do another round of query and replace.

src/analyzing/CFG_build.ml
amchiclet: Yeah, I updated `try_must_not_exit` (aka try must continue) to the current code because what I actually want is to have this arc if `try` may continue.

`Try` may continue when `finaltry` is `Some x`, and `add_arc_from_opt` will take care of.

It's not unconditional, because `finaltry` might be `None` if `try` must exit.

I should add more comments.

amchiclet:
Based on the comments, I'm going to scope this PR.

`may_throw` and `may_return` seems to be adding too much complexity for now.

But I think `may_throw` alone is still suitable to be included and will simplify this PR quite a bit, so I'll do that next, while also addressing relevant comments.

I'd like to spend more time thinking about all interactions between exceptions, returns, and gotos, so better not to tackle those in this PR.

src/analyzing/CFG_build.ml
amchiclet: I'm removing return-related things in this PR. I'll resolve this.

src/analyzing/CFG_build.ml
amchiclet: Good to know!

src/analyzing/CFG_build.ml
amchiclet: I'll removed this logic. It makes this PR too complicated.

src/analyzing/CFG_build.ml
amchiclet: This logic will be removed to simplify this PR.

src/analyzing/CFG_build.ml
amchiclet: Yes, if else or catch throws it should go to finally or propagate. The binding can be reused, but making a new binding to be explicit here. (Or if there's a better name that conveys an environment for both catch and else, that would help.)

src/analyzing/CFG_build.ml
amchiclet: Thanks for catching this bug! It will be fixed in the next update.

src/analyzing/CFG_build.ml
amchiclet: Will be fixed the same way as one of your other comments above.

src/analyzing/CFG_build.ml
amchiclet: Will remove this logic to simplify this PR, but basically, it was needed above because it may be overridden afterwards. Eww, imperative programming :)

tests/rules/taint_exception.py
amchiclet: Sounds good! Next update will include some constant propagation tests.

src/analyzing/CFG_build.ml
amchiclet: Removing this logic to simplify this PR.

tests/rules/taint_exception.py
amchiclet: Will add a comment to explain this.

changelog.d/pa-3054.added
amchiclet: Based on your other comment, I believe you found the code, but happy to discuss more if you have any questions.

I've added a bit more details in the changelog to make it clearer what assumptions are updated in this change.

changelog.d/pa-3054.added
```diff
@@ -0,0 +1,38 @@
...
+any statement inside the try clause may throw an exception, but now only
+function calls are assumed to possibly throw exceptions.
+
+Throw statements always throw an exception as it was before.
+
+This kind of statement is supported in languages including Python, Ruby, and Julia.
+
+Python example:
+
+```py
```
IagoAbal, line 13: After you removed `may_return` etc, please check if this example still behaves as you describe in the current PR.

src/analyzing/CFG_build.ml
```diff
@@ -228,58 +238,115 @@ let rec cfg_stmt : state -> F.nodei option -> stmt -> cfg_stmt_result =
...
+
+      (* Also propagate any throws that may have happened inside this
+       * try-catch-else-finally statement at the end of the finally clause.
+       * If there's no finally clause, the arcs should already be created
+       * from the code above that sets proper throw_destinations before visiting.
+       *)
+      let may_throw =
+        try_may_throw || else_may_throw || catch_may_throw || finally_may_throw
+      in
+      if may_throw then
```
IagoAbal, line 331: Should this be `may_throw && not has_finally` ? Based on

> If there's no finally clause, the arcs should already be created from the code above that sets proper throw_destinations before visiting.

IagoAbal (approved):

changelog.d/pa-3054.added
amchiclet: Yep!

src/analyzing/CFG_build.ml
amchiclet: I think the comment added more confusion than clarity, so I'm removing it.

We definitely want to enter this condition when `has_finally` is true. The comment meant to say something like "don't worry, we're not missing arcs that do the propagation".

amchiclet:
Rebasing and seeing if the tests pass.

amchiclet:
I think the failing pro engine checks are unrelated, so I'm going ahead and merging this.

Thanks @aryx @ihji @IagoAbal for the feedback. I'll try to make my future PRs more scoped.

tests/patterns/cp_exception.py
```diff
@@ -0,0 +1,5 @@
+def f():
```
aryx, line 1: @amchiclet can you move this test file under tests/patterns/python/ so it does not clutter the tests/patterns/ directory?
(now actually OSS/tests/patterns/).

Issue #6163, opened 2022-09-22
Autofix rewrites CRLF line separators to LF line separators 


dmgping:
**Description**
Running a scan with the cli using a rule that has a fix definition and passing in the --autofix flag to implement the fix rewrites CRLF line separators to LF line separators. 

**To Reproduce**

**1** - Create an example.html file with the following contents 

```
<!DOCTYPE html>
#* Sample File *#
<html lang="$locale.getLanguage()" dir="ltr">
<body>
    <label class="class-name">
        <input type="type" name="$unescapedName" value="$unescapedValue"/>
    </label>
</body>
</html>
```

**2** - ensure it has CRLF line separator by running this command:
"find example.html | xargs file" 
This should output "example.html: HTML document text, ASCII text, with CRLF line terminators"

If not, run "unix2dos example.html" to convert it , run  "find example.html | xargs file" again to confirm.

**3** - Create a rule.yaml file with the following contents 
"
```
rules:
- id: HTML-XSS-Escape-all-html-input
  languages: [generic]
  paths:
    include:
      - "*html"
  patterns:
  - pattern-inside: <input ... />
  - pattern-not-inside: $escape.escape(...)
  - pattern-not: $escape...
  - pattern: $$OBJ
  fix: $escape.escape($$OBJ)
"
```
**4** - Run an autofix scan with the following rule:
semgrep --autofix --config rule.yaml example.html

**5** - Run "find example.html | xargs file"  again and observe the "with CRLF line terminators" has been dropped, indicating the file has been converted to LF line separators

**Expected behavior**
CRLF Line Separators are retained during autofixes
  
**Priority**
How important is this to you?
- [✓] P0: blocking me from making progress

**Desktop**
 - OS: macOS
 - Browser: N/A
 - Version: 12.6

dmgping:
html and yaml files couldn't be attached so I tried to paste the bodies but seems the html is being filtered

dmgping:
![Screen Shot 2022-09-22 at 12 44 22 PM](https://user-images.githubusercontent.com/99043843/191837143-ec29bcfc-47a1-46a5-8672-7b378180132d.png)
![Screen Shot 2022-09-22 at 12 43 36 PM](https://user-images.githubusercontent.com/99043843/191837146-2f3ddbd7-de64-45d8-89c2-ccc0cd6088bb.png)

r2c-demo:
[...]

PR #8691, opened 2023-09-12
Issue #6163 Fix - Using \n as the split character consistenly to retain CRLF \r\n line separators

dmgping:
PR checklist:

### DIFF
--- a/cli/tests/e2e/rules/autofix/CRLF-line-endings.yaml
+++ b/cli/tests/e2e/rules/autofix/CRLF-line-endings.yaml
@@ -1,0 +1,16 @@
+rules:
+  - id: java.lang.security.audit.crypto.ssl.defaulthttpclient-is-deprecated.defaulthttpclient-is-deprecated
+    metadata:
+      cwe: "CWE-326: Inadequate Encryption Strength"
+      owasp: "A3: Sensitive Data Exposure"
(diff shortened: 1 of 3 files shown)


cli/tests/e2e/rules/autofix/CRLF-line-endings.yaml
```diff
@@ -0,0 +1,16 @@
...
+    message: |
+      DefaultHttpClient is deprecated. Further, it does not support connections
+      using TLS1.2, which makes using DefaultHttpClient a security hazard.
+      Use SystemDefaultHttpClient instead, which supports TLS1.2.
+    severity: WARNING
+    languages: [java]
+    pattern: new DefaultHttpClient(...);
+    fix-regex:
+      regex: 'DefaultHttpClient\('
+      replacement: "SystemDefaultHttpClient("
```
nmote, line 16: Could you use a small contrived test case instead? Particularly with the test example below, I'd rather not pull in somebody else's copyrighted code even if it is just a simple example.

nmote (commented):
Looks like the tests are failing -- I haven't touched the e2e tests in a while but I think you'll at least need to add a fixed version of the source file so that it knows what to expect after autofix runs. Take a look at the other autofix examples

PR #8694, opened 2023-09-13
fix(swift): implemented key path expression parsing in Swift.

ihji:
PR checklist:

closes PA-3070, closes PA-3087

### DIFF
--- a/changelog.d/pa-3070.fixed
+++ b/changelog.d/pa-3070.fixed
@@ -4,3 +4,5 @@ now be correctly matched by the \$X.isActive pattern:
 ```
 employee.filter(\.isActive)
 ```
+Note that when the implicit type is used, the metavariable $X will bind to the
+backslash character instead of the type name.
--- a/changelog.d/pa-3070.fixed
+++ b/changelog.d/pa-3070.fixed
@@ -1,0 +1,6 @@
+Implemented key path expression parsing in Swift. The following example should
+now be correctly matched by the \$X.isActive pattern:
+
+```
+employee.filter(\.isActive)
+```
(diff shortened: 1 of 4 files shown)


changelog.d/pa-3070.fixed
```diff
@@ -0,0 +1,6 @@
+Implemented key path expression parsing in Swift. The following example should
+now be correctly matched by the \$X.isActive pattern:
```
aryx, line 2: and $X will bind to which content in this example?

languages/swift/generic/Parse_swift_tree_sitter.ml
```diff
@@ -2448,32 +2448,36 @@ and map_primary_expression (env : env) (x : CST.primary_expression) : G.expr =
...
+              (match x with
+              | `Simple_user_type x ->
+                  let id, targs = map_simple_user_type env x in
+                  let name = H2.name_of_id id in
+                  let name = H2.add_type_args_opt_to_name name targs in
+                  G.TyN name |> G.t
+              | `Array_type x -> map_array_type env x
+              | `Dict_type x -> map_dictionary_type env x)
+        | None ->
+            G.T (G.OtherType (("Implicit", Tok.unsafe_fake_tok ""), []) |> G.t)
```
aryx: This is not good, because $X will not have any real token in it, which will make fail certain things.
Maybe you should reuse v1 (the token for \\) at least.
Try running this rule in semgrep instead of semgrep-core, and use --json; I think you would get an error
because semgrep will not be able to generate a content for $X

changelog.d/pa-3070.fixed
ihji: Done.

languages/swift/generic/Parse_swift_tree_sitter.ml
ihji: Thanks for pointing this out. Replaced the unsafe fake token with `v1`.

aryx (approved):

PR #8696, opened 2023-09-13
ci: use PR checklist as comment instead of template

amchiclet:
Added PR checklist to workflow instead of using a PR template. This way the checklist will be added as a comment rather than being in the PR message of every commit (which will clutter the release).

Tested in personal repository.

This PR itself should trigger the new workflow right away, but it seems that the PR template will kick in for the next PR.


amchiclet:
The PR comment shows that at least the commenting part is working.

For the PR template to not include the checklist, I'll have to wait and see the next PR after this is merged to see if it's working or not, but I tried it on my personal repo and it seemed to work that way.

amchiclet:
BTW, I tried removing pull_request_template.md, but it seems to break what goes into the first PR comment and title. When I used the blank file, things seem to be working as expected, so I'm using the blank file in this PR.

.github/workflows/pull-request.yaml
```diff
@@ -0,0 +1,30 @@
+name: PR
```
aryx: maybe rename this workflow pr-checklist.yaml
Most of the workflows in this directory are about doing stuff on pull-requests, so pull-request.yaml is not a very helpful name.

amchiclet: Done.

aryx:
I'll let @mjambon approve the PR; he is the one who started this PR template.

.github/workflows/pr-checklist.yaml
```diff
@@ -0,0 +1,30 @@
+name: PR-checklist
```
aryx, line 22: Maybe add a comment before like
#alt: we used to have a .github/pull_request_template.md file, but then it causes X and Y
# so we opted for a workflow that does not have problem X and Y.

I usually like to put more context on why we decided to do something, and why the obvious naive alternative is not great.

amchiclet: True. Done.

.github/workflows/pr-checklist.yaml
```diff
@@ -0,0 +1,35 @@
+# alt: We could have used the .github/pull_request_template.md file
+# for this checklist, but that means that the checklist will clutter
```
IagoAbal: A bit more context that you can add (rephrasing as you wish). Initially the commit description was the default for "squash and merge", which is made of all the commit messages that you wrote for that PR. But people writes very bad commit messages, "init", "progress", "more work", etc. So while our PR descriptions were pretty good, our Git log looked pretty bad. Then we tried a GitHub option that lets you use the PR title+description as the commit title+message, and I believe everyone appreciated it. However we had the checklist adding noise to every commit message, to remove it you had to do it manually at the time of merging. This PR will fix that, so we can have a nice Git history.

amchiclet: Done.

mjambon (approved):

.github/workflows/pr-checklist.yaml
```diff
@@ -0,0 +1,51 @@
+# Context:
```
aryx, line 1: great comment!

.github/workflows/pr-checklist.yaml
```diff
@@ -0,0 +1,51 @@
...
+#
+# We decided to create a workflow to add the checklist as a comment
+# on the PR instead, so developers will have a checklist, and
+# the PR descriptions won't be cluttered.
+
+name: PR-checklist
+
+on:
+  pull_request:
+    types: [opened, edited]
```
aryx, line 26: no sure we want to repeat that for each edit; I would vote to add the checklist just when the PR is created.

.github/workflows/pr-checklist.yaml
```diff
@@ -0,0 +1,51 @@
...
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  add-checklist-comment:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: pr-checklist
+          only_create: true
```
aryx, line 38: Then I'm confused by those 2 settings; the opened/edited above and this only_create here.
Could you add comments explaining why we need those 2?

PR #8697, opened 2023-09-13
fix(pro): change type signature for semgrep_core_proprietary hook

ihji:
Modify the data type of `diff_targets` from `Fpath.t list` to `Fpath.t list option` to accommodate three distinct scenarios: diff mode with non-empty diff targets, diff mode with empty diff targets, and whole scan mode.

PR checklist:

### DIFF
--- a/src/osemgrep/cli_scan/Scan_subcommand.ml
+++ b/src/osemgrep/cli_scan/Scan_subcommand.ml
@@ -33,13 +33,9 @@ module SS = Set.Make (String)
 (* To run the pro engine, including multistep rules *)
 (*****************************************************************************)
 
-type diff_config = { diff_targets : Fpath.t list; diff_depth : int option }
-
-let default_diff_config = { diff_targets = []; diff_depth = None }
-
 let (invoke_semgrep_core_proprietary :
       (Fpath.t list ->
-      ?diff_config:diff_config ->
+      ?diff_config:Differential_scan_config.t ->
       Engine_type.t ->
       Core_runner.semgrep_core_runner)
       option
@@ -163,7 +159,7 @@ let rules_and_counted_matches (res : Core_runner.result) : (Rule.t * int) list =
 (* Select and execute the Semgrep core engine based on the configured
    engine settings. *)
 let core (conf : Scan_CLI.conf) file_match_results_hook errors targets
-    ?(diff_config = default_diff_config) rules () =
+    ?(diff_config = Differential_scan_config.WholeScan) rules () =
   let invoke_semgrep_core =
     match conf.engine_type with
     | OSS ->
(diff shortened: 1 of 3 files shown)


src/osemgrep/cli_scan/Scan_subcommand.ml
```diff
@@ -33,13 +33,14 @@ module SS = Set.Make (String)
 (* To run the pro engine, including multistep rules *)
 (*****************************************************************************)
 
-type diff_config = { diff_targets : Fpath.t list; diff_depth : int option }
-
-let default_diff_config = { diff_targets = []; diff_depth = None }
+type diff_config = {
```
aryx: maybe add a comment like (* see comment in .mli for explanations *)

aryx: Also it is maybe better to move this type in a separate file; it's becoming an important notion, so maybe it should be in 
a src/core/Differential_scan_config.ml

src/osemgrep/cli_scan/Scan_subcommand.mli
```diff
@@ -26,16 +26,17 @@ val run_scan_conf : Scan_CLI.conf -> Exit_code.t
    `diff_targets` are automatically calculated based on the output
    from the `git diff` command between the baseline commit and the
    head commit. *)
-type diff_config = { diff_targets : Fpath.t list; diff_depth : int option }
-
-val default_diff_config : diff_config
+type diff_config = {
+  diff_targets : Fpath.t list option;
```
aryx: Can you move the comment in the PR about the different situation in the code here too?
Can you explain the different scenario and what are the values for each field in those different situation?
Also it looks like for --baseline-commit, the depth is always 0 and diff_targets represent the whole set of files to analyze whereas in pro depth will be different than 0 (which value actually? where this is set?) and maybe diff_targets are starting point to find the final set of targets? In that case maybe it would be better to introduce a better single type instead of those 2 fields like | BaseLine of Fpath.t list | Depth of Fpath.t list (* starting points *) * int (* depth *) | ...

aryx (changes requested):

src/osemgrep/cli_scan/Scan_subcommand.ml
ihji: Moved to the separate file.

src/osemgrep/cli_scan/Scan_subcommand.mli
ihji: Thank you for the helpful suggestion. I've updated the type of the diff configuration, and it now conveys clearer semantics. It looks like the code has much better readability.

aryx (approved):
I'm still a bit confused between WholeScan and Baseline, because
baseline for me means running on all the files to get a base line.

PR #8695, opened 2023-09-13
Allow customizing the install location of the pro binary

cgdolan:
The pro binary is currently installed in the same directory as the core binary, /usr/local/bin. This is problematic for non root Docker containers since the non root `semgrep` user that performs the install at runtime does not have permissions to write to this directory. Allow the pro binary to be installed somewhere else the non root user does have permissions to (their home directory) by setting an environment variable. Additionally, set this environment variable in non root containers.

PR checklist:

- [ ] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [ ] Tests included or PR comment includes a reproducible test plan
- [ ] Documentation is up-to-date
- [ ] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [ ] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #8705, opened 2023-09-13
Allow customizing the install location of the pro binary

sinat101:
The pro binary is currently installed in the same directory as the core binary, `/usr/local/bin`. This is problematic for non-root Docker containers since the non-root `semgrep` user that performs the install at runtime does not have permissions to write to this directory. Allow the pro binary to be installed somewhere else the non-root user does have permissions to (their home directory) by setting an environment variable. Additionally, set this environment variable in non-root containers.

Credit goes to @cgdolan for the original PR:
https://github.com/returntocorp/semgrep/pull/8695

PR checklist:

### DIFF
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -49,9 +49,7 @@ def determine_semgrep_pro_path() -> Path:
         if os.access(semgrep_pro_path.parent, os.W_OK):
             return semgrep_pro_path
         else:
-            logger.info(
-                "Invalid path, designated SEMGREP_PRO_PATH is not writeable."
-            )
+            logger.info("Invalid path, designated SEMGREP_PRO_PATH is not writeable.")
             sys.exit(FATAL_EXIT_CODE)
 
     core_path = SemgrepCore.path()
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -27,8 +27,14 @@ from semgrep.verbose_logging import getLogger
 
 logger = getLogger(__name__)
 
+SEMGREP_PRO_PATH = os.getenv("SEMGREP_PRO_PATH")
 
 def determine_semgrep_pro_path() -> Path:
+    if SEMGREP_PRO_PATH:
+	    path = Path(SEMGREP_PRO_PATH) / "semgrep-core-proprietary"
+        path.parent.mkdir(parents=True, exist_ok=True)
+        return path
+        
     core_path = SemgrepCore.path()
     if core_path is None:
         logger.info(
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -31,7 +31,7 @@ SEMGREP_PRO_PATH = os.getenv("SEMGREP_PRO_PATH")
 
 def determine_semgrep_pro_path() -> Path:
     if SEMGREP_PRO_PATH:
-	path = Path(SEMGREP_PRO_PATH) / "semgrep-core-proprietary"
+        path = Path(SEMGREP_PRO_PATH) / "semgrep-core-proprietary"
         path.parent.mkdir(parents=True, exist_ok=True)
         return path
         
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -152,7 +152,7 @@ def run_install_semgrep_pro() -> None:
     if semgrep_pro_path.exists():
         semgrep_pro_path.unlink()
     semgrep_pro_path_tmp.rename(semgrep_pro_path)
-    logger.info(f"Successfully installed Semgrep Pro Engine (version {version})!")
+    logger.info(f"\nSuccessfully installed Semgrep Pro Engine (version {version})!")
 
 
 @click.command()
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -40,7 +40,9 @@ def determine_semgrep_pro_path() -> Path:
             semgrep_pro_path.parent.mkdir(parents=True, exist_ok=True)
             return semgrep_pro_path
         else:
-            logger.info("Invalid path, please set a valid path with proper permissions for SEMGREP_PRO_PATH")
+            logger.info(
+                "Invalid path, please set a valid path with proper permissions for SEMGREP_PRO_PATH"
+            )
             sys.exit(FATAL_EXIT_CODE)
 
     core_path = SemgrepCore.path()
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -27,10 +27,13 @@ from semgrep.verbose_logging import getLogger
 
 logger = getLogger(__name__)
 
-SEMGREP_PRO_PATH = os.getenv("SEMGREP_PRO_PATH")
-
 
 def determine_semgrep_pro_path() -> Path:
+    # The SEMGREP_PRO_PATH env var is set for the non-root Docker image to install
+    # the Pro Engine binary in the semgrep user's $HOME/bin directory where it has
+    # the proper permissions to do so.
+    SEMGREP_PRO_PATH = os.getenv("SEMGREP_PRO_PATH")
+    
     if SEMGREP_PRO_PATH:
         path = Path(SEMGREP_PRO_PATH) / "semgrep-core-proprietary"
         path.parent.mkdir(parents=True, exist_ok=True)
--- a/cli/src/semgrep/commands/install.py
+++ b/cli/src/semgrep/commands/install.py
@@ -35,16 +35,20 @@ def determine_semgrep_pro_path() -> Path:
     SEMGREP_PRO_PATH = os.getenv("SEMGREP_PRO_PATH")
 
     if SEMGREP_PRO_PATH:
-        path = Path(SEMGREP_PRO_PATH) / "semgrep-core-proprietary"
-        path.parent.mkdir(parents=True, exist_ok=True)
-        return path
+        semgrep_pro_path = Path(SEMGREP_PRO_PATH) / "semgrep-core-proprietary"
+        if os.access(semgrep_pro_path.parent, os.W_OK):
+            semgrep_pro_path.parent.mkdir(parents=True, exist_ok=True)
+            return semgrep_pro_path
+        else:
+            logger.info("Invalid path, please set a valid path with proper permissions for SEMGREP_PRO_PATH")
+            sys.exit(FATAL_EXIT_CODE)
 
     core_path = SemgrepCore.path()
     if core_path is None:
         logger.info(
             "Could not find `semgrep-core` executable so not sure where to install DeepSemgrep"
         )
-        logger.info("There is something wrong with your semgrep installtation")
+        logger.info("There is something wrong with your semgrep installation")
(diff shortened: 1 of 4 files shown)


sinat101:
I just tested locally and was successful:
```bash
[Sep 13 2023 06:26:59 PM] : sina @ ~/Code/mygithubpage : docker run --rm -v "${PWD}:/src" \
-e SEMGREP_APP_TOKEN="***" \
semgrep \
/bin/bash -c "whoami; semgrep ci --pro"                             
semgrep
Failed to find semgrep-core-proprietary in PATH or in the semgrep package.
You may need to run `semgrep install-semgrep-pro`
Since `semgrep ci` was run, defaulting to legacy semgrep
                  
                  
┌────────────────┐
│ Debugging Info │
└────────────────┘
                  
  SCAN ENVIRONMENT
  versions    - semgrep 1.39.0 on python 3.11.4                        
            
  CONNECTION
  Reporting start of scan for sinat101 (scan_id=12694346)  
  Fetching configuration from Semgrep Cloud Platform       
  Enabled products: Semgrep Code, Semgrep Supply Chain             
        
  ENGINE
Semgrep Pro Engine will be installed in /home/semgrep/bin/semgrep-core-proprietary
Downloading... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 160.5/160.5 MB 8.8 MB/s 0:00:00
Successfully installed Semgrep Pro Engine (version b351eb2)!
Failed to collect contributions. Continuing with scan...
               
               
┌─────────────┐
│ Scan Status │
└─────────────┘
  Scanning 3 files tracked by git with 1677 Code rules, 581 Pro rules:
                                                                                
  <multilang>      61       9          Community    1096                        
  python          349       2          Pro rules     581                        
                                                                                
Semgrep Pro Engine may be slower and show different results than Semgrep OSS.
  2233 rules ran in a within-a-file fashion because `interfile: true` was not   
  specified.                                                                    
  (Use --verbose to see which ones.)

                
                
┌──────────────┐
│ Scan Summary │
└──────────────┘

  Found 0 findings (0 blocking) from 1677 rules.
  Uploading scan results  
  Finalizing scan           View results in Semgrep Cloud Platform:
    https://semgrep.dev/orgs/sinat101/findings
    https://semgrep.dev/orgs/sinat101/supply-chain
```

sinat101:
I don't understand why the circleCI and lint tests are failing. It looks to be the same failure due to https://github.com/returntocorp/semgrep/pull/8701 but I've merged those changes into this branch.

cli/src/semgrep/commands/install.py
```diff
@@ -27,8 +27,15 @@
 
 logger = getLogger(__name__)
 
+SEMGREP_PRO_PATH = os.getenv("SEMGREP_PRO_PATH")
+
 
 def determine_semgrep_pro_path() -> Path:
+    if SEMGREP_PRO_PATH:
```
emjin, line 37: Unfortunately we have some duplicated code. Could you also update `find_semgrep_core_path` in `semgrep/cli/bin`? Or at least leave a note

cgdolan: Do we need to modify that function? That function looks for binaries in PATH and we've made sure to put the location we're storing the pro binary (/home/semgrep/bin) into PATH in the Dockerfile.

cgdolan: To rephrase: If a user sets the `SEMGREP_PRO_PATH` env var to a location that's not on `PATH`, runs the install to put the pro binary in that directory, and then runs the cli, it seems like `find_semgrep_core_path` will correctly raise a `CoreNotFound` error and explain it couldn't find the binary in `PATH`. This behavior seems consistent with the user experience in other cases like installing semgrep OSS with Homebrew if the brew directory isn't in `PATH`, so it's not clear to me what modifications we need to make to `find_semgrep_core_path`.

sinat101: @emjin at first i also thought this file/code might have to be modified as well, but in looking at it, i agree with @cgdolan , all `find_semgrep_core_path` does is locate the path of the binary, not set where to install it.

sinat101:
I figured out why the lint/pre-commit test and CircleCI test were failing. They were failing when running the `semgrep.jsonnet` rules on the PR files, specifically `cli/src/semgrep/commands/install.py`. It was hard to figure out because the check/CI job doesn't say the name of the rule that caused it to fail! I had to finally run it locally to figure out what rule it was and how to fix it. 

See our CI check run vs. my local run:
Screenshot 2023-09-14 at 2 02 07 AM
Screenshot 2023-09-14 at 2 08 30 AM

sinat101:
Testing the image once more, with the latest changes, this time from the docker image for the PR on docker hub:
```bash
[Sep 14 2023 02:37:22 AM] : sina @ ~/Code/mygithubpage : docker run --rm -v "${PWD}:/src" \
-e SEMGREP_APP_TOKEN="***" \
returntocorp/semgrep:pr-8705-nonroot \
/bin/bash -c "whoami; semgrep ci --pro"
semgrep
Failed to find semgrep-core-proprietary in PATH or in the semgrep package.
You may need to run `semgrep install-semgrep-pro`
Since `semgrep ci` was run, defaulting to legacy semgrep
                  
                  
┌────────────────┐
│ Debugging Info │
└────────────────┘
                  
  SCAN ENVIRONMENT
  versions    - semgrep 1.39.0 on python 3.11.4                        
            
  CONNECTION
  Reporting start of scan for sinat101 (scan_id=12707731)  
  Fetching configuration from Semgrep Cloud Platform       
  Enabled products: Semgrep Code, Semgrep Supply Chain             
        
  ENGINE
Semgrep Pro Engine will be installed in /home/semgrep/bin/semgrep-core-proprietary
Downloading... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 160.5/160.5 MB 13.6 MB/s 0:00:00
Successfully installed Semgrep Pro Engine (version b351eb2)!
Failed to collect contributions. Continuing with scan...
               
               
┌─────────────┐
│ Scan Status │
└─────────────┘
  Scanning 3 files tracked by git with 1677 Code rules, 581 Pro rules:
                                                                                
  <multilang>      61       9          Community    1096                        
  python          349       2          Pro rules     581                        
                                                                                
Semgrep Pro Engine may be slower and show different results than Semgrep OSS.
  2235 rules ran in a within-a-file fashion because `interfile: true` was not   
  specified.                                                                    
  (Use --verbose to see which ones.)

                
                
┌──────────────┐
│ Scan Summary │
└──────────────┘

  Found 0 findings (0 blocking) from 1677 rules.
  Uploading scan results  
  Finalizing scan           View results in Semgrep Cloud Platform:
    https://semgrep.dev/orgs/sinat101/findings
    https://semgrep.dev/orgs/sinat101/supply-chain
  ```

File: cli/src/semgrep/commands/install.py (hunk 1 above)
Comment (emjin) [reply cli/src/semgrep/commands/install.py:37]: Ah, I see, I missed that `PATH` was set! In that case, this is fine without. I would still prefer a comment, because I think it would be confusing for a local `semgrep scan` user who had set `SEMGREP_PRO_PATH` but was getting an error that semgrep-pro isn't found, but it's not urgent to implement

File: cli/src/semgrep/commands/install.py [hunk 2]
```diff
@@ -29,6 +29,16 @@
...
 def determine_semgrep_pro_path() -> Path:
+    # The SEMGREP_PRO_PATH env var is set for the non-root Docker image to install
+    # the Pro Engine binary in the semgrep user's $HOME/bin directory where it has
+    # the proper permissions to do so.
+    SEMGREP_PRO_PATH = os.getenv("SEMGREP_PRO_PATH")
+
+    if SEMGREP_PRO_PATH:
+        path = Path(SEMGREP_PRO_PATH) / "semgrep-core-proprietary"
+        path.parent.mkdir(parents=True, exist_ok=True)
+        return path
```
emjin: I would love to see test output for what happens if the user's `SEMGREP_PRO_PATH` is invalid (hoping for a clear error message)

emjin (approved):
This looks reasonable to me but I can't speak to the Docker side of changes. I'm also wondering how much work it would be to add an automated test

cli/src/semgrep/commands/install.py
sinat101: Hey @emjin, if the path is invalid the docker image crashes much like it already does when trying to install the pro binary as a non-root user. I did just add some validation however to exit more gracefully if the path is invalid.

cli/src/semgrep/commands/install.py
sinat101: @emjin sorry I don't exactly follow what you mean 😅 . would it be clearer if we changed `SEMGREP_PRO_PATH` to `SEMGREP_PRO_INSTALL_PATH`?

emjin: What I'm concerned about is:
1. A user doesn't have semgrep pro
2. They set SEMGREP_PRO_PATH and run `semgrep ci --pro`
3. `semgrep ci --pro` installs semgrep pro for the user
4. They later run `semgrep scan --pro`, which cannot find semgrep pro. The user is confused

It's not a super common scenario, so I'm not that worried, but it seems like we could avoid this by also looking for the pro binary in SEMGREP_PRO_PATH

Let me know if I've misunderstood something!

sinat101: I totally get what you mean now, sorry it took me a minute to realize the scenario. I added logic for `semgrep scan --pro` to look for the pro binary in SEMGREP_PRO_PATH if it's set. I also added some more error handling in general and code comments.

aryx:
What about we simply move semgrep-core when in a non-root container to /home/semgrep/bin/ already
and add this to the PATH, so then semgrep install will copy it to the same place of semgrep-core, without
additional logic in /home/semgrep/bin?

aryx (changes requested):
If the core problem is that semgrep-core is in a protected place and semgrep-core-pro can't be there for nonroot container, seems simpler to just move semgrep-core in the first place
in a non-protected place, to get better symetry and remove special logic.

aryx:
We could just alter the Dockerfile and when building the `FROM semgrep-cli AS nonroot` do more stuff that just changing the USER. we can move around semgrep-core to /home/semgrep/bin and modify the PATH.
Hopefully this will solve for free the installation of semgrep-core-proprietary without needing yet another environment variable (we have too many SEMGREP_XXX env var already ... let's not add one).

sinat101:
Hey @aryx that was my initial thought actually, here is my first PR I made before this:
https://github.com/returntocorp/semgrep/pull/8685

However Tom and Chris mentioned they wanted to give users customizability and use an environment variable, so I went down that route and made a new PR based on Chris' PR.

If you think the original idea of placing semgrep-core in a non-root directory is better, and we get rid of the env var, I'm happy with that. Just would have been good to discuss this earlier before all this work I guess 😅 .

Also, I did add some error handling that would be helpful in general, I can keep those changes if we decide on a new PR.

aryx:
Sometimes you need to code the alternate solution to realize the first one was actually better.
Also people have different priorities. Personally I really really don't want cli/bin/semgrep to grow, because it's a critical piece and the more we add code there, the more complex it become, so if we can move the complexity outside, like improving the Dockerfile then it's far better for me.

aryx:
Actually closing in favor of the orginal (and far simpler) https://github.com/returntocorp/semgrep/pull/8685

PR #8686, opened 2023-09-12
Simplify masking of the variable part of test results

mjambon:
* Some of the temporary path masking was done on specific JSON fields. I removed some of it to simplify things. This will make it easier to port this feature to OCaml. Some JSON-specific masking remains. We might want to get rid of it as well eventually.
* I also added a special case for masking paths that are temporary in pysemgrep but constant in osemgrep so as to pass tests (rule files being copied in one case but not the other). This technique could be used to mask other ignorable differences between pysemgrep and osemgrep.

This allowed me to enable two osempass tests.

PR checklist:

- [x] Purpose of the code is [evident to future readers](https://semgrep.dev/docs/contributing/contributing-code/#explaining-code)
- [x] Tests included or PR comment includes a reproducible test plan
- [x] Documentation is up-to-date
- [x] A changelog entry was [added to changelog.d](https://semgrep.dev/docs/contributing/contributing-code/#adding-a-changelog-entry) for any user-facing change
- [x] Change has no security implications (otherwise, ping security team)

If you're unsure about any of this, please see:

- [Contribution guidelines](https://semgrep.dev/docs/contributing/contributing-code)!
- [One of the more specific guides located here](https://semgrep.dev/docs/contributing/contributing/)

PR #8707, opened 2023-09-13
Rewrite rule IDs at rule-parsing time rather than later

mjambon:
See the problem here: https://github.com/returntocorp/semgrep/pull/8686#issuecomment-1716299070

These changes were needed to produce correct rule IDs of the form `path.to.rulename` instead of `rulename` when parsing a rule with `osemgrep scan` and producing an error message at this stage. This will also help in other contexts where we parse rules and want this ID rewriting to take place.

Lessons:
* It was a lot of work for what's essentially moving a function call from one file to another. It shows the value of designing things right upfront (we always try and sometimes we fail, that's ok).
* The use of the coercion operator to turn a rule ID into a string is cute (`(rule_id :> string)`) but was getting in the way of debugging: first, I wanted to trace all the places where we convert a rule ID to a string and this wasn't possible because `:>` is not a function and can't be traced by adding logging instructions. Second, `:> string` could be used for other things than rule IDs. It is less specific than a call to `Rule_id.to_string` which is easier to locate in source code.

PR checklist:


src/core/Rule_ID.mli
```diff
@@ -3,7 +3,7 @@
    and error messages. The coercion operator can be used as an alternative
```
aryx, line 3: Is this comment still valid now that we use an abstract type?

src/core_cli/Core_actions.ml
```diff
@@ -148,11 +148,11 @@ let dump_equivalences file =
 
 let dump_rule file =
   let file = Run_semgrep.replace_named_pipe_by_regular_file file in
-  let rules = Parse_rule.parse file in
+  let rules = Parse_rule.parse ~rewrite_rule_ids:None file in
   rules |> List.iter (fun r -> pr (Rule.show r))
 
 let prefilter_of_rules file =
-  let rules = Parse_rule.parse file in
+  let rules = Parse_rule.parse ~rewrite_rule_ids:None file in
```
aryx, line 155: wouldn't it be better to make rewrite_rule_ids a default parameter like
?(rewrite_rule_ids=None), so you don't need to modify all those call sites?

src/parsing/Parse_rule.mli
```diff
@@ -1,15 +1,33 @@
+(*
+   Rule parsing
+*)
+
+(*
+   For rules that come from a local file, this is the path to the file
+   containing the rule. It's used to create a dotted rule identifier
+   such as 'stuff.rules.foo' for a rule declared as 'foo' in the file
+   'stuff/rules.yml'.
```
aryx, line 9: maybe add a comment like
"This may seem redundant with the second argument of parse_and_filter_invalid_rules() below, but sometimes
this second argument corresponds to a temporary file generated from content fetched from the network (e.g., rules coming from the registry locally stored in /tmp/) in which case we do not want to add this prefix, hence the use of Other_origin in that case, even if we're still parsing a local (temporary) file."

src/osemgrep/cli_scan/Validate_subcommand.ml
```diff
@@ -87,16 +87,20 @@ let run (conf : conf) : Exit_code.t =
...
+                     (* TODO: stricter: warn if no origin (meaning URL or registry) *)
+                     None)
         in
         let in_docker = !Semgrep_envvars.v.in_docker in
         let (config : Rules_config.t) =
           Rules_config.parse_config_string ~in_docker metarules_pack
         in
         let metarules_and_origin =
           Rule_fetching.rules_from_dashdash_config ~token_opt
+            ~rewrite_rule_ids:true (* default *)
```
aryx, line 103: it's a bit confusing that sometimes ~rewrite_rule_ids is a boolean and sometimes it's this Parse_rule.origin type.

src/osemgrep/networking/Rule_fetching.ml
```diff
@@ -32,16 +32,11 @@ module XP = Xpattern
  * a pair, which would remove the need for partition_rules_and_errors.
  *)
 type rules_and_origin = {
-  origin : origin;
   rules : Rule.rules;
   errors : Rule.invalid_rule_error list;
+  origin : Parse_rule.origin; (* used by Validate_subcommand *)
```
aryx, line 37: not sure it was good to move this type to Parse_rule.ml.
Maybe we should just pass a callback ?rewrite_rule_ids_hook:(Rule_id.t -> Rule_id.t) option
in Parse_rule.ml, that we set here in Rule_fetching.ml differently
when we were passed the --rewrite-rule-ids CLI flag.

aryx (approved):
I think the use of the same label with different type is a bit confusing
and I'm not sure it was worth moving the code to Parse_rule.ml.
We could have achieved the same by passing a rewrite_rule_id_hook callback.
set to None by default which would avoid to modify all the other call sites
that don't need to bother about this rewrite_rule_id thing.
I'm approving, but it feels a bit ugly.

aryx:
@mjambon can you rebase and fix the conflict so we can merge. Seems like a good thing to get in.

aryx:
I had a few comments on it, I still think a rewrite_rule_id_hook callback would have been simpler than moving code there.

mjambon:
I haven't had time to revisit today other than fixing the merge conflicts.

> I had a few comments on it, I still think a rewrite_rule_id_hook callback would have been simpler than moving code there.

I'll look into this (possibly after merging).

aryx (approved):
I'll address my own comment in a separate PR I make.
Let's merge this.

PR #8713, opened 2023-09-14
chore: [SCP-965] Limit contributions commits to 30 days

zyannes:
The new `dump_contributions` command collects the contribution from ALL commits in the history.

- Limit the contributions to the commits created in the last 30 days

## Testing
Tested on a few high-traffic public repos (e.g. cpython). 

PR checklist:

### DIFF
--- a/libs/git_wrapper/Git_wrapper.ml
+++ b/libs/git_wrapper/Git_wrapper.ml
@@ -349,19 +349,20 @@ let git_log_json_format =
    \"contributor\": {\"commit_author_name\": \"%an\", \"commit_author_email\": \
    \"%ae\"}}"
 
-let get_month_ago_date () =
-  let yesterday = Common2.yesterday () in
-  let month_ago = Common2.month_before yesterday in
-  let after_date = Unix.gmtime month_ago in
-  let year = after_date.tm_year + 1900 in
-  let month = after_date.tm_mon + 1 in
-  let day = after_date.tm_mday in
+let date_to_year_str (timestamp : Common2.float_time) : string =
+  let date = Unix.gmtime timestamp in
+  let year = date.tm_year + 1900 in
+  let month = date.tm_mon + 1 in
+  let day = date.tm_mday in
   Printf.sprintf "%04d-%02d-%02d" year month day
 
-let get_git_logs () : string list =
-  (* TODO Allow this to be configurable past 30 days? *)
-  let month_ago_date = get_month_ago_date () in
-  let after_arg = Printf.sprintf "--after=\"%s\"" month_ago_date in
+let get_git_logs ?(since = None) () : string list =
+  let since_year =
+    match since with
+    | None -> date_to_year_str (Common2.month_before (Common2.yesterday ()))
+    | Some since_timestamp -> date_to_year_str since_timestamp
+  in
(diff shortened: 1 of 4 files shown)


libs/git_wrapper/Git_wrapper.ml
```diff
@@ -349,8 +349,20 @@ let git_log_json_format =
...
+  let month_ago = Common2.month_before yesterday in
+  let after_date = Unix.gmtime month_ago in
+  let year = after_date.tm_year + 1900 in
+  let month = after_date.tm_mon + 1 in
+  let day = after_date.tm_mday in
+  Printf.sprintf "%04d-%02d-%02d" year month day
+
 let get_git_logs () : string list =
-  let cmd = Bos.Cmd.(v "git" % "log" % git_log_json_format) in
+  (* TODO Allow this to be configurable past 30 days? *)
```
aryx: at this point would be better to make it an argument of get_git_logs
like get_git_logs ~since:(Common2.month_before (Common2.yesterday ()) 
so you specify this at the call site.
and maybe have an helper function here like  time_to_string_for_git_after (date : Common2.float_time) : string

aryx (changes requested):

libs/git_wrapper/Git_wrapper.ml
zyannes: Updated to provide an optional `since` param that defaults to last month timestamp if None.

libs/git_wrapper/Git_wrapper.mli
```diff
@@ -67,7 +67,7 @@ val get_project_url : unit -> string option
     [git ls-remote] or from the [.git/config] file. It returns [None] if it
     found nothing relevant. *)
 
-val get_git_logs : unit -> string list
+val get_git_logs : ?since:Common2.float_time option -> unit -> string list
```
aryx, line 70: then you need to update the comment saying if no since then it defaults to the last 30 days.

aryx (approved):

libs/git_wrapper/Git_wrapper.mli
aryx: and then I'd rather be explicit at the calling site in the dump_contributions() function and still pass a ~since:"last 30 days", even if it's the default, so it's clear in the semgrep code (this libs/ code is supposed to contain general libraries that can be used outside semgrep) that we use a 30 days (and maybe even add a link to the semgrep.dev/docs/usage-limit to explain why we use 30 days in dump_contributions().

zyannes: Yeah that makes sense. Updating!

zyannes: Do you think we need to propagate the `since` param all the way up to Core_CLI?

aryx: no, you can just add the ~since:"30 days" in the call to get_git_logs in Parse_contribution.ml
(and maybe add a comment about it in Parse_contribution.mli, and the link to semgrep.dev/usage-limit there)

aryx: Parse_contributions.ml is part of semgrep (it's not in libs/, it's in src/), and its objective is to parse the contributions for semgrep, for the app, so it's perfectly fine to hardcode the 30 days there;

PR #8719, opened 2023-09-14
feat: modify origin in `semgrep.dev` metadata for non-semgrep sources

kopecs:
When a rule is not downloaded from semgrep.dev by the CLI, strips out the `semgrep.dev` metadata field if present and the rule also has a postprocessor. This is to ensure we are able to perform a check against the origin of the rule in pro so that in normal execution of `semgrep ci` untrusted postprocessors are not run.

This PR also adds a flag to opt-in to allowing untrusted postprocessors.

Test plan: will be present in pro PR.

PR checklist:

- [x] Change has no security implications (otherwise, ping security team) (@LewisArdern do you want to review?)


akuhlens (approved):
This looks good to me. I can't verify this is enough prevent using rules to exfiltrate data, but it seems like it is better than it was.

kurt-r2c (approved):

LewisArdern (commented):
LGTM

